PR #3423, opened 2018-11-06
Infer input/output buffer constraints from schedule directives

steven-johnson:
This is still a work-in-progress that isn't ready for review, but I'm opening a PR to get wider feedback on it; it's not ready for a proper code review yet, but a conceptual review would be welcome.

The idea here (from @dsharlet) is that we can use the existing scheduling language to infer the right constraints for input and output buffers, avoiding the need to use `dim().set_stride()` and such entirely; this is motivated by several things:

- More compact code, less redundant specification (cf the changes to test/interleave)
- Ability to specify constraints for Generators that use `Input<Func>` or `Output<Func>`; putting schedule-methods shims onto these allows you to conveniently use such a Generator both in 'standalone' mode, or via GeneratorStub (in which case the constraints would simply be ignored for inlined Funcs.)

Theoretically, this would allow us to migrate all code to simply use scheduling directives and ~never need to use the set_stride(), etc methods; in practice, there are some holes that aren't addressable with the existing scheduling directives:
- There's no way to set stride explicitly to a strange value (eg if you have a buffer that is referring to some array of structs or some such)
- There's no way to set min without setting extent (eg if you want min=0 but extent aligned to a 16-byte boundary); align_bounds() aligns both min and extent.
- There's no way to clamp an extent to an upper bound (eg if you want to ensure that a buffer is <= a hardware-specific limit)

Also, the implementation has some ugly bits:
- We emit user warnings if you attempt to constrain the bounds of an inline Func (since this is meaningless and probably an error); with these changes, it becomes reasonable to set constraints on the inputs or outputs of a Generator, which might be ignored in some situations. Rather than try to finesse this, I've just inserted a flag to allow suppressing these warnings for a given `Function`, and then ruthlessly set it for Generator inputs and outputs. This is probably ok, but it has a code smell I don't like.
- There's some handwavy TODO stuff in the Generator code (re: limiting access to certain methods) that needs cleaning up.
- There are a handful of existing tests and apps that have failures that need investigating.

PR #3427, opened 2018-11-08
Add generalized Buffer::transpose() and related ctors

steven-johnson:
It was awkward to use Buffer for non-planar data; `make_interleaved` supports only a limited case. This adds an arbitrary `reorder` method, and related ctors, that allow for easily making Buffers that have storage ordering that doesn't match the dimension ordering. (This will be used by https://github.com/halide/Halide/pull/3423 to make the results of `Func::realize()` have the proper layout automatically.)

Also, some drive-by cleanup in Buffer to use delegating ctors in some spots.


src/runtime/HalideBuffer.h
```diff
@@ -1381,6 +1396,34 @@ class Buffer {
         std::swap(buf.dim[d1], buf.dim[d2]);
     }
 
+    /** A generalization of transpose: instead of swapping two dimensions,
+     * pass a vector that lists each dimension index exactly once, in the desired order.
+     * For instance, to reorder a 3-dimensional planar image to be interleaved,
+     * pass {2, 0, 1} for storage_order */
+    void reorder(const std::vector<int> &order) {
```
abadams: I think it would be better to name this "transpose". In Halide::Func, reorder and reorder_storage both have no effect on how something is indexed, but change the underlying representation. Transpose is the complement. It swizzles indexing without changing the underlying representation.

abadams:
LGTM apart from disagreeing with the name

PR #3431, opened 2018-11-09
Make can_prove use CSE and trim dead lets

abadams:
This makes it slightly more powerful. 

Also report the original unsimplified expression too when a proof fails so we know what original state got the simplifier stuck in a local minimum.

### DIFF
--- a/src/Simplify.cpp
+++ b/src/Simplify.cpp
@@ -1,6 +1,7 @@
 #include "Simplify.h"
 #include "Simplify_Internal.h"
 
+#include "CSE.h"
 #include "IRMutator.h"
 #include "Substitute.h"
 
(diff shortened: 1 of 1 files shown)


src/Simplify.cpp
```diff
@@ -274,8 +275,16 @@ bool can_prove(Expr e, const Scope<Interval> &bounds) {
...
 
+    e = common_subexpression_elimination(e);
+
+    Expr orig = e;
+
     e = simplify(e, true, bounds);
 
+    // When terms cancel, the simplifier doesn't always successfully
+    // kill the dead lets.
+    while (const Let *l = e.as<Let>()) e = l->body;
```
steven-johnson: Loops without { curly braces } make me nervous, even trivial ones like this. Don't we (generally) prefer all loopy structures to be embraced?

abadams:
Fixed. There's no way adding braces could break anything, so merging.

PR #3441, opened 2018-11-16
Add Buffer::as_ptr() and as_ptr_const()

steven-johnson:
These are analogous to as() and as_const(), but return pointers rather than refs; this is useful for coding in C++ styles that require pointers for mutable references, when you have (eg)

    void DoSomethingWith(Buffer<const int> *input);

     Buffer<int> foo;
     DoSomethingWith(&foo);  // won't compile
     DoSomethingWith(foo.as_ptr<const int>());  // OK
     DoSomethingWith(foo.as_ptr_const());  // also OK

Note that I deliberately avoid as_const_ptr() as the name, as that would imply to me the pointer returned is always const, which isn't the case.

Also note that I'm deliberately avoiding rvalue overloads for this, as it's not clear to me that it's an interesting use case.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1118,6 +1118,46 @@ public:
     }
     // @}
 
+    /** Return a typed pointer to this Buffer. Useful for safely converting
+     * a reference to a Buffer<void>* to a reference to, for example, a
+     * Buffer<const uint8_t>*, or converting a Buffer<T>* to Buffer<const T>*.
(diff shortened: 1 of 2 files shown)


abadams:
Is `&foo.as<T>()` not legal?

steven-johnson:
> Is &foo.as<T>() not legal?

It's legal, but Certain Build Environments will aggressively warn about taking the address of a temporary.

steven-johnson:
Bah: so far I'm unable to repro the compiler issue I was having re: the warnings/errors. I'm going to close this for now.

PR #3442, opened 2018-11-17
Refactor Argument, add ArgumentEstimates

steven-johnson:
This is a refactoring of Argument to move the auto-schedule estimates into it, as preparation for adding them to the generated Metadata. (There turned out to be substantial amounts of drive-by changes, but I think they are worthwhile.)

### DIFF
--- a/src/Argument.h
+++ b/src/Argument.h
@@ -14,6 +14,21 @@
 
 namespace Halide {
 
+struct ArgumentEstimates {
+    /** If this is a scalar argument, then these are its default, min, max, and estimated values.
+     * For buffer arguments, all should be undefined. */
+    Expr scalar_def, scalar_min, scalar_max, scalar_estimate;
+
+    /** If this is a buffer argument, these are the estimated min and extent for each dimension.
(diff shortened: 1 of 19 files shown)


src/Argument.h
```diff
@@ -53,27 +68,14 @@ struct Argument {
...
-        user_assert(!(is_buffer() && def.defined()))
-            << "Scalar default must not be defined for Buffer Arguments";
-        user_assert(!(is_buffer() && min.defined()))
-            << "Scalar min must not be defined for Buffer Arguments";
-        user_assert(!(is_buffer() && max.defined()))
-            << "Scalar max must not be defined for Buffer Arguments";
-    }
+             const ArgumentEstimates &argument_estimates);
 
+    // TODO: this should really be marked explicit
```
abadams: ... but then you wouldn't be able to (without a cast) put a Buffer in an argument list, to indicate that it shouldn't be baked into the object file, but instead received as an argument at runtime.

steven-johnson: true. it's just my knee-jerk assumption that non-explicit ctors should be regarded with suspicion. I'll update the comment.

PR #3443, opened 2018-11-17
Add auto-scheduling estimate values to metadata output

steven-johnson:
(Note, additive to #3442)


abadams:
lgtm, but clearly something is segfaulting

PR #3446, opened 2018-11-19
Add runtime wrapper for abort()

steven-johnson:
Windows Debug runtimes stall the `abort()` call by presenting a modal dialog. Add a runtime overload for windows that does the equivalent raise/exit combination as documented by MSFT instead and use it instead of raw abort(). Ditto for the compile-time Error.cpp in the compiler. (Note: this makes no attempt to convert apps or tests that explicitly call abort; this is just for the Halide compiler and runtime itself.)


abadams:
I think we *do* want the modal dialog box... just not on the buildbots. If you're debugging code that uses Halide and you fail an abort, people will expect to be able to drop into the debugger in the usual way.

If there's no global environment setting to turn off these damn dialogs, then I think the solution to the buildbot issue is to use taskkill or something similar to nuke all processes holding on to Halide.dll at the start of each build.

steven-johnson:
(Alternately: we could add our own HL_DISABLE_WINDOWS_ABORT_DIALOG and only define it for buildbots.)

dsharletg:
If we're looking for programmatic solutions to this, there's also _set_abort_behavior (https://msdn.microsoft.com/en-us/library/e631wekh.aspx).

src/Error.cpp
```diff
@@ -1,11 +1,32 @@
...
+
 namespace Halide {
 
 namespace {
 
 CompileTimeErrorReporter* custom_error_reporter = nullptr;
 
+void error_abort() {
+#if defined(_MSC_VER) && defined(_DEBUG)
+    if (!IsDebuggerPresent()) {
```
dsharletg: I don't think this really solves the problem. When I worked on windows a lot, my main use of this dialog was when I *wasn't* running in the debugger, and then hitting retry would start the debugger and let you debug the process after the fact (without starting from the debugger).

steven-johnson: Sigh. Yeap. You're right.

src/Error.cpp
```diff
@@ -1,11 +1,30 @@
...
 
 CompileTimeErrorReporter* custom_error_reporter = nullptr;
 
+void error_abort() {
+#ifdef _MSC_VER
+    const std::string s = Internal::get_env_variable("HL_DISABLE_WINDOWS_ABORT_DIALOG");
+    const int disable !s.empty() ? atoi(s.c_str()) : 0;
+    if (disable) {
+        // Debug variants of the MSVC runtime will present an "Abort, Retry, Ignore"
+        // dialog in response to a call to abort(); we ~never want this unless there
```
abadams: This comment is now stale.

src/runtime/windows_abort.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+
+WEAK __attribute__((always_inline)) void halide_abort() {
+    char *s = getenv("HL_DISABLE_WINDOWS_ABORT_DIALOG");
+    if (s && atoi(s)) {
+        // Debug variants of the MSVC runtime will present an "Abort, Retry, Ignore"
+        // dialog in response to a call to abort(); we ~never want this unless there
```
abadams: Comment is stale. Maybe just refer to src/Error.cpp?

abadams:
Env var solution seems reasonable.

PR #3447, opened 2018-11-19
Expand "name not in scope" debug info

steven-johnson:
Many of the rogue `abort()` calls on Windows seem to be from failed Scope lookup during Generator compilation, which is odd, since (in theory) this code should be common across all platforms. Augment the assertion-failure code to dump the entire Scope chain out upon failure to help debug the situation. (Also: drive-by tweak to allow iterators to compile for `Scope<void>`)

### DIFF
--- a/src/Scope.h
+++ b/src/Scope.h
@@ -125,7 +125,7 @@ public:
             if (containing_scope) {
                 return containing_scope->get(name);
             } else {
-                internal_error << "Name not in Scope: " << name << "\n";
+                internal_error << "Name not in Scope: " << name << "\n" << *this << "\n";
             }
         }
         return iter->second.top();
@@ -137,7 +137,7 @@ public:
     T2 &ref(const std::string &name) {
         typename std::unordered_map<std::string, SmallStack<T>>::iterator iter = table.find(name);
         if (iter == table.end() || iter->second.empty()) {
-            internal_error << "Name not in Scope: " << name << "\n";
+            internal_error << "Name not in Scope: " << name << "\n" << *this << "\n";
         }
         return iter->second.top_ref();
     }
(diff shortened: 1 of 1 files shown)


steven-johnson:
e.g., when building generator_aot_buffer_copy with this patch, I get a failure like:

```
         Internal error at c:\build_bot\worker\win-32-distro-700\halide\src\Scope.h:178
         Condition failed: iter != table.end()
         Name not in Scope: x.min.0
         {
           sum$1.s1.r14$x.max
           result.s0.i.min
           result.s2.i.max
           result.s0.i.max
           sum$1.s1.r14$x.min
           x.stride.0.proposed
           result.s2.i.min
           result.s1.i.max
           x.min.0.proposed
           result.s1.i.min
           sum.s1.r9$x.max
           result.stride.0.constrained
           sum.s1.r9$x.min
           f0.s1.r4$x.max
           f0.s1.r4$x.min
           f0.s1.i.max
           x.extent.0.required
           f0.s1.i.min
           f0.s0.i.max
           f0.s0.i.min
           x.extent.0.proposed
           result.extent.0.required
           result.min.0.required
           result.stride.0.required
           x.min.0.required
           x.stride.0.required
           x.stride.0.constrained
           x.min.0.constrained
           result.stride.0.proposed
           result.min.0.proposed
           result.extent.0.proposed
         }
```

abadams:
Odd. I wonder if it's a stack overflow issue - that could cause arbitrary corruption.

steven-johnson:
Scope uses `unordered_map` -- maybe there's a subtle dependency on an order which isn't guaranteed.

abadams:
That's a terrifying thought. Nothing should be depending on that order.

steven-johnson:
An even more interesting question is why it apparently only fails for Debug builds.

steven-johnson:
Changing to `std::map` doesn't appear to move the needle, so presumably that's not it.

steven-johnson:
Unfortunately I've run out of time to debug this (I'm out the rest of this week), but it seems like something potentially real -- if I build bilateral_grid with auto_schedule=true, the `bilateral_grid.dimensions` symbol defined (by AddImageChecks, I think) doesn't seem to get pushed onto the scope correctly. Perhaps getting simplified away? Don't know how that could be compiler-specific. There's definitely something here that needs investigating.

steven-johnson:
(Maybe Win32 specific? Maybe MSVC-Debug specific? Both might be possible.)

PR #3450, opened 2018-11-21
Type mismatch HalideBuffer.h (int32_t -> int)

aankit-ca:
Type mismatch between expected and actual return type (int32_t -> int)


aankit-ca:
@dsharletg PTAL

Issue #3451, opened 2018-11-23
Shouldn't need to specify input sizes to rungen


abadams:
If I specify an output size, it should just use a bounds query to get input sizes. E.g. I'd like to be able to say:

`./my_generator.rungen --output_extents=[1024,1024] input1=zero input2=zero`

Additionally, it would be nice to set a policy for all inputs not otherwise specified:

`./my_generator.rungen --output_extents=[1024,1024] --input_values=random`

steven-johnson:
offline, andrew also suggested using autoscheduler estimates for output sizes (if present) now that they are emitted in the metadata

steven-johnson:
(Note that there isn't currently a way to use `--output_extents` for Generator with multiple Outputs that may have different dims/sizes. Would probably be useful to have a way to handle this case.)

steven-johnson:
It could be useful to allow for either bounds-query or estimates for the 'auto' option, so how about something like this as an option:
```
input1=zero:[100,100,3]    # explicit
input1=zero:bq             # use bounds-query size. alternately, 'calc' or 'auto' for text
input1=zero:estimates      # use estimates provided in code
```

abadams:
lgtm. Another tedious thing was just having to list every single input. It would be nice to be able to specify a value for all inputs not explicitly mentioned elsewhere:

`--input_extents=random:auto`

steven-johnson:
Should the default-input flag apply to scalars as well as buffers?

abadams:
Maybe via a separate flag? Ideally the person has given their input scalars default values that appear in the metadata.

abadams:
(or estimates)

steven-johnson:
Declaring fixed.

PR #3452, opened 2018-11-27
Avoid need to specify inputs for RunGen in some cases (Issue #3451)

steven-johnson:
This adds the ability to specify `auto` for an inputs extents, in which case we use bounds-query mode to calculate a legal size. It also adds `--default_input_buffers` and `--default_input_scalars` to allow for auto-initializing unspecified inputs to certain defaults.

Also removes the defunct-and-useless --experimental_guess mode.


steven-johnson:
(Note that this doesn't deal with estimates at all, as discussed in the issue; that will come in a subsequent PR if this is approved)

PR #3456, opened 2018-11-28
Add features to disable LLVM Loop Vectorization and Unrolling

steven-johnson:
There's some evidence that this increases compile time without improving code generation for *properly-scheduled* Halide code. Adding feature flags to allow opting out so that downstream code can experimentally disable it.

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1086,8 +1086,9 @@ void CodeGen_LLVM::optimize_module() {
     PassManagerBuilder b;
     b.OptLevel = 3;
     b.Inliner = createFunctionInliningPass(b.OptLevel, 0, false);
-    b.LoopVectorize = true;
-    b.SLPVectorize = true;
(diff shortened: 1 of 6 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -1086,8 +1086,9 @@ void CodeGen_LLVM::optimize_module() {
     PassManagerBuilder b;
     b.OptLevel = 3;
     b.Inliner = createFunctionInliningPass(b.OptLevel, 0, false);
-    b.LoopVectorize = true;
-    b.SLPVectorize = true;
+    b.LoopVectorize = !get_target().has_feature(Target::DisableLLVMLoopVectorize);
+    b.DisableUnrollLoops = get_target().has_feature(Target::DisableLLVMLoopUnroll);
+    b.SLPVectorize = true;  // TODO: should this also be based on DisableLLVMLoopVectorize?
```
abadams: No, SLP vectorization has no analogue in the Halide scheduling model.

PR #3458, opened 2018-11-29
Allocate hexagon runtime arguments buffers on the heap if there are too many arguments

white-pony:
I also included some of the binaries, I wasn't exactly sure which one of them are needed, so please let me know if I missed anything or something is not needed.

Also, I don't know if it matters, but Hexagon SDK I have comes with Android NDK r14b instead of r10d from Makefile, so I built with that one (with local modifications to Makefile).


src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -368,6 +385,11 @@ int halide_hexagon_remote_run_v2(handle_t module_ptr, handle_t function,
     // Power HVX off.
     halide_hexagon_remote_power_hvx_off();
 
+    if (allocated_on_heap) {
+        free(buffers);
```
steven-johnson: The `return` statement on line 379 will skip these frees.

steven-johnson:
LGTM

dsharletg:
Looks good, but libhalide_hexagon_host.so doesn't need to change.

white-pony:
> Looks good, but libhalide_hexagon_host.so doesn't need to change.

Thanks, was not sure about this one, reverted it back.

PR #3459, opened 2018-11-29
Remove support for legacy buffer_t via define_extern

steven-johnson:
This is step one towards (eventually) removing support for LegacyBufferWrappers; it keeps support for generating toplevel entry points that use buffer_t (via the LegacyBufferWrappers flag), but removes support for `define_extern` functions using buffer_t.

(This probably isn't ready to land yet, but the Google codebase is getting close to eradicating all legacy `define_extern()` usage, so I wanted to put this out for discussion, to see if other downstream users are also ready to drop this backwards support.)


abadams:
I would love to see this killed with fire, but I'm not the one with legacy usage.

steven-johnson:
Adding @shoaibkamil to comment on Adobe usage.

shoaibkamil:
We don't use legacy `buffer_t` anymore at all, so this should be good from our perspective.

steven-johnson:
I'm ready to land this on my side. Any objections?

PR #3461, opened 2018-11-29
Set estimates for training in Generator

steven-johnson:
Not sure if this is what you had in mind or not, but demonstrates putting all the values into estimates rather than command line. Includes fixes from https://github.com/halide/Halide/pull/3460 as well.

### DIFF
--- a/src/CostModelGenerator.cpp
+++ b/src/CostModelGenerator.cpp
@@ -156,39 +156,6 @@ public:
     Output<Buffer<float>> prediction_output{ "prediction_output", 1 };
     Output<Buffer<float>> loss_output { "loss_output", 0 };
 
-    void set_input_estimates() {
-        // Provide estimates for things that don't have hard bounds
-        num_stages.set_estimate(22);
-        batch_size.set_estimate(1024);
-
-        pipeline_features.dim(0).set_bounds_estimate(0, 56)
-                         .dim(1).set_bounds_estimate(0, 7)
-                         .dim(2).set_bounds_estimate(0, 22);
-
-        schedule_features.dim(0).set_bounds_estimate(0, 1024)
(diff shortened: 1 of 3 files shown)


abadams:
Change to script looks reasonable, but all those inputs and outputs  already have .bound directives. I just started trying to autotune that pipeline locally, and to get estimates I just made Dimension::set_bounds also call Dimension::set_bounds_estimates. It seems to me that if you have a concrete bound on something, that should act as an estimate too.

steven-johnson:
> I just made Dimension::set_bounds also call Dimension::set_bounds_estimates

SGTM, and should arguably land in main as-is. 

But maybe a potential issue with order-of-call for existing code? (Though.... I guess if there are concrete bounds, those should always take precedence over estimates. Maybe issue a warning if both are specified and they don't match?)

steven-johnson:
See: https://github.com/halide/Halide/pull/3462

steven-johnson:
That's not what I found when I tried, but maybe I need to re-sync to the tip of this branch?

abadams:
Possibly. There's an existing block of estimates set already in the if (auto_schedule) section of the schedule. The inputs should certainly not need estimates - bounds inference should work for those.

steven-johnson:
Maybe, but I did this by dumb trial and error: I removed everything and re-added them until RunGen no longer complained about missing estimates.

steven-johnson:
> I think the required estimates are already in the file now - I've been successfully using rungen on it.

If you're still using the bench_cost_model.sh that is checked in to this branch, ~all of the estimates are provided on the command line there.

steven-johnson:
(Note that this includes fixes from https://github.com/halide/Halide/pull/3465)

PR #3462, opened 2018-11-29
bound()/set_bounds() should also call estimate()/set_bounds_estimate()

steven-johnson:
Suggestion from @abadams: if we have hard boundaries, we should set the autoschedule estimate values to that as well.

(Lots of drive-by churn in the test that was updated, as the old version wouldn't show the values when mismatches were found, making debugging painful.)

### DIFF
--- a/src/Dimension.cpp
+++ b/src/Dimension.cpp
@@ -47,15 +47,19 @@ Expr Dimension::stride() const {
 
 Dimension Dimension::set_extent(Expr extent) {
     param.set_extent_constraint(d, extent);
-    // If we have an explicit bound on the extent, set that as the estimate as well.
-    param.set_extent_constraint_estimate(d, extent);
+    // Propagate constant bounds into estimates as well.
+    if (is_const(extent)) {
+        param.set_extent_constraint_estimate(d, extent);
+    }
     return *this;
 }
 
 Dimension Dimension::set_min(Expr min) {
     param.set_min_constraint(d, min);
-    // If we have an explicit bound on the min, set that as the estimate as well.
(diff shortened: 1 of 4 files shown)


steven-johnson:
Whoa, this has many strange failure modes I didn't anticipate. Stay tuned.

steven-johnson:
OK, this is not workable as-is: estimates must be constant expressions, but bounds don't have to be. e.g.

```
class ErrorCodes : public Halide::Generator<ErrorCodes> {
public:
    Input<Buffer<int32_t>>  input{ "input", 2};
    Input<int>              f_explicit_bound{"f_explicit_bound", 1, 0, 64};
    Output<Buffer<int32_t>> output{"output", 2};
    void generate() {
        Var x, y;
        output(x, y) = input(x, y);
        output.bound(x, 0, f_explicit_bound);
    }
};
```

Note that `f_explicit_bound` will never be provably constant, so we can't use it as an estimate.

src/Dimension.cpp
```diff
@@ -47,11 +47,15 @@ Expr Dimension::stride() const {
...
     param.set_extent_constraint(d, extent);
+    // If we have an explicit bound on the extent, set that as the estimate as well.
+    param.set_extent_constraint_estimate(d, extent);
     return *this;
 }
 
 Dimension Dimension::set_min(Expr min) {
     param.set_min_constraint(d, min);
+    // If we have an explicit bound on the min, set that as the estimate as well.
+    param.set_min_constraint_estimate(d, min);
```
abadams: How about

`if (is_const(min)) ...`

PR #3464, opened 2018-12-03
Fix tail selection strategy in autoscheduler

abadams:
I encountered a bug where loads from inputs via a trivially inlined Func weren't discovered by the logic in the autoscheduler that decides if RoundUp is safe for a tail strategy.


steven-johnson:
Ready to land?

PR #3466, opened 2018-12-04
Revert c072dccd8d05b4311ebee9ef69acbba3291db4fd

steven-johnson:
Apparently we have some code that is sensitive to iteration order, as the current Win32Dbg and WIn64Dbg failures were injected here. Reverting to std::map until we track down the order dependency.

Drive-by change to remove the non-const iterator, since the only caller (VectorizeLoops) was actually fine using the const iterator.

### DIFF
--- a/src/Scope.h
+++ b/src/Scope.h
@@ -2,7 +2,7 @@
 #define HALIDE_SCOPE_H
 
 #include <iostream>
-#include <unordered_map>
+#include <map>
 #include <stack>
 #include <string>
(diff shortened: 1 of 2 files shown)


steven-johnson:
Update: in my debugging, this change seemed to unequivocally be the injection point for the WinDbg failures we were seeing; doing further testing today in master, I'm seeing the same failure with this rollback in place, so clearly this isn't the cause (or not the only cause).

PR #3468, opened 2018-12-04
Restrict fused groups to require either all or none of the funcs to be used.

alexreinking:
It used to be allowed to fuse two funcs (via compute_with) in contexts where at least the parent func was used, but not necessarily all of the children. After discussion with @abadams it was decided to disallow this feature since it is confusing and not necessary since a user can introduce intermediate funcs to achieve the  same effect.

This PR adds that restriction.


abadams:
@psuriana  do you remember if this feature was important for any production code? It only shows up when you compute_with a Func that has multiple different realizations.

psuriana:
@abadams Not as far as I can remember. I think it would be safe to disable this feature.

PR #3252, opened 2018-08-29
Use posix_memalign/memalign/_aligned_malloc

steven-johnson:
This changes posix_allocator to use posix_memalign(), and adds android_allocator (using memalign) and windows_allocator (using _aligned_malloc/_aligned_free), rather than using the over-allocation trick we have been using.

This will need testing on Android devices before landing.

Note that I didn't attempt to modify qurt_allocator, as I don't know which API  variants are available there; someone from Qualcomm should probably take a look at possible improvements there.

PR #2162, opened 2017-06-28
sim_remote.cpp cannot rely on memalign(); roll our own for now

steven-johnson:
(For future reference: it appears that the memalign() available in the simulator runtime can return multiple overlapping allocations, which causes havoc and crashiness. Hopefully the simulator will be fixed, but in the meantime, we are using this workaround.)

PR #3471, opened 2018-12-06
Use memalign for aligned_malloc on device and sim for Hexagon

aankit-ca:
This is a part of https://github.com/halide/Halide/pull/3252 effort.
@steven-johnson @dsharletg @pranavb-ca @dpalermo


dsharletg:
Can we verify somehow that the issue addressed by https://github.com/halide/Halide/pull/2162 is not happening again? That failure was subtle and hard to diagnose. I believe the issue is that the 'size' and 'alignment' parameters were swapped, which happens to *mostly* work without issue most of the time.

dsharletg:
Also, should this be a PR to srj-alloc instead of master?

aankit-ca:
Changed the base for the PR.
For verification, we tried running our test examples. I'll confirm with Pranav and Dan to see if we can do anything to address the concerns.

steven-johnson:
> Can we verify somehow that the issue addressed by #2162 is not happening again?

+1

pranavb-ca:
> > Can we verify somehow that the issue addressed by #2162 is not happening again?
> 
> +1

We couldn't find a reproducer from this back from Jun'17, so @dpalermo is writing a synthetic stress test for memalign on the simulator to see if overlapping regions are returned. we'll keep you guys posted.

dpalermo:
Just an update but the memalign test I’m writing hasn’t shown any overlaps being generated.
It currently does:

    // A) For alignments from 2^0 -> 2^MAX_ALIGN,
    //    with fixed allocation size of FIX_SIZE bytes
    //     1) memalign NUM_BUF buffers, checking for overlap, and then free all
    //     2) NUM_RAND times, randomly free buffers and memalign a new buffer
    //        (keeping up to NUM_BUF active buffers at once)

    // B) For alignments from 2^0 -> 2^MAX_ALIGN,
    //    with random allocation sizes up to MAX_SIZE bytes
    //     1) memalign NUM_BUF buffers, checking for overlap, and then free all
    //     2) NUM_RAND times, randomly free buffers and memalign a new buffer
    //        (keeping up to NUM_BUF active buffers at once)

    // C) For random alignments from 2^0 -> 2^MAX_ALIGN,
    //    with random allocation sizes up to MAX_SIZE bytes
    //     1) NUM_RAND times, randomly free buffers and memalign a new buffer
    //        (keeping up to NUM_BUF active buffers at once)

With the following settings (runs for ~30minutes on the simulator):

#define NUM_BUF         100     // number of active buffers
#define MAX_ALIGN       12      // maximum alignment (2^max)
#define MAX_SIZE        100000  // maximum single memalign allocation size
#define FIX_SIZE        1024    // maximum single memalign allocation size
#define NUM_RAND        1000    // number of random allocations in one run 

I'm currently adding one more scenario at the moment (that allocates for a set of specified sizes...to get a mix of large and small in case there is anything going on there).  If you have any other suggested patterns you would like me to exercise, or if you would like the test soruce to try anything yourself, just let me know.

dsharletg:
I think when we looked at this last time, Ron was able to just look at the code and see that the alignment and size parameters were swapped in the implementation. I think if you review the code for that bug and don't see it now, that's probably sufficient.

The test that you've written is good too, but it's hard to be certain about anything unless we also reproduce the issue with the same test when we had this bug.

dpalermo:
Pranav and I have both reviewd the memalign source and neither of us think the arguments are swapped.  We have also talked with the library maintainer and he agree that isn't a problem.  There was however an overflow check added to the code earlier this year that could have been responsible for bad allocations for large alignments & sizes.  It now returns a 0 status if an overflow was detected.  

Do we know what was run before that triggered the problem?...or was it just intermittent?

dpalermo:
Looking at the history for memalign.c, the only other change since 2012 was the addition of an aligned_malloc wrapper for C11 compatibility.  Was Ron possibly referring to arguments swapped in the Halide Hexagon runtime source call to memalign?

dsharletg:
I wondered the same thing, but it wasn't backwards in our code: https://github.com/halide/Halide/pull/2162/files#diff-ee58b82ab840ffa20b0c87d65f893fa7L119

dpalermo:
Updated the random alignment test, and added the scenario selecting sizes randomly from a specified set (using:  16 100 128 1024 1152 1280 1048704) to force a mix of small buffers and large:

    // C) For random alignments from 2^0 -> 2^MAX_ALIGN,
    //    with random allocation sizes up to MAX_SIZE bytes
    //     1) memalign NUM_BUF buffers, checking for overlap, and then free all
    //     2) NUM_RAND times, randomly free buffers and memalign a new buffer
    //        (keeping up to NUM_BUF active buffers at once)

    // D) For alignments from 2^0 -> 2^MAX_ALIGN,
    //    with random allocation sizes from sizes[] buffer
    //     1) memalign NUM_BUF buffers, checking for overlap, and then free all
    //     2) NUM_RAND times, randomly free buffers and memalign a new buffer
    //        (keeping up to NUM_BUF active buffers at once)

All these tests pass as well with no overlap detected...total simulator runtime ~45minutes. Test attached (test-memalign.c) if you want to try anything else.


[test-memalign.c.txt](https://github.com/halide/Halide/files/2665447/test-memalign.c.txt)

dsharletg:
I think we should just go ahead with this. I don't see how we are going to get any more closure without reproducing the original issue, and it's not worth the effort to do that. Dan's test and reviewing the code for whatever Ron saw seem good enough to me. If we start seeing mysterious unexplained problems, it will be easy to rule out this issue early.

pranavb-ca:
> I think we should just go ahead with this. I don't see how we are going to get any more closure without reproducing the original issue, and it's not worth the effort to do that. Dan's test and reviewing the code for whatever Ron saw seem good enough to me. If we start seeing mysterious unexplained problems, it will be easy to rule out this issue early.

+1

PR #3473, opened 2018-12-06
Remove stale comment about halide_set_num_threads()

shoaibkamil:
Trivial, but confuses users.  Now that macOS/iOS use pthreads, `halide_set_num_threads()` does indeed have an effect.


shoaibkamil:
Given there's no actual change to code, merging.

PR #3475, opened 2018-12-06
Add a temporary means to unroll loops only when they turn out to be constant-sized

abadams:


steven-johnson:
Please add an issue to the effect of "Remove temporary unroll hack" so that this doesn't stay in forever.

steven-johnson:
Windows failure is irrelevant buildbot nonsesne.

PR #3480, opened 2018-12-07
Add ability to dynamically add inputs and outputs to a Generator

steven-johnson:
This is a requested feature that should simplify some metaprogramming and autoscheduling concerns.


slomp:
Really sorry that I missed this...
Really weird; I did not get an e-mail notification (unless I accidentally deleted and did not realize).

test/generator/configure_generator.cpp
```diff
@@ -0,0 +1,86 @@
...
+class Configure : public Halide::Generator<Configure> {
+public:
+    GeneratorParam<int> num_extra_buffer_inputs{"num_extra_buffer_inputs", 3};
+
+    Input<Buffer<int>> input{"input", 3};
+    Input<int> bias{"bias"};
+
+    Output<Buffer<int>> output{ "output", 3 };
+
+    void configure() {
```
slomp: In my implementation, I make configure take a `std::string` as a parameter.
This string comes from the registered name of the generator.

For example, if I have a generator class `MyGenClass` with the following specializations:
```
HALIDE_REGISTER_GENERATOR( MyGenClass, mygen_8bpp )
HALIDE_REGISTER_GENERATOR( MyGenClass, mygen_16bpp )
HALIDE_REGISTER_GENERATOR( MyGenClass, mygen_hdr )
```
the `configure()` method would receive either "mygen_8bpp", "mygen_16bpp" or "mygen_8hdr", which I can then inspect to make decisions about how to configure the inputs/outputs.

Either that, or there should be a way to query the base `Generator<>` class about the name of the registered stub that it is generating code for (I could not find an interface for that).

test/generator/configure_generator.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+        // Pointers returned by add_input() are managed by the Generator;
+        // user code must not free them. We can stash them in member variables
+        // as-is or in containers, like so:
+        for (int i = 0; i < num_extra_buffer_inputs; ++i) {
+            auto *extra = add_input<Buffer<>>("extra_" + std::to_string(i), UInt(8), 2);
+            extra_buffer_inputs.push_back(extra);
+        }
+
+        typed_extra_buffer_input = add_input<Buffer<int16_t>>("typed_extra_buffer_input", 2);
```
slomp: `add_input<T>` is fine, but what would be even better is an interface like `add_input(std::string descriptor)`.
This would for a true, flexible and generic meta-programming facility within `configure()`, without the having to rely on the horrendous C++ type system and template machinery.
(btw, we can still have `add_input<T>` as a wrapper around `add_input(std::string descriptor)`).

test/generator/configure_generator.cpp
steven-johnson: Examining GeneratorParams is a better and more granular way to accomplish the same thing.

test/generator/configure_generator.cpp
steven-johnson: What does `descriptor` look like?

abadams: But is the input a Buffer, a Func, or some scalar? If you are going to use the input in the generate() method it's going to need to have a c++ type.

slomp: And yes, we'd need a "generic" `Input<void>` type to allow for fully customizable types.

abadams: There would have to be some piece of code that parses these string descriptions and dynamically creates objects with the correct actual C++ type to use in the generate() method. The idea of add_input is that it's the method that code or something similar would call. It's to enable metaprogramming of generators in a type-safe way. You can always layer something like generators-specified-by-json-strings on top of it, but the underlying fully typed layer would still be necessary.

test/generator/configure_generator.cpp
steven-johnson: No, because `GeneratorParam<string>` isn't an option for `GeneratorParam` :-)

No, what I'm saying is that instead of passing in a string which you have to parse to figure out the meaning(s) of, you should have GeneratorParam(s) controlling what you actually want. (In your example, an int that is 8 or 16 for instance.)

abadams: Though it would be nice to have a generatorparam<string> for the cases where you *do* have something json-like that specifies the structure of your pipeline. Right now you'd have to get it out-of-band, e.g. by doing file io inside configure().

test/generator/configure_generator.cpp
slomp: Ah, indeed, I guess I can just build these abstractions on top of the typed layer.

test/generator/configure_generator.cpp
steven-johnson: Understood. Again, though, if it would it be possible for you to share an example of what your existing configure() method looks like, how it's called, etc., it would help inform the discussion, I think.

slomp: In addition, I am OK having `configure` take no argument. I already cooked up a recipe with `GeneratorParam<>`, as you suggested, to pass the generator name.
(I am basically using `GeneratorParam<uint64_t>` to pass a `const char*` with the registered generator name to the `configure()` method.

abadams:
Is this ready to merge?

steven-johnson:
It could be with trivial work; it hasn't landed because it wasn't clear to me that the discussion had come to a conclusion (especially re @slomp issues). I'll bring it up to date with master.

steven-johnson:
Wanted to ping @slomp on this again to see where we stand on this PR adequately meeting his needs.

abadams:
re-ping @slomp

slomp:
Looks good to me! 👍 
I might have some suggestions in the future, but it covers a lot of what I was currently hacking on my side.

(Sorry, but I don't know why I am not getting notifications for this PR in particular!)

steven-johnson:
Ready to land as soon as the bots go green.

slomp:
**ProTip**: if you stop receiving github notifications on a PR, unsubscribe from the PR notifications and subscribe again.

slomp:
OK to merge?

PR #3485, opened 2018-12-10
Add the ability for generators to load plugins

abadams:
Adds a flag to gengen to ask it to dlopen something for us. Can be used to install a custom autoscheduler (by having a static constructor in the library call Pipeline::set_custom_autoscheduler). I imagine it could be used for other purposes in the future too.

Drive-by fix of line wrapping in the gengen usage message.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -2,6 +2,12 @@
 #include <fstream>
 #include <set>
 
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
 #include "Generator.h"
 #include "Outputs.h"
 #include "Simplify.h"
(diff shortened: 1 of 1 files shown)


src/Generator.cpp
```diff
@@ -786,20 +792,33 @@ std::string halide_type_to_c_type(const Type &t) {
...
+        "     [assembly, bitcode, cpp, h, html, o, static_library,\n"
+        "      stmt, cpp_stub, schedule].\n"
+        "     If omitted, default value is [static_library, h].\n"
+        "\n"
+        " -x  A comma separated list of file extension pairs to substitute during\n"
+        "     file naming, in the form [.old=.new[,.old2=.new2]]\n"
+        "\n"
+        " -p  A comma-separted list of shared libraries that will be loaded before the\n"
+        "     generator is run. Useful for custom auto-schedulers. The generator must\n"
+        "     either be linked against a shared libHalide or compiled with -rdynamic\n"
```
steven-johnson: Linking with a shared libHalide is going to be fun in the google3 environment, but that's my problem, not yours :-)

zvookin:
Relying on static constructors is likely going to bite us sooner or later. An immediate concern is that the plug-in has to make a call across to Halide while it is the middle of constructing itself and all its state may or may not be complete.

The solution would be to use dlysym to get a well known named entrypoint and call it if it is present. (One could also use the entrypoint, but dlsym without failure if the symbol is missing, gives one hte choice of the current behavior or getting a call to a function after all the libraries static constructors have run.)

this also comes up in static linkage in that the call to register the autoscheduler could happen before Pipeline initializes itself. This is likely not possible as the Pipeline global is probably statically initialized, but that requires some assumptions on implementation.

PR #3488, opened 2018-12-11
A few fixes for OSX compatibility

steven-johnson:


apps/autoscheduler/Makefile
```diff
@@ -63,12 +66,6 @@ $(BIN)/demo.a: $(BIN)/demo.generator $(BIN)/auto_schedule.so
 $(BIN)/demo.rungen: $(BIN)/demo.a $(BIN)/runtime.a
 	$(CXX) -std=c++11 -I $(BIN) -DHL_RUNGEN_FILTER_HEADER="\"demo.h\"" -I ../../include ../../tools/RunGenMain.cpp ../../tools/RunGenStubs.cpp  $(BIN)/demo.a -o $@ $(HALIDE_SYSTEM_LIBS) -lpng -ljpeg
 
-$(BIN)/test: test.cpp $(BIN)/auto_schedule.so
```
abadams: Oops, I forgot to check in test.cpp originally. I've added it, so please restore this block of the makefile.

steven-johnson:
Done.

Right now the samples all get output in-tree; IMHO we should either emit them into `bin/samples` or add a `.gitignore` for them. (I'd favor the former but can live with the latter.)

abadams:
Having make clean nuke a bunch of samples gathered overnight is disastrous. There shouldn't be any makefile target capable of deleting them and it should be really hard to delete them accidentally too. So .gitignore?

apps/support/Makefile.inc
```diff
@@ -31,6 +31,20 @@ ifeq ($(UNAME), Darwin)
 CXXFLAGS += -fvisibility=hidden
 endif
 
+ifeq ($(UNAME), Linux)
+USE_DYNAMIC_LOOKUPS=-rdynamic
```
zvookin: I think this would be better called "USE_EXPORT_DYNAMIC".

PR #3489, opened 2018-12-12
Grab bag of minor cleanups

steven-johnson:
(These PR is a result of doing a walkthrough of apps/autoscheduler to be sure I understood the workflow, in anticipation of porting it to Google's internal build system; some of these are probably sensible changes, some are probably pointless nitpickery in code style, and some may actually be counterproductive. Happy to cherrypick any/all of these.)

- AutoSchedule: attempt listing all the env vars used. (Exact definitions needed for each.)
- AutoSchedule: dump the generated C++ schedule to HL_SCHEDULE_FILE, in addition to having it lumped in with the general log
- autotune_loop.sh and RunGen.h: make the benchmark/compile loop output a little less verbose
- autotune.sh: split insanely long lines into multilines
- autotune.sh: give each sample variant a unique function name, this allows us to skip some extra output and to distinguish variants later on more easily. (Not entirely sure about this one, TBH.)
- train_cost_mode: emit the best result in path-ish form as well, trim iteration output to a single line
- train_cost_mode: if HL_BEST_SCHEDULE_FILE is defined, emit the info about 'best schedule' to that file (as well as stdout); autotune_loop.sh defines this to be `samples/best.txt`, the goal here being that the 'best' is kept up to date in the samples folder
- autotuneloop.sh: make GENERATOR, PIPELINE, and HL_TARGET required script arguments

### DIFF
--- a/apps/autoscheduler/train_cost_model.cpp
+++ b/apps/autoscheduler/train_cost_model.cpp
@@ -1,7 +1,9 @@
-#include <string>
-#include <vector>
+#include <iomanip>
 #include <set>
+#include <sstream>
+#include <string>
 #include <unistd.h>
+#include <vector>
 
 #include "ThroughputPredictorPipeline.h"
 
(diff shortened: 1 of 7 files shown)


apps/autoscheduler/train_cost_model.cpp
```diff
@@ -46,6 +48,23 @@ bool ends_with(const string &str, const string &suffix) {
     return true;
 }
 
+void log_best_schedule(int best_id, float best_runtime) {
+    const int batch = best_id / 100;
```
abadams: The /100 %100 thing is a function of the way in which the autotuning script is currently set up, and probably doesn't belong in the binary that trains the cost model. The filename of the best sample file could be tracked below just as easily as its schedule id, so there's no need to recreate it here. In the comparison of `runtime < best_runtime` you can just grab the variable `s` as well to keep track of the sample file that had the lowest runtime in it.

steven-johnson: Done

PR #3490, opened 2018-12-12
halide_type_of<>() should support some non-C99 int types

steven-johnson:
(Not sure if this is correct yet; opening a PR to get complete buildbot testing)

### DIFF
--- a/src/runtime/HalideRuntime.h
+++ b/src/runtime/HalideRuntime.h
@@ -1890,6 +1890,25 @@ HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<int64_t>() {
     return halide_type_t(halide_type_int, 64);
 }
 
+// Some programming environments use aliases like `typedef long long int64`
+// (rather than the C99 integer types); for various reasons, these aren't
+// always considered the same type by C++, but we'd like to quietly accept them
+// in Halide, so add mappings here to the values we expect (along with static_asserts
+// to verify that the situation remains as we expect).
+template<>
+HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<long long>() {
+    static_assert(sizeof(long long) == sizeof(int64_t),
+                  "long long is expected to be a 64-bit signed int here");
+    return halide_type_t(halide_type_int, 64);
+}
+
+template<>
+HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<unsigned long long>() {
+    static_assert(sizeof(unsigned long long) == sizeof(uint64_t),
+                  "unsigned long long is expected to be a 64-bit unsigned int here");
+    return halide_type_t(halide_type_uint, 64);
+}
+
 #endif
 
 #endif // HALIDE_HALIDERUNTIME_H


src/runtime/HalideRuntime.h
```diff
@@ -1890,6 +1890,25 @@ HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<int64_t>() {
     return halide_type_t(halide_type_int, 64);
 }
 
+// Some programming environments use aliases like `typedef long long int64`
+// (rather than the C99 integer types); for various reasons, these aren't
+// always considered the same type by C++, but we'd like to quietly accept them
+// in Halide, so add mappings here to the values we expect (along with static_asserts
+// to verify that the situation remains as we expect).
+template<>
+HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<long long>() {
```
abadams: Does this work out OK in environments in which int64_t == long long? I guess the buildbots will tell us.

zvookin:
This is not a good change. I'll need detail on why this is needed. Please note we do not have a template specialization for e.g. ```int```. There is nothing in C++ that requires long long to be 64-bits so it is inherently non-portable. Also, I expect it would fail if ```int64_t``` ends up as "typedef long long int64_t```, but perhaps that works via some other magic.

Really folks need to use stdint.h types when interoperating with Halide.

We could write template enable fu to match any integral type with the right number of bits and signedness and map it to our type if we wanted to go down this road.

zvookin:
Yep, the redefinition issue is showing up on Travis.

steven-johnson:
Background:

The issue here is some code internal to google that is using `Buffer<int64>` (a google-specific type, which is defined as `long long` in our build env) instead of `Buffer<int64_t>`. Since `long long` doesn't have a `halide_type_of` definition, this didn't compile; rather than seeking our guidance, the authors of that code used some template-fu to substitute `int64_t` for `int64` in certain situations... which happened to work because the move-ctor in Buffer neglected to check for `assert-can-convert-from` at the time. 

We added the assert-can-convert-from in master, but it's been disabled inside google's copy for a while because code had grown up that was incompatible; this was almost all code that was doing illegal `Buffer<const T> -> Buffer<T>` conversions, which I've been whittling away at fixing; this is a different variant, wherein the conversion isn't legal because `int64_t` and `long long` aren't identical types, although they happen to be bit-identical in our build environment.

This PR was really a test to see if failures were likely in any of our well-supported build environments, in hopes of an ugly-but-expedient fix. Alas, that's clearly not gonna happen.

(From our point of view, simply requiring them to use the C99 int types everywhere is the obvious and correct fix, but this is a deeply religious issue that is not one worthy of wading into.)

> We could write template enable fu to match any integral type with the right number of bits and signedness and map it to our type

Yeah. An alternate approach might be to use something like 

```
    std::enable_if<!std::is_same<int64_t, long long>::value && sizeof(long long) == sizeof(int64_t)>
```

steven-johnson:
Here is a revised approach; I'm not sure if this is a good idea or not, but putting it out for comment. Basically: we keep the existing implementations for use when compiling the runtime (where we can't use type_traits), but use an entirely different implementation elsewhere (where we can use type_traits to allow all valid type aliases).

steven-johnson:
Meh: I forgot about float16_t. This is turning out to be less a can of worms and more a can-of-ebola-filled-snakes.

steven-johnson:
(Pushed out another fix just for reference. That said, I'm going to look into a fix on my side instead.)

PR #3493, opened 2018-12-12
Added a #ifdef to use the large llvm CodeModel

benoitsteiner:
By default, Halide relies on the llvm::CodeModel::Small. Added a #ifdef to use the llvm::CodeModel::Large instead.


steven-johnson:
LGTM

src/CodeGen_Internal.cpp
```diff
@@ -459,7 +459,11 @@ std::unique_ptr<llvm::TargetMachine> make_target_machine(const llvm::Module &mod
                                                 mcpu, mattrs,
                                                 options,
                                                 llvm::Reloc::PIC_,
+#ifdef HALIDE_USE_CODEMODEL_LARGE
+                                                llvm::CodeModel::Large,
```
steven-johnson: Worth adding a comment here to the effect of "Halide is not regularly tested with CodeModel::Large; this is is provided for experimentation and may break"

zvookin:
What is the use case here? I'd expect a Target feature might be more useful.

benoitsteiner:
We use llvm::CodeModel::Large internally. I don't know the details, but it looks like there are small memory leaks in LLVM when using llvm::CodeModel::Small. This option enables us to upgrade Halide without having to manually modify the code every time.

PR #3498, opened 2018-12-13
Various fixes for apps/autoscheduler

steven-johnson:
(Things found while bringing up in google build env)

-- revise cost_model_schedule.h to be valid C++ when compiled on its own (ie, with the proper includes, wrapped in a function), rather than relying on inclusion in the proper spot; this is good hygiene in general, and essentially required for C++ modules to work in the future. (Unless there is disagreement with this change, I intend to make a followup PR to have the C++-schedule-dumping-code do this sort of wrapping by default.)

-- remove unused lamba in cost_model_generator.cpp

### DIFF
--- a/apps/autoscheduler/cost_model_generator.cpp
+++ b/apps/autoscheduler/cost_model_generator.cpp
@@ -1,6 +1,8 @@
 #include "Halide.h"
 #include "Derivative.h"
 
+#include "cost_model_schedule.h"
+
 using namespace Halide;
 
 // A model weight is either just an input, or an input and an output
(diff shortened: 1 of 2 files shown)


apps/autoscheduler/Derivative.cpp
```diff
@@ -5,7 +5,6 @@
 #include <set>
 
 #include "Halide.h"
-#include "Errors.h"
```
abadams: Errors.h is supposed to be apps/autoscheduler/Errors.h - there's no Errors.h in Halide proper.

abadams:
The schedule output as currently envisioned is not intended to be a valid entire C++ translation unit. It's a snippet of source code that users will typically copy-paste when they want to freeze a given schedule. We could revise that model to treat it more like generated source - probably a good thing - but we should limit the amount of time we spend on this sort of thing prior to the deadline.

apps/autoscheduler/Derivative.cpp
steven-johnson: Well darn. I misread, that one little "s" caught me. Will fix and update this PR as appropriate.

steven-johnson:
> We could revise that model 

understood.

steven-johnson:
backed out the unnecessary changes to Derivative. The changes to the prebaked schedule for cost_model_schedule.h will make my life easier in the short term when pulling these changes into Google repeatedly; agreed that we can figure out better long-term ways to manage the generated code in the future.

PR #3505, opened 2018-12-13
Fix an issue that device frees are not injected.

jingpu:
PR fixes an issue that device frees are not injected in InjectHostDevBufferCopies.cpp sometimes.

The problem is that the Block visitor of InjectBufferCopiesForSingleBuffer (which uses IRMutator2's Block visitor) will canonicalize nested Block's, which can destroy the "last_use" Stmt pattern we tracked in the AST, resulting in the free injector fails to find a match.

### DIFF
--- a/src/InjectHostDevBufferCopies.cpp
+++ b/src/InjectHostDevBufferCopies.cpp
@@ -435,7 +435,7 @@ private:
         }
     }
 
-    // We want to break things down into a serial sequence of leaf
+    // We break things down into a serial sequence of leaf
     // stmts similar to InjectBufferCopiesForSingleBuffer.
     void visit(const For *op) override {
          check_and_record_last_use(op);


abadams:
Ah, that sounds like a nasty bug, but I'm not a fan of this solution. The point of canonicalizing block layout was that other code can assume blocks are always laid out that way. If it's conditional then it's not an invariant anymore and it's worthless.

The way the non-gpu version of this pass is intended to work is to scan over the IR without mutating anything to find the last uses, then search again in a mutator checking each unmutated IR node to see if it's the designated last use. The GPU version appears to try to take a shortcut by tracking the last use as it injects copies, which seems like it's a bad idea.

jingpu:
The windows buildbot failures don't seem related to me. PTAL.

abadams:
The failing test is related to the performance of async on gpu. It could plausibly be related. I'll rerun to see if it's just a flake.

PR #3511, opened 2018-12-18
Fixed a memory corruption in the generation of tiles

benoitsteiner:


apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -2564,7 +2564,7 @@ struct LoopNest {
 
         if ((!is_root() || f->is_output || !force_only_output_compute_root) &&
             !innermost &&
-            (!in_realization || size.empty() || size[vector_dim] == 1)) {
+            (!in_realization || size.empty() || (vector_dim < size.size() && size[vector_dim] == 1))) {
```
abadams: Nice catch, but I think this should be:
```
!in_realization || size.empty() || vector_dim == -1 || size[vector_dim] == 1
```

vector_dim should never be larger than size.size(). We could also assert that before the if.

steven-johnson: > vector_dim should never be larger than size.size()

Could use sizse.at(vector_dim) in that case.

benoitsteiner: In the instance of the bug that I faced, vector_dim was 2 and size.size() was also 2. I couldn't figure out exactly where it came from, since it looks like vector_dim should always be less than size.size(). However I noticed that the code used to compute the boolean may_slide also handles the case where vector_dim >= size.size(), so I replicated the fix (even though it looks more like a workaround based on your comments)

abadams: Ugh, this code has a bug that I fixed in my more experimental branch where vector_dim refers to the wrong Func. I'll merge my experimental work back into this branch today.

PR #3513, opened 2018-12-19
Made the autoscheduler code a little bit more modular.

benoitsteiner:


apps/autoscheduler/ThroughputPredictor.h
```diff
@@ -0,0 +1,25 @@
+#include "HalideBuffer.h"
+
+namespace Halide {
+class ThroughputPredictor {
+ public:
+  virtual ~ThroughputPredictor() { }
```
steven-johnson: super-nit: is `=default` preferred?

apps/autoscheduler/ThroughputPredictor.h
```diff
@@ -0,0 +1,25 @@
...
+  virtual void set_pipeline_features(const Runtime::Buffer<float> &pipeline_feats, int n) = 0;
+  virtual void enqueue(int ns, Runtime::Buffer<float> *schedule_feats, double *cost_ptr) = 0;
+  virtual void evaluate_costs() = 0;
+  virtual void reset() = 0;
+};
+
+
+class ThroughputPredictorPipeline : public ThroughputPredictor {
```
abadams: I'm confused about this derived class. It appears to do nothing but has the same name as ThroughputPredictorPipeline below

benoitsteiner: I'm not using it yet, but I will in my follow up set of changes. I'll move some of the code that search through the set of possible schedule into a separate file. That code will take a ThroughputPredictor instead of a ThroughputPredictorPipeline in argument. That will allow me to plug other neural network architectures in the scheduler. The code in AutoSchedule.cpp will continue to use the ThroughputPredictorPipeline since that's the best we have atm.

apps/autoscheduler/Featurization.cpp
```diff
@@ -0,0 +1,22 @@
+#include "Featurization.h"
+#include "FunctionDAG.h"
+
+namespace Halide {
+void compute_pipeline_featurization(const Pipeline &pipeline, const Target& tgt, const MachineParams &params, std::unordered_map<Stage, PipelineFeatures, StageHasher> *features) {
```
abadams: I'm not sure this interface makes sense. You can get the pipeline features this way with a temporary internal FunctionDAG, but you also need a FunctionDAG to compute schedule-specific features, and the DAG is expensive to construct - it's where all the front-loaded analysis of the pipeline is done (e.g. bounds of things).

abadams:
I'm really worried that this amount of code motion is going to make it impossible for the MIT people to reconcile their changes against this branch. Is it possible to proceed in smaller steps?

abadams:
I'll start by merging Steven's change to the throughput predictor then putting it behind the abstract base class. That should get us halfway there and doesn't touch AutoSchedule.cpp

abadams:
Ugh, no I need to focus on getting results - I shouldn't be moving code around.

benoitsteiner:
The code is currently stored in a big .cpp file. I need to expose the APIs in header files in order to be able to use parts of the code while leaving others alone. That's why I created a couple header files. I moved the description of the features in the first header, and the FunctionDAG class in the second. I could refactor the code to have the declaration of the FunctionDAG class in FunctionDAG.h and leave the definition of its methods in AutoSchedule.cc, but that's likely to make it even harder to merge any change further down the road.
I'll get rid of the new Featurization.cpp file and move the code directly as a new method of FunctionDAG to avoid the need to create a FunctionDAG just to extract the pipeline features.

abadams:
I made changes this morning that would have invalidated this PR, so I redid it and pushed the result. The one thing I didn't do is add the new API to extract the features from the FunctionDAG.

brom84:
Szvf

PR #3514, opened 2018-12-19
Move getenv() usage towards the edges, Part 1

steven-johnson:
Using getenv() to configure things is super-fast-and-easy, but makes tracking the configuration of which tools need which vars set (and which don't) painful; this is especially painful in some build environments (e.g. Bazel) where changing an env var can decide to rebuild all dependencies, even the ones that don't care.

Removing env vars entirely isn't going to happen soon, but we can move in that direction by pushing the calls to getenv() further outward. This starts the ball rolling by changing ThroughputPredictorPipeline to take a Param structure; this structure can trivially be initialized via getenv(), but the onus is on the TPP's owner to choose that option.


abadams:
I created a massive merge conflict here so I just isolated the use of environment variables myself.

PR #3517, opened 2018-12-19
add $CXXFLAGS to $CCFLAGS in Python bindings

minrk:
to ensure that $CXXFLAGS are respected when compiling the Python bindings


steven-johnson:
Cloned as https://github.com/halide/Halide/pull/3518 for buildbot testing

PR #3520, opened 2018-12-20
Don't crash when env vars aren't set

steven-johnson:
creating a std::string from a NULL ptr tends to be crashy


abadams:
Just got back from dinner to find something I launched before dinner crashed for this very reason. Sigh.

PR #3521, opened 2018-12-20
Rework RunGen registration model

steven-johnson:
Expand to allow for multiple filters to link against RunGen at once, selecting via a flag; this is of limited use for routine purposes, but should give us greater flexibility when experimenting with multiple auto-scheduler models.

### DIFF
--- a/tools/RunGenStubs.cpp
+++ b/tools/RunGenStubs.cpp
@@ -1,13 +1,36 @@
-#define HALIDE_GET_STANDARD_ARGV_FUNCTION halide_rungen_redirect_argv_getter
-#define HALIDE_GET_STANDARD_METADATA_FUNCTION halide_rungen_redirect_metadata_getter
+#define HALIDE_REGISTER_ARGV_AND_METADATA halide_rungen_register_argv_and_metadata
 
-// This is legal C, as long as the macro expands to a single quoted (or <>-enclosed) string literal
+extern "C" void halide_rungen_register_filter(
+    int (*filter_argv_call)(void **),
+    const struct halide_filter_metadata_t *filter_metadata,
+    void* registerer);
+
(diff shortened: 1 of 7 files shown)


tools/RunGenStubs.cpp
```diff
@@ -1,13 +1,36 @@
...
+
+    Linux:      -Wl,--whole-archive "/path/to/library" -Wl,-no-whole-archive
+    Darwin/OSX: -Wl,-force_load,/path/to/library
+    VS2015 R2+: /WHOLEARCHIVE:/path/to/library.lib
+    Bazel:      alwayslink=1
+*/
+
+namespace {
+
+struct Registerer {
```
abadams: What if the Halide runtime maintained a list of all linked filters using something equivalent to the mechanism below (static constructors). The list would be exposed with a getter in HalideRuntime.h in the metadata section.

abadams: This would have to be implemented in codegen_llvm alongside the code that embeds the metadata getter, I think. I don't mean to stall this PR on that suggestion though.

steven-johnson: I thought about implementing this with `halide_register_argv_and_metadata()` being a weak halide runtime method, which defaults to an empty implementation (RunGen and the like could override it), but would be problematic on Windows and of course a setter wouldn't suffice due to initialization-order issues.

Re: having a default mechanism that always registers them, it's an interesting idea that is worth thinking about, but I'd want to be sure we weren't accidentally defeating dead-stripping of genuinely unused filters for shipping code.

steven-johnson:
(I need to do some integration testing inside google before landing this, even once the buildbots go green, to be sure no additional tweaking is needed)

PR #3522, opened 2018-12-20
Fix HalideTraceViz on windows.

LazyDodo:
HalideTraceViz makes really trippy 80's style visualizations on windows.

badviz

This is due to windows by default treating stdin/out as text files, switching them to binary fixes the issue.

goodviz

### DIFF
--- a/util/HalideTraceViz.cpp
+++ b/util/HalideTraceViz.cpp
@@ -12,6 +12,7 @@
 
 #ifdef _MSC_VER
 #include <io.h>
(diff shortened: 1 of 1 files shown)


abadams:
Oh wow. Are you sure this PR is an improvement? :)

LazyDodo:
>Oh wow. Are you sure this PR is an improvement? :)

Perhaps we could expose it as flag for halidetraceviz for people that are into this kind of thing? :)

PR #3524, opened 2018-12-20
Reset unique name counters before the compilation of each target.

jingpu:
The PR makes the function/variable names look consistent during the compilation for different targets when using the multi-target AOT flow. It is for improving the debugging experience.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -999,14 +999,15 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
             Outputs output_files = compute_outputs(targets[0], base_path, emit_options);
             auto module_producer = [&generator_name, &generator_args]
                 (const std::string &name, const Target &target) -> Module {
+                    // Reset the counters so the function/variable names look
+                    // more consistent for different targets.
+                    reset_unique_name_counters();
+
                     auto sub_generator_args = generator_args;
                     sub_generator_args.erase("target");
                     // Must re-create each time since each instance will have a different Target.
                     auto gen = GeneratorRegistry::create(generator_name, GeneratorContext(target));
                     gen->set_generator_param_values(sub_generator_args);
-                    // Reset the counters so the function/variable names look
-                    // more consistent for different targets.
-                    reset_unique_name_counters();
                     return gen->build_module(name);
                 };
             if (targets.size() > 1 || !emit_options.substitutions.empty()) {
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1004,6 +1004,9 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
                     // Must re-create each time since each instance will have a different Target.
                     auto gen = GeneratorRegistry::create(generator_name, GeneratorContext(target));
                     gen->set_generator_param_values(sub_generator_args);
+                    // Reset the counters so the function/variable names look
+                    // more consistent for different targets.
+                    reset_unique_name_counters();
                     return gen->build_module(name);
                 };
             if (targets.size() > 1 || !emit_options.substitutions.empty()) {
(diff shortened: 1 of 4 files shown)


src/Generator.cpp
```diff
@@ -1004,6 +1004,9 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
                     // Must re-create each time since each instance will have a different Target.
                     auto gen = GeneratorRegistry::create(generator_name, GeneratorContext(target));
                     gen->set_generator_param_values(sub_generator_args);
+                    // Reset the counters so the function/variable names look
+                    // more consistent for different targets.
+                    reset_unique_name_counters();
```
abadams: This is not quite safe, because code in another thread could create pipelines with duplicated names in it if this runs while it is being defined. I'm sure there's a simple solution to getting consistent autogenerated names per generator invocation, but it's eluding me. Maybe some sort of thread-local naming context? Zalman or Steven might have an idea.

zvookin:
Probably the right way to do it is to have a pass in lowering that rebases all the temp names to 0. This could either find the lowest value and subtract it from all names or could just rename them in the order encountered. The latter is implementable by not using strings for the names in the first place and doing the conversion to a unique string name as part of codegen.

Other possibilities include using thread locals for the counters or just insisting that Generator.cpp is not thread safe at all.

src/Generator.cpp
steven-johnson: In all the current AOT build systems I'm aware of, there won't be multiple threads in the same process targeting this.... but yeah, it's a bug waiting to happen. Making the unique-name-counter stuff use thread locals would probably be a safe and simple fix (at a nonzero but probably small runtime cost.)

jingpu:
I changed the counters to thread_local. I am not sure if we want it be both thread_local and std::atom. They sounds redundant to me.

@zvookin 
Having another lowering pass to rebase the indices or getting rid of the strings doesn't necessary solves my issue. My main issue is the function name which gets uniquified in the ctor of Func is a user facing value (via Func::name()). I want to make this value consistent across different target.

jingpu:
The failures looks pre-existing on master. PTAL

abadams:
I don't see how thread locals are safe here - what happens when someone mixes pieces of IR created by different threads?

jingpu:
> what happens when someone mixes pieces of IR created by different threads?

Well, it won't be safe in that case. The thread locals are just for addressing Steven's concerns that some AOT system potentially parallels the compilation for different targets. (I don't know there is an intended use case where threads will exchange IRs. Could you explain how it would be useful?)

If this thread local approach is not acceptable, I can try to restrict the change to only affect function names chosen during construction of Func in a generator, which I don't expect to ever become multi-threaded. Any thoughts on it?

abadams:
unique_name is a very general global facility for producing per-process unique identifiers, e.g. for temporary variable in lets. This PR currently changes all symbols ever generated by Halide in a way that makes it impossible to do multi-threaded IR generation. We couldn't ever parallelize slow parts of lowering this way.

There's got to be a more targeted way to fix your problem... if we're in generate_filter_main called from GenGen.cpp we know there isn't anything else going on in the process, because we're inside our own int main, and it would be OK to reset global counters there.... which is what your PR originally did before the thread locals. Sorry, I didn't think of the calling context when I originally complained about resetting the counters. At that call site it's fine. I'd be good with this PR if you reverted from thread locals back to atomic ints.

abadams:
... Nobody would call generate_filter_main inside a parallel loop would they? Probably should comment that it's not thread-safe because it resets global state to discourage that.

steven-johnson:
> Nobody would call generate_filter_main inside a parallel loop would they? 

I'd assume/hope not, but you never know. Agreed that documenting that it isn't thread-safe (and is intended to be called from a trivial real `main` ala GenGen.cpp is an adequate fix.

alexreinking:
> if we're in generate_filter_main called from GenGen.cpp we know there isn't anything else going on in the process, because we're inside our own int main, and it would be OK to reset global counters there....

We just hit a bug that showed this assumption to be incorrect. Globally initialized vars (or other unique named classes) can get their names before the reset. The camera pipeline in apps did this.

I think @zvookin had the right idea with last-minute pass to rebase the generated names. Though it would also be worth considering using longs (maybe uint32s?) for names.

steven-johnson:
Since this was purely for the purpose of improving debugging, we should just roll this back and open an issue.

PR #3527, opened 2018-12-21
Store all the schedule features using an int64. This ensures that the…

benoitsteiner:
… number of features can still be correctly computed by dividing sizeof(ScheduleFeatures) by sizeof(int64).


abadams:
Oops!

PR #3528, opened 2018-12-21
Merge to master + update RunGen changes

steven-johnson:
RunGen registration now allows multiple filters to be linked at once, so autotune_loop.sh now takes advantage of this to link all the variants of a batch into a single executable, which is run with different flags for each sample.


apps/autoscheduler/autotune_loop.sh
```diff
@@ -107,19 +119,43 @@ for ((i=$((FIRST+1));i<1000000;i++)); do
...
     echo Compiling ${BATCH_SIZE} samples for batch_${i}...
     for ((b=0;b<${BATCH_SIZE};b++)); do
         S=$(printf "%d%02d" $i $b)
         FNAME=$(printf "%s_batch_%02d_sample_%02d" ${PIPELINE} $i $b)
-        make_sample "${DIR}/${b}" $S $FNAME &
+        make_sample_lib "${DIR}/${b}" $S $FNAME &
     done
     wait
 
+    echo Compiling ${BATCH_SIZE} stubs for batch_${i}...
```
abadams: Does this encapsulate failures in the same way as the old script? If anything went wrong compiling any one sample in the old script, it didn't affect other samples. I could imagine this bringing down the whole batch.

apps/autoscheduler/autotune_loop.sh
```diff
@@ -107,19 +119,43 @@ for ((i=$((FIRST+1));i<1000000;i++)); do
...
+        ALWAYS=`alwayslink $f`
+        STUBS="${STUBS} ${ALWAYS}"
+    done
+
+    echo Linking batch_${i}...
+    c++ -std=c++11 \
+        ../../tools/RunGenMain.cpp \
+        -I ../../include \
+        ${STUBS} \
+        ${DIR}/*/*.a \
```
abadams: I guess this line only collects the successfully-compiled samples?

apps/autoscheduler/autotune_loop.sh
steven-johnson: Yes, this definitely would bring down the whole batch. But how would that have worked before? ie, if one sample failed to compile, what happened when the script tried to execute it -- wouldn't it fail then?

apps/autoscheduler/autotune_loop.sh
steven-johnson: Yes, but it would fail to link, since the RunGenStubs are basically never going to fail to generate.

apps/autoscheduler/autotune_loop.sh
abadams: Yes, and it results in a truncated sample file, which train_cost_model just rejects at training time.

abadams: (rejects gracefully - it skips it)

steven-johnson: Sounds like I should try injecting deliberate failures to see what happens and mimic the old behavior.

steven-johnson:
I'm going to close this pending the investigation of failure modes. (No special changes are needed to merge master into this branch; existing makefiles will continue to work as-is.)

apps/autoscheduler/autotune_loop.sh
steven-johnson: > If anything went wrong compiling any one sample in the old script, it didn't affect other samples. 

How likely is this in practice, and/or when do you expect it to happen? (Getting Bazel to fail gracefully in this situation is going to require some thought.)

abadams: The random pipeline generator doesn't always generate valid pipelines, and a random schedule on a pipeline might not be valid due to a corner case bug in the new autoscheduler. Both of these are fairly frequent.

steven-johnson: Ouch -- that will make my favored technique harder to make work, because reasons. Better to find out now, though.

PR #3530, opened 2018-12-21
Added a constructor for the pipeline feature class. I…

benoitsteiner:
…'m getting some NaN when training a cost model for a long time. It's not clear yet if the NaNs come from the inputs or from the network itself, so I'm just making sure that the inputs are always initialized to rule this out.


abadams:
It's already zero-initialized (see the initializers on the members)

abadams:
Actually that's only true for the ScheduleFeatures

benoitsteiner:
Right, I deleted the default constructor for ScheduleFeatures.

PR #3531, opened 2018-12-22
Remove RunGenStubs; add registration as a Generator output type

steven-johnson:
This simplifies some use cases with multiple-registration for RunGen when used in conjunction with Bazel/Blaze: we no longer have to do preprocessor tricks (or, for some build systems, sed tricks) to get RunGenStubs to work; the new, generated stubs are 100% standalone C++ code.


steven-johnson:
(Note that this will require some trivial changes to the standalone_autoscheduler branch, but I have those ready to go if/when this lands.)

PR #3537, opened 2018-12-27
Refactor libauto_schedule.so Makefile

steven-johnson:
Move the rules for libauto_schedule.so (and friends) into autoscheduler.inc, so that the apps directories aren't dependent on apps/autoscheduler being built first. (Unbreaks 'make test_apps')


apps/support/autoscheduler.inc
```diff
@@ -0,0 +1,68 @@
+AUTOSCHED_SRC=../autoscheduler
+
+WEIGHT_OBJECTS=\
+$(BIN)/weights_head1_conv1_weight.o \
```
abadams: Doesn't the use of BIN here imply that a separate copy of the autoscheduler is built for each app? Should it be AUTOSCHED_BIN or something?

steven-johnson: Yep, good point. Let me tweak.

steven-johnson: PTAL

apps/random_pipeline/bench_comparison.sh
```diff
@@ -17,7 +17,7 @@ SCHEDULES=1
 RANDOM_DROPOUT=100
 BEAM_SIZE=50
 
-make -C ../autoscheduler bin/libauto_schedule.so
+make -C ../autoscheduler ../autoscheduler/bin/libauto_schedule.so
```
abadams: This change looks fishy to me. make -C already sets the working directory to its argument.

steven-johnson: Will fix.

PR #3539, opened 2018-12-27
 Added apis to compute the number of schedule and pipeline features

benoitsteiner:


apps/autoscheduler/Featurization.h
```diff
@@ -13,6 +13,11 @@ struct PipelineFeatures {
         std::memset(this, 0, sizeof(PipelineFeatures));
     }
 
+    static constexpr int NumFeatures() {
```
abadams: nit: We're doing underscores for methods and variables: num_features

benoitsteiner: Fixed

PR #3541, opened 2018-12-27
Add code formatting and other cleanups. [skip ci]

mbrukman:
Additional cleanups:

* added a top-level heading for the project name to make it stand out in
  the README
* fixed capitalization when referring to project names rather than their
  binary or directory names
* replaced / (forward slash) with \ (backward slash) for sample Windows
  paths, and put them in code formatting as well
* removed escaping of underscores since code formatting avoids Markdown
  interpretation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# Halide
+
 Halide is a programming language designed to make it easier to write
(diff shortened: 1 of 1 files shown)


abadams:
Thanks.

PR #3544, opened 2018-12-28
Autoscheduler gpu2

rbaghdadi:
This is a basic code that allows the standalone autoscheduler to generate GPU code. We will be adding more features to it and making it more robust/general gradually.


apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -113,28 +120,35 @@ vector<vector<int64_t>> generate_tilings(const vector<int64_t> &s, int d, int fa
...
-                }
 
-                // The sequence above (in terms of the inner loop) goes 1 2 4 8 16 ...
-                // but 3 is an important inner tiling factor for matrix multiply ops.
-                int inner3 = (d == vector_dim) ? 3*vector_size : 3;
-                int outer3 = (s[d] + inner3 - 1) / inner3;
-                if (factor == 2 && inner3 < s[d] && outer3 < s[d] && outer3 > 1) {
-                    if (inner3 * outer3 * 7 <= s[d] * 8) {
-                        t.back() = outer3;
+                // Additional tiling options may have inner extents that are not
```
abadams: None of this should be necessary anymore. compute_here has already created an inner loop equal to the vector size, so any tiling produced here is already implicitly a multiple of the vector size

apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -1756,6 +1771,28 @@ struct State {
...
+
+        static int64_t limit = get_shared_memory_limit();
+
+        if (limit == 0) {
+            return false;
+        }
+
+        // If the working set is too large on the GPU, shared memory will be
+        // exhausted, so reject any such schedules
+        for (const auto& c : root->children) {
```
abadams: You can tell if something is a child of root by checking feat.inner_parallelism, so here you could say feat.inner_parallelism > 1 && feat.working_set > limit instead of looping

apps/autoscheduler/AutoSchedule.cpp
abadams: Ah, I said this because I think that the vector loop should be gpu_threads, but I see below that you've just picked the first three loops you find inside the gpu_blocks loops. This is going to create far too many threads a lot of the time due to data dependencies. The loop structure should be blocks -> serial loops -> threads

apps/autoscheduler/FunctionDAG.h
```diff
@@ -15,6 +15,10 @@ namespace Internal {
 
 namespace {
 
+bool gpu_mode() {
+    return atoi(get_env_variable("HL_GPU_MODE").c_str());
```
abadams: No need for a fresh environment variable: use target.has_gpu_feature()

apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -124,14 +132,16 @@ vector<vector<int64_t>> generate_tilings(const vector<int64_t> &s, int d, int fa
...
 
-                // The sequence above (in terms of the inner loop) goes 1 2 4 8 16 ...
-                // but 3 is an important inner tiling factor for matrix multiply ops.
-                int inner3 = 3;
-                int outer3 = (s[d] + inner3 - 1) / inner3;
-                if (factor == 2 && inner3 < s[d] && outer3 < s[d] && outer3 > 1) {
-                    if (inner3 * outer3 * 7 <= s[d] * 8) {
-                        t.back() = outer3;
-                        result.push_back(t);
+                if (!target.has_gpu_feature()) {
```
abadams: What's wrong with 3 as a tiling factor here? It's not like it's the innermost loop.

apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -2243,16 +2285,53 @@ struct State {
...
+                    if (!it->exists || it->extent == 1) continue;
+                    if (!it->parallel) break;
+
+                    n_parallel_loops++;
+                }
+
+                // Tag as GPUBlock and GPUThread only if we have at least two loops.
+                // We want the loop structure to be GPU blocks -> serial loops -> GPU threads.
+                // We tag the first min(3, n_parallel_loops/2) loops to be GPU blocks and
+                // tag the last min(3, n_parallel_loops/2) loops to be GPU threads.
```
abadams: There are no guarantees on the size of anything but the innermost loop, so this looks like it would die a lot of the time. Do you have an idea of what fraction of random pipelines this successfully runs?

abadams: (die due to using too many threads)

abadams:
Taking this off the PRs page for now during review period.

PR #3552, opened 2019-01-04
Add 'make autotune' targets to some more apps

steven-johnson:
This makes apps/autoscheduler/autotune_loop.sh a bit more general, and adds `make autotune` targets to apps/hist and apps/harris. Opening this as a PR rather than pushing directly because there are a couple of minor changes in behavior for apps/autoscheduler (and anything else using the autotune_loop.sh script):

possibly-breaking:
- we now always copy the `weights` directory into `samples/weights` and work from there; the source `weights` directory is **not** updated by default. (If you want to update the results from a run, you must manually copy the weights back.) IMHO this is a feature, not a bug, but I wouldn't want anyone to lose work because of it.

minor:
- The script takes more arguments 
- we lazily add `disable_llvm_loop_unroll` and `disable_llvm_loop_vectorize` to HL_TARGET as needed (caller need not set these)
- fixed wrong estimate in harris
- cleaned up .gitignore to ignore sample/ directories everywhere under apps/


apps/autoscheduler/autotune_loop.sh
```diff
@@ -108,25 +128,25 @@ benchmark_sample() {
...
-    cp weights/* ${SAMPLES}/batch_${i}/
+    mkdir -p ${DIR}/weights_used/
+    cp ${WEIGHTS}/* ${DIR}/weights_used/
 
     echo Compiling ${BATCH_SIZE} samples for batch_${i}...
     for ((b=0;b<${BATCH_SIZE};b++)); do
         S=$(printf "%d%02d" $i $b)
         FNAME=$(printf "%s_batch_%02d_sample_%02d" ${PIPELINE} $i $b)
-        make_sample "${DIR}/${b}" $S $FNAME &
+        make_sample "${DIR}/${b}" $S $FNAME #&
```
abadams: No more parallel compilation?

steven-johnson: whoops, scalpel left in patient

PR #3553, opened 2019-01-04
Add autotune targets to other apps/ that can use it

steven-johnson:
NOTE: several of these reliably fail to compile within the 120-second timeout that is set by default, at least on my local machines:

- ~~interpolate_generator~~
- lens_blur
- local_laplacian
- ~~max_filter~~
- ~~stencil_chain~~

(However, those two local machines are a 12-core Linux box and a 4-core Macbook, so trying to build 32 at the same time is probably not helping. I may follow up with some tweakage to the autotune_loop.sh script to modify the number of parallel builds based on number of cores.)

UPDATE: with the tweaks just mentioned in place, now only two of them reliably time out for me; these are more complex pipelines so it might just be unavoidable for now.

I'm going to leave them in anyway (caveat emptor!) as they may be useful in helping profile speed issues in the autoscheduler.

### DIFF
--- a/apps/interpolate_generator/interpolate_generator.cpp
+++ b/apps/interpolate_generator/interpolate_generator.cpp
@@ -142,7 +142,7 @@ public:
                   .dim(2).set_bounds_estimate(0, 4);
             output_.dim(0).set_bounds_estimate(0, 1536)
                   .dim(1).set_bounds_estimate(0, 2560)
-                  .dim(2).set_bounds_estimate(0, 4);
+                  .dim(2).set_bounds_estimate(0, 3);
         }
     }
 };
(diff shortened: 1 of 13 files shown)


apps/interpolate_generator/interpolate_generator.cpp
```diff
@@ -122,7 +122,7 @@ class Interpolate : public Halide::Generator<Interpolate> {
                 }
                 normalize
                     .reorder(c, x, y)
-                    .bound(c, 0, 3)
+                    .bound(c, 0, 4)
```
abadams: I think interpolate is 4 input channels and 3 output channels, based on its filter.cpp

steven-johnson: Nice catch.

abadams: This also implies the bound(c, 0, 4) calls above should be reverted to bound(c, 0, 3), because I think those are both the output Func.

PR #3554, opened 2019-01-04
Track versions of the featurization.

benoitsteiner:
This helps manage the lifecycle of features extracted ahead of time, for example in order to build a training set for the cost model.


abadams:
I think you accidentally added an extra commit. In any case, I've been changing this file, so I added the version() methods.

PR #3555, opened 2019-01-04
Clarify where registration files come from

shoaibkamil:
This will save others some reverse-engineering I had to do :)


shoaibkamil:
No problem.  Thanks for implementing this, it's a huge timesaver.

PR #3561, opened 2019-01-08
Add filename and line number to debug message

asaadaldien:
Add filename and line number to the debug message. This is useful especially for people who are new to the codebase.


asaadaldien:
I agree with you @zvookin, macros are ugly and actually the more I run debugging the redundant code position text becomes more annoying. I probably overfitted few debugging runs I have early today.

PR #3565, opened 2019-01-09
Fix race conditions during CUDA runtime initialization

jia-kai:
I encountered segfault when calling a Halide cuda function from two threads. This is caused by `halide_cuda_initialize_kernels`, which calls the ctor of `Context`, which calls `load_libcuda` if `cuInit` is NULL. However if `Context::Context` is invoked from two threads, one thread may have resolved `cuInit` but not yet `cuCtxPushCurrent`, and another thread might skip calling `load_libcuda` because it finds `cuInit` to be non-null, and segfault follows when it calls uninitialized `cuCtxPushCurrent`.

This PR fixes this problem by adding a helper `ensure_libcuda_init`, which guards `load_libcuda` call by a spinlock.

### DIFF
--- a/src/runtime/cuda.cpp
+++ b/src/runtime/cuda.cpp
@@ -15,10 +15,15 @@ namespace Halide { namespace Runtime { namespace Internal { namespace Cuda {
 #define CUDA_FN_3020(ret, fn, fn_3020, args) WEAK ret (CUDAAPI *fn)args;
 #define CUDA_FN_4000(ret, fn, fn_4000, args) WEAK ret (CUDAAPI *fn)args;
 #include "cuda_functions.h"
+#undef CUDA_FN
+#undef CUDA_FN_OPTIONAL
+#undef CUDA_FN_3020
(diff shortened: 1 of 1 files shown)


steven-johnson:
LGTM

dsharletg:
This looks OK to me because if I'm reading the code correctly, it's only adding a new synchronization lock/unlock step after it already has skipped a fast already-initialized path: https://github.com/halide/Halide/blob/294133b82ad581d766b3fadfb44d940027866627/src/runtime/cuda.cpp#L141

zvookin:
How is that not double checked locking?

abadams:
It's double-checked locking, but we spent a while analyzing it to make sure it was safe. Hence all the comments.

abadams:
(To be clear - the double-checked locking is in the existing code - not the new code in the PR. The code in the PR unconditionally grabs a spinlock)

abadams:
Clean on the buildbots (#3614 ) . Ping @zvookin

PR #3566, opened 2019-01-09
autoscheduler improvements

benoitsteiner:
Moved the cost model and network size definition into the Halide namespace
Added an API to be able to call the autoscheduler code from c++


apps/autoscheduler/AutoSchedule.h
```diff
@@ -0,0 +1,19 @@
...
+#include "PerfectHashMap.h"
+#include "Featurization.h"
+#include <vector>
+
+namespace Halide {
+namespace Internal {
+
+typedef PerfectHashMap<FunctionDAG::Node::Stage, ScheduleFeatures> StageMapOfScheduleFeatures;
+
+extern "C" {
```
abadams: Why extern "C" linkage?

apps/autoscheduler/CostModel.h
```diff
@@ -3,6 +3,7 @@
 
 #include "HalideBuffer.h"
 
+namespace halide {
```
abadams: halide -> Halide

apps/autoscheduler/AutoSchedule.h
benoitsteiner: This was a quick way to work around the fact that the FunctionDAG class was in an anonymous namespace, and couldn't be linked properly. This isn't required anymore.

apps/autoscheduler/CostModel.h
benoitsteiner: Fixed

PR #3568, opened 2019-01-10
Precompiled test headers

alexreinking:
Building tests (eg. correctness) uses different flags than the default build and this invalidates the precompiled header (PCH) in those targets. This PR produces a test-compatible PCH alongside the default one and speeds up `build_tests` by almost 2x when using GCC.

Also moved `#include "Halide.h"` to the front of any test that includes it to be sure that the PCH will be used. Changed instances of `#include <Halide.h>` to use double quotes for consistency.


abadams:
CXX=g++ make build_tests -j32 for me goes from 1:16 to 0:53 :+1: Let's hope it makes it through the buildbots.

alexreinking:
I should have specified that the 2x speedup was using `-j4` on my laptop :slightly_smiling_face:

PR #3570, opened 2019-01-10
Added support for halide_hexagon_set_thread_priority()

dpalermo:
- Added halide_hexagon_set_thread_priority IDL API to set the
    priority of threads created in the Halide runtime
  - Works similarly to halide_hexagon_set_performance[_mode]
  - A user would call halide_hexagon_set_thread_priority from the
    host app before dispatching a pipeline

    // Increase the default priority of Halide Hexagon threads
    halide_hexagon_set_thread_priority(NULL, 50);

  - Threads will be created with the previously specified priority
  - If not called, the default priority will be used
  - Compatible with old runtimes - priority won't be set w/ old
  - Set qurt_thread_pool.cpp priority default to 100 (to match master thread)
  - Limit priority range from 1 to 255
  - Also set priority on existing master thread
  - Update QURT_THREAD_ATTR_PRIORITY_DEFAULT to match v62 qthread.h
  - Added more documentation to HalideRuntimeHexagonHost.h

    /** Set the priority for Hexagon threads.
     * - Valid priority values range from 1 to 255
     * - Smaller number for higher priority
     * - The highest priority for a user thread is 1, 0 reserved for OS usage
     * - If not set, Halide thread priority will default to 100
     * - This should be called before running a pipeline. */
    // @{
    extern int halide_hexagon_set_thread_priority(void *user_context, int priority);
    // @}

  - Added mmap_dlsym_libs that will search the list of loaded libs as
    8996/sdm820 was failing to find halide_set_default_thread_priority
    using mmap_dlsym
  - For dlopenbuf platforms, using halide_get_symbol instead of dlsym
    directly in order to pick up Dillon's fallback technique of
    checking RTLD_SELF & RTLD_DEFAULT

### DIFF
--- a/src/runtime/HalideRuntimeHexagonHost.h
+++ b/src/runtime/HalideRuntimeHexagonHost.h
@@ -79,7 +79,7 @@ typedef enum halide_hexagon_power_mode_t {
  * @param mipsPerThread - mips requested per thread, to establish a minimal clock frequency per HW thread
  * @param mipsTotal - Total mips requested, to establish total number of MIPS required across all HW threads
  * @param set_bus_bw - Set to TRUE to request bus_bw
- * @param bwMeagabytesPerSec - Max bus BW requested (megabytes per second)
+ * @param bwMegabytesPerSec - Max bus BW requested (megabytes per second)
  * @param busbwUsagePercentage - Percentage of time during which bwBytesPerSec BW is required from the bus (0..100)
  * @param set_latency - Set to TRUE to set latency
  * @param latency - maximum hardware wakeup latency in microseconds.  The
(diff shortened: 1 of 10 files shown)


dsharletg:
Can we first do a PR that just changes the thread priority of thread pool threads, and the pipeline context thread to the same value, something reasonable like 100? Basically, just change these two numbers to the same thing:
https://github.com/halide/Halide/blob/master/src/runtime/qurt_threads.cpp#L47
https://github.com/halide/Halide/blob/master/src/runtime/hexagon_remote/pipeline_context.h#L58

I'm not sure about all of the new features in this PR. The PR doesn't even fix the discrepancy in priority unless one uses halide_hexagon_set_thread_priority, which I don't think we should require people to do. The first thing to do is just make the two constants the same value.

steven-johnson:
What's the status on this?

dpalermo:
This PR does fix the default discrepancy in priority (both master & worker are now defaulting to 100).  You don't need to call halide_hexagon_set_thread_priority() first.  The only other feature this PR adds is the ability for the halide app to specify the priority before dispatching a pipeline...by calling halide_hexagon_set_thread_priority() before the dispatch...in the same way we currently call halide_hexagon_set_performance() before the dispatch.  If there are other comments on the PR, let me know.

dpalermo:
Closing this PR and will break it into multiple/smaller PRs.

PR #3577, opened 2019-01-12
Add crosscompile script for apps

steven-johnson:
Still doing some torture testing but this should be right: use `cross_compile_everything generate` to produce all the .a files to a specific dir, then (on the dest machine) use `cross_compile_everything reassemble` to copy the .a files to the correct bins. (Then you can `make test` as usual.)

### DIFF
--- a/apps/.gitignore
+++ b/apps/.gitignore
@@ -1,1 +1,2 @@
 samples/
+bin1/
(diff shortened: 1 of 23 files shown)


steven-johnson:
Some stupid race conditions, hopefully fixed now

steven-johnson:
Is this still necessary/desirable?

abadams:
It was looking like it right up till the deadline (hence me pinging you Monday morning), but it turns out the AWS arm instances we were using can indeed compile everything on-board. We should use them for buildbots - 16 cores, lots of ram, very cheap.

PR #3580, opened 2019-01-14
Add handling of absd() in Bounds.cpp

steven-johnson:
Formerly this defaulted to bounds-of-type, but that meant that widened expressions never attempted to use the known bounds of the narrow types (e.g. absd(i16(a), i16(b)) would assume [0,65535] even if a and b are known to be i8)

Also, drive-by addition of some debugging infrastructure for Bounds, disabled by default (ala EXPR_MUTATIONS in SImplify)

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -960,6 +960,11 @@ private:
                 b.accept(this);
                 Interval b_interval = interval;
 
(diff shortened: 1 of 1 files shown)


steven-johnson:
Meh, this doesn't handle all the cases I thought it would; not ready for review.

PR #3581, opened 2019-01-14
Add handling of absd() in Bounds.cpp

steven-johnson:
Formerly this defaulted to bounds-of-type, but that meant that widened expressions never attempted to use the known bounds of the narrow types (e.g. absd(i16(a), i16(b)) would assume [0,65535] even if a and b are known to be i8)

Also, drive-by addition of some debugging infrastructure for Bounds, disabled by default (ala EXPR_MUTATIONS in Simplify)

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1,6 +1,7 @@
 #include <iostream>
 
 #include "Bounds.h"
+#include "ConciseCasts.h"
 #include "CSE.h"
 #include "Debug.h"
 #include "Deinterleave.h"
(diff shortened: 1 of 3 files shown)


src/Bounds.cpp
```diff
@@ -861,6 +965,44 @@ class Bounds : public IRVisitor {
...
+                Expr e = abs(op->args[0] - op->args[1]);
+                e.accept(this);
+            } else {
+                // absd() for int types will always produce a uint result
+                internal_assert(t.is_uint());
+
+                // Recover what the matched types for the args were
+                Expr a = op->args[0];
+                Expr b = op->args[1];
+                match_types(a, b);
```
abadams: Why is this necessary if the absd function already did it?

src/Bounds.cpp
```diff
@@ -861,6 +965,44 @@ class Bounds : public IRVisitor {
...
+                b.accept(this);
+                Interval b_interval = interval;
+
+                a_interval.min = simplify(a_interval.min);
+                a_interval.max = simplify(a_interval.max);
+                b_interval.min = simplify(b_interval.min);
+                b_interval.max = simplify(b_interval.max);
+
+                if (a_interval.is_bounded() && b_interval.is_bounded()) {
+                    // Cast to 64-bit type to minimize cast-out-of-range issues;
```
abadams: To avoid introducing wider types, how about:

interval.max = max(absd(a_interval.max, b_interval.min), absd(a_interval.min, b_interval.max))

src/Bounds.cpp
steven-johnson: You are correct, the match_type'd versions of the args are saved in the op.

src/Bounds.cpp
steven-johnson: ah yes, but brings up another fun fact: `absd()` isn't handled in Simplify_Call.cpp, so `max(absd((uint8)20, (uint8)5), absd((uint8)10, (uint8)30))` isn't simplified down to `(uint8)20` as it should be, thus the newly added test fails

steven-johnson: (adding absd() handling to simplify now)

Issue #3582, opened 2019-01-15
compute_function_value_bounds() ignores impure Funcs with const RDom ranges


steven-johnson:
If you have code that is using (e.g.) `sum()` with fixed-at-compile-time values (for convenience), e.g.

```
      const int radius = 8;
      RDom r(-radius, 2 * radius + 1, -radius, 2 * radius + 1);
      return sum(pixel_cost(x_ + r.x, y_ + r.y));
```
Then compute_function_value_bounds() doesn't even attempt to compute a value bounds: the anonymous function inside `sum()` is impure, so it just skips it entirely and gives it an unbounded range.

You can work around this by rewriting the sum as an explicit loop in C++, of course, but it seems like checking for constant RDom ranges and attempting to calculate useful boundaries here would be worthwhile.

steven-johnson:
As a maybe-better-than-nothing thought, Funcs of this sort that have non-overflowing types (int16, int8, etc) could at least be bounded to bounds-of-type rather than unbounded. (With care for tupled funcs.)

PR #3589, opened 2019-01-15
Use bounds-of-type for bounds of some impure functions (Issue #3582)

steven-johnson:
Also, drive-by fix to move the previous test for uint math into check_constant_bound(), where it belonged in the first place

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -6,6 +6,7 @@
 #include "Deinterleave.h"
 #include "ExprUsesVar.h"
 #include "Func.h"
+#include "InlineReductions.h"
 #include "IR.h"
 #include "IREquality.h"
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -2182,6 +2183,18 @@ FuncValueBounds compute_function_value_bounds(const vector<string> &order,
                 }
 
                 fb[key] = result;
+            } else {
+                // If the Func is impure, we may still be able to specify a bounds-of-type here
+                Type t = f.output_types()[j].element_of();
+                if ((t.is_uint() || t.is_int()) && t.bits() <= 16) {
+                    result = Interval(t.min(), t.max());
+                } else {
+                  result = Interval::everything();
```
abadams: indentation

PR #3590, opened 2019-01-16
Made the scheduler more user friendly, by throwing an assertion inste…

benoitsteiner:
…ad of asserting when it can't find a valid schedule

### DIFF
--- a/apps/autoscheduler/AutoSchedule.cpp
+++ b/apps/autoscheduler/AutoSchedule.cpp
@@ -2947,7 +2947,9 @@ IntrusivePtr<State> optimal_schedule_pass(FunctionDAG &dag,
         std::unordered_map<uint64_t, int> hashes;
         q.swap(pending);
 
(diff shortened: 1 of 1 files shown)


apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -2947,7 +2947,9 @@ IntrusivePtr<State> optimal_schedule_pass(FunctionDAG &dag,
         std::unordered_map<uint64_t, int> hashes;
         q.swap(pending);
 
-        internal_assert(!pending.empty());
```
abadams: internal_assert should already throw an exception when WITH_EXCEPTIONS is on. Perhaps it needs to be #defined separately because this is built separately to Halide?

benoitsteiner: Indeed, closing the PR.

PR #3596, opened 2019-01-17
Be more careful about padding out to the narrowest type possible

abadams:


abadams:
This lets us reenable a bunch of simd op check.

64-bit comparisons still don't generate cleanly, but they're not too bad - it just widens to the next type up.

src/CodeGen_X86.cpp
```diff
@@ -94,10 +94,12 @@ void CodeGen_X86::visit(const GT *op) {
         // split it up ourselves.
 
         Type t = op->a.type();
-        int slice_size = 128 / t.bits();
-        if (slice_size < t.lanes()) {
-            slice_size = target.natural_vector_size(t);
-        }
+        int vec_bits = t.lanes() * t.bits();
```
steven-johnson: surely there's a `get_slice_size()` function waiting to be created here

test/correctness/simd_op_check.cpp
```diff
@@ -82,7 +82,9 @@ struct Test {
         target = get_target_from_environment()
             .with_feature(Target::NoBoundsQuery)
             .with_feature(Target::NoAsserts)
-            .with_feature(Target::NoRuntime);
+            .with_feature(Target::NoRuntime)
+            .with_feature(Target::DisableLLVMLoopUnroll)
```
steven-johnson: We should really make it a point to move forward into making these set by default (I presume the major obstacle being risk of injecting bugs and/or perf regressions in existing code)

abadams: Yes, they may cause perf regressions on existing code where there's room for better scheduling (which there very often is). I enabled them here because humans read this assembly, and any automatic unrolling and vectorizing is a red herring.

src/CodeGen_X86.cpp
abadams: done

steven-johnson:
Looks like simd-op-check is still failing.

abadams:
Ah, it passes for all targets except for AVX1-only. Will fix.

PR #3598, opened 2019-01-18
Reorder AddImageCheck asserts by buffer names instead of assert types. 

asaadaldien:
CodeGen for halide functions with multiple inputs takes a long time. -time-pases showing 70% of the time is spent in register allocation. Even for a simple halide pipeline adding 256 inputs takes 43.sec to compile.

This PR reorders asserts into blocks acting on the same buffer. It seem to help codegen time ad running run_performance_lots_of_inputs with 512 input takes 15% less in Regallocation time.


src/AddImageChecks.cpp
```diff
@@ -13,15 +13,50 @@ using std::pair;
 using std::string;
 using std::vector;
 
+namespace {
+
+// Sotres buffer asserts Stmt's ordered by buffer name and assert type.
```
steven-johnson: Did you mean "Sorts"?

src/AddImageChecks.cpp
```diff
@@ -13,15 +13,50 @@ using std::pair;
...
+class BufferAsserts {
+public:
+    enum Type { DimsNoOverflow,
+                Constrained,
+                Proposed,
+                Required,
+                ElemSize,
+                HostAllignment,
```
steven-johnson: HostAlignment

src/AddImageChecks.cpp
```diff
@@ -13,15 +13,50 @@ using std::pair;
...
 /* Find all the externally referenced buffers in a stmt */
 class FindBuffers : public IRGraphVisitor {
 public:
     struct Result {
         Buffer<> image;
         Parameter param;
         Type type;
-        int dimensions{0};
-        bool used_on_host{false};
+        int dimensions{ 0 };
```
steven-johnson: Don't make unrelated formatting changes (here and elsewhere); those are based saved for separate PRs, as it makes the "real" changes much harder to parse out in the diff.

test/performance/lots_of_inputs.cpp
```diff
@@ -0,0 +1,44 @@
+#include "halide_benchmark.h"
+#include <Halide.h>
+
+using namespace Halide;
+using namespace Halide::Tools;
+
+int main(int argc, char **argv) {
+    // TODO(ataei): Increase the size once we have a reasnable compilation time.
```
steven-johnson: reasonable

test/performance/lots_of_inputs.cpp
```diff
@@ -0,0 +1,44 @@
+#include "halide_benchmark.h"
+#include <Halide.h>
+
+using namespace Halide;
+using namespace Halide::Tools;
+
+int main(int argc, char **argv) {
+    // TODO(ataei): Increase the size once we have a reasnable compilation time. 
+    // This takes several inutes for bigger sizes now.
```
steven-johnson: minutes

asaadaldien:
@steven-johnson Its my local clang-format! will revert those!

src/AddImageChecks.cpp
```diff
@@ -13,6 +15,41 @@ using std::pair;
 using std::string;
 using std::vector;
 
+namespace {
+
+// Stores buffer asserts Stmt's ordered by buffer name and assert type.
+class BufferAsserts {
+public:
+    enum Type { DimsNoOverflow,
```
steven-johnson: Nit: indentation doesn't match that of other enums in Halide, please fix

steven-johnson: Nit: new enums should probably use `enum class`

asaadaldien: @steven-johnson BufferAsserts defined in an anonymous namespace we shouldn't have  ambiguity between Halide::Type & BufferAsserts::Type.

steven-johnson: Ambiguity to the human reader, not to the compiler.

asaadaldien:
Reordering by buffer is slightly faster:

Before:
```
Compile time with 1 inputs = 0.554068 s
Compile time with 2 inputs = 0.077662 s
Compile time with 4 inputs = 0.120946 s
Compile time with 8 inputs = 0.210784 s
Compile time with 16 inputs = 0.442599 s
Compile time with 32 inputs = 1.049370 s
Compile time with 64 inputs = 2.981606 s
Compile time with 128 inputs = 8.911081 s
```

After:
```
Compile time with 1 inputs = 0.570248 s
Compile time with 2 inputs = 0.062023 s
Compile time with 4 inputs = 0.099298 s
Compile time with 8 inputs = 0.164235 s
Compile time with 16 inputs = 0.361655 s
Compile time with 32 inputs = 0.834294 s
Compile time with 64 inputs = 2.320249 s
Compile time with 128 inputs = 6.965614 s
```

asaadaldien:
Actually no compile-time improvement by doing this reordering. The real wins are coming from doing int32 cmp for type check.

PR #3599, opened 2019-01-19
Check for signed_integer_overflow in bounds_of_expr_in_scope()

steven-johnson:
I have some (private) code that generates some insanely complicated expressions (by way of multiple Generator concatenations via Stub), which end up creating bounds expressions that include signed_integer_overflow(), which cause compilation to fail. From Gitter discussions, it seems the most likely fix is to treat these as unbounded.

Unfortunately, I don't yet have a test case that I can share; the code that I have that causes this is both quite proprietary and quite complex, and so far my efforts to extract a reasonable sub-part that demonstrates the issue haven't succeeded; that said, I want to go ahead and offer the fix for discussion to see if I'm on the right track. (It does seem to fix issue inside our codebase.)

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -31,6 +31,12 @@ using std::string;
 using std::vector;
 
 namespace {
+
+bool is_signed_integer_overflow(Expr e) {
+  const Call *call = e.as<Call>();
+  return call && call->is_intrinsic(Call::signed_integer_overflow);
+}
+
 int static_sign(Expr x) {
     if (is_positive_const(x)) {
         return 1;
@@ -46,18 +52,6 @@ int static_sign(Expr x) {
     }
     return 0;
 }
-
-class HasSIO : public Halide::Internal::IRVisitor {
-    using IRVisitor::visit;
-
-    void visit(const Halide::Internal::Call *call_expr) override {
-        if (call_expr->is_intrinsic(Halide::Internal::Call::signed_integer_overflow)) count++;
-        IRVisitor::visit(call_expr);
-    }
-public:
-    int count;
-    HasSIO() : count(0) {}
-};
 
 }  // anonymous namespace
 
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -1185,6 +1198,19 @@ Interval bounds_of_expr_in_scope(Expr expr, const Scope<Interval> &scope, const
             << " should have been a scalar of type " << expected
             << ": " << b.interval.max << "\n";
     }
+
+    // If there is signed integer overflow, replace it with unbounded.
+    {
```
abadams: bounds_of_expr_in_scope is used in lots of tight inner loops, and this creates a second entire recursive pass over the returned IR, so I worry about performance.

Is it not the case that signed_integer_overflow propagates upwards, and thus only the output needs to be checked to see if it equals it? Alternatively, the Bounds visitor itself could set a flag if it discovers such a call.

steven-johnson: > Is it not the case that signed_integer_overflow propagates upwards, and thus only the output needs to be checked to see if it equals it?

Apparently not, as the final LetStmt expression I encounter has it as just part of the expression. Will post as followup.

> Alternatively, the Bounds visitor itself could set a flag if it discovers such a call.

That sounds reasonable.

abadams: No, I guess not. That will probably open a can of worms regarding cases where the signed integer overflow is technically unreachable, or doesn't affect the output value.

steven-johnson: Now that we mention it, though, it is a little curious that it isn't propagated outward here. Are there specific guidelines in the simplifier for how SIO should be propagated, or is it just ad hoc?

abadams: It's ad-hoc. I'm really unsure as to whether it should be propagated through casts.

steven-johnson: Just for grins, I'm going to add that behavior to simplify for casts and see what happens.

steven-johnson:
PTAL

PR #3600, opened 2019-01-20
fixed LLVM 8 compilation issue

slomp:


shoaibkamil:
LGTM.  Confirmed that master doesn't compile on LLVM8, but with this patch it does.

steven-johnson:
Interesting, thought I checked against LLVM8, but either I didn't or they updated it with this change too.

PR #3601, opened 2019-01-22
Propagate signed_integer_overflow and indeterminate_expressions thru Casts

steven-johnson:
We're not sure if this is the right thing to do in all cases, but doing this would make checking for inappropriate existence of these in a post-final-lowering IR tree simpler. See also PR #3599

(Appears to still pass all tests locally. Will pull into Google and do some internal testing.)

### DIFF
--- a/src/IR.h
+++ b/src/IR.h
@@ -594,12 +594,6 @@ struct Call : public ExprNode<Call> {
         return make(param.type(), param.name(), args, Image, FunctionPtr(), 0, Buffer<>(), param);
     }
 
-    /** Convenience constructor for indeterminate_expression() */
-    static Expr make_indeterminate_expression(Type type);
-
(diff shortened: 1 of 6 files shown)


steven-johnson:
FWIW, this passes all tests in Halide and Google, so it doesn't break any existing check.

src/IR.h
```diff
@@ -594,6 +594,12 @@ struct Call : public ExprNode<Call> {
         return make(param.type(), param.name(), args, Image, FunctionPtr(), 0, Buffer<>(), param);
     }
 
+    /** Convenience constructor for indeterminate_expression() */
+    static Expr make_indeterminate_expression(Type type);
```
abadams: These feel like they belong better as free functions in IROperator.cpp

steven-johnson: Fair, will fix

PR #3604, opened 2019-01-23
 Add bitwise_not, xor, or to Bounds detection

steven-johnson:
(Note, built on top of PR #3603, which should land first)

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1012,7 +1012,10 @@ private:
             op->args[1].accept(this);
         } else if (op->is_intrinsic(Call::shift_left) ||
                    op->is_intrinsic(Call::shift_right) ||
-                   op->is_intrinsic(Call::bitwise_and)) {
+                   op->is_intrinsic(Call::bitwise_and) ||
+                   op->is_intrinsic(Call::bitwise_not) ||
+                   op->is_intrinsic(Call::bitwise_xor) ||
+                   op->is_intrinsic(Call::bitwise_or)) {
             Expr simplified = simplify(op);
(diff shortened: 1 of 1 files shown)


abadams:
The existing code is a bit of a hack there. It simplifies the IR to canonicalize things like << 2 into *4, but then it throws away the result. Using it for canonicalization is one thing, but using it for constant folding makes it clear just how gross it is. I wonder if we can do better and actually handle the cases that would be canonicalized or folded correctly instead of calling the simplifier. The folding case would look something like:

`if (a.is_single_point() && b.is_single_point()) interval = Interval::single_point(op)`

For the canonicalization cases, I think shift left and shift right are monotonic in both args, so the min and max of the result is just `{a.min << b.min, a.max << b.max}` for shift left, and `{a.min >> b.max, a.max >> b.min}` for shift right. For bitwise and on unsigned types, I think it's `{0, min(a.max, b.max)}`...


Alternatively, we could ensure that simplify_exprs runs before bounds inference.

steven-johnson:
I'm sure we could improve it -- not sure if that's worthy doing here or just marking as a TODO.

abadams:
More concretely, I worry about the effects on compile time of injecting additional calls to the simplifier here. It is going to bite us at some point. E.g. if you have a deeply nested chain of O(n) bitwise math, this calls the simplifier recursively at every node, doing O(n^2) work.

steven-johnson:
Fair. I'll leave this open until I (or someone) gets around to improving it (this was just an opportunistic fix rather than a fix in response to a known issue).

steven-johnson:
Discarding in favor of #3607

PR #3607, opened 2019-01-23
Bounds of bitwise ops

abadams:


abadams:
This PR removes fishy use of the simplifier inside bounds logic (#3604)

src/Bounds.cpp
```diff
@@ -1011,14 +1011,92 @@ class Bounds : public IRVisitor {
...
+                    if (a_interval.has_lower_bound() && b_interval.has_upper_bound()) {
+                        interval.min = a_interval.min >> b_interval.max;
+                    }
+                    if (a_interval.has_upper_bound() && b_interval.has_lower_bound()) {
+                        interval.max = a_interval.max >> b_interval.min;
+                    }
+                }
+            }
+        } else if (op->is_intrinsic(Call::bitwise_and)) {
+            Expr a = op->args[0], b = op->args[1];
```
steven-johnson: There's a fair amount of DRY here, but restructuring might be awkward.

PR #3611, opened 2019-01-23
Check types as uint32_t

abadams:
If we check type codes as uint32s, we never need to unpack the struct fields. This shrinks generated IR slightly by removing two letstmts per input, which helps out llvm. It's also theoretically a little more efficient at runtime, but the effect is too small to measure.

Builds on #3610 for a test

Without this change:

Compile time with 1 inputs = 0.941949 s 
Compile time with 2 inputs = 0.103978 s 
Compile time with 4 inputs = 0.167523 s 
Compile time with 8 inputs = 0.295939 s 
Compile time with 16 inputs = 0.661234 s 
Compile time with 32 inputs = 1.677875 s 
Compile time with 64 inputs = 5.087148 s 
Compile time with 128 inputs = 17.967057 s 

With this change:

Compile time with 1 inputs = 0.919779 s 
Compile time with 2 inputs = 0.102316 s 
Compile time with 4 inputs = 0.161664 s 
Compile time with 8 inputs = 0.284712 s 
Compile time with 16 inputs = 0.596491 s 
Compile time with 32 inputs = 1.425520 s 
Compile time with 64 inputs = 4.242031 s 
Compile time with 128 inputs = 14.277232 s


abadams:
Only failure was just fixed by a different PR.

PR #3612, opened 2019-01-23
Bound CSE calls for function inlining 

asaadaldien:
Only run CSE if there is more than one call to current inlined function. 

This improves compile time for pipelines with large number of inputs

make run_performance_lots_of_inputs

```
Before: 3c84ccb
Compile time with 1 inputs = 0.490005 s
Compile time with 2 inputs = 0.074211 s
Compile time with 4 inputs = 0.121533 s
Compile time with 8 inputs = 0.205918 s
Compile time with 16 inputs = 0.445551 s
Compile time with 32 inputs = 1.155209 s
Compile time with 64 inputs = 3.547876 s
Compile time with 128 inputs = 13.205774 s

After : 3c84ccb 

Compile time with 1 inputs = 0.489706 s
Compile time with 2 inputs = 0.075291 s
Compile time with 4 inputs = 0.115259 s
Compile time with 8 inputs = 0.214029 s
Compile time with 16 inputs = 0.458343 s
Compile time with 32 inputs = 1.102211 s
Compile time with 64 inputs = 3.282386 s
Compile time with 128 inputs = 11.546451 s
```


src/Inline.cpp
```diff
@@ -151,12 +151,12 @@ class Inliner : public IRMutator {
     }
 
     Stmt visit(const Provide *op) override {
-        bool old_found = found;
+        int old_found = found;
```
steven-johnson: ScopedValue old_found(found, 0);

abadams:
Passed all relevant tests before I merged master into it, so just merging this to avoid wasting buildbot time.

PR #3614, opened 2019-01-24
clone #3565 for buildbots

abadams:


steven-johnson:
Is the underlying PR ready to land? This appears clean.

abadams:
Yes, though I was waiting to hear back from Zalman on whether or not the locking seemed kosher

steven-johnson:
ok, let's definitely still wait then

steven-johnson:
I presume this can be killed now

Issue #3616, opened 2019-01-24
Calling round() on an int16 Expr leads to verifyModule failure in LLVM


dsharletg:
I observed this when I accidentally called Halide's `round` function on an Expr with type int16, while vectorized:

Call parameter type does not match function signature!
  %665 = sitofp <8 x i16> %664 to <8 x half>
 <8 x float>  %a70 = call reassoc nnan ninf nsz contract afn <8 x float> @round_f32x8(<8 x half> %665) #6
Both operands to FCmp instruction are not of the same type!
  %666 = fcmp reassoc nnan ninf nsz contract afn ogt <8 x half> <half 0xH7800, half 0xH7800, half 0xH7800, half 0xH7800, half 0xH7800, half 0xH7800, half 0xH7800, half 0xH7800>, <8 x float> %a70
Both operands to FCmp instruction are not of the same type!
  %667 = fcmp reassoc nnan ninf nsz contract afn ogt <8 x float> %a70, <8 x half> <half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800>
Invalid operands for select instruction!
  %668 = select <8 x i1> %667, <8 x float> %a70, <8 x half> <half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800, half 0xHF800>
Call parameter type does not match function signature!

abadams:
Ah, I see the bug

dsharletg:
Fixed by the above referenced PR.

PR #3617, opened 2019-01-24
Types of 32-bit float intrinsics must be a 32-bit float

abadams:
#3616


zvookin:
Why are we allowing these calls on integer types at all? And if we are going to, shouldn't the integer version just do the cast, not call the intrinsic?

dsharletg:
I'd prefer to allow this casting to happen. It makes writing generic helper functions easier. I often run into an annoying issue now where fast_inverse asserts the argument is a float. It seems reasonable to implicitly convert an integer to a float in this situation. It seems like if someone really is making a mistake, the result will be float where they didn't want that anyways.

steven-johnson:
What's the status on this PR?

PR #3623, opened 2019-01-25
Enable some 256-bit integer ops for AVX

abadams:
They don't exist for AVX, but llvm legalizes them to instructions that do exist. If we try to legalize it ourselves to the instructions that do exist instead, then llvm behaves poorly.


src/CodeGen_X86.cpp
```diff
@@ -291,19 +291,19 @@ void CodeGen_X86::visit(const Cast *op) {
          u16(max(wild_i32x_ - wild_i32x_, 0))},
 #else
         // LLVM 8.0+ require using helpers from x86.ll
-        {Target::AVX2, true, UInt(8, 32), 17, "paddusbx32",
+        {Target::AVX, true, UInt(8, 32), 17, "paddusbx32",
```
steven-johnson: Do we need to add comments here about what LLVM8+ is doing under the hood?

src/runtime/x86_avx.ll
```diff
@@ -1,3 +1,56 @@
+; Note that this is only used for LLVM 8.0+
```
steven-johnson: Do we need to add comments here about what LLVM8+ is doing under the hood?

abadams:
All tests passed before I added the comments, then the buildbots decided not to rerun, and travis is doing something random. Merging.

PR #3625, opened 2019-01-25
Revert "Bounds of bitwise ops"

steven-johnson:
This is broken, I shouldn't have approved it:
- shift_right went missing entirely
- bitwise_xor is treated as shift_left
- shift_left has a special case intended for shift_right

Rather than trying to spot-fix, let's back it out entirely and fix it properly (with tests) later.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1011,92 +1011,14 @@ private:
             assert(op->args.size() == 3);
             op->args[1].accept(this);
         } else if (op->is_intrinsic(Call::shift_left) ||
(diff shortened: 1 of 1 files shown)


abadams:
Yikes, sorry. I must have pushed it in a half-done state.

abadams:
Ah, I see what I did. This is epsilon away from making sense due to a late but incorrect DRY-simplification I made. Will fix and reopen.

steven-johnson:
It gave some *very* interesting failures when I tried it out in google code :-)

PR #3626, opened 2019-01-26
Dump Hexagon shared objects in debug builds

dsharletg:
When the debug target flag is set, write the shared objects loaded by Hexagon to the current directory.

Not sure I really like this PR, but it is useful for some Hexagon profiling/debugging tools, and this is a lower impact approach than past PRs aimed at accomplishing this task.

### DIFF
--- a/src/runtime/hexagon_host.cpp
+++ b/src/runtime/hexagon_host.cpp
@@ -182,7 +182,6 @@ WEAK void write_shared_object(void *user_context, const char* path,
     size_t written = fwrite(code, 1, code_size, f);
     if (written != code_size) {
         debug(user_context) << "    bad write of shared object to '" << path << "'\n";
-        return;
     }
     fclose(f);
 }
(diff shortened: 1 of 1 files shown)


steven-johnson:
Have you tried this out inside Google? It might make it painful to build debug runtimes (even experimentally) if Blaze decides to complain about unexpected files being written.

src/runtime/hexagon_host.cpp
```diff
@@ -168,6 +168,27 @@ struct module_state {
...
+WEAK void write_shared_object(void *user_context, const char* path,
+                              const uint8_t* code, uint64_t code_size) {
+    void *f = fopen(path, "wb");
+    if (!f) {
+        debug(user_context) << "    failed to write shared object to '" << path << "'\n";
+        return;
+    }
+    size_t written = fwrite(code, 1, code_size, f);
+    if (written != code_size) {
+        debug(user_context) << "    bad write of shared object to '" << path << "'\n";
```
steven-johnson: fclose(f) here too, or just elide the return and fall thru.

steven-johnson:
> this dumps at runtime

ahhh, I missed that. LGTM.

pranavb-ca:
Thanks, this is very useful!

PR #3628, opened 2019-01-26
Fix bounds of bitwise ops

abadams:


steven-johnson:
LGTM; I'll patch this in to my tests and see how it looks.

steven-johnson:
I'm now getting other failures in google but I think they are likely unrelated to this one. LGTM.

steven-johnson:
So I think this may (still) be causing some new 'unbounded access' issues -- I had patched it in and run the usual massive set of tests without seeing anything, but I must have missed something, because today's attempt at a Halide merge is producing these sorts of issues again, and reverting just this set of changes 'fixes' it. (I'll dive in further and see what I can figure out before proposing to revert this, again...)

PR #3629, opened 2019-01-26
Standalone autoscheduler

benoitsteiner:


steven-johnson:
IMHO It would be even more user-friendly if you made all these changes with a constant, eg

    constexpr int kDebugLevelHereName = 1;

So that subsequent experiments could make it more or less verbose at a single edit point.

PR #3630, opened 2019-01-26
Remove overzealous check when extracting even/odd lanes of a vector

dsharletg:
This check was causing some shuffles to go through an unnecessarily slow path. The check shouldn't be necessary.


steven-johnson:
Is this ready to land (pending buildbots)?

dsharletg:
I'm syncing to master to re-run the build bots with the simd_op_check fixes. I'm pretty sure the failures were unrelated (I ran the Hexagon tests locally) but just in case.

Other than that, it needs approval.

PR #3632, opened 2019-01-27
Remove dead lets more thoroughly. Fixes #3631

### LINKED ISSUE #3631: Simplifier not simplifying dead lets
I have an expression that is not getting simplified as I expect:

expr:

    (let t12073 = ((t11933/2) + ((gained.s0.x.x + yuv.s0.x.min_1.s)*64)) in (((t12073 + -3) + (63*1)) - ((t12073 + -3) + (0*1))))

simplify(expr):

    (let t12073 = ((t11933/2) + ((gained.s0.x.x + yuv.s0.x.min_1.s)*64)) in 63)

I think the issue is that here, var_info.old_uses is greater than 0, but the var is not actually used anymore. If I make this change, I get the simplification that I expect (the expression is simply 63):

    @@ -200,7 +201,7 @@ Body Simplify::simplify_let(const LetOrLetStmt *op, ExprInfo *bounds) {
             result = LetOrLetStmt::make(it->new_name, it->new_value, result);
         }
 
    -    if (info.old_uses > 0 || !remove_dead_lets) {
    +    if ((info.old_uses > 0 && stmt_or_expr_uses_var(result, it->op->name)) || !remove_dead_lets) {
             // The old name is still in use. We'd better keep it as well.
             result = LetOrLetStmt::make(it->op->name, it->value, result);
         }

(Keeping the old check seems to avoid serious compile time issues).

However, I can't seem to find a failing expression to put in test/correctness/simplify.cpp! Reimplementing the expression I gave is simplifying correctly in the context of that test.

dsharletg:


src/Simplify_Let.cpp
```diff
@@ -184,6 +207,9 @@ Body Simplify::simplify_let(const LetOrLetStmt *op, ExprInfo *bounds) {
 
     result = mutate_let_body(result, bounds);
 
+    std::map<std::string, int> vars_used;
+    count_var_uses(result, vars_used);
```
abadams: This seems like a reasonable algorithm, but don't we now have two independent trackers for var usage? I think info.new_uses and info.old_uses were only used for this purpose. They were inexact, because they tried to run at the same time as the rest of simplification, whereas here you use a post-pass in a separate visitor.

src/Simplify_Let.cpp
```diff
@@ -195,14 +221,28 @@ Body Simplify::simplify_let(const LetOrLetStmt *op, ExprInfo *bounds) {
         VarInfo info = var_info.get(it->op->name);
         var_info.pop(it->op->name);
 
-        if (it->new_value.defined() && info.new_uses > 0) {
```
abadams: You need to count loads or stores to "foo" as a usage of the name "foo" in the visitor above. That will make the hack below unnecessary.

abadams:
Please check performance_lots_of_inputs before and after this change to test some long let and letstmt chains.

src/Simplify_Let.cpp
abadams: (i.e. try deleting new_uses and old_uses entirely if you can)

dsharletg: I thought so too, but I saw some issues when I removed the old logic. I'll try again now that you pointed out Loads/Stores can also use a Let.

dsharletg: It seems like the info.new_uses/info.old_uses are protecting against some redundant lets getting generated (which trigger the later 'lets must be unique' assert), from *different* `Frame` struct instances.

steven-johnson:
Is this ready to land (pending buildbots)?

dsharletg:
I think it's ready. As for the build bots, I think the only failures are unrelated.

steven-johnson:
I'll wait for @abadams to approve it, since he did all the reviewing.

abadams:
The windows failure was not one I could explain easily, so I'd like to wait for it to rerun.

dsharletg:
It looks like windows failed again, and it doesn't seem to be failing on other branches. But I can't tell what's going wrong in the log?

dsharletg:
This did finally pass the windows build bot, please take a look at the code. This addresses a performance issue that I'd like to get fixed.

PR #3634, opened 2019-01-28
Restructure LLVM ifdefs

steven-johnson:
Minor, but: Restructure so that all ifdefs are of the form

    #ifdef LLVM_VERSION >= someversion

(vs < someversion); this makes scanning thru code to look for LLVM dependencies less error prone. (Obviously the intent is to standardize on this for future changes as well.)

This does mean that we have a couple if awkward "nothing" clauses, but IMHO it's a minor nit to make version-checking more uniform and reliable.


abadams:
Looks correct, but I probably just created a merge conflict in CodeGen_X86

PR #3635, opened 2019-01-28
Revert "Bound CSE calls for function inlining "

steven-johnson:
Reverts halide/Halide#3612

For reasons that aren't yet clear, this change causes some pipelines inside Google to explode in compiletime (so much so that they timeout and never complete). Would like to revert and re-examine the change in more detail before re-landing.

### DIFF
--- a/src/Inline.cpp
+++ b/src/Inline.cpp
@@ -116,7 +116,7 @@ class Inliner : public IRMutator {
                 body = Let::make(func.name() + "." + func_args[i], args[i], body);
             }
 
-            found++;
+            found = true;
 
             return body;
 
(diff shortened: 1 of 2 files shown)


abadams:
The partition loops change was a straight-up bug. We were producing invalid code in some circumstances, so I'd prefer a softer roll-back. I'll mark the line to change to do the minimal rollback.

src/Inline.cpp
```diff
@@ -151,20 +151,23 @@ class Inliner : public IRMutator {
...
 
     Stmt visit(const Provide *op) override {
-        ScopedValue<int> old_found(found, 0);
+        bool old_found = found;
+
+        found = false;
         Stmt stmt = IRMutator::visit(op);
 
-        if (found > 1) {
+        if (found) {
```
abadams: Try reverting this line alone

src/Inline.cpp
```diff
@@ -180,7 +183,7 @@ Stmt inline_function(Stmt s, Function f) {
 Expr inline_function(Expr e, Function f) {
     Inliner i(f);
     e = i.mutate(e);
-    if (i.found > 1) {
```
abadams: and this one

steven-johnson:
...those two changes are essentially this entire commit. What is this "partition loops" change to which you refer?

abadams:
Looking at files changed, there's a lot of stuff inside PartitionLoops.cpp

steven-johnson:
Huh, that's weird, I thought I had checked and didn't see any other changes (other than a test). Yeah, the changes to inlining are all that's necessary here. I'll drop this and propose a more targeted fix.

PR #3636, opened 2019-01-28
Don't create unnecessary lets in Inlining

abadams:
Possible alternative to #3635

PR #3637, opened 2019-01-28
Partial revert of #3612

steven-johnson:
For reasons that aren't yet clear, this change causes some pipelines inside Google to explode in compiletime (so much so that they timeout and never complete). Would like to revert and re-examine the change in more detail before re-landing.

### DIFF
--- a/src/Inline.cpp
+++ b/src/Inline.cpp
@@ -154,7 +154,9 @@ class Inliner : public IRMutator {
         ScopedValue<int> old_found(found, 0);
         Stmt stmt = IRMutator::visit(op);
 
-        if (found > 1) {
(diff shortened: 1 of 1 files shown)


abadams:
Please check if #3636 solves the problem too first. Otherwise there's some more subtle reason why CSE is necessary even when there's only one call to the func being inlined.

steven-johnson:
(I have narrowed down a somewhat-simpler repro case and communicated it to @abadams privately.)

steven-johnson:
Opened https://github.com/halide/Halide/issues/3639 with a reproducible case so we can investigate and fix the pathology, if possible

PR #3088, opened 2018-06-26
Add support for bounds inference using shifts.

schoenberg-google:
This changelist adds support for performing bounds inference from shifted values. As an example, consider a Generator that produces an image that has been downsampled by a power of two. It is natural to express this downsampling ratio in log-space, because otherwise invalid ratios can be expressed.

Resolves #199, as described by the comments in that issue.

PR #3642, opened 2019-01-29
Bounds inference for shift_left of int32s

abadams:


dsharletg:
See also: #3088

abadams:
Also pushed support for narrower integer types

src/Bounds.cpp
```diff
@@ -1039,13 +1053,25 @@ class Bounds : public IRVisitor {
...
+                            interval.min = make_zero(t);
+                        } else if (t.is_int()) {
+                            if (b_positive) {
+                                interval.min = make_zero(t);
+                                interval.max = b_interval.max;
+                            } else if (a_positive) {
+                                interval.min = make_zero(t);
+                                interval.max = a_interval.max;
+                            } else {
+                                // Smaller than the larger of the two args
```
steven-johnson: So in this case, the min remains unbounded?

abadams: Yeah, if neither arg is known to be positive then you can't say anything about the sign bit. All the others bits could cancel out and leave you with 0x80000000

steven-johnson:
Are we ready to land this?

steven-johnson:
So there is still a subtle bug here (and possibly in some pre-existing code too); namely that the `is_single_point` optimization cases need to always use the values from the intervals themselves (*not* the original Exprs) in case there is a scope-lookup substitution.

I don't have a repro case from source code (yet), but IR that looks like this will tickle it:

```
(let t0 = int32(ceil_f32(log_f32(float32(radius + 1)))) in select((radius == 0), input(row, col), min(some_func(row, (col - radius), t0), some_func(row, (((col + radius) - max(min(shift_left(1, t0), (radius*2)), 0)) + 1), t0))))
```

The value of `t0` goes thru `is_small_enough_to_substitute`, but when we get down to the `shift_left` handler, it uses the unsafe `is_single_point` route mentioned above, meaning that you end up with an Expr that uses `t0`, but no corresponding Let-expr (since we thought we just substituted directly).

I'll make a secondary PR that branches off this one to propose a fix.

steven-johnson:
Update: the fix was actually much more localized and safe than I expected, so I pushed it directly to this PR.

PR #3643, opened 2019-01-29
Improve bounds detection for (a << b)

steven-johnson:
For the case where b's interval is a single point (e.g. a constant), we formerly simplified, allowing us to decompose into a multiply and sometimes do better than bounds-of-type; this restores that behavior, as we have downstream users relying on it.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1029,7 +1029,14 @@ private:
                 // For some of these intrinsics applied to integer
                 // types we can go a little further.
                 if (t.is_int() || t.is_uint()) {
-                    if (op->is_intrinsic(Call::shift_right)) {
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -1029,7 +1029,14 @@ class Bounds : public IRVisitor {
                 // For some of these intrinsics applied to integer
                 // types we can go a little further.
                 if (t.is_int() || t.is_uint()) {
-                    if (op->is_intrinsic(Call::shift_right)) {
```
abadams: The whole point of the change was to remove the use of the simplifier here. Does the other PR (#3642) solve your problem? If not we should canonicalize without the simplifier to avoid a recursive descent into the LHS.

steven-johnson:
Withdrawn in favor of #3642

PR #3644, opened 2019-01-29
Fix falsely promised alignment

dsharletg:
@steven-johnson @abadams

### DIFF
--- a/src/HexagonOptimize.cpp
+++ b/src/HexagonOptimize.cpp
@@ -1771,7 +1771,10 @@ class OptimizeShuffles : public IRMutator {
                 (unaligned_index_bounds.min / align) * align,
                 ((unaligned_index_bounds.max + align) / align) * align - 1
             };
-            ModulusRemainder alignment(align, 0);
+            // TODO: This assumes the host pointer of this buffer is aligned,
+            // which we can't do.
+            //ModulusRemainder alignment(align, 0);
+            ModulusRemainder alignment(1, 0);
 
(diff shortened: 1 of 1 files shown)


dsharletg:
It's possible the better fix is to just leave this and instead work on making codegen understand when the base address of a buffer is aligned or not. As far as I can tell, most usages of load/store alignment only refer to the alignment of the index, and not the overall alignment (buffer + index). So this fix might be just one of several cases where this matters (if not *all* uses of load/store alignments).

steven-johnson:
> the better fix is to just leave this and instead work on making codegen understand when the base address of a buffer is aligned or not

That definitely sounds better; my assumption (given the TODO) was that this was merely an expediency to avoid breaking existing code.

dsharletg:
I want to understand the problem better because CodeGen is already doing the right thing. I think it's the code itself that is the problem (passing an unaligned pointer when it said it would be aligned). It sounds like the better fix is what we should do. I'm going to close this.

pranavb-ca:
We look at the base alignment for external buffers in Codegen. Unfortunately, we don't seem to be doing the same for internal buffers.

abadams:
internal buffers are supposed to be allocated with maximum alignment, aren't they?

dsharletg:
I think we can assume all internal buffers are aligned. The buffer that's causing problems in this case is an external buffer anyways.

There's something else going on here that I'm trying to debug. It's definitely still related to alignment of these boundary condition loads though.

PR #3647, opened 2019-01-30
Simplify shift_right and shift_left for out-of-range shifts too

steven-johnson:


steven-johnson:
Hm, this is arguably not quite right: with this change, `u8(3) << u8(7)` yields `u8(0)` due to the overflow check, even though the infinite-width result (0x180) wouldn't have all of its bits shifted away.

steven-johnson:
See https://github.com/halide/Halide/pull/3649 for fix

PR #3648, opened 2019-01-31
Fix unsafe usage of certain is_single_point() constructs in Bounds

steven-johnson:
The various `is_single_point` optimizations in Bounds need to ensure to only use values from the intervals themselves (not the original Exprs), otherwise scope-substitution errors can occur, resulting in missing Lets.

The actual error was found in an Expr using shift_left, but inspection would seem to reveal that many other operators could potentially have this issue (see commented-out code bits).

Need to add a testcase, but we don't have one yet (the failure was found in some complex private code and we haven't reconstructed a simple one). Also, this fix injects at least one failure into existing test suite (compute_at_split_rvar) that needs investigation.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -191,10 +191,12 @@ private:
         op->value.accept(this);
         Interval a = interval;
 
-        if (a.is_single_point(op->value)) {
-            interval = Interval::single_point(op);
-            return;
-        }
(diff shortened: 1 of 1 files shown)


steven-johnson:
Needlessly aggressive, will offer a simpler one shortly

PR #3649, opened 2019-01-31
Fix PR#3647 for unsigned ints

steven-johnson:
Maximum shift amount for unsigned should have been one larger.


dsharletg:
This looks good for now, but I think even shifts of t.bits() are undefined on most platforms. For example, on ARM I think the legal range of shifts is [0, 2^bits - 1]. So maybe we should just be using t.bits() - 1 always, not just for signed integers...

steven-johnson:
Yep, you are right. Let me close this for now; the issue I wanted to address with it isn't essential.

steven-johnson:
Reopening to add some comments:
- This PR shouldn't actually invoke any undefined shifts; the confusingly-named `max_shift` is actually one-more-than-the-max-shift, and the clause at line 45 only converts into mul/div if the constant shift amount is **less than** that. If it's >= that max-shift, then we reduce to 0 or -1 as seems appropriate. (Did I get a boundary condition wrong?)
- Is your concern that the UB could produce differing results between constant and non-constant shift amounts? e.g. with this change, `u8 << 13` would always produce 0, but `u8 << ParamWithValue13` could vary result depending on platform? That's a valid concern, but it seems like it's one that we already had in place since forever. (And strictly speaking, this change should only reduce the net possible UB by adding well-defined for a subset of constant expressions.)

src/Simplify_Call.cpp
```diff
@@ -40,7 +40,9 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             }
 
             if (ib >= 0) {
-                if (ib < std::min(t.bits(), 64) - 1) {
+                int max_shift = std::min(t.bits(), 64);
+                if (t.is_int()) max_shift -= 1;
```
zvookin: Brace the if or do ```max_shift -= t.is_int(); // One less if there is a sign bit``` I'd also rename ```max_shift``` because it is not actually allowed. E.g. ```effective_type_bits``` or ```shift_too_large_bits```.

src/Simplify_Call.cpp
```diff
@@ -40,7 +40,9 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             }
 
             if (ib >= 0) {
-                if (ib < std::min(t.bits(), 64) - 1) {
+                int max_shift = std::min(t.bits(), 64);
```
zvookin: What is this line doing? It seems it should either be an assert or not limit the size. (128-bit integers are a thing on the cutting edge of computer architecture.) I assume the context has already ensured that t is a scalar, not a vector type.

steven-johnson: Looks to me like it's verifying that the 64-bit constant we produce (to hand off to Mul or Div) won't overflow an int64. That said, now that you point this out, the logic I'm adding would in fact overflow if the type is uint64. Maybe this whole clause would be clearer if split into completely separate int-vs-uint sections.

steven-johnson:
Added tests for 64-bit ints and fixed a couple of corner cases. PTAL.

src/Simplify_Call.cpp
```diff
@@ -50,15 +64,18 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
+                    // just replace with a constant 0 or -1 if possible.
                     if (shift_left) {
                         return make_zero(t);
                     } else {
                         // shift-right-out-of-bounds -> zero for uint.
                         if (t.is_uint()) {
                             return make_zero(t);
+                        } else {
+                            // shift-right-out-of-bounds -> zero or -1 for int.
+                            return mutate(select(a < 0, make_const(t, -1), make_zero(t)), bounds);
```
dsharletg: It seems like leaving this as a shift is better. It's the same number of the same kind of instructions, but doesn't need the constant registers.

steven-johnson: The motivation for this construct is to constant-fold expressions that can result from the recent changes to Bounds (e.g. 32767 << -32768); if we left it as a shift, we'd end up with `shift_right(32767, 32768)` in our Expr tree post-simplify.

dsharletg: So we know a is a constant? If so, how about a compile time if, instead of a runtime if?

dsharletg: Although I see now this code is pre-existing, so maybe my comment isn't relevant in this PR.

steven-johnson: That code is pre-existing by only a few days. 

I suppose we could structure it to only do that if `a` is a constant, or to do something like

   if (can_prove(a < 0)) // -1
   else if (can_prove(a >= 0)) // 0
   else // use shift

steven-johnson: (oh, and `a` is **not** known to be a constant.)

src/Simplify_Call.cpp
steven-johnson: Done.

PR #3654, opened 2019-02-01
Add accessors to RunGen to retrieve the call + metadata

steven-johnson:


steven-johnson:
Yeah, RunGen is sorely undertested (it's mostly "does it build"). Let me add something.

PR #3656, opened 2019-02-04
Extend registration mechanism to be extensible

steven-johnson:
This extends the registration mechanism for AOT-Generators to allow for a build system to add arbitrary additional text as part of the `halide_register_argv_and_metadata` call.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1526,10 +1526,10 @@ $(FILTERS_DIR)/registration_test: $(ROOT_DIR)/test/generator/registration_test.c
 			$(FILTERS_DIR)/blur2x2.registration_extra.o \
 			$(FILTERS_DIR)/cxx_mangling.registration_extra.o \
 			$(FILTERS_DIR)/pyramid.registration_extra.o \
-			$(BIN_DIR)/$(TARGET)/runtime.a \
 			$(FILTERS_DIR)/blur2x2.a \
 			$(FILTERS_DIR)/cxx_mangling.a \
 			$(FILTERS_DIR)/pyramid.a \
+      $(BIN_DIR)/$(TARGET)/runtime.a \
 			$(GEN_AOT_LD_FLAGS) $(IMAGE_IO_LIBS) -o $@
 
 # Test RunGen itself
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1064,12 +1064,14 @@ GENERATOR_BUILD_RUNGEN_TESTS := $(filter-out $(FILTERS_DIR)/nested_externs.runge
 GENERATOR_BUILD_RUNGEN_TESTS := $(filter-out $(FILTERS_DIR)/old_buffer_t.rungen,$(GENERATOR_BUILD_RUNGEN_TESTS))
 GENERATOR_BUILD_RUNGEN_TESTS := $(filter-out $(FILTERS_DIR)/tiled_blur.rungen,$(GENERATOR_BUILD_RUNGEN_TESTS))
 GENERATOR_BUILD_RUNGEN_TESTS := $(filter-out $(FILTERS_DIR)/extern_output.rungen,$(GENERATOR_BUILD_RUNGEN_TESTS))
-GENERATOR_BUILD_RUNGEN_TESTS := $(GENERATOR_BUILD_RUNGEN_TESTS) $(FILTERS_DIR)/multi_rungen $(FILTERS_DIR)/multi_rungen2 $(FILTERS_DIR)/rungen_test
+GENERATOR_BUILD_RUNGEN_TESTS := $(GENERATOR_BUILD_RUNGEN_TESTS) $(FILTERS_DIR)/multi_rungen $(FILTERS_DIR)/multi_rungen2 $(FILTERS_DIR)/rungen_test $(FILTERS_DIR)/registration_test
```
zvookin: Probably time to break this into multiple lines.

src/Module.cpp
```diff
@@ -150,11 +152,19 @@ extern int $SHORTNAME$_argv(void **args);
 extern const struct halide_filter_metadata_t *$SHORTNAME$_metadata();
 $NSCLOSE$
 
+#ifdef HL_REGISTER_EXTRA_KEY_VALUE_PAIRS_FUNC
```
zvookin: I don't see any reason to abbreviate "HALIDE" to "HL" here.

src/Module.cpp
```diff
@@ -150,11 +152,19 @@ extern int $SHORTNAME$_argv(void **args);
...
+extern "C" const char * const *HL_REGISTER_EXTRA_KEY_VALUE_PAIRS_FUNC();
+#endif  // HL_REGISTER_EXTRA_KEY_VALUE_PAIRS
```
zvookin: The closing comment does not match the ifdef (..._FUNC).

PR #3659, opened 2019-02-06
Bounds for shift_right with signed-ints was incorrect

steven-johnson:
The boundaries didn't take into account the differing behavior of signed values, so the constant bounds signed ints was wildly wrong in some cases (eg int16 >> int16 -> [-1, 0])

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1055,11 +1055,31 @@ private:
                             equiv.accept(this);
                         }
                     } else if (op->is_intrinsic(Call::shift_right)) {
-                        if (a_interval.has_lower_bound() && b_interval.has_upper_bound()) {
-                            interval.min = a_interval.min >> b_interval.max;
-                        }
-                        if (a_interval.has_upper_bound() && b_interval.has_lower_bound()) {
-                            interval.max = a_interval.max >> b_interval.min;
+                        // shift_right(a, b) is UB for b < 0, so only try to improve on bounds-of-type
+                        // if we can prove b >= 0.
+                        if (b_interval.is_bounded() && (t.is_uint() || can_prove(b_interval.min >= 0))) {
+                            if (a_interval.has_lower_bound()) {
+                                if (t.is_uint() || can_prove(a_interval.min >= 0)) {
+                                    interval.min = a_interval.min >> b_interval.max;
+                                } else {
+                                    // if a < 0, the smallest value will be a >> b.min
+                                    // if a > 0, the smallest value will be a >> b.max
+                                    interval.min = a_interval.min >> select(a_interval.min < 0,
+                                                                            b_interval.min,
+                                                                            b_interval.max);
+                                }
+                            }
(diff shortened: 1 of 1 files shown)


steven-johnson:
Gentle review ping

src/Bounds.cpp
```diff
@@ -1055,11 +1055,31 @@ class Bounds : public IRVisitor {
...
+                        // shift_right(a, b) is UB for b < 0, so only try to improve on bounds-of-type
+                        // if we can prove b >= 0.
+                        if (b_interval.is_bounded() && (t.is_uint() || can_prove(b_interval.min >= 0))) {
+                            if (a_interval.has_lower_bound()) {
+                                if (t.is_uint() || can_prove(a_interval.min >= 0)) {
+                                    interval.min = a_interval.min >> b_interval.max;
+                                } else {
+                                    // if a < 0, the smallest value will be a >> b.min
+                                    // if a > 0, the smallest value will be a >> b.max
+                                    interval.min = a_interval.min >> select(a_interval.min < 0,
```
abadams: This will probably be accepted more cleanly by other analysis passes as:

``` min(a_interval.min >> b_interval.min, a_interval.min >> b_interval.max)```

File: src/Bounds.cpp [hunk 2]
```diff
@@ -1055,11 +1055,31 @@ class Bounds : public IRVisitor {
...
+                                                                            b_interval.max);
+                                }
+                            }
+                            if (a_interval.has_upper_bound()) {
+                                if (t.is_uint() || can_prove(a_interval.max >= 0)) {
+                                    interval.max = a_interval.max >> b_interval.min;
+                                } else {
+                                    // if a < 0, the largest value will be a >> b.max
+                                    // if a > 0, the largest value will be a >> b.min
+                                    interval.max = a_interval.max >> select(a_interval.max < 0,
```
abadams: ditto

src/Bounds.cpp
steven-johnson: min() is always better than select()

steven-johnson:
I think this is finally clean; PTAL

src/Bounds.cpp
```diff
@@ -1055,11 +1055,27 @@ class Bounds : public IRVisitor {
...
+                                    interval.min = min(a_interval.min >> b_interval.min, a_interval.min >> b_interval.max);
+                                }
+                            }
+                            if (a_interval.has_upper_bound()) {
+                                if (t.is_uint() || can_prove(a_interval.max >= 0)) {
+                                    interval.max = a_interval.max >> b_interval.min;
+                                } else {
+                                    // if a < 0, the largest value will be a >> b.max
+                                    // if a > 0, the largest value will be a >> b.min
+                                    interval.max = min(a_interval.max >> b_interval.max, a_interval.max >> b_interval.min);
```
abadams: I think the RHS here should be a max, not a min.

steven-johnson:
PTAL

Issue #3325, opened 2018-10-08
Define shift by a negative value as a shift in the other direction


dsharletg:
Some architectures support "signed shifts", where `x << -N` is equivalent to `x >> N` and vice-versa. Currently in Halide, shifting by a negative value is undefined behavior. We could define it this way, and then emulate this on targets that do not support this in hardware. This could be made zero cost by only emulating this for signed shift RHS values. 

Currently, shifts require that the LHS and RHS have the same type, including signedness. A prerequisite to this strategy would be to allow the signedness of the operands of a shift to differ. Both signed shifts of unsigned values, as well as being able to shift signed values by unsigned shift amounts would become important.

zvookin:
To clarify, the design would allow mixing the types of operands for shift operators and do a bounds proof on the shift amount. If if the proof comes back known non-negative or non-positive the shift operator can be determined statically at compile time, otherwise a conditional structure would be used. (Some research into the conditional structure that is best is likely warranted. My reading of the LLVM IR docs is one cannot presume LLVM will generate a machine shift instruction on machines which provide the semantics we desire. Though to be honest, I'm not sure I know of such a machine.)

The situation now has this being undefined behavior so I'm happy to define it this way. (I.e. one can basically never use a negative shift amount anyway so if it occurs in any program today, it's a bug.) The only place this might be a bit of an issue is code can be optimized now on the assumption that the negative values don't happen.

zvookin:
One more thing: ```unsafe_promise_clamped``` would get one the current behavior fairly trivially.

steven-johnson:
The current behavior doesn't satisfy anything: the logic for simplifying both shift-left and shift-right assumes that negative shift values are possible, and should be treated as mirroring the operation (e.g. a << -b == a >> b), but we don't seem to actually enforce this at codegen time.

I'd be fine with us ensuring that negative shifts are allowed everywhere, but until/unless that happens, we should probably stick to the lowest-common-denominator of assuming that a shift amount of `0 <= shift-amount < bits-in-type` is legal, and have the simplifier follow that rule. (And also probably add logic to fail at compile time if we can prove that the shift amount is outside that range.)

PR #3660, opened 2019-02-06
Revise Simplify for << and >> to forbid out-of-range constant shifts

steven-johnson:
Formerly we tried to handle negative shifts by flipping the operation, and out-of-range shifts by clipping to zero or -1. This isn't necessarily *wrong* -- these are UB in C++ (and poison-producing in LLVM) -- but I'd argue that if we can statically detect UB/poison values, we're better off producing a hard error at compile time.

See also https://github.com/halide/Halide/issues/3325

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1058,20 +1058,22 @@ private:
                         // shift_right(a, b) is UB for b < 0, so only try to improve on bounds-of-type
                         // if we can prove b >= 0.
                         if (b_interval.is_bounded() && (t.is_uint() || can_prove(b_interval.min >= 0))) {
-                            if (a_interval.has_lower_bound()) {
+                            bool b_min_ok = (t.is_uint() || can_prove(b_interval.min >= 0)) && can_prove(b_interval.min < t.bits());
+                            bool b_max_ok = (t.is_uint() || can_prove(b_interval.max >= 0)) && can_prove(b_interval.max < t.bits());
+                            if (a_interval.has_lower_bound() && b_max_ok) {
                                 if (t.is_uint() || can_prove(a_interval.min >= 0)) {
                                     interval.min = a_interval.min >> b_interval.max;
-                                } else {
+                                } else if (b_min_ok) {
                                     // if a < 0, the smallest value will be a >> b.min
                                     // if a > 0, the smallest value will be a >> b.max
                                     interval.min = min(a_interval.min >> b_interval.min,
                                                        a_interval.min >> b_interval.max);
                                 }
                             }
-                            if (a_interval.has_upper_bound()) {
+                            if (a_interval.has_upper_bound() && b_min_ok) {
                                 if (t.is_uint() || can_prove(a_interval.max >= 0)) {
                                     interval.max = a_interval.max >> b_interval.min;
-                                } else {
+                                } else if (b_max_ok) {
                                     // if a < 0, the largest value will be a >> b.max
                                     // if a > 0, the largest value will be a >> b.min
                                     interval.max = max(a_interval.max >> b_interval.max,
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1043,10 +1043,10 @@ private:
                     if (op->is_intrinsic(Call::shift_left)) {
                         if (t.is_int() && t.bits() >= 32) {
                             // Overflow is UB
-                            if (a_interval.has_lower_bound() && b_interval.has_lower_bound() && can_prove(b_interval.min < t.bits())) {
+                            if (a_interval.has_lower_bound() && b_interval.has_lower_bound() && can_prove(b_interval.min >= 0 && b_interval.min < t.bits())) {
                                 interval.min = a_interval.min << b_interval.min;
                             }
-                            if (a_interval.has_upper_bound() && b_interval.has_upper_bound() && can_prove(b_interval.max < t.bits())) {
+                            if (a_interval.has_upper_bound() && b_interval.has_upper_bound() && can_prove(b_interval.max >= 0 && b_interval.max < t.bits())) {
                                 interval.max = a_interval.max << b_interval.max;
                             }
                         } else if (is_const(b)) {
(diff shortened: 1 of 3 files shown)


steven-johnson:
(Note that there are known failures here that would need addressing; before spending time doing so, it's probably worth getting feedback as to whether this change is desirable in the first place. IMHO it is or I wouldn't have proposed it, but I've been wrong before.)

zvookin:
To be clear: either we need to define the Halide language as supporting negative shift amounts by flipping the shift and implement this in codegen, or we need to assume it is undefined. The particular case Steven is working on is that the idea of flipped shifts seems to be affecting bounds inference and the execution model doesn't work that way so the bounds do not match what is really happening in the code. This is bad.

In general, any place the simplifier doesn't match execution has a possibility of resulting in incorrect bounds. E.g. the simplifier does not know about floating-point rounding and we do have bugs around that as well.

steven-johnson:
FWIW, the interesting failure here is in performance_const_division, which happens to calculate value-bounds on an Expr of the form `x >> y`, where y has a known bounds of `[0, 65535]`; this means we try to calculate the min as `x' >> 65535`, which goes boom.

Now, in this case, the arguably most-correct thing to do is to have Bounds attempt to detect if the shift amount is provably outside of the acceptable range, and if so, revert to bounds-of-type, but `can_prove()` is just a simplify call in hiding, and removing the simplifies from this path was a recent goal.

src/Bounds.cpp
```diff
@@ -1043,10 +1043,10 @@ class Bounds : public IRVisitor {
                     if (op->is_intrinsic(Call::shift_left)) {
                         if (t.is_int() && t.bits() >= 32) {
                             // Overflow is UB
-                            if (a_interval.has_lower_bound() && b_interval.has_lower_bound()) {
+                            if (a_interval.has_lower_bound() && b_interval.has_lower_bound() && can_prove(b_interval.min < t.bits())) {
```
abadams: Do we not also need to know it's >= 0?

src/Simplify_Call.cpp
```diff
@@ -125,56 +125,40 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
-                        } else {
-                            // shift-right-out-of-bounds -> zero or -1 for int.
-                            return mutate(select(a < 0, make_const(t, -1), make_zero(t)), bounds);
-                        }
-                    }
+                    return mutate(select(a < 0, make_const(t, -1), make_zero(t)), bounds);
                 }
             }
+        } else if (const_uint(b, &ub)) {
+            user_assert(ub < (uint64_t) t.bits()) << "bitshift by a constant amount >= the type size is not legal in Halide.";
```
abadams: I predict there are cases that will trip this that aren't due to user code, but rather created by IR manipulation inside the compiler (if not in bounds). We might need to return a quiet failure expression instead (e.g. signed_integer_overflow, though that would probably need to be renamed if used for const uints).

src/Simplify_Call.cpp
```diff
@@ -125,56 +125,40 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
-                constexpr int64_t i64_max = std::numeric_limits<int64_t>::max();
-                if (t.is_int()) {
-                    shift_left = !shift_left;
-                    // Ensure that corner case of ib = lowest() doesn't negate into itself
-                    ib = -std::max(ib, -i64_max);
+        uint64_t ub = 0;
+        if (const_int(b, &ib)) {
+            // LLVM shl and shr instructions produce poison for negative shifts,
+            // or for shifts >= typesize, so we will follow suit in our simplifier.
+            user_assert(ib >= 0) << "bitshift by a constant negative amount is not legal in Halide.";
```
abadams: I think this should possibly simplify to signed_integer_overflow instead of throwing an error. If we want to throw a user error, it should be in `Expr::operator<<(int)`

src/Bounds.cpp
steven-johnson: Done.

src/Simplify_Call.cpp
steven-johnson: That's possible, but I'd be inclined to see how much it happens in practice first; as of right now, public Halide code is clean with this in place. (Checking against the massive Google corpus of Halide code would be instructive.)

src/Simplify_Call.cpp
steven-johnson: See above re: immediate error. 

Re: immediate errors in `operator<<`` and friends, I considered doing this, but it wasn't clear to me how much precedent there was for this sort of thing. (I am a fan of failing hard-and-early for this sort of thing, though.)

steven-johnson:
FWIW: patching this in to Google injected exactly one new failure, which was a chunk of code of the form `select(b != 0, 1<<(b-1), 0)`, which was harmless but (technically) UB for b==0. Fixing that lets all of Google's code appear to build cleanly with this patch.

abadams:
I'm worried about this sort of thing:

```
ImageParam im(UInt(8), 2); // I know all these values are between 1 and 5, but I'm not going to tell Halide that.

lut(x) = sqrt(x);
out(x, y) = lut(cast<uint8_t>(1) << im(x))

```

It's reasonable for bounds inference to evaluate lut(x) over [0-255], but instead might we throw an error about out-of-range shifts, even though the user code would not technically be UB?

In general I worry about UB uses of << injected by the compiler due to conservative bounds inference, where no user code is actually UB. This is the sort of thing that bit us for division and required the introduction of a division intrinsic which is well-defined at zero (but where the value returned at zero can't possibly matter). It's also one reason we don't noisily alert on signed integer overflow in constant folding.

steven-johnson:
> instead might we throw an error about out-of-range shifts

No: this code would only generate an error if it can *prove* that an out-of-range shift will *definitely occur*. It can't prove it here, so you're still at the mercy of potential UB.

> I worry about UB uses of << injected by the compiler

That's fair, but if our bounds inference code is capable of injecting UB which we can easily detect at simplification time, we really need to fix the UB, rather than covering it up.

abadams:
Ah, I see. Bounds only constructs the shift if it can prove it's not UB.

PR #3662, opened 2019-02-07
Improve handling of clz, ctz, popcount

steven-johnson:
- Add special-casing of these to the Simplifier, so that constant values are simplified to constants
- Improve handling of bounds-calculation for clz to allow narrower min and max when the intervals are bounded

### DIFF
--- a/src/Simplify_Call.cpp
+++ b/src/Simplify_Call.cpp
@@ -79,11 +79,11 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
                 // popcount *is* well-defined for ua = 0
                 r = popcount64(ua);
             } else if (op->is_intrinsic(Call::count_leading_zeros)) {
-                user_assert(ua != 0) << "The result of count_leading_zeros(x) is undefined for x = 0.";
-                r = clz64(ua) - (64 - bits);
+                // clz64() is undefined for 0, but Halide's count_leading_zeros defines clz(0) = bits
+                r = ua == 0 ? bits : (clz64(ua) - (64 - bits));
             } else /* if (op->is_intrinsic(Call::count_trailing_zeros)) */ {
-                user_assert(ua != 0) << "The result of count_trailing_zeros(x) is undefined for x = 0.";
-                r = ctz64(ua);
+                // ctz64() is undefined for 0, but Halide's count_trailing_zeros defines clz(0) = bits
+                r = ua == 0 ? bits : (ctz64(ua));
             }
             return make_const(op->type, r);
         }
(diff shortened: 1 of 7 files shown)


src/Simplify_Call.cpp
```diff
@@ -20,6 +20,40 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
         }
+    } else if (op->is_intrinsic(Call::popcount) ||
+               op->is_intrinsic(Call::count_leading_zeros) ||
+               op->is_intrinsic(Call::count_trailing_zeros)) {
+        Expr a = mutate(op->args[0], nullptr);
+
+        uint64_t ua = 0;
+        if (const_int(a, (int64_t *)(&ua)) || const_uint(a, &ua)) {
+            const int bits = op->type.bits();
+            // __builtin_clz and friends only operate on uint;
```
abadams: __builtin_clzll ?

steven-johnson: Heh, didn't occur to me to check for longer variants. I'll sub those in and verify that they work on MSVC.

steven-johnson: Note that in addition to using the intrinsics, this now checks for static clz(0)/ctz(0) in the simplifier and makes it a compile error, and only does the new clz optimization in Bounds if we can prove the value is nonzero.

steven-johnson:
Grr, of course MSVC doesn't support long-long versions of those builtins

steven-johnson:
So @dsharletg and I had been puzzling over why it is that we have code that is calling `count_leading_zeros(0)` and yet seeming to get well-defined results... it looks like the answer is that we've been passing the wrong value to the `Intrinsic::ctlz` and `Intrinsic::cttz` calls in LLVM. The second arg is actually `is_zero_undef` (true -> zero is undef, false -> zero must be well-defined), but we labeled it `zero_is_not_undef` in our code and set it to `false`.

So, we've been generating well-defined results for ctlz(0) and cttz(0) for pretty much forever (at the expense of added test-and-branches in some configs).

I suppose we could change this codegen to actually do what is promised (make zero undef), with the knowledge that we'd have to chase down and fix the inevitable downstream bugs from code that happened to be relying on this behavior. (No idea if the presumably improved performance would be measurable.)

Alternately (and what I'd propose) is that we keep it this way and change the documentation for our code (and the bits in this particular PR) to simply enshrine the behavior as guaranteed.

abadams:
Oops. Yeah, making it well-defined for zero sounds like the best option.

steven-johnson:
(Note that we were already providing this behavior by mistake, we just didn't document or test it as such.)

- fixed comments for count_leading_zeros() and count_trailing_zeros
- rename misnamed arg in Codegen_LLVM
- revised Simplify code to handle zero properly
- revised Bounds code to allow for zero
- added missing definitions of count_trailing_zeros and popcount to the C backend, and added scalarizing

PR #3664, opened 2019-02-07
Fix require() for HVX backend

steven-johnson:
HVX (and possible OpenCL) have code to eliminate bool vectors by converting them into a mask; this breaks scalarizing require() if the condition is a vector, as it assumes the condition is still bool, vector or not. This just inserts a workaround to re-boolify the vector when scalarization is necessary.

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -2478,9 +2478,6 @@ void CodeGen_LLVM::visit(const Call *op) {
         internal_assert(op->args.size() == 3);
         Expr cond = op->args[0];
         if (cond.type().is_vector()) {
-            // This will dramatically slow down the code, so it's worthy
-            // of a warning to user, in case it was inserted in a non-obvious way.
-            user_warning << "The condition for a require() expr is vectorized; it will be split and scalarized.";
             if (!cond.type().is_bool()) {
                 // If we're building for a backend that uses bool_to_mask(),
                 // the vector might no longer be booleans, and scalarizing
@@ -2500,6 +2497,11 @@ void CodeGen_LLVM::visit(const Call *op) {
             create_assertion(c, op->args[2]);
             value = codegen(op->args[1]);
         }
+    } else if (op->is_intrinsic(Call::extract_mask_element)) {
+        internal_assert(op->args.size() == 2);
+        const int64_t *index = as_const_int(op->args[1]);
+        internal_assert(index);
+        value = codegen(Cast::make(Bool(), Shuffle::make_extract_element(op->args[0], *index)));
     } else if (op->is_intrinsic(Call::make_struct)) {
         if (op->type.is_vector()) {
             // Make a vector of pointers to distinct structs
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1163,7 +1163,7 @@ void CodeGen_LLVM::optimize_module() {
 #if LLVM_VERSION >= 80
             pm.add(createThreadSanitizerLegacyPassPass());
 #else
-            pm.add(createThreadSanitizerPass());            
+            pm.add(createThreadSanitizerPass());
 #endif
         };
         b.addExtension(PassManagerBuilder::EP_OptimizerLast, addThreadSanitizerPass);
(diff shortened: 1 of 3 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -2475,7 +2475,23 @@ void CodeGen_LLVM::visit(const Call *op) {
         internal_assert(op->args.size() == 3);
         Expr cond = op->args[0];
         if (cond.type().is_vector()) {
-            scalarize(op);
+            // This will dramatically slow down the code, so it's worthy
```
dsharletg: I wouldn't bother with this warning. We already have lots of things that silently scalarize and have bad performance, and this warning is going to be mostly ignored.

test/correctness/require.cpp
```diff
@@ -64,6 +65,28 @@ static void test(int vector_width) {
...
     test(4);
+
+    {
+        // Verify that the HVX backend can compile vectorized require() correctly
+        Target target("hexagon-32-noos-hvx_64");
+
+        Var x;
+        Func f;
+        f(x) = require(x > 0, x);
+        f.vectorize(x, 8).hexagon();
```
pranavb-ca: Not that it should matter for this testcase nor am I suggesting that 8 is unsupported but idiomatically on Hexagon, one would want ```f.vectorize(x, 32).hexagon()```

pranavb-ca:
Thanks, @steven-johnson. LGTM.

src/CodeGen_LLVM.cpp
```diff
@@ -2475,7 +2475,23 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+            // of a warning to user, in case it was inserted in a non-obvious way.
+            user_warning << "The condition for a require() expr is vectorized; it will be split and scalarized.";
+            if (!cond.type().is_bool()) {
+                // If we're building for a backend that uses bool_to_mask(),
+                // the vector might no longer be booleans, and scalarizing
+                // can fail at codegen time (since the conditions are expected)
+                // to be boolean.
+                cond = (cond != make_zero(cond.type()));
```
dsharletg: I think the right thing to do here might be to to make it so scalarize uses `Call::extract_mask_element`.

steven-johnson: AFACT `extract_mask_element` is only defined for the HVX backend. (Maybe a better fix is to specialize `require` for that backend...)

dsharletg: That just looks like a bug. It looks like the implementation of extract_mask_element in CodeGen_Hexagon should be in CodeGen_LLVM.

dsharletg: Hmm, it looks the same for bool_to_mask, but maybe there was a reason for this. I vaguely recall discussing this with @abadams...

steven-johnson: This is by no means an urgent fix, so I'll wait to see if @abadams chimes in with any recollection about it.

dsharletg: Thought about this more. I think the mask intrinsic implementations should stay in the codegen backends, scalarize should use extract_mask_element, and it should be implemented on any target that uses eliminate_bool_vectors. Since OpenCL doesn't use CodeGen_LLVM, everything I said above wouldn't impact it anyways.

steven-johnson: Moved Call::extract_mask_element from Hexagon to LLVM backend.

test/correctness/require.cpp
steven-johnson: Done

src/CodeGen_LLVM.cpp
steven-johnson: Done

steven-johnson:
PTAL

dsharletg:
I think moving Call::extract_mask_element to CodeGen_LLVM is not right (sorry for the confusion, my last comment was really unclear).

This is also side cleanup that isn't going to affect the issue in this case (CodeGen_Hexagon still had access to extract_mask_element). The issue is that scalarize doesn't use extract_mask_element to get at the scalars of a condition vector.

I've thought even more about this. Let me make an attempt at fixing this. I think what we need to do is:
- Make default implementations of the mask intrinsics that are no-ops (i.e. assume eliminate_bool_vectors never ran).
- Use those in scalarize (and anywhere else)
- Keep the non-default implementations of the mask intrinsics in the target specific backends.

steven-johnson:
> sorry for the confusion

No worries

> Let me make an attempt at fixing this

Sure, just keep in mind this is very-very-low-priority at this point (though the Pipeline::require proposal might bump up the need for this fix).

dsharletg:
I thought it was a bit higher priority because I view it as scalarize being broken, not require.

After looking at it a bit, `Call::if_then_else` has the same problem.

I think that select *would* have the same problem, but `select` gets turned into `select_mask` by `eliminate_bool_vectors`. I think that's the proper fix here: `Call::if_then_else` and `Call::require` also need a `Call::if_then_else_mask` and `Call::require_mask` version.

steven-johnson:
Closing in favor of https://github.com/halide/Halide/pull/3686

PR #3666, opened 2019-02-08
Remove apparently-useless function table in CodegenHexagon

steven-johnson:
The function table only contains ops that begin with `halide.hexagon.`, all of which are special-cased before the function table is ever used.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -1742,18 +1742,6 @@ void CodeGen_Hexagon::visit(const Call *op) {
     internal_assert(op->is_extern() || op->is_intrinsic())
         << "Can only codegen extern calls and intrinsics\n";
 
-    // Map Halide functions to Hexagon intrinsics, plus a boolean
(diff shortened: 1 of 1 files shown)


steven-johnson:
```
    if (starts_with(op->name, "halide.hexagon.")) {
        // Handle all of the intrinsics we generated in
        // hexagon_optimize.  I'm not sure why this is different than
        // letting it fall through to CodeGen_LLVM.
        value = call_intrin(op->type, op->name, op->args);
        return;
    }
```

dsharletg:
That doesn't mean this code is useless. Ops might still be e.g. Call::bitwise_and, which that if would be true for.

I actually just checked, and it doesn't look like HexagonOptimize handles these calls. So I think this is very much not useless.

pranavb-ca:
@steven-johnson  - HexagonOptimize.cpp doesn't handle all (in fact, most) of these Call:xxxxx calls. This table is needed.

steven-johnson:
Ouch, you are right -- I was mistaken about the lookup key. Sorry for the noise.

PR #3667, opened 2019-02-10
Basic bfloat16 support

abadams:
Lowers to uint16s for all backends, with math done by temporarily placing the bits in the MSBs of a float.

### DIFF
--- a/src/Float16.h
+++ b/src/Float16.h
@@ -16,6 +16,11 @@ namespace Halide {
  * */
 struct float16_t {
 
+    static const int mantissa_bits = 10;
+    static const uint16_t sign_mask = 0x8000;
+    static const uint16_t exponent_mask = 0x7c00;
+    static const uint16_t mantissa_mask = 0x03ff;
+
     /// \name Constructors
     /// @{
 
@@ -40,12 +45,6 @@ struct float16_t {
     explicit operator float() const;
     /** Cast to double */
     explicit operator double() const;
-
-    // Be explicit about how the copy constructor is expected to behave
-    float16_t(const float16_t&) = default;
-
-    // Be explicit about how assignment is expected to behave
-    float16_t& operator=(const float16_t&) = default;
 
     /** \name Convenience "constructors"
      */
@@ -122,6 +121,123 @@ static_assert(sizeof(float16_t) == 2, "float16_t should occupy two bytes");
 template<>
 HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<Halide::float16_t>() {
     return halide_type_t(halide_type_float, 16);
+}
+
+namespace Halide {
+
+/** Class that provides a type that implements half precision
+ *  floating point using the bfloat16 format.
+ *
+ *  This type is enforced to be 16-bits wide and maintains no state
+ *  other than the raw bits so that it can passed to code that checks
+ *  a type's size and used for buffer_t allocation. */
+struct bfloat16_t {
+
+    static const int mantissa_bits = 7;
+    static const uint16_t sign_mask = 0x8000;
+    static const uint16_t exponent_mask = 0x7f80;
+    static const uint16_t mantissa_mask = 0x007f;
+
+    /// \name Constructors
+    /// @{
+
+    /** Construct from a float, double, or int using
+     * round-to-nearest-ties-to-even. Out-of-range values become +/-
+     * infinity.
+     */
+    // @{
+    explicit bfloat16_t(float value);
+    explicit bfloat16_t(double value);
+    explicit bfloat16_t(int value);
+    // @}
+
+    /** Construct a bfloat16_t with the bits initialised to 0. This represents
+     * positive zero.*/
+    bfloat16_t();
+
+    /// @}
+
+    // Use explicit to avoid accidently raising the precision
+    /** Cast to float */
+    explicit operator float() const;
+    /** Cast to double */
+    explicit operator double() const;
+
+    /** \name Convenience "constructors"
+     */
+    /**@{*/
+
+    /** Get a new bfloat16_t that represents zero
+     * \param positive if true then returns positive zero otherwise returns
+     *        negative zero.
+     */
+    static bfloat16_t make_zero(bool positive);
+
+    /** Get a new float16_t that represents infinity
+     * \param positive if true then returns positive infinity otherwise returns
+     *        negative infinity.
+     */
+    static bfloat16_t make_infinity(bool positive);
+
+    /** Get a new bfloat16_t that represents NaN (not a number) */
+    static bfloat16_t make_nan();
+
+    /** Get a new bfloat16_t with the given raw bits
+     *
+     * \param bits The bits conformant to IEEE754 binary16
+     */
+    static bfloat16_t make_from_bits(uint16_t bits);
+
+    /**@}*/
+
+    /** Return a new bfloat16_t with a negated sign bit*/
+    bfloat16_t operator-() const;
+
+    /** Arithmetic operators. */
+    // @{
+    bfloat16_t operator+(bfloat16_t rhs) const;
+    bfloat16_t operator-(bfloat16_t rhs) const;
+    bfloat16_t operator*(bfloat16_t rhs) const;
+    bfloat16_t operator/(bfloat16_t rhs) const;
+    // @}
+
+    /** Comparison operators */
+    // @{
+    bool operator==(bfloat16_t rhs) const;
+    bool operator!=(bfloat16_t rhs) const { return !(*this == rhs); }
+    bool operator>(bfloat16_t rhs) const;
+    bool operator<(bfloat16_t rhs) const;
+    bool operator>=(bfloat16_t rhs) const { return (*this > rhs) || (*this == rhs); }
+    bool operator<=(bfloat16_t rhs) const { return (*this < rhs) || (*this == rhs); }
+    // @}
+
+    /** Properties */
+    // @{
+    bool is_nan() const;
+    bool is_infinity() const;
+    bool is_negative() const;
+    bool is_zero() const;
+    // @}
+
+    /** Returns the bits that represent this bfloat16_t.
+     *
+     *  An alternative method to access the bits is to cast a pointer
+     *  to this instance as a pointer to a uint16_t.
+     **/
+    uint16_t to_bits() const;
+
+private:
+    // The raw bits.
+    uint16_t data;
+};
+
+static_assert(sizeof(bfloat16_t) == 2, "bfloat16_t should occupy two bytes");
+
+}  // namespace Halide
+
+template<>
+HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<Halide::bfloat16_t>() {
+    return halide_type_t(halide_type_bfloat, 16);
 }
 
 #endif
--- a/src/Type.cpp
+++ b/src/Type.cpp
@@ -100,16 +100,23 @@ bool Type::is_min(uint64_t x) const {
 }
 
 bool Type::can_represent(Type other) const {
+    if (*this == other) return true;
     if (lanes() != other.lanes()) return false;
     if (is_int()) {
         return ((other.is_int() && other.bits() <= bits()) ||
                 (other.is_uint() && other.bits() < bits()));
     } else if (is_uint()) {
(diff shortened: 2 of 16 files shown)


src/Float16.h
```diff
@@ -24,9 +29,9 @@ struct float16_t {
      * infinity.
      */
     // @{
-    explicit float16_t(float value);
-    explicit float16_t(double value);
-    explicit float16_t(int value);
+    float16_t(float value);
```
steven-johnson: Are we really sure about removing the `explicit` here?

src/Float16.h
```diff
@@ -124,4 +129,127 @@ HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<Halide::float16_t>() {
...
+
+    /// \name Constructors
+    /// @{
+
+    /** Construct from a float, double, or int using
+     * round-to-nearest-ties-to-even. Out-of-range values become +/-
+     * infinity.
+     */
+    // @{
+    bfloat16_t(float value);
```
steven-johnson: are we sure we don't want `explicit` here?

src/Float16.h
```diff
@@ -124,4 +129,127 @@ HALIDE_ALWAYS_INLINE halide_type_t halide_type_of<Halide::float16_t>() {
...
+
+    /** \name Convenience "constructors"
+     */
+    /**@{*/
+
+    /** Get a new bfloat16_t that represents zero
+     * \param positive if true then returns positive zero otherwise returns
+     *        negative zero.
+     */
+    static bfloat16_t make_zero(bool positive);
```
steven-johnson: Nit: reading `make_zero(true)` at a call site isn't very obvious as to what the boolean means. Maybe not used enough to matter, but something like `enum { kPositive, kNegative }` would be more obvious.

src/Type.cpp
```diff
@@ -120,6 +127,13 @@ bool Type::can_represent(int64_t x) const {
         return x >= min_int(bits()) && x <= max_int(bits());
     } else if (is_uint()) {
         return x >= 0 && (uint64_t)x <= max_uint(bits());
+    } else if (is_bfloat()) {
+        switch (bits()) {
+        case 16:
+            return (int64_t)(float)(bfloat16_t)(float)x == x;
```
steven-johnson: I am speechless

src/runtime/HalideRuntime.h
```diff
@@ -379,8 +379,9 @@ typedef enum halide_type_code_t
 {
     halide_type_int = 0,   //!< signed integers
     halide_type_uint = 1,  //!< unsigned integers
-    halide_type_float = 2, //!< floating point numbers
-    halide_type_handle = 3 //!< opaque pointer type (void *)
+    halide_type_float = 2, //!< IEEE floating point numbers
+    halide_type_handle = 3, //!< opaque pointer type (void *)
+    halide_type_bfloat = 4, //!< floating point numbers in the bfloat format
```
steven-johnson: I know this is almost certainly necessary, but I predict we'll find a surprisingly long tail of code that will need to be tweaked for the addition of a new Typecode.

src/Type.cpp
abadams: Haha yeah - I was copying the original code. The idea is to round-trip from int64 to bfloat16 then back to int64 to see if the value was preserved. The bfloat16 <-> int64 conversion must be done via float in both directions though.

src/runtime/HalideRuntime.h
steven-johnson: Probably worth commenting on that to be sure no one in the future tries to "optimize" it

src/Float16.h
abadams: It's to support things like some_bfloat16 > 0, or some_bfloat16 + 1, but now that you mention it it's going to have unwanted side-effects, like some_bfloat16 + 1.0f won't get promoted to float.

I guess I could add more overloads below instead. Do you think that would be better?

abadams:
Tagging @jeffsetter , who requested this feature.

src/Float16.h
steven-johnson: Yeah, non-explicit single-arg ctors that take 'primitive' types (numbers, strings) are a bug waiting to happen; IMHO we'd be better off making them all explicit and seeing how much pain that really is in practice. Before adding more overloads it may be worth just seeing how many explicit casts are necessary; if it's small-ish, I think that's the better option.

abadams: Done. It was smallish.

src/runtime/HalideRuntime.h
abadams: Not sure what you mean - how would one "optimize" it?

abadams:
The problem now is that llvm jumps to a null address for targets without native support for half-conversions. Presumably there's a support library that's not getting linked in. I might need to patch in a call into the runtime conversion routines on targets without explicit support.

src/runtime/HalideRuntime.h
steven-johnson: whoops, that response went on the wrong line. I meant, a comment on the line with the four-in-a-row-cast might be useful in case someone in the future thinks it's redundant.

tools/halide_image_io.h
```diff
@@ -1192,7 +1192,7 @@ bool save_mat(ImageType &im, const std::string &filename) {
             check(false, "unreachable");
         };
         break;
-    case halide_type_handle:
+    default:
```
zvookin: Might be worth also adding the bfloat type case and giving an explicit error message that bfloat is not supported. (The default should remain in place of course.)

jeffsetter:
This code worked properly for some of my examples for emulating bfloat16 execution on a cpu. For hardware that has bfloat16 hardware, lines 68-70 were not completely what I found. Instead, types had to be casted to prevent mismatch between input types.

steven-johnson:
Is this PR still interesting?

abadams:
Yes, it just got stalled because LLVM seems buggy for float16 right now, so I had to emulate it, but my emulate is apparently buggy. I need to get back to it. Current plan is that according to Benoit, Eigen has good vectorizable float16 conversion code I could crib. I just need to wait for a day where I'm procrastinating on something more important :)

jeffsetter:
A suggestion is that even rounding might be an option when converting float to bfloat. Some untested code that I ended up using for even rounding was:
```
     Expr float_to_bfloat(Expr e) {
        e = reinterpret(UInt(32, e.type().lanes()), e);

	// round float to even                                                                                                                                                                            
        Expr half = UIntImm::make(UInt(32), 0x00008000);
        Expr sum = e + half;
        // check if bottom bits are all zero                                                                                                                                                              
        Expr mantissa_mask = UIntImm::make(UInt(32), 0x0000ffff);
        Expr zeroed = (sum & mantissa_mask) == 0;
        // clear last bit (round even) on tie                                                                                                                                                             
        Expr clear_mask = ~(cast(UInt(32, e.type().lanes()), zeroed) << 16);
        e = sum & clear_mask;

        // shift out bottom bits and cast                                                                                                                                                                              
        e = e >> 16;
        e = cast(UInt(16, e.type().lanes()), e);
        return e;
    }
```

abadams:
TODO: Fix #4019

abadams:
TODO: fix #4040

abadams:
Ok, ready for another round of review. PTAAL. The latest approach requires being very explicit in the backends about what LLVM types are used for storage and arithmetic for a given Halide type.

src/CodeGen_LLVM.cpp
```diff
@@ -1739,13 +1875,7 @@ Value *CodeGen_LLVM::codegen_buffer_pointer(Value *base_address, Halide::Type ty
         index = promote_64(index);
     }
 
-    // Handles are always indexed as 64-bit.
-    if (type.is_handle()) {
-        return codegen_buffer_pointer(base_address, UInt(64, type.lanes()), index);
-    } else {
```
shoaibkamil: Not sure I understand why this case won't arise anymore.

abadams:
https://bugs.llvm.org/show_bug.cgi?id=43065

src/CodeGen_LLVM.cpp
abadams: It's now handled in the codegen_buffer_pointer variant that this calls. upgrade_type_for_storage does the promotion of handles to uint64s.

src/CodeGen_X86.cpp
```diff
@@ -504,5 +508,18 @@ int CodeGen_X86::vector_lanes_for_slice(Type t) const {
     return slice_bits / t.bits();
 }
 
+llvm::Type *CodeGen_X86::llvm_type_of(Type t) const {
+    if (t.is_float() && t.bits() < 32) {
+        // LLVM as of August 2019 has all sorts of issues in the x86
+        // backend for half types. It injects expensive calls to
+        // convert between float and half for seemingly no reason
+        // (e.g. to do a select), and bitcasting to int16 doesn't
+        // help, because it simplifies away the bitcast for you.
```
shoaibkamil: Maybe reference the relevant bug(s) here so if they get fixed, this can be changed?

abadams: Done

src/CodeGen_Hexagon.cpp
```diff
@@ -87,7 +87,7 @@ Stmt acquire_hvx_context(Stmt stmt, const Target &target) {
     // register a destructor to call halide_qurt_hvx_unlock.
     Stmt check_hvx_lock = call_halide_qurt_hvx_lock(target);
     Expr dummy_obj = reinterpret(Handle(), cast<uint64_t>(1));
-    Expr hvx_unlock = Call::make(Int(32), Call::register_destructor,
+    Expr hvx_unlock = Call::make(Handle(), Call::register_destructor,
```
steven-johnson: Here and elsewhere: I don't see that the sig of Call::register_destructor changed; was this just always wrong before and we were just getting lucky? (If so, is change actually related to bfloat or is it just a driveby fix?)

src/CodeGen_LLVM.cpp
```diff
@@ -1460,6 +1473,34 @@ void CodeGen_LLVM::codegen(Stmt s) {
     s.accept(this);
 }
 
+Type CodeGen_LLVM::upgrade_type_for_arithmetic(Type t) const {
```
steven-johnson: nit: wouldn't it be better to pass Type by const-ref?

src/CodeGen_X86.cpp
```diff
@@ -504,5 +508,19 @@ int CodeGen_X86::vector_lanes_for_slice(Type t) const {
...
+        // See: https://bugs.llvm.org/show_bug.cgi?id=43065
```
steven-johnson: We should probably also create a Halide issue to track this

src/Type.cpp
steven-johnson: `(int64_t)(float)(bfloat16_t)(float)` is my next band name

src/Type.h
```diff
@@ -361,7 +363,11 @@ struct Type {
 
     /** Is this type a floating point type (float or double). */
     HALIDE_ALWAYS_INLINE
-    bool is_float() const {return code() == Float;}
+    bool is_float() const {return code() == Float || code() == BFloat;}
```
steven-johnson: ...are we *certain* that this is the right thing to do? Without having examined downstream uses, I'd bet that for some cases we need this semantics but for others we don't. (Alternative is presumably to leave as-is and add `is_float_or_bfloat()`)

src/CodeGen_Hexagon.cpp
abadams: This was always wrong before, but there was no checking that the llvm return type of the intrinsic. matched the Halide type. I added checking in this PR to make sure I got the bfloat stuff right, and it necessitated fixes elsewhere.

src/Type.h
abadams: Yes, I think so. bfloats are still floats, it's just that there are two different ways to make a 16-bit float, so setting bits = 16 is ambiguous. I'm treating them as a subtype.

src/CodeGen_LLVM.cpp
abadams: done

src/CodeGen_X86.cpp
abadams: done

PR #3671, opened 2019-02-16
Let generators compute a GCD target when multiple targets are passed to `-r`

alexreinking:
Currently, generators support compiling Halide functions for multiple platform-compatible targets and automatically selecting the "best" one (in order) via runtime checks.

This PR extends a compatible behavior to the generator's runtime generation (via `-r`). Scripts that compile multiple functions into a single static library can now pass a comma-separated list of all the function targets to the generator and a "greatest common denominator" runtime will be selected.

This is a **draft** as there are certain target features that I'm not familiar with. I'm asking for some help deciding whether a GCD target should include a feature if _either_ source target has it (eg. `cuda`), only if _both_ have it (eg. `avx2`), or if it should be dropped (eg. `no_runtime`). The behavior does not change when a single runtime is specified, so tests are still needed, too.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -988,7 +988,7 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
         std::string base_path = compute_base_path(output_dir, runtime_name, "");
 
         Target gcd_target = targets[0];
-        for (int i = 1; i < targets.size(); i++) {
+        for (size_t i = 1; i < targets.size(); i++) {
             if (!gcd_target.get_runtime_compatible_target(targets[i], gcd_target)) {
                 user_error << "Failed to find compatible runtime target for "
                            << gcd_target.to_string()
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -977,7 +977,7 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
     auto target_strings = split_string(generator_args["target"].string_value, ",");
     std::vector<Target> targets;
     for (const auto &s : target_strings) {
-        targets.push_back(Target(s));
+        targets.emplace_back(s);
     }
 
     if (!runtime_name.empty()) {
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -813,10 +813,17 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
         " -x  A comma separated list of file extension pairs to substitute during\n"
         "     file naming, in the form [.old=.new[,.old2=.new2]]\n"
         "\n"
-        " -p  A comma-separted list of shared libraries that will be loaded before the\n"
+        " -p  A comma-separated list of shared libraries that will be loaded before the\n"
         "     generator is run. Useful for custom auto-schedulers. The generator must\n"
         "     either be linked against a shared libHalide or compiled with -rdynamic\n"
-        "     so that references in the shared library to libHalide can resolve.\n";
+        "     so that references in the shared library to libHalide can resolve.\n"
+        "\n"
+        "-r   The name of a standalone runtime to generate. Only honors EMIT_OPTIONS 'o'\n"
+        "     and 'static_library'. When multiple targets are specified, it picks a\n"
+        "     runtime that is compatible with all of the targets, or fails if it cannot\n"
+        "     find one. Flags across all of the targets that do not affect runtime code\n"
+        "     generation, such as `no_asserts` and `no_runtime`, are ignored.\n"
+        ;
 
     std::map<std::string, std::string> flags_info = { { "-f", "" },
                                                       { "-g", "" },
@@ -855,7 +862,7 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
 
     // It's possible that in the future loaded plugins might change
     // how arguments are parsed, so we handle those first.
-    for (auto lib : split_string(flags_info["-p"], ",")) {
+    for (const auto &lib : split_string(flags_info["-p"], ",")) {
         if (lib.empty()) continue;
 #ifdef _WIN32
         if (LoadLibrary(lib.c_str()) != nullptr) {
@@ -885,7 +892,7 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
         // no longer infer the name when only one Generator is registered
         cerr << "Either -g <name> or -r must be specified; available Generators are:\n";
         if (!generator_names.empty()) {
-            for (auto name : generator_names) {
+            for (const auto &name : generator_names) {
                 cerr << "    " << name << "\n";
             }
         } else {
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -981,10 +981,6 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
     }
 
     if (!runtime_name.empty()) {
-        if (targets.size() != 1) {
-            cerr << "Only one target allowed here";
-            return 1;
-        }
         std::string base_path = compute_base_path(output_dir, runtime_name, "");
 
         Target gcd_target = targets[0];
@@ -995,6 +991,10 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
                            << " and "
                            << targets[i].to_string() << '\n';
             }
+        }
+
+        if (targets.size() > 1) {
+            debug(1) << "Building runtime for computed target: " << gcd_target.to_string() << "\n";
         }
 
         Outputs output_files = compute_outputs(gcd_target, base_path, emit_options);
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -986,8 +986,19 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
             return 1;
         }
         std::string base_path = compute_base_path(output_dir, runtime_name, "");
-        Outputs output_files = compute_outputs(targets[0], base_path, emit_options);
-        compile_standalone_runtime(output_files, targets[0]);
+
+        Target gcd_target = targets[0];
+        for (int i = 1; i < targets.size(); i++) {
+            if (!gcd_target.get_runtime_compatible_target(targets[i], gcd_target)) {
+                user_error << "Failed to find compatible runtime target for "
+                           << gcd_target.to_string()
+                           << " and "
+                           << targets[i].to_string() << '\n';
+            }
+        }
+
+        Outputs output_files = compute_outputs(gcd_target, base_path, emit_options);
+        compile_standalone_runtime(output_files, gcd_target);
     }
 
     if (!generator_name.empty()) {
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -1,5 +1,6 @@
 #include <iostream>
 #include <string>
+#include <array>
 
 #include "Target.h"
 
@@ -187,10 +188,7 @@ int get_highest_cuda_capability(const Target& t) {
     if (t.has_feature(Target::CUDACapability30)) {
         return 30;
     }
-
-    // invalid target
-    user_warning << "called get_highest_cuda_capability on invalid target (CUDA set with no capability)\n";
-    return -1;
+    return 20;
 }
 
 }  // namespace
@@ -775,49 +773,42 @@ int Target::natural_vector_size(const Halide::Type &t) const {
 }
 
 bool Target::get_runtime_compatible_target(const Target& other, Target &result) {
-    // A compatible runtime must support these features if _either_ of the base runtimes
-    static const std::vector<Feature> unionFeatures = {
-            // These features assist in debugging.
-            Debug, TSAN, ASAN, MSAN,
-
-            // These features enable access to other devices.
-            CUDA, CUDACapability30, CUDACapability32, CUDACapability35, CUDACapability50, CUDACapability61, OpenCL,
-            OpenGL, OpenGLCompute, MinGW, Metal, HexagonDma, HVX_64, HVX_128, HVX_v62, HVX_v65, HVX_v66,
-            HVX_shared_object, D3D12Compute
+    const std::array<Feature, 25> union_features = {
+            Debug, TSAN, ASAN, MSAN, CUDA, CUDACapability30, CUDACapability32, CUDACapability35, CUDACapability50,
+            CUDACapability61, OpenCL, OpenGL, OpenGLCompute, MinGW, Metal, HexagonDma, HVX_64, HVX_128, HVX_v62,
+            HVX_v65, HVX_v66, HVX_shared_object, D3D12Compute, NoNEON, SoftFloatABI
     };
 
-    static const std::vector<Feature> intersectionFeatures = {
-            // These features specify additional instruction sets available on the target architecture.
-            SSE41, AVX, AVX2, FMA, FMA4, F16C, ARMv7s, VSX, AVX512,AVX512_KNL, AVX512_Skylake, AVX512_Cannonlake
+    const std::array<Feature, 12> intersection_features = {
+            SSE41, AVX, AVX2, FMA, FMA4, F16C, ARMv7s,VSX, AVX512, AVX512_KNL, AVX512_Skylake, AVX512_Cannonlake
     };
 
-    // I don't know how to merge these features because I don't know what they do. Reviewers, please advise :)
-    static const std::vector<Feature> iDontKnowWhatToDoWithTheseFeaturesPleaseHelp = {
-            CLDoubles, CLHalf, UserContext, Matlab, Profile, CPlusPlusMangling, LargeBuffers, FuzzFloatStores,
-            SoftFloatABI, StrictFloat, LegacyBufferWrappers, CheckUnsafePromises, EmbedBitcode,
-            DisableLLVMLoopVectorize, DisableLLVMLoopUnroll, NoNEON, POWER_ARCH_2_07
-    };
+    // bitsets need to be the same width.
+    decltype(result.features) union_mask;
+    decltype(result.features) intersection_mask;
+
+    for (auto& feature : union_features) {
+        union_mask.set(feature);
+    }
+
+    for (auto& feature : intersection_features) {
+        intersection_mask.set(feature);
+    }
 
     if (arch != other.arch || bits != other.bits || os != other.os) {
         user_warning << "runtime targets must agree on platform (arch-bits-os)\n";
         return false;
     }
 
-    std::vector<Feature> merged_features{};
-
-    for (const auto &feature : unionFeatures) {
-        if (has_feature(feature) || other.has_feature(feature)) {
-            merged_features.push_back(feature);
-        }
+    if (has_feature(SoftFloatABI) != other.has_feature(SoftFloatABI)) {
+        user_warning << "runtime targets must agree on SoftFloatABI\n";
+        return false;
     }
 
-    for (const auto &feature : intersectionFeatures) {
-        if (has_feature(feature) && other.has_feature(feature)) {
-            merged_features.push_back(feature);
-        }
-    }
+    Target output{os, arch, bits};
+    output.features = ((features | other.features) & union_mask)
+            | ((features & other.features) & intersection_mask);
 
-    Target output = Target(os, arch, bits, merged_features);
     if (!fixup_gcd_target(output)) {
         return false;
     }
@@ -828,31 +819,13 @@ bool Target::get_runtime_compatible_target(const Target& other, Target &result) 
 
 
 bool Target::fixup_gcd_target(Target& target) {
-    // These features do not affect runtime code generation.
-    static const std::vector<Feature> irrelevantFeatures = {
-            JIT, NoAsserts, NoBoundsQuery, NoRuntime, TraceLoads, TraceStores, TraceRealizations
-    };
-
-    // Ensure that irrelevant features are not set
-    for (const auto &feature : irrelevantFeatures) {
-        target.features.reset(feature);
-    }
-
     // Pick highest CUDA capability. Use fall-through to clear redundant features
     switch (get_highest_cuda_capability(target)) {
-        case -1:
-            user_warning << "cannot merge cuda-enabled targets without knowing capability levels\n";
-            return false;
-        case 61:
-            target.features.reset(CUDACapability50);
-        case 50:
-            target.features.reset(CUDACapability35);
-        case 35:
-            target.features.reset(CUDACapability32);
-        case 32:
-            target.features.reset(CUDACapability30);
-        default:
-            break;
+        case 61: target.features.reset(CUDACapability50);
+        case 50: target.features.reset(CUDACapability35);
+        case 35: target.features.reset(CUDACapability32);
+        case 32: target.features.reset(CUDACapability30);
+        default: break;
     }
 
     // Check to see that there weren't two or more different HVX flags specified.
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -871,7 +871,7 @@ bool Target::get_runtime_compatible_target(const Target& other, Target &result) 
 }
 
 
-    namespace Internal {
+namespace Internal {
 
 void target_test() {
     Target t;
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -796,9 +796,9 @@ bool Target::get_runtime_compatible_target(const Target& other, Target &result) 
     // (b) must be included if both targets have the feature (intersection)
     // (c) must match across both targets; it is an error if one target has the feature and the other doesn't
     const std::array<Feature, 25> union_features = {
-            Debug, TSAN, ASAN, MSAN, CUDA, CUDACapability30, CUDACapability32, CUDACapability35, CUDACapability50,
-            CUDACapability61, OpenCL, OpenGL, OpenGLCompute, MinGW, Metal, HexagonDma, HVX_64, HVX_128, HVX_v62,
-            HVX_v65, HVX_v66, HVX_shared_object, D3D12Compute, NoNEON, SoftFloatABI
+            CUDA, CUDACapability30, CUDACapability32, CUDACapability35, CUDACapability50, CUDACapability61, OpenCL,
+            OpenGL, OpenGLCompute, MinGW, Metal, HexagonDma, HVX_64, HVX_128, HVX_v62, HVX_v65, HVX_v66,
+            HVX_shared_object, D3D12Compute, NoNEON
     };
 
     const std::array<Feature, 12> intersection_features = {
@@ -841,6 +841,7 @@ bool Target::get_runtime_compatible_target(const Target& other, Target &result) 
     // We merge the bits via bitwise or.
     Target output{os, arch, bits};
     output.features = ((features | other.features) & union_mask)
+            | ((features | other.features) & matching_mask)
             | ((features & other.features) & intersection_mask);
 
     if (!fixup_gcd_target(output)) {
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -197,6 +197,8 @@ bool is_using_hexagon(const Target &t) {
            || t.has_feature(Target::HVX_v62)
            || t.has_feature(Target::HVX_v65)
            || t.has_feature(Target::HVX_v66)
+           || t.has_feature(Target::HexagonDma)
+           || t.has_feature(Target::HVX_shared_object)
            || t.arch == Target::Hexagon;
 }
 
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -165,9 +165,7 @@ Target calculate_host_target() {
 #endif
 #endif
 
-    Target t(os, arch, bits, initial_features);
-
-    return Target(os, arch, bits, initial_features);
+    return {os, arch, bits, initial_features};
 }
 
 }  // namespace
@@ -450,7 +448,7 @@ void bad_target_string(const std::string &target) {
     }
     separator = "";
     std::string oses;
-    for (auto os_entry : os_name_map) {
+    for (const auto &os_entry : os_name_map) {
         oses += separator + os_entry.first;
         separator = ", ";
     }
@@ -459,7 +457,7 @@ void bad_target_string(const std::string &target) {
     // assume the first line starts with "Features are ".
     int line_char_start = -(int)sizeof("Features are");
     std::string features;
-    for (auto feature_entry : feature_name_map) {
+    for (const auto &feature_entry : feature_name_map) {
         features += separator + feature_entry.first;
         if (features.length() - line_char_start > 70) {
             separator = "\n";
@@ -508,9 +506,7 @@ Target::Target(const std::string &target) {
     }
 }
 
-Target::Target(const char *s) {
-    *this = Target(std::string(s));
-}
+Target::Target(const char *s) : Target(std::string(s)) {}
 
 bool Target::validate_target_string(const std::string &s) {
     Target t;
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -169,41 +169,41 @@ Target calculate_host_target() {
     return {os, arch, bits, initial_features};
 }
 
-int get_highest_cuda_capability(const Target& t) {
+int get_cuda_capability_lower_bound(const Target &t) {
     if (!t.has_feature(Target::CUDA)) {
         return 0;
-    }
-    if (t.has_feature(Target::CUDACapability61)) {
-        return 61;
-    }
-    if (t.has_feature(Target::CUDACapability50)) {
-        return 50;
-    }
-    if (t.has_feature(Target::CUDACapability35)) {
-        return 35;
-    }
-    if (t.has_feature(Target::CUDACapability32)) {
-        return 32;
     }
     if (t.has_feature(Target::CUDACapability30)) {
         return 30;
     }
+    if (t.has_feature(Target::CUDACapability32)) {
+        return 32;
+    }
+    if (t.has_feature(Target::CUDACapability35)) {
+        return 35;
+    }
+    if (t.has_feature(Target::CUDACapability50)) {
+        return 50;
+    }
+    if (t.has_feature(Target::CUDACapability61)) {
+        return 61;
+    }
     return 20;
 }
 
-int get_highest_hvx_version(const Target& t) {
+int get_hvx_lower_bound(const Target &t) {
     if (!t.has_feature(Target::HexagonDma) && t.arch != Target::Hexagon) {
         // TODO: is this right? how do these options differ
         return 0;
     }
-    if (t.has_feature(Target::HVX_v66)) {
-        return 66;
+    if (t.has_feature(Target::HVX_v62)) {
+        return 62;
     }
     if (t.has_feature(Target::HVX_v65)) {
         return 65;
     }
-    if (t.has_feature(Target::HVX_v62)) {
-        return 62;
+    if (t.has_feature(Target::HVX_v66)) {
+        return 66;
     }
     // TODO: Is there a lower HVX version, like 20 for CUDA?
     return -1;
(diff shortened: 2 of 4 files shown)


src/Target.cpp
```diff
@@ -165,9 +165,32 @@ Target calculate_host_target() {
...
+        return 35;
+    }
+    if (t.has_feature(Target::CUDACapability32)) {
+        return 32;
+    }
+    if (t.has_feature(Target::CUDACapability30)) {
+        return 30;
+    }
+
+    // invalid target
```
abadams: Return 20 if no capabilities are set. That's the minimum.

abadams: (that we support)

src/Target.cpp
```diff
@@ -753,6 +774,103 @@ int Target::natural_vector_size(const Halide::Type &t) const {
     }
 }
 
+bool Target::get_runtime_compatible_target(const Target& other, Target &result) {
+    // A compatible runtime must support these features if _either_ of the base runtimes
+    static const std::vector<Feature> unionFeatures = {
```
abadams: camel -> snake

src/Target.cpp
```diff
@@ -753,6 +774,103 @@ int Target::natural_vector_size(const Halide::Type &t) const {
...
+
+    // Ensure that irrelevant features are not set
+    for (const auto &feature : irrelevantFeatures) {
+        target.features.reset(feature);
+    }
+
+    // Pick highest CUDA capability. Use fall-through to clear redundant features
+    switch (get_highest_cuda_capability(target)) {
+        case -1:
+            user_warning << "cannot merge cuda-enabled targets without knowing capability levels\n";
```
abadams: Should be case 20, and no warning, and just fallthrough

src/Target.cpp
```diff
@@ -753,6 +774,103 @@ int Target::natural_vector_size(const Halide::Type &t) const {
...
+            // These features assist in debugging.
+            Debug, TSAN, ASAN, MSAN,
+
+            // These features enable access to other devices.
+            CUDA, CUDACapability30, CUDACapability32, CUDACapability35, CUDACapability50, CUDACapability61, OpenCL,
+            OpenGL, OpenGLCompute, MinGW, Metal, HexagonDma, HVX_64, HVX_128, HVX_v62, HVX_v65, HVX_v66,
+            HVX_shared_object, D3D12Compute
+    };
+
+    static const std::vector<Feature> intersectionFeatures = {
```
abadams: Most of the features in the following two lists do not meaningfully affect the runtime and should just be stripped. Exceptions:

SoftFloatABI must match across all targets in the list, (like arch, os, bits)
NoNEON is a union feature

src/Target.cpp
```diff
@@ -753,6 +774,103 @@ int Target::natural_vector_size(const Halide::Type &t) const {
...
+        case 32:
+            target.features.reset(CUDACapability30);
+        default:
+            break;
+    }
+
+    // Check to see that there weren't two or more different HVX flags specified.
+    int number_of_hvx_flags_set = 0;
+    number_of_hvx_flags_set += (int)target.features.test(HVX_64);
+    number_of_hvx_flags_set += (int)target.features.test(HVX_128);
```
abadams: I believe it should be *exactly* one of {HVX_64, HVX_128}, and at *most* one of the HVX_v* features.

@dsharletg

src/Target.h
```diff
@@ -106,7 +106,7 @@ struct Target {
         FeatureEnd = halide_target_feature_end
     };
     Target() : os(OSUnknown), arch(ArchUnknown), bits(0) {}
-    Target(OS o, Arch a, int b, std::vector<Feature> initial_features = std::vector<Feature>())
+    Target(OS o, Arch a, int b, const std::vector<Feature> &initial_features = std::vector<Feature>())
```
abadams: I have a feeling that const ref args can't take rvals as their defaults on some older compilers. We had to add Scope::empty_scope to work around.

src/Target.cpp
dsharletg: HVX_64 and HVX_128 aren't required, but they should not both be specified.

HVX_v* are treated like CUDA compute capabilities (specifying HVX_v65 also implies HVX_v62).

abadams:
Also pinging @steven-johnson to see if this logic matches or could replace his internal Google build rules for generating runtimes. Would be a shame if this almost but didn't quite meet his needs.

src/Target.cpp
```diff
@@ -753,6 +772,78 @@ int Target::natural_vector_size(const Halide::Type &t) const {
     }
 }
 
+bool Target::get_runtime_compatible_target(const Target& other, Target &result) {
+    const std::array<Feature, 25> union_features = {
```
steven-johnson: (1) Some commenting about the logic here is appropriate.
(2) As new features are added, they may need to be added to one of these sets; at a minimum, we should add a comment to the declaration in HalideRuntime.h to warn people of this.

src/Target.cpp
```diff
@@ -753,6 +772,78 @@ int Target::natural_vector_size(const Halide::Type &t) const {
...
+    for (auto& feature : intersection_features) {
+        intersection_mask.set(feature);
+    }
+
+    if (arch != other.arch || bits != other.bits || os != other.os) {
+        user_warning << "runtime targets must agree on platform (arch-bits-os)\n";
+        return false;
+    }
+
+    if (has_feature(SoftFloatABI) != other.has_feature(SoftFloatABI)) {
```
steven-johnson: Is this really the only feature that must match? IMHO some of the features in the 'union' set make more sense here (e.g., specifying `debug` for some-but-not-all of the targets should be an error, not a quietly-add-debug-to-everything).

src/Target.h
steven-johnson: > on some older compilers

Any idea which, and if we still need to support them? (In hindsight, should have added comments about versions to e.g. the empty_scope fixes)

src/Target.h
```diff
@@ -189,6 +189,8 @@ struct Target {
       return !(*this == other);
     }
 
+    bool get_runtime_compatible_target(const Target& other, Target &result);
```
steven-johnson: This needs a Doxygen comment.

src/Target.cpp
abadams: ok, so we should add a must_match std::array, and put SoftFloatABI and the debugging flags (debug, msan, tsan, asan) in it.

steven-johnson:
The only internal logic for the Google builds that isn't in public Halide is some logic that ignores 'irrelevant' features when customizing a runtime (ie, features that shouldn't affect the contents of the runtime library). At present, that list is 

```
        "disable_llvm_loop_unroll",
        "disable_llvm_loop_vectorize",
        "no_asserts",
        "no_bounds_query",
        "profile",
        "strict_float",
        "user_context",
```

but that may not be exhaustive.

There is some existing logic that already attempts to deal with this for multitarget usage, in Module.cpp:601 (or thereabouts); separating this logic to allow standalone usage with `-r` is a good idea, but it would probably be useful to combine the logic bits to avoid possible logic drift (presumably by replacing the custom logic in Module.cpp with calls to this code).

Anyway, the main concern I have with this is that it's not clear to me that all of the features in the `union` and `intersection` list are correct choices; if we're truly choosing a GCD runtime, I'm not sure if we want a `union` list at all, as it suggests that we're going to be quietly adding features to the runtime that weren't specified. My instinct would be to instead assume that if a feature needs to be present for all of the variants to work, that feature should be explicitly listed in all of the subtargets (with compilation errors otherwise); yes, this puts more burden on the end user, but it means that they will be more informed about the runtime they need, and I'd argue that using the `-r` flag with this sort of thing should imply the necessary level of knowledge anyway.

(TL;DR: I think this code should only be in the business of removing features in order to achieve GCD; if it needs to add one, that's a user error. Am I missing a detail that makes that unworkable?)

abadams:
I disagree that we don't want a union list. Here's why:

IMO this should act as if we generated separate runtimes for each target and linked them using weak linkage. Anything that would cause ODR violations if we did this should go in the must_match list.

Under this goal, device apis need to be in the union list. If any one generator needs cuda, then the runtime should include support for it it, even if the other generators don't care.

abadams:
Also NoNEON needs to be in the union list because it's a negation of something that would go in the intersection list. I.e. if any generators in the list do not support NEON, then to be safe the runtime had better also not use NEON.

steven-johnson:
> IMO this should act as if we generated separate runtimes for each target and linked them using weak linkage. Anything that would cause ODR violations if we did this should go in the must_match list.... Under this goal, device apis [eg cuda] need to be in the union list.

OK, those are goals I can agree with. But we should document that as the explicit goal of the function, and add a caveat that it needs to be maintained conservatively (with items in the union list each being justified); I realize this may seem to be a picky request, but it's a lot easier to add items to the list later than to remove them after someone inadvertently depends on the behavior.

abadams:
Good point. We should put anything we can't immediately justify in the must_match bucket, because it's easy to move it to the union list later, but moves in the opposite directions will be hard.

alexreinking:
Okay, so what are the final lists going to be? I currently have 25 union features (all the device flags + NoNEON), 12 intersection features (instruction set extensions) and 5 matching features (SoftFloatABI + debug flags). All the rest of the features are discarded (eg. no_runtime, no_asserts)

abadams:
With the idea that it's easy to promote from the match list to the union or intersection lists later:

union:
CUDA, OpenCL, OpenGL, OpenGLCompute, Metal, D3D12Compute, NoNEON

intersection:
SSE41, AVX, AVX2, FMA, FMA4, F16C, ARMv7s,VSX, AVX512, AVX512_KNL, AVX512_Skylake, AVX512_Cannonlake, CUDACapability30, CUDACapability32, CUDACapability35, CUDACapability50, CUDACapability61, HVX_v62, HVX_v65, HVX_v66,

match:
SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX_64, HVX_128, MinGW, HexagonDma, HVX_shared_object

alexreinking:
Wait, why are the cuda capabilities in intersection? If we have `x86-64-linux-cuda-cuda_capability_50` and `x86-64-linux-cuda-cuda_capability_61`, what should the GCD be? I would have thought `x86-64-linux-cuda-cuda_capability_61`, which would correspond to union + max, not intersection (which would result in `x86-64-linux-cuda`, ie. capability 20)

steven-johnson:
If you're building for cuda50 and cuda60, it's likely because you want to use cuda60 and fall back to cuda50 if necessary. Thus you don't want the runtime to possibly require cuda60.

alexreinking:
So the final semantics should be union + min? That way we get `x86-64-linux-cuda-cuda_capability_50` as the gcd rather than CUDA 20?

abadams:
cuda_capability_*** is in the same bucket as sse vs avx vs avx2. Using it turns on increasingly advanced features which are only available on newer hardware. So yeah, take the min for the architecture feature flags.

alexreinking:
Okay, but an intersection operation wouldn't work in this case because we only set _one_ of the flags (the highest) in a target, not _all_ (right?). Union + min is the operation we need to perform to get a lower bound.

If we had Target always set all the lower cuda_capability bits then intersection would work. Would that be a change worth making?

abadams:
No, I guess take the min.

src/Target.cpp
dsharletg: HVX_v* flags are independent of Target::arch. If Target::arch is something other than Hexagon, it still affects the target architecture of offloaded code.

Target::HexagonDMA controls whether or not the Hexagon DMA runtime is included. This is currently only valid if Target::arch is Hexagon.

If none of the HVX_v* flags are set, it defaults to the minimum targeted architecture (v60).

HVX_shared_object looks like it is effectively deprecated, I don't see any uses of it. It used to control whether offloading generated an object or a shared object, but now we always generate a shared object.

FWIW, the motivation for #3673 is to clean up a lot of this (make it easier to deprecate/clean up old deprecated things).

alexreinking:
@dsharletg -- I'm operating under principle of least surprise here. If we discard lots of arch features, then compiling with `-r ... target=host,host` will _not_ compile with `host`. That behavior is really unsettling to me since `meet(x,x)` should `=x` up to equivalence classes over `x` that strip _truly_ irrelevant information (ie. `no_asserts`).

I don't really see any problem with Target being able to compute a meet with another Target (or perform other lattice-y operations). I'm actually not sure what would be a more appropriate location, besides maybe a set of non-member functions which then wouldn't get good implementations that use the bitset (unless they were friends?).

dsharletg:
I guess I should check my understanding: I thought this PR was to assist in compiling standalone runtimes via the '-r' flag of the generator, by choosing a target for the runtime that is something compatible with a list of targets supplied?

alexreinking:
Yes, that is the goal of this PR. While I agree that goal _can_ be achieved by discarding arch flags (eg. `avx2`), I argue above that it is better to preserve them to avoid counter-intuitive behavior. If for some reason there is a bug present for someone compiling with `target=host` but _not_ `target=host,host` (as generated by a script, say), it will likely be very difficult to debug.

steven-johnson:
Preserving `avx2` (& similar) when it is not specified across all the subtargets is an illegal-instruction waiting to happen; the avx2-specific runtime may not *require* avx2 instructions right now, but it's certainly *allowed* to do so. It is absolutely critical we avoid any change that could allow this.

dsharletg:
> Preserving `avx2` (& similar) when it is not specified across all the subtargets is an illegal-instruction waiting to happen; the avx2-specific runtime may not _require_ avx2 instructions right now, but it's certainly _allowed_ to do so. It is absolutely critical we avoid any change that could allow this.

I think that is what the PR does: it only adds arch flags if they are present in all the targets (they are "intersection" flags).

> Yes, that is the goal of this PR. While I agree that goal _can_ be achieved by discarding arch flags (eg. `avx2`), I argue above that it is better to preserve them to avoid counter-intuitive behavior. If for some reason there is a bug present for someone compiling with `target=host` but _not_ `target=host,host` (as generated by a script, say), it will likely be very difficult to debug.

I agree, that would be confusing, especially because I think this could be done to compile a pipeline (and not just a runtime). 

That's why I'm a little wary of this PR. I should have finished my thought on my previous comment: given that shared runtimes should only have the target architecture/OS, and a few other flags, I am wondering if it makes more sense to just require people to specify these manually.

I wonder if the thing we should have is a build helper function (bash script, Makefile, CMake, ... ugh) that computes a suitable target for a shared runtime from a list of targets.

Or, maybe we need a separate 'runtime_target' option for generators, where only the runtime_target supports this GCD feature, and the runtime_target *always* strips off these flags that shouldn't affect the runtime.

dsharletg:
The more I think about it, the more I'm leaning towards the 'runtime_target' option:

> Or, maybe we need a separate 'runtime_target' option for generators for use with the '-r' flag, where only the runtime_target supports this GCD feature, and the runtime_target _always_ strips off these flags that shouldn't affect the runtime (regardless of whether you use a GCD of multiple targets or not).

@steven-johnson what do you think?

alexreinking:
> Preserving `avx2` (& similar) when it is not specified across all the subtargets is an illegal-instruction waiting to happen; the avx2-specific runtime may not _require_ avx2 instructions right now, but it's certainly _allowed_ to do so. It is absolutely critical we avoid any change that could allow this.

This was never going to happen. The instruction flags are kept only if both (by extension, all) targets have them.

> I wonder if the thing we should have is a build helper function (bash script, Makefile, CMake, ... ugh) that computes a suitable target for a shared runtime from a list of targets.

I strongly oppose this. That would necessitate duplicating the host feature detection logic to be able to correctly merge things like `host,x86-64-linux-cuda-cuda_capability50`.

> Or, maybe we need a separate 'runtime_target' option for generators for use with the '-r' flag, where only the runtime_target supports this GCD feature, and the runtime_target _always_ strips off these flags that shouldn't affect the runtime (regardless of whether you use a GCD of multiple targets or not).

I don't see why this is different from `-r .. target`. Especially if using `-r ... target=[one option]` has the same behavior as `-r ... runtime_target=[one option]`. If it's about communication, I guess we could put some clarifying info in the generator's help text.

steven-johnson:
`-runtime_target` seems like safer alternative, but I guess I want to back up for a minute and just get clarity on the motivation for this target -- I presume the idea here is that we want to be able to build a multitarget without a runtime, and have an automated way of computing a 'safe' runtime that can be built separately.

I think the big thing that's missing here is that there isn't currently any well-defined set of rules for what runtime feature(s) are required/forbidden for a given set of targets; maybe a better first step here would be to start by defining such a set of rules by fiat:

```
// Return true if a runtime built for 'this' can be safely used for *all* of the targets in 'targets'
bool Target::is_runtime_compatible_with_all(vector<Target> targets);`
```

This could be used to verify that an arbitrary runtime choice was safe, e.g. we could call `generator -r some-target --ensure_runtime_compatibility target1,target2,target3`, which would fail if `some_target` didn't pass this check.

alexreinking:
> `-runtime_target` seems like safer alternative, but I guess I want to back up for a minute and just get clarity on the motivation for this target -- I presume the idea here is that we want to be able to build a multitarget without a runtime, and have an automated way of computing a 'safe' runtime that can be built separately.

Yes. If we're building multiple pipelines into a single static lib, it should be possible to compute a compatible runtime for all of them (a meet).

> I think the big thing that's missing here is that there isn't currently any well-defined set of rules for what runtime feature(s) are required/forbidden for a given set of targets; maybe a better first step here would be to start by defining such a set of rules by fiat:
> ```
> // Return true if a runtime built for 'this' can be safely used for *all* of the targets in 'targets'
> bool Target::is_runtime_compatible_with_all(vector<Target> targets);`
> ```

Why is this different than:

```
bool Target::is_runtime_compatible_with_all(vector<Target> targets) {
    for (const auto& target : targets) {
        if (!is_runtime_compatible_with(target)) {
            return false;
        }
    }
    return true;
}
```

>This could be used to verify that an arbitrary runtime choice was safe, e.g. we could call `generator -r some-target --ensure_runtime_compatibility target1,target2,target3`, which would fail if `some_target` didn't pass this check.

This seems like something that's more useful for testing an implementation of GCD for targets. ie. for all targets t1, t2, let tGcd = meet(t1, t2), assert `tGcd.is_runtime_compatible_with(t1)`, assert `tGcd.is_runtime_compatible_with(t1)`. I don't really see how it would fit into a build process that wants to automatically select a runtime.

src/Target.cpp
```diff
@@ -165,9 +166,47 @@ Target calculate_host_target() {
...
+    }
+    if (t.has_feature(Target::CUDACapability61)) {
+        return 61;
+    }
+    return 20;
+}
 
-    return Target(os, arch, bits, initial_features);
+int get_hvx_lower_bound(const Target &t) {
+    if (!t.has_feature(Target::HexagonDma) && t.arch != Target::Hexagon) {
```
pranavb-ca: You shouldn't  need to check ``Target::HexagonDma`` in relation with hvx_lower_bound. For compiling a runtime on hexagon ``t.arch`` will _never_ perhaps be anything other than ``Target::Hexagon`` and so, you'll not likely satisfy this condition. I think it is simply safe to leave ``Target::HexagonDma`` in the union list.

alexreinking: It would be really useful if someone could explain the various Hexagon flags / arch setting to me. I have never worked with Hexagon, ever, so I'm well aware there are going to be issues with the implementation until someone helps me fix it.

pranavb-ca: Sure. 

- HVX_64 and HVX_128 allow you to choose between 64 byte vectors and 128 byte vectors. Only **one of these** should be present
- HVX_v62, HVX_v65 and HVX_v65 are successive ISA generations, each a super set of generations before it. So, a tighter (lower) bound is the right thing to do.
- HVX_shared_object - I think it is best to remove this altogether. I don't believe it is used any more.
- HexagonDma - This includes runtime functions to support Direct Memory Access functionality and as such is decoupled from the choice of HVX_v**.

alexreinking: Thanks, @pranavb-ca -- how are these flags affected by the presence / absence of hexagon as the arch? Are these flags even valid if `arch == Linux`, say?

alexreinking: Also, is there a generation _before_ HVX_v62? Is it valid to have none of those three flags set? If so, what does it mean?

pranavb-ca: Typical target values we use most often for compiling Hexagon runtimes are
``hexagon-32-qurt-hvx_128`` or ``hexagon-32-qurt-hvx_128-hexagon_dma`` (or the same ones with ``hvx_64`` in place of ``hvx_128``

pranavb-ca: > Also, is there a generation _before_ HVX_v62, analogous to CUDA Capability 20? Is it valid to have none of those three flags set? If so, what does it mean?

The Hexagon backend in Halide defaults to v60. It isn't a separate target feature though. So, haveing v62,65, 66 reset means you get v60.

alexreinking: Thanks so much again! Very helpful. Could you answer my questions about other architectures? Is it valid to have `x86-64-linux-hvx_128-hexagon_dma`?

pranavb-ca: > Thanks, @pranavb-ca -- how are these flags affected by the presence / absence of hexagon as the arch? Are these flags even valid if `arch == Linux`, say?

``os == Linux`` when ``arch == Hexagon`` will lead to problems because the OS on Hexagon is ``QuRT``. So, it doesn't have posix like functionality. However, you won't be changing any behaviour by not handling this bad combination.

Typically Generators can be compiled with ``Arch != Target::Hexagon``, for eg. ``arm-64-android-hvx_128``. This simply means that the pipeline will be offloaded to the Hexagon however, even in these cases the runtime is compiled using ``hexagon-32-qurt-hvx_128`` (See HexagonOffload.cpp)

alexreinking: Okay, it looks like the compilation process will generate a runtime if necessary using the features on the target. So we can merge `arm-64-android` and `arm-64-android-hvx_128` to get the latter? And `arm-64-android` does not imply hexagon v60, it's just no hexagon in the absence of any other hexagon flag?

alexreinking: Oh wait, any use of Hexagon requires exactly one of `HVX_64` or `HVX_128` to be set. I can just check that, right?

pranavb-ca: > Oh wait, any use of Hexagon requires exactly one of `HVX_64` or `HVX_128` to be set. I can just check that, right?

Yes

alexreinking: Awesome -- could you take another look at the code? I think it matches the description you gave.

src/Target.cpp
alexreinking: Thanks @dsharletg -- I think we can take as a precondition for GCD that the input Targets are valid, though it might be useful to also add a `bool Target::is_valid()` that could be used for sanity checking / assertions. So I'll leave HexagonDMA as a `must_match`.

src/Target.cpp
dsharletg: > > Oh wait, any use of Hexagon requires exactly one of `HVX_64` or `HVX_128` to be set. I can just check that, right?

> Yes

Are you sure about this? It might not be a typical usage of Halide, but I think it should be totally possible to build code with arch == Hexagon and no HVX flags.

alexreinking: @dsharletg - what happens in that case?

pranavb-ca: > > > Oh wait, any use of Hexagon requires exactly one of `HVX_64` or `HVX_128` to be set. I can just check that, right?
> 
> > Yes
> 
> Are you sure about this? It might not be a typical usage of Halide, but I think it should be totally possible to build code with arch == Hexagon and no HVX flags.

Sorry, I misread/misunderstood Alex's comment- I was talking about how they should be mutually exclusive. Neither is necessary. ``arch == Target::Hexagon`` implies v60. Although not intended, but Alex's code still ensures this as the absence of HVX_64 and HVX_128 results in teh resetting of 62,65 and 66. Leaving a default of v60 that is ensured by CodeGen_Hexagon.cpp.

However, to be strictly correct we should also check if ``arch == Hexagon``

pranavb-ca: aargh. @alexreinking  - I am really sorry for making you go over this again :(

@dsharletg  is absolutely right. To `return 0` from ``get_hvx_lower_bound`` your check needs to be
``if (!t.has_feature(Target::HVX_64) && !t.has_feature(Target::HVX_128) && t.arch != Target::Hexagon)``

alexreinking: @dsharletg, @pranavb-ca -- Is this target valid: `arm-32-android-hvx_v62`? If so, is `hvx_64` or `hvx_128` set? What about when we have `hexagon-32-qurt-hvx_v62`?

pranavb-ca: > @dsharletg - what happens in that case?

So, v60, v61 etc are Scalar ISA versions. We don't have these features in Halide because Halide for the Scalar ISA is not interesting. hvx_v6x is Scalar ISA + Hexagon Vector Extensions.

So,  ``hvx_64`` and ``hvx_128`` are useful really only if you use ``.vectorize()``
And a target like ``hexagon-32-qurt-hvx_v62` will result in v62 scalar code.

pranavb-ca: > @dsharletg, @pranavb-ca -- Is this target valid: `arm-32-android-hvx_v62`? If so, is `hvx_64` or `hvx_128` set? What about when we have `hexagon-32-qurt-hvx_v62`?

For ``arm-32-android-hvx_v62`` - This is valid and should result in v62 . If the hexagon target is instantiated, which in this case will happen, then hvx_v64 is the default. I think we should change this default, but that is outside the scope of this discussion. So effectively, you get ``hvx_64-hvx_v62``

``hexagon-32-qurt-hvx_v62`` - Again, hvx_v64 default and v62 ISA.

alexreinking: So the check should _really_ look for _any_ of the `hvx_*` flags or `arch == Hexagon`

alexreinking: Actually, does `arm-32-android-hexagon_dma` equal `arm-32-android-hvx_64-hexagon_dma` with HVX_v60?

alexreinking: Should it really be:

```
bool is_using_hexagon(const Target &t) {
    return t.has_feature(Target::HVX_64)
           || t.has_feature(Target::HVX_128)
           || t.has_feature(Target::HVX_v62)
           || t.has_feature(Target::HVX_v65)
           || t.has_feature(Target::HVX_v66)
           || t.has_feature(Target::HexagonDma)
           || t.has_feature(Target::HVX_shared_object)
           || t.arch == Target::Hexagon;
}

int get_hvx_lower_bound(const Target &t) {
    if (!is_using_hexagon(t)) {
        return 0;
    }
    if (t.has_feature(Target::HVX_v62)) {
        return 62;
    }
    if (t.has_feature(Target::HVX_v65)) {
        return 65;
    }
    if (t.has_feature(Target::HVX_v66)) {
        return 66;
    }
    return 60;
}
```

pranavb-ca: > Should it really be:
> 
> ```
> bool is_using_hexagon(const Target &t) {
>     return t.has_feature(Target::HVX_64)
>            || t.has_feature(Target::HVX_128)
>            || t.has_feature(Target::HVX_v62)
>            || t.has_feature(Target::HVX_v65)
>            || t.has_feature(Target::HVX_v66)
>            || t.has_feature(Target::HexagonDma)
>            || t.has_feature(Target::HVX_shared_object)
>            || t.arch == Target::Hexagon;
> }
> 
> int get_hvx_lower_bound(const Target &t) {
>     if (!is_using_hexagon(t)) {
>         return 0;
>     }
>     if (t.has_feature(Target::HVX_v62)) {
>         return 62;
>     }
>     if (t.has_feature(Target::HVX_v65)) {
>         return 65;
>     }
>     if (t.has_feature(Target::HVX_v66)) {
>         return 66;
>     }
>     return 60;
> }
> ```

Yes (I should submit a PR to remove HVX_shared_object)

alexreinking: > Yes

Excellent. I just pushed

> (I should submit a PR to remove HVX_shared_object)

I look forward to it!

alexreinking:
@dsharletg - I was talking a bit with @jrk about your comments:

>I also think that _most_ of the flags we are doing a lot of work to deal with really should not matter:
> _[...]_
> That just leaves a hodge podge of misc flags, and I don't think any of them are "intersection" flags. Just a few "union" flags, and a few "match" flags.
> _[...]_
> I agree, that would be confusing, especially because I think this could be done to compile a pipeline (and not just a runtime).
> That's why I'm a little wary of this PR. I should have finished my thought on my previous comment: given that shared runtimes should only have the target architecture/OS, and a few other flags, I am wondering if it makes more sense to just require people to specify these manually.

I think I see what you're getting at, but please tell me if I'm wrong. I see two issues. (1) is that runtimes are specified using targets and it is questionable whether all of that detail is necessary. (2) is whether or not we want to be in the business of computing meets in the generators at all (the subject of this PR).

As I see (2), I think we _should_ be, and I touched on this with my comment about duplicating host detection. I've been trying to integrate Halide into msbuild for a project that _will not_ add CMake to their build. In trying to compute these targets externally, I ran into a lot of issues regarding `host` (and spent a lot of @abadams's patience -- thank you!) and it makes sense to me that we should be able to reuse the logic that's already in the compiler. I believe doing this externally would be a maintenance nightmare since it would need to be kept up to date with Halide and would likely go stale.

If you know of some reason I might be wrong about this, I'm open to hearing it.

As for (1), I don't know the right answer to this. I'm of the opinion that whatever we do, some basic invariants should be kept (ie. `target=host,host` should be the same as `target=host`). One way to discard not-useful flags (ie. instruction set flags) in a sound way would be to do so unconditionally. Take the GCD and then discard. Or discard first and then take the GCD (it ought to be the same either way). Then someone could reuse the GCD logic to compute pipeline targets for some other purpose, but the feature I _think_ you want could be built on this more general foundation.

Could I please get your thoughts on this?

src/Target.h
alexreinking: Any update on this? I went ahead and tested a simple function that uses the same construction on Godbolt

```
#include <vector>

int test(int a, const std::vector<int>& b = std::vector<int>()) {
    for (const auto& x : b) {
        a += x;
    }
    return a;
}
```

This is accepted by the following compilers:

[g++ >= 4.7.1](https://godbolt.org/z/w-pJbV) (earlier versions reject `-std=c++11`)
[clang >= 3.4.1](https://godbolt.org/z/OCwJ1_) (earlier versions fail to find the `vector` header... an error on Goldbot's part?)
[msvc >= 19.14](https://godbolt.org/z/JpormX) (no earlier versions for me to test)
[icc >= 13.0.1](https://godbolt.org/z/YneUDC) (no earlier versions for me to test)

It also looks like it works with early versions of ellcc (which I think is based on clang?), djggp (which I think is based on gcc?), and power64(le) AT12.0.

steven-johnson: At this point I think this is a safe construction on all the compilers we're actually using.

src/Target.cpp
```diff
@@ -753,6 +799,88 @@ int Target::natural_vector_size(const Halide::Type &t) const {
     }
 }
 
+bool Target::get_runtime_compatible_target(const Target& other, Target &result) {
```
steven-johnson: Since at least one caller calls with 'this' and 'result' referring to the same instance, we should document that as supported behavior, and double-check this code to be sure there are no issues. (Might be simplest to construct the result in a local tmp and assign to 'result' at the end, since this object is small and cheap.)

alexreinking: @steven-johnson - done! just pushed :)

alexreinking: It's too bad that `std::optional` isn't available in C++11, though! That would certainly be a much cleaner interface to this function.

dsharletg:
> As for (1), I don't know the right answer to this. I'm of the opinion that whatever we do, some basic invariants should be kept (ie. `target=host,host` should be the same as `target=host`). One way to discard not-useful flags (ie. instruction set flags) in a sound way would be to do so unconditionally. Take the GCD and then discard. Or discard first and then take the GCD (it ought to be the same either way). Then someone could reuse the GCD logic to compute pipeline targets for some other purpose, but the feature I _think_ you want could be built on this more general foundation.

I agree that if we have a proper GCD, that `GCD(host, host)` should equal `host`. The reason I like the idea of `runtime_target=` being the only thing supporting multiple targets is that:
- `runtime_target` can be explicitly documented to ignore most target feature flags.
- Then it is only necessary to make the GCD operation deal with a small subset of the feature flags (so it would be a "weak" GCD or something).

One way to look at it is to compare the documentation needed to explain both options:

IMHO, the full general GCD option needs quite a lot of explaining. Maybe it doesn't *need* it, but people could easily have a ton of questions about it.

On the other hand, the `runtime_target` supporting multiple targets is *much* easier to explain: a `runtime_target` generates a runtime that is compatible with all specified targets. The lower level details beyond that (probably) don't need to be explained in much detail.


That said: my opinion on which way to go is not that strong. I just wanted to float the idea and make sure we're on the same page on what it would look like. I think @steven-johnson is the right person to make the call on what to do here.

zvookin:
Please add some text to the gengen usage string describing the RUNTIME_NAME syntax and behavior. That -r is used to generate a standalone runtime should have been given its own line before, but with this added behavior it definitely needs to be documented. (And we really should have all of the usage info in Doxygen somewhere too.)

alexreinking:
> Please add some text to the gengen usage string describing the RUNTIME_NAME syntax and behavior. That -r is used to generate a standalone runtime should have been given its own line before, but with this added behavior it definitely needs to be documented. (And we really should have all of the usage info in Doxygen somewhere too.)

That's a great point @zvookin -- I'll get right on that.

alexreinking:
@dsharletg -- I think I'm missing the philosophical difference between `-r ... target=` and `-r ... runtime_target=`. Why can't we just explain that the behavior for targets is somewhat simplified when applied to runtimes?

I see a few options for adjusting the behavior of `-r`:

1) Do nothing. Reject this PR. `target=` keeps its current behavior of compiling _with_ irrelevant flags (those that we can always drop from a runtime, eg. `avx2`) and rejecting lists of targets longer than n=1.
2) Add `runtime_target=` that is only honored when `-r` is specified. It drops all irrelevant flags before computing a GCD (or after, it's the same). It is incompatible with `target=` and is an error to specify both.
3)  (this PR) Modify `target=` so that it accepts longer lists of targets and computes a GCD for them. Its behavior agrees with the existing functionality because it computes a true GCD.
4) Modify `target=` so that it behaves exactly like `runtime_target=` in (2), ie. it always drops flags.

I'm skeptical of (2) since keeping both parameters (`target` and `runtime_target`) is confusing to me. Why give users the option to compile with irrelevant flags only if they specify one target with a special param? The two parameters would have to behave differently on n=1 (else `runtime_target=host,host` is unexpected). The other three options make more sense to me, though I obviously like (1) less 😉 

Let me know if there are more options, or which one makes the most sense to you.

>On the other hand, the `runtime_target` supporting multiple targets is _much_ easier to explain: a `runtime_target` generates a runtime that is compatible with all specified targets. The lower level details beyond that (probably) don't need to be explained in much detail.

I see communicating how things work to the user as being orthogonal to how they're implemented. Both options (3) and (4) could be explained in almost the same words you used: "when compiling with `-r`, `target` generates a runtime that is compatible with all specified targets.".

src/Generator.cpp
```diff
@@ -977,17 +977,28 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
...
-            return 1;
-        }
         std::string base_path = compute_base_path(output_dir, runtime_name, "");
-        Outputs output_files = compute_outputs(targets[0], base_path, emit_options);
-        compile_standalone_runtime(output_files, targets[0]);
+
+        Target gcd_target = targets[0];
+        for (size_t i = 1; i < targets.size(); i++) {
+            if (!gcd_target.get_runtime_compatible_target(targets[i], gcd_target)) {
+                user_error << "Failed to find compatible runtime target for "
```
steven-johnson: nit: it may be better to print all of the targets here?

src/Target.cpp
steven-johnson: I suspect we'll be able to move to C++17 within a year or so :-)

src/Target.cpp
```diff
@@ -753,6 +799,89 @@ int Target::natural_vector_size(const Halide::Type &t) const {
...
+    // We merge the bits via bitwise or.
+    Target output = Target{os, arch, bits};
+    output.features = ((features | other.features) & union_mask)
+            | ((features | other.features) & matching_mask)
+            | ((features & other.features) & intersection_mask);
+
+    // Pick tight lower bound for CUDA capability. Use fall-through to clear redundant features
+    switch (get_cuda_capability_lower_bound(output)) {
+        default: // no CUDA feature; clear all capability flags
+        case 20: output.features.reset(CUDACapability30);
```
steven-johnson: these should have comments about 'deliberately falling thru', since I presume that's what's happening here.

alexreinking:
@zvookin -- which doxygen comments should I update? The ones in Generator.h don't deal with the command line at all. And this PR doesn't affect the targets as the generators see them.

abadams:
kUsage in generate_filter_main in Generator.cpp

alexreinking:
> kUsage in generate_filter_main in Generator.cpp

I updated that in my latest commit, already. I was having trouble finding the doxygen comments, not gengen's help text 🙂

src/Generator.cpp
alexreinking: My thought was that it would be best to highlight (a) what was computed so far and (b) what failed to unify with it. All of the rest of the targets are already available on the command line, no?

src/Target.cpp
alexreinking: Is that not what "Use fall-through to clear redundant features" is doing? Should I add comments after each line like:

    // Pick tight lower bound for CUDA capability. Use fall-through to clear redundant features
    switch (get_cuda_capability_lower_bound(output)) {
        default: // no CUDA feature; clear all capability flags
        case 20: output.features.reset(CUDACapability30); // fall-thru
        case 30: output.features.reset(CUDACapability32); // fall-thru
        case 32: output.features.reset(CUDACapability35); // fall-thru
        case 35: output.features.reset(CUDACapability50); // fall-thru
        case 50: output.features.reset(CUDACapability61); // fall-thru
        case 61: break;
    }

I personally think that's a little noisy given the comment right in front of the switch.

steven-johnson: I'd err on the side of noise; having fallthru be possible by accident is a misfeature of the language. (Google-internal code requires a specific annotation for deliberate fallthru; omitting it will trigger linter errors.)

alexreinking: Okay, I can add those, then!

PR #3672, opened 2019-02-17
Gpu camera pipe

abadams:
Add a reasonable GPU schedule for camera pipe.

### DIFF
--- a/apps/camera_pipe/camera_pipe_generator.cpp
+++ b/apps/camera_pipe/camera_pipe_generator.cpp
@@ -153,7 +153,18 @@ public:
     void schedule() {
         Pipeline p(output);
 
-        if (!auto_schedule) {
+        if (auto_schedule) {
+            // blank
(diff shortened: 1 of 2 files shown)


apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -426,10 +447,52 @@ void CameraPipe::generate() {
...
+                .bound(y, 0, (out_height / 2) * 2);
+        }
+
+        Var xi, yi, xii, xio;
+
+        /* 1391us on a gtx 980. */
+        processed.compute_root()
+            .reorder(c, x, y)
+            .unroll(x, 2)
+            .gpu_tile(x, y, xi, yi, 28, 12);
```
dsharletg: Might be good to briefly comment where this tile size came from (so things fit in shared memory?).

steven-johnson:
Ready to land?

PR #3673, opened 2019-02-19
Remove explicit target feature numbering

dsharletg:
This PR removes the explicit target feature numbers from the `halide_target_feature_t` enum:
- This ordering is well defined by the C/C++ standard (even without the first zero defined enum value).
- I don't think we support binary compatibility of this compiled code across compiler versions, so inserting/removing enum values shouldn't be an issue.
- Any edits to these features requires either painful renumbering, or 'last added' sorting, both of which are annoying.

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -152,13 +152,16 @@ public:
 /** Link a set of modules together into one module. */
 Module link_modules(const std::string &name, const std::vector<Module> &modules);
 
-/** Create an object file containing the Halide runtime for a given
- * target. For use with Target::NoRuntime. */
+/** Create an object file containing the Halide runtime for a given target. For
+ * use with Target::NoRuntime. Standalone runtimes are only compatible with
+ * pipelines compiled by the same build of Halide used to call this function. */
 void compile_standalone_runtime(const std::string &object_filename, Target t);
 
-/** Create an object and/or static library file containing the Halide runtime for a given
- * target. For use with Target::NoRuntime. Return an Outputs with just the actual
- * outputs filled in (typically, object_name and/or static_library_name).
+/** Create an object and/or static library file containing the Halide runtime
+ * for a given target. For use with Target::NoRuntime. Standalone runtimes are
+ * only compatible with pipelines compiled by the same build of Halide used to
+ * call this function. Return an Outputs with just the actual outputs filled in
+ * (typically, object_name and/or static_library_name).
  */
 Outputs compile_standalone_runtime(const Outputs &output_files, Target t);
 
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -761,7 +761,6 @@ void target_test() {
         t.set_feature(feature.second);
     }
     for (int i = 0; i < (int)(Target::FeatureEnd); i++) {
-        if (i == halide_target_feature_unused_23) continue;
         internal_assert(t.has_feature((Target::Feature)i)) << "Feature " << i << " not in feature_names_map.\n";
     }
     std::cout << "Target test passed" << std::endl;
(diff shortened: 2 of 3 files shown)


steven-johnson:
> * I don't think we support binary compatibility of this compiled code across compiler versions, so inserting/removing enum values shouldn't be an issue.

Might be worth adding an explicit comment to this effect.

dsharletg:
Where do you suggest putting that comment?

FWIW I think the same comment applies to a *lot* of stuff in Halide.

steven-johnson:
> Where do you suggest putting that comment?

In the doxygen comment at the top of the declaration.

> FWIW I think the same comment applies to a _lot_ of stuff in Halide.

True, and yet this is one people seem to get stuck on thinking should remain stable, so we should reiterate the point.

dsharletg:
I looked at this for a while because I really don't think it makes sense to put this comment on the enumeration. It both is unnecessary information for almost everyone reading that comment, and yet at the same time it is not enough information in the right place for someone that really needs to know this.

I decided to put it in Module.h, on `compile_standalone_runtime`. My reasoning is that if you aren't using a standalone runtime, then this really isn't a problem for you, because you are compiling a runtime into all your pipelines anyways. If you are using a standalone runtime, then you'll hopefully read this comment when you learn how to use compile_standalone_runtime.

zvookin:
The main reason for the numbers to be written out is debugging. We also do expose the enum via ```halide_can_use_target_features``` which is C only runtime. I don't have a strict objection to removing the numbering and allowing the values to change somewhat, but it does seem a bit hypocritical for us to hardcode binary values from lower level APIs in order to avoid including their headers, and then not to extend some degree of that ability upward. Obviously we're not going to guarantee the entire set of bits for the whole C++ API to the compiler, etc, but things that are visible to the C only runtime API should stay as stable as we can make them.

steven-johnson:
> I decided to put it in Module.h, on `compile_standalone_runtime`

I can live with this.

dsharletg:
For those lower level APIs, they are guaranteeing binary compatibility, by necessity. Us hardcoding the values of their enumerations doesn't change that, they need a stable binary interface for apps to communicate with drivers. No one from these products is going to come along and tell us we can't hardcode their enums :)

On the other hand, I don't think we claim to provide a stable binary interface anywhere, and it's hard for me to think of a use case where it is necessary. I think the C runtime API is necessary just to enable a stable interface between compilers (us generating code with LLVM, users potentially using any compiler), but that doesn't require stability across versions.

steven-johnson:
Do we have consensus on landing this?

abadams:
I'm weakly ambivalent. On one hand, any minor quality of life changes to maintaining the compiler that delete lines of code have my support. On the other hand Target features are added very rarely, and I really don't think maintaining the numbering is hard. The question is: Is there any reasonable usage scenario which actually relies on this being stable across versions? I believe the target enum is only used at runtime in feature-sniffing code, so linking a filter generated with one version of Halide against an older runtime could get the features wrong (or vice-versa). I think that would be the least of your problems though if you're mixing versions between runtime and filters. We typically only keep the runtime stable to make life easier for people overriding runtime functions, not for binary compatibility across different runtime and filter versions. @slomp would doing this mess anything up at Adobe?

dsharletg:
> I think that would be the least of your problems though if you're mixing versions between runtime and filters.

This is my thinking as well. If nothing else, we don't claim to support this... if it actually works and someone is using this functionality now, we should improve our documentation around this.

There are two reasons I wanted to make this change:
- I wanted to clean up the feature enum (reorder things to be nearby other relevant things, etc.), which would involve manually renumbering it. Assuming what I said above is correct, then the more expedient thing to do was just remove the explicit numbers.
- Any local changes to this enum generate merge conflicts with local changes to this enum, which happens often for me.

steven-johnson:
Are we ready to land this, or do we need more discussion?

steven-johnson:
Since this is approved, I'm going to land it if I don't hear objections in the next 24 hours.

steven-johnson:
24 hours. Landing.

PR #3674, opened 2019-02-19
Fix Makefile building of halide.tgz

steven-johnson:
We weren't including everything from the distrib/ folder in the tgz. Now we should be.


steven-johnson:
attn @pklos_gitlab

PR #3678, opened 2019-02-21
Cuda sticky allocator

abadams:
Opt-in allocation cache for cuda. Speeds up local laplacian and bilateral grid by 3x!

I don't have a good eviction policy yet though. The current limit is just a max of 32 unused allocations, regardless of their size. I tried a high-water mark of memory usage so far, but it was too aggressive at freeing memory, and isn't great for a long-running process anyway. Suggestions welcome.


src/runtime/cuda.cpp
```diff
@@ -136,7 +147,7 @@ WEAK int halide_cuda_acquire_context(void *user_context, CUcontext *ctx, bool cr
         }
 
         {
-            ScopedSpinLock spinlock(&context_lock);
+            ScopedMutexLock spinlock(&context_lock);
```
dsharletg: @zvookin might be interested in this change

src/runtime/cuda.cpp
```diff
@@ -211,11 +222,10 @@ class Context {
...
     INLINE ~Context() {
-        if (error == 0) {
-            CUcontext old;
-            cuCtxPopCurrent(&old);
-            halide_cuda_release_context(user_context);
-        }
+        CUcontext old;
+        cuCtxPopCurrent(&old);
+
+        halide_cuda_release_context(user_context);
```
dsharletg: I think only this should come out of the `if (error == 0)`

src/runtime/cuda.cpp
```diff
@@ -552,19 +605,48 @@ WEAK int halide_cuda_device_free(void *user_context, halide_buffer_t* buf) {
...
     uint64_t t_before = halide_current_time_ns(user_context);
     #endif
 
     halide_assert(user_context, validate_device_pointer(user_context, buf));
 
-    debug(user_context) <<  "    cuMemFree " << (void *)(dev_ptr) << "\n";
-    CUresult err = cuMemFree(dev_ptr);
-    // If cuMemFree fails, it isn't likely to succeed later, so just drop
-    // the reference.
+    CUresult err = CUDA_SUCCESS;
```
dsharletg: If possible, I think this should be in a helper function wrapper around cuMemAlloc/cuMemFree, even if this helper function is not something that is overrideable a custom runtime (re: our gitter conversation about this).

src/runtime/cuda.cpp
```diff
@@ -688,16 +773,85 @@ WEAK int halide_cuda_device_malloc(void *user_context, halide_buffer_t *buf) {
...
-    debug(user_context) << "    cuMemAlloc " << (uint64_t)size << " -> ";
-    CUresult err = cuMemAlloc(&p, size);
-    if (err != CUDA_SUCCESS) {
-        debug(user_context) << get_error_name(err) << "\n";
-        error(user_context) << "CUDA: cuMemAlloc failed: "
-                            << get_error_name(err);
-        return err;
-    } else {
-        debug(user_context) << (void *)p << "\n";
+    CUdeviceptr p = 0;
```
dsharletg: Same comment as on the free side.

src/runtime/cuda.cpp
```diff
@@ -669,7 +754,7 @@ WEAK int halide_cuda_device_malloc(void *user_context, halide_buffer_t *buf) {
         return ctx.error;
     }
 
-    size_t size = buf->size_in_bytes();
+    size_t size = quantize_allocation_size(buf->size_in_bytes());
```
dsharletg: Only want to do this if caching allocations

src/runtime/cuda_functions.h
```diff
@@ -56,6 +56,7 @@ CUDA_FN_4000(CUresult, cuCtxPopCurrent, cuCtxPopCurrent_v2, (CUcontext *pctx));
 CUDA_FN(CUresult, cuPointerGetAttribute, (void *result, int query, CUdeviceptr ptr));
 
 CUDA_FN_OPTIONAL(CUresult, cuStreamSynchronize, (CUstream hStream));
+CUDA_FN_OPTIONAL(CUresult, cuLaunchHostFunc, (CUstream hStream, void (*fn)(void *), void *));
```
dsharletg: Don't see this referenced anywhere else in this PR?

ongjunjie:
I'm watching this PR with great interest. I'm also seeing 3x improvements on some of my filters :)

Some feedback on JIT usage:
- Having to call an internal function is a little weird
- It's not immediately obvious that you must have JIT compiled the shared runtime before calling `halide_reuse_device_allocations()`

steven-johnson:
Given the 3x performance increase in some cases, clearly this PR is still interesting; is the lack of an good eviction policy the gating factor?

abadams:
Yes, on one hand the performance win seems like a must-have, but on the other this just leaves memory allocated indefinitely, and I don't have a good fix for that. There has to be something more clever we could do. E.g. somehow enqueing a callback so that we start freeing things when Halide hasn't used the GPU for a while.

abadams:
In the absence of something cleverer, I think we should merge this as-is. I think I hit the outstanding comments. I just encountered a case where someone concluded Halide was bad for cuda. Their benchmark was spending a lot of its time doing malloc/free.

steven-johnson:
SGTM, but can we go ahead and open an issue to track "may need good eviction policy", and annotate in the relevant chunk of code?

abadams:
#4093

PR #3680, opened 2019-02-24
Refactor is_parallel helper functions

dsharletg:
This PR does a small cleanup of code that checks if a loop is parallel:
- Adds `is_unordered_parallel` and `is_parallel` helper functions
- Renames existing `is_parallel` helper functions on `For` and `Dim` and adds `is_unordered_parallel`
- Possible small bug fix in ScheduleFunctions.cpp, was this check missing `ForType::GPULane`? @abadams

### DIFF
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -422,6 +422,19 @@ enum class ForType {
     GPULane,
 };
 
+/** Check if for_type executes for loop iterations in parallel and unordered. */
+inline bool is_unordered_parallel(ForType for_type) {
+    return (for_type == ForType::Parallel ||
+            for_type == ForType::GPUBlock ||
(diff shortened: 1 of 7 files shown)


src/Expr.h
```diff
@@ -422,6 +422,19 @@ enum class ForType {
...
 };
 
+/** Check if for_type executes for loop iterations in parallel and unordered. */
+inline bool is_unordered_parallel(ForType for_type) {
+    return (for_type == ForType::Parallel ||
+            for_type == ForType::GPUBlock ||
+            for_type == ForType::GPUThread);
+}
+
+/** Returns true if for_type executes for loop iterations is parallel. */
```
abadams: is parallel -> in parallel

Issue #3682, opened 2019-02-26
Supporting Unicode builds on Windows


alexreinking:
In `Generator.cpp:868`, there is the following code:

```
#ifdef _WIN32
        if (LoadLibrary(lib.c_str()) != nullptr) {
            cerr << "Failed to load: " << lib << "\n";
            return 1;
        }
#else
```

That `LoadLibrary` macro is designed to select either `LoadLibraryA` and `LoadLibraryW` depending on whether or not `-DUNICODE` and/or `-D_UNICODE` are/is set. But since `lib.c_str()` returns a `const char*`, it is only compilable with the "A" (ANSI) variant. [It is preferable to use UTF-8 everywhere](http://utf8everywhere.org/#windows) but still compile in Unicode mode to prevent UTF-8 characters from leaking into ANSI-only functions.

As it stands, this code will fail to compile with Unicode support and without it, it will fail if given a path name that contains foreign characters (for example, a user's home directory).

For this code to be correct, we either need to directly call `LoadLibraryA` and declare our lack of support for plugins that use non-ANSI paths or we need to compile with Unicode on Windows and use [`MultiByteToWideChar`](https://docs.microsoft.com/en-us/windows/desktop/api/stringapiset/nf-stringapiset-multibytetowidechar) to do the encoding conversion ourselves. I think the second option is preferable. There is also the Boost.Nowide library that provides good functions to help use UTF-8 in Windows applications. It has a [standalone version](http://cppcms.com/files/nowide/html/index.html#standalone_version) that would allow us to use it without pulling in all of Boost.

steven-johnson:
Ah, the Win32 A-vs-W API nonsense: the gift that keeps on taking :-) I'm just gonna boldly assert that we don't care about supporting any Windows that doesn't have Unicode support, so, my recommendation would be:

-- Here (and everywhere else in Halide) that we rely on any Windows API that has A-vs-W variants, we explicitly call the W version so that we are immune to however the _UNICODE definitions are set
-- For places that we receive single-char strings (like this), use MultiByteToWideChar() (etc) to to the conversion.

Re: Boost, we absolutely positively will not consider using any Boost code in Halide; even a 'standalone' version is problematic, as Halide is very reluctant to bring in dependencies on any other C++ libraries. What, specifically, is the functionality in that library that you would like to use?

alexreinking:
[...]

PR #3684, opened 2019-02-26
Fix LoadLibrary call to use UTF-8 instead of ANSI.

alexreinking:
Addressing issue #3682, fixes the call to `LoadLibrary` to always use `LoadLibraryW` by converting UTF-8 to UTF-16 manually before call.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -865,7 +865,17 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
     for (const auto &lib : split_string(flags_info["-p"], ",")) {
         if (lib.empty()) continue;
 #ifdef _WIN32
-        if (LoadLibrary(lib.c_str()) != nullptr) {
+        HMODULE loaded_lib;
+
+        {
+            int wide_len = MultiByteToWideChar(CP_UTF8, 0, lib.c_str(), -1, nullptr, 0);
+            auto *wide_lib = new wchar_t[wide_len];
+            MultiByteToWideChar(CP_UTF8, 0, lib.c_str(), -1, wide_lib, wide_len);
+            loaded_lib = LoadLibraryW(wide_lib);
+            delete[] wide_lib;
+        }
+
+        if (loaded_lib != nullptr) {
             cerr << "Failed to load: " << lib << "\n";
             return 1;
         }
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -865,17 +865,20 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
     for (const auto &lib : split_string(flags_info["-p"], ",")) {
         if (lib.empty()) continue;
 #ifdef _WIN32
-        HMODULE loaded_lib;
-
-        {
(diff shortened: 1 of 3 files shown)


src/Generator.cpp
```diff
@@ -865,7 +865,17 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
     for (const auto &lib : split_string(flags_info["-p"], ",")) {
         if (lib.empty()) continue;
 #ifdef _WIN32
-        if (LoadLibrary(lib.c_str()) != nullptr) {
+        HMODULE loaded_lib;
+
+        {
+            int wide_len = MultiByteToWideChar(CP_UTF8, 0, lib.c_str(), -1, nullptr, 0);
+            auto *wide_lib = new wchar_t[wide_len];
```
steven-johnson: suggestion: use `std::vector<wchar_t>` here instead to get automatic cleanup; `.data()` is guaranteed to return a pointer to contiguous storage

src/Generator.cpp
```diff
@@ -865,7 +865,17 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
     for (const auto &lib : split_string(flags_info["-p"], ",")) {
         if (lib.empty()) continue;
 #ifdef _WIN32
-        if (LoadLibrary(lib.c_str()) != nullptr) {
+        HMODULE loaded_lib;
+
+        {
+            int wide_len = MultiByteToWideChar(CP_UTF8, 0, lib.c_str(), -1, nullptr, 0);
```
steven-johnson: IIRC this returns 0 if the call fails; we definitely need to check :-)

src/Generator.cpp
```diff
@@ -865,7 +865,17 @@ int generate_filter_main(int argc, char **argv, std::ostream &cerr) {
...
+            auto *wide_lib = new wchar_t[wide_len];
+            MultiByteToWideChar(CP_UTF8, 0, lib.c_str(), -1, wide_lib, wide_len);
```
steven-johnson: also check for error here too, alas

alexreinking:
Thanks @abadams -- the use of "A" there stopped it from breaking my build with `-DUNICODE` so I didn't notice it. :)

alexreinking:
I found one more in lesson 12 that can safely use LoadLibraryA since the strings are constant.

alexreinking:
Turns out I'm not sure how to fix the call in windows_get_symbol.cpp. It won't build if I try to include windows.h, and it would probably be a bad idea to hard code the value of `CP_UTF8`

alexreinking:
I chose to copy in constants because that's what was done with `SEM_NOOPENFILEERRORBOX`. I found that  I got compiler errors when trying to `#include vector` or use `new`/`delete` so I'm using malloc/free. If you know why that might be, please let me know.

alexreinking:
`simd_op_check` failed in mingw-64-testbranch-700. I understand that test is flaky; any chance that's my doing anyway?

steven-johnson:
It's probably not you, those are flaky. We really need to diagnose it.

alexreinking:
Okay, so this is safe to merge, then?

PR #3685, opened 2019-02-26
Add read_entire() and write_entire()

steven-johnson:
Simple utility functions to slurp / spew the entire contents of a file; not much re-use in this PR but I'll need them elsewhere in the future, so this is intended to reduce future PR noise.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -4,6 +4,7 @@
 #include "Introspection.h"
 #include <atomic>
 #include <chrono>
+#include <fstream>
 #include <iomanip>
 #include <map>
 #include <mutex>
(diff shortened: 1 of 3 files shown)


src/Util.h
```diff
@@ -265,6 +265,19 @@ void dir_rmdir(const std::string &name);
 /** Wrapper for stat(). Asserts upon error. */
 FileStat file_stat(const std::string &name);
 
+/** Read the entire contents of a file into a vector<char>. The file
+ * is read in binary mode. Errors trigger an assertion failure. */
+std::vector<char> read_entire(const std::string &pathname);
```
dsharletg: I think these should be named `read_entire_file` and `write_entire_file`, to make the code using it easier to read, and to be consistent with the rest of the file helper functions here.

src/Util.cpp
```diff
@@ -411,6 +412,29 @@ std::string dir_make_temp() {
...
+    f.close();
+    return result;
+}
+
+void write_entire(const std::string &pathname, const void *source, size_t source_len) {
+    std::ofstream f(pathname, std::ios::out|std::ios::binary);
+
+    f.write(reinterpret_cast<const char*>(source), source_len);
+    f.flush();
+    internal_assert(f.good()) << "Unable to write file: " << pathname;
```
dsharletg: I think the error checking should be in the caller, so the assert is more meaningful. Same for the read assert, though maybe it means read_entire needs to return a bool and take a reference to a vector as an argument?

src/Util.h
steven-johnson: Agreed.

src/Util.cpp
steven-johnson: Not sure about this; I like the idea of a "can't forget to check the error" function for these purposes.

steven-johnson: Also, the other file utilities in this file tend to have the same assert-fail behavior, rather than returning an error code.

steven-johnson:
Going to land this as-is.

PR #3686, opened 2019-02-26
Fix if_then_else and require for vectorized conditions on Hexagon

dsharletg:
Alternative to #3664


src/EliminateBoolVectors.cpp
```diff
@@ -141,6 +141,66 @@ class EliminateBoolVectors : public IRMutator {
...
+    // been boolean vectors converted to integer vectors, cast the masks to be the
+    // same type. This is necessary in the case of a select or Call::if_then_else, e.g.:
+    //
+    //    Expr a = float_expr1() < float_expr2();  // promoted to int32xN
+    //    Expr b = uint8_expr1() < uint8_expr2();  // promoted to int8xN
+    //    Expr c = select(a < b, a, b);            // whoops
+    static void unify_bool_vector_types(Expr& a, Expr& b) {
+        if (a.type().bits() != b.type().bits() &&
+            a.type().lanes() == b.type().lanes() &&
+            a.type().is_int() && b.type().is_int()) {
```
steven-johnson: These should be (is_int() || is_uint()). (We should really add Type::is_int_or_uint() as a method.)

dsharletg: I'm not sure that's true, I think bool masks are signed (so widening casts do the right thing).

dsharletg:
The win32 build bot failure must be unrelated, going to merge this.

PR #3687, opened 2019-02-28
Pipeline require

abadams:
Add the ability to make top-level unstructured requirements (assertions) with custom error messages. The simplifier will attempt to exploit them.


src/Pipeline.h
```diff
@@ -455,6 +455,21 @@ class Pipeline {
      * been rescheduled. */
     void invalidate_cache();
 
+    /** Add a top-level precondition to the generated pipeline,
+     * expressed as a boolean Expr. The Expr may depend on parameters
+     * only, and may not call any Func or use a Var. If the condition
```
steven-johnson: > may not call any Func or use a Var

What happens if this is violated -- do we check and produce an error, or is it UB?

steven-johnson: Are the requirements checked in the order added, or is the order (potentially) arbitrary? (Should be documented either way)

steven-johnson:
This appears to be (partly) green -- is it ready for final review?

abadams:
Still not sure if I have pacified msbuild. I checked the MSBuild source, and there is indeed a regexp that looks for warning: or error: or similar.

abadams:
But it *is* ready for review.

steven-johnson:
I have a vague recollection that CMake considers any string with "error" in it to be false, but hey, maybe both are stupid in this regard

src/Pipeline.h
dsharletg: References to Func/Var are checked in Pipeline::add_requirement (produces a user_error).

test/correctness/constraints.cpp
```diff
@@ -8,7 +8,7 @@ using namespace Halide;
 
 bool error_occurred = false;
 void my_error_handler(void *user_context, const char *msg) {
-    printf("%s\n", msg);
+    //printf("%s\n", msg);
```
dsharletg: Intended to commit this change?

dsharletg:
I found a cool use case for this: asserting that a pipeline parameter is a multiple of some constant. Normally I would just use `TailStrategy::RoundUp` on the relevant splits to get Halide to make the necessary simplifications, but in this case, the param is used in a reduction (which is an error to split with `TailStrategy::RoundUp`).

Of course, you can also do this with the `(param/N)*N` pattern, but that silently works even if param is not a multiple of N. Usually the way I do this is to make the parameter implicitly divided by N (e.g. the param would be `tile_size_divided_by_N`). I think `add_requirement` will be nicer.

steven-johnson:
> I found a cool use case for this:

We should find a way to capture this as a pattern for other users to discover. Doc comment here? Or on TailStrategy? Or somewhere in our wiki? Something else?

steven-johnson:
LGTM pending green. (I do think this merits additional documentation somewhere to ensure it gets noticed.)

test/correctness/constraints.cpp
abadams: Yes, it's too noisy with it, but when debugging the test you want to uncomment it.

abadams:
arm failure due to undiagnosed simd_op_check issue. linux failure fixed by #3708

Issue #3683, opened 2019-02-26
Using _ on the LHS without a matching _ on the RHS should be an error


abadams:
Currently it's not. The _ on the LHS just expands to zero Vars, which doesn't seem like ideal behavior.

PR #3688, opened 2019-02-28
Assert no implicit l.h.s vars when there is no implicit r.h.s vars.

asaadaldien:
Fix  for #3683, This should report and error when using _ on the LHS without a matching
_ on the RHS.

Test Plan:
test/error/lhs_implicit_no_rhs.cpp


src/Func.cpp
```diff
@@ -2595,6 +2595,11 @@ vector<Expr> FuncRef::args_with_implicit_vars(const vector<Expr> &e) const {
                 iter++;
             }
         }
+    } else {
+        user_assert(implicit_placeholder_pos == -1)
+          << "Left-hand-side of update definition of " << func.name()
```
abadams: update definition -> definition

This says "update definition", which is copied from above, but I think it's incorrect here - the message above occurs within a has_pure_definition condition, so you know it's an update definition there. Here it could be either.

asaadaldien:
Another issue I found while working on this is Halide::lambda(expr) always assigns implicit  var (_) to lhs. we should do that only | _ vars|  on the r.h.s is > 0. thoughts ?

zvookin:
I'm pretty sure the design as it is was intentional. The entire point of _ is to be convenient for metaprogramming type stuff. The lambda case is an example of this type of convenience.

At some level, I consider the entire placeholder feature a failure anyway. Does anyone really use it?

asaadaldien:
My understanding  of placeholders syntax is to unfolds _ left -> right. 
So frontend should not accept something like 
```
f(x,_) = g(x) * h(x);
```

We can change `Halide::lambda(Expr e)` to return `f()` if e doesn't have any implicit vars.
Instead if returning `f(_)` regardless.

abadams:
My current conception of it is that it's bound on the RHS to match some number of vars, and then available for use on the LHS. Currently if it's unbound on the RHS we unpack it to zero Vars, but I saw that create a confusing error message for someone and thought there should be an explicit use of it on the RHS.

I do see people using it, though I share your misgivings about the value of the feature.

asaadaldien:
> but I saw that create a confusing error message for someone and thought there should be an explicit use of it on the RHS.

What syntax repro an error message ? 
Without this PR the following syntax is valid.
```
f(x, _) = h(x);
```
This PR is to prevent that.

abadams:
The error message was for code similar to the following but more complex:

```
f(_) = 0;
f(_) += .... g(_) ...
```

I think the user assumed that the _ would be left unexpanded until resolved by the update definition. The error was about f being called with two arguments even though it was zero-dimensional. The fact that it was zero-dimensional surprised the user. It seemed to be that we should just disallow that kind of pure definition, because the _ in it is ambiguous.

zvookin:
Please see the docs for ```Var::implicit```. At the very least it needs to be updated to cover this constraint. Also these two test cases need to be verified:
```
    Func f, g;
    Var x, y;
    f(x, y) = 0;
    g(_) = f*3;
```

and

```
    Func f, g;
    Var x, y;
    f(x, y) = 0;
    g = lambda(f*3);
```

(Where by verified, I mean checked that they work currently and still work with the change, or the above referenced documentation, and possibly that for lambda, updated.)

zvookin:
Pretty sure you want "scalar" not "scaler" everywhere here.

asaadaldien:
Correct should be "scalar not scaler" :). 
For some reason I don't get notifications form comments on my  github PRs!!!
I will update the PR to make sure the test cases you mentioned are covered and documentation is updated.

asaadaldien:
@zvookin In the test cases you meant calling f with implicit vars, right ? e.g

```
    Func f, g;
    Var x, y;
    f(x, y) = 1;
    g(_) = f(_)*3;
```
This works on master and this PR.

zvookin:
The doc on Var::implicit claims ```g(_) = f*3``` with no argument list on ```f`` is valid. It is used as an actual example: https://github.com/halide/Halide/blob/master/src/Var.h#L57 . If this is wrong, it needs to be fixed.

asaadaldien:
`g(_) = f*3` doesn't work currently. I will update the documentation.

test/error/lhs_implicit_scaler_rhs.cpp
```diff
@@ -0,0 +1,11 @@
+#include <Halide.h>
+
+using namespace Halide;
+
+int main(int argc, char **argv) {
+  Func f;
+  Var x;
+  // Implicit variables only allowed when rhs is non-scalar.
```
abadams: "rhs is non-scalar" -> "rhs also uses _"

src/Lambda.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    CountImplicitVars(const Expr &e)
+        : count(0) {
+        e.accept(this);
+    }
+
+    using IRGraphVisitor::visit;
+
+    void visit(const Internal::Variable *v) override {
+        int index = Var::implicit_index(v->name);
```
abadams: count = std::max(count, Var::implicit_index(v->name) + 1)

src/Func.cpp
```diff
@@ -2569,36 +2569,42 @@ vector<Expr> FuncRef::args_with_implicit_vars(const vector<Expr> &e) const {
...
-    CountImplicitVars count(e);
+    CountVars count(e);
     for (size_t i = 0; i < a.size(); i++) {
         a[i].accept(&count);
     }
 
-    if (count.count > 0) {
+    CountVars rhs_count(e);
+    if (!func.has_pure_definition() && implicit_placeholder_pos != -1) {
+        user_assert(rhs_count.var_count > 0)
```
abadams: I don't understand this part. If the left-hand-side uses implicit vars, why does it matter whether or not the right hand side uses non-implicit vars? Why track the total number of vars at all?

asaadaldien: Because we want to prevent implicit var in l.h.s and scaler function on the r.h.s. `e.g f(_) = 0`
This PR still allow _ on the l.h.s to expand to zero vars. That seem to be a common pattern (e.g in local_laplacian app)
```
downx(x, y, _) = (f(2*x-1, y, _) + 3.0f * (f(2*x, y, _) + f(2*x+1, y, _)) + f(2*x+2, y, _)) / 8.0f;
```
You can use this to downsample gray or rgb f.

Note:when we are here r.h.s implicit vars already expanded.

abadams:
Damn. I think if the RHS is already expanded, then we have no way to distinguish between:

```
f(x, y, _) = g(x, y, _)
```

and 

```
f(x, y, _) = g(x, y)
```

Ideally the first one would be legal syntax and the second would not. Maybe we should move the checking to FuncRef::operator= where the expansion is done. There we have the opportunity to assert that the unexpanded RHS may only contain _ if the LHS does, and vice-versa.

asaadaldien:
This already in `FuncRef::operator=`, Unfortunately the current implementation I can't see a single place where _ unexpanded in both sides to perform the check. It's already happening at constructing r.h.s calls `Func::operator()(...)`. 

IMHO In order to prevent this assignment `f(x,y,_) = h(x,y)`
We have to push the entier expression down to check if its valid then 
write the expansion as a mutation (not at call time).

abadams:
Oh duh, good point.

Nuts. I'm out of ideas then. Maybe we should shelve this for now.

abadams:
Closing for now, as we hit a show-stopper, and it's not a critical issue so it doesn't seem worth pursuing more aggressively.

PR #3689, opened 2019-02-28
Add support for Fuchsia operating system.

zvookin:
Working on figuring out how to test this.


zvookin:
Ran some code. Basic stuff seems to work. Opening for review.

src/runtime/fuchsia_clock.cpp
```diff
@@ -0,0 +1,34 @@
+#include "HalideRuntime.h"
+
+extern "C" {
+
+typedef int32_t zx_status_t;
+typedef int64_t zx_time_t;
+typedef int64_t zx_duration_t;
+  
+zx_time_t zx_clock_get_monotonic(void);
```
steven-johnson: supernit: we rarely use (void) in Halide, tend to prefer () instead

PR #3691, opened 2019-03-01
Runtime cleanups

steven-johnson:
Cherrypicking some changes from an experimental branch that adds a WebAssembly backend, to simplfiy downstream reviews:

- add a 'fake_get_symbol' module, for supporting platforms where dlsym(), etc aren't supported. (wasm actually theoretically can support this but getting the runtime correct is tricky, so leaving it out is simpler for now.) Yes, this is unused right now, but I think that's OK.

~~- Move `halide_sleep_ms` from the clock modules to the thread modules, since it's thread-related, and platforms using `fake_thread_pool` don't need a real implementation~~

- Add some missing stub functions to `fake_thread_pool`.

- Add some missing #includes to a couple of tests.

- Clean up weird disaster in the cross_compilation test.


abadams:
Most of this LGTM, but sleep not sleeping on platforms without threading support seems pretty odd to me.

steven-johnson:
re: sleep, I'll double-check that there isn't an alternate workaround.

steven-johnson:
OK, so, yes, implementing halide_sleep_ms() is fine in the case I was worried about. I'll revert those parts of this change.

src/runtime/fake_get_symbol.cpp
```diff
@@ -0,0 +1,62 @@
+#include "HalideRuntime.h"
+
+extern "C" {
+
+WEAK void *halide_default_get_symbol(const char *name) {
+    halide_error(NULL, "halide_default_get_symbol not implemented on this platform.");
+    return 0;
```
dsharletg: Super nit: 0 -> NULL?

steven-johnson:
One more change: split `halide_trace_helper` into its own runtime module; this allows us to inject the tracing code itself (part of the runtime) differently from the trace_helper (which needs special handling in some JIT situations).

abadams:
Changes since last approved lgtm

PR #3693, opened 2019-03-02
Hexagon target flag cleanup

dsharletg:
- Deprecate HVX_64 and HVX_128
- Rename HVX_128 to HVX
- Remove HVX_shared_object (unused in the compiler, deprecated for a while now)

I considered removing the HVX_64 tests and apps (and the associated build bot run), but I think we should keep them until we actually remove support.


abadams:
Does this deprecate HVX-64 mode? Might there be current users?

dsharletg:
It does, but Qualcomm has also deprecated HVX_64. Note that HVX_64 is still working after this PR. The main purpose of this PR is to start letting people migrate code from using HVX_64/HVX_128 to just 'HVX'. I don't think we should actually remove support for HVX_64 for a while yet.

steven-johnson:
LGTM

PR #3694, opened 2019-03-04
Fix errors due to -Wno-missing-braces+-Werror

shoaibkamil:
Initializing an std::array with single braces is a warning in certain versions of clang/gcc if in C++11 mode.  Because we error on warnings, this prevents Halide from building.

Specifically, I was unable to build on Apple LLVM version 9.0.0 plus LLVM/clang 7.0 without this change.


alexreinking:
How annoying. Thanks for fixing!

Issue #3699, opened 2019-03-06
FFT generator does not use parallelism param


jrk:
It seems that the FFT codegen logic in fft.h/cpp has an optional flag for internal parallelism (in `struct Fft2dDesc`), and that the top-level FFT generator has a `GeneratorParam parallelism`, but that the latter is never passed into the former so it has no effect on code generation.

dsharletg:
Fixed by the above referenced PR.

PR #3700, opened 2019-03-06
FFT generator should propagate parallel parm (Issue #3699)

steven-johnson:


dsharletg:
Thanks for taking this one :)

PR #3705, opened 2019-03-08
Safer implementation of mapping performance mode to voltage corner.

dsharletg:


abadams:
When you said "safer" in combination with "voltage" I got excited and thought devices were catching fire. Very disappointed to find that it's about enum ordering issues and out-of-bounds reads.

dsharletg:
I'm glad we don't have to worry about that kind of problem :)

dsharletg:
I'm going to merge this because it doesn't get any coverage by the build bots anyways (only used on device). I tested it on device.

PR #3707, opened 2019-03-08
Stack usage improvements

abadams:
grab-bag of stack usage reductions to reduce stack overflows on pathological pipelines. No functional changes intended. No measurable effect on lowering time.

I'm separately exploring ways to just buy a larger stack, but it's unlikely to apply to every platform.


src/BoundSmallAllocations.cpp
```diff
@@ -14,16 +14,38 @@ class BoundSmallAllocations : public IRMutator {
     // Track constant bounds
     Scope<Interval> scope;
 
+    template<typename T, typename Body>
+    Body visit_let(const T *op) {
```
abadams: This is the best example of the idiom I use throughout. Make a vector of frames on the heap storing the state that used to be on the stack, build it up using a do while loop, mutate the body, then unwind by reverse iterating over the frames.

src/AsyncProducers.cpp
```diff
@@ -494,15 +494,30 @@ class ExpandAcquireNodes : public IRMutator {
...
     Stmt visit(const Block *op) override {
-        Stmt first = mutate(op->first), rest = mutate(op->rest);
-        if (const Acquire *a = first.as<Acquire>()) {
-            // May as well nest the rest stmt inside the acquire
-            // node. It's also blocked on it.
-            return Acquire::make(a->semaphore, a->count,
-                                 mutate(Block::make(a->body, op->rest)));
-        } else {
-            return Block::make(first, rest);
+        vector<Stmt> stmts;
```
steven-johnson: Here (and elsewhere) it's worth documenting that this structure is done to conserve stack space.

src/BoundSmallAllocations.cpp
steven-johnson: Is the idiom regular enough that it could be template-ized somehow?

abadams: I thought about that for a while, and I don't believe so. There might be something clever with lambdas to capture state or something, but it escaped me.

dsharletg: What about this (for Block, the same thing could work for Let/LetStmt):

- Add visitor methods `IRMutator(vector<const Block*>)` and `IRVisitor::visit(vector<const Block*>)`
- Implement IRMutator::visit(const Block*) and IRVisitor::visit(const Block*) to flatten Blocks into vectors of Blocks, call the visitors above, and then expand them back into IR Blocks.

This probably doesn't work due to C++ overloading/visitor pattern stuff, but it seems like it *might* work. Maybe it should be in IRVisitor/IRMutator derived classes to avoid the overhead when it isn't necessary.

abadams: It's not the data structure that's the problem. The visitor for consuming a linked list of Blocks looks the same as the visitor for consuming a vector of Blocks - all you need to do is forward-iterate over it. Pre-packing things into a vector just adds overhead.

slomp:
Great work!
With these changes I'm seeing a substantial reduction in overall code-gen times on my side!

(btw, the `HALIDE_NEVER_INLINE` in `Deinterleave.cpp` won't compile on MSVC;
MSVC relies on `__declspec(noinline)` which has to decorate the function at the very beginning)

abadams:
Thanks Marcos, I've moved it.

src/AsyncProducers.cpp
abadams: Done.

shoaibkamil:
Is there some reason this didn't run on the buildbots?  I'm hoping we can merge this sooner rather than later (and I don't see any outstanding review issues)

abadams:
I think the buildbots haven't picked it up yet. I'm surprised this had any effect for anyone except for us - we were getting stack overflows lowering some larger neural networks with weird schedules.

slomp:
I am also getting a bunch of stack overflow issues on my side with the recent masters.
(I still rely on a fairly old build of Halide, and that one works, so I just noticed the issue very recently when trying to upgrade)

abadams:
If you have a simple repro of something that stack overflows with master, that would be very helpful. This PR doesn't change the amount of memory used - it just moves it from stack to heap - so if non-pathological code is stack overflowing even without this PR, we have a different problem.

abadams:
(This PR is specifically to handle the pathological code we're producing with random schedules for larger pipelines)

steven-johnson:
There are a handful of tests that will fail due to stack space if you build with (e.g.) -O1 instead of -O3. correctness_vector_math, IIRC.

slomp:
Ah, one more thing:
I was getting this when trying to build this branch on MacOS
```
(...)/halide/src/Target.cpp:811:13: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            CUDA, OpenCL, OpenGL, OpenGLCompute, Metal, D3D12Compute, NoNEON,
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(...)/halide/src/Target.cpp:820:13: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  ...SSE41, AVX, AVX2, FMA, FMA4, F16C, ARMv7s,VSX, AVX512, AVX512_KNL, AVX512_Skylake, AVX512_Cannonlake
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     {                                                                                                   }
(...)/halide/src/Target.cpp:824:13: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  ...SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX_64, HVX_128, MinGW, HexagonDma, HVX_shared_object
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     {                                                                                           }
(...)/halide/src/Target.cpp:904:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"x86-64-linux-sse41-fma", "x86-64-linux-sse41-fma", "x86-64-linux-sse41-fma"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                                           }
(...)/halide/src/Target.cpp:905:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  ..."x86-64-linux-sse41-fma-no_asserts-no_runtime", "x86-64-linux-sse41-fma", "x86-64-linux-sse41-fma"},
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     {                                                                                                 }
(...)/halide/src/Target.cpp:906:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"x86-64-linux-avx2-sse41", "x86-64-linux-sse41-fma", "x86-64-linux-sse41"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                                        }
(...)/halide/src/Target.cpp:907:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"x86-64-linux-avx2-sse41", "x86-32-linux-sse41-fma", ""},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                      }
(...)/halide/src/Target.cpp:908:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"x86-64-linux-cuda", "x86-64-linux", "x86-64-linux-cuda"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                       }
(...)/halide/src/Target.cpp:909:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"x86-64-linux-cuda-cuda_capability_50", "x86-64-linux-cuda", "x86-64-linux-cuda"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                                               }
(...)/halide/src/Target.cpp:910:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
  ..."x86-64-linux-cuda-cuda_capability_50", "x86-64-linux-cuda-cuda_capability_30", "x86-64-linux-cuda-cuda_capability_30"},
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     {                                                                                                                     }
(...)/halide/src/Target.cpp:911:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"x86-64-linux-cuda", "x86-64-linux-opengl", "x86-64-linux-cuda-opengl"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                                     }
(...)/halide/src/Target.cpp:912:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"hexagon-32-qurt-hvx_v65", "hexagon-32-qurt-hvx_v62", "hexagon-32-qurt-hvx_v62"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                                              }
(...)/halide/src/Target.cpp:913:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"hexagon-32-qurt-hvx_v62", "hexagon-32-qurt", "hexagon-32-qurt"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                              }
(...)/halide/src/Target.cpp:914:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"hexagon-32-qurt-hvx_v62-hvx_64", "hexagon-32-qurt", ""},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                      }
(...)/halide/src/Target.cpp:915:14: error: 
      suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
            {"hexagon-32-qurt-hvx_v62-hvx_64", "hexagon-32-qurt-hvx_64", "hexagon-32-qurt-hvx_64"},
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             {                                                                                   }
```
My quick solution for quick testing was to simply add:
```
#pragma clang diagnostic ignored "-Wmissing-braces"
```
but I guess there is no harm in adding the suggested braces around these initializations.

slomp:
I have a (very convoluted) example that not only causes stack-overflow with master, but also sets `can_prove()` on fire (even on this stack_usage_improvements branch here), but I am not sure if I can share it. @shoaibkamil Any ideas?

shoaibkamil:
@slomp The bracing issue will be fixed when Andrew merges in master (I fixed it last week)

@abadams We could at least share the failed can_prove log, and I can, later this week, try to boil it down to a small repro.

steven-johnson:
synced to master to try to tickle the buildbots

steven-johnson:
Ready to land, I think

Issue #3709, opened 2019-03-10
Support for Float16 in untyped Generator Input<Buffer<>>


JasperJenkins:
There is no command line type for float16 use in Generators. Was this intentional or just an omission? Probably a workaround with uint16_t and a flag that adds a reinterpret cast, but that seems hacky for what afaict is a couple lines added to Generator.cpp.

abadams:
An omission, but in master float16 is still half-implemented. It has been stalled by float16 bugs in llvm. I'll add it as a TODO in my float16 branch #3667

steven-johnson:
It's more than a couple of lines. Partial PR on the way.

steven-johnson:
Please see https://github.com/halide/Halide/pull/3711

steven-johnson:
PR#3713 has a more limited fix (Buffers only, no scalars) that may be adequate for now

JasperJenkins:
Thanks a lot. It works great for my use case. For scalar params I can cast from a f32 param, and implement the undefined conversion symbols myself.

PR #3713, opened 2019-03-12
Support Buffer<float16> in Generators (Issue #3709)

steven-johnson:
Note that this should allow for Input<Buffer<float16>> and Output<Buffer<float16>>, but *not* Input<float16> or Output<float16>; those require a bit more work (see PR #3711)


abadams:
This is all green and apparently already useful to someone.

PR #3715, opened 2019-03-12
Silence irrelevant data race in Introspection.cpp

steven-johnson:
At least one version of libc++ on Linux will report data races inside the std::regex implementation if we call get_variable_name() from multiple threads at the same time (e.g. from correctness_vector_width); it appears that the implementation details of std::regex have a "benign"* data race. Inserting a mutext when in TSAN mode to silence what is likely to be an irrelevant-to-us data race when using TSAN.

* There's no such thing as a benign data race, really, but this is one that's unlikely to be worth chasing down.

### DIFF
--- a/src/Introspection.cpp
+++ b/src/Introspection.cpp
@@ -2201,13 +2201,8 @@ private:
 };
 
 namespace {
-
 DebugSections *debug_sections = nullptr;
-
-#if __has_feature(thread_sanitizer)
 std::mutex get_variable_name_lock;
-#endif
-
 }
 
 bool dump_stack_frame() {
@@ -2219,19 +2214,15 @@ bool dump_stack_frame() {
 }
 
 std::string get_variable_name(const void *var, const std::string &expected_type) {
+    if (!debug_sections) return "";
+    if (!debug_sections->working) return "";
 
-#if defined(__has_feature)
-#if __has_feature(thread_sanitizer)
     // At least one version of libc++ will report data races inside the regex
     // implementation if we call this from multiple threads at the same time
     // (e.g. from correctness_vector_width); this mutex serves more to silence
     // what is likely to be an irrelevant-to-us data race when using TSAN.
     std::lock_guard<std::mutex> lock(get_variable_name_lock);
-#endif
-#endif
 
-    if (!debug_sections) return "";
-    if (!debug_sections->working) return "";
     std::string name = debug_sections->get_stack_variable_name(var, expected_type);
     if (name.empty()) {
         // Maybe it's a member of a heap object.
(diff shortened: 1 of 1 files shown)


src/Introspection.cpp
```diff
@@ -2200,7 +2201,13 @@ class DebugSections {
 };
 
 namespace {
+
 DebugSections *debug_sections = nullptr;
+
+#if __has_feature(thread_sanitizer)
```
dsharletg: I'm guessing this won't work on MSVC. It seems fine to just put the mutex in unconditionally, the performance of this doesn't matter does it?

steven-johnson: You're right, needs to be wrapped in #if defined(__has_feature), as it is in Target.cpp.

It's called from `make_entity_name()`, which is called a lot.

zvookin:
The TSan mantra is "There is no such thing as an irrelevant data race." Either there's a bug in our code or libc++.

zvookin:
And if it's performance critical, does it really want to be using regex anyway?

steven-johnson:
I'm guessing this hasn't been noticed inside Google since the style guidelines forbid usage of std::regex in the first place :-)

src/Introspection.cpp
dsharletg: I think if you just put the lock after the two early outs for missing debug data, it will only ever actually lock the mutex for make_entity_name calls directly from user declared objects. I think all of the internal make_entity_name calls won't get that far.

src/Introspection.cpp
```diff
@@ -2214,6 +2216,13 @@ bool dump_stack_frame() {
 std::string get_variable_name(const void *var, const std::string &expected_type) {
     if (!debug_sections) return "";
     if (!debug_sections->working) return "";
+
+    // At least one version of libc++ will report data races inside the regex
+    // implementation if we call this from multiple threads at the same time
+    // (e.g. from correctness_vector_width); this mutex serves more to silence
+    // what is likely to be an irrelevant-to-us data race when using TSAN.
+    std::lock_guard<std::mutex> lock(get_variable_name_lock);
```
abadams: This entirely serializes something potentially expensive (stack walking code). I don't think it's a good solution.

src/Introspection.cpp
abadams: I think they all do get that far. The early-outs are for when the user code wasn't compiled with -g, so the stack walking also occurs for things internal to libHalide.

steven-johnson:
I'll probably just abandon this PR rather than try to track down the issue -- I can silence the noise locally for my own purposes.

Issue #3718, opened 2019-03-13
x == x not simplified


shoaibkamil:
During lowering, we're seeing a stream of failed proofs that include:
```
...
Trimming loops to the region over which they do something...
Failed to prove, but could not find a counter-example:
 (let v0 = (v1 != 0) in (v0 == v0))
Original expression:
(let t9280 = (t2898 != 0) in (t9280 == t9280))
Failed to prove, but could not find a counter-example:
 (let v0 = (v1 != 0) in (v0 == v0))
Original expression:
(let t9600 = (t2898 != 0) in (t9600 == t9600))
Failed to prove, but could not find a counter-example:
 (let v0 = (v1 != 0) in (v0 == v0))
Original expression:
(let t9915 = (t2898 != 0) in (t9915 == t9915))
Failed to prove, but could not find a counter-example:
 (let v0 = (v1 != 0) in (v0 == v0))
Original expression:
(let t10237 = (t2898 != 0) in (t10237 == t10237))
....
```

Here's a simple example that shows we don't simplify:
```
#include "Halide.h"

using namespace Halide;

int main(int argc, char* argv[]) {

  ImageParam image{UInt(8), 3};
  Var x{"x"}, y{"y"}, c{"c"};
  Func highpass{"highpass"};
  Param<int> k;

  highpass(x, y) = (k != 0) == (k != 0);

  highpass.compile_to_file("boo", {image, k}, "gah");

  return 0;
}
```

And the stmt:
```
  produce highpass {
    let t8 = (p1 != 0)
    let t6 = uint8((t8 == t8))
    let t7 = ((highpass.min.1*highpass.stride.1) + highpass.min.0)
    for (highpass.s0.y, highpass.min.1, highpass.extent.1) {
      let t9 = ((highpass.s0.y*highpass.stride.1) - t7)
      for (highpass.s0.x, highpass.min.0, highpass.extent.0) {
        highpass[(highpass.s0.x + t9)] = t6
      }
    }
  }
```

There may be a reason I'm not understanding, but I think this should just simplify to `highpass[(highpass.s0.x + t9)] = (uint8)1;`

shoaibkamil:
Closed via #3719

PR #3719, opened 2019-03-13
Simplify x == x

shoaibkamil:
For both boolean and numeric values of x.  An attempt to address #3718 in the simplest manner possible.

### DIFF
--- a/src/Simplify_EQ.cpp
+++ b/src/Simplify_EQ.cpp
@@ -42,9 +42,8 @@ Expr Simplify::visit(const EQ *op, ExprInfo *bounds) {
     Expr delta = mutate(op->a - op->b, &delta_bounds);
     const int lanes = op->type.lanes();
 
-    // If the bounds are 0, then it's just x == x
-    if (delta_bounds.min_defined && delta_bounds.min == 0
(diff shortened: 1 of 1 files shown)


src/Simplify_EQ.cpp
```diff
@@ -39,6 +42,12 @@ Expr Simplify::visit(const EQ *op, ExprInfo *bounds) {
     Expr delta = mutate(op->a - op->b, &delta_bounds);
     const int lanes = op->type.lanes();
 
+    // If the bounds are 0, then it's just x == x
```
abadams: I don't think this actually covers floating point x == x, as the bounds won't be defined. I think the missing rewrite (in addition to the one above), is actually 0 == 0

abadams: Though written like that it's obviously not going to be a matcher. Maybe we should just say if (is_zero(delta)) for the condition here.

shoaibkamil: Ah right, I forgot bounds won't be defined for floating point.

slomp:
I remember auditing that piece of code many weeks ago when I first encountered the problem, and also completely missed that case :(

PR #3720, opened 2019-03-13
Handle correlated subtractions

### LINKED ISSUE #3697: Bounds inference is overconservative when hoisting allocations into shared memory
This is an interval-arithmetic-hates-correlated-subexpressions problem. It is best explained with some sample code.

```
#include "Halide.h"

using namespace Halide;

void works() {
    Func f, g;
    Var x, y;

    g(x, y) = 0.0f;
    f(x, y) = g(x, y) * 2.0f;

    // Make a three-level tiling
    Var xi, yi, xii, yii;
    f.tile(x, y, xi, yi, 64, 1024)
        .tile(xi, yi, xii, yii, 32, 4, TailStrategy::GuardWithIf)
        .gpu_blocks(x, y)
        .gpu_threads(xii, yii);

    // Inside a serial loop inside the blocks loop. Will be stored in shared.
    g.compute_at(f, xi).gpu_threads(x, y);

    // Allocates 512 bytes of shared.
    f.realize(1024, 1024);
}

void does_not_work() {
    Func f, g;
    Var x, y;

    g(x, y) = 0.0f;
    f(x, y) = g(x, y) * 2.0f;


    // Exactly the same as above, but tile height is no longer a compile-time-constant
    Param<int> h;

    // Make a three-level tiling
    Var xi, yi, xii, yii;
    f.tile(x, y, xi, yi, 64, h)
        .tile(xi, yi, xii, yii, 32, 4, TailStrategy::GuardWithIf)
        .gpu_blocks(x, y)
        .gpu_threads(xii, yii);

    // N.B, 4 does not necessarily divide h, so the amount of g required by f
    // changes as a function of yi.

    // Inside a serial loop inside the blocks loop. Will be stored in shared.
    g.compute_at(f, xi).gpu_threads(x, y);

    h.set(1024);

    // Kaboom. Tries to allocate 128 * h bytes of shared
    f.realize(1024, 1024);
}

    printf("Trying with statically known no-tail-case...\n");
    works();
    printf("Trying with dynamic loop extent, which requires a tail case and a dynamic-sized allocation...\n");
    does_not_work();
    return 0;
}

```

abadams:
This is a bandaid to avoid the largest weakness of symbolic interval arithmetic: handling of correlated differences (e.g. this fixes #3697). To copy-paste the new docs:

```

/** Symbolic interval arithmetic can be extremely conservative in
 * cases where we analyze the difference between two correlated
 * expressions. For example, consider:
 *
 * for x in [0, 10]:
 *   let y = x + 3
 *   let z = y - x
 *
 * x lies within [0, 10]. Interval arithmetic will correctly determine
 * that y lies within [3, 13]. When z is encountered, it is treated as
 * a difference of two independent variables, and gives [3 - 10, 13 -
 * 0] = [-7, 13] instead of the tighter interval [3, 3]. It
 * doesn't understand that y and x are correlated.
 *
 * In practice, this problem causes problems for unrolling, and
 * arbitrarily-bad overconservative behavior in bounds inference
 * (e.g.h ttps://github.com/halide/Halide/issues/3697 )
 *
 * The function below attempts to address this by walking the IR,
 * remembering whether each let variable is monotonic increasing,
 * decreasing, unknown, or constant w.r.t each loop var. When it
 * encounters a subtract node where both sides have the same
 * monotonicity it substitutes, solves, and attempts to generally
 * simplify as aggressively as possible to try to cancel out the
 * repeated dependence on x. The same is done for addition nodes with
 * arguments of opposite monotonicity. Only index expressions outside of
 * Stores/Provides are considered.
 *
 * This pass is safe to use on code with repeated instances of the
 * same variable name (it must be, because we want to run it before
 * allocation bounds inference).
 */
```

This PR also exploits this to be able to unroll in more cases. If an unrolled loop is not constant sized, but does have a constant upper bound, it rounds up the loop extent to that bound, masks off the additional iterations with an if statement inside the loop, and then unrolls.

In general this makes Funcs compute_at a GuardWithIf split much cleaner, both in terms of their loop and their allocation size (see test/correctness/unroll_dynamic_loop.cpp).

These extra passes are not free. This PR increases lowering time by about 4% on local laplacian. The difference is in the noise on camera pipe.


src/Simplify_Stmts.cpp
```diff
@@ -138,6 +138,18 @@ Stmt Simplify::visit(const For *op) {
 
     if (is_no_op(new_body)) {
         return new_body;
+    } else if (extent_bounds.max_defined &&
```
abadams: RemoveTrivialForLoops amounted to no more than this code here, so I nuked it.

src/VectorizeLoops.cpp
```diff
@@ -158,6 +158,12 @@ Interval bounds_of_lanes(Expr e) {
         if (expr_uses_var(ib.max, let->name + ".max_lane")) {
             ib.max = Let::make(let->name + ".max_lane", ia.max, ib.max);
         }
+        if (expr_uses_var(ib.min, let->name)) {
```
abadams: This fixes a bug in master.

src/Lower.cpp
```diff
@@ -174,6 +174,7 @@ Module lower(const vector<Function> &output_funcs,
     debug(2) << "Lowering after sliding window:\n" << s << '\n';
 
     debug(1) << "Performing allocation bounds inference...\n";
+    s = simplify_correlated_differences(s);
```
zvookin: I'd really like all passes to have their own debug messages here.

zvookin: Hmm... I guess we already lost that property to the lower function. It's probably time to consider adding a little bit of structure here to allow consistent printing of timing and debug info at varying levels of granularity.

abadams: Done

abadams: It's already a little heirarchical, in that some of the passes (e.g. partition loops) are internally arranged as a sequence of sub-passes. We could just break out all the passes here.

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -0,0 +1,130 @@
...
+};
+
+}  // namespace
+
+Stmt simplify_correlated_differences(const Stmt &s) {
+    return SimplifyCorrelatedDifferences().mutate(s);
+}
+
+}
+}
```
zvookin: Is it worth adding an internal test for this?

abadams: I don't think so. It's a pretty simple pass, but it only does something meaningful on large complex pieces of IR, so it would be very cumbersome to test as a unit. The two new end-to-end tests do a good job of covering it, and it also changes the IR generated on many of the existing tests and apps.

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+            e = common_subexpression_elimination(e);
+            e = solve_expression(e, loop_var).result;
+            e = simplify(e);
+
+            if ((debug::debug_level() > 0) &&
+                is_monotonic(e, loop_var, monotonic) == Monotonic::Unknown) {
+                // Might be a missed simplification opportunity. Log to help improve the simplifier.
+                debug(1) << "Warning: expression is non-monotonic in loop variable " << loop_var << ": " << e << "\n";
```
abadams: @shaoibkamil @jn80842 This is intended to be a source of expressions that we failed to cancel a repeated variable in.

abadams:
arm32 failure unrelated.

steven-johnson:
This also seems to have broken `warning_parallel_size_one`

PR #3721, opened 2019-03-13
added parallel build toggle for Halide when cmake generate its Visual Studio project

slomp:
(I would expose the flag to all targets, but that might saturate the build since multiple projects might get built in parallel; Halide itself is a knot point of the build, so compiling multiple files in parallel will not oversubscribe the system)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -604,6 +604,8 @@ add_library(Halide ${HALIDE_LIBRARY_TYPE}
   ${INITIAL_MODULES}
 )
 
+target_compile_options(Halide PUBLIC $<$<CXX_COMPILER_ID:MSVC>:/MP>)
+
 # Define Halide_SHARED or Halide_STATIC depending on library type
 target_compile_definitions(Halide PRIVATE "-DHalide_${HALIDE_LIBRARY_TYPE}")
 # Ensure that these tools are build first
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -604,6 +604,8 @@ add_library(Halide ${HALIDE_LIBRARY_TYPE}
   ${INITIAL_MODULES}
 )
 
+target_compile_options(Halide PUBLIC $<$<CXX_COMPILER_ID:MSVC>:/MP>)
```
steven-johnson: It's worth adding a comment here that basically replicates the PR comment, otherwise LGTM

abadams:
Does this do something in addition to the parallelism flags on MSBuild? Is MSBuild only parallel over projects?

slomp:
AFAIK:
MSBuild parallelism flags are for building non-dependent projects in parallel.
/MP is for compiling source files within a project in parallel

shoaibkamil:
LGTM with the comment suggested by Steven.

slomp:
Is this something I need to worry about?
`buildbot/mingw-64-testbranch-700 `:
```
make: *** [../halide/Makefile:1666: quiet_correctness_simd_op_check] Error 1
```

steven-johnson:
We've been getting intermittent failures in simd_op_check (multiple platforms) for weeks now, so this is nothing new. It really needs fixing but it's not your bug.

PR #3723, opened 2019-03-13
More simplification of IfThenElse stmts

dsharletg:
Add a few missed simplifications of if stmts. Before, common statements were only pulled out of Blocks when both branches were Blocks. This PR handles the case where zero or one of the branches are a Block.

### DIFF
--- a/src/Simplify_Stmts.cpp
+++ b/src/Simplify_Stmts.cpp
@@ -46,6 +46,9 @@ Stmt Simplify::visit(const IfThenElse *op) {
     }
 
     // Pull out common nodes
+    if (equal(then_case, else_case)) {
+        return then_case;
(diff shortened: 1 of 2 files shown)


dsharletg:
I was wondering about how this interacts with `specialize`. If the specialization simplifications don't occur before this runs, then this might undo the specialization.

However, we were *attempting* to do this before, it just wasn't very strong :) So We should either add these extra cases, or we should remove the attempts to simplify Blocks that already exist in master.

abadams:
I think it interacts with specialization fine - the new rules happen after then_case and else_case have already been mutated exploiting the condition. If they're found to partially match afterwards, no useful specialization was done to that piece of code anyway.

PR #3724, opened 2019-03-14
Use invalidate_cache() aggressively in multithreaded tests

steven-johnson:
Make Func::invalidate_cache() public, and use it aggressively in some tests that JIT many Funcs in a ThreadPool: the issue here is that we don't reclaim JIT-related memory until the Func is destroyed, and the structure of correctness_vector_math and correctness_boundary_conditions is such that we can have dozens in existence *per thread*; this greatly increases transient memory pressure, even though we generally never need more than a single one to be in existence at a time.

### DIFF
--- a/test/correctness/vector_math.cpp
+++ b/test/correctness/vector_math.cpp
@@ -142,7 +142,7 @@ bool test(int lanes, int seed) {
     Func f1;
     f1(x, y) = input(x, y) + input(x+1, y);
     f1.vectorize(x, lanes);
-    Buffer<A> im1 = f1.realize(W, H);
+    Buffer<A> im1 = f1.realize(W, H); f1.invalidate_cache();
 
     for (int y = 0; y < H; y++) {
         for (int x = 0; x < W; x++) {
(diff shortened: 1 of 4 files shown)


steven-johnson:
(I noticed this in the wasm-prototype thread, where the current naive JIT approach causes a large amount of address space to be reserved for each extant jitted Func, exhausting memory space quickly in some cases. But reducing memory pressure on 'normal' systems seems like a good idea too.)

test/correctness/vector_math.cpp
```diff
@@ -602,7 +602,7 @@ bool test(int lanes, int seed) {
     Func f22;
     f22(x, y) = absd(input(x, y), input(x+1, y));
     f22.vectorize(x, lanes);
-    Buffer<typename with_unsigned<A>::type> im22 = f22.realize(W, H);
+    Buffer<typename with_unsigned<A>::type> im22 = f22.realize(W, H); f22.invalidate_cache();
```
abadams: Is this equivalent to making a temporary Pipeline to hold the state?
```
im22 = Pipeline(f22).realize(W, H);
```

If so, having an object represent the lifetime seems cleaner than exposing the invalidate method.

steven-johnson: True. (Or, alternately, wrapping the calls in {} to limit the lifespan.

Re: exposing the method, the method was *already* public on Pipeline; given that, adding to Func seems appropriate regardless.

steven-johnson:
Simpler option on the way.

Issue #3669, opened 2019-02-14
Intermittent failures in simd_op_check


steven-johnson:
e.g., https://buildbot.halide-lang.org/master/#/builders/39/builds/340

```
The vector and scalar versions of op_ld3_456 disagree. Maximum error: 4
../halide/Makefile:1656: recipe for target 'quiet_correctness_simd_op_check' failed

```

steven-johnson:
FYI, I happened to get one of the intermittent failures locally (x86-64-osx), and captured the assembly output:

```
The vector and scalar versions of op_shufps_182 disagree. Maximum error: 1.64589e+14
Error assembly:
  .section  __TEXT,__text,regular,pure_instructions
  .section  __TEXT,__literal4,4byte_literals
  .p2align  2               ## -- Begin function error_op_shufps_182
LCPI0_0:
  .long 4286578688              ## float -Inf
  .section  __TEXT,__text,regular,pure_instructions
  .globl  _error_op_shufps_182
  .p2align  4, 0x90
_error_op_shufps_182:                   ## @error_op_shufps_182
  .cfi_startproc
## %bb.0:                               ## %entry
  pushq %r15
  .cfi_def_cfa_offset 16
  pushq %r14
  .cfi_def_cfa_offset 24
  pushq %r13
  .cfi_def_cfa_offset 32
  pushq %r12
  .cfi_def_cfa_offset 40
  pushq %rbx
  .cfi_def_cfa_offset 48
  subq  $16, %rsp
  .cfi_def_cfa_offset 64
  .cfi_offset %rbx, -48
  .cfi_offset %r12, -40
  .cfi_offset %r13, -32
  .cfi_offset %r14, -24
  .cfi_offset %r15, -16
  movq  96(%rsp), %rax
  movq  16(%rax), %r15
  movq  16(%rdi), %rbx
  xorl  %r12d, %r12d
  movl  $393220, %esi           ## imm = 0x60004
  xorl  %edi, %edi
  callq _halide_malloc
  movq  %rax, %r14
  .p2align  4, 0x90
LBB0_1:                                 ## %"for op_shufps_182$1.s0.y"
                                        ## =>This Loop Header: Depth=1
                                        ##     Child Loop BB0_2 Depth 2
  xorl  %ecx, %ecx
  .p2align  4, 0x90
LBB0_2:                                 ## %"for op_shufps_182$1.s0.x.x"
                                        ##   Parent Loop BB0_1 Depth=1
[...]

PR #3725, opened 2019-03-14
Limit simd_op_check to single-threaded testing as stopgap fix (Issue #3669)

steven-johnson:


steven-johnson:
The more I look into this, the more I think that the TSAN issues may be a red herring: we do have data races, and I am getting LLVM crashes in debug builds, but it's not at all clear to me that this is necessarily the cause of our ongoing simd_op_check flakiness (which tends to be failures of the "vector and scalar results did not match").

abadams:
I thought it was just binding distinct buffers to the same imageparam that was inherently not thread safe - they act as uncontrolled shared state if you do that. In this code, the setting is done outside the parallel loop.

steven-johnson:
I'm going to drop this for now: there are definitely some data races going on, but it's not clear to me that they are related to the failures we're actually seeing; furthermore, the assertion-failures I was seeing inside LLVM yesterday were apparently transient to the local trunk I had synced (I can no longer repro on linux or osx with top-of-trunk).

PR #3726, opened 2019-03-14
Reduce JIT memory pressure in correctness_vector_math

steven-johnson:


zvookin:
Why isn't the solution here just to put scopes around each separate part of the test?

steven-johnson:
> Why isn't the solution here just to put scopes around each separate part of the test?

I thought this was cleaner.

zvookin:
Gotta disagree. Generally scoping things to their use is a standard way to control lifetimes in C++. I also think it should be the standard pattern for writing many separate tests in a single C++ function. Not just for resources, but to make sure one doesn't inadvertently reuse things from one test to another. Intentional reuse can be had by declaring common stuff at top-level scope at the start of the function.

(More structured test infrastructure would have each test in its own function with special support for setup and teardown, etc... That requires ... more stuff ... and maybe more lines per test. The braces are simple and a good compromise.)

zvookin:
Oh and while I'm at it, a comment above each open brace saying what the test tests would be really nice too.

steven-johnson:
I disagree substantially -- this PR fixes the issue with minimum of confusion and code delta -- but it's not a point worth arguing. I'll offer something else.

PR #3727, opened 2019-03-14
Fix broken trace code.

dzo:
Using malloc to allocate space for the TraceBuffer object doesn't call the constructor so
the member variables (including the shared spinlock) may not be initialised to zero.
It seems that new isn't allowed in the runtime so I've added some init methods to fix this.

### DIFF
--- a/src/runtime/tracing.cpp
+++ b/src/runtime/tracing.cpp
@@ -65,6 +65,10 @@ public:
         __sync_fetch_and_and(&lock, ~exclusive_held_mask);
     }
 
+    __attribute__((always_inline)) void init() {
+    	lock=0;
+    }
+
     SharedExclusiveSpinLock() : lock(0) {}
 };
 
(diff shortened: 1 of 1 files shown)


src/runtime/tracing.cpp
```diff
@@ -65,6 +65,10 @@ class SharedExclusiveSpinLock {
         __sync_fetch_and_and(&lock, ~exclusive_held_mask);
     }
 
+    __attribute__((always_inline)) void init() {
+    	lock=0;
```
steven-johnson: lock = 0;

src/runtime/tracing.cpp
```diff
@@ -132,6 +136,12 @@ class TraceBuffer {
         lock.release_shared();
     }
 
+    __attribute__((always_inline)) void init() {
+    	cursor=0;
```
steven-johnson: cursor = 0;
overage = 0;

steven-johnson:
https://github.com/halide/Halide/pull/3730 looks clean, ready to merge?

PR #3728, opened 2019-03-14
Grab Bag of minor changes

steven-johnson:
These are cleanups/refactors from the wasm branch that are relatively incidental; proposing to land them in master to reduce deltas from that branch:
- compile_jit() returns void rather than a void* (since not all JITs can easily return a real fn ptr here)
- add_extern_for_export() returns the Symbol it added
- JITCallArgs use new[] / delete[]
- replaced a stray std::cerr with debug(0)
- whitespace nits
- spelling nits
- fake_thread_pool only accepts setting num_threads to exactly 1
- add Type::is_int_or_uint()
- error_broken_promise should use get_jit_target_from_environment() instead of "host"
- simd_op_check logs the host and HL_TARGET values at the start (for ease of tracking down issues later on if failures occur)

### DIFF
--- a/test/correctness/simd_op_check.cpp
+++ b/test/correctness/simd_op_check.cpp
@@ -2148,6 +2148,9 @@ check("v*.w += vrmpy(v*.b,v*.b)", hvx_width, i32_1 + i32(i8_1)*i8_1 + i32(i8_2)*
 int main(int argc, char **argv) {
     Test test;
 
+    printf("host is:      %s\n", get_host_target().to_string().c_str());
+    printf("HL_TARGET is: %s\n", get_target_from_environment().to_string().c_str());
+
     if (argc > 1) {
         test.filter = argv[1];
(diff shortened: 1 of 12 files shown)


test/correctness/simd_op_check.cpp
```diff
@@ -2148,6 +2148,9 @@ check("v*.w += vrmpy(v*.b,v*.b)", hvx_width, i32_1 + i32(i8_1)*i8_1 + i32(i8_2)*
 int main(int argc, char **argv) {
     Test test;
 
+    printf("host is:      %s\n", get_host_target().to_string().c_str());
+    printf("HL_TARGET is: %s\n", get_target_from_environment().to_string().c_str());
```
abadams: +inf

steven-johnson: +NaN

PR #3732, opened 2019-03-15
Add support for WebAssembly

steven-johnson:
This branch represents an effort to support WebAssembly (Wasm) code generation from Halide, using the LLVM backend. There isn't yet a timetable for landing this code in the Halide master branch. It is very much a work-in-progress at this point; be sure to read `README_webassembly.md` all of the current limitations.

As progress is made on this branch, README_webassembly.md will be updated with the status of various features.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,10 @@ endif
 
 SHELL = bash
 CXX ?= g++
+# EMCC is the tool that invokes Emscripten
 EMCC ?= emcc
+# WASM_SHELL is the shell tool used to run AOT-compiled WebAssembly.
+# (Node could be used instead.)
 WASM_SHELL ?= d8
 PREFIX ?= /usr/local
 LLVM_CONFIG ?= llvm-config
@@ -59,7 +62,8 @@ CLANG ?= clang
 CLANG_VERSION = $(shell $(CLANG) --version)
 LLVM_BINDIR = $(shell $(LLVM_CONFIG) --bindir | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')
 LLVM_LIBDIR = $(shell $(LLVM_CONFIG) --libdir | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')
-# TODO: apparently no llvm_config flag to get canonical paths to tools
+# Apparently there is no llvm_config flag to get canonical paths to tools,
+# so we'll just construct one relative to --src-root and hope that is stable everywhere.
 LLVM_LLD_INCLUDE_DIR = $(shell $(LLVM_CONFIG) --src-root | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')/tools/lld/include
 LLVM_SYSTEM_LIBS=$(shell ${LLVM_CONFIG} --system-libs --link-static | sed -e 's/[\/&]/\\&/g')
 LLVM_AS = $(LLVM_BINDIR)/llvm-as
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,8 @@ endif
 
 SHELL = bash
 CXX ?= g++
+EMCC ?= emcc
+WASM_SHELL ?= d8
 PREFIX ?= /usr/local
 LLVM_CONFIG ?= llvm-config
 LLVM_COMPONENTS= $(shell $(LLVM_CONFIG) --components)
@@ -57,10 +59,12 @@ CLANG ?= clang
 CLANG_VERSION = $(shell $(CLANG) --version)
 LLVM_BINDIR = $(shell $(LLVM_CONFIG) --bindir | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')
 LLVM_LIBDIR = $(shell $(LLVM_CONFIG) --libdir | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')
+# TODO: apparently no llvm_config flag to get canonical paths to tools
+LLVM_LLD_INCLUDE_DIR = $(shell $(LLVM_CONFIG) --src-root | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')/tools/lld/include
 LLVM_SYSTEM_LIBS=$(shell ${LLVM_CONFIG} --system-libs --link-static | sed -e 's/[\/&]/\\&/g')
 LLVM_AS = $(LLVM_BINDIR)/llvm-as
 LLVM_NM = $(LLVM_BINDIR)/llvm-nm
-LLVM_CXX_FLAGS = -std=c++11  $(filter-out -O% -g -fomit-frame-pointer -pedantic -W% -W, $(shell $(LLVM_CONFIG) --cxxflags | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g;s/-D/ -D/g;s/-O/ -O/g'))
+LLVM_CXX_FLAGS = -std=c++11  $(filter-out -O% -g -fomit-frame-pointer -pedantic -W% -W, $(shell $(LLVM_CONFIG) --cxxflags | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g;s/-D/ -D/g;s/-O/ -O/g')) -I$(LLVM_LLD_INCLUDE_DIR)
 OPTIMIZE ?= -O3
 OPTIMIZE_FOR_BUILD_TIME ?= -O0
 
@@ -108,6 +112,7 @@ WITH_POWERPC ?= $(findstring powerpc, $(LLVM_COMPONENTS))
 WITH_PTX ?= $(findstring nvptx, $(LLVM_COMPONENTS))
 # AMDGPU target is WIP
 WITH_AMDGPU ?= $(findstring amdgpu, $(LLVM_COMPONENTS))
+WITH_WEBASSEMBLY ?= $(findstring webassembly, $(LLVM_COMPONENTS))
 WITH_OPENCL ?= not-empty
 WITH_METAL ?= not-empty
 WITH_OPENGL ?= not-empty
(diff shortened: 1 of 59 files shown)


steven-johnson:
I think this is ready to review -- for non-SIMD usage I think it's fairly solid, given the constraints in the README.

(Obviously this will need to be followed up with some buildbot attention to ensure we get at least basic build/test coverage.)

Makefile
```diff
@@ -47,6 +47,8 @@ endif
 
 SHELL = bash
 CXX ?= g++
+EMCC ?= emcc
```
zvookin: Can we add a comment mentioning "emscripten"?

Makefile
```diff
@@ -47,6 +47,8 @@ endif
...
+WASM_SHELL ?= d8
```
zvookin: Is there another standard option here? If so, might be worth a comment.

Makefile
```diff
@@ -57,10 +59,12 @@ CLANG ?= clang
 CLANG_VERSION = $(shell $(CLANG) --version)
 LLVM_BINDIR = $(shell $(LLVM_CONFIG) --bindir | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')
 LLVM_LIBDIR = $(shell $(LLVM_CONFIG) --libdir | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')
+# TODO: apparently no llvm_config flag to get canonical paths to tools
```
zvookin: Not sure what the todo is here. If it is at all important, add an issue before merging.

Makefile
zvookin: Also might be worth mentioning that "wasm" is "WebAssembly" since both appear in this file and it is a good intro point for those not already in the know.

README_webassembly.md
```diff
@@ -0,0 +1,104 @@
...
+
+As WebAssembly itself is still under active development, Halide's support has some limitations. Some of the most important:
+
+- We require using LLVM 9+ for Wasm codegen; earlier versions of LLVM may work, but have not been tested, and so are currently deliberately excluded.
+- SIMD can be enabled via Target::WasmSimd128, but is likely to require building with LLVM 9+ and running in V8 7.5 or later to work properly. (Earlier versions of V8 should work for non-SIMD usage.)
+- Multithreading is not yet ready to be supported -- there isn't even a Feature flag to enable it yet -- but we would like to add support for this in the future.
+- Halide's JIT for Wasm is extremely limited and really useful only for internal testing purposes.
+
+# Additional Tooling Requirements:
+- In additional to the usual install of LLVM and Clang, you'll need wasm-ld (via LLVM/tools/lld). All should be v9.x+ (current trunk)
```
zvookin: Probably best to write "current truck as of <date>".

README_webassembly.md
```diff
@@ -0,0 +1,104 @@
...
+
+Note that while some of these limitations may be improved in the future, some are effectively intrinsic to the nature of this problem. Realistically, this JIT implementation is intended solely for running Halide self-tests (and even then, a number of them are fundamentally impractical to support in a hosted-Wasm environment and are blacklisted).
+
+In sum: don't plan on using Halide JIT mode with Wasm unless you are working on the Halide library itself.
+
+# To Use Halide For WebAssembly:
+
+- Ensure WebAssembly is in LLVM_TARGETS_TO_BUILD:
+```
+-DLLVM_TARGETS_TO_BUILD="X86;ARM;NVPTX;AArch64;Mips;PowerPC;Hexagon;WebAssembly
```
zvookin: Is it not included in the "all" option as of LLVM 9?

src/CodeGen_LLVM.cpp
```diff
@@ -499,19 +513,78 @@ MangledNames get_mangled_names(const LoweredFunc &f, const Target &target) {
 
 }  // namespace
 
-std::unique_ptr<llvm::Module> CodeGen_LLVM::compile(const Module &input) {
+// Make a wrapper to call the function with an array of pointer
+// args. This is easier for the JIT to call than a function with an
+// unknown (at compile time) argument list.
+llvm::Function *CodeGen_LLVM::add_trampoline_wrapper(llvm::Function *fn,
```
abadams: How similar is this to the argv entrypoint?

src/Pipeline.cpp
```diff
@@ -345,6 +350,43 @@ vector<Argument> Pipeline::infer_arguments(Stmt body) {
...
+                // call sites which use vectors will have to be scalarized into a
+                // separate call per lane. Not sure there is anywhere to get
+                // information to make a distinction in the current design.
+                std::vector<Type> arg_types;
+                for (Expr e : op->args) {
+                    arg_types.push_back(e.type().element_of());
+                }
+                ExternCFunction f(address, ExternSignature(op->type.element_of(), op->type.bits() == 0, arg_types));
+                JITExtern jit_extern(f);
+debug(0) << "FindExterns adds: " << op->name << "\n";
```
abadams: nit: no indentation here?

steven-johnson: Whups, that shouldn't have been included.

src/CodeGen_LLVM.cpp
steven-johnson: Similar-ish. It should really be reconciled to avoid somewhat-duplicated code.

steven-johnson:
Any other comments on this? I'd like to land this soon, if possible.

src/CodeGen_LLVM.cpp
abadams: The comment should explain why the argv version doesn't serve your purposes. It seems like the only difference is that the scalars are done indirectly. Is that because there are scalar types wider than the pointer type in wasm?

src/CodeGen_LLVM.h
```diff
@@ -70,6 +71,27 @@ class CodeGen_LLVM : public IRVisitor {
     /** Initialize internal llvm state for the enabled targets. */
     static void initialize_llvm();
 
+    /** Setup state for ad-hoc codegen. Used in JIT outside of compiling an entire Halide module.
```
abadams: "TODO: This probably shouldn't exist" is not confidence-inspiring. Could you clarify? E.g. on init_for_codegen something like "only used directly by the wasm backend for ad-hoc JIT testing"

src/CodeGen_LLVM.cpp
abadams: Oh, I think I misunderstood what this is for. I thought it was another wrapper for the pipeline, but it's a wrapper for extern functions?

src/WasmExecutor.cpp
```diff
@@ -0,0 +1,1583 @@
...
+//
+// Note that this means that the functor must be able to instantiate its
+// operator() for all the Halide scalar types; it also means that all those
+// variants *will* be instantiated (increasing code size), so this approach
+// should only be used when strictly necessary.
+template<template<typename> class Functor, typename... Args>
+auto dynamic_type_dispatch(const halide_type_t &type, Args&&... args) ->
+    decltype(std::declval<Functor<uint8_t>>()(std::forward<Args>(args)...)) {
+
+#define HANDLE_CASE(CODE, BITS, TYPE) \
```
abadams: I don't think this macro buys you much in terms of clarity. E.g. none of the args are used more than once, and it's a single line.

src/WasmExecutor.cpp
```diff
@@ -0,0 +1,1583 @@
...
+
+    BDMalloc *bdmalloc = (BDMalloc *) context->GetAlignedPointerFromEmbedderData(kBDMallocPtr);
+    if (!bdmalloc->inited()) {
+        int32_t heap_base = context->GetEmbedderData(kHeapBase)->Int32Value(context).ToChecked();
+
+        Local<Object> memory_value = context->GetEmbedderData(kWasmMemoryObject).As<Object>();  // really a WasmMemoryObject
+        Local<Object> buffer_string = context->GetEmbedderData(kString_buffer).As<Object>();
+        Local<ArrayBuffer> wasm_memory = Local<ArrayBuffer>::Cast(memory_value->Get(buffer_string));
+
+        wdebug(0)<<"heap_base is: "<<heap_base<<"\n";
```
abadams: Please add spaces around << in this file

src/WasmExecutor.cpp
```diff
@@ -0,0 +1,1583 @@
...
+    wdebug(0) << "  padding = " << buf->padding << "\n";
+    wdebug(0) << "}\n";
+#endif
+}
+
+static_assert(sizeof(halide_type_t) == 4, "halide_type_t");
+static_assert(sizeof(halide_dimension_t) == 16, "halide_dimension_t");
+static_assert(sizeof(wasm_halide_buffer_t) == 40, "wasm_halide_buffer_t");
+
+size_t size_in_bytes(const halide_buffer_t *b) {
```
abadams: How does this differ from b->size_in_bytes()?

src/WasmExecutor.cpp
steven-johnson: Heh: I think that this is a fine example of one of the few remaining legit uses of the preprocessor; the 'full' line has the differences lost in the noise when I read it, whereas the macro'ed version makes the differences much more obvious.

I'll change this if there's opposition, of course, but IMHO this is clearly more readable than the alternative.

src/WasmExecutor.cpp
steven-johnson: A lot of the debugging in here started as quick-hack-for-debugging and then migrated to leave-in-long-term; I should have done more cleaning. Done.

src/WasmExecutor.cpp
steven-johnson: ...I forgot that we had some of these methods defined on halide_buffer_t (as well as Buffer<>), so, it doesn't. Removing.

src/CodeGen_LLVM.cpp
steven-johnson: That's correct. (It's basically directly lifted from @zvookin's earlier work on a JS backend.)

src/CodeGen_LLVM.h
steven-johnson: Guilty as charged. Let me do another pass to see if I can make this less bad (perhaps consolidating with the existing argv-wrapper code if possible).

src/WasmExecutor.cpp
abadams: I will defer to your judgement.

PR #3733, opened 2019-03-15
Add RVar to names brought into scope by Generator

dsharletg:


zvookin:
Approving, but I expect this to force some code changes in at least a few places. Hoping I'm wrong.

dsharletg:
Hmm, I wouldn't expect that, I think this just makes `using Halide::RVar` redundant (if that was being done already... I was doing that in some of my code).

PR #3734, opened 2019-03-15
Refactor HL_LLVM_ARGS parsing

steven-johnson:
Currently, we parse the HL_LLVM_ARGS environment variable exactly once (when we init LLVM). This proposes to refactor that to allow resetting the LLVM flags once per compile:

-- defines CodeGen_LLVM::set_llvm_command_line_options() as a virtual method; the default implementation resets all LLVM flags to their default values, then re-parses from HL_LLVM_ARGS as before

-- subclasses are allowed to override this to make additional settings; the existing usage in Codgen_Hexagon uses this approach now (instead of overriding `compile()`, and an override for Codegen_X86 was added to set `enable-mssa-loop-dependency=1`, which can improve compile time substantially in some cases (see @alinas for details)

That said: I'm not 100% satisfied with this approach, because it means we're relying on mutating LLVM flags (aka "global variables") to get the settings we want; if the Halide compiler is being used from multiple threads in the same process, you can guess what sorts of shenanigans could potentially ensue. (That said, the risk is small, because in practice, we ~never use the same process to generate different architectures in multiple threads... at least, so far.)

On the other hand, if we need to set things in LLVM that are only controllable via command-line flags, our options would seem to be:
(1) take this approach and hope things work
(2) restrict LLVM usage to a single thread at a time in Halide
(3) patch LLVM to allow the settings to be controlled per-LLVMContext

This PR is of course approach 1; it's not optimal, but it may be acceptable for the short term. Opinions welcome.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -60,34 +60,25 @@ CodeGen_Hexagon::CodeGen_Hexagon(Target t) : CodeGen_Posix(t) {
 }
 
 std::unique_ptr<llvm::Module> CodeGen_Hexagon::compile(const Module &module) {
-    auto llvm_module = CodeGen_Posix::compile(module);
-
-    // TODO: This should be set on the module itself, or some other
-    // safer way to pass this through to the target specific lowering
-    // passes. We set the option here (after the base class'
-    // implementation of compile) because it is the last
-    // Hexagon-specific code to run prior to invoking the target
-    // specific lowering in LLVM, minimizing the chances of the wrong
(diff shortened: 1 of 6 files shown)


steven-johnson:
Yeah, this approach definitely isn't going to work reliably without more thought; even using multiple threads within the same architecture can trigger failures, e.g. correctness_simd_op_check will often fail with LLVM reporting that you can't specify `-enable-mssa-loop-dependency` more than once. Leaving this open for now, for discussion.

steven-johnson:
Obsoleted by other changes.

PR #3745, opened 2019-03-20
Upgrade of Travis setup

steven-johnson:
- Upgrade Trusty -> Xenial (14.04 -> 16.04)
- Add LLVM8 to test matrik
- Use the Travis+Xenial default of gcc5.4 instead of gcc4.8 (which we no longer build or test with anywhere else)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ compiler:
   - gcc
 env:
   global:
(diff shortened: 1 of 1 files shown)


steven-johnson:
Travis is being cranky; I'm going to try making these changes in smaller steps

PR #1536, opened 2016-10-03
Remove custom strided store code generation.

alinas:
Generate a single shuffle and a single vector store, which is
turned into a strided store by LLVM's interleaved access pass.
This is dependent on the following LLVM patch:
https://reviews.llvm.org/D23646

PR #3746, opened 2019-03-20
Remove custom strided store code generation for ARM

steven-johnson:
Note that this is a port of https://github.com/halide/Halide/pull/1536, which is pretty ancient, but still possibly useful. Both correctness and performance testing needs to be done. Also, it's not clear whether this approach should be adapted for all targets (ie moved into Codegen_LLVM.cpp).

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -12,6 +12,16 @@
 #include "Simplify.h"
 #include "Util.h"
 
+/*
+    Older versions of LLVM (< 4.0) required some custom code to
+    get strided stores to work well on ARM; more recent versions
+    do a good job without the old workarounds. Providing this via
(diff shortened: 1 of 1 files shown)


steven-johnson:
attn @alinas

dsharletg:
I still think this is a good change, but last time I tested it simd_op_check had a few failing cases. It's possible simd_op_check should simply be changed if the generated code is different, but IIRC it was things like vld2/3 not generating, which are important and hard to imagine not being the best instruction for that pattern.

steven-johnson:
Looks like `correctness_deinterleave4` is (also?) failing. This change will really need to be vetted on arm32 and arm64 machines. (My main interest here was in rescuing the change, which had become obscured by the git history rewrite.)

PR #3749, opened 2019-03-21
Silence warnings in Target::get_runtime_compatible_target()

steven-johnson:
We emit user_warnings when returning false, which generates useless and confusing warnings from test_internal. Convert to debug(1) output; the caller should be checking the result in any event.


abadams:
How does this affect the output for things like "-r x86-64,arm-32" as a generator argument?

steven-johnson:
The Generator usage checks the result and fails with user_error for a false result (Generator.cpp:1010)

alexreinking:
Looks good to me. Sorry for the output barfing!

PR #3752, opened 2019-03-22
is_monotonic() should not fail for Exprs with strings

steven-johnson:
require() Exprs can legitimately have these. Add a test case that would catch it.

### DIFF
--- a/src/Monotonic.cpp
+++ b/src/Monotonic.cpp
@@ -28,7 +28,9 @@ class MonotonicVisitor : public IRVisitor {
     }
 
     void visit(const StringImm *) override {
-        internal_error << "Monotonic on String\n";
+        // No: require() Exprs can includes Strings, which
+        // shouldn't affect the outcome. Just ignore them.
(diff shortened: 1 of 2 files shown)


src/Monotonic.cpp
```diff
@@ -28,7 +28,9 @@ class MonotonicVisitor : public IRVisitor {
     }
 
     void visit(const StringImm *) override {
-        internal_error << "Monotonic on String\n";
+        // No: require() Exprs can includes Strings, which
+        // shouldn't affect the outcome. Just ignore them.
+        // internal_error << "Monotonic on String\n";
     }
```
abadams: I think this should be ```result = Monotonic::Constant```, because you know a string doesn't vary wrt a Var.

abadams:
It would be a good idea to handle Call::require in the Call visitor too, because require(x > 3, x) is monotonic in x.

steven-johnson:
PTAL

PR #3756, opened 2019-03-24
Add check for the overflow in Python bindings

yamaguchi1024:
Add check for ssize_t -> int32_t overflow. Add the corresponding test  to correctness/buffer.py.

### DIFF
--- a/python_bindings/src/PyBuffer.cpp
+++ b/python_bindings/src/PyBuffer.cpp
@@ -231,7 +231,9 @@ class PyBuffer : public Buffer<> {
         std::vector<halide_dimension_t> dims;
         dims.reserve(info.ndim);
         for (int i = 0; i < info.ndim; i++) {
-            // TODO: check for ssize_t -> int32_t overflow
+            if (INT_MAX < info.shape[i] || INT_MAX < (info.strides[i] / t.bytes())) {
+                throw py::value_error("Out of range arguments to make_dim_vec.");
+            }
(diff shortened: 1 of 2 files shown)


python_bindings/src/PyBuffer.cpp
```diff
@@ -231,7 +231,8 @@ class PyBuffer : public Buffer<> {
         std::vector<halide_dimension_t> dims;
         dims.reserve(info.ndim);
         for (int i = 0; i < info.ndim; i++) {
-            // TODO: check for ssize_t -> int32_t overflow
+            if (INT_MAX < info.shape[i] || INT_MAX < (info.strides[i] / t.bytes()))
```
steven-johnson: nit: enclose all if clauses in braces, even if a single line.

yamaguchi1024:
@steven-johnson Thanks for your comment! I updated the patch. As said in https://chezjrk.slack.com/archives/C752YMC95/p1553338460001800, I'm trying to contribute and didn't actually hit this bug. However the size of `info.shape[i]` was 64bit on my machine, so there are still a possibility of overflow here.

steven-johnson:
I like this change, but I'd like to see a test added before we land it.

yamaguchi1024:
Hi @steven-johnson , sorry for being so slow! I just added the test to python_bindings/correctness/buffer.py, is it how it supposed to look like?

steven-johnson:
LGTM. Let me run it thru the buildbots.

steven-johnson:
Thanks for the fix!

PR #3759, opened 2019-03-25
Fix sign error in SolveExpression

steven-johnson:


steven-johnson:
I'm a little surprised this hasn't bitten us already.

steven-johnson:
(Note: I did a quick once-over of SolveExpressions to see if anything else was obviously wrong in this same vein, and didn't see anything, but it's easy to miss something, as this fix suggests; it's probably worth having another set of eyes review it)

PR #3760, opened 2019-03-25
Simplify similar IfThenElse even when 'rest' is also Block

dsharletg:


src/Simplify_Stmts.cpp
```diff
@@ -352,8 +352,12 @@ Stmt Simplify::visit(const Block *op) {
     // Check if both halves start with a let statement.
     const LetStmt *let_first = first.as<LetStmt>();
     const LetStmt *let_rest = rest.as<LetStmt>();
+    const Block *block_rest = rest.as<Block>();
     const IfThenElse *if_first = first.as<IfThenElse>();
-    const IfThenElse *if_rest = rest.as<IfThenElse>();
+    const IfThenElse *if_next =
```
abadams: I found this kinda awkward. How about:

```
    const IfThenElse *if_next = block_rest ? block_rest->first.as<IfThenElse> : rest.as<IfThenElse>();
```

src/Simplify_Stmts.cpp
```diff
@@ -379,34 +383,42 @@ Stmt Simplify::visit(const Block *op) {
...
         // first one must also be true.
-        Stmt then_case = mutate(Block::make(if_first->then_case, if_rest));
+        Stmt then_case = mutate(Block::make(if_first->then_case, if_next));
         Stmt else_case = mutate(if_first->else_case);
-        return IfThenElse::make(if_first->condition, then_case, else_case);
+        Stmt result = IfThenElse::make(if_first->condition, then_case, else_case);
+        if (if_rest.defined()) {
+            result = Block::make(result, if_rest);
+        }
+        return result;
```
abadams: What happens with a chain of if statements all with the same condition? I think because we already mutated rest, it'll be collected and OK. But maybe add a new test case with three IfThenElses in a row followed by an evaluate.

Issue #3755, opened 2019-03-22
Recent dramatic slowdown in compute_at + GuardWithIf splits


steven-johnson:
Sometime since ~Mar 13 (commit 305f2cd392c72f62204dcb5af9981aea9b478397), compilation of simple code using these patterns has gotten dramatically slower, with the bulk of the time (apparently) spent in LLVM. A very simple repro:

```
    Var x, xo, xi;
    ImageParam input(UInt(8), 1);

    Func f, output;
    f(x) = input(x);
    output(x) = f(x);

    f.compute_at(output, xo).vectorize(x, 8, TailStrategy::GuardWithIf);
    output.split(x, xo, xi, 256, TailStrategy::GuardWithIf);

    HALIDE_TIC;
    output.compile_jit();
    HALIDE_TOC;
```

On my Mac laptop, the call to compile_jit() now takes ~ 75 seconds. (More complex pipelines can take dramatically longer.)

steven-johnson:
FYI, the Halide-level IR for this is different for 'before' and 'after', the change above, but nothing that looks obviously pathological to my eye. (Interestingly, they both produce bitcode outputs of nearly the same size, but one takes much much longer to do so.)

@alinas is going to take a look from the LLVM side, to rule out the possibility that a minor change on our side has triggered some latent-but-undiscovered pathology in trunk LLVM.

steven-johnson:
git bisect confirms the slowdown injection point is commit 94dd9b35be583a76e51b89ea805ce5cf936475eb from https://github.com/halide/Halide/pull/3720 -- tagging @abadams as an FYI.

@alinas indicates that there is possible pathology in llvm::ScalarEvolution pass (and evidence of at least one Halide-related cap there already), so there's a real possibility that this is a case of Halide tickling a latent LLVM problem via code restructuring; that said, it's not clear yet if the best fix is on the Halide side or the LLVM side (or some combination).

alinas:
We need to fix this on the LLVM side. See: https://bugs.llvm.org/show_bug.cgi?id=41225

abadams:
I don't have access to the IR, but looking at the .ll from that bug report, it looks like a deeply nested min/max tree in the inner loop. I wonder how many of those terms actually depend on the loop variable. We have GroupLoopInvariants in LICM to try to separate the terms that don't depend on the loop variable for chains of additions, but nothing similar for chains of mins/maxs. Might that help?

abadams:
[...]

PR #3761, opened 2019-03-25
Add some min/max simplifier rules

steven-johnson:
- min(min(x,y)+c,x) -> min(x, y+c) [iff c > 0]
- max(max(x,y)-c,x) -> max(x, y-c) [iff c > 0]

This eliminates some very long nested min() expressions seen in https://github.com/halide/Halide/issues/3755.

### DIFF
--- a/src/Simplify_Max.cpp
+++ b/src/Simplify_Max.cpp
@@ -113,6 +113,7 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(min(max(x, y), z), y), max(min(x, z), y)) ||
              rewrite(max(min(max(y, x), z), y), max(y, min(x, z))) ||
              rewrite(max(max(x, c0), c1), max(x, fold(max(c0, c1)))) ||
+             rewrite(max(max(x, y) + c0, x), max(x, y + c0), c0 < 0) ||
 
              (no_overflow(op->type) &&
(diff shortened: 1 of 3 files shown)


src/Simplify_Max.cpp
```diff
@@ -113,6 +113,7 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(min(max(x, y), z), y), max(min(x, z), y)) ||
              rewrite(max(min(max(y, x), z), y), max(y, min(x, z))) ||
              rewrite(max(max(x, c0), c1), max(x, fold(max(c0, c1)))) ||
+             rewrite(max(max(x, y) + c0, x), max(x, y + c0), c0 < 0) ||
```
abadams: Looks correct to me. While you're here, please also add the case where c0 > 0: `max(x, y) + c0`, and versions with x, y swapped in the innermost max, and the equivalent rules to Simplify_Min.

abadams:
In the rule-based simplifier, any rule with two instances of a wildcard on the LHS and only one instance on the RHS is considered safe/profitable. It strictly reduces the number of subexpressions so it can't possible cause an infinite simplifier loop.

steven-johnson:
Added:
- min(min(x,y)+c,x) -> min(x,y)+c [iff c < 0]
- max(max(x,y)-c,x) -> max(x,y)-c [iff c < 0]

Issue #3621, opened 2019-01-25
Missed simplification of predicated store expressions


dsharletg:
I'm writing a schedule that uses GuardWithIf to deal with unaligned output sizes, and seeing lots of if statements with conditions like this:

[...]

PR #3765, opened 2019-03-26
Simplify boolean combinations of inequalities using min/max

abadams:
Spotted some dumb code in vectorized if conditions of the form `(x < y && x < y + 1 && x < y + 2 ...)`

I believe Dillon has complained about this in the past. Here are some new simplifier rules to clean it up a little.


steven-johnson:
> (x < y && x < y + 1 && x < y + 2 ...)

when you *really* want to be *sure* that x is less than y

dsharletg:
Might fix #3621

PR #3766, opened 2019-03-26
Add missing critical point in Simplify_Div

### LINKED ISSUE #3764: fuzz_simplify failure
From https://buildbot.halide-lang.org/master/#/builders/73/builds/10: 

```
FAILED TEST: fuzz_simplify
Simplify fuzz test seed: 1553566478
a = 29
b = 105
c = -36
d = int32(c)
e = int32(a)
((int32(((int16)-24*(int16)-14))/max(int32(int16(b)), int32((uint16)38))) + int32(a))
((336/max(int32(int16(b)), 38)) + int32(a))
In vector lane 0:
((int32(((int16)-24*(int16)-14))/max(int32(int16(b)), int32((uint16)38))) + int32(a)) -> 32
((336/max(int32(int16(b)), 38)) + int32(a)) -> 37
../halide/Makefile:1666: recipe for target 'quiet_correctness_fuzz_simplify' failed
make: *** [quiet_correctness_fuzz_simplify] Error 1
```

abadams:
Fixes #3764 in two ways

- Adds a missing critical point to be considered in the simplify_div bounds logic (the limit as the denominator goes to +/- infinity is zero, provided the denominator is known to be positive or negative). 
- But the denominator shouldn't have been unbounded on one side, because it was the result of constant folding. Fix this by tracking bounds through casts of constants.

### DIFF
--- a/src/Simplify_Cast.cpp
+++ b/src/Simplify_Cast.cpp
@@ -4,7 +4,9 @@ namespace Halide {
 namespace Internal {
 
 Expr Simplify::visit(const Cast *op, ExprInfo *bounds) {
-    // We don't try to reason about bounds through casts for now
+    // We generally don't track bounds through casts, with the
+    // exception of casts that constant-fold to a signed integer, so
+    // we don't need the bounds of the value.
     Expr value = mutate(op->value, nullptr);
 
     if (may_simplify(op->type) && may_simplify(op->value.type())) {
--- a/src/Simplify_Cast.cpp
+++ b/src/Simplify_Cast.cpp
@@ -30,12 +30,8 @@ Expr Simplify::visit(const Cast *op, ExprInfo *bounds) {
                    const_float(value, &f) &&
                    std::isfinite(f)) {
             // float -> int
-            int64_t v = safe_numeric_cast<int64_t>(f);
-            if (bounds) {
-                bounds->min_defined = bounds->max_defined = true;
-                bounds->min = bounds->max = v;
-            }
-            return IntImm::make(op->type, v);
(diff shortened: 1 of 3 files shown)


src/Simplify_Cast.cpp
```diff
@@ -41,6 +46,10 @@ Expr Simplify::visit(const Cast *op, ExprInfo *bounds) {
         } else if (op->type.is_int() &&
                    const_int(value, &i)) {
             // int -> int
+            if (bounds) {
```
steven-johnson: Why don't we need this check elsewhere as well (e.g., the int->uint case below)?

abadams: It's only needed if you want to set the bounds, and it didn't seem worth tracking bounds on non-ints, because you first have to check if they're representable as int64s, and there aren't many (any) rules that exploit the bounds for non-integers.

steven-johnson: Probably worth adding a comment to that effect.

abadams:
Corrected the existing comment at the function entry.

abadams:
Causing new fuzz tester failures - returning correct bounds for casts of constants probably uncovered some other issue in the bounds-based reasoning. Investigating.

abadams:
No, the error was in the new code. I was returning bounds using the values before normalized to the narrower type. Safer to just remutate the generated constant - it's the same code, just plus a function call.

steven-johnson:
LGTM

abadams:
Failure unrelated, and this is a correctness fix.

PR #3768, opened 2019-03-26
performance_vectorize should use natural_vector_size<>()

steven-johnson:
...instead of hardcoded values; this allows experimenting with different HL_JIT_TARGET values more easily.

### DIFF
--- a/test/performance/vectorize.cpp
+++ b/test/performance/vectorize.cpp
@@ -22,7 +22,9 @@ DECL_SOT(float);
 DECL_SOT(double);
 
 template<typename A>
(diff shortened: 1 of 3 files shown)


test/performance/vectorize.cpp
```diff
@@ -95,15 +97,14 @@ int main(int argc, char **argv) {
     bool ok = true;
 
     // Only native vector widths for now
-    ok = ok && test<float>(4);
```
abadams: Note that the original code effectively tested both sse and avx performance avx machines. Not sure if that really matters.

PR #3770, opened 2019-03-26
Grab-bag of fixes from the autoscheduler branch

abadams:
The intent is to reduce the size of the diff with the new autoscheduler branch so that we can merge it soon. These are the main changes in that branch in the src/ folder. The bulk of the remaining changes are all to the apps.


abadams:
The new simplifier rules need tests. Will fix.

PR #3774, opened 2019-03-27
Distinguish more between bounds correct vs bounds loaded

abadams:
Currently computation bounds inference just respects box_touched, which means that for something like `select(x < 100, f(x), g(x))`, f and g are computed over the entire domain. It would be better if we allocated f and g over the entire domain, but just computed f before 100, and just computed g after 100. 

This PR achieves this by treating selects as if_then_else instead when doing computation bounds inference. In allocation bounds inference they're still selects. Also fixed a bug in Bounds along the way (it was checking that the solver terminated with a single copy of the variable, but not that the thing on the LHS was just the variable, and not, e.g. `x%2`).

I'm a little nervous about compile time implications. I would appreciate if people with any particularly sensitive pipelines could try this out.


src/Bounds.cpp
```diff
@@ -2061,17 +2061,21 @@ class BoxesTouched : public IRGraphVisitor {
...
                     const GE *ge = solved.as<GE>();
                     const EQ *eq = solved.as<EQ>();
-                    Expr rhs;
-                    if (lt) {rhs = lt->b;}
-                    if (le) {rhs = le->b;}
-                    if (gt) {rhs = gt->b;}
-                    if (ge) {rhs = ge->b;}
-                    if (eq) {rhs = eq->b;}
+                    Expr lhs, rhs;
+                    if (lt) {lhs = lt->a; rhs = lt->b;}
```
steven-johnson: super-nit: would it be worthwhile to make these if-else since they are all mutually exclusive?

abadams:
It's not urgent. I need this for another feature I'm working on, but I can just merge this into my other branch.

steven-johnson:
Let me pull current master into our code (so that CSE, etc fixes are already in place), then I'll do a test integrate of this if you think it's ready

steven-johnson:
Pulling this into google, I get many hundreds of failures, but all seem to be variants of

```
The pure definition  of Function foo calls function bar in an unbounded way in dimension 0
```

or

```
Symbol not found: baz$3.s0.fx.max
```

abadams:
I'm at a loss for what might be failing. Are any of the failures small enough to turn into a repro? Otherwise I could also cut this into several smaller PRs and we can bisect what's causing the problem.

steven-johnson:
No idea yet -- I was doing bisection on my end but have been sidetracked by more important matters. I hope to get back to it soon.

steven-johnson:
Still getting oodles of 'unbounded' where I didn't before, and I have high-priority tasks ahead of this for the next couple of days. If it's easy to split this PR into smaller pieces that might be an easier test.

steven-johnson:
FYI: if I disable the changes in BoundInference.cpp (but leave the rest), my mystery failures go away.

abadams:
I split the changes in BoundsInference.cpp into two and opened #3859

abadams:
Updated to a form that passes the repro from the other PR (#3859)

steven-johnson:
To test this in Google, I probably need to land #3867 first, then pull Halide in, to avoid unrelated failures.

steven-johnson:
Running a test of this inside Google, and so far so good. Will do a more thorough test and hopefully we can land this later today.

abadams:
Excellent. Getting zero-copy reshapes and concats from the store_with PR would be a good feature to be able to merge, but the concats rely on this PR to work out cleanly.

steven-johnson:
Not seeing any new failures due to this. OK to land.

PR #3777, opened 2019-03-28
Fix bug in posix_error_handler.

matthiaskramm:
This bug caused Halide to swallow error messages. (And, for very long error messages, crash, instead of aborting.)


abadams:
Nice catch.

steven-johnson:
Wow.

PR #3783, opened 2019-03-30
Fix bug in Func::infer_input_bounds()

steven-johnson:
It was doing some dancing on a halide_buffer_t and leaving the host storage in a state inconsistent with the dimensions; in most code this was harmless but in some situations (e.g. WebAssembly JIT mode) it can lead to amusing memory corruption.

### DIFF
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -2908,6 +2908,16 @@ void Func::infer_input_bounds(const std::vector<int> &sizes, const ParamMap &par
     infer_input_bounds(r, param_map);
 }
 
+void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
+                                  const ParamMap &param_map) {
+    vector<int> sizes;
+    if (x_size) sizes.push_back(x_size);
+    if (y_size) sizes.push_back(y_size);
+    if (z_size) sizes.push_back(z_size);
+    if (w_size) sizes.push_back(w_size);
+    infer_input_bounds(sizes, param_map);
+}
+
 OutputImageParam Func::output_buffer() const {
     user_assert(defined())
         << "Can't access output buffer of undefined Func.\n";
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -2897,12 +2897,11 @@ Realization Func::realize(const Target &target,
     return realize(std::vector<int>{}, target, param_map);
 }
 
-void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
-                              const ParamMap &param_map) {
+void Func::infer_input_bounds(const std::vector<int> &sizes, const ParamMap &param_map) {
     user_assert(defined()) << "Can't infer input bounds on an undefined Func.\n";
     vector<Buffer<>> outputs(func.outputs());
     for (size_t i = 0; i < outputs.size(); i++) {
-        Buffer<> im(func.output_types()[i], nullptr, {x_size, y_size, z_size, w_size});
+        Buffer<> im(func.output_types()[i], nullptr, sizes);
         outputs[i] = std::move(im);
     }
     Realization r(outputs);
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -2908,11 +2908,17 @@ void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
         // the halide_buffer_t fields. (We can't use crop because it explicitly
         // disallows expanding the fields in this unsafe manner.)
         Buffer<> im = Buffer<>::make_scalar(func.output_types()[i]);
+        // Deallocate all storage and ensure that host is null; since
+        // we are diddling the dimensions below, the host ptr will no longer
+        // reflect the allocation described by the dimensions, which is usually
+        // fine, unless an extern function wants to try accessing that memory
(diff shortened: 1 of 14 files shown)


src/Func.cpp
```diff
@@ -2908,11 +2908,17 @@ void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
         // the halide_buffer_t fields. (We can't use crop because it explicitly
         // disallows expanding the fields in this unsafe manner.)
         Buffer<> im = Buffer<>::make_scalar(func.output_types()[i]);
```
abadams: I thought the whole point of this code was that the host pointers on the output buffers need to be non-null. Otherwise it could be substantially less clever here - we could just make bounds query buffers. Maybe that's not the case?

abadams: Assuming this is passing tests, maybe at some point in time the outputs needed to be non-null in order to do bounds inference on the inputs.

steven-johnson: > I thought the whole point of this code

That is an excellent question, but I don't know the answer. FWIW, test_correctness seems to pass just fine like this. AFAICT this is only used when we are going to infer the *input* bounds, so the only thing the outputs are even needed for are to provide dimensions.

steven-johnson: It's passing all the tests.

abadams: In that case we can nuke the entire loop below and just say:
```
Buffer<> im(func.output_types()[i], nullptr, sizes);
```
for an appropriate `vector<int>` of sizes.

steven-johnson:
> Yes, the calling convention to infer_input_bounds is goofy

sad face, ok

steven-johnson:
How about I fix that goofy calling convention while I'm at it? It's not a heavily used method.

abadams:
We don't actually want to support people passing zeros there. It's only the way it is to prevent having to write per-dimensionality overloads. Would be better if all the overloads just delegated to one that takes a `vector<int>`

abadams:
(So +1 to fixing it)

steven-johnson:
PTAL

abadams:
I'd rather not break any existing users though. I was thinking the main method would be as you wrote it, but there would also be an overload that takes some number of ints and packs them into a vector.

steven-johnson:
> also be an overload that takes some number of ints and packs them into a vector

We could do that, but it require doing some ugly variadic template hacking to allow for the optional ParamMap at the end -- we *could* do that (and indeed we do in a handful of other places), but I think it's a pretty bad antipattern for a public API, as it makes reading and understanding the API a lot harder than it could be. IMHO the overhead of the transitory vector is insignificant compared to everything else that `infer_input_bounds` is going to do.

steven-johnson:
(That said: we could keep the exactly-4-argument call in existence but marked as 'deprecated', and just add smarts to avoid adding zero entries.)

abadams:
How about just a version that takes a variadic number of ints, and no trailing args. I think that would cover 99.9% of existing use-cases.

steven-johnson:
I'm gonna push back a little bit more on this and say that I think that variadic args in our public API (in general) don't really pay for themselves; as we've seen here, it means that if a method later needs additional arguments, we have to either resort to template-fu to add them (yuck), or add an overload that uses a list in the first place, or just break existing usages. 

Some existing APIs are so heavily used that breaking existing usage is a complete nonstarter (e.g. `realize()`); infer_input_bounds() is very lightly used, so we have the opportunity to fix an API that only really made sense in the pre-`halide_buffer_t` world. I'm fine with adding the deprecated-backwards-compatible call (I have that prepped locally), but not so much a new variadic call.

abadams:
Fair points. 

The original intent of infer_input_bounds is that its signature would match realize exactly, because it's like a dry-run version of realize. Adding a Target never happened, which was a mistake because it implies recompilation if the target differs between realize and infer_input_bounds. Then adding a ParamMap after a bunch of default-zero args was another mistake because it invites users to actually fill in those zeros. At that point to continue to match realize we should have just provided the same overloads. 

I think it's more important to maintain consistency of usage with realize than to use this as an opportunity to deprecate things. So I think the reasonable positions are that the overloads should exactly mirror the realize overloads, or be a subset. The most user-friendly thing seems like mirroring all of them and deprecating anything in lockstep across infer_input_bounds and realize. We should definitely nuke the zero-filling goofiness now though.

abadams:
(If you have better things to do than bikeshed infer_input_bounds APIs, then we could also just remove the deprecation tag from the current state of the PR and kick the can)

steven-johnson:
> better things to do than bikeshed

I'm pretty sure that is the highest possible calling of anyone in our profession

steven-johnson:
And in the WTF category, it seems that correctness/bounds_of_monotonic_math is choosing the wrong overload and complaining that it doesn't compile:

```
/Users/srj/GitHub/Halide/test/correctness/bounds_of_monotonic_math.cpp:14:26: error: braces around scalar initializer [-Werror,-Wbraced-scalar-init]
    f.infer_input_bounds({10});
```

In an effort to just fix the underlying Real Bug, I'm going to open a new PR that doesn't mess with any API and just fixes the issue at hand; we can debate this later, because everyone knows that bikesheds should not be variadic

Issue #3778, opened 2019-03-28
possible bug in simplify_correlated_differences()


steven-johnson:
The recent PR that introduced `simplify_correlated_differences()` may have a latent bug; I am still trying to track it down, but have a relatively simple JIT-based test that will repro it. Essentially, we build two Funcs with a trivially different composition; both should produce identical output, but do not.

- If I eliminate the first call to `simplify_correlated_differences()` in `lower()`, they do.
- If I revert to the commit prior to `simplify_correlated_differences()` being introduced, they do.

The IR being produced is different, but I haven't yet figured out where the 'wrong' path is, well, wrong. Posting here to get more eyes on it.

```
#include "Halide.h"
#include <stdio.h>

using namespace Halide;

constexpr int IN_WIDTH = 1;
constexpr int IN_HEIGHT = 3;
// Interestingly, IN_HEIGHT < 3 doesn't seem to fail

constexpr int OUT_WIDTH = IN_WIDTH*2;
constexpr int OUT_HEIGHT = IN_HEIGHT*2;

Func build_func(const char *fname, ImageParam &in, bool correct) {
    Var x, y, xo, yo, xi, yi;

    Func gr;
    gr(x, y) = in(x, y);

    Func gz;
    if (correct) {
      // This formulation produces 'correct' output.
      gz(x, y) = in(x, y);
    } else {
      // This formulation produces 'wrong' output.
      gz(x, y) = gr(x, y);
    }

    Func intermed;
    intermed(x, y) = gr(x/2, y/2) | gz(x/2, y/2);

    Func output(fname);
    output(x, y) = intermed(x, y);

    // yes, we 'split' by the entire size
    // TailStrategy doesn't seem to matter
    output
        .split(x, xo, xi, OUT_WIDTH, TailStrategy::ShiftInwards)
        .split(y, yo, yi, OUT_HEIGHT, TailStrategy::ShiftInwards);

    // yes, we 'split' into pieces of only 1
    // TailStrategy doesn't seem to matter
    intermed
        .split(x, xo, xi, 1, TailStrategy::ShiftInwards)
        .split(y, yo, yi, 1, TailStrategy::ShiftInwards)
        .compute_at(output, yi);

    {
      gr.compute_at(intermed, xi).store_at(output, yo);
      gz.compute_at(intermed, xi).store_at(output, yo);
    }

    output.bound(x, 0, OUT_WIDTH).bound(y, 0, OUT_HEIGHT);

    return output;
}

[...]

PR #3784, opened 2019-04-01
Avoid passing CSE exprs with shadowed variables

abadams:


abadams:
Should fix #3778

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -94,9 +94,9 @@ class SimplifyCorrelatedDifferences : public IRMutator {
             (ma == Monotonic::Decreasing && mb == Monotonic::Increasing && std::is_same<T, Add>::value)) {
 
             for (auto it = lets.rbegin(); it != lets.rend(); it++) {
-                if (expr_uses_var(e, it->first)) {
-                    e = Let::make(it->first, it->second, e);
-                }
+                // We must avoid making let nodes here, because CSE
+                // does not handle IR with shadowed variable names.
+                e = graph_substitute(it->first, it->second, e);
```
steven-johnson: If CSE cannot handle shadowed names, perhaps we should add a visitor to it to verify that it doesn't encounter duplicate lets?

abadams: Or just fix whatever is broken when there are repeated names. I think this is the minimal PR to unbreak master though, so I'll work on that in a separate PR.

Issue #3096, opened 2018-07-02
Remove Hexagon runtime prebuilts


dsharletg:
We shouldn't be checking in prebuilts. Instead, we should have the buildbots build them, and distribute them as part of Halide releases instead.

ronlieb:
sounds like a great idea

ronlieb:
as i recall, the issue we ran into before, is that we have dependencies on having an installed HEXAGON_SDK. The qaic compiler, elfsigner, include files and libs are all needed to rebuild these libraries. Folks who want to use hexagon as a target are able to do so because of the committed prebuilts.

dsharletg:
I had another idea recently: we could just "prebuild" the output of the RPC compiler, which is just some C code (not a binary) so it can be diffed and such when modifying the RPC definition. The rest could be built as part of the normal halide build as I don't think it would have any dependencies on the Hexagon SDK.

PR #3785, opened 2019-04-01
Add ability to change priority of QURT threads

dpalermo:
Support to:
  - modify the priority of the existing pipeline context thread
  - set the default priority used for creating thread pool threads


dpalermo:
Note that this is just the support for changing the priority...which will be needed by the user visible API (which will be submitted in a separate PR later).

dpalermo:
Changes for the user visible halide_hexagon_set_thread_priority() API is now part of this PR.

steven-johnson:
Is this ready to land?

steven-johnson:
We should land this if it's ready.

src/runtime/hexagon_remote/dlib.cpp
```diff
@@ -508,6 +502,18 @@ void *mmap_dlsym(void *from, const char *name) {
     return (void *)dlib->get_symbol_addr(sym);
 }
 
+// Check all currently loaded libraries for a symbol
+void *mmap_dlsym_libs(const char *name) {
```
dsharletg: Is there a fix here, or is it just refactoring?

dpalermo: This was refactored in order to expose this functionality so it could be called from halide_remote.cpp to lookup the halide_set_default_thread_priority symbol to see if it is present in the shared runtime.  This was needed to support the non-dlopenbuf path on the 8996/Pixel 1.

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -113,6 +113,16 @@ extern int halide_hexagon_set_performance_mode(void *user_context, halide_hexago
 extern int halide_hexagon_set_performance(void *user_context, halide_hexagon_power_t *perf);
 // @}
 
+/** Set the priority for Hexagon threads.
+ * - Valid priority values range from 1 to 255
+ * - Smaller number for higher priority
+ * - The highest priority for a user thread is 1, 0 reserved for OS usage
+ * - If not set, Halide thread priority will default to 100
+ * - This should be called before running a pipeline. */
```
dsharletg: IMO this behavior is problematic. It's especially weird because the run_context thread will get a new priority, but the thread pool threads will not.

IIRC we started down this path of exploring priorities because we were debugging bad parallel performance issues. But, I don't think we ever actually concluded that thread priorities were an issue, correct?

I am wondering if we are adding some new API support/complexity for something that we don't actually have a use case for, and has some quirky behavior.

I know the design has been like this for a while, sorry for not saying this earlier.

dpalermo: If you are talking about a case where the priority gets changed between pipeline dispatches, then yes, existing worker threads will not get the new priority.  If you like, we could avoid this behavior if we added logic to only allow a single call to set the thread priority....or a way to track down the existing worker threads and update their priorities.

This API was really intended to be used once at the beginning of an app.

The use case for this API actually came a while ago from a customer who had previously written Hexagon apps each with different priorities so they could all be running simultaneously and yet still provide the desired quality of service.  They wanted to be able to do the same thing when they used Halide.

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -318,10 +318,53 @@ int halide_hexagon_remote_get_symbol_v4(handle_t module_ptr, const char* name, i
...
+        return 0;
+    }
+
+    // Find the halide_set_default_thread_priority function in the shared runtime,
+    // which we loaded with RTLD_GLOBAL.
+    void (*set_priority)(int) = NULL;
+    if (use_dlopenbuf()) {
+        set_priority = (void (*)(int)) halide_get_symbol("halide_set_default_thread_priority");
+    } else {
+        set_priority = (void (*)(int)) mmap_dlsym_libs("halide_set_default_thread_priority");
```
dsharletg: It seems like this should always use halide_get_symbol, and halide_get_symbol is the thing that should be calling mmap_dlsym_libs.

src/runtime/hexagon_remote/dlib.cpp
dsharletg: I see. I think this logic should go in mmap_dlsym, and that should handle the case when the library argument is NULL, or RTLD_SELF, or RTLD_DEFAULT. The design here was that mmap_dlopen/mmap_dlsym/mmap_dlclose mirror the standard APIs.

src/runtime/HalideRuntimeHexagonHost.h
dsharletg: OK, in that case, we can go ahead with this, modulo some minor comments.

src/runtime/hexagon_remote/dlib.cpp
```diff
@@ -495,25 +496,25 @@ void *mmap_dlopen(const void *code, size_t size) {
 void *mmap_dlsym(void *from, const char *name) {
     if (!from) return NULL;
```
dsharletg: Isn't RTLD_DEFAULT == NULL, at least on some platforms? https://code.woboq.org/userspace/glibc/dlfcn/dlfcn.h.html#_M/RTLD_DEFAULT

I think this check should go after the RTLD_SELF/RTLD_DEFAULT check.

dpalermo: Interesting...on Hexagon it is -2 (HEXAGON_Tools/8.2.07/Tools/target/hexagon/include/dlfcn.h).  Seems like a good change to make though just to be safe.

dpalermo:
Yes, I was avoiding pushing up prebuilts to make it easier to focus on just the actual changes.  Do you want me to add the prebuilts to this PR now?  ...or make that a separate PR to update prebuilts after this PR is done?

dsharletg:
The prebuilts need to be part of the PR, otherwise there is a mismatch between the headers and the files.

Fixing #3096 will be a great improvement...

dsharletg:
BTW I went ahead and pushed rebuilt prebuilts

PR #3786, opened 2019-04-01
Small c++ library that converts ONNX models into Halide pipelines

benoitsteiner:
Also developed a regression test.

A more complete solution that autoschedule these pipelines and run them is next. This will also enable us to run the entire ONNX compliance testsuite.

### DIFF
--- a/apps/onnx/Makefile
+++ b/apps/onnx/Makefile
@@ -4,24 +4,26 @@ BIN ?= bin
 
 HL_TARGET ?= host
 
-ONNX_SRC_DIR ?= $(error ONNX_SRC_DIR must provided)
-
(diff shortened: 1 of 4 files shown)


abadams:
There are some large TODOs here:
- Hooking up any kind of scheduling
- Wrapping it in a Generator so that people can actually use it in an existing build system (also makes autoscheduling it much easier). This may require some restructuring
- Adding the python layer necessary to run the onnx test suite
- Buildbot testing (it has a dependency on protoc, unfortunately)

But this seems like a minimal viable PR for onnx->Halide support so that others may contribute. Would like to get a sanity check from an external person though, e.g. @zvookin

PR #3788, opened 2019-04-02
Add option to disable PIC codegen

steven-johnson:
Currently, all of our LLVM backends generate position-independent code (llvm::Reloc::PIC_); the WebAssembly backend will need to be able to generate Static code in at least some situations, so this PR adds a way to accomplish that. (Adding here rather than in the webassembly branch to simplify downstream diffs.)

Note: I'm not wild about the 'disable_pic' nomenclature, but having one of our virtual methods return an LLVM enum value felt weird (might complicate includes), and having the default value be 'false' (to mirror use_soft_float_abi) seemed right. Open for better suggestions, of course.


abadams:
I don't think there's anything magical about having false be the default. I'd mildly prefer use_pic with a default of true. I don't feel strongly either way though.

steven-johnson:
`use_pic()` it is.

PR #3789, opened 2019-04-03
Don't call cse() in AttemptStorageFoldingOfFunction

steven-johnson:
This was a merge from the autoscheduler branch, but it's not safe: there can be duplicate lets in the Expr, and CSE is known to be unsafe on such Exprs. Backing out this change to unbreak downstream users.

### DIFF
--- a/src/StorageFolding.cpp
+++ b/src/StorageFolding.cpp
@@ -485,7 +485,8 @@ class AttemptStorageFoldingOfFunction : public IRMutator {
             Expr extent_initial = simplify(substitute(loop_var, op->min, max_initial - min_initial + 1), true, bounds);
             Expr extent_steady = simplify(max_steady - min_steady + 1, true, steady_bounds);
             Expr extent = Max::make(extent_initial, extent_steady);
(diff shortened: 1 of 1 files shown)


src/StorageFolding.cpp
```diff
@@ -455,8 +455,9 @@ class AttemptStorageFoldingOfFunction : public IRMutator {
                 continue;
             }
 
-            Expr min = simplify(common_subexpression_elimination(box[dim].min));
-            Expr max = simplify(common_subexpression_elimination(box[dim].max));
+            // TODO: should call cse() here, but there can be duplicate names in the Expr.
+            Expr min = simplify(box[dim].min);
+            Expr max = simplify(box[dim].max);
```
dsharletg: There's another one on line 488

steven-johnson: Done

steven-johnson:
@abadams -- if the proper fix for CSE isn't likely to be imminent, could we land this as a temporary band-aid to unbreak downstream users? (How critical is the CSE that was added here for the autoscheduler branch?)

PR #3791, opened 2019-04-03
Add verification to mandelbrot_aottest

steven-johnson:
Our output is coarse enough that it should match on all systems. (Assuming no meaningful floating point edge effects, of course.)

(I'm seeing some deviations from expected output for some WebAssembly variants, hence the added test)


steven-johnson:
Already fixed upstream in wasm, so going to drop this.

PR #3792, opened 2019-04-04
Avoid overreads/writes in Codegen_C

steven-johnson:
When using the 'native vector' types in GCC/Clang, the sizeof(type) may be larger than what we actually requested (eg 48->64); ensure we use only the logical size when memcpy'ing.


steven-johnson:
Waiting to land until the Mac buildbot runs; it is inexplicably offline (all the others are up)

PR #3795, opened 2019-04-05
Allow GeneratorParam<std::string>

asaadaldien:
This will allow us to pass and AOT arbitrary serialized objects (e.g onnx protobuffs).

### DIFF
--- a/src/Generator.h
+++ b/src/Generator.h
@@ -888,11 +888,35 @@ public:
 };
 
 template<typename T>
+class GeneratorParam_String : public Internal::GeneratorParamImpl<T> {
+public:
+    GeneratorParam_String(const std::string &name, const std::string &value)
+        : GeneratorParamImpl<T>(name, value) {
+    }
+    void set_from_string(const std::string &new_value_string) override {
+        this->set(new_value_string);
+    }
+
+    std::string get_default_value() const override {
+        return this->value();
+    }
+
+    std::string call_to_string(const std::string &v) const override {
+        return v;
+    }
+
+    std::string get_c_type() const override {
+        return "std::string";
+    }
+};
+
+template<typename T>
 using GeneratorParamImplBase =
     typename select_type<
         cond<std::is_same<T, Target>::value,        GeneratorParam_Target<T>>,
         cond<std::is_same<T, MachineParams>::value, GeneratorParam_MachineParams<T>>,
         cond<std::is_same<T, LoopLevel>::value,     GeneratorParam_LoopLevel>,
+        cond<std::is_same<T, std::string>::value,   GeneratorParam_String<T>>,
         cond<std::is_same<T, Type>::value,          GeneratorParam_Type<T>>,
         cond<std::is_same<T, bool>::value,          GeneratorParam_Bool<T>>,
         cond<std::is_arithmetic<T>::value,          GeneratorParam_Arithmetic<T>>,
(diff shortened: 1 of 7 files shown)


abadams:
Deferring to @steven-johnson to check if this is the right way to add std::string

test/generator/string_param_generator.cpp
```diff
@@ -0,0 +1,24 @@
+#include "Halide.h"
```
steven-johnson: The test needs to be augmented to ensure that setting the generatorparam in the Makefile (and also the CMake file) works correctly.

src/Generator.h
```diff
@@ -939,6 +945,30 @@ class GeneratorParam : public Internal::GeneratorParamImplBase<T> {
         : Internal::GeneratorParamImplBase<T>(name, value) {}
 };
 
+// Explicitly initiate std::string type outside to avoid constructor duplicate signature.
```
steven-johnson: This works but is dramatically different in style from how all of the other GeneratorParam subtypes are implemented; mimicking the existing style would arguably be better, though probably more verbose.

steven-johnson: The comment block above here (line 898 or so) is now out of date and needs updating.

test/generator/string_param_generator.cpp
steven-johnson: We also need a test to verify that GeneratorParam<string> works correctly with Stubs; the easiest way to do that is add one to stubtest_generator.cpp and then to reference it in stubtest_aottest.cpp and also in stubuser_generator.cpp

test/generator/string_param_generator.cpp
```diff
@@ -0,0 +1,24 @@
...
+class StringParam : public Halide::Generator<StringParam> {
+public:
+    GeneratorParam<std::string> param{"param", "add_one"};
+
+    Input<Buffer<float>> input{ "input", 1 };
+    Output<Buffer<float>> output{ "output", 1 };
+
+    void generate() {
+        Var x;
+        if (param.value() == "add_one") {
```
steven-johnson: I realize you want a simple test, but this is a very bad example that I fear people will mimic in new code; in this case, using a GeneratorParam<bool> or GeneratorParam<EnumType>, etc would be vastly more appropriate. Can we come up with a test example that actually *requires* the string type, and none of the other existing types will reasonably substitute?

abadams: How about an RPN description of the computation to perform e.g: "3 x + y +". That just requires Halide::Internal::split_string to break it into tokens and a stack (`vector<Expr>`) to parse into some Expr to use as the RHS of a Func.

steven-johnson: SGTM

asaadaldien: I agree with @steven-johnson its a bad example, I added it just want to make sure buildbots are happy with this change. @abadams example is perfect will update the test accordingly.

slomp:
Thaks for considering and implementing this feature 👍

steven-johnson:
LGTM; only piece missing is that we should verify that setting a string value in Make / CMake files works correctly.

test/generator/stubuser_generator.cpp
```diff
@@ -54,6 +54,7 @@ class StubUser : public Halide::Generator<StubUser> {
         gp.untyped_buffer_output_type = int32_buffer_output.type();
         gp.intermediate_level.set(LoopLevel(calculated_output, Var("y")));
         gp.vectorize = true;
+        gp.str_param = "2 * x y +";
```
abadams: Nit: this particular RPN expression would underflow the stack.

asaadaldien: Ops!

steven-johnson:
LGTM when everything is green. Thanks for the work!

asaadaldien:
Thanks for the review!

asaadaldien:
CMake removes double quote"" from argument value! e.g: 
```
  halide_define_aot_test(string_param
                         GENERATOR_ARGS rpn_expr="5 x * y +")
```

Generator is called with `rpn_expr=5\ x\ *\ y\ +`.

steven-johnson:
Ah, CMake. The gift that keeps on giving.

Issue #3793, opened 2019-04-04
CSE() is buggy


steven-johnson:
Per discussions in Gitter, it seems that our CSE pass can produce incorrect results in some cases where that are nested lets with duplicate names:

"I can repro CSE failures now. I can generate a class of pathological expressions where failures occur about 3% of the time. `(let y = (y + (y + (y + y))) in (y + y)) -> (y + y)` The y on the rhs of the let should refer to some hypothetical outer y, because the y defined by the let is not in scope yet. So the result should be 8y, not 2y."

@abadams is working on possible fixes.

PR #3796, opened 2019-04-05
Fix 3793

abadams:
Make CSE work when there are repeated lets with the same name. E.g. pathological things like let (x = x + x) in (x + x) where both (x + x) instances are the same Expr node.

Removed any attempt to not deduplicate extern calls. The simplifier cancels and reorders them anyway, so we don't really provide any guarantees like that within a single Expr.

Also added a fuzzer for CSE, and fixed an Expr-side blowup in the solver (and an incorrect upper instead of lower!)

### DIFF
--- a/src/CSE.cpp
+++ b/src/CSE.cpp
@@ -78,7 +78,7 @@ public:
         map<ExprWithCompareCache, int> uses;
         Entry(const Expr &e) : expr(e) {}
     };
-    vector<Entry> entries;
+    vector<std::unique_ptr<Entry>> entries;
 
     map<const BaseExprNode *, int> shallow_numbering, output_numbering;
     map<ExprWithCompareCache, int> leaves;
@@ -104,8 +104,7 @@ public:
             auto iter = shallow_numbering.find(e.get());
             if (iter != shallow_numbering.end()) {
                 number = iter->second;
-                auto &entry = entries[number];
-                return entry.expr;
+                return entries[number]->expr;
             }
         }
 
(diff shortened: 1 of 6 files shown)


abadams:
Fixes #3793

src/CSE.cpp
```diff
@@ -246,27 +179,55 @@ class ComputeUseCounts : public IRGraphVisitor {
 class Replacer : public IRMutator {
 public:
     map<Expr, Expr, ExprCompare> replacements;
+    Replacer() {}
```
steven-johnson: `Replacer() = default;` is probably better these days

src/IRMutator.cpp
```diff
@@ -355,25 +355,20 @@ Stmt IRMutator::visit(const Acquire *op) {
     }
 }
 
-
 Stmt IRGraphMutator2::mutate(const Stmt &s) {
```
steven-johnson: Note to self: rename IRGraphMutator2 -> IRGraphMutator

src/CSE.cpp
```diff
@@ -84,17 +73,17 @@ class GVN : public IRMutator {
...
+        map<ExprWithCompareCache, int> uses;
+        Entry(const Expr &e) : expr(e) {}
     };
     vector<Entry> entries;
 
-    typedef map<ExprWithCompareCache, int> CacheType;
-    CacheType numbering;
-
-    map<Expr, int, ExprCompare> shallow_numbering;
+    map<const BaseExprNode *, int> shallow_numbering, output_numbering;
```
steven-johnson: I will bet you a dollar that we need to use `IRHandle` here (instead of BaseExprNode*) to avoid possible refcounting bugs.

src/CSE.cpp
```diff
@@ -112,102 +101,44 @@ class GVN : public IRMutator {
...
+
+        auto &use_map = number == -1 ? leaves : entries[number].uses;
+        auto p = use_map.emplace(with_cache(new_e), (int)entries.size());
+        auto iter = p.first;
+        bool novel = p.second;
+        if (novel) {
+            // This is a never-before-seen Expr
+            number = (int)entries.size();
+            entries.emplace_back(new_e);
+            iter->second = number;
```
steven-johnson: This is not safe: iter refers to an item contained in `entries`, but the call to emplace_back could have reallocated the items in entries, making iter invalid. (Wish I could say I saw this from inspection, but ASAN is the hero here.)

abadams: iter refers to an item in the use_map, which is merely pointed to by .... I'm passing maps around by value aren't I. Will fix.

steven-johnson:
At this point I think we're good inside google (all failures appear to be unrelated.)

PR #3800, opened 2019-04-09
Minor changes from wasm branch

steven-johnson:
Grab bag of minor fixes (formatting, etc) from the wasm branch to reduce irrelevant deltas.

### DIFF
--- a/src/JITModule.cpp
+++ b/src/JITModule.cpp
@@ -1,19 +1,25 @@
-#include <string>
-#include <stdint.h>
 #include <mutex>
 #include <set>
+#include <stdint.h>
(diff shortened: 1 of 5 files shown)


src/JITModule.cpp
```diff
@@ -17,28 +17,37 @@
...
+namespace Halide {
+namespace Internal {
+
+using std::string;
+
+#ifdef _MSC_VER
 #define NOMINMAX
 #endif
 #ifdef _WIN32
 #include <windows.h>
```
abadams: Seems weird to include windows.h inside a namespace

PR #3801, opened 2019-04-10
Add basic CTest support

kpet:
This is enough to use most CTest features and makes it easy
to parallelise tests or run only select tests.

A couple of examples:
```
cd build

# List all tests
ctest -N

# Run all tests whose name matches 'reduction' using 4 jobs
ctest -j4 -R 'reduction'
```

Signed-off-by: Kévin Petit <kpet@free.fr>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -316,7 +316,7 @@ define_test_group(run_tests)
 # TODO(srj): add test_avx512 variant
 # TODO(srj): add test_python variant
 # TODO(srj): add test_apps variant
-function(add_test TARGET)
+function(add_halide_test TARGET)
   set(options EXPECT_FAILURE)
   set(oneValueArgs WORKING_DIRECTORY)
   set(multiValueArgs GROUPS)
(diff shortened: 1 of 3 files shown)


steven-johnson:
LGTM

abadams:
LGTM too. Do we need to clone this one for the buildbots? I think it could only possibly affect the windows bot.

steven-johnson:
cloned as https://github.com/halide/Halide/pull/3803 for testing

steven-johnson:
Clone is clean.

steven-johnson:
With this landed, `ctest -N` gives me many many warnings about `Could not find executable`. 

Same is true for `ctest -j4 -R 'reduction'`.

kpet:
Are you running this in the build directory? Have all the test executables been built?

steven-johnson:
yes and yes.

kpet:
Hmm, I don't really know what to say. I've just pulled master and the following just worked:

```
$ mkdir build
$ cd build
$ cmake -G Ninja ..
$ ninja build_tests
$ ctest -N
Test project /path/to/Halide/build
  Test   #1: test_internal
  Test   #2: auto_schedule_cost_function
  Test   #3: auto_schedule_data_dependent
  Test   #4: auto_schedule_extern
  Test   #5: auto_schedule_fibonacci
  Test   #6: auto_schedule_harris
  Test   #7: auto_schedule_histogram
  Test   #8: auto_schedule_iir
  Test   #9: auto_schedule_interpolate
  Test  #10: auto_schedule_large_window
...
```

Did you try from a clean build?

steven-johnson:
> Did you try from a clean build?

Yes, but I used make, not ninja. I'll try again.

steven-johnson:
Ah, apparently the issue is that CMake's default `make` build target doesn't build the tests; if I add an explicit `make build_tests` first then this works. (This is a bit surprising, as I thought that our CMake build was set up to build ~all targets by default, but apparently not.)

kpet:
Oh, yes. I hit that as well but I thought that was the intended behaviour. Glad to hear you got it working.

PR #3802, opened 2019-04-11
Add a generator to AOT onnx models

asaadaldien:
This PR adds a generator that AOT an onnx_model along with a simple prototext model as a test case.


apps/onnx/onnx_converter_generator.cc
```diff
@@ -0,0 +1,95 @@
...
+            model_outputs_[output.first] = add_output<Buffer<>>(
+                output.first,
+                getHalideType(output.second),
+                output.second.rep.dimensions());
+        }
+    }
+
+    void generate() {
+        for (auto &t : converted_model_.tensors) {
+            auto input = model_inputs_.find(t.first);
```
benoitsteiner: There are many tensors, but few are inputs: it would be more efficient to iterate over model_inputs_ and look for the corresponding tensors.

apps/onnx/onnx_converter_generator_test.cc
```diff
@@ -0,0 +1,31 @@
...
+        v = rnd();
+    });
+    B.for_each_value([&](float &v) {
+        v = rnd();
+    });
+    test_model(A, B, C);
+    for (int i = 0; i < 3; ++i) {
+        for (int j = 0; j < 4; ++j) {
+            if (C(i, j) != A(i, j) + B(i, j)) {
+                std::cout << "Unexpcted value for inputs at (" << i << "," << j << ") " << std::endl;
```
benoitsteiner: s/Unexpcted/Unexpected/

apps/onnx/Makefile
```diff
@@ -52,6 +52,34 @@ clean:
 $(BIN)/onnx_converter_test: onnx_converter_test.cc $(ONNX_CONVERTER_LIB)
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) -I$(BIN) $< -o $@ $(LDFLAGS) $(ONNX_CONVERTER_LIB) $(LIB_HALIDE) $(HALIDE_SYSTEM_LIBS)
 
-test: $(BIN)/onnx_converter_test
+
+$(BIN)/onnx_converter.generator : onnx_converter_generator.cc $(GENERATOR_DEPS) $(ONNX_CONVERTER_LIB)
+	@mkdir -p $(@D)
+	$(CXX) $(CXXFLAGS) -I$(BIN) -g -fno-rtti $(filter-out %.h,$^) -o $@ $(LDFLAGS) $(HALIDE_SYSTEM_LIBS)
+
+# Convolution with autopad=SAME_UPPER isn't supported yet (Also depracted). TODO Enable this.
```
steven-johnson: deprecated

apps/onnx/onnx_converter_generator.cc
```diff
@@ -0,0 +1,95 @@
+#include "onnx_converter.h"
+#include <Halide.h>
```
steven-johnson: Only system headers should use <>

apps/onnx/onnx_converter_generator.cc
```diff
@@ -0,0 +1,95 @@
...
+#include <map>
+#include <fstream>
+
+namespace {
+
+// TODO: This should be an onnx_converter API.
+Halide::Type getHalideType(const Tensor &tensor) {
```
steven-johnson: This should be `get_halide_type()` to follow Halide style

apps/onnx/onnx_converter_generator.cc
```diff
@@ -0,0 +1,95 @@
...
+    case onnx::TensorProto_DataType_UINT16:
+        return Halide::UInt(16);
+    case onnx::TensorProto_DataType_UINT32:
+        return Halide::UInt(32);
+    case onnx::TensorProto_DataType_INT64:
+        return Halide::Int(64);
+    case onnx::TensorProto_DataType_BOOL:
+        return Halide::Bool();
+    default:
+        throw std::domain_error("Unsupported or unknown target type");
```
steven-johnson: Halide doesn't use exceptions for error handling

apps/onnx/onnx_converter_generator_test.cc
steven-johnson: Error outputs should go to std::cerr, not std::cout

apps/onnx/onnx_converter_generator.cc
asaadaldien: onnx_converter library use exception. should we keep exceptions here as well for consistency ?

steven-johnson: I didn't review that code, or I would have made the same comment there.

I'll defer to @abadams here, but as of now, the only place we use exceptions in Halide are in the Python wrapper code (where it's essentially required for PyBind11 usage); AFAIK there is no other place in any of our codebase that uses C++ exceptions, and while I'm not intrinsically opposed to them, consistency in the codebase is a good thing. If we want to allow C++ exception usage in Halide code, we should do so deliberately and by consensus.

apps/onnx/onnx_converter_generator_test.cc
```diff
@@ -21,7 +21,7 @@ int main(int argc, char **argv) {
     for (int i = 0; i < 3; ++i) {
         for (int j = 0; j < 4; ++j) {
             if (C(i, j) != A(i, j) + B(i, j)) {
-                std::cout << "Unexpcted value for inputs at (" << i << "," << j << ") " << std::endl;
+                std::err << "Unexpected value for inputs at (" << i << "," << j << ") " << std::endl;
```
steven-johnson: std::cerr, not std::err

apps/onnx/onnx_converter_generator.cc
asaadaldien: onnx_converter also going to have a Pybind11 python extension in a follow up PR. so I thought keeping exceptions here as well.

abadams:
How about we avoid exceptions in the generator entirely, and we can worry about exceptions in the library once it becomes a problem. We may have to make them conditional on a #define if someone really wants to compile this generator with -fno-exceptions.

apps/onnx/onnx_converter_generator_test.cc
```diff
@@ -0,0 +1,31 @@
+#include "HalideBuffer.h"
+#include "HalideRuntime.h"
+#include "test_model.h"
+#include <iostream>
+#include <random>
+
+int main(int argc, char **argv) {
+    std::cout << "Running onnx_converter_generator_test..." << std::endl;
```
abadams: Nit: Please just use '\n' instead of endl. endl also implies a buffer flush, which is unnecessary, so usually endl is more verbose and slightly slower.

apps/onnx/onnx_converter_generator.cc
```diff
@@ -0,0 +1,98 @@
+#include "Halide.h"
+#include "onnx_converter.h"
+#include <fstream>
+#include <map>
+
+namespace {
+
+// TODO: This should be an onnx_converter API.
+Halide::Type get_halide_type(const Tensor &tensor) {
```
abadams: May as well move this into onnx_converter now.

asaadaldien: I have a follow up PR for updating onnx_converter (to sync with internal facebook).Let's move this in that PR.

abadams: sgtm

apps/onnx/onnx_converter_generator.cc
```diff
@@ -0,0 +1,98 @@
...
+                output.second.rep.dimensions());
+        }
+    }
+
+    void generate() {
+        for (auto const &input : model_inputs_) {
+            auto tensor = converted_model_.tensors.find(input.first);
+            if (tensor == converted_model_.tensors.end()) {
+                std::cerr << "Can't bind input " << input.first;
+								abort();
```
abadams: Weird indentation on this line.

abadams:
lgtm apart from the \n and indentation nits.

Issue #3739, opened 2019-03-18
Is -hexagon-small-data-threshold=0 still necessary for HVX backend?


steven-johnson:
Codegen_Hexagon currently uses `llvm::cl::ParseCommandLineOptions()` to set a command line flag (`-hexagon-small-data-threshold=0`) that affects HVX codegen; as discovered in https://github.com/halide/Halide/pull/3734, this isn't threadsafe, and could lead to unpredictable results if using Halide across multiple threads in the same process. (I'm not aware of any existing failures likely to be related to this, but it is definitely not an ideal practice.)

(1) Is this flag still necessary at recent LLVM trunk builds?
(2) Are there any other programmatic ways of getting this behavior, on a per-LLVMContext basis?
(3) If not, could such a way be added?

steven-johnson:
Gentle ping

ronlieb:
I imagine it still is needed, but ask Dan or Pranav

steven-johnson:
@pranavb-ca has never responded to this :-(

pranavb-ca:
I apologise for not responding. This fell through the cracks. My bad completely. This was put in a long time ago and i think it was needed at that time because of problems with PIC code. I am doing an experiment to see what breaks, if it does, if I remove this. I'll circle back shortly.

dsharletg:
I think this can be safely removed, after testing it on the various hexagon code we maintain.

pranavb-ca:
Agreed. I confirmed that this change did not affect any of our tests as well.
(Sorry for my delayed reply - I was out of office today)

dsharletg:
Fixed by the PR referenced above.

PR #3804, opened 2019-04-11
Remove -hexagon-small-data-threshold=0 hack (Issue #3739)

steven-johnson:
We believe this is no longer necessary for recent LLVM builds. https://github.com/halide/Halide/issues/3739

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -57,37 +57,8 @@ CodeGen_Hexagon::CodeGen_Hexagon(Target t) : CodeGen_Posix(t) {
     } else {
         isa_version = 60;
     }
-}
-
-std::unique_ptr<llvm::Module> CodeGen_Hexagon::compile(const Module &module) {
-    auto llvm_module = CodeGen_Posix::compile(module);
-
-    // TODO: This should be set on the module itself, or some other
-    // safer way to pass this through to the target specific lowering
-    // passes. We set the option here (after the base class'
-    // implementation of compile) because it is the last
-    // Hexagon-specific code to run prior to invoking the target
-    // specific lowering in LLVM, minimizing the chances of the wrong
-    // flag being set for the wrong module.
-    static std::once_flag set_options_once;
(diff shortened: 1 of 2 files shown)


src/CodeGen_Hexagon.cpp
```diff
@@ -57,37 +57,8 @@ CodeGen_Hexagon::CodeGen_Hexagon(Target t) : CodeGen_Posix(t) {
...
-    // TODO: This should be set on the module itself, or some other
-    // safer way to pass this through to the target specific lowering
-    // passes. We set the option here (after the base class'
-    // implementation of compile) because it is the last
-    // Hexagon-specific code to run prior to invoking the target
-    // specific lowering in LLVM, minimizing the chances of the wrong
-    // flag being set for the wrong module.
-    static std::once_flag set_options_once;
-    std::call_once(set_options_once, []() {
-        cl::ParseEnvironmentOptions("halide-hvx-be", "HALIDE_LLVM_ARGS",
```
dsharletg: This is unrelated to the small data threshold thing, and might still be necessary. I think @pranavb-ca's team was using this for something.

dsharletg:
FWIW I think the reason this is different now is because we are loading shared objects now, and our linker can handle PIC code (and not because of a change in LLVM).

src/CodeGen_Hexagon.cpp
steven-johnson: Hmm -- we can put this back in, but based on my experimentation in https://github.com/halide/Halide/pull/3734 this can be a dicey proposition, since we (theoretically) support using the same Halide instance from multiple threads; if you had a compilation setup that did some HVX and some non-HVX backends in the same process, you could potentially get different flag values depending on which backend happened to execute first. 

I'd definitely like to hear more about what their team is using it for.

pranavb-ca: I think we are fine to get rid of this. We haven't used this in our team for a long time now.

steven-johnson:
@dsharletg are you ok with this?

PR #3805, opened 2019-04-11
Allow building of the python bindings using cmake. 

LazyDodo:
This is the bare minimum needed to build the bindings on windows, and does not integrate any of the testing suite. (although any tests I ran manually seem to pass successfully).

I don't have access to a working mac/linux build environment, so someone should probably run a sanity check before merging this PR. 

By default bindings are off (you can toggle with `WITH_PYTHON_BINDINGS`) when on, you need to manually point cmake to pybind by passing `-DPYBIND11_PATH=wherever_you_have_it`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,7 @@ check_llvm_target(NVPTX WITH_NVPTX)
 # AMDGPU target is WIP
 check_llvm_target(AMDGPU WITH_AMDGPU)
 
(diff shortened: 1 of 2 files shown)


steven-johnson:
Builds OK on OSX, but needs a `test_python` build target (as is present in Makefile) to build-and-run the python tests.

LazyDodo:
Sadly this is as far as i can take this PR, both that makefile and the testing framework may as well be magic to me.

LazyDodo:
Will resubmit when i have tests working.

PR #3806, opened 2019-04-11
Disable UBSAN signed-integer-overflow warnings

steven-johnson:
some of the int64 constant-fold ops warn about possible signed-integer-overflow when run under UBSAN; disable the check for these.

### DIFF
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -16,6 +16,18 @@
 namespace Halide {
 namespace Internal {
 
+// There is no __has_feature(undefined_sanitizer), so we use asan as
+// a proxy for detecting ubsan.
+#if defined(__has_feature)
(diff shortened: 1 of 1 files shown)


abadams:
Any reason we shouldn't just move the cast to uint64 to before the + or *? Shouldn't change the bits of the result in the non-overflowing cases.

steven-johnson:
Would uint64(a * b) -> uint64(a) * uint64(b) be equivalent for a > 0 and b < 0?

abadams:
Yeah, anything involving *, +, - you can do the cast before or after. Something something ring homomorphism something.

steven-johnson:
well ok then, stand by

steven-johnson:
> (Assuming it avoids tickling ubsan)

Yeppers

PR #3807, opened 2019-04-11
Avoid initialization-order-fiasco with Halide::_

steven-johnson:
If you declare (say) a static ImageParam, that creates a Func internally, which can call Func methods that reference `Halide::_`, which might not be initialized yet. Use a local var of the same name to avoid possible UB.

Note that there may be other risky uses of `Halide::_`, `_0`, etc.: these are only the ones actually reported by UBSAN.

A better fix probably requires (eventually) eliminating these vars from our public API, and using `Var::implicit()` for `_0`, etc... adding some new API for `Halide::_`. (What would be an appropriate name for `Halide::_` ... perhaps `Var::inferred()`?)

### DIFF
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -48,7 +48,9 @@ Func::Func(const string &name) : func(unique_name(name)) {}
 Func::Func() : func(make_entity_name(this, "Halide:.*:Func", 'f')) {}
 
 Func::Func(Expr e) : func(make_entity_name(this, "Halide:.*:Func", 'f')) {
-    (*this)(_) = e;
+    // Don't use Halide::_ (avoid initialization-order-fiasco)
+    Var underscore = Var("_");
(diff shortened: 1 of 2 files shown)


abadams:
So we can't use Halide::_ inside the compiler itself, but it's still fine in user code? Just making sure this doesn't defeat the purpose of _ as syntax.

steven-johnson:
It's potentially risky in user code too; the place it was reported was in correctness_strict_float, which just declares a global-scope `ImageParam`, which is really all it takes. Not sure how much risk of genuine misbehavior there is but having any non-constant-data-globals as part of our API puts us at risk for this.

abadams:
I have a potential alternative fix. Trying it now.

abadams:
See #3808 for an alternative which avoids data globals in the API

PR #3808, opened 2019-04-11
Give the implicit vars unique types instead of state that needs to be initialized

abadams:
This makes the implicit var pieces of syntax (_*) stateless static constexpr things, instead of extern object instances. Should avoid the problems associated with exposing variables in a public API.

Alternative to #3807

### DIFF
--- a/src/Var.h
+++ b/src/Var.h
@@ -167,12 +167,40 @@ public:
     }
 };
 
+template<int N = -1>
+struct ImplicitVar {
+public:
+    Var to_var() const {
+        if (N >= 0) {
+            return Var::implicit(N);
+        } else {
+            return Var("_");
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


steven-johnson:
Looks promising. Let me give it a whirl.

src/Var.h
```diff
@@ -167,12 +167,39 @@ class Var {
...
+
+    operator Var() const {
+        return to_var();
+    }
+    operator Expr() const {
+        return to_var();
+    }
+};
+
 /** A placeholder variable for infered arguments. See \ref Var::implicit */
```
steven-johnson: inferred

steven-johnson:
Breakage seems to be due to https://github.com/halide/Halide/pull/3816

PR #3814, opened 2019-04-12
Add LLVM SCEV option

steven-johnson:
From @alinas, this can reduce compile time for some complex schedules substantially. Requires LLVM9 >= r358304.

Background: Certain Halide applications spend a very long time compiling in forgetLoop, and prefer to forget everything and rebuild SCEV (aka "Scalar Evolution") from scratch. Sample difference in compile time reduction: 21.04 to 14.78 using current ToT release build.

See also https://reviews.llvm.org/rL358304


zvookin:
I think I'd like to get a little more info in the comment about what this does re: LLVM optimization. At the very least, mentioning that "SCEV" expands to "ScalarEvolution" would be useful.

steven-johnson:
Done

PR #3817, opened 2019-04-14
Add autodiff to Python bindings

BachiLi:


python_bindings/src/PyDerivative.cpp
```diff
@@ -0,0 +1,20 @@
+#include "PyDerivative.h"
+
+namespace Halide {
+namespace PythonBindings {
+
+void define_derivative(py::module &m) {
+    auto derivative_class = py::class_<Derivative>(m, "Derivative")
+        .def("__getitem__", [](const Derivative &d, const Func &func) {return d(func);})
```
steven-johnson: The equivalent method in C++ has two optional arguments. Do we want to support those here?

python_bindings/correctness/autodiff.py
```diff
@@ -0,0 +1,20 @@
...
+	f, g, h = hl.Func('f'), hl.Func('g'), hl.Func('h')
+	f[x] = 1.0
+	g[x] = f[x] * 5.0
+	r = hl.RDom([(0, 3)])
+	h[()] = 0.0
+	h[()] += g[r.x]
+
+	d = hl.propagate_adjoints(h)
+	d_f = d[f]
+	d_f_buf = d_f.realize(3)
```
steven-johnson: We should probably test the Buffer variants of these too

BachiLi: tested in the new commit

python_bindings/src/PyDerivative.cpp
BachiLi: Addressed. The second optional argument in the C++ function is actually unnecessary and should be removed.

BachiLi: Actually I am wrong. The bounded argument was there for easier scheduling. I'll add it.

steven-johnson:
testing in https://github.com/halide/Halide/pull/3823

python_bindings/src/PyDerivative.cpp
```diff
@@ -6,6 +6,7 @@ namespace PythonBindings {
 void define_derivative(py::module &m) {
     auto derivative_class = py::class_<Derivative>(m, "Derivative")
         .def("__getitem__", [](const Derivative &d, const Func &func) {return d(func);})
+        .def("__getitem__", [](const Derivative &d, const Func &func, int update_id) {return d(func, update_id);})
```
steven-johnson: This doesn't offer the same default value options as the C++ method does. You can do this easily in PyBind by doing something like

        .def("__getitem__", [](const Derivative &d, const Func &func, int update_id, bool bounded) {
        	return d(func, update_id, bounded);
        }, py::arg("func"), py::arg("update_id") = -1, py::arg("bounded") = true)

steven-johnson: (This would allow you to delete the previous single-arg version)

python_bindings/src/PyDerivative.cpp
```diff
@@ -0,0 +1,21 @@
+#include "PyDerivative.h"
+
+namespace Halide {
+namespace PythonBindings {
+
+void define_derivative(py::module &m) {
+    auto derivative_class = py::class_<Derivative>(m, "Derivative")
+        .def("__getitem__", [](const Derivative &d, const Func &func) {return d(func);})
+        .def("__getitem__", [](const Derivative &d, const Func &func, int update_id) {return d(func, update_id);})
+        .def("__getitem__", [](const Derivative &d, const Buffer<> &buffer) {return d(buffer);});
```
steven-johnson: Even if there aren't default-values for args, you should still specify `py::arg("buffer")` here because it allows using named-argument idiom in Python.

Also, nit: spaces around braces, so `{ return d(buffer); });`

python_bindings/src/PyDerivative.cpp
BachiLi: Turns out that __getitem__ can only have a single argument, so this didn't work.
I used std::tuple instead.

steven-johnson:
Do we believe this is fully ready to review?

BachiLi:
@steven-johnson Sorry, was distracted by other stuffs. I kept the original __getitem__ operators but also added the get operators you suggested. Please take a look.

src/Derivative.h
```diff
@@ -25,20 +25,33 @@ struct Derivative {
...
         }
         auto it = adjoints.find(FuncKey{ name, update_id });
+        if (!bounded && it == adjoints.end()) {
+            // No boundary condition applied, use the original function
+            name = func.name();
+            it = adjoints.find(FuncKey{ name, update_id });
+        }
         internal_assert(it != adjoints.end()) << "Could not find Func " << name << "\n";
-        return it->second;
+        return it->second;
```
steven-johnson: nit: remove trailing whitespace

PR #3822, opened 2019-04-16
Revert "Add operator[] to RDom in the Python binding"

steven-johnson:
Reverts halide/Halide#3816

This has broken test_tutorial_lesson_13_tuples in the Python tests; we didn't notice it because the PR wasn't cloned and run with full tests.

(The nature of the fix required is not obvious to me, so I'm proposing to revert this to unbreak the tests; I encourage it to be re-added, but only after it's clear that `make test_python` is properly passing.)

### DIFF
--- a/python_bindings/correctness/rdom.py
+++ b/python_bindings/correctness/rdom.py
@@ -14,13 +14,12 @@ def test_rdom():
     r.where(r.x <= r.y)
 
     diagonal[r.x, r.y] += 2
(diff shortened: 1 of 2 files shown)


BachiLi:
Sorry for that : (
I'll think about the fix.

steven-johnson:
No worries, it's not like I've never landed changes that broke things :-)

PR #3825, opened 2019-04-17
Handle bounds of lets where the value is a single point more concisely

### LINKED ISSUE #3824: Unexpected "calls function in unbounded way"
Given the following snippet:

```
    Var x, y;

    Param<uint16_t> uint16_val;
    Param<float> float_val;

    Func rate;
    rate() = u32_sat(clamp(float_val, 1.f, 1000.f));
    rate.compute_root();

    Func intermed;
    intermed(x, y) = uint16_val;
    intermed.compute_root();

    Func result;
    result(x, y) = intermed(x, y * rate());
    result.compute_root();

    result.compile_jit();

```

It's pretty clear that `rate()` is well-bounded, but running this results in 

```
The pure definition  of Function result calls function intermed in an unbounded way in dimension 1
```

Using `u32` instead of `u32_sat` makes the error go away.

abadams:
When encountering let(name, value, body) in bounds, we use a symbolic .min/.max for the bounds of the value within the body. This misses a lot of simplifications that happen when the bounds are a single point. This PR changes things so that if the bounds of the value are a single point, we push .min/.min onto the scope instead of .min/.max. Fixes #3824 , though it wouldn't if a pure var were involved too.

Also adds an early-out in func value bounds inference for scalar funcs, which is another incidental fix for the same issue which again wouldn't help if a pure var were involved.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1218,6 +1218,8 @@ private:
                 // Substitute it in
                 var.max = val.max;
                 val.max = Expr();
+            } else if (val.is_single_point()) {
+                var.max = var.min;
             } else {
                 var.max = Variable::make(op->value.type().element_of(), max_name);
             }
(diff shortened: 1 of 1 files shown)


steven-johnson:
This does indeed fix #3824 -- though, alas, it doesn't fix the much-more-complex source from which I derived that example (I think another pure var is indeed involved). This is definitely an improvement, though, so let's land it and I'll attempt to derive another simple-ish example to repro the failure I'm seeing.

PR #3826, opened 2019-04-18
Add `-static-libgcc -static-libstdc++` to the build rules for libhalide_hexagon_host.so

steven-johnson:
This removes a runtime dep on libstdc++.so. Rebuilt all of src/runtime/hexagon_remote. Also note the drive-by change to the Makefile to update ANDROID_NDK_ROOT to r14b (which is the one installed by QC HVX SDK 3.3.3).


dsharletg:
I don't think the bin/v60 files should be changing (you're only changing the host make rules). I think they're just changing due to e.g. SDK version changes (sigh prebuilts...).

steven-johnson:
PTAL

abadams:
This makes the binary a fair bit larger. How often does it change these days? Should we consider an alternative like lfs before putting a large blob in history?

dsharletg:
This binary changes pretty rarely. I also want to remove this prebuilt entirely soon(ish...), and build it on the build bots instead (and include it in the binary releases of Halide).

steven-johnson:
Is there consensus on whether we should land this?

dsharletg:
I think we should land this. While it is a ~4x size increase, it was only ~60kb to start with.

abadams:
This is an extra 250k that we're asking people to download forever more every time they check out Halide. Maybe that's not a big deal, but if we're going to move it elsewhere soon, how about just replacing it with a curl invocation that downloads it from the build master, and manually uploading this file to the build master in some sane location? It should only take 15mins to get that sorted for a googler with access to the build master GCE instance, and we can worry about automating it using the buildbots later.

Note that if we do that, we should squash history on this PR to actually expunge the file.

dsharletg:
I'm not sure I follow the suggestion. That would mean manually building and uploading the prebuilts to the build master somehow? They surely wouldn't be part of the buildbot build outputs? I don't think it makes sense to automate anything at all with that workflow (just let people build these manually always, which is a reasonable temporary solution).

FWIW, the last time we went through this with the large git repo and eliminating large files, we had multiple, multi-megabyte uncompressed image files checked in. IIRC, it was nearly 200MB of binary data files. I agree it is the same problem, but the magnitude of the problem was very, very different.

steven-johnson:
Yeah, I'm agreeing with @dsharletg here; 250k isn't enough to worry too much about. In terms of "will we move it elsewhere soon", the problem is we don't know how soon "soon" is. If we had something ready and imminent elsewhere then sure, but we don't. I'd vote to land this and not sweat the extra.

dsharletg:
I just pulled this in and tested it, and it's broken for an unrelated reason. I think we should give up on these prebuilts now, rather than kick the can down the road.

PR #3831, opened 2019-04-22
Onnx converter updates

asaadaldien:
Upstream onnx_converter library updates.

### DIFF
--- a/apps/onnx/onnx_converter_test.cc
+++ b/apps/onnx/onnx_converter_test.cc
@@ -510,4 +510,5 @@ int main() {
     test_concat();
     test_constant_fill();
     test_model();
+    printf("Success!\n");
 }
--- a/apps/onnx/onnx_converter_test.cc
+++ b/apps/onnx/onnx_converter_test.cc
@@ -1,513 +1,513 @@
+#include "onnx_converter.h"
 #include <cmath>
 #include <random>
-#include "onnx_converter.h"
 
 #define EXPECT_EQ(a, b) \
-  if ((a) != (b)) {     \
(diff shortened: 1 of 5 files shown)


apps/onnx/onnx_converter_test.cc
```diff
@@ -1,11 +1,17 @@
...
+    if ((a) != (b)) {   \
+        exit(-1);       \
+    }
+#define EXPECT_NEAR(a, b, c)         \
+    if (std::abs((a) - (b)) > (c)) { \
+        exit(-1);                    \
+    }
 
-void test_abs() {
+static void testAbs() {
```
steven-johnson: These functions were already using proper Halide naming convention (snake_case rather than camelCase). Why were they changed?

asaadaldien: Ah probably get merged incorrectly

apps/onnx/onnx_converter_test.cc
```diff
@@ -504,6 +510,4 @@ int main() {
     test_concat();
     test_constant_fill();
     test_model();
-    printf("Success!\n");
```
abadams: Please restore these two lines (printing something on success is convention for Halide tests)

apps/onnx/onnx_converter_test.cc
```diff
@@ -505,5 +511,4 @@ int main() {
     test_constant_fill();
     test_model();
     printf("Success!\n");
```
benoitsteiner: You need to return a value since the main function is declared as int main();

PR #3832, opened 2019-04-22
JITModule::Symbol doesn't need llvm_type

steven-johnson:
We build and save the llvm FunctionType for each Symbol, but never actually use that type information anywhere. Stop doing that.

### DIFF
--- a/src/JITModule.h
+++ b/src/JITModule.h
@@ -34,10 +34,9 @@ struct JITModule {
     IntrusivePtr<JITModuleContents> jit_module;
 
     struct Symbol {
-        void *address;
-        llvm::Type *llvm_type;
-        Symbol() : address(nullptr), llvm_type(nullptr) {}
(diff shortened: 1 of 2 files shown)


src/JITModule.h
```diff
@@ -34,10 +34,9 @@ struct JITModule {
     IntrusivePtr<JITModuleContents> jit_module;
 
     struct Symbol {
-        void *address;
-        llvm::Type *llvm_type;
-        Symbol() : address(nullptr), llvm_type(nullptr) {}
-        Symbol(void *address, llvm::Type *llvm_type) : address(address), llvm_type(llvm_type) {}
+        void *address = nullptr;
+        Symbol() : address(nullptr) {}
```
abadams: ```Symbol() = default;```?

steven-johnson: done

PR #3836, opened 2019-04-23
onnx_converter library python bindings 

asaadaldien:
This adds a Pybind11 python module `model` to export onnx_converter functionality.
To use it users need to build the extension and add apps/onnx & apps/onnx/bin to their `PYTHONPATH`


apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+#include "common_types.h"
+#include "onnx_converter.h"
+#include <fstream>
+#include <random>
+#include <sys/time.h>
+#include <unordered_set>
+
+namespace py = pybind11;
+
+static HalideModel convert_onnx_model(
```
steven-johnson: here and elsewere prefer wrapping local functions in an anonymous namespace to "static" declarations

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+    std::string onnx_model_str) {
```
steven-johnson: const std::string &

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+    for (const Halide::Expr &requirement : result.model->requirements) {
+        if (Halide::Internal::is_pure(requirement)) {
+            result.rep->add_requirement(requirement);
+        }
+    }
+
+    return result;
+}
+
+static std::string auto_schedule(HalideModel pipeline) {
```
steven-johnson: pass by ref or const-ref

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+    } else if (ndarray.dtype().is(py::dtype::of<uint32_t>())) {
+        prepare_input<uint32_t>(pipeline, input_name, input_shape, ndarray);
+    } else if (ndarray.dtype().is(py::dtype::of<uint64_t>())) {
+        prepare_input<uint64_t>(pipeline, input_name, input_shape, ndarray);
+    } else if (ndarray.dtype().is(py::dtype::of<float>())) {
+        prepare_input<float>(pipeline, input_name, input_shape, ndarray);
+    } else if (ndarray.dtype().is(py::dtype::of<double>())) {
+        prepare_input<double>(pipeline, input_name, input_shape, ndarray);
+    } else if (ndarray.dtype().kind() == 'i') {
+        // TODO : Figure out why static type casting doesn't work for singed intger!
```
steven-johnson: signed

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+        if (!dim) {
+            throw std::invalid_argument(
+                "Couldn't statically infer dim " + std::to_string(rank - (i + 1)) +
+                " of output " + std::to_string(output_id));
+        }
+        stride *= *dim;
+    }
+    std::reverse(np_strides.begin(), np_strides.end());
+
+    // TODO:Better handeling of scalar outputs.
```
steven-johnson: handling

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+            }
+            mutable_data[np_index] = output_values(halide_coords);
+        });
+        return result;
+    }
+}
+
+static Halide::Type onnx_type_to_halide_type(int t) {
+    switch (t) {
+    case onnx::TensorProto::FLOAT:
```
steven-johnson: This isn't essential for this PR to land, but we have an awful lot of repeated code that maps type-to-type; is there any way this could be consolidated somehow?

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+        return Halide::Type(halide_type_uint, 8 * sizeof(uint32_t), 1);
+    case onnx::TensorProto::UINT64:
+        return Halide::Type(halide_type_uint, 8 * sizeof(uint64_t), 1);
+    default:
+        throw std::domain_error("Unsupported output type");
+    }
+}
+
+static std::vector<py::array> run(
+    HalideModel pipeline,
```
steven-johnson: pass by ref or const-ref

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+                    " of output " + std::to_string(i));
+            }
+            output_shape.push_back(*dim);
+        }
+        outputs[i] = Halide::Buffer<>(
+            onnx_type_to_halide_type(pipeline.output_types[i]), output_shape);
+    }
+    Halide::Realization real(outputs);
+    Halide::Target tgt;
+    tgt.set_feature(Halide::Target::NoAsserts, true);
```
steven-johnson: Using NoAsserts is often a Bad Idea -- if you really need to add it (and have done the performance measurement to verify that is actually worthwhile), it's essential that you document why you know it to be important here (ie how much performance gain it provides), and also how you know that it is safe to do so (ie the pipeline in question is provably safe in all cases).

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+    // Release all the inputs to free the corresponding memory until the next call
+    // to Run.
+    for (auto input : pipeline.model->inputs) {
+        input.second.reset();
+    }
+
+    return results;
+}
+
+double benchmark(HalideModel pipeline, int num_iters, const std::string &device) {
```
steven-johnson: pass by ref or const-ref

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+            }
+            output_shape.push_back(*dim);
+        }
+        outputs[i] = Halide::Buffer<>(
+            onnx_type_to_halide_type(pipeline.output_types[i]), output_shape);
+    }
+
+    Halide::Realization real(outputs);
+    Halide::Target tgt;
+    tgt.set_feature(Halide::Target::NoAsserts, true);
```
steven-johnson: same comment re: NoAsserts here too

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+    clock_gettime(CLOCK_REALTIME, &end);
+
+    double total_runtime =
+        (end.tv_sec - start.tv_sec) * 1e9 + end.tv_nsec - start.tv_nsec;
+
+    // Return the average runtime. TODO: filter the outliers if any.
+    return total_runtime / num_iters;
+}
+
+void print_loop_nest(HalideModel pipeline) {
```
steven-johnson: pass by ref or const-ref

apps/onnx/model.cpp
```diff
@@ -0,0 +1,549 @@
...
+    pipeline.rep->print_loop_nest();
+}
+
+void print_lowered_statement(HalideModel pipeline) {
```
steven-johnson: pass by ref or const-ref

apps/onnx/model.py
```diff
@@ -0,0 +1,49 @@
...
+    def BuildFromOnnxModel(self, onnx_model):
+        assert onnx_model
+
+        if type(onnx_model) is str:
+            onnx_model = onnx_model.encode()
+            self.pipeline = model_cpp.ConvertOnnxModel(onnx_model)
+        elif type(onnx_model) is bytes:
+            self.pipeline = model_cpp.ConvertOnnxModel(onnx_model)
+        else:
+            # protobuf don't support swig, so we have to convert them to string.
```
steven-johnson: doesn't

benoitsteiner: This probably should go away entirely, this is a left over from my initial implementation that was using swig instead of pybind11

apps/onnx/halide_as_onnx_backend_test.py
```diff
@@ -0,0 +1,81 @@
...
+     '|test_vgg.*'  # results are not accurate enough
+     '|test_zfnet.*'      # LRN not supported yet
+     '|test_densenet.*'   # Times out
+     '|test_shufflenet.*'  # Padding type not supported for pooling
+     '|test_squeezenet.*'   # Results are not accurate enough
+     '|test_inception_v1.*'   # LRN not supported yet
+     '|test_inception_v2.*'  # Padding type not supported for pooling
+)
+
+if not torch.cuda.is_available():
```
benoitsteiner: We probably don't want to add a hard dependency on pytorch here. This should probably be replaced with a check for Halide::Target::supports_device_api(CUDA) or similar.

apps/onnx/halide_as_onnx_backend.py
```diff
@@ -0,0 +1,132 @@
...
+
+from onnx.backend.base import Backend as BackendBase
+import onnx
+import model as halide_model
+import signal
+import base64
+import hashlib
+import datetime
+
+SAVE_MODEL = False
```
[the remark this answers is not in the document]
asaadaldien: Will be removed since we don't save models in the oss version yet

apps/onnx/model.cpp
asaadaldien: Agree this and get_tesnsor_type(Tensor& tensor) should be consolidated. Adding a TODO.

apps/onnx/model.cpp
asaadaldien: We actually changed that recently and we have NoAsserts for NDEBUG mode.

steven-johnson: Wait what? I strenuously object to this. Strenuously. This should be discussed widely.

apps/onnx/halide_as_onnx_backend.py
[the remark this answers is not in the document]
benoitsteiner: I've created this mechanism as a quick and dirty way to ingest the onnx compliance tests into an internal DB of onnx models that I use to build a training set for the cost model used in the autoscheduler. A better solution would have been to create a separate app to take care of this distinct use case. As a result it's safe to completely delete this code in the open source version of the codebase.

apps/onnx/model.cpp
asaadaldien: CC:@benoitsteiner IIRC jiting a pipeline with assertions was taking long time so we want to compile fast in NDEBUG mode ?!

steven-johnson: tying Halide::NoAsserts to NDEBUG is not the right answer. If you need to use Halide::NoAsserts, that's a decision that should be made indepdendently from NDEBUG.

abadams: I think using NoAsserts at all was a hold-over from when we were having llvm problems related to the number of assertions. We should remove this now that we've fixed that.

apps/onnx/halide_as_onnx_backend.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+prepare = HalideBackend.prepare
+run_node = HalideBackend.run_node
+run_model = HalideBackend.run_model
+supports_device = HalideBackend.supports_device
+is_compatible = HalideBackend.is_compatible
+
+# TODO Query if JIT Target is cuda and device is available. We currently running on host.
+def is_cuda_avaiable():
```
benoitsteiner: This functionality should be moved to the supports_device class method above to avoid introducing a redundant mechanism. You should then be able to delete the code that filters the cuda tests from halide_as_onnx_backend_test.py.

apps/onnx/halide_as_onnx_backend.py
```diff
@@ -0,0 +1,133 @@
...
+    @classmethod
+    def supports_device(cls, device):  # type: (Text) -> bool
+        """
+        Checks whether the backend is compiled with particular device support.
+        Always return True since we ignore the device and run on the host at the
+        moment.
+        """
+        return True
+
+    @classmethod
```
benoitsteiner: This method isn't needed anymore, and this isn't part of the onnx.Backend api. Please delete.

PR #3837, opened 2019-04-23
Add apps/onnx/test to test_apps

asaadaldien:
This will gracefully exit if protoc isn't installed on the Buildbot

### DIFF
--- a/apps/onnx/Makefile
+++ b/apps/onnx/Makefile
@@ -83,5 +83,5 @@ test: $(BIN)/onnx_converter_test $(BIN)/onnx_converter_generator_test
 # No Protoc
 else
 test:
-	@echo "No protoc in $(PATH), you need to install protocal buffers"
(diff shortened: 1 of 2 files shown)


apps/onnx/Makefile
```diff
@@ -83,3 +80,8 @@ test: $(BIN)/onnx_converter_test $(BIN)/onnx_converter_generator_test
 	LD_LIBRARY_PATH=$(BIN) $(BIN)/onnx_converter_test
 	LD_LIBRARY_PATH=$(BIN) $(BIN)/onnx_converter_generator_test
 
+# No Protoc
+else
+test:
+	@echo "No protoc in $(PATH), you need to install protocal buffers"
```
abadams: protocal -> protocol

abadams:
This should pass tests as-is, but for this to do anything meaningful we need help from Steven - apt-get installing something that provides protoc on some of the buildbots should be sufficient.

PR #3838, opened 2019-04-23
Change dependency on lld to be for WITH_V8 instead of WITH_WEBASSEMBLY.

zvookin:
(Per discussion with Steven.)


zvookin:
I think one has to configure Web Assembly explicitly in LLVM, but not have lld in the LLVM tree, to see the failure. I had Web Assembly configured due to early work I did on that stuff.

PR #3839, opened 2019-04-24
Remove part of https://github.com/halide/Halide/pull/3825

steven-johnson:
Propagating the value of zero-dim functions can result in references to inlined funcs remaining in the IR until codegen time, which makes everything fail. Backing this part out pending possible future investigation.


abadams:
#3825 fixed the issue it was fixing in two independent ways, so backing out one of them shouldn't hurt.

PR #3840, opened 2019-04-24
 Improved conversion of the "Shape" onnx op

benoitsteiner:


apps/onnx/onnx_converter.cc
```diff
@@ -554,23 +554,22 @@ Node convert_metadata_node(
...
+      }
+      result.outputs[0].rep() = Halide::cast<int64_t>(num_elements);
     } else {
-        std::vector<Halide::Var> var(1);
-        result.outputs[0].rep(var) = Halide::cast<int64_t>(0);
-        for (int i = 0; i < dims.size(); ++i) {
-            result.outputs[0].rep(i) = Halide::cast<int64_t>(dims[i]);
-        }
-        result.outputs[0].shape.push_back(static_cast<int>(input_shape.size()));
+      Halide::Var var;
```
steven-johnson: Are we assuming that op == "Shape" here? If so, add `internal_assert`

apps/onnx/onnx_converter.cc
```diff
@@ -554,23 +554,22 @@ Node convert_metadata_node(
...
+      Halide::Expr res = Halide::Expr(static_cast<int64_t>(0));
+      for (int i = 0; i < input_shape.size(); ++i) {
+	res = Halide::select(var == i, Halide::cast<int64_t>(input_shape[i]), res);
```
steven-johnson: nit: indentation

apps/onnx/onnx_converter.cc
```diff
@@ -554,23 +554,22 @@ Node convert_metadata_node(
...
+      }
+      result.outputs[0].rep(var) = res;
```
steven-johnson: I don't know the semantics of the 'shape' op, so I can't comment on the correctness of the code here.

apps/onnx/onnx_converter.cc
benoitsteiner: Fixed

apps/onnx/onnx_converter.cc
benoitsteiner: For some reason internal_assert is not available from this file. I've used an exception instead as done in other places.

PR #3841, opened 2019-04-24
Fix IRPrinter to show non-trivial alignment with modulus 0.

dsharletg:
I'm not sure this is correct, but it does seem like modulus-remainder with modulus 0 is treated as significant information in the simplifier and elsewhere.

### DIFF
--- a/src/IRPrinter.cpp
+++ b/src/IRPrinter.cpp
@@ -560,7 +560,7 @@ void IRPrinter::visit(const Select *op) {
 
 void IRPrinter::visit(const Load *op) {
     const bool has_pred = !is_one(op->predicate);
-    const bool show_alignment = op->type.is_vector() && op->alignment.modulus > 1;
+    const bool show_alignment = op->type.is_vector() && op->alignment.modulus != 1;
     if (has_pred) {
         stream << "(";
     }
@@ -678,7 +678,7 @@ void IRPrinter::visit(const Acquire *op) {
 void IRPrinter::visit(const Store *op) {
     do_indent();
     const bool has_pred = !is_one(op->predicate);
-    const bool show_alignment = op->value.type().is_vector() && (op->alignment.modulus > 1);
+    const bool show_alignment = op->value.type().is_vector() && (op->alignment.modulus != 1);
     if (has_pred) {
         stream << "predicate (" << op->predicate << ")\n";
         indent += 2;


abadams:
I'm confused. Modulus zero just means it's a constant, which is already obvious from the index.

dsharletg:
It's obvious from the index, but it isn't obvious that's what the modulus actually is :)

abadams:
I must be missing something - when the modulus is zero the remainder is equal to the index. With this PR I think every constant-index load is just going to print the index twice. There's no new information printed by printing the modulus/remainder.

dsharletg:
In #3842 for example, I get this IR when I use HL_DEBUG_CODEGEN=2 on this PR and the example program in that issue:

       blur_y[ramp(0, 1, 64) aligned(0, 0)] = (concat_vectors(blur_y[ramp(0, 1, 32) aligned(0, 0)], blur_y[ramp(32, 1, 32) aligned(0, 32)]) + (int32x64(in[ramp((((blur_y.s1.ry$x + t20)*in.stride.1) + t22), 1, 64)])*x64(2)))
        blur_y[ramp(64, 1, 64) aligned(0, 64)] = (concat_vectors(blur_y[ramp(64, 1, 32) aligned(0, 64)], blur_y[ramp(96, 1, 32) aligned(0, 96)]) + (int32x64(in[ramp((((blur_y.s1.ry$x + t20)*in.stride.1) + t23), 1, 64)])*x64(2)))
        blur_y[ramp(128, 1, 64) aligned(0, 128)] = (concat_vectors(blur_y[ramp(128, 1, 32) aligned(0, 128)], blur_y[ramp(160, 1, 32) aligned(0, 160)]) + (int32x64(in[ramp((((blur_y.s1.ry$x + t20)*in.stride.1) + t24), 1, 64)])*x64(2)))

        blur_y[ramp(0, 1, 64) aligned(0, 0)] = (blur_y[ramp(0, 1, 64)] + (int32x64(in[ramp(((((downsample_y.s0.y*4) + blur_y.s1.ry$x)*in.stride.1) + (downsample_y.s0.x.v24.base.s + t14)), 1, 64)])*x64(2)))
        blur_y[ramp(64, 1, 64) aligned(0, 64)] = (blur_y[ramp(64, 1, 64)] + (int32x64(in[ramp(((((downsample_y.s0.y*4) + blur_y.s1.ry$x)*in.stride.1) + (downsample_y.s0.x.v24.base.s + t15)), 1, 64)])*x64(2)))
        blur_y[ramp(128, 1, 64) aligned(0, 128)] = (blur_y[ramp(128, 1, 64)] + (int32x64(in[ramp(((((downsample_y.s0.y*4) + blur_y.s1.ry$x)*in.stride.1) + (downsample_y.s0.x.v24.base.s + t16)), 1, 64)])*x64(2)))
 
 Without this PR, all of these loads are printed without alignment information. With this PR, we can clearly see that some alignment information has been lost.

abadams:
You're saying we're somehow constructing a load at a constant address with alignment that *isn't* modulus = 0, alignment = index? I guess I was assuming that was an invariant. Maybe we should fix it in Load::make and make it *actually* an invariant instead of just making it easier to debug.

dsharletg:
I think this was a real bug that goes beyond just constant ramps. I think the fix for the issue I mentioned is #3844, which I think would affect simplified loads of shuffles that *aren't* constants as well, so just hacking shuffles of constant loads would not fix the bug in that case?

That said, I'm OK with not merging this PR, as I do get that it's going to make IR a lot noisier. But I do think it has value for debugging issues like this one, and I don't think we should hardcode the alignment of constant ramps to avoid hiding bugs that affect more than just that case.

abadams:
It was indeed a real bug. But for the printing thing, I think we have two options: we should either print any non-trivial mod/rem, or we should enforce the invariant that all loads that would have a modulus of zero based on the index always do, so there's no need to print it.

Currently thinking that the second invariant is a little too much to put into Load::make though, and I guess this PR is the other solution, as noisy as it is. Hrm. I don't find either answer particularly satisfying.

dsharletg:
Maybe we should make the printer only print alignment if it is not trivially the same as what the load indicates? So the logic would be:

    print_alignment = false
    if (load.index is a constant) {
      if (modulus_remainder(load.index) != load.alignment)
        print_alignment = true
    } else if (alignment > 1) {
        print_alignment = true
    }

But then we should basically be asserting anyways if that inner if fails...

Maybe we should just not make any change here. We don't need to cover all debugging scenarios all the time. This was useful to debug this bug, but that doesn't mean it needs to be committed.

steven-johnson:
> doesn't mean it needs to be committed

But it might be worth capturing in a comment in the file, so that future debuggers don't have to reinvent the trick.

PR #3843, opened 2019-04-25
Mul div intrinsics codegen refactor.

zvookin:
Allows using full div/mod codegen with non-LLVM backends. Arguably a simplification of the code as it gets rid of virtual method dispatch as al alternative to other means of lowering.


src/CodeGen_LLVM.h
```diff
@@ -332,17 +332,6 @@ class CodeGen_LLVM : public IRVisitor {
      * when multiple Allocate nodes shared the same memory. */
     virtual std::string get_allocation_name(const std::string &n) {return n;}
 
-    /** Helpers for implementing fast integer division. */
-    // @{
-    // Compute high_half(a*b) >> shr. Note that this is a shift in
```
abadams: Did these comments get preserved somewhere? Otherwise it's hard to know what "sorted" means in sorted_avg.

steven-johnson:
Ready to land?

steven-johnson:
Don't land yet, there's another related fix we want to land first.

PR #3847, opened 2019-04-26
Bitwise ops type consistency.

zvookin:
Our current type coercion for bitwise ops has a number of issues. It does not handle matching the number of lanes when a scalar and vector are paired and it is not commutative in terms of type width. It also did not handle mixture of signed and unsigned types well, possibly in ways that could result in errors constructing LLVM bitcode.

This PR fixes the above issues by making the operators require arguments that match in signed/unsigned sense, though constants will be made to match non-constant expressions so ```a & 127``` will get the type of ```a``` even though it may be unsigned and 127 would normally be an ```int``` or signed. Scalars are broadcasted to match a vector argument if needed. Logic for expanding bit width chooses the larger width.

Some fixes were required in the compiler and tests, though these were generally all compilation failures not result mismatches. The PR does change the semantics of the language slightly and will likely break some end user code. The current behavior for such code is questionable however and I believe there are cases where the compiler fails with an LLVM assertion instead of giving a clean error.

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -441,7 +441,7 @@ void match_types(Expr &a, Expr &b) {
         << "Can't do arithmetic on opaque pointer types: "
         << a << ", " << b << "\n";
 
-    // First widen to match
+    // Broadcast scalar to match vector
     if (a.type().is_scalar() && b.type().is_vector()) {
         a = Broadcast::make(std::move(a), b.type().lanes());
     } else if (a.type().is_vector() && b.type().is_scalar()) {
@@ -452,7 +452,7 @@ void match_types(Expr &a, Expr &b) {
 
     Type ta = a.type(), tb = b.type();
 
-    // If type widening has made the types match no additional casts are needed
+    // If type broadcasting has made the types match no additional casts are needed
     if (ta == tb) return;
 
     if (!ta.is_float() && tb.is_float()) {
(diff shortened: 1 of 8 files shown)


src/IROperator.cpp
```diff
@@ -480,6 +480,47 @@ void match_types(Expr &a, Expr &b) {
...
 }
 
+void match_types_bitwise(Expr &x, Expr &y, const char *op_name) {
+    user_assert(x.defined() && y.defined()) << op_name << " of undefined Expr\n";
+    user_assert(x.type().is_int() || x.type().is_uint())
+      << "The first argument to " << op_name << " must be an integer or unsigned integer";
+    user_assert(y.type().is_int() || y.type().is_uint())
+      << "The second argument to " << op_name << " must be an integer or unsigned integer";
+
+    // Give explicit type to immediate constants in expression.
```
abadams: We used to do this with match_types, and it caused problems. Consider:
```
Expr f(Expr a, Expr b) {
  return a & b;
}
```

Now you have a helper function which may error or change behavior in other ways, depending on the *representation* of b but not the value or (Halide or C++) type. E.g. if b is '2+3' then that will produce different behavior to if b is '5'

We went with overloads that take explicit ints for the other operators, so that the difference was determined by C++ type. We haven't had cause to regret that, so I think we should do the same here.

zvookin:
@dsharletg There's a different standard for writing a compiler transformation than for just writing code as an end user. I'm not sure all of the changes are strictly necessary after I got the type inference logic changed in the operator implementation, but before, some of those ```x & 3``` cases were changing the type of the result to be signed when it needed to be unsigned. That's a complete disaster if it happens inside the compiler.

dsharletg:
I agree, but we shouldn't be doing that anywhere, whether it's in the compiler or not :) So I'd personally prefer to see the issue fixed such that the compiler doesn't need this fix (so we can be confident that it is also fixed for end user code).

src/IROperator.cpp
zvookin: Will implement.

zvookin:
Dillon and I have discussed this a bit. I will endeavor to make the tests not require explicit casting if they did not before. (I think this is just one case.) Inside the compiler I feel it is good discipline to use make_const and cast a type would be inferred for two reasons: first one really ought to be completely aware of the types when writing compiler transformations, not just the values. Second, this isolates the compiler from changes to the Halide language in as much as we always preserve the rule that if both arguments to an operator have the same type and lane width, that is the type used for each argument and the result. (And I'd argue that rule really has to always be true.)

steven-johnson:
(To be clear: sounds like this needs further work before being review-ready?)

zvookin:
Um not really. There are two review comments to address, which are not totally trivial but don't change the major gist of the PR.

zvookin:
Changes for int overloads made. Had to tweak fix a couple tests where errors due to this were surfaced.

PR #3848, opened 2019-04-26
Fix a few more portability errors in hexagon_remote

steven-johnson:


src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -177,7 +177,7 @@ void *halide_hexagon_host_malloc(size_t size) {
     handle = ion_alloc(ion_fd, size, alignment, 1 << heap_id, ion_flags);
     if (handle < 0) {
         __android_log_print(ANDROID_LOG_ERROR, "halide", "ion_alloc(%d, %d, %d, %d, %d) failed",
-                            ion_fd, size, alignment, 1 << heap_id, ion_flags);
+                            ion_fd, (int) size, (int) alignment, 1 << heap_id, ion_flags);
```
dsharletg: Can we fix the format string instead here? It's not impossible that these variables could go over 32-bits.

steven-johnson: done

PR #3849, opened 2019-04-26
Set llvm::TargetOptions::UseInitArray = true

steven-johnson:
This causes llvm to put ctors into the `.init_array` section when possible (rather than the .ctors` section.

This can dramatically improve performance for some targets (esp aarch64) when the `lld` linker is used: older linkers (eg, gold) tend to do legacy conversion of .ctors to .init_array, but lld does not.

(It's not clear to me why we default this to false -- historical reasons? mistake? ignorance? -- or if there are any downsides to making this change.)


abadams:
Looks like it was introduced 7 years ago: 

https://github.com/halide/Halide/commit/7d7bb9503366fc8592f23b4a45155857d0713981

and just moved around since. It was briefly set to true for native client only, so maybe there was a reason to not make it universal then, but maybe I was just being conservative. If there was a reason, the buildbots should pick it up.

PR #3850, opened 2019-04-26
updating makefile to process newer ld trace output

KarimaMa:
ld as of version 2.32 no longer prints the members of an archive with a single `-t` flag. Now it must be passed twice. In addition, the new output format precedes lines of the form `(archive)object` with the archive on a single line regardless of whether any objects therein were used. This PR adjusts the command in the Makefile to adapt to this by passing the second `-t` flag (which will be redundant on earlier versions) and dropping lines that lack a closing parenthesis (of which there are none on earlier versions).

Thanks to @alexreinking for figuring this out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -910,7 +910,7 @@ $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
 	# object files in which archives it uses to resolve
 	# symbols. We only care about the libLLVM ones.
 	@mkdir -p $(@D)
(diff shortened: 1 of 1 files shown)


alexreinking:
For reference, this is reported in the `NEWS` file of binutils 2.32:

```
* -t now doesn't report members within archives, unless -t is given twice.
  A single -t is now more useful when generating a list of files that should be
  packaged for a linker bug report.  For example:
    gcc hello.c -save-temps -Wl,-t | xargs realpath | sort | uniq > files
    tar cJf test.tar.xz `cat files`
```


https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob_plain;f=ld/NEWS;hb=refs/tags/binutils-2_32

alexreinking:
Even looking at the logs, I can't tell why win32 failed. Any idea if this is safe to merge, anyway?

abadams:
simd_op_check failed. I don't see how that could possibly be related to this though, especially given the platform.

Issue #3851, opened 2019-04-27
ARM backends are really bad at generating mlal


dsharletg:
I have found that almost none of my code is generating mlal, and instead is generating widening + full width mla instructions. The problem seems to be that LLVM aggressively "optimizes" to zext instead of sext LLVM instructions for most of my operands to mul, and then LLVM's pattern fails to match (https://github.com/earl/llvm-mirror/blob/master/test/CodeGen/AArch64/aarch64-smull.ll).

Looking at arm_neon.h (https://github.com/Cronus-Emulator/clang/blob/master/include/arm_neon.h#L2074-L2085), I think the solution is to match  mull ourselves, and then LLVM will figure out the mlal part.

abadams:
😠 

I agree with your proposed solution

PR #3852, opened 2019-04-27
Improve mlal codegen

dsharletg:
Currently, very little code except for *exact* matches to https://github.com/earl/llvm-mirror/blob/master/test/CodeGen/AArch64/aarch64-smull.ll are generating mlal instructions. This PR makes LLVM more able to generate mlal by emulating Clang's arm_neon.h.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -228,6 +228,12 @@ CodeGen_ARM::CodeGen_ARM(Target target) : CodeGen_Posix(target) {
     negations.push_back(Pattern("vqneg.v4i32", "sqneg.v4i32", 4,  -max(wild_i32x_, -(0x7fffffff))));
 
     // Widening multiplies.
+    multiplies.push_back(Pattern("vmulls.v2i64", "smull.v2i64", 2,
+                                 wild_i64x_ * wild_i64x_,
+                                 Pattern::NarrowArgs));
+    multiplies.push_back(Pattern("vmullu.v2i64", "umull.v2i64", 2,
+                                 wild_u64x_ * wild_u64x_,
+                                 Pattern::NarrowArgs));
     multiplies.push_back(Pattern("vmulls.v4i32", "smull.v4i32", 4,
                                  wild_i32x_ * wild_i32x_,
                                  Pattern::NarrowArgs));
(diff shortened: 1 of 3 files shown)


dsharletg:
Fixes #3851

src/CodeGen_ARM.cpp
```diff
@@ -360,9 +374,57 @@ void CodeGen_ARM::visit(const Mul *op) {
...
+            }
+        }
+    }
+
     // Vector multiplies by 3, 5, 7, 9 should do shift-and-add or
     // shift-and-sub instead to reduce register pressure (the
     // shift is an immediate)
+    // TODO: Verify this is still good codegen.
     if (is_const(op->b, 3)) {
         value = codegen(op->a*2 + op->a);
```
abadams: This is dubious because it codegens op->a twice. Consider the horrific bitcode we'd produce with ((((((x*3 + y)*3 + z)*3 + w)*3 + u)*3). Might be better to just delete this. Probably a task for a separate PR unless you've got nothing better to do on a Saturday morning.

dsharletg: Let's do it in a separate PR. Since both changes will have a pretty wide ranging impact on ARM codegen, I'd like to make them separately with some soak time between them.

steven-johnson:
Have we / should we file an LLVM bug to the effect of "generating these instructions requires a silly amount of knowledge about what kind of pattern matching LLVM does internally and is extremely fragile?" (I'm not particularly hopeful it would make any immediate change but getting the issue into the Official Record would IMHO be worthwhile)

dsharletg:
I actually think this strategy (look at what arm_neon.h/xmmintrin.h/etc. do and emulate it with pattern matching) will generally work. It seems very reliable in the case of mlal when looking at all the code that should generate it.

@abadams what do you think re: doing the same for x86?

dsharletg:
It's also probably what LLVM would tell us to do if the right person saw the bug.

abadams:
I would agree with you if that file reliably produced the equivalent assembly. Recently I have heard complaints that it does not (the specific case was a widening cast being lifted out as a loop invariant). It's certainly a good reference though. I think we just have to open llvm bug reports on pieces of IR that should be generating a particular instruction but are not. I.e. keep adding things to simd_op_check.

dsharletg:
That makes sense, but at least we'd be in the same boat as the intrinsics headers, which would probably make it easier to report/get attention for the bugs.

BTW, this CL needs a review, when anyone gets a chance.

PR #3854, opened 2019-04-29
Drop support for LLVM6

steven-johnson:
Halide policy is to support LLVM trunk + two previous versions, which means it's time for LLVM6 to go. If any downstream users are still relying on LLVM6, please speak up.

(Note: this can't land until https://github.com/halide/build_bot/pull/36 lands to update the buildbots; for now, Travis should pass but buildbots will fail. Opening PR anyway for discussion.)


README.md
```diff
@@ -37,26 +37,26 @@ Building Halide
...
+path. If your OS does not have packages for llvm-7.0, you can find
 binaries for it at http://llvm.org/releases/download.html. Download an
 appropriate package and then either install it, or at least put the
 `bin` subdirectory in your path. (This works well on OS X and Ubuntu.)
 
 If you want to build it yourself, first check it out from subversion:
 
-    % svn co https://llvm.org/svn/llvm-project/llvm/branches/release_60 llvm6.0
-    % svn co https://llvm.org/svn/llvm-project/cfe/branches/release_60 llvm6.0/tools/clang
+    % svn co https://llvm.org/svn/llvm-project/llvm/branches/release_70 llvm7.0
```
abadams: Should probably update the instructions to 80

steven-johnson: dangit, thought I caught all of these :-(

abadams: I has always been the min version because we've only ever updated it when we drop a version :)

PR #3859, opened 2019-05-01
Bounds correct vs bounds loaded reduced

abadams:
A reduced version of #3774 with just the non-significant cleanups and fixes. This should help diagnose the problems with the earlier PR.

In particular I'm wondering if it's the if_then_elses that break things, or the simplification post-inlining.


steven-johnson:
Testing now.

steven-johnson:
Basically identical failure count and type(s) as the previous:

`The pure definition  of Function foo calls function bar in an unbounded way in dimension 0`
or
`Symbol not found: foo$3.s0.fx.max`

abadams:
Interesting! Combined with the fact that reverting BoundsInference.cpp fixes it, it must be due to the simplification. I'll have to think about how that could possibly matter.

abadams:
I'm stumped. Maybe try printing those expressions before and after simplification in a failing case?

steven-johnson:
I'll see what I can do,.

steven-johnson:
I'm trying to narrow it down; one interesting bit I've found so far is that part of my pipeline has something like
```
Func f;
f(x, y) = select(false, VeryComplicatedExpr(x, y), 0.f);
```

where that is indeed a constant `false`, so the result should always be zero.

If I leave it as-is, I fail with `Symbol not found: foo$1.s0.fy.max`
If I remove the `select()` expr with just `0.f`, it compiles fine.

Still trying to narrow it down further.

steven-johnson:
Here's a simple repro case:

```
    Var x, y, xi, yi;

    ImageParam input(Float(32), 2);
    Param<int> radius;

    RDom range(0, radius * 2);

    const int num_phases = 2;

    Func filtered;
    filtered(x, y) = sum(input(x, y / num_phases + range - radius));
    filtered
        .tile(x, y, xi, yi, num_phases, num_phases)
        .compute_root();

    Func result;
    result(x, y) = select(false, filtered(x, y), 0.0f);

    result.compile_jit();
```

abadams:
It turns out that simplification during bounds inference deletes references which still in exist in the stmt being lowered, so you end up with orphan stages already lowered but without defined bounds. These stages may have dependencies on the input, which add_image_checks sees, and constructs bounds expressions that refer to the non-existent bounds of the orphan stages.

It works out better if you just let these orphan stages exist. You get a conservative footprint on the input, but that would be better solved by moving add_image_checks later.

Closing this PR and reviving the other one.

PR #3861, opened 2019-05-01
Fix Codegen_C for msan

steven-johnson:
The inserted calls to msan helpers won't compile under CodegenC; insert some special code to deal with this.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -2376,7 +2376,16 @@ void CodeGen_C::visit(const Call *op) {
         rhs << print_extern_call(op);
     }
 
-    print_assignment(op->type, rhs.str());
+    // Special-case for some calls that have IR that returns int, but really return void
+    if (op->name == "halide_msan_annotate_memory_is_initialized" ||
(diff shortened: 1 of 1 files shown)


abadams:
Would it be cleaner to just make halide_msan_* actually always return 0, instead of lying in the IR about the return type?

abadams:
(For precedent see for example _halide_buffer_set_host_dirty)

steven-johnson:
I thought about that but it looked more invasive... and also kinda weird since it returns a value that will always be ignored. Happy to withdraw this and do that instead though if you think it's better.

abadams:
I think it's better to keep the IR honest, and we have no void in our type system. It would also be consistent with other runtime functions with the same issue. Looking through the runtime we have lots of functions that unconditionally return zero. In many cases they are functions where we might like to return an error code (e.g. in a user override of them), but currently don't from our implementation. These tend to get wrapped in AssertStmt. In other cases they are just utility functions which will never return an error. These tend to get wrapped in Evaluate nodes.

steven-johnson:
Closed in favor of https://github.com/halide/Halide/pull/3862

PR #3862, opened 2019-05-01
Fix msan helper return types

steven-johnson:
This allows Codegen_C to use msan cleanly.

### DIFF
--- a/src/runtime/posix_error_handler.cpp
+++ b/src/runtime/posix_error_handler.cpp
@@ -15,7 +15,7 @@ WEAK void halide_default_error(void *user_context, const char *msg) {
         dst[1] = 0;
         dst += 1;
     }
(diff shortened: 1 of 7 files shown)


src/runtime/posix_error_handler.cpp
```diff
@@ -15,7 +15,7 @@ WEAK void halide_default_error(void *user_context, const char *msg) {
         dst[1] = 0;
         dst += 1;
     }
-    halide_msan_annotate_memory_is_initialized(user_context, buf, dst - buf + 1);
+    (void) halide_msan_annotate_memory_is_initialized(user_context, buf, dst - buf + 1);
```
abadams: why the explicit discards of the return value?

steven-johnson: I think it's better to make it obvious that we know a value is returned but we are deliberately ignoring it.

PR #3867, opened 2019-05-08
Shift amount type change

zvookin:
Per discussion with Dillon, this PR changes Halide to require shift amounts to be of unsigned type. The idea we'll added negative shifts in with defined behavior to reverse the direction of the shift.

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -2402,21 +2402,17 @@ void CodeGen_LLVM::visit(const Call *op) {
         }
     } else if (op->is_intrinsic(Call::shift_left)) {
         internal_assert(op->args.size() == 2);
+        internal_assert(op->args[0].type().bits() == op->args[1].type().bits());
         Value *a = codegen(op->args[0]);
         Value *b = codegen(op->args[1]);
-        if (op->args[0].type() != op->args[1].type()) {
-            llvm::Type *llvm_type_lhs = llvm_type_of(op->args[0].type());
-            b = builder->CreateBitCast(b, llvm_type_lhs);
-        }
+	internal_assert(a->getType() == b->getType()) << "LLVM type mismatch on (" << op->args[0] << ") << (" << op->args[1] << ").\n";
         value = builder->CreateShl(a, b);
     } else if (op->is_intrinsic(Call::shift_right)) {
         internal_assert(op->args.size() == 2);
+        internal_assert(op->args[0].type().bits() == op->args[1].type().bits());
         Value *a = codegen(op->args[0]);
         Value *b = codegen(op->args[1]);
-        if (op->args[0].type() != op->args[1].type()) {
-            llvm::Type *llvm_type_lhs = llvm_type_of(op->args[0].type());
-            b = builder->CreateBitCast(b, llvm_type_lhs);
-        }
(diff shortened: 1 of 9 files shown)


src/IROperator.h
```diff
@@ -1530,20 +1530,16 @@ inline Expr operator~(Expr x) {
  * arguments must have integer type. */
 // @{
 inline Expr operator<<(Expr x, Expr y) {
-    Internal::match_types_bitwise(x, y, "shift left");
+    Expr unsigned_amount = lossless_cast(UInt(x.type().bits(), x.type().lanes()), y);
+    user_assert(unsigned_amount.defined()) << "In shift left expression:\n   (" << x << ") << (" << y << ")\nthe amount must be unsigned and sizeable to the same size as value.\n";
```
dsharletg: sizable -> castable?

src/Simplify_Call.cpp
```diff
@@ -128,37 +128,28 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
         uint64_t ub = 0;
-        if (const_int(b, &ib)) {
+        if (const_uint(b, &ub)) {
             // LLVM shl and shr instructions produce poison for negative shifts,
             // or for shifts >= typesize, so we will follow suit in our simplifier.
-            user_assert(ib >= 0) << "bitshift by a constant negative amount is not legal in Halide.";
-            user_assert(ib < t.bits()) << "bitshift by a constant amount >= the type size is not legal in Halide.";
-            if (ib < t.bits() - 1) {
-                b = make_const(t, ((int64_t) 1LL) << ib);
+          user_assert(ub < (uint64_t)t.bits()) << "bitshift by a constant amount >= the type size is not legal in Halide.";
```
dsharletg: Indentation

src/CodeGen_LLVM.cpp
```diff
@@ -2404,11 +2404,19 @@ void CodeGen_LLVM::visit(const Call *op) {
         internal_assert(op->args.size() == 2);
         Value *a = codegen(op->args[0]);
         Value *b = codegen(op->args[1]);
+        if (op->args[0].type() != op->args[1].type()) {
+            llvm::Type *llvm_type_lhs = llvm_type_of(op->args[0].type());
+            b = builder->CreateBitCast(b, llvm_type_lhs);
```
dsharletg: Is bitcast the right thing here? I would think we need `CreateZExtOrTrunc`.

zvookin: Both arguments have to be the same bit size, which should perhaps be asserted here. The LLVM IR docs say the LLVM types have to match even though the amount is *always* treated as an unsigned value, even if it is signed. So I think bitcast is the right thing, but there's probably some room for improvement.

dsharletg: Hmm, I thought the plan was to allow shifts to have differently sized LHS and RHS types in the front and, and the only requirement was that the RHS be unsigned. Then we would normalize in the LLVM backend (since LLVM requires that) which would need `CreateZExtOrTrunc`, and any other backends that require this.

I see now that isn't the case in the front end (we need to be able to lossless_cast to an unsigned int of the size of the LHS type), but should it be?

Aside from that: LLVM types don't have signed-ness, so this cast will always be a no-op if the front-end doesn't allow differently sized LHS and RHS.

dsharletg: The current design also looks good to me though.

steven-johnson:
Is this ready to land once it goes green?

PR #3870, opened 2019-05-08
Use less stack space in simplify correlated differences visitor

abadams:
It's causing segfaults in #3774


steven-johnson:
(ignore the broken 'wasm' test)

PR #3871, opened 2019-05-09
Add types to left-shift and right-shift assertions

steven-johnson:
This makes debugging violations a bit easier.

### DIFF
--- a/src/CodeGen_OpenCL_Dev.cpp
+++ b/src/CodeGen_OpenCL_Dev.cpp
@@ -207,6 +207,16 @@ void CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::visit(const Call *op) {
         do_indent();
         stream << "barrier(CLK_LOCAL_MEM_FENCE);\n";
         print_assignment(op->type, "0");
+    } else if (op->is_intrinsic(Call::shift_left) || op->is_intrinsic(Call::shift_right)) {
+        // Some OpenCL implementations forbid mixing signed-and-unsigned shift values;
+        // we enforce RHS as unsigned, so quietly cast it back to int if the LHS is int
+        if (op->args[0].type().is_int() && op->args[1].type().is_uint()) {
+            Type t = op->args[0].type().with_code(halide_type_int);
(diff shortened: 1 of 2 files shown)


zvookin:
Please do not mix the types with the expression print in a way that looks like the types might have come from the expression itself. (Long discussion, but this PR exists because Expr printing is somewhat ambiguous to read. If you wanted to make it unambiguous, the way would be to wrap the Expr in a cast before printing, but then that is confusing because it looks like the cast was part of the IR originally, not inserted by the error message.) I'd suggest putting the type info on another line or using English text like "... where the arguments have type <x type> and <y type>".

steven-johnson:
Moved the types to a separate line per your request.

Also fixed a mistake in operator>>, which should have passed the unsigned RHS rather than the original RHS. (operator<< was already correct.) This is arguably a separate issue worthy of its own PR, but in the interest of expedience, including it here.

steven-johnson:
Looks like this is inducing OpenCL failures, e.g.

CUSTOMBUILD : error : CL: clBuildProgram failed: CL_BUILD_PROGRAM_FAILURE [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_boundary_conditions.vcxproj]
  Build Log:
<kernel>(100,24): error GF2BD9111: can't convert between vector values of different size ('int16' and 'uint16') [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_boundary_conditions.vcxproj]
     int16 _6824 = _6821 >> _6823;
                   ~~~~~ ^  ~~~~~
<kernel>(161,23): error GF2BD9111: can't convert between vector values of different size ('int16' and 'uint16') [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_boundary_conditions.vcxproj]
    int16 _6887 = _6884 >> _6886;
                  ~~~~~ ^  ~~~~~

dsharletg:
The OpenCL backend probably needs to cast RHS of shifts to the type of the LHS.

PR #3872, opened 2019-05-09
Fix HVX shl/shr patterns to uniformly assume RHS is unsigned

steven-johnson:


pranavb-ca:
Thanks for this PR, @steven-johnson Curious though, did you find this because simd_op_check broke? i am looking at changing simd_op_check because https://buildbot.halide-lang.org/master/#/builders/24 runs simd_op_check for hexagon (at least it thinks it does) but HL_TARGET for it is host-hvx_128 and simd_op_check runs HVX tests only if arch is hexagon. I could easily change simd_op_check for this but it breaks the build_bot model of having HL_TARGET == HL_JIT_TARGET.

steven-johnson:
@pranavb-ca I discovered it from various tests internal to Google.

dsharletg:
There's unfortunately a few more of these in https://github.com/halide/Halide/blob/master/src/runtime/hvx_128.ll#L240 and hvx_64.ll (a mirror of hvx_128.ll unfortunately).

dsharletg:
It would be nice to figure out how to run this on the build bots, the issue that pranav pointed out is unfortunate.

I think maybe we should also set HL_TARGET on the build bots when they run the various tests, and then the tests that actually need the target arch to match to run can just skip or something?

pranavb-ca:
The only tests in test_correctness that use HL_TARGET are python_extension_gen.cpp & simd_op_check.cpp
```
$Halide/test/correcntess>grep "get_target_from_env" *
python_extension_gen.cpp:    Target t = get_target_from_environment();
simd_op_check.cpp:        target = get_target_from_environment()
simd_op_check.cpp:    printf("HL_TARGET is: %s\n", get_target_from_environment().to_string().c_str());
```

I wonder if we can relax this for hvx? OTOH, we could do somethign about changing simd_op_check specially for HVX, because HVX is the only unhosted (guest/device) feature tested in simd_op_check.

PR #3877, opened 2019-05-14
Support older versions of grep

shubhamp-ca:
Older versions of egrep treat ")" as
having special meaning and the change in
PR #3850 results in Halide build failure.
(Observed in Ubuntu 14.04, GNU grep 2.16)
Replacing egrep with grep as extended
regular expression is not required here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -916,7 +916,7 @@ $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
 	# object files in which archives it uses to resolve
 	# symbols. We only care about the libLLVM ones.
 	@mkdir -p $(@D)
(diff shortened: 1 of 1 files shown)


abadams:
Would be nice to also replace the other egrep on the same line with grep.

shubhamp-ca:
> Would be nice to also replace the other egrep on the same line with grep.

Have replaced the other egrep and force pushed the changes.

pranavb-ca:
All checks passed and change approved by @abadams  & @steven-johnson . Merging

pranavb-ca:
I think I'll wait before merging. I see only one check which is unusual. @abadams , @steven-johnson  - Any ideas?

PR #3878, opened 2019-05-15
RISC-V backend.

zvookin:
Support seems to work for a minimal test. I think it is better to get it merged than to keep dragging a branch along.


src/CodeGen_RISCV.h
```diff
@@ -0,0 +1,34 @@
...
+namespace Internal {
+
+/** A code generator that emits mips code from a given Halide stmt. */
+class CodeGen_RISCV : public CodeGen_Posix {
+public:
+    /** Create a mips code generator. Processor features can be
+     * enabled using the appropriate flags in the target struct. */
+    CodeGen_RISCV(Target);
+
+    static void test();
```
abadams: What's this method?

src/LLVM_Runtime_Linker.cpp
```diff
@@ -941,6 +973,9 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
             if (t.arch == Target::POWERPC) {
                 modules.push_back(get_initmod_powerpc_ll(c));
             }
+            if (t.arch == Target::RISCV) {
+              //                modules.push_back(get_initmod_riscv_ll(c));
```
abadams: Some explanation for why this is commented out but still exists would be nice. Is it because this is where an .ll runtime module will go once we have one?

PR #3881, opened 2019-05-16
apps/ should default to using libHalide.so rather than libHalide.a

steven-johnson:
This is arguably a better choice.

(My ulterior motivation is to make adding wasm benchmarking to several of the apps, and compiling WITH_V8 is currently much easier to manage link-time dependencies of when using libHalide.so)


abadams:
I keep thinking we should make this change, but didn't realize it was so simple. The binaries are absurdly large currently.

steven-johnson:
The mingw failure is odd; one of the apps is failing to load libHalide.dll, but the others aren't. I've just tried on the buildbot that's failing and can't replicate. Going to retry the build.

steven-johnson:
My failure-to-repeat before is now easy-to-repeat after trying again. stay tuned.

steven-johnson:
TL;DR: Windows DLL search order is a nightmare; our existing tests happen to work because they are all in the same `bin/` dir as libHalide.dll, but this isn't the case for `apps/`.

PR #3882, opened 2019-05-16
Add 'make benchmark_apps'

steven-johnson:
This adds a convenience target that builds and benchmarks a handful of apps that are likely to be of interest when evaluating performance; it also does some drive-by cleanup to remove noise from the apps output and make the output more uniform.

This will likely be followed up with more-detailed benchmark output (e.g. mpix/s, etc).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1890,8 +1890,8 @@ TEST_APPS=\
 	HelloMatlab \
 	bilateral_grid \
 	blur \
-	camera_pipe \
 	c_backend \
+	camera_pipe \
 	conv_layer \
 	fft \
 	interpolate \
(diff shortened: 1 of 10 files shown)


steven-johnson:
(My immediate ulterior motive here is to provide a reasonable way to run some benchmarks for aot-compiled wasm code; however, extending some-or-all of these to run on GPU as well as CPU would seem to also be a quite useful thing.)

abadams:
Why not just use the existing 'test' makefile targets and nice the call to make benchmark_apps?

steven-johnson:
> Why not just use the existing 'test' makefile targets

The test targets don't re-run if their output file exists, so subsequent runs are rather fast :-)

steven-johnson:
(I'm actually now pondering if a good followup to this would be to have benchmark targets use the `.rungen` approach instead, as a way to provide more uniformity of benchmark output. Opinions welcome.)

abadams:
rungen might make it easier to get a consistent benchmark target for free on new apps, so +1

steven-johnson:
So should I land this as-is and build off of it?

abadams:
What do you think about the second part of my comment - just nicing the call to benchmark_apps instead of having a make variable prefix

steven-johnson:
> nicing the call to benchmark_apps 

SGTM

steven-johnson:
withdrawing; I have a better version based on RunGen coming up

PR #3885, opened 2019-05-16
Use $(MAKE) instead of make for recursive make calls

steven-johnson:
This allows the `-j` parameter to properly get propagated to sub-makefiles, and avoids the `‘warning: jobserver unavailable: using -j1. Add `+' to parent make rule.’` warning seen in some configurations (e.g. build test_apps on mingw).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1218,7 +1218,7 @@ init_time_compilation_%:
 TIME_COMPILATION ?= /usr/bin/time -a -f "$@,%U,%S,%E" -o
 
 run_tests: $(ALL_TESTS)
-	make -f $(THIS_MAKEFILE) test_performance test_auto_schedule
+	$(MAKE) -f $(THIS_MAKEFILE) test_performance test_auto_schedule
 
 build_tests: $(CORRECTNESS_TESTS:$(ROOT_DIR)/test/correctness/%.cpp=$(BIN_DIR)/correctness_%) \
 	$(PERFORMANCE_TESTS:$(ROOT_DIR)/test/performance/%.cpp=$(BIN_DIR)/performance_%) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1218,7 +1218,7 @@ init_time_compilation_%:
 TIME_COMPILATION ?= /usr/bin/time -a -f "$@,%U,%S,%E" -o
 
 run_tests: $(ALL_TESTS)
-	make -f $(THIS_MAKEFILE) test_performance test_auto_schedule
```
abadams: This one was to explicitly drop any j flag to serialize the performance tests

Makefile
```diff
@@ -1736,7 +1736,7 @@ $(BIN_DIR)/tutorial_%: $(ROOT_DIR)/tutorial/%.cpp $(BIN_DIR)/libHalide.$(SHARED_
 	@ if [[ $@ == *_run ]]; then \
 		export TUTORIAL=$* ;\
 		export LESSON=`echo $${TUTORIAL} | cut -b1-9`; \
-		make -f $(THIS_MAKEFILE) tutorial_$${TUTORIAL/run/generate}; \
+		$(MAKE) -f $(THIS_MAKEFILE) tutorial_$${TUTORIAL/run/generate}; \
```
abadams: I think some of the tutorials need to be serial too

Makefile
```diff
@@ -1884,16 +1884,16 @@ auto_schedule_%: $(BIN_DIR)/auto_schedule_%
...
-	$(TIME_COMPILATION) compile_times_performance.csv make -f $(THIS_MAKEFILE) $(@:time_compilation_performance_%=performance_%)
+	$(TIME_COMPILATION) compile_times_performance.csv $(MAKE) -f $(THIS_MAKEFILE) $(@:time_compilation_performance_%=performance_%)
 
 time_compilation_opengl_%: $(BIN_DIR)/opengl_%
-	$(TIME_COMPILATION) compile_times_opengl.csv make -f $(THIS_MAKEFILE) $(@:time_compilation_opengl_%=opengl_%)
+	$(TIME_COMPILATION) compile_times_opengl.csv $(MAKE) -f $(THIS_MAKEFILE) $(@:time_compilation_opengl_%=opengl_%)
 
 time_compilation_generator_%: $(BIN_DIR)/%.generator
-	$(TIME_COMPILATION) compile_times_generator.csv make -f $(THIS_MAKEFILE) $(@:time_compilation_generator_%=$(FILTERS_DIR)/%.a)
+	$(TIME_COMPILATION) compile_times_generator.csv $(MAKE) -f $(THIS_MAKEFILE) $(@:time_compilation_generator_%=$(FILTERS_DIR)/%.a)
```
abadams: Actually so do these... I guess most uses of recursive make here were done in order to serialize things.

Makefile
```diff
@@ -1918,15 +1918,15 @@ TEST_APPS=\
 test_apps: distrib
 	@for APP in $(TEST_APPS); do \
 		echo Testing app $${APP}... ; \
-		make -C $(ROOT_DIR)/apps/$${APP} test \
+		$(MAKE) -C $(ROOT_DIR)/apps/$${APP} test \
```
abadams: ditto, because they report runtimes

steven-johnson: ...reporting runtimes seems more like a job for benchmarks, not tests

steven-johnson:
Meh. Guess I'll drop this whole thing.

PR #3886, opened 2019-05-16
Add `make benchmarks` target

steven-johnson:
Adds benchmarking to several of the apps, using RunGen's benchmarking capability:
- Some drive-by fixes to RunGen's code and README, which have bitrotted from underuse
- Add new output format for the benchmarks (via --parsable_output) to make output simpler to machine-parse (rather than relying on the human-readable output remaining unchanged forever)
- Ensured that all the Generators being benchmarked had proper estimates or defaults filled in, so we could use those values
- Drive-by fix to put the output of `make test_apps` into the correct build subtree when building out-of-tree

### DIFF
--- a/tools/RunGen.h
+++ b/tools/RunGen.h
@@ -884,24 +884,38 @@ public:
             auto &arg = arg_pair.second;
             switch (arg.metadata->kind) {
             case halide_argument_kind_input_scalar: {
+                if (!strcmp(arg.metadata->name, "__user_context")) {
+                  arg.scalar_value.u.handle = nullptr;
+                  info() << "Argument value for: __user_context is special-cased as: nullptr";
+                  break;
+                }
+                std::vector<std::pair<const halide_scalar_value_t*, const char*>> values;
+                // If this gets any more complex, smarten it up, but for now,
+                // simpleminded code is fine.
                 if (arg.raw_string == "default") {
-                    if (!arg.metadata->scalar_def) {
-                        fail() << "Argument value for: " << arg.metadata->name << " was specified as using the default, but no default was found in the metadata.";
-                    }
-                    info() << "Argument value for: " << arg.metadata->name << " is parsed from metadata as: "
-                           << scalar_to_string(arg.metadata->type, *arg.metadata->scalar_def);
-                    arg.scalar_value = *arg.metadata->scalar_def;
+                    values.push_back({arg.metadata->scalar_def, "default"});
                 } else if (arg.raw_string == "estimate") {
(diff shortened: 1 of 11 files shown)


steven-johnson:
Example of "parsable" output:
```
Benchmarking bilateral_grid...
bilateral_grid  BEST_TIME_MSEC_PER_ITER  13.599
bilateral_grid  SAMPLES                  7
bilateral_grid  ITERATIONS               28
bilateral_grid  TIMING_ACCURACY          0.0517781
bilateral_grid  THROUGHPUT_MPIX_PER_SEC  275.755
bilateral_grid  HALIDE_TARGET            x86-64-osx-avx-avx2-f16c-fma-sse41
```

tools/RunGen.h
```diff
@@ -884,24 +884,38 @@ class RunGen {
             auto &arg = arg_pair.second;
             switch (arg.metadata->kind) {
             case halide_argument_kind_input_scalar: {
+                if (!strcmp(arg.metadata->name, "__user_context")) {
+                  arg.scalar_value.u.handle = nullptr;
```
abadams: nit: indentation

tools/RunGen.h
```diff
@@ -1168,11 +1182,20 @@ class RunGen {
...
+                  << result.iterations << " iterations, "
+                  << "accuracy " << std::setprecision(2) << (result.accuracy * 100.0) << "%).\n"
+                  << "Best output throughput is " << (megapixels_out() / result.wall_time) << " mpix/sec.\n";
+        } else {
+            out() << md->name << "  BEST_TIME_MSEC_PER_ITER  " << result.wall_time * 1000.f << "\n"
+                  << md->name << "  SAMPLES                  " << result.samples << "\n"
+                  << md->name << "  ITERATIONS               " << result.iterations << "\n"
+                  << md->name << "  TIMING_ACCURACY          " << result.accuracy << "\n"
+                  << md->name << "  THROUGHPUT_MPIX_PER_SEC  " << (megapixels_out() / result.wall_time) << "\n"
+                  << md->name << "  HALIDE_TARGET            " << md->target << "\n";
```
abadams: Not for this PR, but I was thinking about distinguishing between performance of different generator param settings (mostly auto_schedule vs not), and it made me wonder if we should be embedding generator param values in the metadata. Or do we already?

tools/RunGen.h
steven-johnson: fixed

tools/RunGen.h
steven-johnson: We don't, but that's a good idea. Please open an issue on it :-)

Issue #3883, opened 2019-05-16
Directly realizing a func with an RDom aborts when realization does not contain RDom bounds.


alexreinking:
If you have a Func (call it `g`) that contains an RDom and you try to realize it directly, then the bounds of that RDom are checked to be within the bounds of the realization (I believe so as not to overwrite the buffer). However, creating an identity Func `h` for `g` and then realizing `h` on those same bounds works as one might expect.

This behavior is confusing especially when calling `realize()` because the Halide API is responsible for allocating a buffer in this case, anyway.

It would be better to allocate a temporary buffer when needed for an output and then copy if the supplied buffer isn't large enough already. The safety check is being performed anyway.

----

The following Halide program demonstrates the behavior:

```
#include <Halide.h>
#include <iostream>
using namespace Halide;

void print_buf2d(Buffer<float> box) {
    int x_lo = box.min(0);
    int x_hi = x_lo + box.extent(0);
    int y_lo = box.min(1);
    int y_hi = y_lo + box.extent(1);
    for (int y = y_lo; y < y_hi; y++) {
        for (int x = x_lo; x < x_hi; x++) {
            std::cout << box(x,y);
        }
        std::cout << "\n";
    }
}

int main() {
    Func f{"f"}, g{"g"}, h{"h"};
    Var x{"x"}, y{"y"};

    f(x, y) = 1.0f;

    RDom r(-10,15,-10,15);
    g(x, y) = 1 + f(x, y);
    g(r.x, r.y) = 1 + g(r.x, r.y);

    h(x,y) = g(x,y);

    Buffer<float> box = h.realize(10, 10);
    print_buf2d(box);

    Buffer<float> box2 = g.realize(10, 10);
    print_buf2d(box2);
}
```

It produces the following output:

```
3333322222
3333322222
3333322222
3333322222
3333322222
2222222222
2222222222
2222222222
2222222222
2222222222
Error:
Output buffer g is accessed at -10, which is before the min (0) in dimension 0
Aborted (core dumped)
```

abadams:
realize's main mode is to evaluate into an existing buffer. It also has a convenient variant that allocates a buffer of the requested size and realizes into it (i.e. you should read `realize(20, 20)` as just `realize(Buffer<float>{20, 20})` followed by returning the new Buffer) I think it would be quite surprising for the returned buffer to not be the size requested. A hard failure seems better to me if it's not possible to realize something at the requested size.

abadams:
[...]

PR #3889, opened 2019-05-16
Permit realizing over smaller sizes than minimum by cropping output

abadams:
This is the 'view' solution from #3883 , though it could be easily adapted to be the copy solution too.

It avoids error messages about writing out of bounds on the output when calling realize with int args. It does an output bounds query, realizes over the upgraded domain, and then returns a crop back to the requested size. This crop may not be contiguous in memory.


abadams:
This change revealed a bug where debug_to_file gets called on outputs even if they're in bounds query mode, so there's a fix for that here too.

alexreinking:
Does this work with split domains? e.g. RDom over [5,10]x[5,10] but a realization over [0,7]x[0x7]? That realization is strictly speaking larger than the minimum, but not in the "right" position.

abadams:
It will realize over the union of the box requested and the box written to, so you'd get a 10x10 output that gets cropped to the top-left 7x7.

alexreinking:
Cool. That's what I understood but wanted to confirm.

alexreinking:
What is causing the failures? Something I can help with?

abadams:
When you make a bounds query call to ask a pipeline about output shapes (which we now do in this PR) it transitively makes bounds queries to any extern stages. My guess is that there's a test with an extern stage that is doing something funny that needs to be accounted for. I'll take a look on Monday.

alexreinking:
Yeah, the extern stage in `skip_stages_external_array_functions` is incrementing a counter for how many bounds queries are made to it. The up-front bounds query adds a couple.

abadams:
simd_op_check is compiled with NoBoundsQuery, so it's not going to work with this path. I'll add code to revert to the existing behavior if we're got NoBoundsQuery on. This will also give people a way to disable this new behavior if they don't want a bounds query for some reason.

alexreinking:
We should also update the documentation in Func.h to make a note of the new behavior.

abadams:
> We should also update the documentation in Func.h to make a note of the new behavior.

done. Take a look and see if it makes sense to you.

src/Func.h
```diff
@@ -758,6 +758,16 @@ class Func {
...
+     * If the Func cannot be realized into a buffer of the given size
+     * due to scheduling constraints on scattering update definitions,
+     * it will be realized into a larger buffer of the minimum size
+     * possible, and a cropped view at the requested size will be
+     * returned. It is thus not safe to assume the returned buffers
+     * are contiguous in memory. This behavior can be disabled with
+     * the NoBoundsQuery target flag, in which case an error about
+     * writing out of bounds on the output buffer will trigger
+     * instead.
+     *
```
alexreinking: Sounds good to me!

steven-johnson:
Is this a WIP or is it ready to land (pending green)?

abadams:
It's ready to land pending green.

steven-johnson:
Failure is unrelated.

PR #3891, opened 2019-05-17
Fix UB in ModulusRemainder

abadams:
Remainders can in fact be negative if the modulus is zero.


abadams:
I don't think so - the case that was a problem was caught by the assert just below. correctness_likely at HL_DEBUG_CODEGEN=1 tripped it.

PR #3893, opened 2019-05-17
More apps/ Makefile cleanup

steven-johnson:
For all the apps that use Makefile.inc:
- remove redundant conditional-defs of BIN and/or HL_TARGET
- build into bin/$HL_TARGET; this allows testing (and benchmarking) for different targets without needing to clean the output
- special-case Generators to always go into bin/host, since they are always host-only regardless of target
- use `$(@D)` for output-directory in many places where appropriate, instead of recapitulating $(BIN) or whatnot

### DIFF
--- a/apps/autoscheduler/Makefile
+++ b/apps/autoscheduler/Makefile
@@ -4,14 +4,14 @@ include ../support/autoscheduler.inc
 # A sample generator to autoschedule. Note that if it statically links
 # to libHalide, then it must be build with $(USE_EXPORT_DYNAMIC), or the
 # autoscheduler can't find the libHalide symbols that it needs.
-$(BIN)/demo.generator: demo_generator.cpp $(GENERATOR_DEPS)
+$(GENERATOR_BIN)/demo.generator: demo_generator.cpp $(GENERATOR_DEPS)
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) -g $(filter-out %.h,$^) -o $@ $(LDFLAGS) $(HALIDE_SYSTEM_LIBS)
 
 # To use the autoscheduler, set a few environment variables and use the -p flag to the generator to load the autoscheduler as a plugin
-$(BIN)/demo.a: $(BIN)/demo.generator $(AUTOSCHED_BIN)/libauto_schedule.so
+$(BIN)/demo.a: $(GENERATOR_BIN)/demo.generator $(AUTOSCHED_BIN)/libauto_schedule.so
 	HL_PERMIT_FAILED_UNROLL=1 HL_WEIGHTS_DIR=weights \
-	$(BIN)/demo.generator -g demo -o $(@D) -f demo target=$(HL_TARGET) auto_schedule=true -p $(AUTOSCHED_BIN)/libauto_schedule.so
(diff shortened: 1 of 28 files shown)


abadams:
+1. Can we set the target GeneratorParam from the directory name instead of the environment though? I believe as is you'd have to specify the target twice if you want to build a specific file (e.g. `HL_TARGET=some_target make bin/some_target/out.png`). HL_TARGET would just set the directory of the dependencies to build for generic targets like make test. I'll make a note on the line I'm talking about...

apps/autoscheduler/Makefile
```diff
@@ -1,20 +1,17 @@
...
+$(GENERATOR_BIN)/demo.generator: demo_generator.cpp $(GENERATOR_DEPS)
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) -g $(filter-out %.h,$^) -o $@ $(LDFLAGS) $(HALIDE_SYSTEM_LIBS)
 
 # To use the autoscheduler, set a few environment variables and use the -p flag to the generator to load the autoscheduler as a plugin
-$(BIN)/demo.a: $(BIN)/demo.generator $(AUTOSCHED_BIN)/libauto_schedule.so
+$(BIN)/demo.a: $(GENERATOR_BIN)/demo.generator $(AUTOSCHED_BIN)/libauto_schedule.so
 	HL_PERMIT_FAILED_UNROLL=1 HL_WEIGHTS_DIR=weights \
-	$(BIN)/demo.generator -g demo -o $(BIN) -f demo target=$(HL_TARGET) auto_schedule=true -p $(AUTOSCHED_BIN)/libauto_schedule.so
+	$(GENERATOR_BIN)/demo.generator -g demo -o $(@D) -f demo target=$(HL_TARGET) auto_schedule=true -p $(AUTOSCHED_BIN)/libauto_schedule.so
```
abadams: Let's say BIN is just bin and doesn't depend on the target, then we could structure it like:

```
$(BIN)/%/demo.a: ...
   $(BIN)/demo.generator ... target=$? ...

test: $(BIN)/$(HL_TARGET)/out.png
```

steven-johnson: Yes, but doing that will make the change much more invasive in terms of lines touched. (Maybe not a big deal, just sayin'). Let me try it.

steven-johnson: So that pattern fails where there are ambiguous matches, e.g. in bilateral_grid, where we would have
```
    bin/%/bilateral_grid.a: bilateral_grid.generator
        bilateral_grid.generator target=$*

    # separate 'viz' build with tracing enabled
    bin/%/viz/bilateral_grid.a: bilateral_grid.generator
        bilateral_grid.generator target=$*-trace_all
```

apparently the match is greedy and things that want the 'viz' version follow the first rule, which fails because the target is `host/viz`

steven-johnson: (obviously this can be solved by doing more rule rearranging but not sure if it is worth it)

abadams: I raised the idea because I did things the way they are done in this PR in some of the autoscheduler makefiles (make target names that depend on makefile variables), and sorely regretted it. I think it's worth getting the alternative of scraping the parameters from the target names working. For viz, we could just use -trace_all instead of viz as the suffix that determines we're going to be tracing things. That might actually simplify things quite a bit. The mp4 would just depend on bin/host-trace_all/process, and we wouldn't need other special rules.

steven-johnson:
Changes were large, so I opened https://github.com/halide/Halide/pull/3898 instead

PR #3894, opened 2019-05-18
re-enable -fexceptions when WITH_EXCEPTIONS is defined

macpete:
Halide's Makefile unconditionally disables exceptions by specifying `-fno-exceptions`, resulting in a compilation error if WITH_EXCEPTIONS is set.

This patch re-enables exceptions if WITH_EXCEPTIONS is set to true.


abadams:
I think -fno-exceptions only shows up if your llvm was compiled without exceptions. Is that not the case? We were assuming people would compile both llvm *and* halide with exceptions, or neither.

macpete:
Do you think it's problematic forcing `-fexceptions` when using an llvm that was built without exceptions? I'd assume, only, if llvm called back into Halide code.

I have configurations where I'm setting `WITH_LLVM_INSIDE_SHARED_LIBHALIDE=''`, because otherwise llvm inside libHalide.so conflicts with the system-provided llvm (pulled in by my use of OpenGL and OpenCL with AMD GPUs).

Ubuntu's output of `llvm-config --cxxflags` contains `-fno-exceptions`.

Is there a way to to still use llvm inside libHalide in the presence of different llvm instances in the same process? Then I'd happily set `LLVM_ENABLE_EH=true` when building llvm for Halide and drop this PR.

abadams:
Hrm, I guess we never call back from llvm into Halide code that might throw exceptions, so it's probably fine. We'll need to test it on the buildbots though, so I'll clone your PR to do that.

dsharletg:
It passed on the build bots, should this be merged?

PR #3895, opened 2019-05-18
fix CUDA debug output

macpete:
Add threads-per-core configurations for newer shader models, and fix computation of total number of cores.


abadams:
Thanks for the fix.

PR #3898, opened 2019-05-20
Clean up apps makefiles

steven-johnson:
Alternateive to #3893

### DIFF
--- a/apps/auto_viz/Makefile
+++ b/apps/auto_viz/Makefile
@@ -1,14 +1,8 @@
 include ../support/Makefile.inc
 
-VARIANTS = \
-naive_up naive_down \
-lessnaive_up lessnaive_down \
-complex_up complex_down
+VARIANTS = naive lessnaive complex
 
-LIBRARIES = $(foreach V,$(VARIANTS),$(BIN)/$(HL_TARGET)/auto_viz_demo_$(V).a)
-OUTPUTS = $(foreach V,$(VARIANTS),$(BIN)/$(HL_TARGET)/out_$(V).png)
-
-TRACE_TARGET=$(HL_TARGET)-trace_all
+OUTPUTS = $(foreach V,$(VARIANTS),$(BIN)/$(HL_TARGET)/out_$(V)_up.png $(BIN)/$(HL_TARGET)/out_$(V)_down.png)
 
 all: $(OUTPUTS)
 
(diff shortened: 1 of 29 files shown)


apps/autoscheduler/Makefile
```diff
@@ -1,50 +1,51 @@
...
 
 # Simple jit-based test
-$(BIN)/test: test.cpp $(AUTOSCHED_BIN)/libauto_schedule.so
+$(BIN)/%/test: test.cpp $(AUTOSCHED_BIN)/libauto_schedule.so
+	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) $< -o $@ $(LDFLAGS) $(LIB_HALIDE) $(HALIDE_SYSTEM_LIBS)
 
-test_perfect_hash_map: $(BIN)/test_perfect_hash_map
-	$(BIN)/test_perfect_hash_map
+test_perfect_hash_map: $(BIN)/$(HL_TARGET)/test_perfect_hash_map
```
abadams: This test doesn't depend on halide - it's pure c++

apps/bilateral_grid/Makefile
```diff
@@ -1,57 +1,53 @@
...
+	$^ -g bilateral_grid -o $(@D) -f bilateral_grid_auto_schedule target=$*-no_runtime auto_schedule=true -e static_library,h,schedule
 
-$(BIN)/viz/bilateral_grid.a: $(BIN)/bilateral_grid.generator
+$(BIN)/%/filter: $(BIN)/%/bilateral_grid.a $(BIN)/%/bilateral_grid_auto_schedule.a filter.cpp
 	@mkdir -p $(@D)
-	$^ -g bilateral_grid -o $(BIN)/viz target=$(HL_TARGET)-trace_all
+	$(CXX) $(CXXFLAGS) -ffast-math -Wall -Werror -I$(BIN)/$* $^ -o $@ $(IMAGE_IO_FLAGS) $(LDFLAGS)
 
-$(BIN)/filter: $(BIN)/bilateral_grid.a $(BIN)/bilateral_grid_auto_schedule.a filter.cpp
+$(BIN)/%/filter_viz: $(BIN)/%-trace_all/bilateral_grid.a filter.cpp ../../bin/HalideTraceViz
```
abadams: Could we just delete this rule entirely, and make the viz rule depend on $(BIN)/$(HL_TARGET)-trace_all/bilateral_grid.mp4, and make the mp4 rules depend on filter, instead of filter_viz?

apps/camera_pipe/Makefile
```diff
@@ -1,51 +1,50 @@
...
+	$^ -g camera_pipe -o $(@D) -f camera_pipe target=$* auto_schedule=false
 
-$(BIN)/camera_pipe_auto_schedule.a: $(BIN)/camera_pipe.generator
+$(BIN)/%/camera_pipe_auto_schedule.a: $(GENERATOR_BIN)/camera_pipe.generator
 	@mkdir -p $(@D)
-	$^ -g camera_pipe -o $(BIN) -f camera_pipe_auto_schedule target=$(HL_TARGET)-no_runtime auto_schedule=true
+	$^ -g camera_pipe -o $(@D) -f camera_pipe_auto_schedule target=$*-no_runtime auto_schedule=true
 
-$(BIN)/viz/camera_pipe.a: $(BIN)/camera_pipe.generator
+$(BIN)/%/process: process.cpp $(BIN)/$(HL_TARGET)/camera_pipe.a $(BIN)/$(HL_TARGET)/camera_pipe_auto_schedule.a
```
abadams: both $(HL_TARGET)s in the deps should be %, I think

apps/bilateral_grid/Makefile
steven-johnson: Not without more work, because filter_viz also builds with -DNO_AUTO_SCHEDULE.

apps/autoscheduler/Makefile
steven-johnson: Fixed

apps/resize/Makefile
```diff
@@ -14,55 +14,62 @@ lanczos_float32_up lanczos_float32_down \
...
-$(BIN)/resize_%.a: $(BIN)/resize.generator
-	@mkdir -p $(@D)
-	$^ -g resize -o $(BIN) -f resize_$* \
-	target=$(HL_TARGET)-no_runtime \
-	interpolation_type=$$(echo $* | cut -d_ -f1) \
-	input.type=$$(echo $* | cut -d_ -f2) \
-	upsample=$$(echo $* | cut -d_ -f3 | sed 's/up/true/;s/down/false/')
 
-$(BIN)/runtime.a: $(BIN)/resize.generator
+# Can't have multiple wildcards in Make, so we'll use a macro
```
abadams: Yuck, but good to know how to do this.

apps/camera_pipe/Makefile
steven-johnson: Fixed

apps/resize/Makefile
steven-johnson: ...in Bazel this would just be a list comprehension :-)

apps/support/Makefile.inc
```diff
@@ -145,7 +158,7 @@ RUNARGS ?=
 #
 #     make foo.run RUNARGS='input=a output=baz'
 #
-$(BIN)/%.run: $(BIN)/%.rungen
+$(BIN)/$(HL_TARGET)/%.run: $(BIN)/$(HL_TARGET)/%.rungen
```
abadams: This one is unfortunate - it means I can't say 'make bin/some_target/foo.run' without also setting HL_TARGET to equal the same. Does it work to say `$(BIN)/%.run: $(BIN)/%.rungen`?

apps/support/Makefile.inc
```diff
@@ -130,13 +143,13 @@ endif
...
+$(BIN)/%/RunGenMain.o: $(HALIDE_DISTRIB_PATH)/tools/RunGenMain.cpp $(HALIDE_DISTRIB_PATH)/tools/RunGen.h
 	@mkdir -p $(@D)
-	@$(CXX) -c $< $(CXXFLAGS) $(IMAGE_IO_CXX_FLAGS) -I$(BIN) -o $@
+	@$(CXX) -c $< $(CXXFLAGS) $(IMAGE_IO_CXX_FLAGS) -I$(BIN)/$* -o $@
 
 # Really, .SECONDARY is what we want, but it won't accept wildcards
-.PRECIOUS: $(BIN)/%.rungen
-$(BIN)/%.rungen: $(BIN)/RunGenMain.o $(BIN)/%.a $(BIN)/%.registration.cpp
+.PRECIOUS: $(BIN)/$(HL_TARGET)/%.rungen
+$(BIN)/$(HL_TARGET)/%.rungen: $(BIN)/$(HL_TARGET)/RunGenMain.o $(BIN)/$(HL_TARGET)/%.a $(BIN)/$(HL_TARGET)/%.registration.cpp
```
abadams: I don't think RunGenMain.o actually depends on the halide target, so how about: 
```
$(BIN)/%.rungen: $(BIN)/RunGenMain.o $(BIN)/%.a $(BIN)/%.registration.cpp
```

apps/support/Makefile.inc
steven-johnson: Even if it did, the rules (above it) for building .rungen and .registration wouldn't work.

apps/auto_viz/Makefile
```diff
@@ -5,47 +5,47 @@ naive_up naive_down \
...
 	upsample=$$(echo $* | cut -d_ -f2 | sed 's/up/true/;s/down/false/')
 
-$(BIN)/runtime.a: $(BIN)/auto_viz_demo.generator
+$(BIN)/%/runtime.a: $(GENERATOR_BIN)/auto_viz_demo.generator
 	@mkdir -p $(@D)
-	$^ -r runtime -o $(BIN) target=$(TRACE_TARGET)
+	$^ -r runtime -o $(@D) target=$(TRACE_TARGET)
 
-$(BIN)/auto_viz_demo: auto_viz_demo.cpp $(LIBRARIES) $(BIN)/runtime.a
+$(BIN)/%/auto_viz_demo: auto_viz_demo.cpp $(LIBRARIES) $(BIN)/$(HL_TARGET)/runtime.a
```
abadams: HL_TARGET -> % in RHS

apps/auto_viz/Makefile
```diff
@@ -5,47 +5,47 @@ naive_up naive_down \
...
 	@mkdir -p $(@D)
-	$(CXX) $(CXXFLAGS) -I $(BIN) $^ -o $@ $(IMAGE_IO_FLAGS) $(LDFLAGS)
+	$(CXX) $(CXXFLAGS) -I$(BIN)/$* $^ -o $@ $(IMAGE_IO_FLAGS) $(LDFLAGS)
 
-$(BIN)/out_%_up.png: $(BIN)/auto_viz_demo
+$(BIN)/$(HL_TARGET)/out_%_up.png: $(BIN)/$(HL_TARGET)/auto_viz_demo
```
abadams: I think it would be better to use a big macro to stamp out the filter-specific rules so that we can avoid all explicit use of HL_TARGET here as you did in the later Makefiles.

apps/support/Makefile.inc
steven-johnson: Actually, it will, for the Wasm case that I'm working on :-)

apps/auto_viz/Makefile
steven-johnson: meh, ok

apps/support/Makefile.inc
abadams: Nuts, I was hoping to avoid multiple wildcards to make the rungen rule below work. I guess another alternative is to just have the name of rungen depend on the filter name: 
```
$(BIN)/%_RunGenMain.o: rungen sources
   ... use string manipulation to extract the target from $*

$(BIN)/%.rungen: $(BIN)/%_RunGenMain.o $(BIN)/%.a $(BIN)/%.registration.cpp
```

abadams: Alternatively:

```
$(BIN)/%_RunGenMain.o: $(BIN)/RunGenMain.o
   ... to avoid repeatedly recompiling RunGenMain, either just cp $(BIN)/RunGenMain.o or do something fancier for wasm

$(BIN)/%.rungen: $(BIN)/%_RunGenMain.o $(BIN)/%.a $(BIN)/%.registration.cpp
```

apps/auto_viz/Makefile
steven-johnson: done

apps/support/Makefile.inc
steven-johnson: let's hold off on that, since the actual implementation of the wasm stuff might change this

abadams: If we're going to hold off, why not use the simpler first suggestion with a single RunGenMain.o? If I'm understanding things correctly, in the current state the following will produce a confusing error about a make target not existing:

```
make bin/x86-64-linux/local_laplacian.run
```

this is what we have in master, but with this PR it will produce a similar error:
```
HL_TARGET=x86-64-linux make bin/local_laplacian.run
```

You'd need to say:
```
HL_TARGET=x86-64-linux make bin/x86-64-linux/local_laplacian.run
```

make targets not existing is very annoying to debug

steven-johnson: OK, just pushed a revision that I think will work; it basically relies on % being greedy. PTAL

apps/resize/Makefile
abadams: I think this is my single favorite feature of bazel/blaze.

steven-johnson:
mingw failure is unrelated: https://github.com/halide/Halide/issues/3909
linux-64 failure is unrelated: https://github.com/halide/Halide/issues/3910

PR #3904, opened 2019-05-22
Add Buffer::copy_interleaved()

steven-johnson:
Convenience function for when you have a non-interleaved source and need an interleaved copy, which happens often enough to be worthwhile. Also, a driveby fix to Halide::Buffer::copy(), which apparently wasn't ever being used for non-void Buffers.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1180,6 +1180,16 @@ public:
         return dst;
     }
 
+    /** Like copy(), but the resulting Buffer is always interleaved.
+     * Requires that the source buffer have exactly 3 dimensions. */
+    Buffer<not_const_T, D> copy_interleaved(void *(*allocate_fn)(size_t) = nullptr,
+                                            void (*deallocate_fn)(void *) = nullptr) const {
+        assert(dimensions() == 3);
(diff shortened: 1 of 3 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -1180,6 +1180,16 @@ class Buffer {
         return dst;
     }
 
+    /** Like copy(), but the resulting Buffer is always interleaved.
+     * Requires that the source buffer have exactly 3 dimensions. */
+    Buffer<not_const_T, D> copy_interleaved(void *(*allocate_fn)(size_t) = nullptr,
+                                            void (*deallocate_fn)(void *) = nullptr) const {
+        assert(dimensions() == 3);
+        Buffer<not_const_T, D> dst = Buffer<not_const_T, D>::make_interleaved(width(), height(), channels(), allocate_fn, deallocate_fn);
```
dsharletg: I don't think this handles the case where `this` buffer has non-zero mins.

steven-johnson:
I've had several folks (over the last year+) ask how to make an interleaved copy, so I thought adding this would clarify the situation. I'm getting pushback from various folk so I'll drop this.

PR #3907, opened 2019-05-23
Add 'const' to C++ backend var decls

steven-johnson:
The code we generate is SSA so all declarations should (conceptually) be const; this is an experiment to see if this matters to codegen quality at all. (I would hope not, but it wouldn't be the first time a compiler has surprised me.)


steven-johnson:
This seems to pass everything (failures are unrelated). It doesn't make any difference in codegen quality (not surprising); is it worth landing just for trivial code 'correctness' improvement?

steven-johnson:
I guess that's a 'yes' :-)

PR #3912, opened 2019-05-23
Augment apps benchmarking rules to work with wasm

steven-johnson:
Added rules to apps/Makefile.inc to provide `benchmark_wasm` targets; the toplevel `make benchmark_apps` now knows how to sniff `HL_TARGET` and run these appropriately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,6 +158,14 @@ else
 	EMCC_SIMD_OPT=0
 endif
 
+ifneq (,$(findstring node,$(WASM_SHELL)))
+	# If 'node' is anywhere in the string, assume Node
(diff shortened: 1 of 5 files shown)


abadams:
I'm happy to stamp this, but I'm not familiar enough with the wasm ecosystem to know if the sort of stuff going on in Makefile.inc (e.g. forking a user config file) is sane.

steven-johnson:
> Makefile.inc (e.g. forking a user config file) is sane

Oh, it's definitely not sane, it's just the least-bad way to do this at the moment (there's no way to override just part of the config file via the command line). I'm going to request such a way.

PR #3916, opened 2019-05-27
Rungen binary size

abadams:
Get rungen binary size down to 1MB by optimizing for size, and moving some templated code in Runtime::Buffer into non-templated helper methods.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1754,12 +1754,28 @@ public:
     static Buffer<T, D> make_with_shape_of(Buffer<T2, D2> src,
                                            void *(*allocate_fn)(size_t) = nullptr,
                                            void (*deallocate_fn)(void *) = nullptr) {
+
+        const halide_type_t dst_type = T_is_void
+            ? src.type()
+            : halide_type_of<typename std::remove_cv<not_void_T>::type>();
+        return Buffer<>::make_with_shape_of_helper(dst_type, src.
+                                                   dimensions(), src.buf.dim,
+                                                   allocate_fn, deallocate_fn);
+    }
+
+private:
+
+    static Buffer<> make_with_shape_of_helper(halide_type_t dst_type,
+                                              int dimensions,
+                                              halide_dimension_t *shape,
+                                              void *(*allocate_fn)(size_t),
+                                              void (*deallocate_fn)(void *)) {
         // Reorder the dimensions of src to have strides in increasing order
         std::vector<int> swaps;
-        for (int i = src.dimensions()-1; i > 0; i--) {
+        for (int i = dimensions - 1; i > 0; i--) {
             for (int j = i; j > 0; j--) {
-                if (src.dim(j-1).stride() > src.dim(j).stride()) {
-                    src.transpose(j-1, j);
+                if (shape[j-1].stride > shape[j].stride) {
+                    std::swap(shape[j-1], shape[j]);
                     swaps.push_back(j);
                 }
             }
@@ -1767,8 +1783,7 @@ public:
 
         // Rewrite the strides to be dense (this messes up src, which
         // is why we took it by value).
-        halide_dimension_t *shape = src.buf.dim;
-        for (int i = 0; i < src.dimensions(); i++) {
+        for (int i = 0; i < dimensions; i++) {
             if (i == 0) {
                 shape[i].stride = 1;
             } else {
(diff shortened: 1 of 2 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -1754,21 +1754,36 @@ class Buffer {
...
+            ? src.type()
+            : halide_type_of<typename std::remove_cv<not_void_T>::type>();
+        return Buffer<>::make_with_shape_of_helper(dst_type, src.
+                                                   dimensions(), src.buf.dim,
+                                                   allocate_fn, deallocate_fn);
+    }
+
+private:
+
+    static Buffer<> make_with_shape_of_helper(halide_type_t dst_type,
```
steven-johnson: This is great in that it reduces the expansion explosion that happens here due to the both-src-and-dst-template-types. But we could reduce it even further if we made (say) `struct BufferHelper` that is outside `struct Buffer<>` in which we placed static-helper methods like this, so it wouldn't vary by any Buffer<> type instantiation. So in theory that would guarantee that helper methods like this got only one instantiation. (Might be overthinking it.)

src/runtime/HalideBuffer.h
```diff
@@ -1938,89 +1942,72 @@ class Buffer {
     // (all of different types), advance the pointers using the
     // strides.
     template<typename Ptr, typename ...Ptrs>
+    HALIDE_ALWAYS_INLINE
```
steven-johnson: Similarly, several of these helpers could be moved into a hypothetical `struct BufferHelper` to eliminate extra instantiations. (Again, not sure if it pays for itself... just sayin'.)

src/runtime/HalideBuffer.h
```diff
@@ -1938,89 +1942,72 @@ class Buffer {
...
-    // The template function that constructs the loop nest for for_each_value
-    template<int d, bool innermost_strides_are_one, typename Fn, typename... Ptrs>
-    static void for_each_value_helper(Fn &&f, const for_each_value_task_dim<sizeof...(Ptrs)> *t, Ptrs... ptrs) {
+    template<typename Fn, typename... Ptrs>
+    HALIDE_NEVER_INLINE
+    static void for_each_value_helper(Fn &&f, int d, bool innermost_strides_are_one,
+                                      const for_each_value_task_dim<sizeof...(Ptrs)> *t, Ptrs... ptrs) {
+        // When we hit a low dimensionality, switch from runtime
+        // recursion to template recursion.
+        asm volatile ("# Hi");
```
steven-johnson: Um, this absolutely needs a comment. Did you mean to leave this in?

abadams: Ha, no I did not.

src/runtime/HalideBuffer.h
abadams: If we only ever call this static method on `Buffer<>`, which we do, then only one version of it is ever instantiated.

src/runtime/HalideBuffer.h
```diff
@@ -2029,35 +2016,46 @@ class Buffer {
...
+    template<typename Fn, typename ...Args, int N = sizeof...(Args) + 1>
+    void for_each_value_impl(Fn &&f, Args&&... other_buffers) const {
+        Buffer<>::for_each_value_task_dim<N> *t =
+        (Buffer<>::for_each_value_task_dim<N> *)HALIDE_ALLOCA((dimensions()+1) * sizeof(for_each_value_task_dim<N>));
+        // Move the preparatory code into a non-templated helper to
+        // save code size.
+        const halide_buffer_t *buffers[] = {&buf, (&other_buffers.buf)...};
+        bool innermost_strides_are_one = Buffer<>::for_each_value_prep(t, buffers);
+
+        Buffer<>::for_each_value_helper(f, dimensions() - 1,
```
abadams: Note that because we enter `Buffer<>` at this point, all the support code (advance_ptrs, etc), is only called from within Buffer<>, so that it doesn't get a separate instantiation per T, D, parameter.

src/runtime/HalideBuffer.h
steven-johnson: Ah, I did not notice that detail. (Worth commenting on?)

src/runtime/HalideBuffer.h
steven-johnson: Pretty sure this is what's breaking the windows buildbots, btw

steven-johnson:
Failure is unrelated, merging

PR #3917, opened 2019-05-28
Add required sed to process msys2/mingw windows path slashes

sinkingsugar:
Not sure why your builds are passing but I think those changes are really needed to fix msys2/mingw windows paths.
They are copy pasted from few lines above where llvm-config is called.
Blaming this line seems that it's a relatively recent merge from wasm changes.

Cheers!


steven-johnson:
testing via https://github.com/halide/Halide/pull/3918

PR #3919, opened 2019-05-28
Uniformly use defined-or-not for -D flags

abadams:
For things like WITH_X86, some of our external flags were -DFLAG=1 and tested using a mix of #if and #ifdef, and some were -DFLAG and tested with #ifdef alone. This PR changes everything to use defined-or-not-defined. Internal-to-a-file debugging flags (e.g. LOG_EXPR_MUTATIONS) were unchanged.


steven-johnson:
Failure is unrelated

PR #3920, opened 2019-05-28
Store with

abadams:
Add a scheduling directive to explicitly alias the memory of two Funcs. Can be used to save memory, perform operations in-place to have better cache properties and less address computation, or to elide copies entirely for zero-copy concat and reshape. See test/correctness/store_with.cpp for a whole ton of use-cases.

Doesn't support input images yet. 

This PR gets zero-copy concats by depending on #3774 . That PR should be tested in Google and merged first, because it affects bounds inference semantics. That aside, this PR shouldn't affect the behavior of existing code.


src/Func.h
```diff
@@ -2079,6 +2080,94 @@ class Func {
...
     Func &compute_root();
 
+    /** Store this Func in the same piece of memory used to back some
+     * other Func given as an argument. This Func must be scheduled
+     * within the store_at level of the argument Func so that its
+     * realization is in scope.
+     *
+     * Halide will attempt to prove that for each site in the shared
+     * memory, all use of that site for the purpose of backing this
+     * Func occurs earlier in time than all of use that site for the
```
shoaibkamil: 'of use' --> 'use of'

src/StoreWith.h
```diff
@@ -0,0 +1,22 @@
+#ifndef HALIDE_INTERNAL_STORE_WITH
+#define HALIDE_INTERNAL_STORE_WITH
+
+#include <map>
+
+#include "IR.h"
+
+// TODO: file docs
```
abadams: Oops, forgot about these TODOs. Will get to them shortly.

steven-johnson:
I hadn't realized that #3774 was ready for trying out again yet. I'll get to that shortly.

src/StoreWith.cpp
```diff
@@ -0,0 +1,1483 @@
...
+    Expr most_recent_substitution;
+
+    // An additional arbitrary term to place non-linear constraints
+    Expr non_linear_term;
+
+    // A heuristic for how close we are to finding infeasibility
+    float c = 0;
+
+    // unique IDs for each system for debugging and training a good heuristic
+    static int id_counter;
```
steven-johnson: Reality check: any chance of this overflowing? What would happen if it did?

steven-johnson:
This is some very cool stuff. It does seem ripe for possible corner-case failures, however. I suspect it might be useful to think about doing fuzz-testing where we generate random-ish schedules that use store_with().

steven-johnson:
(FYI, I landed #3774 then updated this to master.)

src/StoreWith.cpp
abadams: That would require the system to have considered 2^31 different possible systems of equations, which is at least several hours inside this lowering pass. The consequences of overflow are that my script to train the coefficients would get confused about which systems lead to which other systems.

abadams: I'll change it to a size_t so that readers don't have to think about it

steven-johnson: With a uint64_t we still might overflow after ~4M years of running in this lowering pass. I guess that's ok.

abadams: I worried about that, but at least the overflow will be defined behavior. 

More seriously, I changed it so that you don't have to look at it and think "is that going to tickle UBsan when it overflows?"

steven-johnson:
Where do we stand on this PR... does it need more work, more tests, splitting into several PRs...?

abadams:
Needs more tests, as Dillon requested. It's on my todo list after float16.

steven-johnson:
Periodic ping for status update.

abadams:
I'm working on a variety of improvements to the constraint system solver that this branch uses as part of a separate research project. I was going to backport those improvements to this branch once the research project is done. If someone desperately needs this feature I can reprioritize.

steven-johnson:
Sync to master, plus a periodic ping to see if there's any update on where this stands.

abadams:
Still on my TODO list, slightly higher than before.

steven-johnson:
Marking as `skip_buildbots` just in case the Buildbot decides to try building this (it's not gonna build at all until synced to master)

abadams:
This is a potentially useful feature still to eliminate certain memcpys, but the reliance on a constraint system solver made me conclude that it just wouldn't be reliable. Closing until such time as I think of a more straightforward way to solve the problem.

steven-johnson:
Alas! This would be quite cool if we can ever make it work.

Issue #3926, opened 2019-05-31
specialize() and compute_at() interaction failing


shoaibkamil:
Given this code:
```
#include "Halide.h"

using namespace Halide;

int main(int argc, char* argv[]) {
  Func f("f"), g("g");
  Var x("x"), y("y");
  Var tx("tx"), ty("ty");
  Param<bool> param;

  f(x) = x;
  g(x, y) = f(x) + select(param, 1, 2);

#if 0
  g.gpu_tile(x, y, tx, ty, 8, 8, TailStrategy::GuardWithIf);
#else
  g.specialize(param).gpu_tile(x, y, tx, ty, 8, 8, TailStrategy::GuardWithIf);
  g.specialize(!param).gpu_tile(x, y, tx, ty, 8, 8, TailStrategy::GuardWithIf);
  g.specialize_fail("Unknown");
#endif
  f.in().compute_at(g, tx);

  g.compile_to_file("foo", {param});

}
```

This gives an error:
```
Error:
Func "f_global_wrapper" is computed at the following invalid location:
  f_global_wrapper.compute_at(g, tx);
Legal locations for this function are:
  f_global_wrapper.compute_root();
  f_global_wrapper.compute_at(g, Var::outermost());
  f_global_wrapper.compute_at(g, y);
  f_global_wrapper.compute_at(g, x);
"f_global_wrapper" is used in the following places:
  for g.s0.y.y:
    for g.s0.x.x:
      for g.s0.y.ty:
        for g.s0.x.tx:
          g uses f_global_wrapper
  for g.s0.y.y:
    for g.s0.x.x:
      for g.s0.y.ty:
        for g.s0.x.tx:
          g uses f_global_wrapper
  for g.s0.y:
    for g.s0.x:
      g uses f_global_wrapper
```

However, the last use (causing the violation) is spurious, because the `specialize_fail()` ensures this branch will never be reached.  Looking at the IR, I see:
```
...
    } else if ((uint1)1) {
      assert((uint1)0, halide_error_specialize_fail("Unknown"))
    } else {
      // violating usage here
    }
....
```

steven-johnson:
Is this resolved by #3927?

shoaibkamil:
Yup, resolved.

PR #3927, opened 2019-05-31
Don't replicate the original stmt when doing specialize_fail()

shoaibkamil:
Proposed fix for #3926.

### DIFF
--- a/src/ScheduleFunctions.cpp
+++ b/src/ScheduleFunctions.cpp
@@ -358,9 +358,9 @@ Stmt build_provide_loop_nest(const map<string, Function> &env,
     const vector<Specialization> &specializations = def.specializations();
     for (size_t i = specializations.size(); i > 0; i--) {
         const Specialization &s = specializations[i - 1];
-        Stmt then_case;
         if (s.failure_message.empty()) {
-            then_case = build_provide_loop_nest(env, prefix, func, s.definition, start_fuse, is_update);
+            Stmt then_case = build_provide_loop_nest(env, prefix, func, s.definition, start_fuse, is_update);
+            stmt = IfThenElse::make(s.condition, then_case, stmt);
         } else {
             internal_assert(equal(s.condition, const_true()));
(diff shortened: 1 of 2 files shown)


src/ScheduleFunctions.cpp
```diff
@@ -371,8 +373,11 @@ Stmt build_provide_loop_nest(const map<string, Function> &env,
                                      {StringImm::make(s.failure_message)},
                                      Internal::Call::Extern);
             then_case = AssertStmt::make(const_false(), specialize_fail_error);
+            // Since the condition is a constant true, the else case will
+            // never execute
+            else_case = then_case;
         }
-        stmt = IfThenElse::make(s.condition, then_case, stmt);
+        stmt = IfThenElse::make(s.condition, then_case, else_case);
```
abadams: Isn't this constructing IfThenElse::make(s.condition, then_case, then_case)? Why not just set stmt = then_case in the else branch above and only construct the IfThenElse in the empty-failure-message branch.

shoaibkamil: Very true.  Pushed a simpler version that doesn't construct a redundant IfThenElse.

shoaibkamil:
Failures seem to be `simd_op_check`.  Ok to merge @steven-johnson?

test/correctness/issue_3926.cpp
```diff
@@ -0,0 +1,26 @@
...
+int main(int argc, char* argv[]) {
+  Func f("f"), g("g");
+  Var x("x"), y("y");
+  Var tx("tx"), ty("ty");
+  Param<bool> param;
+
+  f(x) = x;
+  g(x, y) = f(x) + select(param, 1, 2);
+
+  //g.gpu_tile(x, y, tx, ty, 8, 8, TailStrategy::GuardWithIf);
```
steven-johnson: Just noticed this, but: did you mean to leave this line commented out?

PR #3929, opened 2019-05-31
Explicitly set SHELL=bash in Makefile.inc

steven-johnson:
(0) Our main Makefile already defines this, so we already rely on bash
(1) The SHELL var isn't inherited by sub-makefiles
(2) apps/Makefile.inc relies on bash (vs plain sh) syntax
(3) Some of our buildbots don't define SHELL


steven-johnson:
Fine with me, but the definition in the toplevel Makefile isn't conditional; we should change it too.

PR #3931, opened 2019-06-04
Remove extra instruction generated in OpenCL backend

BachiLi:
There seems to be an extra print_expr in OpenCL backend when processing Store. This patch removes that line. Sorry if the extra print is intentional.


abadams:
The first line prints the index, and the second line prints the value.

BachiLi:
Yes, but the value is already printed at
https://github.com/halide/Halide/blob/b8c031a07f5296cc0e4eaacb9c51cb7546160999/src/CodeGen_OpenCL_Dev.cpp#L305

abadams:
Wow.

abadams:
Feel free to merge once all of the buildbots are green.

Issue #1017, opened 2016-02-03
Atomic updates


abadams:
We should be able to parallelize associate operations across RDoms in several ones, one of which is using atomic operations, e.g:

```
RDom r(0, 100);
f() = 0
f() += g(r);
f.update().atomic().parallel(r);
```

abadams:
See https://lists.csail.mit.edu/pipermail/halide-dev/2016-February/002194.html for a discussion.

BachiLi:
Hi,

Is there still any plan to add the atomic updates? This would be useful for general scattering operations, e.g.,
```
f(g(r.x)) += h(r.x) 
```
This happens a lot when computing the gradients. We hacked it in the gradient-halide by having atomics for all += in the PTX backend ( https://github.com/jrk/gradient-halide/blob/master/src/CodeGen_PTX_Dev.cpp#L247 ), but something like what is proposed in this issue would be very useful.

BachiLi:
Hi,

Is there still any plan to add the atomic updates? This would be useful for general scattering operations, e.g.,
```
RDom r(0, 1000001);
f(clamp(r.x, 0, 1000000)) += g(r.x) 
```
This happens a lot when computing the gradients. We hacked it in the gradient-halide branch by having atomics for all += in the PTX backend ( https://github.com/jrk/gradient-halide/blob/master/src/CodeGen_PTX_Dev.cpp#L247 ), but something like what is proposed in this issue would be very useful.

PR #3936, opened 2019-06-09
Atomics support

BachiLi:
Hi,

This is an attempt to add basic atomics support to Halide. Following the suggestions in https://github.com/halide/Halide/issues/1017, I added an atomic scheduling directive, such that you can write a schedule like this:
```
hist(x) = 0;
hist(im(r)) += 1;
hist.compute_root();
hist.update().atomic().parallel(r); 
```
(also see test/correctness/atomics.cpp)
I implement this by adding an atomic flag in the Store node, and emit corresponding backend code by checking the flag. LLVM, OpenCL, and CUDA backends are supported. Vectorized store is not supported yet. I didn't implement the coarser-grained locking mechanism discussed in the mailing list, as it requires more complicated backends. All atomics are implemented through an atomicAdd or an atomicCompareAndSwap loop. I don't have machines at the moment to test the OpenGLCompute and DirectX12 backends so I didn't implement atomics backends for them. The PTX backend mostly copied-pasted from the gradient-halide repo.

As I mentioned in the issue, atomics are important, if not essential, for gradients computation, especially on GPUs. It would be nice if you can take a look when you have time.


src/Func.cpp
```diff
@@ -289,7 +289,8 @@ void Stage::set_dim_type(VarOrRVar var, ForType t) {
             // If it's an rvar and the for type is parallel, we need to
             // validate that this doesn't introduce a race condition.
             if (!dims[i].is_pure() && var.is_rvar && is_parallel(t)) {
-                user_assert(definition.schedule().allow_race_conditions())
+                user_assert(definition.schedule().allow_race_conditions() ||
+                            definition.schedule().atomic())
                     << "In schedule for " << name()
                     << ", marking var " << var.name()
                     << " as parallel or vectorized may introduce a race"
```
steven-johnson: We should probably augment the assertion-failure text to also mention `.atomic()` as an option.

src/Func.h
```diff
@@ -1532,6 +1533,9 @@ class Func {
      * different values at different times or on different machines. */
     Func &allow_race_conditions();
 
+    /** Issue atomic store for this Func. This allows parallelization on arbitrary RVars. */
```
steven-johnson: We'll need to substantially augment this docstring before landing.

src/Bounds.cpp
```diff
@@ -2659,7 +2659,7 @@ void boxes_touched_test() {
     Scope<Interval> scope;
     scope.push("y", Interval(Expr(0), Expr(10)));
 
-    Stmt stmt = Provide::make("f", {10}, {x, y, z, w});
+    Stmt stmt = Provide::make("f", {10}, {x, y, z, w}, false);
```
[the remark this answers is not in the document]
BachiLi: I annotated all of them. I am open to default to false or having an enum. I didn't set the default arguments precisely because I needed to pass the flags from Provide to Store.

abadams:
I'm worried about this case:

```
  foo[x] = foo[x] + bar + foo[...];
```
then some optimization pass runs (e.g. the one that does interleaving stores, or LICM, or some form of CSE) and you get:
```
let tmp = bar + foo[...];
foo[x] = foo[x] + tmp;
```
The you lower just the store to be atomic, and it's not correct, because there's a load from foo outside the CAS loop. 

If we want the RHS of the store to be atomic there's an additional constraint that we can't lift Exprs out of it, and we'd have to be careful to respect that constraint everywhere else in the compiler that might move Exprs around. I'm just not sure how to do that reliably. I'll mull it over.

src/CodeGen_LLVM.cpp
```diff
@@ -2021,6 +2021,7 @@ void CodeGen_LLVM::scalarize(Expr e) {
 }
 
 void CodeGen_LLVM::codegen_predicated_vector_store(const Store *op) {
+    internal_assert(!op->is_atomic); // Doesn't support atomic vector store
```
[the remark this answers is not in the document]
BachiLi: Added a user_assert.
Scalarizing the store sounds like a reasonable plan. I can code it up later.

steven-johnson:
Another thought: does this need any special attention if used in conjunction with `.async()`?

src/CodeGen_LLVM.cpp
```diff
@@ -2210,6 +2211,81 @@ void CodeGen_LLVM::codegen_predicated_vector_load(const Load *op) {
...
+    user_assert(op->value.type().is_scalar()) << "Atomic store does not support vectorization.\n";
+
+    // Detect whether we can describe this as an atomic-read-modify-write, 
+    // otherwise fallback to a compare-and-swap loop.
+    // Current only test for atomicAdd
+    Expr val_expr = op->value;
+    Expr equiv_load = Load::make(op->value.type(), op->name, op->index, Buffer<>(), op->param, op->predicate, op->alignment);
+    Expr delta = simplify(common_subexpression_elimination(op->value - equiv_load));
+    bool is_atomic_add = val_expr.type().is_int_or_uint() && !expr_uses_var(delta, op->name);
+    if (is_atomic_add) {
```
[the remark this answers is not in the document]
BachiLi: This is checking if delta = op->value - store[index] is independent of store, so we can rewrite things as store[index] += delta. I added some comments.

src/CodeGen_LLVM.cpp
```diff
@@ -2210,6 +2211,81 @@ void CodeGen_LLVM::codegen_predicated_vector_load(const Load *op) {
...
+        Value *ptr = codegen_buffer_pointer(op->name, op->value.type(), op->index);
+        Value *val = codegen(delta);
+        // llvm 9 has FAdd which can be used for atomic floats,
+        // but currently Halide depends on llvm 7
+        builder->CreateAtomicRMW(AtomicRMWInst::Add, ptr, val, AtomicOrdering::Monotonic);
```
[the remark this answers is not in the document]
BachiLi: Will deal with this later. Is it OK to generate different code for LLVM8- and LLVM9+, without issuing errors/assertions?

src/CodeGen_OpenCL_Dev.cpp
```diff
@@ -7,6 +7,9 @@
 #include "EliminateBoolVectors.h"
 #include "IRMutator.h"
 #include "IROperator.h"
+#include "CSE.h"
```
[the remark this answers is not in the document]
BachiLi: Done

src/CodeGen_OpenCL_Dev.cpp
```diff
@@ -302,6 +305,106 @@ void CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::visit(const Load *op) {
 void CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::visit(const Store *op) {
     user_assert(is_one(op->predicate)) << "Predicated store is not supported inside OpenCL kernel.\n";
 
+    if (op->is_atomic) {
+        // Currently only support scalar atomics
+        user_assert(op->value.type().is_scalar()) << "Atomic store does not support vectorization.\n";
+        user_assert(op->value.type().bits() >= 32) << "OpenCL only support 32 and 64 bits atomics.\n";
```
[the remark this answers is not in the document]
BachiLi: Done

src/CodeGen_OpenCL_Dev.cpp
```diff
@@ -302,6 +305,106 @@ void CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::visit(const Load *op) {
...
+                "Enable feature CLAtomics64 for 64-bit atomics in OpenCL.\n";
+        }
+        // Detect whether we can describe this as an atomic-read-modify-write, 
+        // otherwise fallback to a compare-and-swap loop.
+        // Current only test for atomic_add
+        Expr val_expr = op->value;
+        Type t = val_expr.type();
+        Expr equiv_load = Load::make(t, op->name, op->index, Buffer<>(), op->param, op->predicate, op->alignment);
+        Expr delta = simplify(common_subexpression_elimination(op->value - equiv_load));
+        bool is_atomic_add = t.is_int_or_uint() && !expr_uses_var(delta, op->name);
```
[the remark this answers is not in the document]
BachiLi: Added comments

src/CodeGen_OpenGLCompute_Dev.cpp
```diff
@@ -5,6 +5,8 @@
 #include "IRMutator.h"
 #include "IROperator.h"
 #include "Simplify.h"
+#include "CSE.h"
```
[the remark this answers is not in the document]
BachiLi: This header is not used eventually. Removed.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -10,6 +10,7 @@
 #include "LLVM_Headers.h"
 #include "LLVM_Runtime_Linker.h"
 #include "Simplify.h"
+#include "CSE.h"
```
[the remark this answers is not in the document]
BachiLi: Done.

BachiLi:
> I'm worried about this case:
> 
> ```
>   foo[x] = foo[x] + bar + foo[...];
> ```
> 
> then some optimization pass runs (e.g. the one that does interleaving stores, or LICM, or some form of CSE) and you get:
> 
> ```
> let tmp = bar + foo[...];
> foo[x] = foo[x] + tmp;
> ```
> 
> The you lower just the store to be atomic, and it's not correct, because there's a load from foo outside the CAS loop.
> 
> If we want the RHS of the store to be atomic there's an additional constraint that we can't lift Exprs out of it, and we'd have to be careful to respect that constraint everywhere else in the compiler that might move Exprs around. I'm just not sure how to do that reliably. I'll mull it over.

This sounds quite tricky. My current thought is to add a "deoptimization" pass before the codegen phase, check for all LetStmt, and substitute all let variables in if we detect such situation. If you think this is fine I can proceed to implement this.

BachiLi:
> Another thought: does this need any special attention if used in conjunction with `.async()`?

async should be fine but I should add tests.

src/Func.cpp
BachiLi: Added something.

zvookin:
The same mechanism used by strict_float might be useful to prevent transformations from changing an atomic construct in an invalid way. It wraps an intrinsic around every Expr in the context to make the simplifier, etc. not recognize anything inside the tree of IR. This is the only viable way I could figure out to accomplish this purpose.

src/Func.cpp
steven-johnson: Maybe I'm confused, but: I was assuming that if you specify `.atomic()`, it makes race conditions impossible. Am I incorrect?

If my assumption is correct, then what we want is something like:

    user_assert(allow_race && !atomic, "existing warning test plus comment about how using .atomic() may prevent the race condition")

abadams:
Even if it's only possible to generate non-failing cases from the front-end, many passes have to transform any legal Stmt IR (because who knows what new scheduling directives will come in the future). Atomic should be properly respected.

That said, I believe any tuple reduction will currently generate a failure case for you, because the value gets lifted out into a let stmt (e.g. try a multi-dimensional argmin). Even if each store was individually atomic, they need to be atomic as a group for it to work.

Starting to think it's not stores that need to be atomic, but rather we need an "Atomic" IR node that can contain any sub-stmt.

BachiLi:
Indeed. Guess I'll have to implement general locking then. I'll also take a look at strict_float. Thanks!

abadams:
The Acquire IR node would work for general locking. Freeing the lock requires a semaphore release and making the lock requires a semaphore init, so it's not the cleanest.

src/Bounds.cpp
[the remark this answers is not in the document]
BachiLi: Not applicable now.

src/CodeGen_LLVM.cpp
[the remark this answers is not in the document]
BachiLi: Added vectorization through scalarization of the store. Predicated vector is still not supported since it requires pretty complicated codegen logic for the CAS loop (need to specialize when the predicate does recursively depend on the function).

src/CodeGen_LLVM.cpp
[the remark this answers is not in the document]
BachiLi: Added macros to use FAdd when possible.

src/Func.cpp
BachiLi: This is actually a bit complicated: parallel reduction variable can change the meaning of the algorithm even with atomic, which breaks Halide's philosophy. Therefore we want to prove associativity when specifying atomic(). However, as mentioned in the allows_race_condition error message, sometimes the user may still want parallel reduction even when the reduction is not associative (or Halide can't prove it), so we still want something like allow race condition for atomic updates.

Perhaps the problem is the naming of allow race condition: maybe should call it force_parallel_reduction or something like that.

BachiLi:
Hi,

I added an atomic node in the IR as suggested by Andrew. This allows us to lock a region using an array of mutex (each element of the store location is protected by a mutex). This only works on CPU as GPU requires all threads in a warp to do the same operations and mutex lock leads to deadlock. For a function update I check for whether the reduction requires a mutex lock during the schedule_function lowering pass: this only happens when the function is a tuple and the update recursions are not independent between each tuple elements, as compare-and-swap loop cannot update two disjoint locations at the same time. If it requires the mutex lock, I insert Allocate nodes and calls to memset to initialize the mutex buffer. If target is GPU and we instantiate mutex buffers then assertions are triggered.

For the let lifting problem Andrew mentioned, I added a lowering pass to substitute the let variables back (see SubstituteLetInAtomics.h/.cpp). It also checks if the substitution changes the meaning of IR or not. If we cannot substitute without changing the meaning, assertion is triggered and likely some optimization/scheduling passes have broken the atomic updates.

Haven't added the async test yet. Also I don't plan to support vectorization on GPU yet (CUDA/OpenCL don't have 8-bit or 16-bit atomic operations anyway). Predicated vector is also not implemented as it requires pretty complex codegen logic. The use of predicated vector should be very rare anyway.

src/Func.cpp
BachiLi: My current solution to this is a flag in the atomic function that can turn off the associativity test (see https://github.com/halide/Halide/pull/3936/commits/a26821fa9065d766feedeeaf0f83377963fac6f7).

src/Func.h
BachiLi: Augmented. Please take a look.

BachiLi:
Hi,

I did my best to fix all the issues raised here. Please take a look when you have time.

abadams:
Sorry for the long delay on this - adding a new Stmt node and thinking through whether existing lowering passes respect it is tricky, and I'm frantically preparing for siggraph.

BachiLi:
I know. Take your time ; )

BachiLi:
@abadams Would it be easier if we only support atomic add? We don't need new IR node in that case, and it's probably the most common use case of atomic.

abadams:
How would that remove the need for the IR node? You'd have to lower to the atomic add quite early I guess, before any expressions might be lifted out of the RHS of a provide node in a way that makes things non-atomic.

BachiLi:
Because if the operand in the atomic add doesn't have the source, you don't need to worry about any kind of code motion? We can add check in the beginning to reject code that aren't expressible as atomic add.

BachiLi:
You're right, I was confused. I guess always substituting the let node inside for a Store node marked as atomic is one possible solution, though it might lead to less efficient code.

abadams:
I'm not so sure I was right. I think all the way up to the "split_tuples" phase of lowering we're careful to never pull expressions out of the right-hand-side of a provide node. So if we peephole recognized atomic ops any time before that point we'd be OK.

abadams:
We'd still need to represent the atomic op in the IR in some reasonable way though so that other passes understand that it's both a load and a store (e.g. for box_touched)

BachiLi:
So: an atomic node that basically function like Store or Provide but has a different type so that people need to know what they are doing when they operate on it?

BachiLi:
Summarizing from the meeting today so I don't forget: the overall design of using mutex locks and atomic adds seems to be fine, and the let lifting issue is fine since store is a non-pure operator, and all the optimization passes in Halide are very careful not to lift non-pure operations. Therefore the let substitution in the current implementation can be removed.

We need more tests to make sure this doesn't generate deadlocks or incorrect mutex arrays. In particular:
1) We want tests for interaction between compute_at, rfactor, async, and the mutex locks. Produce tests that has non-trivial dependency between the locks.
2) Make sure store_at/sliding window optimization doesn't produce invalid code.
3) Make sure we handle vectorize() correctly.

In the future we might want to improve the mutex lock allocation code to make it more efficient.

zvookin:
Feel free to move the failing test into test/failing_with_issue .

BachiLi:
Hey @abadams @steven-johnson @zvookin,

I think this is ready for another round of review. Let me provide a summary:

This pull request adds a new scheduling directive `atomic()` to Func and Stage. Once applied, all stores to the corresponding Stage of the Func becomes atomic. This allows us to parallelize associative reduction variables,  and provides a different trade-offs comparing to `rfactor`. This is likely to be more efficient when the Func we are storing to has a huge domain. An example usage:
```
    im(x) = ...
    hist(im(r)) += 1;
    hist.update().atomic().parallel(r);
```

Once a stage is flagged as atomic, the lowered code has three possible outcomes:
1) The stores to the stage are lowered into hardware atomic add instructions. This happens when the add operand is independent to the Func we are storing to. The histogram example above will be lowered into atomic add.
2) There stores to the stage are lowered into hardware compare-and-swap instructions. This happens when the update expression is not expressible as an atomic add, and Func's Tuple values doesn't need to be lowered "atomically" (sorry for the naming overload here, it means the Tuple values need to be computed first in Let variables, and then stored into the Func). 
3) The stores to the stage are protected by an array of mutex locks, where the size of the array is the same as the storage size of the Func. This happens 1) and 2) don't happen.

If the reduction variable being parallelized is not associative, this leads to a compile error, but can be overridden by setting the `override_associativity_test` to true in the atomic function. This is necessary for parallelizing reduction variables on float16 functions since Halide's associativity prover doesn't support float16 yet.

Currently the atomic operation is supported by x86, CUDA, OpenCL backends. Compiling to other backends result in a compile error. If an operation is compiled into a mutex lock, and is vectorized or is compiled to CUDA or OpenCL, it also results in a compile error, since per-element mutex lock on vectorized operation leads to a deadlock. Vectorization of predicated RVars (through rdom.where()) on CPU is also unsupported yet. 8-bit and 16-bit atomics on GPU are also not supported.

I made the following changes to Halide's lowering passes:
* A new statement node `Atomic` is added to the IR. It contains the name of the mutex lock buffer if it exists, and the indices for accessing the mutex lock, and all the `Provide` or `Store` statements in an Atomic node should be executed atomically.
* In the `schedule_function` pass, the `Atomic` node is emitted outside the `Provide` node for the stages marked as atomic. For those stages, we also emit an `Allocate` node with a call to `memset` to initialize the mutex locks just before the `Realize` node.
* A new lowering pass `remove_atomic_mutex_locks`. This pass removes the unnecessary mutex lock buffers by detecting whether the `split_tuple` pass has emitted Tuple with lifted variables.
* Special handling of atomic nodes and the mutex buffers in the `fork_async_producers` pass. When forking the the producers and consumers, the Allocate nodes responsible for initializing the mutex buffers need to be handled correctly, to avoid creating dangling allocate/memset/let variables for the bounds of the mutex buffers.
* In the llvm codegen, when an `Atomic` is visited, if the mutex lock is required, then emit the lock/free code, otherwise for the `Store` nodes inside the `Atomic` node we generate atomic add or atomic compare and swap. For PTX backend we use nvvm's intrinsics for atomic add (in the future this can be replaced by llvm's FAdd flag in the atomic add introduced in llvm 9). For OpenCL a new target flag `cl_atomics64` is introduced for 64-bit atomics.
* Python bindings of atomic is also provided.

Future work:
* half atomic add on CUDA. In particular CUDA can take two half floats and atomic add them at the same time. We can detect and output more optimized code when the vector size is 2.
* Predicated vectorization. The codegen logic seems quite complicated especially with CAS loops.
* Other backends. OpenGLCompute and D3D12 are probably the easiest ones. GLSL and Metal seem very tricky. I know nothing about WebAssembly.
* Hardware transaction memory can be interesting. There might be more efficient ways to do the mutex locking.

BachiLi:
Pinging again (I know this is a huge PR).

src/CodeGen_LLVM.cpp
```diff
@@ -2457,6 +2459,150 @@ void CodeGen_LLVM::codegen_predicated_vector_load(const Load *op) {
...
+                } else {
+                    builder->CreateAtomicRMW(AtomicRMWInst::Add, ptr, v, AtomicOrdering::Monotonic);
+                }
+#else
+                builder->CreateAtomicRMW(AtomicRMWInst::Add, ptr, v, AtomicOrdering::Monotonic);
+#endif
+            }
+        }
+    } else {
+        // We want to create the following CAS loop:
```
abadams: Can this logic be moved into a runtime function, so that it can be written in C or .ll? Codegen_LLVM is already huge.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -263,6 +264,40 @@ void CodeGen_PTX_Dev::visit(const Load *op) {
...
 
 void CodeGen_PTX_Dev::visit(const Store *op) {
+    if (emit_atomic_stores) {
+        user_assert(op->value.type().is_scalar()) << "CUDA atomic update does not support vectorization.\n";
+        user_assert(is_one(op->predicate)) << "Atomic update does not support predicated store.\n";
+        user_assert(op->value.type().bits() >= 32) << "CUDA: 8-bit or 16-bit atomics are not supported.\n";
+#if LLVM_VERSION < 90
+        // Generate nnvm intrinsics for the atomics if this is an float atomicAdd.
+        // Otherwise refer to the llvm codegen. For llvm version >= 90, atomicrmw support floats so we
+        // can also refer to llvm.
```
abadams: refer -> defer

src/IR.h
```diff
@@ -814,6 +814,24 @@ struct Prefetch : public StmtNode<Prefetch> {
     static const IRNodeType _node_type = IRNodeType::Prefetch;
 };
 
+/** Lock all the Store nodes in the body statement.
```
abadams: Nit: our multiline comments include a `*` at the start of each line

src/RemoveAtomicMutexLocks.cpp
```diff
@@ -0,0 +1,102 @@
...
+using std::vector;
+
+namespace {
+
+class FindAtomicTupleProvide : public IRVisitor {
+    using IRVisitor::visit;
+
+    void visit(const Variable *op) override {
+        if (current_provide_name != "" &&
+                op->name == current_provide_name + ".value") {
```
abadams: Finding variables by name like this is very fragile. Is there an alternative?

abadams:
Using an Allocate node with a Block and a memset like that is very fragile. Intervening passes could mess up the structure, and it makes it hard to do anything clever w.r.t packing the mutexes more efficiently. I propose the following alternative abstract interface. Construct a set of runtime functions:
```
struct halide_mutex_array;
halide_mutex_array *halide_mutex_array_create(int sz);
halide_mutex_array *halide_mutex_array_destroy();
void halide_mutex_array_lock(halide_mutex_array *array, int entry);
void halide_mutex_array_unlock(halide_mutex_array *array, int entry);
```

For now these would be implemented very simply in some new runtime module src/runtime/mutex_array.cpp using a literal array of halide_mutex 

Then in the IR instead of an Allocate node around a Block with a call to memset, make it an Allocate node like so:
```
Expr mutex_array = Call::make(type_of<halide_mutex_array *>(), Call::create_mutex_array, {size}, Call::Extern);
Stmt alloc = Allocate::make(some_name, type_of<halide_mutex_array *>(), MemoryType::Stack, {}, const_true(), body, mutex_array, Call::free_mutex_array);
```
Note that this is a scalar allocate node, that makes a single halide_mutex_array * using a custom new/free function. The reason it's not just a letstmt is exception-safety - if there's an error in the body of the allocate node, free_mutex_array will still be called.

src/ScheduleFunctions.cpp
```diff
@@ -1099,9 +1113,45 @@ class InjectFunctionRealization : public IRMutator {
     const LoopLevel &store_level;
 
     Stmt build_realize(Stmt s, const Function &func, bool is_output) {
+        const string &name = func.name();
+        if (_funcs_need_alloc_mutex.find(func) != _funcs_need_alloc_mutex.end()) {
+            // Place an Allocate node here and a call to initialization to create a mutex buffer
+            debug(1) << "Injecting allocation of mutex buffer for " << name << '\n';
```
abadams: I think this is way too early to be injecting the mutex buffers. Can they be introduced after storage flattening in a dedicated pass? Then you won't have to inject dependencies on symbols that don't exist yet.

src/AsyncProducers.cpp
```diff
@@ -623,7 +679,12 @@ class TightenForkNodes : public IRMutator {
 
     Stmt visit(const LetStmt *op) override {
         Stmt body = mutate(op->body);
-        if (in_fork && !stmt_uses_var(body, op->name)) {
+        // Special case handling for .min & .stride fields.
+        // Atomics generate these let variables for the mutexes
+        // before the storage flatten stage insert the usages of them.
+        bool is_storage_var = op->name.find(".min.") != std::string::npos ||
```
abadams: This string-matching is a red flag. If the mutex buffers are injected later in lowering, could the modifications to AsyncProducers go away or at least be simpler?

src/CodeGen_LLVM.cpp
BachiLi: Hmm might be difficult since we want to inline the value computation inside the loop...

src/CodeGen_PTX_Dev.cpp
BachiLi: fixed

src/IR.h
BachiLi: fixed

src/RemoveAtomicMutexLocks.cpp
BachiLi: Should be fixed. Now we search for let bindings variables that references the Provide nodes inside the same Atomic node.

src/ScheduleFunctions.cpp
BachiLi: Yes they can. This is done.

src/AsyncProducers.cpp
BachiLi: This goes away now.

BachiLi:
I implemented most of @abadams's suggestions. Now the mutex locks/allocation are generated in a separate pass called `AddAtomicMutex`. The mutex is only added if we detect a let binding references to the buffer marked atomic.

There are two caveats:
1. I couldn't figure out how to make the CAS loop codegen a runtime library -- the computation inside the CAS loop needs to be inlined and is different for every atomic expressions.
2. The runtime mutex array creation is done, but the interface is a bit different. It looks like this now:
```
extern struct halide_mutex_array** halide_mutex_array_create(int sz);
extern void halide_mutex_array_destroy(void *user_context, void *array);
extern int halide_mutex_array_lock(struct halide_mutex_array *array, int entry);
extern int halide_mutex_array_unlock(struct halide_mutex_array *array, int entry);
```
The biggest difference being that halide_mutex_array_create generates a double pointer instead of just a pointer to the generated halide_mutex_array. This is related to how Halide's lowering is done and llvm's type system. If we want to make `halide_mutex_array_lock` takes `halide_mutex_array` instead of `halide_mutex_array*` as input, we need to generate a Halide Load statement with type `halide_mutex_array`, which is not supported in the current Halide typing system. If on the other hand we create the Load statement with a type with equivalent size, say, `uint64_t`, this generates a type error in the llvm codegen since llvm doesn't cast `uint64_t` to `halide_mutex_array` automatically.

Let me know if this is okay or if you have any suggestions to make this better. Thanks!

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    Stmt visit(const Atomic *op) override {
+        if (op->mutex_name == "") {
+            return IRMutator::visit(op);
+        }
+
+        // op->mutex_indices.size() could be 0.
+        Expr index = op->mutex_indices.size() == 1 ? op->mutex_indices[0] : Expr(0);
+        Stmt body = op->body;
+        Expr mutex = Load::make(type_of<halide_mutex_array *>(),
```
abadams: I think you can avoid the double-pointer. An Allocate node of type T with name "foo" lets you make load nodes from foo of type T, but it also places a Variable "foo" of type T* in scope. So you don't need to have a load node here - you can just make a Variable and pass it to the lock/unlock calls.

BachiLi: Ah great idea let me do that

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+                                op->mutex_name,
+                                Expr(0),
+                                Buffer<>(),
+                                Parameter(),
+                                const_true(),
+                                ModulusRemainder());
+        // Add mutex locks & unlocks
+        body = Block::make(
+            Evaluate::make(Call::make(type_of<int>(),
```
abadams: This is not exception-safe. If the body throws an error then the mutex is never unlocked. However, halide_mutex_array_destroy *will* be called, because it's the custom free function on an Allocate node. I guess  it just has to handle the case where some of the mutexes are still locked? I guess this is OK then, but you should add a comment.

src/CodeGen_LLVM.cpp
abadams: Fair enough.

src/CodeGen_LLVM.cpp
```diff
@@ -4136,6 +4288,22 @@ void CodeGen_LLVM::visit(const Shuffle *op) {
...
+    if (op->mutex_name != "") {
+        internal_assert(op->mutex_indices.size() <= 1) << "Atomic mutex access index should be flattened.";
+        Expr index = op->mutex_indices.size() == 1 ? op->mutex_indices[0] : Expr(0);
+        user_assert(index.type().is_scalar()) << "The vectorized atomic operation in " <<
+            op->producer_name << " is lowered into a mutex lock, which does not support vectorization.\n";
+
+        // Mutex locking & unlocking are handled by function calls generated by previous lowering passes.
+        codegen(op->body);
+    } else {
+        emit_atomic_stores = true;
```
abadams: I think this will break with nested atomic nodes (if that's even allowed). Use:
```
{
ScopedValue<bool> old_emit_atomic_stores(emit_atomic_stores, true);
codegen(op->body);
}
```

That will restore the old value of emit_atomic_stores instead of setting it to false.

Alternatively, assert !emit_atomic_stores before setting it to true if nested atomic nodes are not allowed.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+class FindAtomicLetBindings : public IRGraphVisitor {
+public:
+    using IRVisitor::visit;
+
+    FindAtomicLetBindings(const set<string> &store_names)
+        : store_names(store_names) {}
+
+    void visit(const Let *op) override {
+        include(op->value);
+        let_bindings.push(op->name, op->value);
```
abadams: Instead of push/pop pairs, use ScopedBinding:
```
include(op->value);
{
  ScopedBinding<Expr> bind(let_bindings, op->name, op->value);
  include(op->body);
}
```

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+        include(op->body);
+        if (op->producer_name == producer_name &&
+                op->mutex_name != "") {
+            if (found) {
+                // Multiple atomics inside the producer,
+                // make sure they have consistent information
+                internal_assert(mutex_name == op->mutex_name &&
+                    tuple_size == op->tuple_size &&
+                    dimensions == op->dimensions) <<
+                    "Inconsistent informatino of atomics inside a Producer node.\n";
```
abadams: information

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    using IRGraphVisitor::visit;
+
+    void visit(const Atomic *op) override {
+        for (Expr i : op->mutex_indices) {
+            include(i);
+        }
+        include(op->body);
+        if (op->producer_name == producer_name &&
+                op->mutex_name != "") {
```
abadams: Use string::empty() instead of comparing it to "" here and elsewhere

src/IR.h
```diff
@@ -814,6 +814,30 @@ struct Prefetch : public StmtNode<Prefetch> {
...
+ *  Typically the lock is implemented by an atomic operation
+ *  (e.g. atomic add or atomic compare-and-swap).
+ *  However, if necessary, the node can access a mutex buffer through
+ *  mutex_name and mutex_args, by lowering this node into
+ *  calls to acquire and release the lock. */
+struct Atomic : public StmtNode<Atomic> {
+    std::string producer_name;
+    std::string mutex_name; // empty string if not using mutex
+    std::vector<Expr> mutex_indices;
```
abadams: I don't think this field should be necessary, which would simplify the insertion of the mutexes (you won't have to worry about multiple dimensions and storage flattening interaction). Instead you can just lock the mutexes at each same single-dimensional coordinate that is accessed by Store nodes inside the body of the atomic

src/CodeGen_LLVM.cpp
```diff
@@ -2457,6 +2459,150 @@ void CodeGen_LLVM::codegen_predicated_vector_load(const Load *op) {
...
+    Expr equiv_load = Load::make(value_type, op->name,
+                                 op->index,
+                                 Buffer<>(),
+                                 op->param,
+                                 op->predicate,
+                                 op->alignment);
+    Expr delta = simplify(common_subexpression_elimination(op->value - equiv_load));
+#if LLVM_VERSION >= 90
+    bool is_atomic_add = !expr_uses_var(delta, op->name);
+    if (is_ptx_atomic_store) {
```
abadams: Rather than putting ptx-specific stuff in the base codegen class, how about a method on CodeGen_LLVM:

`virtual bool supports_atomic_add(const Type &t) const;`

Where the default implementation just checks if it's an Int, and Codegen_PTX_Dev overrides it to allow floats too.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+class AddAtomicMutex : public IRMutator {
+public:
+    using IRMutator::visit;
+
+    const Target &target;
+
+    AddAtomicMutex(const Target &target) : target(target) {}
+
+    Stmt visit(const ProducerConsumer *op) override {
```
abadams: What happens with sliding window? Should this instead happen at the Allocate node?

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+            }
+        } else {
+            return IRMutator::visit(op);
+        }
+
+        const std::string &producer_name = finder.producer_name;
+        const std::string &mutex_name = finder.mutex_name;
+        int tuple_size = finder.tuple_size;
+        int dimensions = finder.dimensions;
+        const char *extent_field = tuple_size == 1 ? ".extent." : ".0.extent.";
```
abadams: Constructing these by name is a red flag. I think if you 1) Do this at the allocate node and get the extent from the allocate node's extent, and 2) remove the indices from the atomic node in the IR and just lock whatever single-dimensional indices are stored to inside the body, then you won't need any of these.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,266 @@
...
+        }
+        Expr mutex_array = Call::make(type_of<halide_mutex_array **>(),
+                                      "halide_mutex_array_create",
+                                      {buffer_size},
+                                      Call::Extern);
+        Stmt body = mutate(op->body);
+        // Allocate a scalar of halide_mutex_array.
+        // This generate halide_mutex_array *mutex[1];
+        body = Allocate::make(mutex_name,
+                              type_of<halide_mutex_array *>(),
```
abadams: On second thought, the double-pointer may be necessary, because I don't think you can make an Allocate of `type_of<halide_mutex_array>()`

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: I made the type a Handle() and it was fine. The codegen doesn't seem to check the type of the buffers.

src/AddAtomicMutex.cpp
BachiLi: Not sure if I fully understand this but I added some comments in https://github.com/halide/Halide/commit/19a6bd69f4e0d15ba89f597b6877bf4f2316a3af

src/CodeGen_LLVM.cpp
BachiLi: Nested atomics without mutex locks is fine. Nested mutex locks lead to deadlocks. I added checks for the nested mutex locks case.

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Done

src/AddAtomicMutex.cpp
BachiLi: Done.

src/IR.h
BachiLi: Done. I noticed that there could be another Producer inside the Atomic due to inlining. So I still have to do some string matching to check whether the Store nodes is from the same producer.

src/CodeGen_LLVM.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Hmm. Directly modifying this code to take place in Allocate doesn't work since there is no allocate node for the output buffer. There indeed is a problem with sliding window that we over-allocate mutex buffer. Let me think.

src/AddAtomicMutex.cpp
BachiLi: I don't see an easy way to get rid of this at the moment if the corresponding buffer is the output.

BachiLi: Also I remember that the llvm codegen assumes all buffers have these fields when packing the closure. If I remove these it will likely trigger an assertion inside the codegen.

BachiLi: I'll store the extent in the atomic node if the producer is the output.

BachiLi: That doesn't work either because I'm just essentially moving these statements to where ever I construct that extent. In general Halide doesn't store the output extents in any IR nodes except these variables which you can only find by name.

BachiLi: I'll pass the output funcs/output buffers inside the AddAtomicMutex pass.

BachiLi: Done. See my comment below.

src/AddAtomicMutex.cpp
BachiLi: Some. See my comment below.

BachiLi:
Did another pass. Now the mutex buffer allocations are done at either an Allocate node or a ProducerConsumer node, depending on whether the mutex buffer is associated with an output buffer or not. If it is not, it reuses the extent of the Allocate node. If it is an output buffer, it takes the associated Parameter's extent. Currently there is no method to extract the extent field from the Parameter object directly, but it should be easy to add in the future and also simplifies passes such as `add_image_checks()`.

Not fully sure about exception safety since I have no idea how it works in Halide.

src/IR.h
BachiLi: I removed the string matching.

BachiLi:
I further cleanup the code such that there is almost no partial string matching and so during lowering.
(Sorry for the push, I just want to get this out of my todo list asap.)

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+using std::set;
+using std::string;
+
+namespace {
+
+class FindLoad : public IRGraphVisitor {
+public:
+    using IRGraphVisitor::visit;
+
+    set<string> symbols;
```
abadams: The set is not modified, so this should be a `const set<string> &` and initialized in the constructor to avoid a copy. Also most of this class and the classes below should be protected, not public

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    void visit(const Load *op) override {
+        if (symbols.find(op->name) != symbols.end()) {
+            found = true;
+        }
+        include(op->predicate);
+        include(op->index);
+    }
+
+    bool find_load(Expr e, const set<string> &symbols) {
```
abadams: This should be a constructor so that the set can be taken by reference

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+            include(op->body);
+        }
+    }
+
+    void visit(const Variable *op) override {
+        if (inside_store) {
+            if (let_bindings.contains(op->name)) {
+                Expr e = let_bindings.get(op->name);
+                FindLoad finder;
+                if (finder.find_load(e, store_names)) {
```
abadams: What happens with:
```
let a = f[x]
let b = 2*a
f[x] = b + 1
```

abadams: I think you want to maintain a Scope of all the vars that transitively depend on one of the store names, rather than your let_bindings scope

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+        FindAtomicLetBindings finder(collector.store_names);
+        op->body.accept(&finder);
+        if (finder.found) {
+            // Can't remove mutex lock. Leave the Stmt as is.
+            return IRMutator::visit(op);
+        } else {
+            remove_mutex_lock_names.insert(op->mutex_name);
+            Stmt body = mutate(op->body);
+            return Atomic::make(op->producer_name,
+                                "",
```
abadams: use std::string() instead of ""

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    bool in_atomic_mutex = false;
+    const std::set<std::string> &store_names;
+    bool found = false;
+    std::string producer_name;
+    std::string mutex_name;
+};
+
+class FindStoreIndex : public IRGraphVisitor {
```
abadams: What does this class even do?

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        if (allocated_mutexes.find(finder.mutex_name) != allocated_mutexes.end()) {
+            // We've already allocated a mutex.
+            return IRMutator::visit(op);
+        }
+
+        allocated_mutexes.insert(finder.mutex_name);
+
+        // We assume all output buffers have the same extent
```
abadams: More correctly:

// We assume all output buffers in a Tuple have the same extent

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,344 @@
...
+        Parameter output_buffer = f.output_buffers()[0];
```
abadams: Change Parameter to OutputImageParam and you'll get your extent methods to use below.

src/IR.cpp
```diff
@@ -810,6 +810,16 @@ bool Shuffle::is_interleave() const {
     return true;
 }
 
+Stmt Atomic::make(const std::string &producer_name,
+                  const std::string &mutex_name,
+                  Stmt body) {
+    Atomic *node = new Atomic;
```
abadams: Should probably assert body.defined()

src/IRMutator.cpp
```diff
@@ -392,6 +392,21 @@ Stmt IRMutator::visit(const Acquire *op) {
     }
 }
 
+Stmt IRMutator::visit(const Atomic *op) {
+    bool changed = false;
+    Stmt body = mutate(op->body);
+    if (!body.same_as(op->body)) {
```
abadams: Don't need the changed variable here, just make `if (!body.same_as(op->body))` the condition

src/runtime/fake_thread_pool.cpp
```diff
@@ -78,6 +78,21 @@ WEAK void halide_mutex_lock(halide_mutex *mutex) {
 WEAK void halide_mutex_unlock(halide_mutex *mutex) {
 }
 
+WEAK halide_mutex_array* halide_mutex_array_create(int sz) {
```
abadams: Didn't notice this was the fake_thread_pool and wondered for a long time how this could possibly work :)

src/runtime/synchronization_common.h
```diff
@@ -1138,4 +1138,36 @@ WEAK void halide_cond_wait(struct halide_cond *cond, struct halide_mutex *mutex)
    fast_cond->wait(fast_mutex);
 }
 
+WEAK halide_mutex_array* halide_mutex_array_create(int sz) {
```
abadams: Not for this PR, but if sz is huge, we should probably hash it down to something smaller in the accessors below. Is there any harm in using the same mutex for multiple buckets (e.g. risk of deadlock)?

src/runtime/synchronization_common.h
```diff
@@ -1138,4 +1138,36 @@ WEAK void halide_cond_wait(struct halide_cond *cond, struct halide_mutex *mutex)
...
+    halide_mutex_array *array = (halide_mutex_array*)halide_default_malloc(
+        NULL, sizeof(halide_mutex_array));
+    if (array == NULL) {
+        return NULL;
+    }
+    array->array = (halide_mutex*)halide_default_malloc(
+        NULL, sz * sizeof(halide_mutex));
+    if (array->array == NULL) {
```
abadams: halide_default_free(NULL, NULL) doesn't seem worth calling.

abadams: Also, should probably use halide_malloc/halide_free, not the default implementation. These allocations could be large, and users might want them to come from a custom allocator.

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Done. I uses `expr_uses_vars` with the let binding scope to check it.

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Fixed. This is a bug. Added comments to each class regarding what they do.
The reason why this didn't trigger an error in the tests is because this bug makes all the threads access mutex at location 0, which means it just serializes all the atomic mutex access.

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Done! Can't believe I miss this (probably because Function doesn't expose it and it's only in Func).

src/IR.cpp
BachiLi: Done.

src/IRMutator.cpp
BachiLi: Done.

src/runtime/fake_thread_pool.cpp
BachiLi: Added a comment to avoid confusion from future readers.

src/runtime/synchronization_common.h
BachiLi: I don't think there's a risk of deadlock but I added some comments based on your comments.

src/runtime/synchronization_common.h
BachiLi: Done. I meant to call halide_free(NULL, array) so that was a bug.

BachiLi:
Did another around and addressed all comments. PTAL when you have time.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,340 @@
...
+    void visit(const Store *op) override {
+        // Ideally we want to insert equal() checks here for different stores,
+        // but the indices of them actually are different in the case of tuples,
+        // since they usually refer to the strides/min/extents of their own tuple
+        // buffers. However, different elements in a tuple would have the same
+        // strides/min/extents so we are fine.
+        if (index.defined()) {
+            return;
+        }
+        index = op->index;
```
abadams: This lifts the index out of any containing LetStmts which may define variables it relies on. Add a LetStmt visitor:

```
void visit(const LetStmt *op) override {
  IRGraphVisitor::visit(op);
  if (index.defined()) {
    if (expr_uses_var(index, op->name)) {
      index = Let::make(op->name, op->value, index);
    }
  }
}
```

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,340 @@
...
+        // Lock the mutexes using the indices from the Store nodes inside the body.
+        FindStoreIndex find;
+        op->body.accept(&find);
+
+        Expr index = find.index;
+        if (!index.defined()) {
+            // scalar output
+            index = Expr(0);
+        }
+        Stmt body = op->body;
```
abadams: Consider:
```

    f(x) = 0;
    f(clamp(f(r), 0, 100)) = f(r) + 1;

```
You need a check:
```
user_assert(!expr_uses_var(index, name of the buffer)) 
    << "Can't use atomic() on an update where the index written to depends on the current value of the Func";
```

src/Simplify_Stmts.cpp
```diff
@@ -501,5 +501,22 @@ Stmt Simplify::visit(const Fork *op) {
     }
 }
 
+Stmt Simplify::visit(const Atomic *op) {
+    bool changed = false;
```
abadams: Don't need the changed variable, as in the simplification to IRMutator.cpp

src/AddAtomicMutex.cpp
BachiLi: Done.

src/AddAtomicMutex.cpp
BachiLi: Done. I did this in a separate pass (see CheckAtomicValidity) since it also applies to atomic stores. Also added a test in error.

src/Simplify_Stmts.cpp
BachiLi: Done.

BachiLi:
Did a pass.

src/runtime/HalideRuntime.h
```diff
@@ -110,10 +110,20 @@ extern void halide_mutex_lock(struct halide_mutex *mutex);
 extern void halide_mutex_unlock(struct halide_mutex *mutex);
 extern void halide_cond_signal(struct halide_cond *cond);
 extern void halide_cond_broadcast(struct halide_cond *cond);
-extern void halide_cond_signal(struct halide_cond *cond);
 extern void halide_cond_wait(struct halide_cond *cond, struct halide_mutex *mutex);
 //@}
 
+/** Functions for constructing/destroying/locking/unlocking arrays of mutexes. */
+struct halide_mutex_array {
+    struct halide_mutex *array;
```
abadams: Does this struct need to be defined here? Would be nice if it were opaque to make the data type abstract.

i.e.
```struct halide_mutex_array;```

and then put the actual definition alongside the implementation of the methods below. This would let us change the implementation of the mutex array depending on the target.

Comment (BachiLi) [reply src/runtime/HalideRuntime.h]: Done. Ran into some stupid C/C++ error when I defined an empty struct in fake pool thread.

**Comment by BachiLi** [comment]
Sorry, I might have broke the linux-64 buildbot since it's running out of disk space. I'll try not to push too frequently in the future.

**Comment by abadams** [comment]
The buildbots are configured to only take the latest commit to test whenever they're free, so don't worry too much about pushing a lot. Running out of space on the linux bots is a configuration issue I think, not a usage issue. I'm not sure why it happens - we shouldn't be saving state from old test runs. They possibly just need larger drives.

File: src/runtime/HalideRuntime.h [hunk 45]
```diff
@@ -114,9 +114,7 @@ extern void halide_cond_wait(struct halide_cond *cond, struct halide_mutex *mute
 //@}
 
 /** Functions for constructing/destroying/locking/unlocking arrays of mutexes. */
-struct halide_mutex_array {
-    struct halide_mutex *array;
-};
+struct halide_mutex_array;
```
abadams: Nice, now we have license to be clever in the implementation in future.

abadams:
I have been testing out a new clang-format file and decided to try it on this branch. Is it cool if I push a formatting change directly to this branch? I don't want to cause a merge conflict with any local changes you haven't pushed yet.

BachiLi:
Yes please test it. I don't have anything local right now.

abadams:
At HL_NUM_THREADS=2, on my machine, the atomics test takes 36 seconds. This is reasonable.

At HL_NUM_THREADS=4, it takes 96 seconds.

At HL_NUM_THREADS=8, it takes 339 seconds

At HL_NUM_THREADS=16, it takes 2831 seconds (47 minutes!)

At HL_NUM_THREADS=32 (native for my machine), I didn't even try.

The increase in runtime with thread count seems fairly uniformly distributed across the tests. So this test is unrunnable on large machines.

I guess this is just due to contention, because mostly the test does parallel reductions into things of size 7 or 13 or 1, so these are just poor schedules for machines with lots of cores. We want lots of contention for the tests to uncover races, but we don't want the runtime to scale with the number of cores on the test machine. I guess I'll just limit the number of threads Halide uses.

abadams:
pinging @steven-johnson regarding linux-worker-3 being out of disk space

test/correctness/atomics.cpp
```diff
@@ -0,0 +1,1223 @@
...
+        Buffer<int> out0 = out[0];
+        Buffer<int> out1 = out[1];
+        for (int i = 0; i < 2 * img_size; i++) {
+            check(__LINE__, out0(i), correct_consumer1(i));
+            check(__LINE__, out1(i), correct_consumer1(i));
+        }
+    }
+}
+
+int main(int argc, char **argv) {
```
steven-johnson: This test seems large. How long is the runtime? (Wondering if it would be sensible to split it into several tests)

test/correctness/atomics.cpp
```diff
@@ -0,0 +1,1223 @@
...
+    Target target = get_jit_target_from_environment();
+    // Most of the schedules used in this test are terrible for large
+    // thread count machines, due to massive amounts of
+    // contention. We'll just set the thread count to 4. Unfortunately
+    // there's no JIT api for this yet.
+    #ifdef _WIN32
+    _putenv_s("HL_NUM_THREADS", "4");
```
steven-johnson: Does `halide_set_num_threads()` not work for the JIT? If not, that's a major oversight. (Could you open an issue on it?)

python_bindings/correctness/atomics.py
```diff
@@ -0,0 +1,22 @@
...
+
+    ref = [0, 0, 0, 0, 0]
+    for i in range(100):
+        idx = (i * i) % 5
+        ref[idx] += 1
+    for i in range(5):
+        assert(b[i] == ref[i])
+
+if __name__ == "__main__":
+    test_atomics()
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,395 @@
...
+namespace Halide {
+namespace Internal {
+
+using std::map;
+using std::set;
+using std::string;
+
+namespace {
+
+/** Collect all store's names inside a statement. */
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,395 @@
...
+    // The set of producers that have allocated a mutex buffer
+    set<string> allocated_mutexes;
+
+    Stmt allocate_mutex(const string &mutex_name, Expr extent, Stmt body) {
+        Expr mutex_array = Call::make(type_of<halide_mutex_array *>(),
+                                      "halide_mutex_array_create",
+                                      {extent},
+                                      Call::Extern);
+        // Allocate a scalar of halide_mutex_array.
+        // This generate halide_mutex_array mutex[1];
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/AddAtomicMutex.h
```diff
@@ -0,0 +1,27 @@
...
+#define HALIDE_ADD_ATOMIC_MUTEX_H
+
+#include "Expr.h"
+#include "Function.h"
+#include <map>
+
+/** \file
+ * Defines the lowering pass that insert mutex allocation code & locks
+ * for the atomic nodes that require mutex locks. It also checks whether
+ * the atomic operation is valid. It rejects algorithms that has indexing
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/CodeGen_C.cpp
```diff
@@ -2378,6 +2379,18 @@ void CodeGen_C::visit(const Store *op) {
...
+
+    if (inside_atomic_mutex_node) {
+        user_assert(t.is_scalar())
+            << "The vectorized atomic operation for the store" << op->name
+            << " is lowered into a mutex lock, which does not support vectorization.\n";
+    }
+
+    // Issue atomic store if we are in the designated producer.
+    if (emit_atomic_stores) {
+        stream << "#pragma omp atomic\n";
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/CodeGen_LLVM.cpp
```diff
@@ -2457,6 +2461,136 @@ void CodeGen_LLVM::codegen_predicated_vector_load(const Load *op) {
     }
 }
 
+void CodeGen_LLVM::codegen_atomic_store(const Store *op) {
+    // TODO: predicated store
```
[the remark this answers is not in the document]
BachiLi: Done.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -263,6 +264,41 @@ void CodeGen_PTX_Dev::visit(const Load *op) {
...
 
 void CodeGen_PTX_Dev::visit(const Store *op) {
+    // Issue atomic store if we are inside an Atomic node.
+    if (emit_atomic_stores) {
+        user_assert(is_one(op->predicate)) << "Atomic update does not support predicated store.\n";
+        user_assert(op->value.type().bits() >= 32) << "CUDA: 8-bit or 16-bit atomics are not supported.\n";
+#if LLVM_VERSION < 90
+        user_assert(op->value.type().is_scalar())
+            << "CUDA atomic update does not support vectorization with LLVM version < 9.\n";
+        // Generate nnvm intrinsics for the atomics if this is an float atomicAdd.
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/Func.h
```diff
@@ -1548,6 +1549,56 @@ class Func {
...
+    /** Issue atomic updates for this Func. This allows parallelization
+     * on associative RVars. The function throws a compile error when
+     * Halide fails to prove associativity. Use override_associativity_test
+     * to disable the associativity test if you believe the function is
+     * associative or the order of reduction variable execution does not
+     * matter. 
+     * Halide compiles this into hardware atomic operations whenever possible, 
+     * and falls back to a mutex lock per storage element if it is impossible 
+     * to atomically update.
+     * There are three possible outcome of the compiled code:
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/Func.h
```diff
@@ -1548,6 +1549,56 @@ class Func {
...
+     * Expr new_max   = max(im(r), old_max);
+     * arg_max() = {new_index, new_max};
+     * arg_max.compute_root();
+     * arg_max.update().atomic().parallel();
+     *
+     * will be compiled to updates guarded by a mutex lock,
+     * since it is impossible to atomically update two different locations.
+     *
+     * Currently the atomic operation is supported by x86, CUDA, and OpenCL backends.
+     * Compiling to other backends result in a compile error.
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/Func.h
```diff
@@ -1548,6 +1549,56 @@ class Func {
...
+     * If an operation is compiled into a mutex lock, and is vectorized or is
+     * compiled to CUDA or OpenCL, it also results in a compile error,
+     * since per-element mutex lock on vectorized operation leads to a
+     * deadlock.
+     * Vectorization of predicated RVars (through rdom.where()) on CPU
+     * is also unsupported yet.
```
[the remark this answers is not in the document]
BachiLi: Done.

src/Schedule.h
```diff
@@ -573,6 +573,21 @@ class StageSchedule {
     bool &allow_race_conditions();
     // @}
 
+    /** Use atomic update? */
+    // @{
+    bool atomic() const;
+    bool &atomic();
+    // @}
+
+    /** Atomic updates are only allowed on associative reduction.
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/Schedule.h
```diff
@@ -573,6 +573,21 @@ class StageSchedule {
...
+     *  We try to prove the associativity, but the user can override
+     *  the associativity test and supress compiler error if the prover
```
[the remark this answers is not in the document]
BachiLi: Fixed.

src/runtime/synchronization_common.h
```diff
@@ -1138,4 +1138,43 @@ WEAK void halide_cond_wait(struct halide_cond *cond, struct halide_mutex *mutex)
...
+    struct halide_mutex *array;
+};
+
+WEAK halide_mutex_array* halide_mutex_array_create(int sz) {
+    // TODO: If sz is huge, we should probably hash it down to something smaller 
+    // in the accessors below. Check for deadlocks before doing so.
+    halide_mutex_array *array = (halide_mutex_array*)halide_malloc(
+        NULL, sizeof(halide_mutex_array));
+    if (array == NULL) {
+        return NULL;
```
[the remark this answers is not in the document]
BachiLi: halide_default_malloc in `posix_allocator.cpp` returns NULL and says
`// Will result in a failed assertion and a call to halide_error`
so I did the same here.

test/correctness/atomics.cpp
BachiLi: See Andrew's comment: around a minute.

test/correctness/atomics.cpp
BachiLi: It looks like it should works. @abadams is it fine if I change it to use `halide_set_num_threads`?

abadams: I don't think it works for JIT. It's a runtime function. I went looking for an accessor but could not find it. Should probably be a method on JITSharedRuntime

src/runtime/synchronization_common.h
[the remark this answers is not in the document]
abadams: For that to happen you need to inject an AssertStmt into the IR that the return value of halide_mutex_array_create was not null. Probably a good idea.

test/correctness/atomics.cpp
abadams: I got it to go substantially faster than that by trimming down the number of iterations for the code that tries to make it more likely to detect race conditions.

test/correctness/atomics.cpp
BachiLi: OK. Opened an issue.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,395 @@
...
+    void visit(const Atomic *op) override {
+        // Collect the names of all Store nodes inside.
+        CollectStoreNames collector;
+        op->body.accept(&collector);
+
+        // Find the indices from the Store nodes inside the body.
+        FindStoreIndex find;
+        op->body.accept(&find);
+
+        Expr index = find.index;
```
abadams: If evaluating the index (including the containing lets) has side-effects, this will trigger them twice. Maybe lift the index computation into a letstmt outside of the atomic node.

BachiLi: Fixed.

BachiLi:
Andrew said he was reasonably satisfied this morning. I will merge tomorrow if the tests are passed.

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,431 @@
...
+        if (!index.defined()) {
+            // Scalar output.
+            index = Expr(0);
+        } else {
+            // Lift the index outside of the atomic node.
+            string name = unique_name('t');
+            index_let = index;
+            index = Variable::make(index.type(), name);
+            body = ReplaceStoreIndexWithVar(index).mutate(body);
+        }
```
abadams: All of this code assumes that all stores inside the atomic node are to the Func that the atomic node is guarding. What happens if additional stores get injected here, e.g. due to tracing or profiling or other future instrumentation? Should there be some sort of name check against the name on the atomic node?

src/AddAtomicMutex.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    void visit(const Store *op) override {
+        // Ideally we want to insert equal() checks here for different stores,
+        // but the indices of them actually are different in the case of tuples,
+        // since they usually refer to the strides/min/extents of their own tuple
+        // buffers. However, different elements in a tuple would have the same
+        // strides/min/extents so we are fine.
+        if (index.defined()) {
```
abadams: I think we could check here that the stores are at least to the Func the Atomic node is supposed to be guarding. This would have to be with string matching, unfortunately. ```(op->name == guarded_func_name || starts_with(op->name, guarded_func_name + "."))```

BachiLi: Added the suggested change.

PR #3938, opened 2019-06-10
Typed variables from Exprs

jn80842:
Top-down type inference; guesses integer when type is ambiguous


steven-johnson:
Can you expand a bit more on what this is trying to achieve? 

Where are the test(s)?

abadams:
Note that this is not a pull request into master - it's a thing from a research branch. Annoying that it shows up in the same list as the other pull requests. I'll close it and grab the code myself as needed.

PR #3939, opened 2019-06-10
Increase thread priority whenever halide_hexagon_remote_poll_profiler…

white-pony:
…_state is called

Priority of the working threads (running on Hexagon) was recently increased to 100 (from 255). However, profiler thread which runs separately and requests state updates via RPC had lower priority (192), which led to the inaccurate profiling results. This PR increases priority of the thread to match priority of the working threads, so profiler thread has a better chance to run.


dsharletg:
Can you add to the pull request description a brief explanation what this fixes?

white-pony:
> Can you add to the pull request description a brief explanation what this fixes?

I've updated the PR description.

PR #3941, opened 2019-06-13
Remove incorrect GCD reference from README

jrk:
Delete what I believe is long-outdated and incorrect information.


abadams:
Failure is unrelated trunk llvm build failure.

Issue #3943, opened 2019-06-14
Merging Update Definitions in the Schedule?


Mx7f:
Is there any way to merge update definitions without changing the algorithm code (so just using scheduling constructs)? Put another way, given a 1D `RDom` r, and 2 `Func`s f0,f1, can I convert this code:

```
Halide::Func cost("cost");
cost() = 0.f;
cost() += f0(r.x);
cost() += f1(r.x);
```

To use a single loop for the updates, like the default schedule for:

```
Halide::Func cost("cost");
cost() = 0.f;
cost() += f0(r.x) + f1(r.x);
```

abadams:
Interesting question. This transformation requires commutativity/associativity of the reduction, and the only transformation we have that exploits those is rfactor, so if such a schedule existed it would have to exploit rfactor somehow. This would be a degenerate rfactor because you don't actually want to factor the reduction....

Ok, so you can use an empty rfactor to lift each reduction out into its own Func. Then you can merge the now independent Funcs' loops with compute_with:

```
#include "Halide.h"

using namespace Halide;

int main(int argc, char **argv) {
    Func f0, f1, cost;
    Var x;
    f0(x) = x;
    f1(x) = x;

    RDom r(0, 100);
    cost() = 0.f;
    cost() += f0(r.x);
    cost() += f1(r.x);

    f0.compute_root();
    f1.compute_root();

    // Move the reductions into their own Funcs
    Func tmp1 = cost.update(0).rfactor({});
    Func tmp2 = cost.update(1).rfactor({});

    tmp1.compute_root();
    tmp2.compute_root();

    // Now that they're independent funcs, we can fuse the loops using compute_with

    tmp1.update().compute_with(tmp2.update(), r.x);

    cost.realize();

    return 0;
}

```

This produces the IR:

```
  allocate cost_intm$1[float32 * 1]
  allocate cost_intm[float32 * 1]
  produce cost_intm {
    produce cost_intm$1 {
      cost_intm$1[0] = 0.000000f
      consume f1$0 {
        consume f0$0 {
          for (cost_intm$1.s1.r$x, 0, 100) {
            cost_intm$1[0] = (cost_intm$1[0] + float32(f1$0[cost_intm$1.s1.r$x]))
            cost_intm[0] = (cost_intm[0] + float32(f0$0[cost_intm$1.s1.r$x]))
          }
        }
      }
      free f0$0
      free f1$0
      cost_intm[0] = 0.000000f
    }
  }
  produce cost {
[...]

PR #3947, opened 2019-06-14
Fix floated pure stage

alexreinking:
Issue #3943 discovered that pure definitions for fused groups could be injected after an update definitions. This change ensures that all the pure definitions are injected before the `.update(0)` definitions, which are injected before the `.update(1)` definitions and so on.

I'm not sure this is actually right, but the tests are passing. The main issue is that this is now the first and only test case for compute_with behavior when applied to an update stage, rather than to a whole Func.

### DIFF
--- a/src/RealizationOrder.cpp
+++ b/src/RealizationOrder.cpp
@@ -18,14 +18,14 @@ using std::vector;
 
 namespace {
 
-void find_fused_groups_dfs(string current,
+void find_fused_groups_dfs(const string& current,
                            const map<string, set<string>> &fuse_adjacency_list,
                            set<string> &visited,
                            vector<string> &group) {
     visited.insert(current);
     group.push_back(current);
 
-    map<string, set<string>>::const_iterator iter = fuse_adjacency_list.find(current);
+    auto iter = fuse_adjacency_list.find(current);
     internal_assert(iter != fuse_adjacency_list.end());
 
     for (const string &fn : iter->second) {
@@ -59,7 +59,7 @@ find_fused_groups(const map<string, Function> &env,
     return {fused_groups, group_name};
 }
 
-void realization_order_dfs(string current,
+void realization_order_dfs(const string& current,
                            const map<string, vector<string>> &graph,
                            set<string> &visited,
                            set<string> &result_set,
@@ -210,7 +210,7 @@ pair<vector<string>, vector<vector<string>>> realization_order(
 
     // Collect all indirect calls made by all the functions in "env".
     map<string, map<string, Function>> indirect_calls;
-    for (const pair<string, Function> &caller : env) {
+    for (const pair<const string, Function> &caller : env) {
         map<string, Function> more_funcs = find_transitive_calls(caller.second);
         indirect_calls.emplace(caller.first, more_funcs);
     }
@@ -225,7 +225,7 @@ pair<vector<string>, vector<vector<string>>> realization_order(
     map<string, vector<FusedPair>> fused_pairs_graph;
     map<string, set<string>> fuse_adjacency_list;
 
-    for (const pair<string, Function> &caller : env) {
+    for (const pair<const string, Function> &caller : env) {
         // Find all compute_with (fused) pairs. We have to look at the update
         // definitions as well since compute_with is defined per definition (stage).
         vector<FusedPair> &func_fused_pairs = fused_pairs_graph[caller.first];
@@ -256,7 +256,7 @@ pair<vector<string>, vector<vector<string>>> realization_order(
     std::tie(fused_groups, group_name) = find_fused_groups(env, fuse_adjacency_list);
 
     // Compute the DAG representing the pipeline
-    for (const pair<string, Function> &caller : env) {
+    for (const pair<const string, Function> &caller : env) {
         const string &caller_rename = group_name.at(caller.first);
         // Create a dummy node representing the fused group and add input edge
         // dependencies from the nodes representing member of the fused group
@@ -265,7 +265,7 @@ pair<vector<string>, vector<vector<string>>> realization_order(
         // Direct the calls to calls from the dummy node. This forces all the
         // functions called by members of the fused group to be realized first.
         vector<string> &s = graph[caller_rename];
-        for (const pair<string, Function> &callee : find_direct_calls(caller.second)) {
+        for (const pair<const string, Function> &callee : find_direct_calls(caller.second)) {
             if ((callee.first != caller.first) && // Skip calls to itself (i.e. update stages)
                 (std::find(s.begin(), s.end(), callee.first) == s.end())) {
                 s.push_back(callee.first);
@@ -278,7 +278,7 @@ pair<vector<string>, vector<vector<string>>> realization_order(
     vector<string> temp;
     set<string> result_set;
     set<string> visited;
-    for (Function f : outputs) {
+    for (const Function& f : outputs) {
         if (visited.find(f.name()) == visited.end()) {
             realization_order_dfs(f.name(), graph, visited, result_set, temp);
         }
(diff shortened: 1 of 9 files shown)


src/RealizationOrder.cpp
```diff
@@ -18,14 +18,14 @@ using std::vector;
 
 namespace {
 
-void find_fused_groups_dfs(string current,
+void find_fused_groups_dfs(const string& current,
```
abadams: nit: we put the ampersand after the space

test/correctness/compute_with_bug.cpp
```diff
@@ -0,0 +1,34 @@
...
+    cost_intm.compute_root();
+    cost_intm_1.compute_root();
+
+    // Now that they're independent funcs, we can fuse the loops using compute_with
+    cost_intm.update().compute_with(cost_intm_1.update(), r.x);
+
+    Buffer<float> result = cost.realize();
+    const float expected_result = 9900.f; // 2 * sum(0..99)
+
+    return result(0) == expected_result ? EXIT_SUCCESS : EXIT_FAILURE;
```
abadams: It should print something rather than just returning a non-zero exit status (Makes it easier to figure out what's going on in the logs when running lots of tests). Also by convention we print "Success!" when a test succeeds :)

src/ScheduleFunctions.cpp
```diff
@@ -1381,12 +1381,21 @@ class InjectFunctionRealization : public IRMutator {
                                                                   replacements, add_lets);
                 producer = inject_stmt(producer, produceDef, f.definition().schedule().fuse_level().level);
             }
+        }
+
+        bool some_updated = false;
+        for (size_t j = 0; j == 0 || some_updated; j++) {
```
abadams: Minor suggestions: Why not just initialize some_updated to true and simplify the loop condition?

src/ScheduleFunctions.cpp
```diff
@@ -1381,12 +1381,21 @@ class InjectFunctionRealization : public IRMutator {
...
+            some_updated = false;
+            for (auto iter = funcs.rbegin(); iter != funcs.rend(); iter++) {
+                const auto &f = *iter;
 
-            for (size_t j = 0; j < f.updates().size(); ++j) {
-                string defPrefix = f.name() + ".s" + std::to_string(j + 1) + ".";
-                const Definition &def = f.updates()[j];
-                const Stmt &updateDef = build_produce_definition(f, defPrefix, def, true, replacements, add_lets);
-                producer = inject_stmt(producer, updateDef, def.schedule().fuse_level().level);
+                if (j < f.updates().size()) {
```
abadams: So does this interleave the update definitions of the compute_with things for all the stages that aren't actually compute_with each other? Worth a comment explaining the strategy.

abadams:
Thanks for doing this. Here's a case to test: f and g both have ~10 update definitions, and you say: f.update(8).compute_with(g.update(1), x);

src/ScheduleFunctions.cpp
alexreinking: I think it will, yeah. There might be an alternative fix in RealizationOrder that adds an appropriate edge. But it wasn't obvious to me how to do that (whereas this was quick), and I'm pretty sure this interleaving preserves the actual values.

It would be good to really write down what compute_with is supposed to do / guarantee. And then write some tests for that.

test/correctness/compute_with_bug.cpp
alexreinking: Done! I'm having it print "Success!" as you suggested, and a message with the actual and expected values to stderr on failure.

alexreinking:
> Thanks for doing this. Here's a case to test: f and g both have ~10 update definitions, and you say: f.update(8).compute_with(g.update(1), x);

That was a good test. The current "fix" is still wrong. It's moving f.update(8) too early. This is probably going to need changes in RealizationOrder, too. Gonna take a bit more work, but I have some ideas.

alexreinking:
Closing this for now so the buildbots don't freak out while I work on this.

alexreinking:
# Status Update

I'm shelving this to focus on my PLDI submission, but I wanted to jot down the progress I made working on this...

## The Problem

The issue with `compute_with` right now is that it makes the incorrect assumption that there is a total order of funcs within a fused group that allows lowering to proceed by lowering each function individually, and fusing loops for stages as they are encountered. There is also an ill-defined convention of identifying a "parent" of each fused group that has to come first in its fused stages.

A failing example is the following. Consider 2D Funcs `f` and `g`, each with two update stages. Fuse `f`'s update stage into `g`'s pure stage, making `g` the parent, and thus lowered first. 

```
f.compute_root();
g.compute_root();
f.update(0).compute_with(g, y);
```

Then the existing ScheduleFunctions sees a fused group [g, f] and proceeds to lower g:

```
for g.y:
  for g.x:
    g(x, y) = ...  # pure
for g.y:
  for g.x:
    g(x, y) = ...  # update(0)
```

Then f gets lowered stage by stage. First goes the pure stage:

```
for g.y:
  for g.x:
    g(x, y) = ...  # pure
for g.y:
  for g.x:
    g(x, y) = ...  # update(0)
for f.y:
  for f.x:
    f(x, y) = ... # pure
```

And now ScheduleFunctions sees that `f`'s update is fused to `g`'s pure definition and spits this out:

```
for fused.y:
  for g.x:
    g(x, y) = ...  # pure
  for f.x:
    f(x, y) = ... # update(0)
for g.y:
  for g.x:
    g(x, y) = ...  # update(0)
for f.y:
  for f.x:
    f(x, y) = ... # pure
```

Oops! The whole algorithm is unfortunately compromised thanks to this and there's no simple fix since the basic assumptions are flawed.

## A Solution

Instead, I propose the following scheduling algorithm:

1. Identify Fused Groups of Funcs in the environment and arrange them into a producer-consumer DAG. This is the _Pipeline Graph_.
2. Within each Fused Group, break apart the functions into their stages and build Fused Stages from the compute_with directives. Validate that there are no P/C dependencies within a Fused Stage and arrange those Fused Stages into a DAG over the stage-order edges (ie. pure def comes before update(0), comes before update(1), etc.)
3. Recognize that the Fused Stages are each a tree of normal stages with edges labeled by the loop level (var) at which they are fused.

Thus we must build a three-level deep data structure.

This is greatly accelerated by using a [union-find / disjoint-set forest data structure](https://en.wikipedia.org/wiki/Disjoint-set_data_structure). I believe there's one in the PR already that does path halving and union by size.

To determine the fuse level of two stages in a Fused Stage, find their LCA and walk up the tree, picking the rightmost edge label. This can be accelerated for many pairs using [Tarjan's offline LCA algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_off-line_lowest_common_ancestors_algorithm) but there might be a better way of incrementally tracking this information.

Lowering should then work broadly as before, but will need to lower entire Fused Stages at a time within each Fused Group. Bounds inference will need to use the above method of computing fusion levels. I believe bounds inference is still sound with this interpretation of compute_with.

## Another Perspective

When scheduling, we might wish to build up abstract Loop Nest objects directly via the scheduling directives. We could form a sort of algebra where compute_with, split, reorder, etc. modify loop nests, and there is a many-to-one relationship between loop nests and the stages they compute. Compute_at / store_at would refer to a particular loop nest and var. Parallel/vectorized, etc. would annotate the loops in a nest. Then we could lower much more quickly by validating the schedule and stitching everything together by walking a DAG with fused stages at the top level. A prioritized topological sort could (probably, I haven't proved this yet) keep update stages close to their predecessors.

## Temporary Measures

The original scheduling algorithm works for funcs that are only fused at their pure definitions and it seems that fusing update stages is not common. None of our tests currently cover the case of fusing update stages, so maybe we should just disallow it? Any program that a user currently has that uses compute_with either doesn't actually work or they lucked out.

PR #3960, opened 2019-06-25
adds labels to command buffers

mikewoodworth:
Sets the label string ivar on metal command buffers to:
- the calling halide function name for blit commands
- the entry_name arg for compute commands created from halide_metal_run

These labels can be seen at debug time in Apple's Xcode and Instrument tools.


zvookin:
Have we verified there's no performance impact here? @shoaibkamil, any concerns?

shoaibkamil:
LGTM, no concerns here. 

A higher-level question is whether we should modify the Metal codegen to produce more meaningful names for the kernel functions (eg by prefixing the Func name).

mikewoodworth:
The func names are human readable, but not pretty (at least for JITed funcs, haven't looked at AOT)

PR #3961, opened 2019-06-25
Marking IntrusivePtr move copy/assign as noexcept

alexreinking:
This is a nasty C++ subtlety... custom move and copy constructors aren't `noexcept` without explicitly saying so. This means that inferred move constructors aren't `noexcept` either. Vectors of such classes will silently _copy_ upon expansion or transfer.

We should look around to see where else there might be similar issues.

### DIFF
--- a/src/IntrusivePtr.h
+++ b/src/IntrusivePtr.h
@@ -118,7 +118,7 @@ public:
     }
 
     HALIDE_ALWAYS_INLINE
-    IntrusivePtr(IntrusivePtr<T> &&other) noexcept: ptr(other.ptr) {
+    IntrusivePtr(IntrusivePtr<T> &&other) noexcept : ptr(other.ptr) {
         other.ptr = nullptr;
     }
 
--- a/src/IntrusivePtr.h
+++ b/src/IntrusivePtr.h
@@ -118,7 +118,7 @@ public:
     }
 
     HALIDE_ALWAYS_INLINE
-    IntrusivePtr(IntrusivePtr<T> &&other) : ptr(other.ptr) {
+    IntrusivePtr(IntrusivePtr<T> &&other) noexcept: ptr(other.ptr) {
         other.ptr = nullptr;
     }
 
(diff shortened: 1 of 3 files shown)


src/IntrusivePtr.h
```diff
@@ -118,7 +118,7 @@ struct IntrusivePtr {
     }
 
     HALIDE_ALWAYS_INLINE
-    IntrusivePtr(IntrusivePtr<T> &&other) : ptr(other.ptr) {
+    IntrusivePtr(IntrusivePtr<T> &&other) noexcept: ptr(other.ptr) {
```
abadams: micronit: space between noexcept and colon

alexreinking: I'm looking around for more noexcept move constructors / assigns. I'll put in the space when I fix up those.

alexreinking:
mingw failure is unrelated: simd_op_check

BachiLi:
I got the following compile error on MacOS (using Apple LLVM version 10.0.1) after this patch:

Halide/src/Util.h:326:5: error: exception specification of explicitly defaulted move
      constructor does not match the calculated one
    ScopedValue(ScopedValue&& that) noexcept = default;
    ^
/Users/bachi/Programs/Halide/src/EarlyFree.cpp:94:31: note: in instantiation of template class
      'Halide::Internal::ScopedValue<Halide::Internal::Stmt>' requested here
            ScopedValue<Stmt> old_containing_stmt(containing_stmt, block->first);

abadams:
Seems like it's related to standard versions:
https://stackoverflow.com/questions/29483120/program-with-noexcept-constructor-accepted-by-gcc-rejected-by-clang

We could remove the noexcept from ScopedBinding, because it requires
noexcept for the underlying type, which isn't always the case. But it's
nice to get the error in cases where the underlying type should be noexcept
but isn't.

Hrm.

Thoughts @alexreinking ?

abadams:
In this case I believe it's because Stmt/Expr have default constructors that should be noexcept but aren't. If you fix that, then you get an error for `ScopedBinding<string>`, which is clearly out of our control.

abadams:
I'm going to open a PR that removes noexcept from ScopedValue, but also defaults a bunch of currently-explicitly defined default constructors so that the compiler can infer noexcept for itself.

Issue #3955, opened 2019-06-24
AOT calls will accept/write to Buffer<const T>, but shouldn't


steven-johnson:
H::R::B provides `operator halide_buffer_t*()` to make it easy to pass as arguments to AOT code, but this means you can inadvertently pass a `Buffer<const T>` as an output buffer arg, which is a mistake that should be preventable. 

Random thoughts:

- We could add overloads to the AOT calls that handle HRB as well as halide_buffer_t (and fail for const outputs); basically a much simpler version of https://github.com/halide/Halide/pull/3206. As a bonus this would add some static type checking. Drawback is that this would require HalideBuffer.h to be included before the generated header (or make the calls guarded with ifdef).

- To avoid that we could define two new types that are used in the aot-generated code declarations (`halide_buffer_input` and `halide_buffer_output` or whatnot); halide_buffer_t* would autoconvert to either, and HRB would get new overloads to return these, with appropriate static_assert failures for the const -> output conversion.

- Or we could leave as-is and decide it's not worth the trouble.

steven-johnson:
I resurrected & simplified https://github.com/halide/Halide/pull/3255 to see how it would look. Generated header basically adds:

```
// C++ wrappers to accept Halide::Runtime::Buffers of the appropriate types
#if defined(HALIDE_RUNTIME_BUFFER_WRAPPERS)
template<
  typename T__input1_buffer, 
  typename T__input2_buffer, 
  typename T__scalar, 
  typename T__output_buffer>
inline int constinput(
  ::Halide::Runtime::Buffer<T__input1_buffer> &_input1_buffer, 
  ::Halide::Runtime::Buffer<T__input2_buffer> &_input2_buffer, 
  int32_t _scalar, 
  ::Halide::Runtime::Buffer<T__output_buffer> &_output_buffer) HALIDE_FUNCTION_ATTRS {
    return constinput(
        _input1_buffer.template as<const int32_t>().raw_buffer(), 
        _input2_buffer.template as<const int32_t>().raw_buffer(), 
        _scalar, 
        _output_buffer.template as<int32_t>().raw_buffer());
}
#endif  // defined(HALIDE_RUNTIME_BUFFER_WRAPPERS)

```

PR #3255, opened 2018-08-31
Add optional wrappers (V2) to AOT headers to take Halide::Runtime::Bu…

steven-johnson:
…ffers as arguments

This is a reworked variant of #3206 that dispenses with the Google-style convention and requires all Buffer arguments to be mutable, whether by ref or ptr. Note also that (unlike the previous patch) it disables the implicit `operator halide_buffer_t*` provided by the `Buffer` class, in an effort to ensure that it's harder to pass a wrong-static-type Buffer to a given function (and, of course, with the idea that implicit conversions are always preferable to avoid when not absolutely necessary).

Not sure if this is ready or desirable to land, but putting it out there for discussion purposes.

PR #3962, opened 2019-06-26
Fix Issue #3955

steven-johnson:
This is a simpler alternative to https://github.com/halide/Halide/pull/3255; it doesn't support non-Buffer classes quite as easily, but is likely to be much easier to apply to existing code.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -46,11 +46,12 @@ struct halide_typed_buffer_t {
     assert(p->type == halide_type_of<not_void_T>());
   }
 
-  HALIDE_ALWAYS_INLINE halide_buffer_t *raw_buffer() { return ptr; }
-
-private:
   halide_buffer_t * const ptr;
 };
+
+template<typename T>
+HALIDE_ALWAYS_INLINE
+halide_buffer_t *halide_typed_buffer_t_to_halide_buffer_t(const halide_typed_buffer_t<T> &t) { return t.ptr; }
 
 #endif  // HALIDE_RUNTIME_BUFFER_WRAPPERS
 
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -34,6 +34,27 @@
 #pragma GCC diagnostic ignored "-Warray-bounds"
 #endif
 
+#if defined(HALIDE_RUNTIME_BUFFER_WRAPPERS)
+
+template<typename T>
+struct halide_typed_buffer_t {
+  HALIDE_ALWAYS_INLINE explicit halide_typed_buffer_t(halide_buffer_t *p) : ptr(p) {
+    using not_const_T = typename std::remove_const<T>::type;
+    constexpr bool T_is_void = std::is_same<not_const_T, void>::value;
+    static_assert(!T_is_void, "Cannot create halide_typed_buffer_t<void>");
+    using not_void_T = typename std::conditional<T_is_void, uint8_t, not_const_T>::type;
+    assert(p->type == halide_type_of<not_void_T>());
+  }
+
+  HALIDE_ALWAYS_INLINE halide_buffer_t *raw_buffer() { return ptr; }
+
+private:
+  halide_buffer_t * const ptr;
+};
+
+#endif  // HALIDE_RUNTIME_BUFFER_WRAPPERS
+
+
 namespace Halide {
 namespace Runtime {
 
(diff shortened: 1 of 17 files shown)


steven-johnson:
PTAL

src/runtime/HalideBuffer.h
```diff
@@ -34,6 +34,28 @@
...
+    using not_void_T = typename std::conditional<T_is_void, uint8_t, not_const_T>::type;
+    assert(p->type == halide_type_of<not_void_T>());
+  }
+
+  halide_buffer_t * const ptr;
+};
+
+template<typename T>
+HALIDE_ALWAYS_INLINE
+halide_buffer_t *halide_typed_buffer_t_to_halide_buffer_t(const halide_typed_buffer_t<T> &t) { return t.ptr; }
```
abadams: What are the advantages of this intermediate type over just a free function that takes whatever Buffer type the user has provided?

steven-johnson: If we have a free function, what will the type of the argument be?

- If it's `BufType&`, the arg has to be an lvalue, so you can't pass temporaries.
- If it's `const BufType&`, then we have to do a `const_cast` internally, and I'm convinced that every time you do a `const_cast`, God kills a kitten.
- if it's just plain `BufType`, you may end up with unnecessary copies (if the value isn't provably movable by the compiler).

abadams: I tried `BufType &&` but I can't get the return type to work. A temporary object seems to be necessary.

test/correctness/halide_buffer.cpp
```diff
@@ -75,7 +77,60 @@ void test_copy(Buffer<float> a, Buffer<float> b) {
     check_equal(a_window, b_window);
 }
 
+// // Fake "buffer" class that is shaped enough like Halide::Runtime::Buffer
```
abadams: Did you intend to come back to this commented out code?

src/runtime/HalideBuffer.h
steven-johnson: There may well be a slicker way to do it, but until/unless we find someone who really needs the extra flexibility, this works.

test/correctness/halide_buffer.cpp
steven-johnson: Oops. Nope. Scalpel left in patient.

src/runtime/HalideBuffer.h
```diff
@@ -1069,11 +1091,25 @@ class Buffer {
...
+    // Only Buffer<non-const-T> can produce halide_typed_buffer_t<non-const-T>
+    operator halide_typed_buffer_t<typename std::remove_const<T>::type> () {
+        static_assert(!std::is_const<T>::value, "Can't convert from a Buffer<const T> to a Buffer<T>");
+        return halide_typed_buffer_t<typename std::remove_const<T>::type>(raw_buffer());
+    }
+
+#else
     /** Provide a cast operator to halide_buffer_t *, so that
      * instances can be passed directly to Halide filters. */
     operator halide_buffer_t *() {
```
steven-johnson: Per comment from @abadams in https://github.com/halide/Halide/issues/3963, should we mark `operator halide_buffer_t *` with HALIDE_ATTRIBUTE_DEPRECATED here as well?

abadams: Probably should port over all the production code we're aware of first, in case it takes a while to get done.

steven-johnson:
Update: testing this change inside of Google breaks many many things, but mostly because it was formerly legal to do

    Buffer<> b;   // no static type!
    call_halide(b);

whereas this PR requires a correctly-statically-typed-Buffer<Something> at the call site -- which is undoubtedly the correct thing to do in the long run, but it might be simpler to soften this requirement (at least internal to Google) before this can land.

abadams:
Oh yeah, it would be good to be able to pass dynamically typed buffers. There's a runtime type check done inside the kernel, so it's really just our job here to check the constness of the element type (void vs const void). Checking the type at compile-time is a nice-to-have.

zvookin:
Ultimately, it would be really handy for us to have a way to generate type casing filters for some uses. Ideally this would avoid a dynamic check if the buffer is statically typed and do the dynamic check and dispatch to a specific implementation if not.

steven-johnson:
> it would be good to be able to pass dynamically typed buffers

That would doable with this; just call `.as<whatever>()` at the call site.

steven-johnson:
> That would doable with this; just call .as<whatever>() at the call site.

...except of course for `float16`, since we don't currently have a *runtime* type that corresponds to this (`Halide::float16_t` is compile-time, not runtime), and thus you can't make a statically-correct Buffer<> for this :-(

abadams:
I think it's desirable to continue to be able to pass dynamically-typed buffers in general without statically declaring their type at the call-site using as()

steven-johnson:
I disagree, but will go with what the consensus thinks.

steven-johnson:
Another interesting byproduct of this change (in its current state): currently it's easy to mix `Buffer<>` and actual `halide_buffer_t*` in calls; in its current form, this change requires you to use all-of-one or all-of-the-other. Not sure yet if that's a dealbreaker (i.e., if it's ok to require explicit calls to `.raw_buffer()` in the mixed case). Thoughts?

abadams:
Mixing seems weird to me, but I would be loathe to disallow it without seeing why someone wrote something that way.

steven-johnson:
This also turns into a big bag of hurt for a lot of our existing code that was (unintentionally) taking advantage of this bug. This will probably take some hygiene to land.

steven-johnson:
So after looking at more use of this PR in our code... this can be made to work but is going to be a lot of churn. In particular, we have a surprisingly large amount of code that defines function pointers with `halide_buffer_t*` arguments and uses them in bottlenecks in a lot of places; the fact that `Buffer<>` implicitly converts to this makes this work on most places, and adding explicit calls to `.raw_buffer()` adds a lot of verbosity.

Obvious answer #1 is presumably "Use Buffer<>* instead of halide_buffer_t*", and that's not a bad answer in any event, but here's a thought:

- Define `halide_const_buffer_t` as a type in HalideRuntime; it's identical to `halide_buffer_t` except that the host field is const
- keep the implicit `Buffer<T> -> halide_buffer_t*` and add an implicit `Buffer<const T> -> halide_const_buffer_t*`
- modify header file generation to require `halide_const_buffer_t*` for inputs

Pro:  would allow for cleaner migration of existing code since inserting `.raw_buffer()` no longer needed
Con: Doesn't do element-type checking, which is desireable
Con: Actually adds more implicit conversions rather than reducing them

PR #3969, opened 2019-06-27
Use default constructors where appropriate

abadams:
Also made the ref_count accessor no_except so that when it's defined in a different translation unit we don't falsely think incrementing a refcount might throw.

This removed most of the errors resulting from noexcept ScopedValues. The string move constructor is not noexcept though, so I had to remove noexcept from ScopedValue because we used `ScopedValue<string>` in a few places.


steven-johnson:
LGTM

BachiLi:
Compiles fine on my machine.

alexreinking:
LGTM

PR #3970, opened 2019-06-28
Validating a pointer requires having the OpenCL API loaded

abadams:
device_wrap_native, without debug, just sets device field appropriately and returns, so it doesn't need a CL context and it doesn't need to load libopencl. With debug on, it also validates the pointer for you, which still doesn't need a CL context, but which does require calling clGetMemObjectInfo, which may be a null function pointer if this is the first thing you've done with Halide's OpenCL runtime in the process.


abadams:
mingw failure unrelated

PR #3984, opened 2019-07-08
Add basic support for ARM SVE and SVE2 target features.

zvookin:


steven-johnson:
You also need to update halide.cmake and PyEnums.cpp

halide.cmake
```diff
@@ -497,6 +497,8 @@ function(_halide_runtime_target_name HALIDE_TARGET OUTVAR)
         disable_llvm_loop_unroll
         wasm_simd128
         wasm_signext
+	sve
```
steven-johnson: needs one more space

PR #3985, opened 2019-07-08
Allow taking derivative with respect to Param

BachiLi:


src/Derivative.cpp
```diff
@@ -847,6 +847,15 @@ void ReverseAccumulationVisitor::visit(const Variable *op) {
     internal_assert(expr_adjoints.find(op) != expr_adjoints.end());
     Expr adjoint = expr_adjoints[op];
 
+    if (op->param.defined()) {
+        // This is a reference to a Parameter.
+        // Create a call expression to that parameter and propagate the call expression.
+        Expr call = Call::make(op->param, {});
```
abadams: Could you explain the purpose of this call node? I don't think we ever construct call nodes to non-buffer params, so it seems pretty dicey to me. Can it just be passed through as a Variable instead?

BachiLi: This is just for reusing the code in ```ReverseAccumulationVisitor::visit(const Call *op)```.
I can refactor that part as a function and call the function here.

abadams: 👍

BachiLi: Done. We'll see if it passes the test.

BachiLi:
@abadams Please take a look when you have time.

Issue #3972, opened 2019-06-28
CMake/CPack 'distrib' target omits static libraries


steven-johnson:
'make distrib' (using the Makefile) includes both bin/libHalide.so and lib/libHalide.a, whereas 'make distrib' (using CMake) includes only the former, not the latter. This should be fixed so that both build systems produce identical distrib folders.

(See also the ancient PR https://github.com/halide/Halide/pull/2516)

steven-johnson:
https://github.com/halide/Halide/pull/2516 addresses part of this issue (by cleaning up static-library usage in our CMake rules), but is incomplete, as it requires rebuilding the world to switch between static and shared; in theory, I think that it should be simple-ish to produce both at once (by just adding a new `add_library` rule with a different name), but nothing with CMake is ever simple, so that should probably be done as a followup.

alexreinking:
This can be done in CMake by creating two separate targets. Otherwise, we'll have to build twice and set `HALIDE_SHARED_LIBRARY=YES/NO` appropriately. Clients would link to either `Halide::Halide` or `Halide::HalideStatic`

Issue #3971, opened 2019-06-28
CMake/CPack 'distrib' target isn't correct for Windows


steven-johnson:
A while back, someone thoughtfully contributed CPack support to our CMake files, so that `make distrib` worked correctly for CMake as well as make; unfortunately, it needs some extra love for Windows builds in order to match the status quo for Windows:

- Windows requires separate Debug and Release builds; the CMake rules only provide a single build (presumably Release)
- The CMake rules only provide shared-library outputs (bin/Halide.dll), whereas our existing Windows distribs provide Halide.lib as well 

This would be really nice to fix, as currently, Windows distribs are produced by custom code in the buildbots, which can (and does) easily fall out of sync with the rest of Halide.

alexreinking:
This is getting overhauled in #4644 -- I'll need to look into package managers for Windows, eg. vcpkg, chocolatey, Nuget, to see which would be easiest to publish Halide to.

alexreinking:
It turns out that the vcpkg people are currently packaging an old version of Halide. We should work with them as we cut a new release to make sure the new CMake improvements are available to their users.

steven-johnson:
See also https://github.com/halide/Halide/pull/4539

alexreinking:
This can only be reasonably done with CMake 3.16 and only becomes sane in the absolute newest version, 3.17. This is because CPack only gained the ability to produce multi-configuration packages recently. We should just keep CMake on the Windows buildbots as up-to-date as possible and write `packaging/CMakeLists.txt` to detect newer features.

alexreinking:
This will be corrected once CMake 3.18 is released due to a bug in 3.17. We will need to use the `Ninja Multi-Config` generator and then run:

```
> cmake -G "Ninja Multi-Config" ...
> cmake --build . --config Debug
> cmake --build . --config Release
> cpack -C "Debug;Release"
```

See #4980 and [the related issue on CMake](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4825)

steven-johnson:
Hopefully we can continue with our current CMake min of 3.16 and just require 3.18 for building windows distribs?

alexreinking:
Yes, this is about the tool versions, not anything that's in the CMakeLists.txt files.

PR #3986, opened 2019-07-09
 Revamp CMake to build both static and shared libraries

steven-johnson:
Revamp to build both shared and static libraries at the same time; HALIDE_SHARED_LIBRARY now controls which is actually used for building tests and apps. This is (essentially) the fix for https://github.com/halide/Halide/issues/3972, and part of the necessary fix for https://github.com/halide/Halide/issues/3971.


steven-johnson:
This is ancient and CMake is getting major reworking effort; closing.

PR #3987, opened 2019-07-09
Auto detect and set target feature CUDA on Linux

vmiheer:
This replaces default of OpenCl with cuda if cuda found.
I hope this is okay.

### DIFF
--- a/tutorial/lesson_12_using_the_gpu.cpp
+++ b/tutorial/lesson_12_using_the_gpu.cpp
@@ -309,7 +309,19 @@ Target find_gpu_target() {
         target.set_feature(Target::Metal);
     }
 #else
-    if (dlopen("libOpenCL.so", RTLD_LAZY) != NULL) {
+    if (void *handle = dlopen("libcuda.so", RTLD_LAZY)) {
+        int numDevices = 0;
+        int (*cuInit)(int) = (int (*)(int)) dlsym(handle, "cuInit");
+        int (*cuDeviceGetCount)(int *) =
+            (int (*)(int *)) dlsym(handle, "cuDeviceGetCount");
+        if (!dlerror()) {
+            (*cuInit)(0);
+            (*cuDeviceGetCount)(&numDevices);
+            if (numDevices) {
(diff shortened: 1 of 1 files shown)


tutorial/lesson_12_using_the_gpu.cpp
```diff
@@ -309,7 +309,20 @@ Target find_gpu_target() {
         target.set_feature(Target::Metal);
     }
 #else
-    if (dlopen("libOpenCL.so", RTLD_LAZY) != NULL) {
+    if (void *handle = dlopen("libcuda.so", RTLD_LAZY)) {
+        int numDevices = 0;
+        int (*cuDeviceGetCount)(int *);
```
steven-johnson: Declare-and-initialize in one line is more appropriate.

abadams:
This is a tutorial, which means we must minimize the number of lines of code that aren't directly teaching something, especially if they're hard to understand.

So I don't think this is a good change. If people want to use cuda they can uncomment the lines that enable cuda.

tutorial/lesson_12_using_the_gpu.cpp
vmiheer: Fixed. Thank you!

vmiheer:
@abadams, I didn't know that. I though it would benefit people with cuda devices available.
The tutorial otherwise detects nvidia OpenCL but the nvprof doesn't even show the executed kernels (seems like nvidia dropped support for OpenCL in tools.) So I thought the change would be on the lines of how other api's are detected.
I leave the decision of whether to accept this change to you 😃.
I was also wondering existence of binary which sets HALIDE_TARGE/HALIDE_JIT_TARGET and runs the generator process as sub-process OR halide api to auto-populate target properties (refactoring of the code in the tutorial) if there is any value in it.

steven-johnson:
I suspect we should probably close this, but will leave that up to @abadams

abadams:
using host-cuda already sniffs the cuda host capability. I looked into trying to figure out an API to say "Does this machine support this target?", but I didn't get very far. It requires some design thought.

PR #3989, opened 2019-07-10
Don't export Interval::pos_inf/neg_inf

steven-johnson:
Experiment to see if we can remove the need for HALIDE_EXPORT for data members

### DIFF
--- a/src/Interval.cpp
+++ b/src/Interval.cpp
@@ -16,11 +16,13 @@ IRMatcher::WildConst<1> c1;
 Expr make_max_helper(const Expr &a, const Expr &b) {
     auto rewrite = IRMatcher::rewriter(IRMatcher::max(a, b), a.type());
 
+    Expr pos_inf = Interval::pos_inf();
+    Expr neg_inf = Interval::neg_inf();
     if (rewrite(max(x, x), x) ||
-        rewrite(max(x, Interval::pos_inf), Interval::pos_inf) ||
-        rewrite(max(Interval::pos_inf, x), Interval::pos_inf) ||
-        rewrite(max(x, Interval::neg_inf), x) ||
-        rewrite(max(Interval::neg_inf, x), x) ||
+        rewrite(max(x, pos_inf), pos_inf) ||
+        rewrite(max(pos_inf, x), pos_inf) ||
+        rewrite(max(x, neg_inf), x) ||
+        rewrite(max(neg_inf, x), x) ||
         rewrite(max(c0, c1), fold(max(c0, c1))) ||
         rewrite(max(c0, x), max(x, c0)) ||
         rewrite(max(max(x, c0), c1), max(x, fold(max(c0, c1)))) ||
@@ -39,11 +41,13 @@ Expr make_max_helper(const Expr &a, const Expr &b) {
 Expr make_min_helper(const Expr &a, const Expr &b) {
     auto rewrite = IRMatcher::rewriter(IRMatcher::min(a, b), a.type());
 
+    Expr pos_inf = Interval::pos_inf();
+    Expr neg_inf = Interval::neg_inf();
     if (rewrite(min(x, x), x) ||
-        rewrite(min(x, Interval::pos_inf), x) ||
-        rewrite(min(Interval::pos_inf, x), x) ||
-        rewrite(min(x, Interval::neg_inf), Interval::neg_inf) ||
-        rewrite(min(Interval::neg_inf, x), Interval::neg_inf) ||
+        rewrite(min(x, pos_inf), x) ||
+        rewrite(min(pos_inf, x), x) ||
+        rewrite(min(x, neg_inf), neg_inf) ||
+        rewrite(min(neg_inf, x), neg_inf) ||
         rewrite(min(c0, c1), fold(min(c0, c1))) ||
         rewrite(min(c0, x), min(x, c0)) ||
         rewrite(min(min(x, c0), c1), min(x, fold(min(c0, c1)))) ||
@@ -95,8 +99,11 @@ Interval Interval::make_intersection(const Interval &a, const Interval &b) {
 
 // Use Handle types for positive and negative infinity, to prevent
 // accidentally doing arithmetic on them.
-Expr Interval::pos_inf = Variable::make(Handle(), "pos_inf");
-Expr Interval::neg_inf = Variable::make(Handle(), "neg_inf");
+Expr Interval::pos_inf_expr = Variable::make(Handle(), "pos_inf");
+Expr Interval::neg_inf_expr = Variable::make(Handle(), "neg_inf");
+
+Expr Interval::pos_inf_noinline() { return Interval::pos_inf_expr; }
+Expr Interval::neg_inf_noinline() { return Interval::neg_inf_expr; }
 
 }  // namespace Internal
 }  // namespace Halide
--- a/src/Interval.h
+++ b/src/Interval.h
@@ -6,6 +6,7 @@
  */
 
 #include "Expr.h"
+#include "Util.h"
 
 namespace Halide {
 namespace Internal {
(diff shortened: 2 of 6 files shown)


abadams:
+1

src/Interval.cpp
```diff
@@ -98,5 +98,8 @@ Interval Interval::make_intersection(const Interval &a, const Interval &b) {
 Expr Interval::pos_inf = Variable::make(Handle(), "pos_inf");
 Expr Interval::neg_inf = Variable::make(Handle(), "neg_inf");
 
+/*static*/ Expr Interval::get_pos_inf() { return pos_inf; }
```
zvookin: I really don't like the commented out static as it is ambiguous as to whether it is an annotation or commented out syntax. If you want to add a useful comment like ```/* Static getter methods to avoid exporting constants. */``` above the two functions, that seems more explanatory.

src/Interval.h
```diff
@@ -14,7 +14,13 @@ namespace Internal {
 struct Interval {
 
     /** Exprs to represent positive and negative infinity */
-    HALIDE_EXPORT static Expr pos_inf, neg_inf;
+    static Expr pos_inf, neg_inf;
```
zvookin: Why not just make getter methods ```pos_inf()``` and ```neg_inf()``` and move the constants into an anonymous namespace in the file?

zvookin: Or make them static in the function itself.

src/Interval.cpp
steven-johnson: Interesting; I always like it because I feel that it makes the result *more* readable (since the staticness is only reflected syntactically in the .h file); the fact that recapitulation of it here isn't required by C++ strikes me as a language defect. But I can live with a more verbose way of doing the same thing.

src/Interval.h
steven-johnson: I did it this way to avoid code churn -- if there's no reluctance to that approach I'll do that instead.

abadams: We have lots of semi-bounded intervals. Within the compiler it could be a performance hit to need to call a function to construct one.

steven-johnson: In that case I can make the actual `Expr` values private, which will allow for efficient construction.

abadams: True, the main requirement is for has_upper_bound / has_lower_bound to be fast. Explicit use of pos_inf / neg_inf elsewhere in the compiler could maybe be replaced with helpers to construct intervals that are unbounded above/below

steven-johnson:
Yeah, this is just a mess -- I was trying to finesse with `HALIDE_NO_USER_CODE_INLINE` so that internal code could get the 'fast' inlined implementations and external code would fall back to slow-but-safe... but of course that only works if there is a non-inlined implementation for the external code to actually fall back to use (otherwise it uses its own and we are back to square one). Looks like `HALIDE_NO_USER_CODE_INLINE` is currently really only used with templated code, which makes sense.

The obvious answer here is to actually measure compilation time for the current code vs. a fully-not-inlined-version to see if it really moves the needle enough to worry about; are there good reference tests we use for this?

steven-johnson:
Various finessed attempts at this didn't work, so I just reverted to bottlenecking all access thru static functions (even internally). It's not clear what performance hit during simplification/compilation this might be; any suggestions for a useful benchmark would be welcomed.

abadams:
Check local laplacian compile time.

steven-johnson:
Running `make clean && time make bin/host/local_laplacian.a` three times:

On my crazy-fast Linux desktop:

* master *

```
real    0m13.284s
user    0m12.894s
sys 0m0.390s

real    0m13.388s
user    0m12.972s
sys 0m0.415s

real    0m13.256s
user    0m12.788s
sys 0m0.468s
```

* srj-export-inf *

```
real    0m13.757s
user    0m12.878s
sys 0m0.878s

real    0m13.406s
user    0m12.988s
sys 0m0.417s

real    0m13.490s
user    0m13.071s
sys 0m0.419s
```


On my almost-unusably-slow MacBook Air (which is fortunately a temporary loaner):

* master *

```
real    0m22.228s
user    0m21.146s
sys 0m0.595s

real    0m23.536s
user    0m22.416s
sys 0m0.634s

real    0m27.352s
user    0m25.163s
sys 0m0.765s
```

* srj-export-inf *

```
real    0m21.343s
user    0m20.150s
sys 0m0.608s

real    0m21.742s
user    0m20.590s
sys 0m0.613s

real    0m22.494s
user    0m21.317s
sys 0m0.641s
```

abadams:
Any effect is probably being dominated by llvm compile time noise. Try just -e stmt to be sure.

steven-johnson:
Running `time bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host auto_schedule=false` three times:

On my crazy-fast Linux desktop:

* master *

```
real  0m0.686s
user  0m0.666s
sys 0m0.026s

real  0m0.687s
user  0m0.663s
sys 0m0.029s

real  0m0.687s
user  0m0.673s
sys 0m0.019s
```

* srj-export-inf *

```
real  0m0.694s
user  0m0.663s
sys 0m0.037s

real  0m0.691s
user  0m0.661s
sys 0m0.036s

real  0m0.687s
user  0m0.663s
sys 0m0.030s
```


On my almost-unusably-slow MacBook Air (which is fortunately a temporary loaner):

* master *

```
real  0m1.008s
user  0m0.954s
sys 0m0.043s
srj@srj-macbookair:~/GitHub/Halide/apps/local_laplacian$ time bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host auto_schedule=false

real  0m0.983s
user  0m0.935s
sys 0m0.043s
srj@srj-macbookair:~/GitHub/Halide/apps/local_laplacian$ time bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host auto_schedule=false

real  0m0.983s
user  0m0.939s
sys 0m0.040s
```

* srj-export-inf *

```
real  0m1.240s
user  0m1.135s
sys 0m0.048s

real  0m1.513s
user  0m1.359s
sys 0m0.053s

real  0m1.268s
user  0m1.215s
sys 0m0.035s
```

abadams:
Odd that this is bad on osx but not linux. Perhaps on linux we're getting some sort of link-time inlining.

One thought is that inline functions can safely appear in multiple translation units provided the definitions match, so you could copy-paste the definitions of get_pos/neg_inf() into the header guarded by #if COMPILING_HALIDE

steven-johnson:
Yeah, I was hoping to avoid injecting specializations based on COMPILING_HALIDE or not, as it seems like a dubious precedent. Maybe it's worth it in this case.

Re; linux vs osx, I suspect the much bigger difference is the CPU; my linux box is a dual-Xeon GOLD 6154 running at something like 3.5ghz with a crapload of cache, whereas my MBAir is running a 1.6GHz 2-core i5

steven-johnson:
> so you could copy-paste the definitions of get_pos/neg_inf() into the header guarded by #if COMPILING_HALIDE

Also: for this to actually work, wouldn't we have to do it for all the other inlined methods in Interval.h (otherwise they'd end up getting inlined outside Halide and referencing the globals?)

abadams:
If the other methods could get to the globals via the static getters, then when they get inlined into user code they'll still go via the static getters, right?

steven-johnson:
OK, so this approach "works" -- it has a bit of a code smell, but I guess I can live with it. Thoughts?

steven-johnson:
The one failure is unrelated random fuzz-simplify fail: https://github.com/halide/Halide/issues/4003

steven-johnson:
(Rebased to avoid all the weird experiments along the way.)

steven-johnson:
Gentle Friday morning review ping

PR #3991, opened 2019-07-11
Make rhs argument of the shift operators in Codegen_C templatized

white-pony:
This is to handle the case when lhs is signed and rhs is unsigned.


steven-johnson:
Waiting on https://github.com/halide/Halide/pull/3992 to go green

PR #3993, opened 2019-07-11
Hexagon: Fixes the slicing bounds for the last vector in interleave_vector func

aankit-ca:
@pranavb-ca @dpalermo @dsharletg I don't have a reproducer for this but seems like a trivial change.


steven-johnson:
A testcase would surely be nice.

dsharletg:
This seems correct, but as Steven says, this one is really hard to verify without a test case. It doesn't seem trivial to me, I think you should either explain why it should be what you are changing it to, or, a test case that fails at master and passes with this, would be really helpful.

aankit-ca:
Sure, I'll get a testcase for this.

dsharletg:
Re-reading this, it's clear this is correct. Going to merge this.

dsharletg:
That said, if you have a test case that exercises this code path, that would still be much appreciated.

PR #3995, opened 2019-07-11
Add HALIDE_HIDDEN

steven-johnson:
This is a wrapper for `__attribute__((visibility("hidden")))`, intended for use with data members that shouldn't ever need to be exported from libHalide.  Initially applying to the buffer accessor set of strings from Call::make, since they aren't currently used outside libHalide and really never should be.


abadams:
What's our rationale for not just exporting everything? I thought that's where we landed last time we considered visibility settings.

I'm loathe to tell people trying to do work outside of the compiler (e.g. using custom lowering passes) what they can and cannot use from within the compiler.

steven-johnson:
> What's our rationale for not just exporting everything

Because AFAICT, there isn't a way to tell MSVC to export all *data* by default.  (You can get it to export all *functions* by default.) If we go this route, we're going to have a long tail of whack-a-mole of MSVC breakages. IMHO, better to declare that all functions are fair game, but no data is; selectively hiding things this way will make the world easier down the road.

abadams:
Ugh, I think that means we need to make the intrinsics an enum instead.

steven-johnson:
> we need to make the intrinsics an enum instead.

Working on it now :-)

steven-johnson:
Clean. PTAL.

abadams:
Do you think HALIDE_NO_EXPORT would be clearer? (More obviously the opposite of HALIDE_EXPORT)?

steven-johnson:
> Do you think HALIDE_NO_EXPORT would be clearer? (More obviously the opposite of HALIDE_EXPORT)?

 IMHO, the end-goal of this stuff should be HALIDE_EXPORT going away entirely, so it would be the opposite of something that doesn't exist. HALIDE_HIDDEN would seem to be more obviously reflective of the gcc/clang notation. But I could live with either one.

abadams:
I think we should only do this in the public headers that go into Halide.h (e.g. no need to do it in the llvm stuff)

steven-johnson:
> I think we should only do this in the public headers that go into Halide.h (e.g. no need to do it in the llvm stuff)

Agreed 100%. Aside from what's in this PR, only other thing in this area I want to do is a quick followup to hide the stuff in binary2cpp output (ie our runtime bitcode) by default; this is not at all essential, but is arguably good hygiene to minimize the exported data footprint from our SO.

abadams:
I was assuming CodeGen_LLVM.h and IntegerDivisionTable.h were not public headers, but to my surprise they are. Do you remember why? I guess we've just been making things public by default.

abadams:
I'm confused though - by introduced HALIDE_HIDDEN as the inverse of HALIDE_EXPORT, don't we have the same whack-a-mole game with msvc, where if we forget to say HALIDE_HIDDEN and then depend on it in an inlined method it will only break on msvc?

steven-johnson:
> I was assuming CodeGen_LLVM.h and IntegerDivisionTable.h were not public headers, but to my surprise they are. Do you remember why? 

Nope.

> I guess we've just been making things public by default.

Yep. (Didn't you say you wanted ~everything available externally for ease of experimentation?)

(Note that even if they weren't public, the data members would still be exported by default in our .so; this arguably isn't a real problem if the headers aren't exposed, but minimizing the symbols exposed in our .so trivially improves load times, I suppose)

steven-johnson:
> I'm confused though 

Yeah, you have a good point. But in defense of this plan, if we're going to have to whack moles, IMHO it's better to do so by limiting access rather than expanding it.

abadams:
So if it's whack-a-mole either way we need a solid reason in favor of hiding things. Making things public has the advantage of making it easier to implement certain things (e.g. Interval), and making it easier to experiment from outside libHalide. IIRC the reason to not like data members as part of the public API is that you get very little control over static initialization order (I think that was the problem with _0, _1, _2, and friends)? Are there other reasons to not want data members exported?

steven-johnson:
> the reason to not like data members as part of the public API is that you get very little control over static initialization order

That was indeed the impetus for _0 and friends; the fact that it happened there makes other globals potentially susceptible to the same thing (e.g. Interval::pos_inf and neg_inf). (The potential may be small, but if we can reduce small -> literally zero without extra overhead, that's a good thing.)

But I'm gonna argue in favor of it for a more philosophical reason: it's preferable to have a public API (even one as fluid as Halide's) be all functions-and-constants, and not rely on arbitrary static data being exported as an implicit part of that API. 

From recent inspection, it turns out that this was already close to being the case for Halide; aside from pos_inf/neg_inf, the only "exposed" data members in our API were either accidental (via inlining in the client) or "internal-but-needed-for-tests" (e.g. the Call::some_instrinsic_name strings). 

This is not a hill I think is worth dying on, but it's a baby step in making our API cleaner.

steven-johnson:
(Note also that if we decide against hiding data in the service of this goal, we may want to revert https://github.com/halide/Halide/pull/3998, as that was the main impetus for that change as well.)

abadams:
Your philosophical reason doesn't really seem to be philosophical in nature. There should be an engineering reason for that preference that we can state. Constant global variables with meaningful unique addresses in public APIs are harder to refactor, and can cause linkage and constructor/destructor order problems depending on the platform. Those are good reasons to avoid them.

I think we'd be better of explicitly encapsulating each case though, rather than adding a new whack-a-mole macro. The integer division tables are also accessible via FastIntegerDivide.h, so we can just drop IntegerDivisionTable.h from Halide.h (and see if we can just have the other in-compiler uses go via that too). The llvm target-enabled things could be functions that return bool instead of globals, conditionally compiled to either return true or false using the macro junk in CodeGen_LLVM.cpp

steven-johnson:
None of the changes in this PR are actually required to achieve my short-term goal (namely, avoiding the need to use HALIDE_EXPORT); rather, these were intended as pre-emptive ways to prevent new dependencies on internal data structures from growing new external uses. I'll abandon this effort for now.

abadams:
I did a hunt through ld flags to see if there's some easy to way to set default visibility of data symbols, but it doesn't look possible without nasty nm-based scripts.

steven-johnson:
Pretty sure we could do this with a custom linker script, but not gonna worry about it until/unless we have a clear need at this point.

PR #3999, opened 2019-07-12
Capture rval exprs by value in irmatcher

abadams:
IRMatcher patterns captured referenced expressions without incrementing their ref count. This is not great for rval Exprs used in a pattern. Naively capturing them by value instead blows up runtime and code size because the SpecificExpr class, which is used to hold an Expr as a node in an IRMatcher pattern is also used by value to pass through expressions when matching. This PR uses const BaseExprNode & for that purpose instead, and converts an attempted match on a SpecificExpr to matching on the underlying const BaseExprNode & as early as possible in the matching tree using a new unwrap helper.

### DIFF
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -200,24 +200,30 @@ struct SpecificExpr {
 
     constexpr static uint32_t binds = 0;
 
-    const BaseExprNode &expr;
+    Expr expr;
 
     template<uint32_t bound>
     HALIDE_ALWAYS_INLINE
-    bool match(SpecificExpr e, MatcherState &state) const noexcept {
-        return equal(expr, e.expr);
+    bool match(const SpecificExpr &e, MatcherState &state) const noexcept {
+        return match<bound>(*e.expr.get(), state);
+    }
+
+    template<uint32_t bound>
+    HALIDE_ALWAYS_INLINE
+    bool match(const BaseExprNode &e, MatcherState &state) const noexcept {
+        return equal(*expr.get(), e);
     }
 
     HALIDE_ALWAYS_INLINE
     Expr make(MatcherState &state, halide_type_t type_hint) const {
-        return &expr;
+        return expr;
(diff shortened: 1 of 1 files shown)


abadams:
Thanks to @LazyDodo for catching this

src/IRMatch.h
```diff
@@ -200,24 +200,24 @@ struct SpecificExpr {
 
     constexpr static uint32_t binds = 0;
 
-    const BaseExprNode &expr;
+    Expr expr;
```
steven-johnson: Given the subtlety of the failure, it's probably worth adding something similar to the PR description here for future maintainers

src/IRMatch.h
```diff
@@ -552,9 +552,27 @@ Const pattern_arg(int64_t x) {
...
+    return {e};
+}
+
+// Helpers to deref SpecificExprs to const BaseExprNode & rather than
+// passing them by value anywhere (incurring lots of refcounting)
+template<typename T,
+         // T must be a pattern node
+         typename = typename std::decay<T>::type::pattern_tag,
+         // But T may not be SpecificExpr
+         typename = typename std::enable_if<!std::is_same<typename std::decay<T>::type, SpecificExpr>::value>::type>
```
steven-johnson: (Unrelated: On the one hand, I'm glad that C++11, etc lets us do this sort of thing, but geez is the syntax painful.)

steven-johnson:
Is there any reasonable way to write a test for this?

Any obvious reason that it would fail in the MSVC config but not under ASAN? (Perhaps just allowed compiler latitude?)

src/IRMatch.h
abadams: Capturing an Expr seems totally reasonable. Really the original should have had a good justification for why capturing a const BaseExprNode & was safe.

steven-johnson:
Is this ready to land?

PR #4000, opened 2019-07-12
Add suffixes to integer constants in CodeGen_C

white-pony:
I think it should work without it most of the time, but we have some weird compiler which can't handle properly some of the generated statements, such as: 

int64_t const _200 = (int64_t)(-2147483648);

and will tell something like:

warning: this decimal constant is unsigned only in ISO C90

which means, if I understand correctly, that it treats 2147483648 as a signed int (and minus is just a unary operator), which won't fit into 32 bit and will overflow. If I add suffixes it works fine and I think it won't change anything for all other compilers (the value field in IntImm is always int64), so hopefully it's ok with you (if not maybe, we can conditionally add suffixes based on value itself?).

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -2064,12 +2064,12 @@ void CodeGen_C::visit(const IntImm *op) {
     if (op->type == Int(32)) {
         id = std::to_string(op->value);
     } else {
-        print_assignment(op->type, "(" + print_type(op->type) + ")(" + std::to_string(op->value) + ")");
+        print_assignment(op->type, "(" + print_type(op->type) + ")(" + std::to_string(op->value) + "ll)");
     }
 }
 
 void CodeGen_C::visit(const UIntImm *op) {
-    print_assignment(op->type, "(" + print_type(op->type) + ")(" + std::to_string(op->value) + ")");
+    print_assignment(op->type, "(" + print_type(op->type) + ")(" + std::to_string(op->value) + "ull)");
 }
 
 void CodeGen_C::visit(const StringImm *op) {
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -39,6 +39,7 @@ const string headers =
     "#include <math.h>\n"
     "#include <float.h>\n"
     "#include <assert.h>\n"
+    "#include <limits.h>\n"
     "#include <string.h>\n"
     "#include <stdio.h>\n"
     "#include <stdint.h>\n";
@@ -194,6 +195,14 @@ public:
     }
 };
 } // namespace
+
+#if LONG_MAX == 0x7fffffff
+#define ADD_INT64_T_SUFFIX(x) x##ll
+#define ADD_UINT64_T_SUFFIX(x) x##ull
+#else
+#define ADD_INT64_T_SUFFIX(x) x##l
+#define ADD_UINT64_T_SUFFIX(x) x##ul
+#endif
 
 )INLINE_CODE";
 }  // namespace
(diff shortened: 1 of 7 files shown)


steven-johnson:
Best branchname ever

steven-johnson:
Fails on WIn64 MSVC build:

```
CUSTOMBUILD : error : CL: clBuildProgram failed: CL_BUILD_PROGRAM_FAILURE [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_widening_reduction.vcxproj]
  Build Log:
<kernel>(66,22): error G3AE38AB9: 'long long' type is not supported [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_widening_reduction.vcxproj]
    short _5 = (short)(0ll);
                       ^
<kernel>(95,23): error G3AE38AB9: 'long long' type is not supported [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_widening_reduction.vcxproj]
    short _25 = (short)(4ll);
                        ^
<kernel>(97,23): error G3AE38AB9: 'long long' type is not supported [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_widening_reduction.vcxproj]
    short _27 = (short)(255ll);
                        ^
<kernel>(99,23): error G3AE38AB9: 'long long' type is not supported [C:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\run_correctness_widening_reduction.vcxproj]
    short _29 = (short)(0ll);
```

white-pony:
Ugh, sorry about that. 

I went through the build log and it looks like this change is breaking OpenCL tests on Windows (huh). If I read the spec correctly, OpenCL LL is reserved for 128-bit and L is for 64-bit, but in C++ L can be 32-bit and only LL is guaranteed to be 64-bit (https://en.cppreference.com/w/cpp/language/types). I think one potential way to make it work is to override these two functions in CodeGen_OpenCL and emit correct suffix for OpenCL there, but I am not sure if this is a good solution. What do you think?

steven-johnson:
> override these two functions in CodeGen_OpenCL and emit correct suffix for OpenCL there

SGTM

zvookin:
The other way to go would be to make a macro that is defined to use the right suffix based on either sizeof or stdint.

white-pony:
I added a macro to append the suffixes: it uses LONG_MAX, which is supported by OpenCL and is in limits.h (which I had to add) for C code. Is there a way for me to trigger windows build to check?

steven-johnson:
Still unhappy on Win32, alas

white-pony:
Just checked it on my Mac and it indeed fails now, but in different way. Compilation error is that the new macros I added are not defined for both Metal and OpenCL. Sorry about that, I got confused, they are actually added to the output stream in CodeGen_C constructor, however, for GPU backends it gets cleared later on (https://github.com/white-pony/Halide/blob/master/src/CodeGen_OpenCL_Dev.cpp#L659). 

It seems like there is no way to add macros (well, I couldn't find it), such that it's shared across all backends which use CodeGen_C. Maybe, I should go with the other option we've considered (override visit(IntImm) for OpenCL codegen).

steven-johnson:
Maybe add a way to add such macros?

white-pony:
> Maybe add a way to add such macros?

Yep, I'll do that if it's preferred. Also, trying to set up OpenCL on my machine to avoid bothering you with back and forth on broken tests.

white-pony:
Fixed OpenCL and other tests, please, take a look.

white-pony:
Added a special case for HLSL, please take a look.

PR #4001, opened 2019-07-12
Don't read out of bounds in simd_op_check

abadams:
The shufps test was reading out of bounds. We didn't notice because we were compiling with noasserts to aid with grepping the output assembly, and then running with the same target (oops). This PR changes things to run with assertions on, and use bounds querying to size the inputs instead of guessing.

Drive-by cleanup to target handling in infer_input_bounds - it was using the jit target from the environment and pointlessly recompiling, even if a working jit module had already been compiled.

Thanks to @LazyDodo for the diagnosis


src/Pipeline.cpp
```diff
@@ -1105,17 +1105,18 @@ void Pipeline::realize(RealizationArg outputs, const Target &t,
 }
 
 void Pipeline::infer_input_bounds(RealizationArg outputs, const ParamMap &param_map) {
-    Target target = get_jit_target_from_environment();
-
-    compile_jit(target);
+    if (!contents->jit_module.compiled()) {
```
steven-johnson: I know that this call is used very infrequently, but the implicit read-from-the-env makes me nervous. Could we at least add `const Target &target = Target()` as an argument, as we do everywhere else?

steven-johnson: (And yes, I realize this is pre-existing behavior, so it's definitely scope creep.)

test/correctness/simd_op_check.cpp
```diff
@@ -254,7 +226,43 @@ struct Test {
 
         bool can_run_the_code = can_run_code();
         if (can_run_the_code) {
-            Realization r = error.realize(target.without_feature(Target::NoRuntime));
+            Target run_target = target;
+            run_target.set_feature(Target::NoRuntime, false);
```
steven-johnson: style nit:  might be easier to read as a chained series of `without_feature()` calls?

steven-johnson:
Where was NoAsserts being set?

src/Pipeline.cpp
abadams: I initially did this, but the thing is that setting the target in the call to realize and infer_input_bounds is sticky - it influences the target that will be used later if one is not specified. I think for people using non-default targets, calling compile_jit(my_target) to set it is clearer.

abadams:
noasserts (and no-lots-of-other-things) was set because the primary purpose of this test is to grep emitted assembly for the pattern. Minimizing that assembly minimizes the risk of a false positive.

Then we were using that same target to actually run the code, which was a mistake.

steven-johnson:
Nevermind -- the place we were setting it wasn't in the diff here (just the new clearing of it) so it was confusing. All good.

src/Pipeline.cpp
zvookin: Am I correct that this change here is just a performance optimization? As in it does not change any behavior. If so, I'm fine with not combining any possible change in behavior (or performance regression) with a fix that possibly gets rid of a long standing testing failure.

zvookin: Hmmm... actually I think this does change behavior. Before, infer_input_bounds always returned the bounds for the global JIT target. Now it returns that or whatever ```compile_jit``` was last called with if the Func is already compiled. This is indeed problematic. The previous operation was wrong as evidenced by the bug. We're now changing that to a strangely stateful thing. At the very least I would expect that to require documentation updates. But I do sort of agree that the consistent thing would be to add a target argument.

Another way to go would be to disallow rejitting a Func without some much more explicit mechanism. What we have now is also an invitation to thread issues.

abadams: The schedule is fixed, so the bounds inferred cannot possibly depend on the target at this point. It's simply cheapest to use already-compiled code.

abadams: ... unless no_bounds_query is set.

abadams: ok, with my latest push the only user-visible change in behavior (ignoring compile time) is that whereas HL_JIT_TARGET=host-no_bounds_query would previously cause havoc, now it cause rejitting if you do an input bounds query.

steven-johnson:
So this still seems to be cratering on Windows.

abadams:
I think I messed up the alignment assumption (setting it after compilation is not particularly helpful). I have attempted a fix.

steven-johnson:
Apparently simd_op_check is still unhappy :-/

abadams:
It looks like it's generating different-but-still-correct instruction sequences. I'm not sure why this wasn't happening before though - my change shouldn't have affected that. I'll puzzle it out.

steven-johnson:
(synced to head to hopefully reduce irrelevant failures recently fixed)

steven-johnson:
woot, everything is green! Do we think this is ready to land?

PR #2580, opened 2017-11-30
Metal: pre-create pipeline states

shoaibkamil:
This PR creates `MTLComputePipelineState` objects for each function in a Halide Metal library in `halide_metal_initialize_kernels()`.  Previously, these objects were created for each kernel launch in `halide_metal_run()`, but according to the [Metal Best Practices Guide ](https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/PersistentObjects.html#//apple_ref/doc/uid/TP40016642-CH4-SW1), the `MTLComputePipelineState` objects can be created once and reused.  We pay a greater cost at initialization time, but the cost is amortized across multiple calls to the kernel.

For an unscientific example, we can run a test with the debug runtime and just compare (average) reported times for `halide_metal_run()`:

On master:
```
$ HL_JIT_TARGET=host-metal-debug  make -j5 performance_wrap 2>&1| grep run | awk '{sum += $4} END {print sum/NR}'
0.0455317
```

With this branch:
```
$ HL_JIT_TARGET=host-metal-debug  make -j5 performance_wrap 2>&1| grep run | awk '{sum += $4} END {print sum/NR}'
0.0243639
```

I *believe* I'm correctly managing Objective C memory, but please take a look at that in particular in this code.

CC @zvookin @slomp

PR #4008, opened 2019-07-12
cache metal compute pipelines

mikewoodworth:
instead of recreating in each invocation of halide_metal_run

"Creating compute pipeline state objects can be expensive, so you usually create these pipeline objects when your app is initialized and reuse them throughout its lifetime."
- https://developer.apple.com/documentation/metal/mtlcomputepipelinestate?language=objc

This pull also moves the call to halide_start_clock out of halide_metal_acquire_context so that clients that overload that method don't experience crashes when running in debug.

### DIFF
--- a/src/runtime/metal.cpp
+++ b/src/runtime/metal.cpp
@@ -221,6 +221,12 @@ struct device_handle {
     mtl_buffer *buf;
     uint64_t offset;
 };
+    
+struct cached_pipeline_state {
+    const char* function_name;
+    mtl_compute_pipeline_state *precompiled_pipeline_state;
+    cached_pipeline_state *next;
+};
 
 // Structure to hold the state of a module attached to the context.
 // Also used as a linked-list to keep track of all the different
@@ -228,6 +234,7 @@ struct device_handle {
 // when then context is released.
 struct module_state {
     mtl_library *library;
+    cached_pipeline_state *cached_pipeline_state_list;
     module_state *next;
 };
 WEAK module_state *state_list = NULL;
@@ -281,9 +288,6 @@ WEAK int halide_metal_acquire_context(void *user_context, mtl_device **device_re
     halide_assert(user_context, &thread_lock != NULL);
     while (__sync_lock_test_and_set(&thread_lock, 1)) { }
 
-#ifdef DEBUG_RUNTIME
-        halide_start_clock(user_context);
-#endif
 
     if (device == 0 && create) {
         debug(user_context) <<  "Metal - Allocating: MTLCreateSystemDefaultDevice\n";
@@ -341,6 +345,10 @@ public:
 
 WEAK void MetalContextHolder::save(void *user_context_arg, bool create) {
     user_context = user_context_arg;
+    #ifdef DEBUG_RUNTIME
+        halide_start_clock(user_context);
+    #endif
+    
     pool = create_autorelease_pool();
     error = halide_metal_acquire_context(user_context, &device, &queue, create);
 }
@@ -481,6 +489,7 @@ WEAK int halide_metal_initialize_kernels(void *user_context, void **state_ptr, c
     if (!(*state)) {
         *state = (module_state*)malloc(sizeof(module_state));
         (*state)->library = NULL;
+        (*state)->cached_pipeline_state_list = NULL;
         (*state)->next = state_list;
         state_list = *state;
     }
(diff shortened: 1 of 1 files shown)


src/runtime/metal.cpp
```diff
@@ -714,18 +734,42 @@ WEAK int halide_metal_run(void *user_context,
     halide_assert(user_context, state_ptr);
     module_state *state = (module_state*)state_ptr;
 
-    mtl_function *function = new_function_with_name(state->library, entry_name, strlen(entry_name));
-    if (function == 0) {
-        error(user_context) << "Metal: Could not get function " << entry_name << "from Metal library.\n";
-        return -1;
+    
+    cached_pipeline_state *cached_pipeline = state->cached_pipeline_state_list;
+    while (cached_pipeline != NULL && strcmp(cached_pipeline->function_name, entry_name) != 0) {
```
shoaibkamil: I'd change this to `strncmp`.

src/runtime/metal.cpp
```diff
@@ -714,18 +734,42 @@ WEAK int halide_metal_run(void *user_context,
...
+        }
+ 
+        cached_pipeline_state *new_pipeline = (cached_pipeline_state*)malloc(sizeof(cached_pipeline_state));
+        new_pipeline->function_name = entry_name;
+        new_pipeline->precompiled_pipeline_state = pipeline_state;
+        new_pipeline->next = state->cached_pipeline_state_list;
+        state->cached_pipeline_state_list = new_pipeline;
+        
+        cached_pipeline = new_pipeline;
+    }else{
```
shoaibkamil: Spacing

src/runtime/metal.cpp
```diff
@@ -714,18 +734,42 @@ WEAK int halide_metal_run(void *user_context,
...
+    while (cached_pipeline != NULL && strcmp(cached_pipeline->function_name, entry_name) != 0) {
+        cached_pipeline = cached_pipeline->next;
     }
 
-    mtl_compute_pipeline_state *pipeline_state = new_compute_pipeline_state_with_function(metal_context.device, function);
-    if (pipeline_state == 0) {
-        error(user_context) << "Metal: Could not allocate pipeline state.\n";
-        return -1;
+    
+    if(!cached_pipeline){
```
shoaibkamil: Why not just create the pipeline states during `halide_metal_initialize_kernel()`?

shoaibkamil: Also, spacing.

shoaibkamil:
In principle, I agree with this change-- I had a PR that did this (#2580), but at the time we didn't merge it in, so I defer to @zvookin.

mikewoodworth:
> In principle, I agree with this change-- I had a PR that did this (#2580), but at the time we didn't merge it in, so I defer to @zvookin.

I'll wait to see what @zvookin says before making changes.

zvookin:
Will review shortly. I need to go page back in some knowledge on this. I recall a PR for OpenCL recently where there was some debate about what is supposed to be cheap and what is supposed to be expensive and what state has to be recreated under certain circumstances. E.g. is there any situation were we have to recreate the pipeline state for the kernel invocation to be valid? (E.g. if the machine switches from discrete GPU to integrated between two calls? Usually Apple's stuff takes care of that internally, but I have to go check the docs.)

Ultimately I'd like to surface this stuff via an explicit (and general) API and allow end user control, but it is clear we will need some changes like this for a while.

mikewoodworth:
In the default implementation halide_default_metal_acquire_context will return the preexisting device for all later invocations... we should be on the same device for all subsequent calls to acquire. But I suppose that problem could arise when overloading halide_metal_acquire_context.

We can cache device and reuse for later invocations, or do a simple match (cached == acquired) and if not recompile. 

If we're worrying about device changing mid use, I suspect there are other concerns with buffers being on the wrong device, but I'd have to examine the source more closely.

steven-johnson:
Ready for final review?

mikewoodworth:
I haven't made any of @shoaibkamil fixes yet, waiting to hear if @zvookin is ok with the general direction first.

steven-johnson:
Ping to see if still waiting on feedback.

steven-johnson:
What's the status on this PR -- still waiting on feedback?

mikewoodworth:
confirm with @zvookin, but I think its a non-starter

Issue #4010, opened 2019-07-15
Func name "head1_filter.extent.0_d__" is invalid. Func names may not contain the character '.', as it is used internally by Halide as a separator


OuHangKresnik:
When I tried to compile autoscheduler, such error occurs.

Linux Platform

HL_PERMIT_FAILED_UNROLL=1 ../autoscheduler/bin/cost_model.generator -g train_cost_model -o ../autoscheduler/bin/cost_model -f train_cost_model target=host-no_runtime auto_schedule=false -e stmt,static_library,h,assembly

OuHangKresnik:
Does anyone know what's happening here ? I'm the first time to try this out.

thiesmoeller:
Did a git a bisect run ...:

```
46abc24b8a62cd7d3d58ac066df8896201208485 is the first bad commit
commit 46abc24b8a62cd7d3d58ac066df8896201208485
Author: Tzu-Mao Li <tzumao@mit.edu>
Date:   Mon Jul 8 17:32:37 2019 -0400

    Allow taking derivative with respect to Param

```

OuHangKresnik:
Thank you @thiesmoeller  I will try this out today.

BachiLi:
: ( I'll take a look

OuHangKresnik:
After I roll back to 6e99decbf0876bcfbf856d86afe6d76fe4019a7e, the one just before 46abc24b8a62cd7d3d58ac066df8896201208485, the compilation still fails:

../autoscheduler/train_cost_model.cpp:358:0: error: ignoring #pragma omp parallel [-Werror=unknown-pragmas]
             #pragma omp parallel for
 ^
cc1plus: error: unrecognized command line option ‘-Wno-unknown-warning-option’ [-Werror]

Not enable openmp ? -fopenmp  in ./support/Makefile.inc

thiesmoeller:
use clang.
This was fixed later..

ouhang <notifications@github.com> schrieb am Mi., 17. Juli 2019, 09:01:

OuHangKresnik:
what libpng and libz version number does Halide use ?

OuHangKresnik:
Also, when I tried to run:

Now make -j32 autotune has passed, but

make -j32 demo in apps/autoscheduler is failed

bin/host/demo.registration.cpp: In constructor ‘halide_nsreg_demo::{anonymous}::Registerer::Registerer()’:
[...]

PR #4016, opened 2019-07-17
Fix issue #4010

BachiLi:
In autodiff, legalize adjoint function naming, and don't propagate to non float parameters.


test/correctness/autodiff.cpp
```diff
@@ -1246,7 +1246,8 @@ void test_select_guard() {
 }
 
 void test_param() {
-    Param<float> param(2.f);
+    // Give the parameter an illegal name to test the function renaming.
+    Param<float> param("param.0", 2.f);
```
abadams: We may forbid explicit param names with dots at some point. Instead, make it a buffer param, and use its width where you use param below.

BachiLi: Like this? https://github.com/halide/Halide/pull/4016/commits/bac5e9be0e113b65cecce2cb4a01d02d5223356b

BachiLi: Accidentally turned out all other tests. Fixed in https://github.com/halide/Halide/pull/4016/commits/1e4076d653b7bc5459b99a3205efde923d0acf9d

BachiLi: Sorry, my brain wasn't working. Fixed the test by using an ImageParam. See https://github.com/halide/Halide/pull/4016/commits/e7ca5649a079d66ab72a24af582cfe76ed3aaa7c

steven-johnson:
Is this ready to land?

BachiLi:
I think it is ready but @abadams should take a quick look first.

PR #4021, opened 2019-07-18
Upgrade make files to understand LLVM 10+

steven-johnson:


CMakeLists.txt
```diff
@@ -24,8 +24,17 @@ if("${HALIDE_REQUIRE_LLVM_VERSION}" STREQUAL "")
   find_package(LLVM REQUIRED CONFIG)
 else()
   # Find a specific version.
-  string(SUBSTRING "${HALIDE_REQUIRE_LLVM_VERSION}" 0 1 MAJOR)
-  string(SUBSTRING "${HALIDE_REQUIRE_LLVM_VERSION}" 1 1 MINOR)
+  string(REGEX REPLACE
```
abadams: Does llvm even use minor versions anymore?

steven-johnson: I don't know.

abadams: Or, to put it another, is it enough for us to just check the major version?

steven-johnson: According to http://blog.llvm.org/2016/12/llvms-new-versioning-scheme.html: no, it will always be zero henceforth, so the version will be MAJOR.zero.patch. Want me to revise?

abadams: I mean, it's breaking the build right? So only if you have time to kill.

PR #4023, opened 2019-07-19
Add Target::TracePipeline feature flag

dsharletg:
This PR adds a `Target::TracePipeline` target feature flag. The motivation for this is that I would like to get a `halide_trace` call corresponding to the beginning and end of a pipeline, but no more than this. I don't think it is possible to do this because the next possible thing to trace is realizations, but if a pipeline contains multiple compute_root stages (or even outputs), the realizations are hard to predict and may not cover the whole pipeline.

I'm throwing this out there just as a suggestion for now. This PR doesn't have a lot of time invested in it, but, I really would like to get a way to get traces at the pipeline level somehow. The motivation for this is I want to generate android traces corresponding to *most* of our pipelines.

### DIFF
--- a/src/Generator.h
+++ b/src/Generator.h
@@ -2990,6 +2990,10 @@ public:
         get_pipeline().add_requirement(condition, std::forward<Args>(args)...);
     }
 
+    void trace_pipeline() {
+        get_pipeline().trace_pipeline();
+    }
+
 protected:
     GeneratorBase(size_t size, const void *introspection_helper);
     void set_generator_names(const std::string &registered_name, const std::string &stub_name);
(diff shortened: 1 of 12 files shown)


steven-johnson:
Not opposed to this PR, but given your goal, could you achieve the same thing with the Debug flag and sniffing the logs?

zvookin:
I would also expect there to be a method on ```Pipeline``` itself to turn this on for a single pipeline without modifying the build setup. (Akin to ```Func::trace_realizations```. Perhaps ```Pipeline::trace_calls``` or maybe just ```Pipeline::trace```.)

abadams:
I think this is a good change. +1 to Zalman's suggestion though.

dsharletg:
The Debug flag adds a lot of overhead, and sniffing the logs after the fact is not sufficient. I want to call Android's tracing APIs when pipelines start and end, so the information is included in a broader view of the program (the android trace).

I will add a Pipeline::trace_pipeline() call.

steven-johnson:
All green. Ready to land?

PR #4025, opened 2019-07-19
Improve codegen in HalideBuffer.h

steven-johnson:
Casual inspection of disassembly from clang showed that the template expansion of variadic methods can be pretty stupid: initialize_shape() wasn't flattening out into tail recursion at compile time, and we got separate (but equivalent) versions instantiated for all Buffer variants (ie Buffer<float> and Buffer<int> get their own versions).

This PR refactors the variadic stuff into an almost-entirely-inlined helper class that just flattens into stack storage, then makes initialize_shape into a single implementation.

EDIT: I had assumed my test code was being compiled -O3, but in fact it was using OPTIMIZE_FOR_BUILD_TIME (-O0) instead of OPTIMIZE (-O3); compiling at -O3 does the reasonable smart thing.

That said, this change gives *much* better code at -O0, so it's probably worth considering (though I'm not gonna spend time searching for other similar optimizations).

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -89,35 +89,6 @@ struct DeviceRefCount {
     BufferDeviceOwnership ownership{BufferDeviceOwnership::Allocated};
 };
 
-namespace Internal {
-
-// This namespace contains some helpers for Buffer that don't depend
-// on Buffer's templated types, and thus are pulled out here to avoid
-// possible unnecessary code replication via template expansion.
-
-template<typename ...Args>
-struct GatheredInts {
-    const int size = (int) sizeof...(Args);
-    int values[sizeof...(Args)];
-    const bool any_zero;
-
-    inline explicit GatheredInts(Args... args) : any_zero(gather(values, args...)) {
-    }
-
-private:
-    template<typename ...Rest>
-    HALIDE_ALWAYS_INLINE static bool gather(int *v, int first, Rest... rest) {
-        *v++ = first;
-        return (!first) || gather(v, rest...);
-    }
-
-    HALIDE_ALWAYS_INLINE static bool gather(int *v) {
-        return false;
-    }
-};
-
-}  // namespace Internal
-
 /** A templated Buffer class that wraps halide_buffer_t and adds
  * functionality. When using Halide from C++, this is the preferred
  * way to create input and output buffers. The overhead of using this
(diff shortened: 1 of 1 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -855,8 +860,9 @@ class Buffer {
         buf.type = t;
         buf.dimensions = 1 + (int)(sizeof...(rest));
         make_shape_storage();
-        initialize_shape(0, first, rest...);
-        if (!any_zero(first, rest...)) {
+        Internal::GatheredInts<int, Args...> extents(first, rest...);
```
abadams: Can gatheredints just be a stack array of ints?

Here's an alternative with no helpers:
```
    bool any_zero = false;
    int extents[] = {first, rest...};
    for (int e : extents) {
        any_zero |= (e == 0);
    }
    if (!any_zero) {
        int idx = 0;
        for (int e : extents) {
            buf.dim[idx].min = 0;
            buf.dim[idx].extent = e;
            idx++;
        }
    }

```

abadams: I guess that would still imply some code duplication, but you could just stuff it in a non-recursive helper and call it like: `as<const void>().initialize_shape(first, rest...);`
to avoid duplication across template instantiations.

steven-johnson: I don't think `int extents[] = {first, rest...};` is legal. When I try using it I get `‘extents’ has incomplete type`.

steven-johnson:
PTAL

src/runtime/HalideBuffer.h
abadams: orly? I copied pasted that out of a test file that worked. I've been using that idiom for a while now as a simpler alternative to template-recursively gathering something. E.g. see for_each_value_impl. Here's my whole test:

```

struct d {
    int min, extent;
};
struct b {
    d dim[100];
} buf;

template<typename... Args>
void test(int first, Args... rest) {
    bool any_zero = false;
    int extents[] = {first, rest...};
    for (int e : extents) {
        any_zero |= (e == 0);
    }
    if (!any_zero) {
        int idx = 0;
        for (int e : extents) {
            buf.dim[idx].min = 0;
            buf.dim[idx].extent = e;
            idx++;
        }
    }
}

int foo(int a, int b, int c, int d) {
    test(a, b, c, d);
    return 0;
}
```
Maybe it needs an explicit type `int extents[] = {first, (int)rest...};` ?

steven-johnson: Hm, good thought, but `const int extents[] = {first, (int) rest...};` fails in the same way.

steven-johnson:
Withdrawn in favor of https://github.com/halide/Halide/pull/4027

PR #4026, opened 2019-07-19
Smarten code expansion in Buffer::copy_from()

steven-johnson:
When both types are statically known (and match), we can skip generating code for every possible size (and doing the runtime dispatching).

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1194,16 +1194,74 @@ public:
     }
 
 private:
+    template<typename T2, int D2>
+    struct CopyImplHelper {
+        template<typename MemType>
+        HALIDE_ALWAYS_INLINE static void copy_impl_typed(Buffer<const T2, D2> &src, Buffer<T2, D2> &dst) {
+            auto &typed_dst = (Buffer<MemType, D> &)dst;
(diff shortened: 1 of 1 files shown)


steven-johnson:
(Note that I made more changes after you approved this; I'll wait for you to LGTM them, or not, before landing)

abadams:
Doesn't seem to change the code-size of RunGenMain. I assume you're reducing code size of something internal. Could you test the following arguably simpler alternative and see if it would also meet your needs?

In the if-else statement, replace clauses of the form:
 `(type.bytes() == X)`
with 
 `T_is_void ? (type.bytes() == X) : (sizeof(not_void_T) == X)`

The idea is that if T is statically known, one of the branch conditions will fold to true and the others will fold to false. Dead code will then be eliminated, and the unwanted template instantiations will be dead-stripped.

steven-johnson:
RunGenMain has to include every possible permutation since it can't statically know what types are actually going to be used by what it links to. The use case for this change is code that uses only (say) Buffer<uint8_t>::copy_from(); previously it had to carry around the variants for 16, 32, 64 bit buffer copy as well.

Your dead-stripping solution does sound much simpler than mine, though (as usual). Will give it a try.

steven-johnson:
see https://github.com/halide/Halide/pull/4032

PR #4027, opened 2019-07-19
Slightly simplified #4025

abadams:


abadams:
@steven-johnson I was seeing a similar incomplete type error when I hadn't removed all instances of extents.any_zero or extents.size yet, so I think it was just a terrible C++ error message.

PR #4028, opened 2019-07-20
Minor optimization to make_shape_storage

steven-johnson:
Make a constexpr variant of make_shape_storage(), so that call sites that know at compiletime that dimensions <= D don't have to insert code to consider a dynamic allocation

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -268,20 +268,17 @@ private:
         }
     }
 
-    void make_shape_storage() {
-        if (buf.dimensions <= D) {
-            buf.dim = shape;
-        } else {
-            buf.dim = new halide_dimension_t[buf.dimensions];
-        }
+    void make_shape_storage(const int dimensions) {
+        // This should usually be inlined, so if dimensions is statically known,
+        // we can skip the call to new
+        buf.dimensions = dimensions;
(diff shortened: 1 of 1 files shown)


abadams:
The other option is inlining make_shape_storage so that the branch can resolve statically in cases where it is known statically. Is it not getting inlined in some build?

steven-johnson:
Yeah, that probably is happening -- I got sucked into a late Friday afternoon premature optimization rathole. I'll recheck the statusquo and if the optimization is happening at reasonable optimizer levels, I'll drop this.

abadams:
You could always force-inline it if you're getting code size issues in a debug build

steven-johnson:
Nah, the only thing I really wanted to accomplish was streamlining copy_from() (PR#4026) and verifying my changes via disassembly led me to this temptation .

steven-johnson:
Yeah, the templatized call was overkill (an 'ordinary' method gets inlined as you suspect), but the previous code wasn't able to elide the new, as the compile-time knowledge got lost via the store-and-load from buf.dimensions. PTAL.

Also note the drive-by change to use std::copy; at least for my compiler, this gets translated into a call to memmove (since the things being copied are POD).

steven-johnson:
(rebased to minimize cruft)

Issue #4030, opened 2019-07-22
Stage constructor seems to leave `dim_vars` uninitialized.


alexreinking:
In the string-args constructor for Stage, a local variable named `dim_vars` is created that shadows the class field.

https://github.com/halide/Halide/blob/fa25a98119c9a8dc4a3b4f758569966c5bd50b35/src/Func.h#L95-L99

That sequence should be replaced with a `reserve` and `emplace_back` loop a la:

```cpp
dim_vars.reserve(args.size());
for (const auto &arg : args) {
    dim_vars.emplace_back(arg);
}
```

If this looks right to everyone else, I'll open a PR. Regardless, the code sequence is a no-op unless the assertion fails. If we want the assertion, the loop could be dropped while maintaining the existing behavior.

When might this cause a problem, I wonder?

abadams:
Yikes. Code that uses that constructor doesn't seem like it could work. Maybe try just deleting that constructor first?

alexreinking:
The only usages are in FuncRef's operator=, here:

https://github.com/halide/Halide/blob/fa25a98119c9a8dc4a3b4f758569966c5bd50b35/src/Func.cpp#L2641-L2648

These could be modified to create Var vectors instead of string vectors and use the first constructor.

alexreinking:
Actually, I'm 99% convinced that fourth argument is unnecessary and it can always be derived from the `Function`'s `args()` method. I'm going to try that change locally.

alexreinking:
Yup, l I was right. Updated the PR. PTAL :slightly_smiling_face:

steven-johnson:
Is this fixed?

alexreinking:
@steven-johnson Yes. Fixed in #4031

PR #4031, opened 2019-07-22
Consolidate Stage ctors, fixing an uninitialized field in the process.

alexreinking:
Ref #4030 - this fixes a shadowed class field and replaces some copies with moves. Both changes are in the constructors for `Stage`.


abadams:
looks good to me, but please check if anyone is actually using that second broken constructor. If not, just delete it.

alexreinking:
@abadams - I found a better solution! Every call to the non-broken constructor created a fresh vector from the args of the Function parameter anyway. So consolidating the two constructors by dropping the `args` argument and creating that very same vector from the Function parameter fixes the initialization bug and simplifies every call site!

alexreinking:
Unsure what part of the change surfaced this bug (which I believe should be fixed independently). But the camera pipe app is failing because the globally defined shared variables get automatic names.

https://github.com/halide/Halide/blob/1104144d576f89c3172453ff54b08a4932ed60b1/apps/camera_pipe/camera_pipe_generator.cpp#L12-L13

In particular `c` gets automatically assigned to `v2`. But then `reset_unique_name_counters` is called by the Generator's main.

https://github.com/halide/Halide/blob/1104144d576f89c3172453ff54b08a4932ed60b1/src/Generator.cpp#L1041-L1043

Later when the camera pipeline detects that a GPU is available and tries to unroll the `x` variable here on line 164...

https://github.com/halide/Halide/blob/1104144d576f89c3172453ff54b08a4932ed60b1/apps/camera_pipe/camera_pipe_generator.cpp#L158-L166

... the temp var in here gets the exact same name of `v2` ...

https://github.com/halide/Halide/blob/1104144d576f89c3172453ff54b08a4932ed60b1/src/Func.cpp#L1458-L1462

... and we get the following error.

```
In schedule for output, can't create var v2 using a split or tile, because v2 is already used in this Func's schedule elsewhere.
```
I gave the camera pipe shared variables names, which fixes this app and all the others on my machine. I also had to fix the apps support Makefile to allow custom HL_TARGETs to be set and use the default CXX/CXXFLAGS/LDFLAGS.

alexreinking:
I did some testing on master. The global var `c` gets the name `v2` as before, but now the temp var in `split()` gets the name `v67`. The `Var` constructor gets pumped enough that a collision is avoided. The allocation of the `dim_vars` array on L95 has the side-effect of incrementing that counter a whole lot. When this PR removed it, the gap was closed and the names collided.

https://github.com/halide/Halide/blob/fa25a98119c9a8dc4a3b4f758569966c5bd50b35/src/Func.h#L95-L99

I believe this to be a pre-existing bug. I see two solutions:

1) Rather than reset the initialization counts to zero, freeze them at the start of the generator main, then reset them to _those values_ instead.
2) Disallow globally initialized Vars / other automatically named Halide constructs.

Maybe there's a better third option out there.

src/Func.cpp
```diff
@@ -1150,9 +1151,9 @@ Stage Stage::specialize(Expr condition) {
     // The user may be retrieving a reference to an existing
     // specialization.
     const vector<Specialization> &specializations = definition.specializations();
-    for (size_t i = 0; i < specializations.size(); i++) {
-        if (equal(condition, specializations[i].condition)) {
-            return Stage(function, specializations[i].definition, stage_index, dim_vars);
+    for (const auto & specialization : specializations) {
```
steven-johnson: `const autio &specialization`1

steven-johnson:
The call to reset_unique_name_counters() was inserted in https://github.com/halide/Halide/pull/3524 and was intended mainly for improving the debugging experience; we should simply remove this change entirely as the best quick fix.

alexreinking:
@steven-johnson Ok! I'll add that to this PR.

alexreinking:
@steven-johnson actually, maybe it's better to land this now and revert the other change as its own PR. I can fix the one space without needing to re-run the buildbots, right?

steven-johnson:
> maybe it's better to land this now and revert #3524 separately.

sure, but may I suggest the other order: revert #3524 first. I just opened https://github.com/halide/Halide/pull/4035 to do so.

src/Func.cpp
alexreinking: Just fixed the space. Didn't write `autio`, though. :wink:

alexreinking:
> sure, but may I suggest the other order: revert #3524 first. I just opened #4035 to do so.

For sure. This PR doesn't actually touch any of that code, so I don't think the order is important.

steven-johnson:
In that case, LGTM, but will defer to @abadams

apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -10,7 +10,7 @@ using namespace Halide;
 using namespace Halide::ConciseCasts;
 
 // Shared variables
-Var x, y, c, yi, yo, yii, xi;
```
abadams: Are the changes in this file necessary, or the result of intermediate states while debugging?

alexreinking: Not strictly necessary, but I don't think they're hurting anything, either.

abadams: I would prefer without then. I find "input" clearer than "func_in", and the explicit names are in theory redundant (at least on os x and linux, where we should be sniffing the dwarf syms. If we're not, that's a separate bug.)

alexreinking: @abadams -- Okay, I will do that, but just so you know "input" shadows a class field and the names were _not_ being properly sniffed, which is why making them explicit worked around the name-clashing-after-reset bug.

alexreinking: Done. The camera pipeline is now untouched by this PR.

alexreinking:
Not sure why Travis builds are failing... @abadams any idea?

steven-johnson:
travis timeouts should be fixed once https://github.com/halide/Halide/pull/4045 lands

alexreinking:
This is one for you, @abadams. I don't know if this is something that this PR could have introduced. It's the simplifier fuzz test.

```
FAILED TEST: fuzz_simplify
.Simplify fuzz test seed: 1563913770
a = (uint1)0
b = uint1(c)
c = (uint1)0
d = (uint1)1
e = uint1(e)
uint1x2(select((ramp((uint16(d)/(uint16(a) + (uint16)29933)), uint16(b), 2) != ramp(uint16(c), (uint16)40807, 2)), max(x2(uint16(a)), (ramp(uint16(d), (uint16)20756, 2) - x2(uint16(a)))), min(((x2(uint16(d)) - x2(uint16(d))) % (ramp(uint16(a), uint16(b), 2) % ramp(uint16(a), (uint16)739, 2))), select(((uint16(b) + uint16(a)) == uint16(e)), (x2((uint16)1243) % x2(uint16(d))), (x2(uint16(a)) + x2(uint16(d)))))))
uint1x2(select((ramp(((uint16(d)/(uint16(a) + (uint16)29933)) - uint16(c)), (uint16(b) - (uint16)40807), 2) == x2((uint16)0)), x2((uint16)0), max(ramp((uint16(d) - uint16(a)), (uint16)20756, 2), x2(uint16(a)))))
In vector lane 0:
uint1(max((uint16(d) - uint16(a)), uint16(a))) -> (uint1)1
uint1(select(((uint16(d)/(uint16(a) + (uint16)29933)) == uint16(c)), (uint16)0, max((uint16(d) - uint16(a)), uint16(a)))) -> (uint1)0
```

https://buildbot.halide-lang.org/master/#builders/8/builds/187

steven-johnson:
opened as https://github.com/halide/Halide/issues/4047

steven-johnson:
linux failure is spurious, see https://github.com/halide/Halide/issues/4049

PR #4032, opened 2019-07-22
Smarten code expansion in Buffer::copy_from() (V2)

steven-johnson:
Simpler alternative to https://github.com/halide/Halide/pull/4026, per @abadams suggestion


steven-johnson:
FYI, this is definitely smarter, but at the end of the day we're still copying one element at a time (I don't see the optimizer doing anything smart about combining). I have no evidence that this function *needs* to be faster, but adapting the code from device_buffer_utils.h would probably be faster. Not sure if worth doing or not.

PR #4037, opened 2019-07-22
move add_common_macros to after the version directive in OpenGL Compute

wangcheng22:
GLSL "version" directive must be first statement.  "add_common_macros" added macros before the "version" directive in OpenGL compute, which causes the following error:

Could not compile shader:
0(14) : error C0204: version directive must be first statement and may not be repeated

This PR move "add_common_macros" to after generating "version" directive in OpenGLCompute.


steven-johnson:
attn @white-pony

steven-johnson:
LGTM, waiting for build to go green

PR #4038, opened 2019-07-22
Fix add_common_macros for OpenGLCompute

steven-johnson:
From @wangcheng22:

Compute shader version directive must be first statement. The "add_common_macro" adds macros before "version" directive in OpenGLCompute code generation.  This CL moves "add_common_macro" after generating "version" directive


steven-johnson:
attn @white-pony

PR #4045, opened 2019-07-23
Improve CMake build time for tests

steven-johnson:


steven-johnson:
I don't think so; this should match what Makefile already does.

PR #4059, opened 2019-07-27
Clean up AutoScheduler outputs

steven-johnson:
Make the autoscheduler Featurization output a first-class Generator output, so we don't have to use HL_FEATURE_FILE to generate it. 

Smarten the existing ScheduleSource Generator output to work properly with custom autoschedulers (as well as the 'classic' built-in one). 

Modify the autotune scripts to use these outputs rather than setting HL_FEATURE_FILE or HL_SCHEDULE_FILE.


apps/autoscheduler/autotune_loop.sh
```diff
@@ -204,12 +203,13 @@ for ((BATCH_ID=$((FIRST+1));BATCH_ID<$((FIRST+1+NUM_BATCHES));BATCH_ID++)); do
         # benchmark them serially using rungen
         for ((SAMPLE_ID=0;SAMPLE_ID<${BATCH_SIZE};SAMPLE_ID++)); do
             S=$(printf "%d%02d" $BATCH_ID $SAMPLE_ID)
-            benchmark_sample "${DIR}/${SAMPLE_ID}" $S $EXTRA_ARGS_IDX
+            FNAME=$(printf "%s_batch_%02d_sample_%02d" ${PIPELINE} $BATCH_ID $SAMPLE_ID)
```
abadams: There are often going to be many more than 100 batches. I'd use %04d for the batch id

src/Generator.cpp
```diff
@@ -132,6 +132,10 @@ Outputs compute_outputs(const Target &target,
     if (options.emit_schedule) {
         output_files.schedule_name = base_path + get_extension(".schedule", options);
     }
+    if (options.emit_featurization) {
+        // TODO: apparently we use '.sample' for featurization files?
```
abadams: A sample is a featurization + a runtime + some ids, all together in one file. benchmark_sample above concats the runtime and ids.

apps/autoscheduler/autotune_loop.sh
steven-johnson: OK, but this line is an exact copy of line 198, so that limitation has already been in place for months. I'll make the change both places.

src/Generator.cpp
steven-johnson: Mm, so it would perhaps be clearer to have the 'featurization' output to have a different suffix (e.g. `.featurization`), and to have `augment_sample` use that to produce a `.sample` file?

steven-johnson: Update: pushed a commit that distinguishes between featurization and sample files; PTAL

steven-johnson:
(To be clear: although this has approval and is green, the last change I pushed, which distinguished between featurization and schedule files, should probably be explicitly reviewed and approved by someone who is doing active work on the autoscheduler.)

steven-johnson:
Update: I have some downstream code that is (partly) blocked by this, so I'm going to land it in this form; if there are issues with the featurization-vs-sample change, we can back that part out.

Issue #3507, opened 2018-12-15
Build error with msvc 2017 (15.9.3)


LazyDodo:
Out of the box the ms compiler will enable the latest C++ standard available, `uniform_int_distribution<int8_t>`  is not allowed in the C++17 standard and vs2017/19 will give the following build error:

```
------ Build started: Project: _halide_library_from_generator_rungen, Configuration: Release x64 ------
RunGenMain.cpp
c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\random(2401): error C2338: invalid template argument for uniform_int_distribution: N4659 29.6.1.1 [rand.req.genl]/1e requires one of short, int, long, long long, unsigned short, unsigned int, unsigned long, or unsigned long long
k:\halide\halide\tools\rungen.h(485): note: see reference to class template instantiation 'std::uniform_int_distribution<T2>' being compiled
        with
        [
            T2=int8_t
        ]
k:\halide\halide\tools\rungen.h(479): note: see reference to function template instantiation 'void Halide::RunGen::FillWithRandom<int8_t>::fill<int8_t,0x0>(Halide::Runtime::Buffer<int8_t,4> &,std::mt19937 &)' being compiled
k:\halide\halide\tools\rungen.h(479): note: see reference to function template instantiation 'void Halide::RunGen::FillWithRandom<int8_t>::fill<int8_t,0x0>(Halide::Runtime::Buffer<int8_t,4> &,std::mt19937 &)' being compiled
k:\halide\halide\tools\rungen.h(476): note: while compiling class template member function 'void Halide::RunGen::FillWithRandom<int8_t>::operator ()(Halide::Runtime::Buffer<void,4> &,int)'
k:\halide\halide\tools\rungen.h(188): note: see reference to function template instantiation 'void Halide::RunGen::FillWithRandom<int8_t>::operator ()(Halide::Runtime::Buffer<void,4> &,int)' being compiled
k:\halide\halide\tools\rungen.h(188): note: see reference to class template instantiation 'Halide::RunGen::FillWithRandom<int8_t>' being compiled
k:\halide\halide\tools\rungen.h(678): note: see reference to function template instantiation 'void Halide::RunGen::dynamic_type_dispatch<Halide::RunGen::FillWithRandom,Halide::Runtime::Buffer<void,4>&,int&>(const halide_type_t &,Halide::Runtime::Buffer<void,4> &,int &)' being compiled
[...]

Issue #3747, opened 2019-03-21
How to link runtime into shared library?


trivialfis:
Originally the question was posted on [SO](https://stackoverflow.com/questions/55234614/how-to-generate-position-independent-code-in-halide-with-ahead-of-time-compilati).  I failed to link AOT generated object into a shared object.  I extracted a minimum example:
```c++
// file: impl.cc
#include <Halide.h>
#include <vector>

namespace ft {

class Test : public Halide::Generator<Test> {
 public:
  Test() : vars(4) {}

  Input<Halide::Buffer<>> lhs{"lhs"};
  Input<Halide::Buffer<>> rhs{"rhs"};
  Output<Halide::Buffer<>> out{"out"};

  std::vector<Var> vars;

  void generate() {
    out(vars) = lhs(vars) + rhs(vars);
  }
};

}  // namespace ft

HALIDE_REGISTER_GENERATOR(ft::Test, test_generator)
```

``` c++
// file: lib.cc
#include <Halide.h>
#include "test_generator.h"

void test() {
  Halide::Runtime::Buffer<uint8_t> input(640, 480);
  test_generator(input, input, input);
}

```

``` makefile
# file: makefile
all:
	clang++ -fno-rtti -o impl.generator impl.cc GenGen.cpp -fpie -L/usr/local/bin -lHalide
	LD_LIBRARY_PATH=/usr/local/bin ./impl.generator \
		-e cpp,o,h -g test_generator \
		-o . target=x86-64-linux  \
		lhs.type=float32 lhs.dim=4 rhs.type=float32 \
		rhs.dim=4 out.type=float32
	LD_LIBRARY_PATH=/usr/local/bin ./impl.generator \
		-e cpp,o,h,bitcode,registration \
		-g test_generator \
		-o . \
		-r halide_runtime_x86-64-linux target=x86-64-linux \
		lhs.type=float32 lhs.dim=4 rhs.type=float32 rhs.dim=4 out.type=float32
	clang++ -c lib.cc
	clang++ -g  -o lib.so lib.o test_generator.o -L/usr/local/bin/ -lHalide -pthread -ldl -shared -fuse-ld=lld

clean:
	-rm *.bc *.o *.s *.so
	-rm impl.generator
	-rm *.registration.cpp
	-rm test_generator.cpp
	-rm test_generator.h
```

Running make with above example results in:
```
[...]

PR #4060, opened 2019-07-28
CMake build script for apps/autoscheduler

ghost-luck:
CMake version of Makefile.

In order to compile demo application under modern MSVC compilers, we need to make a change of `tools/RunGen.h`. This is closely related to issue #3507 

If I understood [2326(i)](https://cplusplus.github.io/LWG/issue2326) correctly `uniform_int_distribution<char>` and `uniform_int_distribution<unsigned char>` will be acceptable again in the future. By now MSVC 2015...2019 stops with an error on these lines.


steven-johnson:
FWIW, C++11 says that uniform_int_distribution<T> is UB for T in char or unsigned char; MSVC may 'fix' this but IMHO we'd be better off just adapting the code to do something that isn't UB

steven-johnson:
See https://github.com/halide/Halide/pull/4061 for attempted separate fix for UB in RunGen.h

steven-johnson:
Also see https://github.com/halide/Halide/pull/4063; the 'test' target for the apps/autoscheduler Makefile needed to do more testing, so a similar change might be needed here

ghost-luck:
Microsoft defines `int8_t` as `signed char` so I am made a little addition to `tools\RunGen.h`.

ghost-luck:
During Linux testing of this PR, I discovered several problems. Therefore, it is better not to merge before I figure them out.

1. There is a problem similar to #3747:

When I try to link `libauto_schedule` library under Linux (no matter with Makefile or CMake) I got the following error:
```
/usr/bin/ld: ../.././genfiles/halide_rt_host/halide_rt_host.a(halide_rt_host.a.o): relocation R_X86_64_PC32 against symbol `_ZN6Halide7Runtime8Internal13custom_mallocE' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
apps/autoscheduler/CMakeFiles/libauto_schedule.dir/build.make:336: recipe for target 'lib/liblibauto_schedule.so' failed
```

Since that library should be linked trouble-free, the problem is in my setup. Up to this point, I did not do manual builds of Halide under Linux (I had used buildbot builds). Now I build with `gcc 7.4.0` and `LLVM 8.0.1`. At a time when Travis-CI uses `gcc 5.4.0` and `LLVM 8.0.0`.

2. I just noticed that all libraries in Linux build have a `lib` prefix. I will need to remove the extra prefix.

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
...
+                              GENERATOR
+                              train_cost_model.generator)
+target_link_libraries(train_cost_model_process
+                      PRIVATE cost_model train_cost_model)
+add_dependencies(train_cost_model_process cost_model train_cost_model)
+
+# =======================================================
+
+# libauto_schedule
+add_library(libauto_schedule
```
steven-johnson: In theory we should need to somehow specify the equivalent of USE_EXPORT_DYNAMIC here (ie, `-rdynamic` for linux, `-undefined dynamic_lookup` for osx)... except I'm not sure if such a thing exists for Windows. Does this actually link?

ghost-luck:
Although the project builds without problems with the `HALIDE_SHARED_LIBRARY=ON` option, there are still some problems with `HALIDE_SHARED_LIBRARY=OFF` configuration. I will try to fix them soon.

ghost-luck:
Both `ON` and `OFF` options of `HALIDE_SHARED_LIBRARY` builds fine now under Linux.

Windows version of `retrain_cost_model`, on the other hand, is broken since there is no `getopt.h`.

steven-johnson:
> there is no getopt.h

There is a BSD-licensed getopt port available here: https://github.com/kimgr/getopt_port -- I'd be amenable to bringing it over so we can just rely on using getopt everywhere instead of perpetually rolling our own flag handling.

steven-johnson:
See also https://github.com/halide/Halide/pull/4200

steven-johnson:
https://github.com/halide/Halide/pull/4200 removed the usage of getopt(); if you sync to master you should be closer to getting this working

ghost-luck:
Looks ready to merge in my opinion.

steven-johnson:
I'll pull and test locally.

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+               DefaultCostModel.cpp
+               Weights.cpp
+               retrain_cost_model.cpp
+               ${WF_CPP})
+set_target_properties(retrain_cost_model_process
+                      PROPERTIES OUTPUT_NAME retrain_cost_model)
+
+set(COST_MODEL_GENERATOR_FILES cost_model_generator.cpp)
+halide_generator(cost_model.generator SRCS ${COST_MODEL_GENERATOR_FILES})
+halide_generator(train_cost_model.generator SRCS ${COST_MODEL_GENERATOR_FILES})
```
steven-johnson: You shouldn't need to make a separate halide_generator for these; you should be able to just define cost_model.generator and invoke it with a separate GENERATOR argument (this is what we do in autoscheduler.inc)

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+halide_library_from_generator(cost_model
+                              FUNCTION_NAME
```
steven-johnson: Style nit: prefer ARGNAME and argvalue on the same line, which is how we do this elsewhere, e.g.

```
halide_library_from_generator(train_cost_model
                              FUNCTION_NAME train_cost_model
                              GENERATOR train_cost_model.generator)
```

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+if(HALIDE_LIBRARY_TYPE STREQUAL STATIC)
+  target_link_libraries(auto_schedule PRIVATE cost_model train_cost_model)
+else()
+  target_link_libraries(auto_schedule
+                        PRIVATE cost_model train_cost_model Halide)
+endif()
+
+if(NOT MSVC)
+  set_target_properties(auto_schedule PROPERTIES LINK_FLAGS "-rdynamic")
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL Clang)
```
steven-johnson: I think `-undefined dynamic_lookup` is specific to OSX, not Clang, which would mean this would break for someone using Clang on Linux

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+# demo_apps_autoscheduler
+halide_generator(demo.generator SRCS demo_generator.cpp)
+halide_library_from_generator(demo
+                              FUNCTION_NAME
+                              demo
+                              GENERATOR
+                              demo.generator
+                              GENERATOR_ARGS
+                              auto_schedule=true
+                              -p
```
steven-johnson: collapse onto one line: `-p $<TARGET_FILE:auto_schedule>`

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,169 @@
...
+                   COMMAND demo_apps_autoscheduler
+                           --benchmarks=all
+                           --benchmark_min_time=1
+                           --estimate_all
+                   WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+
+# =======================================================
+
+# included_schedule_file
+halide_generator(included_schedule_file.generator SRCS
```
steven-johnson: Nit: for generators that only get executed to form a single output, it's terser and cleaner to use a single `halide_library` rule, rather than separate `halide_generator` + `halide_library_from_generator` rules

apps/support/cmdline.h
```diff
@@ -106,11 +118,24 @@ Target lexical_cast(const Source &arg)
 
 static inline std::string demangle(const std::string &name)
 {
+#if !defined(_WIN32)
```
steven-johnson: Thanks for the Win32 fix. Looking over the broader use of this, I'm a little skeptical as to whether this demangling stuff actually pays for itself in the first place (I had forgotten this library does this stuff). But as long as it works, hey, why not :-)

apps/support/cmdline.h
```diff
@@ -106,11 +118,24 @@ Target lexical_cast(const Source &arg)
...
+#else
+  char demangled_name[8192];
+  if (UnDecorateSymbolName(name.c_str(), demangled_name, sizeof(demangled_name),
+                           UNDNAME_COMPLETE)) {
+    std::string ret(demangled_name);
+    return ret;
+  } else {
+    DWORD error = GetLastError();
+    std::cout << "UnDecorateSymbolName error: " << error << std::endl;
+	return name;
```
steven-johnson: Nit: indentation

src/CMakeLists.txt
```diff
@@ -175,7 +175,7 @@ function(add_runtime_modules TARGET_RUNTIME_FILES TARGET_RUNTIME_DIR)
       add_custom_command(OUTPUT "${LL_D}"
                          # COMMENT "${SOURCE} -> ${LL_D}"
                          DEPENDS "${SOURCE}"
-                         COMMAND ${CLANG} ${CXX_WARNING_FLAGS} ${RUNTIME_DEBUG_FLAG} -DDEBUG_RUNTIME -O3 -fno-ms-compatibility -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables -std=gnu++98 -m${j} -target "${TARGET}" "-I${TARGET_RUNTIME_DIR}" -DCOMPILING_HALIDE_RUNTIME "-DLLVM_VERSION=${LLVM_VERSION}" -DBITS_${j} -emit-llvm -S "${SOURCE}" -o "${LL_D}"
+                         COMMAND ${CLANG} ${CXX_WARNING_FLAGS} ${RUNTIME_DEBUG_FLAG} -DDEBUG_RUNTIME -O3 -fno-vectorize -fpic -fno-ms-compatibility -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables -std=gnu++98 -m${j} -target "${TARGET}" "-I${TARGET_RUNTIME_DIR}" -DCOMPILING_HALIDE_RUNTIME "-DLLVM_VERSION=${LLVM_VERSION}" -DBITS_${j} -emit-llvm -S "${SOURCE}" -o "${LL_D}"
```
steven-johnson: Can you expand on why the new compiler flags are needed here now? They don't seem to be obviously related to the other changes here.

steven-johnson:
Failing in Win32 builds with

```
train_cost_model.lib(train_cost_model.lib.obj) : error LNK2005: _ceil_f64 already defined in cost_model.lib(cost_model.lib.obj) [D:\build_bot\worker\win-32-testbranch-800\halide-build-Release\apps\autoscheduler\retrain_cost_model_process.vcxproj]
```

apps/autoscheduler/CMakeLists.txt
ghost-luck: It seems to be impossible to do this without changing the `halide.cmake` file. Currently, `GENERATOR_NAME` is not a parameter but is a property. We can look at the example of the `linear_algebra/blas_generators` where are also creating a lot of generators.

src/CMakeLists.txt
ghost-luck: I had linking problems of `libauto_schedule` without `-fpic`. I add `-fno-vectorize` flag here to match Makefile flags. Most likely `-fno-vectorize` flag can be omitted. But I have not tested this.

ghost-luck:
> Failing in Win32 builds with
> 
> `train_cost_model.lib(train_cost_model.lib.obj) : error LNK2005: _ceil_f64 already defined in cost_model.lib(cost_model.lib.obj) [D:\build_bot\worker\win-32-testbranch-800\halide-build-Release\apps\autoscheduler\retrain_cost_model_process.vcxproj]`

Is a full log available?

apps/autoscheduler/CMakeLists.txt
steven-johnson: Yep, you are correct -- my bad here. I'll ponder whether making it configurable makes sense.

steven-johnson:
> P.S. Most likely it was this build: https://buildbot.halide-lang.org/master/#/builders/13/builds/316

that looks like the one

ghost-luck:
I added a linker option to support the win32 version.

But perhaps it’s worth creating a separate issue. In the case of win32, LLVM adds math functions to the symbol table.
https://github.com/halide/Halide/blob/d1eb604163e336d4407d0a2ff3897d3c2bc6c3c8/src/runtime/win32_math.ll#L92
https://github.com/halide/Halide/blob/d1eb604163e336d4407d0a2ff3897d3c2bc6c3c8/src/runtime/win32_math.ll#L106
etc.

As a result, we get a conflict if linking object files from two generators at the same time.

apps/autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,157 @@
...
+                              GENERATOR train_cost_model.generator)
+if(MSVC AND (CMAKE_SIZEOF_VOID_P EQUAL 4))
+  set_target_properties(retrain_cost_model_process
+                        PROPERTIES LINK_FLAGS "/FORCE:MULTIPLE")
+endif()
+target_include_directories(retrain_cost_model_process
+                           PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../support)
+target_link_libraries(retrain_cost_model_process
+                      PRIVATE cost_model train_cost_model)
+add_dependencies(retrain_cost_model_process cost_model train_cost_model)
```
steven-johnson: This should be unnecessary; IIUC, `target_link_libraries` implicitly does `add_dependencies`. (I don't think this is causing the Win32 problem we are seeing, but it still shouldn't be necessary, here and a few other places)

steven-johnson:
OK, I think the change in https://github.com/halide/Halide/pull/4238 is what we need for the Win32 issue (it sould make your last change (adding /FORCE:INLINE) undesirable and needing to be backed out).

PR #4063, opened 2019-07-29
Add apps/autoscheduler to test_apps target

steven-johnson:
Also augment the test target to ensure that there is basic coverage for single-shot and looped autoscheduling


steven-johnson:
The Mac buildbot needs to have `gtimeout` installed, but Chromote is unhappy right now; this will have to wait

PR #4064, opened 2019-07-29
Zero out 'other' buffer in Buffer's move ctor/op=

steven-johnson:
move is allowed to leave the moved-from item in an unspecified state; we zero out the device fields but not the host fields, so code downstream could inadvertently use a 'stale' moved-from Buffer in a flaky way. This PR is an experiment to see if fully zeroing out the moved-from Buffer's halide_buffer_t will inject any failures.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -675,9 +675,8 @@ public:
                                    dev_ref_count(other.dev_ref_count) {
         other.dev_ref_count = nullptr;
         other.alloc = nullptr;
-        other.buf.device = 0;
-        other.buf.device_interface = nullptr;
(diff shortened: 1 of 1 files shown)


steven-johnson:
Basically, the failure mode is code like this:

        Buffer<int> a(16, 16);
        a.fill(42);

        Buffer<int> b = std::move(a); // a.host still points to b.host but isn't holding a reference to it

        b = Buffer<int>(32, 32);  // a.host now points to deallocated memory

        assert(a.all_equal(42));  // Flaky Badness happens here, since a.host is a bad pointer

This PR merely tries to upgrade the failure from "flaky" to "predictably crashy" (since ~any attempt to use `a` will likely deref a null pointer)

PR #4065, opened 2019-07-29
Fix Makefile for LLVM 7.1

longjon:
As explained in the original text here, LLVM versions usually read like `major.zero.patch`, and this fact is used in the `sed` expression in the `Makefile`.

Unfortunately there is now LLVM 7.1 which breaks this, even though Halide seems to build fine. This patch simply adjusts the comment and `sed` expression accordingly. `LLVM_VERSION` is left as `70` in this case; alternatively, it could be `71` and explicitly checked below.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,11 +55,12 @@ WASM_SHELL ?= d8
 PREFIX ?= /usr/local
 LLVM_CONFIG ?= llvm-config
 LLVM_COMPONENTS= $(shell $(LLVM_CONFIG) --components)
-# Since LLVM5, the LLVM version always uses a major.zero.patch format
-# (ie there is no 'minor' version). We ignore the patch version and
-# just use the major version. For historical reasons, we continue to multiply
-# by 10, to avoid changing all the ifdefs in our code.
-LLVM_VERSION = $(shell $(LLVM_CONFIG) --version | sed 's/\([0-9][0-9]*\)\.0.*/\10/')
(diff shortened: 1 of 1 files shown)


steven-johnson:
Ouch, I didn't know that version existed -- I don't see it on http://releases.llvm.org, where are you finding it?

Thanks for this PR, but I think just reverting https://github.com/halide/Halide/pull/4022 is a better solution (since that would also correct the issue in CMake, which this PR doesn't consider)

longjon:
> I don't see it on http://releases.llvm.org, where are you finding it?

I see it as the second line on that page. (In fact, I notice that the version before #4022 explicitly checks for `71` as `LLVM_VERSION_TIMES_10` on line 2063!)

Anyway proceed as you see fit, the reversion looks good for the `Makefile` but I'm not so familiar with the CMake syntax.

steven-johnson:
That second line is 7.0.1, not 7.1.0.

longjon:
> That second line is 7.0.1, not 7.1.0.

Here is what I see:

Screenshot from 2019-07-29 15-31-59

steven-johnson:
Interesting, here's what I see (on OSX Chrome):

Screen Shot 2019-07-29 at 4 44 51 PM

Maybe 7.1 is specific to (say) Linux?

steven-johnson:
Reverting as https://github.com/halide/Halide/pull/4066

steven-johnson:
Should be obsoleted by #4066

PR #4073, opened 2019-07-30
Rename Python's "in" methods to "in_", so that calls are not a syntax error

longjon:
The Python bindings expose `Func::in` and `ImageParam::in`, but these are currently not callable, because code like `f.in(g)` is not syntactical Python. (Technically, `in` can be called with circumlocutions like `getattr(f, 'in')(g)`.)

This patch suggests a fix by renaming `in` to `in_`, a common if somewhat inelegant way to deal with keyword conflicts. (Other suggestions welcome.)

The readme gets updated to reflect this change.

### DIFF
--- a/python_bindings/readme.md
+++ b/python_bindings/readme.md
@@ -20,6 +20,7 @@ in Python, only the second variant is provided.
 - No mechanism is provided for overriding any runtime functions from Python.
 - No mechanism is provided for supporting `Func::define_extern`.
 - `Buffer::for_each_value()` is hard to implement well in Python; it's omitted entirely for now.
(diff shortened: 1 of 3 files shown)


steven-johnson:
Clearly, we need a test for this in the python_bindings tests; could you add one as part of this PR?

I'm also not wild about `in_`, but it will do as a stopgap until a better suggestion is made.

steven-johnson:
(LGTM once a test is added)

longjon:
(Will update with a test, most likely next week.)

steven-johnson:
Gentle ping. Would love to land once test is there :-)

steven-johnson:
Hi again, would love to land this once a test is added.

steven-johnson:
This change ended up being done in https://github.com/halide/Halide/pull/4356. Closing.

PR #4075, opened 2019-07-31
Don't declare all targets unsupported when LLVM < 9 is built with wasm

longjon:
Oddly(?), #3732 added a compiler-conditional `bad = true` line to `Target::supported`. When `llvm-config` claims webassembly support, and the LLVM version is less than 9, `Target::supported` returns `false` for _all_ targets.

This does not seem right; in particular, the Python tests fail, because they check that the host target is supported.

This patch consolidates this case with the existing wasm check, simplifying the logic and declaring wasm targets but no others unsupported for LLVM < 9. This fixes the Python target tests for me with LLVM 7.1.

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -626,7 +626,10 @@ bool Target::supported() const {
 #if !defined(WITH_HEXAGON)
     bad |= arch == Target::Hexagon;
 #endif
-#if !defined(WITH_WEBASSEMBLY)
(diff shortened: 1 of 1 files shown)


steven-johnson:
Ouch. Another good catch. Thanks for the fix.

steven-johnson:
Testing via https://github.com/halide/Halide/pull/4076

Issue #3454, opened 2018-11-27
The API for setting estimates for auto-schedule has inconsistent naming


steven-johnson:
e.g., from local_laplacian:

```
        input.dim(0).set_bounds_estimate(0, 1536);
        input.dim(1).set_bounds_estimate(0, 2560);
        input.dim(2).set_bounds_estimate(0, 3);
        // Provide estimates on the parameters
        levels.set_estimate(8);
        alpha.set_estimate(1);
        beta.set_estimate(1);
        // Provide estimates on the pipeline output
        output.estimate(x, 0, 1536)
              .estimate(y, 0, 2560)
              .estimate(c, 0, 3);
```

Why do we have three slightly different names? Can we / should we settle on a single convention?

abadams:
I think it was mirroring .bound and .set_bounds. I'd also prefer if it were consistent.

steven-johnson:
Bumping this again because the status quo is needlessly weird; if no one else wants to fix this I'll deal with it soon.

PR #4078, opened 2019-07-31
Rationalize API for setting autoscheduler estimates.

steven-johnson:
Fix https://github.com/halide/Halide/issues/3454:
- settles on `set_estimate()` as the one preferred API name; deprecate `estimate()` and `set_bounds_estimate()`
- deprecate `min_estimate()` and `extent_estimate()`, since they are unused and there weren't equivalent getters for ImageParam, Param, etc.
- Add `Func::set_estimates()` and `ImageParam::set_estimates()`, plus the equivalent Generator wrappers; these take a vector of Expr pairs, to allow setting the estimates for all dimensions at once. (In addition to being terser, this allows a quick fail if you don't set all of the dimensions.)
- Convert all extant use to use the new APIs; prefer to use `set_estimates()` rather than repeated calls to `set_estimate()` for reasons given above.

### DIFF
--- a/src/Func.h
+++ b/src/Func.h
@@ -1465,7 +1465,16 @@ public:
      * generated schedules might break when the sizes of the dimensions are
      * very different from the estimates specified. These estimates are used
      * only by the auto scheduler if the function is a pipeline output. */
-    Func &estimate(Var var, Expr min, Expr extent);
+    Func &set_estimate(Var var, Expr min, Expr extent);
+
+    HALIDE_ATTRIBUTE_DEPRECATED("Use set_estimate() instead")
+    Func &estimate(Var var, Expr min, Expr extent) { return set_estimate(var, min, extent); }
+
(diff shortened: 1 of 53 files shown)


src/Func.h
```diff
@@ -1465,7 +1465,16 @@ class Func {
      * generated schedules might break when the sizes of the dimensions are
      * very different from the estimates specified. These estimates are used
      * only by the auto scheduler if the function is a pipeline output. */
-    Func &estimate(Var var, Expr min, Expr extent);
+    Func &set_estimate(Var var, Expr min, Expr extent);
+
+    //HALIDE_ATTRIBUTE_DEPRECATED("Use set_estimate() instead")
```
abadams: I assume this is just commented out to be sure you catch all the cases in the initial branch?

steven-johnson: This is a draft PR. :-)

steven-johnson:
Ready for review. Any comments on the subsequent cleanups?

PR #4084, opened 2019-08-01
Build benchmark apps in parallel (run in serial)

ngzhian:
See https://www.gnu.org/software/make/manual/make.html#Phony-Targets for
an explanation of this idea.


steven-johnson:
Cloned into https://github.com/halide/Halide/pull/4085 for testing.

steven-johnson:
Thanks for the fix!

PR #4087, opened 2019-08-01
fix simd op check hvx

pranavb-ca:
simd_op_check failures fixed here

valign/vlalign tests need the min of the innermost dimension set to meaningfully reason about alignment
shift tests were failing because we no longer support signed shift amounts.


test/correctness/simd_op_check.cpp
```diff
@@ -195,6 +195,7 @@ struct Test {
         for (auto p : image_params) {
             p.reset();
             p.set_host_alignment(128);
+            p.dim(0).set_min(0);
```
steven-johnson: Why is this additional constraint necessary?

pranavb-ca: This additional constraint is necessary for testing 'valign' that relies on Halide being able to figure out the alignment of addresses in loads/stores. Without setting the min to a value like this, Halide cannot figure out the alignment of loads and stores.

pranavb-ca:
@dsharletg  & @steven-johnson  - Yes, the failing tests are from adding the min constraint. I didn't get a chance to revisit but I should try to find a way to fix this this week. Will post patches shortly.

steven-johnson:
What's the status on this PR?

pranavb-ca:
@steven-johnson  - This is mostly ready except, I need to fix up the apps side drive (apps/simd_op_check) to be aware of the splitting up of simd_op_check.cpp. I am OOO until Wednesday and hope to have something by the end of this week (Sept 13).

steven-johnson:
(there are some spurious python-related failures in your presubmits that should get fixed if you re-sync to master.)

pranavb-ca:
@steven-johnson and @dsharletg  - This is now ready for review.

pranavb-ca:
@steven-johnson  - I really couldn't tell why buildbot/win-64-testbranch-800 is failing. the correctness tests step is failing but I couldn't find a failing test.

steven-johnson:
Maybe this line?

```
d:\build_bot\worker\win-64-testbranch-800\halide\test\correctness\simd_op_check.h(6): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory [D:\build_bot\worker\win-64-testbranch-800\halide-build-Release\test\correctness_simd_op_check_hvx.vcxproj]
```

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -0,0 +1,664 @@
+// width and height of test images
```
steven-johnson: This comment makes no sense

test/correctness/simd_op_check.cpp
```diff
@@ -1,13 +1,12 @@
 #include "Halide.h"
+#include "simd_op_check.h"
```
steven-johnson: I presume this refactoring was done so you could have a new simd_op_check_hvx.cpp test, but it's not clear to me why it's a separate test now -- can you add some comment to this PR about why that's desirable?

test/correctness/simd_op_check.h
```diff
@@ -0,0 +1,301 @@
+#ifndef SIMD_OP_CHECK_H
+#define SIMD_OP_CHECK_H
+
+#include "Halide.h"
+#include "test/common/halide_test_dirs.h"
+#include "sys/time.h"
+#include <fstream>
```
steven-johnson: all <system> includes should go before any "non-system" includes.

test/correctness/simd_op_check.h
```diff
@@ -0,0 +1,301 @@
+#ifndef SIMD_OP_CHECK_H
+#define SIMD_OP_CHECK_H
+
+#include "Halide.h"
+#include "test/common/halide_test_dirs.h"
+#include "sys/time.h"
```
steven-johnson: (1) Should be <sys/time.h>
(2) Isn't available on Win32 systems. What do you need from this file?

test/correctness/simd_op_check.h
```diff
@@ -0,0 +1,301 @@
...
+#include <fstream>
+
+using namespace Halide;
+using namespace Halide::ConciseCasts;
+
+//namespace {
```
steven-johnson: (1) anonymous namespaces should ~never go in .h files
(2) commented-out code should never be left in checked-in code (unless there's a comment explaining why)

I presume this and the matching closure later were left in by accident

test/correctness/simd_op_check.h
```diff
@@ -0,0 +1,301 @@
...
+};
+
+struct Task {
+    string op;
+    string name;
+    int vector_width;
+    Expr expr;
+};
+
+size_t num_threads = Halide::Internal::ThreadPool<void>::num_processors_online();
```
steven-johnson: Declaring a global in a .h file in this manner is ~never OK

test/correctness/simd_op_check.h
```diff
@@ -0,0 +1,301 @@
...
+#include "sys/time.h"
+#include <fstream>
+
+using namespace Halide;
+using namespace Halide::ConciseCasts;
+
+//namespace {
+
+// This tests that we can correctly generate all the simd ops
+using std::vector;
```
steven-johnson: You should ~never put `using` statements in the toplevel namespace in a .h file like this

test/correctness/simd_op_check.h
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        return success;
+    }
+};
+
+    //}  // namespace
+
+
+#endif
```
steven-johnson: It's good form to have a comment indicating the matching ifdef , ie `#endif  // SIMD_OP_CHECK_H`

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -0,0 +1,664 @@
...
+        check("vunpack(v*.b)", hvx_width/1, i16(i8_1));
+        check("vunpack(v*.h)", hvx_width/2, u32(i16_1));
+        check("vunpack(v*.h)", hvx_width/2, i32(i16_1));
+
+        check("vunpack(v*.ub)", hvx_width/1, u32(u8_1));
+        check("vunpack(v*.ub)", hvx_width/1, i32(u8_1));
+        check("vunpack(v*.b)", hvx_width/1, u32(i8_1));
+        check("vunpack(v*.b)", hvx_width/1, i32(i8_1));
+
+#if 0
```
steven-johnson: I realize that the `#if 0` sections already existed, but now seems like a good time to open Issues for them (if they actually need to get dealt with eventually) or just deleted (if they don't)

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -0,0 +1,664 @@
...
+        check("vcl0(v*.uh)", hvx_width/2, count_leading_zeros(u16_1));
+        check("vcl0(v*.uw)", hvx_width/4, count_leading_zeros(u32_1));
+        check("vnormamt(v*.h)", hvx_width/2, max(count_leading_zeros(i16_1), count_leading_zeros(~i16_1)));
+        check("vnormamt(v*.w)", hvx_width/4, max(count_leading_zeros(i32_1), count_leading_zeros(~i32_1)));
+        check("vpopcount(v*.h)", hvx_width/2, popcount(u16_1));
+    }
+};
+
+int main(int argc, char **argv) {
+
```
steven-johnson: No blank line between opening brace and first statement in a function

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -0,0 +1,664 @@
...
+    Target host = get_host_target();
+    Target hl_target = get_target_from_environment();
+    printf("host is:      %s\n", host.to_string().c_str());
+    printf("HL_TARGET is: %s\n", hl_target.to_string().c_str());
+
+    Target t(Target::NoOS, Target::Hexagon, 32);
+    for (const auto &f : {Target::HVX_64, Target::HVX_128,
```
steven-johnson: style nit: weird indentation of the stuff in braces, make regular somehow (maybe one per line?)

test/correctness/simd_op_check.cpp
pranavb-ca: The Halide buildbots run correctness tests by setting ```HL_TARGET``` to the same value as ```HL_JIT_TARGET```. When testing hexagon, they are host-hvx_128. This works for all correctness tests except simd_op_check.cpp which is not an 'offload' tests at all - hexagon tests are run if the host processor is hexagon. In other words, it'll run hexagon tests only if ```HL_TARGET``` has a hexagon host arch like ```hexagon-32-noos-hvx_128```.

As a result the buildbot never really tested HVX correctly and over time we have discovered a number of failures in simd_op_check that the buildbots didn't catch simply because they weren't testing simd_op_check with the right target for HVX.

One solution would be to change the buildbot recipe but after I looked at the buildbot recipe, I rejected that idea because it would mean one exception for one test on architecture. Instead, I decided to refactor this way, so HVX is free to do its own thing and the buildbot can keep its model of ```HL_TARGET=HL_JIT_TARGET```

test/correctness/simd_op_check.h
pranavb-ca: Yikes! remnants from an older version of this branch. I don't need this. Will remove.

test/correctness/simd_op_check.h
pranavb-ca: Fixed.

test/correctness/simd_op_check.h
pranavb-ca: Fixed.

pranavb-ca:
@steven-johnson : Thanks for your review. Updated PR. PTAL.
linux-64-gcc53-testbranch-trunk is failing test_apps because of x86 isel issues. Unrelated to this PR.

test/correctness/simd_op_check.cpp
steven-johnson: Could you add this detail as a comment in simd_op_check_hvx.cpp, so that future readers won't be tempted to recombine them without realizing this?

steven-johnson:
LGTM pending the one change about inserting a comment.

pranavb-ca:
@steven-johnson : you mean the comment about opening issues for the ```#if 0``` part?

pranavb-ca:
or why we have a separate simd_op_check_hvx.cpp? Sorry, I may have misunderstood and put the comment in the PR conversation.

steven-johnson:
> or why we have a separate simd_op_check_hvx.cpp? 

This one.

test/correctness/simd_op_check.cpp
pranavb-ca: Done.

test/correctness/simd_op_check_hvx.cpp
pranavb-ca: Done

test/correctness/simd_op_check_hvx.cpp
pranavb-ca: Done

pranavb-ca:
@steven-johnson  - PTAL.

PR #4090, opened 2019-08-05
Avoid testing expressions that exercise UB

abadams:


steven-johnson:
mac failure is unrelated

PR #4092, opened 2019-08-05
Allow Input<SomeType*>::set_estimate()

steven-johnson:
Previouslty this didn't compile. Add a wrapper that allows it to be set (but only allow nullptr as the estimate value for now).

Also, drive-by add of set_estimates calls to two Generators.


abadams:
Seems like at odd thing to enable. Is this to support generic code?

steven-johnson:
The autoscheduler requires that estimates be set on all inputs.

PR #4094, opened 2019-08-06
Upgrade autogenerated schedule outputs into legal C++ headers

steven-johnson:
Currently, autogenerated `.schedule` files are C++ fragments that must be copy/pasted in order to be useful; this is error prone and awkward, especially for novice users trying to use the Halide autoscheduler.

This PR changes schedule output to be a legal C++ header file; for the generator `Foo`, the file might look something like:

```

#ifndef FOO_SCHEDULE_H
#define FOO_SCHEDULE_H

// MACHINE GENERATED -- DO NOT EDIT
// This schedule was automatically generated by apps/autoscheduler/AutoSchedule
// for target=x86-64-osx-avx-avx2-f16c-fma-sse41
// with machine_params=16,16777216,40

#include "Halide.h"


inline void apply_schedule_Foo(
    ::Halide::Pipeline pipeline,
    ::Halide::Target target
) {
    using ::Halide::Func;
    using ::Halide::MemoryType;
    using ::Halide::RVar;
    using ::Halide::TailStrategy;
    using ::Halide::Var;

    Func relu = pipeline.get_func(4);
    Func conv = pipeline.get_func(3);
    Var c(relu.get_schedule().dims()[0].var);
    Var ci("ci");
    Var n(relu.get_schedule().dims()[3].var);
    Var x(relu.get_schedule().dims()[1].var);
    Var xi("xi");
    Var y(relu.get_schedule().dims()[2].var);
    Var yi("yi");
    RVar r4_x(conv.update(0).get_schedule().dims()[0].var);
    RVar r4_y(conv.update(0).get_schedule().dims()[1].var);
    RVar r4_z(conv.update(0).get_schedule().dims()[2].var);
    relu
        .split(x, x, xi, 2, TailStrategy::ShiftInwards)
        .split(c, c, ci, 8, TailStrategy::ShiftInwards)
        .split(y, y, yi, 4, TailStrategy::ShiftInwards)
        .unroll(xi)
        .unroll(yi)
        .vectorize(ci)
        .compute_root()
        .reorder(ci, xi, yi, c, y, x, n)
        .fuse(x, n, x)
        .parallel(x);
    conv.update(0)
        .split(c, c, ci, 8, TailStrategy::GuardWithIf)
        .unroll(x)
        .unroll(y)
        .vectorize(ci)
        .reorder(ci, c, x, y, n, r4_x, r4_y, r4_z);
    conv
        .store_in(MemoryType::Stack)
        .split(c, c, ci, 8, TailStrategy::ShiftInwards)
        .unroll(x)
        .unroll(y)
        .vectorize(ci)
        .compute_at(relu, c)
        .reorder(ci, c, x, y, n);

}

#endif  // FOO_SCHEDULE_H
```

(The Generator would then just #include this header and call the function, see code for example.)

The intended workflow here is that the coder would run the autoscheduler on a Generator via the appropriate commands, then *add this generated file to their source control* so that subsequent builds need not invoke the autoscheduler at all. (Of course, people who prefer to copy-n-paste could still use that approach.)

One nice thing about this is that all the machine-generated code becomes concentrated into its own file(s) -- if you intend to have some manual scheduling in addition to the autoscheduler, it becomes easier to avoid accidentally mixing them.

Another nice thing is that this cleanly produces autoscheduler output that is multitarget-friendly, so if you want to generate custom schedules for (say) AVX512 vs AVX2 vs AVX, you can get them (basically) for free (try it and see).

I think this is ready for review, but as is, this does have a few known warts:
- The chicken-and-egg need to build the Generator both with-and-without the #included schedule is... not a pleasant looking code pattern, and additionally, some build systems may find it awkward to support cleanly.
- Some people may also object to the prospect of adding machine-generated code to source control (rather than generating it every time).  I'd argue that this PR is strictly a step forward since previously the options were either manual copy-n-paste (which amounts to the same thing), or re-running the autoscheduler at every build with a stable set of weights (which seems like a nontrivial waste of build machine cycles, and also an avenue for possible performance changes/regressions as the autoscheduler and/or weight set are upgraded over time).
- Currently, the `apply_schedule_Foo` function always has `Foo` matching the generated function name; it may be desirable to allow custom names so that different variants can be experimented with more easily.


apps/autoscheduler/Makefile
```diff
@@ -51,3 +51,51 @@ clean:
...
+# which has been added to our local source control.
+$(BIN)/%/included_schedule_file.a: $(GENERATOR_BIN)/included_schedule_file.generator
+	@mkdir -p $(@D)
+	$< -g included_schedule_file -o $(@D) -f included_schedule_file target=$*
+
+$(BIN)/%/included_schedule_file.rungen: $(BIN)/%/RunGenMain.o $(BIN)/%/included_schedule_file.registration.cpp $(BIN)/%/included_schedule_file.a
+	@mkdir -p $(@D)
+	$(CXX) $(CXXFLAGS) -I$(BIN)/$* $^ -o $@ $(HALIDE_SYSTEM_LIBS) $(IMAGE_IO_FLAGS)
+
+# included_schedule_filenstrates single-shot use of the autoscheduler
```
abadams: filenstrates

src/Module.cpp
```diff
@@ -199,19 +200,91 @@ static Registerer registerer;
...
+#define $CLEANNAME$_SCHEDULE_H
+
+// MACHINE GENERATED -- DO NOT EDIT
+// This schedule was automatically generated by $SCHEDULER$
+// for target=$TARGET$
+// with machine_params=$MACHINEPARAMS$
+
+#include "Halide.h"
+
+$NSOPEN$
```
abadams: nit: had to think about this too long. NS->NAMESPACE

abadams:
Seems like a good change to me.

apps/autoscheduler/Makefile
steven-johnson: it's a perfectly cromulent word

src/Module.cpp
steven-johnson: done

PR #4102, opened 2019-08-06
Strip 'irrelevant' features from the target comment when generating schedule.h files

steven-johnson:


src/Module.cpp
```diff
@@ -247,6 +247,21 @@ inline void apply_schedule_$SHORTNAME$(
 #endif  // $CLEANNAME$_SCHEDULE_H
 )INLINE_CODE";
 
+    // For logging in the comment, strip out features that are almost
+    // certainly irrelevant to scheduling issues, to make for easier reading
+    const Target::Feature irrelevant_features[] = {
+        Target::CPlusPlusMangling,
+        Target::DisableLLVMLoopVectorize,
```
abadams: This is the target that this schedule was generated for, right?

If you generate a schedule via autotuning, then anything here that affects performance is relevant, and you'd want it to match the target you ultimately compile for. So I think we should remove anything affecting performance from this list, which just leaves CPlusPlusMangling, LegacyBufferWrappers, NoRuntime, UserContext

steven-johnson: This is true... except for DisableLLVMLoopVectorize & DisableLLVMLoopUnroll, which we set for all autotuning, so that really is just noise, isn't it? (Removing the others)

steven-johnson: (On a related note, we should probably move toward disabling those by default...)

steven-johnson: PTAL

abadams: I'm still uneasy about DisableLLVMLoopVectorize and DisableLLVMLoopUnroll. We enable those flags for autotuning because it makes the performance landscape smoother and easier to predict, but they do potentially change performance. Final use of the generated schedule should either leave them enabled too, or toggle them and see which is faster. Silently dropping them from the comment seems dangerous.

steven-johnson: Hmmm... yeah, I suppose that's fair. (The right answer in the long run is to set these by default and remove the features entirely, since our assumption has long been that these are likely to be usually be wasted effort on LLVM's part; I'll open an issue.)

PR #4103, opened 2019-08-07
Fix comment on FFT App Generator file

gasparitiago:
Fix the comment that documents the direction of the FFT on generator file.


gasparitiago:
It's just a fix on a comment in order to avoid mistakes on using the method.

abadams:
Yeah, it looks right to me, but I was hoping Dillon (@dsharletg) could double-check, in case there's anything weird about the API usage that we're missing.

abadams:
@dsharletg ping

PR #4104, opened 2019-08-07
[Hexagon] LUT32 implementation using lut16

aankit-ca:
Hi @dsharletg @pranavb-ca 
The PR is not complete but I'ld appreciate your comments on the changes. I'll add the lut32 to the existing lut testcases. For now test/correctness/gather.cpp without the hvx_v65 flag is good test for this PR .


src/CodeGen_Hexagon.cpp
```diff
@@ -1080,23 +1080,45 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
...
+        vector<Value *> indices;
+        Value *replicate_val = ConstantInt::get(idx->getType(), replicate);
+        Value *piece = builder->CreateMul(idx, replicate_val);
+        for (int i = 0; i < replicate; i++) {
+            Value *pos = ConstantInt::get(idx->getType(), i);
+            indices.emplace_back(builder->CreateAdd(piece, pos));
+        }
+        idx = interleave_vectors(indices);
+        min_index = replicate * min_index;
+        max_index = replicate * max_index + replicate - 1;
```
dsharletg: I find this easier to read/understand:

    (max_index + 1) * replicate - 1;

src/CodeGen_Hexagon.cpp
```diff
@@ -1080,23 +1080,45 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
...
 
-Value *CodeGen_Hexagon::vlut(Value *lut, Value *idx, int min_index, int max_index) {
+Value *CodeGen_Hexagon::vlut256(Value *lut, Value *idx, int min_index, int max_index) {
     bool is_128B = target.has_feature(Halide::Target::HVX_128);
     llvm::Type *lut_ty = lut->getType();
+    llvm::Type *result_ty = llvm::VectorType::get(lut_ty->getVectorElementType(),
+                                   idx->getType()->getVectorNumElements());
+
+    int replicate = lut_ty->getScalarSizeInBits()/16;
+    int lut_max = (replicate > 2) ? 256/replicate : 256;
```
dsharletg: Why does this not go down for replicate = 2?

I'd expect this to need to be:

    lut_max = 256 / replicate

Also, I think there is a way to do this without this reduced lut_max requirement. If we deinterleaved the LUT instead of interleaving the indices, we wouldn't have this constraint. But this approach might have more overhead? Either way, perhaps leave it as a TODO.

src/CodeGen_Hexagon.cpp
```diff
@@ -1395,6 +1411,77 @@ Value *CodeGen_Hexagon::vdelta(Value *lut, const vector<int> &indices) {
...
+        // Make a vector of the indices shifted such that the min of
+        // this range is at 0. Use 16-bit indices for this.
+        Value *min_index_i_val = create_vector(i16x_t, min_index_i);
+        Value *indices = builder->CreateSub(idx16, min_index_i_val);
+        // Upcast the indices if required to match the lut_elems_size as
+        // use_index mask element size should be same as lut_elem_size.
+        indices = builder->CreateZExtOrBitCast(indices, cmp_index_ty);
+
+        // Create a condition value for which elements of the range are valid
+        // for this index. We can't make a constant vector of <1024 x i1>, it
```
dsharletg: This comment re: constant 1024 x 1-bit vectors now seems misplaced.

src/CodeGen_Hexagon.cpp
```diff
@@ -1395,6 +1411,77 @@ Value *CodeGen_Hexagon::vdelta(Value *lut, const vector<int> &indices) {
...
+    return ConstantVector::getSplat(ty->getVectorNumElements(), value);
+}
+
+Value *CodeGen_Hexagon::vlut(Value *lut, Value *idx, int min_index, int max_index) {
+    const unsigned idx_elems_size = idx->getType()->getScalarSizeInBits();
+    const unsigned lut_elems_size = lut->getType()->getScalarSizeInBits();
+    const unsigned idx_elems = idx->getType()->getVectorNumElements();
+    llvm::Type *i8x_t  = VectorType::get(i8_t,  idx_elems);
+    llvm::Type *i16x_t = VectorType::get(i16_t, idx_elems);
+	llvm::Type *i32x_t = VectorType::get(i32_t, idx_elems);
```
dsharletg: Indentation is off

src/CodeGen_Hexagon.cpp
```diff
@@ -1395,6 +1411,77 @@ Value *CodeGen_Hexagon::vdelta(Value *lut, const vector<int> &indices) {
...
+
+    internal_assert(idx_elems_size <= 16)
+        << "Index element for lookup tables must be <= 16 bits in size.\n";
+
+    int replicate = lut->getType()->getScalarSizeInBits()/16;
+    const int lut_max = (replicate > 1) ? 256/replicate : 256;
```
dsharletg: 256/replicate. Also note this is different from the expression above.

src/CodeGen_Hexagon.cpp
```diff
@@ -1395,6 +1411,77 @@ Value *CodeGen_Hexagon::vdelta(Value *lut, const vector<int> &indices) {
...
+    const unsigned lut_elems_size = lut->getType()->getScalarSizeInBits();
+    const unsigned idx_elems = idx->getType()->getVectorNumElements();
+    llvm::Type *i8x_t  = VectorType::get(i8_t,  idx_elems);
+    llvm::Type *i16x_t = VectorType::get(i16_t, idx_elems);
+	llvm::Type *i32x_t = VectorType::get(i32_t, idx_elems);
+
+    internal_assert(idx_elems_size <= 16)
+        << "Index element for lookup tables must be <= 16 bits in size.\n";
+
+    int replicate = lut->getType()->getScalarSizeInBits()/16;
```
dsharletg: replicate = max(1, .)

would make things simpler (no special case at 0).

src/CodeGen_Hexagon.cpp
```diff
@@ -1080,23 +1080,45 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
...
 }
 
-Value *CodeGen_Hexagon::vlut(Value *lut, Value *idx, int min_index, int max_index) {
+Value *CodeGen_Hexagon::vlut256(Value *lut, Value *idx, int min_index, int max_index) {
     bool is_128B = target.has_feature(Halide::Target::HVX_128);
     llvm::Type *lut_ty = lut->getType();
+    llvm::Type *result_ty = llvm::VectorType::get(lut_ty->getVectorElementType(),
+                                   idx->getType()->getVectorNumElements());
+
+    int replicate = lut_ty->getScalarSizeInBits()/16;
```
dsharletg: max(1, .)

eliminates the need for a special case at 0.

aankit-ca:
Thanks Dillon for the comments. I've tried to address all the comments. Please review the patch once more

src/HexagonOptimize.cpp
```diff
@@ -1777,11 +1777,12 @@ class OptimizeShuffles : public IRMutator {
                 Expr index_span = span_of_bounds(index_bounds);
                 index_span = common_subexpression_elimination(index_span);
                 index_span = simplify(index_span);
-
-                if (can_prove(index_span < 256)) {
```
aankit-ca: What do you think would be a good number here? I don't think we need not restrict dynamic_shuffle to 256.

src/CodeGen_Hexagon.cpp
```diff
@@ -1395,6 +1384,94 @@ Value *CodeGen_Hexagon::vdelta(Value *lut, const vector<int> &indices) {
     return vlut(lut, indices);
 }
 
+Value *create_vector(llvm::Type *ty, int val) {
```
dsharletg: Make this `static`

src/HexagonOptimize.cpp
dsharletg: I think we should keep it 256 until we know a better value for it. There must be some crossover point at which it is not more efficient to use dynamic_shuffle, but I don't know what it is.

aankit-ca:
@dsharletg ping

dsharletg:
This looks good, but IIRC, we were nervous about relying on LLVM HVX instruction codegen (e.g. using `IRBuilder::Create*` vs. `call_intrin(...)`) on older LLVMs. Do you know what version of LLVM it was first safe to do this in? Alternatively, is there a version of LLVM you don't currently support? It would make sense for the Halide Hexagon backend to inherit the policy you use for the Hexagon LLVM backend.

Going forward: while I think it is a great change to use LLVM's IRBuilder instead of intrinsics where possible, it would be best not to make changes like that in unrelated PRs. I think we should make separate PRs that deliberately makes this change, and avoid making these changes in other feature PRs (like this one).

aankit-ca:
I reverted the IRBuilder changes to intrinsics in the code that was already there. However, I thought that using IRBuilder made sense atleast for all the new pieces of code but I can change that to use intrinsics too.
Also the PR does not specifically depend on any recent LLVM change and should work on older LLVM versions as well. However, I've  tested the PR only with the latest LLVM.

dsharletg:
A test appeared to fail on the build bots with this PR:

FAILED TEST: gather
Hexagon simulator executed function 0x5ba90 in 2000 cycles
Hexagon simulator executed function 0x60a90 in 2000 cycles
Hexagon simulator executed function 0x63850 in 4000 cycles
output(2, 0) = 1117274333 instead of 1198408891
../halide/Makefile:1855: recipe for target 'quiet_correctness_gather' failed

https://buildbot.halide-lang.org/master/#/builders/18/builds/613

aankit-ca:
@dsharletg I've removed the IRBuilder calls to use intrinsics instead.

aankit-ca:
@dsharletg It solves the failing correctness test as well

dsharletg:
#4358 passed, the ARM tests aren't relevant for this PR.

PR #4106, opened 2019-08-07
Wrappers need distinct names

abadams:


src/Func.cpp
```diff
@@ -1839,6 +1839,7 @@ Func get_wrapper(Function wrapped_fn, string wrapper_name, const vector<Func> &f
     // have any wrappers. Otherwise, throw an error. If 'fs' is empty, then
     // it is a global wrapper.
     const map<string, FunctionPtr> &wrappers = wrapped_fn.wrappers();
+    wrapper_name += ("$" + std::to_string(wrappers.size()));
```
steven-johnson: Is it possible that this could collide with a uniquified name, especially if wrappers.size() is tiny?

abadams:
I think it's safe. It's a suffix on a Func name that has already been uniqued, so it looks like: foo$3_wrapper$7. The stem is unique, so the wrappers just need to be numbered within that namespace.

PR #4107, opened 2019-08-07
Fix for trunk LLVM changes

steven-johnson:


abadams:
Hooray for race conditions. I propose we use #4105 , because there were a few other places I found that should use the wider type (they would silently narrow with a 64-bit offset)

abadams:
I'll steal your typedef name though...

Issue #4110, opened 2019-08-08
Crash in FunctionDAG.h


steven-johnson:
While trying to autoschedule a rather complicated pipeline, I got a crash in LoadJacobian::operator(); specifically on line 136:

```
        // coeffs[producer_storage_dim].size() == 0, so we are
        // accessing a nonexistent entry
        return coeffs[producer_storage_dim][consumer_loop_dim];
```

(The pipeline in question is proprietary and can't be shared publicly)

I'm unfamiliar with the semantics here; would it be reasonable to do something like:

```
    OptionalRational operator()(int producer_storage_dim, int consumer_loop_dim) const {
        if (coeffs.empty()) {
            // The producer is scalar, so all strides are zero.
            return {true, 0, 1};
        }
        internal_assert(producer_storage_dim < coeffs.size());
        const auto &p = coeffs[producer_storage_dim];
        if (p.empty()) {
            // The consumer is scalar, so all strides are zero.
            return {true, 0, 1};
        }
        internal_assert(consumer_loop_dim < p.size());
        return p[consumer_loop_dim];
    }
```

?

PR #4112, opened 2019-08-08
Avoid crash in FunctionDAG.h (Issue #4110)

steven-johnson:


steven-johnson:
(This fixes the crash I noted in #4110, but I may well be getting lucky here; an expert on this code should review)

abadams:
Pretty sure that's right.

PR #4117, opened 2019-08-08
PerfectHashMap.h should be standalone

steven-johnson:
We don't want it to require libHalide, but including Errors.h implicitly does that. Move the 'standalone' asserter into the template class and out of the test, so that the .h file will compile standalone (regardless of what is #defined before it).


abadams:
I thought it was standalone? At some point the dependency must have crept in.

steven-johnson:
yeah, that was my bad from yesterday, when I added Errors.h

Issue #4121, opened 2019-08-08
FR: consolidate the autoscheduler weights into a single file


steven-johnson:
Since it may be desirable to save a custom-tuned set of weights (from running autotune_loop), it would be nice to make this a single file instead of a group of files. (I assume that the set-of-files was much handier during prototyping, but at this point I suspect we could consolidate them).

(While we're on that topic, would it be sensible to consider making said file text instead of binary? It might make merging/diffing less painful.)

abadams:
They'll never be partially updated, so diffing merging shouldn't be an issue

steven-johnson:
> diffing merging shouldn't be an issue

yeah, I was thinking more in terms of being easy to tell that there is a difference when looking at diffs in commit-reviews. 

(text format is really just gravy; consolidation into a single file is what I'm curious about.)

PR #4123, opened 2019-08-09
Convert autoscheduler weights to a single file

steven-johnson:
The main gist of this change is to put all of the weights used by the autoscheduler into a single file (instead of a directory of multiple files); the goal is to make saving 'overnight' tuned weights for a pipeline simpler, via having only a single `.weights` file rather than a directory of things.

There are various drive-by fixes, of course:
- The 'Weights' struct moved into its own file, and now manages its own loading/saving, in both the classic dir-of-files format, and new, single-file format. (The single-file format is very simpleminded but likely more than adequate; it has a trivial signature at the front so some smart upgrading over time should be easy to do.)
- DefaultCostModel.cpp got its own .h file, instead of being incestuously connected to the ABC it implements
- existing `weights` folder was deleted; the equivalent data is now in `baseline.weights`
- added the `weightsdir_to_weightsfile` utility for any downstream consumers that have custom weights in the 'old' format that they need to update to the 'new' format.
- Removed some unused code here and there.

Note that HL_WIEGHTS_DIR (and HL_WEIGHTS_OUT_DIR) are still used, but are 'smart' for now; if you pass a pathname that ends in `.weights` it loads as a single file, otherwise it assumes an old-style directory. (I didn't try to smarten the env var names because we really need to eradicate the use of env vars for this purpose anyway. but that's a job for another day.)


steven-johnson:
(Issue #4121)

steven-johnson:
(If there is a better -- existing -- format for combining the weights into one file, I'm all ears. Only one that springs to mind that supports multiple float32 images is MultiTIFF, which seems vastly too heavyweight for our needs here.)

abadams:
Would be good to also include PipelineFeatures::version and ScheduleFeatures::version, and to warn on mismatch. If either the pipeline or schedule features get a version bump, the weights are invalid, but are probably still a good starting point for retraining in the number of features hasn't changed (hence warning but not error).

steven-johnson:
> include PipelineFeatures::version and ScheduleFeatures::version

Done. Also reworked code to be more resilient to failures when reading weight files, and improved some deps in the makefile. PTAL.

abadams:
Regarding nomenclature: compiled-in weights are reasonable to call baseline, because they can't be modified, but as a weights these are the trained x86 weights. Should we be calling them that? Or maybe baseline_x86?

steven-johnson:
> baseline_x86

More likely `x86-64-avx2`, since I suspect that was used (it's the default in autotune_loop.sh)... which (btw) now that I look at it might be suboptimal, since I don't think that including the AVX2 feature also includes AVX and SSE41 implicitly...

steven-johnson:
> compiled-in weights are reasonable to call baseline, because they can't be modified

FYI, the 'compiled-in-weights' were just an in-code copy of the old `weights` folder, which is now called `baseline.weights`, or possibly `baseline_x86-64-avx2.weights`, so....?

PR #4125, opened 2019-08-09
Re-introduce Bazel support

### LINKED ISSUE #4009: Add Bazel support back?
I want to use Halide with Bazel. I'm currently hacking on reverting the PRs that deleted support for it earlier this year.

Is there interested in merging this work back into this repository? I could also keep this separate without too much trouble.

If there is interest in porting the whole build system to Bazel and not just the library user side I could potentially help a bit with that too. I'm a Googler so could potentially help with open sourcing the internal Blaze setup. It seems like Windows support in Bazel has gotten a bit better in the last year. Building Halide with Blaze would be very useful for the project I'm using it in, since we will likely need to hack at least a bit in the Halide code base.

Details:
* I use Linux and am not very interested in working with Windows stuff, but if I could build all of Halide with Bazel maybe I can help there too.
* I don't fully understand how the old Bazel rules worked wrt LLVM; Unlike CMake, Bazel cannot build LLVM so it needs to be fed with a binary distribution of it. Linking directly against the LLVM libraries is a bit sensitive to which toolchain is used by the Bazel build. The official LLVM binary distribution for Ubuntu links against libstdc++, but the [Bazel compiler toolchain that seems most well supported](https://github.com/grailbio/bazel-toolchain) is hard coded to use libc++... The approach I am going for at the moment is to have my own custom built LLVM where the LLVM libraries are built against libc++ in order to be able to use the gailbio toolchain setup. All in all it just seems a bit tricky to get hermetic builds up and running in a user friendly way.

per-gron:
This PR resurrects the deleted Bazel support. This doesn't build Halide with Bazel, it just makes it possible to use Halide from projects that are built with Bazel, pointing Bazel to a binary build of Halide+LLVM.

In this PR, the commit that I'm the most unsure about is "Update halide.bzl to work with the Grail Bazel clang toolchain". I don't really understand how this code worked before, and I don't know if it was broken now because of changes in Bazel or just because `halide.bzl` did not work with the grailbio toolchain. Without that commit, whenever I try to build the example code, it would attempt to generate Android output.

Fixes #4009


steven-johnson:
In general, this is heading in the right direction, but it's likely to need more work before landing:
- How did we go from requiring Bazel 0.6 to Bazel 0.28? Surely that can't be correct?
- It's likely that halide.bzl will need some attention to work well, as the Generator command-line interface has morphed some since this stuff was deleted.
- I see that the readme says "linux only has been tested"; I'd really like for at least an attempt at testing on OSX and Windows be made, to see where we stand.
- I have no idea what "Grail Bazel" is or what the implications are. Can you point me at relevant docs?
- "LLVM needs to be built with the same C++ standard library as the one used by Bazel" -- not saying that's your fault, but that's an entirely unreasonable requirement. Why should the build system's library have any effect on what is being built?

per-gron:
> In general, this is heading in the right direction, but it's likely to need more work before landing:
> 
> * How did we go from requiring Bazel 0.6 to Bazel 0.28? Surely that can't be correct?

Is there any point in supporting old versions of Bazel? 0.28 is the current version and the one I tested with. Bazel is a fast moving system with frequent API breaks and users are in practice required to do the work to stay up to date. Bazel is not like CMake in this regard. I don't know when PACKAGE_NAME was removed and I don't think it's a good use of anyone's time to binary search back to find out.

> * It's likely that halide.bzl will need some attention to work well, as the Generator command-line interface has morphed some since this stuff was deleted.

Do you have any concrete suggestions for what to test? The example seems to work.

> * I see that the readme says "linux only has been tested"; I'd really like for at least an attempt at testing on OSX and Windows be made, to see where we stand.

I don't have access to a Windows computer so I can't help with that. I could perhaps try on Mac.

> * I have no idea what "Grail Bazel" is or what the implications are. Can you point me at relevant docs?

The Grail bazel toolchain is just a Clang toolchain that is nicely packaged, available [here](https://github.com/grailbio/bazel-toolchain). This Bazel code in Halide does not depend on it, it's just used in the example code. Bazel by default uses whatever compiler is installed on the system, which is not how I want to set up my builds.

Especially with Halide, which by necessity (see below) will only work with a Bazel toolchain that uses a specific C++ standard library, I don't think it's a good idea to rely on whatever the system default is.

AFAIK it would work to, for example, build Halide against libstdc++ and use Bazel's default toolchain on Ubuntu, which is GCC with libstdc++, but the same configuration would break on a machine that uses libc++ by default. So it seems like a good idea to show how to do it in a more portable way in the example.

> * "LLVM needs to be built with the same C++ standard library as the one used by Bazel" -- not saying that's your fault, but that's an entirely unreasonable requirement. Why should the build system's library have any effect on what is being built?

There is no realistic way to avoid this requirement.

It's not LLVM as in clang and the build system, it's LLVM as in the static LLVM library files that are contained within libHalide.a.

For Halide as it is configured inside of Google this is a non-issue because everything is built from source and consequently everything will be built with the same toolchain and C++ standard library automatically.

Halide with CMake can be (I don't know if this is the common way to do it though) configured to build LLVM and Halide from source with CMake which also will implicitly make sure that everything has the same C++ standard library.

For this Bazel setup, there is no realistic way to achieve this; LLVM can only be built with CMake.

Halide has a C++ API which seems to have C++ stdlib types like `std::string` in its API boundary, so any code that uses Halide **must** be compiled against the same C++ stdlib as Halide. Similarly, Halide directly uses LLVM C++ APIs, which I assume include C++ standard library types in the API boundary and perhaps also uses exceptions, so Halide **must** be built against the same C++ stdlib as LLVM.

It's painful, but this is just the way C++ is when distributing binary libraries. The only real way to avoid this is to meticulously avoid any use of the C++ standard library in the API boundary and never let any exceptions pass through, not even from libraries that you use. But this is not really a realistic option, as it would make the API quite a bit less convenient to use.

steven-johnson:
> Is there any point in supporting old versions of Bazel? 

No, absolutely not. I just don't understand how 0.28 is "newer" than 0.6 -- if that's like "prerelease v28" vs "prerelease v6" then, uh, ok, I guess? But in the parts of the world where the decimal separator is `.` that's pretty weird. (But in sum: if 0.28 is newer than 0.6 then yes, let's use that)

> It's not LLVM as in clang and the build system, it's LLVM as in the static LLVM library files that are contained within libHalide.a.

Ah, in that case we need to clarify the README instructions to make that clear, as the former is *exactly* what I inferred from the wording. It makes perfect sense that LLVM and Halide must be built with matching C++ libs.

> The Grail bazel toolchain is just a Clang toolchain ...  Bazel by default uses whatever compiler is installed on the system, which is not how I want to set up my builds.

I'd be happier if our one-and-only example used those defaults; I'm reluctant to point people at a toolchain that I can't vouch for. (Having additional docs or examples describing the Grail toolchain and why you'd consider it are completely reasonable though.)

> I don't have access to a Windows computer so I can't help with that. 

To be more on-point: we should aggressively remove support for platforms that aren't tested. If Linux is the only platform that the resurrected code has been tested on, we should prune out support for the other stuff rather than let people try something that may well fail. I'd rather have limited-but-tested support instead of maybe-this-works-but-who-knows, as the latter really presents a bad look to the project. (We can always re-add support for other things later, when we have the resources to actually test them.)

> Do you have any concrete suggestions for what to test?

The example works but is very limited; looking over the README, there are a number of things that have changed over time (e.g. the possible values for `extra_outputs` in `halide_library()` are much larger now). This isn't necessarily a call to action, but a heads-up that you may well find rough edges as you use it more.

BTW: sorry if I'm sounding overly critical here; I'm happy that the Bazel work is being resurrected! My main goal here is to make sure that what we land in this PR is limited to what we have confidence in actually working. 

(In particular, I would be stunned if the Windows support came even close to working still, as it was a barely-working hack-filled mess back then; hopefully Bazel's support for Windows is better now and it can eventually be resurrected in a cleaner way.)

per-gron:
Note: I intend to finish this PR as soon as I can. Unfortunately we discovered nasty insects in our home so I will have to spend the time I spend on this kind of coding to sanitize our apartment. I will get back to this, but not in the coming couple of days.

laurentlb:
> if 0.28 is newer than 0.6 then yes, let's use that

Yes, it's much newer. Version numbers are tuples, `0.10.0 > 0.9.0` (see [semver](https://semver.org/#semantic-versioning-specification-semver)).

If you need to integrate with CMake (i.e. part of the build is done by CMake), you may take a look at https://github.com/bazelbuild/rules_foreign_cc

per-gron:
@steven-johnson sorry about the slowness here. I still intend to do this, but it seems to take longer than I anticipated before I'll get back to this. I'll close this now and re-open when I have an update.

PR #4131, opened 2019-08-13
Change halide_toc_impl to use debug(1) instead of debug(0)

steven-johnson:
While handy for debugging, this means the autoscheduler currently spams stdout with timing info by default, which is puzzling to people not actually working on the autoscheduler.

(alternately: add debug-level option to HALIDE_TIC, or just comment out the calls to it in the autoscheduler)


steven-johnson:
(Linux failure is unrelated trunk breakage)

PR #4132, opened 2019-08-13
Tweak schedule.h output

steven-johnson:
- avoid unnecessary blank lines
- avoid lines that are whitespace-only

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -202,13 +202,15 @@ static Registerer registerer;
 
 std::string indent_string(const std::string &src, const std::string &indent) {
     std::ostringstream o;
-    o << indent;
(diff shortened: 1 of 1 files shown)


src/Module.cpp
```diff
@@ -202,13 +202,15 @@ static Registerer registerer;
...
-    o << indent;
+    bool prev_was_newline = true;
     for (size_t i = 0; i < src.size(); i++) {
         const char c = src[i];
-        o << c;
-        if (c == '\n' && i != src.size() - 1) {
-            o << indent;
+        const bool is_newline = (c == '\n');
+        if (prev_was_newline && !is_newline) {
+          o << indent;
```
abadams: nit: indentation

PR #4133, opened 2019-08-13
Clean up default autotuner output

steven-johnson:
This restructures so that the default output of the autoscheduler (at HL_DEBUG_AUTOSCHEDULE=0) is much smaller, with just some output for each pass. (All the previous output is still present for HL_DEBUG_AUTOSCHEDULE=1.) Not entire sure if this is the best "per-pass" summary; definitely open for suggestions here (even shorter/terser would be even nicer here IMHO).

Also, a little drive-by cleanup to the progress bar (to avoid leaving them dangling) and to give slightly better feedback during the autotune script.

### DIFF
--- a/apps/autoscheduler/AutoSchedule.cpp
+++ b/apps/autoscheduler/AutoSchedule.cpp
@@ -107,6 +107,41 @@ using std::map;
 using std::set;
 using std::pair;
 
+struct ProgressBar {
+    void set(double progress) {
+        if (!draw_progress_bar) return;
+        counter++;
+        const int bits = 11;
+        if (counter & ((1 << bits) - 1)) return;
+        const int pos = (int) (progress * 78);
+        aslog(0) << '[';
+        for (int j = 0; j < 78; j++) {
+            if (j < pos) {
+                aslog(0) << '.';
+            } else if (j - 1 < pos) {
+                aslog(0) << "/-\\|"[(counter >> bits) % 4];
+            } else {
+                aslog(0) << ' ';
+            }
+        }
+        aslog(0) << ']';
+        for (int j = 0; j < 80; j++) {
+            aslog(0) << '\b';
+        }
+    }
+
+    void clear() {
+        if (counter) {
+            for (int j = 0; j < 80; j++) aslog(0) << ' ';
+            for (int j = 0; j < 80; j++) aslog(0) << '\b';
+        }
+    }
+private:
+    uint32_t counter = 0;
+    const bool draw_progress_bar = isatty(2);
+};
+
+
 // Get the HL_RANDOM_DROPOUT environment variable. Purpose of this is described above.
 uint32_t get_dropout_threshold() {
     string random_dropout_str = get_env_variable("HL_RANDOM_DROPOUT");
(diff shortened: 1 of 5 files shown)


apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -2650,9 +2697,21 @@ struct State {
     }
 
     void dump() const {
-        aslog(0) << "State with cost " << cost << ":\n";
+        if (aslog::aslog_level() == 0) {
```
abadams: Not sure I understand this part. It just prints which Funcs are compute_root?

steven-johnson: Yeah, pretty much. I'm not sure that there is a good, terse way to describe each pass; this was my arbitrary first stab at it. Maybe we'd be better off just skipping that entirely and saying `Pass N of M, Cost Foo`?

abadams: Yeah, I think that's sufficient. The logging here is a sketch of the loop nest. There isn't really a useful condensed version of that.

apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -2389,22 +2424,22 @@ struct State {
...
-            // aslog(0) << "Skipping over scheduling input node: " << node->func.name() << "\n";
+            // aslog(1) << "Skipping over scheduling input node: " << node->func.name() << "\n";
             auto child = make_child();
             child->num_decisions_made++;
             accept_child(std::move(child));
             return;
         }
 
         if (!node->outgoing_edges.empty() && !root->calls(node)) {
-            aslog(0) << "In state:\n";
```
abadams: This is all printing an elaborate error message, rather than verbose logging. I'd say to just revert to aslog(0), but the dump() won't trigger. Maybe dump should all be aslog(0) too, and  the other calls to dump() should be conditioned on the logging level?

apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -3098,11 +3110,11 @@ IntrusivePtr<State> optimal_schedule_pass(FunctionDAG &dag,
             // The user has set HL_CYOS, and wants to navigate the
             // search space manually.  Discard everything in the queue
             // except for the user-chosen option.
-            aslog(0) << "\n--------------------\n";
```
abadams: This is asking for human input, and should always be printed

apps/autoscheduler/DefaultCostModel.cpp
```diff
@@ -176,18 +176,18 @@ class DefaultCostModel : public CostModel {
             *(cost_ptrs(i)) = dst(i);
             if (std::isnan(dst(i))) {
                 any_nans = true;
```
abadams: This is all an elaborate error message for when the network predicts nan, and should be printed unconditionally

apps/autoscheduler/AutoSchedule.cpp
steven-johnson: I still shudder at the "ask for human input" part here :-)

apps/autoscheduler/FunctionDAG.h
```diff
@@ -185,24 +185,24 @@ class LoadJacobian {
 
     void dump(const char *prefix) const {
```
abadams: I don't think the bodies of any of the dump calls should be changed in this PR, but rather the call to dump should be conditional on the debug level

abadams: (I'm imagining adding new dump calls while debugging and getting confused regarding why things aren't being dumped)

apps/autoscheduler/AutoSchedule.cpp
steven-johnson: Ouch, yeah, I guess. Let me see what I can do.

steven-johnson: Better yet might be to pass an ostream (or similar) in to dump methods, and let the caller decide... but I'll save that for another day

apps/autoscheduler/AutoSchedule.cpp
abadams: It's a debugging tool for people working on the autoscheduler

steven-johnson:
PTAL

apps/autoscheduler/DefaultCostModel.cpp
abadams: bump

steven-johnson: PTAL

apps/autoscheduler/FunctionDAG.h
steven-johnson: done, I think

steven-johnson:
(The Linux failure is unrelated LLVM-trunk issue)

PR #4138, opened 2019-08-15
Fix for https://reviews.llvm.org/rL368647, from alinas@

steven-johnson:


src/CodeGen_LLVM.cpp
```diff
@@ -3858,7 +3858,14 @@ void CodeGen_LLVM::codegen_asserts(const vector<const AssertStmt *> &asserts) {
 
     // Now switch on the bitmask to the correct failure
     Expr case_idx = cast<int32_t>(count_trailing_zeros(bitmask));
-    auto *switch_inst = builder->CreateSwitch(codegen(case_idx), no_errors_bb, asserts.size(), very_likely_branch);
+    llvm::SmallVector<uint32_t, 32> weights;
```
abadams: Given that asserts.size() is at most 63, may as well use 64 as the template parameter for the size of the small vector.

src/CodeGen_LLVM.cpp
```diff
@@ -3858,7 +3858,14 @@ void CodeGen_LLVM::codegen_asserts(const vector<const AssertStmt *> &asserts) {
...
+    weights.push_back(1<<30);
+    for (int i = 0; i < (int) asserts.size(); i++) {
+      weights.push_back(0);
```
abadams: indentation

src/CodeGen_LLVM.cpp
steven-johnson: Done. (I hadn't realized we have this limit -- have we ever exceeded this limit in a real-world pipeline?)

src/CodeGen_LLVM.cpp
steven-johnson: Done

src/CodeGen_LLVM.cpp
abadams: All the time, but there's an outer loop that processes blocks of 63 asserts. The general strategy is that we process blocks of asserts in batches of size 63. The conditions go into a bitmask (one bit is reserved to represent all asserts passing). We then count leading zeros to generate a switch index so that we can branch to the correct error handler for the first false condition.

abadams: I should add: the nice thing about this is that by computing the conditions together without any control flow in the way, the checks get vectorized by llvm's SLP vectorizer.

PR #4145, opened 2019-08-16
Improve (re)train_cost_model tool

steven-johnson:
- Rename to retrain_cost_model, since (1) that's what it does, and (2) this avoids confusion with the Generator-produces filter of the same name
- Remove all use of env vars for inputs; explicit commandline flags are now required instead. Update autotune_loop.sh accordingly.
- add an option to copy the best .schedule.h file somewhere at the end (note that the previous HL_BEST_SCHEDULE_FILE env var actually output benchmark results, not the schedule file)
- Check the return value of calls to train_cost_model() and cost_model(), just in case someone uses a nonstandard override of halide_error.
- Tighten some verbosity of output (display just filenames instead of pathnames in some cases)
- Some minor drive-by style cleanups.


abadams:
Not sure I buy the renaming. The main use of this binary is to train the cost model. This can start from random initialization (the default at one point at least), or it can optionally continue from an existing set of weights. We should make sure that usage stays simple. So the old HL_WEIGHTS_DIR is primarily an output parameter, not an input parameter. It started from any existing compatible weights in that folder, or randomly initialized weights if there wasn't anything there already. This sort of implies that --weights-out is the mandatory parameter, and --weights is optional (and should maybe be --initial-weights/-i and mutually exclusive with -r). Also, the binary should probably refuse to continue if --weights-out and -r are set but that weights file already exists, to avoid clobbering a good set of weights. This behavior was previously impossible (I believe).

steven-johnson:
> Not sure I buy the renaming. 

Fair, but I do think that having a different name from the generator-output filter of the same name would be useful. (Maybe rename that filter instead?) This not a hill I'm willing to die on, but more a point of minor frustration when searching and opening files.

> We should make sure that usage stays simple. So the old HL_WEIGHTS_DIR is primarily an output parameter, not an input parameter. It started from any existing compatible weights in that folder, or randomly initialized weights if there wasn't anything there already. 

In theory, sure, but in practice, no; AFAICT all the callers always set HL_WEIGHTS_DIR (intending it as an in+out param); HL_WEIGHTS_OUT_DIR existed but nothing was using it (I added it months ago thinking I would need it for some google-build-system-specific purposes).

> This sort of implies that --weights-out is the mandatory parameter, and --weights is optional (and should maybe be --initial-weights/-i and mutually exclusive with -r). 

SGTM.

> Also, the binary should probably refuse to continue if --weights-out and -r are set but that weights file already exists, to avoid clobbering a good set of weights. This behavior was previously impossible (I believe).

That is absolutely not the case. If you set HL_WEIGHTS_DIR and HL_RANDOMIZE_WEIGHTS in the old model, it would read in the weights from HL_WEIGHTS_DIR, then overwrite them with random values, then save the results in HL_WEIGHTS_DIR, overwriting the original. (I agree that it could be a problem, but it is definitely not new behavior.)

abadams:
That all sounds reasonable. Maybe just change the bit you SGTM'd?

steven-johnson:
I just pushed the SGTM bit. 

I'm fine to back out the renaming here (and take it up in a separate PR as a minor issue); LMK.

steven-johnson:
> --initial-weights/-i and mutually exclusive with -r

On that note: currently, the loading code in DefaultCostModel tries to load a file, but if it fails, it reverts to randomizing (with just a warning). I know the motivation for this but it still kinda scares me (ie overwriting useful weights is bad); I wonder if we'd be better off having it be a hard-fail by default, with some flag to allow this quiet fallback for when it's needed.

steven-johnson:
FYI: the OSX buildbot is failing sometimes because we are training for x86-64-avx2, and... that machine doesn't have AVX2. (Training doesn't do multitarget with graceful fallback. Whoops.)

steven-johnson:
Update: Guess What? Neither of our Linux buildbots have AVX2, either. (One doesn't even have AVX.) How this test has ever worked on the buildbot is puzzling. But more importantly, we have zero coverage as written. We need to figure out whether we can test usefully with lower features or not before landing this.

abadams:
We could change it to target=host. I think it was set to avx2 because avx512 is not a clear win and that's what host was on my benchmarking machine.

steven-johnson:
Do we have a convenient way to sniff what target=host is from a makefile? (If not, maybe we should add one?)

steven-johnson:
PTAL

steven-johnson:
Gentle review ping :-)

apps/autoscheduler/get_host_target.cpp
```diff
@@ -0,0 +1,24 @@
...
+            exit(1);
+        }
+        t = t.without_feature(f);
+    }
+    printf("%s", t.to_string().c_str());
+    return 0;
+}
+
+
+
```
abadams: I've been doing: 
```
any.generator -o foo -r bar target=pants 2>&1 | grep "the host target is" | cut -d: -f2
```

Perhaps any generator should be able to print the host target? Not worth changing for this PR though.

steven-johnson: Sniffing the host target by grepping an error message is expedient, but perhaps suboptimal as a long-term solution :-)

abadams: Oh absolutely, but I wondered if generator binaries should serve as the utilities for these sorts of queries (much as they can generate runtimes).

PR #4151, opened 2019-08-19
Smarten UnrollLoops; remove HL_PERMIT_FAILED_UNROLL=1 defaults

steven-johnson:
Fix an issue where UnrollLoops failed to recognize a constant extent as such, then remove HL_PERMIT_FAILED_UNROLL=1 as the default in autoscheduler. (We think this should be (mostly) unnecessary in most cases, so we'll now issue a user-warning if we trigger it.)

### DIFF
--- a/src/UnrollLoops.cpp
+++ b/src/UnrollLoops.cpp
@@ -72,6 +72,7 @@ class UnrollLoops : public IRMutator {
             if (e == nullptr && permit_failed_unroll) {
                 // Still no luck, but we're allowed to fail. Rewrite
                 // to a serial loop.
+                user_warning << "HL_PERMIT_FAILED_UNROLL is allowing us to unroll a non-constant loop into a serial loop. Did you mean to do this?\n";
                 body = mutate(body);
                 return For::make(for_loop->name, for_loop->min, for_loop->extent,
                                  ForType::Serial, for_loop->device_api, std::move(body));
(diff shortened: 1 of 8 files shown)


src/UnrollLoops.cpp
```diff
@@ -11,6 +11,18 @@ using std::pair;
 namespace Halide {
 namespace Internal {
 
+struct RemoveLikelies : public IRMutator {
```
abadams: Should we perhaps move this to the internal section of IROperator? I feel like I've needed to copy-paste it once already.

steven-johnson: done

abadams:
RemoveLikelyTags in PartitionLoops can use it too. It's only different because it's called after likely_if_innermost has been lowered to likely.

steven-johnson:
FYI: with this fix in place, I'm getting another failed-unroll... but this one is across a for-loop named `repeat_edge$1.s0.y`, so I suspect this is one where the unroll shouldn't have been chosen in the first place (rather than a simplification failure). Trying to figure out the logic for choosing an unroll now.

abadams:
It's not so much should-unroll logic as it is loop-is-known-to-have-constant-bounds logic. Whether or not to unroll is up to the mysterious cost model, but it's only allowed to do it for loops that are known to have constant bounds (so the inside of a split, or a loop that comes in with a known constant extent e.g. due to a Func::bound directive)

steven-johnson:
Pretty sure it's not a `.bound()` directive.

PR #4154, opened 2019-08-19
Enable bit shifts by negative amounts.

kpassarella:
A bit shift by a negative amount is now defined as a shift by a positive
amount in the opposite direction.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1039,11 +1039,29 @@ private:
                     if (op->is_intrinsic(Call::shift_left)) {
                         if (t.is_int() && t.bits() >= 32) {
                             // Overflow is UB
-                            if (a_interval.has_lower_bound() && b_interval.has_lower_bound() && can_prove(b_interval.min >= 0 && b_interval.min < t.bits())) {
+                            if (a_interval.has_lower_bound() &&
+                                b_interval.has_lower_bound() &&
+                                can_prove(b_interval.min >= 0 &&
+                                          b_interval.min < t.bits())) {
                                 interval.min = a_interval.min << b_interval.min;
+                            } else if (a_interval.has_lower_bound() &&
+                                       b_interval.has_lower_bound() &&
+                                       !b_interval.min.type().is_uint() &&
+                                       can_prove(b_interval.min < 0 &&
+                                                 b_interval.min > -t.bits())) {
+                                interval.min = a_interval.min >> abs(b_interval.min);
                             }
-                            if (a_interval.has_upper_bound() && b_interval.has_upper_bound() && can_prove(b_interval.max >= 0 && b_interval.max < t.bits())) {
+                            if (a_interval.has_upper_bound() &&
+                                b_interval.has_upper_bound() &&
+                                can_prove(b_interval.max >= 0 &&
+                                          b_interval.max < t.bits())) {
                                 interval.max = a_interval.max << b_interval.max;
+                            } else if (a_interval.has_upper_bound() &&
+                                       b_interval.has_upper_bound() &&
+                                       !b_interval.max.type().is_uint() &&
+                                       can_prove(b_interval.max < 0 &&
+                                                 b_interval.max > -t.bits())) {
+                                interval.max = a_interval.max >> abs(b_interval.max);
                             }
                         } else if (is_const(b)) {
                             // We can normalize to multiplication
@@ -1054,26 +1072,48 @@ private:
                         // Only try to improve on bounds-of-type if we can prove 0 <= b < t.bits,
                         // as shift_right(a, b) is UB for b outside that range.
                         if (b_interval.is_bounded()) {
-                            bool b_min_ok = can_prove(b_interval.min >= 0 && b_interval.min < t.bits());
-                            bool b_max_ok = can_prove(b_interval.max >= 0 && b_interval.max < t.bits());
+                            bool b_min_ok_positive = can_prove(b_interval.min >= 0 &&
+                                                               b_interval.min < t.bits());
+                            bool b_max_ok_positive = can_prove(b_interval.max >= 0 &&
+                                                               b_interval.max < t.bits());
+                            bool b_min_ok_negative =
+                                !b_interval.min.type().is_uint() &&
+                                can_prove(b_interval.min < 0 && b_interval.min > -t.bits());
+                            bool b_max_ok_negative =
+                                !b_interval.max.type().is_uint() &&
+                                can_prove(b_interval.max < 0 && b_interval.max > -t.bits());
                             if (a_interval.has_lower_bound()) {
-                                if (can_prove(a_interval.min >= 0) && b_max_ok) {
+                                if (can_prove(a_interval.min >= 0) && b_max_ok_positive) {
                                     interval.min = a_interval.min >> b_interval.max;
-                                } else if (b_min_ok && b_max_ok) {
+                                } else if (can_prove(a_interval.min < 0) && b_max_ok_negative) {
+                                    interval.min = a_interval.min << abs(b_interval.max);
+                                } else if (b_min_ok_positive && b_max_ok_positive) {
                                     // if a < 0, the smallest value will be a >> b.min
                                     // if a > 0, the smallest value will be a >> b.max
                                     interval.min = min(a_interval.min >> b_interval.min,
                                                        a_interval.min >> b_interval.max);
+                                } else if (b_min_ok_negative && b_max_ok_negative) {
+                                    // if a < 0, the smallest value will be a << abs(b.min)
+                                    // if a > 0, the smallest value will be a << abs(b.max)
+                                    interval.min = min(a_interval.min << abs(b_interval.min),
+                                                       a_interval.min << abs(b_interval.max));
                                 }
                             }
                             if (a_interval.has_upper_bound()) {
-                                if (can_prove(a_interval.max >= 0) && b_min_ok) {
+                                if (can_prove(a_interval.max >= 0) && b_min_ok_positive) {
                                     interval.max = a_interval.max >> b_interval.min;
-                                } else if (b_min_ok && b_max_ok) {
+                                } else if (can_prove(a_interval.max < 0) && b_min_ok_negative) {
+                                    interval.max = a_interval.max << abs(b_interval.min);
+                                } else if (b_min_ok_positive && b_max_ok_positive) {
                                     // if a < 0, the largest value will be a >> b.max
                                     // if a > 0, the largest value will be a >> b.min
                                     interval.max = max(a_interval.max >> b_interval.max,
                                                        a_interval.max >> b_interval.min);
+                                } else if (b_min_ok_negative && b_max_ok_negative) {
+                                    // if a < 0, the largest value will be a << abs(b.max)
+                                    // if a > 0, the largest value will be a << abs(b.min)
+                                    interval.max = max(a_interval.max << abs(b_interval.max),
+                                                       a_interval.max << abs(b_interval.min));
                                 }
                             }
                         }
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -1712,10 +1712,13 @@ void CodeGen_Hexagon::visit(const Call *op) {
             if (value) return;
         } else if (op->is_intrinsic(Call::shift_left) ||
                    op->is_intrinsic(Call::shift_right)) {
-
             internal_assert(op->args.size() == 2);
             string instr = op->is_intrinsic(Call::shift_left) ? "halide.hexagon.shl" : "halide.hexagon.shr";
             Expr b = maybe_scalar(op->args[1]);
+            if (b.type().is_int()) {
+                CodeGen_Posix::visit(op);
+                return;
+            }
             internal_assert(b.type().is_uint());
             value = call_intrin(op->type,
                                 instr + type_suffix(op->args[0], b),
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -5,6 +5,8 @@
 #include "IRMutator.h"
 #include "IROperator.h"
 #include "LLVM_Headers.h"
+#include "Simplify.h"
+#include "Simplify_Internal.h"
 
 namespace Halide {
 namespace Internal {
@@ -405,6 +407,50 @@ Expr lower_euclidean_mod(Expr a, Expr b) {
     }
 }
 
+Expr lower_signed_shift_left(Expr a, Expr b) {
+    assert(b.type().is_int());
+    const int64_t *const_int_b = as_const_int(b);
+    if (const_int_b) {
+        Type t = UInt(a.type().bits(), a.type().lanes());
+        Expr val;
+        const uint64_t b_unsigned = std::abs(*const_int_b);
+        if (*const_int_b >= 0) {
+            val = a << make_const(t, b_unsigned);
+        } else if (*const_int_b < 0) {
+            val = a >> make_const(t, b_unsigned);
+        }
+        return common_subexpression_elimination(val);
+    } else {
+        // The abs() below uses Halide's abs operator. This eliminates the overflow
+        // case for the most negative value because its result is unsigned.
+        Expr b_unsigned = abs(b);
+        Expr val = select(b >= 0, a << b_unsigned, a >> b_unsigned);
+        return simplify(common_subexpression_elimination(val));
+    }
+}
+
+Expr lower_signed_shift_right(Expr a, Expr b) {
+    assert(b.type().is_int());
+    const int64_t *const_int_b = as_const_int(b);
+    if (const_int_b) {
+        Type t = UInt(a.type().bits(), a.type().lanes());
+        Expr val;
+        const uint64_t b_unsigned = std::abs(*const_int_b);
+        if (*const_int_b >= 0) {
+            val = a >> make_const(t, b_unsigned);
+        } else if (*const_int_b < 0) {
+            val = a << make_const(t, b_unsigned);
+        }
+        return common_subexpression_elimination(val);
+    } else {
+        // The abs() below uses Halide's abs operator. This eliminates the overflow
+        // case for the most negative value because its result is unsigned.
+        Expr b_unsigned = abs(b);
+        Expr val = select(b >= 0, a >> b_unsigned, a << b_unsigned);
+        return simplify(common_subexpression_elimination(val));
+    }
+}
+
 namespace {
 
 // This mutator rewrites predicated loads and stores as unpredicated
@@ -686,3 +732,4 @@ void embed_bitcode(llvm::Module *M, const string &halide_command) {
 
 }  // namespace Internal
 }  // namespace Halide
+
--- a/src/CodeGen_Internal.h
+++ b/src/CodeGen_Internal.h
@@ -84,6 +84,13 @@ Expr lower_euclidean_div(Expr a, Expr b);
 Expr lower_euclidean_mod(Expr a, Expr b);
 ///@}
 
+/** Given a Halide shift operation with a signed shift amount (may be negative), define
+ * an equivalent expression using only shifts by unsigned amounts. */
+///@{
+Expr lower_signed_shift_left(Expr a, Expr b);
+Expr lower_signed_shift_right(Expr a, Expr b);
+///@}
+
 /** Replace predicated loads/stores with unpredicated equivalents
  * inside branches. */
 Stmt unpredicate_loads_stores(Stmt s);
@@ -110,3 +117,4 @@ void embed_bitcode(llvm::Module *M, const std::string &halide_command);
 }  // namespace Halide
 
 #endif
+
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -515,6 +515,38 @@ void match_types(Expr &a, Expr &b) {
     }
 }
 
+// Cast to the wider type of the two. Already guaranteed to leave
+// signed/unsigned on number of lanes unchanged.
+void match_bits(Expr &x, Expr &y) {
+    if (y.type().is_int() == x.type().is_int()) {
+        // The signedness matches, so do a full cast.
+        if (x.type().bits() < y.type().bits()) {
+            x = cast(y.type(), x);
+        } else if (y.type().bits() < x.type().bits()) {
+            y = cast(x.type(), y);
+        }
+    } else {
+        // The signedness doesn't match, so just match the bits.
+        if (x.type().bits() < y.type().bits()) {
+            Type t;
+            if (x.type().is_int()) {
+              t = Int(y.type().bits(), y.type().lanes());
+            } else {
+              t = UInt(y.type().bits(), y.type().lanes());
+            }
+            x = cast(t, x);
+        } else if (y.type().bits() < x.type().bits()) {
+            Type t;
+            if (y.type().is_int()) {
+              t = Int(x.type().bits(), x.type().lanes());
+            } else {
+              t = UInt(x.type().bits(), x.type().lanes());
+            }
+            y = cast(t, y);
+        }
+    }
+}
+
 void match_types_bitwise(Expr &x, Expr &y, const char *op_name) {
     user_assert(x.defined() && y.defined()) << op_name << " of undefined Expr\n";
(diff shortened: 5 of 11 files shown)


src/CodeGen_Hexagon.cpp
```diff
@@ -1716,7 +1716,6 @@ void CodeGen_Hexagon::visit(const Call *op) {
             internal_assert(op->args.size() == 2);
             string instr = op->is_intrinsic(Call::shift_left) ? "halide.hexagon.shl" : "halide.hexagon.shr";
             Expr b = maybe_scalar(op->args[1]);
-            internal_assert(b.type().is_uint());
```
dsharletg: I think this assert should remain, until we know that Hexagon's shift operators do the right thing with negative shifts. Until then, this should use `lower_signed_shift_*` to avoid shifts by signed RHS values.

I think you could implement this by inserting this before the assert:

    if (b.type().is_int()) {
        CodeGen_Posix::visit(op);
        return;
    }

zvookin:
Addresses issue: https://github.com/halide/Halide/issues/3325 .

src/CodeGen_Internal.h
```diff
@@ -84,6 +84,13 @@ Expr lower_euclidean_div(Expr a, Expr b);
 Expr lower_euclidean_mod(Expr a, Expr b);
 ///@}
 
+/** Given a Halide shift operation with negative shift amount, define it as
+ * a positive shift in the opposite direction. */
```
dsharletg: I think this comment sounds more like it is handling a shift operation with a concretely negative amount, which isn't quite what it is doing. I think the comment should be a description more like roughly "Given a Halide shift operation with a signed shift amount (may be negative), define an equivalent expression using only shifts by unsigned amounts."

src/CodeGen_Hexagon.cpp
zvookin: Do we need to assert ```b.type().is_uint() || b.type().is_int()``` now?

src/CodeGen_Internal.cpp
```diff
@@ -407,6 +407,16 @@ Expr lower_euclidean_mod(Expr a, Expr b) {
     }
 }
 
+Expr lower_signed_shift_left(Expr a, Expr b) {
+  Expr b_unsigned = abs(b);
+  return select(b >= 0, a << b_unsigned, a >> b_unsigned);
```
abadams: I'd modify these to do both simplification and then common_subexpression_elimination on the output. So something like:
```
Expr val = select(b >= 0, a << b_unsigned, a >> b_unsigned);
return simplify(common_subexpression_elimination(val));
```
Halide is better at proving things are positive than LLVM, so the switch might simplify away, and if a/b are large Exprs, duplicating them needlessly creates extra codegen work.

src/IROperator.h
```diff
@@ -1562,20 +1545,35 @@ inline Expr operator<<(Expr x, int y) {
...
-        << "In shift right expression:\n"
-        << "   (" << x << ") >> (" << y << ")\n"
-        << "   with types " << x.type() << " >> " << y.type() << "\n"
-        << "the RHS must be unsigned and losslessly castable to the same size as the LHS.\n";
     Type t = x.type();
-    return Internal::Call::make(t, Internal::Call::shift_right, {std::move(x), std::move(unsigned_amount)}, Internal::Call::PureIntrinsic);
+    return Internal::Call::make(t, Internal::Call::shift_right, {std::move(x), std::move(y)}, Internal::Call::PureIntrinsic);
+}
+inline Expr operator<<(Expr x, int y) {
+    const bool flip_shift = y < 0;
```
dsharletg: I think that because we should handle shifts of constant negative values in the simplifier, that we shouldn't *also* handle it here, which should make things simpler (only having this logic in one place).

src/Simplify_Call.cpp
```diff
@@ -125,28 +127,39 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             return a;
         }
 
-        const bool shift_left = op->is_intrinsic(Call::shift_left);
+        //const bool const_shift_left = op->is_intrinsic(Call::shift_left);
```
dsharletg: Remove commented code.

test/auto_schedule/mat_mul.cpp
```diff
@@ -126,7 +126,7 @@ int main(int argc, char **argv) {
     std::cout << "======================" << std::endl;
 
     if (!get_jit_target_from_environment().has_gpu_feature() &&
-        (auto_time > manual_time * 3.5)) {
+        (auto_time > manual_time * 5.0)) {
```
dsharletg: I think this change should be reverted, as it is unrelated to this PR. I think the autoschedule tests shouldn't be blocking this (or any) PR.

src/Simplify_Call.cpp
```diff
@@ -125,28 +127,39 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
         uint64_t ub = 0;
-        if (const_uint(b, &ub)) {
+        bool is_const = const_uint(b, &ub);
+        if (!is_const) {
+            // Only do the abs(b) operation if b is signed to avoid warning
+            // messages that abs of an unsigned type is a no-op.
+            is_const = const_uint(abs(b), &ub);
+        }
+        if (is_const) {
             // LLVM shl and shr instructions produce poison for negative shifts,
```
dsharletg: I think this comment is stale now.

src/CodeGen_Internal.cpp
```diff
@@ -407,6 +407,16 @@ Expr lower_euclidean_mod(Expr a, Expr b) {
     }
 }
 
+Expr lower_signed_shift_left(Expr a, Expr b) {
+  Expr b_unsigned = abs(b);
```
zvookin: Might be worth adding a comment here that this is Halide's abs operator which eliminates the overflow case for the most negative value because its result is unsigned.

src/CodeGen_Internal.cpp
abadams: (and assumes that llvm with successfully CSE them away again, which it almost certainly will, but still... in theory we have more information about what is and isn't pure than llvm)

src/Simplify_Call.cpp
steven-johnson: Did you mean to leave this commented out? If so, add a line explaining why. If not, remove.

src/Simplify_Call.cpp
```diff
@@ -125,28 +127,39 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             return a;
         }
 
-        const bool shift_left = op->is_intrinsic(Call::shift_left);
+        //const bool const_shift_left = op->is_intrinsic(Call::shift_left);
+        //const bool const_shift_right = true;
+        const bool const_shift_left = ((op->is_intrinsic(Call::shift_left) && can_prove(b >= 0)) ||
```
steven-johnson: `can_prove()` is an expensive operation; might be good to pull these out into 

```
const bool b_pos = can_prove(b >= 0);
const bool b_neg = can_prove(b < 0);
```

dsharletg: @abadams might be best to comment on this, but this might be overkill at this point in the simplifier.

It might be best to just try to handle constant shifts (replacing them with Mul/Div), like the old code did (but only for unsigned shift amounts). 

If nothing else, it might be best to leave more aggressive simplification to a follow-up PR.

test/correctness/bitwise_ops.cpp
```diff
@@ -129,22 +129,142 @@ int main(int argc, char **argv) {
...
+            printf("%d >> %d -> %d instead of %d\n",
+                   input(x), shift_amount, im9(x), correct);
             return -1;
         }
     }
 
-    // bitwise and on mixed types
+    // arithmetic shift left with signed type (positive value)
     Func f10;
+    Expr a10 = reinterpret<int>(input(x));
```
dsharletg: I think this should just be a cast<int32_t>, rather than a reinterpret (and below). reinterpret is more like bitcast/reinterpret_cast, which I think is not what you want.

test/correctness/bitwise_ops.cpp
```diff
@@ -129,22 +129,142 @@ int main(int argc, char **argv) {
...
+    Expr b10 = reinterpret<int>(input(x+1));
+    f10(x) = a10 << (b10 & 0xf);
```
dsharletg: Why not 31 here, instead of 15? (for 32-bit integers).

src/Simplify_Call.cpp
abadams: Actually instead of these can_proves, you likely just want to pass an ExprInfo last arg for the mutation of b above instead of the nullptr, then check b_info.min_defined() && b_info.min >= 0. It's both cheaper, but more importantly it exploits information from the context of this call (e.g. maybe b is a variable and we're inside a for loop over that variable that starts at 3, or maybe we're inside an if statement with a condition of the form b > 0)

abadams: Oh, but the results are never going to be used unless b is also const. So my second reason doesn't hold. Just use is_positive_const  / is_negative_const instead of can_prove

dsharletg: Andrew already commented, going to resolve this and let Andrew's comment be the one for discussion.

src/CodeGen_Hexagon.cpp
kpassarella: I think doing Dillon's suggestion fixed the issue here.

src/IROperator.h
kpassarella: I tried to do that, but I either got cast errors going to unsigned t or segfaults later if I made t signed. Is this ok to handle it both places?

src/Simplify_Call.cpp
kpassarella: Removed, my bad.

src/Simplify_Call.cpp
kpassarella: All makes sense. I switched it to is_positive_const / is_negative_const.

test/auto_schedule/mat_mul.cpp
kpassarella: Done. It might fail build/presubmit tests until https://github.com/halide/Halide/pull/4153 is submitted, but then it'll be fine.

src/CodeGen_Internal.cpp
kpassarella: I added common_subexpression_elimination, but simplify seemed to segfault places. I'm not sure if that's expected? I can investigate more if that should have worked.

test/correctness/bitwise_ops.cpp
kpassarella: Yup, it should be 31. Changed it to that.

src/IROperator.h
dsharletg: I think we need to understand this better. I would be worried that if the simplifier produced a constant negative shift, that it would seg fault later too.

By removing this logic to handle this early, it would exercise the rest of the compiler with more tests (and avoid hiding future bugs), which needs to handle negative strides too.

src/CodeGen_Internal.cpp
abadams: That's not expected and is concerning. It would be good to investigate more.

steven-johnson: As a good rule of thumb, segfaults are never expected or acceptable -- we'll have to run this to ground and fix it.

kpassarella: I think I figured out what was happening here. There was an infinite loop happening between Simplify_Call and the new stuff in LLVM_Internal, which I found a way around in Simplify_Call. I'm still trying to resolve some weird output related to Dillon's comment in IROperator, so I'll update the working code when that's resolved, too.

kpassarella: Updated now with the fix. Please take another look when you get a chance!

src/IROperator.h
kpassarella: Updated now with the fix. Please take another look when you get a chance!

src/Simplify_Call.cpp
```diff
@@ -125,28 +127,39 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             return a;
         }
 
-        const bool shift_left = op->is_intrinsic(Call::shift_left);
         const Type t = op->type;
+        const bool b_pos = is_positive_const(b);
```
dsharletg: From here to ~10 lines below is a little bit confusing because there's some redundant things happening. It took me quite a while to read it, I think it should be simplified somehow.

Some ideas:
- After checking is_const_int/is_const_uint, just determine b_pos and b_neg from the resulting constants, rather than calling is_positive_const/is_negative_const.
- Move this const_shift_left/const_shift_right stuff inside the first if below.

src/CodeGen_Internal.cpp
```diff
@@ -407,6 +409,58 @@ Expr lower_euclidean_mod(Expr a, Expr b) {
     }
 }
 
+Expr lower_signed_shift_left(Expr a, Expr b) {
+    const uint64_t *const_uint_b = as_const_uint(b);
```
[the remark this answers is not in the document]
kpassarella: Good point. Done.

src/Simplify_Call.cpp
kpassarella: You're right -- I cleaned this up a little. Please let me know what you think now!

src/CodeGen_Hexagon.cpp
```diff
@@ -1716,6 +1716,10 @@ void CodeGen_Hexagon::visit(const Call *op) {
             internal_assert(op->args.size() == 2);
             string instr = op->is_intrinsic(Call::shift_left) ? "halide.hexagon.shl" : "halide.hexagon.shr";
             Expr b = maybe_scalar(op->args[1]);
+            if (b.type().is_int()) {
+              CodeGen_Posix::visit(op);
```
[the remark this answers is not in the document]
kpassarella: One of these days I'll remember the 4 spaces!
Although, this was on an old patch, so it was already fixed.

src/Bounds.cpp
```diff
@@ -1039,11 +1039,33 @@ class Bounds : public IRVisitor {
...
+                                b_interval.max.type().is_int() ? abs(b_interval.max) : b_interval.max;
+                            if (a_interval.has_lower_bound() &&
+                                b_interval.has_lower_bound() &&
+                                can_prove(b_interval.min >= 0 &&
+                                          b_interval.min < t.bits())) {
                                 interval.min = a_interval.min << b_interval.min;
+                            } else if (a_interval.has_lower_bound() &&
+                                       b_interval.has_lower_bound() &&
+                                       can_prove(b_interval.min < 0 &&
+                                                 abs_b_interval_min < t.bits())) {
```
dsharletg: Should this possibly be b_interval.min > -t.bits()? Then the abs_b_interval_min (and abs_b_interval_max) would be unused and could be removed. I'm also thinking maybe the simplifier will be more likely to be able to prove that than an expression with abs in it.

src/Bounds.cpp
```diff
@@ -1054,26 +1076,52 @@ class Bounds : public IRVisitor {
                         // Only try to improve on bounds-of-type if we can prove 0 <= b < t.bits,
```
dsharletg: abs(t.bits) in comment

src/Bounds.cpp
```diff
@@ -1054,26 +1076,52 @@ class Bounds : public IRVisitor {
...
+                            bool b_max_ok_positive = can_prove(b_interval.max >= 0 &&
+                                                               b_interval.max < t.bits());
+                            // The following two Exprs are to avoid warning
+                            // messages that abs of an unsigned type is a no-op.
+                            Expr abs_b_interval_min =
+                                b_interval.min.type().is_int() ? abs(b_interval.min) : b_interval.min;
+                            Expr abs_b_interval_max =
+                                b_interval.max.type().is_int() ? abs(b_interval.max) : b_interval.max;
+                            bool b_min_ok_negative = can_prove(b_interval.min < 0 &&
+                                                               abs_b_interval_min < t.bits());
```
dsharletg: Same comment as above.

src/Bounds.cpp
dsharletg: Eh that's a no-op, never mind.

src/Bounds.cpp
kpassarella: Done

src/Bounds.cpp
kpassarella: Done

src/Bounds.cpp
```diff
@@ -1039,11 +1039,29 @@ class Bounds : public IRVisitor {
                     if (op->is_intrinsic(Call::shift_left)) {
                         if (t.is_int() && t.bits() >= 32) {
                             // Overflow is UB
-                            if (a_interval.has_lower_bound() && b_interval.has_lower_bound() && can_prove(b_interval.min >= 0 && b_interval.min < t.bits())) {
+                            // The following two Exprs are to avoid warning
```
dsharletg: This comment should now be removed too

src/Bounds.cpp
```diff
@@ -1054,26 +1072,48 @@ class Bounds : public IRVisitor {
                         // Only try to improve on bounds-of-type if we can prove 0 <= b < t.bits,
                         // as shift_right(a, b) is UB for b outside that range.
                         if (b_interval.is_bounded()) {
-                            bool b_min_ok = can_prove(b_interval.min >= 0 && b_interval.min < t.bits());
-                            bool b_max_ok = can_prove(b_interval.max >= 0 && b_interval.max < t.bits());
+                            bool b_min_ok_positive = can_prove(b_interval.min >= 0 &&
+                                                               b_interval.min < t.bits());
+                            bool b_max_ok_positive = can_prove(b_interval.max >= 0 &&
+                                                               b_interval.max < t.bits());
+                            // The following two Exprs are to avoid warning
```
dsharletg: Stale comment

src/Bounds.cpp
```diff
@@ -1039,11 +1039,29 @@ class Bounds : public IRVisitor {
...
+                            // messages that abs of an unsigned type is a no-op.
+                            if (a_interval.has_lower_bound() &&
+                                b_interval.has_lower_bound() &&
+                                can_prove(b_interval.min >= 0 &&
+                                          b_interval.min < t.bits())) {
                                 interval.min = a_interval.min << b_interval.min;
+                            } else if (a_interval.has_lower_bound() &&
+                                       b_interval.has_lower_bound() &&
+                                       can_prove(b_interval.min < 0 &&
+                                                 cast<int>(b_interval.min) > -t.bits())) {
```
dsharletg: Casting this to int seems possibly dangerous, if it is an unsigned int. Maybe adding to the condition: `... && !b_interval.min.type().is_uint() && ...` and then skipping the cast which is hopefully now unnecessary?

src/Bounds.cpp
kpassarella: Done

src/Bounds.cpp
kpassarella: Done

src/Bounds.cpp
kpassarella: I figured it would be fine since the one before is checking that it's < 0 and an unsigned min would fail there, but I agree that this is even safer.

src/Bounds.cpp
```diff
@@ -1039,11 +1039,29 @@ class Bounds : public IRVisitor {
...
-                            if (a_interval.has_upper_bound() && b_interval.has_upper_bound() && can_prove(b_interval.max >= 0 && b_interval.max < t.bits())) {
+                            if (a_interval.has_upper_bound() &&
+                                b_interval.has_upper_bound() &&
+                                can_prove(b_interval.max >= 0 &&
+                                          b_interval.max < t.bits())) {
                                 interval.max = a_interval.max << b_interval.max;
+                            } else if (a_interval.has_upper_bound() &&
+                                       b_interval.has_upper_bound() &&
+                                       can_prove(b_interval.max < 0 &&
+                                                 !b_interval.max.type().is_uint() &&
```
dsharletg: This is a constant so I think it should go outside (before) the can_prove (there's no need to make the simplifer deal with this, it's probably not a problem but it also makes the code clearer IMO).

steven-johnson:
Note 1: I just landed a PR that moved most of the function bodies in IROperator.h into IROperator.cpp, so this will need some resolution there, sorry for the trouble.

Note 2: may I offer/suggest that before we land this, we pull it into google3 and do a global presubmit test, to help flush out any possible gotchas.

src/Bounds.cpp
kpassarella: Done

kpassarella:
Steven, Thanks for the heads up! The merge conflict has been resolved.

How would we go about pulling it into google3? Is there a specific process I should do for that?

Thanks!

src/Bounds.cpp
```diff
@@ -1039,11 +1039,29 @@ class Bounds : public IRVisitor {
...
+                            if (a_interval.has_lower_bound() &&
+                                b_interval.has_lower_bound() &&
+                                can_prove(b_interval.min >= 0 &&
+                                          b_interval.min < t.bits())) {
                                 interval.min = a_interval.min << b_interval.min;
+                            } else if (a_interval.has_lower_bound() &&
+                                       b_interval.has_lower_bound() &&
+                                       !b_interval.min.type().is_uint() &&
+                                       can_prove(b_interval.min < 0 &&
+                                                 cast<int>(b_interval.min) > -t.bits())) {
```
dsharletg: You don't need the cast anymore, and it would be better not to have it if possible.

steven-johnson:
> How would we go about pulling it into google3? Is there a specific process I should do for that?

It's a little fiddly, I'll describe it off-GitHub. LMK when this PR is ready and I'll get it started.

src/Bounds.cpp
kpassarella: Good point. Removed.

steven-johnson:
FYI, I'm seeing some failures of the form `Both operands to a binary operator are not of the same type!` inside google. I'll point you at the relevant bits internally.

dsharletg:
It would be helpful to push new changes as new commits, rather than rewriting the previous commit, so we can see the incremental changes.

abadams:
Oh yeah, I've been meaning to say that as a general practice we don't tend to squash commits or rebase around here, so we very rarely force-push.

steven-johnson:
I will sometimes force-push, but generally only if I have a work-in-progress with lots of changes that hasn't really been reviewed yet -- once reviews start on it, I avoid it.

kpassarella:
Good to know -- I'll avoid rebasing the commits together in the future. Sorry for making it unnecessarily difficult to see the incremental changes!

src/IROperator.cpp
```diff
@@ -515,6 +515,38 @@ void match_types(Expr &a, Expr &b) {
     }
 }
 
+// Cast to the wider type of the two. Already guaranteed to leave
+// signed/unsigned on number of lanes unchanged.
+void match_bits(Expr &x, Expr &y) {
+    if (y.type().is_int() == x.type().is_int()) {
+        // The signedness matches, so do a full cast.
```
dsharletg: It seems like this case is handled by the below code too, and it would be better to minimize branches like this (which reduce test coverage, and it's just more logic we need to ensure is correct). So, I'd remove this branch and just handle every case with the else branch.

src/IROperator.cpp
```diff
@@ -536,13 +568,8 @@ void match_types_bitwise(Expr &x, Expr &y, const char *op_name) {
...
 
-    // Cast to the wider type of the two. Already guaranteed to leave
-    // signed/unsigned on number of lanes unchanged.
-    if (x.type().bits() < y.type().bits()) {
-        x = cast(y.type(), x);
-    } else if (y.type().bits() < x.type().bits()) {
-        y = cast(x.type(), y);
-    }
+    // Cast to the wider type of the two.
+    match_bits(x, y);
```
dsharletg: Do we want this change in all cases `match_types_bitwise` is used, or just shifts? This change will affect bitwise and, or, and xor too.

kpassarella: I had been trying to reduce repeated code, but it's probably simpler to let this stay as it's own case.

src/IROperator.cpp
kpassarella: Done.

steven-johnson:
Latest version looks good in Google. LGTM.

kpassarella:
Thanks for all the reviews!

PR #4157, opened 2019-08-20
Made model.cpp compile again

benoitsteiner:


apps/onnx/Makefile
```diff
@@ -76,11 +76,11 @@ test: $(BIN)/$(HL_TARGET)/onnx_converter_test $(BIN)/$(HL_TARGET)/onnx_converter
 	LD_LIBRARY_PATH=$(BIN) $(BIN)/$(HL_TARGET)/onnx_converter_test
 	LD_LIBRARY_PATH=$(BIN) $(BIN)/$(HL_TARGET)/onnx_converter_generator_test
 
-PYTHON ?= python3
-PYBIND11_CFLAGS = $(shell $(PYTHON) -m pybind11 --includes) -frtti
+PYTHON ?= python
+PYBIND11_CFLAGS = $(shell $(PYTHON)-config --includes) -frtti -undefined dynamic_lookup
```
steven-johnson: "-undefined dynamic_lookup" is probably only going to compile on OSX.

apps/onnx/Makefile
```diff
@@ -76,11 +76,11 @@ test: $(BIN)/$(HL_TARGET)/onnx_converter_test $(BIN)/$(HL_TARGET)/onnx_converter
 	LD_LIBRARY_PATH=$(BIN) $(BIN)/$(HL_TARGET)/onnx_converter_test
 	LD_LIBRARY_PATH=$(BIN) $(BIN)/$(HL_TARGET)/onnx_converter_generator_test
 
-PYTHON ?= python3
-PYBIND11_CFLAGS = $(shell $(PYTHON) -m pybind11 --includes) -frtti
+PYTHON ?= python
```
steven-johnson: Are we going to support python2 at this point in time?

apps/onnx/model.cpp
```diff
@@ -333,7 +333,7 @@ std::vector<py::array> run(
     }
     Halide::Realization real(outputs);
     Halide::Target tgt;
-    tgt.set_feature(Halide::Target::DisableLLVMLoopOpt);
+    tgt.set_feature(Halide::Target::DisableLLVMLoopUnroll);
```
steven-johnson: DisableLLVMLoopUnroll is not a Feature anymore; this could not possibly compile at top-of-tree. DisableLLVMLoopOpt is correct.

benoitsteiner: You are correct. I updated my code locally, but I didn't rerun make test_apps from the top level so I ended up using the old version of the header files. This is fixed now.

apps/onnx/Makefile
benoitsteiner: There is no point since it's going to be supported only for a semester or so. I reverted to python3

apps/onnx/Makefile
benoitsteiner: Indeed,  this should be fixed now

PR #4159, opened 2019-08-21
Fix build when LLVM is built with libc++

pranavb-ca:
When LLVM is built using libc++, then ```llvm-config``` returns ```-stdlib=libc++``` as its value when called with ```--cxxflags```
g++ barfs at this.  So, the **easy** alternative is to use clang++ as the host compiler. However, this causes a few build issues that i have fixed in this PR.
@steven-johnson  - Could you PTAL. This is blocking my fixes for https://github.com/halide/Halide/pull/4087


src/Generator.h
```diff
@@ -1257,9 +1257,9 @@ class StubInput {
 public:
     // *not* explicit.
     template<typename T2>
-    StubInput(const StubInputBuffer<T2> &b) : kind_(IOKind::Buffer), parameter_(b.parameter_) {}
-    StubInput(const Func &f) : kind_(IOKind::Function), parameter_(), func_(f) {}
-    StubInput(const Expr &e) : kind_(IOKind::Scalar), parameter_(), expr_(e) {}
+      StubInput(const StubInputBuffer<T2> &b) : kind_(IOKind::Buffer), parameter_(b.parameter_), func_(), expr_() {}
```
steven-johnson: Why are these changes necessary?

src/Var.h
```diff
@@ -171,6 +171,8 @@ struct ImplicitVar {
     operator Expr() const {
         return to_var();
     }
+
+    constexpr ImplicitVar(){};
```
steven-johnson: Why is this change necessary?

src/Generator.h
pranavb-ca: Sorry, I think I had needed this when i was using an older compiler. Will push an update

pranavb-ca:
The failure in ```buildbot/linux-64-gcc53-testbranch-trunk``` appears to be a flake, or at least not related to this PR. The failure is 
```
/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-trunk/halide/src/Introspection.cpp:942:31: error: no matching function for call to ‘llvm::object::SectionRef::getName(llvm::StringRef&) const’
             iter->getName(name);
```
But, this line should have got #ifdef'd out because of ```-DLLVM_VERSION``` is 100 (See https://github.com/halide/Halide/blob/master/src/Introspection.cpp#L947).
@steven-johnson  ok to merge?

abadams:
That doesn't sound like a flake, but it does sound unrelated to this PR. We
must be broken with trunk again.

PR #4160, opened 2019-08-21
Default user errors and user warnings to output on a single line

steven-johnson:
Tweak the formatting of ErrorReport to use only a single line of output (aside from newlines inserted by the user <<), to make it more tractable to filter warnings/errors via grep.

(I'm not 100% sold on this change; my motivation is to be able to silence spammy warnings during autoscheduler, but there may be a better / cleaner way.)


src/Error.cpp
```diff
@@ -63,41 +63,29 @@ InternalError _internal_error("");
 }  // namespace
 
 ErrorReport::ErrorReport(const char *file, int line, const char *condition_string, int flags) : flags(flags) {
-
+    // Note that we deliberately put the entire message into a single line
+    // (aside from newlines inserted by user code) to make it easy to filter
+    // specific warnings or messages via (e.g.) grep.
     const std::string &source_loc = Introspection::get_source_location();
-
+    const char *what = (flags & Warning) ? "Warning" : "Error";
```
abadams: How about picking a separator like so: 
```
#if defined(WITH_EXCEPTIONS) || defined(_MSC_VER)
const bool use_newlines = false;
#else 
const bool use_newlines = (custom_error_reporter == nullptr) && isatty(2);
#endif  
```

steven-johnson: why WITH_EXCEPTIONS here?

abadams: I just looked at the destructor for the circumstances in which something would be printed to stdout. If exceptions are on it throws instead of printing, so isatty is not meaningful.

steven-johnson:
PTAL

PR #4163, opened 2019-08-22
Simplifications for unrolling sliding window

abadams:
Unrolling happens after sliding but before partitioning, so the extents of the unrolled loop are complex, but should have an upper bound that enables unrolling. It can be hard for the compiler to see this through all the selects. 

This PR makes things way more aggressive in the about-to-fail case in unrolling. It now exhaustively substitutes in everything for maximally aggressive simplification, and there are a bunch of new simplifier rules to help cancel things through select nodes. This exhaustive substitution is somewhat dangerous, but this is a case where the compiler is about to throw an error if it doesn't simplify down to almost a constant.

### DIFF
--- a/src/Simplify_Add.cpp
+++ b/src/Simplify_Add.cpp
@@ -73,12 +73,23 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
              rewrite(x + (y + c0), (x + y) + c0) ||
              rewrite((c0 - x) + c1, fold(c0 + c1) - x) ||
              rewrite((c0 - x) + y, (y - x) + c0) ||
+
              rewrite((x - y) + y, x) ||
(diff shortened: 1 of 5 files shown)


steven-johnson:
This PR LGTM, but applying it to the original source where I found this bug still fails with a 'can't unroll over constant extent'... but now the Expr for the non-constant-extent is much much much more complex. :-(

abadams:
I plugged your monster Expr into the super-simplifier and it's spitting out more missing rules, so stay tuned.

abadams:
Added some more rules. Please try again.

steven-johnson:
Looks good so far; let me do some more torture testing.

steven-johnson:
OK, looks good.

Issue #1582, opened 2016-10-27
Hexagon backend miscompiles non-native vector width interleaving and deinterleaving


dsharletg:
When compiling a widening operation where the input is not a multiple of the native vector size, or a narrowing operation where the output is not a multiple of the native vector size, results in a miscompile.

For example, consider a narrowing cast that interleaves the operand, vectorized for one vector of input (half a vector of output):

```
y = narrow(x)
```

To make this operation correct, we deinterleave x:

```
y = narrow(deinterleave(x))
```

`deinterleave` only works on pairs of native vectors, as does narrow. The actual sequence of operations that occurs is:

```
y = narrow(pad(slice(deinterleave(pad(x))))
```

Where `pad` and `slice` either double the vector width by padding with undef, or slicing in half.

The problem is that the `pad(slice(.))` step throws away useful vector lanes, replacing them with undef, which can lead to miscompiles.

ckun5:
I met this error when I scheduled with hexagon:

Error at
 ～/Qualcomm/Hexagon_SDK/3.3.0/tools/HEXAGON_Tools/8.1.04/Tools/Halide/tools/GenGen.cpp:4:
Interleave or deinterleave will result in miscompilation, see https://github.com/halide/Halide/issues/1582

I did as you said that to narrow the input as beblow:
       Img(w,h) = narrow(pad(slice(deinterleave(pad(input(w,h)))));
but it tells me:
       error: ‘pad’ was not declared in this scope
       error: ‘deinterleave’ was not declared in this scope
       error: ‘slice’ was not declared in this scope
       error: ‘pad’ was not declared in this scope
       error: redeclaration of ‘<typeprefixerror>pad’
       error: ‘narrow’ was not declared in this scope

So how can deal with that the input size  is not a multiple of 64/128?

dsharletg:
Sorry, that was pseudo-code to explain the problem, not a fix.

The fix is to vectorize your code such that all expressions involved are one native vector (i.e. 64 or 128 bytes, for HVX64/HVX128).

If your data is not a multiple of this vector size, you can:

- Use TailStrategy::GuardWithIf when you vectorize. This will hopefully generate semi-vectorized code in the last vector (vectorized arithmetic, scalarized load/store).
- Use TailStrategy::ShiftInwards. This is makes it hard to generate code that assumes the data is aligned though.
- Pad the input and/or output so it is a multiple of the vector size. This is what we do in practice.

dsharletg:
[...]

PR #4165, opened 2019-08-22
Restrict interleaving Hexagon pattern matches

shubhamp-ca:
This prevents matching interleaving hexagon patterns if the result can't be interleaved. Checks that the
op type is atleast 2 vector widths for widening ops.

Currently, compile-time assert is thrown in Codegen (see: https://github.com/halide/Halide/issues/1582) when the interleaved hexagon intrinsic op is not equal to two vector width. This is a recurring issue in widening ops, where the widening ops have to be vectorized by 2 * native vector width for valid codegen. This change lets the backend ISEL handle half-native vectors, instead of applying interleaved widening Hexagon intrinsics which the backend miscompiles.
@dsharletg PTAL

### DIFF
--- a/src/HexagonOptimize.cpp
+++ b/src/HexagonOptimize.cpp
@@ -246,6 +246,11 @@ Expr replace_pattern(Expr x, const vector<Expr> &matches, const Pattern &p) {
     return x;
 }
 
+bool is_double_vector(Expr x, const Target &target) {
+    int native_vector_bits = target.natural_vector_size(Int(8)) * 8;
+    return (x.type().lanes() / ( native_vector_bits / x.type().bits()) >= 2);
+}
+
 // Attempt to apply one of the patterns to x. If a match is
 // successful, the expression is replaced with a call using the
 // matched operands. Prior to substitution, the matches are mutated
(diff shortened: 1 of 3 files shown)


src/HexagonOptimize.cpp
```diff
@@ -246,6 +246,11 @@ Expr replace_pattern(Expr x, const vector<Expr> &matches, const Pattern &p) {
     return x;
 }
 
+bool is_double_vector(Expr x, const Target &target) {
+    int native_vector_bits = target.natural_vector_size(Int(8)) * 8;
+    return (x.type().lanes() / ( native_vector_bits / x.type().bits()) >= 2);
```
dsharletg: I think this needs to be % 2 == 0, not >= 2? Or more clearly (IMO):

    int native_vector_lanes = target.natural_vector_size(x.type());
    return x.type.lanes() % (2 * native_vector_lanes) == 0;

shubhamp-ca: Changed. Thanks.

shubhamp-ca:
> I'm not seeing any false positives now, thanks!

If this is the case, I guess having is_double_vector() check before expr_match() is the better thing to do, as it saves the effort of having to pattern match an expression which shouldn't be. Which is to say -- `(expr_match && is_double_vector)`   is same as -- `(is_double_vector && expr_match)`, but the latter may be a bit more efficient?

src/HexagonOptimize.cpp
```diff
@@ -269,6 +274,11 @@ Expr apply_patterns(Expr x, const vector<Pattern> &patterns, const Target &targe
                 continue;
             }
 
+            // Don't apply pattern if it does not involve an
```
dsharletg: This comment is a bit confusing. I think it should be rewritten "Don't apply pattern if it involves an interleave, and it is not a multiple of two vectors" and link to the issue #1582.

shubhamp-ca: Done. Thanks.

steven-johnson:
clone-for-build looks clean. ready to land?

PR #4169, opened 2019-08-23
Remove min_iter and max_iter from benchmark & RunGen

steven-johnson:
These aren't really that useful; min_time is usually the only knob of this sort that is interesting to fiddle with. Remove them entirely to simplify code and reduce the temptation to mess with them. (Some versions of gbenchmark don't support min_iter and max_iter at all, only min_time.)

### DIFF
--- a/tools/RunGen.h
+++ b/tools/RunGen.h
@@ -1208,9 +1208,7 @@ public:
         }
     }
 
-    void run_for_benchmark(double benchmark_min_time,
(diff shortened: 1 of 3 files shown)


abadams:
SGTM. I would really like a way to be able to set number of iterations and number of samples directly. If you're comparing two schedules and checking the performance counters with perf (e.g. which schedule has more page faults?), you want them both to run exactly the same number of times.

PR #4175, opened 2019-08-26
HelloPyTorch: allow for mixed python2 + python3 installs

steven-johnson:
I'm sure we don't want to bother supporting python2 anymore, but we probably still want to support running on systems that have both py2 and py3 installed (at least until Jan 1 2020); therefore, bottleneck python into a Makefile var and default it to be explicitly 'python3'.

attn @mgharbi


steven-johnson:
Ping to @mgharbi for feedback on this

mgharbi:
Looks good to me. The prints are ugly, but the app builds and test pass also with Python2.7.

steven-johnson:
Asking from ignorance: does PyTorch still support Python2?

mgharbi:
Maybe we can add `from __future__ import print_function` atop `test.py`.
Yes, PyTorch still support python2.7

steven-johnson:
> Maybe we can add `from __future__ import print_function` atop `test.py`.

PTAL

mgharbi:
I confirm this works for python2/3 on my end.

steven-johnson:
cool, ping to J. Random Reviewer for approval

steven-johnson:
Monday Morning Review Ping

PR #4186, opened 2019-08-27
Fix More LLVM DataLayouts

steven-johnson:


abadams:
If you remember, how did you determine the datalayout for platforms for which we don't have testing (e.g. x86-32-ios)? Is there some place that lists all llvm datalayouts in the llvm tree?

PR #4192, opened 2019-08-28
clang-tidy: avoid 'code will never be executed' static checks

steven-johnson:
Putting `false` inside extra parens, ie `(false)`, is apparently the trick that clang-tidy wants. (The comment is just for emphasis.)

(Ordinarily I'd say code of this sort should be removed or commented out with explanation. Maybe that should happen here too?)


abadams:
These things are both if (false) with an explanation for why you might want to turn them on while testing alternatives, so they shouldn't be removed. The second one could be commented out instead, but that would be slightly more involved for the first one, and would make it slightly more annoying to toggle between alternatives.

We could also do a preprocessor if 0, but I find those much uglier than if (false)

My vote is for if ((false)) {  // Intentional dead code. Extra parens to pacify clang-tidy.

steven-johnson:
Done

PR #4200, opened 2019-08-29
Replace getopt() with header-only flags parser

steven-johnson:
getopt() doesn't exist for MSVC; there are ports available, but if we have to add code, let's add something that doesn't include weird extra globalvars for parsing, the way getopt does. I somewhat arbitrarily chose a simple header-only BSD/MIT-licensed library that seems to be reasonable (https://github.com/tanakh/cmdline). Thoughts?

### DIFF
--- a/apps/autoscheduler/retrain_cost_model.cpp
+++ b/apps/autoscheduler/retrain_cost_model.cpp
@@ -1,6 +1,5 @@
 #include <cmath>
 #include <fstream>
-#include <getopt.h>
 #include <iomanip>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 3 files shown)


shubhamp-ca:
This library seems to require RTTI to be turned on. Building retrain_cost_model doesn't explicitly specify -frtti flag and is failing with -fno-rtti (set by HALIDE_RTTI).

```
../support/cmdline.h:129:19: error: use of typeid requires -frtti
  return demangle(typeid(T).name());

../support/cmdline.h:379:35: error: use of dynamic_cast requires -frtti
    const option_with_value<T> *p=dynamic_cast<const option_with_value<T>*>(options.find(name)->second);
```

steven-johnson:
We should enable rtti there then.

PR #4203, opened 2019-09-03
Add OpenCL program caching with clCreateProgramWithBinary

brotherofken:
Purpose of this pull request is to speed up `halide_opencl_initialize_kernels` by caching of CL program intermediate representation on disk.

User have to call `halide_opencl_set_compiled_programs_cache_dir("path")` in runtime before any generated function call in order to specify directory where OpenCL runtime stores compiled programs.

OpenCL program files have the following form: `halide_opencl_program_binary_%x.bin`

There are two new functions in `opencl.cpp`:
1. `cl_load_program_from_cache` - checks if program exists in cache and loads it
2. `cl_cache_program_binary` - checks if program exists in cache and saves program

P.S. Good desciption of what's happening: https://community.amd.com/thread/223853 

Also I added stub test `test/generator/opencl_program_caching_aottest.cpp` but not sure how to test caching behaviour properly.


brotherofken:
Seems that I mixed internal and external changed in this pull request. It's my bad. I'm going to close this request and create new one.

PR #4206, opened 2019-09-03
Add Linker script to hide all non-Halide symbols in .so files.

steven-johnson:
Probably doesn't move the needle much, but the goal here is to hide all of the non-Halide symbols (in particular, llvm entry points) to reduce the surface area of our shared library and (perhaps) slightly improve load time. Not entirely sure if worthwhile. Comments welcome.


abadams:
I guess the main reason for this is to avoid conflicts between Halide's llvm and some other llvm used in the same process? I'm very curious to know if it impacts load time or library size.

steven-johnson:
Update: apparently OSX linker doesn't support `--version-script`? Oh well.

steven-johnson:
Was mainly hoping it would impact load time and/or binary size. 

Binary size seems to be only trivially smaller (3% or so on linux).

As for load time, not sure of the best way to test it; using `time make -j72 test_correctness` (after ensuring everything already built) before-and-after showed a slight improvement; typical results on master are

```
real	0m52.058s
user	14m46.420s
sys	0m40.145s
```

while on this branch I get results like

```
real    0m47.778s
user    13m19.411s
sys     0m45.674s
```

steven-johnson:
BTW: this exposes all of the `Internal` stuff too, because (1) a lot of the tests need it, and (2) various branch experimenters might need it. In a good and just world, though, we'd be hiding all that by default; if it isn't really internal, we shouldn't label it as such.

steven-johnson:
Screen Shot 2019-09-03 at 3 47 03 PM

steven-johnson:
I'm going to abandon this, as there doesn't seem to be any compelling interest in it. (I do think that hiding the LLVM symbols is good pre-emptive hygiene though; we may find reasons to revisit this someday.)

PR #4207, opened 2019-09-03
Add a way for binary2cpp to specify visibility=hidden; use this for all our halide_internal_initmod data

steven-johnson:
(I proposed a similar change a while back, but with the hiding on by default; this swaps it to be opt-in, enabling it only for the initmod stuff. The goal here is similar to https://github.com/halide/Halide/pull/4206, i.e., to painlessly hide symbols that shouldn't ever be necessary outside of Halide; handling these in this way is slightly more robust and portable than the linker script option.)

### DIFF
--- a/tools/binary2cpp.cpp
+++ b/tools/binary2cpp.cpp
@@ -13,49 +13,79 @@
 // chars. Similar to the xxd utility.
 
 static int usage() {
-    fprintf(stderr, "Usage: binary2cpp identifier [-header]\n");
+    fprintf(stderr, "Usage: binary2cpp identifier [-header] [-hidden]\n");
     return -1;
 }
 
+void emit_hidden_macro(const char *target) {
+    // Note that visibility=hidden isn't supported under MinGW, regardless of compiler
+    printf("#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)\n");
+    printf("#define HALIDE_BINARY2CPP_%s_HIDDEN_DATA\n", target);
+    printf("#else\n");
+    printf("#define HALIDE_BINARY2CPP_%s_HIDDEN_DATA __attribute__((visibility(\"hidden\")))\n", target);
+    printf("#endif\n");
+}
+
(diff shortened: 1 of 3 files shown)


tools/binary2cpp.cpp
```diff
@@ -13,49 +13,79 @@
...
+    }
+
+    if (header) {
+        printf("#ifndef _H_%s_binary2cpp\n", target);
+        printf("#define _H_%s_binary2cpp\n", target);
+        if (hidden) {
+            emit_hidden_macro(target);
+        }
+        printf("extern \"C\" {\n");
+        printf("extern %s unsigned char %s[];\n", vis, target);
```
dsharletg: I don't think you need the visibility attribute in the header, only on the definition (possibly only on one of the declarations regardless of whether it is a definition or not). Does that make it easier to generate this file at all? It seems like it might?

steven-johnson: Yeah, probably. (Though AFAICT we don't currently ever use the -header mode, so it's probably moot...)

steven-johnson: Interestingly, from a quick google, I can't find definitive documentation on what happens if you get these attributes mismatched between declaration and definition. Though I guess the answer is likely "it's fine" since we're essentially doing extern redeclarations of these, without the attribute, in LLVM_Runtime_Linker.cpp.

steven-johnson:
Monday morning review ping: any thoughts on this?

abadams:
I'm not sure I understand the benefit of hidden visibility in general, for things that are already namespaced. Could you elaborate on the motivation?

steven-johnson:
The motivations for this (and other related-but-now-closed PRs) was:
- pre-emptive cleanliness; external code can't rely on internal apis and symbols if they are literally hidden (rather than merely namespaced into a 'please don't use me' space)
- extremely minor incremental improvements in code size and load times

I'm not sure I see any downside to hiding these symbols, but not gonna keep arguing for it; we can resurrect this if we ever end up needing the hygiene.

PR #4209, opened 2019-09-04
Improve code in Codegen_Hexagon.cpp

steven-johnson:
Reworked the intrinsic_wrappers table into a table of constant data (instead of one that must be laboriously constructed at runtime); previously, Codegen_Hexagon::init_module() was ~40k of code (the second largest function in all of libHalide); now, it's ~1k of code and ~9k of constant data.

As part of this:
- marked some halide_type_t methods as `constexpr` to allow for efficient use of `halide_type_t` in compile-time constant data sections
- added `halide_type_t::with_lanes()`
- replaced IPICK with MAKE_ID_PAIR, since we now may need to compute both ids at compile time and select one at runtime


src/CodeGen_Hexagon.cpp
```diff
@@ -326,350 +313,386 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
...
+constexpr halide_type_t i8v2 = i8v1.with_lanes(i8v1.lanes * 2);
+constexpr halide_type_t i16v2 = i16v1.with_lanes(i16v1.lanes * 2);
+constexpr halide_type_t i32v2 = i32v1.with_lanes(i32v1.lanes * 2);
+constexpr halide_type_t u8v2 = u8v1.with_lanes(u8v1.lanes * 2);
+constexpr halide_type_t u16v2 = u16v1.with_lanes(u16v1.lanes * 2);
+constexpr halide_type_t u32v2 = u32v1.with_lanes(u32v1.lanes * 2);
+
+// LLVM's HVX vector intrinsics don't include the type of the
+// operands, they all operate on vectors of 32 bit integers. To make
+// it easier to generate code, we define wrapper intrinsics with
```
dsharletg: This comment seems partially misplaced, the wrapper intrinsics are still defined in init_module. This table just provides a new name and the arg/return types of the intrinsics.

src/CodeGen_Hexagon.cpp
```diff
@@ -326,350 +313,386 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
...
+        Intrinsic::ID id = i.ids.get(is_128B);
+        internal_assert(id != Intrinsic::not_intrinsic);
+        // Get the real intrinsic.
+        llvm::Function *intrin = Intrinsic::getDeclaration(module.get(), id);
+        arg_types.clear();
+        for (const auto &a : i.arg_types) {
+            if (a.bits == 0) {
+                break;
+            }
+            const int lanes_actual = a.lanes > 1 ?
```
dsharletg: Doesn't the return type need this treatment too?

src/CodeGen_Hexagon.cpp
steven-johnson: Done.

src/CodeGen_Hexagon.cpp
steven-johnson: Good catch!

steven-johnson:
Looks like MSVC is complaining about the `constexpr` declarations... not sure if glitch in MSVC or subtle bug in this patch.

dsharletg:
Do you need all of the new constexpr stuff? I think the compiler is likely to optimize the code the same way without it. This change still avoids the large control flow graph that the old code had, which was probably the main issue, regardless of the constexpr change.

steven-johnson:
The constexpr stuff means that the LUT can boil down to a strict chunk of compile-time read-only data baked into the executable, so yes, it's worth keeping.

dsharletg:
Are you sure the compiler doesn't just do all that via non-explicit optimization anyways? All the compiler has to do is inline IdPair's constructor, halide_type_t's constructor, and halide_type_t::with_lanes (and do some constant folding), all of which seem very likely. 

it would be better to make it explicit with constexpr of course, I'm just suggesting that we go ahead with this PR without it, and come back to it later, after we have MSVC2019.

pranavb-ca:
@steven-johnson  - Thanks for the PR. Can we wait for a few days before i have a chance to finish https://github.com/halide/Halide/pull/4087 off, which fixes simd_op_check and this PR really needs the simd_op_check test, IMHO. I am OOO and I am hoping to finish the PR after I am back later this week (hopefully, by Sept 13).

steven-johnson:
@dsharletg -- yes, good point; current compilers might be smart enough to do this anyway. I'll look at output and see.

@pranavb-ca -- absolutely, there is no reason to rush this at all, it was purely code tweaking on my part.

dsharletg:
> yes, good point; current compilers might be smart enough to do this anyway. I'll look at output and see.

Even if it doesn't currently, this PR is still a good step towards what we want. Basically there's 3 options I see.
1. Let this PR dangle until we use MSVC2019.
2. Close it and try again when we have MSVC2019.
3. Merge it and just try adding constexpr when we have MSVC2019.

I think 3 is the best thing to do.

steven-johnson:
Trying option 4: make the constexpr stuff conditional, so that we only use it for VC if we know it is >= VC2019.

abadams:
I was under the impression that because constexpr has to be checked my the compiler, it has to be able to perfectly detect when something could be constexpr too, and thus constexpr *never* enables optimizations that would not otherwise have happened. It's strictly a request to throw an error if the optimization doesn't occur. So if in doubt, why not leave it out?

PR #4210, opened 2019-09-04
De-inline most of IROperator.h

steven-johnson:
IROperator.h has way too much code in it; this moves almost all of the non-templated inline functions into IROperator.cpp (along with some drive-by anon-namespacing of internal functions). 

This is mostly just an attempt at hygiene, as ~none of these functions really needed to be inlined; that said, this does seem to produce a small but measurable improvement in compilation performance on my Linux box (recompiling libHalide.a from scratch drops to about 97% of user time).

Note that I didn't attempt to update any calling conventions; I suspect that updating various calls to pass Expr and/or Type by const-ref (rather than value) might be profitable.

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <atomic>
 #include <cmath>
 #include <iostream>
 #include <sstream>
@@ -9,6 +10,7 @@
 #include "IRMutator.h"
 #include "IROperator.h"
 #include "IRPrinter.h"
+#include "Util.h"
 #include "Var.h"
 
 namespace Halide {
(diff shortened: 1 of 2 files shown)


abadams:
I'd be shocked if this showed anything, but it would be good to check something like local_laplacian.generator -e stmt to confirm that none of these functions benefit from inlining.

steven-johnson:
Uh, ok, I'll try that, but I will be genuinely shocked if it shows anything

abadams:
While user-code calls to these functions aren't going to show up, things like operator+ are used very heavily inside the compiler, and there could be some benefit to inlining them.

steven-johnson:
Oh, wait, you mean *profiling* that, not comparing the output. Yes, that makes sense.

steven-johnson:
Testing with `time for i in {1..100}; do nice -n10 bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host; done` on my Linux box:

current master:

```
real	1m4.725s
user	1m3.729s
sys	0m0.946s
```

this PR:

```
real	1m4.108s
user	1m3.285s
sys	0m0.779s
```

PR #4219, opened 2019-09-06
Python: import imread/imsave from imageio instead of scipy.misc

steven-johnson:
imread and imsave are deprecated and not present in current versions of scipy; using equivalent functions in imageio is the recommended replacement.


steven-johnson:
Looks like `imageio` needs to be installed on the Mac buildbot... but it's responding properly to Chromote logins (it probably got wedged on a permission screen during a recent update to Chromoting). I'll go by in person on Monday and fix the issue; we shouldn't land this until then.

Issue #3957, opened 2019-06-24
Tracking issue for disablement of LLVM NewPassManager


steven-johnson:
The changes made in #3868 are being temporarily disabled: they have injected a pathological compilation slowdown in some code inside Google, and while @alinas has been investigating, she's going OOO for a bit. Since this change isn't urgent (ie LLVM isn't removing the old pass managaer API anytime soon) we're going to temporarily disable it to make downstream merges simpler.

steven-johnson:
@alinas -- where does this stand? Have there been changes upstream that would warrant retrying this?

alinas:
IIRC, last we tried re-testing the tests that used to timeout, it was no longer the case.
So, yes, it's worth trying to re-enable this.

steven-johnson:
Re-enabling it inside Google is producing some numerically different results in some cases (ie, not a slowdown but a change in computation).

steven-johnson:
Looks like we needed to set `pto.SLPVectorization = true` in the new case to match the old case

PR #4221, opened 2019-09-06
Re-enable LLVM NewPassManager

steven-johnson:
See https://github.com/halide/Halide/issues/3957 for details.


steven-johnson:
Seems to be clean on builds inside Google; PTAL

steven-johnson:
Failures are unrelated.

PR #4223, opened 2019-09-08
fix python_bindings bad float cast

mugwort-rc:
always double to float cast.

----------

Test code

```python
import halide
import numpy

def run_test():
    test = halide.Func("test")
    x = halide.Var("x")
    test[x] = x * halide.f64(0.12345678901234)
    for i, val in enumerate(numpy.asarray(test.realize(10))):
        print("{}: {:.20f}".format(i, val))
```

before:
```
>>> run_test()
0: 0.00000000000000000000
1: 0.12345679104328155518
2: 0.24691358208656311035
3: 0.37037037312984466553
4: 0.49382716417312622070
5: 0.61728395521640777588
6: 0.74074074625968933105
7: 0.86419753730297088623
8: 0.98765432834625244141
9: 1.11111111938953399658
```

after:
```
>>> run_test()
0: 0.00000000000000000000
1: 0.12345678901234000135
2: 0.24691357802468000271
3: 0.37037036703701997631
4: 0.49382715604936000542
5: 0.61728394506170003453
6: 0.74074073407403995262
7: 0.86419752308637998173
8: 0.98765431209872001084
9: 1.11111110111105992893
```

### DIFF
--- a/python_bindings/correctness/float_precision_test.py
+++ b/python_bindings/correctness/float_precision_test.py
@@ -1,3 +1,5 @@
+import sys
+
 import halide as hl
 import numpy
 
@@ -7,8 +9,11 @@ def test():
         x = hl.Var("x")
         f = hl.Func("f")
         f[x] = x * c
-        for i, v in enumerate(numpy.asarray(f.realize(10))):
-            assert v == (i * c), "{}[{}]: {} != {}".format(i, c, v, (i * c))
+        for i, hl_value in enumerate(numpy.asarray(f.realize(10))):
+            py_value = (i * c)
+            delta = hl_value - py_value
+            check = delta < sys.float_info.epsilon
+            assert check, "{}[{}]: {} != {}".format(i, c, hl_value, py_value)
 
     test_pattern(0.123456789012345678)
     test_pattern(0.987654321098765432)
--- a/python_bindings/correctness/float_precision_test.py
+++ b/python_bindings/correctness/float_precision_test.py
@@ -1,4 +1,4 @@
-import sys
+import math
 
 import halide as hl
 import numpy
@@ -11,8 +11,7 @@ def test():
         f[x] = x * c
         for i, hl_value in enumerate(numpy.asarray(f.realize(10))):
             py_value = (i * c)
-            delta = hl_value - py_value
-            check = delta < sys.float_info.epsilon
+            check = math.isclose(hl_value, py_value)
             assert check, "{}[{}]: {} != {}".format(i, c, hl_value, py_value)
 
     test_pattern(0.123456789012345678)
--- a/python_bindings/correctness/float_precision_test.py
+++ b/python_bindings/correctness/float_precision_test.py
@@ -1,0 +1,18 @@
+import halide as hl
+import numpy
+
+
+def test():
+    def test_pattern(c):
+        x = hl.Var("x")
+        f = hl.Func("f")
+        f[x] = x * c
+        for i, v in enumerate(numpy.asarray(f.realize(10))):
+            assert v == (i * c), "{}[{}]: {} != {}".format(i, c, v, (i * c))
+
+    test_pattern(0.123456789012345678)
+    test_pattern(0.987654321098765432)
+
+
+if __name__ == "__main__":
+    test()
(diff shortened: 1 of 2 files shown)


python_bindings/src/PyExpr.cpp
```diff
@@ -22,7 +22,7 @@ void define_expr(py::module &m) {
             // PyBind11 searches in declared order,
             // int should be tried before float conversion
             .def(py::init<int>())
-            .def(py::init<float>())
+            //.def(py::init<float>())
```
abadams: Please add a comment explaining why the line is commented out, or just delete it.

steven-johnson:
Can you expand a bit more on why the existing code was bad? It's not clear to me from casual inspection.

mugwort-rc:
OK.

This example uses the following change:

```cpp
            .def(py::init([](int v) -> Expr {
                std::cout << "Expr(int)" << std::endl;
                return v;
            }))
            .def(py::init([](float v) -> Expr {
                std::cout << "Expr(float)" << std::endl;
                return v;
            }))
            .def(py::init([](double v) -> Expr {
                std::cout << "Expr(double)" << std::endl;
                return Expr(v);
            }))
            .def(py::init([](const std::string &v) -> Expr {
                std::cout << "Expr(std::string)" << std::endl;
                return v;
            }))
```

### Example of implicit case

Python float -> C float -> Halide::Expr(float)

```
>>> x = halide.Var("x")
>>> x + 1.0
Expr(float)
<halide.Expr of type float32>
```

### Example of explicit case

Python float -> (C float (?!)) -> Halide Expr

```
>>> halide.f64(1.0)
Expr(float)
<halide.Expr of type float64>
```

because `halide.f64()` is defined [here](https://github.com/halide/Halide/blob/master/python_bindings/src/PyConciseCasts.cpp#L7)

```
    m.def("f64", Halide::ConciseCasts::f64);
```

and `Halide::ConciseCasts::f64` is defined [here](https://github.com/halide/Halide/blob/master/src/ConciseCasts.h#L18)

```
inline Expr f64(Expr e) {
    return cast(Float(64, e.type().lanes()), e);
}
```

By the way, according to the Python document:

https://docs.python.org/3/library/stdtypes.html#typesnumeric

> Floating point numbers are usually implemented using double in C

Therefore, python_bindings should be as follows:

### implicit case

Python float -> C double -> Halide::Expr(double)

```
>>> x = halide.Var("x")
>>> x + 1.0
Expr(double)
<halide.Expr of type float64>
```

### explicit case

Python float -> (C double) -> Halide Expr

```
>>> halide.f64(1.0)
Expr(double)
<halide.Expr of type float64>
```

steven-johnson:
OK, so if I understand correctly, this really sums up as: 

> The Python `float` type is ~always a double in C++; if we have a `float` conversion, we may incorrectly lose precision, so don't even attempt it.

If so, I'd like to see something like that added as a comment, so that future maintainers don't think we accidentally omitted it. Otherwise, LGTM.

abadams:
In Halide we never implicitly create a double constant, as double constants in imaging pipelines are almost always unintentional. I think the current implicit behavior is fine, but clearly the explicit f64 construction going via a C float is a bug.

steven-johnson:
> the explicit f64 construction going via a C float is a bug.

We definitely need test cases added to this patch, to ensure that the new implicit and explicit behaviors are what is documented in the comment above.

python_bindings/correctness/float_precision_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+def test():
+    def test_pattern(c):
+        x = hl.Var("x")
+        f = hl.Func("f")
+        f[x] = x * c
+        for i, v in enumerate(numpy.asarray(f.realize(10))):
+            assert v == (i * c), "{}[{}]: {} != {}".format(i, c, v, (i * c))
+
+    test_pattern(0.123456789012345678)
```
steven-johnson: Thanks for adding a test, but I suspect this will be a fragile test; testing exact matches for numeric operations on doubles is usually fraught with flakiness, unless you know you are always running on exactly the same runtime environment. Comparisons of this sort really need to be done with an epsilon of some sort.

PR #4224, opened 2019-09-09
Fix minor clang-tidy warnings

steven-johnson:


steven-johnson:
Added a few more calls to internal_assert() for non-null ptrs

abadams:
What is your clang-tidy invocation that's finding these? Should we be adding a clang-tidy makefile test target, and eventually a clang-tidy presubmit test?

steven-johnson:
> What is your clang-tidy invocation that's finding these? 

I ran tests inside google using the Static Analyzer with `clang-analyzer-core.*` enabled; I've been working thru the warnings at odd moments.

> Should we be adding a clang-tidy makefile test target, and eventually a clang-tidy presubmit test?

Probably at some point... but since this is super-low-importance, I figured it would be sensible to get things (mostly) clean inside the google builds first, to minimize possible noise later on.

PR #4227, opened 2019-09-09
Ensure that ~ErrorReport is marked as noreturn

steven-johnson:
C++ wasn't aware that ~ErrorReport() never returns; as a result, clang-tidy can issue complaints about e.g. "possible null pointer access" that aren't possible because the code is guarded by `internal_assert` or similar. This PR adds the C++11 [[noreturn]] annotation as needed to ensure that the compiler does in fact know this; whether this has meaingful impact on code quality isn't clear, but at a minimum, it allows us to enable clang-tidy checks that previously would have sprayed too many false-positive warnings to be usable.

(Note of course that the 'never returns' doesn't apply to warnings; to allow for this, I split the functionality into ErrorReport and WarningReport accordingly.)


src/Error.h
```diff
@@ -92,13 +94,31 @@ struct ErrorReport {
      * this by only actually throwing if there isn't an exception in
      * flight already.
      */
-#if __cplusplus >= 201100 || _MSC_VER >= 1900
-    ~ErrorReport() noexcept(false);
+#if (__cplusplus >= 201100 || _MSC_VER >= 1900) && defined(WITH_EXCEPTIONS)
+    [[ noreturn ]] ~ErrorReport() noexcept(false);
 #else
-    ~ErrorReport();
+    [[ noreturn ]] ~ErrorReport();
```
abadams: If I'm reading it right, this says:

if C++11 and WITH_EXCEPTIONS
   use C++11 thing and exception thing
else
   use C++11 thing
endif

So in the case where C++11 is off, we're falling back to using a C++11 construct. But we require C++11 now, so maybe we should just remove the C++11 clause of the if?

steven-johnson: Heh, yeah, good point; that clause must be there for quite a while. I'll double-check the edit history but will likely remove the C++11 check.

steven-johnson:
Does anyone actually use CustomErrorReporter? AFAICT the self-test is the only use inside Halide.

src/Error.h
steven-johnson: ...it wasn't even the correct check for C++11; canonically, it should be `__cplusplus >= 201103`

PR #4229, opened 2019-09-10
Alter default extension for generated C/C++ source

steven-johnson:
We currently default to either ".c" (when calling compile_to_c()) or ".cpp" (via Generator). The former is actually wrong (since we generate C++ code, not C).

This changes both to use ".halide_generated.cpp" as the extension; this is admittedly long but ensures uniqueness for the output file, as well as making the contents clearer.

(I also considered ".generated.cpp" and ".halide.cpp" as shorter extensions.)


steven-johnson:
Failure is unrelated.

PR #4232, opened 2019-09-11
Allow `make test_apps` to run in parallel

steven-johnson:
We currently run the sub-makes serially; running them in parallel is much faster. (I'm assuming that none of the tests are time-sensitive.)


abadams:
The apps tests report benchmarking results, so running them in parallel is not great. Is it possible to configure it so that the apps *build* in parallel and then run serially? I.e. do "make some_build_target" in each app folder in parallel then "make test" serially?

steven-johnson:
Yes, but we'd have to restructure things nontrivially; right now the only contract is that each app is expected to have a `test` target that builds and runs tests. We'd have to expand that to have separate targets for building and running.

Note that we already have a `benchmark_apps` target that does exactly this; perhaps we should move performance-sensitive tests there, and add that to the mix?

steven-johnson:
> The apps tests report benchmarking results

Just to push back on this a little bit -- is anything (or anyone) actually checking/verifying the benchmark results when run via `test_apps`? If so, why is the target still passing with this change in place? Or, if not, should checking be added?

abadams:
Fair enough.

PR #4238, opened 2019-09-12
Make everything in win32_math.ll alwaysinline

steven-johnson:
We used to mark some of these as noinline, with the claim that "otherwise llvm sneakily peephole optimizes things like (float)(ceil((double)f)) -> ceilf(f), even though ceilf doesn't exist.""

However, this causes another problem; since this module is always inlined into generated code by LLVM_Runtime_Linker, anything that is marked 'noinline' gets duplicated in each AOT section, making it impossible to link multiple AOT outputs into a single binary without using the /FORCE:MULTIPLE flag, which is suboptimal.

Since posix_math.ll seems to be avoiding the peepholing issue mentioned above, we're marking everything in this file as alwaysinline (as expected by LLVM_Runtime_Linker) for now, and will re-examine the peephole issues if/when they appear.

attn @ghost-luck


steven-johnson:
Monday Morning Review Ping

steven-johnson:
> Hopefully the original comment refers to a long-fixed bug.

Crossing my fingers

PR #4240, opened 2019-09-12
Proposal: replace `Outputs` with `map<Output, string>`

steven-johnson:
The idea here is to generalize the Outputs mess:
- the Outputs struct is replaced by a map, where keys are enums analogous to the existing Output struct names (e.g. "object") and the values are the pathnames
- we have a single table of Truth for all the known outputs, and the filename extensions for each one

This makes adding new output types much much simpler (currently there are many places you must update by adding the field, checking the right extension, etc); it also puts the source of truth for filename extensions into a data structure rather than code, so things like verifying validity, fixing extensions, etc becomes trivial. Downside is it introduces a breaking change in the API; since it's such an easy thing to update, I didn't bother including a deprecated wrapper yet (though that could be done if desired).

(Normally I prefer to avoid this sort of dynamic structure for a fixed set of options; in a language with introspection the struct would continue to be preferred, but since this is C++. this approach is I think the less-bad one.)


abadams:
Wouldn't `map<some enum, string>` be better than `map<string, string>`?

steven-johnson:
> map<some enum, string>

Hmm... In C++, yes, arguably. (And maybe also in Python.) We'd have to add an enum-to-string mapping for the Generator cmd line args but that's easy. Let me take a stab at that.

abadams:
I like it. It's still marked as a draft, so I didn't go line-by-line. LMK if I should.

steven-johnson:
Still some details to fix, will update when ready to review

steven-johnson:
I think this is ready for review; the one thing I'm unsure about is naming conventions for the C/C++ generation options. Right now we use 
- `cpp_stub` for C++ Generator Stub creation
- `c_source` for the C++ backend
- `c_header` for the C/C++ header files for AOT compilation. 

I'd argue that using `cpp_source` and `cpp_header` for the latter two are probably the best option (even though the header files are intended to be plain-C compatible if you build without the cpp-name-mangling feature). Thoughts?

steven-johnson:
All green, PTAL

Makefile
```diff
@@ -1,9 +1,9 @@
 # 'make' builds libHalide.a, the internal test suite, and runs the internal test suite
 # 'make run_tests' builds and runs all the end-to-end tests in the test subdirectory
 # 'make {error,performance}_foo' builds and runs test/{...}/foo.cpp for any
-#     cpp file in the corresponding subdirectory of the test folder
+#     c_source file in the corresponding subdirectory of the test folder
 # 'make correctness_foo' builds and runs test/correctness/foo.cpp for any
-#     cpp file in the correctness/ subdirectoy of the test folder
+#     c_source file in the correctness/ subdirectoy of the test folder
```
abadams: subdirectoy

apps/linear_algebra/Makefile
```diff
@@ -6,7 +6,7 @@ CXXFLAGS += -Wall -march=native -I /opt/local/include
 HL_TARGET ?= host
 LIBHALIDE_BLAS = $(BIN)/libhalide_blas.a
 HALIDEBLAS_FLAGS ?= -DUSE_HALIDE
-EMIT_OPTIONS = stmt,assembly,o,h
+EMIT_OPTIONS = stmt,assembly,object,h
```
abadams: h -> c_header, based on what you have elsewhere

src/Generator.cpp
```diff
@@ -944,55 +896,48 @@ int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
...
-                     << " not one of [assembly, bitcode, cpp, h, html, o, static_library, stmt, cpp_stub, registration, pytorch_wrapper], ignoring.\n";
-            }
+        // Build a reverse lookup table. Allow some legacy aliases on the command line,
+        // to allow legacy build systems to work more easily.
+        std::map<std::string, Output> output_name_to_enum = {
+            {"cpp", Output::c_source},
+            {"h", Output::c_header},
+            {"html", Output::stmt_html},
+            {"o", Output::object},
+            {"py.c", Output::python_extension},
```
abadams: I wonder if we should use the extension map to autopopulate this. E.g. so that I can say "s" or "ll" instead of assembly,llvm_assembly. It's easier to remember the file extensions that I want than the enum names.

steven-johnson: Hmm... worth thinking about, but easy enough to add later if we want.

Makefile
steven-johnson: done

apps/linear_algebra/Makefile
steven-johnson: done

steven-johnson:
any thoughts on the cpp-vs-c naming convention?

PR #4246, opened 2019-09-13
Revert "fix python_bindings bad float cast"

steven-johnson:
Reverts halide/Halide#4223

Unfortunately, I neglected to run the python tests before landing this (bad on me), and it has injected some failures that I think we need to discuss before trying to fix-forward; therefore I'm going to roll back this change for now, and re-open it with the issues for discussion.


steven-johnson:
attn @mugwort-rc

PR #4247, opened 2019-09-13
fix python_bindings bad float cast (v2)

steven-johnson:
https://github.com/halide/Halide/pull/4223 was landed, then reverted, because it injected some failures into the python tests; mainly, this makes the default value for float expressions float64 instead of float32, and it's not clear whether that's what we want or not.

e.g. given

```
    x = hl.Var('x')

    # This previously succeeded, but with this patch, it fails
    assert ((x/2.0)).type() == hl.Float(32)

    # This previously would have failed, but with this patch, it succeeds
    assert ((x/2.0)).type() == hl.Float(64)
```

attn: @mugwort-rc


abadams:
This is what my complaint was about on the other PR. We don't want the literal 2.0 to be implicitly a double.

mugwort-rc:
hmmm...

According to the C++ reference:
<https://en.cppreference.com/w/cpp/language/floating_literal>

> suffix, if present, is one of f, F, l, or L. The suffix determines the type of the floating-point literal:
>
> * (no suffix) defines `double`

As described #4223, Python is running with `double`.

Both language is implicitly using `double` by default...

mugwort-rc:
Or, should be like this?

```diff
diff --git a/python_bindings/src/PyConciseCasts.cpp b/python_bindings/src/PyConciseCasts.cpp
index 9a08b458c..93a38d377 100644
--- a/python_bindings/src/PyConciseCasts.cpp
+++ b/python_bindings/src/PyConciseCasts.cpp
@@ -4,7 +4,9 @@ namespace Halide {
 namespace PythonBindings {
 
 void define_concise_casts(py::module &m) {
-    m.def("f64", Halide::ConciseCasts::f64);
+    m.def("f64", [](double v) {
+        return Halide::ConciseCasts::f64(Expr(v));
+    });
     m.def("f32", Halide::ConciseCasts::f32);
     m.def("i64", Halide::ConciseCasts::i64);
     m.def("i32", Halide::ConciseCasts::i32);
```

abadams:
C++ and Python use double for numeric types when no suffix is present. Halide is an embedded language within C++/Python. There are certain things we lift into Halide from the host language (e.g. arithmetic operators), and certain things we can't (e.g. if statements). Then for some things there's a choice. For double literals in the host language we have three choices:

1) lifting them implicitly. So (x + 2.0) is a Halide Expr with Halide type Float(64).
2) lifting them only explicitly. So x + 2.0 is an error, but x + f64(2.0) or x + Expr(2.0) are ok and have type Float(64).
3) implicitly narrowing it double literals to floats. So (x + 2.0) is the same as (x + 2.0f), and is a Halide Expr of type Float(32).

Using a double in an imaging pipeline is almost always more precision than you need, and it results in slow code with no obvious reason unless you go read the assembly. It's also an easy mistake to make. The C++ embedding goes with option 2 to avoid this sort of accident. If you want a double you should say so.

It seems currently the python embedding goes with option 3. Plus there's a bug: f64(some_double) first truncates the double to a float and then converts it back to a Halide Expr, losing precision. I'm all in favor of fixing the bug, but am strongly opposed to changing the python bindings to act like option 1 above. I don't think python has 32-bit float literals, so that leaves us with option 3.

mugwort-rc:
OK. I understand.

But incorrectly lose precision is **very bad**. It is unexpected behavior for users.
It should indicate warning (by `PyErr_WarnEx`) like this.

```python
>>> x = halide.Var("x")
>>> e = x + 0.123456789012345678
HalideWarning: lose precision detected. Use `f64()` blablabla...
```

```cpp
.def(py::init([](double v) {
    float f = v;
    something_nice_detect_lose_precision(v, f);
    return Expr(f);
}))
```
(But I have no idea of `something_nice_detect_lose_precision()`... haha...)

And, I think https://github.com/halide/Halide/pull/4247#issuecomment-531555509 `f64`etc hook is not bad.

Can I fix with this plan?

abadams:
Yes, I think that's a great idea. Cast the literal to a float and then back to double. If the bits are not the same, we need to throw an error and tell the user to use f32() or f64() explicitly. So something like:

```
if (reinterpret_bits<uint64_t>(v) != reinterpret_bits<uint64_t>((double)f)) throw error;
```

We should make sure things like f32(0.3) don't error though. The intent there is clear.

steven-johnson:
+1 to @abadams plan of attack.

mugwort-rc:
Sorry for late.

I tried throw an exception, but it is not working like this.

```
  File "correctness/float_precision_test.py", line 57, in <module>
    test()
  File "correctness/float_precision_test.py", line 48, in test
    x + 0.123456789012345678
TypeError: unsupported operand type(s) for +: 'halide.Var' and 'float'
```

Because pybind11 ignore exception [here](https://github.com/pybind/pybind11/blob/f6c4c1047a293ea77223c21f84f888e062212d78/include/pybind11/pybind11.h#L1742).

Sequence:
1. `Ver + float`given
2. `Ver::operator +` is require `Expr`
3. pybind11 try convert float -> Expr by `py::implicitly_convertible<double, Expr>();`
4. throw pybind11::value_error
5. ignore exception by `PyError_Clear()`
6. float to Expr conversion failed
7. TypeError!

So I changed to throw a warning instead of an exception.

abadams:
Looks like linux-worker-3 has a full drive (also blocking the atomics merge)

steven-johnson:
Ugh, I'll go clean it up (again) but clearly we need bigger drives (not to mention faster machines)...

steven-johnson:
done, please restart failing tests

steven-johnson:
Finally clean! PTAL

PR #4252, opened 2019-09-18
Explicitly enable RTTI in retrain_cost_model (cmdline.h requires it)

steven-johnson:
attn @shubhamp-ca


steven-johnson:
FWIW: ensuring that we can compile cleanly with exceptions disabled is a hard requirement, for obvious reasons, but AFAIK there isn't nearly the same pushback on RTTI; if there are downstream users for whom enabling RTTI is a dealbreaker, it would be good to know about it.

steven-johnson:
(We may also just decide to replace cmdline.h with something that doesn't require RTTI -- cmdline.h was chosen somewhat arbitrarily.)

PR #4253, opened 2019-09-18
Re-Enable NewPassManager, plus ASAN fixes

steven-johnson:
Setting recover = false is apparently necessary in some cases for this variant of ASAN in order to meet existing results.


steven-johnson:
(Please don't land this yet, even if tests pass; @alinas is still doing some investigation into some underlying issues)

steven-johnson:
What's the latest on this -- are we still waiting on upstream fixes, or should we land this?

alinas:
Please hold off a little longer on this.
Enabling this with asan and recovery on triggered what appears to be a register allocator bug in LLVM. I'm still tracking this down.

steven-johnson:
Cool, no rush.

alinas:
This is no longer blocked. Let me know where or if we should document the reason for delaying this switch.
Thanks!

steven-johnson:
There is still an internal-to-Google issue that we need to resolve before we can enable it there, as discussed via internal email, but yes, we should be able to enable it here.

PR #4255, opened 2019-09-19
Various followup cleanups to 'Outputs' change

steven-johnson:
- Code in Module.cpp should just require the c_header to be generated along with PyTorch (and PythonExtension) rather than trying to fake one
- Removed `compile_to_python_extension()` as a public API entirely; since you must generate at the same time as the header file, it's more sensible to just require using the `compile_to()` method for this instead
- strip the header path into 'leaf' form uniformly
- don't put 'using' into root namespace in generated header files
- fix ad-hoc .pytorch.h extension usage in favor of putting the right thing into the table (overlooked in recent Outputs overhaul)
- nuked unnecessary member vars, virtual decls, suboptimal grammar, other detritus

attn: @mgharbi


steven-johnson:
Failure is unrelated.

PR #4259, opened 2019-09-23
Add draft code of conduct

### LINKED ISSUE #4256: Halide should adopt an explicit Code Of Conduct
I don't think should be a controversial assertion at this point of time.

@zvookin suggested (offline) that the Contributor's Covenant would be a good starting point (https://www.contributor-covenant.org/version/1/4/code-of-conduct); I happen to agree (it's nicely short and to-the-point). In point of fact I was tempted to making it a fait accompli by opening a PR to land it as CODE_OF_CONDUCT.md, but discussion outside of a PR is surely a better way to go about doing this.

abadams:
Here's a proposed code of conduct. Comments/changes welcome.

It's important to note that this is not a reaction to anything. We haven't had any problems that would require this and I hope we never do, but better to have one and not need it than need one and not have it.

Fixes #4256

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,32 @@
+The Halide community has always worked to be a welcoming and respectful community, and we want to ensure that doesn’t change as we grow and evolve. To that end, we have a few ground rules that we ask people to adhere to:
+
+* Be friendly and patient.
+
+* Be welcoming.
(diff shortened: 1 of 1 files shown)


steven-johnson:
LGTM

shoaibkamil:
LGTM as well

abadams:
ok, seems uncontroversial. I'll hit merge and if someone wants a change we can make a new PR for it.

PR #4261, opened 2019-09-23
Minor hygiene in Derivative

steven-johnson:
- Hide internal map struct in class Derivative, make it const, add a move ctor and use it where appropriate
- Move various functions declared in the .h file into the .cpp file
- Wrap giant glob of helper code in anonymous namespace to reduce surface area

Question: why does Derivative expose both get() and operator() for identical operations? Is there a good reason not to choose just one or the other?


steven-johnson:
@BachiLi @abadams Question about Derivative.h: the `get(Func)` method allows you to pass an optional `bool bounded` argument, which returns an 'unbounded' variant of the Func... but (1) I'm not sure what this is intended to be useful for (there is a comment in Derivative.cpp about it being "Useful for scheduling" but it's not clear to be why/how, and (2) I cannot find a single use of this argument anywhere in the Halide master branch (in tests/, apps/, or python_bindings/). 

Is this a holdover from early work, or is it still in use in some branch? (If the former, we should consider deleting it. If the latter, we need to add tests for it.)

abadams:
Each Func in the forwards pipeline maps to multiple Funcs in the backwards pipeline. You get a bunch of boundary conditions injected to make the derivatives correct. The bounded flag lets you grab the Func before or after this boundary condition is applied, so if you want to just compute_root everything you might want to schedule both. So it's meaningful, if currently unused. Manually scheduling the reverse pipelines is... difficult.

src/Derivative.h
```diff
@@ -19,71 +19,37 @@ namespace Halide {
...
-        }
-        internal_assert(it != adjoints.end()) << "Could not find Func " << name << "\n";
-        return it->second;
-    }
+    Func get(const Func &func, int update_id = -1, bool bounded = true) const;
+    Func get(const Buffer<> &buffer) const;
+    Func get(const Param<> &param) const;
 
-    Func operator()(const Func &func, int update_id = -1, bool bounded = true) const {
+    inline Func operator()(const Func &func, int update_id = -1, bool bounded = true) const {
```
abadams: I thought member functions defined within the class were automatically inline.

steven-johnson: Open https://github.com/halide/Halide/issues/4264 to track adding tests for the unbounded case.

re: member functions automatically being inlined, yes, this is redundant, I'll eliminate them.

(I'd still like to know why we have both `get()` and `operator()` and whether we can eliminate one or the other.)

BachiLi:
I'm not sure about get() and operator(). As discussed in https://github.com/halide/Halide/pull/3817
, get() makes more sense in Python and operator() makes more sense in C++. If we must pick one I would pick get(), but that breaks compatibility.

steven-johnson:
> If we must pick one I would pick get(), but that breaks backward compatibility.

How much code (outside of the tests) is using the current API? There seems to be relatively little in the current Halide master branch.

abadams:
There's some amount of research code that uses it. The current idiom in C++ is:
```
auto d_y_d = propagate_adjoints(y);

... d_y_d(x) ...

```

This is intended to read like conventional calculus -"dx/dy"

Switching to just get would be pretty awful, IMO. Now you have "d y d get x"

steven-johnson:
I'd be happy with either one, but IMHO we should choose one or the other; I don't see the upside to having both.

Also: may I suggest that instead of having

    Func operator()(Func, int update_id = -1, bool bounded = true);

we split this into two calls, since I'm guessing that ~all calls that specify `bounded` are gonna be hardcoded to false (and ~all of those will be in scheduling code), and thus we could more readably have:

    Func operator()(Func, int update_id = -1);  // always bounded
    Func get_unbounded(Func, int update_id = -1);  // try unbounded, fall back to bounded

steven-johnson:
See https://github.com/halide/Halide/pull/4272 for a draft of the revision I proposed above

BachiLi:
Looks great!

abadams:
lgtm

PR #4269, opened 2019-09-25
Use Buffer instead of Func in HelloPyTorch, to avoid needing to pass explicit sizes for the bounds of the output

steven-johnson:


steven-johnson:
attn @mgharbi

mgharbi:
Looks good to me; I did not know the type-less buffer syntax was valid.

steven-johnson:
> the type-less buffer syntax

It's totally fine, but you have to specify a concrete type in the Makefile (as you do for Func).

PR #4274, opened 2019-09-25
Don't emit float16 constants as bfloat16s

### LINKED ISSUE #4273: cast<float16_t>() outputs wrong number
The following code triggers the assertion:
```
#include "Halide.h"
#include <cmath>
using namespace Halide;
int main(int argc, char *argv[]) {
    Func im;
    im() = cast<float16_t>(100);
    Buffer<float16_t> b = im.realize();
    std::cout << "b(0):" << b(0) << std::endl;
    assert(fabs(float(b(0)) - 100.f) < 1e-3f);
    return 0;
}
```
The output is
```
b(0):3.390625h
Assertion failed: (fabs(float(b(0)) - 100.f) < 1e-3f), function main, ...
```
Changing the number from 100 to 100.f does not help either.
Changing `im`'s definition to
```
im() = float16_t(100);
```
outputs the same number (3.390625h)

abadams:
Fixes #4273


abadams:
Sadly, I think I should bloat this PR with a test.

PR #4275, opened 2019-09-25
Add operator<<(Func)

steven-johnson:
Based on code from @BachiLi, this is meant to be fairly simpleminded chunk of code for dumping Func contents for debugging purposes.

### DIFF
--- a/src/IRPrinter.cpp
+++ b/src/IRPrinter.cpp
@@ -5,8 +5,13 @@
 
 #include "AssociativeOpsTable.h"
 #include "Associativity.h"
+#include "DerivativeUtils.h"
+#include "FindCalls.h"
+#include "Func.h"
 #include "IROperator.h"
 #include "Module.h"
+#include "RealizationOrder.h"
+#include "Simplify.h"
 #include "Target.h"
 
 namespace Halide {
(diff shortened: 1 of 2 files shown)


src/IRPrinter.cpp
```diff
@@ -165,6 +170,71 @@ ostream &operator<<(ostream &stream, const Target &target) {
...
+            // Assume that Tuples have the same (or no) RDom across all values.
+            ReductionDomain rdom = extract_rdom(vals.at(0));
+            if (rdom.defined()) {
+                stream << " with RDom: ";
+                std::vector<Expr> e;
+                for (const auto &d : rdom.domain()) {
+                    e.push_back(d.min);
+                    e.push_back(d.extent);
+                }
+                emit_with_commas(stream, e);
```
BachiLi: might be nice to also print out the predicate of the RDom

steven-johnson: done

steven-johnson:
Monday morning review ping

src/IRPrinter.h
```diff
@@ -50,6 +50,10 @@ struct Target;
 /** Emit a halide Target in a human readable form */
 std::ostream &operator<<(std::ostream &stream, const Target &);
 
+/** Emit a halide Func in a human readable form. */
```
BachiLi: Emit a Halide Func "and its dependencies". Currently the code find all the dependent Funcs and print all of them.
Not sure if we want an option to only print one Func. The print all dependencies version was more useful for me since it prints both the boundary condition and the actual content of the gradients.

abadams: Dumping the entire DAG seems like a weird default behavior to me. I would expect a << on a Pipeline to do that, and << on a Func to just dump that one Func

BachiLi: Sounds good to me!

steven-johnson:
I'm not sure I agree that Pipeline should be full-dag but Func should not; to avoid opening a debate on that, I've made it so that `operator<<(Func)` is single-Func, and added syntactic-sugar `operator<<(FuncWithDependencies)` for full-DAG.

(Alternatively, I thought about adding a `print_func()` public method to offer the full-DAG option; I suspect that full-DAG will only be occasional code for debugging, so it may not matter all that much.)

steven-johnson:
Gentle review ping on the last revision pushed.

steven-johnson:
Monday Morning Review Ping

src/IRPrinter.cpp
```diff
@@ -165,11 +170,89 @@ ostream &operator<<(ostream &stream, const Target &target) {
...
+                emit_with_commas(stream, f.args());
+                stream << " = ";
+                vals = f.values().as_vector();
+            }
+            if (vals.size() > 1) {
+                stream << "tuple<" << vals.size() << ">";
+            }
+            emit_with_commas(stream, vals);
+            // Assume that Tuples have the same (or no) RDom across all values.
+            ReductionDomain rdom = extract_rdom(vals.at(0));
```
abadams: This assumption isn't true in general. An RVar may only be used in one val or even just in one arg (e.g. zeroing out a portion of a tensor). Also we really need the RDom name for this to be completely specified. Otherwise you can't distinguish RVars from Params.

src/IRPrinter.cpp
```diff
@@ -165,11 +170,89 @@ ostream &operator<<(ostream &stream, const Target &target) {
...
+    std::vector<std::string> order = include_dependencies
+        ? realization_order({ func.function() }, env).first
+        : std::vector<std::string>{func.function().name()};
+
+    for (int i = (int) order.size() - 1; i >= 0; i--) {
+        Func f(env[order[i]]);
+        for (int update_id = -1; update_id < f.num_update_definitions(); update_id++) {
+            std::vector<Expr> vals;
+            if (update_id >= 0) {
+                stream << "  " << f.name() << ".update[" << update_id << "]";
```
abadams: ```.update[...]``` seems superfluous. Wouldn't it be better to print it in the way that most closely matches the Halide source code?

src/IRPrinter.cpp
```diff
@@ -198,21 +281,77 @@ void IRPrinter::test() {
...
+  out[x] = (buf((x % 3)) + 1)
+ }
+}
+func some_func = {
+ some_func(xx) = (0)
+  some_func.update[0](rr$x) = ((some_func(rr$x) + rr$x)) with RDom(1, 99)
+  some_func.update[1](rr$x) = ((some_func(rr$x)*rr$x)) with RDom(1, 99)
+}
+func tuple_func = {
+ tuple_func(xx) = tuple<2>(0, 0)
```
abadams: Would prefer braces: `{0, 0}`, to more closely match Halide source.

BachiLi:
Any chance this can be merged recently? I found this functionality to be very useful personally (and has been copy pasting code from this branch several times).

PR #4276, opened 2019-09-25
Add backend support for negative bit shifts on Hexagon.

kpassarella:
The existing intrinsics already supported this, so now the signed shifts are being passed through them, too.


steven-johnson:
Looks like apps/linear_algebra is failing:

```
make[1]: Leaving directory '/home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide/apps/linear_algebra'
/home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide-build/bin/apps/nl_means/bin/host/nl_means.generator -g nl_means -e static_library,h,registration,stmt,assembly -o /home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide-build/bin/apps/nl_means/bin/host-hvx_128 -f nl_means_auto_schedule target=host-hvx_128-no_runtime auto_schedule=true
Internal Error at /home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide/src/CodeGen_Hexagon.cpp:1527 triggered by user code at : Condition failed: fn: Function 'halide.hexagon.shr.vh.h' not found
Makefile:12: recipe for target '/home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide-build/bin/apps/camera_pipe/bin/host-hvx_128/camera_pipe.a' failed
make[1]: *** [/home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide-build/bin/apps/camera_pipe/bin/host-hvx_128/camera_pipe.a] Aborted (core dumped)
make[1]: *** Waiting for unfinished jobs....
```

src/CodeGen_Hexagon.cpp
```diff
@@ -1713,11 +1713,6 @@ void CodeGen_Hexagon::visit(const Call *op) {
             internal_assert(op->args.size() == 2);
             string instr = op->is_intrinsic(Call::shift_left) ? "halide.hexagon.shl" : "halide.hexagon.shr";
             Expr b = maybe_scalar(op->args[1]);
```
pranavb-ca: If you remove this code here, then you'll have to handle signed variants of the shift amount in ```intrinsic_wrappers``` (https://github.com/halide/Halide/blob/master/src/CodeGen_Hexagon.cpp#L372) because we won't benefit anymore from the canonicalization done by ```lower_signed_shift```_xxx

dsharletg: I think what needs to happen here is that we just change the wrappers to have signed RHSes, and cast unsigned RHSes to signed RHSes (might already happen automatically).

BTW, I think you meant to link to https://github.com/halide/Halide/blob/master/src/CodeGen_Hexagon.cpp#L604-L612.

We do need to be careful re: all of the patterns that involve shifts. It might be these basic shifts are the only ones with vector RHSes, and so this isn't actually tricky to handle.

I *think* it might only be v62 or later where the RHS can be signed. This isn't hard to support either, we basically just need to keep the old behavior if `get_target().has_feature(Target::v60)`. @pranavb-ca can you confirm? I read the HVX instruction docs but it was quite tricky to find exactly where this is supported. I'm pretty sure vasl, vasr, vlsr do after v62.

There's also this: https://github.com/halide/Halide/blob/master/src/runtime/hvx_64.ll#L288-L324 and https://github.com/halide/Halide/blob/master/src/runtime/hvx_128.ll#L289-L325. This is going to be a little annoying to handle. Given the nature of HVX intrinsics/type "safety" in LLVM, I think we can just make this all have a signed RHS, but only ever actually pass signed RHSes on v62 or later.

To summarize, I think what needs to happen is:
1. Change https://github.com/halide/Halide/blob/master/src/CodeGen_Hexagon.cpp#L604-L612 to have signed RHSes
2. Change https://github.com/halide/Halide/blob/master/src/runtime/hvx_64.ll#L288-L324 and https://github.com/halide/Halide/blob/master/src/runtime/hvx_128.ll#L289-L325 to have signed RHSes
3. Conditionalize the change you made here to only happen if `!get_target().has_feature(Target::v60)`
4. If necessary, we need to cast unsigned RHSes to signed RHSes here, but I think `call_intrin` already does this.

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -225,10 +225,18 @@ class SimdOpCheckHVX : public SimdOpCheckTest {
         check("vpacko(v*.h,v*.h)", hvx_width/1, u8(i16_1 >> 8));
         check("vpacko(v*.h,v*.h)", hvx_width/1, i8(u16_1 >> 8));
         check("vpacko(v*.h,v*.h)", hvx_width/1, i8(i16_1 >> 8));
-        check("vpacko(v*.w,v*.w)", hvx_width/2, u16(u32_1 >> 16));
-        check("vpacko(v*.w,v*.w)", hvx_width/2, u16(i32_1 >> 16));
-        check("vpacko(v*.w,v*.w)", hvx_width/2, i16(u32_1 >> 16));
-        check("vpacko(v*.w,v*.w)", hvx_width/2, i16(i32_1 >> 16));
+        check("vpacko(v*.w,v*.w)", hvx_width/2, u16(u32_1 >> cast<unsigned>(16)));
```
dsharletg: I think the ones that need testing are the vector RHS tests above (vasl, vasr, vlsr). Your change only affects vector code (as it should), so it might be best to revert the changes to these shifts.

steven-johnson:
Looks like we still have a failure:

https://buildbot.halide-lang.org/master/#/builders/18/builds/540

Internal Error at /home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-trunk/halide/src/CodeGen_Hexagon.cpp:1547 Condition failed: fn: Function 'halide.hexagon.shr.vub.vub' not found
../halide/Makefile:1852: recipe for target 'quiet_correctness_simd_op_check_hvx' failed
make: *** [quiet_correctness_simd_op_check_hvx] Error 1

src/CodeGen_Hexagon.cpp
pranavb-ca: Hi @dsharletg  - i like your suggestion. you are right, v60 doesn't have bidirectional shifts but every ISA after that does. So, yes, we should add a check for v60 and simply do 1, 2 above.

kpassarella: This has been fixed now. Please take another look, thanks!

test/correctness/simd_op_check_hvx.cpp
kpassarella: These have been reverted, thanks.

steven-johnson:
LGTM, but @dsharletg and @pranavb-ca are the must-review reviewers

src/runtime/hvx_128.ll
```diff
@@ -237,12 +237,31 @@ define weak_odr <128 x i8> @halide.hexagon.shl.vub.ub(<128 x i8> %a, i8 %b) noun
   ret <128 x i8> %r
 }
 
+define weak_odr <128 x i8> @halide.hexagon.shl.vub.b(<128 x i8> %a, i8 %b) nounwind uwtable readnone alwaysinline {
+  %a_32 = bitcast <128 x i8> %a to <32 x i32>
+  %bw = zext i8 %b to i32
```
dsharletg: `sext` (sign extend) instead of `zext` (zero extend)

src/runtime/hvx_128.ll
```diff
@@ -256,6 +275,19 @@ define weak_odr <128 x i8> @halide.hexagon.shr.vub.ub(<128 x i8> %a, i8 %b) noun
   ret <128 x i8> %r
 }
 
+define weak_odr <128 x i8> @halide.hexagon.shr.vub.b(<128 x i8> %a, i8 %b) nounwind uwtable readnone alwaysinline {
+  %a_32 = bitcast <128 x i8> %a to <32 x i32>
+  %bw = zext i8 %b to i32
```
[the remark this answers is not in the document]
kpassarella: Done, thanks

src/runtime/hvx_128.ll
```diff
@@ -269,6 +301,19 @@ define weak_odr <128 x i8> @halide.hexagon.shr.vb.ub(<128 x i8> %a, i8 %b) nounw
   ret <128 x i8> %r
 }
 
+define weak_odr <128 x i8> @halide.hexagon.shr.vb.b(<128 x i8> %a, i8 %b) nounwind uwtable readnone alwaysinline {
+  %a_32 = bitcast <128 x i8> %a to <32 x i32>
+  %bw = zext i8 %b to i32
```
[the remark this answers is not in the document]
kpassarella: Done

src/runtime/hvx_128.ll
kpassarella: Done

kpassarella:
I tried to add the tests to simd_op_check_hvx.cpp (Dillon's comment above), but I can't get it to pass.

It fails with the following error (and should in the continuous tests here, too):
Internal Error at [...]/halide/src/CodeGen_Hexagon.cpp:2647
Condition failed: op->condition.type().is_scalar():
select(((int8x128)x0 >= x128((int8)0)), (int8x128)x0, (x128((int8)0) - (int8x128)x0))
bash: line 1: 190855 Aborted                 [...]/halide/bin/correctness_simd_op_check_hvx
make: *** [Makefile:1851: correctness_simd_op_check_hvx] Error 134

I've been a little stumped on if the issue is in the test type (vlsr/vasr?) or somewhere else. Any pointers would be very helpful. Thanks!

pranavb-ca:
@kpassarella  - The problem you are running into is that ```lower_signed_shift_{right/left}```  is resulting in the creation of a select with a boolean vector. This for the HVX backend is too late for a boolean vector - the select should have been converted to a select_mask. See EliminateBoolVectors and

kpassarella:
@pranavb-ca Thanks for the pointer -- I hadn't known about this. Was there more to your sentence after "See EliminateBoolVectors and"?

pranavb-ca:
> @pranavb-ca Thanks for the pointer -- I hadn't known about this. Was there more to your sentence after "See EliminateBoolVectors and"?

Sorry for the confusion, @kpassarella . No, the "and" is misplaced and shouldn't have been there.

steven-johnson:
I just synced this to master to get an updated build. Where does this stand? Is there still a BoolVector issue to wrap up?

kpassarella:
> I just synced this to master to get an updated build. Where does this stand? Is there still a BoolVector issue to wrap up?

Yes, the bool vector issue still remains -- I haven't gotten it to work yet. Are there any examples I could look at that use EliminateBoolVectors? I don't think I've been trying to use it correctly.

dsharletg:
It's currently only used on the whole Stmt by Hexagon and OpenCL:
- https://github.com/halide/Halide/blob/ac91d3557d747477f29fc9b7b64290a939a64cf6/src/CodeGen_Hexagon.cpp#L321
- https://github.com/halide/Halide/blob/092f5282ffbea660e833688679c65cfc0d800a99/src/CodeGen_OpenCL_Dev.cpp#L670

But there's also a version of it that works on Exprs, so I think you can just call it on the result of lower_signed_shift_*.

Another option would be to write a mutator that lowers signed shifts and run it before we call eliminate_bool_vectors on the Stmt.

Another factor to consider here: we might be about to drop support for HVX v60 (it's 4 years old now), which was the last architecture to not support signed shifts. So this problem might just go away shortly...

kpassarella:
@dsharletg When I wrap the output of lower_signed_shift_* with eliminate_bool_vectors(), it still gets the same error, which is why I wasn't sure if that was the correct thing to do.
Is there something else I need, like using select_mask instead of select? It says select_mask isn't declared in this scope, and I'm not sure where to get it from.

dsharletg:
I just sent #4465, which might simply remove the need for this codepath on Hexagon :)

dsharletg:
#4465 is merged, so we simply don't need a fallback path for signed shifts on Hexagon anymore.

kpassarella:
@dsharletg Thanks for the heads up! Just to confirm, none of the changes in this PR would be needed, so it can be closed? Or do you mean that this fallback can be removed:
if (b.type().is_int()) {
    CodeGen_Posix::visit(op);
    return;
}

dsharletg:
We can now assume that isa_version >= 62, so the ifs that check if that is not true (and the associated branch) can be removed.

dsharletg:
This is almost all @kpassarella's work, I only just tweaked it after we dropped support for v60 :)

kpassarella:
Thanks for the tweaks/fixes, Dillon! Do you happen to know why it's still failing the one testbranch-trunk check?

steven-johnson:
> Thanks for the tweaks/fixes, Dillon! Do you happen to know why it's still failing the one testbranch-trunk check?

See https://github.com/halide/Halide/pull/4572

steven-johnson:
Remaining breakage is just https://github.com/halide/Halide/issues/4559 (unrelated). Ready to land?

kpassarella:
Got it, since the breakage is unrelated, I'll merge. Thanks!

pranavb-ca:
> This is almost all @kpassarella's work, I only just tweaked it after we dropped support for v60 :)

Indeed. Sorry about that and thank you @kpassarella! :)

PR #4277, opened 2019-09-26
Clean IRPrinter indentation code

steven-johnson:
It's always bugged me that IRPrinter used a separate call (`do_indent()`) to indent outputs, rather than some approach that mixed into the stream flow of operator<<.

This is a completely nonessential change that nevertheless (I think) improves the code; it creates a new type, `Indentation`, which has an `operator<<` overload. This means that stream writes that begin with an indent are more obvious (since it's part of the ordinary stream syntax).

### DIFF
--- a/src/IRPrinter.cpp
+++ b/src/IRPrinter.cpp
@@ -328,6 +328,11 @@ std::ostream &operator<<(std::ostream &stream, const LinkageType &type) {
     return stream;
 }
 
+std::ostream &operator<<(std::ostream &stream, const Indentation &indentation) {
+    for (int i = 0; i < indentation.indent; i++) stream << ' ';
+    return stream;
+}
+
 IRPrinter::IRPrinter(ostream &s) : stream(s), indent(0) {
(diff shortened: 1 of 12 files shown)


src/IRPrinter.cpp
```diff
@@ -328,6 +328,11 @@ std::ostream &operator<<(std::ostream &stream, const LinkageType &type) {
     return stream;
 }
 
+std::ostream &operator<<(std::ostream &stream, const Indentation &indentation) {
+    for (int i = 0; i < indentation.indent; i++) stream << ' ';
```
abadams: Nit: for for loops, I think we use braces and multiple lines.

steven-johnson: Will fix, but look at the deleted version of do_indent(), from which this was directly copied :-/

abadams: Yeah, I noticed that. I spent some time yesterday trying to wrangle a clang-format file that I could blindly apply, but it's still the case that it fixes trivial things like this that aren't a big deal for readability, and then makes a horrible mess in other cases.

PR #4278, opened 2019-09-26
Fix HelloPyTorch/setup.py

steven-johnson:
include_dirs should be an arg to the Extensions, not to setup(). (Apparently we were often getting lucky with the include path, but on my local OSX box it fails to build without this change.)


steven-johnson:
attn @mgharbi

Issue #4100, opened 2019-08-06
Halide should provide a is_inf method


benoitsteiner:
Halide provides a very convenient is_nan function to check for NaN. It would be very convenient to also have the ability to check for infinite floating point values.

steven-johnson:
+1 -- mimicking the name and semantics of C++ `std::isfinite()` seems like the right thing.

Issue #2497, opened 2017-10-30
Add Halide::is_finite()


steven-johnson:
We provide is_nan() but not is_finite(); notwithstanding https://github.com/halide/Halide/issues/1440 and the -ffast-math promises, we should probably provide is_finite().

steven-johnson:
https://github.com/halide/Halide/pull/4280

PR #4280, opened 2019-09-27
Add is_inf() and is_finite() (Issues #4100 and #2497)

steven-johnson:
attn @benoitsteiner


steven-johnson:
(These are not particularly urgent at all but I needed a nice end-of-day task...)

src/CodeGen_LLVM.cpp
```diff
@@ -3156,6 +3156,37 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+        llvm::FastMathFlags safe_flags;
+        safe_flags.clear();
+        builder->setFastMathFlags(safe_flags);
+        builder->setDefaultFPMathTag(strict_fp_math_md);
+
+        // isinf(e) -> (fabs(e) == infinity)
+        Expr e = op->args[0];
+        internal_assert(e.type().is_float());
+        Expr inf = e.type().max();
+        codegen(abs(e) == inf);
```
abadams: would codegen(strict_float(abs(e) == inf)) be equivalent but shorter?

src/CodeGen_LLVM.cpp
```diff
@@ -3156,6 +3156,37 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+        llvm::FastMathFlags safe_flags;
+        safe_flags.clear();
+        builder->setFastMathFlags(safe_flags);
+        builder->setDefaultFPMathTag(strict_fp_math_md);
+
+        // isfinite(e) -> (fabs(e) != infinity && !isnan(e)) -> (fabs(e) != infinity && e == e)
+        Expr e = op->args[0];
+        internal_assert(e.type().is_float());
+        Expr inf = e.type().max();
+        codegen(abs(e) != inf && e == e);
```
abadams: Same comment here.

src/CodeGen_LLVM.cpp
steven-johnson: That's a good question -- I was following the pattern used by is_nan here. Maybe. Though I think a "righter" answer is probably to follow this with a fix for https://github.com/halide/Halide/issues/4281 (which I filed this morning), as the use of these calls outside a strict_float context in the first place is likely an error on the coder's part anyway.

benoitsteiner:
Thanks for adding this.

PR #4282, opened 2019-09-27
Clang format

abadams:
Stylistic inconsistencies have crept into our codebase. I've been opposed to it in the past, but I think we're probably at the point where it's worth just clang-formatting all the things. This PR clang-formats most of the things. I accepted every suggested change that was not outright horrifying, which was fortunately nearly all of them after some tuning of the clang-format file.

Note that many of these changes are slightly pessimizations, but in net I think it's positive, and it's not really feasible to do the full triage.


steven-johnson:
LGTM in principle; how do we ensure that it stays this way? e.g. add a presubmit formatting check?

steven-johnson:
While we're on the topic of style-related things... can we / should we settle on a preferred approach for std:: things? Some files like to put (e.g.) `using std::string` etc at top, then use unadorned `string`; others use `std::string` everywhere. I could live with either, but stating a preferred approach would be nice.

src/CodeGen_OpenGLCompute_Dev.cpp
```diff
@@ -157,10 +159,9 @@ void CodeGen_OpenGLCompute_Dev::CodeGen_OpenGLCompute_C::visit(const For *loop)
             const IntImm *int_limit = loop->extent.as<IntImm>();
             user_assert(int_limit != nullptr) << "For OpenGLCompute workgroup size must be a constant integer.\n";
             int new_workgroup_size = int_limit->value;
-            user_assert(workgroup_size[index] == 0 || workgroup_size[index] == new_workgroup_size) <<
-                "OpenGLCompute requires all gpu kernels have same workgroup size, "
-                "but two different ones were encountered " << workgroup_size[index] << " and " << new_workgroup_size <<
-                " in dimension " << index << ".\n";
+            user_assert(workgroup_size[index] == 0 || workgroup_size[index] == new_workgroup_size) << "OpenGLCompute requires all gpu kernels have same workgroup size, "
```
steven-johnson: Another weird one.

abadams:
Mechanically ensuring it stays this way I think has to wait until it's workable to just blindly apply the clang-format. There's still a bunch of code where it makes a mess.

I don't have a strong opinion on explicit std:: vs using

src/Error.h
```diff
@@ -104,11 +106,12 @@ struct ErrorReport {
...
- public:
-  HALIDE_ALWAYS_INLINE Voidifier() = default;
-  // This has to be an operator with a precedence lower than << but
-  // higher than ?:
-  HALIDE_ALWAYS_INLINE void operator&(ErrorReport&) {}
+public:
+    HALIDE_ALWAYS_INLINE Voidifier() = default;
+    // This has to be an operator with a precedence lower than << but
+    // higher than ?:
+    HALIDE_ALWAYS_INLINE void operator&(ErrorReport &) {
```
steven-johnson: Yuckity yuck.

zvookin:
I glanced at it and will approve, but I'm not planning to read every line here.

Re: using and std::, they definitely are not allowed at top-level in header files. Personally I find using individual types more hassle than it is worth. Perhaps we could consider importing a standard set of names into e.g. Halide::Internal.

src/IR.cpp
```diff
@@ -855,96 +851,360 @@ bool Shuffle::is_extract_element() const {
...
-template<> Stmt StmtNode<Allocate>::mutate_stmt(IRMutator *v) const { return v->visit((const Allocate *)this); }
-template<> Stmt StmtNode<Free>::mutate_stmt(IRMutator *v) const { return v->visit((const Free *)this); }
-template<> Stmt StmtNode<Realize>::mutate_stmt(IRMutator *v) const { return v->visit((const Realize *)this); }
-template<> Stmt StmtNode<Block>::mutate_stmt(IRMutator *v) const { return v->visit((const Block *)this); }
-template<> Stmt StmtNode<IfThenElse>::mutate_stmt(IRMutator *v) const { return v->visit((const IfThenElse *)this); }
-template<> Stmt StmtNode<Evaluate>::mutate_stmt(IRMutator *v) const { return v->visit((const Evaluate *)this); }
-template<> Stmt StmtNode<Prefetch>::mutate_stmt(IRMutator *v) const { return v->visit((const Prefetch *)this); }
-template<> Stmt StmtNode<Acquire>::mutate_stmt(IRMutator *v) const { return v->visit((const Acquire *)this); }
-template<> Stmt StmtNode<Fork>::mutate_stmt(IRMutator *v) const { return v->visit((const Fork *)this); }
+template<>
```
steven-johnson: I'd be tempted to use a temporary #define here to make it clear this is all boilerplate...

    #define MAKE_ACCEPT(OP) \
        template<> void ExprNode<OP>::accept(IRVisitor *v) const { v->visit((const OP *)this); }

src/IRMutator.cpp
```diff
@@ -47,21 +56,51 @@ Expr mutate_binary_operator(IRMutator *mutator, const T *op) {
...
-Expr IRMutator::visit(const Max *op)     {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const EQ *op)      {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const NE *op)      {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const LT *op)      {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const LE *op)      {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const GT *op)      {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const GE *op)      {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const And *op)     {return mutate_binary_operator(this, op);}
-Expr IRMutator::visit(const Or *op)      {return mutate_binary_operator(this, op);}
+Expr IRMutator::visit(const Add *op) {
```
steven-johnson: see previous thought about using a temporary #define

src/CodeGen_C.cpp
```diff
@@ -2892,14 +2892,22 @@ int test1(struct halide_buffer_t *_buf_buffer, float _alpha, int32_t _beta, void
...
+            diff--;
+        }
+        while (diff_end < (int)src.size() && src[diff_end] != '\n') {
+            diff_end++;
+        }
 
         internal_error
-            << "Correct source code:\n" << correct_source
-            << "Actual source code:\n" << src
+            << "Correct source code:\n"
```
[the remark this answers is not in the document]
abadams: I think it's the presence of the newline characters.

src/CodeGen_OpenGLCompute_Dev.cpp
abadams: Fixed. clang-format doesn't seem to remove line breaks, so I just added some extra breaks to tame this one.

Issue #4285, opened 2019-09-29
f[g[1]]=2 in Python bindings triggers wrong operator


BachiLi:
Consider the following code:
```
    x = hl.Var('x')
    f = hl.Func('f')
    g = hl.Func('g')
    g[x] = 1
    f[x] = 0
    f[g[1]] = 2
```
This triggers the error
```
RuntimeError: Error: Can't index into a reference to Func "g", because it does not return a Tuple.
```

I don't know why but `f[g[1]] = 2` triggers the `__getitem__` operator in FuncRef, which then calls the Tuple element selection function. I printed the index in the `__getitem__` function and it was 0.

Changing the last line to `f[g[1]+0] = 2` or `f[hl.Expr(g[1])] = 2` makes the runtime error disappear and calls the correct `__setitem__` operator in Func. Removing the `__getitem__` function in FuncRef makes the error disappear too. Adding a `__setitem__` operator in Func that takes FuncRef and Expr as arguments doesn't help. Changing the last line to `f.__setitem__(g[1], 2)` doesn't help either (still directed to `FuncRef.__getitem__`).

I couldn't figure out why after an hour of investigation, so I'm posting the issue here. This might be a pybind11 bug.

BachiLi:
After some tracing into Pybind11 code, I think what happened is, pybind11 is trying to convert the FuncRef (`g[1]`) into a list of Var. To do this it has to check whether `g[1][0]` is a Var or not. But `g[1][0]` has a totally different meaning that it’s selecting the Tuple element so it fails.

PR #4286, opened 2019-09-29
Allowing f[g[1]] = 2 in Python bindings

BachiLi:
An attempt to fix https://github.com/halide/Halide/issues/4285. As mentioned the issue is that when pybind11 is visiting all possible overloaded function for `Func.__setitem__(key, value)`, it tries to convert the key `g[1]` into a list of `Var` or `Expr`. To do this it checks whether `g[1][0]` is a Var/Expr or not, but this triggers a runtime error inside Halide, since `g` is not a Tuple.

My proposed solution to this is to define a set of overloaded functions so that the overloading test ends before it enters the list testing conversion phase. Another possible fix is to rewrite `__getitem__` in FuncRef such that it doesn't trigger runtime error even when `g` is not a Tuple.

### DIFF
--- a/python_bindings/correctness/basics.py
+++ b/python_bindings/correctness/basics.py
@@ -160,7 +160,7 @@ def test_basics4():
     g[x] = 1
     f[x] = 0
     r = hl.RDom([(0, 100)])
-    f[g[1]] = 2
+    f[g[r]] = 2
     f.compute_root()
     f.compile_jit()
 
--- a/python_bindings/correctness/basics.py
+++ b/python_bindings/correctness/basics.py
@@ -160,9 +160,9 @@ def test_basics4():
     f = hl.Func('f')
     g = hl.Func('g')
     g[x] = 1
-    f[x] = 0
+    f[x] = 0.0
     r = hl.RDom([(0, 100)])
-    f[g[r]] = 2
+    f[g[r]] = 2.3 # This triggers a warning of double-to-float conversion
     f.compute_root()
     f.compile_jit()
 
--- a/python_bindings/correctness/basics.py
+++ b/python_bindings/correctness/basics.py
@@ -154,6 +154,8 @@ def test_basics3():
     left += ss
 
 def test_basics4():
+    # Test for f[g[r]] = ...
+    # See https://github.com/halide/Halide/issues/4285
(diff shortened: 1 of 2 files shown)


python_bindings/src/PyFunc.cpp
```diff
@@ -36,6 +36,25 @@ void define_set(py::class_<Func> &func_class) {
...
+        });
+}
+
+// Special case: there is no implicit conversion of double in C++ Halide
+// but there is implicit conversion of double in Python.
+template<>
+void define_set_func_ref<double>(py::class_<Func> &func_class) {
+    func_class
+        .def("__setitem__", [](Func &func, const FuncRef &lhs, const double &rhs) -> Stage {
+            return func(lhs) = Expr(rhs);
```
abadams: Should perhaps throw an error if rhs is not exactly representable as a float (we shouldn't implicitly lose precision). Not sure what the best way to check that is in a fast-math world.

abadams: See #4247 for context

python_bindings/correctness/basics.py
```diff
@@ -153,6 +153,17 @@ def test_basics3():
     left += 5
     left += ss
 
+def test_basics4():
+    x = hl.Var('x')
```
steven-johnson: Should probably add a comment about what this is testing for (and perhaps a link to the GitHub issue) for future reference

BachiLi: Fixed.

python_bindings/src/PyFunc.cpp
BachiLi: Added some warnings. It seems that the precise content of warnings depend on #4247. @steven-johnson do you need help on that PR?

steven-johnson:
LGTM, but will leave for @abadams to approve.

python_bindings/src/PyFunc.cpp
steven-johnson: I wasn't working on the PR (I had assumed that @mugwort-rc would finish it with the relevant comments, since they authored the original), but nothing much has happened. I'd be happy for you to resurrect it and apply a similar fix there :-)

Issue #4284, opened 2019-09-28
[RFC] CMake buildsystem QoL improvements ?


LebedevRI:
Hello.
To preface, i will understand if this isn't something that is wanted for the project..

I'd like to try using halide in some my future toy project.
As it stands now, halide is largely not packaged in distributions,
certainly not in debian; and this is a dealbreaker for me :/

After looking around, i understand why that might be the case - the cmake build system
is, uhm, making some very unusual and interesting choices.

The question is: is there any interest in trying to straighten out the cmake build,
make it more likely to be packaged, and easier to use in cmake-based projects?

Right now i see 3 main issues:
1. There should be a switch to **OPTIONALLY** link to shared LLVM libraries (trivial, i already have a patch)
2. The installation destinations need to be fixed, the current install directory structure is _not ideal_. (Likely simple. Should the defaults (that are likely in-parity with other build systems?) be fixed, or should the defaults remain?)
3. A proper Halide-Config.cmake magic needs to be added, so it can be found with simple `find_package(halide)`.

abadams:
These sorts of improvements are generally welcome, provided they don't break existing usage. @steven-johnson is the person who knows the most about the existing cmake-based build. I generally avoid it and just use the makefiles.

BachiLi:
A related issue is that the Python bindings also need a CMake file (especially important for Windows users).

steven-johnson:
Improvements of the sort you describe would be quite welcome. We have historically used CMake primarily for Windows testing, but as you can see, it often lags behind; getting it to 100% parity with Makefile would be a highly desirable goal in my opinion. Contributions in this vein would be great, especially because none of the core contributors use CMake much outside of this project and thus things like the installation options are easy for us to overlook.

LebedevRI:
> provided they don't break existing usage

Yeah, that's the fun part i suppose.
Do i understand correctly that if one needs to opt-in into new behavior explicitly,
for example via a CMake option that you'd pass when building Halide itself,
then the existing usage would not be considered broken?

zvookin:
We're totally into helping to make stuff like this better. The main thing I'll add to the discussion so far is that future success rides on making sure the things you care about are covered by tests somehow so that if anything breaks it gets caught before it gets checked in.

I'd also like to see us have regular packaging for various environments as part of our release process.

LebedevRI:
> > provided they don't break existing usage
> 
> Yeah, that's the fun part i suppose.
[...]

PR #4287, opened 2019-09-30
CMake: add optional LLVM_USE_SHARED_LLVM_LIBRARY option.

LebedevRI:
As discussed in https://github.com/halide/Halide/issues/4284
one important thing is to be able to link to shared versions
of LLVM libraries. This is off-by-default, so i don't believe
this should have any impact on existing users.

The important [existing] caveat to note is that i don't see
where those LLVM libraries-to-be-linked are specified anywhere
in the `halide_config.cmake`, so as long as you are
a. building static halide
and
b. not building halide as a cmake sub-project (via `add_subdirectory()`)
i don't see how the linking can succeed, both either with
old static LLVM libraries, and now with shared LLVM library.

The best i can tell `llvm-config --system-libs` is only needed
when linking to static libraries, to propagate their deps,
so let's not do that for shared LLVM libs.

And the switch to `llvm_config()` is the main thing here,
it internally handles `USE_SHARED`, and does `target_link_libraries()`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ file(TO_NATIVE_PATH "${LLVM_TOOLS_BINARY_DIR}/llvm-nm${CMAKE_EXECUTABLE_SUFFIX}"
 file(TO_NATIVE_PATH "${LLVM_TOOLS_BINARY_DIR}/clang${CMAKE_EXECUTABLE_SUFFIX}" CLANG)
 file(TO_NATIVE_PATH "${LLVM_TOOLS_BINARY_DIR}/llvm-config${CMAKE_EXECUTABLE_SUFFIX}" LLVM_CONFIG)
 
-if(HALIDE_SHARED_LLVM_LIBRARY)
+if(LLVM_USE_SHARED_LLVM_LIBRARY)
   # Since we will be linking to the shared LLVM library,
   # we will get all the transitive dependencies for free automagically.
   set(HALIDE_SYSTEM_LIBS)
@@ -191,7 +191,7 @@ option(TARGET_OPENGL "Include OpenGL/GLSL target" ON)
 option(TARGET_OPENGLCOMPUTE "Include OpenGLCompute target" ON)
 option(TARGET_D3D12COMPUTE "Include Direct3D 12 Compute target" ON)
 option(HALIDE_SHARED_LIBRARY "Build as a shared library" ON)
-option(HALIDE_SHARED_LLVM_LIBRARY "Use shared versions of LLVM libraries" OFF)
+option(LLVM_USE_SHARED_LLVM_LIBRARY "Use shared versions of LLVM libraries" OFF)
 option(HALIDE_ENABLE_RTTI "Enable RTTI" ${LLVM_ENABLE_RTTI})
 option(HALIDE_ENABLE_EXCEPTIONS "Enable exceptions" ${LLVM_ENABLE_EH})
 option(HALIDE_USE_CODEMODEL_LARGE "Use the Large LLVM codemodel" OFF)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -185,6 +191,7 @@ option(TARGET_OPENGL "Include OpenGL/GLSL target" ON)
 option(TARGET_OPENGLCOMPUTE "Include OpenGLCompute target" ON)
 option(TARGET_D3D12COMPUTE "Include Direct3D 12 Compute target" ON)
 option(HALIDE_SHARED_LIBRARY "Build as a shared library" ON)
+option(HALIDE_SHARED_LLVM_LIBRARY "Use shared versions of LLVM libraries" OFF)
```
LebedevRI: Please bikeshed on the name, are there any better suggestions?

src/CMakeLists.txt
```diff
@@ -758,17 +758,19 @@ else()
...
-# When building a shared library the LLVM libraries will be
-# embedded in the Halide library. When building a static library
-# LLVM is not embedded but CMake knows that when building an executable
-# against the Halide static library that it needs to link LLVM too so
-# PRIVATE scope is the correct choice here.
-target_link_libraries(Halide PRIVATE ${LIBS})
+# When building a shared library the LLVM libraries will either be
+# embedded or linked into the Halide library. When building a static library
+# LLVM is not embedded/linked but CMake knows that when building an executable
+# against the Halide static library that it needs to link LLVM too, IFF the
```
steven-johnson: nit: s/the halide/Halide/

src/CMakeLists.txt
```diff
@@ -758,17 +758,19 @@ else()
...
+# halide is being built as a subproject. (which means, statically-built halide
```
steven-johnson: Nit: make a proper sentence: "Which means, statically-built Halide doesn't work out-of-installation directory."

CMakeLists.txt
steven-johnson: LLVM_USE_SHARED_LLVM_LIBRARY

src/CMakeLists.txt
```diff
@@ -758,17 +758,19 @@ else()
...
+# doesn't work out-of-installation directory)
+# FIXME: migrate to proper cmake config magic.
```
steven-johnson: I'd prefer not to land this change until seeing what config magic is actually necessary; it's not clear to me what the FIXME implies.

LebedevRI:
> > The important [existing] caveat to note is that i don't see
> > where those LLVM libraries-to-be-linked are specified anywhere
> > in the `halide_config.cmake`
> 
> Sounds like we need to upgrade the halide_config.cmake contents to specify this, then upgrade halide.cmake to reference whatever we add there as necessary.

Err, not exactly. 

The thing is, `HALIDE_SHARED_LIBRARY` is *already* broken.
As the comment notes, CMake won't link `libLLVM.a` into `libHalide.a`,
nor do you tell the users which LLVM libs should be linked when
linking to `libHalide.a` 'manually'.
`HALIDE_SHARED_LIBRARY` is only not broken when you have access
to the `Halide` CMake target, i.e. when you are building Halide as
a cmake sub-project via `add_subdirectory()` or such.

So `LLVM_USE_SHARED_LLVM_LIBRARY` doesn't change anything here.
This is a separate problem. I believe it **will** be solved further down the road
with later changes, but i'm trying to keep thinks small and simple here.

src/CMakeLists.txt
LebedevRI: Please note that these issues are not being introduced by this change, they are preexisting.
The FIXME simply records that they are there, and that they likely can be solved.

steven-johnson:
Fair enough. I'll clone this for testing.

steven-johnson:
Thanks for the fix!

LebedevRI:
Thanks!

PR #4293, opened 2019-10-02
throw warning when detect lose precision

mugwort-rc:
see https://github.com/halide/Halide/pull/4247#issuecomment-537322606

### DIFF
--- a/python_bindings/src/PyExpr.cpp
+++ b/python_bindings/src/PyExpr.cpp
@@ -22,7 +22,17 @@ void define_expr(py::module &m) {
             // PyBind11 searches in declared order,
             // int should be tried before float conversion
             .def(py::init<int>())
-            .def(py::init<double>())
+            // Python float is implemented by double
+            // But Halide prohibits implicitly construct by double.
+            .def(py::init([](double v) {
+                float f = static_cast<float>(v);
+                double check = static_cast<double>(f);
+                if ( Internal::reinterpret_bits<uint64_t>(v) != Internal::reinterpret_bits<uint64_t>(check) ) {
+                    throw py::value_error("The halide.Expr(" + std::to_string(v) + ") loses precision."
+                        "If this error occurs construct by double-range float, consider construct by the `f32`/`f64` casts instead.");
+                }
+                return Expr(f);
+            }))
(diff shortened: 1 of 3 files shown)


python_bindings/src/PyExpr.cpp
```diff
@@ -22,7 +23,25 @@ void define_expr(py::module &m) {
...
+                // otherwise, precision will be lost.  e.g. 0.1, 0.3, ...
+                using Internal::reinterpret_bits;
+                if ( reinterpret_bits<uint64_t>(v) != reinterpret_bits<uint64_t>(check) ) {
+                    std::ostringstream oss;
+                    oss.precision(17);
+                    oss << std::fixed << v;
+                    PyErr_WarnEx(
+                        PyExc_RuntimeWarning,
+                        ("The halide.Expr(" + oss.str() + ") loses precision."
+                         " If this error occurs construct by double-range float, consider construct by the `f32`/`f64` casts instead.").c_str(), 0);
```
steven-johnson: This wording is a little awkward; may I suggest:

"The floating-point value " + oss.str() + " will be interpreted as a float32 by Halide and lose precision; add an explicit `f32()` or `f64()` cast to avoid this warning."

mugwort-rc:
Thanks for reviews. Fixed it.

steven-johnson:
LGTM, but adding @abadams for opinion on whether a warning is sufficient (vs. throwing a runtime error).

mugwort-rc:
I fixed it a little because I was doing something stupid.
Sorry...

steven-johnson:
Clone of build is clean. I'm not sure yet whether I think throwing a runtime exception would be preferred to logging a warning. Anyone else have strong opinions?

abadams:
Really an error would be best, but this PR is certainly an improvement over the status-quo, and it seems like there's a technical blocker to throwing an exception. I'm OK with merging it.

steven-johnson:
Fair enough. Landing. Thanks for the fix!

steven-johnson:
whoops, this merged into a branch of mine, not master. Need to follow up by landing https://github.com/halide/Halide/pull/4247

PR #4301, opened 2019-10-08
Link libHalide for building libauto_schedule.so

shubhamp-ca:
Passing libHalide.so as an input while building the autoscheduler
inside apps/autoscheduler is creating a dependency on libHalide for
the shared plugin library where libauto_schedule.so looks for
libHalide.so in the absolute path and fails to open if the libHalide
is not found in that path. This makes the plugin not portable, as it
warrants the libHalide.so to be present in the same location while using
the plugin.

Using readelf -d on the built libauto_schedule.so --

`(NEEDED)             Shared library: [/mnt/workspace/Halide/distrib/bin/libHalide.so]`

While using the plugin, if libHalide is loaded from other path, the
plugin fails to load as it searches for libHalide in the absolute
path used while building autoscheduler.

This change links to the built libHalide.so instead of passing it as
an input file while building experimental autoscheduler plugin
libauto_schedule.so.

### DIFF
--- a/apps/autoscheduler/Makefile
+++ b/apps/autoscheduler/Makefile
@@ -6,7 +6,7 @@ include ../support/autoscheduler.inc
 # autoscheduler can't find the libHalide symbols that it needs.
 $(GENERATOR_BIN)/demo.generator: demo_generator.cpp $(GENERATOR_DEPS)
 	@mkdir -p $(@D)
-	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) -g $(filter-out %.h,$^) -o $@ $(LDFLAGS) $(HALIDE_SYSTEM_LIBS)
+	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) -g $(filter-out %.h,$^) -o $@ $(LDFLAGS) $(HALIDE_SYSTEM_LIBS) $(GENERATOR_LDFLAGS)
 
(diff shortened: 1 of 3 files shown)


steven-johnson:
This is failing on the buildbots: https://buildbot.halide-lang.org/master/#/builders/33/builds/83/steps/19/logs/stdio

shubhamp-ca:
> This is failing on the buildbots: https://buildbot.halide-lang.org/master/#/builders/33/builds/83/steps/19/logs/stdio

Aah, seems like the plugin can't find libHalide.so as it is no longer pointing to the absolute path. Didn't catch this because my LD_LIBRARY_PATH was pointing to libHalide.so and it was running okay for me. The solution seems to be to add the link./biner option 'rpath' pointing to libHalide.so for building the generator which is going to use the autoscheduler plugin (in this case demo.generator) when it is executed. Will add a patch to address this.

shubhamp-ca:
build fails because clang complains '-rpath=<some_path>' is an invalid linker option. Modifying the GENERATOR_LDFLAGS to --
`-Wl,-rpath,$(dir $(LIB_HALIDE))` 
 as both clang and GNU man pages seem to indicate this as the preferred route of passing options with arguments to the linker.

steven-johnson:
OSX builds are failing now: https://buildbot.halide-lang.org/master/#/builders/9/builds/482

steven-johnson:
The clone is clean, the travis failure is a red herring. LGTM.

PR #4303, opened 2019-10-08
Plumb LLVM_CXXFLAGS to apps/support_sa/Makefile.inc via halide_config.make

pranavb-ca:
This is done so that if Halide is built using an LLVM that needs libc++ then the same is needed to build generators.
@steven-johnson : PTAL


apps/support/Makefile.inc
```diff
@@ -47,6 +47,10 @@ OPTIMIZE ?= -O3
 CFLAGS += $(OPTIMIZE) -I $(HALIDE_DISTRIB_PATH)/include/ -I $(HALIDE_DISTRIB_PATH)/tools/ -I $(HALIDE_DISTRIB_PATH)/apps/support/
 CXXFLAGS += $(OPTIMIZE) -std=c++11 -I $(HALIDE_DISTRIB_PATH)/include/ -I $(HALIDE_DISTRIB_PATH)/tools/ $(SANITIZER_FLAGS) -Wall -Werror -Wno-unused-function -Wcast-qual -Wignored-qualifiers -Wno-comment -Wsign-compare -Wno-unknown-warning-option -Wno-psabi
 
+ifeq ($(CXX), clang++)
```
steven-johnson: What is CXX is g++?

pranavb-ca: Using libc++ with Clang is easier. I am expecting someone who built LLVM with libc++  to use clang++ as CXX. It is an assumption, but I think a reasonable one.

shubhamp-ca: This doesn't reliably detect whether CXX is clang++ (when CXX is a path to clang++ or, clang++ is symlinked to foo). Can use something like --
`CXX_VERSION = $(shell $(CXX) --version | head -n1)`
`ifneq (,$(findstring clang,$(CXX_VERSION)))`

shubhamp-ca: Have created a PR for this -- https://github.com/halide/Halide/pull/4309

PR #4312, opened 2019-10-10
Update README description to match the website

BachiLi:
The website now advertises Halide as an "image and array processing language". We should do the same for the readme.

In general we should make the readme friendlier, give some hints regarding what Halide can do, who are the target audiences, what does Halide code look like, etc. We should also mention the Python bindings. I'll draft something these days.


abadams:
No reason to let the buildbots run this one...

PR #4314, opened 2019-10-11
Minor changes in CMake build system

kasper93:
Hi,

I share few patches that fixes Halide building in my environment. Apart from that sync the RT compilation flags between CMake and Makefiles.

Best Regards,
Kacper


src/CMakeLists.txt
```diff
@@ -143,6 +143,7 @@ endif()
 # Commands to build initial module objects.
 file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/${PROJECT_NAME}.build")
 set (CXX_WARNING_FLAGS -Wall -Werror -Wno-unused-function -Wcast-qual)
+set (RUNTIME_CXX_FLAGS -O3 -fno-vectorize -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables -fpic -std=gnu++98)
```
steven-johnson: Why did you remove `-fno-ms-compatibility`?

halide.cmake
```diff
@@ -305,7 +305,6 @@ endfunction()
 
 # Set the C++ options necessary for using libHalide.
 function(_halide_set_cxx_options TARGET)
-  set_target_properties("${TARGET}" PROPERTIES CXX_STANDARD 11 CXX_STANDARD_REQUIRED YES CXX_EXTENSIONS NO)
```
steven-johnson: I presume this is being removed to allow you to use something later than C++11... I suppose it's ok to just remove this and require the downstream consumer to configure at-least-C++11.

CMakeLists.txt
```diff
@@ -105,6 +109,18 @@ else()
   set(CMAKE_CXX_STANDARD 11)
 endif()
 
+string(FIND "${LLVM_CONFIG_CXXFLAGS}" "-stdlib=libc++" LLVM_LIBCXX)
```
steven-johnson: Will this work for non-Clang environments?

src/CMakeLists.txt
kasper93: Like the commit message says, I want to keep the same flags for both Makefile and CMake. See https://github.com/halide/Halide/blob/aa33d0a/Makefile#L1039 there is not `-fno-ms-compatibility`. We can add it in both if you prefer, but we need to keep them in sync.

halide.cmake
kasper93: Exactly, cmake function shouldn't overside C++ std options like that. Some people (like me) might prefer to use different version.

CMakeLists.txt
kasper93: No. But it wouldn't link anyway. If you have LLVM built with libc++ you need to build Halide with it also.

src/CMakeLists.txt
steven-johnson: Fair enough -- I don't recall offhand why that flag is here in the first place, and all the tests are passing.

PR #4317, opened 2019-10-12
Out-of-source wheel-based build for HelloPyTorch

mgharbi:


apps/HelloPyTorch/Makefile
```diff
@@ -12,30 +12,36 @@ CUDA_OPS =
...
 all: test
 
 # Run the PyTorch tests to verify the module is compiled correctly
 # .wrapper is a dummy file whose timestamps allow Make to track the dependency
-# on the Python side of the build
-test: .wrapper
-	@PYTHONPATH=lib:${PYTHONPATH} $(PYTHON) test.py
+# on the Python side of the build 
+
+HALIDE_DISTRIB_PATH=$(shell pwd)/../../distrib
```
steven-johnson: HALIDE_DISTRIB_PATH should already be defined in Makefile.inc... if that definition isn't working, we should probably fix it there rather than redefine it here

apps/HelloPyTorch/Makefile
```diff
@@ -12,30 +12,36 @@ CUDA_OPS =
...
-	@mkdir -p lib
-	@HAS_CUDA=$(HAS_CUDA) \
-		HALIDE_DISTRIB_PATH=$(HALIDE_DISTRIB_PATH) \
-		BIN=$(BIN)/$(HL_TARGET) \
-		PYTHONPATH=lib:${PYTHONPATH} \
-		$(PYTHON) setup.py build -b $(BIN)/python install --install-lib lib
-	@touch .wrapper
+$(BIN)/.wrapper: $(OPS) $(CUDA_OPS) setup.py
+	@mkdir -p $(EXT_LIB)
+	@HAS_CUDA=$(HAS_CUDA) HALIDE_DISTRIB_PATH=$(HALIDE_DISTRIB_PATH) \
```
steven-johnson: IMHO keeping the one-definition-per-line formatting makes this slightly more readable (or maybe slightly less unreadable)

apps/HelloPyTorch/Makefile
mgharbi: So, the issues comes up in the `pip install` build: the build system creates a temp directory and since the path is relative, it looks for Halide in the wrong directory. Making HALIDE_DISTRIB_PATH absolute would fix it.

steven-johnson: In that case I suspect that the best fix is changing Makefile.inc to use `realpath`:

    HALIDE_DISTRIB_ATH ?= $(realpath ../../distrib)

mgharbi: I would agree for linux machines, but I think e.g. `realpath` is not installed by default on OSX..

apps/HelloPyTorch/Makefile
mgharbi: fixed in #96aa9a6

apps/HelloPyTorch/Makefile
steven-johnson: `realpath` as a bash command is not available on OSX, but `$(realpath)` is part of standard make ~everywhere, and works fine on OSX (we already use it in `Makefile`)

mgharbi: fixed in 93c2db901

PR #4320, opened 2019-10-14
Warn user if using JIT + MSAN

steven-johnson:
MSAN is well-documented as not support JIT compilers of any sort, and will report false positives in that case. We should (at least) emit a warning if the user tries to use the JIT if libHalide is compiled with MSAN enabled (regardless of whether 'msan' is in the Halide::Target).

(I'm tempted to make this a user_error instead of a warning. Thoughts?)


abadams:
+1 to a warning, but infer_input_bounds may be called before realize and will have already failed. Maybe put the error in call_jit_code?

PR #4325, opened 2019-10-18
make_const -> make_zero or make_one in Derivative.cpp

BachiLi:
A cosmetic change in Derivative.cpp that replaces make_const with make_zero and make_one when possible.


BachiLi:
Tested in #4323. The failure in #4323 will be addresed in #4326.

PR #4326, opened 2019-10-18
Fix deprecated LLVM function calls (setAlignment)

BachiLi:
`setAlignment` now takes `MaybeAlign` objects instead of an `uint64_t`, in llvm 10. This is causing compilation failures in buildbots.


abadams:
Thanks. Something very bad is happening on the linux builder though with a double-free. I don't see how it could be related to this PR.

BachiLi:
It's failing on the Hexagon tests. I have no idea what's happening.

PR #4330, opened 2019-10-22
Drop support for Python 2.x

steven-johnson:
Most of the world is dropping support for Python 2.x no later than January 1, 2020; in theory our Python bindings still support 2.x, but in practice we haven't been reliably testing that configuration. Let's go ahead and drop any pretense of support now; this removes the last dregs of code that assumes that Python 2.x is a possibility.


python_bindings/readme.md
```diff
@@ -27,7 +27,7 @@ in Python, only the second variant is provided.
 
 ## Prerequisites ##
 
-The bindings (and demonstration applications) should work well both for python2.7 and python3.4 (or higher), on Linux and OSX platforms. Windows is not yet supported, but could be with CMake work. (The Makefile defaults to using Python 3.x; to use Python 2, set `PYTHON = python` before building.)
+The bindings (and demonstration applications) should work well for Pytho n3.4 (or higher), on Linux and OSX platforms. Windows is not yet supported, but could be with CMake work. (We have dropped support for Python 2.x and will not accept patches to re-enable it.)
```
abadams: Pytho n -> Python

python_bindings/src/PyVar.cpp
```diff
@@ -29,11 +29,6 @@ void define_var(py::module &m) {
             .def_static("outermost", &Var::outermost)
             .def("__repr__", &var_repr)
             .def("__str__", &Var::name)
-
-            // Cannot overload a method as both static and instance with PyBind11 and Python2.x
-            // .def_static("is_implicit", (bool (*)(const std::string &)) &Var::is_implicit)
```
abadams: Does the comment imply that dropping python 2 support means we can uncomment these?

steven-johnson: No, this comment is in error -- PyBind11 doesn't support this out of the box. (We could probably find a way to make it work but since no one has complained about it so far, I'm inclined to just remove it permanently.)

steven-johnson:
Failures are unrelated

Issue #4335, opened 2019-10-23
mul_div_mod crashing for HVX backend


steven-johnson:
Appears to have been injected by https://reviews.llvm.org/rL375240 but not sure why yet -- only happens when running with multiple threads so presumably it's injecting some new shared global state that isn't threadsafe.

(As a short-term workaround I'm going to switch that test to be single-threaded)

pranavb-ca:
@dpalermo  - FYI

Seems to be a problem in ```Nfatranscriber```
```
(gdb) bt
#0  0x00007ffff3058eb3 in llvm::internal::NfaTranscriber::transition(llvm::ArrayRef<llvm::NfaStatePair>) () from libHalide.so
#1  0x00007ffff3058cd9 in llvm::internal::NfaTranscriber::transition(unsigned int) () from libHalide.so
#2  0x00007ffff3054df5 in llvm::Automaton<unsigned long>::add(unsigned long const&) () from libHalide.so
#3  0x00007ffff3054b08 in llvm::DFAPacketizer::reserveResources(llvm::MCInstrDesc const*) () from libHalide.so
#4  0x00007ffff3054f15 in llvm::DFAPacketizer::reserveResources(llvm::MachineInstr&) () from libHalide.so
#5  0x00007ffff1910a17 in llvm::HexagonPacketizerList::tryAllocateResourcesForConstExt(bool) () from libHalide.so
#6  0x00007ffff1919225 in llvm::HexagonPacketizerList::addToPacket(llvm::MachineInstr&) () from libHalide.so
#7  0x00007ffff3056a96 in llvm::VLIWPacketizerList::PacketizeMIs(llvm::MachineBasicBlock*, llvm::MachineInstrBundleIterator<llvm::MachineInstr, false>, llvm::MachineInstrBundleIterator<llvm::MachineInstr, false>) () from libHalide.so
#8  0x00007ffff191d483 in (anonymous namespace)::HexagonPacketizer::runOnMachineFunction(llvm::MachineFunction&) () from libHalide.so
#9  0x00007ffff31b3e3f in llvm::MachineFunctionPass::runOnFunction(llvm::Function&) () from libHalide.so
#10 0x00007ffff4c5ed59 in llvm::FPPassManager::runOnFunction(llvm::Function&) () from libHalide.so
#11 0x00007ffff4c5f165 in llvm::FPPassManager::runOnModule(llvm::Module&) () from libHalide.so
[...]

PR #4336, opened 2019-10-23
Make mul_div_mod singlethreaded for HVX (Issue #4335)

steven-johnson:


dsharletg:
@pranavb-ca note that this is the host side threading, indicating that it is the runtime that is doing something that is not thread safe. It is surprising that this regressed recently.

PR #4338, opened 2019-10-24
Remove boundary conditions in autodiff

BachiLi:
Currently, for each function f' generated by autodiff, we add a boundary condition setting everything outside of f to zero. I suspect that this is totally unnecessary. The correctness of the autodiff system at boundaries can be proved by induction:

Assuming we have all the derivative functions before f defined correctly, WLOG, for each function `g` that depends on `f`,  `f'`'s update can be written as
`f'(A(r)) += g'(B(r)) * e(r)`
where `r` is g's (multi-dimensional) domain and `A`, `B` are some indexing expressions, `e` is the derivative term generated by autodiff.
The scatter-gather transform only applies when `B` is identity (i.e. `g`'s left hand side are pure variables). In this case we rewrite the update above as
`f'(x) += g'(x) * e(x)`
where `x = A^(-1)(r)`
To show that the two updates are the same, we need to show that, for all `x` outside of `r`'s bounds, the update is zero. However this is trivial: because by induction we assume `g'` outputs zero correctly outside it's bounds. Therefore we don't need boundary conditions. Sometimes we only perform scatter-gather transform on a subset of `r`, where in that subset `B(r)` is identity. The argument above also applies to that because as long as one of the variable is out-of-bounds, `g'` would return zero.

The reason why we thought we need boundary condition, I think, was that we were confused by the more conservative bounds generated by the shear transformation. Even though the shear transformation generates a larger bounds compared to `r`, we are fine because the out-of-bounds access returns 0 safely in Halide.

I would love @abadams to take a look at this and try to figure out counterexamples as this could be totally wrong.

This PR also fixes a bug in autodiff's bounds inference when there is self reference. Bounds inference is still required when scatter-gather transformation fails.


BachiLi:
This also makes me realize that `e(x)` can produce out-of-bounds access to some input buffers, but it's not a new bug introduced by this PR, but a bug that was always there. Will fix it later.

src/Derivative.cpp
```diff
@@ -1365,8 +1330,8 @@ void ReverseAccumulationVisitor::propagate_halide_function_call(
     // We then substitute the original variable names back to get
     // f'(x, y, z) += g'(x, x + 1, z - 1)
     //
-    // Currently we don't want to mess with system solving yet,
-    // So we gather all arguments that contains multiple pure variables,
+    // Currently we don't want to mess with system solving yet.
```
steven-johnson: Nit: "Currently... yet" is redundant. Maybe omit the "yet" at the end.

src/Derivative.cpp
```diff
@@ -553,36 +553,6 @@ void ReverseAccumulationVisitor::propagate_adjoints(
...
 
         FuncKey func_key{func.name(), func.num_update_definitions() - 1};
-        // Set up boundary condition for the last adjoint, for
-        // non overwriting scans, we delay the boundary condition
-        // setup since the gradients depend on itself.
-        auto add_boundary_condition = [&](const FuncKey &func_key) {
-            Func &adjoint_func = adjoint_funcs[func_key];
-            const Box &bounds = func_bounds[func.name()];
-            // Save a pointer to the unbounded def. Useful for scheduling
-            FuncKey unbounded_func_key{func.name() + "_unbounded", func_key.second};
```
steven-johnson: If we are no longer saving/using the "_unbounded" def, shouldn't we also remove the public API for accessing it?

src/Derivative.cpp
BachiLi: Fixed.

src/Derivative.cpp
BachiLi: Fixed.

steven-johnson:
Code changes LGTM. You probably want @abadams to approve before landing though.

BachiLi:
> This also makes me realize that `e(x)` can produce out-of-bounds access to some input buffers, but it's not a new bug introduced by this PR, but a bug that was always there. Will fix it later.

This should fixed now. I do it by inserting `clamp` inside the adjoints using `g's` bounds, except for the access to `g'`. See https://github.com/halide/Halide/blob/b261efa62501b62966fa848088947821af4743da/test/correctness/autodiff.cpp#L1363 for related test.

I feel that *this* was the boundary condition in our mind and I implemented it wrong.

src/Derivative.h
```diff
@@ -35,10 +35,6 @@ class Derivative {
     Func operator()(const Buffer<> &buffer) const;
     Func operator()(const Param<> &param) const;
 
-    /** Get the unbounded Func, if any. If no such Func (ie no boundary condition
-     * was applied), equivalent to just calling operator(). */
-    Func get_unbounded(const Func &func, int update_id = -1) const;
-
```
abadams: The funcs method below can go away now. The fit_function example will need fixing

abadams:
TODO as discussed offline: add a zero boundary condition to adjoints passed in as inputs to the backwards pipeline.

src/Derivative.h
BachiLi: Fixed.

BachiLi:
> TODO as discussed offline: add a zero boundary condition to adjoints passed in as inputs to the backwards pipeline.

Fixed. See https://github.com/halide/Halide/pull/4338/files#diff-df5d7ab2494c349b53b54de3ca6a615fR1937

BachiLi:
Also added some comments in `Derivative.h`: 
```
*  For each Func the output depends on, and for each update of that Func,
*  including the pure definition, it generates a derivative Func stored in
*  the Derivative."
```

BachiLi:
Comments addressed. PTAL.

steven-johnson:
@abadams I'd like to land this as soon as we're comfortable with is.

PR #4340, opened 2019-10-25
Gradient autoscheduler

BachiLi:
This is a conservative autoscheduler that `compute_root` most Funcs except for the trivial ones (think of it as a -O1 optimizer for Halide). It recognizes large reduction patterns and does `rfactor` or `atomic` to parallelize on associative reduction when there's not enough parallelism in the pure variable domain. This strategy works reasonable well for gradient pipelines, and is suitable as a default option for decent but not optimal performance. This is currently the only autoscheduler that generate GPU schedules.

Running some benchmarks in the app directory gives me the following statistics (all use `halide_reuse_device_allocations(nullptr, true)` for GPU)

app | manual (CPU) | gradient-autoscheduler (CPU) | manual (GPU) | gradient-autoscheduler (GPU)
-- | -- | -- | -- | --
bilateral filter | 7.93 ms | 12.92 ms | 0.29 ms | 1.05 ms
camera_pipe | 8823.33 us | 25126 us | 605.03 us | 3347.44 us
lens_blur | 7.77 ms | 22.41 ms | 0.73 ms | 5.60 ms
local_laplacian | 42.29 ms | 128.31 ms | 0.81 ms | 14.30 ms
nl_means | 145.003 ms | out-of-memory | N/A | 82.93 ms
conv_layer | 15.46 ms | 6.89 ms | N/A | 1.90 ms
stencil_chain | 18.86 ms | 21.46 ms | N/A | 6.35 ms

Tested on a 8 core Intel CPU (16 with HT) and TITAN Xp.

I'm investigating the OOM in nl_means but it's not entirely surprising. 

I can keep improving this autoscheduler, e.g. by making inlining more aggressive. However the hope is that the ML-based GPU autoscheduler will replace this.


apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
+#include "Halide.h"
+
+#ifndef user_error
```
steven-johnson: We need a better answer here -- as a starting point, maybe copy apps/autoscheduler/Errors.h here so that the evilness is contained in a header. (But really as a separate PR we need to figure out how to consolidate.)

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+template <typename T>
+std::vector<int> sort_indices(const std::vector<T> &v) {
+    std::vector<int> idx(v.size());
+    std::iota(idx.begin(), idx.end(), 0);
+    std::sort(idx.begin(), idx.end(),
+         [&v](int i1, int i2) {return v[i1] < v[i2];});
+    return idx;
+}
+
+// Taken from src/AutoSchedule.cpp
```
steven-johnson: As per previous comment, we need to think about ways to avoid redundancy (or at least copy-n-paste) between different schedulers... I don't think it should block this PR, but perhaps it would be good to open a tracking issue of "stuff that needs to be reused by multiple scheduler plugins" so we know what the candidates are.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+
+#ifndef internal_error
+#define internal_error            Halide::Internal::ErrorReport(__FILE__, __LINE__, nullptr, 0)
+#endif
+
+namespace Halide {
+namespace Internal {
+namespace Autoscheduler {
+
+std::map<std::string, Box> inference_bounds(const std::vector<Function> &functions,
```
steven-johnson: ~most of these functions aren't used outside this file, and so should be in an anonymous namespace

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+            schedule_source << ",";
+        }
+    }
+    schedule_source << ")\n";
+}
+
+void reorder_storage(Stage stage,
+                     const std::vector<Var> &all_vars,
+                     std::ostringstream &schedule_source) {
+    // Can't reorder storage of a stage.
```
steven-johnson: Does that imply this should be a compile error? Or emit a runtime error? Or...?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+}
+
+int natural_vector_size(const Target &target, const Type &t) {
+    const bool is_integer = t.is_int() || t.is_uint();
```
steven-johnson: `Target::natural_vector_size` already exists -- why aren't we using it? (If it doesn't handle AVX512 we should augment it to do so rather than reinvent the wheel here)

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+        bool is_pure_def,
+        const std::vector<Var> &vars,
+        const std::vector<int> &var_bounds,
+        const std::vector<RVar> &rvars,
+        const std::vector<int> &rvar_bounds,
+        TailStrategy tail,
+        std::ostringstream &schedule_source) {
+    // Find the first variable that has bounds larger or equal than 64,
+    // this is our GPU thread.
+    int split_size = 64;
```
steven-johnson: Why 64, what's special about it?
Also, here and elsewhere, split_size never seems to change, should it be const?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+                r_gpu_blocks.push_back(rvars[i]);
+            }
+        }
+    } else {
+        serial_rvars = rvars;
+    }
+
+    // Fuse all gpu blocks into a single variable
+    Var fused_var("");
+    if (gpu_blocks.size() > 0) {
```
steven-johnson: here and elsewhere: prefer `!foo.empty()` to `foo.size() > 0`

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+        FuncOrStage func_or_stage,
+        int natural_vector_size,
+        bool is_pure_def,
+        const std::vector<Var> &vars,
+        const std::vector<int> &var_bounds,
+        const std::vector<RVar> &rvars,
+        const std::vector<int> &rvar_bounds,
+        TailStrategy tail,
+        std::ostringstream &schedule_source) {
+    // Find the first variable that has bounds larger or equal than 16,
```
steven-johnson: comment says 16, but I don't see where 16 is actually referenced in code

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+        std::vector<RVar> rvars;
+        rvars.reserve(reduction_vars.size());
+        for (ReductionVariable r : reduction_vars) {
+            rvars.push_back(RVar(r.var));
+        }
+        int rdomain_size = 1;
+        for (int b : rvar_bounds) {
+            rdomain_size *= b;
+        }
+        int cpu_max_domain_size = 8 * params.parallelism;
```
steven-johnson: Please add comments about why these constants (8, 4096, etc) were chosen.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+                        } else {
+                            inner_rvars.push_back(rvars[i]);
+                            inner_rvar_sizes.push_back(rvar_bounds[i]);
+                        }
+                    }
+                    schedule_source << ";\n";
+                    if (outer_rvars.size() > 0 && inner_rvars.size() > 0) {
+                        // Rfactor all the outer RVars.
+                        std::vector<std::pair<RVar, Var>> preserved;
+                        std::vector<Var> interm_vars;
```
steven-johnson: Is this supposed to be "intern_vars" or maybe "interim_vars"?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+                        preserved.reserve(outer_rvars.size());
+                        interm_vars.reserve(outer_rvars.size());
+                        for (RVar r : outer_rvars) {
+                            Var v;
+                            preserved.push_back({r, v});
+                            interm_vars.push_back(v);
+                        }
+
+                        Var factored;
+                        Func interm =
```
steven-johnson: I don't know what "interm" means.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+    // Make an environment map which is used throughout the auto scheduling process.
+    std::map<std::string, Function> env;
+    for (const auto &func : outputs) {
+        std::map<std::string, Function> local_env =
+            find_transitive_calls(func);
+        env.insert(local_env.begin(), local_env.end());
+    }
+    // Compute the topological order
+    std::vector<std::string> top_order = topological_order(outputs, env);
+    // Run a pre-pass that inline all trivial Funcs (i.e. the cost of
```
steven-johnson: inlines

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+    // computing a Func <= calling that Func).
+    // XXX: Note that the cost is estimated using heuristics based on CPU statistics
```
steven-johnson: What does XXX mean here? Is it a TODO?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,1008 @@
...
+        }
+        output_bounds_expr.push_back(Box(b));
+    }
+
+    std::map<std::string, Box> func_bounds = inference_bounds(outputs, output_bounds_expr);
+    for (const auto &it : func_bounds) {
+        const Box &bounds = it.second;
+        for (int d = 0; d < (int)bounds.size(); d++) {
+            user_assert(bounds[d].is_bounded()) << "Access to function or buffer " << it.first << " at dimension " << d << " is not bounded. "
+                                                << "We can only differentiate bounded accesses.\n";
```
steven-johnson: This seems like an odd error message -- are we doing a differentiation here that I don't see?

src/AutoScheduleUtils.h
```diff
@@ -47,8 +47,8 @@ int string_to_int(const std::string &s);
 /** Substitute every variable in an Expr or a Stmt with its estimate
  * if specified. */
 //@{
-Expr subsitute_var_estimates(Expr e);
-Stmt subsitute_var_estimates(Stmt s);
+Expr substitute_var_estimates(Expr e);
```
steven-johnson: I am in favor of fixing the spelling error in this name, but doing this sort of thing as a standalone PR in the future is probably better, to avoid confusion in reviewing.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Moved.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: I refactored these functions and expose them into Internal namespace.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Done.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Done. Issuing an internal assert.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Now we use target.natural_vector_size (it does handle AVX512). The issue was that in my test the target is unknown.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: It's 32*2 ; ) Made them constexpr.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Done

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Fixed.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Done. In general these numbers can be improved.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: interim, sorry ; )

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: interim ; )

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Fixed.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Fixed.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Fixed.

src/AutoScheduleUtils.h
BachiLi: Sorry, was too lazy.

BachiLi:
Addressed all comments. There is a known issue in the bounds inference that should be fixed by https://github.com/halide/Halide/pulls.

BachiLi:
Added a README and an example of using this with the Python bindings. I think we can do the same for the ML-based autoscheduler @abadams @aekul

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -535,8 +535,14 @@ void apply_schedule(const MachineParams &params,
...
             rdomain_size *= b;
         }
-        int cpu_max_domain_size = 8 * params.parallelism;
-        int gpu_max_domain_size = 4096;
+        // Define the thresholds for the pure domain.
+        // For CPU we want at least params.parallelism number of elements
+        // to launch threads. For GPU we want to launch at least 64 GPU blocks.
+        // We don't use a larger domain size for GPU since we can also use atomic
+        // to increase parallelism and atomics are faster on GPU.
+        // These numbers can be better tuned.
```
steven-johnson: Does this comment mean that it's worthy of an Issue being opened (and referenced in the comment)?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -204,7 +69,7 @@ void reorder_storage(Func func,
 void reorder_storage(Stage stage,
                      const std::vector<Var> &all_vars,
                      std::ostringstream &schedule_source) {
-    // Can't reorder storage of a stage.
+    internal_assert(false) << "Can't reorder storage of a stage.";
```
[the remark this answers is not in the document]
BachiLi: Fixed.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: opened and updated comments.

apps/gradient_autoscheduler/README.md
```diff
@@ -0,0 +1,21 @@
+This is a conservative autoscheduler that `compute_root` most Funcs except for the trivial ones (think of it as a -O1 optimizer for Halide). It recognizes large reduction patterns and does `rfactor` or `atomic` to parallelize on associative reduction when there's not enough parallelism in the pure variable domain. This strategy works reasonable well for gradient pipelines, and is suitable as a default option for decent but not optimal performance. This is currently the only autoscheduler that generate GPU schedules.
```
steven-johnson: "reasonably well"

steven-johnson: "generates GPU schedules"

apps/gradient_autoscheduler/test.py
```diff
@@ -0,0 +1,32 @@
...
+    params = hl.MachineParams(32, 0, 0);
+    result = p.auto_schedule(target, params)
+    print('Schedule:')
+    print(result.schedule_source)
+
+    p.compile_jit() # compile
+    buf = p.realize(1000) # compute and get the buffer
+
+if __name__ == '__main__':
+    main()
```
steven-johnson: need a trailing line-feed here

apps/gradient_autoscheduler/README.md
BachiLi: Fixed ; )

apps/gradient_autoscheduler/test.py
BachiLi: Fixed.

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -0,0 +1,874 @@
...
+    for (int i = 0; i < (int)vars.size(); i++) {
+        if (vectorized_var.name().empty() && var_bounds[i] >= split_size) {
+            vectorized_dim = i;
+            Var outer, inner;
+            func_or_stage.split(vars[i],
+                                outer,
+                                inner,
+                                split_size,
+                                tail);
+            schedule_source << "    .split(" <<
```
abadams: Style elsewhere in the codebase is to put << at the start of each line, not the end

apps/gradient_autoscheduler/Makefile
```diff
@@ -0,0 +1,53 @@
+include ../support/Makefile.inc
+
+WITH_PYTHON ?= 0
```
abadams: Why does WITH_PYTHON default to zero?

abadams: (at least add a comment explaining it)

abadams:
We should fix that manual conv layer schedule. I opened an issue: #4359

Maybe reference the issue in the timings in the readme, so that we remember to fix the timing when we fix the issue.

abadams:
LGTM apart from my comments

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Fixed.

apps/gradient_autoscheduler/Makefile
BachiLi: It defaults to 1 now. If we can't find pybind11 we prompt the user to set WITH_PYTHON to 0 if they wish to.

BachiLi:
Comments addressed. Waiting for the buildbots.

PR #4342, opened 2019-10-25
Generator outputs

kasper93:
Hello,

Another small fix for cmake, the problem was that custom command outputs were not defined correctly forcing regeneration on every build even if nothing was changed in source code.

Also minor change in generator code to exclude legacy aliases from printed list. No need to duplicate things there :)


src/Generator.cpp
```diff
@@ -918,16 +918,20 @@ int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
...
             auto it = output_name_to_enum.find(opt);
             if (it == output_name_to_enum.end()) {
                 cerr << "Unrecognized emit option: " << opt << " is not one of [";
-                for (auto k : output_name_to_enum) {
-                    cerr << k.first << " ";
+                auto end = output_info.cend();
+                auto last = std::prev(end);
+                for (auto iter = output_info.cbegin(); iter != end; ++iter) {
+                    cerr << iter->second.name;
+                    if (iter != last)
```
steven-johnson: Even single-line if statements should be enclosed in braces in Halide

steven-johnson:
Thanks for the fixes! LGTM, going to clone to run thru the testbots first though.

src/Generator.cpp
kasper93: Done

Issue #4348, opened 2019-10-25
Should halide_reuse_device_allocations default to be true?


BachiLi:
It seems to work pretty well for me. I can understand the reason to set this to false by default but still want to open this issue for discussion.

abadams:
I think it should default to true until such time as we have a better solution. I left it as false in the initial PR to keep is small.

BachiLi:
Now the default is true.

PR #4350, opened 2019-10-25
Let halide_reuse_device_allocations defaults to be true

BachiLi:
See https://github.com/halide/Halide/issues/4348. This PR depends on https://github.com/halide/Halide/pull/4347


BachiLi:
Failure due to no space left on device. Will merge once we're sure the tests are passing.

steven-johnson:
Looks like the culprit for the recent 'filling up space' issues is that the apps/autoscheduler/samples directly is never deleted; we need to add some step to testing to make sure that happens. Manually cleaning out for now.

PR #4352, opened 2019-10-25
Improved the onnx converter

benoitsteiner:


apps/onnx/benchmarking_utils.h
```diff
@@ -0,0 +1,21 @@
+#ifndef BENCHMARKING_UTILS_H_
+#define BENCHMARKING_UTILS_H_
+
+class CacheEvictor {
```
steven-johnson: I don't understand the purpose of this class. We update buffer but never do anything with it.

apps/onnx/denormal_disabler.h
```diff
@@ -0,0 +1,46 @@
...
+#include <xmmintrin.h>
+#endif
+
+class DenormalDisabler {
+public:
+    DenormalDisabler() {
+        // Record how denormals are currently handled so we can restore the current
+        // behavior when we exit the current scope and the class destructor is
+        // called. Note that the current code relies on the CSR register, which is
+        // available on all new x86 cpus.
```
steven-johnson: Nit: replace 'new' with the actual baseline you expect.

apps/onnx/model.cpp
```diff
@@ -2,483 +2,573 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
-#include "common_types.h"
-#include "onnx_converter.h"
+#include <sys/time.h>
```
steven-johnson: Pretty sure this won't compile on MSVC builds

apps/onnx/model.cpp
```diff
@@ -2,483 +2,573 @@
...
+  if (onnx_model.graph().output_size() == 0) {
+    throw std::invalid_argument("No output specified in the model");
+  }
+
+  std::unordered_set<std::string> dflt_values;
+  for (const auto& dflt : onnx_model.graph().initializer()) {
+    dflt_values.insert(dflt.name());
+  }
+  HalideModel result;
+  for (const auto& input : onnx_model.graph().input()) {
```
steven-johnson: here and many other places: `type &name` is preferred style in Halide (rather than `type& name`)

apps/onnx/model.cpp
```diff
@@ -2,483 +2,573 @@
...
+std::string auto_schedule(const HalideModel& pipeline) {
+  // Generate a schedule for the pipeline.
+  Halide::Target tgt = Halide::get_host_target();
+  auto schedule = pipeline.rep->auto_schedule(tgt);
+  return schedule.schedule_source;
 }
 
-template<typename T>
+template <typename T>
 struct Distribution {
```
steven-johnson: Here and many other places: proper indentation in Halide is 4 spaces, not two

apps/onnx/model.cpp
```diff
@@ -2,483 +2,573 @@
...
+  std::ifstream is(tmp_file);
+  std::string line;
+  while (std::getline(is, line)) {
+    std::cout << line << std::endl;
+  }
+  std::remove(tmp_file.c_str());
 }
 
-}  // namespace
+} // namespace
```
steven-johnson: Here and elsewhere: for a trailing `// comment` put exactly two spaces before the //

apps/onnx/model.py
```diff
@@ -1,3 +1,7 @@
+from __future__ import absolute_import
```
steven-johnson: FWIW, we just switched ~all the code in Halide to assume/require Py3; is this still necessary?

benoitsteiner: This is only necessary with python v2. I've removed the imports.

apps/onnx/benchmarking_utils.h
benoitsteiner: I am simply flushing the content of all the CPU caches by updating more data than what would fit in cache. This is simply needed when benchmarking performance in order to get more reliable performance numbers

apps/onnx/model.cpp
benoitsteiner: I forgot to apply clang-format to this file. This is fixed now.

apps/onnx/denormal_disabler.h
benoitsteiner: Done

apps/onnx/benchmarking_utils.h
benoitsteiner: Done

apps/onnx/model.cpp
benoitsteiner: I'll eventually switch to std::chrono::high_resolution_clock but if you don't mind and since the problem exists in the merged code I'd rather handle this in a separate PR.

steven-johnson: SGTM

apps/onnx/model.cpp
```diff
@@ -439,24 +476,65 @@ double benchmark(const HalideModel &pipeline, int num_iters, const std::string &
...
+void compile(
+    const HalideModel &pipeline,
+    const std::string &func_name,
+    const std::string &lib_name) {
+    std::vector<Halide::Argument> inputs;
+    for (const std::string &input_name : pipeline.input_names) {
+        inputs.push_back(pipeline.model->inputs.at(input_name));
+    }
+    Halide::Target tgt = Halide::get_host_target();
+    // tgt.set_feature(Halide::Target::Debug, true);
```
steven-johnson: Why are these commented out? Add comment or remove entirely.

apps/onnx/model.cpp
```diff
@@ -439,24 +476,65 @@ double benchmark(const HalideModel &pipeline, int num_iters, const std::string &
...
+    // tgt.set_feature(Halide::Target::NoBoundsQuery, true);
+    // tgt.set_feature(Halide::Target::TracePipeline, true);
+    // tgt.set_feature(Halide::Target::TraceRealizations, true);
+    // pipeline.rep->compile_to_lowered_stmt(std::string("/tmp/")+lib_name+".stmt",
+    // inputs, Halide::Text, tgt);
+    pipeline.rep->compile_to_file(
```
steven-johnson: /tmp/ doesn't exist on Windows systems

PR #4354, opened 2019-10-26
Add Buffer.defined() checks in Python bindings

BachiLi:
Currently the following Python code causes segmentation fault:
```
import halide as hl
b = hl.Buffer()
print(b)
```
Is there any reason why we want the users to construct an empty buffer? If no I propose just to remove these constructors in Python.

### DIFF
--- a/apps/bilateral_grid/filter.cpp
+++ b/apps/bilateral_grid/filter.cpp
@@ -27,11 +27,6 @@ int main(int argc, char **argv) {
     Buffer<float> input = load_and_convert_image(argv[1]);
     Buffer<float> output(input.width(), input.height());
 
-    // Let the Halide runtime hold onto GPU allocations for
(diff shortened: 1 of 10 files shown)


steven-johnson:
I don't want to just disable it without finding out why it is broken.

BachiLi:
> I don't want to just disable it without finding out why it is broken.

I think it is because `Buffer.__repr__()` (https://github.com/halide/Halide/blob/master/python_bindings/src/PyBuffer.cpp#L534) calls `get_buffer_shape` and `Buffer.buf` is `nullptr`. I can fix that too. Which one do you think is better?

steven-johnson:
I'd strongly vote for having `get_buffer_shape` just return an empty vector if the `Buffer` is invalid.

BachiLi:
I messed up the commits when trying to rebase. Let me open a new PR.

PR #4356, opened 2019-10-28
Change Func.in to Func.inside in Python

BachiLi:
Currently `Func.in()` in Python is not callable since `in` is a Python keyword. I propose changing it to `Func.inside()`.

### DIFF
--- a/python_bindings/correctness/basics.py
+++ b/python_bindings/correctness/basics.py
@@ -164,6 +164,24 @@ def test_basics4():
     f.compute_root()
     f.compile_jit()
 
+def test_basics5():
+    x, y = hl.Var('x'), hl.Var('y')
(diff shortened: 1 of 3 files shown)


abadams:
I would prefer the name from #4073 , but that contributor never added a test. What do you think? "inside" sounds a bit odd to me.

BachiLi:
> I would prefer the name from #4073 , but that contributor never added a test. What do you think? "inside" sounds a bit odd to me.

Totally missed that PR. Changed to `in_` now.

BachiLi:
Error seems unrelated. I will let you decide whether to merge this.

steven-johnson:
I'm not wild about `in_` either but it's better than being broken, so LGTM

PR #4357, opened 2019-10-28
Hexagon: Fix VTCM support on targets > v65

kasper93:


steven-johnson:
LGTM

src/CodeGen_Hexagon.cpp
```diff
@@ -2869,7 +2869,7 @@ void CodeGen_Hexagon::visit(const Allocate *alloc) {
         }
     } else if (alloc->memory_type == MemoryType::VTCM &&
                !alloc->new_expr.defined()) {
-        if (!target.has_feature(Target::HVX_v65)) {
+        if (!is_hvx_v65_or_later()) {
             user_error << "VTCM store_in requires hvx_v65 target feature.\n";
```
dsharletg: This error message is now a bit stale.

src/LLVM_Runtime_Linker.cpp
```diff
@@ -1026,7 +1026,7 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
                 } else if (t.has_feature(Target::HVX_128)) {
                     modules.push_back(get_initmod_hvx_128_ll(c));
                 }
-                if (t.has_feature(Target::HVX_v65)) {
+                if (t.features_any_of({Target::HVX_v65, Target::HVX_v66})) {
```
dsharletg: Would it be better to rewrite this to `!t.features_any_of({Target::HVX_v60, ...})` so future new architectures won't need this fix?

And the following 3 more of the same pattern.

shubhamp-ca: > Would it be better to rewrite this to `!t.features_any_of({Target::HVX_v60, ...})` so future new architectures won't need this fix?
> 
> And the following 3 more of the same pattern.

@dsharletg I doubt that would work, as currently we pick a default isa_version (v60 presently) if there is no "HVX_v**" feature present. So it is possible that {Target::HVX_v60, ...} is not present, but hvx defaulting to an earlier version.

dsharletg: That makes sense, unfortunately. I'm OK with this as-is, though it would be nice to fix the error message.

kasper93: > That makes sense, unfortunately.

Target feature handling is little bit clunky imo. But this is a story for another time.

> though it would be nice to fix the error message.

Done.

Issue #4281, opened 2019-09-27
is_nan/is_inf/is_finite should produce errors if used outside strict_float


steven-johnson:
is_nan (and is_inf/is_finite, after they land) aren't meaningful or reliable unless they are used inside a strict_float context; we should probably add checks post-lowering to verify that they are evaluated inside such a context, and error if not.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the
> reason below. The reasoning was checked against the current source, and a second pass
> tried and failed to refute it, but no human verified this specific text.
> **If the reason is wrong, please reopen the issue and say so** — some fraction of these
> are expected to be wrong.

**Reason (CLOSE_FIXED):** Rather than erroring, the resolution taken was to quietly wrap the argument in strict_float() so the results are always meaningful, which satisfies the issue's intent.

**Evidence:** Commit 8fa7c46697b "Make is_nan/is_inf/is_finite enforce strict_float() (Issue #4281)"; current code at src/IROperator.cpp:2419-2425 (and the is_inf/is_finite equivalents) does `if (!is_const(x)) { x = strict_float(x); }`; documented at src/IROperator.h:1058-1071.

Closing: fixed by 8fa7c46697b — is_nan/is_inf/is_finite now wrap their argument in strict_float() automatically (src/IROperator.cpp:2419), so they're meaningful regardless of the surrounding float mode.

PR #4360, opened 2019-10-28
Make is_nan/is_inf/is_finite require strict_float() (Issue #4281)

steven-johnson:
If the Expr being evaluated doesn't requires strict_float, the results are meaningless; this adds sniffing to require this to be the case, making it a compile-time error to call these if the Expr can't be proven to be using strict_float.

(Also: drive-by cleanup of code in StrictifyFloat that was a holdover from when it was possible to have contexts in which strict-float was not allowed)

### DIFF
--- a/src/IROperator.h
+++ b/src/IROperator.h
@@ -929,15 +929,18 @@ Expr round(Expr x);
 Expr trunc(Expr x);
 
 /** Returns true if the argument is a Not a Number (NaN). Requires a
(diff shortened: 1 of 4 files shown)


abadams:
Can't we just automatically inject a strict_float instead of making it an error?

steven-johnson:
Yeah, @zvookin suggested that too. Stay tuned.

PR #4365, opened 2019-10-29
gradient_autoscheduler fixes

steven-johnson:
- Migrate the aslog() utility from apps/autoscheduler and use aslog(1) instead of std::cerr (schedulers need to be *absolutely silent* by default to avoid build-system spam)
- Add gradient_autoscheduler to 'make test_apps'
- build with -fvisibility=hidden to save code size
- default WITH_PYTHON to 0... as written, this requires every piece of code using to link to Python, which is definitely not what we want. (The code should be modified to either have a separate library for Python usage, or to find a way to safely lazy-load the Python extension library.)
- Use `$(PYTHON)-config --cflags` to ensure path to Python.h can be found (if WITH_PYTHON=1 is set)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1964,6 +1964,7 @@ TEST_APPS=\
 # (see https://github.com/halide/Halide/issues/4069)
 ifneq ($(OS), Windows_NT)
 	TEST_APPS += autoscheduler
+	TEST_APPS += gradient_autoscheduler
 endif
 
 TEST_APPS_DEPS=$(TEST_APPS:%=%_test_app)
--- a/apps/gradient_autoscheduler/ASLog.cpp
+++ b/apps/gradient_autoscheduler/ASLog.cpp
@@ -1,0 +1,52 @@
+#include "ASLog.h"
+
+namespace Halide {
+namespace Internal {
+
+namespace {
+
+std::string get_env_variable(char const *env_var_name) {
+    if (!env_var_name) {
+        return "";
+    }
+
+    #ifdef _MSC_VER
+    // call getenv_s without a buffer to determine the correct string length:
+    size_t length = 0;
+    if ((getenv_s(&length, NULL, 0, env_var_name) != 0) || (length == 0)) {
+        return "";
+    }
+    // call it again to retrieve the value of the environment variable;
+    // note that 'length' already accounts for the null-terminator
+    std::string lvl(length - 1, '@');
+    size_t read = 0;
+    if ((getenv_s(&read, &lvl[0], length, env_var_name) != 0) || (read != length)) {
+        return "";
+    }
+    return lvl;
+    #else
+    char *lvl = getenv(env_var_name);
+    if (lvl) return std::string(lvl);
+    #endif
+
+    return "";
+}
+
+}  // namespace
+
+int aslog::aslog_level() {
+    static int cached_aslog_level = ([]() -> int {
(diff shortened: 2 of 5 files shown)


steven-johnson:
FWIW, I can't get test.py to work at all, and I'm not sure I understand how it's *supposed* to work:
- It fails for me at `import halide`; you clearly have PYTHONPATH set up in a way that I don't
- I presume that the intent of making `libgradient_autoscheduler` a Python module is so that it becomes implicitly registered as 'the' active autoscheduler, but may I gently suggest that this is misguided; if nothing else, this method of use means you couldn't use it in conjunction with any other autoscheduler (which is not an unreasonable expectation).

BachiLi:
> FWIW, I can't get test.py to work at all, and I'm not sure I understand how it's _supposed_ to work:
> 
> * It fails for me at `import halide`; you clearly have PYTHONPATH set up in a way that I don't
> * I presume that the intent of making `libgradient_autoscheduler` a Python module is so that it becomes implicitly registered as 'the' active autoscheduler, but may I gently suggest that this is misguided; if nothing else, this method of use means you couldn't use it in conjunction with any other autoscheduler (which is not an unreasonable expectation).

Both the Halide Python modules and the autoscheduler module need to be in the PYTHONPATH for this (as recommended in the readme in python_bindings). 

I guess we can have a `.active()` method for each autoscheduler to resolve this.

steven-johnson:
> I guess we can have a .active() or .register() method for each autoscheduler to resolve the second issue.

I'm currently working on a PR to deal with this a bit more sanely. Stand by.

steven-johnson:
> Both the Halide Python modules and the autoscheduler module need to be in the PYTHONPATH for this (as recommended in the readme in python_bindings). For test.py I added ./bin to the system path so it should find the autoscheduler module.

For tests, you should augment PYTHONPATH (or other lookup paths) in the Makefile and avoid needing system settings (the python_bindings Makefile does this).

Additionally, `test.py` should really be built/tested as part of `make test`.

steven-johnson:
Also, it's worth noting that 'test.cpp' isn't really a good automated test (it requires human examination to verify correctness). Some tests that automatically verify output vs some expected schedule is really needed.

steven-johnson:
So, looking at this further, it looks like wrapping `libgradient_autoscheduler` in a Python module here is pretty much entirely a convenience for getting the library loaded; it doesn't introduce *any* new Python API, but it introduces a new Python dependency even when not needed by its consumers.

May I propose that instead (after this and https://github.com/halide/Halide/pull/4367 land), we remove that part entirely, and instead add a Python API for `load_plugin`.

BachiLi:
> So, looking at this further, it looks like wrapping `libgradient_autoscheduler` in a Python module here is pretty much entirely a convenience for getting the library loaded; it doesn't introduce _any_ new Python API, but it introduces a new Python dependency even when not needed by its consumers.
> 
> May I propose that instead (after this and #4367 land), we remove that part entirely, and instead add a Python API for `load_plugin`.

That sounds perfect for me.

steven-johnson:
> That sounds perfect for me

OK, I'll leave this as WITH_PYTHON=0 by default for now (you can override locally); making that change should allow it to go away entirely.

steven-johnson:
(Needs review approval tho)

BachiLi:
Sorry, not familiar with these yet ; )

PR #4366, opened 2019-10-29
Add Buffer.defined() checks in Python bindings when printing buffers

BachiLi:
Current Python bindings crashes when doing
```
import halide as hl
b = hl.Buffer()
print(b)
```
This is because the `__repr__` function in hl.Buffer does not check if the Buffer is defined when accessing its member. This PR fixes that.


steven-johnson:
Failure is unrelated.

Issue #4057, opened 2019-07-25
AutoScheduler nomenclature needs clarity


steven-johnson:
At present, we have two autoschedulers in our codebase, but there don't appear to be well-defined terms for each (e.g. “classic” vs “ML hotness” or whatnot); we should really settle on some terms to avoid confusion going forward. (Since there is the possibility of additional autoschedulers being added in the future, avoiding terms like 'old', 'new', etc are probably wise.)

abadams:
One option is labelling them via the papers: Mullapudi_2016, Adams_2019. Or maybe siggraph_2016, siggraph_2019. Precise, and the most useful for academic comparisons, but not intuitive. Also using names is possibly unfair to all the other authors and developers, e.g. Patricia ultimately had as much influence on the Mullapudi code/algorithm as Ravi, and tons of people other than me worked on the newer one. We could also name them by the core algorithm involved: clustering, beam_search.

steven-johnson:
In a private branch, I'm currently using the placeholder names "classic" and "v2019" for now.

steven-johnson:
Update: as of https://github.com/halide/Halide/pull/4094, the 'name' of the scheduler used now is embedded in the generated `.schedule.h` file; for now, the names being used are `src/AutoSchedule` and `apps/autoscheduler/AutoSchedule`, which are... suboptimal.

steven-johnson:
Update: soon we'll also have apps/gradient_autoscheduler, so 'v2019' is clearly a bad choice

abadams:
I'm in favor of citation-style then: Adams2019, Mullapudi2016, Li2018.

BachiLi:
Sounds good to me

PR #4367, opened 2019-10-29
Allow multiple autoschedulers to be registered (Issue #4057)

steven-johnson:
Currently, the plugin autoschedulers take over all autoschedule hooks once loaded; there's no way to mix and match between them, and no API-level way to control which one you get. This reworks this nontrivially:

- Names for the autoschedulers are revised per the comments in #4057
- We now keep a global map of all known schedulers, with the 'classic' (aka 'Mullapudi2016') always available for now
- When you load a plugin (either explicitly via dlopen, or manually via the `-p` option to Generator_main), it does *not* change the default scheduler, it just adds itself to the map.
- There is a new API to set the default autoscheduler to use (`Pipeline::set_default_autoscheduler_name`)
- There is also a new generator_main flag (`-s`) to set the default autoscheduler to use
- The existing call to `Pipeline::auto_schedule()` uses the default autoscheduler name; there is also a new variant in which you can pass the autoscheduler name explicitly.

Stuff to consider:
- I contemplated adding the `autoscheduler_name` as a new field to MachineParams, but decided that was a bad fit.
- I contemplated avoiding the `-s` argument for Generators entirely and instead somehow overloading `auto_schedule=` instead, but this seems fraught with weirdness to make legacy code work without changes. (It still feels like something that might make more sense as a GeneratorParam, but not sure yet.)
- It might make sense to (someday) have plugin loading be done implicitly when `-s` is specified, but definitely not yet.


steven-johnson:
Also moved the plugin-loading code to Util.cpp, to avoid spraying this error-prone code across even more places than it already is located

apps/autoscheduler/test.cpp
```diff
@@ -1,50 +1,13 @@
...
-        LPVOID last_err_msg;
-        FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM |
-                           FORMAT_MESSAGE_IGNORE_INSERTS,
-                       nullptr, last_err, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
-                       reinterpret_cast<LPSTR>(&last_err_msg), 0, nullptr);
-        std::cerr << "Failed to load: " << lib << "\n";
-        std::cerr << "LoadLibraryA failed with error " << last_err << ": "
-                  << static_cast<char *>(last_err_msg) << "\n";
-        LocalFree(last_err_msg);
+    if (!::Halide::Internal::load_plugin("auto_schedule", std::cerr)) {
```
BachiLi: Would be nice to have a comment here explaining what "auto_schedule" means.

src/Pipeline.h
```diff
@@ -144,16 +151,30 @@ class Pipeline {
...
+    static void add_autoscheduler(const std::string &autoscheduler_name, const AutoSchedulerFn autoscheduler);
+
+    /** Globally set the default autoscheduler method to use whenever
+     * autoscheduling any Pipeline when no name is specified. If the autoscheduler_name isn't in the
+     * current table of known autoschedulers, assert-fail.
+     *
+     * At this time, well-known autoschedulers include:
+     *  "Mullapudi2016" -- heuristics-based; the first working autoscheduler; currently built in to libHalide
+     *  "Adams2019"     -- aka "the ML autoscheduler"; currently located in apps/autoscheduler
+     *  "Li2019"        -- aka "the gradient autoscheduler"; currently located in apps/gradient_autoscheduler
```
BachiLi: Li2018 (it's the gradient Halide paper)

apps/gradient_autoscheduler/test.cpp
```diff
@@ -1,50 +1,13 @@
...
-    } else {
-        lib = "libgradient_autoscheduler.so";
-    }
 
-    if (dlopen(lib.c_str(), RTLD_LAZY) == nullptr) {
-        std::cerr << "Failed to load: " << lib << ": " << dlerror() << "\n";
-        return 1;
-    }
-#endif
+    Pipeline::set_default_autoscheduler_name("Li2019");
```
BachiLi: Li2018

apps/gradient_autoscheduler/test.py
```diff
@@ -15,13 +15,13 @@ def main():
...
 
-    # Setup 
+    # Setup
     f_2.set_estimate(x, 0, 1000)
     p = hl.Pipeline(f_2)
     target = hl.Target()
     # Only first parameter is used (number of cores on CPU)
     params = hl.MachineParams(32, 0, 0);
-    result = p.auto_schedule(target, params)
+    result = p.auto_schedule('Li2019', target, params)
```
BachiLi: Li2018

src/Pipeline.h
```diff
@@ -144,16 +151,30 @@ class Pipeline {
...
+    /** Add a new the autoscheduler method with the given name. Does not affect the current default autoscheduler.
+     * It is an error to call this with the same name multiple times. */
+    static void add_autoscheduler(const std::string &autoscheduler_name, const AutoSchedulerFn autoscheduler);
+
+    /** Globally set the default autoscheduler method to use whenever
+     * autoscheduling any Pipeline when no name is specified. If the autoscheduler_name isn't in the
+     * current table of known autoschedulers, assert-fail.
+     *
+     * At this time, well-known autoschedulers include:
+     *  "Mullapudi2016" -- heuristics-based; the first working autoscheduler; currently built in to libHalide
```
BachiLi: Maybe link to the paper: http://graphics.cs.cmu.edu/projects/halidesched/
so people know what Mullapudi2016 means.

src/Pipeline.h
```diff
@@ -144,16 +151,30 @@ class Pipeline {
...
+     *  "Adams2019"     -- aka "the ML autoscheduler"; currently located in apps/autoscheduler
```
BachiLi: Maybe link to the paper: https://halide-lang.org/papers/autoscheduler2019.html
so people know what Adams2019 means.

src/Pipeline.h
BachiLi: Maybe link to the paper: https://people.csail.mit.edu/tzumao/gradient_halide/
so people know what Li2018 means.

src/Util.h
```diff
@@ -455,6 +455,11 @@ struct IsRoundtrippable {
 /** Emit a version of a string that is a valid identifier in C (. is replaced with _) */
 std::string c_print_name(const std::string &name);
 
+/** Load a plugin (e.g. for custom autoschedulers). If the string doesn't contain
```
BachiLi: Maybe add that the plugin is in the form of a dynamic library.

apps/autoscheduler/test.cpp
steven-johnson: done

src/Pipeline.h
steven-johnson: done

BachiLi:
Looks good to me at least.

Issue #4368, opened 2019-10-30
Python bindings should be included in the distrib/ folder


steven-johnson:
In theory, everything in apps/ should rely only on things available in the distrib/ folder; a few things are peeking 'up' into the Halide tree for minor things (e.g. binary2cpp), but some apps are now using the python bindings, so we really need to migrate these.

alexreinking:
There is some support for this upstream already and more fully realized in #5135. However, there's more to distribute than just the `.so` module, and even that should be built using the "anylinux" CentOS docker image if we want to publish on PIP. More work needs to be done.

alexreinking:
Python bindings are only built by CMake now, so this issue isn't relevant anymore.

PR #4369, opened 2019-10-30
Add Halide::load_plugin

steven-johnson:
Promote this to a first-class Halide API, along with Python wrapper. Remove the python-related code from apps/gradient_autoscheduler and use this instead. Also drive-by cleanup in gradient_autoscheduler to run all the tests properly.


BachiLi:
Seems that `test.py` in the gradient autoscheduler is failing at `import halide as hl` : (

steven-johnson:
> Seems that test.py in the gradient autoscheduler is failing at import halide as hl : (

ouch, yeah, there is likely a missing dependency; python_bindings have to be built beforehand since it's not currently part of distrib/ (see https://github.com/halide/Halide/issues/4368)

PR #4373, opened 2019-11-01
Fix OSX+Homebrew+Python issue for HelloPyTorch

steven-johnson:
Some combinations have a known issue and the `pip install` part of the Makefile will fail with `DistutilsOptionError: must supply either home or prefix/exec-prefix — not both`

Adding this dummy setup.cfg is (apparently) a reasonably safe solution, and works on my setup.

attn @mgharbi


steven-johnson:
(Also added some drive-by reformatting of the Makefile, to make some upcoming changes I'm working on easier to review)

steven-johnson:
Monday Morning Review Ping

Issue #4371, opened 2019-10-31
Autoscheduler ARM64 build broken


slamcorejon:
The num_features() function in ScheduleFeatures, /apps/autoscheduler/Featurization.h, currently returns an int.  This fails compilation in apps/autoscheduler/AutoSchedule.cpp line 2380.  Fix is to return a size_t from this function.

steven-johnson:
fixed

PR #4375, opened 2019-11-01
ScheduleFeatures::num_features() and PipelineFeatures::num_features()…

steven-johnson:
… should return size_t (Issue #4371)

### DIFF
--- a/apps/autoscheduler/Featurization.h
+++ b/apps/autoscheduler/Featurization.h
@@ -13,7 +13,7 @@ namespace Internal {
 // The algorithm-specific features. For legacy reasons these are
 // called PipelineFeatures in the code.
 struct PipelineFeatures {
-    static constexpr int num_features() {
+    static constexpr size_t num_features() {
         return sizeof(PipelineFeatures) / sizeof(int);
     }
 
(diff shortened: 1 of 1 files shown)


abadams:
I think this value gets embedded in featurization files, so it should probably not use an architecture-varying-size. How about an explicit int cast instead?

steven-johnson:
literally every caller I found was assigning the result to a size_t immediately (or using it as an index). Where do you see it being written?

abadams:
Disregard, I misread this as the version field.

alexreinking:
Nit: Based on my reading of the standard, both `size_t` and `int` have architecture-varying sizes. Though if we only support running the Halide compiler on a fixed set of architectures and compiled with a fixed set of compilers, it might be that `size_t` is equivalent to `uint64_t` and `int` is `int32_t`.

Regarding `size_t` the standard says in 18.2.6-7
> The type size_t is an implementation-defined unsigned integer type that is large enough to contain the size in bytes of any object.
>[Note:It is recommended that implementations choose types for ptrdiff_t and size_t whose integer conversion ranks (4.13) are no greater than that of signed long int unless a larger size is necessary to contain all the possible values.— end note]

And in 3.9.1.2, `int` is defined as:
> Plain `int`s have the natural size suggested by the architecture of the execution environment [44];
> ...
> [44] that is, large enough to contain any value in the range of INT_MIN and INT_MAX, as defined in the header <climits>.

If we need a fixed size for serialization and we want standards compliance, we have to use the `cstdint` types.

steven-johnson:
Again: where are we serializing this value?

alexreinking:
@steven-johnson -- I don't think we are doing so here, but if we are using `int` or `size_t` elsewhere for serialization, then we shouldn't and I thought having the standard references around would be useful.

PR #4380, opened 2019-11-05
Fix offload mode autoscheduler results.

shubhamp-ca:
Currently, when autoscheduler is invoked in offload mode, the
the emit options {featurization and schedule} are not working
as expected and bailing out due to auto_scheduler_results not
present in the lowered module, resulting in empty feature and
schedule files produced. This happens because, when offloaded
submodules are recursively resolved, the autoscheduler results
are not copied back into the lowered module. This change fixes
this issue by setting back the auto_scheduler_results in the
ModuleContents of the module that is lowered after resolving
submodules.

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -530,7 +530,8 @@ Module Module::resolve_submodules() const {
         auto buf = copy.compile_to_buffer();
         lowered_module.append(buf);
     }
-
+    // Copy the autoscheduler results back into the lowered module after resolving the submodules.
+    lowered_module.set_auto_scheduler_results(*contents->auto_scheduler_results.get());
     return lowered_module;
 }
 
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -531,7 +531,9 @@ Module Module::resolve_submodules() const {
         lowered_module.append(buf);
     }
     // Copy the autoscheduler results back into the lowered module after resolving the submodules.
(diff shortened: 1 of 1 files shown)


steven-johnson:
Thanks for the fix! Would it be possible to add a test that would catch this (since clearly we don't have one now)?

shubhamp-ca:
> Thanks for the fix! Would it be possible to add a test that would catch this (since clearly we don't have one now)?

I considered adding a test with the change, but I am not sure if either the default autoscheduler or the Adams2019 are currently capable of emitting schedules which lead to creation of submodules within the module which is evaluated. This bug was only hit when the apps/autoscheduler was forced to emit ".hexagon()" calls at the root stage. Any ideas on how this could be tested are welcome.

Though, this bug probably doesn't affect anything currently, it seemed like an "okay" fix for whenever autoschedulers are able to autonomously emit offload calls.

shubhamp-ca:
Also, not really sure what is causing the buildbot failure here. The error is -- 

`CodeGen_LLVM.cpp:993:66: error: ‘void llvm::GlobalObject::setAlignment(unsigned int)’ is deprecated [-Werror=deprecated-declarations]
storage->setAlignment((unsigned)sizeof(halide_scalar_value_t));`

steven-johnson:
re: buildbot failure, that looks like an ongoing LLVM change we're aware of; we can ignore it here.

steven-johnson:
Thanks for the fix!

PR #4384, opened 2019-11-06
Revert PR #4336

pranavb-ca:
Revert PR #4336 as underlying issue in LLVM has been addressed by LLVM commit 041f35c
  Link - https://github.com/llvm/llvm-project/commit/041f35c468088d315bae6c2a71ec901a12cca1b5
This should fix Issue #4335


steven-johnson:
Failure in linux trunk:

bash: line 1: 15533 Segmentation fault      (core dumped) /home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-trunk/halide-build/bin/correctness_mul_div_mod 2> stderr_mul_div_mod.txt > stdout_mul_div_mod.txt

dpalermo:
Bummer.  Are we sure that the buildbot picked up this change?  We aren't seeing any failures locally with this fix.  Can you verify that it was present in the llvm source tree used in this build?  If so, something must still not be thread safe (and unfortunately/apparently harder to hit).

commit 041f35c468088d315bae6c2a71ec901a12cca1b5
Author: James Molloy <James Molloy jmolloy@google.com>
Date:   Tue Nov 5 22:53:56 2019 +0000

    [Automaton] Make Automaton thread-safe

steven-johnson:
From https://buildbot.halide-lang.org/master/#/builders/18/builds/628/steps/5/logs/stdio:
```
At revision 375507.
```
Should that version of LLVM trunk have the fix?

dpalermo:
That looks a bit too old to me:

<commit
   revision="375505">
<author>dmgreen</author>
<date>2019-10-22T15:39:47.756967Z</date>
</commit>

We need to be beyond Nov 5.  Is your internal llvm repo lagging tip by that much?  If so we will need to keep the workaround in place until you've caught up to:

commit 041f35c468088d315bae6c2a71ec901a12cca1b5
Author: James Molloy <James Molloy jmolloy@google.com>
Date:   Tue Nov 5 22:53:56 2019 +0000

pranavb-ca:
In rev 375507, Automaton.h at include/llvm/Support is stuck on Rev 375240 (See https://llvm.org/viewvc/llvm-project/llvm/trunk/include/llvm/Support/?pathrev=375507) which is the patch that introduced this problem.

dpalermo:
The reason you are stuck on this revision, might be because of the llvm.org transition from subversion to github (which was supposed to complete on Oct 21, 2019)

        https://llvm.org/docs/Proposals/GitHubMove.html

The revision you are pointing to happens to be the last commit that was tagged with a llvm-svn version.  If you are still checking out sources from the old subversion repo, you will need to transtition to github.

steven-johnson:
Failing test is pending buildbot update per https://github.com/halide/build_bot/pull/47

steven-johnson:
arm failure is unrelated

PR #4391, opened 2019-11-11
avoid reinterpret_cast in onnx_converter

steven-johnson:
Existing code will fail when compiled with -Werror=strict-aliasing


steven-johnson:
Gentle review ping

PR #4392, opened 2019-11-11
Refined reduction order in check on synthesized rules

jn80842:


apps/super_simplify/reduction_order.cpp
```diff
@@ -65,18 +65,153 @@ std::set<std::string> find_divisors(const Expr &e) {
...
+
+int get_nonlinear_op_count(const Expr &e) {
+    NonlinearOpsCount nl;
+    e.accept(&nl);
+    return nl.counter;
+}
+
+class IsConstant : public IRVisitor {
+    void visit(const Variable *op) override {
+        if (op->name[0] == 'c') {
```
steven-johnson: Is this some convention specific to this app?

abadams: Yeah, it's a bit of a hack, but note that this pr is not into master

Issue #4381, opened 2019-11-05
Halide's use of pair<Expr, Expr> in API is inconsistent


steven-johnson:
In some places in Halide (e.g. Func::set_estimates), the pair is `(min, extent)`, but in others (e.g. Derivative.h), the pair is `(min, max)` -- this is a confusion that is too easy to make. Can we standardize on one or the other everywhere?

(Adding even more fun is that `ArgumentEstimates` uses a bespoke `struct MinAndExtent` instead of `pair<Expr, Expr>` -- maybe that would be clearer to use in more places?)

abadams:
We have structs for each already: Anything that's a literal `pair<Expr, Expr>` should probably either use an Interval (min, max), or a Range (min, extent)

steven-johnson:
I took a look at using `Range` for the various `set_estimates` and `BoundaryConditions` calls, but there is one minor gotcha: `BoundaryConditions` deliberately allows undefined Exprs (== 'no boundary here'), but `Range` is confused on this: you can construct a Range with both values undefined (via the default ctor), but the pairwise-ctor will crash (since it calls `.type()` on each to match types). Wouldn't be hard to modify `Range` to allow undefined value(s) via that ctor, but I'm not sure if that would open up issues elsewhere. Thoughts?

PR #4393, opened 2019-11-11
Use Halide::Range instead of std::pair<Expr, Expr> (Issue #4381)

steven-johnson:
This replaces the use of `pair<Expr, Expr>` with `Range` (or `vector<pair>` -> `Region`) in the public APIs for `Func::set_estimates()`, `BoundaryConditions`, and `RDom`; the intent here is to make it clearer that these take (min, extent) rather than (min, max) pairs. Mostly pretty clean conversion, but a couple of things to note:
- previously, `Range` would crash if you tried to construct one with one (or two) undefined Exprs; this was relaxed, as it is required for certain BoundaryCondition operations
- previously, there was no enforcement that the two values in `pair<Expr, Expr>` has matching types; now, there is. (Interestingly, I found only one case in all of Halide where this wasn't already the case.)


steven-johnson:
(Note that `Derivative` is still using `pair<Expr, Expr>`; I intend to consider how/whether to deal with that after this lands.)

steven-johnson:
arm failure is unrelated, PTAL

steven-johnson:
> I am increasingly of the opinion that std::pair is a code smell.

Yeah, much like using a `bool` argument as one of a list of arguments.

Issue #4394, opened 2019-11-12
Update README + buildbots for LLVM to use GitHub


steven-johnson:
Per https://llvm.org/docs/Proposals/GitHubMove.html, trunk LLVM is now using GitHub rather than Subversion; we need to update the buildbots to use this source, and update the README to reference the new path(s).

PR #4395, opened 2019-11-12
Update README for LLVM build instructions (Issue #4394)

steven-johnson:


README.md
```diff
@@ -331,26 +333,21 @@ For examples of using the `hexagon` scheduling directive on both the simulator a
...
-    # Or:
-    #    git clone http://llvm.org/git/llvm.git .
-    #    git clone http://llvm.org/git/clang.git llvm/tools
-    mkdir build
-    cd build
-    cmake -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_TARGETS_TO_BUILD="X86;ARM;NVPTX;AArch64;Mips;PowerPC;Hexagon" -DLLVM_ENABLE_ASSERTIONS=ON -DCMAKE_BUILD_TYPE=Release ..
-    make -j8
-    export LLVM_CONFIG=<path to llvm>/build/bin/llvm-config
+#### 1. Obtain and build LLVM and clang v8.0 or later from llvm.org
+The Hexagon backend is currently under development, so it's best to use trunk llvm. (These are the same instructions as above for building Clang/LLVM.
```
pranavb-ca: I really think we should make this more instructive rather than a suggestion to use trunk llvm. The reason is that for the longest time we have had a buildbot that has tested Hexagon only with trunk llvm and nothing else. So, I have no idea what'll happen if we even try to stick to the policy of supporting the last two releases of LLVM.

steven-johnson: I'm confused; what is the recommendation we are trying to make here?

pranavb-ca: "If you wish to use Halide to target Hexagon with HVX, you should use trunk LLVM"

steven-johnson: PTAL

PR #4398, opened 2019-11-12
Fix set_host_alignment() in simd_op_check

steven-johnson:
We have been calling set_host_alignment(128) on our input params for ~ever, but this has always been wrong; as the argument is in bytes (not bits), and most of our targets provide only a 32-byte guarantee at the memory allocator level (x86 does 64-bytes, see LLVM_Runtime_Linker.cpp)

(Note that I'm not sure if the _hvx variant really should be 128 bytes or not, please review carefully.)

### DIFF
--- a/test/correctness/simd_op_check.cpp
+++ b/test/correctness/simd_op_check.cpp
@@ -1,8 +1,6 @@
 #include "Halide.h"
 #include "simd_op_check.h"
 
-#include <fstream>
-#include <future>
 #include <stdarg.h>
 #include <stdio.h>
 #include <string.h>
--- a/test/correctness/simd_op_check.h
+++ b/test/correctness/simd_op_check.h
@@ -261,9 +261,6 @@ class SimdOpCheckTest {
     virtual void setup_images() {
         for (auto p : image_params) {
             p.reset();
-            constexpr int kHostAlignmentBits = 128;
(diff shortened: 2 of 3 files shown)


dsharletg:
I'm pretty sure this is actually correct. Hexagon needs 128 byte alignment, and it seems harmless on other targets (i.e. we should just call set_host_alignment with the biggest alignment we need for any target).

steven-johnson:
It's not harmless on targets that only promise (e.g.) 32-byte alignment for halide_malloc, because this call injects an assert-check that makes us fail if the allocation isn't 128-aligned.

dsharletg:
I guess this needs to be target dependent (`natural_vector_size<uint8_t>()`?), or the buffers allocated should always be allocated with 128 byte alignment even if the target's halide_malloc doesn't do that.

steven-johnson:
Not sure I understand why changing the allocator is preferable to this fix.

dsharletg:
This fix is incorrect for Hexagon, which actually needs 128 *byte* alignment, not 128 bits?

steven-johnson:
> needs 128 byte alignment

fixed

dsharletg:
The current fix looks good. But the more I think about it, we probably just don't need set_host_alignment at all on most targets. How about just deleting set_host_alignment from the non-HVX codepath?

steven-johnson:
Done

PR #4399, opened 2019-11-12
Fix gradient autoscheduler's usage of output estimates

BachiLi:
There is a bug in the gradient autoscheduler, that makes it taking the wrong output size estimates since it assumed the estimates have the same order as the output arguments. This PR fixes it.

Not sure how to test this. @steven-johnson any idea?


BachiLi:
OK. Added a test.

BachiLi:
I'm ready to merge this, we just need to trigger the buildbots.

steven-johnson:
arm failure is unrelated

PR #4402, opened 2019-11-13
Calculate Expr bounds using function_value_bounds

### LINKED ISSUE #4397: Incorrect Expr bounds calculated in FunctionDAG by Adams2019 autoscheduler
https://github.com/halide/Halide/blob/a47d58e99724f282031655d5295fc9b904f565c0/apps/autoscheduler/FunctionDAG.cpp#L648

The code referenced above tries to calculate bounds of an expression for a given scope with rvar bounds, within FunctionDAG.cpp. It doesn't take into account function value bounds and fails if a Func is used as an expression. Concretely,
```
void generate() {
        input_clamped(x, y) = clamp(input(x, y), 0, 255);
        RDom my_r(0, input.width(), 0, input.height());
        hist(i) = cast<uint32_t>(0);
        hist(input_clamped(my_r.x, my_r.y)) += cast<uint32_t>(1);
        output(i) = hist(i);
    }
```
In the above code, the autoscheduler complains that region of hist is unbounded, though it is clearly bounded as input_clamped is clamped between (0, 255). Compilation aborts with the following error --
```
Condition failed: in.is_bounded():
Region computed of hist is unbounded: [((void *))neg_inf ((void *))pos_inf]
```
This is confirmed by simply replacing input_clamped reference inside hist update definition by "clamp(input(....))", in which case the autoscheduler is able to calculate the correct bounds for region of hist.

By adding function value bounds while calculating the bounds_of_expr_in_scope (tried locally), autoscheduler correctly calculates the bounds for the region to be computed. The problem is func_value_bounds are computed after this step by the FunctionDAG and are not available at this point. See --
https://github.com/halide/Halide/blob/a47d58e99724f282031655d5295fc9b904f565c0/apps/autoscheduler/FunctionDAG.cpp#L881-L885

Is there any reason they are calculated here (the above point being inside of a loop), as opposed to only once somewhere outside the loop? If the computation of func_value_bounds is hoisted out of the loop, it can be made available while calculating the bounds of expressions.

shubhamp-ca:
This fixes #4397 where FunctionDAG.cpp computes an unbounded
region for the update definitions of a Func when the region for
the consumer Func is defined by another Func (call f2), because
the bounds for f2 are not taken into account while calculating
the bounds of expr which defines a dimension of the consumer.

This change --
	1. func_value_bounds are calculated earlier as required by bounds_of_expr...()
	2. uses the computed func_value_bounds while calculating
	   bounds_of_expr_in_scope()
        3. Adds a test to catch the issue #4397


abadams:
There's a problem with using apply_param_estimates there. The autoscheduler needs to know if things are constant or not in order to know if they can be unrolled. If you apply param estimates, it replaces variable parameters with constants, so the autoscheduler incorrectly assumes that things that are runtime parameters are actually constant, and fails on loop unrolling. This is an existing bug, but I worry this PR makes it harder to fix by letting the bad constant bounds creep into more places, as scalar parameters often show up in FuncValueBounds. Maybe see what goes wrong if you remove apply_param_estimates, and try to fix it by applying them as late as possible?

shubhamp-ca:
> There's a problem with using apply_param_estimates there. The autoscheduler needs to know if things are constant or not in order to know if they can be unrolled. If you apply param estimates, it replaces variable parameters with constants, so the autoscheduler incorrectly assumes that things that are runtime parameters are actually constant, and fails on loop unrolling. This is an existing bug, but I worry this PR makes it harder to fix by letting the bad constant bounds creep into more places, as scalar parameters often show up in FuncValueBounds. Maybe see what goes wrong if you remove apply_param_estimates, and try to fix it by applying them as late as possible?

Thanks for that info. Realized applying param estimates is not needed for issue #4397 as it relates to the correctness of function bounds, independent of the estimates. Pushed a change which doesn't apply_param_estimates early. Also, not hoisting computation of func_value_bounds out of the loop, but only calculating earlier as required by bounds_expr_check. I realized hoisting it above the loop meant that it is calculated even if there are no update definitions (only place func_value_bounds are used) and hence possibly unused. Though, calculating inside the loop (as is done currently) will lead to redundant computations if multiple updates are present.

Have also added a test for this as it wasn't caught before.

steven-johnson:
(you should probably re-sync to master to fore a rebuild)

shubhamp-ca:
> (you should probably re-sync to master to force a rebuild)

@steven-johnson do you mean merge master into the branch? If yes, is this the required workflow to follow for any future PRs? (merge Halide/master into feature branch in forked repo)

steven-johnson:
Yes

steven-johnson:
The cloned build is green; this should probably await review by @abadams as it's code I'm not too familiar with.

shubhamp-ca:
> The cloned build is green; this should probably await review by @abadams as it's code I'm not too familiar with.

@steven-johnson Is there a way to trigger the build-bot tests without the extra effort of cloning from your side? I thought merging of master into the branch would have done it. Am I missing something here?

steven-johnson:
buildbot clone is green, LGTM

steven-johnson:
Pretty sure this is just waiting on approval from @abadams

PR #4404, opened 2019-11-14
Adds the ability to pass extra build options for OpenCL kernels.

vksnk:
It is controlled by HL_OCL_BUILD_OPTIONS env. variable right now, would it be useful to add similar API call like halide_opencl_set_build_options as well?

Also, fixed typo.


src/runtime/opencl.cpp
```diff
@@ -593,6 +593,11 @@ WEAK int halide_opencl_initialize_kernels(void *user_context, void **state_ptr,
         options << "-D MAX_CONSTANT_BUFFER_SIZE=" << max_constant_buffer_size
                 << " -D MAX_CONSTANT_ARGS=" << max_constant_args;
 
+        const char *extra_options = getenv("HL_OCL_BUILD_OPTIONS");
```
zvookin: It would probably be best to add an overridable runtime routine like ```halide_opencl_get_device_type``` here.

steven-johnson:
In general, using env variables to do things like this is probably not a good idea; they are a very coarse global setting (what if you need different options at the same time?) and support varies by system/OS. 

What are the options we want to add here? I'd *much* prefer to add an explicit API of some sort than to add any new env var usage.

zvookin:
We're already using env vars for similar so as much as I dislike it, it seems like the right tool for the purpose. It's mostly for performance experiments and such.

steven-johnson:
> We're already using env vars for similar so as much as I dislike it, it seems like the right tool for the purpose. It's mostly for performance experiments and such.

I will bet you US$1 that the existing similar env vars are being used for non-experimental purposes, and this one will be, too.

vksnk:
Thanks for the comments.

> What are the options we want to add here? I'd _much_ prefer to add an explicit API of some sort than to add any new env var usage.
You can have some additional options to enable/disable various optimizations, change floating-point behavior, etc (see some of the default ones here: https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clBuildProgram.html, but there are going to be more, which are vendor specific).

I think it's quite useful to have option to be able to try it, but wasn't sure it's THAT useful to justify an API for this, hence the question in PR description. Although, it seems like everyone is in favor of it, so  I'll add runtime function for this.

vksnk:
I've added runtime functions to set the build options, please take a look.

src/runtime/opencl.cpp
```diff
@@ -130,6 +134,24 @@ WEAK const char *halide_opencl_get_device_type(void *user_context) {
     return device_type;
 }
 
+WEAK void halide_opencl_set_build_options(const char *n) {
+    if (n) {
+        strncpy(build_options, n, 255);
```
steven-johnson: I know we're already using strncpy() somewhere else, but we probably shouldn't, since it has a bad failure mode (it can leave the destination not-null-terminated if the src string is too long).

src/runtime/opencl.cpp
```diff
@@ -130,6 +134,24 @@ WEAK const char *halide_opencl_get_device_type(void *user_context) {
     return device_type;
 }
 
+WEAK void halide_opencl_set_build_options(const char *n) {
+    if (n) {
```
steven-johnson: Why don't we need to acquire build_options_lock here?

src/runtime/opencl.cpp
vksnk: I've added explicit check to make it null-terminated here and in other places in this file.

src/runtime/opencl.cpp
vksnk: I think we do, added here and in other two existing functions.

vksnk:
Thanks for the review, I've addressed comments, PTAL.

src/runtime/opencl.cpp
```diff
@@ -84,52 +84,116 @@ WEAK char device_type[256];
...
-extern "C" {
-
-WEAK void halide_opencl_set_platform_name(const char *n) {
+namespace {
+void halide_opencl_set_platform_name_internal(const char *n) {
     if (n) {
-        strncpy(platform_name, n, 255);
+        size_t buffer_size = sizeof(platform_name) / sizeof(platform_name[0]);
+        strncpy(platform_name, n, buffer_size);
+        if (strlen(n) > buffer_size - 1) {
```
steven-johnson: nit: strncpy() actually zeros out the part of the buffer that it doesn't copy into, so I suspect it would be safe to just do `platform_name[buffer_size - 1] = 0` without the if

steven-johnson:
LGTM, but we need to add a test for it.

vksnk:
> I presume you added this just to make things compile -- you can probably just remove this entirely if you move the test to test/correctness

Yeah, that's right. The thing is that tests from correctness folder don't get runtime.a generated and linked, this tripped me in the beginning (and I got lost in Makefile). That was easy to fix though, sorry about that! 

It's ready for review now, could you please take a look?

steven-johnson:
LGTM but the CMake build might need attention too (to ensure runtime.a is linked)

PR #4406, opened 2019-11-14
Update PyTorch support to work with v1.3

steven-johnson:
attn @mgharbi


src/runtime/HalidePyTorchHelpers.h
```diff
@@ -48,6 +48,31 @@ inline void check_type(at::Tensor &tensor) {
     AT_ERROR("Scalar type ", tensor.scalar_type(), " not handled by Halide's PyTorch wrapper");
 }
 
+// TODO: if PyTorch exposes any variable with the API version,
+// I haven't found it in source or documentation; for now, we'll sniff
+// this macro's existence to infer that we are building with v1.3+ (vs 1.2)
+#ifdef AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS
```
shoaibkamil: This is unfortunate but seems unavoidable.

src/runtime/HalidePyTorchHelpers.h
```diff
@@ -59,12 +84,18 @@ AT_FORALL_SCALAR_TYPES_WITH_COMPLEX(HL_PT_DEFINE_TYPECHECK);
 
 #undef HL_PT_DEFINE_TYPECHECK
 
+#endif
```
shoaibkamil: Nit: It'd be nice to see what this `#endif` corresponds to.

shoaibkamil:
Minor nit, otherwise LGTM

src/runtime/HalidePyTorchHelpers.h
steven-johnson: Done

src/runtime/HalidePyTorchHelpers.h
mgharbi: I don't know of a better solution, either.

PR #4407, opened 2019-11-14
Add Buffer::copy_to_interleaved() and Buffer::copy_to_planar()

steven-johnson:
I've found lots of code reinventing this wheel inside Google (e.g., tests that load interleaved images and need to make a planar copy to run a particular Halide filter on it); while this is purely syntactic sugar on top of existing functionality, it does handle some cases that are easy to miss (e.g. nonzero mins), so I think having these as canonical functions is worthwhile.


steven-johnson:
arm failure is unrelated

PR #4408, opened 2019-11-15
Bugfix: Empty device buffer on corner case

timo-zeiss-research:
**Scenario**
1. Buffer is created and filled with data.
2. CUDA is **later** added to the JIT target ```set_feature(Halide::Target::CUDA)```
3. Use buffer on the device for computation

**Observation**
Buffer is not copied to Device before use.

**Solution**
Mark buffer as ```set_host_dirty``` when device memory is allocated.

**Discussion**
In the corner case, when a buffer is created and filled on the host before CUDA is added to the JIT target, the buffer is not marked as host_dirty and thus not copied on first use. Our proposed fix is to mark the host as dirty on device memory allocation.  This one-line change fixes the issue for our use cases. Is there additional tests we should do to check is validity?

### DIFF
--- a/src/Lower.cpp
+++ b/src/Lower.cpp
@@ -294,6 +294,13 @@ Module lower(const vector<Function> &output_funcs,
         s = select_gpu_api(s, t);
         debug(2) << "Lowering after selecting a GPU API for extern stages:\n"
                  << s << "\n\n";
+    } else {
+        // Always mark buffers host dirty. Buffers will otherwise not be correctly copied for
+        // other pipelines with device feature enabled.
+        debug(1) << "Injecting host <-> dev buffer copies...\n";
+        s = inject_host_dev_buffer_copies(s, t);
+        debug(2) << "Lowering after injecting host <-> dev buffer copies:\n"
+                    << s << "\n\n";
     }
 
     if (t.has_feature(Target::OpenGL)) {


steven-johnson:
LGTM so far (cloned as https://github.com/halide/Halide/pull/4409 for testing); is there a way to add a test to verify the fix?

Also, do other device backends need a similar fix?

abadams:
Won't this cause copies to device for fresh buffers though? That seems like a massive performance hit. I think the better solution is to mark all outputs as host_dirty, even in pipelines where the target doesn't contain a GPU feature.

timo-zeiss-research:
Thanks a lot for your feedback. @abadams I agree that the performance hit from always copy is unacceptable. I updated the pull request replacing my previous commit with a commit following your advice. Please let me know if you agree on the implementation.

abadams:
Yes, I think this is the right solution

steven-johnson:
FYI, this seems to have made it possible to have a buffer with both host-dirty and device-dirty bits set; this has caused some internal-to-Google code to fail. I'd argue that it *should* be illegal to have both set.

timo-zeiss-research:
We have observed the same on our end when fill() was called to clear a temp buffer that has not been copied. Our solution was to explicitly set both host and device dirty flags to false before fill() outside of Halide. Not sure if that is the only case, but that might actually also be done in side Halide as a follow up change?

steven-johnson:
See https://github.com/halide/Halide/issues/4431 for followup.

steven-johnson:
As a followup:

(1) The first case I found this in I now have definitive evidence of Bad Code on the callers part: they were calling a `.define_extern()` function with `DeviceAPI::Hexagon`, but the underlying function being called could sometimes actually call a CPU-based Halide function. This got the injected copy code confused and led to this situation.

(2) I think there still might be an issue here, in that the newly-added `else` clause should probably deliberately exclude `t.arch == Hexagon`; I'm seeing a case where we (apparently) get `copy_to_device()` calls injected into the HVX codegen; this causes all sorts of amusing failure modes as the hexagon backend tries to make calls that can't be satisfied on that side. I'm going to prep a PR to fix that.

PR #4412, opened 2019-11-19
Add option to create gradient-descent variant of a Generator

steven-johnson:
This adds a new flag to Generator main (-d) that will look at the inputs and outputs of the Generator and produce a new pipeline (via propagate_adjoint) that is suitable for using as the gradient-descent stage for PyTorch, TensorFlow, etc.

A few notes:
- obviously this is lightly tested, but the heavy lifting is done by Derivative, which is in pretty good shape.
- doesn't support inputs or outputs with Tuple values; we probably could make this work with more effort, but since (AFAICT) neither PT nor TF allow tuple-valued Tensors, it's of dubious value to support
- In general the nomenclature here is dubious and may need rethinking
- Updated apps/HelloPyTorch to use this (in conjunction with the manually-written variant). Note to reviewers, there's some horrible code in the PyTorch support that needs to be improved, but it's not clear to me how to do so within the confines of PyTorch limitations; I've reached out to @mgharbi for suggestions.
- Note that the synthesized gradient-descent pipeline requests autoscheduling, but be aware that this is defaulting to the **classic** autoscheduler; what we really want is the gradient autoscheduler, but that currently isn't packaged in a way to make it easy to use outside of apps/gradient_autoscheduler. (My intent is to rectify that once the long-awaited autoscheduler refactoring is done.)

attn @mgharbi

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1507,8 +1507,8 @@ Module GeneratorBase::build_gradient_module(const std::string &function_name) {
             // output_name is something like "funcname_i"
             const std::string grad_in_name = replace_all(grad_input_pattern, "$OUT$", output_name);
             // TODO(srj): does it make sense for gradient to be a non-float type?
-            // For now, assume it's always float32 (unless the output was float64).
-            const Type grad_in_type = output->type() == Float(64) ? Float(64) : Float(32);
+            // For now, assume it's always float32 (unless the output is already some float).
+            const Type grad_in_type = output->type().is_float() ? output->type() : Float(32);
             const int grad_in_dimensions = f.dimensions();
             const ArgumentEstimates grad_in_estimates = f.output_buffer().parameter().get_argument_estimates();
             internal_assert((int) grad_in_estimates.buffer_estimates.size() == grad_in_dimensions);
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1546,10 +1546,27 @@ Module GeneratorBase::build_gradient_module(const std::string &function_name) {
                 const auto &f = input->funcs_[i];
                 const auto &p = input->parameters_[i];
 
-                const std::string grad_out_name = replace_all(replace_all(grad_output_pattern, "$OUT$", output_name), "$IN$", input_name);
+                Func d_f = d(f);
+
+                std::string grad_out_name = replace_all(replace_all(grad_output_pattern, "$OUT$", output_name), "$IN$", input_name);
+                if (!d_f.defined()) {
+                    grad_out_name = "_dummy" + grad_out_name;
+                }
 
                 Func d_out_wrt_in(grad_out_name);
-                d_out_wrt_in(Halide::_) = d(f)(Halide::_);
+                if (d_f.defined()) {
+                    d_out_wrt_in(Halide::_) = d_f(Halide::_);
+                } else {
+                    debug(DBG) << "    No Derivative found for output " << output_name << " wrt input " << input_name << "\n";
+                    // If there was no Derivative found, don't skip the output;
+                    // just replace with a dummy Func that is all zeros. This ensures
+                    // that the signature of the Pipeline we produce is always predictable.
+                    std::vector<Var> vars;
+                    for (int i = 0; i < d_output.dimensions(); i++) {
+                        vars.push_back(Var::implicit(i));
+                    }
+                    d_out_wrt_in(vars) = make_zero(d_output.type());
+                }
 
                 std::vector<std::pair<Expr, Expr>> est_pairs;
                 for (const auto e : p.get_argument_estimates().buffer_estimates) {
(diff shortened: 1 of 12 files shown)


BachiLi:
This may have issues with integer Buffer inputs and so on, and this can be important, e.g., when the generator takes an integer buffer for indexing the values. I'm not sure what is the best way to handle this. Currently, I think you won't find the corresponding Funcs for the integer buffers from propagate_adjoint and this might crash with assertion. For PyTorch & Tensorflow they expect the users to return None for those inputs.

src/Generator.cpp
```diff
@@ -1429,6 +1441,156 @@ Module GeneratorBase::build_module(const std::string &function_name,
...
+        const Func &original_output = original_outputs.at(i);
+        const ImageParam &d_output = d_output_imageparams.at(i);
+        std::vector<std::pair<Expr, Expr>> bounds;
+        for (const auto &e : d_output.parameter().get_argument_estimates().buffer_estimates) {
+            user_assert(e.min.defined() && e.extent.defined())
+                << "build_gradient_module: you must provide complete estimates for all inputs and outputs";
+            bounds.emplace_back(e.min, e.min + e.extent - 1);
+        }
+        Func adjoint_func = BoundaryConditions::constant_exterior(d_output, make_zero(d_output.type()));
+        Derivative d = propagate_adjoints(original_output, adjoint_func, bounds);
```
BachiLi: I think you can simply do
`Derivative d = propagate_adjoints(original_output, d_output);`
The boundary condition is automatically added.

BachiLi: Actually no since d_output is an ImageParam...never mind.

BachiLi: This can generate incorrect gradients if the estimates are different from the actual output size. You should use the actual bounds, i.e., `d_output.dim(i).min()` and `d_output.dim(i).max()`.

steven-johnson:
> This may have issues with integer Buffer inputs and so on, and this can be important, e.g., when the generator takes an integer buffer for indexing the values. I'm not sure what is the best way to handle this. Currently, I think you won't find the corresponding Funcs for the integer buffers from propagate_adjoint and this might crash with assertion. For PyTorch & Tensorflow they expect the users to return None for those inputs.

good point, I'll add a test case and ensure we can handle it gracefully somehow.

src/Generator.cpp
steven-johnson: done

src/Generator.cpp
```diff
@@ -1429,6 +1441,156 @@ Module GeneratorBase::build_module(const std::string &function_name,
...
+    std::vector<ImageParam> d_output_imageparams;
+    for (const auto *output : pi.outputs()) {
+        for (size_t i = 0; i < output->funcs().size(); ++i) {
+            const Func &f = output->funcs()[i];
+            const std::string output_name = output->array_name(i);
+            // output_name is something like "funcname_i"
+            const std::string grad_in_name = replace_all(grad_input_pattern, "$OUT$", output_name);
+            // TODO(srj): does it make sense for gradient to be a non-float type?
+            // For now, assume it's always float32 (unless the output was float64).
+            const Type grad_in_type = output->type() == Float(64) ? Float(64) : Float(32);
```
BachiLi: We also support Float(16) if that matters.

steven-johnson: done

steven-johnson:
> good point, I'll add a test case and ensure we can handle it gracefully somehow.

hm, does propagate_adjoint() have a way to gracefully report "there is no well-defined derivative"? If not, we definitely need to add one.

BachiLi:
> > good point, I'll add a test case and ensure we can handle it gracefully somehow.
> 
> hm, does propagate_adjoint() have a way to gracefully report "there is no well-defined derivative"? If not, we definitely need to add one.

Not yet. I guess instead of the binary "found the corresponding Func" and "couldn't find it", we should add the third one "the Func's derivative is not well defined" (because it's type is an integer).

steven-johnson:
> Not yet. I guess instead of the binary "found the corresponding Func" and "couldn't find it", we should add the third one "the Func's derivative is not well defined" (because it's type is an integer).

I'm going to modify Derivative to just have it return an undefined Func when not found, and use that for now; I suspect you'd be better suited to making a clean three-way result to distinguish between not-found and not-differentiable.

steven-johnson:
That last commit adds test cases and the func-as-undefined change I mentioned. It may need more work, but take a look at it and see if it looks plausible.

BachiLi:
> That last commit adds test cases and the func-as-undefined change I mentioned. It may need more work, but take a look at it and see if it looks plausible.

LGTM!

steven-johnson:
OK, I will probably go ahead and land this (pending approval from @zalman) as is; it's probably worth adding logic to Derivative to allow distinguishing the two possible not-found conditions, but I think that can be done later.

PR #4415, opened 2019-11-19
Use std::abs instead of abs to prevent ambigious match

ashishUthama:


steven-johnson:
LGTM pending green tests

steven-johnson:
Thanks for the fix!

PR #4420, opened 2019-11-20
Small code simplifications at the beginning of the FunctionDAG constructor

benoitsteiner:
Avoid setting node.func in two separate locations. The second initialization is redundant and confusing
 Streamlined the initialization of node.stages by avoiding extra stage creation and movements.

Also applied clang-format, which accounts for all the other code changes

### DIFF
--- a/apps/autoscheduler/FunctionDAG.cpp
+++ b/apps/autoscheduler/FunctionDAG.cpp
@@ -6,7 +6,9 @@ namespace Halide {
 namespace Internal {
 
 template<>
-RefCount &ref_count<Autoscheduler::BoundContents>(const Autoscheduler::BoundContents *t) noexcept {return t->ref_count;}
+RefCount &ref_count<Autoscheduler::BoundContents>(const Autoscheduler::BoundContents *t) noexcept {
+    return t->ref_count;
+}
 
 template<>
 void destroy<Autoscheduler::BoundContents>(const Autoscheduler::BoundContents *t) {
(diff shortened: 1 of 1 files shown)


apps/autoscheduler/FunctionDAG.cpp
```diff
@@ -330,8 +333,10 @@ void BoundContents::validate() const {
         if (p.max() < p.min()) {
             aslog(0) << "Bad bounds object:\n";
             for (int j = 0; j < layout->total_size; j++) {
-                if (i == j) aslog(0) << "=> ";
-                else aslog(0) << "   ";
+                if (i == j)
```
steven-johnson: use {} even for single-line ifelse bodies

apps/autoscheduler/FunctionDAG.cpp
```diff
@@ -588,11 +594,9 @@ FunctionDAG::FunctionDAG(const vector<Function> &outputs, const MachineParams &p
     int stage_count = 0;
 
     for (size_t i = order.size(); i > 0; i--) {
-        Function consumer = env[order[i-1]];
-
         Node &node = nodes[order.size() - i];
+        Function consumer = node.func;
```
steven-johnson: I presume this is intentionally changing from the i'th-from-the-end to always-at-the-end?

apps/autoscheduler/FunctionDAG.cpp
```diff
@@ -983,7 +988,8 @@ void FunctionDAG::featurize() {
 
             // Pick a dimension to vectorize over - the innermost pure loop
             size_t vector_dim = 0;
-            while (vector_dim < stage.loop.size() && !stage.loop[vector_dim].pure) vector_dim++;
+            while (vector_dim < stage.loop.size() && !stage.loop[vector_dim].pure)
```
steven-johnson: add {} around single-line while bodies

apps/autoscheduler/FunctionDAG.cpp
benoitsteiner: The loop starting at line 586 ensures that nodes[order.size() - i] and env[order[i-1]] refer to the same function. I can therefore use the function assigned to the node line 588 to reduce the mental load needed to understand this part of the code.

steven-johnson: ahhh, got it

PR #4426, opened 2019-11-21
Fix lld::wasm::link() call for trunk LLVM

steven-johnson:


steven-johnson:
(Don't merge this just yet; I am now seeking a little more info on this from an LLVM engineer.)

PR #4427, opened 2019-11-22
Revert "Bugfix: Empty device buffer on corner case"

steven-johnson:
Reverts halide/Halide#4408 -- unfortunately, this can get us into a situation where a buffer can get both host-dirty and device-dirty bits set; I don't have the exact scenario but it involves complex `define_extern` usage that calls from CPU->Device or vice versa. For now we need to roll this back as it's breakage some obscure code.


steven-johnson:
After further review and assistance from Dillon, we want to keep this; see https://github.com/halide/Halide/issues/4431 for followup.

PR #4428, opened 2019-11-22
Improved generation of Pipeline features for extern functions

benoitsteiner:
Leverage the proxy (if any) to generate pipeline features for extern functions. There's still work to do to model the memory accesses, but we can already have a sense of how much computation is done by the black box (assuming that the proxy is accurate), which should help the auto-scheduler further down the road. Note that this PR doesn't address the generation of schedule features, which will be handled in a follow up PR.

### DIFF
--- a/apps/autoscheduler/CMakeLists.txt
+++ b/apps/autoscheduler/CMakeLists.txt
@@ -153,7 +153,7 @@ endif()
 
 add_executable(test_perfect_hash_map test_perfect_hash_map.cpp)
 
-add_executable(test_function_dag test_function_dag.cpp FunctionDAG.h FunctionDAG.cpp ASLog.h ASLog.cpp)
+add_executable(test_function_dag test_function_dag.cpp FunctionDAG.cpp ASLog.cpp)
 target_include_directories(
   test_function_dag
   PRIVATE "${HALIDE_INCLUDE_DIR}" "${HALIDE_TOOLS_DIR}")
--- a/apps/autoscheduler/CMakeLists.txt
+++ b/apps/autoscheduler/CMakeLists.txt
@@ -153,14 +153,14 @@ endif()
 
 add_executable(test_perfect_hash_map test_perfect_hash_map.cpp)
 
-add_executable(test_function_dag test_function_dag.cpp)
+add_executable(test_function_dag test_function_dag.cpp FunctionDAG.h FunctionDAG.cpp ASLog.h ASLog.cpp)
 target_include_directories(
   test_function_dag
   PRIVATE "${HALIDE_INCLUDE_DIR}" "${HALIDE_TOOLS_DIR}")
-target_link_libraries(test_apps_autoscheduler PRIVATE Halide)
-if (NOT WIN32)
-  target_link_libraries(test_apps_autoscheduler PRIVATE dl)
-endif()
+target_link_libraries(test_function_dag PRIVATE Halide)
+#if (NOT WIN32)
+#  target_link_libraries(test_function_dag PRIVATE dl)
+#endif()
 
 add_executable(weightsdir_to_weightsfile weightsdir_to_weightsfile.cpp
                Weights.cpp)
--- a/apps/autoscheduler/CMakeLists.txt
+++ b/apps/autoscheduler/CMakeLists.txt
@@ -153,5 +153,14 @@ endif()
 
 add_executable(test_perfect_hash_map test_perfect_hash_map.cpp)
 
+add_executable(test_function_dag test_function_dag.cpp)
+target_include_directories(
+  test_function_dag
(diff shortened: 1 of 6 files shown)


apps/autoscheduler/FunctionDAG.cpp
```diff
@@ -996,66 +996,83 @@ void FunctionDAG::featurize() {
...
-            for (auto v : def.values()) {
-                featurizer.visit_store_args(node.func.name(), v.type(), def.args());
-                v = common_subexpression_elimination(simplify(v));  // Get things into canonical form
-                v.accept(&featurizer);
-            }
-            for (auto v : def.args()) {
-                v = common_subexpression_elimination(simplify(v));  // Get things into canonical form
+            if (node.func.extern_definition_proxy_expr().get()) {
+                // Extern function call with a proxy implementation specified: generate the featurization from the proxy
+                Expr v = common_subexpression_elimination(simplify(node.func.extern_definition_proxy_expr()));  // Get things into canonical form
```
steven-johnson: Nit: this line seems awfully long.

apps/autoscheduler/Makefile
```diff
@@ -38,6 +38,10 @@ $(BIN)/test_perfect_hash_map: test_perfect_hash_map.cpp PerfectHashMap.h
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $< -o $@
 
+$(BIN)/test_function_dag: test_function_dag.cpp FunctionDAG.h FunctionDAG.cpp ASLog.h ASLog.cpp
```
steven-johnson: don't forget to update the CMakeLists.txt too

apps/autoscheduler/test_function_dag.cpp
```diff
@@ -0,0 +1,86 @@
+#include "FunctionDAG.h"
+#include "Halide.h"
+#include <sstream>
+
+using namespace Halide;
+
+extern "C" int generateOutputVals(
```
steven-johnson: generate_output_vals

benoitsteiner: fixed.

apps/autoscheduler/FunctionDAG.cpp
benoitsteiner: For some reason clang-format won't improve the formatting. I manually split the code.

apps/autoscheduler/Makefile
benoitsteiner: Done

apps/autoscheduler/CMakeLists.txt
```diff
@@ -153,5 +153,14 @@ endif()
 
 add_executable(test_perfect_hash_map test_perfect_hash_map.cpp)
 
+add_executable(test_function_dag test_function_dag.cpp FunctionDAG.h FunctionDAG.cpp ASLog.h ASLog.cpp)
+target_include_directories(
+  test_function_dag
+  PRIVATE "${HALIDE_INCLUDE_DIR}" "${HALIDE_TOOLS_DIR}")
+target_link_libraries(test_function_dag PRIVATE Halide)
+#if (NOT WIN32)
```
steven-johnson: Either remove entirely or add comment about why it should stay commented out

steven-johnson:
In the future, may I suggest doing the clang-format step as a separate PR; it makes the meat of the real changes harder to review

apps/autoscheduler/CMakeLists.txt
benoitsteiner: I deleted the code.

benoitsteiner:
Understood. I'll refrain from using clang-format on code that doesn't already follow the formatting standard.

steven-johnson:
> Understood. I'll refrain from using clang-format on code that doesn't already follow the formatting standard.

Oh, please fix the formatting, just consider opening up a *separate* PR to do that with no other changes :-)

(I'm surprised we still have files that haven't been thru the formatter, but perhaps the giant formatting party didn't include apps/?)

apps/autoscheduler/Makefile
```diff
@@ -38,6 +38,10 @@ $(BIN)/test_perfect_hash_map: test_perfect_hash_map.cpp PerfectHashMap.h
...
+	@mkdir -p $(@D)
+	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) $^ -o $@ $(LDFLAGS) $(LIB_HALIDE) $(HALIDE_SYSTEM_LIBS)
```
steven-johnson: This should be `$(filter-out %.h,$^)` instead of `$^`

benoitsteiner: Thanks for the tip, done

PR #4429, opened 2019-11-22
FreeBSD support for halide

Martinfx:


Martinfx:
Hi @yurivict , do you have some idea ?

yurivict:
It still fails with the same way these patches.

yurivict:
What llvm version are you able to build with?
In my case it's llvm90 from ports that is used.

Martinfx:
@yurivict i am using 
```
-- Found LLVM 9.0.0
-- Using LLVMConfig.cmake in: /usr/local/llvm90/lib/cmake/llvm
-- Using llvm-as at /usr/local/llvm90/bin/llvm-as
-- Using llvm-nm at /usr/local/llvm90/bin/llvm-nm
-- Using clang at /usr/local/llvm90/bin/clang
-- Building Halide as a shared library
-- WARNINGS_AS_ERRORS enabled
-- Found OpenMP
-- Building tests enabled
-- Internal tests enabled
-- Building apps enabled
-- linear_algebra: No CBLAS header, skipping CBLAS tests
-- linear_algebra: Eigen3 Missing, skipping Eigen3 benchmarks
-- linear_algebra: No CBLAS header, skipping BLAS benchmarks
-- Building tutorials enabled
-- Building utils enabled
-- Configuring done
-- Generating done
-- Build files have been written to: /home/maxfx/Documents/Halide

```

yurivict:
I shared my version of the port here: https://people.freebsd.org/~yuri/halide.shar that fails.

Martinfx:
@yurivict if you disable in cmake test and apps and next stuff compilation is good without bugs.

yurivict:
@Martinfx I've updated the shar at the same link with the options but the build still fails with thee options disabled.

Martinfx:
@yurivict this changes you can try https://github.com/Martinfx/FreeBSD-Ports/commit/51de259c2a48bb4fff8ed7438736a526807ec78c

yurivict:
This version of the port still has the same build-time error:
> Internal Error at /usr/ports/lang/halide/work/Halide-de12425a/src/CodeGen_LLVM.cpp:610 triggered by user code at : Condition failed: semaphore_t_type: Did not find halide_semaphore_t in initial module
Child aborted

abadams:
If we merge support for FreeBSD, we're going to need to be clearer about how well tested things are here and what the level of support is. I don't think it's feasible for us to add FreeBSD build bots to the CI and the core devs wouldn't be able to fix things if they broke anyway, or be able to help people if problems arise. On the other hand, we also don't have android and ios build bots, but we do have substantial production use of those targets so we have confidence in those backends for that reason. I'm not sure what the QuRT situation is. It's possible Qualcomm tests that themselves. We should perhaps add a section to the readme somewhere describing what the testing and production usage story is for each CPU/GPU target. I'll ping some people for their thoughts @zvookin @steven-johnson @shoaibkamil @slomp @pranavb-ca. Maybe just a table showing which targets are tested on every PR and which have serious production usage?

yurivict:
Just being able to build it on FreeBSD would be a good first step.

steven-johnson:
It's a multifaceted problem... we should really try to add proper iOS + Android test coverage in the public buildbots too (though even basic coverage on our buildbots has been a problem recently), but they do get a lot of non-public testing elsewhere.

I'd be willing to accept changes that enable FreeBSD support, but with the caveat that they are likely to go untested on mainline buildbots anytime soon, thus breakage *will* occur sometimes. (This is probably better than trying to maintain patchsets, I'm guessing)

steven-johnson:
To add to Andrew's list of stuff-not-tested-enough(-or-at-all) is the WebAssembly backend. (I've had it on my TODO list for a while, but have been reluctant to add it to our buildbots until we get additional machines, as the Linux boxes are overtaxed as it is.)

pranavb-ca:
@abadams - Yes, we do have in-house QuRT testing (It has been down for a couple of reasons for a little bit but we are working on bringing our bot back up). FWIW, we recently added a line to the README saying Hexagon is supported only by the use LLVM trunk because there are no bots testing Halide with other LLVM branches.

Martinfx:
@abadams, @steven-johnson @pranavb-ca  Thank you for your response. I would like make first step how said @yurivict

yurivict:
Somehow, the type ```struct.halide_semaphore_t``` defined in src/runtime/HalideRuntime.h isn't found at https://github.com/halide/Halide/blob/master/src/CodeGen_LLVM.cpp#L626 on FreeBSD.

Martinfx:
Pull requestwith changes here: https://github.com/halide/Halide/pull/4955

Issue #3534, opened 2018-12-26
Recent LLVM trunk builds seem to have broken predicated_store_load


steven-johnson:
Haven't investigated yet, just opening an issue for tracking.

abadams:
I noticed massive performance issues with masked vector store instructions on x86. It's not supposed to be non-temporal anymore, but subsequence loads of an address that you do a masked store to take forever, so it seems like it has left cache. I just turned them off entirely in the autoscheduler branch and was going to propose turning them off in master for x86 for now.

steven-johnson:
> propose turning them off in master for x86 for now

+1, make it so

steven-johnson:
>  I just turned them off entirely in the autoscheduler branch

Did you do that locally? I don't see it in the branch.

steven-johnson:
Looks like we never followed up on this -- probably worth revisiting to see if the 'recent LLVM trunk breakage' is still a thing.

PR #4432, opened 2019-11-25
Bring back predicated store/load

BachiLi:
Predicated store/load was disabled due to a "llvm trunk breakage" (https://github.com/halide/Halide/issues/3534). I tried running the tests on my machine and it runs fine now with the current llvm trunk (also one of my research projects needs predicated store/load), so I propose to revive it.

### DIFF
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -268,10 +268,7 @@ class PredicateLoadStore : public IRMutator {
         } else if (target.arch == Target::X86) {
             // Should only attempt to predicate store/load if the lane size is
             // no less than 4
-            // TODO: disabling for now due to trunk LLVM breakage.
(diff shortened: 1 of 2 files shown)


abadams:
The autoscheduler work also revealed it to be a massive performance hit whenever you do a masked store followed by a reload of the same address. It's like all masked stores are also non-temporal or something. We got a free performance boost just by turning it off in that branch. So we should be very cautious about reintroducing it.

BachiLi:
OK. I will add some performance tests. It seems that I broke atomics, will fix it.

steven-johnson:
Is this still an active PR? No action on it for a while.

BachiLi:
Sorry, didn't find time to do this. Likely won't have time to work on this until Feburary. I'll close it for the moment.

PR #4436, opened 2019-12-02
Tutorial comment fixes from igarcia@mathworks.com

abadams:


abadams:
https://travis-ci.org/halide/Halide/jobs/619742900?utm_medium=notification&utm_source=github_status

Looks like a failure due to the Region PR? Not sure how that slipped through.

PR #4437, opened 2019-12-02
Fix dangling MinAndExtent

abadams:
See #4393

### DIFF
--- a/src/Argument.h
+++ b/src/Argument.h
@@ -19,12 +19,11 @@ struct ArgumentEstimates {
      * For buffer arguments, all should be undefined. */
     Expr scalar_def, scalar_min, scalar_max, scalar_estimate;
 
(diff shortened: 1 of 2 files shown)


steven-johnson:
Looks like the recently-added generator_aot_autograd test is cratering (on Travis at least).

steven-johnson:
Looks like the issue with autograd is that:
- because gradient-scheduler isn't yet easily usable by tests, it's using the 'classic' autoscheduler
- the classic autocheduler is generating a schedule that causes out-of-bounds access due to vectorization, but apparently only under LLVM9

steven-johnson:
Update: it might also be specific to running on an AVX512 host; if I set HL_TARGET to avx2-or-less it works fine, so perhaps the classic autoscheduler has a subtle bug about vectorizing where native-vector-width > 32?

PR #4438, opened 2019-12-02
Increase autograd estimates/sizes from 32 -> 64

steven-johnson:
The classic autoscheduler can over-vectorize (e.g., on Skylake it will vectorize to width=64, which will fail for inputs of width<64). This is a somewhat-known issue; rather than try to fix it here, I am just going to widen the buffers to the autograd test to avoid failure on Skylake+ machines. (Note that autograd shouldn't be using the classic autoscheduler in the first place; it should be using the gradient autoscheduler, but packaging issues make that impractical just yet.)

### DIFF
--- a/test/generator/autograd_aottest.cpp
+++ b/test/generator/autograd_aottest.cpp
@@ -9,7 +9,7 @@
 
 using namespace Halide::Runtime;
 
-const int kSize = 32;
+constexpr int kSize = 64;
 
 int main(int argc, char **argv) {
     int result;
--- a/test/generator/autograd_generator.cpp
+++ b/test/generator/autograd_generator.cpp
@@ -2,6 +2,8 @@
 
 namespace {
 
+constexpr int kSize = 64;
+
 class Autograd : public Halide::Generator<Autograd> {
 public:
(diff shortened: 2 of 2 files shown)


test/generator/autograd_aottest.cpp
```diff
@@ -9,7 +9,7 @@
 
 using namespace Halide::Runtime;
 
-const int kSize = 32;
+constexpr int kSize = 64;
```
abadams: At global scope is constexpr int meaningfully different to const int? Any reason to prefer it?

steven-johnson: Not really, it's just reflexive change at this point. Want me to back it out?

abadams: Was mostly wondering if we should make this change more broadly as a matter of style. The internet seems to suggest that constexpr is very slightly preferable because it guarantees you can use it in other compile-time constructs like static_assert. I think uniformity of style is a good thing, and this seems like mostly a stylistic choice, but I don't care enough to delay a PR over it, so lgtm.

Issue #4423, opened 2019-11-21
RoundUp behavior on integer funcs can cause arithmetic exceptions


alexreinking:
Consider the two pipelines

    output_i(x) = 100 / (27 - x)
    output_f(x) = 100.f / (27.f - x)

Then in either case writing

    output_*.split(x, xo, xi, 8, TailStrategy::RoundUp);

If you then try to realize via:

    output_*.realize(24);

In either case everything is fine. But if you try to realize via:

    output_*.realize(25);

Then `output_i` crashes with a floating point exception (ie. integer division by zero) while `output_f` returns a buffer containing the correct values (because FP div-by-zero doesn't trap).

This behavior is slightly unexpected since the compute bounds don't contain the problematic value. While this particular example is designed to be pathological, here is a program that actually runs into this issue:

```
#include <Halide.h>
using namespace Halide;

class RoundUp : public Generator<RoundUp>
{
public:
    Output<Buffer<int>> output{"output", 1};

    void generate() {
        Func f, g;
        Var x, xo, xi;

        g(x) = 5;
        f(x) = 100 / g(x);
        output(x) = f(x);

        g.compute_root();
        f.compute_root();
        f.split(x, xo, xi, 8, TailStrategy::RoundUp);
    }
};

HALIDE_REGISTER_GENERATOR(RoundUp, round_up)
```

I get:

```
[alex@reinking-arch round_up]$ ./run_round_up output=foo.mat --output_extents=[25]
Floating point exception (core dumped)
```

One fix would be to reject this tail strategy for rules that divide by an integer which contains zero in the range.

alexreinking:
Note that this issue also arises with multiplication / division by -1 when the other value is `INT_MIN`

abadams:
Yeah, I've seen this pop up in practice. I think we just need to define
division by zero as returning zero.

alexreinking:
How will that work in the implementation? Won't that make integer division by an unknown less efficient? JRK and I briefly looked online to see if there was any integer divide by zero that didn't trap and we couldn't find one.

abadams:
Yes, that would make integer division by an unknown that can't statically be proven to be not zero less efficient. Fortunately that doesn't seem to happen much in image processing. Division by non-constant integers is very rare to begin with, and when it happens it usually includes an explicit guard against zero, e.g. x / max(y, 1), so we can use a if (can_prove(denominator != 0)) to skip the extra logic.

[...]

PR #4439, opened 2019-12-02
Define division by zero to return zero

abadams:
Define div/mod by zero to return zero
    
Also define things like -128/-1 to just wrap, for types that wrap.
    
This makes div/mod side-effect-free, which removes the need for quiet_div/quiet_mod, and avoids runtime faults due to scheduling (see #4423 )
    
Not yet ready to merge, as the simplifier rules need to be reverified under this new constraint (there may now be division by zero on the LHS, where previously we occasionally assumed that there was not, because it was UB).

The performance implications of this will be negligible. Division by non-constants in the critical path is rare to begin with, and already scalarizes and takes tens of cycles. This just adds a probably-perfectly-predicted test that the denominator is non-zero a corresponding branch.

Note that division of things that are already guarded against being zero is not affected (e.g. foo/max(bar, 1)), because there's a can_prove that skips the additional logic if the denominator is provably != 0


steven-johnson:
I'm assuming this isn't ready for review yet...

abadams:
It's ready for people to object to defining division in this way, and I opened it early for that purpose, but actual code review can wait.

steven-johnson:
Eh, honestly, I'm not sure, I think my spidey sense is tingling a bit. Defining division this way is useful for various reasons, but I'm hesitating to call it "division". If I were a newcomer to the project, I'd be pretty surprised to have N/0 -> 0, though maybe it wouldn't matter much. It would probably be appropriate to have a section of the documentation that explains exactly why we define it this way, and what the pros and cons of such a choice are.

alexreinking:
> Not yet ready to merge, as the simplifier rules need to be reverified under this new constraint (there may now be division by zero on the LHS, where previously we occasionally assumed that there was not, because it was UB).

This is the only threat to soundness I'm aware of for this. Are there any other trapping expressions? E.g. do we have integer square roots?

abadams:
I can't think of any other trapping expressions other than the require intrinsic, whose whole purpose is to trap. I guess if you require(x >= 0) and then shiftinwards your way to failure then it shouldn't be very mysterious what has happened.

abadams:
The list of incorrect simplifier rules (thanks @jn80842 ) revealed that we actually want x % 0 to return x, not zero, to maintain the euclidean identity. So for the purposes of div and mod, zero as the second argument should be interpreted as infinity. Anything mod infinity is unchanged. Anything divided by infinity is zero. The rules that are broken are any that assume multiplication can be undone by division by the same value. So (x * y) / y is no longer x, as y might be zero. These rules will be changed to only for for constant non-zero y.

alexreinking:
@abadams -- which is the Euclidean Identity?

abadams:
(x/y)*y + x%y = x

alexreinking:
Ah, great, that makes sense. `x / 0 = 0` and `x % 0 = x`.

abadams:
Ugh. x % 0 == x means that foo(x % foo.width()) is no longer a bounded access

steven-johnson:
(needless to say: if/when you get this to a stable point that has consensus, let me run it thru the Google corpus before landing, to see what unexpected issues may crop up there.)

abadams:
This is ready for testing in Google

alexreinking:
@abadams -- so is `x % 0 == 0` again? Do we not need the Euclidean identity?

abadams:
0 <= (x % y) <= max(0, abs(y) - 1) is more important (e.g. for bounds inference) than the euclidean identity is.

steven-johnson:
> This is ready for testing in Google

Underway

steven-johnson:
This is triggering many assertion failures in HVX code in google, e.g.

```
Internal Error at third_party/halide/halide/src/CodeGen_Hexagon.cpp:2684 triggered by user code at : Condition failed: op->condition.type().is_scalar(): select((x128(input2.extent.0) == x128(0)), x128(-1), x128(0))
*** SIGABRT received by PID 8831 (TID 8831) from PID 8831; stack trace: ***
PC: @     0x7f194db25602  (unknown)  raise
    @     0x55f1a4c2a4d1       1728  FailureSignalHandler()
    @     0x7f194dca39a0  964522416  (unknown)
    @     0x55f1a13219f0        112  Halide::Internal::ErrorReport::~ErrorReport()
    @     0x55f1a15394b5        416  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ecff5         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed037         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed037         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14d9afd        224  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a1533382        160  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a15350b3       1728  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a1533382        160  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14dce9c        816  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a1534d48       1728  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a1534d48       1728  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a1533382        160  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f4b24        144  Halide::Internal::CodeGen_LLVM::call_intrin()
    @     0x55f1a1532ea4        400  Halide::Internal::CodeGen_Hexagon::call_intrin()
    @     0x55f1a1534d48       1728  Halide::Internal::CodeGen_Hexagon::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ecff5         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d6179        368  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14f21dd        960  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed0f7         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14edf1f        480  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed0f7         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed0f7         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed0f7         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14edf1f        480  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed0f7         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @     0x55f1a14ed0f7         64  Halide::Internal::CodeGen_LLVM::visit()
    @     0x55f1a14d78f3        320  Halide::Internal::CodeGen_LLVM::codegen()
    @ ... and at least 214 more frames
```

abadams:
Hopefully fixed. eliminate_bool_vectors as a pass doesn't work reliably when codegen has the option of reintroducing them while lowering things.

steven-johnson:
OK, much better now: I get an unexpected "success" from the `error_modulo_constant_zero` test, and tweaky failures from a dozen gcam-related pipelines. Investigating.

steven-johnson:
error_modulo_constant_zero looks like a legit mistake. (are we not running test_errors on the buildbots?)

the others look like teeny tiny golden-image variation of the sort we regularly see from compiler/optimization tweaks; I haven't verified all of them but it looks good so far.

we'll definitely want to wait post-holidays to land this but otherwise is likely LGTM.

abadams:
arm-32 buildbot is failing to configure llvm

steven-johnson:
I will run this thru Google again, just to be sure, but I think we are just about ready to land. (One caveat is that the armbots are still down...)

abadams:
Builds finally ran and are green.

src/Simplify_Select.cpp
```diff
@@ -35,18 +35,7 @@ Expr Simplify::visit(const Select *op, ExprInfo *bounds) {
...
-             // signalling error. Call it UB and assume it can't
-             // happen. The tricky case to consider is:
-             // select(x > 0, a/x, select(x < 0, b/x, indeterminate()))
-             // If we use a signalling error and x > 0, then this will
-             // evaluate indeterminate(), because the top-level select
-             // evaluates both sides.
-
-             rewrite(select(x, y, IRMatcher::Indeterminate()), y) ||
-             rewrite(select(x, IRMatcher::Indeterminate(), y), y))) {
+             false)) {
```
alexreinking: Is the `|| false` necessary?

src/IROperator.h
```diff
@@ -397,7 +414,10 @@ Expr &operator*=(Expr &a, Expr b);
 /** Return the ratio of two expressions, doing any necessary type
  * coercion using \ref Internal::match_types. Note that signed integer
  * division in Halide rounds towards minus infinity, unlike C, which
- * rounds towards zero. */
+ * rounds towards zero. Division by zero returns zero. For types where
+ * overflow is defined behavior, division of the largest negative
+ * signed integer by -1 returns the larged negative signed integer for
+ * the type (i.e. it wraps). */
 Expr operator/(Expr a, Expr b);
```
alexreinking: I would like to see this sort of documentation added to the tutorial since it's such an unusual behavior.

src/Simplify_Select.cpp
abadams: I do this a bunch in the simplifier. It makes it easier to insert, remove, or comment out rewrite rules without worrying about messing up the closing parens and open brace. It's like using a trailing comma at the end of an initializer list.

steven-johnson:
I think this is good to land -- I see no red flags in Google so far.

I agree with @alexreinking that a more-detailed description of this is probably useful (maybe in the tutorials) since it's definitely gonna be a surprising corner case to newcomers.

PR #4442, opened 2019-12-03
Use Halide::Range instead of std::pair<Expr, Expr> in Derivative (Issue #4381)

steven-johnson:
This finishes the work stated in PR#4393, converting Derivative to use Range/Region in its API rather than pair<Expr, Expr>. Note that aside from the structure change, this is also a change from {min,max} to {min,extent}.


BachiLi:
LGTM.

PR #4444, opened 2019-12-03
Smarten Range(min, extent) ctor

steven-johnson:
It's common to pass a literal-0 for min (especially via the RDom ctor), so quietly coerce it to a zero of the same type as extent in these cases, to make migrating old code simpler.

### DIFF
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -391,13 +391,7 @@ struct Range {
     Expr min, extent;
 
     Range() = default;
-    Range(const Expr &min, const Expr &extent)
-        : min(min), extent(extent) {
-        internal_assert(!min.defined() ||
-                        !extent.defined() ||
-                        min.type() == extent.type())
-            << "Region min and extent must have same type\n";
-    }
+    Range(const Expr &min_in, const Expr &extent_in);
 };
 
 /** A multi-dimensional box. The outer product of the elements */
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -2211,4 +2211,18 @@ Expr undef(Type t) {
                                 Internal::Call::PureIntrinsic);
 }
 
+Range::Range(const Expr &min_in, const Expr &extent_in)
+    : min(min_in), extent(extent_in) {
+    // It's common to pass literal-0 for min, so add a little coercion
+    // to ensure that it matches the type of extent in that case, to avoid
+    // noise at the call site.
+    if (Internal::is_zero(min) && extent.defined()) {
+        min = Internal::make_zero(extent.type());
+    }
+    internal_assert(!min.defined() ||
+                    !extent.defined() ||
+                    min.type() == extent.type())
+        << "Region min and extent must have same type\n";
+}
+
 }  // namespace Halide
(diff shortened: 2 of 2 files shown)


abadams:
The way we approached this in IROperator.h was to add overloads that take int instead of an is_const or is_zero check. This means that for example Range(0, some_integer_expr) will work, but Range(const_false(), some_integer_expr) will throw a type error, which seems desirable.

steven-johnson:
> add overloads

We could do that, but we'd have to put similar overload pairs into RDom::initialize_from_region(), BoundaryConditions::collect_region(), etc., since those all just take Exprs. (Or maybe those type could be templatized...)

steven-johnson:
Adding those overload pairs solves that problem, but we still end up with a host of ambiguous-overload issues. It might be simpler to just clean things up at the call sites...

steven-johnson:
Instead of trying to propagate, I just pushed an alternate approach: call `match_types()`, which is what those IROperator cases currently do for the two-Expr case.

abadams:
match_types assumes the existing of int overloads, so it will do things like Range(0, some_uint8) and make both min and extent int32. That's what we're trying to avoid, right?

steven-johnson:
Alas, yes. Let me do some more thinking on this. It might be simplest to bite the bullet and insert explicit casts.

steven-johnson:
While we're on this topic -- what used to happen for mismatched RDom types? Did we normalize them to (say) all ints in the backend somehow? (The only case I'm really seeing this as an issue is in RDom / BoundaryCondition construction, which use Range internally now.)

abadams:
For those uses it doesn't make sense to pass anything but an int, so there's probably just an assert somewhere.

steven-johnson:
> For those uses it doesn't make sense to pass anything but an int

There are a decent number of cases in google3 where Exprs are passed for both min and extent, with different types (which is how I found this in the first place)

steven-johnson:
For instance:
```
  Expr height = image_.height();  // int32
  ...
  Expr maximum_disk_radius = cast<int16_t>(-image2_.dim(0).min());
  ...
  RDom r(-maximum_disk_radius, height + maximum_disk_radius * 2 + 1);
```

abadams:
I see. How about this: Any Expr passed to the RDom or Range constructor that can be losslessly cast to int32 should be, and all others should throw an error.

steven-johnson:
Are 64-bit RDoms/Ranges not supported?

abadams:
No, currently all indices are int32. It's a bit of a problem, but the way out is tricky.

steven-johnson:
OK, that makes it easy then.

steven-johnson:
Heh: would you believe we also have plenty of code that passes floats into `RDom`, and it works fine? e.g.

```
  Halide::Expr kernel_taps_x = ceil(kernel_taps / kernel_scaling_x);
  ...
  Halide::RDom r_x(0, kernel_taps_x);
```

steven-johnson:
Using `lossless_cast()` still causes a surprising amount of failure in the real world; e.g. places that were constructing an RDom from a simple `uint32` expression that can't be proven to fit into an int32. I suspect that this is probably the best we can do (we'll just have to require explicit casts elsewhere).

PR #4445, opened 2019-12-04
Apps from autoscheduler

abadams:
Pull over more apps from the autoscheduler work, and fix their schedules. People like to use harris and unsharp mask as benchmarks, so we should make sure we have properly scheduled implementations of them available. The stuff in tests/auto_schedule had rotted. Also pulled over the iir blur.

This also switches interpolate to a generator. It was nice to have as an example of a jit-based app, but it made inclusion in benchmark suites annoying for research work, and there are plenty of other examples of jit usage.

I also did a pass over the Makefiles making the linker flags for the generators more uniform and more sane. We were passing libHalide.so with its absolute path as a compile input along with the cpp files, instead of using -lHalide.

There are still a few apps to do after this, but this seemed like a nice mergeable unit of work.


CMakeLists.txt
```diff
@@ -265,7 +265,7 @@ if (NOT MSVC)
                       -Wcast-qual
                       -Woverloaded-virtual
                       -Wignored-qualifiers)
-  if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1)
+  if (CMAKE_CXX_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1)
```
abadams: Turns out this was broken if you're unintentionally using clang for c++ and gcc for c like I was.

apps/interpolate/CMakeLists.txt
```diff
@@ -1,4 +0,0 @@
-halide_project(interpolate "apps" interpolate.cpp)
```
steven-johnson: Can we have CMake support for this one too?

abadams: Oops, forgot to commit the new version of it. Done.

apps/interpolate/filter.cpp
```diff
@@ -0,0 +1,40 @@
+#include <cassert>
```
steven-johnson: A low-priority thought for followup work: it would be nice to have a consistent naming convention for the apps; we have variants of `filter.cpp`, `test.cpp`, `run.cpp`, `process.cpp`, and probably others, with no conventions I can see.

abadams: At the dawn of history, bilateral_grid used filter, local_laplacian used process, and blur used test. Apps since then have copied one of the above, resulting in the factions seen today. There was never any good reason for these to be different.

steven-johnson:
mingw needs some love, otherwise LGTM

PR #4447, opened 2019-12-04
Fixed the types of the args of extern functions

benoitsteiner:
Made sure the arguments of extern functions are of type Var instead ofmstring immediates.


abadams:
The only place that calls this function goes to some trouble to unpack a vector<Var> into a vector of strings... I'm not sure what the point of that is. Maybe it would be better to change this define_extern to accept a vector of Var as the args argument, and then not mess with it?

benoitsteiner:
I've updated Function::define_extern to take a vector<Var>.

benoitsteiner:
Done

PR #4452, opened 2019-12-05
Added a test to cover the featurization of reductions

benoitsteiner:


apps/autoscheduler/test_function_dag.cpp
```diff
@@ -47,11 +44,11 @@ int main(int argc, char **argv) {
         std::vector<Var> vars = {x, y};
         Halide::Type input_type = Halide::Float(32);
         g.define_extern(
-            "generate_output_vals",
+            "mul_by_two",
             {arg},
             input_type,
             vars,
-            Halide::NameMangling::C);
+            Halide::NameMangling::C /*PlusPlus*/);
```
steven-johnson: Why is PlusPlus commented out?

apps/autoscheduler/test_function_dag.cpp
```diff
@@ -81,6 +78,102 @@ int main(int argc, char **argv) {
 
     // Disabled for now: there is still work to do to populate the jacobian
     //assert(with_extern.str() == without_extern.str());
+}
+
+extern "C" int matmul(
+    halide_buffer_t *input1,
+    halide_buffer_t *input2,
+    halide_buffer_t *output) {
+
```
steven-johnson: Nit: first line of function should not be blank

apps/autoscheduler/test_function_dag.cpp
```diff
@@ -81,6 +78,102 @@ int main(int argc, char **argv) {
...
+                (*out) += (*in1) * (*in2);
+            }
+        }
+    }
+    return 0;
+}
+
+void test_matmul(const MachineParams &params, const Target &target) {
+
```
steven-johnson: ditto

apps/autoscheduler/test_function_dag.cpp
benoitsteiner: It's a mistake, fixed :)

apps/autoscheduler/test_function_dag.cpp
benoitsteiner: done

apps/autoscheduler/test_function_dag.cpp
benoitsteiner: done

PR #4454, opened 2019-12-05
Add HALIDE_MUST_USE_RESULT macro to HalideRuntime.h and AOT .h

steven-johnson:
This is designed to make it easier to enforce examination of the result code of calls into Halide AOT code; by providing the HALIDE_MUST_USE_RESULT as part of the generated .h files, you can enable this by simply adding `-DHALIDE_FUNCTION_ATTRS=HALIDE_MUST_USE_RESULT` to your build options (rather than having to ensure that other dependencies ahead of other includes).

This mimics the definition used by Abseil, in which the C++17 [[nodiscard]] attribute is preferred when available.

(It's quite possible that HALIDE_MUST_USE_RESULT could be profitably applied to Halide APIs as well, both internal and external; I haven't attempted to look for candidates at this time.)

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -35,6 +35,24 @@ extern "C" unsigned char halide_internal_runtime_header_HalideRuntimeD3D12Comput
 
 namespace {
 
+// HALIDE_MUST_USE_RESULT defined here is intended to exactly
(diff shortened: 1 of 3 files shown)


steven-johnson:
Failure is unrelated.

PR #4455, opened 2019-12-06
Improve the accuracy of the pipeline features for math functions

benoitsteiner:
Instead of modeling tanh as an extern function call, count the number of primitive
floating point operations required for an actual implementation of the function.
This PR only applies this approach to tanh, but if it seems reasonable to the Halide maintainers I will apply it more broadly to other math functions in follow up PRs.


abadams:
The possible gotcha is that all of the ops in the featurization vectorize cleanly, while the tanh call does not, and I'm not sure if this scalarization is captured anywhere else in the featurization. A property of extern calls that's important for the autoscheduler in addition to their performance is that it's a non-vectorizable chunk of code.

Maybe we should add to buckets according to the op count. and *also* add to the extern call bucket?

benoitsteiner:
Another issue is that my modeling of the tanh function is very different from the implementation of tanh provided by llvm, so it's not very accurate. I'll close this PR and try to come up with a better solution.

PR #4458, opened 2019-12-07
Avoid type punning in the generated C for buffer shapes

abadams:
We're playing pretty fast-and-loose with make_struct to generate halide_dimension_t arrays inline by making an array of ints and reinterpreting the pointer as a smaller array of halide_dimension_t. Newer gcc versions complain that the dimension array may be uninitialized if you do this. This PR specializes make_struct emission for the halide_dimension_t and prints a nicer array of initialized halide_dimension_t objects instead.


abadams:
The c_backend app exercises this, so it's at least legal. That app fails with gcc 9 before this change.

Issue #4122, opened 2019-08-08
apps/fft fails to compile on AVX512 systems


steven-johnson:
When doing `make test` on an AVX512-enabled system, this fails to build at compiletime, with many errors of the form

```
This pipeline is guaranteed to fail an assertion at runtime:
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 8, 0, 15)
```

Presumably the larger size of `natural_vector_size<float>` (64 vs 32 on AVX2) is somehow not accounted for in the code?

PR #4461, opened 2019-12-09
Fix small FFTs.

### LINKED ISSUE #4450: bugs in fft2d_r2c fft2 apps/fft
Hi Team,

a bug was found in the fft library implemented in Halide. 
function fft2d_r2c in the directory apps/fft would trigger some crashes/exceptions when N0 and N1 is smaller than 16. 

A short example is given below. 

'
    Fft2dDesc fwd_desc;
    Fft2dDesc inv_desc;
    inv_desc.gain = 1.0f / (W * H);
    int W=8, H=8;
    ComplexFunc dft_in = fft2d_r2c(make_real(in), W, H, target, fwd_desc);
    ComplexFunc dft_out = fft2d_c2r(dft_in, W, H, target, inverse_desc);
    dft_out.compute_root();
    dft_out.realize();
'

Or replace the Makefile in the folder fft with the zipped Makefile, and run 
'
make test
'
the following messages will pop up.
'
bin/host/bench_fft 8 8 bin/host
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 4, 0, 7)
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 4, 0, 7)
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 4, 0, 7)
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 4, 0, 7)
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 4, 0, 7)
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior", 0, 4, 0, 7)

'

[Makefile.zip](https://github.com/halide/Halide/files/3927637/Makefile.zip)

dsharletg:
This fixes two issues:
- When the FFT had only one stage, the schedule was bad (vectorized an inline stage).
- When the size of an FFT -> inverse FFT pipeline is smaller than 2x the vector size, the inverse FFT would access part of the forward FFT that is not defined. These values probably weren't used in the result, but triggered asserts anyways.

Fixes #4450 and #4122


dsharletg:
Probably also fixes #4122

abadams:
For avx512 with this branch I get:

bin/host/bench_fft 8 8 bin/host
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("x", "constant_exterior", 0, 7, -8, 7)
...
Error:
Bounds given for constant_exterior in x (from 0 to 7) do not cover required region (from -8 to 7)

Probably with avx2 you can repro with a 4x4 fft

dsharletg:
Can you try the current branch on an AVX512 machine?

FWIW, with 4x4 FFTs, I'm seeing correctness issues. I'm not sure if this is due to algorithm issues or vectorization. I'm working on debugging it, but it's a challenging issue and I suspect something deeper than just a bug in the FFT code. It would be useful information to know what the current branch does on an AVX512 machine.

dsharletg:
Also: This is all perhaps a bit of a wild goose chase. It doesn't make sense to try to implement these super small (size ~= SIMD size) FFTs with all of the fancy optimizations we have (like zipping 2 real FFTs into one complex FFT) because it just adds overhead and doesn't actually execute any faster.

I am not sure if we should build a special case for these into this library or not. The problem is that it changes the API in a visible way: fft2d_c2r would consume an N0 x N1 FFT, rather than the N0 x (N1 + 1) / 2. 

Maybe what we should do is fft2d_r2c should do this optimization, because producing a bigger output than necessary should be OK. However, consuming a larger input than expected (in the case of fft2d_c2r) is not.

dsharletg:
Hmm, even just doing this for fft2d_r2c is problematic. I think we should just make it functional and I'll add some docs/comments that if you are doing FFTs with the size ~= the SIMD width, that you're probably better off using fft2d_c2c.

abadams:
The current state of the branch is that size 8, 16, 32, 64 work, but size 48 fails with:

This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("y", "constant_exterior$1", 0, 24, 0, 31)
Error:
Bounds given for constant_exterior in y (from 0 to 24) do not cover required region (from 0 to 31)

dsharletg:
I think these cases are now fixed. However, this code is seriously growing some needs for some ugly heuristics that change the algorithm/schedule based on vector sizes and FFT sizes...

abadams:
You keep making it harder for yourself! Now 48 works but the newly-added 24 shows similar errors: 

This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("x", "c2c_fft1_S8_R3_y$2", 0, 23, 0, (max((f0.extent.0 + f0.min.0), 32) + -1))
Warning:
This pipeline is guaranteed to fail an assertion at runtime: 
halide_error_explicit_bounds_too_small("x", "c2c_fft1_S8_R3_x$2", 0, 23, 0, 31)
Error:
Bounds given for c2c_fft1_S8_R3_y in x (from 0 to 23) do not cover required region (from 0 to 31)
Aborted (core dumped)

dsharletg:
Fun observation: For 8x8 FFTs, I think LLVM is mem2reg-ing *everything*, and optimizing it... a lot:

    DFT type  Time (us)    MFLOP/s  Time (us)    MFLOP/s      Ratio
         c2c      0.007  293533.10      0.000        inf          0
         r2c      0.003  305052.43      0.000        inf          0
         c2r      0.004  273348.52      0.000        inf          0
bin/host/bench_fft 12 12 bin/host
                       Halide                  FFTW     
    DFT type  Time (us)    MFLOP/s  Time (us)    MFLOP/s      Ratio
         c2c      0.360   14337.50      0.000        inf          0
         r2c      0.322    8016.76      0.000        inf          0
         c2r      0.384    6718.60      0.000        inf          0
bin/host/bench_fft 16 16 bin/host
                       Halide                  FFTW     
    DFT type  Time (us)    MFLOP/s  Time (us)    MFLOP/s      Ratio
         c2c      0.433   23627.41      0.000        inf          0
         r2c      0.203   25193.01      0.000        inf          0
         c2r      0.204   25045.62      0.000        inf          0

abadams:
Wow, that's a lot more MFLOP/s than I have on one core, so I'm guessing the bulk of the computation is being hoisted out of the benchmarking loop?

abadams:
Also, all even sizes seem to work for me now.

dsharletg:
I don't think it's possible for the benchmark to get optimized away like that. I think the issue is that FLOP number for FFTs is an approximation from fftw: http://www.fftw.org/speed/method.html. 

In particular, for small FFTs, it starts to become really wrong. For example, an 8 point FFT per the approximation has 5 * 8 * log2(8) = 120 FLOP. But in reality, this size FFT has far fewer flops, because so many of the operations are multiply by 1, or 0 (and so eliminating an associated add/subtract too).

Basically, I think these crazy numbers are how many FLOPs would be computed *if* the FFT were executed by a general algorithm, and not one specialized for this size.

steven-johnson:
Are we just awaiting a green build here?

steven-johnson:
Looks ready to merge?

abadams:
I checked with Dillon, and he said it was fine to merge.

Issue #4459, opened 2019-12-08
OpenCL. Segmentation Fault on machine with NumDevices > 4. 


akokoshn:
I have stack with an issue: in case of using OpenCL target my application crash on machine with 8 devices.
Seems `clGetDeviceIDs(platform, device_type, maxDevices, devices, NULL);` maxDevices (which equals 4) ignored by OpenCL:
https://github.com/halide/Halide/blob/cc5af85131a97b4fa3fbd50ef53d5ac3e43dad1b/src/runtime/opencl.cpp#L409

I have added work around:

```
const cl_uint maxDevices = 8;
cl_device_id devices[maxDevices];
cl_uint deviceCount = 0;
err = clGetDeviceIDs(platform, device_type, 0, NULL, &deviceCount);
if (err != CL_SUCCESS || deviceCount > maxDevices) {
    error(user_context) << "CL: clGetDeviceIDs failed: "
                        << get_opencl_error_name(err);
    return err;
}
err = clGetDeviceIDs(platform, device_type, maxDevices, devices, NULL);
if (err != CL_SUCCESS) {
    error(user_context) << "CL: clGetDeviceIDs failed: "
                        << get_opencl_error_name(err);
    return err;
}
```

It's fix problem!

As option for right fix I can suggest after first call clGetDeviceIDs for fill in deviceCount, get IDs for devices in loop:

```
clGetDeviceIDs(... &deviceCount)
for (int i = 0; i < deviceCount; i++) {
    clGetDeviceIDs
}
```

-Alexey

dsharletg:
Which OpenCL implementation is it that's crashing here? That seems like it's worth a bug report to that OpenCL vendor.

I don't see how the loop solution you proposed would work, `clGetDeviceIDs` doesn't seem to have an option to get a device ID by index. I think the "correct" pattern here is to call `clGetDeviceIDs` with a null `devices` pointer to get the number of devices, then allocate an array of that size and call it again.

However, I think it would be OK here to just make `maxDevices` really big, like 100+.

akokoshn:
Thanks for quick response.
The issue observed on OpenCL 1.2, Driver - 418.87.00.
Yes, generally fail looks like OpenCL issue.

You are right, `clGetDeviceIDs(platform, device_type, 1, devices, NULL)` applicable only in case of using first device (ignore another). So, I see only one way for avoid issue (without change OpenCL) - dynamically allocate `devices` array (size = actual `deviceCount`)

[...]

PR #4464, opened 2019-12-10
Increase max device count to query in OpenCL runtime

vksnk:
To address https://github.com/halide/Halide/issues/4459 .


abadams:
Failure is unrelated (LLVM trunk not building)

Issue #4466, opened 2019-12-10
Bounds inference for quiet_div


vksnk:
I'm working on a GPU schedule for the image upsampling function and in some cases it asserts inside of the fuse_gpu_thread_loops. I think the issue is caused by bounds inference which returns pos_inf while estimating shared memory size. It looks like the reason is that expression which defines shared memory size contains quiet_div, which is not supported by the bounds inference (at least I couldn't find it mentioned anywhere in Bounds.cpp; there is handling of Div, though and if it matters divider in my case is constant and greater than zero, so it should work for my case). 

I saw https://github.com/halide/Halide/pull/4439, which seems to remove quiet_div completely and probably will fix this issue, but I figured still would post just in case.

abadams:
quiet_div has indeed been removed

PR #4467, opened 2019-12-11
Check if shared memory allocation is bounded

vksnk:
If unchecked it fails down the road with some obscure message. 

Related to https://github.com/halide/Halide/issues/4466

### DIFF
--- a/src/FuseGPUThreadLoops.cpp
+++ b/src/FuseGPUThreadLoops.cpp
@@ -311,7 +311,10 @@ class ExtractSharedAllocations : public IRMutator {
             // Expand the inner allocations using the loop bounds.
             for (SharedAllocation &s : allocations) {
                 if (expr_uses_var(s.size, op->name)) {
-                    s.size = bounds_of_expr_in_scope(s.size, scope).max;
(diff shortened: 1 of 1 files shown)


src/FuseGPUThreadLoops.cpp
```diff
@@ -311,7 +311,10 @@ class ExtractSharedAllocations : public IRMutator {
             // Expand the inner allocations using the loop bounds.
             for (SharedAllocation &s : allocations) {
                 if (expr_uses_var(s.size, op->name)) {
-                    s.size = bounds_of_expr_in_scope(s.size, scope).max;
+                    auto interval_bounds = bounds_of_expr_in_scope(s.size, scope);
+                    user_assert(!interval_bounds.max.same_as(Interval::pos_inf()))
```
abadams: `interval_bounds.has_upper_bound()` would be a little cleaner.

vksnk: Thanks!

steven-johnson:
build failure is unrelated, OK to merge

PR #4480, opened 2019-12-18
Reschedule conv layer

abadams:
Our conv layer schedule was pretty tragic, for both CPU and GPU. This PR fixes them (on at least one machine). This conv schedule hits 88% of peak flops on my GPU, and 91.7% of peak flops on my CPU.

### DIFF
--- a/apps/conv_layer/conv_layer_generator.cpp
+++ b/apps/conv_layer/conv_layer_generator.cpp
@@ -54,6 +54,11 @@ public:
 
         bias.dim(0).set_bounds(0, CO).set_stride(1);
 
+        input.set_host_alignment(128);
+        bias.set_host_alignment(128);
+        filter.set_host_alignment(128);
+        relu.set_host_alignment(128);
+
         if (auto_schedule) {
             input.dim(0).set_estimate(0, CI);
             input.dim(1).set_estimate(0, W + 2);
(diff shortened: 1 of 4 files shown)


apps/conv_layer/conv_layer_generator.cpp
```diff
@@ -10,72 +10,185 @@ class ConvolutionLayer : public Halide::Generator<ConvolutionLayer> {
...
+        input.dim(3).set_bounds(0, N).set_stride(CI * (W + 2) * (H + 2));
+
+        filter.dim(0).set_bounds(0, CO).set_stride(1);
+        filter.dim(1).set_bounds(0, 3).set_stride(CO);
+        filter.dim(2).set_bounds(0, 3).set_stride(CO * 3);
+        filter.dim(3).set_bounds(0, CI).set_stride(CO * 3 * 3);
+
+        bias.dim(0).set_bounds(0, CO).set_stride(1);
+
+        input.set_host_alignment(128);
```
dsharletg: I'm curious, this is new in this PR (so not a remanent of a Hexagon schedule?), did this have some impact on  CPU/GPU performance?

abadams: I was just copying over the code from the autoscheduler branch. I'm not sure how important it is, but it certainly makes it easier to understand the generated code. Probably there are some alignment-related benefits. (CUDA only does vector loads from aligned addresses)

abadams: Oh yeah, if you're referring specifically to the host alignment constraints, it's so that pairs of 32-bit loads can be done as a single 64-bit load instead, to reduce the number of memory transactions. 128 is probably overkill.

dsharletg: Yeah, I was wondering specifically about set_host_alignment. It hadn't occurred to me that it might significantly affect CPU codegen. Maybe I should try it on more code...

abadams: I don't think it matters for modern CPUs - the memory transaction minimizing trick is just for CUDA.

dsharletg: OK that makes more sense. I was confused about this but now it's clear.

I think the alignment requirement should be reduced to what is necessary, and moved inside the GPU specific schedule.

abadams: Turned out to be unnecessary (still not entirely sure why), so I just deleted the alignment constraints.

Issue #4481, opened 2019-12-20
Is there any plan for supporting Tensor Core?


jinderek:
NVIDIA Volta and Turing GPUs have Tensor Cores, which can massively accelerate large matrix operations.(https://www.nvidia.com/en-us/data-center/tensorcore/)

So is there any plan to support Tensor Core?

abadams:
If someone needed it today, then I'd support it via a define_extern that just does a matrix multiply. What we'd really like to do though is be able to map more general tensor contractions to the wmma instruction, and that's considerably harder.

vinodgro:
@abadams are there examples of something similar in halide? if so I can check if i can get someone here interested in something like this.

abadams:
I don't think there is anything close currently. Here's a strawman sketch of how to support it:

Add a tensor_core scheduling directive that takes two vars and an RVar. It is expected that there are outer dimensions that are marked GPUBlocks. There may also be GPUThreads dimensions, but there may not be GPULanes. tensor_core uses the full warp, so it's implicitly a GPULanes dimension of size 32. Here's a rough example of what a schedule might look like:

RDom k(0, 100);
C(i, j) += cast<float>(A(i, k)) * B(k, j);
C.update().tile(i, j, ii, ji, 4, 4).split(k, ko, ki, 4).gpu_blocks(i, j).reorder(ii, ji, ki, ko, i, j).tensor_core(ii, ji, ki)

We'd then assert a few things.
1) The stage's definition must be exactly the one written above.
2) The extent of the Var, Var, and RVar passed to tensor_core must all be statically known to be four.

Over time the goal would be to relax these constraints as much as possible. Supporting constant values other than four is easy by just unrolling/masking. Supporting different patterns is harder. Commutative reshufflings is fine, as are col major vs row major. But it would be nice to do something sensible in the face of things like convolutions: A(k) * B(i + k, j), by broadcasting A and shearing B as necessary.

vinodgro:
To get something with ok performance your sketch seems like a good start, though it will still be a lot of work to get it to work. We have prototyped some GPU scheduling primitives in our own standalone DSL. I gave a lightening talk at the TVM conf earlier this month. I'd like to port that over to Halide and recruiting some folks internally to help. or if I get a good intern.

rioffe:
> I don't think there is anything close currently. Here's a strawman sketch of how to support it:
> 
[...]

PR #4483, opened 2019-12-21
MSAN: check input buffers for poison (Issue #4481)

steven-johnson:
Augment our MSAN support to ensure that input buffers (and the output of extern stages) are checked for uninitialized data.

Note 1: msan_aottest has become an ugly mess of code; it should probably be changed to use a table-driven state machine. Not sure if that blocks this or not.

Note 2: this should probably get checked inside Google before landing (which will probably happen post-holidays), as we have a fair number of complex pipelines that are running with msan enabled.

### DIFF
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -143,6 +143,7 @@ Stmt add_image_checks(Stmt s,
     vector<Stmt> asserts_host_alignment;
     vector<Stmt> asserts_host_non_null;
     vector<Stmt> buffer_rewrites;
+    vector<Stmt> msan_checks;
 
     // Inject the code that conditionally returns if we're in inference mode
     Expr maybe_return_condition = const_false();
@@ -257,6 +258,13 @@ Stmt add_image_checks(Stmt s,
         // Come up with a name to refer to this buffer in the error messages
         string error_name = (is_output_buffer ? "Output" : "Input");
         error_name += " buffer " + name;
+
+        if (!is_output_buffer && t.has_feature(Target::MSAN)) {
+            Expr buffer = Variable::make(type_of<struct halide_buffer_t *>(), name + ".buffer");
+            Stmt check_contents = Evaluate::make(
+                Call::make(Int(32), "halide_msan_check_buffer_is_initialized", {buffer}, Call::Extern));
+            msan_checks.push_back(check_contents);
+        }
 
         // Check the type matches the internally-understood type
         {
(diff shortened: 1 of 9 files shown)


src/AddImageChecks.cpp
```diff
@@ -650,6 +658,11 @@ Stmt add_image_checks(Stmt s,
         s = LetStmt::make(lets_required[i - 1].first, lets_required[i - 1].second, s);
     }
 
+    // Inject the code that checks that does msan checks. (Note that this ignores no_asserts.)
```
abadams: There are people running with msan-no_asserts ?!

src/ScheduleFunctions.cpp
```diff
@@ -600,21 +603,31 @@ Stmt build_extern_produce(const map<string, Function> &env, Function f, const Ta
...
-            if (!is_no_op(annotate)) {
-                annotate = Block::make(annotate, mark_contents);
+            if (!is_no_op(pre_call)) {
+                pre_call = Block::make(pre_call, mark_contents);
             } else {
-                annotate = mark_contents;
+                pre_call = mark_contents;
+            }
+        }
+        // Check the output buffer(s) to be sure they are fully initialized.
```
abadams: Reading upwards, I think these are the output buffers of any extern calls used, rather than output buffers to the pipeline. If that's correct, please clarify in the comment.

src/runtime/HalideRuntime.h
```diff
@@ -934,10 +934,33 @@ extern void halide_memoization_cache_release(void *user_context, void *host);
...
 
+/** Verify that a given range of memory has been initialized; only used when Target::MSAN is enabled.
+ *
+ * The default implementation simply calls the LLVM-provided __msan_check_mem_is_initialized() function.
+ *
+ * The return value should always be zero.
+ */
+extern int halide_msan_check_memory_is_initialized(void *user_context, const void *ptr, uint64_t len);
+
+/** Verify that the data pointed to by the buffer_t is initialized (but *not* the buffer_t itself),
```
abadams: s/buffer_t/halide_buffer_t

test/generator/msan_aottest.cpp
```diff
@@ -19,25 +19,87 @@ int main(int argc, char **argv) {
 using namespace std;
 using namespace Halide::Runtime;
 
+enum {
+    kAnnotate_bounds_inference_buffer,
```
abadams: elsewhere for enums we just use ThisSortOfThing

src/AddImageChecks.cpp
steven-johnson: I haven't seen that yet, but I will bet you $1 someone does at some point (even inadvertently)...

src/ScheduleFunctions.cpp
steven-johnson: done

src/runtime/HalideRuntime.h
steven-johnson: done

test/generator/msan_aottest.cpp
steven-johnson: done

PR #4488, opened 2019-12-26
Use pointer access for aligned load/store in OpenCL codegen

vksnk:
Currently, OpenCL codegen will produce vloadn/vstoren for vectorized load/stores. These functions require alignment to the base type, but not vector, so will allow unaligned access. In some cases, we have information about alignment of the load/store op and can generate cast to the vector + access through the pointer, which has to be aligned to the vector and might be more efficient on some platforms.  

I've tested it on two different platforms and one of them my kernel is significantly faster, while on the other there is no difference.


src/CodeGen_OpenCL_Dev.cpp
```diff
@@ -235,42 +235,58 @@ string CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::print_extern_call(const Call *op) {
...
     if (ramp_base.defined()) {
         internal_assert(op->type.is_vector());
-        string id_ramp_base = print_expr(ramp_base);
 
         ostringstream rhs;
-        rhs << "vload" << op->type.lanes()
-            << "(0, (" << get_memory_space(op->name) << " "
-            << print_type(op->type.element_of()) << "*)"
-            << print_name(op->name) << " + " << id_ramp_base << ")";
+        if((op->alignment.modulus % op->type.lanes() == 0) && (op->alignment.modulus % op->type.lanes() == 0)) {
```
steven-johnson: Nit: this might be easier to read with a line break after the &&

src/CodeGen_OpenCL_Dev.cpp
```diff
@@ -413,15 +429,23 @@ void CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::visit(const Store *op) {
...
         internal_assert(op->value.type().is_vector());
-        string id_ramp_base = print_expr(ramp_base);
 
-        stream << get_indent() << "vstore" << t.lanes() << "("
-               << id_value << ","
-               << 0 << ", (" << get_memory_space(op->name) << " "
-               << print_type(t.element_of()) << "*)"
-               << print_name(op->name) << " + " << id_ramp_base
-               << ");\n";
+        if((op->alignment.modulus % op->value.type().lanes() == 0)
```
steven-johnson: space between `if` and `(` in Halide code style

src/CodeGen_OpenCL_Dev.cpp
abadams: I think this is intended to be (modulus, remainder), not (modulus, modulus).1

src/CodeGen_OpenCL_Dev.cpp
abadams: modulus/modulus check again

src/CodeGen_OpenCL_Dev.cpp
vksnk: > Nit: this might be easier to read with a line break after the &&

Fixed.

> I think this is intended to be (modulus, remainder), not (modulus, modulus).

Ah, that's embarrassing. Thanks for catching this!

src/CodeGen_OpenCL_Dev.cpp
vksnk: Done.

steven-johnson:
Is this ready to approve (pending green)?

vksnk:
> Is this ready to approve (pending green)?

Yes, I've addressed comments.

PR #4493, opened 2019-12-30
Reapply clang-format to src/*

steven-johnson:
We did this in a giant batch once before, but changes have introduced drift over time. Bringing it all back up to standard.

### DIFF
--- a/src/Schedule.cpp
+++ b/src/Schedule.cpp
@@ -287,7 +287,7 @@ struct StageScheduleContents {
     StageScheduleContents()
         : fuse_level(FuseLoopLevel()), touched(false),
           allow_race_conditions(false), atomic(false),
-          override_atomic_associativity_test(false) {};
+          override_atomic_associativity_test(false){};
 
     // Pass an IRMutator through to all Exprs referenced in the StageScheduleContents
     void mutate(IRMutator *mutator) {
--- a/src/Simplify_Add.cpp
+++ b/src/Simplify_Add.cpp
@@ -49,100 +49,99 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
             return rewrite.result;
         }
 
-        if (EVAL_IN_LAMBDA
-            (rewrite(x + x, x * 2) ||
-             rewrite(ramp(x, y) + ramp(z, w), ramp(x + z, y + w, lanes)) ||
-             rewrite(ramp(x, y) + broadcast(z), ramp(x + z, y, lanes)) ||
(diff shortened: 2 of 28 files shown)


src/Schedule.cpp
```diff
@@ -287,7 +287,7 @@ struct StageScheduleContents {
     StageScheduleContents()
         : fuse_level(FuseLoopLevel()), touched(false),
           allow_race_conditions(false), atomic(false),
-          override_atomic_associativity_test(false) {};
+          override_atomic_associativity_test(false){};
```
abadams: I think maybe the trailing semicolon confused it here

src/Simplify_Add.cpp
```diff
@@ -49,100 +49,99 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
             return rewrite.result;
         }
 
-        if (EVAL_IN_LAMBDA
```
abadams: Let's skip Simplify_*.cpp. It makes a mess.

src/VaryingAttributes.cpp
```diff
@@ -1339,15 +1339,15 @@ class CreateVertexBufferHostLoops : public IRMutator {
             // Insert two new for-loops for vertex buffer generation on the host
```
abadams: We should also skip this file, or add no-clang-format comments

src/WasmExecutor.cpp
```diff
@@ -16,8 +16,8 @@
 #include <vector>
 
 #ifdef WITH_V8
-#include "v8.h"
 #include "libplatform/libplatform.h"
+#include "v8.h"
```
abadams: I have a feeling this order might be important. There was a reason I skipped it last time

src/Simplify_Add.cpp
steven-johnson: Sure, but we should insert no-clang-format here (and elsewhere) to avoid that; I'd like to get to a point where we can run clang-format as a presubmit check, so avoiding special cases is good.

src/WasmExecutor.cpp
steven-johnson: In general, relying on #include order being important is a code smell -- aside from preprocessor-list-of-stuff hacks, which are a special case and should definitely be called out with no-clang-format.

steven-johnson: edit: ah, no, you are probably correct here, as v8 is kinda special. will wrap.

steven-johnson:
Closed in favor of https://github.com/halide/Halide/pull/4498

PR #4494, opened 2019-12-30
Apply clang-format to python_bindings/src

steven-johnson:
Aint no party like a clang-format party


python_bindings/src/PyBinaryOperators.h
```diff
@@ -6,33 +6,34 @@
...
-        BINARY_OP(|, or)
-        BINARY_OP(^, xor)
-        BINARY_OP(<, lt)
-        BINARY_OP(<=, le)
-        BINARY_OP(==, eq)
-        BINARY_OP(!=, ne)
-        BINARY_OP(>=, ge)
-        BINARY_OP(>, gt)
-    ;
+            BINARY_OP(%, mod)
```
abadams: Perhaps the macro needs to accommodate a trailing ;

python_bindings/src/PyExpr.cpp
```diff
@@ -13,7 +13,7 @@ void define_expr(py::module &m) {
         std::ostringstream o;
         o << e;
         throw py::value_error("The halide.Expr (" + o.str() + ") cannot be converted to a bool. "
-            "If this error occurs using the 'and'/'or' keywords, consider using the '&'/'|' operators instead.");
+                                                              "If this error occurs using the 'and'/'or' keywords, consider using the '&'/'|' operators instead.");
```
abadams: Break after + to avoid silly indent. Also below.

python_bindings/src/PyFunc.cpp
```diff
@@ -106,256 +101,214 @@ void define_func(py::module &m) {
...
-        .def("dimensions", &Func::dimensions)
-        .def("args", &Func::args)
-        .def("value", &Func::value)
-        .def("values", [](Func &func) -> py::tuple {
-            return to_python_tuple(func.values());
-        })
-        .def("defined", &Func::defined)
-        .def("outputs", &Func::outputs)
-        .def("output_types", &Func::output_types)
+                          .def(py::init<>())
```
abadams: Perhaps we should uniformly break after the = in these definitions to avoid the massive indent

PR #4499, opened 2019-12-30
Tweaks to clang-format

steven-johnson:
- Remove duplicate `Cpp11BracedListStyle` declaration (some tools will barf on duplicate keys, apparently)
- Sort all the entries
- Remove the misleading `BasedOnStyle` line (it was commented out and so we were really based on 'default')


steven-johnson:
(please ignore the build failures; I manually stopped them all to avoid wasting cycles on this since it won't affect any existing tests)

PR #4501, opened 2020-01-02
Clang format makefile target

abadams:
Add a makefile target to clang-format everything except for the tutorials, and apply it.


steven-johnson:
A point of clarity we should document somwhere while we're at it: let's document the version of clang-format we expect. (I'd vote for most-recent-stable, thus clang-format-9 for now.)

src/runtime/hexagon_remote/bin/src/halide_hexagon_remote.h
```diff
@@ -3,51 +3,51 @@
 #include "AEEStdDef.h"
```
steven-johnson: Is this machine-generated? If so, we should exclude it

src/runtime/hexagon_remote/bin/src/halide_hexagon_remote_skel.c
```diff
@@ -20,216 +20,215 @@
 #ifdef _WIN32
```
steven-johnson: Is this machine-generated? If so, we should exclude it

src/runtime/hexagon_remote/bin/src/halide_hexagon_remote_stub.c
```diff
@@ -20,216 +20,215 @@
 #ifdef _WIN32
```
steven-johnson: Is this machine-generated? If so, we should exclude it

(maybe we should exclude all the files in this dir anyway?)

Makefile
```diff
@@ -9,6 +9,11 @@
 #     For correctness and performance tests this include halide build time and run time. For
 #     the tests in test/generator/ this times only the halide build time.
 
+# Disable built-in makefile rules for all apps to avoid pointless file-system
+# scanning and general weirdness resulting from implicit rules.
+MAKEFLAGS += --no-builtin-rules --no-builtin-variables
```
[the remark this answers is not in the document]
abadams: This makes make intentionally dumber. It prevents a bunch of silly built-in rules from running and failing. The two main effects are a slight increase in performance, and also when you typo a target name, or try for a target that doesn't exist, it errors out instead of doing something weird based on implicit magic. 

For example in iir_blur, if I say "make bin/host/iir_blur.generator" it makes that file. Correct. But if I say "make iir_blur_generator" it finds iir_blur_generator.cpp, then decides to invoke a c++ compiler on it with the wrong set of flags to try to compile that file as a standalone tool into the current directory. You get a spew of linker errors. With implicit rules deleted, it instead says "no rule to make target iir_blur_generator", which is more helpful.

apps/linear_algebra/Makefile
```diff
@@ -71,13 +71,9 @@ BENCHMARKS = \
 all: $(BENCHMARKS)
 	make run_benchmarks
 
-ifneq ("$(wildcard /usr/include/cblas.h)","")
+.PHONY: test
```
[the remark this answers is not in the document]
abadams: Because cblas isn't always installed in that location. On current ubuntu it's in /usr/include/openblas/cblas.h

steven-johnson:
> I think we still need to whitelist the directories to apply it to - otherwise all the .cpp build intermediates in bin cause clang-format to take forever, and I didn't want to break someone manually invoking clang-format in that folder.

Oh, good point. Maybe just exclude `bin/*` from the find.

abadams:
To do my cmake build I just made a cmake_build subdirectory, which contains generated source... I think I'm more comfortable sticking with the whitelist for now.

steven-johnson:
fair enough. we can revisit later if needed.

abadams:
The last test run of this branch that actually ran on all of the buildbots wasn't green. Do we have any clue as to why the buildbots seem to ignore some PRs?

steven-johnson:
> The last test run of this branch that actually ran on all of the buildbots wasn't green. 

Ouch -- sorry, I misremembered the last state. Should we back it out?

> Do we have any clue as to why the buildbots seem to ignore some PRs?

Nope. It may be that newer versions of Buildbot have fixed this, but we've avoided doing the upgrade on the (foolish) hope we'd migrate to something better.

(About a couple of years ago I dove into the Buildbot code to try to figure it out; I didn't succeed, but did find one dodgy bit that looked like it might be related -- so I offered a fix for that as a patch. Unfortunately, it wasn't accepted, as I didn't have a testcase for it, and the maintainers apparently weren't willing to entertain the fix without one.)

PR #4503, opened 2020-01-02
Run clang-format as a Presubmit check

steven-johnson:
This is the first attempt to use GitHub Actions with Halide; to dip our toe in the water, this adds an Action that runs on pull requests and verifies that all C/C++ code is conformant with our clang-format rules.

I'm honestly not sure how this checking will feel in practice; I suspect we may need to land it on an experimental basis for a few days and see what reactions are.

A few notes:
- this job uses a user-contributed action from the GitHub Marketplace; while the source is available for inspection (https://github.com/DoozyX/clang-format-lint-action), it does include binary blobs for the clang-format tools, and these could contain malicious stuff. That said, I think our exposure to risk here is slim, even if it does have malice, as the Action is spun up in a separate Ubuntu instance (hosted by GitHub, not us), which operates on a locally-cloned copy of the source, and which is never given the token necessary to push any changes. So if it's going to be naughty, I think the exposure is mainly to GitHub (for something abusing its VMs) rather than the integrity of our source. That said, I am not a security expert.
- If we're concerned about the security risk, we could just roll our own action -- it would just take some fiddling with Docker to produce the setup we need. (In this case I think it may make more sense to just use the existing one, especially since GitHub is listing it in its Actions Marketplace.)
- Currently, this fails with an error if anything isn't compliant with our clang-format rules, and there's no obvious way to make it a "warning" instead (which might be desirable as a transition period); that said, as a practical matter, we have no rules that prevent landing an approved PR even if there are build failures, so I'd argue that this 'error' is already tantamount to a warning.


steven-johnson:
(Note that this PR will generate check-clang-format failures until https://github.com/halide/Halide/pull/4501 lands!)

PR #4507, opened 2020-01-03
BoundsInference needs to mark buffer shapes too

steven-johnson:
BoundsInference.cpp was missing the logic to mark newly-created buffer shapes as initialized for MSAN, so pipelines with `.define_extern()` usage could get erroneous MSAN failures.


steven-johnson:
Failures are unrelated.

PR #4509, opened 2020-01-03
Remove --no-builtin-variables from MAKEFLAGS

steven-johnson:
The additions of MAKEFLAGS is causing my local Linux box to fail:

```
$ make -j72
mkdir -p bin/build
Found a new enough version of clang
touch bin/build/rtti_ok
/usr/local/google/home/srj/GitHub/Halide/tools/binary2cpp.cpp -o bin/binary2cpp
mkdir -p bin/build
bash: /usr/local/google/home/srj/GitHub/Halide/tools/binary2cpp.cpp: Permission denied
```

Let's back this out until it can be debugged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,8 @@
 
 # Disable built-in makefile rules for all apps to avoid pointless file-system
 # scanning and general weirdness resulting from implicit rules.
-#MAKEFLAGS += --no-builtin-rules --no-builtin-variables
-#.SUFFIXES:
+MAKEFLAGS += --no-builtin-rules 
+.SUFFIXES:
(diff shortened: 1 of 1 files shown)


abadams:
Looks like you don't have CXX set, which is something I should have anticipated. Could you try just deleting --no-builtin-variables? That flag isn't so important.

steven-johnson:
That's the ticket!

steven-johnson:
apps/linear_algebra is failing on OSX builds, because <cblas.h> isn't found. (The ifeq removed from the Makefile seems likely to be the culprit.)

PR #4513, opened 2020-01-06
inject_host_dev_buffer_copies() should be a no-op for t.arch = Hexagon

steven-johnson:


src/Lower.cpp
```diff
@@ -294,7 +294,7 @@ Module lower(const vector<Function> &output_funcs,
         s = select_gpu_api(s, t);
         debug(2) << "Lowering after selecting a GPU API for extern stages:\n"
                  << s << "\n\n";
-    } else {
+    } else if (t.arch != Target::Hexagon) {
```
abadams: I don't understand what's different about Hexagon that makes this necessary. For hexagon-native pipelines, shouldn't there be no device allocations or device_dirty flags anyway?

steven-johnson: Correct -- which is why it's bad to inject a call to `halide_copy_to_device()`, because that will eventually try to go thru the hexagon runtime to call out to the runtime to load the helper .so, etc.

abadams: I'm still confused - is copy_to_device getting called in this scenario? It shouldn't be, right? Or is just the existence of the call a problem somehow?

steven-johnson: > I'm still confused - is copy_to_device getting called in this scenario? 

Yes

> It shouldn't be, right? 

Correct.

> Or is just the existence of the call a problem somehow?

Correct -- I don't think it's possible to successfully call that from HVX code; copies have to be handled on the host side.

abadams: copy_to_device should only be injected and called if there's device code that uses the buffer. If something is a pure Hexagon pipeline, that shouldn't be the case, because it's all host code. So why is copy_to_device injected at all? It sounds like maybe inject host copies pass is confused about the hexagon device api flag on the loops?

steven-johnson: Hm, good point. Let me dig a little further first.

abadams: I'm not opposed to solving it this way if it's simpler - I just want to understand exactly what's going on so that it can be captured in a comment so I don't get confused again later.

steven-johnson: OK, did a little digging. I think the issue is that (under the hood) InjectBufferCopiesForInputsAndOutputs() calls InjectBufferCopiesForSingleBuffer() with `is_external=true` (which is reasonable), but that means that the initial `state.current_device` for each buffer stays at its default of `None` (ie unknown), so it generates calls to `copy_to_device()` just to be safe. 

Of course, when generating for Hexagon, this is a bad thing; the hexagon code expects the host shim to do all the necessary copying, and a callout from HVX code to do this is doomed to failure without major reworking (which would be silly).

Arguably, a cleaner fix here might be to have InjectBufferCopiesForSingleBuffer() special-case for target.arch=Hexagon and in that case set the initial `state.current_device` to DeviceAPI::Hexagon -- which would allow any other potential call sites to avoid this problem. I'll give that a try.

steven-johnson: So, as an update: there appear to be a number of other places in inject_host_dev_buffer_copies() that have the implicit assumption that the copies injected always start (and end) at the host; this is hardly surprising, given the name of the function. (e.g., `FindBufferUsage()` always assumes the starting point is Host, not current-device).

I'm sure we clean up all these corner cases, but I think  a more targeted fix of just bailing early for Hexagon (with better documentation) is simpler and safer.

steven-johnson:
PTAL

src/Lower.cpp
zvookin: There is not one host necessarily. Host simply means "the memory associated with the buffer." There's no reason Hexagon couldn't be talking to it's own GPU. And thus that we're trying to do this indicates there are architectural issues elsewhere. That said, fixing the root problem may be untenable and ```inject_host_dev_buffer_copies``` has long been a source of issues that could use a rethink.

I was going to comment on the PR pushing back on having the logic in the lower path. E.g. moving it into ```inject_host_dev_buffer_copies``` might be cleaner.

abadams:
The change LGTM, but in the hexagon path I was thinking that everything was host-only (host meaning hexagon memory in this instance), so while there should be copy-to-host conservatively injected, there should never be a copy-to-device injected.

There must be a non-trivial device_api on a loop somewhere inside the hexagon code to make it think its entering some device code. Presumable that device_api is hexagon, so the code is claiming that it's on hexagon, and is going to recursively transfer control to a sub-DSP?!

steven-johnson:
> There must be a non-trivial device_api on a loop somewhere inside the hexagon code

I haven't found any such thing yet, but I'll dig a bit more while I still have nicely implemented test cases at hand.

steven-johnson:
From my debugging I haven't found *any* unexpected device_apis in here. 

Again, I think the issue is that when we enter `InjectBufferCopiesForSingleBuffer` with `is_external=true`, we leave everything as 'unknown' and assume we have to copy from 'device=host' to 'device=hexagon' -- but in this case, those are essentially the same thing.

steven-johnson:
Since this is approved, I'm going to land this so an unbroken merge for Google is possible. Happy to circle back if someone can suggest how my reasoning is wrong.

abadams:
I think we may have different definitions of "unexpected". No objection to merging - the thing I'm confused about is how the to 'device=hexagon' arose. Once you're inside a hexagon loop, all the device apis on the inner loops should be None. So it should look like vanilla host IR at that point. copy_to_device is only injected if touched_on_device is true, which means that FindBufferUsage has current_device_api set to Hexagon at some point, which I think it can only do if it hits a for loop with that device_api. So there must be an inner loop flagged as Hexagon....

Except reading SelectGPUAPI again I see that all inner loops are indeed probably flagged as Hexagon. Maybe this isn't a great design.

steven-johnson:
> Except reading SelectGPUAPI again I see that all inner loops are indeed probably flagged as Hexagon. Maybe this isn't a great design.

:-/

PR #4515, opened 2020-01-07
Reinsert check for cblas.h

abadams:
But add more search locations


abadams:
I believe the cuda tests may now be flaky because they're being run in parallel, and the buildbots do not have enough GPU memory to support that reliably.

PR #4517, opened 2020-01-07
Permit more memory types for GPU intermediate data

abadams:
This PR provides several new places to store intermediate data in GPU kernels, by making several types of store_in invocations that currently return compile-time errors do something sensible instead.

* A Func can now always be stored at a consumer's threads loop and store_in MemoryType::Auto. Previously this only worked if the allocation was constant-sized. Otherwise it threw an error about using shared memory instead. This error was misleading because just switching the memory type to shared was also illegal without changing the compute_at location too (See next bullet). Now this just works by falling back to heap memory (global memory) if it's dynamically sized, by allocating a global pool outside the kernel invocation and statically partitioning it among the blocks and threads. This adds overhead compared to using registers, of course. Not clear how much worse it is than local memory though. Either way it's living in L1.

* A Func can now be stored at a consumer's threads loop and store_in MemoryType::Shared. Previously this threw a compile-time error about shared allocations inside the threads loop not being supported. It allocates a piece of shared memory and statically partitions it across the threads. This was a long-standing TODO in FuseGpuThreadLoops.cpp

* A Func can now be stored at either the block or threads loop of a consumer and store_in MemoryType::Heap. This is the big one - you can now store GPU intermediates in global memory, which means we can deal with GPU intermediates that can't fit in shared (see the new schedule for the IIR blur app). As with the first case above, this is supported by allocating a global buffer outside of the kernel invocation and statically partitioning it across blocks, and also striping it across threads if the store_at location is inside the threads loop.

An alternative considered was just calling malloc/free inside the kernel, but this has a two problems: First it requires GPU API-specific support for this, whereas the solution in this PR is agnostic to the GPU API. Second, for CUDA at least if you're going to malloc more than 8MB of heap (which we do, because this functionality is for allocations too big to fit into shared) you need to tell the cuda context the peak size before you launch *any* kernels, and this is not really feasible for us because we don't control the entire lifetime of the GPU context.

Another alternative considered was allocating a smaller pool sized according to the maximum number of blocks that can be resident on the GPU, rather than the total number of blocks to be launched, and then dynamically providing memory out of it using a free list, instead of statically partitioning a big pool. This would use less global memory in cases where you are launching many more blocks than the GPU can keep resident, but it would add a significant chunk of runtime overhead to manage the free list, and a bunch of API-specific work to implement it. I decided it's probably better to keep things simple and statically partition a single allocation. In the cases where it matters, people can always just launch fewer blocks by scheduling one of their outer blocks loops as a CPU-side serial loop, or scheduling one of their inner blocks loops as an in-kernel serial loop to keep fewer blocks resident for longer.

Finally note that the approach of allocating and freeing a pool (represented as a halide_buffer_t) to serve as the heap around every kernel invocation that uses heap memory assumes that something is caching allocations under the hood to make this cheap, as is the case with the cuda backend.


abadams:
@aekul

src/CodeGen_PTX_Dev.cpp
```diff
@@ -218,22 +218,38 @@ void CodeGen_PTX_Dev::visit(const Allocate *alloc) {
...
-            << "Only fixed-size allocations are supported on the gpu. "
-            << "Try storing into shared memory instead.";
 
-        BasicBlock *here = builder->GetInsertBlock();
-
-        builder->SetInsertPoint(entry_block);
-        Value *ptr = builder->CreateAlloca(llvm_type_of(alloc->type), ConstantInt::get(i32_t, size));
-        builder->SetInsertPoint(here);
-        sym_push(allocation_name, ptr);
+        if (size) {
```
abadams: oops, this code was not supposed to be committed. Will revert.

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -57,7 +60,11 @@ class SimplifyCorrelatedDifferences : public IRMutator {
         result = mutate(result);
 
         for (auto it = frames.rbegin(); it != frames.rend(); it++) {
-            result = LetStmtOrLet::make(it->op->name, it->op->value, result);
```
abadams: The changes in this file are because sizing the global pool well requires careful handling of correlated difference expressions. I discovered that the pass to simplify correlated difference expressions was not handling let values.

src/runtime/cuda.cpp
```diff
@@ -1181,6 +1181,9 @@ WEAK int halide_cuda_run(void *user_context,
         int result = halide_cuda_get_stream(user_context, ctx.context, &stream);
         if (result != 0) {
             error(user_context) << "CUDA: In halide_cuda_run, halide_cuda_get_stream returned " << result << "\n";
+            free(dev_handles);
```
abadams: drive-by fix. This seemed to throw an error but didn't bail out.

src/FuseGPUThreadLoops.cpp
```diff
@@ -347,51 +381,93 @@ class ExtractSharedAllocations : public IRMutator {
...
+        user_assert(!op->new_expr.defined())
+            << "Allocate node inside GPU kernel has custom new expression.\n"
+            << "(Memoization is not supported inside GPU kernels at present.)\n";
+
+        bool fixed_size_thread_allocation = (op->constant_allocation_size() != 0) && in_threads;
+
+        if ((fixed_size_thread_allocation &&
+             op->memory_type != MemoryType::Heap &&
+             op->memory_type != MemoryType::GPUShared) ||
+            op->memory_type == MemoryType::Register ||
```
shoaibkamil: Alignment is weird here (clang-format issue?)

abadams: Alignment seems legit to me - the structure of the expression is (A && B && C) || D || E, so A, B, C, are one space further in to line up with the open paren

BachiLi:
Should probably change the comments in `FuseGPUThreadLoops.h`. Right now it says it "hoists allocations at the block level out into a single shared memory array."

src/FuseGPUThreadLoops.cpp
```diff
@@ -868,7 +1087,11 @@ class FuseGPUThreadLoops : public IRMutator {
                      << loop << "\n\n";
 
             // Mutate the inside of the kernel
-            return FuseGPUThreadLoopsSingleKernel(block_size, shared_mem).mutate(loop);
+            loop = FuseGPUThreadLoopsSingleKernel(block_size, shared_mem).mutate(loop);
```
BachiLi: nit: I found the variable name "shared_mem" slightly confusing since it can now be shared memory *or* global memory. Maybe call it `shared_alloc`.

src/FuseGPUThreadLoops.cpp
```diff
@@ -570,44 +664,168 @@ class ExtractSharedAllocations : public IRMutator {
...
+        }
+        if (stmt_uses_var(s, num_threads_var_name)) {
+            s = LetStmt::make(num_threads_var_name, num_threads, s);
+        }
+
+        return s;
+    }
+
+    Stmt rewrap_heap(Stmt s, const ExtractBlockSize &bs, DeviceAPI device_api) {
+        if (!heap_bytes_per_block.defined()) {
```
BachiLi: could it be 0?

src/FuseGPUThreadLoops.cpp
```diff
@@ -570,44 +664,168 @@ class ExtractSharedAllocations : public IRMutator {
...
+                // Define an offset for each allocation. The offsets are in
+                // elements, not bytes, so that the stores and loads can use
+                // them directly.
+                for (int i = (int)(mem_allocs.size()) - 1; i >= 0; i--) {
+                    if (mem_allocs[i].memory_type != memory_type) {
+                        continue;
+                    }
+                    Expr group_offset = Variable::make(Int(32), "group_" + std::to_string(i) + ".offset");
+
+                    for (SharedAllocation &alloc : mem_allocs[i].group) {
```
BachiLi: const?

BachiLi: or just remove the &

src/FuseGPUThreadLoops.cpp
```diff
@@ -278,55 +306,61 @@ class ExtractSharedAllocations : public IRMutator {
...
-                if (expr_uses_var(s.size, op->name)) {
-                    auto interval_bounds = bounds_of_expr_in_scope(s.size, scope);
-                    user_assert(interval_bounds.has_upper_bound())
-                        << "Couldn't infer bounds for " << s.name << " shared memory allocation\n";
-                    s.size = interval_bounds.max;
-                }
+        // Expand any new shared allocations found in the body using the loop bounds.
+        Scope<Interval> scope;
+        scope.push(op->name, Interval(op->min, simplify(op->min + op->extent - 1)));
+        for (SharedAllocation &s : allocations) {
```
BachiLi: `const SharedAllocation &s` or `SharedAllocation s`

abadams:
This is green and ready to go

apps/iir_blur/iir_blur_generator.cpp
```diff
@@ -63,33 +63,41 @@ Func blur_cols_transpose(Func input, Expr height, Expr alpha, bool skip_schedule
...
-                .split(x, x, xi, 32)
-                .gpu_blocks(x, c)
-                .gpu_threads(xi);
+                .unroll(ry, 8)
+                .gpu_lanes(x);
+
+            // Stage the transpose input through shared so that we do
+            // strided loads out of shared instead of global.
+            blur.in()
+                .align_storage(x, 33)  // Avoid bank conflicts. Saves 0.05 ms
```
steven-johnson: 33 is a weird magic number here; would be worthy of naming and explaining?

apps/iir_blur/iir_blur_generator.cpp
```diff
@@ -63,33 +63,41 @@ Func blur_cols_transpose(Func input, Expr height, Expr alpha, bool skip_schedule
...
-            // 3.39ms on a 2060 RTX
+            // parallelism from the scan dimension. Most GPUs will be
+            // heavily under-utilized with this schedule and thus
+            // unable to hide the memory latencies to L2.
 
+            // 2.06ms on a 2060 RTX
             Var xi, yi;
-            blur.compute_root();
+            transpose.compute_root()
+                .tile(x, y, xi, yi, 32, 32)
```
steven-johnson: magic numbers like 32 are probably worth pulling into named constants and explaining why, at least in intended-as-example code like this

abadams: Done. It's the cuda warp size.

apps/iir_blur/iir_blur_generator.cpp
abadams: Done, it's the smallest integer greater than the warp size which is coprime with the warp size - isn't that obvious!? Honestly I'm not even sure bank conflicts are still a thing, but it did reduce runtime slightly.

PR #4518, opened 2020-01-07
Fix solve canonicalization

abadams:
Found a silly bug in solve_expression. We were unpacking a, then swapping the sense of a and b to canonicalize things. That should happen in the other order.

@aekul


abadams:
Note that this bug meant that an allocation was undersized, and so a legal Halide schedule wrote out of bounds, so it's a pretty serious miscompilation.

PR #4520, opened 2020-01-10
Make the Simplifier handle PureExtern functions via LUT

steven-johnson:
There were a number of PureExtern functions we didn't handle (notably sin and cos, but many others too); rather than add even more if-else clauses, I refactored the PureExtern case into a set of LUTs to reduce redundant code.

Note that the actual motivation for this change was to ensure that sin(CONSTANT) and cos(CONSTANT) got simplified to a real constant (rather than putting us at the mercy of LLVM).

(With more cleverness we could make it a single LUT, but my first approach involved template-fu that was overly complex to read; I suspect that this approach is a better balance of compile time vs code complexity, but would welcome someone more clever to simplify it further.)

### DIFF
--- a/src/Simplify_Call.cpp
+++ b/src/Simplify_Call.cpp
@@ -479,15 +479,15 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
         {
             static const std::unordered_map<std::string, std::function<bool(double)>>
                 pure_externs_f1b = {
-                    {"is_finite_f16", [](double a) { return std::isfinite(a); }},
-                    {"is_finite_f32", [](double a) { return std::isfinite(a); }},
-                    {"is_finite_f64", [](double a) { return std::isfinite(a); }},
-                    {"is_inf_f16", [](double a) { return std::isinf(a); }},
-                    {"is_inf_f32", [](double a) { return std::isinf(a); }},
-                    {"is_inf_f64", [](double a) { return std::isinf(a); }},
-                    {"is_nan_f16", [](double a) { return std::isnan(a); }},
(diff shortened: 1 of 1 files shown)


steven-johnson:
(Removing review request -- thought this was ready, but it isn't.)

steven-johnson:
OK, stupid bug fixed. 

Note that this is definitely a win for some code; for an internal bit of code that is basically a Wiener filter, the constifying of the sin and cos calls produces a measurable improvement of -8% for x86-64-avx2 (though, interestingly, no measurable change at all for arm64).

src/Simplify_Call.cpp
```diff
@@ -577,7 +470,143 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
+    } else if (op->call_type == Call::PureExtern) {
+        // TODO: This could probably be simplified into a single map-lookup
+        // with a bit more cleverness; not sure if the reduced lookup time
+        // would pay for itself (in comparison with the possible lost code clarity).
+
+        // Handle all the PureExtern cases of float -> bool
+        {
+            static const std::unordered_map<std::string, std::function<bool(double)>>
+                pure_externs_f1b = {
+                    {"is_finite_f16", [](double a) { return std::isfinite(a); }},
```
abadams: Can you construct a std::function<bool(double)> from std::isfinite directly without the wrapping lambda? (and similar below)

steven-johnson: Done. (Note: if approved, please don't merge yet -- I want to re-run a global reality check at Google to be sure there are no injections from this.)

PR #4522, opened 2020-01-10
Reschedule and benchmark nl_means app

abadams:
I am continuing to roll through the apps, making sure everything has a decent schedule and a benchmark on the same machine. This PR addresses the nl_means app.

### DIFF
--- a/apps/nl_means/nl_means_generator.cpp
+++ b/apps/nl_means/nl_means_generator.cpp
@@ -90,6 +90,12 @@ public:
             // 22 ms on a 2060 RTX
             Var xii, yii;
 
+            // We'll use 32x16 thread blocks throughout. This was
+            // found by just trying lots of sizes, but large thread
+            // blocks are particularly good in the blur_d stage to
+            // avoid doing wasted blurring work at tile boundaries
+            // (especially for large patch sizes).
+
             non_local_means.compute_root()
                 .reorder(c, x, y)
                 .unroll(c)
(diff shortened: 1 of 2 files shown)


apps/nl_means/nl_means_generator.cpp
```diff
@@ -86,51 +86,69 @@ class NonLocalMeans : public Halide::Generator<NonLocalMeans> {
...
-                .gpu_threads(x, y);
-            blur_d_y.compute_at(non_local_means_sum, s_dom.x)
-                .unroll(x, 2).gpu_threads(x, y);
-            blur_d.compute_at(non_local_means_sum, s_dom.x)
-                .gpu_threads(x, y);
-            non_local_means_sum.compute_at(non_local_means, x)
-                .gpu_threads(x, y)
+                .reorder(c, x, y)
+                .unroll(c)
+                .gpu_tile(x, y, xi, yi, 32, 16);
```
steven-johnson: Once again, a please for explanation of magic numbers in the comments. Future You will thank Present You.

abadams: In this case there's no meaning behind these numbers. It's just a size that worked out OK performance-wise.

steven-johnson: ...that sounds like a useful comment then.

dsharletg: Isn't that basically the "default comment" on every schedule? Schedules inherently involve a ton of these constants that are mostly just found via trial and error, or simply set arbitrarily and forgotten about. Unless there's some specific reason for a constant to be something (e.g. a tile size needs to fit in shared/local memory, satisfies some other constraint), I think it's OK to not explain schedule constants.

abadams: Probably overthinking it, but I've been internally holding these two same points of view since Steven's comment. For production code, I think a comment would be unnecessary here for the reason Dillon states. Future me is not going to be confused about tile sizes in schedules - I'll just assume that I tried a few things and this one was fast. But this is example code for people learning Halide too, and I think a comment here would add some value for those people, even if it just says "A 32x16 tile seems to work well. Found through trial and error".

steven-johnson: What he said. This is (at least partly) example code. Overdoing the explanation is arguably worthwhile in this case.

PR #4524, opened 2020-01-13
Rename/clarify the Hexagon argument 'buffer_t' type

steven-johnson:
We currently use an odd subset of the classic 'buffer_t' type in the HVX arg-passing code; despite its name, it doesn't appear to ever we used as an old 'buffer_t' (which is good, since it's too small and would likely cause crashes). As far as I can tell, it's just a two-entry struct that is used exclusively to pass arg info from host->hvx (or host->sim). Since the old buffer_t type is long-deprecated (and hopefully going to go away entirely soon), let's avoid any possible confusion by renaming it to `hexagon_arg_buffer_t` in hexagon_remove and sim_remote, and by declaring and using it in hexagon_host.cpp (rather than unpacking the fields directly).

HexagonOffload.cpp could arguably use the same struct, but I didn't bother, instead just documenting the correspondence with a comment.

### DIFF
--- a/src/runtime/hexagon_remote/halide_remote.cpp
+++ b/src/runtime/hexagon_remote/halide_remote.cpp
@@ -378,19 +378,17 @@ int halide_hexagon_remote_run_v2(handle_t module_ptr, handle_t function,
     // Get a pointer to the argv version of the pipeline.
     pipeline_argv_t pipeline = reinterpret_cast<pipeline_argv_t>(function);
 
-    // Construct a list of arguments. This is the way that HexagonOffload
-    // packages arguments for us; despite the name, it really has nothing
-    // to do with the old 'buffer_t' type (or the current 'halide_buffer_t' type).
-    struct hexagon_arg_buffer_t {
+    // Construct a list of arguments.
+    struct hexagon_device_pointer {
         uint64_t dev;
         uint8_t *host;
     };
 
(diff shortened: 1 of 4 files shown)


src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -378,36 +378,35 @@ int halide_hexagon_remote_run_v2(handle_t module_ptr, handle_t function,
...
 
-    // Construct a list of arguments. This is only part of a
-    // buffer_t. We know that the only field of buffer_t that the
-    // generated code should access is the host field (any other
-    // fields should be passed as their own scalar parameters) so we
-    // can just make this dummy buffer_t type.
-    struct buffer_t {
+    // Construct a list of arguments. This is the way that HexagonOffload
+    // packages arguments for us; despite the name, it really has nothing
+    // to do with the old 'buffer_t' type (or the current 'halide_buffer_t' type).
```
dsharletg: This comment has a lot of "history" that is irrelevant to the current code. How about just calling this struct "hexagon_device_pointer" or something like that, and removing all references to "buffer_t"? Instead there should be a comment on the code that translates buffers to pointers (and the other args).

steven-johnson: PTAL

PR #4525, opened 2020-01-13
Remove all support for the legacy `buffer_t` type

steven-johnson:


abadams:
Lots of incorrect usage of the old type in the repo revealed here. If we decide we have to hang onto the legacy support for a bit longer we should at least merge the fixes.

steven-johnson:
Urk. Want me to split this into two PRs (one to fix all existing bogus usage and comments, and another to remove the support entirely)?

abadams:
In theory it would be cleaner, but doesn't seem worth the effort. Hopefully we can just remove support entirely.

steven-johnson:
(Note that the hexagon-related 'buffer_t' type is handled separately, in https://github.com/halide/Halide/pull/4524)

steven-johnson:
Ready for review, but not yet ready to land (I'd prefer to do a little preflight work inside google first).

python_bindings/correctness/the_sort_function.c
```diff
@@ -3,72 +3,24 @@
  * gcc -std=c99 the_sort_function.c -shared -o the_sort_function.so
  */
 
+#include "HalideRuntime"
```
abadams: Does this work without the .h?

abadams:
Deleted code is best code.

python_bindings/correctness/the_sort_function.c
steven-johnson: No, but apparently it's not currently actively built and tested :-/

PR #4526, opened 2020-01-14
Use string literals instead of byte arrays for embedded blobs

abadams:
This is a PR to speed up compilation of the compiler very slightly, using a trick I read about recently (but am sadly unable to find again, or I would cite the page).

Using string literals instead of arrays for embedded blobs speeds up compilation of the runtime modules, because a string literal is a single AST node in C++, but an array initializer needs an AST node per array entry and all the compiler overhead that entails.

For our largest runtime module (d3d12compute_64_debug), compiling the generated .cpp file takes 0.98s in master and hits 205k peak memory use. In this branch it takes 0.05s and only uses 57k, so it's a 20x speedup.

Of course this is far from the long pole in compilation of the Halide compiler, so any end-to-end benefit is lost in the noise of a big parallel compilation. Might save a few seconds, or more on our less beefy buildbots.

### DIFF
--- a/tools/binary2cpp.cpp
+++ b/tools/binary2cpp.cpp
@@ -48,7 +48,7 @@ int main(int argc, const char **argv) {
         if (c == EOF) break;
         printf("\\x%02x", c);
         // Not necessary, but makes a bit easier to read
-        if (++line_break > 12) {
+        if (++line_break > 16) {
             printf("\"\n\"");
             line_break = 0;
         }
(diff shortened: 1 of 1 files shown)


abadams:
fatal error C1091: compiler limit: string exceeds 65535 bytes in length

Welp. I guess not then.

steven-johnson:
Ah yes, I remember that now. MSVC is the gift that keeps on giving.

PR #4530, opened 2020-01-16
Allow out-of-memory errors for max filter app to reduce build bot flakes

abadams:


abadams:
I opened this under the assumption that the apps tests were already benchmarking serially, but perhaps that's not the case?

abadams:
I'll open a PR that builds the apps in parallel but tests them serially instead.

PR #4531, opened 2020-01-16
Test apps serially, build them in parallel

abadams:
The idea is to reduce flakes on the buildbots from running multiple apps at once consuming too much GPU memory (or some other resource). Required adding build targets to all the app makefiles.


abadams:
Let's check how long test_apps takes on the buildbots on this PR, and compare it to master, and use that to estimate the cost of doing this.

Makefile
```diff
@@ -2039,7 +2050,7 @@ benchmark_apps: $(BENCHMARK_APPS)
 	@for APP in $(BENCHMARK_APPS); do \
 		echo ;\
 		echo Benchmarking $${APP} for ${HL_TARGET}... ; \
-		make -C $(ROOT_DIR)/apps/$${APP} \
+		make -C $(ROOT_DIR)/apps/$${APP} test \
```
steven-johnson: Why did you add this to the `benchmark_apps` target? This one was fine before.

apps/autoscheduler/Makefile
```diff
@@ -65,17 +65,16 @@ run_test: $(BIN)/$(HL_TARGET)/test
 # to avoid eventually consuming all disk space on the buildbot...)
 test: AUTOSCHED_SAMPLES_OUT = $(BIN)/test_samples_out
 
-# Note that 'make test' is used by Halide buildbots to spot-check changes,
-# so it's important to try a little of each of the important paths here,
-# including single-shot and autotune-loop
+# Note that 'make build' and 'make test' is used by Halide buildbots
+# to spot-check changes, so it's important to try a little of each of
+# the important paths here, including single-shot and autotune-loop
+build: $(BIN)/$(HL_TARGET)/test $(BIN)/test_perfect_hash_map $(BIN)/test_function_dag $(BIN)/$(HL_TARGET)/included_schedule_file.rungen $(GENERATOR_BIN)/demo.generator $(AUTOSCHED_BIN)/featurization_to_sample $(AUTOSCHED_BIN)/get_host_target $(AUTOSCHED_BIN)/retrain_cost_model $(AUTOSCHED_BIN)/libauto_schedule.so
```
steven-johnson: IMHO this would be easier to read if broken into multiple lines with \

Makefile
abadams: Accidental. Removed.

apps/autoscheduler/Makefile
abadams: Done.

Makefile
```diff
@@ -2147,7 +2151,7 @@ $(BUILD_DIR)/clang_ok:
 	@exit 1
 endif
 
-ifneq (,$(findstring $(LLVM_VERSION_TIMES_10), 80 90 100))
+ifneq (,$(findstring $(LLVM_VERSION_TIMES_10), 80 90 100 110))
```
steven-johnson: Ah yes, LLVM 10 is now at rc1, so I guess trunk is at 11?

steven-johnson:
test_apps still seems to be failing: https://buildbot.halide-lang.org/master/#/builders/15/builds/41

steven-johnson:
The mac build is failing in test_python, but in 'using_the_gpu', which seems odd.

PR #4536, opened 2020-01-17
Refactor of configure_pipeline_features() to avoid libHalide dependency

steven-johnson:
This is my attempt at an alternate version of https://github.com/halide/Halide/pull/4462 that avoids the need to add a libHalide dependency to retrain_cost_model; this achieves the same immediate goal, though I'm not sure if it has the same long-term goal. Please take a look.

### DIFF
--- a/apps/autoscheduler/AutoSchedule.cpp
+++ b/apps/autoscheduler/AutoSchedule.cpp
@@ -932,39 +932,52 @@ public:
     }
 };
 
+namespace {
+
+// Custom version of PipelineInfo to extract the info we need from
+// FunctionDAG + MachineParams for the CostModel interface.
+class FunctionDAGPipelineInfo : public PipelineInfo {
+    const FunctionDAG &dag;
+    const MachineParams &params;
+
+public:
+    FunctionDAGPipelineInfo(const FunctionDAG &d,
+                            const MachineParams &p)
+        : dag(d), params(p) {
+    }
+
+    std::vector<PipelineFeatures> pipeline_features() const override {
+        std::vector<PipelineFeatures> v;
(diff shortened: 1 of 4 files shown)


apps/autoscheduler/AutoSchedule.cpp
```diff
@@ -932,39 +932,52 @@ class StateQueue {
...
+    const FunctionDAG &dag;
+    const MachineParams &params;
+
+public:
+    FunctionDAGPipelineInfo(const FunctionDAG &d,
+                            const MachineParams &p)
+        : dag(d), params(p) {
+    }
+
+    std::vector<PipelineFeatures> pipeline_features() const override {
```
benoitsteiner: My main concern with this is that the communication between the AutoScheduler and the CostModel still happens through a vector of PipelineFeatures, which isn't sufficient for our needs. By Having the cost model take the FunctionDAG as argument, the cost model had the ability to represent the FunctionDAG as it sees fit.

A secondary concern is that the implementation of FunctionDAGPipelineInfo resides in AutoScheduler.cpp, so we can't change anything there without having to fork the AutoScheduler code. For example, if I wanted to multiply all the feature values I pass to my cost model by 2, I wouldn't be able to push that to Halide without making incompatible changes (this is obviously not a good example since I could multiply everything by 2 in a custom cost model itself)

One way to avoid this would be to extract from FunctionDAG the subsets that don't depend on the runtime, encode them as the FunctionDAGPipelineInfo, and pass the FunctionDAGPipelineInfo to the cost model. However, since the exact representation needed for accurate cost modeling is still the subject of active research, it would be hard to figure out at this point what the FunctionDAGPipelineInfo class would need to look like.

PR #4539, opened 2020-01-19
Add vcpkg installation instructions

LilyWangL:
Halide is available as a port in vcpkg, a C++ library manager that simplifies installation for Halide and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build Halide, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,18 +71,6 @@ then to point Halide to it:
 
 (Note that you *must* add `clang` to `LLVM_ENABLE_PROJECTS`; adding `lld` to `LLVM_ENABLE_PROJECTS` is only required when using WebAssembly, but we recommend enabling it in all cases, to simplify builds.)
 
-#### Building Halide - Using vcpkg
-
-You can download and install Halide using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
-
-    git clone https://github.com/Microsoft/vcpkg.git
-    cd vcpkg
-    ./bootstrap-vcpkg.sh
-    ./vcpkg integrate install
-    ./vcpkg install halide
-
-The Halide port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
-
 #### Building Halide with make
 
 With `LLVM_CONFIG` set (or `llvm-config` in your path), you should be
@@ -142,6 +130,18 @@ To configure and build Halide:
     % cd C:\Code\halide-build
     % cmake -DLLVM_DIR=../llvm-install/lib/cmake/llvm -DCMAKE_BUILD_TYPE=Release -G "Visual Studio 14 Win64" ../halide
     % MSBuild.exe /m /t:Build /p:Configuration=Release .\ALL_BUILD.vcxproj
+
+#### Building Halide - Using vcpkg
+
+You can download and install Halide using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install halide
+
+The Halide port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
 
 #### Building Halide and LLVM on Windows using mingw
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,6 +71,18 @@ then to point Halide to it:
 
 (Note that you *must* add `clang` to `LLVM_ENABLE_PROJECTS`; adding `lld` to `LLVM_ENABLE_PROJECTS` is only required when using WebAssembly, but we recommend enabling it in all cases, to simplify builds.)
 
+#### Building Halide - Using vcpkg
```
abadams: Can we nest this under "Building Halide and LLVM on Windows" below? Windows is a minority of our users, so it would be strange to put it above make and cmake.

I will test the instructions on my Windows machine tonight. I haven't used vcpkg there before so it should be a good test of a clean install.

LilyWangL: Yes, I changed it.

abadams:
Thanks for doing this! I followed the instructions in powershell and after a few failed starts got it working. The snags I hit were:

- My visual studio install (2015) didn't work, as I didn't have the ATL/MFC component. The existing error message was fine - I just got a newer visual studio and made sure to include that component. Is it the case that using vcpkg on windows will require a newer visual studio than 2015 in general? Our readme states that 2015 is sufficient, so if you need a newer one we should state that in the vcpkg section.

- vcpkg seems to default to x86-windows, which is not ideal for Halide. It would be good to add text to tell people how to list available versions and select the one they want. For me it was x64-windows

- A clean vcpgk setup with Halide installed uses 90 gigabytes! I ran out of disk space a few times before I realized what was going wrong. We should warn people to expect this, and maybe provide a way to avoid it if it's not necessary. The main culprit seems to be the llvm buildtree in debug mode. Should we tell people to clean the build tree afterwards? I'm not sure of the correct way to do that with vcpkg.

- There's a dependency on openblas, but Halide doesn't actually depend on openblas. One of the example apps uses it for some benchmarking comparisons, so I think it can be a dependency of the app target instead. The "app" feature is really example applications that use Halide, not "app support" as it is described in the port file.

- The configuration seems to default all targets to off except for the host target (x86 or arm). Most serious use of Halide is as a cross-compiler. x86,arm,aarch64,ptx,opencl,opengl,metal should all default to on. nativeclient doesn't exist as a target anymore, and can be deleted. Also, I don't think there should be any dependency on an opencl or opengl package to use the opencl or opengl targets. We don't depend on any opencl/gl headers or libraries - we talk to the driver directly. Some of the tests do though, so a dependency on opencl might be necessary for building the tests.

steven-johnson:
No rush, but what's the status of this PR? Looks like there are a number of minor changes requested.

LilyWangL:
Sorry for my late reply. I will confirm these problems and fix soon.

alexreinking:
In reference to #4890 maybe we should bump the min build in these vcpkg instructions to 2017.

alexreinking:
> * A clean vcpgk setup with Halide installed uses 90 gigabytes! I ran out of disk space a few times before I realized what was going wrong. We should warn people to expect this, and maybe provide a way to avoid it if it's not necessary. The main culprit seems to be the llvm buildtree in debug mode. Should we tell people to clean the build tree afterwards? I'm not sure of the correct way to do that with vcpkg.
> 
> * There's a dependency on openblas, but Halide doesn't actually depend on openblas. One of the example apps uses it for some benchmarking comparisons, so I think it can be a dependency of the app target instead. The "app" feature is really example applications that use Halide, not "app support" as it is described in the port file.
> 
> * The configuration seems to default all targets to off except for the host target (x86 or arm). Most serious use of Halide is as a cross-compiler. x86,arm,aarch64,ptx,opencl,opengl,metal should all default to on. nativeclient doesn't exist as a target anymore, and can be deleted. Also, I don't think there should be any dependency on an opencl or opengl package to use the opencl or opengl targets. We don't depend on any opencl/gl headers or libraries - we talk to the driver directly. Some of the tests do though, so a dependency on opencl might be necessary for building the tests.

@abadams -- all of these would be good issues to raise with the vcpkg maintainers

abadams:
I am under the impression the PR author *is* the vcpkg maintainer for Halide.

alexreinking:
Hi again @LilyWangL -- I just opened microsoft/vcpkg#13580. We would love to add vcpkg instructions to the README once the newest release is available on your platform. Thanks!

alexreinking:
@LilyWangL -- are you still involved with vcpkg / the Halide port? We are very interested in getting our new release published on vcpkg.

LilyWangL:
> @LilyWangL -- are you still involved with vcpkg / the Halide port? We are very interested in getting our new release published on vcpkg.

Yes, I just came back from vacation, I will check PR microsoft/vcpkg#13580 soon.

alexreinking:
Thanks, @LilyWangL!

PR #4540, opened 2020-01-21
Fix duplicate allocation names in gpu kernels

abadams:
This PR fixes two bugs that showed up in random fuzzing of gpu schedules in the gpu autoscheduler work. 

- Register-level allocations within a GPU kernel are hoisted outwards to the top of the kernel, which may cause them to nest with other allocations of the same name (e.g. if something was unrolled). This is usually benign - the original use of these allocations was non-overlapping, so all uses just use the innermost one. Still, it's weird. If the sizes were different somehow it could cause real bugs. This PR uniquely numbers all register-level allocations to avoid collisions.

- let stmts that are necessary for size expressions are also lifted outwards when computing shared/heap memory size outside of the kernel launch. If multiple definitions of the same name exist in non-overlapping places (again, probably due to unrolling), then they get repeatedly rewrapped around a size expression that uses it, causing the simplifier to complain later about things like "let foo = bar in let foo = bar in ... foo ..." This was fixed by only wrapping the lets that would have been in scope.


abadams:
Failure merits investigation, but can't possibly be related.

PR #4541, opened 2020-01-22
Fix for llvm trunk

abadams:


abadams:
Passing a unsigned for an alignment arg is getting deprecated. Currently only CreateMaskedLoad warns, but the others are coming. CreateMaskedStore still takes an unsigned for some reason.

PR #4542, opened 2020-01-22
Fix API breakage for trunk LLVM

steven-johnson:


abadams:
Collision with #4541 

Which do you prefer?

steven-johnson:
Yours

PR #4544, opened 2020-01-22
Fix for trunk LLVM

steven-johnson:


abadams:
Ugh, I hope this isn't a difference between llvm 10 and llvm 11. If so we might need a little class that can implicitly cast to either llvm::Align or unsigned.

PR #4546, opened 2020-01-22
Deprecate Hexagon v60 in the Makefile too

pranavb-ca:
FWIW, we use install_qc internally


pranavb-ca:
@dsharletg, @steven-johnson  - The 2 failing checks are not related, is this ok to merge?

PR #4547, opened 2020-01-22
[Hexagon/HVX] - fix vtmpy fall through

pranavb-ca:
While trying to generate vtmpy, we sort loads based on their respecive start indices. However, it doesn't see to be easy to write a comparator function that'll implement a strict weak ordering.
So, we use stable_sort instead of sort so at the very least, the relative order of tied elements in the vector to be sorted is not changed.

Also, if we cannot match some of the mpys to vtmpy, then check if we should broadcast any of them before using simple multiplies.


src/HexagonOptimize.cpp
```diff
@@ -2046,8 +2049,19 @@ class VtmpyGenerator : public IRGraphMutator {
...
                         }
-                        Expr mpy_a = lossless_cast(op->type, mpys[i].first);
-                        Expr mpy_b = lossless_cast(op->type, mpys[i].second);
+                        // We put expressions in mpys after un-broadcasting them. So, first broadcast
+                        // then call lossless_cast.
+                        Expr a = mpys[i].first;
+                        Expr b = mpys[i].second;
+                        int lanes = op->type.lanes();
+
+                        if (a.type().lanes() < lanes)
```
dsharletg: I think this should be a.type().lanes() == 1. Broadcast only makes sense for lanes == 1, and maybe there are some ops (shuffles) that would pass this check but break when broadcasted.

Same for b below.

abadams: Nit: a.type().is_scalar()

pranavb-ca: Done.

pranavb-ca:
@dsharletg : apps/hexagon_benchmarks/conv3x3a32 was triggering asserts because of the lack of broadcast. What's happening here is that the part where we need broadcast is the part that handles expressions that didnt match vtmpy. This path wasn't tested/exercised before. However, my guess is that difference between test/correctness/convolution.cpp and apps/hexagon_benchmarks/conv3x3a32 is large enough that the former doesnt even exercise this larger path (vtmpy generation).

steven-johnson:
Is this ready to land, or is more work needed?

steven-johnson:
This has approval -- is it ready to land?

steven-johnson:
This PR has been sitting 'approved' for nearly a month now. Is it, in fact, ready to land?

PR #4551, opened 2020-01-23
Just assert not device dirty if no device enabled, don't copy to host

abadams:
We recently made all pipelines do a copy to host for all inputs and outputs, even with no device support. This was more expensive than we realized. A call to copy-to-host is not only a function call, but the body of the function grabs a global mutex. If you have a Halide-generated tiny kernel (say an 8x8 fft) being called inside some outer non-Halide parallel loop, or as an extern stage, there's going to be a lot of pointless contention on this mutex.

This PRs proposes just backing off from our earlier strategy and just asserting !device_dirty if a pipeline was compiled without device support instead of doing a copy-to-host automatically. If you want an automatic copy-to-host then enable a device flag in your target.

I replaced the defunct divide-by-zero error with a new one for this assert.


src/AddImageChecks.cpp
```diff
@@ -577,17 +591,22 @@ Stmt add_image_checks(Stmt s,
         }
     }
 
-    // Inject the code that checks the host pointers.
     if (!no_asserts) {
+        // Inject the code that checks the host pointers.
         for (size_t i = asserts_host_non_null.size(); i > 0; i--) {
```
steven-johnson: nit: is is worth adding an "append vector<Stmt> in reverse order" utility?

abadams: Done

src/Lower.cpp
```diff
@@ -156,10 +156,17 @@ Module lower(const vector<Function> &output_funcs,
     debug(1) << "Computing bounds of each function's value\n";
     FuncValueBounds func_bounds = compute_function_value_bounds(order, env);
 
+    bool will_inject_host_copies =
```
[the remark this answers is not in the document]
abadams: Hrm. I'm not sure where elsewhere needs this. If it's a lowering pass we can pass the flag in.

src/Lower.cpp
```diff
@@ -156,10 +156,17 @@ Module lower(const vector<Function> &output_funcs,
...
+        (t.has_gpu_feature() ||
```
dsharletg: A helper function `Target::has_device_interface` would be nice, at least to wrap up the weird Hexagon logic.

steven-johnson:
Failure is unrelated. LGTM.

PR #4552, opened 2020-01-24
Add support and tests for pmulhrsw ssse3 and avx2 x86 instructions

dzo:
The tests are:
  correctness_simd_op_check pmulhrsw
  correctness_simd_op_check vpmulhrsw*ymm

### DIFF
--- a/src/CodeGen_X86.cpp
+++ b/src/CodeGen_X86.cpp
@@ -283,7 +283,7 @@ void CodeGen_X86::visit(const Cast *op) {
          u16((wild_u32x_ * wild_u32x_) / 65536)},
         {Target::AVX2, true, Int(16, 16), 9, "llvm.x86.avx2.pmul.hr.sw",
          i16((((wild_i32x_ * wild_i32x_) + 16384)) / 32768)},
- 
+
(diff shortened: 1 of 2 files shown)


abadams:
Thanks! But I'm confused by https://www.felixcloutier.com/x86/pmulhrsw, which seems to say the divisor is 2^14, not 2^15. Are you sure those patterns are right?

dzo:
> Thanks! But I'm confused by https://www.felixcloutier.com/x86/pmulhrsw, which seems to say the divisor is 2^14, not 2^15. Are you sure those patterns are right?

Yes, if you look at that page it does:
temp0[31:0] = INT32 ((DEST[15:0] * SRC[15:0]) >>14) + 1;
DEST[15:0] = temp0[16:1];

i.e. i16((((i32(i16_1) * i32(i16_2))>>14)+1)>>1) 
Which should be the same as i16((((i32(i16_1) * i32(i16_2)) + (1<<14)) >>15)

It works for my application and gives exactly the same result as the ARM sqrdmulh instruction (without saturating arithmetic though)

abadams:
Ah, I missed that. Thanks for the clarification.

Issue #4424, opened 2019-11-21
CMake builds should probably use CTest instead of hand-rolled test rules


steven-johnson:
CMake has some build-in support for building/running tests (https://cmake.org/cmake/help/latest/manual/ctest.1.html); I wasn't aware of this when I wrote the stuff we have now, and testing might be simpler and cleaner if we just used CTest instead. Not sure how much work it would be.

steven-johnson:
(To be clear: it sorta-kinda does use CTest, but not in a very useful way.)

alexreinking:
Closed by #4556

PR #4556, opened 2020-01-27
Update CMake to use modern features in testing.

alexreinking:
This is something of a big PR, but it only touches the CMake build, with one notable exception: `apps/support/cmdline.h` has been modified to (crudely) support non-RTTI builds.

The main contribution is this: pseudo-targets for running tests have been removed in favor of using CTest, which has a number of advantages. Notably, it has a native notion of test _labels_, which allow us to define and select groups of tests to run. It also allows developers to run only those tests which failed previously; this is especially useful after an incremental build.

The CMake minimum version requirement has been bumped from 3.3 to 3.14. A _lot_ has changed in CMake in recent years and an initial effort has been made to modernize. For instance, the `halide_use_image_io` function has been replaced with an alias target `Halide::ImageIO` to which a normal target can link. A subsequent PR will focus on modernizing the rest of the CMake build so we can be more easily integrated into other projects, packaged by popular package managers, and installed into standard system locations.

A large number of new lines list source files. Globbing source files in CMake breaks incremental builds and leads to frustrating scenarios where the lack of changes to CMakeLists after a pull requires a developer to do a _full_ rebuild. The CMake [developers strongly caution against source file globbing](https://cmake.org/cmake/help/latest/command/file.html?highlight=file%20glob#glob). It is the [prevailing opinion on StackOverflow](https://stackoverflow.com/q/1027247/2137996) and in talks from the maintainers.

----

Expect build failures while I learn how Travis works. Will also need to ask about the buildbots.


CMakeLists.txt
```diff
@@ -473,6 +414,7 @@ endif()
 # ------------------------------------------------
 # install
 
+# TODO: the use of glob here is almost certainly brittle if not outright wrong.
```
abadams: These are not build deps. How is globbing more brittle that explicitly listing the files, and then forgetting to update the list when a new HalideRuntimeFoo.h is added?

CMakeLists.txt
```diff
@@ -494,32 +436,33 @@ install(DIRECTORY tutorial
...
 
 # ---- README
-file(GLOB FILES "${HALIDE_BASE_DIR}/*.md")
-install(FILES ${FILES}
+install(FILES
+        CODE_OF_CONDUCT.md
+        README_cmake.md
+        README.md
+        README_rungen.md
+        README_webassembly.md
```
abadams: We're going to forget to add new files to this list every single time.

alexreinking: It might be fine to glob for non-source files, since it won't break incremental building. That said, the CMakeLists for src/ already use explicit file lists.

CMakeLists.txt
alexreinking: Because file(GLOB) runs at build generation time and the files in "${CMAKE_BINARY_DIR}/include/" don't exist until after the build has run. I'm actually surprised this works at all.

abadams:
ok, so the explicit file lists are a problem, because people (me) going to forget to add new files to them. I add new tests frequently, and don't use the cmake build.

But I did read the docs and stack overflow things you linked.

We could add a presubmit check that every source file in a list of directories is mentioned explicitly in the corresponding CMakeLists file?

CMakeLists.txt
abadams: I remember to do that because the Makefile uses an explicit list there too, and I've been disciplined about it. Better if we can enforce this automatically though.

CMakeLists.txt
abadams: Ooh, I didn't catch that. We should just list them out here then, if that's the path we're going down (or at least update the comment it include what you said).

alexreinking: Ah, apparently the CMakeLists.txt in src/ copies them to the build output directory at configure time. There's no reason to do this. We can issue the relevant install directives directly from there, remove the glob here, and end up with less CMake code overall.

alexreinking:
> ok, so the explicit file lists are a problem, because people (me) going to forget to add new files to them. I add new tests frequently, and don't use the cmake build.
> 
> But I did read the docs and stack overflow things you linked.
> 
> We could add a presubmit check that every source file in a list of directories is mentioned explicitly in the corresponding CMakeLists file?

That seems like a good compromise.

alexreinking:
I added the presubmit check. Make sure the file lists are in alphabetical order.

CMakeLists.txt
```diff
@@ -370,56 +350,17 @@ function(add_halide_test TARGET)
...
+          COMMAND ${TARGET}
+          WORKING_DIRECTORY "${args_WORKING_DIRECTORY}")
 
+  set_tests_properties(${TARGET} PROPERTIES LABELS "${args_GROUPS}")
+  if (${args_EXPECT_FAILURE})
+    set_tests_properties(${TARGET} PROPERTIES WILL_FAIL true)
+  endif ()
 endfunction()
 
-# Only groups that are defined in *this file* will have projects reliably
```
abadams: Does the CTest stuff mean this comment is no longer a concern?

alexreinking: I assume so.

alexreinking: Reason being that these commands were creating targets at the root of the project, which are visible to CMakeLists in subdirectories. CTest handles all that sort of thing for you.

alexreinking:
Gonna need to reconfigure the build bots for the CMake build.

steven-johnson:
Buildbots will need to be uniformly upgraded to CMake 3.14 (they are a mishmash of versions). I can get to that today.

CMakeLists.txt
steven-johnson: > We're going to forget to add new files to this list every single time.

Not if we all have the discipline to test with CMake as well before pushing :-)

CMakeLists.txt
steven-johnson: > I add new tests frequently, and don't use the cmake build

As a thought experiment: what would it take to get all the Makefile holdouts (myself included) to commit to using CMake regularly (or at least to get in the habit of building it before PRs)?

CMakeLists.txt
steven-johnson: I didn't know CTest existed back when I wrote that explicit test stuff.

CMakeLists.txt
abadams: Maybe switching the buildbots to it?

alexreinking:
> 
> 
> Buildbots will need to be uniformly upgraded to CMake 3.14 (they are a mishmash of versions). I can get to that today.

I would just update them to whatever the latest available version newer than 3.14 is. Version 3.16 is what's actually executing on Travis, and what I'm using to test locally. These days you can install the latest CMake via either Kitware's APT repo (see the travis.yml file) or, in fact, pip.

CMakeLists.txt
alexreinking: > I didn't know CTest existed back when I wrote that explicit test stuff.

If you were actually on version 3.3, it might not have!

alexreinking:
One more todo: we can use the Ninja-build generator on Travis instead of the Makefile generator. It's a good bit faster on my computer.

alexreinking:
One MORE todo, but not blocking this PR. The build is broken on windows if your source or build dir has a space in the path. Eg. "C:\Users\Alex Reinking"... fml why did Windows default to that?

steven-johnson:
> why did Windows

Just leave it like that. It's cleaner.

test/common/expect_abort.cpp
```diff
@@ -0,0 +1,15 @@
...
+    std::_Exit(EXIT_FAILURE);
+}
+
+struct hl_override_abort {
+    hl_override_abort() noexcept {
+        std::signal(SIGABRT, hl_error_test_handle_abort);
+    }
+};
+
+hl_override_abort handler{};
```
steven-johnson: nit, here and elsewhere: please end source files with an EOL

test/generator/CMakeLists.txt
```diff
@@ -0,0 +1,301 @@
...
+    target_link_libraries("${TARGET}" PRIVATE "${NAME}.generator")
+    add_halide_test("${TARGET}"
+            GROUPS generator travis)
+endforeach ()
+
+# ------ Generator tests for ahead-of-time mode: ------
+# Find all the files of form "foo_aottest.cpp" in test/generator
+# and declare a rule for them, each of which depends on foo.generator.
+# Note that there are many special cases.
+# TODO(srj)
```
steven-johnson: ...is this still a WIP?

tutorial/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 include(CMakeParseArguments)
 
-# Detect if we need supress -Wunused-but-set-variable warning.  This warning
+# Detect if we need suppress -Wunused-but-set-variable warning.  This warning
```
steven-johnson: to suppress

steven-johnson:
All the buildbots should be at CMake 3.14+ now. (Actually 3.15+ I think)

alexreinking:
> > why did Windows
> Just leave it like that. It's cleaner.

Eh, the original point of the CMake build was to support Windows. There's no reason not to support spaces in paths if it's not a nightmare to do.

halide.cmake
```diff
@@ -15,18 +15,22 @@ cmake_minimum_required(VERSION 3.3)
...
 
 # Add the include paths and link dependencies for halide_image_io.
-function(halide_use_image_io TARGET)
-  foreach(PKG PNG JPEG)
-    find_package(${PKG} QUIET)
-    if(${PKG}_FOUND)
-      target_compile_definitions(${TARGET} PRIVATE ${${PKG}_DEFINITIONS})
-      target_include_directories(${TARGET} PRIVATE ${${PKG}_INCLUDE_DIRS})
-      target_link_libraries(${TARGET} PRIVATE ${${PKG}_LIBRARIES})
+add_library(halide_image_io INTERFACE)
```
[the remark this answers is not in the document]
alexreinking: No, that should be bumped.

.travis.yml
```diff
@@ -26,18 +35,14 @@ env:
     # Just one version of LLVM8 should suffice.
     - LLVM_VERSION=8.0.0 BUILD_SYSTEM=MAKE
 
-cache: apt ccache
-dist: xenial
+cache: ccache
+dist: bionic
 install:
   # Travis Xenial uses gcc5.4 / cmake 3.12.4 by default, which are fine
```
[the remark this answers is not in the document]
alexreinking: Will do. Need bionic for newer cmake.

alexreinking:
>  Also probably an update to the README for how to build/run tests for both Makefile and CMake would be useful, since they now differ more.

Absolutely. Will do.

test/generator/CMakeLists.txt
alexreinking: Unclear. I haven't really touched the generator files. To do so would require me to first update Halide's cmake rules (halide.cmake).

halide.cmake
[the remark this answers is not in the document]
steven-johnson: Nit: Do we have any meaningful concern about people being unwilling to upgrade CMake? (LLVM only requires 3.4.3, amazingly...)

test/generator/CMakeLists.txt
steven-johnson: Gotcha. (Obviously we can't land this until we're sure the Generator rules and tests are good to go.)

halide.cmake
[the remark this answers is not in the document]
alexreinking: No, I don't think we do. The latest version can be installed locally with pip, no need for a system-wide install. VS2019 comes bundled with 3.15. Older VS users can install locally and generate vcxproj files instead of using built-in support, as they always have.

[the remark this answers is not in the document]
alexreinking: I'm not aware of any backwards compatibility issues with CMake. To their credit, cmake is pretty much infinitely backwards compatible.

test/generator/CMakeLists.txt
alexreinking: I might not have been clear -- the generator rules and tests have already run successfully on Travis. I just haven't modified the build rules much. that TODO existed when I started on this project.

test/common/expect_abort.cpp
alexreinking: Does clang-format have support for that check?

test/CMakeLists.txt
```diff
@@ -1,408 +1,90 @@
...
-  foreach(FILE ${JITTEST_SRCS})
-    string(REPLACE "_jittest.cpp" "" NAME "${FILE}")
-    set(TARGET "generator_jit_${NAME}")
-    halide_project("${TARGET}" "generator" "${GEN_TEST_DIR}/${FILE}")
-    target_link_libraries("${TARGET}" PRIVATE "${NAME}.generator")
-    add_halide_test("${TARGET}"
-             GROUPS test_generator run_tests)
-  endforeach()
-
-  # ------ Generator tests for ahead-of-time mode: ------
```
alexreinking: @steven-johnson  this was commented out to begin with.

test/CMakeLists.txt
```diff
@@ -1,408 +1,90 @@
...
-  #                            GENERATOR_ARGS "${args_GENERATOR_ARGS}"
-  #                            FILTER_DEPS "${args_FILTER_DEPS}")
-  #     target_link_libraries("generator_aot_${NAME}" PUBLIC "${NAME}")
-  #   endif()
-  # endforeach()
-
-  # Create tests for ahead of-time-compiled generators. This will produce two
-  # executables, one containing the generator itself
-  # (e.g. from example_generator.cpp) and used at build time, and the other, the
-  # test that executes the generated code (e.g. from example_aottest.cpp).
```
alexreinking: @steven-johnson This was commented out to begin with.

steven-johnson: ah, ok

steven-johnson:
> It's a good bit faster on my computer.

I've heard this many times, but literally every time I've tried benchmarking ninja vs makefile on a linux or osx box, it's been in the noise range of difference. I'm puzzled as to what I've been doing wrong.

abadams:
IIRC I saw an improvement on windows, because make touches the file system too much.

alexreinking:
The buildbots need to be updated to run `ctest -L correctness`, `ctest -L generator`, and `ctest -L performance` instead of "compiling" the pseudo-target test vcxproj-es

steven-johnson:
> @steven-johnson is this config stored somewhere in the repo?

Unfortunately the buildbot is in a separate repo entirely (https://github.com/halide/build_bot); there is no staging server and no way to push a Halide PR and a buildbot PR simultaneously. 

For this case we can either push the change to the buildbots first (allowing us to verify this change but breaking everything until it lands) or land this as-is, hoping we can fix any issues that turn up later. Neither of these are good options but that's what we have. Which would you prefer?

alexreinking:
I think we should change the buildbots first, then quickly get this PR finalized, if you and @abadams are comfortable with the changes so far. Forthcoming changes to the CMake build shouldn't require so many changes to the build infrastructure. Getting all of this set up seems most important.

Would also be a good idea to have the Linux targets run CMake/CTest, too.

steven-johnson:
> Would also be a good idea to have the Linux targets run CMake/CTest, too.

I agree, but they are already pretty overloaded. If I ever get the chance to get back to https://github.com/halide/Halide/pull/4500 (using GitHub Actions for build/test) hopefully we can offload some of that onto them.

alexreinking:
Glad to know my CMake file lists check is working!

steven-johnson:
On my Mac, when I pull this and try to build it fails with `LLVM version must be 10.0 or older`, but I am using trunk LLVM:

```
$ $LLVM_CONFIG --version
11.0.0git
```

alexreinking:
Btw, this closes #4424

alexreinking:
> So it's finding the right thing... I guess LLVM_VERSION isn't being set based on LLVM_CONFIG. (FYI, we never required LLVM_VERSION to be explicitly set on the build line before, so something has changed.)

I'm 90% sure that LLVM_VERSION is being set in LLVMConfig.cmake, which is included in the LLVM repos. If something has changed in version 11, we'll have to take a look at adapting our build to that.

steven-johnson:
> > So it's finding the right thing... I guess LLVM_VERSION isn't being set based on LLVM_CONFIG. (FYI, we never required LLVM_VERSION to be explicitly set on the build line before, so something has changed.)
> 
> I'm 90% sure that LLVM_VERSION is being set in LLVMConfig.cmake, which is included in the LLVM repos. If something has changed in version 11, we'll have to take a look at adapting our build to that.

Problem turned out to be that the CMake build assumed 10 was the max version :-)

steven-johnson:
OK, so `ctest -L correctness` (etc) works correctly for me on OSX... but on our Windows buildbots, it consistently fails to do anything, with `No tests were found!!!`.

alexreinking:
> 
> 
> OK, so `ctest -L correctness` (etc) works correctly for me on OSX... but on our Windows buildbots, it consistently fails to do anything, with `No tests were found!!!`.

Hmm. I can't repro. What directory is that running from? I get:

```
C:\Users\Alex Reinking\source\repos\halide\Halide\out\build\x64-Release>ctest -L correctness
Test project C:/Users/Alex Reinking/source/repos/halide/Halide/out/build/x64-Release
        Start  18: correctness_align_bounds
  1/318 Test  #18: correctness_align_bounds .................................   Passed    1.09 sec
        Start  19: correctness_argmax
  2/318 Test  #19: correctness_argmax .......................................   Passed    1.05 sec
        Start  20: correctness_assertion_failure_in_parallel_for
  3/318 Test  #20: correctness_assertion_failure_in_parallel_for ............   Passed    0.65 sec
        Start  21: correctness_async_copy_chain
  4/318 Test  #21: correctness_async_copy_chain .............................   Passed    1.30 sec
        Start  22: correctness_async
^C
```

steven-johnson:
It's from the build directory. But in the MinGW shell.

alexreinking:
Is the entire build done in the MinGW shell?

steven-johnson:
> Is the entire build done in the MinGW shell?

Yes.

alexreinking:
@steven-johnson -- Figured it out. ctest defaults to the debug configuration, which actually matters when using multi-configuration generators like Visual Studio or Xcode. You need to specify which set of tests to run via:
```
ctest -C Release -j 16 -L correctness
```

alexreinking:
linux-64-gcc53-testbranch-900 is failing a (likely flaky) performance test.

```
cd bin/build/tmp ; /home/halidenightly/build_bot_new/worker/linux-64-gcc53-testbranch-900/halide-build/bin/performance_boundary_conditions
unbounded           : 3496.999556 us
constant_exterior   : 3473.564500 us
repeat_edge         : 3472.768200 us
repeat_image        : 3622.762278 us
mirror_image        : 3633.548333 us
mirror_interior     : 3609.099056 us
unbounded           : 42553.048000 us
constant_exterior   : 47084.469000 us
repeat_edge         : 46546.262000 us
repeat_image        : 94008.252000 us
Error: repeat_image is 2.209201 times slower than unbounded
../halide/Makefile:1891: recipe for target 'performance_boundary_conditions' failed
make: *** [performance_boundary_conditions] Error 255
```

abadams:
Is the performance test failure related?

abadams:
The performance tests running in parallel change with ctest was windows only, right?

alexreinking:
> 
> 
> The performance tests running in parallel change with ctest was windows only, right?

Yes and as of this morning, Windows doesn't run the performance tests in parallel.

alexreinking:
Make build with LLVM trunk is failing correctness_math. I doubt it's related to this PR

abadams:
Yes, that's also a known issue we haven't figured out yet. Feel free to merge once the builds are green except for the flake and the math failure.

alexreinking:
One compelling reason to use CMake on the buildbots: CTest will continue to run tests even after one fails, enabling us to catch multiple failures at a time.

abadams:
I mean, we can tune make to do that too, it just never occured to me as something to tune. If all tests are going to fail, you should kill the build and let the buildbot do something else sooner rather than later.... maybe you want to stop after N failures for some small value of N?

steven-johnson:
> Make build with LLVM trunk is failing correctness_math. I doubt it's related to this PR

https://github.com/halide/Halide/issues/4559

alexreinking:
No unexpected failures. Time to land!

PR #4565, opened 2020-01-29
Fix of "jpeglib.h" vs. <windows.h> issue

### LINKED ISSUE #4564: Windows build "jpeglib.h" vs. <windows.h>
I tried to build Halide with the latest commit under Windows and have the following problem with these projects:
1. `tutorial\lesson_12_using_the_gpu.vcxproj`
2. `test\correctness\correctness_image_io.vcxproj`

```
C:\Program Files (x86)\Windows Kits\8.1\Include\shared\basetsd.h(72):
error C2371: 'INT32': redefinition; different basic types  
```

Going further into the rabbit hole we will see the following:
```
E:\github\libjpeg\9c\install\include\jmorecfg.h(216):
note: see declaration of 'INT32' (TaskId:97)
```

It looks to me if we want to use both `jpeglib.h` and `windows.h` then we should include `windows.h` first and `jpeglib.h` secondarily.

We do quite the opposite.

ghost-luck:
This commit fixes #4564

### DIFF
--- a/test/correctness/image_io.cpp
+++ b/test/correctness/image_io.cpp
@@ -1,8 +1,10 @@
 #include <fstream>
 
+#include "test/common/halide_test_dirs.h"
+//
+
 #include "Halide.h"
 #include "halide_image_io.h"
-#include "test/common/halide_test_dirs.h"
 
 using namespace Halide;
 
--- a/test/correctness/image_io.cpp
+++ b/test/correctness/image_io.cpp
@@ -1,10 +1,8 @@
 #include <fstream>
 
-#include "test/common/halide_test_dirs.h"
-//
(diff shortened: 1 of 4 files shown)


test/correctness/image_io.cpp
```diff
@@ -1,8 +1,10 @@
 #include <fstream>
 
+#include "test/common/halide_test_dirs.h"
```
steven-johnson: This will not pass our lint rules; #include files should be in alphabetical order; this implies that they must not be order-dependent.

ghost-luck: Yep, it was ugly. Now it’s a little better.

Issue #3737, opened 2019-03-18
print_loop_nest on compute_with Funcs does not work


dsharletg:
For example:

    Func output_y, output_u, output_v;
    // ...

    output_u.tile(x, y, x_outer, y_outer, x_inner, y_inner, vector_width, 1);
    output_v.tile(x, y, x_outer, y_outer, x_inner, y_inner, vector_width, 1);
    output_y.tile(x, y, x_outer, y_outer, x_inner, y_inner, vector_width * 2, 2);

    output_u.compute_with(output_y, x_outer);
    output_v.compute_with(output_y, x_outer);

    output_u.print_loop_nest();

Produces:

    Error:
    Illegal compute_with: "output_u" is scheduled to be computed with "output_y" which is not 
    used anywhere.

This kind of makes sense: output_u and output_v don't have their own loop nests anymore, and so they can't be printed. But, output_y.print_loop_nest does not print output_u and output_v's loops, so it is impossible to print output_u and output_v's loop nests. This seems like an issue/bug.

PR #4566, opened 2020-01-29
print_loop_nest for the schedule with compute_with

vksnk:
This is in some way related to the #3737. 

I've been experimenting with compute_with and wanted to print loop nest, however, similarly to what described in the issue above, behavior is not obvious. For example, If you have the following schedule (taken from the #3737):

```
Func output_y, output_u, output_v;
// ...

output_u.tile(x, y, x_outer, y_outer, x_inner, y_inner, vector_width, 1);
output_v.tile(x, y, x_outer, y_outer, x_inner, y_inner, vector_width, 1);
output_y.tile(x, y, x_outer, y_outer, x_inner, y_inner, vector_width * 2, 2);

output_u.compute_with(output_y, x_outer);
output_v.compute_with(output_y, x_outer);

```

but instead of calling print_loop_nest on one of the functions, wrap them into the pipeline like so:

```
Pipeline p({output_y, output_u, output_v});
p.print_loop_nest();
```

it will print the following:

```
produce output_y:
  produce output_v:
    produce output_u:
      for y.fused.y_outer:
        for x.fused.x_outer:
          for y.y_inner in [0, 1]:
            for x.x_inner in [0, 7]:
              output_y(...) = ...
          for y.fused.y_outer:
            for x.fused.x_outer:
              for y.y_inner in [0, 0]:
                for x.x_inner in [0, 3]:
                  output_v(...) = ...
          for y.fused.y_outer:
            for x.fused.x_outer:
              for y.y_inner in [0, 0]:
                for x.x_inner in [0, 3]:
                  output_u(...) = ...
```

which, I believe, is correct representation of the underlying IR, but a little bit unexpected (?) given that y.fused.y_outer/x.fused.x_outer are printed twice for the functions scheduled with compute_with. I've checked and these loops will be trimmed off during later stages of the lowering, but that part is not included in the PrintLoopNest.

This PR, basically, adds a tracking of loop variables as we go through them and skips print if we've seen this var already. It's kind of hack, but I've checked how compute_with functions are scheduled and it seems that this should be correct, because there is a requirement of outer dims to match. I am not sure if there are any other cases beside compute_with when matching var names can happen. though.

After the changes, output will look like:

```
produce output_y:
  produce output_v:
    produce output_u:
      for y.fused.y_outer:
        for x.fused.x_outer:
          for y.y_inner in [0, 1]:
            for x.x_inner in [0, 7]:
              output_y(...) = ...
          for y.y_inner in [0, 0]:
            for x.x_inner in [0, 3]:
              output_v(...) = ...
          for y.y_inner in [0, 0]:
            for x.x_inner in [0, 3]:
              output_u(...) = ...
```


dsharletg:
This PR is indicating a bit of an issue, it is unfortunate that we have to replicate logic about loop fusion in two places.

I wonder if we should change print_loop_nest to actually perform part of lowering (maybe up through the first simplification?) and then print the actual IR, omitting most of it? This would avoid replicating such significant logic as this.

In the mean time, this PR seems OK to me.

vksnk:
> I wonder if we should change print_loop_nest to actually perform part of lowering (maybe up through the first simplification?) and then print the actual IR, omitting most of it? This would avoid replicating such significant logic as this.

I agree and did try that at first by adding some of the later lowering stages (basically, making it quite similar to Lower.cpp), but PrintLoopNest kind of doesn't work after that, because it relies on Provide which is gone by then (maybe it's easy to fix it by making PrintLoopNest to do it in some other way, but I am not sure, suggestions are appreciated).

dsharletg:
Provides are replaced with Stores by storage flattening, so I think we could avoid that by getting the loop nest IR from before storage flattening.

vksnk:
I was going to say that I tried to include everything up to storage folding and loops were not removed quite yet, but I tried it again and it does remove loops indeed, so I guess I made a mistake before.

Now the output looks like:

```
produce output_y:
  produce output_v:
    produce output_u:
      for y.fused.y_outer:
        for x.fused.x_outer:
          for y.y_inner in [0, 1]:
            for x.x_inner in [0, 7]:
              output_y(...) = ...
          for x.x_inner in [0, 3]:
            output_v(...) = ...
          for x.x_inner in [0, 3]:
            output_u(...) = ...
```

One thing to note is that "for y.y_inner in [0, 0]:" loop was removed as well, but I don't know if it's bad thing or good. 

PTAL.

steven-johnson:
Is this ready to land, or is more work needed?

vksnk:
This is ready for review.

steven-johnson:
Reviewers: please review :-)

alexreinking:
Generally speaking, we should consider having golden outputs for some of these tests unless all we care about is truly that calling the function doesn't crash. But I'm not about to block this PR on it.

Looks good for now, though it will probably change when I (or someone else) loop(s) back and fixes compute_with

vksnk:
Thanks for reviewing.

What do you mean by fixing compute_with?  Are there any specific issues?

alexreinking:
@vksnk -- tons. It's broken to fuse two independent update stages (or any two stages without fusing all previous stages). There's a project board for Halide's semantics that tracks some relevant issues.

vksnk:
Thanks, I looked up the project board: seems like the main thing, as you ssaid, is compute_with for update stages. That being said, changes in this PR  
use the same set of lowering passes, where all the work is done, so I don't think it will need a lot of extra handling. In any case, I hope  it won't require too many changes, once your fixes land, but if you have any suggestions for now, I'll be happy to incorporate them to avoid extra work in the future.

alexreinking:
No, no. We can go ahead and merge

vksnk:
Thanks again!

Issue #4559, opened 2020-01-27
correctness_math is broken for exp() on pre-AVX2 x86-64


steven-johnson:
Based on experimentation, it appears that the Halide-vs-C comparison for exp() breaks when HL_JIT_TARGET doesn't include avx2. 

~My guess here is that the "built-in" exp() we're using is now using avx2 when possible and getting a different result thanks to FMA, but I haven't yet proven this.~ This is definitely not the case, since the failure is happening on buildbots which don't have AVX2 at all. Note that on my local boxes, it fails for HL_JIT_TARGET=x86-64-linux[-sse41[-avx]] (but works for avx2 or host or anything that includes avx2).

Not sure how/why the change in behavior began -- ie, whether it was an LLVM change or a Halide change.

steven-johnson:
Injection point appears to be LLVM commit 3180af4362be22d416464f5f3700c456b2f124b9:

```
commit 3180af4362be22d416464f5f3700c456b2f124b9
Author: Sanjay Patel <spatel@rotateright.com>
Date:   Wed Jan 15 08:23:46 2020 -0500

    [InstCombine] reassociate fsub+fsub into fsub+fadd

    As discussed in the motivating PR44509:
    https://bugs.llvm.org/show_bug.cgi?id=44509

    ...we can end up with worse code using fast-math than without.
    This is because the reassociate pass greedily transforms fsub
    into fneg/fadd and apparently (based on the regression tests
    seen here) expects instcombine to clean that up if it wasn't
    profitable. But we were missing this fold:

    (X - Y) - Z --> X - (Y + Z)

    There's another, more specific case that I think we should
    handle as shown in the "fake" fneg test (but missed with a real
    fneg), but that's another patch. That may be tricky to get
    right without conflicting with existing transforms for fneg.

    Differential Revision: https://reviews.llvm.org/D72521
```

PR #4568, opened 2020-01-30
Makefile: support standard $LDFLAGS, $CXXFLAGS environment variables

minrk:
used by build tools like conda-build to make sure the right flags are passed.


steven-johnson:
Breakage is due to https://github.com/halide/Halide/issues/4559

Issue #4574, opened 2020-01-31
Halide build does not properly package LLVM objects on macOS


shoaibkamil:
Our process for grabbing the objects from the LLVM static libraries does not correctly construct the command for some objects, resulting in not all objects being included in libHalide.a.

The below is on Xcode 11.3, macOS 10.15.2, LLVM 9.0.1 compiled from source:

```
$ g++ lesson_01*.cpp -g -I ../include -L ../lib -lHalide -o lesson_01 -std=c++11
Undefined symbols for architecture x86_64:
  "llvm::DAGTypeLegalizer::SplitRes_UNDEF(llvm::SDNode*, llvm::SDValue&, llvm::SDValue&)", referenced from:
      llvm::DAGTypeLegalizer::ExpandFloatResult(llvm::SDNode*, unsigned int) in libHalide.a(llvm_1344_LegalizeFloatTypes.cpp.o)
      llvm::DAGTypeLegalizer::SplitVectorResult(llvm::SDNode*, unsigned int) in libHalide.a(llvm_1345_LegalizeVectorTypes.cpp.o)
      llvm::DAGTypeLegalizer::ExpandIntegerResult(llvm::SDNode*, unsigned int) in libHalide.a(llvm_1346_LegalizeIntegerTypes.cpp.o)
  "llvm::DAGTypeLegalizer::ExpandRes_VAARG(llvm::SDNode*, llvm::SDValue&, llvm::SDValue&)", referenced from:
      llvm::DAGTypeLegalizer::ExpandFloatResult(llvm::SDNode*, unsigned int) in libHalide.a(llvm_1344_LegalizeFloatTypes
<many more similar errors>
```

shoaibkamil:
I attempted the following fix:

```
diff --git a/Makefile b/Makefile
index d7c9b0f70..b17efec96 100644
--- a/Makefile
+++ b/Makefile
@@ -964,11 +964,12 @@ $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
        # object files in which archives it uses to resolve
        # symbols. We only care about the libLLVM ones.
        @mkdir -p $(@D)
[...]

PR #4576, opened 2020-02-03
Fix macOS static library issues

shoaibkamil:
An attempt to address #4574.

This uses map files from the linker to determine the object files needed for
the static library,instead of the output from -t, which was being truncated on
macOS, likely due to the linker not flushing standard output.

It's possible some versions of the gnu tools require `-Map` instead of `-map`, but both seem to work on my system.

If this works for Makefiles, then a remaining todo is to apply this to cmake as well.

CC @mikewoodworth

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -962,9 +962,9 @@ all: distrib test_internal
 # we need a different invocation to get the
 # linker map file.
 ifeq ($(UNAME), Darwin)
-    MAP_FLAG=-map
+    MAP_FLAGS= -Wl,-map -Wl,$(BUILD_DIR)/llvm_objects/list.all
 else
-    MAP_FLAG=-Map
+    MAP_FLAGS= -Wl,-Map=$(BUILD_DIR)/llvm_objects/list.all
 endif
 
 $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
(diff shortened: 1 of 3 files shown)


steven-johnson:
Out of curiosity: are we having the same failures with using the CMake build with shared_library=off?

shoaibkamil:
> Out of curiosity: are we having the same failures with using the CMake build with shared_library=off?

Given the comments in src/CMakeLists.txt, I think we don't package up the LLVM dependencies into the static library as we do via the Makefile.  Trying to use the CMake-generated static library gives similar errors, though many more, as the issue here.

shoaibkamil:
mingw failure is genuine, but the others are unrelated.  Will figure out the right incantation for mingw.

shoaibkamil:
All green, PTAL.

Makefile
```diff
@@ -958,17 +958,39 @@ endif
...
+    MAP_FLAGS= -Wl,-map -Wl,$(BUILD_DIR)/llvm_objects/list.all
+    LIST_OUTPUT = > /dev/null
+else
+ifeq ($(OS), Windows_NT)
+    # This is for MinGW: the map file gets written, but the
+    # compilation fails with a file truncation error.  Instead,
+    # we use the old strategy.
+    # Note: we do in fact have to pass in this flag twice.
+    MAP_FLAGS= -Wl,-t -Wl,-t
+    LIST_OUTPUT = 2>&1 | grep "libLLVM" | grep ")" > $(BUILD_DIR)/llvm_objects/list.all
```
abadams: Why are we grepping libLLVM twice in the mingw case?

shoaibkamil: If we don’t grep initially in MinGW, the file size grows too large and causes an error, so it’s necessary during the compile command (the same error is the reason we can’t use the linker map). 

For the others, we filter post-compile. I figured it would be ok to redundantly grep for MinGW rather than introducing another difference between platforms.

abadams: Might be worth recording that so that someone doesn't "fix" it later.

steven-johnson:
I guess we must not be regularly testing these on the buildbots?

Makefile
shoaibkamil: Will do.

shoaibkamil:
I think our tests (all?) use the shared library by default, so we're likely not testing static builds.

steven-johnson:
If we're not testing it, we shouldn't have it.
If we need it, we should be testing it.

abadams:
It used to be the case that many of the apps used the static library, but this was suboptimal from a binary size and build dependency perspective. Looks like only mingw still uses it, so I guess it makes sense that this was broken elsewhere. Should we arbitrarily switch one of the tested apps to static linkage? That's easy but ugly, just add "-static -L $(HALIDE_DISTRIB_PATH)/lib" to the rule that builds the generator binary.

shoaibkamil:
I propose doing something even simpler: I'll add a `test_static_lib` target that simply compiles one of the correctness tests or tutorials against the static library, and runs it.  It's as simple as changing the `-L <dir>` to `/lib` rather than `/bin`.

abadams:
This broke the static library build on ubuntu for me :( I think it's because i'm using lld, so the map file isn't getting parsed right. I'll try to tweak the sed lines.

shoaibkamil:
Ah, crap.  Let me know if you'd rather I fix it-- I could spin up a docker image locally if you let me know which Ubuntu release and lld package.

PR #4577, opened 2020-02-03
Add GitHub Action to do initial build-and-test for Ubuntu and MacOS

steven-johnson:
This is intended to be (at least) equivalent to the Travis testing we have now. Adding Windows (and 32-bit Linux) testing should be simple to add if we can obtain suitable LLVM prebuilts (which we'll probably have to handle on our own).

Note that even in this incomplete state, this could be useful, as it would allow us to remove these configurations from testing on our buildbots, reducing their load.


abadams:
Once again I should have hit "unsubscribe" much sooner, but I'm enjoying the degeneration of your commit messages.

steven-johnson:
> Once again I should have hit "unsubscribe" much sooner, but I'm enjoying the degeneration of your commit messages.

khsfk

steven-johnson:
Closing this, starting a new draft for cleanliness

PR #4578, opened 2020-02-03
Allow optional explicit names for lambdas

steven-johnson:
This can make debugging easier in complex pipelines with many lambdas.

### DIFF
--- a/src/Lambda.h
+++ b/src/Lambda.h
@@ -14,8 +14,8 @@ namespace Halide {
 /** Create a zero-dimensional halide function that returns the given
  * expression. The function may have more dimensions if the expression
  * contains implicit arguments. */
-inline Func lambda(Expr e) {
-    Func f("lambda" + Internal::unique_name('_'));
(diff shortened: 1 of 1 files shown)


abadams:
Not necessarily opposed to this functionality, but giving a lambda an explicit name seems like a serious code smell - the whole point of lambdas is that they're anonymous. If something deserves a name the Func should have a name in the C++ source too.

zvookin:
I'm gonna guess the idea is you put the names in while debugging and then remove them. I.e. generally the feature isn't used outside of debugging. No objection, but I agree with Andrew's comment. Though it means lambdas are probably a maintainability lose for anything complex. (Which is the case in all programming languages that have them, so it isn't at all surprising...)

steven-johnson:
I'm gonna withdraw this, because (1) Andrew is not wrong, (2) the Python wrappers would need attention too, and (3) for a debug-only feature this is overkill.

PR #4582, opened 2020-02-04
Smarten tutorial lessons 12 and 19 for OpenCL

steven-johnson:
When running on buildbots/VMs, we might have OpenCL "present" but no GPU to use; do more sniffing at start time so we can fail more gracefully

### DIFF
--- a/tutorial/lesson_12_using_the_gpu.cpp
+++ b/tutorial/lesson_12_using_the_gpu.cpp
@@ -317,7 +317,7 @@ Target find_gpu_target() {
         typedef int32_t (*clGetPlatformIDsFunc)(uint32_t, void**, uint32_t*);
         clGetPlatformIDsFunc clGetPlatformIDs = (clGetPlatformIDsFunc) dlsym(ocl, "clGetPlatformIDs");
         if (!clGetPlatformIDs) {
-            printf("Unable to find clGetPlatformIDs, assuming no GPU\n");
+            printf("Unable to find clGetPlatformIDs(), assuming no GPU\n");
             return target;
         }
 
--- a/tutorial/lesson_12_using_the_gpu.cpp
+++ b/tutorial/lesson_12_using_the_gpu.cpp
@@ -310,7 +310,24 @@ Target find_gpu_target() {
         target.set_feature(Target::Metal);
     }
 #else
-    if (dlopen("libOpenCL.so", RTLD_LAZY) != NULL) {
+    void* ocl = dlopen("libOpenCL.so", RTLD_LAZY);
+    if (ocl != NULL) {
+        // Do a little more sniffing, as some Linux systems might have libOpenCL installed,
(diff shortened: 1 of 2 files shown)


tutorial/lesson_12_using_the_gpu.cpp
```diff
@@ -310,7 +310,24 @@ Target find_gpu_target() {
...
 #else
-    if (dlopen("libOpenCL.so", RTLD_LAZY) != NULL) {
+    void* ocl = dlopen("libOpenCL.so", RTLD_LAZY);
+    if (ocl != NULL) {
+        // Do a little more sniffing, as some Linux systems might have libOpenCL installed,
+        // but no GPU available (e.g. on buildbots / VMs)
+        typedef int32_t (*clGetPlatformIDsFunc)(uint32_t, void**, uint32_t*);
+        clGetPlatformIDsFunc clGetPlatformIDs = (clGetPlatformIDsFunc) dlsym(ocl, "clGetPlatformIDs");
+        if (!clGetPlatformIDs) {
+            printf("Unable to find clGetPlatformIDs, assuming no GPU\n");
```
shoaibkamil: nit: `clGetPlatformIDs()` so it matches line 328

steven-johnson: This is still a Draft PR, but thanks anyway :-)

abadams:
I know it's a draft, but this is a lot of scary-looking boilerplate for a tutorial, and it's not going to be necessary on the machines users are learning Halide on, so an alternative strategy would be great if there is one.

steven-johnson:
true (though there's already scary-looking boilerplate in lesson-12).

alternatives:
- skip testing lesson 12 and 19 entirely on the systems in question. (cons: there isn't really a way to do this for the Makefile without skipping all of the tutorials, which seems suboptimal).
- add some env var ("HL_SKIP_GPU_TESTS"?) that the tutorials can sniff. (cons: env vars are awful; also, it's not clear that this is any less weird for a tutorial)
- migrate the scary boilerplate into a tools/ utility ("find_gpu_target") that can be reused between these (cons: the boilerplate is still there, it's just hidden in a place that it doesn't get in your face)
- something else?

abadams:
This came up with getting the test to use CUDA too. I think we need some way of asking libHalide whether or not a given device api (or more generally, a target including device apis) is going to work on the host machine, then hide the boilerplate in there.

steven-johnson:
Alternately: what if we just added an error code that unambiguously meant "device not supported"? The tutorials (etc) could check for this and treat it as a non-fatal error.

(AFAICT, the OpenCL path returns an OpenCL error code back out to the caller.)

steven-johnson:
> what if we just added an error code that unambiguously meant "device not supported"? 

...except that catching errors in JIT mode requires a custom error handler. Phooey.

steven-johnson:
Abandoning for now.

PR #4583, opened 2020-02-05
Tutorials should use halide_benchmark instead of clock.h

steven-johnson:
We have a really dodgy clock.h file in the tutorials used for timing a few thngs. We should use halide_benchmark.h instead.

### DIFF
--- a/tutorial/clock.h
+++ b/tutorial/clock.h
@@ -1,31 +1,16 @@
+#ifndef HALIDE_TUTORIAL_CLOCK_H
+#define HALIDE_TUTORIAL_CLOCK_H
+
 // A current_time function for use in the tests.  Returns time in
 // milliseconds.
 
-#ifdef _WIN32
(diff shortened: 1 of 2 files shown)


abadams:
I think despite being better engineering, this risks making the tutorials worse, because now there are  new abstractions for people to understand that aren't related to what the tutorial is trying to teach (e.g. the return type of benchmark_now() is some weird C++ STL monster, or what samples/iterations means for the benchmark call). It's better if they use a verbose-but-totally-explicit benchmarking methodology. tick() is clearly goofy though, and is similarly a distraction.

What do you think of this plan: Introduce the halide benchmarking helper in lesson 16 instead of tick, with a comment explaining what samples/iterations means. Leave the earlier lessons alone, but change the implementation of current_time to be built on top of benchmark_now, and return the duration since the first call to current_time.

steven-johnson:
PTAL

PR #4588, opened 2020-02-05
Add host_supports_device_api()

steven-johnson:
This adds a new call, `host_supports_device_api()`, which is intended to allow JIT callers to determine whether a given DeviceAPI is likely supported on the current host. Modified tutorial code to use it as appropriate. (Note that existing logic in e.g. lesson_12 that tries to load driver libraries isn't adequate, as a system may have the right software installed but no suitable hardware.)

### DIFF
--- a/src/runtime/opencl.cpp
+++ b/src/runtime/opencl.cpp
@@ -339,6 +339,11 @@ WEAK int create_opencl_context(void *user_context, cl_context *ctx, cl_command_q
     halide_assert(user_context, ctx != NULL && *ctx == NULL);
     halide_assert(user_context, q != NULL && *q == NULL);
 
+    if (clGetPlatformIDs == NULL) {
+        error(user_context) << "CL: clGetPlatformIDs not found\n";
+        return -1;
+    }
+
     cl_int err = 0;
 
     const cl_uint max_platforms = 4;
--- a/src/runtime/opencl.cpp
+++ b/src/runtime/opencl.cpp
@@ -263,7 +263,11 @@ public:
 #endif
 
         error = halide_acquire_cl_context(user_context, &context, &cmd_queue);
-        halide_assert(user_context, context != NULL && cmd_queue != NULL);
+        // don't abort: that would prevent host_supports_device_api() from being able
+        // work properly. Setting the error code here should be sufficient.
+        if (!context || !cmd_queue) {
+            error = -1;
+        }
     }
 
     INLINE ~ClContext() {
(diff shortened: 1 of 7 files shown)


src/runtime/opencl.cpp
```diff
@@ -720,7 +724,9 @@ WEAK int halide_opencl_device_sync(void *user_context, halide_buffer_t *) {
     debug(user_context) << "CL: halide_opencl_device_sync (user_context: " << user_context << ")\n";
 
     ClContext ctx(user_context);
-    halide_assert(user_context, ctx.error == CL_SUCCESS);
+    if (ctx.error != CL_SUCCESS) {
```
abadams: Generally on failure we call halide_error *and* return a non-zero integer, so this should do error(user_context) << ... like below.

steven-johnson: Done

tutorial/lesson_12_using_the_gpu.cpp
```diff
@@ -283,37 +283,32 @@ int main(int argc, char **argv) {
...
-        target.set_feature(Target::D3D12Compute);
-    } else if (LoadLibraryA("OpenCL.dll") != nullptr) {
-        target.set_feature(Target::OpenCL);
-    }
-#elif __APPLE__
-    // OS X doesn't update its OpenCL drivers, so they tend to be broken.
-    // CUDA would also be a fine choice on machines with NVidia GPUs.
-    if (dlopen("/System/Library/Frameworks/Metal.framework/Versions/Current/Metal", RTLD_LAZY) != NULL) {
-        target.set_feature(Target::Metal);
+    std::vector<std::pair<DeviceAPI, Target::Feature>> devices_to_try;
```
abadams: The parallel between DeviceAPI and Target::Feature is unfortunately boilerplatey. Is there a way to clean it up? Maybe it would be cleaner as host_supports_feature(Target::Feature), which dispatches internally to host_supports_device_api for features that correspond device APIs, and otherwise just calls get_host_target().has_feature(...);

steven-johnson: Yeah, we could do that, but I'm a little concerned that would invite misuse; host_supports_device_api() is clear that it is "special" (e.g. not threadsafe), whereas host_supports_feature() sounds more innocuous. 

Alternate/related would be to add something like `Target::required_device_api()`, which would return the DeviceAPI needed for a given featureset (or None if n/a); that would make the call much simpler.

steven-johnson:
PTAL

PR #4591, opened 2020-02-06
Small code cleanup and simplification

benoitsteiner:


apps/autoscheduler/FunctionDAG.cpp
```diff
@@ -601,12 +601,11 @@ FunctionDAG::FunctionDAG(const vector<Function> &outputs, const MachineParams &p
...
         for (int j = 0; j < consumer.dimensions(); j++) {
-            Expr min_var = Variable::make(Int(32), consumer.name() + "." + consumer.args()[j] + ".min");
-            Expr max_var = Variable::make(Int(32), consumer.name() + "." + consumer.args()[j] + ".max");
-            Expr extent = max_var - min_var + 1;
+            Halide::Var min_var(consumer.name() + "." + consumer.args()[j] + ".min");
+            Halide::Var max_var(consumer.name() + "." + consumer.args()[j] + ".max");
             Interval interval(min_var, max_var);
             scope.push(consumer.args()[j], interval);
-            node.region_required.push_back(interval);
+            node.region_required.push_back(SymbolicRange{min_var, max_var});
```
steven-johnson: Would this be better as `emplace_back()`?

apps/autoscheduler/FunctionDAG.h
```diff
@@ -348,6 +348,11 @@ struct FunctionDAG {
     // An edge is a producer-consumer relationship
     struct Edge;
 
+    struct SymbolicRange {
```
abadams: Halide::Range is min, extent, and Halide::Interval is min, max. In keeping with that convention, how about naming this SymbolicInterval?

benoitsteiner: Good idea, done

apps/autoscheduler/FunctionDAG.cpp
benoitsteiner: Indeed, fixed

Issue #4596, opened 2020-02-07
debug_to_file() can fail to compile for some schedules


steven-johnson:
The following snippet:

```
        Func f, g;
        Var x, y, xo, yo;

        f(x, y) = (x + y);
        g(x, y) = f(x, y);

        g.tile(x, y, xo, yo, x, y, 4, 4, TailStrategy::RoundUp)
         .vectorize(x);

        f.compute_at(g, x);

        f.debug_to_file("f.tmp");

        Buffer<int32_t> im = g.realize(8, 8);
```

will fail at compile time with:

```
Symbol not found: debug_to_file_result
```

abadams:
Oh yeah... vectorizing a debug_to_file call isn't going to work well at all - when an allocation gets vectorized it grows an additional dimension, because each vector lane notionally has its own little allocation. debug_to_file is probably going to be garbage. It might be best to just throw a user_error if we encounter a debug_to_file call inside a vector loop.

steven-johnson:
Even without vectorization, will it work reliably for anything that isn't compute_root()?

(FYI, the proximate cause of this is that the AssertStmt handler in VectorSubs doesn't call `mutate()` on either its condition or its message -- both of which are vectorized -- so the names get mismatched. Not sure if that should also be fixed.)

PR #4597, opened 2020-02-07
Forbid debug_to_file() on vectorized Funcs (Issue #4596)

steven-johnson:


src/VectorizeLoops.cpp
```diff
@@ -514,6 +514,10 @@ class VectorSubs : public IRMutator {
     }
 
     Expr visit(const Call *op) override {
+        user_assert(!op->is_intrinsic(Call::debug_to_file))
```
abadams: It's not that the Func is vectorized, it's that it's compute_at inside some other Func's vectorized loop.

Issue #4593, opened 2020-02-07
Consider dropping MinGW support


shoaibkamil:
It's unclear whether anyone uses Halide via MinGW anymore.  We should drop support if nobody is actively using it, given the fact that we have to employ specific workarounds for that environment.

steven-johnson:
So far haven't heard about a lot of use. Might be worth also explicitly pinging the halide-dev mailing list. Since this is "working" now we don't need to be in a hurry to remove it, but it always has been an odd special case. (I wonder, have people who previously used it switched to using the Linux-on-Windows environment instead?)

alexreinking:
Speaking for myself, I haven't used MinGW since real GCC became available on Windows via WSL. Instead, I prefer clang-cl or Intel's compiler on Windows if I don't want to use MSVC.

PR #4601, opened 2020-02-10
Drop support for MinGW (Issue #4593)

steven-johnson:
Not sure if we are ready for this yet or not, but here's a PR that removes all the MinGW if we decide to go that way. (Note that a separate PR for the buildbots will be necessary as well.)


README.md
```diff
@@ -133,8 +133,7 @@ To configure and build Halide:
 
 #### Building Halide and LLVM on Windows using mingw
 
-The makefile method above should work from inside a "mingw64" shell
-(not the default shell) in an msys2 installation.
+Halide no longer supports building via the Makefile under mingw. Windows users should use the CMake build process instead.
```
abadams: This comment can be misinterpreted to mean that mingw is supported via cmake. We should just delete the mingw section.

Makefile
```diff
@@ -964,21 +956,10 @@ all: distrib test_internal
...
-    MAP_FLAGS= -Wl,-t -Wl,-t
-    LIST_OUTPUT = 2>&1 | grep "libLLVM" | grep ")" > $(BUILD_DIR)/llvm_objects/list.all
 else
     MAP_FLAGS= -Wl,-Map=$(BUILD_DIR)/llvm_objects/list.all
     LIST_OUTPUT = > /dev/null
 endif
-endif
 
 $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
 	# Determine the relevant object files from llvm with a dummy
```
shoaibkamil: This could be more simplified if we don't support MinGW: We can get rid of `LIST_OUTPUT` and in line 985 always pipe to `/dev/null`.

shoaibkamil:
I'm not sure if we should do this yet, but the PR does make a good case that removing MinGW simplifies several pieces of the build process.  Otherwise LGTM.

Perhaps we should send a message on the halide-dev list as you suggested?

Makefile
steven-johnson: Done. Also the removal of xargs later in the file.

steven-johnson:
> Perhaps we should send a message on the halide-dev list as you suggested?

Done, a couple days ago. No responses as of yet.

steven-johnson:
So far, not a single complaint about this proposal. How long should we wait for feedback?

shoaibkamil:
I don't think we should wait too long, if nobody speaks up.  I'm ok with merging this, but maybe not on a Friday afternoon :)

steven-johnson:
I just landed the change to the buildbots, so it's definitely not gonna get tested anymore. Might as well approve this one too :-)

shoaibkamil:
Unfortunately LLVM looks like it's failing to build on the buildbots?

shoaibkamil:
Oh DOH, that's because they were running when you landed the changes to the buildbot config.

Issue #3767, opened 2019-03-26
Tiled compute_with allocates entire outer dimension for tile case.


zvookin:
The correctness_compute_with Yuv 420 test case is massively over-allocating and computing an intermediate buffer scheduled on tiles. In the test case, the `u_part` and `v_part` `Funcs` are recomputed at each tile but the height (`y` dimension) is computed for the size of the entire input, not the size of the tile.

https://github.com/halide/Halide/tree/compute_with_excessive_bounds has the `correctness_compute_with` test modified to fail due to this issue via counting the total number of loads and stores and comparing it to the `compute_root` reference version.

I spent a little time trying to see if this was happening when `compute_with` was first merged, but ran into issues getting a compatible llvm for compiling Halide at various times. (Thus git bisect is a bit trickier.) Wanted to get the issue filed quickly in case the cause is obvious or easily finable for others.

alexreinking:
I'm working through some compute_with bugs right now. I'll see if this happens to get fixed along with the others.

vksnk:
(Background: I've been looking into this issue lately and I think I understand what's happening, but would really appreciate an input from someone with more experience)

Context:

[Note: I simplified the test case from Zalman a bit to make it more tractable for now, but I believe it's still representative]. It might be easier to look at the [test itself](https://github.com/halide/Halide/commit/7eddb00823b4315e3734ae7e0927dbac837455e4#diff-dc08b4fbd878cc0dcb2a9552b0142673L492), but here is the key elements:

- there are three functions y_part, u_part, v_part
- u_part, v_part are scheduled to compute_with(y)
- there is an intermediate rgb_x function, which is compute_at(y)
- (this is important bit) rgb_x is only used at u_part and v_part

The problem is that currently it will allocate entire image for rgb_x and recompute it on every iteration of outer loop (y_part.y). As far as I can tell, the cause of this is that during bounds inference, when producing bounds for rgb_x, it will consider only y_part, but not u_part or v_part and given that  y_part doesn't depend on rgb_x, bounds for rgb_x loops won't be tightened.

[...]

PR #4602, opened 2020-02-12
Produce boxes for every function in the fused group

vksnk:
This PR is addressing the issue described in #3767. The basic idea is that in some cases when compute_at is used with compute_with, we need to provide boxes for every function in the fused group, so inner producer (the one scheduled with compute_at) can get proper loop bounds, so it computes only whatever is necessary (the issue referenced above provides more detailed background).

Andrew, I've added some of the optimizations you mentioned in the discussion (like using boxes_provided only if there are actually fused functions at this loop level). Also, I've shamelessly stolen your comment from the bug above and used it to describe what's going on in the source code, sorry about that, but I hope it's ok.


steven-johnson:
OK to land once all buildbots are green?

vksnk:
Yes, ready on my side.

vksnk:
Bots are green, I'll submit.

PR #4603, opened 2020-02-12
Fix -Map output parsing for lld

abadams:
Here's my attempt at fixing scraping llvm objects when the linker is lld. Let's see if it still works with the other linkers. The sed invocations delete up to the first slash, and then delete everything after ):(

@shoaibkamil

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -992,7 +992,7 @@ $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
 	# files and re-extract the required object files
 	cd $(BUILD_DIR)/llvm_objects; \
 cat list.all |  grep "libLLVM" | grep ")"  | sed "s/^[^/]*//" | sed 's/):(.*/)/' | egrep "^/|^\(" | sort | uniq > list.new; \
-	cat list.new; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -991,8 +991,8 @@ $(BUILD_DIR)/llvm_objects/list: $(OBJECTS) $(INITIAL_MODULES)
 	# is no list from a previous build, then delete any old object
 	# files and re-extract the required object files
 	cd $(BUILD_DIR)/llvm_objects; \
-	cat list.all |  grep "libLLVM" | grep ")"  | sed "s/\[.*\] //" | egrep "^/|^\(" > list.new; \
-	rm list.all; \
+cat list.all |  grep "libLLVM" | grep ")"  | sed "s/^[^/]*//" | sed 's/):(.*/)/' | egrep "^/|^\(" | sort | uniq > list.new; \
+	cat list.new; \
```
shoaibkamil: No longer deleting `list.all`?  Is that intentional or accidental?

shoaibkamil:
LGTM pending buildbots.

shoaibkamil:
Didn't seem to run on our buildbots for some reason.

PR #4605, opened 2020-02-13
Add set(CMAKE_FIND_FRAMEWORK LAST)

steven-johnson:
This addresses some issues with using libpng/libjpeg on OSX systems that have Mono installed.


alexreinking:
Probably worth filing a bug against the FindPNG and FindJPEG cmake modules. This is a workaround, not an intended behavior.

PR #4609, opened 2020-02-14
Smarten python-config logic

steven-johnson:
Our makefiles look for `python3-config` by default, but some python3-only installs may not provide this (providing only `python-config`). Smarten our logic to check and fall back to that if the expected name isn't present.


steven-johnson:
I now suspect this might have been just a misconfigured system and this change isn't needed -- leaving open (as draft) for now though

steven-johnson:
yeah, I think this is bogus. withdrawing.

PR #4614, opened 2020-02-17
Can't use substitute on Expr graphs. Use graph_substitute.

### LINKED ISSUE #4612: can_parallelize_rvar is slow for large expression
The following code takes long to compile in Halide:
```
    Func f, g;
    Var x;
    f(x) = 0;
    // Build a huge indexing expression
    RDom r(0, 10);
    Expr index_expr = r;
    for (int i = 0; i < 100; i++) {
        index_expr *= index_expr;
    }
    g(index_expr) += f(index_expr);
```

The bottleneck is in `FindLoads` in `can_parallelize_rvar`: the substitution generates enormous expressions and takes really long time to process. In contrast, changing the last line to
```
    g(x) += f(index_expr);
```
compiles instantly.

I ran into this during differentiating some code with complex indexing: the forward pass compiles quickly (since it has the pattern of the second case) while the differentiation takes forever.

Not entirely sure how to fix this yet but I think checking the expressions more carefully in `can_parallelize_rvar` can avoid the combinatorial explosion.

abadams:
Fixes #4612

### DIFF
--- a/src/ParallelRVar.cpp
+++ b/src/ParallelRVar.cpp
@@ -34,7 +34,7 @@ class FindLoads : public IRVisitor {
         IRVisitor::visit(op);
         for (size_t i = 0; i < loads.size(); i++) {
             for (size_t j = 0; j < loads[i].size(); j++) {
-                loads[i][j] = substitute(op->name, op->value, loads[i][j]);
+                loads[i][j] = graph_substitute(op->name, op->value, loads[i][j]);
             }
         }
     }


BachiLi:
Wow this is a lot simpler than I expected. Thanks a lot!
This works in the particular case I reported.

abadams:
Failure seems to be a performance flake

steven-johnson:
...delete branch :-)

PR #4618, opened 2020-02-18
Add a GitHub Action that does nightly LLVM builds.

steven-johnson:
I created this to do experimenting with using GHA as buildbots; prebuilding LLVM is essential for this task (due to time/etc constraints). At present these LLVM builds aren't being used outside my experimental branch, but I'd like to check this in to test the cron functionality (which apparently won't run as a presubmit).

Note that this only builds x86 (32 and 64) targets for osx/linux/windows; it doesn't attempt to build arm32/64 targets yet, as GHA doesn't offer native VMs for this architecture, and trying to crosscompile LLVM via qemu might be.... slow. (That said, we should probably try it.)

Some details:
- The LLVM archives are stored in a password-protected location in our GCE instance (which has 200GB of storage), so only authorized users can read or write them.
- the password is securely stored in secrets.LLVM_USER_PASSWORD in the Halide repo (accessible to anyone with Halide admin or collaborator permissions).
- We store the commit hash of each build in the archive; for each build, we compare the current LLVM top-of-tree (for a given branch) with what we have built, and so non-trunk builds are usually no-ops.
- Note that we don't use the term 'trunk' here, and just stick with the version number, so that we don't have to rename things when a new branch is declared (so currently, '11' is trunk)
- Currently this is set up to run once a day, at 1AM. (Builds typically take from 1+ hours on OSX to 2.5+ hours on Windows.) I'd be surprised if we need more frequent builds than this.

Likely TODOs:
- verify what sort of notifications are sent upon build failure (or completion).
- add some simple way to force a rebuild right away (e.g. if the commit that the nightly trunk build happened to choose was a bad commit); as-is, you would need uncomment the 'on push' part of the script and push it out to a PR to trigger it, which is suboptimal.
- Modify our existing buildbots to use these archives instead of rebuilding locally; this could potentially be a big timesaver for trunk testbranch builds. (The big caveat here of course is the lack of ARM builds at present.)

### DIFF
--- a/.github/workflows/llvm_builder.yml
+++ b/.github/workflows/llvm_builder.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         arch: [x86]
         bits: [32, 64]
-        os: [windows-latest, ubuntu-18.04, macos-latest]
+        os: [windows-2019, ubuntu-18.04, macos-10.15]
         llvm_version: [8, 9, 10, 11]
         include:
           - llvm_version: 8
@@ -33,7 +33,7 @@ jobs:
           - llvm_version: 11
             llvm_branch: master
 
-          - os: macos-latest
+          - os: macos-10.15
             cc: clang
             cxx: clang++
             halide_os: osx
@@ -43,14 +43,14 @@ jobs:
             cxx: g++
             halide_os: linux
 
-          - os: windows-latest
+          - os: windows-2019
             cc: cl.exe
             cxx: cl.exe
             halide_os: windows
 
         exclude:
           # We don't support 32-bit macos builds
-          - os: macos-latest
+          - os: macos-10.15
             bits: 32
 
     steps:
--- a/.github/workflows/llvm_builder.yml
+++ b/.github/workflows/llvm_builder.yml
@@ -5,7 +5,7 @@ on:
     - cron:  '0 1 * * *'
   # Uncomment this for debugging purposes:
   # Run immediately after every push (only for debugging)
-  push:
+  # push:
 
 jobs:
   build_llvm:
--- a/.github/workflows/llvm_builder.yml
+++ b/.github/workflows/llvm_builder.yml
@@ -1,0 +1,208 @@
+name: Halide LLVM Builder
+on:
+  # Run every day at 1AM (...presumably US Pacific Time?)
+  schedule:
+    - cron:  '0 1 * * *'
+  # Uncomment this for debugging purposes:
+  # Run immediately after every push (only for debugging)
+  push:
+
+jobs:
+  build_llvm:
+    name: llvm-${{matrix.llvm_version}}-${{matrix.arch}}-${{matrix.bits}}-${{matrix.halide_os}}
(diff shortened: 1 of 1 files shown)


.github/workflows/llvm_builder.yml
```diff
@@ -0,0 +1,208 @@
+name: Halide LLVM Builder
+on:
+  # Run every day at 1AM (...presumably US Pacific Time?)
+  schedule:
+    - cron:  '0 1 * * *'
+  # Uncomment this for debugging purposes:
+  # Run immediately after every push (only for debugging)
+  push:
```
abadams: Intentionally currently uncommented?

steven-johnson: No, but will be fixed shortly :-)

steven-johnson:
> add some simple way to force a rebuild right away

To expand on this more, based on comments on GHA forums, the 'best' way to do this at present is to set up a `repository_dispatch` event (basically just a custom webhook), which would be triggered by via `curl`; this is still suboptimal but manageable. (See http://www.btellez.com/posts/triggering-github-actions-with-webhooks.html, https://help.github.com/en/actions/reference/events-that-trigger-workflows#external-events-repository_dispatch for more info)

steven-johnson:
> trying to crosscompile LLVM via qemu might be.... slow

Expanding more on this:
- How doable might it be to crosscompile for arm32/arm64 on an x86-64 linux box? (This would almost certainly be the fastest option if so)
- GHA allows for 'self-hosted' runners, so we could provision our own (say) RasPi box to do the arm builds... or maybe using an AWS Graviton instance?
- I have no idea what the order of magnitude slowdown for qemu (etc) emulated arm instances are; I'm assuming they are Too Damn Slow To Compile LLVM, but would appreciate feedback on it

steven-johnson:
> > trying to crosscompile LLVM via qemu might be.... slow
> 
> Expanding more on this:
> 
> * How doable might it be to crosscompile for arm32/arm64 on an x86-64 linux box? (This would almost certainly be the fastest option if so)
> * GHA allows for 'self-hosted' runners, so we could provision our own (say) RasPi box to do the arm builds... or maybe using an AWS Graviton instance?
> * I have no idea what the order of magnitude slowdown for qemu (etc) emulated arm instances are; I'm assuming they are Too Damn Slow To Compile LLVM, but would appreciate feedback on it

abadams:
I am currently trying to get a working llvm inside qemu to test sdot and udot instructions. I'll let you know how I go. Currently it seems like compiling llvm would be an overnight thing.

steven-johnson:
> I am currently trying to get a working llvm inside qemu to test sdot and udot instructions. I'll let you know how I go. Currently it seems like compiling llvm would be an overnight thing.

Per my previous comment, looks like LLVM has explicit instructions on doing x86->arm crosscompiles. If/when this PR lands, I'll follow up with a change to try adding those, which should allow you to pull prebuilt llvms directly to avoid this.

steven-johnson:
PTAL

PR #4620, opened 2020-02-19
Add repository_dispatch support to llvm builder

steven-johnson:
This is intended to allow force rebuilding of some/all of the llvm targets via CURL


steven-johnson:
Apparently you can't trigger these events in a non-master branch, so I have to land this in order to even test it...

Issue #4619, opened 2020-02-19
test/generator/multitarget* uses host/host-debug incorrectly


alexreinking:
`debug` is a _matching_ feature, meaning that it must occur in all targets that share a multitarget runtime. The multitarget test doesn't do this, though, and should use a different differentiating feature like "host", "host-no_sse41" or something.

alexreinking:
Fixed by #4623

PR #4623, opened 2020-02-19
Fix test/generator_aot_multitarget (Issue #4619)

steven-johnson:
Need to use a feature that doesn't need a custom runtime.

(Also, drive-by addition of ::Debug to the multitarget sniff test, which it should have been in already)

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -756,9 +756,10 @@ void compile_multitarget(const std::string &fn_name,
             user_error << "All Targets must have matching arch-bits-os for compile_multitarget.\n";
         }
         // Some features must match across all targets.
-        static const std::array<Target::Feature, 8> must_match_features = {{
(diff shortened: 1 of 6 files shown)


src/Module.cpp
```diff
@@ -756,9 +756,10 @@ void compile_multitarget(const std::string &fn_name,
             user_error << "All Targets must have matching arch-bits-os for compile_multitarget.\n";
         }
         // Some features must match across all targets.
-        static const std::array<Target::Feature, 8> must_match_features = {{
+        static const std::array<Target::Feature, 9> must_match_features = {{
```
alexreinking: Should we keep this in sync with Target::get_runtime_compatible_target?

alexreinking: https://github.com/halide/Halide/blob/212b9e5ec33b8563aa7ef878fffa93cdabdb99f7/src/Target.cpp#L889

PR #4625, opened 2020-02-19
Fix #4624 by adding include guards around the actual includes in Runtime headers

### LINKED ISSUE #4624: Generated C headers try to #include HalideRuntime.h as well as including a full copy of it
This can be easily seen by building the tutorials:

```
[alex@reinking-arch Halide_build]$ make -f ../Halide/Makefile -j4 test_tutorial
[alex@reinking-arch Halide_build]$ cd bin
[alex@reinking-arch bin]$ HL_TARGET=host-cuda HL_JIT_TARGET=host-cuda ./tutorial_lesson_10_aot_compilation_generate 
Halide pipeline compiled, but not yet run.
[alex@reinking-arch bin]$ cat lesson_10_halide.h | grep HalideRuntime
 * device API instead (e.g. HalideRuntimeCuda.h). */
#include "HalideRuntime.h"
```

This file is created here:

https://github.com/halide/Halide/blob/212b9e5ec33b8563aa7ef878fffa93cdabdb99f7/src/CodeGen_C.cpp#L367-L425

And the offending section of the cuda runtime is here:

https://github.com/halide/Halide/blob/212b9e5ec33b8563aa7ef878fffa93cdabdb99f7/src/runtime/HalideRuntimeCuda.h#L1-L5

This forces our end users to have an installation of Halide somewhere. Linking to our generated libraries means you need to have $HALIDE_DISTRIB_DIR/include in your include path.

alexreinking:
All in the title. Clang-format saw fit to change a line, too.

### DIFF
--- a/src/runtime/HalideRuntimeCuda.h
+++ b/src/runtime/HalideRuntimeCuda.h
@@ -1,6 +1,7 @@
 #ifndef HALIDE_HALIDERUNTIMECUDA_H
 #define HALIDE_HALIDERUNTIMECUDA_H
 
+// Don't include HalideRuntime.h if the contents of it were already pasted into a generated header above this one
 #ifndef HALIDE_HALIDERUNTIME_H
 
 #include "HalideRuntime.h"
--- a/src/runtime/HalideRuntimeCuda.h
+++ b/src/runtime/HalideRuntimeCuda.h
@@ -1,7 +1,11 @@
 #ifndef HALIDE_HALIDERUNTIMECUDA_H
 #define HALIDE_HALIDERUNTIMECUDA_H
 
+#ifndef HALIDE_HALIDERUNTIME_H
+
 #include "HalideRuntime.h"
+
+#endif
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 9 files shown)


abadams:
Please add a comment explaining why this is necessary. E.g: "Don't include HalideRuntime.h if the contents of it were already pasted into a generated header above this one"

src/runtime/HalideRuntimeCuda.h
```diff
@@ -1,8 +1,12 @@
 #ifndef HALIDE_HALIDERUNTIMECUDA_H
 #define HALIDE_HALIDERUNTIMECUDA_H
 
+#ifndef HALIDE_HALIDERUNTIME_H
+
```
steven-johnson: IMHO we'd be better off without the blank lines before and after, but I won't withhold approval because of it

alexreinking: clang-format wanted them there, so I left them there. I agree they're ugly, though. Maybe there's a switch we can put in .clang-format

alexreinking:
I added the comments. I'll wait for green to merge.

alexreinking:
Failure is LLVM-trunk related.

PR #4626, opened 2020-02-19
Don't keep unnecessary data around.

benoitsteiner:
The node map is only used in the constructor of FunctionDAG: make it a local datastructure instead of keeping it as a class member.

### DIFF
--- a/apps/autoscheduler/FunctionDAG.h
+++ b/apps/autoscheduler/FunctionDAG.h
@@ -544,11 +544,6 @@ struct FunctionDAG {
     vector<Node> nodes;
     vector<Edge> edges;
 
-    // We're going to be querying this DAG a lot while searching for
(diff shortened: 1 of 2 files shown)


apps/autoscheduler/FunctionDAG.h
```diff
@@ -544,11 +544,6 @@ struct FunctionDAG {
     vector<Node> nodes;
     vector<Edge> edges;
 
-    // We're going to be querying this DAG a lot while searching for
```
abadams: Interesting, I guess this comment is no longer true because it was slow to index things by Function.

abadams:
@aekul does your branch use the node_map outside of FunctionDAG?

aekul:
No, it only uses it in FunctionDAG's constructor.

PR #4628, opened 2020-02-19
Within-vector reductions

abadams:
Various ISAs have vector instructions that perform a full or partial reduction across the vector lanes, and we have had no good way to reach these. Of particular note is the instruction variously referred to as udot/dp4a/vrmpy, which does a 4 x int8_t dot product into a 32-bit result, all potentially within a wider vector. This instruction is important for quantized neural networks.

This PR reaches such ops by composing atomic() with vectorize(). These operations as viewed as atomic (i.e. semantics are serial but unordered) reductions along the vectorized dimension. 

I added a new IR node VectorReduce, which reduces groups of adjacent lanes in a wide vector down to a narrower vector using some operator. This is equivalent to a binary operation tree on slices of the vector, but that's much harder to peephole match in backends. The reduction operators are our primitive IR nodes that happen to be associative and commutative (add, mul, min, max, and, or). This node is introduced by the vectorize_loops pass when it would otherwise emit a pattern of the flavor: `f(ramp()/4) = f(ramp()/4) + some_vector;` or for total reductions `f(broadcast()) = f(broadcast()) + some_vector;`

Backends can specialize their handling of this node to hit the appropriate ops. It gets a little more complex than that because some of these ops can also incorporate an existing accumulator. Much of the new logic in this PR is in various backends. I added support for x86, arm, and cuda. I did not touch the hexagon backend, as I figured @pranavb-ca or @dsharletg would be better at handling that. For the arm udot/sdot instructions I added codegen that I believe is correct but again have no way to test it (our buildbots don't have these instructions, which need a qualcomm 855 or similar). I added a feature flag to guard the emission of these instructions. Maybe @pranavb-ca or @dsharletg can sanity-check with a quick test on an 855? 

For d3d and opencl there are intrinsics for this that should reach the same underlying instruction (dp4a on nvidia cards) but I didn't want to continue to drag on this PR. Those can be addressed later. Also related but not considered here is reductions along the warp via composing atomic() and gpu_lanes().

As a final note, atomic_vectorization is my favorite branch name to date.


abadams:
FYI @alexreinking that cmake presubmit just caught me forgetting to add two tests to the cmakelist, so it was worthwhile.

steven-johnson:
Is this ready to review? (I'm assuming not from the failed buildbots.)

abadams:
Triaging the failed buildbots now. Looks like the main issue is an llvm bug that I'll need to work around: https://bugs.llvm.org/show_bug.cgi?id=44976

But there's also a crash on arm that I need to figure out. Feel free to wait until I get the buildbots green - this isn't urgent.

abadams:
Craig Topper jumped on my bug report and so the llvm bug was just fixed in master. We just need to figure out what to do for earlier llvms. A workaround is probably still valuable if there's a simple one.

steven-johnson:
For now, you could just declare (and enforce) that this feature is only available in LLVM 11+

abadams:
Might be hard to do, because this isn't a peephole optimization that's breaking. It happens for vanilla codegen of a VectorReduce IR node, and these could appear for arbitrary reasons. I think I'll just try padding out the vectors to the next power of two for llvms < 11.

abadams:
Actually this bug triggers for just widening multiplication of non-power-of-two vector sizes:

```
    ImageParam in1(Int(16), 1), in2(Int(16), 1);
    Func f;
    Var x;
    f(x) = cast<int32_t>(in1(x)) * in2(x);
    f.vectorize(x, 12);
    f.compile_jit();
```

Yikes. I guess nobody uses those. I'll have to inject a workaround into the Mul visitor

abadams:
FYI @alinas for the just-fixed llvm bug, in case you encounter something similar. It goes away if you turn on the sse4.1 feature flag, so I doubt this ever triggers when building for real systems.

alinas:
Got it, thanks for the heads-up.

steven-johnson:
Something is wonky with this PR -- it's been showing a travis test as in progress for days now. Maybe due to the unresolved conflicts?

abadams:
I'll try a merge with master

abadams:
This is now green and ready for review

steven-johnson:
(I know normally we don't bother rebasing commits, but in this case, maybe consider doing so -- it's spread over 80 or so commits.)

abadams:
Sure. There are a few things I can probably carve off into separate PRs too

src/CodeGen_LLVM.cpp
```diff
@@ -1472,6 +1472,14 @@ Value *CodeGen_LLVM::codegen(const Expr &e) {
     value = nullptr;
     e.accept(this);
     internal_assert(value) << "Codegen of an expr did not produce an llvm value\n";
+
+    // Halide doesn't distinguish between scalars and vectors of size 1.
```
BachiLi: might be good to expand this comment: the code below is converting a vector of size 1 to a scalar?

src/CodeGen_LLVM.cpp
```diff
@@ -4298,11 +4352,246 @@ void CodeGen_LLVM::visit(const Shuffle *op) {
...
+            equiv = binop(equiv, init);
+        }
+        equiv = cast(op->type, equiv);
+        equiv.accept(this);
+        return;
+    }
+
+#if LLVM_VERSION >= 90
+    if (output_lanes == 1 &&
+        // As of the release of llvm 10, the 64-bit experimental total
```
BachiLi: should these criteria be moved to "llvm_has_intrinsic"?

src/VectorizeLoops.cpp
```diff
@@ -994,6 +1240,169 @@ class VectorSubs : public IRMutator {
         : var(std::move(v)), replacement(std::move(r)), target(t), in_hexagon(in_hexagon) {
         widening_suffix = ".x" + std::to_string(replacement.type().lanes());
     }
+};  // namespace
+
+class FindVectorizableExprsInAtomicNode : public IRMutator {
+    Scope<> poisoned_names;
```
BachiLi: I probably know what this means but might be good to comment about the definition of "poisoned".

steven-johnson:
> Sure. There are a few things I can probably carve off into separate PRs too

Yeah, with this amount of change, any easy carving will likely make our life simpler later (if we have to go back and debug injections)

src/Bounds.cpp
```diff
@@ -1340,6 +1340,34 @@ class Bounds : public IRVisitor {
...
+        case VectorReduce::Add:
+            if (interval.has_upper_bound()) {
+                interval.max *= factor;
+            }
+            if (interval.has_lower_bound()) {
+                interval.min *= factor;
+            }
+            break;
+        case VectorReduce::Mul:
+            interval = Interval::everything();
```
steven-johnson: Nit: aren't there *some* cases we could exploit here? e.g., what if the type is float and we know the bounds are [0.0, 1.0]?

src/CodeGen_ARM.cpp
```diff
@@ -1065,6 +1062,165 @@ void CodeGen_ARM::visit(const LE *op) {
...
+    }
+
+    // ARM has a variety of pairwise reduction ops for +, min,
+    // max. The versions that do not widen take two 64-bit args and
+    // return one 64-bit vector of the same type. The versions that
+    // widen take one arg and return something with half the vector
+    // lanes and double the bit-width.
+
+    int factor = op->value.type().lanes() / op->type.lanes();
+    if ((op->type.is_int() ||
```
steven-johnson: style nit: this expr is really hard to read. Might be better to pull out as a separate expr, perhaps even like

bool has_reduce = is_int || is_uint || is_float;
if (!case1) has_reduce = false;  // reason why
if (!case2) has_reduce = false;  // reason why

etc

src/CodeGen_ARM.cpp
```diff
@@ -1100,6 +1256,10 @@ string CodeGen_ARM::mattrs() const {
             arch_flags = "+sve";
         }
 
+        if (target.has_feature(Target::ARMDotProd)) {
```
steven-johnson: Does this need to be gated by LLVM_VERSION?

src/CodeGen_LLVM.cpp
steven-johnson: Do they work in trunk

src/CodeGen_LLVM.cpp
```diff
@@ -4298,11 +4352,246 @@ void CodeGen_LLVM::visit(const Shuffle *op) {
...
+            equiv = binop(equiv, init);
+        }
+        equiv = common_subexpression_elimination(equiv);
+        equiv.accept(this);
+        return;
+    }
+
+    if (factor > 2 && ((factor & 1) == 0)) {
+        // Factor the reduce into multiple stages. If we're going to
+        // be widen the type by 4x or more we should also factor the
```
steven-johnson: widening

src/CodeGen_LLVM.cpp
```diff
@@ -4298,11 +4352,246 @@ void CodeGen_LLVM::visit(const Shuffle *op) {
...
+        // widening into multiple stages.
+        Type intermediate_type = op->value.type().with_lanes(op->value.type().lanes() / 2);
+        Expr equiv = VectorReduce::make(op->op, op->value, intermediate_type.lanes());
+        if (op->op == VectorReduce::Add &&
+            (op->type.is_int() || op->type.is_uint()) &&
+            op->type.bits() >= 32) {
+            Type narrower_type = op->value.type().with_bits(op->type.bits() / 4);
+            Expr narrower = lossless_cast(narrower_type, op->value);
+            if (!narrower.defined() && narrower_type.is_int()) {
+                // Maybe we can narrow to an uint instead.
```
steven-johnson: I would have used "a uint" -- I guess it depends on how you mentally pronounce 'uint'?

src/CodeGen_X86.cpp
```diff
@@ -107,6 +107,33 @@ void CodeGen_X86::visit(const Sub *op) {
     }
 }
 
+void CodeGen_X86::visit(const Mul *op) {
+
+#if LLVM_VERSION < 110
```
steven-johnson: nit: the pattern we use pretty much exclusively in existing code is `LLVM_VERSION >= someconstant`, to make searching for obsolete code easier.

src/CodeGen_X86.cpp
```diff
@@ -390,6 +417,32 @@ void CodeGen_X86::visit(const Call *op) {
...
+        Type narrower = Int(16, mul->type.lanes());
+        Expr a = lossless_cast(narrower, mul->a);
+        Expr b = lossless_cast(narrower, mul->b);
+        if (op->type.is_int() &&
+            op->type.bits() == 32 &&
+            a.defined() &&
+            b.defined() &&
+            factor == 2 &&
+            op->op == VectorReduce::Add) {
+            if (target.has_feature(Target::AVX2) && op->type.lanes() > 4) {
```
steven-johnson: Is this really the only VectorReduce opportunity on x86? (If not, TODO + issue would be appropriate)

src/Deinterleave.cpp
```diff
@@ -313,24 +314,18 @@ class Deinterleaver : public IRGraphMutator {
...
-            return Shuffle::make({op}, indices);
         }
+
+        // Keep the same set of vectors and extract every nth numeric
+        // arg to the shuffle.
+        std::vector<int> indices;
+        for (int i = 0; i < new_lanes; i++) {
+            int idx = i * lane_stride + starting_lane;
+            indices.push_back(op->indices[idx]);
+            debug(0) << indices.back() << "\n";
```
steven-johnson: I'm sure we don't want this `debug(0)` left in :-)

src/IRMatch.h
```diff
@@ -239,7 +239,7 @@ struct WildConstInt {
             halide_scalar_value_t val;
             halide_type_t type;
             state.get_bound_const(i, val, type);
-            return op->type == type && value == val.u.i64;
+            return e.type == type && value == val.u.i64;
```
steven-johnson: eep. this looks like a scary pre-existing bug?

src/ModulusRemainder.cpp
```diff
@@ -489,11 +490,17 @@ void ComputeModulusRemainder::visit(const Let *op) {
 }
 
 void ComputeModulusRemainder::visit(const Shuffle *op) {
-    // It's possible that scalar expressions are extracting a lane of a vector - don't fail in this case, but stop
+    // It's possible that scalar expressions are extracting a lane of
+    // a vector - don't faiql in this case, but stop
```
steven-johnson: fail

src/ScheduleFunctions.cpp
```diff
@@ -354,9 +354,21 @@ Stmt build_provide_loop_nest(const map<string, Function> &env,
...
+        for (auto d : def.schedule().dims()) {
+            any_unordered_parallel |= is_unordered_parallel(d.for_type);
+        }
+        if (any_unordered_parallel) {
+            // If required, we will allocate a mutex buffer called func.name() + ".mutex"
+            // The buffer is added in the AddAtomicMutex pass.
+            body = Atomic::make(func.name(), func.name() + ".mutex", body);
+        } else {
+            // No mutex is required if there is no parallelism, and it
+            // doesn't work if all parallelism is synchronous
```
steven-johnson: does "doesn't work at all" mean "will fail to compile" or "will give bad results unexpectedly"? if the latter, we should inject checks

src/VectorizeLoops.cpp
```diff
@@ -994,6 +1240,169 @@ class VectorSubs : public IRMutator {
...
+
+    Expr visit(const Let *op) override {
+        return visit_let(op);
+    }
+
+    Expr visit(const Load *op) override {
+        // Even if the load is bad, maybe we can lift the index
+        IRMutator::visit(op);
+
+        // TODO: tuples?
```
steven-johnson: this TODO could probably use a bit more meat on the bone

src/VectorizeLoops.cpp
```diff
@@ -1022,6 +1431,7 @@ class VectorizeLoops : public IRMutator {
             // Replace the var with a ramp within the body
             Expr for_var = Variable::make(Int(32), for_loop->name);
             Expr replacement = Ramp::make(for_loop->min, 1, extent->value);
+            stmt = for_loop->body;
```
steven-johnson: ...is this a pre-existing bug?

src/runtime/arm_cpu_features.cpp
```diff
@@ -20,6 +20,8 @@ WEAK CpuFeatures halide_get_cpu_features() {
     //    features.set_available(halide_target_feature_armv7s);
     // }
 
+    // TODO: add runtime detection for ARMDotProd extension
```
steven-johnson: Open an issue on this

test/correctness/cuda_8_bit_dot_product.cpp
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+}
+
+int main(int argc, char **argv) {
+    Target t = get_jit_target_from_environment();
+    if (!t.has_feature(Target::CUDACapability61)) {
+        printf("Not running test because no cuda (with compute capability 6.1) is not enabled in the target\n");
+        return 0;
+    }
```
steven-johnson: suggestion: print the target here

test/correctness/simd_op_check.cpp
```diff
@@ -187,7 +187,7 @@ class SimdOpCheck : public SimdOpCheckTest {
 
         // SSE 2
 
-        for (int w = 2; w <= 4; w++) {
+        for (int w : {2, 4}) {
```
steven-johnson: I presume you intend to skip the w=3 case here

src/CodeGen_ARM.cpp
abadams: llvm 8 has it, so we're good

src/CodeGen_X86.cpp
abadams: In this case there's no code for the case where LLVM_VERSION >= 110, so it would be:
```
#if LLVM_VERSION >= 110
#else 
some code
#endif  
```

steven-johnson: SGTM

src/IRMatch.h
abadams: Yes. Will pull into separate PR.

src/ScheduleFunctions.cpp
abadams: I'll clarify the comment. It's more that it wouldn't work so we don't do it rather than it doesn't work. 

This PR would just scalarize such code during vectorization anyway.

src/VectorizeLoops.cpp
abadams: Hrm, that's something I genuinely forgot to finish. Will need to add a test.

src/VectorizeLoops.cpp
abadams: No, I think this additional line is a no-op, because the next line doesn't use stmt and just clobbers it. I'll delete it.

test/correctness/simd_op_check.cpp
abadams: Yeah, LLVM does different things with different versions, and it's not a case worth testing anyway

src/CodeGen_LLVM.cpp
abadams: No

abadams:
Moving to top-level comment: Not finished until I address the TODO: tuples? in VectorizeLoops.cpp

joesavage:
Do we have a story in this pull request for hoisting the total reductions that typically follow these kinds of partial reduction operations outside the inner loop? Today, using this branch I'm finding that Halide code like the following to perform the dot product:

```c
Func dot("dot");
RDom rv(0, num_elems);
dot() += cast(UInt(32), A_u8(rv)) * B_u8(rv);
dot.update().atomic().vectorize(rv, 16);
```

Produces assembly code like the following for AArch64, where the `movi` vector initialization and `addv` horizontal reduction are in the innermost loop:

```
ldr	q0, [x12], #16
ldr	q1, [x10], #16
movi	v2.2d, #0x0
subs	x9, x9, #0x1
udot	v2.4s, v0.16b, v1.16b
addv	s0, v2.4s
fmov	w13, s0
add	w8, w13, w8
b.ne	374 <dot+0x374>
```

Presumably the same is also true when generating Hexagon's `vrmpy` and the like, producing code that doesn't make the most of the accumulating capabilities of these instructions. Is there some way to get Halide to continually accumulate into a single vector using this `atomic().vectorize()` paradigm, or is it expected that these capabilities make up a separate unit of work and thus would likely come from a separate pull request? The following sequence that feels like it perhaps should work currently seems to hit an internal error ("Only Serial and Parallel For nodes should survive down to codegen"):

```c
Func dot("dot");
Func partial("partial");
Var x("x"), y("y"), k("k");

// Define the function
const int vec = 16;
RDom rvi(0, vec, "rvi");
RDom rvo(0, num_elems/vec, "rvo");
partial(k) += cast(UInt(32), A_u8(vec * k + rvi)) * B_u8(vec * k + rvi);
dot() += partial(rvo);

// Schedule the function
partial.update().atomic().vectorize(rvi);
dot.update().atomic().vectorize(rvo, 4);
```

abadams:
The scheduling directive for that is already in master: rfactor. It's a bit tricky to use unfortunately. Here's how you'd use it for a vectorized dot product:

```
// Define the function
RDom r(0, num_elems);
dot() += cast(UInt(32), A_u8(vec * k + rvi)) * B_u8(vec * k + rvi);

RVar rvo, rvi;
Var k;
const int vec = 16;
dot.update().split(r, rvo, rvi, vec);
// In partial, each value of rvi gets a separate accumulator indexed by the new pure variable k.
Func partial = dot.update().rfactor(rvi, k); 
partial.compute_root().vectorize(k).update().vectorize(k);
```

joesavage:
Thanks for the pointer towards `rfactor`. Unfortunately, though, it looks like this particular case might be a bit more complicated. If we try and simply use `rfactor` on a reduction domain of extent 16 as suggested, this implies that we want to compute over 16 independent partial sums, producing a `VectorReduce` node with source type `uint8x16` and target type `uint32x16`. Since this reads 16 elements from each input buffer and produces 16 outputs, this isn't really a `VectorReduce` in the true sense, and thus doesn't generate `UDOT`.

Ultimately, what we really want here is an `rfactor` of extent four – each independent accumulation for which contains an inner loop of size four – and then to vectorize over these loops with a factor of 16 to product a `VectorReduce` with source type `uint8x16` and target type `uint32x4`. I tried to have a stab at doing this with the following:

```c
// Define the function
Func dot("dot");
RDom r(0, num_elems, "r");
dot() += cast(UInt(32), A(r)) * B(r);

// Schedule the function
Var k("k");
RVar rvo("rvo"), rvoo("rvoo"), rvio("rvio"), rvii("rvii"), fused("fused");
dot.update().split(r, rvo, rvii, 4).split(rvo, rvoo, rvio, 4);
Func partial = dot.update().rfactor(rvio, k);
partial.compute_root().vectorize(k);
partial.update().fuse(rvii, k, fused).atomic().vectorize(fused);
```

Unfortunately, though, while this does produce a `UDOT` instruction, it seems to be doing some really weird stuff such that it definitely isn't generating the desired result. I need to look into the 'why' in more detail, but I'm seeing lots of weird if statements in the pseudocode for the inner loop as a result of the vectorization on `fused` for whatever reason.

Happy to move this discussion into a separate issue if that's helpful as I don't want to drown out any actual review comments for the pull request, but I figure that getting this use case nailed down is probably quite important for this patch and I'm curious to see if there's a good answer here.

abadams:
Current state of the branch is a mess, because I discovered a case where atomic nodes were being incorrectly stripped and had to add them back everywhere temporarily, resulting in CAS loops. I temporarily hacked them back off to look at this case.

Everything works out cleanly if you promise num_elems is a multiple of 16. When it's not you get tail cases. One problem is that both of those splits are GuardWithIf, because they're splits of an RVar with unknown extent. You can make the inner one a RoundUp by reassociating the splits:

```
Var k("k");
RVar rvo("rvo"), rvoo("rvoo"), rvio("rvio"), rvii("rvii"), fused("fused");
dot.update().split(r, rvo, rvii, 4).split(rvo, rvoo, rvio, 4);
Func partial = dot.update().rfactor(rvio, k);
partial.compute_root().vectorize(k);
partial.update().fuse(rvii, k, fused).atomic().vectorize(fused);
```

That gives you the hot loop:

```
.LBB0_2:                                // %"for dot_intm.s1.r$x.rvo"
                                        // =>This Inner Loop Header: Depth=1
	ldr	q1, [x10], #16
	ldr	q2, [x11], #16
	subs	x9, x9, #1              // =1
	udot	v0.4s, v1.16b, v2.16b
	b.ne	.LBB0_2

```

If num_elems is not known to be a multiple of 16, you get tail cases and if statements, and the hot loop becomes:
```
.LBB0_2:                                // %true_bb
                                        //   in Loop: Header=BB0_4 Depth=1
	lsl	x2, x15, #4
	sub	x3, x2, x11
	sub	x2, x2, x13
	ldr	q0, [sp]
	ldr	q1, [x8, x3]
	ldr	q2, [x10, x2]
	udot	v0.4s, v1.16b, v2.16b
	str	q0, [sp]
.LBB0_3:                                // %after_bb
                                        //   in Loop: Header=BB0_4 Depth=1
	add	x15, x15, #1            // =1
	cmp	x15, x16
	b.eq	.LBB0_37
.LBB0_4:                                // %"for dot_intm.s1.r$x.rvo"
	sxtw	x2, w15
	lsl	x3, x2, #4
	add	w2, w3, #16             // =16
	cmp	w2, w12
	b.le	.LBB0_2
```

Not sure why LLVM sees the need to spill the accumulator (q0). But in any case we shouldn't have this if statement. I'll see if I can get it to go away.

abadams:
Found the problem. Will be fixed in the final version of the branch.

abadams:
ok, ready for more review

src/ModulusRemainder.cpp
vksnk: Fail is still misspelled.

src/IREquality.cpp
```diff
@@ -589,6 +590,14 @@ void IRComparer::visit(const Atomic *op) {
     compare_stmt(s->body, op->body);
 }
 
+void IRComparer::visit(const VectorReduce *op) {
+    const VectorReduce *e = expr.as<VectorReduce>();
+
+    compare_scalar(op->op, e->op);
+    // We've already compared types, so it's enough to compare the value
```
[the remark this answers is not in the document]
abadams: types get compared before this function is called in compare_expr

src/IRMatch.cpp
```diff
@@ -282,6 +282,16 @@ class IRMatch : public IRVisitor {
             result = false;
         }
     }
+
+    void visit(const VectorReduce *op) override {
+        const VectorReduce *e = expr.as<VectorReduce>();
+        if (result && e && op->op == e->op) {
```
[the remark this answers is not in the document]
abadams: I think we need to compare types here. Will fix.

test/correctness/vector_reductions.cpp
```diff
@@ -0,0 +1,126 @@
...
+                            in(x) = cast(src_type, random_float());
+                        } else {
+                            in(x) = cast(src_type, random_int());
+                        }
+                        in.compute_root();
+
+                        Expr rhs = cast(dst_type, in(x * reduce_factor + r));
+                        Expr rhs2 = cast(dst_type, in(x * reduce_factor + r + 32));
+
+                        if (op == 4 || op == 5) {
```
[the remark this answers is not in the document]
abadams: These aren't reduction operator types, they're the test cases in the switch statement below. I'll add a comment.

src/IRMatch.cpp
```diff
@@ -413,6 +423,10 @@ bool equal_helper(const BaseExprNode &a, const BaseExprNode &b) noexcept {
     case IRNodeType::Shuffle:
         return (equal_helper(((const Shuffle &)a).vectors, ((const Shuffle &)b).vectors) &&
                 equal_helper(((const Shuffle &)a).indices, ((const Shuffle &)b).indices));
+    case IRNodeType::VectorReduce:
+        return (((const VectorReduce &)a).op == ((const VectorReduce &)b).op &&
```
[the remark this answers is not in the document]
abadams: Hrm, at this point we already know the types match (it's checked outside equal_helper) but we don't know that the types of value match before recursively calling equal_helper, so I'd better check that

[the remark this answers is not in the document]
abadams: I think the cast handler has the same bug

src/VectorizeLoops.cpp
```diff
@@ -941,6 +1029,171 @@ class VectorSubs : public IRMutator {
         return Allocate::make(op->name, op->type, op->memory_type, new_extents, op->condition, body, new_expr, op->free_function);
     }
 
+    Stmt visit(const Atomic *op) override {
+        // Recognize a few special cases that we can handle as within-vector reduction trees.
+        do {
```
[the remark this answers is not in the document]
abadams: I considered it, but having an explicit scope with breaks makes me more confident that RAII things are going to have the lifetime I expect

src/Simplify_Exprs.cpp
```diff
@@ -42,6 +42,141 @@ Expr Simplify::visit(const Broadcast *op, ExprInfo *bounds) {
...
+    }
+    case VectorReduce::Min: {
+        auto rewrite = IRMatcher::rewriter(IRMatcher::h_min(value, lanes), op->type);
+        if (rewrite(h_min(min(x, broadcast(y))), min(h_min(x, lanes), broadcast(y, lanes))) ||
+            rewrite(h_min(min(broadcast(x), y)), min(h_min(y, lanes), broadcast(x, lanes))) ||
+            rewrite(h_min(max(x, broadcast(y))), max(h_min(x, lanes), broadcast(y, lanes))) ||
+            rewrite(h_min(max(broadcast(x), y)), max(h_min(y, lanes), broadcast(x, lanes))) ||
+            rewrite(h_min(broadcast(x)), broadcast(x, lanes)) ||
+            rewrite(h_min(ramp(x, y)), x + min(y * (arg_lanes - 1), 0)) ||
+            false) {
```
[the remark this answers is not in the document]
abadams: It makes it possible to copy-paste simplifier rules in at the end or move them around without worrying about getting the number of closing parens right

src/IROperator.cpp
```diff
@@ -463,6 +463,31 @@ Expr lossless_cast(Type t, Expr e) {
                 return Expr();
             }
         }
+
+        if (const VectorReduce *red = e.as<VectorReduce>()) {
+            const int factor = red->value.type().lanes() / red->type.lanes();
+            switch (red->op) {
+            case VectorReduce::Add:
+                if (t.bits() >= 16 && factor < (1 << (t.bits() / 2))) {
```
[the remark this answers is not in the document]
abadams: You are right. will delete

src/VectorizeLoops.cpp
```diff
@@ -170,21 +171,95 @@ Interval bounds_of_lanes(const Expr &e) {
...
+    } else {
+        Expr min_lane = VectorReduce::make(VectorReduce::Min, e, 1);
+        Expr max_lane = VectorReduce::make(VectorReduce::Max, e, 1);
+        return {min_lane, max_lane};
     }
-    return {min_lane, max_lane};
 };
 
+// A ramp with the lanes repeated (e.g. <0 0 2 2 4 4 6 6>)
+struct InterleavedRamp {
```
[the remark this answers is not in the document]
abadams: Done

abadams:
woo!

PR #4633, opened 2020-02-20
Prevent a memory corruption when featurizing a schedule.

benoitsteiner:
When featurizing a schedule for a computation like
   f(x, y, x) = g(x) + h(x, y, z)
the scheduler might try to attempt to vectorize on the 2nd or 3rd
dimension of f. However there is no such dimension for g.

### DIFF
--- a/apps/autoscheduler/LoopNest.cpp
+++ b/apps/autoscheduler/LoopNest.cpp
@@ -347,7 +347,7 @@ void LoopNest::compute_features(const FunctionDAG &dag,
                 // Extent of the innermost dimension in the storage layout
                 int64_t innermost_storage_extent = 1;
                 int v = p->vector_dim;
-                if (v >= 0 && node->dimensions > 0) {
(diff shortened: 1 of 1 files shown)


abadams:
I don't think this is a schedule that corrupts memory - this is just fixing an incorrect range check when accessing an array during autoscheduling. Still, adding something to apps/autoscheduler/test.cpp would be nice.

steven-johnson:
(Feel free to land at will. Adding a test that would have caught this would still be nice.)

benoitsteiner:
I tried to add a test over the week-end, but it turns out to be fairly challenging. There are 2 ways to push to autoscheduler to reproduce the bug: giving it a pipeline that would be best scheduled along the outermost dimension, and tweaking the cost model to encourage vectorization of outer dimensions. We can't really do the second at the moment, and I failed to come up with a reasonably sized pipeline that would trigger the bug.

PR #4635, opened 2020-02-21
Remove eliminate_bool_vectors from the Hexagon backend

### LINKED ISSUE #4634: Hexagon boolean vectors changed in LLVM
https://reviews.llvm.org/rGb1d47467e261 changed Hexagon boolean vectors in LLVM, which is breaking Halide codegen. Boolean vectors now have one bit per byte of operands, instead of one bit per bit.

This is a bit tricky for us to support, because the intrinsics still only use 'i32' types, so these new boolean vectors add a third "kind" of type to deal with in intrinsics (i32 containers, the actual vector type, and now the i1 per byte predicate vectors).

dsharletg:
LLVM change https://reviews.llvm.org/rGb1d47467e261 made it so the Hexagon backend now expects actual bool vectors, rather than masks. This means we need to change our masks from having one bit per bit of vector data to one bit per byte. However, this is somewhat complicated to implement.

At the same time, the Hexagon LLVM backend is now better at handling standard LLVM assembly like `select`. Therefore, a workaround for the above commit is to just stop using `eliminate_bool_vectors` for the Hexagon backend, and change some uses of intrinsics over to be standard LLVM assembly (e.g. use `select` instead of `vmux` intrinsics).

This PR is currently potentially a performance regression because we can't deinterleave quite as many operations as we used to (because we can't interleave/deinterleave bool vectors any more). This can probably be worked around by making the `EliminateInterleaves` visitor a bit smarter, but I think this should wait until later. The above LLVM change is a breaking change, we should get things back to functional, and improve performance later.

Fixes #4634

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -988,53 +988,6 @@ static const HvxIntrinsic intrinsic_wrappers[] = {
      {i16v1, i16},
      HvxIntrinsic::BroadcastScalarsToWords},
 
-    // Select/conditionals. Conditions are always signed integer
-    // vectors (so widening sign extends).
-    {MAKE_ID_PAIR(Intrinsic::hexagon_V6_vmux),
-     i8v1,
-     "mux.vb.vb",
-     {i1v1, i8v1, i8v1}},
-    {MAKE_ID_PAIR(Intrinsic::hexagon_V6_vmux),
-     i16v1,
-     "mux.vh.vh",
-     {i2v1, i16v1, i16v1}},
-    {MAKE_ID_PAIR(Intrinsic::hexagon_V6_vmux),
-     i32v1,
-     "mux.vw.vw",
-     {i4v1, i32v1, i32v1}},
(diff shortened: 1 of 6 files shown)


abadams:
Doesn't think break everything for older llvms? What's the expectation for using non-trunk llvm for hvx work?

dsharletg:
Note that this PR is not a direct API workaround for the breaking change, rather this change could have been made any time after the Hexagon LLVM backend started handling things like vector `select` (instead of having to use `vmux` intrinsics). So this likely works going back quite a ways.

That said, we also discussed on a recent PR that we only test and use the Hexagon backend on recent LLVMs currently, and that QC only wants to support that (but I can't find that issue/PR at the moment).

dsharletg:
Confirmed this works prior to the breaking LLVM commit at least.

abadams:
Ah ok, that's a much less dramatic change then

dsharletg:
I rewrote the PR description to be a bit more clear.

PR #4639, opened 2020-02-24
Increase stack size for wasm tests as needed

steven-johnson:
CodegenLLVM deliberately makes many small (register-sized) calls to alloca, on the plausible assumption that CPU-ish backends will promote these into registers. This isn't true for the WebAssembly backend, since (as a stack machine) it doesn't do registers; in degenerate cases (e.g. a func with ~everything inlined, and tracing enabled to increase stack usage), we can overflow the default wasm stack of ~64k, which crashes in amusing ways.

To mitigate this, add some code to CodegenLLVM to track the total alloca usage, then have WasmExecutor add this amount to the stack-size specified at link time.

Note 1: AFAICT there isn't a pre-baked LLVM method for "how much stack space does this Module require"; if there is a cleaner way to do this without tracking it ourself, that would be great.

Note 2: would probably also be profitable to see if we can avoid the 'assume alloca -> register' design for the wasm backend; even if we do that, this fix is probably still desirable for other possible degenerate cases of wasm.

Note 3: drive-by fix in WasmExecutor for LLVM >= 11.0 error case of linker.

### DIFF
--- a/src/CodeGen_LLVM.h
+++ b/src/CodeGen_LLVM.h
@@ -438,9 +438,18 @@ protected:
                                         bool zero_initialize = false,
                                         const std::string &name = "");
 
-    /** Total size of all alloca() storage requested (including alignment padding).
-     * Note that the actual stack space used might be less, as LLVM will promote
-     * register-sized allocas into registers in many cases. */
+    /** A (very) conservative guess at the size of all alloca() storage requested
+     * (including alignment padding). It's currently meant only to be used as
+     * a very coarse way to ensure there is enough stack space when testing
+     * on the WebAssembly backend.
+     *
+     * It is *not* meant to be a useful proxy for "stack space needed", for a
+     * number of reasons:
+     * - allocas with non-overlapping lifetimes will share space
+     * - on some backends, LLVM may promote register-sized allocas into registers
+     * - while this accounts for alloca() calls we know about, it doesn't attempt
+     *   to account for stack spills, function call overhead, etc., so
+     */
     size_t requested_alloca_total = 0;
 
     /** Which buffers came in from the outside world (and so we can't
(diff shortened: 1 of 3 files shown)


src/CodeGen_LLVM.h
```diff
@@ -434,6 +438,11 @@ class CodeGen_LLVM : public IRVisitor {
                                         bool zero_initialize = false,
                                         const std::string &name = "");
 
+    /** Total size of all alloca() storage requested (including alignment padding).
```
abadams: It will be substantially less, because allocas with non-overlapping lifetimes will share space. So in a pipeline with n stages doing stack allocations of size m, we'll see O(n*m) bytes allocated on the stack, but probably only use O(m) stack. Register promotions is a smaller effect.

So I suggest tweaking the comment to make it clear that this is a conservative upper bound.

dsharletg: I think this isn't guaranteed to be an upper bound, there are probably other things that go on the stack that aren't being accounted for (function call overhead, stack spills from register allocation, probably other things).

steven-johnson: PTAL

src/CodeGen_LLVM.h
```diff
@@ -434,6 +438,20 @@ class CodeGen_LLVM : public IRVisitor {
...
+     * (including alignment padding). It's currently meant only to be used as
+     * a very coarse way to ensure there is enough stack space when testing
+     * on the WebAssembly backend.
+     *
+     * It is *not* meant to be a useful proxy for "stack space needed", for a
+     * number of reasons:
+     * - allocas with non-overlapping lifetimes will share space
+     * - on some backends, LLVM may promote register-sized allocas into registers
+     * - while this accounts for alloca() calls we know about, it doesn't attempt
+     *   to account for stack spills, function call overhead, etc., so
```
abadams: lgtm except for trailing "so"

steven-johnson: done

PR #4644, opened 2020-02-25
Modernize CMake

### LINKED ISSUE #870: Teach Halide CMake build to export Halide library target
This will allow other CMake projects to import the Halide library (however it's built, e.g. dynamic or static). It also allows a required libraries/compiler flags to be exported to the client if need be. We can also have the `HalideGenerator.cmake` automatically imported for the client when they do `find_package(Halide CONFIG REQUIRED)`.

A use case for this is `app/wavelet` which tries to be standalone but doesn't quite succeed due to the static in source build assumption.

I've done this before for the [STP project](https://github.com/stp/stp) so it should be fairly straight forward to do.

### LINKED ISSUE #2643: python binding can't be built
Halide/python_bindings can not be built with VisualStudio 2017, please fix.

### LINKED ISSUE #2821: Mac OSX cmake libHalide.dylib path error
I'm working on mac OSX and trying to compile my small project with cmake. I wrote my CMakeLists.txt as follows:
```shell
# my CMakeLists.txt
cmake_minimum_required(VERSION 3.9)
set(CMAKE_CXX_STANDARD 11)

# path to distribution of halide
set(HALIDE_DISTRIB_DIR "/my/halide/path")
include("${HALIDE_DISTRIB_DIR}/halide_config.cmake")
include("${HALIDE_DISTRIB_DIR}/halide.cmake")

halide_library(blf
    SRCS bilateral_filter_generator.cpp
    GENERATOR_ARGS auto_schedule=true)

add_executable(process process.cpp)
target_link_libraries(process PUBLIC blf)
```

The problem is, when I finished cmake and start make, it reports an error:
```shell
$ make
Scanning dependencies of target halide_library_runtime.generator_binary
[ 12%] Building CXX object CMakeFiles/halide_library_runtime.generator_binary.dir/path/to/halide/tools/GenGen.cpp.o
[ 25%] Linking CXX executable halide_library_runtime.generator_binary
[ 25%] Built target halide_library_runtime.generator_binary
Scanning dependencies of target halide_rt_host_runtime_gen
dyld: Library not loaded: /Users/halidenightly/build_bot_new/worker/mac-64-trunk/halide-build/bin/libHalide.dylib
  Referenced from: /path/to/my/project/build/./halide_library_runtime.generator_binary
  Reason: image not found
make[2]: *** [genfiles/halide_rt_host/halide_rt_host.a] Abort trap: 6
make[1]: *** [CMakeFiles/halide_rt_host_runtime_gen.dir/all] Error 2
make: *** [all] Error 2
```

Note that it tried to find `libHalide.dylib` at `/Users/halidenightly/build_bot_new/worker/mac-64-trunk/halide-build/bin/libHalide.dylib`, which obviously is *NOT* path on my mac. Looks like it's a path on a building server.

I'm new to cmake and I have no idea about how to figure out what is going wrong. I guess there may be some mistakes in those cmake functions defined in `halide.cmake`? Any one has met similar problems?

P.S. It is totally OK and works well when I use hand-write g++ command to build my project, as follows:
```shell
# my build.sh
g++ bilateral_filter_generator.cpp /path/to/halide/tools/GenGen.cpp -g -std=c++11 -fno-rtti \
    -I /path/to/halide/include -L /path/to/halide/bin -lHalide -lpthread -ldl -o generate
DYLD_LIBRARY_PATH=/path/to/halide/bin ./generate -g blf -o . target=host auto_schedule=true
g++ process.cpp blf.a -g -std=c++11 -fno-rtti -I /path/to/halide/include \
    -I /path/to/halide/tools -L /path/to/halide/bin -lHalide -o process
```

### LINKED ISSUE #2852: auto schedule for GPU
I was wondering can I do auto scheduling for GPUs?
I have simple halide code:
```cpp
  for (int y = 0; y < size; y++) {
    for (int x = 0; x < size; x++) {
      A(x, y) = rand() & 0xfff;
    }
  }
  Var j("j"), i("i"), k("k");
  Func out4("out4");
  out4(i, j) = A(i, j) * 2;

  Target target = get_target_from_environment();
  target.set_feature(Halide::Target::CUDACapability35);
  target.set_feature(Halide::Target::CUDA);

  Pipeline p(out4);
  out4.estimate(i, 0, size).estimate(j, 0, size);
  cout << p.auto_schedule(target);
```
I get schedule:
``` cpp
// Target: x86-64-linux-avx-avx2-cuda-cuda_capability_35-f16c-fma-sse41
// MachineParams: 16,16777216,40
Var i_vi("i_vi");
Var i_vo("i_vo");
Func out4 = pipeline.get_func(0);
{
    Var i = out4.args()[0];
    Var j = out4.args()[1];
    out4
        .compute_root()
        .split(i, i_vo, i_vi, 8)
        .vectorize(i_vi)
        .parallel(j);
}
```
Now even though the Target has feature: cuda, the code is not running on GPU.
Am I missing something?

### LINKED ISSUE #2942: Error at apps/linear_algebra/benchmarks/CMakeLists.txt:41 (add_executable):   Target "eigen_benchmarks" links to target "Eigen3::Eigen" but the target   was not found.
Getting the following error when I try to build with CMake:

CMake Error at apps/linear_algebra/benchmarks/CMakeLists.txt:41 (add_executable):
  Target "eigen_benchmarks" links to target "Eigen3::Eigen" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

CMake Error at apps/linear_algebra/benchmarks/CMakeLists.txt:41 (add_executable):
  Target "eigen_benchmarks" links to target "Eigen3::Eigen" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

### LINKED ISSUE #3658: windows releases do not include halide.cmake (and related files)
Windows builds do not include halide.cmake (and related files). Including these files would make working with halide on windows much easier (particularly now VS2017 has native cmake support). But it seems to depend on some missing config files, which are not generated for windows?

### LINKED ISSUE #4009: Add Bazel support back?

### LINKED ISSUE #4284: [RFC] CMake buildsystem QoL improvements ?

### LINKED ISSUE #4476: -std=gnu++98 does not imply "no thread-syncing on static locals"
A Googler comments:

>-std=gnu++98 has no effect on whether the compiler produces synchronization for function-local statics.  This is because gcc had the support prior to the standard anyway.  c++11 just mandated the behavior for everyone; this affected primarily Microsoft.
>
>Example on Haswell x64: https://gcc.godbolt.org/z/Dzx9tU
>
>Example on RISC-V: https://gcc.godbolt.org/z/ATXUYH
>
>If you don't want synchronization, the flag you should use is -fno-threadsafe-statics, example => https://gcc.godbolt.org/z/8ej1iO

### LINKED ISSUE #4581: CMake static-library build is broken on Linux
On Linux, configure a build with HALIDE_SHARED_LIBRARY=OFF, then "make all -k". You'll fail a fair number of targets with "cannot find -lxml2"

### LINKED ISSUE #4890: Deprecate Visual Studio 2015
@steven-johnson and I were contemplating dropping support for Visual Studio 2015.

LLVM 10 has dropped VS 2015 support, so we're going to be forced to do so soon, anyway.

Does anyone still need to build Halide with VS 2015?

### LINKED ISSUE #4893: Unable to run autoscheduler program in Windows using Visual Studio
The following program 
[https://halide-lang.org/tutorials/tutorial_lesson_21_auto_scheduler_generate.html](url)

is compiled on Linux with the following command:
g++ lesson_21_auto_scheduler_generate.cpp ../tools/GenGen.cpp -g -std=c++11 -fno-rtti -I ../include -L ../bin -lHalide -lpthread -ldl -o lesson_21_generate

No instructions are given to compile it in Windows.

### LINKED ISSUE #4895: recompile with -fPIC on test_correctness_generator in CMake
I am aware that there is a new CMake PR in progress via https://github.com/halide/Halide/pull/4644

For the time being I had similar errors like https://github.com/halide/Halide/issues/3747 with the suggestion to "recompile with -fPIC".

This happened during the build of Halide on a fresh Fedora 32. The errors appeared on the python bindings tests for `addconstant` and `user_context` in `python_bindings/correctness/CMakeLists.txt`

Adding the `POSITION_INDEPENDENT_CODE ON` property for the static libs within `halide.cmake` fixed it for me.

```
  set_target_properties("${BASENAME}_cc" PROPERTIES
    POSITION_INDEPENDENT_CODE ON
    EXCLUDE_FROM_ALL TRUE)
```

### LINKED ISSUE #4943: CMake apps build changes on cmake-modern
In an effort to isolate the apps from the rest of the build, and to test our eventual distribution packages, the apps' CMakeLists.txt files were made _entirely standalone_ from the rest of the build. They call `find_package(Halide)` to load Halide and its special build rules for generators. They also define tests to run the apps on input images and ensure that nothing crashes.

The issue is that _building_ the apps therefore becomes part of _testing_ the app from the main build's perspective and @steven-johnson wants a way to validate that everything builds in a single command (ie. `cmake --build .`). Installing yourself and triggering external builds is not really ergonomic with CMake unless we want to switch to a super-build structure, which would require some filesystem rearrangement I don't think anyone particularly wants.

The compromise I came up with was to add a flag, `APPS_INTERNAL`, which changes the behavior to adding the apps (and their tests) to the main build and NOT testing them externally. By default, it is off, meaning that apps only rebuild when they are run as tests.

If anyone has a better solution, this issue is the place to discuss it before landing #4644 .

### LINKED ISSUE #4948: Standardize test output
Our tests (those in `/tests/`) try to report success via return code 0 and printing a "Success" message. Other tests "succeed" but actually don't run for environmental reasons (e.g. no cuda in HL_TARGET). Detecting these cases should be easy to do by hand.

When a test succeeds, it should print a line containing the string "Success!". When a test fails, it should return non-zero and not print "Success!". When a test chooses not to run, it should print a line beginning with "[SKIP]" and listing a reason for skipping.

In the special case of error tests, they should "succeed" by returning non-zero and "fail" by printing "Success!". The surrounding test infrastructure should reverse the actual condition for these tests.

In the special case of error tests, they should never print "Success!" but should instead print at least one line beginning with "Warning:".

alexreinking:
This PR overhauls the CMake build to use modern features and improve our ability to be used/included by other projects. Note that a recent version of **CMake (3.16+)** is now required.

### New features:

1. Clients now use `find_package(Halide)` to use Halide.
2. Clients can alternately include Halide as a Git submodule and simply `add_subdirectory` our root. This also works with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html), which automates cloning a git repo and downloading it to the build dir.
3. Apps are built in the same way our clients would build them. Each one has an associated test that builds it and confirms that the app runs without crashing. See [apps/lens_blur/CMakeLists.txt](https://github.com/halide/Halide/blob/cmake-modern/apps/lens_blur/CMakeLists.txt) for an example.
4. Python bindings are now built on all platforms.
5. Tests build with precompiled headers.
6. The new autoschedulers are packaged with the build and conveniently available from CMake.
7. All of the tutorials are now built and tested (they weren't before!)

### Changes outside CMake:

1. Documentation is handled by `doc/CMakeLists.txt`. There is no more Doxyfile.
2. Tests no longer have include access to the whole source tree. They can see the runtime includes and the test/common includes only. The affected sources and the Makefile have been modified accordingly.
3. OpenGL tests use system headers (as a consequence of the previous)
4. Tests in the `tests/` folder must now obey the following rules:
  a. Print "Success!" upon success unless it is an error or warning test.
  b. Print a message beginning with "[SKIP]" if the test will not run for some reason (e.g. no GPU).

### Breaking changes:

1. `halide.cmake` and its associated functions are gone. 
2. Generator stubs are no longer supported from CMake.
3. CMake 3.16+ required. This is the default version on Ubuntu 20.04 LTS and Visual Studio 2019.

### TO-DOs (sorted by priority):

These are safe to delay to subsequent PRs.

- [x] Update buildbots to use CMake build.
- [ ] Packaging needs to be fleshed out and tested
  -  [ ] Test exporting and installing the filter libraries produced by generators.
- [ ] Suppress harmless warnings on Windows
  - [ ] CRT library calls
  - [ ] "object specified more than once; extras ignored"
- [ ] Suppress harmless warnings on Linux
  - [ ] `-Wno-psabi`

----

Fixes #870
Fixes #2643
Fixes #2821
Fixes #2852
Fixes #2942
Fixes #3658
Fixes #4009
Fixes #4284
Fixes #4476
Fixes #4581
Fixes #4890
Fixes #4893
Fixes #4895
Fixes #4943
Fixes #4948


Makefile
```diff
@@ -2163,17 +2161,6 @@ $(BUILD_DIR)/rtti_ok:
 	@exit 1
 endif
 
-.PHONY: doc
```
abadams: What's the story here?

alexreinking: The top-level Doxyfile and Doxyfile.in were... confusing. I'm not sure `Doxyfile` was actually used for anything. The Makefile just sed-ed into Doxyfile.in, but when that was removed to use CMake's newer Doxygen module, I removed it from the Makefile, too.

alexreinking: See `doc/CMakeLists.txt` for the list of non-default options that were set by Doxyfile.in.

test/correctness/debug_to_file_multiple_outputs.cpp
```diff
@@ -51,15 +51,17 @@ int main(int argc, char **argv) {
     assert(f && g && h);
 
     int header[5];
-    assert(fread((void *)(&header[0]), 4, 5, f) == 5);
```
abadams: Might be good to pull the fix to the side-effecty asserts into a separate PR. I believe it's the only substantial code change (the rest is just messing with headers)?

alexreinking: Yes. I can cherry pick those changes and move them to a separate PR.

steven-johnson:
(Note: normally I know we don't bother rebasing commits before landing, but in this case, I suspect it will make tracking down issues after the fact simpler; once this gets close(r) to landing, definitely consider collapsing this all into a single commit)

Makefile
abadams: Does this leave us unable to build the docs using make?

alexreinking:
> (Note: normally I know we don't bother rebasing commits before landing, but in this case, I suspect it will make tracking down issues after the fact simpler; once this gets close(r) to landing, definitely consider collapsing this all into a single commit)

Ok!

Makefile
abadams: I currently use make to generate docs for releases (you know, once a year or so when I actually do one). I suppose I could switch to scraping them from a CI distro, or just use cmake. But this makes make/cmake not parallel build systems. The cmake build would be necessary because we cut a feature from the make build. Not sure how I feel about that.

test/correctness/CMakeLists.txt
```diff
@@ -1,324 +1,378 @@
...
-        widening_reduction.cpp
-        )
+      SOURCES
+      align_bounds.cpp
+      argmax.cpp
+      assertion_failure_in_parallel_for.cpp
+      async_copy_chain.cpp
+      async.cpp
+      async_device_copy.cpp
+#      atomics.cpp
```
steven-johnson: Why is this commented out?

tools/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+##
+# Build time tools
```
steven-johnson: I read this as "build (time tools)" and thought WHOA WE HAVE TIME TOOLS

alexreinking:
I opened a bug with LLVM that inquires about their lack of a ClangConfigVersion.cmake

https://bugs.llvm.org/show_bug.cgi?id=45027

steven-johnson:
> Prefer generator expressions to conditional execution of commands

Can you expand a bit on why these are preferred?

steven-johnson:
> Test on macOS

I just tried this on mine, with my usual cmake invocation: `cmake -DLLVM_DIR=`${LLVM_CONFIG} --obj-root`/lib/cmake/llvm -DCMAKE_BUILD_TYPE=Release -DHALIDE_SHARED_LIBRARY=ON ..`

and it fails quickly (for recent-trunk LLVM) with

```
CMake Error at dependencies/clang/CMakeLists.txt:1 (find_package):
  By not providing "FindClang.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Clang", but
  CMake did not find one.

  Could not find a package configuration file provided by "Clang" with any of
  the following names:

    ClangConfig.cmake
    clang-config.cmake

  Add the installation prefix of "Clang" to CMAKE_PREFIX_PATH or set
  "Clang_DIR" to a directory containing one of the above files.  If "Clang"
  provides a separate development package or SDK, be sure it has been
  installed.
```

alexreinking:
@steven-johnson for now add:
```
-DClang_DIR=$(${LLVM_CONFIG} --obj-root)/lib/cmake/clang
```
to the cmake command line. That's what the bug I opened with LLVM is about.

alexreinking:
> > Prefer generator expressions to conditional execution of commands
> 
> Can you expand a bit on why these are preferred?

@steven-johnson 

Yes, it's because they can be installed for end-users as-is. An example where this is _necessary_ is with the Halide::ImageIO target that only links to PNG if it is available and only defines HALIDE_NO_PNG if it isn't.

.clang-format
```diff
@@ -1,17 +1,17 @@
 ---
 AccessModifierOffset: -4
```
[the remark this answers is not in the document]
alexreinking: Cherry picked in #4654

CMakeLists.txt
```diff
@@ -1,485 +1,109 @@
...
-  if(COMMAND add_link_options)
-    add_link_options(-stdlib=libc++)
-  else()
-    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -stdlib=libc++")
-    set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -stdlib=libc++")
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++")
-  endif()
-endif()
+# TODO: reevaluate this. We should probably be selective about what we export.
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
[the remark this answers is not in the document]
alexreinking: Discussion started in #4651

test/correctness/CMakeLists.txt
alexreinking: Because it depends on knowing the LLVM_VERSION Halide is being built against and I haven't gotten that properly propagated through yet. Adding that to the top-level TODOs on this PR.

steven-johnson:
> @steven-johnson for now add:
> 
> ```
> -DClang_DIR=$(${LLVM_CONFIG} --obj-root)/lib/cmake/clang
> ```
> 
> to the cmake command line. That's what the bug I opened with LLVM is about.

Now I get about halfway thru the build, then fail with

`error: unknown warning option '-Wsuggest-override'; did you mean '-Wshift-overflow'?`

alexreinking:
> `error: unknown warning option '-Wsuggest-override'; did you mean '-Wshift-overflow'?`

What compiler are you using? AppleClang? I can remove that flag for that compiler.

steven-johnson:
> What compiler are you using? AppleClang? I can remove that flag for that compiler.

Yep, the version that comes with XCode 10.x

alexreinking:
> (Note: normally I know we don't bother rebasing commits before landing, but in this case, I suspect it will make tracking down issues after the fact simpler; once this gets close(r) to landing, definitely consider collapsing this all into a single commit)

I rebased into a much smaller number of commits that at least preserves some history of the broader changes I made. We can squash down in to one mega-commit later if you still want that.

Makefile
alexreinking: Being a build target integrates better with CPack for packaging the documentation along with the binary releases.

Makefile
```diff
@@ -2231,8 +2218,7 @@ $(DISTRIB_DIR)/halide.tgz: $(LIB_DIR)/libHalide.a \
                            $(RUNTIME_EXPORTED_INCLUDES) \
                            $(ROOT_DIR)/README*.md \
                            $(BUILD_DIR)/halide_config.cmake \
-                           $(BUILD_DIR)/halide_config.make \
-                           $(ROOT_DIR)/halide.cmake
+                           $(BUILD_DIR)/halide_config.make
```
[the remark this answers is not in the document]
alexreinking: Sort of. It's now called `packaging/HalideConfig.cmake`, which is a standard name. It is run when a user calls `find_package(Halide)`.

apps/CMakeLists.txt
```diff
@@ -1,30 +1,58 @@
...
-# add_subdirectory(wavelet)
+##
+# Test apps from the perspective of a consuming project.
+##
+
+# Set up a test fixture for installing the main project into a build directory
+
+set(HALIDE_DIR "${CMAKE_CURRENT_BINARY_DIR}/_Halide_install")
+
+# TODO: if we can support CMake 3.15+, we can use ${CMAKE_COMMAND} --install ...
```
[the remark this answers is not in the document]
alexreinking: I am in favor of depending on the newest version, sure. 3.16 adds cool stuff like cross-platform precompiled headers.

apps/CMakeLists.txt
```diff
@@ -1,30 +1,58 @@
...
+add_app_test(conv_layer)
+add_app_test(cuda_mat_mul)
+#add_app_test(glsl)  # TODO: bugged on my computer; not built by Makefile
+add_app_test(harris)
+add_app_test(hist)
+add_app_test(iir_blur)
+add_app_test(interpolate)
+add_app_test(lens_blur)
+#add_app_test(linear_algebra)  # TODO: halide blas has issues; see #4631
+#add_app_test(linear_blur)  # TODO: uses stubs, for which we're dropping support
```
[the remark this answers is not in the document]
alexreinking: You're right that this comment is misleading. It isn't clear how to add them to the CMake support. They appear to make the common case (not using stubs) harder to use.

(also I'm getting deja vu here... why aren't my old responses appearing?)

apps/linear_algebra/CMakeLists.txt
```diff
@@ -1,54 +1,46 @@
...
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED YES)
+set(CMAKE_CXX_EXTENSIONS NO)
+
+# Find Halide
+find_package(Halide REQUIRED COMPONENTS Halide)
+
+# Find BLAS-es
+set(BLAS_TARGETS)
+set(BLAS_VENDORS OpenBLAS)  # TODO: add ATLAS and its weird extra linking requirements that
```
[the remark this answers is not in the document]
alexreinking: should read "[that] aren't included by the FindBLAS module"

dependencies/jpeg/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if (APPLE)
+    # This changes how find_xxx() commands work; the default is to find frameworks before
+    # standard libraries or headers, but this can be a problem on systems that have Mono
+    # installed, as it has a framework with the libjpeg and libpng  headers present -- so
+    # CMake finds the headers from Mono but the libraries from Homebrew, and hilarity ensues.
+    # Setting this to "last" means we always try the standard libraries before the frameworks.
+    set(CMAKE_FIND_FRAMEWORK LAST)
```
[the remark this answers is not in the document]
alexreinking: Seems like a good idea. Regardless, how many other packages' bad installation practices do we want to work around? I'm pretty sure you can point cmake at the right png/jpeg at the command line. Someone should file a bug with Mono.

apps/CMakeLists.txt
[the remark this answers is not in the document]
abadams: To second Benoit's complaint: Is 3.10 totally untenable? Requiring something not in the apt repo of the latest ubuntu LTS release is an extra burden for users.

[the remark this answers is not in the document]
alexreinking: Well, the Makefile build is still around so we aren't requiring it.

Version 3.10 doesn't support the packaging features I'm using and intend to use. Object libraries (like those used to build the runtime) are broken before 3.14, as well. Kitware provides official support for up-to-date CMake on Ubuntu LTS and when admin access isn't available, pip can be used to install the most recent version locally.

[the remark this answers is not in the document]
alexreinking: Also, there's a new Ubuntu LTS coming out in 8 weeks that will come with CMake 3.16.3.

Makefile
abadams: Building the docs is kinda niche, but as a general rule I think we should not be cutting features from the Makefile in this PR.

apps/CMakeLists.txt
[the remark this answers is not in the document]
abadams: It also means we have to keep supporting and testing everything in the makefile build for the forseeable future. I'm not willing to require people to manually acquire newer versions of build tools before they can even get started - that's one yak too many.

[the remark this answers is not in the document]
abadams: ok, so let's use 3.16.3 and just plan to continue to support and test the makefile build for a long time too.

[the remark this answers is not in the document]
alexreinking: > I'm not willing to require people to manually acquire newer versions of build tools before they can even get started - that's one yak too many.

Get started _using_ Halide or _contributing_ to it? Those are two different scenarios.

Our python bindings build explodes if `imageio` isn't installed. And it's included by default, so the Makefile build fails if you haven't satisfied the requirements.txt tucked away in `python_bindings`. We could literally add `cmake>=3.14` to that file and developers would have to run the _same_ number of commands.

Makefile
alexreinking: @abadams - just pushed doc target to Makefile.

apps/CMakeLists.txt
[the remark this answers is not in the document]
abadams: Building Halide. Currently building a recent llvm is the main hurdle. It would be great if we could use an older llvm, but it's not feasible for us to support lots of llvm versions.

I don't consider the state of the python bindings to be a good target :)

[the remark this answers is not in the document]
alexreinking: Ah, well, that story would change pretty substantially if we produced an APT or PIP package of Halide.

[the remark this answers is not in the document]
abadams: For context on why I'm so conservative on build stuff: I recently had to build a Halide branch from scratch on a quad-core cortex-a53 arm dev board with ubuntu 16 to test something. Anything not in the ubuntu repo I have to build from source, and it takes approximately forever. I reimaged it as ubuntu 18 just to avoid building llvm, because Steven already had a prebuilt for arm ubuntu 18. It was then just a matter of building libHalide. It was still quite painful. And then of course my bug wouldn't repro. Halide, even as a jit compiler, is popular in these sorts of constrained environments, so we have to be very careful about requiring anything new.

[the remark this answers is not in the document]
abadams: I'm not convinced that apt or pip would change life for the people currently building Halide. Aren't those the people who would use one of our releases?

steven-johnson:
> Wire through LLVM_VERSION to Halide's dependents somehow.  As HALIDE_LLVM_VERSION?

I think that `LLVM_VERSION` is already a halide-specific thing (LLVM never defines it IIRC) so either way is likely fine. Changing it would require changing all the usage in Halide so if we want to do that, probably better to do as a separate PR to reduce churn.

abadams:
Not sure if you're already addressed this, but the windows halide.cmake is slightly wrong - it assumes there are bin and lib folders that contain Halide.dll and Halide.lib, but in the distros there aren't. We may want to fix the distros. Not sure how that interacts with shipping Debug and Release in one distro though.

alexreinking:
Properly configuring CPack should just handle that.

abadams:
So after this PR we should change the buildbot script to use CPack to generate the windows distro?

alexreinking:
Yes, absolutely. I'll figure out a script that works for me first and then work with Steven to make sure the buildbots can run it.

steven-johnson:
(Can you merge from master at your convenience? This won't build with current LLVM trunk, but there are enough merge conflicts that it's probably wiser for you to make the changes.)

alexreinking:
> 
> 
> (Can you merge from master at your convenience? This won't build with current LLVM trunk, but there are enough merge conflicts that it's probably wiser for you to make the changes.)

Done!

alexreinking:
Win32 buildbot is failing because we need to install 32-bit Python 3.8, create a 32-bit virtualenv and load it before building. We might possibly need to set `Python3_ROOT_DIR`, too.

alexreinking:
I'm happy to say that building Halide on Ubuntu 20.04 is really as easy as installing the dependent packages from `apt` and running the most basic command possible:

```console
dev@host:~/Halide-build$ cmake -DCMAKE_BUILD_TYPE=Release ../Halide
dev@host:~/Halide-build$ cmake --build .
dev@host:~/Halide-build$ ctest
```

test/opengl/save_state.cpp
```diff
@@ -7,47 +7,21 @@ int main() {
...
 #else
 
-#include <cstring>
-#include <stddef.h>
-#include <stdio.h>
-#include <stdlib.h>
-
 #include "Halide.h"
 
-#include "runtime/mini_opengl.h"
```
abadams, line 17: This change seems orthogonal to the rest of the PR? The decision to not rely on the existence of functioning GPU API headers is intentional.

alexreinking: I understand that for the runtime, but why should the tests have to access to the internal runtime headers? It intersects with this PR since the CMake build detects the appropriate paths for functioning GPU API headers anyway, and this test wouldn't build without exposing the runtime headers to the tests.

Furthermore, since the change works with the Makefile build, where tested (only Linux), I'm not sure why we _shouldn't_ use a system header for a test.

All this is moot if we end up dropping OpenGL support, though. It isn't tested on Windows (where it doesn't even work) or macOS (where it also doesn't even work #4885 )

abadams: Reasonable points (and it probably *is* moot), but I didn't see how it was related to this PR, which is why I commented. 

Is the reason it's in this PR because the tests no longer have access to internal runtime headers? That would make sense.

apps/CMakeLists.txt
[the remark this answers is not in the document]
alexreinking: I wonder if cross compiling Halide would have been a better option in that case.

test/opengl/save_state.cpp
abadams: I believe it worked at one point in time on all platforms. It wasn't added to the buildbots, so it rotted.

alexreinking: > Is the reason it's in this PR because the tests no longer have access to internal runtime headers? That would make sense.

Yes. `mini_opengl.h` is in `src/runtime`, so it's not visible to the tests anymore.

alexreinking:
Windows builds are failing due to mis-configured Python. Need to set up two virtualenvs (one for 64-bit, another for 32-bit) and install the required packages in each.

shoaibkamil:
Should we explicitly mention #4581 as a known issue?  I likely won't be able to start fixing it for a few weeks.

alexreinking:
@shoaibkamil - is this still an issue with this branch? I could swear it worked fine when I tested it (at least on Linux)

shoaibkamil:
I just tried to check, but I'm having issues with cmake (cmake 3.14.5 on macOS):
```
...
-- Configuring done
CMake Error at src/CMakeLists.txt:396 (target_compile_options):
  Error evaluating generator expression:

    $<CXX_COMPILER_ID:GNU,Clang,AppleClang>

  $<CXX_COMPILER_ID> expression requires one or zero parameters.


CMake Error at src/CMakeLists.txt:403 (target_compile_options):
  Error evaluating generator expression:

    $<COMPILE_LANG_AND_ID:CXX,MSVC>

  Expression did not evaluate to a known generator expression
```

alexreinking:
@shoaibkamil  - oops that's a CMake 3.15 feature I let slip in. I should download an old version to test against...

alexreinking:
Well, actually, in a separate discussion, we talked about requiring 3.16 since that's the version in both Visual Studio 2019 and Ubuntu 20.04. Is it cumbersome for you to upgrade? If you have pip, you can `pip install cmake` or (better) you can use Kitware's APT repo for CMake if you're on Ubuntu.

shoaibkamil:
Easy to do, but of course the cmake files should be updated to require that version.  I'll upgrade and report back as to whether static libs are still a problem.

shoaibkamil:
With CMake 3.17.2, I get:
```
-- Configuring done
CMake Error in dependencies/llvm/CMakeLists.txt:
  Target "Halide_LLVM" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/Users/kamil/Documents/work/adobe/halide_master/dependencies/llvm/"

  which is prefixed in the source directory.


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```
Do I need to rebuild LLVM with this version of cmake?

alexreinking:
Huh. I've never kept LLVM inside the Halide source tree! Let me try to repro that locally and I'll push some fixes.

Thanks for spending some time on this!

shoaibkamil:
Well, my LLVM_DIR _isn't_ in my Halide source tree, so perhaps there's something wrong with my invocation:

```
 cmake -DWITH_TESTS=off -DWITH_APPS=off -DWITH_TUTORIALS=off -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=<my llvm dir> ..```

**Comment by alexreinking** [comment]
Can you do `ls <my llvm dir>`?

File: README.md [hunk 13]
```diff
@@ -92,45 +99,129 @@ like so:
...
 
-If you wish to use cmake to build Halide, the build procedure is:
+Follow the above instructions to build LLVM or acquire a suitable binary release.
+Then create a separate build folder for Halide and run CMake, pointing it to your LLVM installation.
 
-    % mkdir cmake_build
-    % cd cmake_build
-    % cmake -DLLVM_DIR=/path-to-llvm-build/lib/cmake/llvm -DCMAKE_BUILD_TYPE=Release /path/to/halide
-    % make -j8
+    % mkdir Halide-build
```
steven-johnson, line 110: Is `$PROJECTNAME-build` the convention for CMake? I've never seen this before.

README.md
```diff
@@ -92,45 +99,129 @@ like so:
...
+    % cd Halide-build
```
steven-johnson, line 111: Nit: Should we make Ninja the default recommendation here, rather than Make? (IMHO it is superior, but I wonder if it would be less-surprising to out users to assume they are using Make, and maybe add a comment that "of course it works with Ninja too"). Or is Ninja basically required for Windows usage?

README.md
```diff
@@ -92,45 +99,129 @@ like so:
...
+
+Create a separate build tree and call CMake with vcpkg's toolchain.
+This will build in either 32-bit or 64-bit depending on the environment script (`vcvars`) that was run earlier. 
+
+    D:\> md Halide-build
+    D:\> cd Halide-build
+    D:\Halide-build> cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=D:/vcpkg/scripts/buildsystems/vcpkg.cmake ..\Halide
+
+**Note:** If building with Python bindings on 32-bit (enabled by default), be sure to point CMake to the installation 
+path of a 32-bit Python3.
```
steven-johnson, line 164: Nit: line endings are a bit wonky. Might be nice to run all the edited .md files thru a formatter if you have one handy. If not, no worries.

README_cmake.md
```diff
@@ -2,226 +2,287 @@
...
 
-Halide code is produced in a two-stage build that may seem a bit unusual at
-first, so before documenting the build rules, let's recap: a *Generator* is a
-C++ class in which you define a Halide pipeline and schedule, as well as
-well-defined input and output parameters; it can be compiled into an executable
-(often referred to as a *Filter*) that efficiently runs the Halide pipeline.
+CMake provides robust support for both use cases. These new features rely on recent
+developments in CMake, so a recent version is necessary. In order to build Halide,
+you would ideally have the newest version of CMake installed on your system (3.17
+at time of writing). However, everything should work with version 3.16 or newer.
```
steven-johnson, line 11: Be specific here as to what the minimum CMake version required is (not just "should work")

packaging/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+
+set(HALIDE_INSTALL_CMAKEDIR
+    "${CMAKE_INSTALL_LIBDIR}/cmake/Halide"
+    CACHE STRING "Path to Halide cmake files")
+
+##
+# Main library exports
+##
+
+# TODO: fix when autoschedulers are refactored
```
steven-johnson, line 16: Are there tracking issues for the new TODOs inserted? If not, now is probably a good time to add them (and add the issue URL as a comment)

test/generator/CMakeLists.txt
```diff
@@ -1,301 +1,276 @@
...
-# executables, one containing the generator itself
-# (e.g. from example_generator.cpp) and used at build time, and the other, the
-# test that executes the generated code (e.g. from example_aottest.cpp).
-
-# Tests with no special requirements
 halide_define_aot_test(acquire_release)
 halide_define_aot_test(argvcall)
 halide_define_aot_test(can_use_target)
-halide_define_aot_test(cleanup_on_error)
+#halide_define_aot_test(cleanup_on_error)  # requires access to internal header runtime/device_interface.h
```
steven-johnson, line 189: this presumably needs a TODO?

README.md
alexreinking: I don't think it's a convention. You can name your build directory anything you want. This is what I have taken to doing because I prefer to do fully out-of-tree builds.

README.md
alexreinking: Is there one you recommend? I don't have a Markdown formatter handy.

README.md
alexreinking: > Should we make Ninja the default recommendation here, rather than Make?

Debatable. I like it much better for the faster incremental builds and cleaner output. It's also the only build system that works just as well on all major platforms.

> Or is Ninja basically required for Windows usage?

It is not required on Windows at all. I've tested with both the Ninja and the Visual Studio generators.

packaging/CMakeLists.txt
alexreinking: There are not, yet. I'll have to go through and make those.

test/generator/CMakeLists.txt
alexreinking: Oops, yes it does.

README.md
steven-johnson: Maybe just add text to the effect of what you said here ("We recommend Ninja because blah, but Make and MSBuild work fine too")

README.md
alexreinking: I just reformatted this file manually... 80 col max for prose, 120 for code.

README.md
alexreinking: I dropped the `-G Ninja` in the main command and just mentioned that it's an option for Mac OS & Linux.

alexreinking:
`linux-64-gcc53-testbranch-cmake-1000` can't find CUDA for some reason. It is not installed in the usual spot (`/usr/local/cuda`)?

steven-johnson:
I think this is finally ready to land, pending green

alexreinking:
# Current failures:

1. One of the Linux bots (linux-worker-2) has an old GPU / broken CUDA 9.0 installation and we get "Error: CUDA: cuModuleLoadData failed: CUDA_ERROR_INVALID_PTX" in the `cuda_mat_mul` app.
  a. The other Linux bot uses the newest driver (440) and CUDA (10.2) and works fine.
  b. Apparently it's a Quadro 600 (GF108GL), which only has CUDA Compute Capability 2.1 while this test requires 5.0
2. On the same buildbot, the test `halide_dgemm_transB_64 ` fails with the error: "Error: Input buffer B_ is accessed at -4, which is before the min (0) in dimension 0"
3. Windows 32-bit is missing numpy and imageio dependencies in Python. This likely requires a virtualenv.

abadams:
It just occurred to me that for cuda, opencl, and d3d12, we don't need to mention them in the build files at all (with the exception of cuda_mat_mul). The runtime dlopens those if it needs them. You don't need to link the generated code against them. So it should be possible to just delete a large chunk of the generator cmake file and avoid any problems with people having libcuda installed somewhere cmake can't find it.

alexreinking:
> It just occurred to me that for cuda, opencl, and d3d12, we don't need to mention them in the build files at all (with the exception of cuda_mat_mul). The runtime dlopens those if it needs them. You don't need to link the generated code against them. So it should be possible to just delete a large chunk of the generator cmake file and avoid any problems with people having libcuda installed somewhere cmake can't find it.
> 
> Edit: two of the generator tests also use the cuda or opencl API directly: acquire_release and define_extern_opencl

Handled.

Makefile
```diff
@@ -2276,7 +2261,7 @@ $(DISTRIB_DIR)/halide.tgz: $(LIB_DIR)/libHalide.a \
 		halide/tutorial \
 		halide/README*.md \
 		halide/halide_config.* \
-		halide/halide.*
+#		halide/halide.*
```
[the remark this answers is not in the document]
abadams: What's going on here?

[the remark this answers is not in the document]
alexreinking: No more `halide.cmake`. Will just delete that line and preceding backslash before merging.

alexreinking:
The only outstanding failure is with Halide's dgemm transB on size 64 when compiling with (say) `HL_TARGET=x86-64-linux`. Landing...

PR #4649, opened 2020-02-26
Work around llvm bug

abadams:
simd_op_check fails on arm64 with newer llvms

See https://bugs.llvm.org/show_bug.cgi?id=45036


abadams:
FYI @alinas

steven-johnson:
Failure is a flaky performance test, ready to land.

PR #4656, opened 2020-02-27
Fix JIT-on-x86-32-under-CPython runtime error

steven-johnson:
This is a workaround for an odd corner case (cross-compiling + testing Python bindings x86-32 on an x86-64 system): 

`__udivdi3` is a helper function that GCC uses to do u64/u64 division on 32-bit systems; it's usually included by the linker on these systems as needed. 

When we JIT, LLVM will include references to this call; MCJIT fixes up these references by doing (essentially) `dlopen(NULL)` + `dlsym()` to look up the symbol. For normal JIT tests, this works fine, as `dlopen(NULL)` finds the test executable, which has the right lookups to locate it inside libHalide.so. 

If, however, we are running a JIT-via-Python test, `dlopen(NULL)` returns the CPython executable... which apparently *doesn't* include this as an exported function, so the lookup fails and very confusing crashiness ensues. 

Our workaround here is a bit icky, but expedient: check for this name if we can't find it elsewhere, and if so, return the one we know should be present.

(Also: drive-by fix to test/correctness/buffer.py for 32-bit systems to avoid allocation failures.)

### DIFF
--- a/python_bindings/correctness/buffer.py
+++ b/python_bindings/correctness/buffer.py
@@ -2,6 +2,7 @@
 import halide as hl
 import numpy as np
 import gc
+import sys
 
 def test_ndarray_to_buffer():
     a0 = np.ones((200, 300), dtype=np.int32)
@@ -85,6 +86,9 @@ def test_buffer_to_ndarray():
 def _assert_fn(e):
     assert e
 
+def _is_64bits():
+    return sys.maxsize > 2**32
+
 def test_for_each_element():
     buf = hl.Buffer(hl.Float(32), [3, 4])
     for x in range(3):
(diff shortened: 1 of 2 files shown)


abadams:
Do we know if there are other functions in the same class? It seems unlikely that there would be just one.

steven-johnson:
There are a dozen or more in that class, but that's the only one that I know turns up. I could pre-emptively add the others, but I'd rather err on the side of fixing as needed here.

abadams:
If you have the list, then I'd just add them all rather than wait for a new bug, but as you like.

steven-johnson:
This is the first time in $NUM years we've ever encountered this, I'm willing to gamble :-)

PR #4659, opened 2020-02-27
Add 'override' for virtual dtors as needed

steven-johnson:
Clang's `-Winconsistent-missing-destructor-override` warning catches these; we don't use this yet, but will after #4644 lands.


alexreinking:
We should consider adding some clang-tidy presubmit checks for this sort of thing.

PR #4662, opened 2020-02-29
Fix python tutorials

steven-johnson:
This is mostly a stylistic fix, but there is a real bugfix in lesson12 (which led to all of this):

-- the bugfix: in addition to PR #4661 (which this depends on), we needed to add an explicit call to `set_host_dirty()` for lesson12. (The C++ image loading code always does this under the hood, but in Python we just use imread.) This is an unsatisfying fix (users shouldn't have to call this), but apparently we don't bother calling copy-to-device if host-dirty isn't set... even if the buffer has no device bits. (This feels like a separate potential bug...)

The style fixes:
- remove weird leftover C++isms and commented-out dreck
- format everything according to PEP8
- ensure that all sanity checks are uniformly handled by assert-fail (rather than raising exceptions, quietly returning -1, etc)
- lesson12 had drifted in nontrivial ways from the C++ version; besides the bugfix, I brought them back into rough alignment

### DIFF
--- a/python_bindings/tutorial/lesson_12_using_the_gpu.py
+++ b/python_bindings/tutorial/lesson_12_using_the_gpu.py
@@ -5,39 +5,26 @@
 # This lesson demonstrates how to use Halide to run code on a GPU.
 
 # This lesson can be built by invoking the command:
-#    make tutorial_lesson_12_using_the_gpu
-# in a shell with the current directory at the top of the halide source tree.
-# Otherwise, see the platform-specific compiler invocations below.
+#    make test_tutorial_lesson_12_using_the_gpu
+# in a shell with the current directory at python_bindings/
 
-# On linux, you can compile and run it like so:
-# g++ lesson_12*.cpp -g -std=c++11 -I ../include -L ../bin -lHalide `libpng-config --cflags --ldflags` -lpthread -ldl -o lesson_12
-# LD_LIBRARY_PATH=../bin ./lesson_12
-
-# On os x:
-# g++ lesson_12*.cpp -g -std=c++11 -I ../include -L ../bin -lHalide `libpng-config --cflags --ldflags` -o lesson_12
-# DYLD_LIBRARY_PATH=../bin ./lesson_12
-
-#include "Halide.h"
-#include <stdio.h>
-#using namespace Halide
 import halide as hl
 
-# Include some support code for loading pngs.
-#include "image_io.h"
 import imageio
 import os.path
 
 # Include a clock to do performance testing.
-#include "clock.h"
 from datetime import datetime
 
-
 # Define some Vars to use.
-x, y, c, i, ii, xi, yi = hl.Var("x"), hl.Var("y"), hl.Var("c"), hl.Var("i"), hl.Var("ii"), hl.Var("xi"), hl.Var("yi")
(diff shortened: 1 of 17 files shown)


abadams:
no device bits and host not dirty means that the buffer is uninitialized host memory, suitable for writing an output into

abadams:
Can we permit longer lines? Some of the line breaks make things harder to read.

python_bindings/tutorial/lesson_12_using_the_gpu.py
```diff
@@ -255,72 +226,85 @@ def test_correctness(self, reference_output):
...
 
         print("CPU and GPU outputs are consistent.")
 
 
 def main():
     # Load an input image.
-    image_path = os.path.join(os.path.dirname(__file__), "../../tutorial/images/rgb.png")
+    image_path = os.path.join(os.path.dirname(
+        __file__), "../../tutorial/images/rgb.png")
     input = hl.Buffer(imageio.imread(image_path))
```
abadams: The Buffer constructor that takes whatever imread returns should probably set host dirty.

steven-johnson:
> Can we permit longer lines? Some of the line breaks make things harder to read.

Let me check and see -- I just applied standard PEP8 and didn't look to see if there were options.

> no device bits and host not dirty means that the buffer is uninitialized host memory, suitable for writing an output into

...this seems like an error condition for an input buffer. (or at the very least, something worth sniffing for in debug-runtime builds and warning about.)

> The Buffer constructor that takes whatever imread returns should probably set host dirty.

Yeah -- the only thing that gives me pause there is that the existing `Buffer<>` ctors in C++ that take pointer-to-existing-data all typically document themselves as "does not set dirty bit", so we have a discontinuity.

steven-johnson:
Sigh, it seems that PEP8 is also in the school of "80 columns should be enough for anybody", though it graciously allows up to 99 columns for team that agree on it. But that does mean the tooling allows varied width, so I'll crank it higher.

abadams:
Hrm. I wonder if not setting host dirty is really the most helpful default for the constructors that take pointers. Do you think it's possible to change without breaking people though?

steven-johnson:
> Hrm. I wonder if not setting host dirty is really the most helpful default for the constructors that take pointers. 

In C++ it seems to be "not taking ownership -> not setting host dirty"; of course, in Python, there isn't an equivalent concept, as we are required to take (shared) ownership of the `py::buffer` passed in.

> Do you think it's possible to change without breaking people though?

In C++? It seems certain this would somehow break things somewhere. In Python? Nah, we could almost certainly change it there. Just uncertain about the possible discontinuity.

abadams:
The case that would break in C++ is that we'd trigger a copy-to-device of an output buffer wrapping externally allocated uninitialized memory. This would be a silent performance drop. Hrm. 

I guess our options would be deprecating the current constructor, or just making the change and then also changing lowering to not copy-to-device buffers that are going to be entirely clobbered on device anyway. The second change seems like a win, but the reason we don't do that already is in case we're only writing to a crop of some larger buffer, and the copy would copy the larger buffer. It's possible that we can detect this. Outputs are only not entirely clobbered if undef() was used.

abadams:
I'll open an issue

abadams:
#4664

steven-johnson:
OK. In the meantime, I changed the Python code so that initializing one of our wrapped Buffer<>'s from a `py::buffer` (aka "whatever imread returns") defaults to setting host-dirty. PTAL.

PR #4663, opened 2020-02-29
Use clang-tidy to check for override flags

alexreinking:
This PR adds a presubmit check that validates that specific clang-tidy (version 9) checks pass. Currently this is limited to `modernize-use-override` which will let us get `-Wsuggest-override` (and equivalents) off the compiler command line. If there are other checks we want to run in the future, this will help enforce those without having to wait for a build to complete.

Will probably fiddle with this branch until I get the GitHub Workflow script right.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,7 @@
+---
+Checks: '-*,modernize-use-override'
+WarningsAsErrors: '*'
+HeaderFilterRegex: '.*'
+AnalyzeTemporaryDtors: false
+FormatStyle: 'file'
+...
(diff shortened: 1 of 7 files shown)


alexreinking:
This is basically an enhancement to #4659

steven-johnson:
So there's a bit of an issue with this PR: CMake has to have LLVM present because it looks for llvm-nm and llvm-as during processing. 

As written, we haven't bothered to install any version of LLVM on the bots... but it happens to work because the VMs have LLVM8 preinstalled. 

https://github.com/halide/Halide/pull/4665 revealed this to be an issue, because it fails if the LLVM version is 8.

The question is: how to fix this? I suppose we could install LLVM9 on the VMs, but that seems potentially heavyweight for this.

Issue #4606, opened 2020-02-13
tutorial lesson_12_using_the_gpu errors on macOS


mikewoodworth:
trying to invoke GPU tutorial on macOS Catalina, errors with the following output:

Running pipeline on CPU:
Running pipeline on GPU:
Target: x86-64-osx-avx-avx2-f16c-fma-metal-sse41
Testing GPU correctness:
Error: Metal: cannot allocate system default device.
Abort trap: 6

shoaibkamil:
On master, I see:
```
$ HL_JIT_TARGET=host-metal ./lesson_12
Running pipeline on CPU:
Running pipeline on GPU:
Requested GPU(s) are not supported. (Do you have the proper hardware and/or driver installed?)
No GPU target available on the host
Testing performance on CPU:
6.6550 milliseconds
```

It's not a crash, but it doesn't work.  I tried compiling with `-framework Metal` but that didn't seem to change anything.

mikewoodworth:
I haven't tried pre catalina yet, will do so just to confirm it isnt a security thing

shoaibkamil:
Also:
```
$  HL_JIT_TARGET=host-metal-debug make correctness_hello_gpu
Defining function...
Realizing function...
Entering Pipeline f
Target: x86-64-osx-avx-avx2-debug-f16c-fma-jit-metal-sse41-user_context
 Input (void *) __user_context: 0x7ffeea224e50
 Output Buffer f: buffer(0, 0x0, 0x0, 0, float32, {0, 32, 1}, {0, 32, 32})
Metal - Allocating: MTLCreateSystemDefaultDevice
Error: (at test/correctness/hello_gpu.cpp:23)
Metal: cannot allocate system default device.
bash: line 1: 38344 Abort trap: 6           /Users/kamil/Documents/work/adobe/halide_master/bin/correctness_hello_gpu
make: *** [correctness_hello_gpu] Error 134
```

Seems like something is broken on Catalina for Metal via JIT.

steven-johnson:
Could be an injection from https://github.com/halide/Halide/pull/4588 -- I don't get this locally on my Mac, but I'm not running Catalina.

shoaibkamil:
Using LLDB, I see this:

```
Process 23235 launched: '/Users/kamil/Documents/work/adobe/halide_master/bin/correctness_hello_gpu' (x86_64)
[...]

PR #4668, opened 2020-03-03
Fix Metal on macOS Catalina

shoaibkamil:
As far as I can tell, this is not documented (shocking!), but the debug messages (see #4606) point to the fact that command-line apps should not use `MTLCreateSystemDefaultDevice()`.  This just uses `MTLCopyAllDevices()` as fallback when `MTLCreateSystemDefaultDevice()` fails, and selects the first device returned.

### DIFF
--- a/src/runtime/metal.cpp
+++ b/src/runtime/metal.cpp
@@ -10,6 +10,7 @@
 
 extern "C" {
 extern objc_id MTLCreateSystemDefaultDevice();
+extern objc_id MTLCopyAllDevices();
 extern struct ObjectiveCClass _NSConcreteGlobalBlock;
 }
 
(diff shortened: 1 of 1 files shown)


shoaibkamil:
CC @mikewoodworth

steven-johnson:
LGTM but someone with more Metal expertise should probably approve

zvookin:
Also my guess is this is somewhat of a permissions thing. But it really doesn't matter to the change.

steven-johnson:
FYI: after this fix landed, my work-in-progress tests on the GitHubActions OSX VMs (running 10.15) *started* failing where they were working before; the failure looks like:

```
38/41 Test #480: lesson_12_using_the_gpu ....................Child aborted***Exception:   1.26 sec
2020-03-04 18:55:02.752 lesson_12_using_the_gpu[10479:38477] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArray0 objectAtIndex:]: index 0 beyond bounds for empty NSArray'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff3554638b __exceptionPreprocess + 250
	1   libobjc.A.dylib                     0x00007fff6b8bd552 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff35572b81 CFArrayExchangeValuesAtIndices + 0
	3   ???                                 0x0000000108638b49 0x0 + 4435708745
	4   ???                                 0x000000010863885b 0x0 + 4435707995
	5   ???                                 0x00000001086373dc 0x0 + 4435702748
	6   ???                                 0x00000001085cef65 0x0 + 4435275621
	7   ???                                 0x00000001085cee20 0x0 + 4435275296
	8   ???                                 0x00000001085cefd2 0x0 + 4435275730
	9   libHalide.dylib                     0x000000010a276738 _ZN6Halide27host_supports_target_deviceERKNS_6TargetE + 264
	10  lesson_12_using_the_gpu             0x0000000107df6c35 _Z15find_gpu_targetv + 277
	11  lesson_12_using_the_gpu             0x0000000107df47b4 _ZN10MyPipeline16schedule_for_gpuEv + 36
	12  lesson_12_using_the_gpu             0x0000000107df2fee main + 974
	13  libdyld.dylib                       0x00007fff6cc1f7fd start + 1
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

steven-johnson:
...they *seemed* to be passing. I can try locally reverting your change and forcing a re-run to verify.

steven-johnson:
I'm getting a report from someone indicating that they can't link (!) for iOS because `MTLCopyAllDevices` isn't defined there. If this is true, does this mean that we've been broken for iOS+Metal for ~2 months?

shoaibkamil:
If that's occurring, it's new behavior.  I did indeed test on iOS.

Let me craft a new PR that gates this via preprocessor macros <sigh>.

zvookin:
Likely a case of their minimum required OS being further backward than what
we're testing with or anyone is using.

PR #4670, opened 2020-03-03
Clang tidy performance

abadams:
Turn on a bunch of clang-tidy checks, add a makefile target for it, and accept a whole bunch of modernization fixes. I'm a little bit leery of having these checks on for the presubmits. We can roll them back if they get annoying.

Many of the tidy checks are aimed at performance. There is indeed a tiny effect: this PR speeds up lowering by about 0.2%. woo.


abadams:
The changes are mechanical and are only in src/. They are:

- Replace old-style for loops with range for loops where the index is not used
- Turn values into const refs if they're only used as a const ref
- Add std::move around the last use of a value
- Use default initializers for members, defaulting the default constructor where this covers everything that was initialized in it
- Make deleted operator= and copy constructors public
- Replace push_back(Foo(args)) with emplace_back(args)
- Replace typedef with using
- If a constructor is going to take a copy of an arg, take it by value and std::move it into the member instead of taking it by const ref and copying it into the member. This potentially saves a copy if the argument was an rvalue.

It required some hand-tweaks at the end to do things like name loop variants sanely. I think I can probably reproduce the mechanical transformations as separate PRs if you don't mind things like silly loop names temporarily, then we can return to this PR for the hand-tweaks. The diff will be much smaller.

abadams:
Closing these. I'm going to be much more choosy about what we turn on.

PR #4674, opened 2020-03-04
Alternative solution for #4668

shoaibkamil:
Use `firstObject` instead of `objectAtIndex:` so if the `NSArray` of devices is empty, it'll just return nil.

CC @steven-johnson


steven-johnson:
There was a buildbot burp; you might want to sync to head to force a rebuild

steven-johnson:
That seems to have fixed the issue for me.

PR #4676, opened 2020-03-04
turn on clang-tidy's modernize-pass-by-value

abadams:
This is a scaled-back #4670, just turning on a single additional check.

In constructors that take a copy of an argument, pass by value and
std::move instead of passing by const ref and making a copy. This is
strictly better for two reasons.

1) This avoids a copy in the case where the argument was an rval.
2) The type signature loses some line noise, and makes it more obvious
to callers that a copy is going to happen.

There were a healthy number of cases this revealed where the object
should *not* have been making a copy of the argument. These were largely
local IRVisitors and IRMutators doing some sort of tightly-scoped
analysis or mutation. so I changed those to have const ref members and
continue to take the argument by const ref.

Also added a makefile target to get clang-tidy to automatically apply
suggested changes.

### DIFF
--- a/src/AutoSchedule.cpp
+++ b/src/AutoSchedule.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <regex>
+#include <utility>
 
 #include "AutoSchedule.h"
 #include "AutoScheduleUtils.h"
@@ -92,7 +93,7 @@ struct FStage {
     Function func;
     uint32_t stage_num;
     FStage(Function func, uint32_t stage_num)
-        : func(func), stage_num(stage_num) {
+        : func(std::move(func)), stage_num(stage_num) {
     }
 
     bool operator==(const FStage &other_stage) const {
@@ -156,9 +157,9 @@ struct DependenceAnalysis {
         set<string> prods;
         bool only_regions_computed;
 
-        RegionsRequiredQuery(const string &f, int stage, const set<string> &prods,
+        RegionsRequiredQuery(string f, int stage, set<string> prods,
                              bool only_regions_computed)
-            : f(f), stage(stage), prods(prods),
+            : f(std::move(f)), stage(stage), prods(std::move(prods)),
               only_regions_computed(only_regions_computed) {
         }
 
@@ -190,17 +191,17 @@ struct DependenceAnalysis {
         // Regions required to compute 'bounds' given a particular
         // RegionsRequiredQuery.
         map<string, Box> regions;
-        RegionsRequired(const DimBounds &b, const map<string, Box> &r)
-            : bounds(b), regions(r) {
+        RegionsRequired(DimBounds b, map<string, Box> r)
+            : bounds(std::move(b)), regions(std::move(r)) {
         }
     };
     // Cache for bounds queries (bound queries with the same parameters are
     // common during the grouping process).
     map<RegionsRequiredQuery, vector<RegionsRequired>> regions_required_cache;
 
-    DependenceAnalysis(const map<string, Function> &env, const vector<string> &order,
-                       const FuncValueBounds &func_val_bounds)
-        : env(env), order(order), func_val_bounds(func_val_bounds) {
+    DependenceAnalysis(map<string, Function> env, vector<string> order,
+                       FuncValueBounds func_val_bounds)
+        : env(std::move(env)), order(std::move(order)), func_val_bounds(std::move(func_val_bounds)) {
     }
 
     // Return the regions of the producers ('prods') required to compute the region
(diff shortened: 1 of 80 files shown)


steven-johnson:
```
Internal Error at /home/travis/build/halide/Halide/src/CPlusPlusMangle.cpp:810 triggered by user code at :
Condition failed: mangled_name == expecteds[expected_index].expected:
Mangling for int32_t foo::test_function(int32_t, struct halide_buffer_t *) expected
    ?test_function@bar@foo@@YAHHPEAUhalide_buffer_t@@@Z got
    ?test_function@bar@foo@@YAHHPAUhalide_buffer_t@@@Z
for target x86-64-windows
```

src/CPlusPlusMangle.cpp
```diff
@@ -133,7 +134,7 @@ struct QualsState {
     bool last_is_pointer{false};
 
     const Type &type;
-    const std::string base_mode;
+    const std::string &base_mode;
```
steven-johnson: Gotta say that having const& member vars makes me a little nervous; I know it's safe with proper usage, but will clang-tidy (etc) notice danger if these classes get refactored to have different lifespans?

steven-johnson: ...and in fact, this specific change seems to have injected the name-mangling test failure; we are stored a reference to a *temporary* std::string in the instantiation on line 191. IMHO we should undo all changes of this sort unless there is also a clang-tidy check of some sort that will verify temporaries aren't ever used.

steven-johnson:
clang-tidy is giving us bad advice here.

src/CPlusPlusMangle.cpp
abadams: This change was me being dumb, not clang-tidy - clang-tidy wanted to std::move into a std:string member. I'll see if there's a checker for it though, as that would be useful.

steven-johnson: There were a number of other cases where a member var was changed to a const-ref too; these should probably changed back (unless/until we find a robust checker)

abadams: Those were all done manually by me after reading the source and checking if it the usage was scoped correctly. I just got tired and sloppy, so I'll recheck them. This idiom is already used throughout the compiler and provides a real performance benefit, so I don't want to back it out. E.g. I found a place where a pointless copy was being made of a large std::map.

abadams: One possible safer idiom: have the things accept const pointers instead, so that it's obvious at the call-site that a non-owning reference is being taken, but then immediately dereference them in the constructor to store in the const ref member variable.

abadams:
```
struct MyHelper {
  const Foo &foo;
  MyHelper(const Foo *foo) : foo(*foo) {}
};

...

Foo foo;
MyHelper helper(&foo);
```

abadams: The downside is that it's possible to pass nullptr, and the signature hints that it might be meaningful to do so.

abadams: In the absence of a checker (I couldn't find one) I think this might be a good solution. You're not going to write &(some rvalue) at the call-site. LMK if you agree and I'll make the change.

abadams: Actually belay that. This gets gross-looking in cases where you're just passing through an existing const-ref. Sigh. I'll keep thinking.

steven-johnson: Yeah, this is a place where C++ just doesn't have a good option (I presume Rust's borrow semantics would be the solution here). 
- Agreed that pointless copy of maps, large structs (etc) is bad, but for "small" things like std::string I'd err on the side of safety unless we have hard numbers showing the performance win
- Agreed that const pointers instead is a mixed blessing here

May I suggest that you decouple the other clang-tidy changes from this one (for now at least) since the others seem less controversial? There's no urgency to land this yet.

src/AutoSchedule.cpp
```diff
@@ -190,17 +191,17 @@ struct DependenceAnalysis {
...
         }
     };
     // Cache for bounds queries (bound queries with the same parameters are
     // common during the grouping process).
     map<RegionsRequiredQuery, vector<RegionsRequired>> regions_required_cache;
 
-    DependenceAnalysis(const map<string, Function> &env, const vector<string> &order,
-                       const FuncValueBounds &func_val_bounds)
-        : env(env), order(order), func_val_bounds(func_val_bounds) {
+    DependenceAnalysis(map<string, Function> env, vector<string> order,
```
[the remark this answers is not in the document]
abadams: It's a stylistic change only, to conform to the current common wisdom for C++. The idea is that if you uniformly pass these sorts of things by value as a matter of convention, then you'll occasionally avoid a copy when you pass an rvalue.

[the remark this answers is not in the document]
steven-johnson: Ugh. OK.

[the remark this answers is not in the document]
steven-johnson: To borrow a meme, "that sounds like the status quo, but with more steps"

steven-johnson:
LGTM pending green. I think the arm/win breakages are due to LLVM10.

src/AutoSchedule.cpp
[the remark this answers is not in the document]
abadams: I mean, we don't need to turn this one on - I was assuming turning on clang-tidy checkers would be a good thing and a good mostly-brainless task for a sick day, but yeah looking at the diff again I think I might just fix that one unintentional copy and move on to the next clang-tidy check.

PR #4677, opened 2020-03-04
Clang tidy use default member initializers

abadams:
Use default member initializers, and either remove or explicitly default the default constructors where possible.

Builds on #4676, so the diff will be large until that is merged.


abadams:
One additional change: I deleted two unused classes in JITModule.cpp rather than update them.

steven-johnson:
LGTM pending green buildbots

Issue #4678, opened 2020-03-05
auto_schedule_unbounded_nonpure is broken


steven-johnson:
At current top of tree, this fails with `Error: The pure definition  of Function g calls function sum in an unbounded way in dimension 0`.

steven-johnson:
strangely, https://github.com/halide/Halide/commit/c924ea43efd317b933f6428dc53cf55142bb5324 (from https://github.com/halide/Halide/pull/4566) seems to be the injection point. Assigning to @vksnk for investigation.

steven-johnson:
OK, a different commit from that same PR is being fingered now:

```
b5c8d8f52adb9a32dd79200654700990d4bafb6d is the first bad commit
commit b5c8d8f52adb9a32dd79200654700990d4bafb6d
Author: Volodymyr Kysenko <vksnk@google.com>
Date:   Thu Feb 6 11:21:03 2020 -0800

    Include more lowering stages in PrintLoopNest

 src/PrintLoopNest.cpp | 33 +++++++++++++++++++++++----------
 1 file changed, 23 insertions(+), 10 deletions(-)
```

reassigning to vksnk@

vksnk:
Thanks. It does call print_loop_nest, I'll investigate.

vksnk:
It fails during bounds inference and also fails in the same way if you do `compile_jit . Is it expected for this test?  I would guess no, but the test name has "unbounded" in it, which is exactly the error:

`The pure definition  of Function g calls function sum in an unbounded way in dimension 0`

so maybe it's intentional?

abadams:
Did print_loop_nest run bounds inference before your change?

abadams:
I'm wondering what the test is supposed to be verifying. I guess it was just a case that was expected to fail, but written in a way where it didn't?

I'd be inclined to just delete the test.

vksnk:
No, it didn't run bounds inference before. It started to fail after I added bounds inference and some couple other passes into print_loop_nest.

steven-johnson:
> I'd be inclined to just delete the test.

Aside from this outright failure, several of the `auto_schedule_xxx` tests are flaky. At some point we should decide whether this autoscheduler is still paying for itself.

vksnk:
Test fails even with the default schedule (when autoscheduler is skipped) and in general, programs seems to be invalid.

[...]

PR #4679, opened 2020-03-05
Remove test/auto_schedule/unbounded_nonpure.cpp

vksnk:
Discussion is at #4678.


vksnk:
There are some failing tests, but they don't seem to be related to this change. I will merge once remaining are done (assuming that nothing else breaks).

PR #4680, opened 2020-03-05
Add GitHub Action to do build-and-test

steven-johnson:
This is an initial attempt at using GitHubActions (GHA) as a buildbot supplement. It's a work in progress still, but I think it's useful enough at this point to consider reviewing. (More details shortly.)


abadams:
Looks like one of the performance tests (well, a tutorial) is flaky doing it this way. Should we skip performance tests for these? I don't see how we can guarantee we don't have a noisy neighbor on the test machine.

What's the strategy for reducing load on the physical bots? Just cut all the build steps that are covered by the cloud builders? We could also somehow gate the physical bots triggering on all the cloud builders passing?

steven-johnson:
> Looks like one of the performance tests (well, a tutorial) is flaky doing it this way. Should we skip performance tests for these? I don't see how we can guarantee we don't have a noisy neighbor on the test machine.

We'l probably have to. I tried having all the benchmarky tests run serially to reduce this but it's still flaky. (though to be fair we get intermittent flakes of this sort even on our bespoke buildbot machines.)
 
> What's the strategy for reducing load on the physical bots? Just cut all the build steps that are covered by the cloud builders? We could also somehow gate the physical bots triggering on all the cloud builders passing?

Yep and yep.

abadams:
Looks like there are 30 things in the testing matrix, and then three copies of each thing? 

Also looks like the arm builders are taking a full hour to cross-compile libHalide, which seems surprising.

steven-johnson:
So, more details:

- This can't fully replace our physical machine setup anytime soon, but the goal is to be able to remove the load from them where possible. 

- This relies on the nightly LLVM prebuilts we've been doing for a while, so tests for LLVM trunk can be up to ~24 hours out of date, so LLVM breakages might take longer to notice.

- Things that aren't possible and/or practical include:
    - benchmarking
    - GPU testing 
    - any nontrivial testing on non-x86 targets (e.g., running the full test suite for aarch64 via QEMU takes ~5 hours)

- Note that there is a way to specify "self-hosted runners" for tests (ie, point at our existing armbots, or AWS Graviton instance, etc); I haven't investigated this yet. (If we used our physical armbots rather than some VM instance, we'd have to be *very* careful about what we allowed to run, to avoid possible compromise of those machines.)

- OTOH, the fact that we can spawn a bunch of VMs in parallel means we can add stuff to the testing matrix without adding much latency (though this depends on how many simultaneous VM instances we are allowed per-PR and per-project; the latter limit isn't clear to me yet)
   
- Currently this PR takes the approach of starting from the "bare" VMs that GHA provides and explicitly installing stuff that's needed every time; this adds a couple minutes or so to the test latency that could probably be reduced by (eg) providing our own Docker images.

- One liability here is that there isn't any logic to try to avoid redundant tests; as currently written, this will fire off a pull build-n-test suite for every new commit pushed to a PR, rather than waiting 5 minutes to allow for "oops" fixes to get collected. (I'm sure this could be made to work, but GHA doesn't obviously provide any built-in support for such things, so some ingenuity may be needed.)

- The build-and-test matrix in place right now may be larger than we need. Easy enough to trim down later.

- GHA is occasionally flaky and some VMs just fail (eg when installing software), but there isn't a way to restart just one of the jobs (it's currently all-or-none)... so if this happens, you'll have to rebuild the whole thing. (This often may not be any slower than just rebuilding one, but it is a waste of resources.)

- In the long run, an ideal approach would likely involve using this for build+CPU-correctness testing, with performance and GPU testing done on our dedicated machines. (We could likely ship the build to those machines after success, so they don't need to spend time rebuilding.)

- Note that all of the steps are done using bash -- even on Windows -- but this doesn't use MinGW; apparently Git-for-Windows provides a bash shell that is good enough for most purposes. (It might well be desirable to refactor the contents of this action into standalone bash scripts to allow for easier replication of tests on an ad hoc basis.)

Summary: My goal for this PR is to consider landing it and running it *in addition* to the existing buildbots for a while, so see how the reliability compares; if it seems generally positive, we can then consider removing redundant work from the existing buildbots; if not, we can revise or abandon it.

steven-johnson:
> Looks like there are 30 things in the testing matrix, and then three copies of each thing?

Yeah, something is wonky with the replication-by-three. Stand by.

> Also looks like the arm builders are taking a full hour to cross-compile libHalide, which seems surprising.

Yeah, I haven't tried to figure that part out yet. *Some* of it is likely to to the fact that a couple of helper apps (binary2cpp) are build for arm and emulated, but there's no way that should add 20 minutes (linux x86 is ~24 mins, arm 64 is ~45 mins).

steven-johnson:
> > Looks like there are 30 things in the testing matrix, and then three copies of each thing?

I think the three-copies thing is a side-effect of me force-pushing a rebase. I'll verify but I think that's a red herring.

PR #4682, opened 2020-03-05
Allow overriding `-march=native`

steven-johnson:
Some crosscompilers don't understand this; allow us to selectively override with (e.g.) GCC_ARCH_FOR_TESTS=armv7-a


abadams:
Why GCC in the name - this isn't a gcc-specific flag, right?

steven-johnson:
> Why GCC in the name - this isn't a gcc-specific flag, right?

Because I'm tired and cranky. Will fix.

PR #4665, opened 2020-03-02
Drop support for LLVM8 (Issue #4560)

steven-johnson:
Drop LLVM8 support (in anticipation of LLVM10 being tagged later this week).

PR #4683, opened 2020-03-06
Disable CudaVectorize atomics test to unbreak windows build

abadams:
Cherry-picked from #4665 to unbreak the bots


alexreinking:
Since #4628 landed, should this be reverted? This is what we currently have in master:

https://github.com/halide/Halide/blob/26ebef3f0480eb65d1b79fefdac48d5682ec6021/test/correctness/atomics.cpp#L1249-L1255

PR #4685, opened 2020-03-06
Remove pointless std::moves, or make them not pointless

abadams:
Given the last clang-tidy PR added std::moves everywhere and was met
with a meh, how about a PR that removes pointless std::moves?

These std::move were all from const args or being passed to things that
expect a const ref arg, so they just made a copy anyway and the
std::move gave a false sense that no copy occurred.

This PR either removes the std::move, or fixes the reason that a copy
was occurring. The reasons were:

- A missing move assignment operator for Buffer<>
- some of the things in IROperator taking const-refs where most didn't,
but ~all definitions assumed things were passed by value and used
std::move internally. Just made these uniformly accept Exprs by value.
- LetStmt::make takes a string by const ref
- The simplifier cleverly passing rvalues to mutate, even though mutate
takes a const ref
- We were std::moving an int in IR.cpp

Also made our move constructors / assignment operators noexcept, which
they must be if you want STL containers to move them instead of making
copies, e.g. when a vector resizes.

Enforce both things with a clang-tidy checker. We don't want to
introduce std::moves that do nothing in future, or accidentally leave
off noexcept from our move constructors.

### DIFF
--- a/src/Buffer.h
+++ b/src/Buffer.h
@@ -128,6 +128,9 @@ public:
     /** Trivial copy assignment operator. */
     Buffer &operator=(const Buffer &that) = default;
 
+    /** Trivial move assignment operator. */
+    Buffer &operator=(Buffer &&) noexcept = default;
+
     /** Make a Buffer from a Buffer of a different type */
     template<typename T2>
     Buffer(const Buffer<T2> &other)
@@ -137,7 +140,7 @@ public:
 
     /** Move construct from a Buffer of a different type */
     template<typename T2>
-    Buffer(Buffer<T2> &&other) {
+    Buffer(Buffer<T2> &&other) noexcept {
         assert_can_convert_from(other);
         contents = std::move(other.contents);
     }
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -20,7 +20,7 @@ namespace Halide {
 // terms, which is the degree plus one.
 namespace {
 
-Expr evaluate_polynomial(const Expr &x, float *coeff, int n) {
+Expr evaluate_polynomial(Expr x, float *coeff, int n) {
     internal_assert(n >= 2);
 
     Expr x2 = x * x;
@@ -1201,7 +1201,7 @@ Expr operator*(Expr a, Expr b) {
     return Internal::Mul::make(std::move(a), std::move(b));
 }
 
-Expr operator*(const Expr &a, int b) {
+Expr operator*(Expr a, int b) {
     user_assert(a.defined()) << "operator* of undefined Expr\n";
     Type t = a.type();
     Internal::check_representable(t, b);
@@ -1262,7 +1262,7 @@ Expr operator%(Expr a, int b) {
     return Internal::Mod::make(std::move(a), Internal::make_const(t, b));
 }
 
-Expr operator%(int a, const Expr &b) {
+Expr operator%(int a, Expr b) {
     user_assert(b.defined()) << "operator% of undefined Expr\n";
     Type t = b.type();
     Internal::check_representable(t, a);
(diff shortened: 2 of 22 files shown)


src/Buffer.h
```diff
@@ -128,6 +128,9 @@ class Buffer {
     /** Trivial copy assignment operator. */
     Buffer &operator=(const Buffer &that) = default;
 
+    /** Trivial move assignment operator. */
```
zvookin: What does this do with the source object? (Yes, I should figure it out myself. If you are sure this doesn't drop the reference in destructing the temporary being assigned from, then all good.)

abadams: A Buffer<> is just an IntrusivePtr, which is safe to move.

src/IROperator.cpp
```diff
@@ -20,7 +20,7 @@ namespace Halide {
 // terms, which is the degree plus one.
 namespace {
 
-Expr evaluate_polynomial(const Expr &x, float *coeff, int n) {
+Expr evaluate_polynomial(Expr x, float *coeff, int n) {
```
steven-johnson: I'm genuinely puzzled as to how this is an improvement. Doesn't this require making a copy to pass in? (Even though it's a cheap-ish refcount adjustment, that still seems more costly than a pass-by-ref.) Not trying to be argumentative, trying to understand.

abadams: This function had to change one way or the other, because the std::move at the end of the body of it was a lie. The options were to remove the std::move, or change the signature so that the std::move actually did something useful. I changed the signature because there was one case where an rvalue was passed, and I figured we avoid a refcount bump in that case, and it's a wash in the lvalue case because we get charged a refcount bump on the call, but save one in the function body due to the std::move.

That's supposed to be the new general rule - pass by value and use std::move inside the function, but I find this hard to think about, so I'll just check the disassembly to see what happens...

Yeah, the body of evaluate_polynomial has two fewer refcount adjustments (on different if paths, so only one triggers), and the call-sites in the lvalue cases have an additional ref count bump, so it's a wash. The call-site in the rvalue case is unchanged, because the temporary x*x just becomes the argument directly without having to make a copy of it, so it's a savings in that case.

There's a little more complexity to it though, involving destructors:

In the pass-by-value case, the extra refcount bump at the lvalue call-sites is due to creating a temporary for the value argument, which then needs to be destroyed after the call (even though it has been moved from inside evaluate_polynomial and is actually just a default-constructed Expr() at that point), so there's also conditional refcount subtract and call to delete if it's zero. The condition is always false, so there's no actual atomic subtract or call to delete, but still, there's an extra branch, and a little extra code size, and the optimizer doesn't seem to realize it's dead code (to do that, it would have to look inside evaluate_polynomial and realize that the arg is unconditionally moved from, and that means the pointer is now null). In the pass-by-const-ref case the extra refcount bump is just because the returned value of evaluate_polynomial captures the argument one additional time, so it's just an add, there's no extra branch.

So the value case does fewer copies total, by moving some out of evaluate_polynomial to the call site instead, but the const-ref case does fewer destructions of moved-from-Exprs(), so it saves some branches and some code size. For types where a copy is expensive and destruction of a moved-from instance is cheap, accepting args by value is a big win if you ever pass an rvalue. Is Expr one of these types? Well, an atomic add is probably much more expensive that a perfectly-predicted branch, so I think it's a good trade-off.

I did actually experiment a while ago with making all of IR.h and IROperator.h take Exprs by const ref instead of by value and then not internally std::move, and it did indeed slow down the compiler due to the extra ref count bumps. Passing Exprs by value is better (slightly) when it's possible they could be rvalues, as they often are when writing expressions out.

abadams: Shorter version: 

- A copy is inevitable because the Expr gets captured. 
- Passing by value instead of const-ref does one more copy at the call-site to create the Expr arg and one fewer copy inside the function (because now you can use std::move).
- It's better to have the copy at the call-site, because it can be optimized away in the case where the arg is a temporary.
- It does, however, cost additional code size at the call site.

steven-johnson: Thanks for the detailed investigation. LGTM.

abadams: That was also the reasoning behind all the std::moves in the previous PR that changed Constructors to accept things by value. That was less worthwhile than having IROperator accept Exprs by value though, because those were almost never rvalues. It's very common for the functions in IROperator.

PR #4686, opened 2020-03-07
Don't create a fresh Expr from a Var on every use

abadams:
A Var silently converts to an Expr in most contexts. In master, this
creates a fresh IR node each time pointlessly, which then makes things
like expression matching more expensive, because you have to actually
compare the strings rather than notice the pointers are identical.

This PR makes a Var construct the expression eagerly on construction,
and just deliver references to it on demand. Because the expression
stores the name in its Variable node, we no longer need to store the
name in the Var too. This makes Var (and VarOrRVar) a little smaller.

Had to clean up some code in the autoscheduler which was making Vars
with empty names. This would have thrown an error if they were ever used
as Exprs (they were sentinels of a sort). Now that Expr construction is
eager Vars with empty names can't be constructed.

This was a minor inefficiency a noticed a while ago that I've been
wanting to clean up.


steven-johnson:
Ready to land? Only failure looks unrelated

PR #4687, opened 2020-03-07
Avoid pointless copies

abadams:
Functions that don't do anything non-const to an argument should take it
by const ref, not by value. We're currently extremely inconsistent on
this front. We used to take Exprs and Stmts by value, then switched
largely to const refs in cases where we weren't mutating or capturing
them. But there's a ton of old code still taking them by value.

Functions that take a value argument and do non-const things to it (e.g.
copy it) should std::move it on the last use.

There were also a few places where we were just making a copy to give a
local variable a different name. A const ref suffices.

Each of these changes avoids a copy, so it should be a strict
performance win in every case.

Total time to run "make test_correctness -j32" with this change:

real	1m7.819s
user	16m40.727s
sys	0m18.393s

real	1m8.215s
user	16m45.404s
sys	0m21.906s

real	1m7.138s
user	16m41.605s
sys	0m19.413s

and in master:

real	1m7.568s
user	16m48.871s
sys	0m21.086s

real	1m8.193s
user	16m44.842s
sys	0m19.516s

real	1m8.056s
user	16m43.921s
sys	0m19.823s

libHalide.so in this branch: 107891072
libHalide.so in master: 113506576

So this saves 0.3% total compile time and (ignoring the part of
libHalide that's just LLVM), 0.2% code size.


steven-johnson:
The python bindings are going to need fixing here.

Issue #4638, opened 2020-02-24
Simplifier failure (fuzz test)


dsharletg:
A random PR presubmit failed with this failure: https://buildbot.halide-lang.org/master/#/builders/21/builds/79

```
FAILED TEST: fuzz_simplify
Simplify fuzz test seed: 1581536154
a = int32((int0)b)
b = int32((int0)b)
c = int32((int0)b)
d = int32((int0)c)
e = int32((int0)b)
((-61/81)/(int32((int0)e) % (max(int32((int0)d), int32((int0)e)) % (int32((int0)e) % int32((int0)b)))))
-1
In vector lane 0:
((-61/81)/(int32((int0)e) % (max(int32((int0)d), int32((int0)e)) % (int32((int0)e) % int32((int0)b))))) -> 0
-1 -> -1
../halide/Makefile:1893: recipe for target 'quiet_correctness_fuzz_simplify' failed
```

dsharletg:
Something weird is going on here... I don't see how this test possibly generates an `int0` type...

PR #4689, opened 2020-03-08
delete rewrite rule that is incorrect under new div by zero definition

jn80842:
After the change to the semantics of division, this rewrite rule is incorrect when y = 0. Deleting this rule should fix issue #4638.


dsharletg:
It looks like the failures in the travis check are real simplification failures:

Input: (-1/((x*2) + 1))
Output: (-1/((x*2) + 1))
Expected output: select((x < 0), 1, -1)

abadams:
The travis check hints at a precondition that could save this rule: If ExprInfo struct tells us y could not possibly be zero, the rule is fine. This is already captured in the bool "denominator_non_zero", we can probably just add that to the precondition.

steven-johnson:
We should add the breakage from #4638 to a test somewhere, to ensure it stays fxed.

steven-johnson:
Clone is clean except for the known arm32 failures. LGTM.

PR #4694, opened 2020-03-10
Prefer emplace_back(args) to push_back(Constructor(args))

abadams:
You can avoid a copy this way some of the time, so it's the preferred style according to the conventional wisdom right now.

This also replaced:
```push_back({args}) ```
with 
```emplace_back(args)```

and 
```Vector<T>::push_back(something_that_can_convert_to_T)```
with
```Vector<T>::emplace_back(something_that_can_convert_to_T)```

It looks a little odd to say:
```
vector<Expr> exprs;
exprs.emplace_back(0);
exprs.push_back(some_expr);
```
instead of 
```
vector<Expr> exprs;
exprs.push_back(0);
exprs.push_back(some_expr);
```
But it is indeed a good idea - an int can be passed by value. A temporary Expr, even an rvalue, must be placed on the stack and passed by reference. push_back(0) has to destroy this temporary (even though it has been moved from in the common case), so there's extra code. This hurts even in the case of -O3 when push_back and emplace_back both get inlined! The issue is that in the slow path of adding something to a vector you need to reallocate memory, and the reallocation might throw, so you have a temporary Exprs that genuinely needs a cleanup. In the emplace_back case you call the constructor *after* you realloc, and do a placement new, so if realloc throws, so there's no cleanup required.

In summary, a function that just does push_back(0) onto a vector of Expr has 58 lines of assembly and a function that does emplace_back(0) has 29 lines of assembly.

The advantage of just accepting all of these changes is that it can be mechanically checked. But other opinions welcome. E.g. we could take a subset of these changes and not mechanically check them

I also turned on another performance check - don't hide trivial destructors in cpp files, or they can't be optimized away. But we never did that, so it was a no-op.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,5 @@
 ---
-Checks: '-*,modernize-use-override,performance-move-const-arg,performance-noexcept-move-constructor,performance-unnecessary-value-param,performance-unnecessary-copy-initialization'
+Checks: '-*,modernize-use-override,performance-move-const-arg,modernize-use-emplace,performance-noexcept-move-constructor,performance-unnecessary-value-param,performance-unnecessary-copy-initialization,performance-trivially-destructible'
 WarningsAsErrors: '*'
 HeaderFilterRegex: '.*'
 AnalyzeTemporaryDtors: false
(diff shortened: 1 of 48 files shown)


steven-johnson:
LGTM

PR #4695, opened 2020-03-10
Remove const that messes up move constructor

abadams:
This const prevents the default move constructor from moving the member, because you can't move from something const. This in turns makes the noexcept a lie whenever T is something with a constructor that might throw an exception (e.g. std::string, which might allocate).


steven-johnson:
LGTM, but: is this move ctor ever used? I wonder if it could just be `= delete` instead.

abadams:
I believe it is because we put Scoped things in Frame objects in let visitors, and the vector of Frames might need to resize and move things around.

Issue #4691, opened 2020-03-09
compute_with(LoopAlignStrategy::AlignStart) throws an error


vksnk:
Following code will throw an assert:

 ```
   const int f_size = 1280;
    const int g_size = 2560;
    Buffer<float> f_im(f_size, f_size, 5), g_im(g_size, g_size);

    Var x("x"), y("y"), c("c"), xi("xi"), yi("yi"), yii("yii"), yo("yo");
    Func f("f"), g("g"), h("h"),  input("input");


    input(x, y) = x;//cos(sin(cos(sin(Halide::cast<float>(x + y)))));
    f(x, y, c) =  c * input(x, y);
    h(x, y, c) = f(x, y, c);
    
    Func inl("inl");
    inl(x, y) = f(x/2, y/2, 0);
    inl(x, y) += f(x/2, y/2, 2);
    g(x, y) = inl(x, y);

    g
      .split(y, yo, y, 32 * 2, TailStrategy::RoundUp)
      .split(y, y, yi, 2, TailStrategy::RoundUp)
      .vectorize(x, 4, TailStrategy::GuardWithIf)
      .compute_with(h, y, LoopAlignStrategy::AlignStart);

    h
      .reorder(x, c, y)
      .split(y, yo, y, 32, TailStrategy::RoundUp)
      .vectorize(x, 4, TailStrategy::GuardWithIf)
      .compute_root();

    g.bound(y, 0, g_size);
    h.bound(y, 0, f_size).bound(c, 0, 5);
    
    Pipeline p({h, g});
    p.compile_jit();
```

error:

`Internal Error at /Users/vksnk/Work/Halide/src/ScheduleFunctions.cpp:1221 triggered by user code at : Condition failed: parent_min != bounds.end():`

It will work if you change align strategy of compute_with to LoopAlignStrategy::NoAlign.

**Comment by vksnk** [comment]
It seems that the cause of this is that one of the functions has one extra split and var names of the parent function (in the context of compute_with) won't match when exact string look up is done in compute_shift_factor. For example, in this case, fused vars are:

```
g y.y.y
g y.yo
h y.y
h  y.yo

```
(notice extra y here). 

I have a change which fixes it, but it breaks one other test, so will do some extra debugging.

vksnk:
Has been fixed by #4691.

PR #4696, opened 2020-03-10
Consider variable names of fused function when looking for the loop min/max in compute_shift_factor

vksnk:
This fixes the issue from #4691. As described there, fused var names may not match var names in the given function, so search in the list fails. This PR adds an extra check for the case, when we can't find the name and tries to search under different name of fused var.


src/ScheduleFunctions.cpp
```diff
@@ -1193,10 +1193,17 @@ class InjectFunctionRealization : public IRMutator {
             internal_assert(iter != dims.end());
             start_fuse = (int)(iter - dims.begin());
         }
+
+        const auto &env_iter = env.find(fuse_level.func());
+        internal_assert(env_iter != env.end());
+        const auto &parent_func = env_iter->second;
+
+        const vector<Dim> &parent_dims = (fuse_level.stage_index() == 0) ? parent_func.definition().schedule().dims() : parent_func.update(fuse_level.stage_index() - 1).schedule().dims();
```
steven-johnson: style nit: long line should be broken, eg

```
const vector<Dim> &parent_dims = (fuse_level.stage_index() == 0) 
    ? parent_func.definition().schedule().dims() 
    : parent_func.update(fuse_level.stage_index() - 1).schedule().dims();
```

src/ScheduleFunctions.cpp
```diff
@@ -1193,10 +1193,17 @@ class InjectFunctionRealization : public IRMutator {
...
+        int fused_vars_num = dims.size() - start_fuse - 1;
```
steven-johnson: Does the code above guarantee this won't underflow?

src/ScheduleFunctions.cpp
```diff
@@ -1211,17 +1218,27 @@ class InjectFunctionRealization : public IRMutator {
             }
 
             string parent_prefix = fuse_level.func() + ".s" + std::to_string(fuse_level.stage_index()) + ".";
+            int parent_var_index = (i - start_fuse) + (int)parent_dims.size() - 1 - fused_vars_num;
```
steven-johnson: Ditto underflow question. Is it overkill to add `internal_assert(parent_var_index >=0)`?

src/ScheduleFunctions.cpp
vksnk: Done, I was hoping that 'make format' would fix it, but apparently not.

src/ScheduleFunctions.cpp
vksnk: There is a check above to make sure that start_fuse is in range, however, I moved this line up to make it more visible.

src/ScheduleFunctions.cpp
vksnk: Good point, I've added assert.

src/ScheduleFunctions.cpp
vksnk: Ugh, now clang-format pre-check complains, is there a way to suppress it?

abadams: I would trying adding parenthesis around the entire expression and then just accept whatever 'make format' decides to do with it.

vksnk: Thanks, 'make format' would just make it one line no matter what I do. I've split it into two lines, one of them is still pretty long, but maybe closer to acceptable.

vksnk:
I've addressed the comments, please take a look.

steven-johnson:
LGTM but I don't know this part of the code well enough to be the best reviewer.

steven-johnson:
ping to @abadams for review

abadams:
I'm not sure I understand what's going on here or why there are multiple possible names. I think the parent func is the one it's fused with? Could you add more comments that explain it?

vksnk:
This is based on the test case from this PR, but say you have two functions (A and B) fused together (`A.compute_with(B, y)`) and A is split like this:

```
.split(y, yo, y, 32 * 2, TailStrategy::RoundUp)
.split(y, y, yi, 2, TailStrategy::RoundUp)
```
while B like this:

`.split(y, yo, y, 32, TailStrategy::RoundUp)`

if you print the names of dimensions:

for A:

```
x.v0
x.x
y.y.yi
y.y.y <- slightly different name
y.yo
```

for B:

```
x.v1
x.x
c
y.y <- slightly different name
y.yo
```

You still can fuse them, because check that fuse dimensions match is done by comparing them [both ways with ends_with](https://github.com/halide/Halide/blob/master/src/ScheduleFunctions.cpp#L54), so it's sort of symmetrical. However, this specific function I've modified in this PR is doing exact string match/look-up of parent's loop var using the name of child's var, so in case like this, it fails to find anything and asserts.

abadams:
So shouldn't it always use the parent var name then, if it's looking for the parent's loop?

vksnk:
> So shouldn't it always use the parent var name then, if it's looking for the parent's loop?

Good point, this should be enough. Let me fix it.

vksnk:
> > So shouldn't it always use the parent var name then, if it's looking for the parent's loop?
> 
> Good point, this should be enough. Let me fix it.

Done, please take a look.

PR #4697, opened 2020-03-10
Support optional ARCH_FOR_TEST in CMake too

steven-johnson:
Was recently added to Makefile

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -18,7 +18,11 @@ else ()
     # Also allow tests, via conditional compilation, to use the entire
     # capability of the CPU being compiled on via -march=native. This
     # presumes tests are run on the same machine they are compiled on.
-    target_compile_options(halide_test INTERFACE "$<$<CONFIG:RELEASE>:-O0>" "-march=native")
+    set(ARCH_FOR_TESTS "$ENV{ARCH_FOR_TESTS}")
+    if ("${ARCH_FOR_TESTS}" STREQUAL "")
+      set(ARCH_FOR_TESTS "native")
+    endif()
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -18,7 +18,11 @@ else ()
     # Also allow tests, via conditional compilation, to use the entire
     # capability of the CPU being compiled on via -march=native. This
     # presumes tests are run on the same machine they are compiled on.
-    target_compile_options(halide_test INTERFACE "$<$<CONFIG:RELEASE>:-O0>" "-march=native")
+    set(ARCH_FOR_TESTS "$ENV{ARCH_FOR_TESTS}")
+    if ("${ARCH_FOR_TESTS}" STREQUAL "")
+      set(ARCH_FOR_TESTS "native")
+    endif()
+    target_compile_options(halide_test INTERFACE "$<$<CONFIG:RELEASE>:-O0>" "-march=${ARCH_FOR_TESTS}")
```
alexreinking: -march=native isn't supported by MSVC. Use `$<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-march=${ARCH_FOR_TESTS}>`

alexreinking: Would also be good to issue a warning that this setting is GNU-compatible only when used with MSVC.

steven-johnson: This whole thing is in an if(MSVC)...else()... clause, so, no, it won't :-)

(But I'll clean it up anyway)

steven-johnson:
PTAL

test/CMakeLists.txt
alexreinking: > This whole thing is in an if(MSVC)...else()... clause, so, no, it won't :-)

Ugh, misleading diff viewer. Sorry.

Issue #4690, opened 2020-03-09
arm32 breakage in LLVM10


steven-johnson:
correctness_simd_op_check is failing for arm32 builds under the LLVM10 branch ~(but not LLVM9 or trunk)~ and also LLVM11 (aka trunk); specifically, we segfault once trying to test the `vqadd.*` instructions.

The injection point is:

```
commit 9f15fcc2718f95f1dac9e6e57aa93d84e9709930
Author: David Green <david.green@arm.com>
Date:   Wed Nov 27 11:01:27 2019 +0000

    [ARM] Replace arm_neon_vqadds with sadd_sat

    This replaces the A32 NEON vqadds, vqaddu, vqsubs and vqsubu intrinsics
    with the target independent sadd_sat, uadd_sat, ssub_sat and usub_sat.
    This helps generate vqadds from standard IR nodes, which might be
    produced from the vectoriser. The old variants are removed in the
    process.

    Differential Revision: https://reviews.llvm.org/D69350
```

steven-johnson:
(Yes, this was back in November 2019. Clearly our armbots weren't testing trunk properly.)

steven-johnson:
cc @alinas

dsharletg:
Meta question here: should we stop supporting arm32? Every ARM target I'm aware of is now 64-bit. We saw significant performance advantages of arm64 over arm32, so the only reason to continue supporting it is if there are users on 32-bit only systems.

PR #4699, opened 2020-03-11
Fix LLVM10/11 breakage on arm32 (Issue #4690)

steven-johnson:
Some arm-specific intrinsics got removed (in favor of portable LLVM versions) and we didn't notice because the armbots never tested on trunk.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -115,10 +115,10 @@ CodeGen_ARM::CodeGen_ARM(Target target)
 #if LLVM_VERSION >= 100
         if (t.is_int()) {
             p.intrin32 = "llvm.sadd.sat" + t_str;
-            p.intrin64 = "llvm.aarch64.neon.sqadd" + t_str;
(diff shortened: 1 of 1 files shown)


src/CodeGen_ARM.cpp
```diff
@@ -112,13 +112,23 @@ CodeGen_ARM::CodeGen_ARM(Target target)
         casts.push_back(p);
 
         // Saturating add
+#if LLVM_VERSION >= 100
+        if (t.is_int()) {
+            p.intrin32 = "llvm.sadd.sat" + t_str;
+            p.intrin64 = "llvm.aarch64.neon.sqadd" + t_str;
```
dsharletg: The standard LLVM intrinsic is probably also supported on arm64, and seems better to use if it's available. It's probably only a matter of time before the aarch64 intrinsic is removed and we'll get another breakage.

steven-johnson: Done, let's try it out and see how it goes

PR #4700, opened 2020-03-11
Fix deps and COMMAND for binary2cpp and build_halide_h

steven-johnson:
Some of the targets that relied on these didn't list them in their DEPENDS (so build order variance could break).

All of them should have used $<TARGET_FILE> for the COMMAND setup; the fact that it worked currently was basically luck.

(Note: the need to $<TARGET_FILE> will go away once the `cmake-modern` PR lands -- the newer CMake version used there changes the behavior -- but until then, this is needed to make some cross-compile targets work.)


alexreinking:
lgtm even though, as you noted, I'll be blowing over these changes soon.

PR #4711, opened 2020-03-13
Add cmake to apps/gradient_autoscheduler

seanprime7:
`apps/gradient_autoscheduler` currently has only Makefile.

This change allows the project-wide cmake and ctest to build apps/gradient_autoscheduler and run tests.


BachiLi:
LGTM but Alex should take a better look.

vinodgro:
I was able to build the gradient_autoscheduler on Windows with this change.

abadams:
Steven is the other reasonable reviewer for cmake stuff. I'll add him in case Alex is busy.

steven-johnson:
FYI, building this PR on OSX and running `ctest -R run_gradient_autoscheduler_test_cpp --output-on-failure` fails with `Error: Failed to load: libgradient_autoscheduler.so: dlopen(libgradient_autoscheduler.so, 1): image not found`

apps/gradient_autoscheduler/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+add_library(gradient_autoscheduler
+            SHARED
+            GradientAutoscheduler.cpp
+            ASLog.cpp)
+target_link_libraries(gradient_autoscheduler PRIVATE "${HALIDE_COMPILER_LIB}")
+target_include_directories(gradient_autoscheduler PRIVATE "${HALIDE_INCLUDE_DIR}" "${HALIDE_TOOLS_DIR}")
+install(TARGETS gradient_autoscheduler LIBRARY DESTINATION bin)
```
steven-johnson: You need to add these lines before `install(...)`:

```
if(NOT MSVC)
  set_target_properties(gradient_autoscheduler PROPERTIES LINK_FLAGS "-rdynamic")
elseif(APPLE)
  set_target_properties(gradient_autoscheduler PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
endif()

if(NOT HALIDE_ENABLE_RTTI)
  if(NOT MSVC)
    target_compile_options(gradient_autoscheduler PRIVATE "-fno-rtti")
  else()
    target_compile_options(gradient_autoscheduler PRIVATE "/GR-")
  endif()
endif()
```

seanprime7: Okay.  I will add this to the change.

seanprime7:
> 
> 
> FYI, building this PR on OSX and running `ctest -R run_gradient_autoscheduler_test_cpp --output-on-failure` fails with `Error: Failed to load: libgradient_autoscheduler.so: dlopen(libgradient_autoscheduler.so, 1): image not found`

I have tested it on Linux and Windows, but not on macos.  Let me try to reproduce it on macos and fix it.

steven-johnson:
Please see https://github.com/halide/Halide/pull/4716 for my version with the changes needed for OSX -- feel free to add them here (or alternately we can move to that one)

seanprime7:
@steven-johnson I have cherry-picked your commit into this branch.

I have one question though about the following snippet:

```
if(NOT MSVC)
  set_target_properties(gradient_autoscheduler PROPERTIES LINK_FLAGS "-rdynamic")
elseif(APPLE)
  set_target_properties(gradient_autoscheduler PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
endif()
```

Shouldn't we change this to the following?  On `APPLE`, `if(NOT MSVC)` will be `true` and `-rdynamic` would be added and the control will not go into `elseif(APPLE)` due to `else`, will it?

```
if(APPLE)
  set_target_properties(gradient_autoscheduler PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
elseif(NOT MSVC)
  set_target_properties(gradient_autoscheduler PROPERTIES LINK_FLAGS "-rdynamic")
endif()
```

apps/gradient_autoscheduler/CMakeLists.txt
seanprime7: Cherry-picked your e28ccd87e57c174a1c9afd9413e5d816400f84a3 with said change.

steven-johnson:
> I have one question though about the following snippet:

Heh, yeah, I suspect you are right -- this is code that I took from apps/autoscheduler, so maybe it's wrong there too. 

That said, probably better to leave them the same since both will be obviated soon by the cmake-modern branch.

steven-johnson:
(still waiting on the clone-for-testing to complete)

alexreinking:
This looks good to me for now. I'll have to update it to work on `cmake-modern`, but this still saves me some work. Thanks!

apps/autoscheduler/CMakeLists.txt
```diff
@@ -41,8 +41,10 @@ target_link_libraries(retrain_cost_model_process
 # =======================================================
 
 # libauto_schedule
+# Note: must use MODULE here (not SHARED) to get .so (instead of .dylib) on OSX.
+# This means that this can only be opened dynamically (not linked directly), but that's ok.
 add_library(auto_schedule
-            SHARED
+            MODULE
```
alexreinking: Good catch.

PR #4713, opened 2020-03-14
Move first simplification before allocation bounds inference

abadams:
This PR improves a few related things in the compiler. The main change is that I removed the dependence of allocation bounds inference and storage folding (and boxes_touched) on magic names, which means we can run full simplification earlier. This should generally make allocation bounds tighter, and also speeds up the compiler by giving allocation bounds inference less IR to chew through (by a full 10% on local laplacian lowering, and a similar amount on a 32x32 fft, 5% on lens blur).

The full story: 

I picked at a thread and a bunch of stuff unraveled.  I wanted to remove a superfluous-looking max in ApplySplit. This broke something in allocation bounds inference related that would have been fine if we could just run full simplification before we did allocation bounds inference. But we can't, because allocation bounds inference relies on magic names from bounds inference (foo.y.min/max) these magic names shadow each other, and haven't been unique'd yet, so we can't run the simplifier, which assumes two expressions that are syntactically identical have the same value, so shadowing breaks it.

But why does allocation bounds inference need these magic names? The only reason it needs them is to ensure it knows that an extern stage is going to write to the whole region required. We can just tell it that earlier instead with a marker in the IR that says "you should treat this marker as a write to this buffer over this region". Presto, no magic name.

However the other way in which allocation bounds inference (well, boxes_touched) relies on magic names was in its handling of getting the min/max out of a bounds query result. It just lifted them out of their containing scope (see the deleted code in Bounds.cpp below), assuming a bounds query buffer *of the same name* also exists in the containing scope, and also assuming that the result of that outer bounds query was a valid interval containing the result of the inner bounds query. Yuck. Instead I removed that special handling from boxes_touched, and instead explicitly clamped the inner bounds query result to be within the outer region required, which comes from the outer bounds query. That gives boxes_touched a way to bound the impure call with something. 

So now if we have a misbehaving extern stage, which lies in its outer bounds query and says it's going to ask for less than it does in the inner bounds query, it simply doesn't get what it asks for in the inner bounds query. It gets the intersection of the outer bounds query result and the inner bounds query result. Too bad. This behavior is illustrated in the new test. Formerly I'm not sure what happened, because we were just assuming that it couldn't occur. I guess that's UB. One of the tests in storage_folding actually broke this rule, but it didn't matter before because it never actually accessed the input.

I'm hoping that moving simplification earlier means we can unstick the PR that moves add_image_checks earlier too, because we'll be able to run a full simplification before it happens.

This PR needs testing inside Google, where there are lots of complex pipelines with extern stages. The interaction with storage folding of extern stages is the part I'm most worried about.

### DIFF
--- a/src/AllocationBoundsInference.cpp
+++ b/src/AllocationBoundsInference.cpp
@@ -31,19 +31,6 @@ class AllocationInference : public IRMutator {
 
         Scope<Interval> empty_scope;
         Box b = box_touched(op->body, op->name, empty_scope, func_bounds);
-        if (touched_by_extern.count(f.name())) {
-            // The region touched is at least the region required at this
-            // loop level of the first stage (this is important for inputs
-            // and outputs to extern stages).
-            Box required(op->bounds.size());
-            for (size_t i = 0; i < required.size(); i++) {
-                string prefix = op->name + ".s0." + f_args[i];
-                required[i] = Interval(Variable::make(Int(32), prefix + ".min"),
-                                       Variable::make(Int(32), prefix + ".max"));
-            }
-
-            merge_boxes(b, required);
-        }
 
         Stmt new_body = mutate(op->body);
         Stmt stmt = Realize::make(op->name, op->types, op->memory_type, op->bounds, op->condition, new_body);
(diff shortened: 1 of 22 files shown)


steven-johnson:
(1) I am ecstatic that we are lessening (somewhat) our reliance on magic names.
(2) I see various breakages on the buildbots -- I am going to hold off on the Google test until they are clean (or you tell me to go ahead anyway)

abadams:
I'm going to break this into some smaller PRs.

steven-johnson:
Besides the build failures, now there are merge conflicts -- I will definitely hold off on Google tests until I hear from you

abadams:
Yes, we want the smaller PRs in first, then I'll probably rebase this.

abadams:
Simplified and rebased in #4766

PR #4718, opened 2020-03-16
Bounds on integer division could be a little tighter

abadams:
Even when nothing is known about the divisor, integer division can't
produce a result larger in magnitude than the numerator.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -552,7 +552,16 @@ private:
             int min_sign = static_sign(b.min);
             int max_sign = static_sign(b.max);
             if (min_sign != max_sign || min_sign == 0 || max_sign == 0) {
-                interval = Interval::everything();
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -552,7 +552,16 @@ class Bounds : public IRVisitor {
             int min_sign = static_sign(b.min);
             int max_sign = static_sign(b.max);
             if (min_sign != max_sign || min_sign == 0 || max_sign == 0) {
-                interval = Interval::everything();
+                if (op->type.is_float()) {
+                    interval = Interval::everything();
+                } else {
```
shoaibkamil: Does this need to ensure we're dealing with no_overflow types?

You may already be accounting for this, but I think `INT_MIN/-1 > INT_MAX` for signed types.

PR #4719, opened 2020-03-16
Fix invalid extern stage in storage_folding test

abadams:
Bounds queries are supposed to nest. If a region O of the output
requires region I of the input, then a subset of O is supposed to
require a subset of I. Not sure what was going on with this extern
stage, but it didn't obey that property. I rewrote it.


steven-johnson:
Would it be possible to add some extra code to `debug` mode to sniff for bad externs? (Would be useful for diagnosing possible flakiness inside Google)

abadams:
Hrm, I don't think that would be easy. However #4713 makes this always true rather than relying on the extern stage to behave - it gives the extern stage the intersection of the outer bounds query and the inner bounds query. So if the extern stage hasn't done the right thing, it will get an input smaller than it expected and throw a bounds error (or just read out of bounds and crash).

abadams:
ptal, this one is green

PR #4720, opened 2020-03-16
Add a promise_clamped intrinsic

abadams:
This intrinsic has proved useful in two branches as a better way to
lower GuardWithIf. Instead of relying on bounds inference being able to
understand the if statement, we just declare the bounds of the variable
inside the if using a version of clamp that is guaranteed to be a no-op
at the place it occurs.

This PR adds the intrinsic and simplification rules for it.


src/IROperator.cpp
```diff
@@ -1136,6 +1136,20 @@ Expr unsafe_promise_clamped(const Expr &value, const Expr &min, const Expr &max)
                                 Internal::Call::Intrinsic);
 }
 
+namespace Internal {
+Expr promise_clamped(const Expr &value, const Expr &min, const Expr &max) {
```
steven-johnson: Worth adding a comment emphasizing that this is really really really not for external use

dsharletg:
Is there any difference between this and `unsafe_promise_clamped` other than the name? I couldn't find one in the code.

abadams:
There are two. First, this is never checked, even when the CheckUnsafePromises target flag is on. We know it's true. Second, it's context-dependent. We assert that the thing is bounded at that point in the code, but it might not be true elsewhere (e.g. maybe we're inside an if). For unsafe_promise_clamped the arg is known to be bounded everywhere, though you can't really exploit that because when you're in checked mode you don't want to trigger UB before you hit the check itself. I'll extend the comment to clarify.

steven-johnson:
> I'll patch this into Google to do some testing

no failures, but of course there aren't, since nothing is injecting the these calls yet :-)

abadams:
Every guardwithif schedule is injecting the calls! So that's actually great news.

steven-johnson:
Wait, I might be speaking too soon: a handful of things seem to be failing with an ~infinite amount of time spent in HexagonOptimize::EliminateInterleaves. Not sure how that could be related (yet).

steven-johnson:
Yeah, we have some hexagon cases that now seem to take waaaaay longer than before -- not sure if literally infinite, or just exponentially slower such that we hit timeouts. Puzzling though, as these changes don't seem to obviously affect HVX output.

dsharletg:
GuardWithIf does far more dramatic things to HVX code than CPU code, due to the vector size. I've seen 128 nested ifs in HVX code (1024 bit SIMD), where the equivalent would be 16 or 32 on typical targets (128 or 256 bit SIMD).

steven-johnson:
FYI: one case I'm seeing it in Google is in our version of apps/nn_ops/DepthwiseConvolution; it's very similar (but not identical), but so far I can't replicate the slowdown running the public version directly (in this branch). Still investigating.

steven-johnson:
OK, I think what's going on is that our internal version has a different schedule, but one that *already* resulted in pathologically slow times (which just happened to be under the timeout limit for the builders); this change has made is slower-enough to time out now, but I'm not sure how much slower (ie whether it's significant or just The Last Straw). Let me get some numbers.

abadams:
Happy to chew on a repro if you can port the pathological schedule to the open source version.

steven-johnson:
This case is pathological even on master; opened as new issue https://github.com/halide/Halide/issues/4735

(Currently running on this branch as well to see if it's substantially slower or not)

steven-johnson:
FYI, on this branch it's *slightly* slower:

```
real	47m37.052s
user	47m17.677s
sys	0m19.158s
```

...but that could be just noise. (Clearly there is an underlying pathology to fix.)

abadams:
Hopefully better in master in #4744

steven-johnson:
register_shuffle failing on windows:

```
317/319 Test #259: correctness_register_shuffle .............................***Failed    5.59 sec
Error: Access pattern for f68_global_wrapper$0.0 does not meet the requirements for its store_at location. All access to an allocation scheduled inside a loop over GPU threads and outside a loop over GPU lanes must obey the following constraint:
The index must be affine in .__thread_id_x with a consistent linear term across all stores, and a constant term which, when divided by the stride (rounding down), becomes a multiple of the warp size (32).
f68_global_wrapper$0.0 is stored to at the following indices by multiple lanes:
  ramp((.__thread_id_x*2), 1, 2)
  ramp(((.__thread_id_x*2) + 128), 1, 2)
  ramp(((.__thread_id_x*2) + 256), 1, 2)
  ramp(((.__thread_id_x*2) + 384), 1, 2)
  ramp(((.__thread_id_x*2) + 512), 1, 2)
  ramp(((.__thread_id_x*2) + 640), 1, 2)
  ramp(64, 1, 2)
  ramp(192, 1, 2)
  ramp(320, 1, 2)
  ramp(448, 1, 2)
  ramp(576, 1, 2)
  ramp(704, 1, 2)
And loaded from at the following indices:
  (.__thread_id_x*2)
  ((.__thread_id_x*2) + 1)
  ((.__thread_id_x*2) + 2)
  ((.__thread_id_x*2) + 1)
  ((.__thread_id_x*2) + 1)
  ((.__thread_id_x*2) + 2)
  ((.__thread_id_x*2) + 3)
  ((.__thread_id_x*2) + 2)
  ((.__thread_id_x*2) + 128)
  ((.__thread_id_x*2) + 129)
  ((.__thread_id_x*2) + 130)
  ((.__thread_id_x*2) + 129)
  ((.__thread_id_x*2) + 129)
  ((.__thread_id_x*2) + 130)
  ((.__thread_id_x*2) + 131)
  ((.__thread_id_x*2) + 130)
  ((.__thread_id_x*2) + 256)
  ((.__thread_id_x*2) + 257)
  ((.__thread_id_x*2) + 258)
  ((.__thread_id_x*2) + 257)
  ((.__thread_id_x*2) + 257)
  ((.__thread_id_x*2) + 258)
  ((.__thread_id_x*2) + 259)
  ((.__thread_id_x*2) + 258)
  ((.__thread_id_x*2) + 384)
  ((.__thread_id_x*2) + 385)
  ((.__thread_id_x*2) + 386)
  ((.__thread_id_x*2) + 385)
  ((.__thread_id_x*2) + 385)
  ((.__thread_id_x*2) + 386)
  ((.__thread_id_x*2) + 387)
  ((.__thread_id_x*2) + 386)
  ((.__thread_id_x*2) + 512)
  ((.__thread_id_x*2) + 513)
  ((.__thread_id_x*2) + 514)
  ((.__thread_id_x*2) + 513)
  ((.__thread_id_x*2) + 513)
  ((.__thread_id_x*2) + 514)
  ((.__thread_id_x*2) + 515)
  ((.__thread_id_x*2) + 514)
  ((.__thread_id_x*2) + 640)
  ((.__thread_id_x*2) + 641)
  ((.__thread_id_x*2) + 642)
  ((.__thread_id_x*2) + 641)
  ((.__thread_id_x*2) + 641)
  ((.__thread_id_x*2) + 642)
  ((.__thread_id_x*2) + 643)
  ((.__thread_id_x*2) + 642)
The problematic indices are:
  ramp(64, 1, 2)
  ramp(192, 1, 2)
  ramp(320, 1, 2)
  ramp(448, 1, 2)
  ramp(576, 1, 2)
  ramp(704, 1, 2)
```

abadams:
An important if statement went awol. Not sure why this didn't fail the linux builds, because it does for me. Perhaps only the windows builder has a new enough GPU for warp shuffles.

abadams:
Wait, belay that, the if statement is there, but it has changed to a form that the pass doesn't recognize. Easy to fix.

abadams:
Am I good to merge this, or should I wait for #4744 ?

steven-johnson:
No need to wait.

PR #4722, opened 2020-03-16
Add simplifier rules for for/if combinations

abadams:
First, in the following:
```
if (extent > 0) { for (x = 0; x < extent; x++) {...}}

```
The if is redundant and can be removed.

Next, we can fold away `if (likely(true/false))` in the same way we fold
away `if (true/false)`

Finally, we used to transform:
```
for (x = 0; x < min(1, foo); x++)
```
to
```
if (0 < foo)
```
Because a for loop where the extent is at most one is just an if.

However, this is not a good mutation if done in a vectorized loop,
because a for loop with an extent that varies across the vector lanes is
considered a likely-if, and an all-true vectorized case is generated.
This doesn't happen for the regular if. So I just squashed this
transformation if we're inside a vector loop.

### DIFF
--- a/src/Simplify.cpp
+++ b/src/Simplify.cpp
@@ -14,7 +14,7 @@ using std::pair;
 using std::string;
 using std::vector;
 
-#if LOG_EXPR_MUTATIONS || LOG_STMT_MUTATIONS
+#if (LOG_EXPR_MUTATIONS || LOG_STMT_MUTATIONS)
 int Simplify::debug_indent = 0;
 #endif
 
--- a/src/Simplify_Internal.h
+++ b/src/Simplify_Internal.h
@@ -90,9 +90,12 @@ public:
         }
     };
 
-#if LOG_EXPR_MUTATIONS
+#if (LOG_EXPR_MUTATORIONS || LOG_STMT_MUTATIONS)
     static int debug_indent;
+#endif
 
+
+#if LOG_EXPR_MUTATIONS
     Expr mutate(const Expr &e, ExprInfo *b) {
         const std::string spaces(debug_indent, ' ');
         debug(1) << spaces << "Simplifying Expr: " << e << "\n";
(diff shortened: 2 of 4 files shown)


steven-johnson:
```
FAILED TEST: gpu_specialize
Internal Error at /home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-1000/halide/src/PartitionLoops.cpp:925 triggered by user code at : Unexpected construct inside if statement: if ((uint1)p1) {
 let f$1.s0.y.yi.base = min(g$1.s0.y.yi.base, (g$1.s0.y.yi.base + -2))
 let f$1.s0.x.xi.base = min(g$1.s0.x.xi.base, (g$1.s0.x.xi.base + -2))
 for (f$1.s0.y.yi, 0, 4) {
  for (f$1.s0.x.xi, 0, 4) {
   __shared[((((f$1.s0.y.yi + f$1.s0.y.yi.base) - (t163 + g$1.s0.y.yi.base))*t164) + ((f$1.s0.x.xi + f$1.s0.x.xi.base) - (t163 + g$1.s0.x.xi.base)))] = ((f$1.s0.x.xi + f$1.s0.x.xi.base) + (f$1.s0.y.yi + f$1.s0.y.yi.base))
  }
 }
} else {
 let f$1.s0.y.yi.base = min(g$1.s0.y.yi.base, (g$1.s0.y.yi.base + -6))
 let f$1.s0.x.xi.base = min(g$1.s0.x.xi.base, (g$1.s0.x.xi.base + -6))
 for (f$1.s0.y.yi, 0, 8) {
  for (f$1.s0.x.xi, 0, 8) {
   __shared[((((f$1.s0.y.yi + f$1.s0.y.yi.base) - (t163 + g$1.s0.y.yi.base))*t164) + ((f$1.s0.x.xi + f$1.s0.x.xi.base) - (t163 + g$1.s0.x.xi.base)))] = ((f$1.s0.x.xi + f$1.s0.x.xi.base) + (f$1.s0.y.yi + f$1.s0.y.yi.base))
  }
 }
}
```

PR #4730, opened 2020-03-16
Bump the minimum cmake vesion to 3.15.7

seanprime7:
ece4783c1a introduced a use of $<CXX_COMPILER_ID:compiler_ids> where
compiler_ids is a comma-separated list.
As $<CXX_COMPIER_ID:compiler_id> accepted only one compiier ID until
3.14, 3.15.x or higher is now required.

3.15.7 seems to be the latest documented 3.15 release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.15.7)
 project(Halide)
 
 set(CPACK_PACKAGE_VENDOR "Halide")


steven-johnson:
There's been a discussion about CMake version (some argue 3.14 is too new, some argue we should go all the way to 3.16). I'll defer to Alex here. (In the meantime, I'll see what version(s) our buildbots have installed.)

steven-johnson:
In the meantime, a quicker fix might be just to split that line into three lines instead (PR welcomed)

seanprime7:
Okay.. Then I will close this PR as you guys are already having a separate discussion on the cmake minimum version requirement.

I will create a new PR for what you suggested above.

alexreinking:
Ah, lovely... thanks for the catch. I didn't notice that in the documentation.

PR #4731, opened 2020-03-16
Parallelize clang-tidy runs

abadams:
Using a script provided by llvm. This makes it way faster to clang-tidy a branch before opening a PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2298,23 +2298,37 @@ $(BIN_DIR)/HalideTraceDump: $(ROOT_DIR)/util/HalideTraceDump.cpp $(ROOT_DIR)/uti
 format:
 	find "${ROOT_DIR}/apps" "${ROOT_DIR}/src" "${ROOT_DIR}/tools" "${ROOT_DIR}/test" "${ROOT_DIR}/util" "${ROOT_DIR}/python_bindings" -name *.cpp -o -name *.h -o -name *.c | xargs ${CLANG}-format -i -style=file
 
-# Run clang-tidy on the core source files
+# run-clang-tidy.py is a script that comes with LLVM for running clang
+# tidy in parallel. Assume it's in the standard install path relative to clang.
+RUN_CLANG_TIDY ?= $(shell dirname $(CLANG))/../share/clang/run-clang-tidy.py
+
+# Run clang-tidy on everything in src/. In future we may increase this
+# surface. Not doing it for now because things outside src are not
+# performance-critical.
+CLANG_TIDY_TARGETS= $(addprefix $(SRC_DIR)/,$(SOURCE_FILES))
+
+INVOKE_CLANG_TIDY ?= $(RUN_CLANG_TIDY) -p $(BUILD_DIR) $(CLANG_TIDY_TARGETS) -clang-tidy-binary $(CLANG)-tidy -clang-apply-replacements-binary $(CLANG)-apply-replacements -quiet
+
 $(BUILD_DIR)/compile_commands.json:
+	mkdir -p $(BUILD_DIR)
 	echo '[' >> $@
-	BD=$(realpath $(BUILD_DIR)); \
-	SD=$(realpath $(SRC_DIR)); \
+	BD=$$(realpath $(BUILD_DIR)); \
+	SD=$$(realpath $(SRC_DIR)); \
+	ID=$$(realpath $(INCLUDE_DIR)); \
 	for S in $(SOURCE_FILES); do \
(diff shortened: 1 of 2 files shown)


steven-johnson:
On my Mac, `make clang-tidy-fix` fails with `/Users/srj/llvm-trunk-install/bin/../share/clang/run-clang-tidy.py: No such file or directory`

abadams:
Does /Users/srj/llvm-trunk-install/bin/clang-tidy exist for you? On top of the instructions in the readme I think you need to do a make install and enable the clang-tools-extra project.

steven-johnson:
> Does /Users/srj/llvm-trunk-install/bin/clang-tidy exist for you? On top of the instructions in the readme I think you need to do a make install and enable the clang-tools-extra project.

Nope. We should add that to the README.

abadams:
Done.

steven-johnson:
After doing that, I now can run it on OSX, but it is unable to find ~any of the system <includes>, e.g.  `/Users/srj/GitHub/Halide/src/IR.h:8:10: error: 'string' file not found [clang-diagnostic-error] #include <string>`

steven-johnson:
I also get a number of warnings of the form 

```
/Users/srj/GitHub/Halide/src/IR.h:666:68: error: the parameter 'param' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
    static Expr make(Type type, const std::string &name, Parameter param) {
```

abadams:
Can you try adding libcxx;libcxxabi to the list of enabled projects? For this to work you need to be able to actually compile c++ with the llvm that you built.

Those warnings sound like things that should be fixed. Not sure why I don't get them with my clang-tidy.

steven-johnson:
> libcxx;libcxxabi

Trying now, but FWIW, that adds a *lot* to the build (> 1000 files). 

Since clang-tidy might be available in prebuilt form, it might be worth smartening our script to do something like:
(1) if $CLANG_TIDY is defined, use that
(2) else, if `which clang-tidy` reports it is available, use that
(3) else, use the logic based on $LLVM_CONFIG that we do now

(AFAICT, Homebrew doesn't offer a clang-tidy formula on Mac, but the prebuilts you can download include it)

abadams:
I think the Makefile is already smart in that way. For any complete llvm install, prebuilt or not, setting LLVM_CONFIG correctly will find the rest of it. It sounds like to run clang-tidy you need a more complete llvm install than you currently have? Or maybe the homebrew prebuilt is only a partial build of llvm? 

The default value of RUN_CLANG_TIDY points to llvm's standard install location for run-clang-tidy.py using the location of CLANG, which comes from querying llvm-config where to look.

steven-johnson:
> I think the Makefile is already smart in that way. For any complete llvm install, prebuilt or not, setting LLVM_CONFIG correctly will find the rest of it. It sounds like to run clang-tidy you need a more complete llvm install than you currently have? Or maybe the homebrew prebuilt is only a partial build of llvm?
> 
> The default value of RUN_CLANG_TIDY points to llvm's standard install location for run-clang-tidy.py using the location of CLANG, which comes from querying llvm-config where to look.

I'm building from scratch, with `LLVM_ENABLE_PROJECTS="clang;lld;clang-tools-extra"`  (still rebuilding the libcxx stuff)

steven-johnson:
(FWIW: I never bothered trying `make clang-tidy` locally before now anyway, so it was likely broken already, and I will probably rely on the presubmit checks anyway most of the time. That said, might as well figure out the failure modes while we are here.)

abadams:
Yeah, I'm assuming this PR doesn't actually change the requirements for you. If  run-clang-tidy.py doesn't work then probably clang-tidy didn't either.

steven-johnson:
FYI: after rebuilding llvm with the libcxx;libcxxabi projects enabled, `make clang-tidy` is unchanged for me on this branch (still can't find system headers).

FYI #2: on the master branch, `make clang-tidy` just fails outright for me like so:

```
echo '[' >> bin/build/compile_commands.json
bash: bin/build/compile_commands.json: No such file or directory
make: *** [bin/build/compile_commands.json] Error 1
```

so, I guess this is a net improvement? Something is wonky but fine with me to land if you like.

abadams:
master is missing a key mkdir. It would be interesting to know if clang-tidy works if you manually create bin/build first.

steven-johnson:
(OK to land this without the buildbot testing IMHO)

PR #4733, opened 2020-03-17
Use CXX_COMPILER_ID with a single ID in CMakeLists.txt

seanprime7:
cmake 3.14, which is the minimum requirement for build, accepts a single
compiler ID in the CXX_COMPILER_ID expression.


seanprime7:
@steven-johnson per your suggestion, I have created this PR to replace https://github.com/halide/Halide/pull/4730

alexreinking:
Thanks for the catch. This is very annoying... I wish newer versions of CMake would warn when the minimum set version is too low.

PR #4734, opened 2020-03-17
Cmake python bindings

seanprime7:
Add cmake to `python_bindings`

This change requires `-DWITH_PYTHON_BINDINGS=ON` to be passed to `cmake` to enable the build of `python_bindings`.


BachiLi:
This is also done in https://github.com/halide/Halide/pull/4644, better coordinate with @alexreinking .

python_bindings/CMakeLists.txt
```diff
@@ -0,0 +1,247 @@
+find_package(pybind11 REQUIRED)
```
steven-johnson: Our existing code assumes/requires that PyBind11 is installed via `pip install requirements.txt`, so this fails for me on my OSX machine. What installation process is required to make this work? Documentation and/or useful error reporting is necessary. 

Since this is an all-header library, could we perhaps use FetchContent to automate this? https://cmake.org/cmake/help/v3.14/module/FetchContent.html

src/PythonExtensionGen.cpp
```diff
@@ -221,7 +225,14 @@ static __attribute__((unused)) int _convert_py_buffer_to_halide(
         }
         const char* type_codes = "bB?hHiIlLqQfd";  // integers and floats
         if (strchr(type_codes, *p)) {
+#if defined(WIN32) || defined(_WIN32)
+#pragma warning(push)
+#pragma warning(disable: 4244)
+#endif
             out->type.bits = buf.itemsize * 8;
```
steven-johnson: Instead of the pragmas, inserting an explicit cast would be cleaner and better (I assume that 4244 is the warning about possible-loss-of-data)

src/PythonExtensionGen.cpp
```diff
@@ -145,7 +145,11 @@ void PythonExtensionGen::compile(const Module &module) {
 extern "C" {
 #endif
 
-static __attribute__((unused)) int _convert_py_buffer_to_halide(
+static
+#if !defined(WIN32) && !defined(_WIN32)
+__attribute__((unused))
```
steven-johnson: Sometimes I just really hate C++. I know why this is here, but damn is it ugly (even for generated code). If we must have it, we should probably conditionalize on the compiler, not the API, i.e. `#ifndef _MSC_VER` instead of `#if !defined(WIN32) && !defined(_WIN32)`

python_bindings/CMakeLists.txt
seanprime7: I am using a miniconda environment and the pybind11 conda package comes with the make scripts whereas the pip package doesn't.

I will look into `FetchContent` instead then.

src/PythonExtensionGen.cpp
seanprime7: Yes.  The warning is about possible loss of data.  I will insert an explicit cast.

src/PythonExtensionGen.cpp
seanprime7: There is another occurrence of `#if !defined(WIN32) && !defined(_WIN32)` in the same source file.  Would you still prefer to have `#if !defined(_MSC_VER)` for this change?

steven-johnson: The other occurrence should go away once the explicit cast is added.

python_bindings/CMakeLists.txt
alexreinking: I'm already using FetchContent on cmake-modern

src/PythonExtensionGen.cpp
seanprime7: No. there is another occurrence in the same file that is not from this PR.

```
/* Older Python versions don't set up PyMODINIT_FUNC correctly. */
#if defined(WIN32) || defined(_WIN32)
#    define HALIDE_PYTHON_EXPORT __declspec(dllexport)
#else
#    define HALIDE_PYTHON_EXPORT __attribute__((visibility("default")))
#endif
```

alexreinking:
This is completely redundant with the Python bindings build in `cmake-modern`. It's way more complicated than it has to be because it doesn't use the modern CMake features introduced in that branch. And it doesn't tackle the issue of distributing functions to our clients for their own usage.

I would welcome PRs and bugfixes against the Python bindings build in `cmake-modern`, but I'm inclined to reject this for master.

seanprime7:
@alexreinking what's the ETA for `cmake-modern`?

alexreinking:
A few weeks. COVID-19 has delayed me lately.

seanprime7:
And you don't want this PR even as an interim solution?

steven-johnson:
> @alexreinking what's the ETA for `cmake-modern`?

@seanprime7 How painful would it be for you to maintain this as a patch on your local work until then?

python_bindings/CMakeLists.txt
```diff
@@ -0,0 +1,247 @@
...
+                                           LINK_FLAGS "${HALIDE_PY_LINK_FLAG}")
+target_link_libraries(halide_py PRIVATE "${HALIDE_COMPILER_LIB}")
+target_include_directories(halide_py PRIVATE "${HALIDE_INCLUDE_DIR}" "${HALIDE_TOOLS_DIR}")
+target_compile_options(halide_py PRIVATE $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wno-cast-qual>
+                                         $<$<CXX_COMPILER_ID:GNU>:-Wno-suggest-override>)
+
+install(TARGETS halide_py LIBRARY DESTINATION bin)
+
+_halide_library_runtime(host HALIDE_TARGET)
+
```
alexreinking: This should be in its own CMakeLists.txt in `src` and then `add_subdirectory(src)` here. This issue repeats below with the tests.

python_bindings/correctness/compile_to.py
```diff
@@ -37,7 +37,10 @@ def main():
 
         f.compile_to_file(os.path.join(tmpdir, "f_all"), args)
         assert os.path.isfile(os.path.join(tmpdir, "f_all.h"))
-        assert os.path.isfile(os.path.join(tmpdir, "f_all.o"))
+        if hl.get_target_from_environment().os == hl.TargetOS.Windows:
+            assert os.path.isfile(os.path.join(tmpdir, "f_all.obj"))
+        else:
+            assert os.path.isfile(os.path.join(tmpdir, "f_all.o"))
```
alexreinking: This would be good (also along with the PythonExtensions.cpp changes) to pull into its own PR.

src/PythonExtensionGen.cpp
```diff
@@ -259,10 +270,10 @@ static __attribute__((unused)) int _convert_py_buffer_to_halide(
...
 static struct PyModuleDef _moduledef = {
     PyModuleDef_HEAD_INIT,
-    .m_name=MODULE_NAME,
-    .m_doc=NULL,
-    .m_size=-1,
-    .m_methods=_methods,
+    MODULE_NAME,
+    NULL,
+    -1,
+    _methods,
```
alexreinking: Why was this deleted?

seanprime7:
> 
> 
> > @alexreinking what's the ETA for `cmake-modern`?
> 
> @seanprime7 How painful would it be for you to maintain this as a patch on your local work until then?

Since I have a local mirror/fork of halide, it's not too bad to maintain this patch locally.  And that's what I will have to do if this PR cannot be accepted as an interim solution.
But I think others in the community might benefit from this while waiting for `cmake-modern`.

src/PythonExtensionGen.cpp
seanprime7: `.<member name>=<value>` syntax isn't supported by MSVC as it is not part of the ISO C++ until C++20.

see https://en.cppreference.com/w/cpp/language/aggregate_initialization

alexreinking: Ah, that's good to know. Thanks for the fix, then!

python_bindings/correctness/compile_to.py
seanprime7: Done.  Created https://github.com/halide/Halide/pull/4743

src/PythonExtensionGen.cpp
seanprime7: Addressed it in https://github.com/halide/Halide/pull/4743

src/PythonExtensionGen.cpp
seanprime7: Addressed it in https://github.com/halide/Halide/pull/4743

seanprime7:
So.. if you still want this PR as an interim solution until `cmake-modern` is ready, then I will make the requested changes.
@alexreinking and @steven-johnson, please let me know.

steven-johnson:
> So.. if you still want this PR as an interim solution until `cmake-modern` is ready, then I will make the requested changes.

Thanks -- let us do some pondering and see what we think.

BachiLi:
I would vote for merging this to master. From the user side, the cmake building process shouldn't be too different.

steven-johnson:
If the end-user building experience is basically the same, I would be ok with landing this for now. If you make the changes discussed above I'll be happy to review it.

seanprime7:
Okay.  Then I will make the requested changes soon.

seanprime7:
* Turned python bindings build on by default
* Modularized `python_bindings/CMakeLists.txt` as requested
* Added FetchContent for pybind11

steven-johnson:
LGTM pending green at https://github.com/halide/Halide/pull/4771

steven-johnson:
The buildbots look good except for the Windows machines, which have never been properly configured for this before -- let me see if I can get them into shape. (We'll probably have to add logic to skip python on 32-bit Windows; AFAICT, it's painful to have both 32-bit and 64-bit Pythons installed on the same Windows box)

steven-johnson:
Update: can you just add something to skip building python bindings *entirely* on Windows for now? This will take some fiddling to get right on the buildbot side and I don't want to block it for that reason.

seanprime7:
I can turn off the python bindings build by default on Windows but leave it on on other platforms.
Does it sound good?

seanprime7:
I have turned off the python_bindings build by default on Windows

steven-johnson:
One more change needed to land this without breakage: in `build_travis.sh`, add `-DWITH_PYTHON_BINDINGS=OFF` to the cmake config; Travis doesn't have the right Python setup and for now we should just skip them there.

steven-johnson:
OK, one more clone for testing, hopefully we'll be good to go

steven-johnson:
(Just waiting on @alexreinking)

alexreinking:
Approved @steven-johnson - do buildbots need to run, or safe to merge?

steven-johnson:
I ran them in a clone.

PR #4736, opened 2020-03-17
Build python module with rpath on linux

Infinoid:
Add an rpath to the python module, to make libHalide.so easier to find.

The python test cases run much better with this on my ubuntu 18.04 machine.

### DIFF
--- a/python_bindings/Makefile
+++ b/python_bindings/Makefile
@@ -48,6 +48,9 @@ CCFLAGS := $(filter-out -Wstrict-prototypes,$(CCFLAGS))
 # DON'T link libpython* - leave those symbols to lazily resolve at load time
 # Cf. https://github.com/pybind/pybind11/blob/master/docs/compiling.rst#building-manually
 LDFLAGS += -lz $(USE_EXPORT_DYNAMIC)
+ifeq ($(UNAME), Linux)
+  LDFLAGS += -Wl,-rpath=$(dir $(LIBHALIDE))
+endif
 
 PY_SRCS=$(shell ls $(ROOT_DIR)/src/*.cpp)
 PY_OBJS=$(PY_SRCS:$(ROOT_DIR)/src/%.cpp=$(BIN)/src/%.o)


steven-johnson:
Can you expand a bit on what "runs much better" means?

Infinoid:
Sure.

The general problem I see is that libHalide.so is not in the dynamic linker's path by default.  Without the rpath and without setting LD_LIBRARY_PATH specially, any attempt to run a generator executable, or to import the halide module in python, results in an error message.

Generator failure looks like:
```
/home/infinoid/workspace/pappa/apps/halide/python_bindings/bin/runtime.gen -r runtime -o /home/infinoid/workspace/pappa/apps/halide/python_bindings/bin target=host
/home/infinoid/workspace/pappa/apps/halide/python_bindings/bin/runtime.gen: error while loading shared libraries: libHalide.so: cannot open shared object file: No such file or directory
```

Python module failure looks like:
```
  File "/home/infinoid/workspace/pappa/apps/halide/python_bindings/correctness/atomics.py", line 1, in <module>
    import halide as hl
ImportError: libHalide.so: cannot open shared object file: No such file or directory
```

With the patch, `make test` runs successfully as long as HALIDE_PATH points at a valid halide build.

I would be curious whether this works on Mac or other platforms.

steven-johnson:
Ah, so by "better" you mean "doesn't fail" :-)

We run the python tests on Linux for every presubmit, so clearly it's not failing *everywhere*. (They also run just fine on my local Linux box without and special values for LD_LIBRARY_PATH).

I'm not opposed to this change, but would like for us to get more clarity about why things seem to be working for us and not for you.

Infinoid:
Agreed, sounds good to me.

What does "ldd" say about your python module after you build it?

Here's mine (without the patch):
```
halide/python_bindings$ ldd bin/halide.cpython-36m-x86_64-linux-gnu.so 
	linux-vdso.so.1 (0x00007ffdf65cf000)
	libHalide.so => not found
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f53ae450000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f53ae238000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f53ade47000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f53adaa9000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f53aecf1000)
```

steven-johnson:
at current top-of-tree:

```
$ ldd bin/halide.cpython-37m-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007ffdd7387000)
	/usr/local/google/home/srj/GitHub/Halide/python_bindings/../distrib/bin/libHalide.so (0x00007f9748ad3000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f97488d0000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f97488b6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f97486f6000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f97486d9000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f97486d2000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f97486b1000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f974856c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f974eee5000)
```

Infinoid:
Details from discussion in gitter channel, for posterity:

steven-johnson builds halide by running "make" directly in the source folder.  I built it in a separate build folder with cmake.  The resulting `libHalide.so` files are subtly different.  The cmake one has a SONAME in the `objdump -x` output, the make one does not.  The SONAME field causes later links to just include the name of the library, not the full path, causing the issues I mentioned above.

steven-johnson:
I am fine with this change. Adding @alexreinking to comment on any CMake implications.

Infinoid:
Yeah, I don't know what the right solution is.

If libHalide.so is installed system-wide, then SONAME is a good idea, and rpath is not needed.  If libHalide.so lives in a build folder under $HOME somewhere, and it has a SONAME, then rpath or $LD_LIBRARY_PATH is needed.

alexreinking:
I'm pretty sure CMake sets up RPath during the build for this sort of thing already. Shouldn't affect me in any way.

steven-johnson:
I suspect this will be good to go once https://github.com/halide/Halide/pull/4741 finishes; you probably want to sync to current master to clean up the red x's.

Issue #4708, opened 2020-03-13
Target(string) ctor invokes JIT if the string contains "cuda"


steven-johnson:
If you execute a generator binary with `target=anything-cuda`, this happens to call `Target(string)`, which calls `merge_string`, which calls `get_host_cuda_capability()` if the string contains `cuda`... which ends up calling `get_device_interface_for_device_api()`, which ends up calling `JITSharedRuntime::get`, which basically initializes all of the JIT's shared runtime code.

This is a nontrivial and unexpected cost for simply constructing a `Target` from a string. Surely we can circumvent this?

abadams:
We need some way to sniff the host cuda capability to avoid using lowest-common-denominator features. Initializing the runtime is simplest, because that's where the code capable of loading libcuda is, and it's very likely to if you're saying host-cuda, you're about to compile cuda code for the host anyway. We could try to sniff the capability some other way, but it will at least involve loading a shared library, possibly redundantly.

When we didn't do this it was akin to always using the target x86-64 no matter what vector instruction set exists on the host target. Sniffing the CPU feature set is much simpler than sniffing the host cuda feature set, unfortunately.

steven-johnson:
I get all of that, but I think that invoking the JIT by way of a Target() ctor is extremely suprising -- Target is meant to be a pretty lightweight class, this is not a side effect I'd expect. Perhaps a more explicit API to request this behavior would be appropriate.

(BTW: I only even noticed this because certain arm32-under-qemu generators are crashing somewhere inside this point... but the arm32 jit (in general) seems to work fine under qemu, so something else here is odd.)

abadams:
Yeah, I had trouble thinking of any other way to do it. We resolve "host" eagerly in the target constructor, and that requires some sniffing.

We could defer that somehow. We could make the sniffing cheaper.

Honestly it's weird that we JIT-compile the host runtimes at all. They could be statically compiled and linked into libHalide and things would be faster. Would clean up a bunch of symbol resolution nonsense when jitting. It's just annoying to set up the build for that.

steven-johnson:
The eager host resolution is pretty cheap though - lazily sniff in `calculate_host_target()` and cache it, and the sniffing itself is arguably not much more expensive than the string ops we do elsewhere.

Agreed re: jit-compiling the host runtimes. Is it really hard to set up the build for that? (Out of curiosity, was it attempted long ago and abandoned?)

abadams:
[...]

PR #4739, opened 2020-03-17
Add a way to disable host-cuda-capability sniffing

steven-johnson:
This is a workaround for Issue #4708, for the cases in which it causes outright problems rather than merely a performance hit (e.g. detecting CUDA while running under QEMU user-mode detection is crashy).

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -472,7 +472,8 @@ bool merge_string(Target &t, const std::string &target) {
         !t.has_feature(Target::CUDACapability32) &&
         !t.has_feature(Target::CUDACapability35) &&
         !t.has_feature(Target::CUDACapability50) &&
-        !t.has_feature(Target::CUDACapability61)) {
+        !t.has_feature(Target::CUDACapability61) &&
+        Internal::get_env_variable("HL_DISABLE_DETECT_HOST_CUDA_CAPABILITY") != "1") {
         // Detect host cuda capability
         t.set_feature(get_host_cuda_capability(t));
     }


steven-johnson:
(Failed builds are due to buildbot restart; I think this is OK to land with the test coverage already completed. PTAL.)

abadams:
What makes it crashy - using the JIT inside an emulator?

steven-johnson:
> What makes it crashy - using the JIT inside an emulator?

I haven't been able to nail down a specific reason, but it's not the JIT.

In this mode, it emulates the user-supplied (eg) arm code, but captures all syscalls and forwards them to the host kernel. This isn't perfect, of course: e.g., if you read from `/proc/cpuinfo` you'd see the host info, not the emulated-cpu info. It also (presumably) means that if you have arm code (eg our runtime) that attempts to load a shared library or driver that is only present for the host, you're gonna end up trying to call x86-64 code from emulated-arm code, and I'm skeptical that QEMU handles that well.

This could possibly be made to work by installing a full root of arm-native stuff and point QEMU at that (instead of letting it use the host stuff), but it would be fiddly to get right, and I'm pretty sure that testing CUDA under QEMU on a VM that has ~no useful GPU hardware is probably not important.

abadams:
So there's no reasonable OS sniffing we could do then. Rewriting host-cuda sniffing to load libcuda.so directly would also not work, because it wouldn't be able to call the code it finds. Using "host" works because we don't even try to do host feature sniffing on arm (and if we did, it would probably be inline assembly in Target.cpp). 

Why are we even constructing a host-cuda target inside qemu?

steven-johnson:
> Why are we even constructing a host-cuda target inside qemu?

An excellent question! The reason is that (as of now) our Make and CMake build systems take an all-or-nothing approach to crosscompiling; there isn't a way to distinguish that some things (eg Generators, libHalide when used in AOT mode) should always be build for the 'host' system for testing, while other things (all JIT-based tests, libHalide when used by JIT tests) must be compiled for the 'target' system. As a result, *everything* is build for the target system, and even the AOT parts (like Generators) are run via emulation.

Making this distinction is doable, of course (our Blaze rules already make this distinction), but a nontrivial amount of work. Running everything under QEMU initially is (by far) the fastest and simplest way to get crosscompilation up and running.

abadams:
Ok, I *think* that makes sense, but it still doesn't tell me why we're trying to run a cuda test in this setup.

steven-johnson:
> why we're trying to run a cuda test in this setup.

We're not. (That's why this is so amusing.)

There are a handful of AOT tests that AOT-generate Cuda code (eg `generator_aot_cxx_mangling`). Since they specify `target=host-cuda` to the Generator, the *Generator* ends up trying to load Cuda (via the JIT, via the Target-from-string ctor) at AOT compilation time.

(Even more ironically, `generator_aot_cxx_mangling` never even attempts to call the Cuda aot code. It just wants to check for link errors.)

abadams:
But I don't see anywhere that sets host-cuda in aot_cxx_mangling, so we must be setting HL_TARGET=host-cuda somewhere outside the build system?

abadams:
Regardless of where it's set, you can explicitly tell it to use a fixed cuda version instead of sniffing by just setting one of the cuda_capability features. E.g. host-cuda-cuda_capability_30 is unambiguous and should skip any sniffing code for the GPU.

steven-johnson:
> But I don't see anywhere that sets host-cuda

Makefile:1418

> you can explicitly tell it to use a fixed cuda version instead of sniffing

D'oh. Yeah. Let me give that a try first.

abadams:
Ah! I was grepping for host-cuda. Yeah let's just make that test use an unambiguous target.

steven-johnson:
Using an unambiguous target is a simper workaround.

PR #4742, opened 2020-03-17
python_bindings should generate directly to static_library, not c_source

steven-johnson:


steven-johnson:
Failures are unrelated performance flakes

PR #4743, opened 2020-03-17
Fix python binding build and test issues on Windows

seanprime7:
This PR fixes python bindings build and test issues on Windows.


seanprime7:
cc @alexreinking @steven-johnson

PR #4758, opened 2020-03-19
Add helper functions for a Tuple-like select construction over a channel dimension

BachiLi:
From a Gitter discussion, quoting @jrk 

> Wanting to use an extra channel vs. a Tuple is so common and similar, a few related thoughts > spring to mind:
> -- Could we have initializer-list-like syntax to make a channel select almost the same way as that Tuple? So like pixels(x,y,c) = select{c, x, y, depth(x,y), 1.f}` would work? Or maybe it could be made even closer to just writing the tuple version?
> -- Can we allow translation or desugaring between tuple-valued Funcs and Funcs with a channel dimension entirely after the fact, assuming types are consistent? Can we unify them more completely?

This PR added a helper function called "select_by_id" that implements the first proposal. Ideas for names and other related helper functions are very welcome.


abadams:
Naming things is hard. What if you could index a tuple using an Expr? Examples:
```
f(x, y, c) = Tuple{x, y, z}[c]

f(x, y) = {a, b};
g(x, y, c) = f(x, y)[c];
```

Could be useful instead of or in addition to select_by_id

Other name ideas: select_by_index, index, pack_channels, channel_select, mux, switch_expr, switch_

src/IROperator.h
```diff
@@ -797,6 +797,33 @@ inline Tuple tuple_select(const Expr &c0, const Tuple &v0, const Expr &c1, const
...
+ *                               25); // Blue
+ * This is tedious when the list is long. The following functions
+ * provide convinent syntax that allow one to write:
+ * img(x, y, c) = select_by_id(c, 100, 50, 25);
+ * or
+ * img(x, y, c) = select_by_id(c, {100, 50, 25});
+ */
+// @{
+inline Expr select_by_id(const Expr &id, const std::vector<Expr> &values) {
+    user_assert(values.size() > 0) << "select_by_id only accepts values with size > 0.\n";
```
steven-johnson: should be `size >= 2`

src/IROperator.h
```diff
@@ -797,6 +797,33 @@ inline Tuple tuple_select(const Expr &c0, const Tuple &v0, const Expr &c1, const
...
+    Expr result = values.back();
+    for (int i = (int)values.size() - 2; i >= 0; i--) {
+        result = select(id == i, values[i], result);
+    }
+    return result;
+}
+template<typename... Args,
+         typename std::enable_if<Halide::Internal::all_are_convertible<Expr, Args...>::value>::type * = nullptr>
+inline Expr select_by_id(const Expr &id, Args &&... values) {
```
steven-johnson: IMHO, we should move away from this pattern, in favor of just a vector. For one, that's mostly what the Python API does (no varargs, only lists); for another, in C++11 etc adding the extra {brace} wrapping is trivial, and the runtime overhead of a temporary vector construction is unlikely to be meaningful.

steven-johnson:
syntax is reasonable, but `select_by_id` isn't the best (it's definitely not an "id"). `select_by_index` perhaps.

Indexing a Tuple is tempting in terms of terseness, but IMHO makes for some really nonobvious-to-read code possibilities; I'd much prefer we err on the side of readability here.

src/IROperator.h
BachiLi: Fixed.

src/IROperator.h
BachiLi: Removed.

BachiLi:
Thanks! I went with Steven's suggestion (name it `select_by_index` and didn't add the fancy Tuple indexing). PTAL

abadams:
if you git grep for "select(c == 0" you'l find a bunch of places where this could be applied immediately. I think it be worthwhile to do so so that people don't get lead astray by bad examples.

BachiLi:
> if you git grep for "select(c == 0" you'l find a bunch of places where this could be applied immediately. I think it be worthwhile to do so so that people don't get lead astray by bad examples.

Nice suggestion. Just pushed the changes.

dsharletg:
+1 for the name `mux`, this is the name we already use for similar helper functions we have in our codebase. `select_by_index` is OK too but IMO a bit verbose for such a basic built-in function.

src/IROperator.h
```diff
@@ -797,6 +797,26 @@ inline Tuple tuple_select(const Expr &c0, const Tuple &v0, const Expr &c1, const
...
+ * img(x, y, c) = select(c == 0, 100, // Red
+ *                       c == 1, 50,  // Green
+ *                               25); // Blue
+ * This is tedious when the list is long. The following function
+ * provide convinent syntax that allow one to write:
+ * img(x, y, c) = select_by_index(c, {100, 50, 25});
+ */
+// @{
+inline Expr select_by_index(const Expr &id, const std::vector<Expr> &values) {
+    user_assert(values.size() >= 2) << "select_by_index only accepts values with size >= 2.\n";
```
dsharletg: It might be good to catch type mismatchs among the values early here, so the error is from `select_by_index`, rather than from the internal `select`.

BachiLi: Done. Thanks!

BachiLi:
I agree with Dilon that shorter names are better so I went for "mux". I'll merge this tomorrow if the tests are passing.

Issue #4749, opened 2020-03-18
All performance-sensitive tests should be under performance/


steven-johnson:
There are a handful of things under tutorial/ and autoschedule/ that do performance tests, and thus can be flaky when run on VMs or shared systems. We should ensure that any tests that rely on uniform system performance are either segregated under the performance/ test suite, or changed into non-fatal warnings if they shouldn't be moved (e.g. for tutorials), to avoid flakiness on test systems.

PR #4759, opened 2020-03-19
Don't even try to run test_performance on GHA VMs

steven-johnson:
Even running serially, too much flakiness.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,7 @@
 # - python is skipped for arm-32, arm-64
 # - apps are skipped for x86-32, arm-32, arm-64
 # - build-and-test matrix is probably too large
+# - test_performance is skipped everywhere (too much flakiness)
 #
 # TODO (stuff that could be usefully added, perhaps)
 # - build + test of WASM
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -447,10 +448,12 @@ jobs:
         fi
 
         # tutorial has some performance measurements that can be flaky if we run them in parallel
-        TEST_GROUPS_SERIAL="performance tutorial"
+        TEST_GROUPS_SERIAL="tutorial"
```
abadams: There's one performance test in the tutorials I think. How about changing it to not be an error in this PR?

steven-johnson: also at least one in `auto_schedule` (see https://github.com/halide/Halide/issues/4749). Not sure what to do about that one.

steven-johnson:
PTAL

tutorial/lesson_16_rgb_run.cpp
```diff
@@ -23,6 +23,14 @@
 
 #include "halide_benchmark.h"
 
+#define check_timing(faster, slower) \
```
abadams: The advantage of the macro is that you get the variable names in the printf, but I think we should avoid it in a tutorial. People might be thrown by the stringification of arguments and think they're not understanding something important.

There's a printf just before each assert already, so how about just:
```
void check_timing(double faster, double slower) {
  if (faster < slower) {
    fprintf(stderr, "Warning: performance was worse than expected. %f should be less than %f\n", faster, slower);
  }
}
```

steven-johnson: Done

steven-johnson:
PTAL

Issue #4756, opened 2020-03-19
GHA buildbots need to coalesce commits for builds


steven-johnson:
Right now, the GHA buildbots fire off a new full test setup for every commit. We should put some slack into the system (e.g. 5 minutes) to allow for the "oops I forgot something" syndrome.

Alternately, would it be possible to set up an easy way to require manual triggering of builds? (eg a button on the PR page?)  This could be done via `curl` but that is not a reasonable solution.

abadams:
If they can trigger based on which reviewers are assigned, there's a user called "halidebuildbots" that you can assign reviews to. I created it when I was attempted to get our buildmaster to do the same thing.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** The ask was to avoid burning a full build on every intermediate push ("oops I forgot something"), or to allow manual gating. Every testing workflow now declares a `concurrency` group with `cancel-in-progress: true`, so a follow-up push immediately cancels the superseded run, and each job is gated on a `skip_buildbots` PR label plus `paths-ignore` filters.

**Evidence:** .github/workflows/testing-linux.yml:23-25 (concurrency/cancel-in-progress) and :32 (`if: "!contains(github.event.pull_request.labels.*.name, 'skip_buildbots')"`); same pattern in testing-windows.yml, testing-macos.yml, testing-arm-linux.yml, testing-ecosystem.yml, clang-tidy.yml, pre-commit.yml

Closing: superseded runs are now canceled automatically via `concurrency` + `cancel-in-progress: true` in every testing workflow, and a `skip_buildbots` label plus `paths-ignore` filters let you opt out entirely. That covers the "oops I forgot something" case without needing an artificial delay.

PR #4762, opened 2020-03-19
Fix incorrect deinterleave of a shuffle

abadams:
Bug was introduced by a recent PR

### DIFF
--- a/src/Deinterleave.cpp
+++ b/src/Deinterleave.cpp
@@ -329,7 +329,7 @@ private:
                 int idx = i * lane_stride + starting_lane;
                 indices.push_back(op->indices[idx]);
             }
-            return Shuffle::make({op}, indices);
+            return Shuffle::make(op->vectors, indices);
         }
     }
 };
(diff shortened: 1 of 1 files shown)


shoaibkamil:
Yikes.

abadams:
The GHA bots seem to have double-triggered. Perhaps they are also triggered by the approval?

steven-johnson:
yeah, I still haven't figured out what the cause for multi triggering

abadams:
Oh you're right, this dates back to the origin of this code but never triggered. I have a PR somewhere that changes this code but it hasn't been merged yet.

steven-johnson:
So the two presubmit runs are because we are triggering when a PR is "opened" and again when a PR is "review requested". I really need to figure out how to avoid this in general (see https://github.com/halide/Halide/issues/4756). 

AFAICT, there isn't a baked-in tool in GHA to avoid redundant actions like this (I've made inquiries to verify I didn't miss something). 

I'm sure I can hand-roll something (the info I need seems to be available) -- I didn't think it would be a deal-breaker before I landed it, but it looks like it's a more significant issue than I thought. I'll get to work on it straightaway.

PR #3037, opened 2018-06-08
Move add_image_checks after bounds inference (See #3036)

abadams:
This seemingly minor change could have large consequences. I'll watch the buildbots. This would also need checking on Google pipelines.

PR #4766, opened 2020-03-19
Move first simplification before allocation bounds inference

abadams:
Removed the dependence of allocation bounds inference and storage
folding (and boxes_touched) on magic names, which means we can run full
simplification earlier. This should generally make allocation bounds
tighter, and also speeds up the compiler by giving allocation bounds
inference less IR to chew through (by a full 10% on local laplacian
lowering, and a similar amount on a 32x32 fft, 5% on lens blur).

See #4713 for the previous version of this PR.

### DIFF
--- a/src/AllocationBoundsInference.cpp
+++ b/src/AllocationBoundsInference.cpp
@@ -31,19 +31,6 @@ class AllocationInference : public IRMutator {
 
         Scope<Interval> empty_scope;
         Box b = box_touched(op->body, op->name, empty_scope, func_bounds);
-        if (touched_by_extern.count(f.name())) {
-            // The region touched is at least the region required at this
-            // loop level of the first stage (this is important for inputs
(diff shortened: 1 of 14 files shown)


steven-johnson:
```
/Users/halidenightly/build_bot_new/worker/mac-64-testbranch-1000/halide-build/bin/apps/gradient_autoscheduler/bin/host/demo.rungen --benchmarks=all --benchmark_min_time=1 --estimate_all
halide_error: Applying the constraints on Input buffer input to the required region made it smaller in dimension 1. Required size: -299 to 1. Constrained size: 0 to 100.
```

abadams:
Funnily enough, this is fixed by moving add_image_checks to where it should be: alongside allocation bounds inference. Maybe I need to do both changes in the one PR.

abadams:
Closing in favor of finishing #3037

PR #4768, opened 2020-03-20
Add SDK_ROOT control to hexagon_remote/Makefile

dpalermo:
- allows specifying only SDK_ROOT in a buildbot to point
    to where SDK storage is located without naming a specific
    SDK revision
  - needed to minimize buildbot script complexity when building
    multiple release streams that use differing SDK revisions
  - no change in behavior if SDK_ROOT env var is not present

### DIFF
--- a/src/runtime/hexagon_remote/Makefile
+++ b/src/runtime/hexagon_remote/Makefile
@@ -8,7 +8,8 @@
 # In order to build this, you'll need to ensure that the following
 # env vars are set correctly:
 #
-# HEXAGON_SDK_ROOT   : path to Qualcom Hexagon SDK
+# SDK_ROOT           : path to Qualcomm SDK storage
+# HEXAGON_SDK_ROOT   : path to Qualcomm Hexagon SDK
 # HEXAGON_TOOLS_ROOT : path to Qualcomm Hexagon Tools
 # HEXAGON_QAIC       : path to Qualcomm qaic compiler
 #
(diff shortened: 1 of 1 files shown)


dpalermo:
Just to give a bit more detail on the motivation for this change...

Our buildbot can't have SDKs installed in $HOME/Qualcom/Hexagon_SDK/...

Previously, what we did was specify the full path to Hexagon_SDK in our buildbot builders.py for each release/branch we maintained...for example:

   tip  
     export HEXAGON_SDK_ROOT=/prj/dsp/austin/llvm/test/SDK/Hexagon_SDK/3.3.0    
  
   BRANCH_HALIDE_2.1  
     export HEXAGON_SDK_ROOT=/prj/dsp/austin/llvm/test/SDK/Hexagon_SDK/3.4.1    
  
   BRANCH_HALIDE_2.2  
     export HEXAGON_SDK_ROOT=/prj/dsp/austin/llvm/test/SDK/Hexagon_SDK/3.5.0    
  
   master (BRANCH_HALIDE_2.3)
     export HEXAGON_SDK_ROOT=/prj/dsp/austin/llvm/test/SDK/Hexagon_SDK/3.5.1    
 
What we can now do with this change is just specify the top directory for SDK storage and let the Makefile on the branch select the version it wants to use:

  export SDK_ROOT=/prj/dsp/austin/llvm/test/SDK
 
In addition to simplifying the buildbot configuration, this also avoids problems where we move the Hexagon_SDK version forward in the Makefile and forget to have our QA team update the builders.py to match...or if we have QA update builders.py to a newer version and forget to move the Makefile forward.

dpalermo:
@dsharletg @pranavb-ca PTAL

dpalermo:
And related to SDK, @dsharletg, the next version of the SDK that will be coming out this year (SDK 4.0.0) will contain the pointer to implementation fix you suggested for the simulator HexagonWrapper.h...all private data has been replaced with a single pImpl pointer:
```
class DLL_A_EXPORT HexagonWrapper
{
private:
    void *pImpl;
```

dpalermo:
Sorry about the accidental close, pushed the wrong button the last comment.

dpalermo:
By just setting the SDK_ROOT env var then the specific version gets determined by the Makefile/branch...e.g. for tip:

`    HEXAGON_SDK_ROOT ?= ${SDK_ROOT}/Hexagon_SDK/3.3.3`

and you no longer need to set HEXAGON_SDK_ROOT just to override $HOME/Qualcomm with where the SDK storage is.

Having the SDK version in the Makefile is a good thing since we want to keep the version stable for any work on a given branch....rather than rely on the user or buildbot environment to have to specify it.

PR #4769, opened 2020-03-20
mux accepts tuples

abadams:
I had a sudden horrifying vision of someone having to write:

`mux(c, {f(x, y)[0], f(x, y)[1], f(x, y)[2]})`

So this lets mux take a tuple as the second argument. Also needed to add
an initializer list overload to prevent ambiguity.

I still think it would be a good idea to permit f(x, y)[c], for
uniformity with Tuple's existing operator[] for integers. I can imagine
people wondering why they can't just pass an Expr there.

### DIFF
--- a/python_bindings/correctness/iroperator.py
+++ b/python_bindings/correctness/iroperator.py
@@ -60,6 +60,19 @@ def test_mux():
     assert b[2] == 2
     assert b[3] == 3
 
+def test_mux_tuple():
+    f = hl.Func()
+    g = hl.Func()
+    x = hl.Var()
+    c = hl.Var()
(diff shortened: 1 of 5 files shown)


steven-johnson:
Please be sure to update the python_bindings version too (if necessary)

re: the operator[] overload, let's open that as a separate PR, with some examples to see how it looks in practice. My hunch is that it will make code using it less obvious when reading, but one way to find out. (Also: consider that the python bindings use [] for function subscripting, so this might be hard/impossible to do in python without ambiguity.)

steven-johnson:
"failures" are actually cancellations (all relevant tests were done so I cancelled to save buildbot time).

PR #4772, opened 2020-03-20
Add simplifications of min(x, y) < min(x, z), and similar

abadams:
In a comparison between two mins, if a term on one side dominates a term
on the other, we can drop it. Consider:
```
min(x, y) < min(z, w)
```

First we'll assume we can statically prove w <= y (e.g. because they are
equal). Then:
```
min(x, y) < min(z, w)
= (x < min(z, w)) || (y < min(z, w))
= (x < min(z, w)) || (y < z && y < w)
= (x < min(z, w)) || (y < z && false)
= (x < min(z, w)) || false
= x < min(z, w)
```
Alternatively if we can prove y < w
```
min(x, y) < min(z, w)
= (min(x, y) < z) && (min(x, y) < w)
= (min(x, y) < z) && (x < w || y < w)
= (min(x, y) < z) && (x < w || true)
= (min(x, y) < z) && true
= min(x, y) < z
```

You can do the same reasoning for max by flipping the sign on everything.

These rules were helpful #3037. Carving them out into a separate PR
here. The basic form of them existed in the rule synthesis branch. I
added variants with constants and more commutations.

### DIFF
--- a/src/Simplify_LT.cpp
+++ b/src/Simplify_LT.cpp
@@ -202,6 +202,38 @@ Expr Simplify::visit(const LT *op, ExprInfo *bounds) {
               rewrite(c1 < min(y, c0), fold(c1 < c0) && c1 < y) ||
               rewrite(c1 < max(y, c0), fold(c1 < c0) || c1 < y) ||
 
+              // Cases where we can remove a min on one side because
+              // one term dominates another. These rules were
+              // synthesized then extended by hand.
+              rewrite(min(z, y) < min(x, y), z < min(x, y)) ||
+              rewrite(min(z, y) < min(y, x), z < min(y, x)) ||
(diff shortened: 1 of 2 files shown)


steven-johnson:
I'll do the obligatory Google test

abadams:
@jn80842

steven-johnson:
no obvious issues in Google

abadams:
The windows bots don't seem to have fully functioning python installs right now. Passes on the other bots. Just needs a signoff.

abadams:
I've formally verified these using the tools in the super_simplify branch. Seemed like a thing to do on a Sunday afternoon. I need to get those tools into master at some point.

steven-johnson:
Yeah, I'm upgrading the Windows python stuff

PR #4773, opened 2020-03-20
Speed up add_image_checks

abadams:
by pre-trimming down to the portion of the stmt that actually touches
input/output buffers.

On master, timing add_image_checks for a few apps:

local_laplacian: 16ms
lens_blur: 18ms
interpolate: 11ms
fft: 37ms

On this branch:

local_laplacian: 2.4ms
lens_blur: 11ms
interpolate: 3.5ms
fft: 3.9ms

This change was pulled from #3037

### DIFF
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -101,7 +101,7 @@ class TrimStmtToPartsThatAccessBuffers : public IRMutator {
     using IRMutator::visit;
 
     Expr visit(const Call *op) override {
-        touches_buffer |= (buffers.find(op->name) != buffers.end());
+        touches_buffer |= (buffers.count(op->name) > 0);
         return IRMutator::visit(op);
     }
     Stmt visit(const Provide *op) override {
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -1,4 +1,5 @@
 #include "AddImageChecks.h"
+#include "IRMutator.h"
 #include "IRVisitor.h"
 #include "Simplify.h"
 #include "Substitute.h"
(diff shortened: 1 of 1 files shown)


src/AddImageChecks.cpp
```diff
@@ -93,6 +94,51 @@ class FindBuffers : public IRGraphVisitor {
...
 };
 
+class TrimStmtToPartsThatAccessBuffers : public IRMutator {
+    bool touches_buffer = false;
+    const map<string, FindBuffers::Result> &buffers;
+
+    using IRMutator::visit;
+
+    Expr visit(const Call *op) override {
+        touches_buffer |= (buffers.find(op->name) != buffers.end());
```
steven-johnson: nit: would this be easier to read as `buffers.count(op->name) > 0` ?

src/AddImageChecks.cpp
```diff
@@ -93,6 +94,51 @@ class FindBuffers : public IRGraphVisitor {
...
+    }
+    Expr visit(const Variable *op) override {
+        if (op->param.defined() && op->param.is_buffer()) {
+            touches_buffer |= (buffers.find(op->param.name()) != buffers.end());
+        }
+        return IRMutator::visit(op);
+    }
+
+    Stmt visit(const Block *op) override {
+        bool old_touches_buffer = touches_buffer;
```
steven-johnson: Is `ScopedValue<bool>` not a good option here?

abadams: I don't think so. The function mutates the value of old_touches_buffer as it goes - it doesn't just restore it at the end.

src/AddImageChecks.cpp
abadams: Interesting point. Currently in the compiler we have 77 instances of foo.find(bar) != foo.end() and a similar total number of instances of foo.count(bar) > 0, foo.count(bar) == 0, if (foo.count(bar)). Should we be changing them all? I guess it types out "foo" one fewer time.

abadams:
Only failing on the windows bots due to python issues. This change is not os-specific though, so I'm going to go ahead and merge.

PR #4774, opened 2020-03-20
Simplifier improvements from #3037

abadams:
Some simplifier improvements harvested from #3037. Bundled into a single PR to avoid overloading the bots, but done as separate commits to facilitate bisection. The are:

1) simplify min(x, likely(x)) -> x

I.e. the likely goes away. Augmented with better tracking of bounds of likely expressions. I believe this is more correct than the status quo, because a likely is supposed to mark which branch of the immediately containing min/max/select is the one to optimize for. This may result in fewer partitioned loops than before though.

2) The remove_dead_lets flag for the simplifier is now remove_dead_let_stmts. You always want to remove dead let exprs.

3) Two new rules to simplify select expressions

### DIFF
--- a/src/Simplify_Max.cpp
+++ b/src/Simplify_Max.cpp
@@ -25,9 +25,21 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
 
     // Early out when the bounds tells us one side or the other is smaller
     if (a_bounds.max_defined && b_bounds.min_defined && a_bounds.max <= b_bounds.min) {
+        if (const Call *call = b.as<Call>()) {
+            if (call->is_intrinsic(Call::likely) ||
+                call->is_intrinsic(Call::likely_if_innermost)) {
+                return call->args[0];
+            }
+        }
         return b;
     }
     if (b_bounds.max_defined && a_bounds.min_defined && b_bounds.max <= a_bounds.min) {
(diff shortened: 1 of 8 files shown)


src/Simplify_Max.cpp
```diff
@@ -69,10 +81,10 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(min(y, x), x), b) ||
              rewrite(max(min(x, c0), c1), b, c1 >= c0) ||
 
-             rewrite(max(intrin(Call::likely, x), x), a) ||
-             rewrite(max(x, intrin(Call::likely, x)), b) ||
-             rewrite(max(intrin(Call::likely_if_innermost, x), x), a) ||
-             rewrite(max(x, intrin(Call::likely_if_innermost, x)), b) ||
+             rewrite(max(intrin(Call::likely, x), x), b) ||
```
steven-johnson: Is the reordering here significant?

abadams: It's not reordering, it's changing it to return the other arg. So max(likely(x), x) now returns x instead of likely(x)

abadams:
@jn80842 for the select rule. Unfortunately there's non-local reasoning.

abadams:
Clean except for broken windows bot

PR #4776, opened 2020-03-21
Prefer calling count() and treating result as bool to find() == end()

abadams:
For std::map, map.count(foo) just inlines to (map.find(foo) == map.end()
? 0 : 1), so the assembly is identical. This is purely a stylistic
choice that I'm raising for discussion because Steven brought it up as
being clearer in one context in another PR.

Opinions? Is this clearer on average? It's certainly fewer characters.
I'm leaning towards yes.

### DIFF
--- a/src/AddAtomicMutex.cpp
+++ b/src/AddAtomicMutex.cpp
@@ -235,7 +235,7 @@ protected:
 
     void visit(const Store *op) override {
         if (in_atomic_mutex) {
-            if (store_names.find(op->name) != store_names.end()) {
+            if (store_names.count(op->name)) {
                 found = true;
             }
         }
@@ -315,7 +315,7 @@ protected:
             return IRMutator::visit(op);
         }
 
-        if (allocated_mutexes.find(finder.mutex_name) != allocated_mutexes.end()) {
+        if (allocated_mutexes.count(finder.mutex_name)) {
             // We've already allocated a mutex.
             return IRMutator::visit(op);
         }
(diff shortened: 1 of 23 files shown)


steven-johnson:
My nit is that it's better to explicitly add `> 0` or `== 0` everywhere to make it clearer that we have a bool result that's being evaluated; I know it shouldn't make any difference to codegen either way. (I won't go to war over this, but am in the habit of preferring actual bool expressions for `if`, etc conditions)

steven-johnson:
(But LGTM)

steven-johnson:
Do we want to land this? (No urgency but would be good to decide)

abadams:
I'm dithering on if I want to defend treating count as a bool or whether I agree with you. I wish std::map had a contains() method.

steven-johnson:
> I wish std::map had a contains() method.

...it does in C++20 :-)

(Side note: does clang-tidy have a mode for "enforce that all ifs use boolean conditions rather than implicit int checks"? Would be interesting to know just how frequently we already fail that.)

abadams:
We convert from pointers to bool everywhere:

```
if (const LT *op = foo.as<LT>()) {
}
```

steven-johnson:
Ouch. OK, problem solved.

steven-johnson:
Did we ever decide if this change was worthwhile?

abadams:
I don't think it's important and have no strong feelings one way or the other. I'll just close it.

PR #4778, opened 2020-03-22
Instead of just assuming bounds query results can be lifted, enforce nesting

abadams:
TL;DR: This is an improvement from #3037 / #4713 that can be tested and
checked in independently. It removes one type of magic name usage.

Allocation bounds inference (well, boxes_touched) relies on magic names
in its handling of getting the min/max out of a bounds query result. It
just lifts them out of their containing scope (see the deleted code in
Bounds.cpp below), assuming a bounds query buffer of the same name also
exists in the containing scope, and also assuming that the result of
that outer bounds query was a valid interval containing the result of
the inner bounds query. Yuck. I removed that special handling from
boxes_touched, and instead explicitly clamped the inner bounds query
result to be within the outer region required, which comes from the
outer bounds query. That gives boxes_touched a way to bound the impure
call with something.

So now if we have a misbehaving extern stage, which lies in its outer
bounds query and says it's going to ask for less than it does in the
inner bounds query, it simply doesn't get what it asks for in the inner
bounds query. It gets the intersection of the outer bounds query result
and the inner bounds query result. Too bad. This behavior is illustrated
in the new test. Formerly I'm not sure what happened, because we were
just assuming that it couldn't occur. I guess that's UB. One of the
tests in storage_folding actually broke this rule, but it didn't matter
before because it never actually accessed the input.

Storage folding needed adjusting because it incorrectly didn't even try
to fold now that the bound seems to be pure (it's a .outer_min Variable,
whereas before it was the get_min call referencing the outer bounds
query result).

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1263,20 +1263,6 @@ private:
                            op->func, op->value_index, op->image, op->param),
                 Call::make(t, op->name, {interval.max}, op->call_type,
                            op->func, op->value_index, op->image, op->param));
-
-        } else if (!const_bound &&
-                   (op->name == Call::buffer_get_min ||
-                    op->name == Call::buffer_get_max)) {
-            // Bounds query results should have perfect nesting. Their
-            // max over a loop is just the same bounds query call at
-            // an outer loop level. This requires that the query is
-            // also done at the outer loop level so that the buffer
-            // arg is still valid, which it is, so it is.
-            //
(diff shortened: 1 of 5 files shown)


steven-johnson:
testing in google now

steven-johnson:
No red flags from google testing

abadams:
linux failure is a performance flake.

PR #4779, opened 2020-03-22
Don't depend on magic names in allocation bounds inference

abadams:
This is another independent improvement carved off of #3037 / #4713

Allocation bounds inference for a Func foo depends on the magic names
foo.x.min/foo.x.max at the realization site. The only reason it needs
them is to ensure it knows that an extern stage is going to write to the
whole region required. We can just tell it that earlier instead with a
marker in the IR that says "you should treat this marker as a write to
this buffer over this region". Presto, no magic name dependence.

This is desirable because it will let us do more types of IR transformation
before allocation bounds inference.


steven-johnson:
testing in google now

steven-johnson:
(my approval is pending google testing)

steven-johnson:
No red flags from google testing

PR #4780, opened 2020-03-22
Fix a bug in autodiff caused by the implicit capture of lower_random

BachiLi:
This PR fixes a bug in autodiff when using with `random_float` or `random_int`. During `lower_random`, the implicit capture creates references to reduction variables without defining their `reduction_domain`, causing issues in autodiff. I fixed this by reintroducing the reduction domain when required. I also removed a few warnings when using random_float caused by propagating the derivative through some bit manipulation intrinsic functions.


abadams:
Wouldn't it be better to not drop the domain reference in the first place? I.e. in Function.cpp make the free_vars vector a `vector<VarOrRVar>` instead of a `vector<string>`? It looks like it would actually slightly simplify LowerRandom, which takes a vector of string and reconstructs a vector of Var

BachiLi:
> Wouldn't it be better to not drop the domain reference in the first place? I.e. in Function.cpp make the free_vars vector a `vector<VarOrRVar>` instead of a `vector<string>`? It looks like it would actually slightly simplify LowerRandom, which takes a vector of string and reconstructs a vector of Var

Good point. Made the suggested changes.

src/Function.cpp
```diff
@@ -427,8 +428,13 @@ void Function::define(const vector<string> &args, vector<Expr> values) {
 
     // Tag calls to random() with the free vars
     int tag = rand_counter++;
+    vector<VarOrRVar> free_vars;
+    free_vars.reserve(args.size());
```
abadams: Why not:
```
vector<VarOrRVar> free_vars(args.size());
```

src/Function.cpp
```diff
@@ -595,16 +601,20 @@ void Function::define_update(const vector<Expr> &_args, vector<Expr> values) {
     }
 
     // Tag calls to random() with the free vars
-    vector<string> free_vars;
+    vector<VarOrRVar> free_vars;
+    int num_free_vars = (int)pure_args.size();
+    if (check.reduction_domain.defined()) {
+        num_free_vars += (int)check.reduction_domain.domain().size();
+    }
+    free_vars.reserve(num_free_vars);
```
abadams: similarly:
```
vector<VarOrRVar> free_vars(num_free_vars);
```

src/Function.cpp
```diff
@@ -427,8 +428,13 @@ void Function::define(const vector<string> &args, vector<Expr> values) {
 
     // Tag calls to random() with the free vars
     int tag = rand_counter++;
+    vector<VarOrRVar> free_vars;
+    free_vars.reserve(args.size());
+    for (size_t i = 0; i < args.size(); i++) {
+        free_vars.emplace_back(Var(args[i]));
```
abadams: then free_vars[i] = Var(args[i])

src/Function.cpp
abadams: Oh, no default constructor on VarOrRVar?

src/Function.cpp
```diff
@@ -427,8 +428,13 @@ void Function::define(const vector<string> &args, vector<Expr> values) {
 
     // Tag calls to random() with the free vars
     int tag = rand_counter++;
+    vector<VarOrRVar> free_vars;
+    free_vars.reserve(args.size());
+    for (size_t i = 0; i < args.size(); i++) {
```
abadams: Prefer range-based for loops when the index isn't used:
```
for (const auto &arg : args) {
  free_vars.emplace_back(Var(arg));
}
```

BachiLi: I propagated the changes to a few loops in proximity. Let me know if I changed too much.

abadams:
Thanks for cleaning up more of the for loops

BachiLi:
@steven-johnson FYI the windows builtbots are having issues.

PR #4781, opened 2020-03-23
speed up prove_associativity by not substituting lets in

BachiLi:
I believe at least two of the `substitute_in_all_lets` in associativity proving are unnecessary. They are the ones before the stage where we replace self-reference function calls to a Var. Removing them makes proving significantly faster. I leave the one before the pattern matching there since it might affect the pattern matching result.

I have some nasty huge expressions generated from autodiff that are taking ~30s in `prove_associativity`, and after this change it only takes 0.02s.


BachiLi:
I seem to be wrong. Closing until I figure things out.

PR #4783, opened 2020-03-23
Calls can't be considered constant if they are impure

abadams:
Currently is_monotonic treats calls with constant args as pure. But impure calls could return a different value on each call, even if they have the same args, so they shouldn't be considered constant.

### DIFF
--- a/src/Monotonic.cpp
+++ b/src/Monotonic.cpp
@@ -517,7 +517,7 @@ void is_monotonic_test() {
     check_constant(select(y > 3, y + 23, y - 65));
 
     std::cout << "is_monotonic test passed" << std::endl;
-}  // namespace Internal
+}
 
 }  // namespace Internal
 }  // namespace Halide
--- a/src/Monotonic.cpp
+++ b/src/Monotonic.cpp
@@ -491,6 +491,9 @@ void is_monotonic_test() {
     check_unknown(x != y);
     check_unknown(x * y);
 
+    // Not constant despite having constant args, because there's a side-effect.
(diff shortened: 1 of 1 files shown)


steven-johnson:
failure is 

```
Error: CUDA: cuMemAlloc failed: CUDA_ERROR_OUT_OF_MEMORY
Makefile:30: recipe for target '/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-1000/halide-build/bin/apps/max_filter/bin/host-cuda/out.png' failed
make[2]: Leaving directory '/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-1000/halide/apps/max_filter'
make[2]: *** [/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-1000/halide-build/bin/apps/max_filter/bin/host-cuda/out.png] Aborted (core dumped)
/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-1000/halide/Makefile:1998: recipe for target 'max_filter_test_app' failed
make[1]: Leaving directory '/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-1000/halide-build'
make[1]: *** [max_filter_test_app] Error 1
```

abadams:
Failure isn't related. I think something is soaking up GPU memory on one of the linux builders. The allocation pattern is the same in this branch as master. The largest allocation is 342MB. Changing the schedule to use less GPU memory would mess up using the app as a benchmark. Probably should just swap in a less ancient GPU at some point.

steven-johnson:
In that case, LGTM, merging

PR #4785, opened 2020-03-23
Generalize uniquify_variable_names

abadams:
Currently uniquify_variable_names only works on exactly the kind of
duplicate names we tend to generate in bounds inference, and only if you
run it on the entire stmt. In other circumstances it does odd things
(e.g. attach negative integer suffixes to free variables!), and we risk
introducing bugs, as I discovered in another PR.

This PR fixes uniquify_variable_names to do the right thing in the
general case, and adds tests.

Also added some needed functionality to Scope - you can now query the
number of nested definitions of the same thing, and you can move a
Scope.

### DIFF
--- a/src/Scope.h
+++ b/src/Scope.h
@@ -64,6 +64,10 @@ public:
     bool empty() const {
         return _empty;
     }
+
+    size_t size() const {
+        return _empty ? 0 : (_rest.size() + 1);
+    }
 };
 
 template<>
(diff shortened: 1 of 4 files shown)


steven-johnson:
LGTM pending tests

steven-johnson:
```
D:\build_bot\worker\win-64-testbranch-1000\halide\src\UniquifyVariableNames.cpp(114): error C2280: 'Halide::Internal::Scope<std::string>::Scope(const Halide::Internal::Scope<std::string> &)': attempting to reference a deleted function [D:\build_bot\worker\win-64-testbranch-1000\halide-build-Release\src\Halide.vcxproj]
  d:\build_bot\worker\win-64-testbranch-1000\halide\src\Scope.h(108): note: see declaration of 'Halide::Internal::Scope<std::string>::Scope'
  d:\build_bot\worker\win-64-testbranch-1000\halide\src\Scope.h(108): note: 'Halide::Internal::Scope<std::string>::Scope(const Halide::Internal::Scope<std::string> &)': function was explicitly deleted (compiling source file D:\build_bot\worker\win-64-testbranch-1000\halide\src\UniquifyVariableNames.cpp)
```

abadams:
ok, I don't understand why the move constructor isn't triggering on windows. I guess I'll just avoid it.

steven-johnson:
```
Failure in uniquify_variable_names
Input:
let x = 3
let y = x
let x = (x + y)
let y = (x + y)
let x = (x + y)
let y = (x + y)
0
Produced:
let x = 3
let y = x
let x_1 = (x_1 + y)
let y_1 = (x_1 + y_1)
let x_2 = (x_2 + y_1)
let y_2 = (x_2 + y_2)
0
Correct output:
let x = 3
let y = x
let x_1 = (x + y)
let y_1 = (x_1 + y)
let x_2 = (x_1 + y_1)
let y_2 = (x_2 + y_1)
0
```

abadams:
Issue was fixed. ptaal

PR #4786, opened 2020-03-23
Use graph_equal instead of equal in associativity proving

BachiLi:
The `equal` in associativity proving is operating on non-CSE'd expressions, making it exponentially slow. This PR changes it to `graph_equal`. I observed significant speedup in some huge expressions generated by autodiff when calling `prove_associativity`(30s -> 0.02s).


BachiLi:
Urgh wrong branch. One moment.

Issue #4675, opened 2020-03-04
compute_with and loops with mismatching loop types


vksnk:
Currently you can schedule a function to be fused with other function only if _for types_ match for every fused variable. This seems to be a bit too strict and requires you to duplicate schedule for outer variables. 

For example:

```
        Var x("x"), y("y");
        Func f("f"), g("g");

        f(x, y) = 1;
        g(x, y) = 2;

        f.parallel(y);
        g.compute_with(f, x);
       
        Pipeline p({f, g});
        p.compile_jit();

```

will fail with:

```
Invalid compute_with: dims 1 of f.s0(y) and g.s0(y) do not match.
```

It fails because scheduler will not only check that vars are matching, but also that their _for types_ do (and in this case only one of them is marked as Parallel). Changing it to ` g.parallel(y).compute_with(x) ` will make it work. 

I think it should be safe to allow at least some combinations of for types (or maybe all?) [here](https://github.com/halide/Halide/blob/master/src/ScheduleFunctions.cpp#L2030). Any objections/potential pitfalls?

At the very least, the error message needs to be improved because right now it would just say that vars don't match (see message above for example), which doesn't really help, because they do match and problem is with something else.

I can do either way, just wanted to seek opinions on what is the best option before starting to work on it.

PR #4788, opened 2020-03-23
More specific compute_with error message

vksnk:
See #4675.

### DIFF
--- a/src/ScheduleFunctions.cpp
+++ b/src/ScheduleFunctions.cpp
@@ -2031,17 +2031,20 @@ void validate_fused_group_schedule_helper(const string &fn,
                                                         << p.stage_1 << "(" << d1.var << ") and " << p.func_2
                                                         << ".s" << p.stage_2 << "(" << d2.var << ") do not match.\n";
             user_assert(d1.for_type == d2.for_type) << "Invalid compute_with: for types of dim "
-                                                    << i << " of " << p.func_1 << ".s"
-                                                    << p.stage_1 << "(" << d1.var << ") and " << p.func_2
-                                                    << ".s" << p.stage_2 << "(" << d2.var << ") do not match.\n";
+                                                    << i << " of " << p.func_1 << ".s" << p.stage_1 << "("
+                                                    << d1.var << " is " << d1.for_type << ") and " << p.func_2
+                                                    << ".s" << p.stage_2 << "(" << d2.var << " is " << d2.for_type
+                                                    << ") do not match.\n";
             user_assert(d1.device_api == d2.device_api) << "Invalid compute_with: device APIs of dim "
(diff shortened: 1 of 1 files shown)


src/ScheduleFunctions.cpp
```diff
@@ -2026,15 +2026,22 @@ void validate_fused_group_schedule_helper(const string &fn,
...
-                         (d1.device_api == d2.device_api) &&
-                         (d1.dim_type == d2.dim_type);
-            if (!equal) {
-                user_error << "Invalid compute_with: dims " << i << " of " << p.func_1 << ".s"
-                           << p.stage_1 << "(" << dims_1[start_fuse_1 + i].var << ") and " << p.func_2
-                           << ".s" << p.stage_2 << "(" << dims_2[start_fuse_2 + i].var << ") do not match.\n";
-            }
+            user_assert(var_name_match(d1.var, d2.var)) << "Invalid compute_with: names of dim "
+                                                        << i << " of " << p.func_1 << ".s"
+                                                        << p.stage_1 << "(" << dims_1[start_fuse_1 + i].var << ") and " << p.func_2
```
abadams: The repetition is unfortunate but I don't see a simple way to remove it. You don't want to eagerly produce the string in the non-error case.

But there's something weird in the existing code that looks like it could be shorter: dims_1[start_fuse_1 + i].var could be d1.var

vksnk: Good point, fixed.

src/ScheduleFunctions.cpp
```diff
@@ -2026,15 +2026,22 @@ void validate_fused_group_schedule_helper(const string &fn,
...
-            if (!equal) {
-                user_error << "Invalid compute_with: dims " << i << " of " << p.func_1 << ".s"
-                           << p.stage_1 << "(" << dims_1[start_fuse_1 + i].var << ") and " << p.func_2
-                           << ".s" << p.stage_2 << "(" << dims_2[start_fuse_2 + i].var << ") do not match.\n";
-            }
+            user_assert(var_name_match(d1.var, d2.var)) << "Invalid compute_with: names of dim "
+                                                        << i << " of " << p.func_1 << ".s"
+                                                        << p.stage_1 << "(" << d1.var << ") and " << p.func_2
+                                                        << ".s" << p.stage_2 << "(" << d2.var << ") do not match.\n";
+            user_assert(d1.for_type == d2.for_type) << "Invalid compute_with: for types of dim "
```
abadams: Suggest adding a final sentence: d1.var << " is " << d1.for_type << " and " << d2.var << " is " << d2.for_type.

Similarly below with device_api and dim_type

vksnk: Done.

PR #4790, opened 2020-03-24
constructor for 'Halide::Internal::StubInput' must explicitly initial…

dpalermo:
…ize const members

  - Seen when compiling Halide apps with certain versions of clang++
  - Seems like a valid complaint (just not reported by all compilers)

### DIFF
--- a/src/Generator.h
+++ b/src/Generator.h
@@ -1377,13 +1377,13 @@ public:
     // *not* explicit.
     template<typename T2>
     StubInput(const StubInputBuffer<T2> &b)
-        : kind_(IOKind::Buffer), parameter_(b.parameter_) {
+        : kind_(IOKind::Buffer), parameter_(b.parameter_), func_(), expr_() {
     }
     StubInput(const Func &f)
-        : kind_(IOKind::Function), parameter_(), func_(f) {
+        : kind_(IOKind::Function), parameter_(), func_(f), expr_() {
     }
     StubInput(const Expr &e)
-        : kind_(IOKind::Scalar), parameter_(), expr_(e) {
+        : kind_(IOKind::Scalar), parameter_(), func_(), expr_(e) {
     }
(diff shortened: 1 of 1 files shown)


dpalermo:
The full error that is reported looks like:

```
In file included from Halide/tools/GenGen.cpp:1:
Halide/include/Halide.h:24116:5: error: constructor for 'Halide::Internal::StubInput' must explicitly initialize the const member 'expr_'
  StubInput(const Func &f)
  ^
Halide/include/Halide.h:24108:16: note: 'expr_' declared here
  const Expr expr_;
  ^
1 error generated.
```

dpalermo:
This was caught by one of our test systems that used an older version of clang++ in it's environment for the x86 host compiler....to run x86 + Hexagon offload simulator based tests (apps/hexagon_benchmarks).  The clang++ version that flagged the error is:

$ clang++ --version
Ubuntu clang version 3.7.1-svn253742-1~exp1 (branches/release_37) (based on LLVM 3.7.1)
Target: x86_64-pc-linux-gnu
Thread model: posix

Interestingly enough, though, newer versions of clang++ don't produce this error.  Both of the following compiled the same test (hexagon_benchmarks) without this error:

$ clang++ --version
clang version 7.0.0-3~ubuntu0.18.04.1 (tags/RELEASE_700/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

$ clang++ --version
clang version 7.0.1 (tags/RELEASE_701/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /pkg/qct/software/llvm/build_tools/clang+llvm-7.0.1-x86_64-linux-gnu-ubuntu-14.04/bin

The hexagon_benchmarks test was run via make.

We are currently using make to build Halide.

PR #4798, opened 2020-03-26
Fix Func name

mroavi:
The `Func` name used in `tutorial/lesson_05_scheduling_1.cpp` in the row-major example is `gradient` and not `gradient_row_major`.


steven-johnson:
Thanks for the fix!

PR #4801, opened 2020-03-30
Fix #4297 by preventing Ifs being pushed between GPUBlocks

### LINKED ISSUE #4297: build_loop_nests() lifts if statements to between GPUBlocks
The following code fails with an internal assertion in PartitionLoops:
```
#include "Halide.h"
using namespace Halide;

    Var x, y, z;
    Func f;
    f(x, y, z) = 0;
    Var yo, yi;
    f.split(y, yo, yi, 32, TailStrategy::GuardWithIf)
     .reorder(x, z, yi, yo)
     .gpu_blocks(yo)
     .gpu_blocks(yi)
     .gpu_blocks(z)
     .gpu_threads(x);

    Target target = get_jit_target_from_environment();
    Buffer<int> imf = f.realize(100, 100, 100, target);
    return 0;
}
```
The output is
```
Internal Error at ... Halide/src/PartitionLoops.cpp:874
Condition failed: op->else_case.defined():
PartitionLoops should only introduce if statements with an else branch
```

This is because the build_loop_nest() function in the schedule function pass tries to lift the if statement through the following code:
```
    for (int i = (int)nest.size() - n_predicates; i < (int)nest.size(); i++) {
        // Only push up IfThenElse.
        internal_assert(nest[i].value.defined());
        internal_assert(nest[i].type == Container::If);

        // Cannot lift out the 'if' if it contains call to non-pure function
        if (contains_impure_call(nest[i].value)) {
            continue;
        }

        for (int j = i - 1; j >= 0; j--) {
            // Try to push it up by one.
            internal_assert(nest[j + 1].value.defined());

            if (!expr_uses_var(nest[j + 1].value, nest[j].name)) {
                std::swap(nest[j + 1], nest[j]);
            } else {
                break;
            }
        }
    }
```

It then constructs the following loop nest:
```
  for (f.s0.__outermost, f.s0.__outermost.loop_min, f.s0.__outermost.loop_extent) {
   gpu_block<Default_GPU> (f.s0.x.xi, f.s0.x.xi.loop_min, f.s0.x.xi.loop_extent) {
    gpu_block<Default_GPU> (f.s0.y.yo, f.s0.y.yo.loop_min, f.s0.y.yo.loop_extent) {
     let f.s0.y.yi.base = ((f.s0.y.yo*split_size) + f.s0.y.loop_min)
     gpu_block<Default_GPU> (f.s0.y.yi, f.s0.y.yi.loop_min, f.s0.y.yi.loop_extent) {
      let f.s0.y.rebased = ((f.s0.y.yo*split_size) + f.s0.y.yi)
      if ((uint1)likely((f.s0.y.rebased < f.s0.y.loop_extent))) {
       let f.s0.y = (f.s0.y.yi.base + f.s0.y.yi)
       gpu_block<Default_GPU> (f.s0.z, f.s0.z.loop_min, f.s0.z.loop_extent) {
        gpu_thread<Default_GPU> (f.s0.x.xo, f.s0.x.xo.loop_min, f.s0.x.xo.loop_extent) {
         let f.s0.x.rebased = ((f.s0.x.xo*split_size) + f.s0.x.xi)
         if ((uint1)likely((f.s0.x.rebased < f.s0.x.loop_extent))) {
          let f.s0.x.xi.base = ((f.s0.x.xo*split_size) + f.s0.x.loop_min)
          let f.s0.x = (f.s0.x.xi.base + f.s0.x.xi)
          f((f.s0.x.rebased + f.s0.x.loop_min), (f.s0.y.rebased + f.s0.y.loop_min), f.s0.z) = 0
         }
        }
       }
      }
     }
    }
   }
```
Notice there is an if statement between gpu blocks.

This triggers an assertion in the partition loop pass when it's trying to renormalize the if statements generated by the loop partition pass:
```
    Stmt visit(const IfThenElse *op) override {
        if (!in_gpu_loop || in_thread_loop) {
            return IRMutator::visit(op);
        }

        internal_assert(op->else_case.defined())
            << "PartitionLoops should only introduce if statements with an else branch\n";
```

Note that it is okay (though not great) to generate an if statement between GPUBlock and GPUThread because the `fuse_gpu_thread_loops` pass will create a new GPUThread loop outside of the if statement.

My proposed solution is to add a tag for For loops in the `Container` structure in `ScheduleFunction.cpp` and avoid lifting If statements outside of a GPUBlock for loop, if you can't just lift it all the way out to a CPU loop.

BachiLi:
The proper refactor approach by doing this in a separate pass suggested by @alexreinking requires complex logic with regard to bound variables (see discussion in #4800). This is a quick patch for solving the issue directly without changing the lowering code too much.


test/correctness/CMakeLists.txt
```diff
@@ -169,6 +169,7 @@ tests(GROUPS correctness travis
         inverse.cpp
         isnan.cpp
         issue_3926.cpp
+        issue_4297.cpp
```
steven-johnson: Nit: can we avoid naming tests this way? It goes against precedent (previous line notwithstanding) and is jarring to read.

BachiLi: Yes. Fixed!

BachiLi:
Ping. Let me know what you think about this when you have time.

steven-johnson:
LGTM. abadams@ should probably also approve.

abadams:
It feels a bit fishy for this to be handled here because it means every later pass needs to maintain this invariant too. For the first half of lowering GPU loops are fairly unstructured and treated as if you can have code just about anywhere, and then in fuse_gpu_loops they get normalized to something stricter. Apparently this doesn't include fixing up if statements in between blocks loops, but maybe it should. What would this look like as an additional mutator inside FuseGPUThreadLoops.cpp instead?

BachiLi:
Seems like this passes some GPU tests. PTAL.

PR #4802, opened 2020-03-30
Gradient autoscheduler improvements for GPU scheduling

BachiLi:
1) Fixed a bug in the condition for assigning GPU blocks/threads for reduction variables.
2) Originally the autoscheduler did not assign GPU threads if it couldn't find any loop larger than a warp size. Now it assigns GPU threads to the loop with the largest estimated extent.
3) Split reduction GPU blocks when they are larger than hardware restrictions.

This PR depends on #4801.


apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -154,6 +154,59 @@ void parallelize_vars_and_rvars_gpu(
         serial_rvars = rvars;
     }
 
+    if (gpu_threads.empty() && r_gpu_threads.empty()) {
+        // If we can't find any loop large enough for a GPU thread,
```
steven-johnson: Nit: this comment seems paradoxical: if none of the loops are large enough for a gpu thread, then presumably even the largest remaining is still not large enough. Clarify?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -154,6 +154,59 @@ void parallelize_vars_and_rvars_gpu(
...
+        }
+        if (largest_loop_id != -1) {
+            if (largest_loop_id < (int)vars.size()) {
+                // The largest loop is a pure variable
+                const Var &v = vars[largest_loop_id];
+                Var inner;
+                func_or_stage.split(v,
+                                    v,
+                                    inner,
+                                    32,  // warp size
```
steven-johnson: How is 32 chosen? Does it deserve to be a named constant?

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
```diff
@@ -179,6 +232,18 @@ void parallelize_vars_and_rvars_gpu(
...
     }
+    // CUDA places rather strict restriction on the second dimension of the GPU blocks (usually 65536),
+    // so we want to split it if it is too large
+    int rdomain_size = 1;
+    for (int b : rvar_bounds) {
+        rdomain_size *= b;
+    }
+    std::string fused_rvar2;
+    if (rdomain_size >= 65536) {
+        RVar r;
```
steven-johnson: Suggestion: give basename like "cuda_gpu_block_split" or something, just to make future debugging easier.

src/ScheduleFunctions.cpp
```diff
@@ -45,9 +45,11 @@ struct Container {
     int dim_idx;
     string name;
     Expr value;
+    // Only meaningful if type==For.
+    ForType for_type;
```
steven-johnson: This looks like it could be const.

test/correctness/CMakeLists.txt
```diff
@@ -169,6 +169,7 @@ tests(GROUPS correctness travis
         inverse.cpp
         isnan.cpp
         issue_3926.cpp
+        issue_4297.cpp
```
steven-johnson: see several previous comments about this filename

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Fixed!

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Named!

apps/gradient_autoscheduler/GradientAutoscheduler.cpp
BachiLi: Given!

src/ScheduleFunctions.cpp
abadams: I think const members might be a bad idea - you can't move from them, which means the whole class doesn't get a default move constructor.

BachiLi: Making this (or other members) const requires changing large amount of code. In particular there are several std::swap() calls with Containers.

test/correctness/CMakeLists.txt
BachiLi: Fixed!

Issue #4804, opened 2020-03-30
python bilateral_grid.py example does not run on GPU


Infinoid:
The bilateral_grid.py example does JIT compilation, but it does not obey the HL_JIT_TARGET environment variable.

When determining whether the target is a GPU, it should call hl.get_jit_target_from_environment(), not hl.get_target_from_environment().

The algorithm defines some Vars but overwrites them with Exprs later on, causing errors in the GPU schedule.

Issue #3082, opened 2018-06-25
Issue with GPU and python binding on apps/bilateral_grid.py


cheng-chi:
Hi! I have just downloaded Halide and trying to verify it works. With the Halide and python binding both build against LLVM-6.0 without error, and majority of CPU examples working, when I tried to test GPU by setting environment variables:
```
export HL_TARGET=cuda
export HL_JIT_TARGET=cuda
```
`python3 bilateral_grid.py` returned following error:
```
Compiling for GPU.
Traceback (most recent call last):
  File "bilateral_grid.py", line 162, in <module>
    main()
  File "bilateral_grid.py", line 148, in main
    bilateral_grid = get_bilateral_grid(input, r_sigma, s_sigma)
  File "bilateral_grid.py", line 69, in get_bilateral_grid
    histogram.compute_root().reorder(c, z, x, y).gpu_tile(x, y, 8, 8);
TypeError: gpu_tile(): incompatible function arguments. The following argument types are supported:
    1. (self: halide.Func, x: halide.VarOrRVar, bx: halide.VarOrRVar, tx: halide.VarOrRVar, x_size: halide.Expr, tail: halide.TailStrategy=TailStrategy.Auto, device_api: halide.DeviceAPI=DeviceAPI.Default_GPU) -> halide.Func
    2. (self: halide.Func, x: halide.VarOrRVar, tx: halide.VarOrRVar, x_size: halide.Expr, tail: halide.TailStrategy=TailStrategy.Auto, device_api: halide.DeviceAPI=DeviceAPI.Default_GPU) -> halide.Func
    3. (self: halide.Func, x: halide.VarOrRVar, y: halide.VarOrRVar, bx: halide.VarOrRVar, by: halide.VarOrRVar, tx: halide.VarOrRVar, ty: halide.VarOrRVar, x_size: halide.Expr, y_size: halide.Expr, tail: halide.TailStrategy=TailStrategy.Auto, device_api: halide.DeviceAPI=DeviceAPI.Default_GPU) -> halide.Func
    4. (self: halide.Func, x: halide.VarOrRVar, y: halide.VarOrRVar, tx: halide.VarOrRVar, ty: halide.VarOrRVar, x_size: halide.Expr, y_size: halide.Expr, tail: halide.TailStrategy=TailStrategy.Auto, device_api: halide.DeviceAPI=DeviceAPI.Default_GPU) -> halide.Func
[...]

PR #4805, opened 2020-03-30
GPU fixes for python_bindings/apps/bilateral_grid.py

Infinoid:
Issues #4804, #3082.

### DIFF
--- a/python_bindings/apps/bilateral_grid.py
+++ b/python_bindings/apps/bilateral_grid.py
@@ -9,14 +9,11 @@ import numpy as np
 import imageio
 import os.path
 
-def get_bilateral_grid(input, r_sigma, s_sigma):
+def get_bilateral_grid(input, r_sigma, s_sigma, aot=False):
     x = hl.Var('x')
     y = hl.Var('y')
     z = hl.Var('z')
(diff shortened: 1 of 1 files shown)


Infinoid:
The script has one additional problem, though I don't know whether I caused it or not.  It looks like the pixels are output in the wrong order.

Here is the output when run on a CPU target:
bilateral_grid


Here is the output when run on a GPU target:
bilateral_grid-gpu

steven-johnson:
Interesting -- does the C++ version of apps/bilateral_grid have the same issue for you?

Infinoid:
No.  The C++ bilateral_grid app output is correct for both CPU and GPU builds.

steven-johnson:
I think it's just missing copy_to_host, try adding:

    # do the actual computation
    bilateral_grid.realize(output_image)
    **output_image.copy_to_host()**

Infinoid:
Yep, that worked.  steven-johnson++

Infinoid:
With that, the script works properly for me in both CPU and GPU mode.

It still breaks if you set `generate` to `True`.  But that's a problem for another PR.

PR #4812, opened 2020-03-31
AlignLoads.h includes stuff it doesn't need

abadams:
Inspired by #4811, I played around with include-what-you-use. Mostly it made a mess and I ultimately gave up, but one particularly egregious header was AlignLoads.h, which includes some stuff it doesn't use at all.


abadams:
Subsumed by #4813

PR #4816, opened 2020-04-02
Minor optimization in SmallStack::push

steven-johnson:
This method is instantiated a lot, so a couple of minor tweaks:
- skip a branch by unconditionally setting _empty=false in all cases
- use std::move for the push_back, to avoid a copy in case the type of object is movable

### DIFF
--- a/src/Scope.h
+++ b/src/Scope.h
@@ -41,12 +41,11 @@ public:
     }
 
     void push(const T &t) {
-        if (_empty) {
-            _empty = false;
-        } else {
-            _rest.push_back(_top);
+        if (!_empty) {
+            _rest.push_back(std::move(_top));
         }
         _top = t;
+        _empty = false;
     }
 
     T top() const {


abadams:
The large piece of low-hanging Scope fruit is fixing up our idiom:

`if (scope.contains(foo)) { ... scope.get(foo) ... }`

which does a redundant map lookup. I guess we need a scope.find(foo) which returns... a pointer? E.g:

`if (auto *bar = scope.find(foo)) { ... *bar ... }`

steven-johnson:
> which does a redundant map lookup. I guess we need a scope.find(foo) which returns... a pointer? E.g:
> 
> `if (auto *bar = scope.find(foo)) { ... *bar ... }`

...or have `get` return a `std::optional` or `pair<T, bool>`? (`optional` is C++17 but might not be hard to roll enough of our own to make it work. Not sure about code efficiency of implementaton.)

(On that note: is it time to think about moving to C++17 yet? Even google has moved to allowing it internally...)

abadams:
Anything that can cast to bool would enable using it in an if like that. I believe that's the case for std::optional.

I sort of like the pointer approach because we use it elsewhere to mean a reference to a thing if it exists, or null if it doesn't (e.g. Expr::as), so it minimizes the total number of idioms to reuse it.

Issue #4333, opened 2019-10-22
GPU memory model is poorly documented. Copy_to_device does nothing unless you set_host_dirty.


SuTanTank:
I was following the tutorials and since my Windows PC doesn't has `libJpeg` or `libPng`, I decided to use OpenCV for image I/O. Then I learnt that in Halide, RGB buffer is default to be planar, and I checked out the `test_interleaved.cpp`, and write two small tools to convert between an interleaved buffer and a planar buffer. Which works fine in CPU version, like `lesson_07`. 

However, when it came to `lesson_12`, the GPU part, the result is all 0 in the GPU version pipeline. 

More specifically, I tried two strategies: 
1. add two **Func** to convert a interleaved input the planar and a Func to convert back to interleaved at the beginning and the end of the original pipeline. 
2. use two additional **pipelines**. The first one converts an interleaved buffer to a planar buffer first, with realize() called, then feed the planar buffer to the untouched original pipeline. Then add another pipeline to convert  the planar result to an interleaved buffer and then create the corresponding cv::Mat form it. 

These two strategies works fines on CPU. But on GPU, the first one (use two more Func) is too slow, so I tried the second strategy. However, the result is all 0.

I have tried `copy_to_device()` `copy_to_host` before and after the realize(), but it seems make no difference. 

The function I use to convert buffer in below
```
template <class T>
Halide::Buffer<T> interleaved2planar(Halide::Buffer<T> in) {
    ImageParam src(in.type(), 3);
    Halide::Func plane("plane");
    Halide::Var x, y, c;

    plane(x, y, c) = src(x, y, c);

    src.dim(0).set_stride(3).dim(2).set_stride(1).set_bounds(0, 3);
    plane.output_buffer().dim(0).set_stride(1).dim(2).set_extent(3);

    plane.reorder(c, x, y).unroll(c);
    // plane.vectorize(x, 64);

    src.set(in);
    plane.compile_jit();

    auto out_buf = Halide::Buffer<T>(in.width(), in.height(), 3);
    plane.realize(out_buf);
    return out_buf;
}
```
and this is how I use it:
```
auto mat = cv::imread(data_root + "images/rgb.png");
Buffer<uint8_t> interleaved_in = Buffer<uint8_t>::make_interleaved(mat.data, mat.cols, mat.rows, mat.channels());
Buffer<uint8_t> input1 = interleaved2planar(interleaved_in);
```

abadams:
[...]

PR #4820, opened 2020-04-02
Improve set_host_dirty error message

BachiLi:
Currently, if someone forgot to `copy_to_host` before accessing the Buffer in the host, this assertion would be triggered:
```
assert((!v || !device_dirty()) && "Cannot set host dirty when device is already dirty.");
```

This message is not very friendly to Halide newbies (like me). I propose changing it to
```
assert((!v || !device_dirty()) && "Cannot set host dirty when device is already dirty. Call copy_to_host before accessing the buffer from host.");
```

Ideally we also want to print out the name of the buffer, but doing it inside an assertion with zero overhead might be difficult.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1570,7 +1570,7 @@ public:
     // access. Must be inlined so it can be hoisted out of loops.
     HALIDE_ALWAYS_INLINE
     void set_host_dirty(bool v = true) {
-        assert((!v || !device_dirty()) && "Cannot set host dirty when device is already dirty.");
+        assert((!v || !device_dirty()) && "Cannot set host dirty when device is already dirty. Call copy_to_host before accessing the buffer from host.");
         buf.set_host_dirty(v);
     }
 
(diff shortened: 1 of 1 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -1570,7 +1570,7 @@ class Buffer {
     // access. Must be inlined so it can be hoisted out of loops.
     HALIDE_ALWAYS_INLINE
     void set_host_dirty(bool v = true) {
-        assert((!v || !device_dirty()) && "Cannot set host dirty when device is already dirty.");
+        assert((!v || !device_dirty()) && "Cannot set host dirty when device is already dirty. Call copy_to_host before accessing the buffer from host.");
```
steven-johnson, line 1573: nit: s/copy_to_host/copy_to_host()/

abadams:
Our GPU memory model is confusing and poorly documented ( #4333 )

src/runtime/HalideBuffer.h
BachiLi: Fixed!

PR #4821, opened 2020-04-02
Add Expr.cpp, move nontrivial function bodies into it

steven-johnson:
Expr.h is included by ~everything in Halide, so we want to minimize its complexity. There are several nontrivial function bodies in Expr.h that are likely too large to inline anyway, so add Expr.cpp and move them there. (Also move the Range ctor there since that's where it belonged anyway.)

Linux build shows a slight decrease in libHalide size (110872632 -> 110830336). Didn't try to measure performance changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /apps/HelloMatlab/blurred.png
 /apps/HelloMatlab/iir_blur.mex
 bin/*
+build/*
 python_bindings/bin/*
 build-64/*
 build-ios/*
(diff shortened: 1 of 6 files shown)


abadams:
IntImm::make is called *a lot*, so it's definitely worth measuring performance. I doubt it was inlined though, so I don't expect to see an effect.

steven-johnson:
> IntImm::make is called _a lot_, so it's definitely worth measuring performance.

What are the best checks for halide compiler performance? (Maybe running fuzz_simplify a bunch?)

abadams:
For a quick check of a realistic workload I usually do:
```
make distrib -j32
cd apps/local_laplacian
make bin/host/local_laplacian.generator
time bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host auto_schedule=false
```

PR #4823, opened 2020-04-02
Add ExternFuncArgument.h

steven-johnson:
Most that need Function don't need ExternFuncArgument,  but putting both in Function.h adds extra includes (eg Buffer.h) in places that it isn't needed. Split it out into a separate file and forward-declared in all headers (except Param.h, which will be addressed in the future).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -667,6 +667,7 @@ HEADER_FILES = \
   Expr.h \
   ExprUsesVar.h \
   Extern.h \
+  ExternFuncArgument.h \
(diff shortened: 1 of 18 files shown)


steven-johnson:
> I think ExternFuncArgument.h needs to go into Halide.h, so the header lists in the Makefile and CMakeLists.txt need adjusting

...that's weird: (1) I definitely had done that in some version of this PR, and (2) how are the tests compiling without it?

steven-johnson:
(Fixed)

PR #4824, opened 2020-04-02
Remove unnecessary #includes from Parameter.h

steven-johnson:
(Note that this required adding some in other places that were relying on transitive includes)


src/Parameter.h
```diff
@@ -4,26 +4,33 @@
...
+struct Type;
 
 namespace Internal {
 
 struct ParameterContents;
 
 /** A reference-counted handle to a parameter to a halide
  * pipeline. May be a scalar parameter or a buffer */
-class Parameter {
+class Parameter final {
```
abadams, line 28: Why final?

steven-johnson: Whoops! Scalpel left in patient. (There was a dangling virtual dtor left over from long ago -- I wanted to be sure no one was still overriding it. Will remove.)

PR #4830, opened 2020-04-04
Move most Interval method bodies to the .cpp file

steven-johnson:
Saves ~12k from libHalide.so on x86-64-linux; timing on local_laplacian.generator doesn't show any performance impact.

### DIFF
--- a/src/Interval.cpp
+++ b/src/Interval.cpp
@@ -65,15 +65,6 @@ Expr make_min_helper(const Expr &a, const Expr &b) {
 
 }  // namespace
 
-Interval::Interval()
-    : min(neg_inf()), max(pos_inf()) {
-}
-
-Interval::Interval(const Expr &min, const Expr &max)
-    : min(min), max(max) {
-    internal_assert(min.defined() && max.defined());
-}
-
 Interval Interval::everything() {
     return Interval(neg_inf(), pos_inf());
 }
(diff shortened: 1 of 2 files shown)


src/Interval.cpp
```diff
@@ -65,6 +65,63 @@ Expr make_min_helper(const Expr &a, const Expr &b) {
...
+
+bool Interval::is_bounded() const {
+    return has_upper_bound() && has_lower_bound();
+}
+
+bool Interval::same_as(const Interval &other) const {
+    return min.same_as(other.min) && max.same_as(other.max);
+}
+
+bool Interval::operator==(const Interval &other) const {
```
abadams, line 121: While you're here: this operator== method looks fishy. It's the same as same_as. Is anyone using it? Maybe it can be deleted.

abadams:
I'm surprised by this one. Most of these methods are incredibly cheap, e.g. two pointer comparisons. I had assumed they were being inlined. Can we hold off on merging this one until I can investigate?

abadams:
Slows down allocation bounds inference on local laplacian by 5% or so. To claw it back I just had to inline both constructors. That in fact gave a minor speed boost (1%) over master for that stage.

Turns out operator== exists because the built-in autoscheduler wants to test to map<string, Interval> for equality in order to cache computations. I'll add a note.

steven-johnson:
No rush.

steven-johnson:
Huh, clearly I didn't time carefully enough, my local-laplacian run seemed to be within typical noise variation.

abadams:
I just put tic/toc around allocation bounds inference, which is the stage that uses intervals the most. The 5%/1% refers to that stage alone. It is indeed in the noise for overall compilation.

PR #4832, opened 2020-04-04
Reduce number of copies in Scope::push

abadams:
pushing something onto a scope currently copies it. If we accept rvals appropriately then we can avoid those copies when pushing an rval (e.g. scope.push(name, expr_a + expr_b))

This also makes it possible to use scopes with non-copyable types as values. 

Speeds up lowering about half a percent on local laplacian. Increases the size of libHalide by two pages (8k), so I guess the speed-up is due to inlining.

### DIFF
--- a/src/Scope.h
+++ b/src/Scope.h
@@ -35,16 +35,16 @@ public:
             _empty = true;
             _top = T();
         } else {
-            _top = _rest.back();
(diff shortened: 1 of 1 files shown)


steven-johnson:
LGTM pending green

PR #4834, opened 2020-04-06
Pimpl-ify most of ParamMap

steven-johnson:


steven-johnson:
Note that this also removed the ParamMap support from the Python bindings... except they were already broken (it wasn't possible to construct a *useful* ParamMap via Python). Rather than try to fix it here, removed entirely for now.

abadams:
Slows down performance_realize_overhead test results by ~7% in the low-argument-count case (easier to measure if you change all the benchmark calls to benchmark(1000, 1000, ...)

Not entirely sure what part of the PR causes the regression. There's a new heap allocation in the default-construction case, but I don't see any default constructor in the code path that got slower.

steven-johnson:
> Slows down performance_realize_overhead test results by ~7% in the low-argument-count case

Wowzers. That's unexpected. Stay tuned.

steven-johnson:
Hm, only difference that seems likely relevant is that `empty_map` can be defined *locally* in the test file in the old code (since it's inlined in the header), whereas it is defined *in libHalide* in the new code (so it has to call externally). Seems surprising that it would be that costly.

abadams:
The overhead of realize is in the tens of nanoseconds (and we want to keep it there so that JIT is still useful for tiny pipelines like an 8x8 dct), so an extra function call is significant.

steven-johnson:
I'm going to close this for now.

steven-johnson:
(I'll

PR #4837, opened 2020-04-07
Require that rewrite rules are in canonical form

abadams:
For every commutative binary op, we know that the weaker IR node is on the right, because the simplifier does that before applying any rules using should_commute. We also normalize >= and > to <= and <. 

We should never make a right-hand-side that doesn't obey these constraints. It just makes more work for the simplifier.

This PR deletes 8 simplifier rules that could never have matched. The checks are done at compile-time using static asserts.

@jn80842

### DIFF
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -443,7 +443,7 @@ struct Wild {
     constexpr static uint32_t binds = 1 << (i + 16);
 
     constexpr static IRNodeType min_node_type = IRNodeType::IntImm;
-    constexpr static IRNodeType max_node_type = IRNodeType::Shuffle;
+    constexpr static IRNodeType max_node_type = StrongestExprNodeType;
     constexpr static bool canonical = true;
 
     template<uint32_t bound>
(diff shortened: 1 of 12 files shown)


src/IRMatch.h
```diff
@@ -1755,6 +1861,10 @@ struct Overflow {
 
     constexpr static uint32_t binds = 0;
 
+    constexpr static IRNodeType min_node_type = IRNodeType::Call;
+    constexpr static IRNodeType max_node_type = IRNodeType::Call;
+    constexpr static bool canonical = true;
```
shoaibkamil, line 1866: Add a comment as to why this has min/max type of `Call`?

src/IRMatch.h
```diff
@@ -1823,6 +1937,10 @@ struct CanProve {
 
     constexpr static uint32_t binds = bindings<A>::mask;
 
+    constexpr static IRNodeType min_node_type = IRNodeType::UIntImm;
+    constexpr static IRNodeType max_node_type = IRNodeType::UIntImm;
+    constexpr static bool canonical = true;
+
```
shoaibkamil, line 1943: Add a comment why this has min/max nodetype of `UIntImm`?

src/IRMatch.h
```diff
@@ -1854,6 +1972,10 @@ struct IsFloat {
...
```
shoaibkamil, line 1978: Ditto for short comment explaining

shoaibkamil:
LGTM.  There's a few places where the min/max IR node type isn't super-obvious, so a one-line comment explaining would be good.  The only other suggestion I have is to add an explicit comment in `Expr.h` that changing the order of the `IRNodeType` enum will require (potentially) changing the matching min/max nodetypes in the matching machinery.

abadams:
Failure unrelated.

PR #4838, opened 2020-04-08
Remove recursion in trim_scope_pop()

steven-johnson:
The current implementation recurses to let_bounds.size(), but a simple loop has the same result.


steven-johnson:
(trim_scope_push could probably benefit as well)

PR #4839, opened 2020-04-09
Avoid adding duplicate children in trim_scope_push

steven-johnson:
The current implementation doesn't try to avoid dupes, which can lead to size explosion in some cases (e.g. a case where ~200 unique children added ~16000 entries to the Scope).

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -2103,19 +2103,41 @@ private:
         }
     };
 
-    void trim_scope_push(const string &name, const Interval &bound, vector<LetBound> &let_bounds) {
+    // Build an ordered list of the unique children of 'name', such that every
+    // parent in the list comes before all its children, but without duplicates.
+    void unique_ordered_children(const string &name, set<string> &unique,
+                                 vector<string> &ordered) {
+        auto result = unique.insert(name);
+        if (result.second) {
+            ordered.push_back(name);
+        }
+        for (const auto &v : children[get_var_instance(name)]) {
+            unique_ordered_children(v.var, unique, ordered);
+        }
+    }
+
+    void trim_scope_push(const string &name, const Interval &bound,
+                         vector<LetBound> &let_bounds) {
         scope.push(name, bound);
 
-        for (const auto &v : children[get_var_instance(name)]) {
+        vector<string> ordered;
+        {
+            set<string> unique;
+            for (const auto &v : children[get_var_instance(name)]) {
+                unique_ordered_children(v.var, unique, ordered);
+            }
+        }
+        for (const string &v : ordered) {
             string max_name = unique_name('t');
             string min_name = unique_name('t');
 
-            let_bounds.insert(let_bounds.begin(), LetBound(v.var, min_name, max_name));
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -2103,19 +2103,38 @@ class BoxesTouched : public IRGraphVisitor {
         }
     };
 
+    // Build an ordered list of the unique children of 'name', such that every
```
abadams, line 2106: This looks like it still does a full depth-first traversal graph, visiting nodes multiple times even though we know their children will never be pushed onto 'ordered'. I think if (result.second) is false you can just return, with no need to visit the children again. Or how about the following non-recursive alternative. I think the only different is that the children get visited in reverse order, but the children are still visited before the parents.

```

    void trim_scope_push(const string &name, const Interval &bound, vector<LetBound> &let_bounds) {
        scope.push(name, bound);

        vector<string> pending;
        set<string> unique;
        pending.push_back(name);
        unique.insert(name);
        while (!pending.empty()) {
            string next = std::move(pending.back());
            pending.pop_back();
            for (const auto &v : children[get_var_instance(next)]) {
                if (unique.insert(v.var).second) {
                    pending.push_back(v.var);
                    string max_name = unique_name('t');
                    string min_name = unique_name('t');

                    let_bounds.insert(let_bounds.begin(), LetBound(v.var, min_name, max_name));

                    Type t = let_stmts.get(v.var).type();
                    Interval b = Interval(Variable::make(t, min_name), Variable::make(t, max_name));
                    scope.push(v.var, b);
                }
            }
        }
    }

```

steven-johnson: Yeah, it visits everything, it just uses a set to filter out the dupes, so it isn't as efficient as it could be. I'll give your suggestion a try.

steven-johnson: One possible gotcha with yours: the toplevel 'name' passed in needs to be pushed onto scope, but I don't think it should be added to let_bounds, as while let_stmts are expected for all the children, I don't think it's expected for the name itself.

steven-johnson: Edit: ah, nevermind, I see this explicitly only processes the children. This looks better than mine, I'll update the PR.

abadams:
Actually I think my algorithm (and the original one) doesn't guarantee the property that children come after their parents. Not sure how this would manifest in a compilation. According to wikipedia, to do a topological sort with a depth-first traversal you need to prepend the node to the list *after* recursively visiting all children. I'll try fixing it.

abadams:
ok, I *think* this is a correct topological sort now.

steven-johnson:
> Actually I think my algorithm (and the original one) doesn't guarantee the property 

...if the original one didn't guarantee it either, was that property actually important? Or were we potentially miscalculating bounds in some cases where that property wasn't maintained?

abadams:
I think the original one was OK because it allowed duplicates. If a child was visited before a parent that's fine, because it'll just visit the child again! If we're removing duplicates we need to also be more careful about the order.

steven-johnson:
gotcha. testing in google now.

steven-johnson:
LGTM. (Ironically, I can't approve this since I opened it, even though the change is now your code...)

PR #4840, opened 2020-04-10
Add RDom[i] method to Python bindings.

Infinoid:
This allows indexed access to the RVars contained by an RDom, just like the C++ API.

### DIFF
--- a/python_bindings/correctness/rdom.py
+++ b/python_bindings/correctness/rdom.py
@@ -27,14 +27,17 @@ def test_rdom():
     assert r.y.name() == r[1].name()
     try:
         r[-1].name()
+        raise Exception("underflowing index should raise KeyError")
     except KeyError:
         pass
     try:
         r[2].name()
+        raise Exception("overflowing index should raise KeyError")
     except KeyError:
         pass
     try:
         r["foo"].name()
+        raise Exception("bad index type should raise TypeError")
     except TypeError:
         pass
 
--- a/python_bindings/correctness/rdom.py
+++ b/python_bindings/correctness/rdom.py
@@ -23,6 +23,9 @@ def test_rdom():
             else:
                 assert output[ix, iy] == 1
 
+    assert r.x.name() == r[0].name()
+    assert r.y.name() == r[1].name()
+
     return 0
 
 if __name__ == "__main__":
--- a/python_bindings/correctness/rdom.py
+++ b/python_bindings/correctness/rdom.py
@@ -25,6 +25,18 @@ def test_rdom():
 
     assert r.x.name() == r[0].name()
     assert r.y.name() == r[1].name()
+    try:
+        r[-1].name()
+    except KeyError:
+        pass
+    try:
+        r[2].name()
+    except KeyError:
+        pass
+    try:
+        r["foo"].name()
+    except TypeError:
+        pass
 
     return 0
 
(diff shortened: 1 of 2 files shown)


Infinoid:
I'm not sure if the constness is right, or if it needs any special bounds checking logic.

Infinoid:
It does need explicit KeyError throwing for out of bounds requests, I'll add that.

steven-johnson:
Wow, this was quite an oversight on my part. Testing now.

steven-johnson:
Fails on mac with 

```
  File "/Users/halidenightly/build_bot_new/worker/mac-64-testbranch-1000/halide/python_bindings/correctness/atomics.py", line 22, in <module>
    test_atomics()
  File "/Users/halidenightly/build_bot_new/worker/mac-64-testbranch-1000/halide/python_bindings/correctness/atomics.py", line 10, in test_atomics
    f[hl.Expr(im[r])] += 1
ValueError: allocator<T>::allocate(size_t n) 'n' exceeds maximum supported size
```

https://buildbot.halide-lang.org/master/#/builders/14/builds/143

Infinoid:
Is size_t the right index type for all platforms?

Infinoid:
Wait, I don't think this patch affects atomics.

steven-johnson:
> Wait, I don't think this patch affects atomics.

Maybe it's not supposed to, but apparently it is :-)

BachiLi:
The bug is related to https://github.com/halide/Halide/pull/4286

Infinoid:
Okay.  If pybind11 is calling __getitem__ methods in completely unrelated classes, I don't know how to solve that.  I'm open to suggestions :)

Infinoid:
I was able to confirm the failure steven-johnson saw, and a couple other failures.  This patch currently causes 3 test failures for me, on linux amd64:

correctness/basics:
```
$ python3 correctness/basics.py 
  File "correctness/basics.py", line 280, in <module>
    test_basics4()
  File "correctness/basics.py", line 163, in test_basics4
    f[g[r]] = 2.5
ValueError: vector::reserve
```

correctness/atomics:
```
$ python3 correctness/atomics.py 
  File "correctness/atomics.py", line 22, in <module>
    test_atomics()
  File "correctness/atomics.py", line 10, in test_atomics
    f[hl.Expr(im[r])] += 1
ValueError: vector::reserve
```

tutorial/lesson_13_tuples:
```
$ python3 tutorial/lesson_13_tuples.py 
  File "tutorial/lesson_13_tuples.py", line 302, in <module>
    main()
  File "tutorial/lesson_13_tuples.py", line 166, in main
    new_index = hl.select(old_max > input[r], r, old_index)
ValueError: vector::reserve
```

In all 3 cases, it looks like an RDom is being used as a subscript index.

Infinoid:
I think that the existence of a `__getitem__` method convinces pybind11 that RDom is a container type, so it tries to convert it into a vector, and fails.

I spent some time trying to convince pybind11 to pass it directly and let the C++ code figure it out.  In the end, I couldn't make that work.  But I noticed adding a `__len__` method allows pybind11 to do the vector conversion successfully, and that allows all of the tests to pass.

```
100% tests passed, 0 tests failed out of 40
```

With this change, f(r) will never pass the RDom object down to the C++ code directly.  Instead, it will extract the individual RVars and pass those down, instead.  This seems like it will add some memory churn and be a little slower, but otherwise it doesn't seem like a problem to me.

PR #4841, opened 2020-04-10
Permit "safe" parallel scatters, even when they race

### LINKED ISSUE #4658: update stages fail associativity when rhs is const value
The following code works, because prove_associativity() correctly evaluates +=1

```
//existing hist sample from lesson_09_update_definitions

Func histogram("histogram");
histogram(x) = 0;
RDom r(0, input.width(), 0, input.height());
histogram(input(r.x, r.y)) += 1;
        
//this schedule is added to test associativity
histogram.update(0).reorder(r.y,r.x);

Buffer<int> halide_result = histogram.realize(256);

```

Changing the updates stage to a const assignment fails

```
histogram(input(r.x, r.y)) = 1;
```

With the error 

```
In schedule for histogram.update(0), can't reorder RVars r$y and r$x because it may change the meaning of the algorithm.
```

abadams:
It's safe(*) to parallelize/split/reorder a reduction domain if all
possible races are races to store the same value to the same index. For
example:
```
f(g(r)) = 4;
```

It's also okay to use the pure vars on the RHS, because for any possible
write-after-write hazard we know the value of the pure var is the same
because we require the pure var to appear unadorned on the LHS:
```
f(g(r, x), x) = h(x);
```

(*) In C/C++ a "benign race condition" is a contradiction in terms,
because it's UB. I'm think LLVM IR makes similar assumptions in its
memory model. The closest example I could find to this race condition
that has a concrete failure mode in C/C++ is section 2.4 in
https://www.usenix.org/legacy/event/hotpar11/tech/final_files/Boehm.pdf.
However that example isn't a straight-up write-after-write hazard where
the writes have the same value. There are also unsynchronized reads
involved, and the reads are essential to get a failure (how else could
you observe one?). This PR explicitly prohibits cases that read the
values being written to. No reads can occur until the parallel loop
ends, which acts as a synchronization point. I'm not going to claim that
a failure is impossible, but I can't think of one. If we manage to
construct a failure for this kind of race we'll have to revisit.

Note that Halide currently permits this kind of race condition in
another context: We redundantly re-evaluate points near the end of a
split domain with shiftinwards, even when the outer loop is
parallelized:

```
f(x) = h(x);
f.split(x, xo, xi, 8, TailStrategy::ShiftInwards).parallel(xo);
```

In practice this sort of race is *not* benign with floating point
right-hand-sides, as the parallel loop could be partitioned and compiled
as two separate racing pieces of code, which undergo different floating
point optimizations, and then the values we're racing to store could then
actually be different. This often manifests as a strip of
nondeterministic values near the bottom or right edge of an image in
algorithms that are sensitive to floating point nonsense. The
strict-float target flag is useful for people who worry about such
things. Or just don't write schedules with this class of race condition
in them. We try to not let users shoot themself in the foot, but let's
not put them in a padded cell.

Fixes #4658


zvookin:
Have you tried vetting write/write of the same value with TSan? Either via Halide or using a C++ test case?

If it can cause TSan failures, I'd at least like to gate this. E.g. a scheduling directive to enable it similar to "allow_race_conditions".

abadams:
tsan errors on redundant write races in C++ even when no read of the memory location ever occurs in the entire program. The race itself is UB in C++ so this is reasonable. A correct but malicious compiler could detect the race and inject nasal demons.

In Halide called from tsan-instrumented C++, with target flag tsan enabled, no error occurs because the writes don't get instrumented. So if you're worried about triggering tsan false positives, this doesn't.

steven-johnson:
> A correct but malicious compiler

At this point I think it's fair to assume that all C++ compilers have at least a little malice in them

abadams:
The "correct" way to express this in C++ is with atomic stores with memory order release (or maybe relaxed - there's a full memory fence before any loads take place). tsan doesn't complain in those circumstances and you get the same assembly as with naive stores. This does affect the llvm bitcode though - you get the atomic flag on the stores. This presumably prevents optimizations that would be unsafe in C++ (register promotion?) So I guess maybe I should treat this sort of thing as associative but not safe to parallelize, and gate parallelizing it on atomic(). Unfortunately the implementation is going to be a little more complex this way.

abadams:
Done. Let's see what the buildbots say.

zvookin:
I would expect the stipulation that the assembly is the same using atomic instructions to be architectures specific. x86 has a much more restrictive memory model than most other architectures.

abadams:
Yeah I nearly went back and qualified my comment with x86. I wanted to ensure there were no memory fences inside the loop in the emitted code on that platform, as it really doesn't need them. I imagine ARM is the same though. The point of the atomic store is to prevent compiler optimizations inside llvm that assume other threads aren't also trying to store the same value to the same site (Though I can't imagine what they could be that would cause a failure). The atomic flag on the store node should do that.

abadams:
Buildbots are happy. Ready for review.

steven-johnson:
ping to @zvookin

abadams:
Bump @zvookin 

This PR changed a little from the original description. It lets the atomic() scheduling directive also apply to simple assignments in addition to associative commutative operators, e.g. hist(f(r.x), x) = g(x) is safe to parallelize over r.x if the stores are atomic, because the RHS doesn't depend on the hist or r.x

alexreinking:
Are we still waiting on this for Halide 12?

alexreinking (approved):

alexreinking:
CUDA failures are `cudaErrorCompatNotSupportedOnDevice = 804`
> This error indicates that the system was upgraded to run with forward compatibility but the visible hardware detected by CUDA does not support this configuration. Refer to the compatibility documentation for the supported hardware matrix or ensure that only supported hardware is visible during initialization via the CUDA_VISIBLE_DEVICES environment variable.

Smells like driver issue, here. Maybe these bots need a reboot.

steven-johnson:
Uh, there were 5 failures on three different OSes, are we sure this is ready to land?

abadams:
They were all unrelated (e.g. cuda driver issues, issues that we have since fixed). This PR doesn't affect existing code, it just makes something that used to be illegal legal.

PR #4843, opened 2020-04-10
In trunk llvm it is no longer legal to call a Value directly

abadams:


abadams:
It should work. Other uses of createcall always used llvm functions as the callee. This call site was an anomaly.

steven-johnson:
Verified this fix locally for LLVM9/10/trunk; landing now (in advance of buildbots) to unbreak things.

PR #4844, opened 2020-04-10
fix eigen3 build on Gentoo Linux by using pkg-config

pelikan:
pkg-config is a standard way of discovering flags needed to build and
link your dependencies; all the dependency package needs to do is create
a small text file in /usr/share/pkgconfig/ and then you can either run:

$ pkg-config --cflags eigen3
$ pkg-config --libs eigen3

These commands will show you the CFLAGS and LDFLAGS as intended by the
library author (for instance on Debian, apt install libeigen3-dev will
create /usr/share/pkgconfig/eigen3.pc).

CMake can then find these dependencies too.  Tested on 3.14.5 and 3.17.

I'm keeping it optional in case some other platforms don't support it,
but according to cmake.org documentation:
- find_library(<VAR>)		sets <VAR>-NOTFOUND
- pkg_check_modules(<prefix>)	sets <prefix>_FOUND
so I fixed and unified those as well.


alexreinking:
I'm fine with using PkgConfig, but not by default. We also strongly prefer using imported targets whenever possible. This needs some reworking before I'm comfortable merging it. In particular, pkg_check_modules has an argument `IMPORTED_TARGET` that would create a target named `PkgConfig::<prefix>` (in your PR, prefix is EIGEN3). Then that should be linked instead.

apps/linear_algebra/benchmarks/CMakeLists.txt
```diff
@@ -34,19 +34,25 @@ target_compile_definitions(halide_benchmarks PRIVATE -DUSE_HALIDE ${MISC_DEFINIT
 target_link_libraries(halide_benchmarks PRIVATE halide_blas)
 list(APPEND BLAS_NAMES halide)
 
-find_package(Eigen3 QUIET)
-if (NOT Eigen3_FOUND)
+find_package(PkgConfig)
```
alexreinking, line 37: I much prefer using find_package _first_ and _then_ checking via PkgConfig if it isn't found.

apps/linear_algebra/benchmarks/CMakeLists.txt
```diff
@@ -34,19 +34,25 @@ target_compile_definitions(halide_benchmarks PRIVATE -DUSE_HALIDE ${MISC_DEFINIT
...
+if (PkgConfig-NOTFOUND)
+  find_package(Eigen3 QUIET)
+  set(EIGEN3_FOUND NOT Eigen3-NOTFOUND)
+  if (EIGEN3_FOUND)
+    set(EIGEN3_LDFLAGS Eigen3::Eigen)
+  endif()
+else()
+  pkg_check_modules(EIGEN3 eigen3)
```
alexreinking, line 45: Use IMPORTED_TARGET here... https://cmake.org/cmake/help/v3.14/module/FindPkgConfig.html#command:pkg_check_modules

apps/linear_algebra/benchmarks/CMakeLists.txt
```diff
@@ -34,19 +34,25 @@ target_compile_definitions(halide_benchmarks PRIVATE -DUSE_HALIDE ${MISC_DEFINIT
...
 
-find_package(Eigen3 QUIET)
-if (NOT Eigen3_FOUND)
+find_package(PkgConfig)
+if (PkgConfig-NOTFOUND)
+  find_package(Eigen3 QUIET)
+  set(EIGEN3_FOUND NOT Eigen3-NOTFOUND)
+  if (EIGEN3_FOUND)
+    set(EIGEN3_LDFLAGS Eigen3::Eigen)
+  endif()
```
alexreinking, line 43: Prefer `if (TARGET Eigen3::Eigen)` ... no need for this `set()` call.

alexreinking: Also just call the variable `EIGEN3` or maybe `EIGEN3_TARGET`, but not `LDFLAGS`

alexreinking:
This might be a better starting point:

```
find_package(Eigen3)
set(Eigen3 Eigen3::Eigen)
if (NOT TARGET ${Eigen3})
  find_package(PkgConfig)
  pkg_check_modules(Eigen3 QUIET IMPORTED_TARGET eigen3)
  if (Eigen3_FOUND)
    set(Eigen3 PkgConfig::Eigen3)
  endif()
endif()

if (NOT TARGET ${Eigen3})
  ...
else()
  ...
  target_compile_options(eigen_benchmarks PRIVATE -Wno-error=unused-variable)
  target_link_libraries(eigen_benchmarks PRIVATE ${Eigen3})
  ...
endif()
```

alexreinking:
Though I should also note that it is worth checking whether the Gentoo package comes with any of  `FindEigen3.cmake`, `Eigen3Config.cmake` or `Eigen3-config.cmake`. If it does, then you should be able to pass the appropriate command line flags and not have to modify the build files at all.

The ebuild makes it look like it does... https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-cpp/eigen/eigen-3.3.7.ebuild

pelikan:
Thank you for the input, you're probably the most knowledgeable person in CMake that I have met so far and it feels reassuring finally knowing what to do.  Unfortunately I don't think using pkg-config as a second option makes sense, since the FindEigen3.cmake file is present, so it would always try to use that.  On second thought, I don't think target_link_libraries is the right place for it since Eigen doesn't seem to actually *link* anything, just adds -I/usr/include/eigen.

I'll find a way to make it work with find_library, even though they appear to find different versions on my system (2.9.something vs. 3.5.5).  Second commit coming soon.

apps/linear_algebra/benchmarks/CMakeLists.txt
pelikan: I only called it _LDFLAGS for consistency with pkgconfig, but that's gone now.

alexreinking:
> Thank you for the input, you're probably the most knowledgeable person in CMake that I have met so far and it feels reassuring finally knowing what to do. 

Well, thanks :)

> Unfortunately I don't think using pkg-config as a second option makes sense, since the FindEigen3.cmake file is present, so it would always try to use that.

What goes wrong when it uses it? It might be a bug with the Gentoo package. Since I don't run Gentoo, I can't repro.

>  On second thought, I don't think target_link_libraries is the right place for it since Eigen doesn't seem to actually _link_ anything, just adds -I/usr/include/eigen.

CMake's naming is confusing here. The `target_link_libraries` command handles not only link flags, but definitions, include paths, compiler flags, and more. It's a full system for propagating usage requirements between dependencies. Look up "interface targets" and "interface properties" for more.

apps/linear_algebra/benchmarks/CMakeLists.txt
pelikan: fair enough, it seems we can do this without pkgconfig anyway.

apps/linear_algebra/benchmarks/CMakeLists.txt
pelikan: I didn't fully understand the IMPORTED keywords, so hopefully it won't be relevant after the next commit.

apps/linear_algebra/benchmarks/CMakeLists.txt
```diff
@@ -35,18 +35,16 @@ target_link_libraries(halide_benchmarks PRIVATE halide_blas)
 list(APPEND BLAS_NAMES halide)
 
 find_package(Eigen3 QUIET)
-if (NOT Eigen3_FOUND)
+if (NOT EIGEN3_FOUND)
```
alexreinking, line 38: This should really be `if (NOT TARGET Eigen3::Eigen)`

pelikan:
> What goes wrong when it uses it? It might be a bug with the Gentoo package. Since I don't run Gentoo, I can't repro.

First, there's a big error during the configure phase, saying "Target eigen_benchmarks links to target Eigen3::Eigen but the target was not found."  Then, when it starts compiling the code, the <Eigen/Eigen> header isn't found because it's in /usr/include/eigen3/ which isn't mentioned in -I.  

pkgconfig puts it there, and that's the only thing it does.  FindEigen3.cmake also only exports "found", "version" and "include_dir".


> CMake's naming is confusing here. The `target_link_libraries` command handles not only link flags, but definitions, include paths, compiler flags, and more. It's a full system for propagating usage requirements between dependencies. Look up "interface targets" and "interface properties" for more.

I'll read up on it.  My next commit still doesn't work, as the Eigen3::Eigen doesn't exist.  I'll have to send a third one :-(

apps/linear_algebra/benchmarks/CMakeLists.txt
```diff
@@ -35,18 +35,16 @@ target_link_libraries(halide_benchmarks PRIVATE halide_blas)
...
   add_executable(eigen_benchmarks
     eigen_benchmarks.cpp
   )
   target_compile_definitions(eigen_benchmarks PRIVATE -DEIGEN_DONT_PARALLELIZE ${MISC_DEFINITIONS})
   target_compile_options(eigen_benchmarks PRIVATE -Wno-error=unused-variable)
-  target_link_libraries(eigen_benchmarks
-    PRIVATE
-      Eigen3::Eigen
-  )
+  target_include_directories(eigen_benchmarks PRIVATE ${EIGEN3_INCLUDE_DIR})
```
alexreinking, line 46: This shouldn't be necessary and should be handled by `target_link_libraries` below... see https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html

Also see [here](https://bitbucket.org/eigen/eigen/src/94875feeeeb9abe5509b314197da1991ba2070f5/cmake/FindEigen3.cmake#lines-103:107) where `${EIGEN3_INCLUDE_DIR}` is added to the INTERFACE_INCLUDE_DIRECTORIES property of `Eigen3::Eigen`

alexreinking:
> First, there's a big error during the configure phase, saying "Target eigen_benchmarks links to target Eigen3::Eigen but the target was not found." Then, when it starts compiling the code, the <Eigen/Eigen> header isn't found because it's in /usr/include/eigen3/ which isn't mentioned in -I.
> 
> pkgconfig puts it there, and that's the only thing it does. FindEigen3.cmake also only exports "found", "version" and "include_dir".

What versions of CMake and Eigen do you have installed? This might be a packaging error on Gentoo's end.

Did you try the code I suggested above? It should work for you since your FindEigen3.cmake apparently isn't creating the `Eigen3::Eigen` target.

pelikan:
> What versions of CMake and Eigen do you have installed? This might be a packaging error on Gentoo's end.

CMake 3.17 and Eigen 3.3.5-r1, just upgrading to 3.3.7.


> Did you try the code I suggested above? It should work for you since your FindEigen3.cmake apparently isn't creating the `Eigen3::Eigen` target.

Will do.  I've attached the FindEigen3.cmake that I have with the 3.3.7 version:

[FindEigen3.cmake.txt](https://github.com/halide/Halide/files/4463867/FindEigen3.cmake.txt)

apps/linear_algebra/benchmarks/CMakeLists.txt
pelikan: This just got a lot more interesting.  Without the target_include_directories() it (meaning your code) doesn't work for me, and with target_link_libraries(... ${Eigen3}) it fails when it tries to add "-lEigen3" to the linker, which of course doesn't exist.

I'm curious about your platform where this Eigen3::* stuff actually exists, and where Eigen is a dynamic library.  Since I found a second bug in Eigen (a missing const triggering Halide's -Werror) so I guess that will be my next stop :-(

apps/linear_algebra/benchmarks/CMakeLists.txt
pelikan: OK, I'll switch to using the TARGET instead of the BOOL variable they defined.

apps/linear_algebra/benchmarks/CMakeLists.txt
pelikan: Ha, the -lEigen3 was my mistake.  I might have a working version now...

pelikan:
PTAL, this version works and is the most like the one you sent.  The eigen3 website confirms they don't have any libraries to link, yet the target_link_libraries() works (adds -isystem to the clang and nothing to the link).

Thank you for the patience!

alexreinking:
@pelikan - great! glad I could help. I'll run our buildbots on it and merge when those pass.

apps/linear_algebra/benchmarks/CMakeLists.txt
```diff
@@ -34,19 +34,24 @@ target_compile_definitions(halide_benchmarks PRIVATE -DUSE_HALIDE ${MISC_DEFINIT
 target_link_libraries(halide_benchmarks PRIVATE halide_blas)
 list(APPEND BLAS_NAMES halide)
 
-find_package(Eigen3 QUIET)
-if (NOT Eigen3_FOUND)
+find_package(Eigen3 QUIET NO_MODULE)
```
steven-johnson, line 37: Would be nice to add some comments here for future readers, since this took some back and forth to get right...

pelikan: Added comments summarizing our discussion.

PR #4845, opened 2020-04-10
fix build (const& vs. const) with LLVM/clang 11 and -Werror

pelikan:
The error message I get (LLVM and clang from today's HEAD):

error: loop variable 'l' creates a copy from type 'const Halide::Internal::BoxesTouched::LetBound' [-Werror,-Wrange-loop-construct]


abadams:
Oops, nice catch, thanks.

PR #4847, opened 2020-04-11
fix build when -stdlib=libc++ was passed into the plain C test

pelikan:
When LLVM was built with it, the -stdlib=libc++ flag would get passed to
correctness_plain_c_includes, which is (correctly) only using the clang
C compiler driver, not clang++, where this is an invalid option.

Since CMake doesn't seem to have per-language options apart from
CMAKE_CXX_FLAGS which people discourage from directly modifying,
we use a conditional instead of the string itself.

Note that passing the option to the linker is fine as is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,9 @@ endif()
 
 string(FIND "${LLVM_CONFIG_CXXFLAGS}" "-stdlib=libc++" LLVM_LIBCXX)
 if (LLVM_LIBCXX GREATER -1)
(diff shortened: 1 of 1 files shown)


steven-johnson:
LGTM but deferring to @alexreinking since he's all over the CMake build files right now

steven-johnson:
gentle ping to @alexreinking

pelikan:
If you want SSH access to a VM where this can be reproduced, tell me and I can arrange something.

alexreinking:
Lgtm

PR #4849, opened 2020-04-11
We want reverse topological order for let bounds

abadams:
#4839 fixed let_bounds in trim_scope_push to use topological order instead of a redundant depth-first traversal of the entire graph. However, by printing out the order of the let_bounds array before and after we made that change (something I should have done with the original PR), I discovered that we actually want *reverse* topological order.

Turns out nothing in the public test suite was sensitive to topological order, but in the add_image_checks_after_bounds_inference branch the compute_with test is.


abadams:
It would be a rathole. compute_with bounds inference checking in #3037 is a decent test of it

PR #4850, opened 2020-04-11
Add image checks simplifier changes

abadams:
This PR carves off some independently mergeable pieces from #3037. It makes simplify_correlated_differences more aggressive, adds some new simplifier rules, and adds some division handling to solve_expression. 

Also includes #4849

### DIFF
--- a/src/SimplifyCorrelatedDifferences.cpp
+++ b/src/SimplifyCorrelatedDifferences.cpp
@@ -2,6 +2,7 @@
 
 #include "CSE.h"
 #include "ExprUsesVar.h"
+#include "IRMatch.h"
 #include "IRMutator.h"
 #include "IROperator.h"
 #include "Monotonic.h"
@@ -25,7 +26,12 @@ class SimplifyCorrelatedDifferences : public IRMutator {
 
     Scope<Monotonic> monotonic;
 
-    vector<pair<string, Expr>> lets;
+    struct OuterLet {
+        string name;
+        Expr value;
+        bool may_substitute;
+    };
+    vector<OuterLet> lets;
 
     template<typename LetStmtOrLet, typename StmtOrExpr>
     StmtOrExpr visit_let(const LetStmtOrLet *op) {
@@ -45,14 +51,33 @@ class SimplifyCorrelatedDifferences : public IRMutator {
         std::vector<Frame> frames;
         StmtOrExpr result;
 
+        // Note that we must add *everything* that depends on the loop
+        // var to the monotonic scope and the list of lets, even
+        // things which we can never substitute in (e.g. impure
+        // things). This is for two reasons. First this pass could be
+        // used at a time when we still have nested lets under the
+        // same name. If we decide not to add an inner let, but do add
+        // the outer one, then later references to it will be
+        // incorrect. Second, if we don't add something that happens
+        // to be non-monotonic, then is_monotonic finds a variable
+        // that references it in a later let, it will think it's a
+        // constant, not an unknown.
         do {
             result = op->body;
-            if (op->value.type() == Int(32) && is_pure(op->value)) {
+            if (loop_var.empty()) {
+                frames.emplace_back(op);
+                continue;
+            }
+
+            bool pure = is_pure(op->value);
+            if (!pure || expr_uses_vars(op->value, monotonic)) {
                 frames.emplace_back(op, loop_var, monotonic);
                 Expr new_value = mutate(op->value);
-                lets.emplace_back(op->name, new_value);
+                bool may_substitute_in = new_value.type() == Int(32) && pure;
+                lets.emplace_back(OuterLet{op->name, new_value, may_substitute_in});
                 frames.back().new_value = std::move(new_value);
             } else {
+                // Pure and constant w.r.t the loop var
                 frames.emplace_back(op);
             }
         } while ((op = result.template as<LetStmtOrLet>()));
@@ -101,7 +126,10 @@ class SimplifyCorrelatedDifferences : public IRMutator {
             tmp_monotonic.swap(monotonic);
             tmp_lets.swap(lets);
             loop_var = op->name;
-            s = IRMutator::visit(op);
+            {
+                ScopedBinding<Monotonic> bind(monotonic, loop_var, Monotonic::Increasing);
+                s = IRMutator::visit(op);
+            }
             loop_var.clear();
             tmp_monotonic.swap(monotonic);
             tmp_lets.swap(lets);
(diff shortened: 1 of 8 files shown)


src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -45,14 +51,33 @@ class SimplifyCorrelatedDifferences : public IRMutator {
...
+        // var to the monotonic scope and the list of lets, even
+        // things which we can never substitute in (e.g. impure
+        // things). This is for two reasons. First this pass could be
+        // used at a time when we still have nested lets under the
+        // same name. If we decide not to add an inner let, but do add
+        // the outer one, then later references to it will be
+        // incorrect. Second, if we don't add something that happens
+        // to be non-monotonic, then is_monotonic finds a variable
+        // that references it in a later let, it will think it's a
+        // constant, not an unknown.
```
steven-johnson, line 64: Excellent comment, thanks

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -45,14 +51,33 @@ class SimplifyCorrelatedDifferences : public IRMutator {
...
+                continue;
+            }
+
+            bool pure = is_pure(op->value);
+            if (!pure || expr_uses_vars(op->value, monotonic)) {
                 frames.emplace_back(op, loop_var, monotonic);
                 Expr new_value = mutate(op->value);
-                lets.emplace_back(op->name, new_value);
+                bool may_substitute_in = new_value.type() == Int(32) && pure;
+                lets.emplace_back(OuterLet{op->name, new_value, may_substitute_in});
```
steven-johnson, line 77: for emplace_back, shouldn't it just be emplace_back(op->name, new_value, may_substitute_in)?

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -125,16 +199,28 @@ class SimplifyCorrelatedDifferences : public IRMutator {
...
             }
             e = common_subexpression_elimination(e);
             e = solve_expression(e, loop_var).result;
+            e = PartiallyCancelDifferences().mutate(e);
             e = simplify(e);
 
             if ((debug::debug_level() > 0) &&
-                is_monotonic(e, loop_var, monotonic) == Monotonic::Unknown) {
+                is_monotonic(e, loop_var) == Monotonic::Unknown) {
                 // Might be a missed simplification opportunity. Log to help improve the simplifier.
```
steven-johnson, line 221: Suggestion: add an Issue for "things we could log in a structured way during batch compiles to gather useful data", and add a TODO for it here

abadams:
I'm hoping this will divide the Google failures from #3037 in two so that we can diagnose separately. Will address your comments tomorrow morning.

steven-johnson:
A few dozen failures in google, but the ones I've looked at all seem to be from one common cause (a generator failing with a no-longer-constant loop unroll):

```
Error: Can only unroll for loops over a constant extent.
Loop over composited.s0.c has extent (4 - result.s0.c).
```

steven-johnson:
Hm, this is especially odd because the schedule explicitly bounds that:

```
  Func composited = ...;
  Func result = ...expression involving composited...;

  // schedule of composited:
  composited.update(0)
      .specialize(result_.width() >= kVectorSizeFloat)
      .vectorize(x, kVectorSizeFloat);
  composited.compute_at(result_, tile)
      .reorder(c, x, y)
      .reorder_storage(c, x, y)
      .bound(c, 0, 4)
      .unroll(c)
      .parallel(y, 16)
      .specialize(result_.width() >= kVectorSizeFloat)
      .vectorize(x, kVectorSizeFloat);



```

steven-johnson:
Only relevant output from HL_DEBUG_CODEGEN=1:

```
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
Warning: expression is non-monotonic in loop variable result.s0.c: (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)
```

steven-johnson:
The 'fix' for this is simple enough (change `unroll(c)` to `unroll(c, 4)`) -- but this seems unnecessary, since the bounds is clearly constant. (I checked the codegen before and it was indeed unrolling at 4)

abadams:
unroll(c, 4) adds an extra split and loop level, so it's not a great solution. The bound should be enough. A way to diagnose might be to see which stage of lowering is making the loop bound non-constant, if it ever was.

steven-johnson:
For both old and new, we have `let composited.s0.c.loop_extent = ((composited.s0.c.max + 1) - composited.s0.c.min)` just before the first call to `simplify_correlated_differences()`; in this branch, it changes to `let composited.s0.c.loop_extent = (max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)` after that call.

abadams:
What are composited.s0.c.max and composited.s0.c.min defined as though? That bound directive should mean that they're constants, but it seems like that's not the case?

steven-johnson:
both with and without this PR, and both before and after `simplify_correlated_differences()`:

```
let composited.s0.c.max = max(max(composited.s1.c.max, result.s0.c.max), 3)
let composited.s0.c.min = min(min(composited.s1.c.min, result.s0.c.min), 3)
let composited.s0.y.max = composited.s2.y.max
let composited.s0.y.min = composited.s2.y.min
let composited.s0.x.max = composited.s2.x.max
let composited.s0.x.min = composited.s2.x.min
let composited.s0.c.max_unbounded = composited.s0.c.max
let composited.s0.c.min_unbounded = composited.s0.c.min
let composited.s0.c.max = ((0 + 4) - 1)
```

abadams:
Oh, the bound injects a shadowed let, and then simplify_correlated_differences uses the wrong outer version! That's a bug.

abadams:
I can only repro if I have some of result's c loop outside of its loop over tiles, which makes this a very silly schedule. Note that `(max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)` is the extent required, and it's definitely less than four, so the schedule must be doing wasted work. Is it possible that's going on in your case?

The bug is still a bug and needs fixing, but there could additionally be an opportunity to improve this schedule.

Here's my repro:

```
    Func f("f"), g("g");
    Var x, y, c("c");

    f(x, y, c) = x + y + c;

    g(x, y, c) = f(x, y, c) + f(x, y, 3);

    Var xi, yi, ci;
    g.compute_root().tile(x, y, xi, yi, 32, 32);
    f.compute_at(g, x).bound(c, 0, 4).unroll(c);

    g.realize(1024, 1024, 4);
```

abadams:
Fix pushed.

steven-johnson:
Seems to address the issue I saw (plus another one I was puzzling over but hadn't posted about here yet). I'm gonna look at the before-and-after the last commit just to verify what's going on but so far this looks good.

steven-johnson:
> I can only repro if I have some of result's c loop outside of its loop over tiles, which makes this a very silly schedule. Note that `(max(max((3 - result.s0.c), (result.s0.c + -3)), 0) + 1)` is the extent required, and it's definitely less than four, so the schedule must be doing wasted work. Is it possible that's going on in your case?

Not that I can tell, but I do see several warnings of the form:

```
Warning: expression is non-monotonic in loop variable result.s0.x.xo.tile: (let t1245 = min((result.s0.x.xo.tile*128), (result.extent.1 + -128)) in max(((min((downsampled_foreground.extent.1*2), ((result.min.1 + t1245) + 128)) - t1245) - result.min.1), 0))
```

note that `result` has a schedule something like

```
  result_.tile(x, y, xo, yo, xi, yi, kTileSize, kTileSize)
      .fuse(xo, yo, tile)
      .parallel(tile);
```

abadams:
That looks a lot like my repro. If result has a c channel, then that c loop was never reordered into the tile loop and is still outermost as a serial loop outside the parallel loop over tiles, which is probably not the intent.

steven-johnson:
> If result has a c channel,

Ah, gotcha. Yeah, result has a c channel, but apparently the storage layout and extent is unspecified (strides are made undefined), so it could be 1, 3, 4, interleaved, planar.

abadams:
That makes sense, so this is definitely doing wasted work then. In the three-channel output case, for channel zero of result it computes channels 0-3 of composited. Then for channel 1 it computes 1-3, and then for channel 2 it computes channels 2-3. So there's a lot of redundant work that would go away with result.reorder(xi, yi, c, tile).

steven-johnson:
> result.reorder(xi, yi, c, tile).

That makes sense re: a better schedule, but FWIW, it doesn't make the non-monotonic warning go away.

abadams:
Yeah that's independent. See if the new simplifier rules help.

steven-johnson:
Before b248f94 was pushed, LGTM in google. With that commit I need to re-run everything as a reality check.

The new rules make that warning go away but I still have two remaining (sorry, I should have listed all of them before):

```
Warning: expression is non-monotonic in loop variable result.s0.x.xo.tile: (let t1247 = min((result.s0.x.xo.tile*128), (result.extent.1 + -128)) in min(((t1247 - min((downsampled_foreground.extent.1*2), ((result.min.1 + t1247) + 128))) + result.min.1), 0))
Warning: expression is non-monotonic in loop variable result.s0.x.xo.tile: (let t1251 = min((result.s0.x.xo.tile*128), (result.extent.1 + -128)) in min(((t1251 - min((downsampled_foreground.extent.1*2), ((result.min.1 + t1251) + 128))) + result.min.1), 0))
```

test/correctness/simplify.cpp
```diff
@@ -932,6 +932,24 @@ void check_bounds() {
     check(min(min(x + z, y), w) - x, min(min(w, y) - x, z));
     check(min(min(y, x + z), w) - x, min(min(w, y) - x, z));
 
+    // More deeply-nested cancellations
```
abadams, line 935: Beginning to question the value of these tests in the machine-generated-rule case - I just copy-pasted the rules and replaced "rewrite" with "check". I think what they usefully check is that there isn't an earlier rule that prevents the given rule from triggering on an expression that exactly matches the LHS. That has some value still - it has saved me from committing a rule that can't ever trigger a few times.

steven-johnson: I defer to your judgement there.

abadams:
As a test I'm going to copy-paste those into the rule synthesis machinery without even reading them.

steven-johnson:
> That makes sense, so this is definitely doing wasted work then.

On the benchmarks on my local box, that produces something like -30% CPU, -50% to allocations, so, yeah, big win :-)

abadams:
The rule it generated took two steps in one specific rule instead of just applying a more simple more general rule and then letting the other rules work from there. So I just added the missing rules by hand and verified them.

steven-johnson:
That cleared up those two. One other (from the other failure case) that I'm still seeing is below, but this is just an FYI... this is LGTM as soon as a retest with the new rules finishes in google.

```
Warning: expression is non-monotonic in loop variable disparity.s0.x.x: (let t78 = min((disparity.s0.x.x*16), (disparity.extent.0 + -16)) in (let t80 = min(max((disparity.min.0 + t78), (max_disparity + -1)), ((disparity.min.0 + t78) + 16)) in (t80 - max(min(min_disparity, t80), (disparity.min.0 + t78)))))
```

steven-johnson:
LGTM

PR #4854, opened 2020-04-14
Remove 'weak' from extern-Function declarations in AOT

steven-johnson:
This addresses an issue seen when building with Clang-for-Windows (the Chrome toolchain): when building AOT code (but not the runtime), ensure that all the 'weak' llvm::Functions are remarked as non-weak. The problem this addresses is that when we generate a call to one of the external-weak functions (e.g. `halide_string_to_string`), LLVM will declare that symbol as .weak -- on Unixy systems this is fine, but on Windows (where weak symbols don't exist), you can end up with multiple non-weak references to the same symbol, causing linkage to fail.

This fix *could* be made Windows specific, but I'd argue that the caller of an extern Func shouldn't know or care about the eventual weakness of the linked-in symbol, so we just de-weakify them for all platforms (I haven't found any apparent injections from doing so).

### DIFF
--- a/src/LLVM_Runtime_Linker.cpp
+++ b/src/LLVM_Runtime_Linker.cpp
@@ -520,7 +520,11 @@ llvm::Triple get_triple_for_target(const Target &target) {
     return triple;
 }
 
-void convert_weak_to_strong(llvm::GlobalValue &gv) {
+}  // namespace Internal
+
+namespace {
+
+void convert_weak_to_linkonce(llvm::GlobalValue &gv) {
     llvm::GlobalValue::LinkageTypes linkage = gv.getLinkage();
     if (linkage == llvm::GlobalValue::WeakAnyLinkage) {
         gv.setLinkage(llvm::GlobalValue::LinkOnceAnyLinkage);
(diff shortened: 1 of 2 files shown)


abadams:
It looks like this doesn't just change extern declarations - it also changes the linkage type of internal functions in a way that could lead them to be stripped. I think this will cause stripping of any function that:
A) is included with every piece of compiled code rather than appearing in the standalone runtime
B) isn't called by the Halide code directly
That sounds like the empty set, because such a function couldn't possibly work in contexts where you're linking multiple pipelines but weak linkage isn't a thing. 

Still... if the intent is to just mark all extern declarations as non-weak, that could be done in a more targeted way with:
```
if (f.getLinkage() == llvm::GlobalValue::ExternalWeakLinkage) {
  f.setLinkage(llvm::GlobalValue::ExternalLinkage);
}
```

The difference (on systems with weak linkage) between extern and extern weak is that if *no* definition for the symbol is provided, it quietly becomes null instead of being a link-time error. I don't think that behavior is useful to us? Certainly seems like a good way to get mysterious segfaults when your code branches to null.

steven-johnson:
PTAL

abadams:
I think the terminology here is backwards. For the non-extern symbols, they get changed from weak to linkonce, which is *even weaker*, not stronger. linkonce is weak plus the compiler is permitted to discard them if they're unused. This is why the original code was in terms of "can_strip" or not, not weak or strong. linkonce is the linkage type you get from template instantiations and inline functions, so windows must support it.

One approach to clarify would be to unconditionally make extern weak symbols just plain old extern, so that can be called unconditionally, and convert_weak_to_strong should instead be convert_weak_to_linkonce and do nothing to extern linkage types. Then the parameter to link_modules is just whether we convert *all* weak symbols to linkonce, or just the ones that aren't part of the halide_* api

steven-johnson:
> This is why the original code was in terms of "can_strip" or not, not weak or strong.

fwiw: the previous code was calling a function called `convert_weak_to_strong`, so the wrong nomenclature was already there. (no reason not to fix it now, of course.) stand by.

steven-johnson:
PTAL

abadams:
I think this is way clearer.

Issue #4853, opened 2020-04-14
associativity proof depends on scalar coefficients?


Infinoid:
I am having some problems with rfactor, and I managed to reduce it down to this simple program:

```
#include <Halide.h>
using namespace Halide;
#include "../tools/halide_image_io.h"
using namespace Halide::Tools;

int main(int argc, char **argv) {

    Halide::Buffer<uint8_t> in = load_image("rgb.png");
    Var i{"i"}, j{"j"}, rx{"rx"}, ry{"ry"}, rz{"rz"};
    RDom r{0, 8, 0, 8, 0, 3, "r"};
    Func f{"f"};
    f(i,j) = Expr((double)0.0);
    f(r.x,r.y) += in(r.x, r.y, r.z) * Expr(-5.0);

    Func f_intm = f.update(0).rfactor({{r.x, rx}, {r.y, ry}});

    Buffer<double> output = f.realize(100, 100);
}
```

When you compile and run it, it dies with the following error:

```
Error:
Failed to call rfactor() on f.update(0) since it can't prove associativity of the operator
Aborted
```

If you remove the negative sign, changing `-5.0` to `5.0`, it completes successfully.

Is this a bug?  I don't understand why rfactor would care what the scalar coefficient is.

abadams:
Yes this is a bug. Looks like it gets simplified to f(r.x, r.y) -= in(r.x, r.y, r.z) * 5.0, and then because subtraction isn't an associative/commutative binary operator it gives up. Thinking about how to fix...

abadams:
Actually that "simplification" is in the process of being removed (hopefully) in #4857 , so += should not get rewritten to -= in future. 

Would be better if we could recognize -= as a valid summation too, just with the output sign flipped, but rfactor can't currently do that sort of rewrite.

Infinoid:
After #4857 was merged, the original test program runs correctly now.  But as @abadams mentioned, associativity still fails for -=.  These two lines should be equivalent:

```
    f(r.x,r.y) += in(r.x, r.y, r.z) * Expr(-5.0);
    f(r.x,r.y) -= in(r.x, r.y, r.z) * Expr(5.0);
```

But the second one fails to compile.  `Error: Failed to call rfactor() on f.update(0) since it can't prove associativity of the operator`

PR #4857, opened 2020-04-14
Delete six dubious simplifier rules

abadams:
The deleted rules try to make negative constants positive, but don't
reduce the total op count, and replace an addition with a subtraction,
which doesn't count as a strength reduction.

Making negative constants positive may be defensible as a strength
reduction, but other rules may make positive constants negative again,
and elsewhere we treat replacing subtraction with addition as a strength
reduction. So we were unable to prove that these rules don't cause
cycles with these other rules (and hence infinite recursion). It would
be good to just delete them if possible.

In the open source tests, this required adding a few more variants of an
EQ rule on selects. These were formally verified. In the strength
ordering the new rules are cycle-free, even though they increase the op
count in two cases. They reduce the total number of leaf nodes by one
(c0 goes away) and we're treating the leaf count as higher priority than
the op count in our total ordering over expressions.


abadams:
Needs testing in Google. It's possible these rules were important for some reason. If so, it would be good to come up with alternatives that *do* obey the ordering.

steven-johnson:
I found exactly *one* test in google that is failing with this. Let me dig into it.

abadams:
Incidentally fixes #4853, which breaks because we change a += to a -=, and it stops looking like an associative and commutative reduction.

PR #4858, opened 2020-04-15
Add HALIDE_STATIC_CONSTEXPR

steven-johnson:
For clients compiling with C++17, using `static constexpr` (rather than `static const` for static member variables initialized inline.

(This has little effect on our current codebase -- the issues it fixes are ones we don't encounter, obviously -- but it tracks minor cleanup work we'll want to do once we decide to require C++17 as a baseline for Halide.)

### DIFF
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -217,7 +217,7 @@ struct IntImm : public ExprNode<IntImm> {
 
     static const IntImm *make(Type t, int64_t value);
 
-    static const IRNodeType _node_type = IRNodeType::IntImm;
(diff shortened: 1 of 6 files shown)


abadams:
I'm not sure I understand the motivation for this. Why would one pass an enum by reference anyway? Can't we just continue to use static const until it's time to switch wholesale to c++17?

steven-johnson:
Yeah, I made this PR in response to some proposed internal google changes, but it's pointless to land this now. I'll keep it locally as a placeholder/reminder.

(IMHO we should start considering when we want to move to C++17 as a baseline, though.)

Issue #4859, opened 2020-04-17
Python bindings leak memory in AOT mode


Infinoid:
The generated "python_extension" code leaks buffers.  The function `_convert_py_buffer_to_halide` accesses a buffer's memory by calling `PyObject_GetBuffer`.  At that point it holds a reference to the input buffer, and that reference is never released.

https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer says:

> Successful calls to PyObject_GetBuffer() must be paired with calls to PyBuffer_Release(), similar to malloc() and free(). Thus, after the consumer is done with the buffer, PyBuffer_Release() must be called exactly once.

The generated code never calls PyBuffer_Release.

Infinoid:
Here is a reproducer: https://gist.github.com/Infinoid/a90efd3b8b10c0859009f780b6033ac0

This reproducer calls an AOT-generated function 3000 times, with a newly allocated input buffer for each call.  The input buffers are allocated on line 30, and leaked on line 32.

It prints a progress bar as it works.  Once every 1000 calls, it prints how much memory it is using. 
 The output looks like this:

```
$ python3 leak.py 
clang-9 $(python3-config --cflags) -g -ggdb -shared -fPIC -Wno-unused-command-line-argument -o hlcopy$(python3-config --extension-suffix) hlcopy.py.cpp hlcopy.o $(python3-config --ldflags)
leaking memory:   0%|                                | 0/3001 [00:00<?, ?calls/s]
rss (bytes): 109780992
leaking memory:  32%|██████▊              | 973/3001 [00:02<00:05, 398.51calls/s]
rss (bytes): 3059380224
leaking memory:  65%|█████████████       | 1959/3001 [00:04<00:02, 441.93calls/s]
rss (bytes): 6008684544
leaking memory: 100%|███████████████████▉| 2995/3001 [00:07<00:00, 439.51calls/s]
rss (bytes): 8958259200
leaking memory: 100%|████████████████████| 3001/3001 [00:07<00:00, 426.85calls/s]
```

Every call adds about 3 gigs of memory usage.

abadams:
IIRC this was @mgharbi 's code. I'll tag him.

mgharbi:
Hum, @abadams this does not seem related to the PyTorch bindings. I am not very familiar with the pure Python bindings.

Infinoid:
Same test script, with the #4861 patch applied:

```
$ python3 leak.py 
[...]

PR #4861, opened 2020-04-17
Fix a leak in python extension wrapper code (Issue #4859)

Infinoid:
Add calls to PyBuffer_Release() where appropriate.
Add a buffer parameter to _convert_py_buffer_to_halide, so the caller can release it later.


steven-johnson:
LGTM, but could you add a before-and-after code snippet in the comments here to verify what the generated code looks like?

Infinoid:
https://gist.github.com/Infinoid/fee9a73a7958b2759e8f3bffb0ee8660 contains generated code before and after the patch, and a diff between the two.

This code was generated by the reproducer mentioned in #4859.

before: https://gist.github.com/Infinoid/fee9a73a7958b2759e8f3bffb0ee8660#file-1-hlcopy-py-cpp-before-patch
after: https://gist.github.com/Infinoid/fee9a73a7958b2759e8f3bffb0ee8660#file-2-hlcopy-py-cpp-after-patch
diff: https://gist.github.com/Infinoid/fee9a73a7958b2759e8f3bffb0ee8660#file-0-diff

PR #4864, opened 2020-04-19
Fix for llvm trunk

abadams:


src/LLVM_Output.cpp
```diff
@@ -156,7 +156,15 @@ void write_symbol_table(std::ostream &out,
         }
         llvm::object::SymbolicFile &obj = *obj_or_err.get();
         for (const auto &sym : obj.symbols()) {
+#if LLVM_VERSION >= 110
+            auto flags = sym.getFlags();
+            if (!flags) {
+                internal_error << llvm::toString(flags.takeError()) << "\n";
```
dsharletg, line 162: Should this really be an error, or does it just mean there are no flags? I don't understand why LLVM would make this "optional" if it's guaranteed to be an error if it isn't present.

abadams: Hrm. I guess I need to figure out what it means for this field to be an error.

abadams: It looks like it could be non-existent for ELF files if the section of the file that defines the metadata for the symbol is missing (.symtab), so it's a genuine error condition.

abadams: The LLVM implementation just has the helpful comment "TODO: Test this error.", so I'm not sure how much thought has even gone into this.

PR #4865, opened 2020-04-20
Add explicit casts to VectorType* in Codegen_Hexagon

steven-johnson:
Not breaking at current llvm-trunk, but breaking at the close-to-trunk version inside google, and from casual inspection it appears that a cast is approprate

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -1623,8 +1623,9 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
                 native_ty, MAKE_ID_PAIR(Intrinsic::hexagon_V6_hi).get(is_128B), {a});
             a = call_intrin_cast(
                 native_ty, MAKE_ID_PAIR(Intrinsic::hexagon_V6_lo).get(is_128B), {a});
-            a_ty = a->getType();
-            b_ty = b->getType();
+            a_ty = cast<llvm::VectorType>(a->getType());
+            b_ty = cast<llvm::VectorType>(b->getType());
+            internal_assert(a_ty && b_ty);
             a_elements = get_vector_num_elements(a_ty);
         }
         if (start == 0 && result_ty == a_ty) {


src/CodeGen_Hexagon.cpp
```diff
@@ -1623,8 +1623,9 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
                 native_ty, MAKE_ID_PAIR(Intrinsic::hexagon_V6_hi).get(is_128B), {a});
             a = call_intrin_cast(
                 native_ty, MAKE_ID_PAIR(Intrinsic::hexagon_V6_lo).get(is_128B), {a});
-            a_ty = a->getType();
-            b_ty = b->getType();
+            a_ty = cast<llvm::VectorType>(a->getType());
```
abadams, line 1626: I'm confused. a_ty is an llvm::Type * local variable. Why would it be necessary to cast it the temporary to VectorType?

steven-johnson: In our local LLVM, this fails to compile with `incompatible pointer types assigning to 'llvm::VectorType *' from 'llvm::Type *'`... but yeah, something is definitely wonky here. Let me figure out what is up internally.

steven-johnson:
Ugh, please ignore -- several (unreviewed) changes were made by the LLVM folk when they integrated and some of them were stupid. Let me go fix them and attempt to chastise them.

abadams:
When I made the get_vector_num_elements change I continued to use llvm::Type instead of just making them llvm::VectorType, because I wasn't sure this is never called on vectors of size 1, which are not llvm::VectorType. Some of our intrinsics code is expected to work in the degenerate case of vectors of size 1.

abadams:
I mean Halide vectors of size one map to llvm Scalars - you *can* have llvm vectors of size one and they're distinct from scalars, but we don't use em.

steven-johnson:
Yeah, the local changes made by LLVM folks to get our stuff compiling were... inconsistent, I'll deal with it.

PR #4869, opened 2020-04-23
Adjust simplifier rules so that they're provably cycle-free

abadams:
Several of these division rules weren't strength reducing - they created
more ops (or stronger ops) than they removed.

This commit tries just adjusting them to be strength-reducing. I'm a
little nervous of it because it means in (x*4 - y)/2 we no longer move
the x out of the numerator. Needs thorough testing.


steven-johnson:
No red flags inside google.

steven-johnson:
Ready to land?

PR #4870, opened 2020-04-23
Reenable introspection by default, and add support for the endbr64 in…

abadams:
…struction

Was accidentally deleted in b0b6a4a84c53b47a91a05b4487ef504fe1345e73

Also add handling for newer gccs. They now inject endbr64 instructions
at function entry.


steven-johnson:
Ready to land?

PR #4871, opened 2020-04-24
Add basic Compiler Logging engine to Halide compiler

steven-johnson:
Up to now, we've typically gathered metadata about Halide compilation on an ad-hoc basis (e.g. via debug() statements); this adds a Telemetry output to Halide to allow us to gather data in a more structured way, so that we can do batch analysis of things like Simplifier rules more easily.

This is modeled as a Generator output for now (though it could be used manually with the JIT if desired); the `telemetry` output is a JSON file that contains basic information about the Generator being driven, along with information about:
- Expressions that failed `can_prove` (including a text version of the original and failed version)
- Loop vars that are non-monotonic
- Eventually, the usage rate for each Simplifier rule (this requires some additional scaffolding in the Simplifier to enable)

The CMake build file was modified to emit telemetry files by default for all Generators (I did not attempt to modify the Makefile in this way, though).

For debugging purposes, you can use HL_DEBUG_TELEMETRY=1 to dump all Telemetry to stderr (whether or not telemetry output is specified for a specific Generator).

You can also optionally specify HL_TELEMETRY_ANONYMIZE=1, which will cause all emitted telemetry to try to scrub interesting identifiers away; this is intended to allow gathering data on proprietary code and contributing it to a public analyzer (e.g. to find missed simplifier opportunities).

The JSON looks something like this (**UPDATED**):

```
{
 "generator_name" : "interpolate",
 "function_name" : "interpolate",
 "target" : "x86-64-linux-avx-no_runtime-sse41",
 "generator_args" : "auto_schedule=false",
 "non_monotonic_loop_vars" : {
  "downsampled_1.s0.y.v10" :
  [
   "(((downsampled_1.s0.y.v10*-2) + ((t140.s*-2) + ((t140.s*2) + (select((0 < downsampled_1.s0.y.v10), -14, -15) + (downsampled_1.s0.y.v10*2)))))*-1)",
   "(((downsampled_1.s0.y.v10*-2) + ((t142.s*-2) + ((t142.s*2) + (select((0 < downsampled_1.s0.y.v10), -14, -15) + (downsampled_1.s0.y.v10*2)))))*-1)",
   "((downsampled_1.s0.y.v10*-2) + (select((0 < downsampled_1.s0.y.v10), -14, -15) + (downsampled_1.s0.y.v10*2)))"
  ],
  "downsampled_1.s0.y.v9" :
  [
   "(let t140.s = min(((downsampled_1.s0.y.v9*8) + 4), downsampled_1.s0.y.max) in (((downsampled_1.s0.y.v10*2) + ((t140.s*2) - (select((0 < downsampled_1.s0.y.v10), -14, -15) + ((downsampled_1.s0.y.v10 + t140.s)*2)))) + -14))",
   ...etc...
  ],
  ...etc...
 },
 "failed_to_prove" : {
  "(v0 <= (min(v0, 16) + (((max(v0, 16) + -1)/16)*16)))" :
  [
   "(let t2839 = min(input.extent.0, 16) in (let t2840 = (normalize.s0.y.v10 + normalize.s0.y.v10.base) in (((((((uint1)1 && ((0 + (t2839 + -16)) <= (t2839 + -16))) && ((15 + (min((t2839 + (((max(input.extent.0, 16) + -1)/16)*16)), input.extent.0) + -16)) >= (min((((input.extent.0 + -1)/16)*16), (input.extent.0 + -16)) + 15))) && (t2840 <= t2840)) && (t2840 >= t2840)) && (0 <= 0)) && (3 >= 3))))"
  ]
 },
 "version": "HalideJSONTelemetryV1"
}
```

If you specify HL_TELEMETRY_ANONYMIZE=1, this will instead look something like  (**UPDATED**):

```
{
 // names omitted
 "non_monotonic_loop_vars" : {
  "anon0" :
  [
   "(((anon0*-2) + ((anon1*-2) + ((anon1*2) + (select((0 < anon0), -14, -15) + (anon0*2)))))*-1)",
   "(((anon0*-2) + ((anon2*-2) + ((anon2*2) + (select((0 < anon0), -14, -15) + (anon0*2)))))*-1)",
   "((anon0*-2) + (select((0 < anon0), -14, -15) + (anon0*2)))"
  ],
  "anon11" :
  [
   "((((((select((0 < anon11), 1022, 0) + anon11) + anon12)/512)*-512) + anon12) + anon11)",
   ...etc...
  ],
  ...etc...
 },
 "failed_to_prove" : {
  "(anon58 <= (min(anon58, 16) + (((max(anon58, 16) + -1)/16)*16)))" :
  [
   "(let anon59 = min(anon60, 16) in (let anon61 = (anon11 + anon62) in (((((((uint1)1 && ((0 + (anon59 + -16)) <= (anon59 + -16))) && ((15 + (min((anon59 + (((max(anon60, 16) + -1)/16)*16)), anon60) + -16)) >= (min((((anon60 + -1)/16)*16), (anon60 + -16)) + 15))) && (anon61 <= anon61)) && (anon61 >= anon61)) && (0 <= 0)) && (3 >= 3))))"
  ]
 },
 "version": "HalideJSONTelemetryV1"
}
```

If generating a multitarget, the JSON will be an array of these objects (one per subtarget):

```
[
{
 "generator_name" : "interpolate",
 "target" : "x86-64-linux-avx-sse41",
 ...
},
{
 "generator_name" : "interpolate",
 "target" : "x86-64-linux-avx",
 ...
},
{
 "generator_name" : "interpolate",
 "target" : "x86-64-linux",
 ...
}
]
```

There are definitely more things that we want to add to this that aren't yet implemented; off the top of my head:
- Halide IR-building/lowering time
- LLVM codegen time
- total compilation time
- final code size

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -8,6 +8,7 @@
 
 #include <functional>
 #include <map>
+#include <memory>
 #include <set>
 #include <string>
 
@@ -29,6 +30,7 @@ enum class Output {
     bitcode,
     c_header,
     c_source,
+    compiler_log,
     cpp_stub,
     featurization,
     llvm_assembly,
@@ -109,7 +111,8 @@ struct LoweredFunc {
 
 namespace Internal {
 struct ModuleContents;
-}
+class CompilerLogger;
+}  // namespace Internal
 
 struct AutoSchedulerResults;
 
@@ -200,12 +203,14 @@ void compile_standalone_runtime(const std::string &object_filename, Target t);
  */
 std::map<Output, std::string> compile_standalone_runtime(const std::map<Output, std::string> &output_files, Target t);
 
-typedef std::function<Module(const std::string &, const Target &)> ModuleProducer;
+using ModuleFactory = std::function<Module(const std::string &fn_name, const Target &target)>;
+using CompilerLoggerFactory = std::function<std::unique_ptr<Internal::CompilerLogger>(const std::string &fn_name, const Target &target)>;
 
 void compile_multitarget(const std::string &fn_name,
                          const std::map<Output, std::string> &output_files,
                          const std::vector<Target> &targets,
-                         const ModuleProducer &module_producer);
+                         const ModuleFactory &module_factory,
+                         const CompilerLoggerFactory &compiler_logger_factory = nullptr);
 
 }  // namespace Halide
 
--- a/src/SimplifyCorrelatedDifferences.cpp
+++ b/src/SimplifyCorrelatedDifferences.cpp
@@ -1,6 +1,7 @@
 #include "SimplifyCorrelatedDifferences.h"
 
 #include "CSE.h"
+#include "CompilerLogger.h"
 #include "ExprUsesVar.h"
 #include "IRMatch.h"
 #include "IRMutator.h"
@@ -218,9 +219,13 @@ class SimplifyCorrelatedDifferences : public IRMutator {
             e = PartiallyCancelDifferences().mutate(e);
             e = simplify(e);
 
-            if ((debug::debug_level() > 0) &&
+            const bool check_non_monotonic = debug::debug_level() > 0 || get_compiler_logger() != nullptr;
+            if (check_non_monotonic &&
                 is_monotonic(e, loop_var) == Monotonic::Unknown) {
                 // Might be a missed simplification opportunity. Log to help improve the simplifier.
+                if (get_compiler_logger()) {
+                    get_compiler_logger()->record_non_monotonic_loop_var(loop_var, e);
+                }
                 debug(1) << "Warning: expression is non-monotonic in loop variable "
                          << loop_var << ": " << e << "\n";
             }
--- a/Makefile
+++ b/Makefile
@@ -476,6 +476,7 @@ SOURCE_FILES = \
   CodeGen_RISCV.cpp \
   CodeGen_WebAssembly.cpp \
   CodeGen_X86.cpp \
+  CompilerLogger.cpp \
   CPlusPlusMangle.cpp \
   CSE.cpp \
   Debug.cpp \
@@ -648,6 +649,7 @@ HEADER_FILES = \
   CodeGen_RISCV.h \
   CodeGen_WebAssembly.h \
   CodeGen_X86.h \
+  CompilerLogger.h \
   ConciseCasts.h \
   CPlusPlusMangle.h \
   CSE.h \
--- a/halide.cmake
+++ b/halide.cmake
@@ -159,7 +159,7 @@ function(halide_library_from_generator BASENAME)
     endif()
   endforeach()
 
-  set(OUTPUTS static_library c_header registration)
+  set(OUTPUTS static_library c_header registration compiler_log)
   foreach(E ${args_EXTRA_OUTPUTS})
     # Convert legacy aliases
     set(cpp_current "c_source")
@@ -220,6 +220,7 @@ function(halide_library_from_generator BASENAME)
   set(static_library_ext ${CMAKE_STATIC_LIBRARY_SUFFIX})
   set(stmt_ext ".stmt")
   set(stmt_html_ext ".stmt.html")
+  set(compiler_log_ext ".halide_compiler_log")
 
   set(OUTPUT_FILES )
   foreach(OUTPUT ${OUTPUTS})
--- a/python_bindings/src/PyEnums.cpp
+++ b/python_bindings/src/PyEnums.cpp
@@ -165,7 +165,8 @@ void define_enums(py::module &m) {
         .value("schedule", Output::schedule)
         .value("static_library", Output::static_library)
         .value("stmt", Output::stmt)
-        .value("stmt_html", Output::stmt_html);
+        .value("stmt_html", Output::stmt_html)
+        .value("compiler_log", Output::compiler_log);
 }
 
 }  // namespace PythonBindings
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -321,6 +321,7 @@ set(HEADER_FILES
   CodeGen_RISCV.h
   CodeGen_WebAssembly.h
   CodeGen_X86.h
+  CompilerLogger.h
   ConciseCasts.h
   CPlusPlusMangle.h
   CSE.h
@@ -502,6 +503,7 @@ add_library(Halide ${HALIDE_LIBRARY_TYPE}
   CodeGen_RISCV.cpp
   CodeGen_WebAssembly.cpp
   CodeGen_X86.cpp
+  CompilerLogger.cpp
   CPlusPlusMangle.cpp
   CSE.cpp
   Debug.cpp
--- a/src/CompilerLogger.cpp
+++ b/src/CompilerLogger.cpp
@@ -1,0 +1,326 @@
+#include "CompilerLogger.h"
+
+#include <algorithm>
+#include <fstream>
+#include <iostream>
+#include <vector>
+
+#include "IRMutator.h"
+#include "Util.h"
+
+namespace Halide {
+namespace Internal {
+
+namespace {
+
+// TODO: for now we are just going to ignore potential issues with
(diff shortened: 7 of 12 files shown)


src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -218,9 +219,15 @@ class SimplifyCorrelatedDifferences : public IRMutator {
...
 
-            if ((debug::debug_level() > 0) &&
+            const bool check_non_monotonic = debug::debug_level() > 0 || get_telemetry() != nullptr;
+            if (check_non_monotonic &&
                 is_monotonic(e, loop_var) == Monotonic::Unknown) {
                 // Might be a missed simplification opportunity. Log to help improve the simplifier.
+                if (get_telemetry()) {
+                    get_telemetry()->record_non_monotonic_loop_var(loop_var, e);
+                }
+                // TODO: should we remove the debug statement, in favor of
```
abadams, line 229: The debug statements are still useful because they tell you *when* during compilation the non-monotonic expr was found. E.g. just before allocation bounds inference is more interesting than after.

src/Telemetry.cpp
```diff
@@ -0,0 +1,243 @@
...
+    using IRMutator::visit;
+
+    std::map<std::string, std::string> remapping;
+
+    // TODO: is this the right way to anonymize extern calls and image input names?
+    Expr visit(const Call *op) override {
+        if (op->call_type == Call::Extern || op->call_type == Call::ExternCPlusPlus) {
+            return Variable::make(op->type, remap(op->name, "define_extern_"));
+        } else if (op->call_type == Call::Image) {
+            return Variable::make(op->type, remap(op->name, "image"));
```
abadams, line 30: I think this might confuse analysis to map calls to variables. Better to leave them as calls but anonymize the name.

src/Telemetry.cpp
```diff
@@ -0,0 +1,243 @@
...
+        } else {
+            return op;
+        }
+    }
+
+    Expr visit(const Variable *op) override {
+        return Variable::make(op->type, remap_var_name(op->name));
+    }
+
+    std::string remap(const std::string &var_name, const char *replacement) {
```
abadams, line 40: I think anonymizing the names of let variables would be a good idea too, because then we get canonicalization of the inner lets and hence better deduplication.

src/Telemetry.h
```diff
@@ -0,0 +1,100 @@
+#ifndef HALIDE_TELEMETRY_H_
+#define HALIDE_TELEMETRY_H_
+
+/** \file
+ * Defines an interface used to gather compile-time information, stats, etc
+ * for use in evaluating internal Halide compilation rules and efficiency.
+ *
+ * There is a 'standard' implementation that simply logs information to
+ * a stream, but the entire implementation can be replaced to log in arbitrary ways.
+ */
```
abadams, line 10: "Telemetry" has some spooky connotations, so we should make it super clear here that this does nothing unless you ask for the telemetry output in a generator, in which case it writes it to a file. In no circumstances does it ever send any information about your code to anyone else, e.g. over the internet.

src/Telemetry.h
```diff
@@ -0,0 +1,100 @@
...
+    const std::string function_name;
+    const std::string autoscheduler_name;
+    const Target target;
+    const std::string generator_args;
+    const bool anonymize_exprs{false};
+
+    std::map<std::string, int64_t> matched_simplifier_rules;
+    // TODO: are we likely to see enough duplicates here to make
+    // it worth considering a set instead of a vector?
+    std::vector<std::pair<std::string, Expr>> non_monotonic_loop_vars;
```
abadams, line 90: Yeah, the same ones will be encountered multiple times during lowering.

src/SimplifyCorrelatedDifferences.cpp
steven-johnson: Will remove the TODO

src/Telemetry.cpp
steven-johnson: Ack, I intended to do so, how embarassing

src/Telemetry.cpp
steven-johnson: SGTM

src/Telemetry.h
steven-johnson: SGTM

src/Telemetry.h
steven-johnson: will fix then

steven-johnson:
PTAL

steven-johnson:
After more testing, I decided to revise the way that `non_monotonic_loop_vars` and `failed_to_prove` are stored and output; previously, we could have duplicated data, which (aside from being wasteful) is technically illegal, as valid JSON isn't allowed to have duplicate keys in an object; even if a forgiving JSON parser accepted it, we could have had values dropped on the floor from merging.

steven-johnson:
PTAL in particular at the failed_to_prove code; this is now a map from "expr passed to can_prove()" -> "set of exprs that still could not be proven"; I'm honestly not sure if it might be useful to invert this instead (i.e., map from "unprovable expr" -> "set of original exprs that reduced to this".)

src/Module.h
```diff
@@ -187,25 +190,24 @@ class Module {
...
+using TelemetryFactory = std::function<std::unique_ptr<Internal::Telemetry>(const std::string &fn_name, const Target &target)>;
 
 /** Create an object and/or static library file containing the Halide runtime
  * for a given target. For use with Target::NoRuntime. Standalone runtimes are
  * only compatible with pipelines compiled by the same build of Halide used to
  * call this function. Return a map with just the actual outputs filled in
  * (typically, Output::object and/or Output::static_library).
  */
-std::map<Output, std::string> compile_standalone_runtime(const std::map<Output, std::string> &output_files, Target t);
+std::map<Output, std::string> compile_standalone_runtime(const std::map<Output, std::string> &output_files, Target t,
```
abadams, line 201: I'm confused by this signature change. I don't think it's legal to compile a standalone runtime to anything other than a single object file, is it? Also telemetry doesn't seem like it makes any sense for a standalone runtime. If either overload was deleted I would think it would be this one. We should at least preserve the existing one - IMO it's the one with the signature that make sense for users.

steven-johnson: It's ~always been legal to compile a runtime into static_library form too.

That said, removing this overload is probably more sensible to do separately; it was done here to test the telemetry for runtimes, which (as you mentioned) doesn't make a lot of sense. I'll drop those changes.

abadams: Oh right, static library too.

steven-johnson: Removed the runtime changes and Telemetry, also rebased to master

src/Telemetry.h
alexreinking: "compiler tracing" or "bug report helper" or something

steven-johnson: @alexreinking not sure where you are suggesting those phrases be inserted

alexreinking: Oh, sorry, I was just suggesting some was to rephrase "telemetry" without the aforementioned spooky connotations. It would be worth explaining the feature in the Generator help text, now that you mention it.

abadams:
sliding window change LGTM

src/Telemetry.h
dsharletg: I wonder if this would be better named simply something to do with logging, (and we could consider routing more of our logging/error reporting through it later) rather than the spooky sounding telemetry.

Similarly, "anonymize" might be better expressed as "obfuscate".

steven-johnson: I'm fine with whatever nomenclature we settle on -- this was the first name that sprang to mind. (spookiness didn't occur to me but since multiple other people say it does, it does :-)

Something along the lines of "CompilerLogger" maybe? (Not very terse but maybe less spooky sounding I guess.) Other specific suggestions before this thing lands?

Re: 'Obfuscate', SGTM.
-

steven-johnson: Update: going with `CompilerLogger` for now.

steven-johnson:
Renaming party + rebase to master.

Issue #4616, opened 2020-02-18
Support for multi-dim vectorization


KnowingNothing:
I want to port Halide to a new hardware backend. As far as I know, Halide only supports 1-D vectorization for now. Is there any support for multi-dim vectorization or is there any plan to support this? It is a valuable feature as there are more and more special accelerators that use tensorization to gain high performance. I'd be grateful if anyone can provide some useful information.

abadams:
This is a very interesting question. We can currently vectorize multiple loops, but only using the Func::fuse directive to turn them into a single loop first. This gives you vector values that encode multi-dimensional ramps in awkward ways, e.g:

Broadcast of a ramp: Ramp(0, 1, 32) % 4 
Ramp of a broadcast: Ramp(0, 1, 32) / 4

A 2D ramp (e.g. the index when loading a dense 2D tile of a matrix) would be the sum of two things that look like that. That's the sort of IR you get if you vectorize multiple dimensions by first fusing them. Backends that support things like 2D loads could just peephole recognize those patterns. Operations that do meaningfully 2D things (e.g. take a 2D crop or transpose of a 2D vector) can use ShuffleVector to get the lanes it needs. Partial reductions across the vector (e.g. a horizontal add) are coming in my atomic_vectorization branch. With that partial reduction node, an entire matrix multiply of two vectors that are actually 4x4 matrices would look something like:
```
let a_idx = ramp(0, 1, 64) % 4 + ramp(0, 1, 64) / 16
let b_idx = ramp(0, 1, 64) % 16
let c = horizontal_add(a[a_idx] * b[b_idx], 4);
```

Now I think we could make things a little cleaner if we allowed ramps and broadcasts to take vector args, instead of requiring that they be scalar, with the semantics that you produce a multi-dimensional vector and then linearize it with the args to the broadcast or ramp innermost. So a broadcast of a vector repeats that entire vector some number of times. Then my example would be:

```
let a_idx = ramp(broadcast(ramp(0, 1, 4), 4), broadcast(1, 4), 4)
let b_idx = broadcast(ramp(ramp(0, 1, 4), broadcast(4, 4), 4), 4)
let c = horizontal_add(a[a_idx] * b[b_idx], 4);
```
which is maybe easier on the backends? The multi-dimensional structure is more clearly expressed in the index as ramp of broadcast of ramp, or broadcast of ramp of ramp. A backend that wants to peephole optimize a dense 2D load could look for a load of a ramp that has ramp base and broadcast stride.

[...]

PR #4873, opened 2020-04-24
Support for multi-dim vectorization

vksnk:
This PR adds support for multi-dimensional vectorization as described in #4616.

There are two main pieces here:
 
* nested ramps/broadcasts:
   * Ramp and Broadcast can be nested now, i.e. it is allowed for them to have a vector base, stride and value;
   * updates simplifier rules to account for that + extended fuzz_simplify to test it.
   * changes to Deinterleave, so it supports extract_lane and others on nested vector expressions;
   * smaller changes in some other place.
* actual multi-dimensional vectorization:
   * this is the place where nested ramps/broadcasts can be introduced;
   * I have added a number of tests, which exercise most of the code branches inside of the VectorizeLoops to ensure it works correctly; 
   * extra pass in the end of the lowering which flattens nested ramps/broadcasts, so existing codegens can work without changes

Notes/questions:
* I updated every simplifier rule which involves ramp/broadcast, such that they are only applied when types of the arguments are matching. For some rules, which explicitly use lanes (for example, trying to prove ranges of the ramps), I've added an extra check if argument is scalar to make sure it's only applied in the case when lanes == type.lanes(). I think in order to make it general, we'd need to turn lanes into wildcard.
* In some cases lanes argument is passed to the result of the rewriting rule, even though it's not present in the input. I've removed it and it works everywhere except Simplify_LT, I don't quite understand why, so would greatly appreciate any suggestions.
* ~~fuzz_simplify is failing for the type UInt(1) when vector width is more than 2 even before the changes, so I've disabled it for now. The error message is `"Integer constant 2 will be implicitly coerced to type uint1, which changes its value to (uint1)0.`. I am going to look into it next, but figured I ask in case this is known or anyone has any suggestions.~~
* I am not sure how to trigger VectorizeLoops::scalarize(Expr), so put an assert if it's called from nested vectorization.

Otherwise, I think, it should be ready for review. Thanks!


test/correctness/fuzz_simplify.cpp
```diff
@@ -242,61 +255,6 @@ bool test_expression(Expr test, int samples) {
     return true;
 }
 
-Expr ramp(Expr b, Expr s, int w) {
```
dsharletg, line 245: FYI, these were here to make it possible to copy and paste failing expressions from failure outputs into the test to debug them (in place of randomly generated expressions). 

(There should have been a comment on them to explain this...)

vksnk: Ah, that's why! Sorry about that, restored them back.

dsharletg: Can you also add a comment that explains them?

// These are here to enable copy failed output expressions and pasting them into the test for debugging.

vksnk: Done

test/correctness/fuzz_simplify.cpp
```diff
@@ -36,6 +37,21 @@ Type random_type(int width) {
     return T;
 }
 
+int get_random_divisor(Type t) {
+    // extract_lane won't work on the ramp of boolean type with width > 2.
```
dsharletg, line 41: This seems like a bug if I'm interpreting this comment correctly. Should it have an issue?

test/correctness/fuzz_simplify.cpp
```diff
@@ -319,11 +339,11 @@ int main(int argc, char **argv) {
     rng.seed(fuzz_seed);
     std::cout << "Simplify fuzz test seed: " << fuzz_seed << "\n";
 
-    int max_fuzz_vector_width = 4;
+    int max_fuzz_vector_width = 8;
 
     for (int i = 0; i < fuzz_type_count; i++) {
         Type T = fuzz_types[i];
-        for (int w = 1; w < max_fuzz_vector_width; w *= 2) {
+        for (int w = 1; w <= max_fuzz_vector_width; w *= 2) {
```
dsharletg, line 346: I think the test as-is will only test power-of-2 width vectors, but we do support other vectorization widths and it might be good to test those with the new ramp-of-ramps logic. 6 would be a good one (to get 3 x 2), so maybe this should just be a loop like: `for (int w : {1, 2, 3, 4, 6, 8})`. Or maybe we should just use random vector widths with a large number of trials.

IIRC the run time of this test was an issue, which is why I'm guessing we tried to minimize the "branching factor" here.

test/correctness/vectorize_nested.cpp
```diff
@@ -0,0 +1,261 @@
...
+using namespace Halide;
+
+int vectorize_2d_round_up() {
+    const int width = 32;
+    const int height = 24;
+
+    Func f("f");
+    Var x("x"), y("y"), xi("xi"), yi("yi");
+
+    f(x, y) = x + y;
```
dsharletg, line 13: It might be good to make this function x + 3 * y or something so bugs that swap x and y are visible to the test (same below).

vksnk: Done

test/correctness/fuzz_simplify.cpp
vksnk: I've changed it to randomly pick a vector width (and type) and included 3 and 6 in the list. Also, increased number of trials.

src/Lower.cpp
```diff
@@ -420,6 +421,14 @@ Module lower(const vector<Function> &output_funcs,
     s = remove_dead_allocations(s);
```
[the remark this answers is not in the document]
vksnk: Done.

src/FlattenNestedRamps.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    Expr visit(const Broadcast *op) override {
+        if (op->value.type().is_vector()) {
+            Expr value = mutate(op->value);
+            std::vector<Expr> broadcast_elems;
+            for (int ix = 0; ix < op->lanes; ix++) {
+                broadcast_elems.push_back(value);
+            }
+
+            return Shuffle::make_concat(broadcast_elems);
```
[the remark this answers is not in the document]
vksnk: Sorry if I misunderstood your suggestion, but value can be a vector, so I think `std::vector<int>(op->lanes, 0)` won't work. However, value is indeed repeating in the broadcast, so we can use that with something like Shuffle::make({value}, [vector with numbers from 0 to value->lanes] repeated op->lanes times), which might be more efficient than concat.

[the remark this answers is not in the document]
dsharletg: `std::vector<int>(op->lanes, 0)` is a vector of 0s with length op->lanes, i.e. this implements what you are suggesting :)

I think it would also be good to add a `Shuffle::make_broadcast` helper that does this.

test/correctness/fuzz_simplify.cpp
vksnk: I think the problem here is expression like ramp((uint1)a, (uint1)b, 4), for which one of the components will be (uint1)a + 2 * (uint1). This multiplication by 2 seems to be problematic and triggers `Integer constant 2 will be implicitly coerced to type uint1, which changes its value to (uint1)0.`error. I assumed that this is expected behaviour, but if not I would definitely file a bug.

vksnk:
@dsharletg Thank you for the review! PTAL.

test/correctness/fuzz_simplify.cpp
dsharletg: Oh, this makes sense, just the comment is a bit confusing.

It does seem like there's a bit of an issue here. It seems like either ramps of uint1 should be OK (because uint overflow is well defined), and then `extract_lane` should work. Or ramps of uint1 are not OK, and in that case, `extract_lane` is not broken (the comment is confusing).

But I'm not sure it's something we should spend time on (seems like a case we don't ever actually hit)

dsharletg: Maybe the easiest thing to do is actually just try fixing `extract_lane` so it doesn't generate this error (i.e. explicitly implement overflow for unsigned types). If that works, then there is no special case/confusing comment required.

vksnk: Hm, it was enough to put explicit cast. Also, removed special handling of uint1 from fuzz_simplify.

src/FlattenNestedRamps.cpp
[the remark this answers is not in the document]
vksnk: Right, this will indeed work for scalar value, but not for vector :)

I've added Shuffle::make_broadcast.

vksnk:
clang-tidy is complaining that "the parameter 'vector' is copied for each invocation but only used as a const reference; consider making it a const reference" (https://github.com/halide/Halide/runs/627872519), however pretty much every function in this file is passing Expr as a value. Should I make it const Expr& only in the new function (probably, no)? Everywhere? Any other way to pacify clang-tidy?

abadams:
The other functions in the file std::move their Expr args to avoid a copy.

vksnk:
> The other functions in the file std::move their Expr args to avoid a copy.

Ah, that makes sense. Thank you!

steven-johnson:
This needs to be merged with master to resolve some conflicts. How close are we to being ready to land it?

vksnk:
> This needs to be merged with master to resolve some conflicts. How close are we to being ready to land it?

Merged with master. 

It's not ready to land yet. I've addressed Dillon's comments so far, but based on offline discussion VectorizeLoops.cpp hasn't been reviewed yet and this is a biggest change in this PR. There are also some questions regarding simplifier changes and how to trigger VectorizeLoops::scalarize(Expr).

joesavage:
In terms of actually generating tensor operations from this, is the expectation that target-specific back-ends will peephole optimise over the large combination of broadcasts, ramps, and shuffles that get generated from ramp flattening, or is there some larger story for how this should work?

If we look at trying to generate something like NEON's `UMMLA`, for instance, in the simplest case we're probably talking about matching against something like this:

```c
AB[(shuffle(concat_vectors(x8(0), x8(1)), 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) + (concat_vectors(x16(0), x16(1)) * x32(AB.stride.1)))] +=
	(uint32x32)(A_inp_u8[shuffle(shuffle(ramp(0, 1, 8), 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7), 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) + concat_vectors(x16(0), x16(1)) * x32(A_inp_u8.stride.1)] *
	            B_trn_u8[shuffle(shuffle(ramp(0, 1, 8), 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7), 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) +
	                     shuffle(concat_vectors(x8(0), x8(1)), 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) * x32(B_trn_u8.stride.1)])
```

Or with `VectorReduce` from #4628, maybe the slightly simpler:

```c
AB[x2(ramp(0, 1, 2)) + concat_vectors(x2(0), x2(1)) * x4(AB.stride.1)] += (uint32x4)vector_reduce(Add, (uint32x32)<RHS_FROM_ABOVE>)
```

Just wondering whether there might be a nicer way for back-ends to match against this stuff. (Perhaps that constitutes a separate unit of work, but figured it was worth asking anyway.)

vksnk:
Thank you, this is very useful feedback!

> In terms of actually generating tensor operations from this, is the expectation that target-specific back-ends will peephole optimise over the large combination of broadcasts, ramps, and shuffles that get generated from ramp flattening, or is there some larger story for how this should work?

Yes, this is correct, expectation is that we can do peephole optimizations on these. I think one of the important changes here is support of nested ramps and broadcasts, which hopefully will make pattern matching of such cases a little bit easier. 

That being said, in this PR there is an extra lowering pass which will flatten all of the nested ramps in the end of lowering. This is sort of intermediate step and is only there, so we don't need to teach all codegens separately about nested ramps at once (to keep already large PR from getting even larger). Later on, each codegen can attempt to recognize these multi-dimensional vector patterns first instead and only apply flattening if it failed to match the pattern.

> Or with `VectorReduce` from #4628, maybe the slightly simpler:
> 
> ```c
> AB[x2(ramp(0, 1, 2)) + concat_vectors(x2(0), x2(1)) * x4(AB.stride.1)] += (uint32x4)vector_reduce(Add, (uint32x32)<RHS_FROM_ABOVE>)
> ```

I agree with you that it can be much more useful when VectorReduce lands. However, in this specific case, before the ramp flattening pass, I think it will generate something like:

`AB[x2(ramp(0,1,2)) + ramp(x2(0), x2(1), 2) * x4(stride)] = `

which, if I am not mistaken, simplifier can (or at least should be able to) fold it into one expression, something like:

`AB[ramp(ramp(0,1,2), x2(1) * x4(stride), 2)] = `

which I think is not too horrible to match.

(I realize that I am only addressing left part of the expression, which is much smaller and simpler compared to RHS. I think I need to understand this instruction and it's pattern better before commenting about it. It looks like a great test case for this change though, I definitely will get back with the update. Sorry about that :) )

Thanks again for the feedback, any suggestions on how to make this better are very welcome!

abadams:
Yes, I think vector reduce + not flattening nested ramps until later will make this tractable

src/VectorizeLoops.cpp
```diff
@@ -364,22 +377,26 @@ class PredicateLoadStore : public IRMutator {
...
+    Expr min;
+    int lanes;
+};
+
 // Substitutes a vector for a scalar var in a Stmt. Used on the
 // body of every vectorized loop.
 class VectorSubs : public IRMutator {
-    // The var we're vectorizing
-    string var;
+    std::vector<VectorizedVar> vectorized_vars;
```
dsharletg, line 389: When reading the code below, it seems the order of this vector is significant. Can you add a comment explaining this ordering?

src/VectorizeLoops.cpp
```diff
@@ -1019,10 +1127,9 @@ class VectorizeLoops : public IRMutator {
                            << "constant extent > 1\n";
             }
 
-            // Replace the var with a ramp within the body
-            Expr for_var = Variable::make(Int(32), for_loop->name);
-            Expr replacement = Ramp::make(for_loop->min, 1, extent->value);
-            stmt = VectorSubs(for_loop->name, replacement, in_hexagon, target).mutate(for_loop->body);
+            vectorized_vars.push_back({for_loop->name, for_loop->min, (int)extent->value});
+            stmt = VectorSubs(vectorized_vars, in_hexagon, target).mutate(for_loop->body);
+            vectorized_vars.clear();
```
dsharletg, line 1132: Shouldn't this be a `pop_back()` ?

test/correctness/vectorize_nested.cpp
```diff
@@ -0,0 +1,261 @@
...
+    f(x, y) = inlined(x) + 2 * y;
+
+    f.compute_root()
+        .tile(x, y, x, y, xi, yi, 8, 4, TailStrategy::GuardWithIf)
+        .vectorize(xi)
+        .vectorize(yi);
+
+    Buffer<int> result = f.realize(width, height);
+
+    // for (int iy = 0; iy < height; iy++) {
```
dsharletg, line 76: Instead of adding commented code (4x), I think we should add a `print_image` function (not in this PR) alongside `check_image`, so adding this check is a one liner.

src/VectorizeLoops.cpp
vksnk: I've added comments here and in some other places in this file.

test/correctness/vectorize_nested.cpp
vksnk: Removed debugging code for now, but +1 to print_image function - I'll add it in separate PR.

src/Lower.cpp
```diff
@@ -421,9 +422,18 @@ Module lower(const vector<Function> &output_funcs,
     debug(2) << "Lowering after lowering unsafe promises:\n"
              << s << "\n\n";
 
+    debug(1) << "Removing dead allocations and moving loop invariant code...\n";
     s = remove_dead_allocations(s);
     s = simplify(s);
     s = loop_invariant_code_motion(s);
+
```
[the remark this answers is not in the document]
vksnk: Done

src/VectorizeLoops.cpp
vksnk: This specific vector never had more than var in it, so clear() worked, however, vector actually is not needed here, so I just removed it.

vksnk:
> Is there a missing test that exercises this logic?

No, I don't think that was a bug, clear() was also correct, because that vector had at most one var in it. I removed vector itself in the follow-up commit.

vksnk:
@abadams this is overlaps quite a bit with the atomic vectorization PR (#4628), so some merge conflicts may need to be resolved. Should I wait until it lands or is it ok to merge this PR first? Either way works for me.

abadams:
#4628 is done and has been sitting ready for more review for a while now. I'd like to get it in asap so that we can emit udot/sdot and friends, so I'd rather get it in before this PR. Would you mind reviewing it? You've had the most recent experience in that code.

vksnk:
> #4628 is done and has been sitting ready for more review for a while now. I'd like to get it in asap so that we can emit udot/sdot and friends, so I'd rather get it in before this PR. Would you mind reviewing it? You've had the most recent experience in that code.

Sounds good. Sure, I'll review #4628.

steven-johnson:
Where do we stand on this -- are we still pending other PRs?

vksnk:
I already merged in within-vector reductions PR, so it should be good as soon as tests failing on build bots are fixed. I was going to take a look at it today.

steven-johnson:
Cool -- no rush, just checking :-)

src/Deinterleave.cpp
```diff
@@ -228,7 +245,27 @@ class Deinterleaver : public IRGraphMutator {
...
 
     Expr visit(const Ramp *op) override {
-        Expr expr = op->base + starting_lane * op->stride;
+        int base_lanes = op->base.type().lanes();
+        if (base_lanes > 1) {
+            if (new_lanes == 1) {
+                int index = starting_lane / base_lanes;
+                Expr expr = op->base + cast(op->base.type(), index) * op->stride;
+
+                int old_starting_lane = starting_lane;
```
abadams, line 254: using ScopedValue could be cleaner here

src/VectorizeLoops.cpp
```diff
@@ -191,7 +191,7 @@ struct InterleavedRamp {
     int lanes, repetitions;
 };
 
-bool is_interleaved_ramp(const Expr &e, const Scope<Expr> &scope, InterleavedRamp *result) {
+bool is_interleaved_ramp(const Expr &e, const std::map<string, Expr> &widened_vars, InterleavedRamp *result) {
```
abadams, line 194: Why was this changed from a Scope to a map?

src/VectorizeLoops.cpp
```diff
@@ -563,6 +605,12 @@ class VectorSubs : public IRMutator {
         return mutate_binary_operator(op);
     }
 
+    Expr visit(const Ramp *op) override {
```
abadams, line 608: This looks the same as the base-class implementation

vksnk: Removed.

src/Deinterleave.cpp
vksnk: Updated, it indeed looks much cleaner now, thank you for suggestion!

src/VectorizeLoops.cpp
vksnk: For the case when there are multiple vectorized loops, it's possible that the same let variable may be vectorized differently depending on the loop level it's used and the exact form may not be known when we first see it (for example, if it's used inside of the inner vectorized loop), so map is used to keep track of all possible variant which were used.

abadams: I'm confused because a Scope is a map where you can multiple values per key, and the most recently added value for the key hides the inner ones. So a Scope should be strictly more general than a map.

vksnk: Huh, I guess I misunderstood the way it works; based on name and some usages I saw, I assumed that it's more like named stack and push/pop should go in last-in-first-out order, but in this case push and pop may happen in different functions (and maybe in different orders with other lets), so I thought it won't work correctly. However, if it's map of stacks instead, I guess this shouldn't be an issue. I can change map to scope and track widened names for Let/LetStmt separately, would this be preferable?

src/VectorizeLoops.cpp
```diff
@@ -1015,19 +1087,30 @@ class VectorSubs : public IRMutator {
...
-        scope.pop(v);
+        for (const auto &v : vs) {
+            scope.pop(v);
+        }
 
         // Replace the widened 'v' with the actual ramp
-        // foo[x*lanes + widened_v] -> foo[x*lanes + ramp(0, 1, lanes)]
-        body = substitute(v + widening_suffix, Ramp::make(0, 1, lanes), body);
+        // foo[x*lanes + widened_v] -> foo[x*lanes + ramp(...)]
+        for (size_t ix = 0; ix < vectorized_vars.size(); ix++) {
```
abadams, line 1105: Why do the widened vars get substituted in at this point? Haven't references to them already been replaced when visiting them?

src/VectorizeLoops.cpp
abadams: The reason to use a Scope instead is if the same name might show up multiple times in a way that a Scope would just handle correctly. Currently vectorization just happens on code with unique var names, but maybe it would be a useful to someone to be able to run it on stmts that don't obey that constraint. But if you're pushing/popping in non-stack order, then maybe that wouldn't work transparently. I need to figure out why you're pushing/popping these in non-stack order though to make sure I understand this code. I added a question about it elsewhere.

vksnk: Thanks for the explanation, I agree this is a good reason to use Scope! The reason I did it this way is because widened value of Let may change in the case when there is a vectorized loop inside of the Let body and this Let is used in there, so instead of mutating it in advance, it's mutated (sort of) lazily only when needed inside of the visit(const Variable *op) (see get_mutated_var_from_scope() function) to make sure that we've seen all of the vectorized loops. Maybe, as an alternative, it would be better to update all variables in scope every time we encounter a new vectorized loop, replace mutated variables in the scope, mutate loop body, and on return from the mutate(body), create widened variables if needed. That might be quite a bit of a change, but sounds like it's worth doing. What do you think?

src/VectorizeLoops.cpp
vksnk: What is replaced here is not one of the vars from Let/LetStmt, but a variable which was created in this function to sort of "vectorize" allocation. They are not strictly "widened", but they also are added to scope and such, and overall are quite similar to true widened vars, so I reused infrastructure and kept them in the same map.

steven-johnson:
Periodic ping to see where this PR stands

abadams:
I believe it's stalled on me continuing my review. Sorry for the delay, I'm currently overloaded with things I have told people I will do.

steven-johnson:
No worries, just making sure there isn't stuff that needed something trivial to unblock :-)

src/VectorizeLoops.cpp
vksnk: I've updated it to use Scope again using approach described above, and I think the code is better now (there are fewer changes for sure), thanks for the the suggestions!

src/VectorizeLoops.cpp
vksnk: Update: this is handled differently now, vars which were are added inside of the visit(Allocate* ) node are just directly replaced with proper combination of ramps/broadcasts whenever we encounter them in visit(Var* ).

vksnk:
All tests are green again.

src/Lower.cpp
```diff
@@ -432,9 +433,17 @@ Module lower(const vector<Function> &output_funcs,
...
+    debug(1) << "Removing dead allocations and moving loop invariant code...\n";
     s = remove_dead_allocations(s);
     s = simplify(s);
     s = hoist_loop_invariant_values(s);
+    debug(2) << "Lowering after removing dead allocations and hoisting loop invariant values:\n"
+             << s << "\n\n";
+
+    debug(1) << "Flattening nested ramps...\n";
+    s = flatten_nested_ramps(s);
+    s = simplify(s);
```
abadams, line 445: We currently intentionally don't simplify after the last LICM pass. Please move this lowering pass up a little if it needs simplification after it.

src/VectorizeLoops.cpp
```diff
@@ -327,6 +327,19 @@ bool uses_gpu_vars(const Expr &s) {
     return uses.uses_gpu;
 }
 
+class SerializeLoops : public IRMutator {
+    using IRMutator::visit;
+
+    Stmt visit(const For *op) override {
+        if (op->for_type == ForType::Vectorized) {
+            return For::make(op->name, mutate(op->min), mutate(op->extent),
```
abadams, line 335: Why would you need to mutate the min and extent here?

abadams:
lgtm apart from my two minor comments.

abadams (approved):

vksnk:
Thank you for reviewing!

PR #4874, opened 2020-04-24
no_asserts was hiding genuine bounds errors

abadams:
This PR turns it back on, except for the L1 blas generators where it
makes a performance difference, and changes the set of sizes to ones
that are actually compatible with the schedule used.


steven-johnson:
Muahahaha! ~~Trogdor~~ **no_asserts** strikes again!

steven-johnson:
More seriously: I realize there are cases where we need no_asserts, but this isn't the first time this has bitten us in exactly this way (either in GitHub or inside Google). 

Inside Google, I'm tempted to add some special-purpose code to elide `no_asserts` for non-opt builds. Not sure if there is an equivalent approach here.

abadams:
Yeah for real. The issue was that the original author was very careful to make these work for all possible sizes, and shaving some cycles with no_asserts made sense. Later I rescheduled some things without providing the same guarantees, so now no_asserts is a problem.

steven-johnson:
> shaving some cycles with no_asserts made sense

Is there any more shaving to be done in our assert implementation (ie, can we give less reason for people to turn them off)?

abadams:
Hrm, we have some optimizations for when there are lots inputs and hence lots of asserts, but I'm not sure if the codegen could be better for the case where there are a small number of inputs. I'll take a look and see if I see anything obvious. I doubt it though - it's just a bunch of struct unpacking and comparisons. LLVM should be good at it.

steven-johnson:
Looks like there is (still) an OOB access for linux-64-gcc53

steven-johnson:
Interestingly, dgemv_notrans fails depending on the target: `HL_TARGET=x86-64-linux-sse41` fails with `Testing dgemv_notrans: Error: Input buffer A is accessed at 1279, which is beyond the max (767) in dimension 0`, but adding `avx` (or later) to the target seems to make it pass.

alexreinking:
Does this fix the problem / is it ready to merge?

apps/linear_algebra/tests/test_halide_blas.cpp
```diff
@@ -303,7 +303,7 @@ int main(int argc, char *argv[]) {
             d.run_tests(size);
         }
     } else {
-        int size = 64 * 7;
+        int size = 768;
```
alexreinking, line 306: 64*7 = 448... what happened here? This is now crashing sdot on Windows for me. I get:
```
Testing sdot: Error: Output buffer result is accessed at 0, which is beyond the max (-1) in dimension 0
```

steven-johnson: ...sounds like we still have a bug. :-) Dunno why Windows would be special though. What is get_host_target*(?

PR #4875, opened 2020-04-27
Don't inject min/max exprs in bounds of a promise_clamped of constant

abadams:
No effect in public Halide, but I hear it causes at least one regression
inside Google. The reason for injecting the min/max exprs in the case
where the first arg doesn't vary has since gone away now that we push
producers inside the guard-with-if if statement of a consumer.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1080,6 +1080,23 @@ private:
             Interval upper = interval;
             op->args[0].accept(this);
 
+            if (op->is_intrinsic(Call::promise_clamped) &&
+                interval.is_single_point()) {
+                // It's not safe to lift a promise_clamped
+                // intrinsic. They make a claim that holds true at
+                // that specific point in the IR. Just drop the
+                // annotation and return the bounds of the first arg.
+                return;
+            }
+
+            if (op->is_intrinsic(Call::unsafe_promise_clamped) &&
+                interval.is_single_point(op->args[0]) &&
+                lower.is_single_point(op->args[1]) &&
+                upper.is_single_point(op->args[2])) {
+                // unsafe_promise_clamp holds globally, so we can lift the entire intrinsic.
+                interval = Interval::single_point(op);
+            }
+
             interval.min = Interval::make_max(interval.min, lower.min);
             interval.max = Interval::make_min(interval.max, upper.max);
         } else if (op->is_intrinsic(Call::likely) ||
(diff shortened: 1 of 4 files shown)


test/correctness/CMakeLists.txt
```diff
@@ -271,6 +271,7 @@ tests(GROUPS correctness travis
         skip_stages_external_array_functions.cpp
         skip_stages_memoize.cpp
         sliding_backwards.cpp
+        slicing_over_guard_with_if.cpp
```
dsharletg, line 274: sliding

src/Bounds.cpp
```diff
@@ -1080,6 +1080,23 @@ class Bounds : public IRVisitor {
             Interval upper = interval;
             op->args[0].accept(this);
 
+            if (op->is_intrinsic(Call::promise_clamped) &&
+                interval.is_single_point()) {
+                // It's not safe to lift a promise_clamped
+                // intrinsic. They make a claim that holds true at
+                // that specific point in the IR. Just drop the
+                // annotation and return the bounds of the first arg.
```
dsharletg, line 1088: ... but only when it's a single point?

src/Bounds.cpp
```diff
@@ -1080,6 +1080,23 @@ class Bounds : public IRVisitor {
...
+                return;
+            }
+
+            if (op->is_intrinsic(Call::unsafe_promise_clamped) &&
+                interval.is_single_point(op->args[0]) &&
+                lower.is_single_point(op->args[1]) &&
+                upper.is_single_point(op->args[2])) {
+                // unsafe_promise_clamp holds globally, so we can lift the entire intrinsic.
```
dsharletg, line 1096: Why is this true only when the operands are a single point?

src/Bounds.cpp
```diff
@@ -1080,6 +1080,23 @@ class Bounds : public IRVisitor {
...
+                interval = Interval::single_point(op);
```
dsharletg, line 1097: Would it be better to return `Interval::single_point(op->args[0])`?

dsharletg:
This has fixed the performance issues I saw, and it looks like it's even making some things slightly faster than from before the regression :) But others may be getting a bit slower too.

src/Bounds.cpp
steven-johnson: It's counterintuitive to the casual reader that promise_clamped and unsafe_promise_clamped have different global behaviors; it's probably worth adding a sentence or two clarifying why that is

steven-johnson:
Is this ready to land (pending green bots) or does it need more pondering?

abadams:
I'm happy with it. Up to Dillon I guess.

dsharletg:
It's ready, and we should land it as soon as we can. Thanks for the fix!

steven-johnson:
Will land as soon as everything goes green.

PR #4877, opened 2020-04-27
Switch to using dlsym for MTLCopyAllDevices()

shoaibkamil:
Should fix reported iOS link error-- my device still _seems_ to work fine using the previous approach, but since this API is not supposed to be on iOS, we should use this approach anyway.

### DIFF
--- a/src/runtime/metal.cpp
+++ b/src/runtime/metal.cpp
@@ -10,7 +10,6 @@
 
 extern "C" {
 extern objc_id MTLCreateSystemDefaultDevice();
-extern objc_id MTLCopyAllDevices();
 extern struct ObjectiveCClass _NSConcreteGlobalBlock;
 }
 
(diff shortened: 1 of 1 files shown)


shoaibkamil:
Failure is unrelated, but I'm not sure if this is a known issue.  Looks like a CUDA out-of-memory error on `max_filter`,

steven-johnson:
Apparently this still fails on iPhone with

```
Undefined symbols for architecture arm64:
  "_halide_get_symbol", referenced from:
      _halide_metal_acquire_context in libhalide_library_runtime_metal.a(halide_library_runtime_metal_88a399c7d19333d2196d685e99851a6c.o)
```

...which makes sense, as we don't link in any `halide_get_symbol` implementation for IOS (we do for OSX). But how did it pass the tests elsewhere?

PR #4886, opened 2020-05-02
Autofix all clang-tidy errors

steven-johnson:
Currently, our clang-tidy checks only run on src/*.cpp, but CMake makes it easy to run on ~everything, via defining `-DCMAKE_CXX_CLANG_TIDY=path/to/clang-tidy`, so cleaning up everything is desirable.

This PR autofixed everything by (roughly):
- configure CMake with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DWARNINGS_AS_ERRORS=OFF`
- `run-clang-tidy.py -p . -quiet -fix`

With a little manual cleanup (it decided to 'fix' the captive PyBind11 library too, which didn't work).

If accepted, there should probably be a followup PR to revise our clang-tidy presubmit check to check everything (not just src).

Note that these are 100% autofixed (not manually fixed), followed by clang-format; I did a quick skim to see if any fixes looked dubious or undesirable, but there may be things I overlooked.


alexreinking:
Everything looks reasonable to me, but we'll see what the buildbots have to say. Glad to see `clang-tidy` pulling its weight around here 😉

apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -13,22 +15,22 @@ using namespace Halide::ConciseCasts;
 Var x, y, c, yi, yo, yii, xi;
 
 // Average two positive values rounding up
-Expr avg(Expr a, Expr b) {
```
abadams, line 16: I think the changes here are a net readability loss for performance penny-pinching. Happy to do this inside the compiler, but it's distracting in code intended to demonstrate how Halide works.

abadams:
I vote we turn off the checkers that inject const-refs and std::moves (i.e. those designed to avoid copies) for apps/ and tutorials/

Other checkers like emplace_back, override, etc all look good to turn on globally.

steven-johnson:
> I vote we turn off the checkers that inject const-refs and std::moves (i.e. those designed to avoid copies) for apps/ and tutorials/

Fair. Let me see if we can just add .clang-tidy files there that selectively override.

steven-johnson:
Getting clang-tidy to vary checking between trees is a little odder than I expected, and this isn't very important (just a late Friday diversion); I think I'm gonna close this and check in the fixes in smaller chunks to make it more tractable.

PR #4899, opened 2020-05-05
Enable dynamic-sized allocations in kernels for openglcompute

abadams:
By making one global heap allocation per alloc group in the kernel, and then only
coalescing those allocations that have matching type.

This should also work for d3d, if someone wants to generalize the == OpenGLCompute checks.

### DIFF
--- a/test/correctness/dynamic_allocation_in_gpu_kernel.cpp
+++ b/test/correctness/dynamic_allocation_in_gpu_kernel.cpp
@@ -4,7 +4,7 @@ using namespace Halide;
 
 int main(int argc, char **argv) {
     Target t(get_jit_target_from_environment());
-    if (!t.has_gpu_feature() || t.has_feature(Target::OpenGLCompute)) {
+    if (!t.has_gpu_feature() && !t.has_feature(Target::OpenGLCompute)) {
         printf("Not running test because no gpu target enabled\n");
         return 0;
     }
--- a/test/correctness/dynamic_allocation_in_gpu_kernel.cpp
+++ b/test/correctness/dynamic_allocation_in_gpu_kernel.cpp
@@ -4,41 +4,51 @@ using namespace Halide;
 
 int main(int argc, char **argv) {
     Target t(get_jit_target_from_environment());
-    if (!t.has_gpu_feature()) {
+    if (!t.has_gpu_feature() || t.has_feature(Target::OpenGLCompute)) {
         printf("Not running test because no gpu target enabled\n");
         return 0;
     }
 
-    Func f, g;
+    Func f1, f2, f3, f4, f5, f6, g;
     Var x, y;
     Param<int> p;
 
-    f(x, y) = x + y;
-    g(x, y) = f(x, y) + f(x + p, y + p);
+    f1(x, y) = cast<float>(x + y);
+    f2(x, y) = cast<int>(f1(x, y) + f1(x + 1, y + 1));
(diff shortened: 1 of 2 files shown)


test/correctness/dynamic_allocation_in_gpu_kernel.cpp
```diff
@@ -4,41 +4,51 @@ using namespace Halide;
 
 int main(int argc, char **argv) {
     Target t(get_jit_target_from_environment());
-    if (!t.has_gpu_feature()) {
+    if (!t.has_gpu_feature() || t.has_feature(Target::OpenGLCompute)) {
```
slomp, line 7: Is this `t.has_feature(Target::OpenGLCompute` intentional?

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
```diff
@@ -4,41 +4,51 @@ using namespace Halide;
...
     for (int i = 0; i < 10; i++) {
         p.set(i);
-        Buffer<int> result = g.realize(W, H);
+        Buffer<float> result = g.realize(W, H);
         result.copy_to_host();
         for (int y = 0; y < H; y++) {
             for (int x = 0; x < W; x++) {
-                int correct = x + y + (x + i) + (y + i);
-                int actual = result(x, y);
+                float correct = 64 * x + 64 * y + 64 * i + 320;
```
slomp, line 48: is this hard-coded 64 constant intentional?

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
abadams: Yes, OpenGLCompute doesn't otherwise count as has_gpu_feature because it has such a terrible type system.

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
abadams: Yes it's what the expression above happens to work out to

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
shoaibkamil: Yes but I believe this conditional is wrong-- I think you _do_ want to run the test for OpenGLCompute.

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
slomp: Oh, OK, now I see.
I had to read and reflect on the code a few times until it clicked!

steven-johnson:
Do our buildbots ever attempt any testing with OpenGLCompute?

abadams:
No, I don't believe so. It was just people testing it on their personal machines.

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
abadams: oops

abadams: Fixed.

test/correctness/dynamic_allocation_in_gpu_kernel.cpp
abadams: I took a different approach - I knew it had to be linear, so I just wrote down a linear function of x, y, and i and found the constants with trial-and-error :)

shoaibkamil:
LGTM now :)

steven-johnson:
Ready to land?

PR #4900, opened 2020-05-06
Support Ubuntu 20.04 LTS

alexreinking:
Ubuntu 20.04 LTS packages slightly older versions of two Python dependencies:

* `python3-imageio` **2.4.1-3**. Unfortunately this is the most recent version that conservatively rejects numpy-compatible types. This PR wraps our buffers in two tutorials with `np.asanyarray()`.
* `python3-pybind11` **2.4.3-2build2**. This version works fine for our purposes.

It would be nice to support building Halide against these versions so that building with a virtual environment is optional.

The changes to the requirements file ensure that testing is done against these versions when running in a virtual environment.

### DIFF
--- a/python_bindings/requirements.txt
+++ b/python_bindings/requirements.txt
@@ -8,5 +8,8 @@
 numpy
 scipy
 pillow
+
+# Keep versions of these requirements equal to the versions in Ubuntu 20.04 LTS
+# because newer versions fix bugs we work around.
 imageio==2.4.1
 pybind11==2.4.3
--- a/python_bindings/requirements.txt
+++ b/python_bindings/requirements.txt
@@ -8,5 +8,5 @@
 numpy
 scipy
 pillow
-imageio
-pybind11
+imageio==2.4.1
+pybind11==2.4.3
--- a/python_bindings/tutorial/lesson_02_input_image.py
+++ b/python_bindings/tutorial/lesson_02_input_image.py
@@ -86,7 +86,7 @@ def main():
     assert output_image.type() == hl.UInt(8)
 
     # Save the output for inspection. It should look like a bright parrot.
-    imageio.imsave("brighter.png", output_image)
+    imageio.imsave("brighter.png", np.asanyarray(output_image))
     print("Created brighter.png result file.")
 
     print("Success!")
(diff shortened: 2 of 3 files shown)


python_bindings/requirements.txt
```diff
@@ -8,5 +8,5 @@
 numpy
 scipy
 pillow
-imageio
-pybind11
+imageio==2.4.1
```
steven-johnson, line 11: Adding comments here (from the PR description) would be useful for future maintainers

python_bindings/tutorial/lesson_02_input_image.py
```diff
@@ -86,7 +86,7 @@ def main():
     assert output_image.type() == hl.UInt(8)
 
     # Save the output for inspection. It should look like a bright parrot.
-    imageio.imsave("brighter.png", output_image)
+    imageio.imsave("brighter.png", np.asanyarray(output_image))
```
steven-johnson, line 89: ditto re: adding comment

python_bindings/tutorial/lesson_07_multi_stage_pipelines.py
```diff
@@ -82,7 +83,7 @@ def main():
         # parrot, and it should be two pixels narrower and two pixels
         # shorter than the input image.
 
-        imageio.imsave("blurry_parrot_1.png", result)
+        imageio.imsave("blurry_parrot_1.png", np.asanyarray(result))
```
steven-johnson, line 86: ditto re: adding comment

alexreinking:
> LGTM but not sure I understand the numpy-type issue. What versions have we been using up to now and what is changed?

I think the latest version of imageio is 2.8. The requirements file before this patch will use whatever the newest version is.

Versions less than 2.5 checked explicitly to see if the argument to imsave is a (sub)type of numpy.ndarray via instanceof. If not, it throws an exception. This causes test failures because Halides Python bindings don't return ndarrays.

Newer versions (2.5+) actually call np.asanyarray on the input unconditionally, which converts any array-like to an ndarray and is smart about not copying when it doesn't have to.

alexreinking:
> (approved with nits)

Added comments as requested.

PR #4902, opened 2020-05-07
Keep constants outermost in LICM

abadams:
LICM is prone to making a bit of a mess. I found it can be tamed by not
letting it sink constants to the bottom of an expression, and by running
CSE before grouping loop invariants instead of after to stop it from
reassociating things in way that break up common subexpressions.

This provides small speedups on several of the apps. The largest gain is
1-2% on local_laplacian.

### DIFF
--- a/src/LICM.cpp
+++ b/src/LICM.cpp
@@ -338,6 +338,10 @@ class GroupLoopInvariants : public IRMutator {
     };
 
     int expr_depth(const Expr &e) {
+        // We want to keep constants outermost
+        if (is_const(e)) {
+            return std::numeric_limits<int>::max();
+        }
         ExprDepth depth(var_depth);
         e.accept(&depth);
         return depth.result;
@@ -496,8 +500,8 @@ class GroupLoopInvariants : public IRMutator {
 };
 
 Stmt loop_invariant_code_motion(Stmt s) {
-    s = GroupLoopInvariants().mutate(s);
     s = common_subexpression_elimination(s);
+    s = GroupLoopInvariants().mutate(s);
     s = LICM().mutate(s);
     s = simplify_exprs(s);
     return s;


src/LICM.cpp
```diff
@@ -338,6 +338,10 @@ class GroupLoopInvariants : public IRMutator {
     };
 
     int expr_depth(const Expr &e) {
+        // We want to keep constants outermost
+        if (is_const(e)) {
+            return std::numeric_limits<int>::max();
```
steven-johnson, line 343: Is there any chance that a caller could increment the result ("I want it outside this other thing") and overflow?

abadams: No, this is only used immediately below to sort and isn't exposed out of this translation unit.

steven-johnson:
I think this has injected a bug into some Google code -- a test timeout (not sure if infinite-loop or merely much much slower), which goes away if I revert this change. Let me do a bit more testing but we may need to revert this pending more examination.

steven-johnson:
FYI, https://github.com/halide/Halide/pull/4908 appears to fix the slowdown injection mentioned above.

abadams:
Interesting. This change must have made it more likely for the load and store to get CSE'd differently. Does your slowdown case do a +=?

steven-johnson:
No, not explicitly. Plenty of addition though. I'll keep trying to get a very simple repro just to try to run it to ground.

abadams:
Another possibility for GPU kernels is the size of the closure. Small changes to LICM can result in changes to the number of uniform params getting passed in, as the number of loop invariants changes.

steven-johnson:
Here's a reasonably simple snippet that is ~2x as slow with this PR in place:

```
#include "Halide.h"
#include "halide_benchmark.h"
#include <stdio.h>

using namespace Halide;
using Halide::Tools::benchmark;

constexpr int kTileSize = 8;
const int kSize = 2048;

    Var x, y, tx, ty;

    ImageParam input(UInt(16), 2);

    Func all_zero;
    all_zero(x, y) = 0;
    all_zero.compute_root();

    Func image;
    image(x, y) = select(
        all_zero(x, y) == 0,  // note: this is always true, but necessary for slowdown
        input(x, y),
        u16(0));

    RDom tile_r(0, kTileSize, 0, kTileSize);

    Func packed_sum;
    packed_sum(x, y) = sum(image(
        x / kTileSize * kTileSize + tile_r.x,
        y / kTileSize * kTileSize + tile_r.y));
    packed_sum.compute_root();

    Func result;
    result(tx, ty) = packed_sum(tx * kTileSize, ty * kTileSize);

    // --------------------

    Buffer<uint16_t> input_buf(kSize, kSize);
    input_buf.fill(0);

    Buffer<uint16_t> output_buf(kSize / kTileSize, kSize / kTileSize);

    result.compile_jit();
    input.set(input_buf);
    double t = Halide::Tools::benchmark(1, 100, [&]() {
        result.realize(output_buf);
    });

    printf("benchmark time: %f\n", t);

    return 0;
}
```

abadams:
Ah, ok, there's a += hiding in the sum helper. Was this case fixed by 4908, or is it still broken?

steven-johnson:
> Was this case fixed by 4908, or is it still broken?

Interestingly, still broken, so my claim about that being the fix was either mistaken or not the whole story.

abadams:
I'll take a look then

abadams:
Yeah this "fix" is actually broken. Any expression with a constant in it counts as max depth, not just constant offset. I have a better fix. Stand by.

PR #4905, opened 2020-05-08
Use correct arguments list in the bounds inference for fused functions

vksnk:
Currently it always uses arg list of the parent function, which is not correct if they differ for the fused function.


vksnk:
Thanks for the review!

PR #4910, opened 2020-05-11
Fix nasty bug in solver.

abadams:
It was mutating min(x - y, x - z) to x - min(y, z), when it should be x - max(y, z). No idea how this wasn't causing crashes all over the place.

We really should be using an IR rewriter here so we can verify these
transformations.

### DIFF
--- a/src/Solve.cpp
+++ b/src/Solve.cpp
@@ -362,6 +362,7 @@ private:
                 expr = a * b;
             }
         }
+
         return expr;
     }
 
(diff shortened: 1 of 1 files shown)


shoaibkamil:
🙃

abadams:
fuzz_simplify never calls the solver, so it wouldn't have. We could make a fuzz_solver... Probably easier to just use rewrite rules for the solver though.

abadams:
Failure is the usual max-filter-on-gpu flake.

steven-johnson:
Failures appear to be the flaky CUDA failures, good to land

abadams:
The GPU max filter crashes were on different workers. I'm going to run it locally to check if the amount of memory it allocates changes with this PR.

abadams:
Allocations are the same before and after.

PR #4912, opened 2020-05-11
add HL_DISABLE_CAN_PROVE_RULES

steven-johnson:
This adds a simple way to disable all rules using can_prove via an env var (useful in this branch).

Also, a drive-by fix to allow correctness_simplify to pass.

### DIFF
--- a/src/Simplify.cpp
+++ b/src/Simplify.cpp
@@ -40,10 +40,18 @@ Simplify::Simplify(bool r, const Scope<Interval> *bi, const Scope<ModulusRemaind
 
     use_synthesized_rules = get_use_synthesized_rules_from_environment();
 
-    // Don't require this one to be set
-    if (get_env_variable("HL_DISABLE_CAN_PROVE_RULES") == "1") {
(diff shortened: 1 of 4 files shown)


src/Simplify.cpp
```diff
@@ -40,6 +40,12 @@ Simplify::Simplify(bool r, const Scope<Interval> *bi, const Scope<ModulusRemaind
 
     use_synthesized_rules = get_use_synthesized_rules_from_environment();
 
+    // Don't require this one to be set
+    if (get_env_variable("HL_DISABLE_CAN_PROVE_RULES") == "1") {
```
abadams, line 44: Simplify instances are constructed somewhat frequently, so I'm a little worried this might hurt compile times, which we are benchmarking. Not sure how fast an env var lookup is. Can this be made static?

steven-johnson: Ah, yes: I didn't notice the static usage for HL_USE_SYNTHESIZED_RULES.

PR #4913, opened 2020-05-12
Fix tutorial 9. Align function definition.

leifhelm:
The definition of `consumer` and `consumer_2` did not align in the Case 4 example.


abadams:
Thanks for the fix

PR #4914, opened 2020-05-12
Revert "Keep constants outermost in LICM"

abadams:
Reverts halide/Halide#4902


abadams:
Added a commit that fixes the issue in a better way.

PR #4923, opened 2020-05-16
More portable handling of unused declarations in CodeGen_C

### LINKED ISSUE #4918: Tutorial lesson_12 fails to run, reporting compile failure.
Hi

I'm working my way through the halide tutorials on Odroid XU4, Linux 18.04.4, armv7l.
Lesson_12 seems to compile ok but when running, aborts with a build failure.  Tried running with  
HL_DEBUG_CODEGEN=3 and the last part of debug is as follows:

Testing GPU correctness:
Realizing Pipeline for target(arch_unknown-0-os_unknown)
jit-compiling for: target(arm-32-linux-jit-opencl-user_context)
Reusing old jit module compiled for :
target(arm-32-linux-jit-opencl-user_context)
custom_print: 0xb6f20fbc
custom_malloc: 0xb6f22000
custom_free: 0xb6f2205c
custom_do_task: 0xb6f1ffac
custom_do_par_for: 0xb252d130
custom_error: 0xb2deec11
custom_trace: 0xb2526000
JIT input Image argument b0 @ 0x55d2c0
__user_context @ 0xbef8ea44
JIT output buffer @ 0x7fac90, 0
Calling jitted function
Error: CL: clBuildProgram failed: CL_BUILD_PROGRAM_FAILURE
Build Log:
<source>:127:8: error: casting to void is not allowed
 (void)_38;
       ^

error: Compiler frontend failed (error code 59)

Is armv7l supported??

Regards

dsharletg:
This PR makes it so CodeGen_C uses some C++ trickery to avoid unused declaration warnings, but I can't find a reliable way to avoid unused variable declarations in OpenCL C. I think we should just let unused declarations be in OpenCL C. I doubt any OpenCL implementations treat them as errors, and warnings should never be visible.

Fixes #4918

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -183,9 +183,12 @@ inline T halide_cpp_max(const T &a, const T &b) {return (a > b) ? a : b;}
 template<typename T>
 inline T halide_cpp_min(const T &a, const T &b) {return (a < b) ? a : b;}
 
+template<typename T>
+inline void halide_unused(const T&) {}
+
 template<typename A, typename B>
 const B &return_second(const A &a, const B &b) {
-    (void) a;
+    halide_unused(a);
     return b;
 }
 
@@ -2803,7 +2806,7 @@ void CodeGen_C::visit(const IfThenElse *op) {
 void CodeGen_C::visit(const Evaluate *op) {
     if (is_const(op->value)) return;
     string id = print_expr(op->value);
-    stream << get_indent() << "(void)" << id << ";\n";
+    stream << get_indent() << "halide_unused(" << id << ");\n";
 }
 
 void CodeGen_C::visit(const Shuffle *op) {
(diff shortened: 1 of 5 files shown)


alexreinking:
What's wrong with the cast to void?

abadams:
The issue is that apparently the opencl driver in #4918 is non-conformant, and others might be too.

alexreinking:
Have we opened a ticket with the OpenCL driver vendor?

dsharletg:
I did ask the reporter of #4918 to report the conformance issue to the OpenCL implementation they are using (https://github.com/halide/Halide/issues/4918#issuecomment-629466494). But given the nature of OpenCL, even if it's an easy fix for the vendor, it will likely be impossible to assume the fix actually lands on every possible device it runs on.

Regarding a comment: note that CodeGen_OpenCL_Dev.cpp is the only C usage (C++ is handled differently) that actually drops the (void) annotation, and there is a comment there explaining (and referring to #4918).

steven-johnson:
Failure is unrelated.

PR #4926, opened 2020-05-18
Fuzz simplify wasn't generating conditions on non-bools

abadams:
Unless they were the arg to a select


steven-johnson:
Remaining failures are unrelated.

PR #4927, opened 2020-05-18
Add stringification for Dim.dim_type enum.

Infinoid:
This makes a .compute_with() error message easier to read.

Before:

`Invalid compute_with: types of dim 0 of f_intm.s1(r$z is 2) and f_intm$1.s1(r$z is 1) do not match.`

After:

`Invalid compute_with: types of dim 0 of f_intm.s1(r$z is ImpureRVar) and f_intm$1.s1(r$z is PureRVar) do not match.`

### DIFF
--- a/src/AutoScheduleUtils.cpp
+++ b/src/AutoScheduleUtils.cpp
@@ -116,7 +116,7 @@ Definition get_stage_definition(const Function &f, int stage_num) {
 
 vector<Dim> &get_stage_dims(const Function &f, int stage_num) {
     static vector<Dim> outermost_only =
-        {{Var::outermost().name(), ForType::Serial, DeviceAPI::None, Dim::Type::PureVar}};
+        {{Var::outermost().name(), ForType::Serial, DeviceAPI::None, DimType::PureVar}};
     if (f.has_extern_definition()) {
         return outermost_only;
     }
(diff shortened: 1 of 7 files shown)


dsharletg:
I think it would be better to have a helper function to convert a dim type to a string, rather than a new member function.

abadams:
Something is going wrong with the "before". That's supposed to print via an overload operator<<(std::ostream &, const DimType &) in IRPrinter.h, but no such overload exists! It must have been lost from the branch that updated the error message.

So the code that needs updating is IRPrinter.h / IRPrinter.cpp (See for example how MemoryType is handled)

Infinoid:
Is DimType a thing?  It looks like `enum Type` is declared locally within `struct Dim`, to me.  And it has nothing to do with the globally defined `Type` class.

Infinoid:
Sounds like you're saying we can turn `enum Type` into `enum class Type` and then do similar things to what `enum clas MemoryType` does.  I'll look into it, thanks.

abadams:
Oh I see, it's Dim::Type, not DimType. Yeah IMO it should be an enum class alongside Dim, not an enum within it.

steven-johnson:
Buildbot run is clean, merging.

PR #4928, opened 2020-05-18
Fix capture by reference of temporary

shoaibkamil:
Fix capture by reference of the temporary string.

### DIFF
--- a/src/FuseGPUThreadLoops.cpp
+++ b/src/FuseGPUThreadLoops.cpp
@@ -765,9 +765,7 @@ public:
                         const string &group_name;
 
                     public:
-                        // Because we're passed in a temporary to construct this, don't capture
-                        // alloc_name by reference
-                        RewriteGroupAccess(const string alloc_name, const string &group_name)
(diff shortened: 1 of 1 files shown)


src/FuseGPUThreadLoops.cpp
```diff
@@ -761,11 +761,13 @@ class ExtractSharedAndHeapAllocations : public IRMutator {
...
                         }
-                        const string &alloc_name;
+                        const string alloc_name;
                         const string &group_name;
 
                     public:
-                        RewriteGroupAccess(const string &alloc_name, const string &group_name)
+                        // Because we're passed in a temporary to construct this, don't capture
+                        // alloc_name by reference
+                        RewriteGroupAccess(const string alloc_name, const string &group_name)
```
abadams, line 770: Doesn't this make an extra copy now? I think clang-tidy might complain unless you throw a std::move in the initialization. Alternatively, the argument type could remain a const ref.

abadams:
cmake-based tests failing as expected, because this isn't based on cmake-modern. Merging.

Issue #4940, opened 2020-05-21
host_supports_target_device('cuda') crashes on linux-arm32


steven-johnson:
Calling host_supports_target_device() with a target that includes `cuda` will crash inside MCJIT on linux-arm32 (with a null pointer somewhere). We should be debugged to figure out why. At least one test (for PyTorch) is partially skipped as a temporary workaround.

steven-johnson:
Note: so does 'get_device_interface_for_device_api(cuda)' -- basically, anything that invokes the JIT to try to get Cuda-related code.

steven-johnson:
I suspect that Issue #4946 is at least part of the issue here.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_STALE):** The report is a null-pointer crash "inside MCJIT" on linux-arm32 when JITting anything CUDA-related. MCJIT is gone from Halide entirely — the JIT is now ORC `LLJIT` — and the specific arm32 failure mode (LLVM lowering atomics to `__sync_*` libgcc symbols that aren't resolvable in the JIT, which @steven-johnson himself flagged as the likely cause via #4946) has since been fixed by mapping those symbols explicitly.

**Evidence:** `src/JITModule.cpp:265` uses `std::unique_ptr<llvm::orc::LLJIT>`; `grep -rn MCJIT src/` returns nothing. Commit `1d5c1e2b8` ("Move ARM Linux workflows to GitHub Actions (#8977)") added `arm32_sync_builtins` (`src/JITModule.cpp:50-108`) mapping `__sync_synchronize`, `__sync_fetch_and_*_4/8`, `__sync_val_compare_and_swap_4/8` into the JIT symbol table, and added `.github/workflows/testing-arm-linux.yml`. No PyTorch-test skip referencing this remains.

Closing as stale: MCJIT no longer exists in Halide (the JIT is ORC `LLJIT`, `src/JITModule.cpp:265`), and the arm32 JIT symbol-resolution failure this was traced to (#4946) was fixed in `1d5c1e2b8`, which also added arm32 CI. Please reopen with a fresh backtrace if `host_supports_target_device` still crashes on linux-arm32.

PR #4941, opened 2020-05-21
Disable PyTorch test on arm32 (Issue #4940)

steven-johnson:


steven-johnson:
(FYI: we never noticed this before because apparently the buildbots were never running test_internal until now...)

src/CodeGen_PyTorch.cpp
```diff
@@ -310,6 +310,14 @@ inline int test1_th_(at::Tensor &_buf, float _alpha, int32_t _beta) {
         compare_src(src.str(), correct_src);
     }
 
+    // TODO: host_supports_target_device('cuda') crashes on arm32 inside MCJIT
+    // (at least on our buildbots). Skip there pending investigation.
+    // https://github.com/halide/Halide/issues/4940
+    if (get_host_target().arch == Target::ARM && get_target().bits == 32) {
```
abadams, line 316: s/get_target/get_host_target?

steven-johnson: Yeah, it's wrong... but I'm going to withdraw this PR; there are other tests broken similarly on arm32 and I need to either fix the underlying issue, or find a similar stopgap.

PR #4945, opened 2020-05-21
Add 'test_correctness_multi_gpu' test target

steven-johnson:
This simplifies an ugly special case in our buildbot testing setup


alexreinking:
Note that `-L correctness` is a regex, so it will still run both sets. Generally speaking, labels should not be substrings of one another unless they intentionally represent subsets.

Issue #4946, opened 2020-05-21
Migrate usage of `__sync` intrinsics to `__atomic` intrinsics


steven-johnson:
Our runtime uses a mix of both, but the older `__sync` versions are deprecated, and apparently not supported for some combinations of the toolsets we use on linux-arm32 (LLVM will map them to symbols that are known to be not present, so some of our code will fail at runtime in JIT situations). We could hack around this issue, but at this point it's probably more sensible to just migrate everything to use the newer `__atomic` intrinsics, as they are likely to be properly supported by the recent versions of gcc/clang/etc we use.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_DUPLICATE):** The migration was done for 64-bit targets and deliberately not done for 32-bit, and the remaining 32-bit question was moved to a dedicated issue that the source code points at by number. #7431 covers exactly the same ground ("Halide uses the deprecated `__sync` primitives on 32-bit targets ... hold discussion as to why/whether we can use `__atomic` everywhere").

**Evidence:** `src/runtime/runtime_atomics.h:4-9`: "This file provides an abstraction layer over the `__sync`/`__atomic` builtins in Clang; for various reasons, we use `__sync` for 32-bit targets, and `__atomic` for 64-bit. ... See https://github.com/halide/Halide/issues/7431 for a discussion of the history and issues as to why we work this way." The `__sync` uses in the file are all under `#ifdef BITS_32` (line 31 onward). The arm32 JIT motivation was separately handled in `1d5c1e2b8` (`src/JITModule.cpp:50-108`).

Closing as superseded by #7431, which tracks the remaining question. The migration to `__atomic` was completed for 64-bit targets; 32-bit deliberately still uses `__sync` (see the header comment at `src/runtime/runtime_atomics.h:4-9`, which cites #7431), and the arm32 JIT symbol failure that motivated this was fixed separately in `1d5c1e2b8`.

PR #4949, opened 2020-05-21
Upgrade ScopedSpinLock to use __atomic (Issue #4946)

steven-johnson:


abadams:
Looks equivalent to existing code to me, but I'd rather defer to Zalman for checking this sort of thing.

Issue #4884, opened 2020-04-30
segfault using define_extern without -rdynamic


Infinoid:
If you use Func.define_extern(), and you forget to add `-Wl,-export-dynamic` or `-rdynamic` to your LDFLAGS, the app segfaults.

It makes sense that dynamic symbols are required, as the JIT code needs to get the function address from the parent executable.  But is it possible to emit a clearer error message about it?

Here is what happens when you build `test/correctness/extern_sort.cpp` without those flags.

```
$ clang++-10 -O2 -Wall -g -ggdb -o extern_sort extern_sort.cpp -lHalide
$ ./extern_sort
Segmentation fault
```

```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007ffff7f3703c in f1 ()
#2  0x00007ffff36fd25f in Halide::Pipeline::call_jit_code (this=0x7fffffffd818, target=..., args=...)
    at /home/infinoid/workspace/halide/git/src/Pipeline.cpp:1054
#3  0x00007ffff3703528 in Halide::Pipeline::realize (this=this@entry=0x7fffffffd818, outputs=..., t=..., param_map=...)
    at /home/infinoid/workspace/halide/git/src/Pipeline.cpp:1146
```

abadams:
Yeah at some point things changed from throwing an error when a symbol could not be resolved to just resolving it to null and then crashing if it gets called. Not particularly helpful. We should track this down.

alexreinking:
Is this maybe something where we wanted to avoid a null check for efficiency? It would be ideal to move the check earlier, perhaps directly into define_extern.

abadams:
No, the resolution is a compile-time thing, so efficiency isn't really a concern. I think maybe we need a null check in HalideJITMemoryManager::getSymbolAddress. I think we set LLVM up to do all symbol resolution through there.

abadams:
Fixed by the referenced PR

PR #4953, opened 2020-05-21
Add assert-fail in HalideJITMemoryManager::getSymbolAddress()

steven-johnson:
If we can't get the address of a symbol, it's likely that we're going to crash later on anyway when we try to jump to a null address, which is hard to debug; changing to an assert-fail makes the issue easier to debug.

(Note that I *think* this is always a fatal error, but it may be that this needs to be a warning instead in some cases.)

### DIFF
--- a/src/JITModule.cpp
+++ b/src/JITModule.cpp
@@ -229,6 +229,8 @@ public:
             result = (uint64_t)&__udivdi3;
         }
 #endif
+        internal_assert(result != 0)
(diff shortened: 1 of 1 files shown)


abadams:
Yes, I think this is a fatal error. Fixes #4884

abadams:
No buildbot runs. Are they currently functional?

steven-johnson:
"functional", yes, but lots of things broken. We plan on landing the CMake update by end of today at which point things should get better.

PR #4954, opened 2020-05-21
Fix clang-format issues

steven-johnson:
I don't know how/why the master branch isn't clean this way. (Maybe I have a different clang-format locally?)

### DIFF
--- a/apps/autoscheduler/FunctionDAG.cpp
+++ b/apps/autoscheduler/FunctionDAG.cpp
@@ -458,9 +458,11 @@ FunctionDAG::Edge::BoundInfo::BoundInfo(const Expr &e, const Node::Stage &consum
     const Mul *mul = add ? add->a.as<Mul>() : expr.as<Mul>();
     const IntImm *coeff_imm = mul ? mul->b.as<IntImm>() : nullptr;
     const IntImm *constant_imm = add ? add->b.as<IntImm>() : nullptr;
+    // clang-format off
     Expr v = (mul ? mul->a :
-                    add ? add->a :
-                          expr);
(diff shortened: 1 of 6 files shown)


steven-johnson:
...hmm, yeah, something *is* different between my local make-format and the one running on GHA. Which is the one that has changed?

abadams:
I don't think the remote one has changed recently. Are you using a different clang-format version?

steven-johnson:
`make clang-format` uses your active llvm install, whereas GHA uses some stable version; maybe clang-format in trunk changed recently? I'll investigate.

steven-johnson:
Looks like the change got injected here:

https://github.com/llvm/llvm-project/commit/4db94094b469b4715d08ef37f1799bf3ea7ca8ea

basically, it affects how nested ternary expressions are formatted. 

I'd argue that the new format is definitely *better*, but there doesn't appear to be a setting that allows us to preserve the old formatting, so this means that clang-format-11 can't be identical to previous clang-format versions for this.... sigh

steven-johnson:
I added clang-format as necessary (and/or tweaking other code) to deal with this variation.

abadams:
Yeah, definitely an improvement for ternary operators. At some point we'll just have to switch clang-format versions and update everything.

steven-johnson:
So from the comments on the LLVM bug above, it appears that Team Clang-Format makes a deliberate decision *not* to try to maintain backwards compatibility across versions of clang-format, so we should probably consider requiring a specific version of clang-format (both in our presubmits and via `make format`, etc) to ensure drift doesn't occur.

PR #4955, opened 2020-05-21
FreeBSD compatibility patches.

yurivict:


src/JITModule.cpp
```diff
@@ -109,6 +109,17 @@ void load_opengl() {
         llvm::sys::DynamicLibrary::LoadLibraryPermanently("/System/Library/Frameworks/OpenGL.framework/OpenGL", &error);
         user_assert(error.empty()) << "Could not find OpenGL.framework\n";
     }
+#elif defined(__FreeBSD__)|| defined(__OpenBSD__) || defined( __NetBSD__) || defined(__DragonFly__)
```
abadams, line 112: Looks like clang-format doesn't like the space before `__NetBSD__`

steven-johnson:
A complete rewrite of the CMake build system will be landing soon (possibly later today), so the change to halide.cmake will need attention then

abadams:
Thanks for doing this, looks like the changes are small. We don't really have any way of testing BSD though so I guess we'll have to rely on BSD users to fix things after the fact if something unique to BSD breaks.

krytarowski:
Please *never* put all BSDs into a shared type. This code is invalid for !FreeBSD, at least for NetBSD. If you want to add FreeBSD support, please define it as FreeBSD only.

krytarowski:
OpenBSD is also broken by this.

krytarowski:
Also, nobody will ever test it on other BSD variants such as BSD/OS.

abadams:
Seems like changing the enum and target flag to FreeBSD/freebsd would address that?

abadams:
I guess also the ifdef

krytarowski:
Please treat FreeBSD as a separate OS. They diverged 25 years ago!

yurivict:
I've updated the pull request.

src/Util.cpp
```diff
@@ -20,6 +20,12 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#if defined(__FreeBSD__)
+#define CAN_GET_RUNNING_PROGRAM_NAME
+#include <sys/types.h>
```
steven-johnson, line 25: To satisfy our clang-format rules, these must be alphabetically sorted:

```
#include <sys/sysctl.h>
#include <sys/syslimits.h>  // For PATH_MAX
#include <sys/types.h>
```

steven-johnson:
Changes look good (pending buildbot testing in https://github.com/halide/Halide/pull/4962), with the caveat that we need to add information to the README (and to a few files, e.g. Target.h) indicating that FreeBSD isn't regularly tested by the core contributors, and so may have bugs, yadda yadda

steven-johnson:
This needs attention as of now, the halide.cmake file no longer exists

steven-johnson:
(Can you re-sync this with master and resolve the conflicts?)

steven-johnson:
Reminder: we would welcome this patch if it's synced to master and updated appropriately. 

(If not, we may close it in a few days due to inattention; if that happens, feel free to reopen it in the future.)

steven-johnson:
Closing due to inactivity. (Feel free to re-open.)

PR #4957, opened 2020-05-22
Support NoOS bare-metal code generation for Arm targets

joesavage:


steven-johnson:
LGTM pending green buildbots.

steven-johnson:
Clone is clean aside from unrelated failures.

PR #4958, opened 2020-05-22
Add two new rewrite rules to the solver

abadams:
Over random schedules on lens blur reduces average peak memory by a full
third! No effect on the other apps.


abadams:
@shoaibkamil @jn80842 We should look at the solver more. Seems like there are opportunities there.

abadams:
Not sure what to make of the buildbot results. There are non-cmake builders broken but I'm not sure why.

PR #4961, opened 2020-05-25
Respect disable_llvm_loop_opt in the ptx backend

abadams:


src/CodeGen_PTX_Dev.cpp
```diff
@@ -449,8 +449,9 @@ vector<char> CodeGen_PTX_Dev::compile_to_src() {
     PassManagerBuilder b;
     b.OptLevel = 3;
```
steven-johnson, line 450: Unrelated note: we should really upgrade this code to use the 'non-legacy' PassManager (Codegen_LLVM.cpp was upgraded long ago).

src/CodeGen_PTX_Dev.cpp
```diff
@@ -449,8 +449,9 @@ vector<char> CodeGen_PTX_Dev::compile_to_src() {
...
     b.Inliner = createFunctionInliningPass(b.OptLevel, 0, false);
-    b.LoopVectorize = true;
+    b.LoopVectorize = !target.has_feature(Target::DisableLLVMLoopOpt);
```
steven-johnson, line 452: The logic used in Codegen_LLVM.cpp also uses the EnableLLVMLoopOpt feature; we should probably mimic that logic here too, for consistency.

steven-johnson: (I took the liberty of pushing a change that added that logic. PTAL)

abadams: Thanks

steven-johnson:
Failure is unrelated

Issue #4965, opened 2020-05-26
FR: add helpful error messages for inadequate-cuda-capability detection


steven-johnson:
Currently, if you compile for (say) `host-cuda-cuda_capability_50`, but run on a system that only supports lower versions of Cuda, you'll fail at runtime with a PTX_INVALID failure, which isn't very helpful. It would be nice if the runtime assertion code we generate sniffed the version-found vs the version-required and emitted a more useful error message.

PR #4966, opened 2020-05-26
cuda_mat_mul should skip test for Cuda < 5.0

steven-johnson:
Otherwise it fails with inscrutable error messages later.

(See also: https://github.com/halide/Halide/issues/4965)


abadams:
In CUDA, version is not the same thing as compute capability, so the error message should say compute capability.

steven-johnson:
> compute capability

Fixed

PR #4968, opened 2020-05-27
Add -I/usr/local/cuda/include to cflags for TEST_CUDA

steven-johnson:


alexreinking:
That looks right to me for the acquire_release test which uses CUDA directly. Not sure how it ever found the `cuda.h` headers. Maybe old packages put the header in the system paths? It hasn't been that way for a while.

PR #4972, opened 2020-05-28
Make Buffer always do type checks, even with NDEBUG

### LINKED ISSUE #4971: Garbage buffers possible from imageio with NDEBUG
The following Halide program exhibits some bad behavior. The load_image call loads the buffer as 8-bit when using the `apps/images/grey_small.png` input image, breaking type-safety when `NDEBUG` is defined.

Furthermore, `load_and_convert_image` fails with an assertion check in this case when `NDEBUG` is not defined. It isn't clear to me what the right way is to run the homepage code (a fairly serious problem, imo) on a simple greyscale PNG.

```
#include "Halide.h"
#include "halide_image_io.h"

#include <iostream>

using namespace Halide;
using namespace Halide::Tools;

// Code from halide-lang.org
Func blur3x3(Func input) {
    Func blur_x, blur_y;
    Var x, y, xi, yi;

    // The algorithm - no storage or order
    blur_x(x, y) = (input(x - 1, y) + input(x, y) + input(x + 1, y)) / 3;
    blur_y(x, y) = (blur_x(x, y - 1) + blur_x(x, y) + blur_x(x, y + 1)) / 3;

    // The schedule - defines order, locality; implies storage
    blur_y.tile(x, y, xi, yi, 256, 32).vectorize(xi, 8).parallel(y);
    blur_x.compute_at(blur_y, x).vectorize(x, 8);

    return blur_y;
}

    if (argc < 3) {
        std::cout << "Usage: " << argv[0] << " <input-file> <output-file>\n";
        return 1;
    }

    // 1. Load the image from disk.
    Buffer<float> input = load_image(argv[1]);

    // 2. Define the points outside the image.
    Func with_boundary = BoundaryConditions::repeat_edge(input);

    // 3. Build the blur program.
    Func output = blur3x3(with_boundary);

    // 4. JIT compile and run the blur.
    Buffer<float> result = output.realize(input.width(), input.height());
    
    // 5. Write the code back out.
    save_image(result, argv[2]);

    return 0;
}
```

abadams:
The type checks in Runtime::Buffer go away with NDEBUG on (for good reasons - the performance of a hot loop that uses operator() drops 10x if you include the assert). So Buffer shouldn't rely on them for its type checking.

Plus, if we use user_assert we can print the types in question and the actual line number.

Fixes #4971


abadams:
See #4971

src/Buffer.h
```diff
@@ -78,6 +78,29 @@ std::vector<int> get_shape_from_start_of_parameter_pack(Args &&... args) {
 template<typename T, typename T2>
 using add_const_if_T_is_const = typename std::conditional<std::is_const<T>::value, const T2, T2>::type;
 
+// Helpers to produce the name of a Buffer element type (a Halide
+// scalar type, or void, possibly with const). Useful for an error
+// messages.
+template<typename T>
+void buffer_type_name_non_const(std::ostringstream &oss) {
```
steven-johnson, line 85: Nit: using `ostream` instead of `ostringstream` would be fine and more general

src/Buffer.h
```diff
@@ -107,7 +130,10 @@ class Buffer {
                               std::is_void<T2>::value,
                           "type mismatch constructing Buffer");
         } else {
-            Runtime::Buffer<T>::assert_can_convert_from(*(other.get()));
+            user_assert(Runtime::Buffer<T>::can_convert_from(*(other.get())))
```
steven-johnson, line 133: Worth adding a comment here about why we don't delegate to assert_can_convert_from()

tools/halide_image_io.h
```diff
@@ -71,7 +71,7 @@ typedef bool (*CheckFunc)(bool condition, const char *msg);
 inline bool CheckFail(bool condition, const char *msg) {
     if (!condition) {
         fprintf(stderr, "%s\n", msg);
-        exit(-1);
+        abort();
```
steven-johnson, line 74: Why is abort() better than exit(-1) here?

abadams: So that you can drop in with gdb and figure out what the call stack was at the point of error. With exit(-1) you don't get that.

steven-johnson: Fair. (FWIW, CTest treats abort() as 'unrecoverable failure' that it refuses to treat as 'ordinary test failure'.)

src/Buffer.h
abadams: Done.

src/Buffer.h
abadams: Done.

tools/halide_image_io.h
alexreinking: @steven-johnson -- the error tests get around that by catching the abort signal and translating them to exit(1)'s. See https://github.com/halide/Halide/blob/master/test/common/expect_abort.cpp

steven-johnson:
failure is unrelated

PR #4974, opened 2020-05-28
Make LLVM_DIR take precedence when searching for Clang.

alexreinking:
@abadams reported a find_package issue with Clang locally. This patch works for me and for him. Let's see if the buildbots like it.


alexreinking:
@steven-johnson is there some way we can stop the non-cmake bots from running on PRs that change only files named "CMakeLists.txt" or "*.cmake"?

steven-johnson:
Possible? Certainly. I'd have to go dive in to figure out how, though.

alexreinking:
I think doing so will save us some real time in the short-medium term as people open issues against the CMake build.

steven-johnson:
I'm upgrading the buildbot setup to use the latest versions first.

alexreinking:
Superceded by #4980

PR #4978, opened 2020-05-28
Fix DWARF symbols in runtime, no-exceptions flags

alexreinking:
Fixes two issues I found with @abadams 

1. `-g` was applied to non-debug runtime files, bloating the binary.
2. Somehow I deleted the code I had to disable exceptions (during a rebase?)

Based on #4974


alexreinking:
Superceded by #4980

PR #4980, opened 2020-05-29
CMake "Day 1" patch

alexreinking:
Fixes several issues found on initial deployment of new CMake build, and adds a few new features:

* `Clang_DIR`-inferred-from-`LLVM_DIR` is now preferred when only `LLVM_DIR` is defined
* Binary size issues resolved
* Apps now print "Success!" messages to ensure they ran correctly and didn't just print 0-status usage messages (affected `camera_pipeline`)
  * Linear algebra tests now know their underlying CBlas implementation (missing command-line define)
* Option added for enabling WASM backend (no WASM tests, yet, though)
* Add multi-config packaging logic to Win32.
* Fixing Python apps that crash with `host-cuda`.

----

Regarding packaging, the intended workflow is to use a _multi-config_ generator like **Visual Studio** or **Ninja Multi-Config** (currently broken on Windows, see [this discussion on CMake Discourse](https://discourse.cmake.org/t/build-failure-with-ninja-multi-config/1290/8)).

Then you just:

```
> cmake --build . --config Debug
> cmake --build . --config Release
> cpack -C "Debug;Release"
```


alexreinking:
Ninja Multi-Config bug was confirmed by maintainers. See https://gitlab.kitware.com/cmake/cmake/-/issues/20775

alexreinking:
Looks like `_alloca` doesn't exist when you disable exceptions. Odd. Including the `<malloc.h>` header seems to fix this. I pushed that fix and stopped the buildbots on the previous push.

alexreinking:
Python build requires exceptions to be enabled in Halide. Otherwise the correctness test "test_compiletime_error" calls `abort()` which can't be caught by Python. Similarly with `tuple_data` and `pystub`.

@steven-johnson -- we should add `HALIDE_ENABLE_EXCEPTIONS=YES` to our buildbot flags.

PR #4992, opened 2020-06-03
Fix Doxygen config and output paths in Makefile.

### LINKED ISSUE #4991: Docs build issues with cmake-modern
Two minor issues with building the docs:

1) There's now a file in doc/, which "make clean" deletes. Previously "doc" was an output directory like "bin", not a source directory. Now it's awkwardly both. 

2) The config can't have been correctly copied over from the old Doxyfile, because the title of the project in the generated docs is "My Project", not "Halide"

alexreinking:
Makefile Doxygen output now appears in `share/doc/Halide`, which removes conflict with top-level `doc` and conforms to usual Linux directory layout in `/usr`.

Clean target now removes generated Doxyfile and share directory.

Fixes #4991

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1131,7 +1131,7 @@ clean:
 	rm -rf $(SHARE_DIR)
 	rm -rf $(DISTRIB_DIR)
 	rm -rf $(ROOT_DIR)/apps/*/bin
-	rm Doxyfile
+	rm -f Doxyfile
 
 .SECONDARY:
 
--- a/Makefile
+++ b/Makefile
@@ -431,7 +431,8 @@ LIB_DIR     = lib
 BIN_DIR     = bin
 DISTRIB_DIR = distrib
 INCLUDE_DIR = include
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1127,9 +1128,10 @@ clean:
 	rm -rf $(TMP_DIR)
 	rm -rf $(FILTERS_DIR)
 	rm -rf $(INCLUDE_DIR)
-	rm -rf $(DOC_DIR)
+	rm -rf $(SHARE_DIR)
 	rm -rf $(DISTRIB_DIR)
 	rm -rf $(ROOT_DIR)/apps/*/bin
+	rm -f Doxyfile
```
abadams, line 1134: If Doxyfile is a build intermediate (which I guess it is now?) then it should go in TMP_DIR

alexreinking: Done

steven-johnson:
Remaining failures are unrelated

PR #4995, opened 2020-06-03
README: add line for booststrapping vcpkg

shoaibkamil:
A fresh checkout will need to run the bootstrap script to build `vcpkg.exe`.


alexreinking:
Thanks for the catch!

PR #4997, opened 2020-06-03
Lower tuples to separate atomic nodes where possible

abadams:
If tuple components have no hazards between them, they can be computed
one after another in separate atomic blocks instead of in a single
combined atomic block. This lets us use atomic adds or CAS loops more
often and mutexes less often.

As a side-effect, this also means that update definitions that don't
modify one of the tuple components now entirely elide that store as a
no-op (see the change to the rfactor test).

This is part of the atomic_vectorization branch that can be separately merged.

### DIFF
--- a/src/SplitTuples.cpp
+++ b/src/SplitTuples.cpp
@@ -134,7 +134,7 @@ class SplitTuples : public IRMutator {
         }
     }
 
-    Stmt visit_provide(const Provide *op, const Atomic *atomic = nullptr) {
+    Stmt visit_provide(const Provide *op, const Atomic *atomic) {
 
         if (op->values.size() == 1) {
             if (atomic) {
@@ -317,6 +317,10 @@ class SplitTuples : public IRMutator {
     }
 
     Stmt visit(const Atomic *op) override {
+        // At this point in lowering, the only child of an atomic node
+        // should be a single provide node. We haven't many any
+        // statement mutations yet that would put things in between
+        // the provide and the atomic.
         if (const Provide *p = op->body.as<Provide>()) {
             return visit_provide(p, op);
         } else {
--- a/src/SplitTuples.cpp
+++ b/src/SplitTuples.cpp
@@ -1,8 +1,11 @@
 #include "SplitTuples.h"
+
 #include "Bounds.h"
+#include "ExprUsesVar.h"
 #include "Function.h"
 #include "IRMutator.h"
 #include "IROperator.h"
+#include "Simplify.h"
 
 namespace Halide {
 namespace Internal {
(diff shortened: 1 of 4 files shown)


steven-johnson:
```
/home/halidenightly/build_bot/worker/linux-64-gcc53-testbranch-trunk/halide/Makefile:1930: recipe for target 'error_atomics_vectorized_mutex' failed
make: *** [error_atomics_vectorized_mutex] Error 1
```

abadams:
Curious. I didn't think to test that error case locally because it's deleted in the branch I took this change from. I'll investigate...

abadams:
A case we formerly could not compile (it required vectorized mutexes) we can now compile correctly (it does not require mutexes at all). I'll make the test harder.

steven-johnson:
Windows failure looks unrelated; LGTM to land pending approval from @BachiLi

src/SplitTuples.cpp
```diff
@@ -165,31 +155,173 @@ class SplitTuples : public IRMutator {
...
         }
+    }
 
-        return result;
+    Stmt visit(const Provide *op) override {
+        return visit_provide(op, nullptr);
+    }
+
+    Stmt visit(const Atomic *op) override {
+        if (const Provide *p = op->body.as<Provide>()) {
```
BachiLi, line 320: Is it possible to have other nodes between the Atomic node and the Provide node? Also can we have multiple Provides inside an Atomic node?

test/correctness/atomic_tuples.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+        f(x, y) = {x, y, 0};
+        f(x, y) = {f(x, y)[1] + x,
+                   f(x, y)[0] + y,
+                   f(x, y)[2] + 1};
+
+        // The summation is coupled across the first two tuple
+        // components and there are two stores, so we need a mutex
+        // there. The last store could in principle be a separate atomic
+        // add, but we instead just pack it into the critical section.
```
BachiLi, line 104: is this a missing optimization opportunity that should be documented as a github issue?

src/SplitTuples.cpp
```diff
@@ -165,31 +155,173 @@ class SplitTuples : public IRMutator {
...
+            if (atomic && !separate_atomic_nodes_per_store) {
+                s = Atomic::make(atomic->producer_name, atomic->mutex_name, s);
+            }
+            return s;
         }
+    }
 
-        return result;
+    Stmt visit(const Provide *op) override {
+        return visit_provide(op, nullptr);
```
BachiLi, line 316: Maybe just `visit_provide(op)`? I don't have strong preference.

test/correctness/atomic_tuples.cpp
abadams: I'm not sure that it's actually an optimization

BachiLi: ah yes i agree

src/SplitTuples.cpp
abadams: I removed the default arg from visit_provide to make it unambiguous

src/SplitTuples.cpp
abadams: Not at this point in lowering. I added a comment to that effect.

steven-johnson:
Windows failures are unrelated, OK to land when other bots go green

PR #4998, opened 2020-06-03
Don't wrap a likely around an RDom predicate if it's already got one.

abadams:
GuardWithIf on RVars seems to be producing conditions of the form
likely(likely(condition)).

The predicate is a mix of stuff from the user and stuff from splits, so
there's no single source of likelies for me to go fix instead.

This was causing an issue in the atomic_vectorization branch. It can be separately fixed though.


steven-johnson:
Build failures are irrelevant. LGTM.

PR #5003, opened 2020-06-04
Fix D3D12Compute apps tests

shoaibkamil:
Mostly minor fixes.  All `correctness|generator|apps` tests now work on my machines (both running Win10Pro, one with an AMD GPU and the other with an Nvidia GPU).


src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -896,22 +896,79 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::add_kernel(Stmt s,
...
+                allocs.push_back(new_alloc);
+                replacements.erase(op->name);
+                return new_alloc;
+            } else {
+                return IRMutator::visit(op);
             }
         }
+
+        Stmt visit(const Free *op) override {
+            if (replacements.count(op->name)) {
```
abadams, line 934: Nit: 
```
auto it = replacements.find(op->name);
if (it != replacements.end()) {
   ... it->second ...
}
```

Instead of double a double map lookup

apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -459,11 +459,22 @@ void CameraPipe::generate() {
 
         Var xi, yi, xii, xio;
 
-        /* 1391us on a gtx 980. */
+        /* These tile factors obtain 1391us on a gtx 980. */
+        int tile_x = 28; 
+        int tile_y = 12;
+
+        if (get_target().has_feature(Target::D3D12Compute)) {
+            // D3D12 can only utilize a limited amount of
```
slomp, line 467: nitpick: it's not so much d3d12, but the minimum shader model version (5.1) that we're adhering to that has this limitation.

src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -896,22 +896,79 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::add_kernel(Stmt s,
...
             if (is_shared_allocation(op)) {
-                allocs.push_back(op);
+                // Because these will go in global scope,
+                // we need to ensure they have unique names.
+                std::string new_name = unique_name(op->name);
+                replacements[op->name] = new_name;
+
+                std::vector<Expr> new_extents;
+                for (size_t i = 0; i < op->extents.size(); i++) {
+                    new_extents.push_back(mutate(op->extents[i]));
```
slomp, line 912: Do we need to mutate these things? Can't we just pass op->extents directly to Allocate::make?

src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -896,22 +896,79 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::add_kernel(Stmt s,
...
+                if (op->new_expr.defined()) {
+                    new_new_expr = mutate(op->new_expr);
+                }
+
+                Stmt new_alloc = Allocate::make(new_name, op->type, op->memory_type, new_extents,
+                                                std::move(new_condition), std::move(new_body),
+                                                std::move(new_new_expr), op->free_function);
+
+                allocs.push_back(new_alloc);
+                replacements.erase(op->name);
```
slomp, line 926: Shouldn't we keep the replacements vector populated for Loads and Stores, and only remove it from the vector when visiting Free?

src/CodeGen_D3D12Compute_Dev.cpp
abadams: It seems extremely unlikely (impossible?) that the size of one shared allocation would depend on a load from another shared allocation. Still, safer to just always mutate than to make assumptions that may change in the future, especially if the cost of mutation is low (which it should be, as extent is probably a constant int).

src/CodeGen_D3D12Compute_Dev.cpp
abadams: body was already mutated above.

shoaibkamil: An `Allocate` node's body contains the lifetime of the allocation, so we only need to keep the mapping in `replacements` for the mutation to the body.

src/CodeGen_D3D12Compute_Dev.cpp
shoaibkamil: Probably, but I just wanted to be safer & reflect the logic in `IRMutator::visit(const Allocate*)`.

shoaibkamil:
> 
> 
> LGTM pending green -- has the D3D12 stuff been pretty thoroughly tested elsewhere? If so, I'll re-enable it on the buildbots after this lands.

I'd like to have at least one other person test this out (@slomp, do you mind trying on yours locally?) before reenabling testing.  I've been running `ctest --build-config Release --output-on-failure --label-regex "correctness|generator|apps" --parallel %NUMBER_OF_PROCESSORS%` on my machines.

apps/camera_pipe/camera_pipe_generator.cpp
shoaibkamil: Changed the comment.

src/CodeGen_D3D12Compute_Dev.cpp
shoaibkamil: Done.

apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -459,11 +459,22 @@ void CameraPipe::generate() {
 
         Var xi, yi, xii, xio;
 
-        /* 1391us on a gtx 980. */
+        /* These tile factors obtain 1391us on a gtx 980. */
+        int tile_x = 28;
```
abadams, line 463: Trailing whitespace is irritating clang-format

slomp:
> > LGTM pending green -- has the D3D12 stuff been pretty thoroughly tested elsewhere? If so, I'll re-enable it on the buildbots after this lands.
> 
> I'd like to have at least one other person test this out (@slomp, do you mind trying on yours locally?) before reenabling testing. I've been running `ctest --build-config Release --output-on-failure --label-regex "correctness|generator|apps" --parallel %NUMBER_OF_PROCESSORS%` on my machines.

Doing that now, but having issues with libpng and friends.

abadams:
> > > LGTM pending green -- has the D3D12 stuff been pretty thoroughly tested elsewhere? If so, I'll re-enable it on the buildbots after this lands.
> > 
> > 
> > I'd like to have at least one other person test this out (@slomp, do you mind trying on yours locally?) before reenabling testing. I've been running `ctest --build-config Release --output-on-failure --label-regex "correctness|generator|apps" --parallel %NUMBER_OF_PROCESSORS%` on my machines.
> 
> Doing that now, but having issues with libpng and friends.

I'll try too.

abadams:
I get the existing "expected label after instruction" failures, and a new failure in performance_boundary_conditions:

```
D:\Halide4\src\runtime\d3d12compute.cpp:2292 Assert failed: buf->dim[i].stride > 0
```

shoaibkamil:
> 
> 
> I get the existing "expected label after instruction" failures, and a new failure in performance_boundary_conditions:
> 
> ```
> D:\Halide4\src\runtime\d3d12compute.cpp:2292 Assert failed: buf->dim[i].stride > 0
> ```

Should be fixed now.

alexreinking:
I am trying to repro windows failures locally.

slomp:
I think it's all good here for me (@shoaibkamil I sent you a direct message with the results of running the apps, the generators and the correctness tests on this branch).

steven-johnson:
Yeah, just waiting for bots to go green

PR #5007, opened 2020-06-05
Fix linear algebra schedules for old x86

abadams:
Vectors of size 2 aren't compatible with the rest of this schedule. Set
the min vector size to 4.


abadams:
Everything expected to pass passed.

PR #5008, opened 2020-06-05
Add support to locate the last index for argmin/argmax.

derek-inteon:
'argmax' and 'argmin' finds the index (and value) of the max/min value in expression. This works fine for cases when there is only one max/minima.  But for the cases when there are multiple max/minimas, the current implementation picks the first found element.  For some use cases, such as in the ONNX specification (https://github.com/onnx/onnx/blob/master/docs/Operators.md#ArgMax), requires finding last max/minima as well. The proposed fix introduces a additional option to argmax/min for support of this use case.

### DIFF
--- a/src/InlineReductions.cpp
+++ b/src/InlineReductions.cpp
@@ -190,7 +190,13 @@ Tuple argmax(const RDom &r, Expr e, const std::string &name) {
     update_tup[value_index] = e;
 
     f(v.free_vars) = initial_tup;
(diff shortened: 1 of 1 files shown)


abadams:
That name is really for debugging purposes - I don't think it's a good idea to have it affect the behavior of the operator. Maybe we could instead add an argmin_last variant that does this, with most of the implementation shared. Or perhaps an enum argument to argmin that specifies how to handle ties.

PR #5010, opened 2020-06-07
Large stack printer objects break llvm debuginfo

### LINKED ISSUE #4996: Windows buildbots hitting assertion failures in LLVM
We're seeing the following error on a few tests:

```
Assertion failed: AfterInsn && "Expected label after instruction", file C:\build_bot\worker\win-64-testbranch-1000\llvm-project\llvm\lib\CodeGen\AsmPrinter\DebugHandlerBase.cpp, line 136
```

These seem to occur in tests that are manually setting the `Target::Debug` feature, and they've been showing up in OpenCL and D3D12Compute.

CC @alexreinking: May be related to Day 1 changes to remove debug info.

abadams:
This fixes several test failures on master with expected-label-after-instruction 

I tracked the error down to instances of large StackPrinter objects in the d3d12 runtime. We probably shouldn't be making large stack allocations anyway, so I avoided these by in one case using malloc, and in the other case (tracing) using a global buffer guarded by a spinlock (trace was already using globals guarded by the same spinlock).

Fixes #4996

### DIFF
--- a/src/runtime/d3d12compute.cpp
+++ b/src/runtime/d3d12compute.cpp
@@ -87,32 +87,46 @@ static void *const user_context = NULL;
 //
 #if HALIDE_D3D12_TRACE
 static volatile ScopedSpinLock::AtomicFlag trace_indent_lock = 0;
+#define TRACE_BUF_SIZE 4096
 static const char trace_indent_pattern[] = "   ";
-static char trace_indent[4096] = {};
-static char trace_buf[4096] = {};
+static char trace_indent[TRACE_BUF_SIZE] = {};
+static char trace_buf[TRACE_BUF_SIZE] = {};
 static int trace_indent_end = 0;
 #define TRACEINDENT ((const char *)trace_indent)
-#define TRACEPRINT(msg) Printer<BasicPrinter, 4096>(user_context, trace_buf) << TRACEINDENT << msg;
+#define TRACEPRINT(msg) { \
+        trace_scope___.lock();                                          \
+        Printer<BasicPrinter, TRACE_BUF_SIZE>(user_context, trace_buf) << TRACEINDENT << msg; \
+        trace_scope___.unlock();                                        \
+}
 struct TraceLogScope {
     void *user_context;
-    TraceLogScope(void *user_context, const char *function) :
-        user_context(user_context) {
-        Printer<BasicPrinter, 4096>(user_context, trace_buf) << TRACEINDENT << "[@] " << function << "\n";
+
+    void lock() {
         while (__atomic_test_and_set(&trace_indent_lock, __ATOMIC_ACQUIRE)) {
         }
+    }
+
+    void unlock() {
+        __atomic_clear(&trace_indent_lock, __ATOMIC_RELEASE);
+    }
+
+    TraceLogScope(void *user_context, const char *function) :
+        user_context(user_context) {
+        lock();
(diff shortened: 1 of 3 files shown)


steven-johnson:
Wow, that is a weird way for stack corruption to get reported.

src/runtime/d3d12compute.cpp
```diff
@@ -86,40 +86,52 @@ class StackPrinter : public Printer<type, length> {
...
+        trace_scope___.lock();                                                                \
+        Printer<BasicPrinter, TRACE_BUF_SIZE>(user_context, trace_buf) << TRACEINDENT << msg; \
+        trace_scope___.unlock();                                                              \
+    }
 struct TraceLogScope {
-    TraceLogScope() {
-        while (__atomic_test_and_set(&indent_lock, __ATOMIC_ACQUIRE)) {
+    void *user_context;
+
+    void lock() {
```
steven-johnson, line 105: nit: it's too bad we can't reuse the guts of ScopedSpinLock here somehow, rather than recaptitulating the __atomic calls

abadams:
It's compiling it that fails. My guess is that large stack frames send llvm down a different path for debug info and that triggers the llvm bug.

src/runtime/d3d12compute.cpp
```diff
@@ -86,40 +86,52 @@ class StackPrinter : public Printer<type, length> {
...
-// a halide_malloc failure, and user code should not affect debug-tracing
-#define trace(x) StackPrinter<4096, BasicPrinter>(x)
-
-static volatile ScopedSpinLock::AtomicFlag indent_lock = 0;
-static const char indent_pattern[] = "   ";
-static char indent[2048] = {};
-static int indent_end = 0;
-#define TRACEINDENT ((const char *)indent)
-#define TRACEPRINT(msg) trace(user_context) << TRACEINDENT << msg;
+static volatile ScopedSpinLock::AtomicFlag trace_indent_lock = 0;
```
steven-johnson, line 89: nit: probably worth adding a comment here about why we are explicitly avoiding StackPrinter

abadams: I just deleted StackPrinter entirely and made Printer use the stack when things are small instead.

src/runtime/d3d12compute.cpp
abadams: Good point. Now uses a ScopedSpinLock.

abadams:
The errors in master were actually in opencl, not d3d12! But it's the same error for the same reason. I removed a large stack allocation and it seemed to heal it.

src/runtime/printer.h
```diff
@@ -32,10 +32,18 @@ class Printer {
...
 
     Printer(void *ctx, char *mem = NULL)
         : user_context(ctx), own_mem(mem == NULL) {
-        buf = mem ? mem : (char *)halide_malloc(user_context, length);
+        if (mem != NULL) {
+            buf = mem;
+        } else if (length <= sizeof(scratch)) {
+            buf = scratch;
+        } else {
+            buf = (char *)malloc(length);
```
shoaibkamil, line 44: Maybe add a comment explaining why not to use halide_malloc?  Otherwise someone will surely change this in a few weeks :)

shoaibkamil:
LGTM, thanks for tracking this down!

PR #5011, opened 2020-06-08
cmake: test: fix PCH compile error with Clang v10

acolinisi:
Clang (10.0) does not like re-using pre-compiled headers across targets
built with and without -pthread compiler arg:

	FAILED: test/correctness/CMakeFiles/correctness_thread_safety.dir/thread_safety.cpp.o

	/usr/lib/llvm/10/bin/clang++ -DHALIDE_ENABLE_RTTI -DWITH_EXCEPTIONS -I../test/common -I../tools
	-Iinclude -O0 -march=native -pthread -std=c++14 -Xclang -include-pch -Xclang
	/.../Halide/build/test/CMakeFiles/_test_internal.dir/cmake_pch.hxx.pch
	-Xclang -include -Xclang /.../Halide/build/test/CMakeFiles/_test_internal.dir/cmake_pch.hxx
	-MD -MT test/correctness/CMakeFiles/correctness_thread_safety.dir/thread_safety.cpp.o
	-MF test/correctness/CMakeFiles/correctness_thread_safety.dir/thread_safety.cpp.o.d
	-o test/correctness/CMakeFiles/correctness_thread_safety.dir/thread_safety.cpp.o
	-c ../test/correctness/thread_safety.cpp

	error: POSIX thread support was disabled in PCH file but is currently enabled

Apparently, GCC (10.1) does not care.

This patch fixes the issue by intorducing another set of pre-compiled
headers for re-use across the targets that are built with -pthread.

### DIFF
--- a/cmake/HalideTestHelpers.cmake
+++ b/cmake/HalideTestHelpers.cmake
@@ -81,6 +81,7 @@ function(tests)
         add_executable("${TARGET}" "${file}")
         target_link_libraries("${TARGET}" PRIVATE Halide::Test)
         if ("${file}" MATCHES ".cpp$")
+            # Note: source for re-use may be overriden by tests_threaded
             target_precompile_headers("${TARGET}" REUSE_FROM _test_internal)
         endif ()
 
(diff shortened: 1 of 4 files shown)


acolinisi:
Didn't see there is also #4990 . In any case this is another way to fix it, fwiw.

steven-johnson:
I just landed #4990 to master -- if that doesn't address the issue, please let us know. (If it *does*, please close this PR.)

PR #5014, opened 2020-06-08
Fix args to nvvm.shfl.up

abadams:
The args to nvvm.shfl.up were wrong

This doesn't actually get buildbot coverage, because it requires a newer GPU than the buildbots have. Tested locally with llvms 9, 10, and master


abadams:
No test coverage of this anyway on the bots, so merging to get it out of the queue

Issue #4113, opened 2019-08-08
Deprecate DisableLLVMLoopVectorize and DisableLLVMLoopUnroll


steven-johnson:
Per the original commit: "There's some evidence that this increases compile time without improving code generation for *properly-scheduled* Halide code. Adding feature flags to allow opting out so that downstream code can experimentally disable it."

At present, some variants of the autoscheduler set these (ie, disables llvm loop opt), but some don't. Since we believe that "properly scheduled" Halide code should be mostly indifferent to this (ie it's just wasted cycles), we should move to disabling them by default; the gotcha (of course) is that there may be downstream code that relies on these for performance, so we'll have to be cautious.

Proposal:
- Add EnableLLVMLoopVectorize and EnableLLVMLoopUnroll features to force these on (as a backdoor for code that may have issues)
- Leave in DisableLLVMLoopVectorize/DisableLLVMLoopUnroll for now (which will continue to force them off)
- Instruct downstream integrators to try adding DisableLLVMLoopVectorize+DisableLLVMLoopUnroll to all their build targets and see if it impacts performance
- After suitable length of time (2 weeks? 1 month?), remove DisableLLVMLoopVectorize+DisableLLVMLoopUnroll and make that the default

benoitsteiner:
FYI: we disable as many LLVM optimizations as possible in order to train a more accurate cost model. For example, it becomes really hard to predict the impact of vectorization when the compiler is allowed to vectorize loops behind your back, and does in a very unpredictable fashion. Therefore when an optimization can be driven directly by Halide through a specific scheduling instruction, we try to give Halide full control over its application.

steven-johnson:
As an update: recent work continues to suggest that the LLVM loop optimizations are generally undesirable:

- they increase compile time
- they rarely improve performance, and when they do, it's usually a suboptimal schedule
- they can add unpredictability to final code size (e.g. because LLVM may decided to unroll or not based on internal metrics that are hard to predict)

Per my initial suggestion above, at this point I think we should move to making llvm-loop-optimizations an opt-in rather than an opt-out.

mcourteaux:
I agree with @steven-johnson here. The code generated is extremely hard to analyze. Especially for CUDA kernels where the code becomes incredibly long and everything remains scalar anyway. Maybe this LLVM loop optimization is less sensible for GPU targets in general.

PR #5019, opened 2020-06-09
Disable LLVM loop optimzation by default (Issue #4113)

steven-johnson:
Up to now, Halide defaulted to having LLVM run its own loop optimization passes. These are designed for typical human-written C/C++ code (and do a good job there), but recent work suggests that it's rarely very useful for Halide-produced code (it tends to expand code size and/or compile time without usually improving performance in any meaningful way). For that reason, Halide should flip to make this 'opt-in' instead of 'opt-out'.

This change removes the DisableLLVMLoopOpt flag entirely. You can still opt-in to the old behavior via EnableLLVMLoopOpt.

(Landing this assumes that most downstream users are OK with this change. I've already made the appropriate changes inside Google and we can land this anytime from our perspective.)

### DIFF
--- a/apps/autoscheduler/autotune_loop.sh
+++ b/apps/autoscheduler/autotune_loop.sh
@@ -63,14 +63,6 @@ else
     echo Copying starting weights from ${START_WEIGHTS_FILE} to ${WEIGHTS}
 fi
 
-# We could add this unconditionally, but it's easier to wade thru
-# results if we only add if needed
-for F in disable_llvm_loop_opt; do
-    if [[ ! ${HL_TARGET} =~ .*${F}.* ]]; then
-        HL_TARGET="${HL_TARGET}-${F}"
-    fi
-done
-
 # A batch of this many samples is built in parallel, and then
 # benchmarked serially.
 BATCH_SIZE=32
--- a/apps/onnx/model.cpp
+++ b/apps/onnx/model.cpp
@@ -344,8 +344,6 @@ std::vector<py::array> run(
     }
     Halide::Realization real(outputs);
     Halide::Target tgt = Halide::get_host_target();
-    // Don't allow LLVM to mess with the code.
-    tgt.set_feature(Halide::Target::DisableLLVMLoopOpt, true);
     // Don't create buffers larger than 2GB since we use 32bit signed indices to
     // index the data stored in them.
     tgt.set_feature(Halide::Target::LargeBuffers, false);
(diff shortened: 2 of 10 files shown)


abadams:
I'm worried about discoverability of this change for people currently using the flag. It will suddenly become an invalid target. Do you think we should briefly keep the flag disable_llvm_loop_opt and have it emit a warning that this flag is now the default behavior and that they should use enable_llvm_loop_opt to restore the old behavior?

Or perhaps the only people using this flag are insiders anyway.

steven-johnson:
> Or perhaps the only people using this flag are insiders anyway.

This is what I'd suspect. But maybe announcing this on halide-dev@ and Gitter to get more input would be worthwhile. (We don't need to be in a hurry to land this.)

steven-johnson:
Hm, maybe this really is moving the needle in this case:

```
cd bin/build/tmp ; /home/halidenightly/build_bot/worker/linux-64-gcc7-testbranch-trunk/halide-Release/bin/performance_boundary_conditions
unbounded           : 88274.299000 us
constant_exterior   : 1574301.813000 us
Error: constant_exterior is 17.834204 times slower than unbounded
/home/halidenightly/build_bot/worker/linux-64-gcc7-testbranch-trunk/halide/Makefile:1921: recipe for target 'performance_boundary_conditions' failed
make: *** [performance_boundary_conditions] Error 255
```

abadams:
I can't repro that slowdown locally. Probably a flake.

abadams:
The sum helper is going to cause problems. You want to unroll them when they're small, but the helper doesn't give you access to the Func. Right now it's fine because LLVM unrolls it anyway. We may need to redesign the helper.

steven-johnson:
Bringing this back up to see if we (finally) want to land this for Halide 13.

dsharletg:
I found that this target flag speeds up many of the correctness tests substantially (e.g. correctness_vector_reductions goes from 42s to 27s).

This change might dramatically speed up buildbot testing if that pattern holds, and I think it would for the correctness tests.

steven-johnson:
Since we're about to cut a branch for Halide 12, I propose we land this for Halide 13, since it's a potentially-breaking change.

abadams:
Sounds good, but I believe there are one or two things to merge into master before we have a complete 12.0 branch.

steven-johnson:
> Sounds good, but I believe there are one or two things to merge into master before we have a complete 12.0 branch.

Yep, just want to get it reviewed so we can land it after the 12 branch is cut

abadams:
Oh sorry, I thought this was a different PR. I'm still concerned that this will cause regressions for any use of the sum helper over small reduction domains, where currently we can rely on llvm unrolling it if sensible to do so. These regressions would be impossible to fix in the schedule because the sum helper sucks that way.

Should we try fixing those helpers first?

steven-johnson:
> Oh sorry, I thought this was a different PR. 

Probably thinking of https://github.com/halide/Halide/pull/5900 (which really should land regardless of this one, hint hint)

> Should we try fixing those helpers first? 

Sure, but I don't recall the discussion you mention -- are you sure it was me?

abadams:
It was a small part of a dev meeting a few months ago, not a direct discussion with you.

PR #5022, opened 2020-06-10
Small wording improvements.

zvookin:


abadams:
Only changes comments so merged it without letting the bots run (to attempt to keep their load low while we shake out some recent changes)

PR #5023, opened 2020-06-10
New simplifier rules necessary for the gpu autoscheduler

abadams:
These cancellations were necessary to get tight bounds in places. 

Also I deleted one parens too many in the previous paren-deletion PR


abadams:
Failure unrelated.

Issue #4967, opened 2020-05-26
Semantics of gpu_thread_barrier()


shoaibkamil:
The intrinsic `gpu_thread_barrier()` currently has different memory fence semantics on different platforms.  We should decide & document what the semantics are.  In addition, if we want fences for global/device memory as well as shared memory, then we need to change backends to respect the semantics correctly.

Currently, I believe the semantics are:
- CUDA: threadgroup barrier, fence for shared & global memory
- OpenCL: threadgroup barrier, fence for shared mem only
- Metal: threadgroup barrier, fence for shared mem only
- D3D: threadgroup barrier, fence for shared mem only
- OpenGLCompute: threadgroup barrier, fence for shared & global memory (_I think_)

abadams:
Ugh, so if we have a Func compute_at gpu blocks, and we elect to store it in MemoryType::Heap instead of MemoryType::GPUShared, because it doesn't fit in shared, then we may not be currently generating correct code?

shoaibkamil:
Yes, I believe that's correct.  It would be correct for CUDA and OpenGLCompute, but not anything else :-/

PR #5025, opened 2020-06-10
Make gpu_thread_barrier() semantics consistent

shoaibkamil:
This PR attempts to address #4967 by implementing a required parameter for `gpu_thread_barrier()`: a mask that determines the memory fence behavior.

Note that this PR does not increase/decrease the number of barriers inserted.  We may need to revisit the logic that determines when to insert barriers, based on other conversations with @slomp and @abadams.

I'll add or modify a test if this approach seems reasonable to everyone.

### DIFF
--- a/src/FuseGPUThreadLoops.cpp
+++ b/src/FuseGPUThreadLoops.cpp
@@ -1057,9 +1057,8 @@ class InjectThreadBarriers : public IRMutator {
 
     using IRMutator::visit;
 
-    Stmt barrier;
-    ExtractSharedAndHeapAllocations &block_allocs;
-    ExtractRegisterAllocations &register_allocs;
+    const ExtractSharedAndHeapAllocations &block_allocs;
+    const ExtractRegisterAllocations &register_allocs;
 
     std::set<std::string> shared_stores;
     std::set<std::string> device_stores;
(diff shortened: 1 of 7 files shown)


src/FuseGPUThreadLoops.cpp
```diff
@@ -1093,6 +1049,155 @@ class ExtractRegisterAllocations : public IRMutator {
...
+                }
+            }
+            for (auto &st : device_stores) {
+                auto elem = device_loads.find(st);
+                if (elem != device_loads.end()) {
+                    mask |= CodeGen_GPU_Dev::MemoryFenceType::Device;
+                    break;
+                }
+            }
+            return Block::make(Block::make(first, make_barrier(mask)), rest);
```
abadams, line 1187: If "first" is a serial for loop and mask is zero, then it already has a barrier after each iteration, and we don't also need a barrier here. But that for loop may have things like LetStmts or ProducerConsumer nodes wrapped around it...

Is there some obvious way to avoid an extra barrier? Does it even matter?

shoaibkamil: Not an obvious way I could think of.  Metal and PTX seem to elide extra barriers, but I'm not so confident about the other GPU languages... it may not matter.

shoaibkamil:
Both of the failures seem unrelated.  Ok to merge?  Or should wait to figure out what's going on with the Windows and arm32 buildbots?

PR #5026, opened 2020-06-10
Combine visit(Cast) for GLSL and OpenGLCompute

steven-johnson:
These are the only two overrides of `visit(Cast) from GLSLBase and they both have identical implementations; combine them into one and move into GLSLBase to save code.


steven-johnson:
Failure is unrelated.

Issue #5024, opened 2020-06-10
OpenGLCompute seems to be broken


steven-johnson:
The recent improvements to typing that were made in GLSL codegen seem to have injected errors into existing code; the issue is that trying to add (sub, mul, etc) a `float` and a `uint` will fail with  something like

```
Could not compile shader:
0(94) : error C7011: implicit cast from "uint" to "float"
0(94) : error C7011: implicit cast from "float" to "uint"
```

The code causing this appears to be something like

```
    uint _vsum_0[1];
    ...
    float _587 = float(_586);
    uint _588 = _vsum_0[int(0)];
    uint _589 = _587 + _588;   // Note, this is float + uint
    _vsum_0[int(0)] = uint(_589);
```

In this case, it appears that previously some of intermediates that are now typed as `uint` were getting typed as `float`, so it happened to work.

I don't have a good public repro case yet, but it looks like we need to add more smarts to GLSL to ensure that binops ensure matching types.

PR #5027, opened 2020-06-10
Fix codegen for absd() in GLSLBase

steven-johnson:
It was emitting as a float, which is *never* correct, since absd() is only used for int or uint types. (This happened to work before because GLSL was previously also incorrectly using float for uint in some cases.)

Also did a drive-by removal of code in Codegen_C that recapitulated the logic from IROperator.cpp; maybe the type field of absd() was incorrect at some point in the past, but this calculation seems redundant and wrong now.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -294,9 +294,6 @@ protected:
             for (auto &a : op->args) {
                 include_lerp_types(a.type());
             }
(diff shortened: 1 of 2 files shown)


steven-johnson:
(Fixes Issue #5024)

steven-johnson:
Failure is unrelated.

Issue #5017, opened 2020-06-09
Merging loops with parallelizable and non-parallelizable reduction variables using compute_with


BachiLi:
We (mainly @Infinoid) have the following code (simplified):
```
f(i, j) = ...;
RDom r(...);
r.where((r.x <  r.y)
         && (r.z <  r.w)
         && (r.x <= r.z)
         && (r.x*N + r.y) < (r.z*N + r.w));
f(r.x,r.y) += g(r.x, r.y, r.z, r.w);
f(r.x,r.z) += g(r.x, r.y, r.z, r.w);
```
`g` is an expensive function, but we don't want to compute root because it requires enormous amount of memory. Instead, we want to merge the two updates, inline g and form the following loop nest
```
for r.w
  for r.z
    for r.y
      for r.x
        if (r.predicate())
          g_ = g(r.x, r.y, r.z, r.w)
          f(r.x, r.y) += g_
          f(r.x, r.z) += g_
```

It seems natural to use compute_with to achieve this:
```
f.update(1).compute_with(f.update(0), r.x);
```
but apparently `compute_with` does not allow merging loops between updates of the same function (why?). We get:
```
Cannot schedule f.update(1) to be computed with f.s1.r$x
```

An arcane trick to get handles for merging the updates it through `rfactor`:
```
Func f0_intm = f.update(0).rfactor({});
Func f1_intm = f.update(1).rfactor({});
f1_intm.update(0).compute_with(f0_intm.update(0), r.x);
```

Now we pass the same function check, but get the following error
```
Invalid compute_with: types of dim 1 of f_intm.s1(r$y is PureRVar) and f_intm$1.s1(r$y is ImpureRVar) do not match.
```

The Pure/Impure judge is correct: for the first update, r.x & r.y are parallelizable and the rest are not. For the second update, r.x & r.z are parallelizable and the rest are not. However, I think this shouldn't be preventing us from merging the loop.

I have two questions for this:
1) Should we allow `compute_with` between update stages for a single Func? What are the conditions that this shouldn't be allowed?
2) Should we allow `compute_with` between different loop nests by "downgrading" the PureRVar into ImpureRVar?

@abadams would love to have your inputs. If you think 1) and 2) are reasonable I can draft a pull request for the change.

abadams:
That's a wild trick. I'm not sure it's going to work out well though - it seems to confuse bounds inference.

[...]

PR #5029, opened 2020-06-11
Make it harder for the associativity test to get confused

abadams:
Minimal fix for #5017 

Longer term we need to be smarter about our representation of +=, and other associative definitions.


BachiLi:
LGTM!

PR #5031, opened 2020-06-11
Fix some "MachO doesn't support COMDAT" issues in runtime

steven-johnson:
Runtime code that will be instantiated for OSX/iOS needs to ensure that there are no plain 'inline' functions -- they must be either WEAK or __attribute__((always_inline)) -- otherwise, some compiler configurations can produce the error above. (Note that this also applies to member functions that are defined inline, even without an explicit 'inline' keyword).

(Note also that the vagaries of C++ mean that declaring a ctor implies that a dtor will be auto-created; in some of these we must explicitly declare the dtor so that it too is always-inlined, even if it is empty...)


abadams:
Ah, so that's why C++ classes in the runtime have behaved oddly. Inline method definitions are implicitly a different linkage type.

PR #5032, opened 2020-06-11
atomic vectorization tweaks

abadams:
Unrelated pieces of the atomic vectorization PR which can be merged separately.

1) Make lossless_cast more aggressive. It now looks through some adds, subs, and muls too
2) You can fuse a PureRVar and an ImpureRVar now. You get an ImpureRVar.
3) Some logic in Deinterleave.cpp handling shuffles was needlessly complex
4) 16-bit floats were never added to associative ops
5) CSE had an error message that benefited from also printing the Expr
6) The Makefile wasn't passing LLVM_VERSION to tests. Some of them care (because expectations about generated code might depend on LLVM version)


abadams:
Failures unrelated.

PR #5034, opened 2020-06-12
Fix inverted assert

abadams:
This assert in cuda_mat_mul is inverted. err == 0 means success.

### DIFF
--- a/apps/cuda_mat_mul/runner.cpp
+++ b/apps/cuda_mat_mul/runner.cpp
@@ -1,5 +1,5 @@
 // We always want asserts
-#ifndef NDEBUG
+#ifdef NDEBUG
 #undef NDEBUG
 #endif
 
@@ -18,7 +18,7 @@ int main(int argc, char **argv) {
     // test doesn't have at least that, quietly skip the test.
 
     fprintf(stderr, "%d\n", __LINE__);
-    fflush(stderr);    
+    fflush(stderr);
 
     const auto *interface = halide_cuda_device_interface();
     assert(interface);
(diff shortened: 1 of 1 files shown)


shoaibkamil:
Oh jeez.

steven-johnson:
...no wonder it's been failing so much :-(

abadams:
Seems like it's still failing. On windows it's being done in release mode, so the assert probably wasn't triggering anyway.

abadams:
Looks like the crash is inside cublas :(

abadams:
Now that the asserts are actually on, I'll want to see if any of them trigger.

abadams:
Also, can we globally disable NDEBUG for all our apps and tests? I don't think we ever want it (there's no use of Runtime::Buffer::operator() inside anything we benchmark)

apps/cuda_mat_mul/runner.cpp
```diff
@@ -1,3 +1,8 @@
+// We always want asserts
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
```
alexreinking, line 4: BTW - it's always safe to unconditionally `#undef NDEBUG`

steven-johnson:
Do we want to land this with the fprintfs (etc) included?

abadams:
No, it's in a temporary state. I can't tell if it has run yet with the asserts on for real.

abadams:
commit history seems to have gotten messed up.  squashing and force-pushing.

steven-johnson:
Failure is unrelated.

steven-johnson:
(I converted this to 'draft' so it doesn't accidentally get landed yet)

abadams:
At this point this PR was just to help diagnose #5053 

Not worth keeping open as a PR.

alexreinking:
@abadams - do you want to keep the `#undef NDEBUG` in this file, though?

abadams:
I'd rather universally make sure we don't have NDEBUG defined for app testing on the buildbots. I can't guarantee that there aren't asserts elsewhere.

PR #5035, opened 2020-06-12
It's worth cancelling correlated subexpressions in load/store indices

abadams:
In particular, this makes warp shuffles much more reliable, because any
dependence of a load or store index on the block id is more likely to
get cancelled out. This PR massively simplifies the generated code for
cuda_mat_mul, and makes it about 30% faster (although it's still
mysteriously 2x slower than cublas on my card).

Also reduces the amount of IR in some other apps very slightly.

Doesn't seem to affect compile times.

### DIFF
--- a/src/SimplifyCorrelatedDifferences.cpp
+++ b/src/SimplifyCorrelatedDifferences.cpp
@@ -109,17 +109,6 @@ class SimplifyCorrelatedDifferences : public IRMutator {
         return visit_let<LetStmt, Stmt>(op);
     }
 
-    Stmt visit(const Store *op) override {
-        // We only care about the expressions that determine the sizes
-        // of allocations and loop extents, so no need to look inside
-        // stores.
-        return op;
-    }
-
-    Stmt visit(const Provide *op) override {
-        return op;
-    }
-
     Stmt visit(const For *op) override {
         Stmt s = op;
         // This is unfortunately quadratic in maximum loop nesting depth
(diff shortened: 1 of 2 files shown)


src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -109,17 +109,6 @@ class SimplifyCorrelatedDifferences : public IRMutator {
         return visit_let<LetStmt, Stmt>(op);
     }
 
-    Stmt visit(const Store *op) override {
```
steven-johnson, line 112: Maybe worth capturing this PR's rationale in a comment here, so no tries to restore this logic in the future.

abadams: I improved and expanded the top-level comment on the pass.

apps/cuda_mat_mul/runner.cpp
```diff
@@ -1,3 +1,8 @@
+// We always want asserts
+#ifndef NDEBUG
+#undef NDEBUG
+#endif
```
alexreinking, line 4: This is definitely wrong. Should be `#ifdef NDEBUG`

abadams: Could have sworn I just fixed that. Must have forgotten to push.

abadams: Ah, that's not part of this PR. Please see #5034, on which this was based. Once that's done I'll rebase this on top of the result.

abadams: Rebased on master instead. This change has been removed from this PR

alexreinking: Great, thanks!

PR #5036, opened 2020-06-15
Constant extents inferred pre-storage flattening

abadams:
Consider an allocation that has a dynamic extent, but needs to have a
constant extent because it's stored in MemoryType::Register (e.g. see
the test). We take an upper bound in these cases to get a constant
allocation size. This PR changes things to take that upper bound
*before* storage flattening instead of after. This way the individual
per-dimension extents are all constant, instead of just their product.
If you do it after storage flattening then you get dynamic strides
within a constant-sized allocation, which is silly and not compatible
with hoisting values into registers anyway (because access is at
non-constant coords).

Also fixed the assumption that MemoryType::Register on the GPU means
that there must be a GPULanes loop.

Builds on #5035

### DIFF
--- a/src/BoundSmallAllocations.cpp
+++ b/src/BoundSmallAllocations.cpp
@@ -71,6 +71,33 @@ class BoundSmallAllocations : public IRMutator {
         return IRMutator::visit(op);
     }
 
+    bool must_be_constant(MemoryType memory_type) const {
+        return (memory_type == MemoryType::Register ||
+                (device_api == DeviceAPI::OpenGLCompute &&
+                 memory_type == MemoryType::GPUShared));
+    }
+
+    Stmt visit(const Realize *op) override {
+        // Called pre-storage-flattening. At this point we just want
+        // to ensure any extents on allocations which *must* be
+        // constant *are* constant.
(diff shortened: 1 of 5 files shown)


apps/cuda_mat_mul/runner.cpp
```diff
@@ -1,3 +1,8 @@
+// We always want asserts
+#ifndef NDEBUG
```
steven-johnson, line 2: Shouldn't this be `ifdef` rather than `ifndef`

abadams: Oops. This is a bug in the underlying branch so I'll fix it there.

abadams:
Can't repro the failure. 99.9% sure it's a flake. I'm going to rebase one last time and merge it.

abadams:
btw I had to rebase again because the old version of the same commits got merged into this branch. I think when you pulled it you also merged the rebased commits and the old commits.

I generally avoid rebasing and force-pushing to avoid this sort of nonsense. I don't recall why I did it in this case - I think to give the same commits a new hash to kick the buildbots.

PR #5041, opened 2020-06-16
Add an explicit pass to lift loop invariant if statements

abadams:
If statements can be injected by GuardWithIf, RDom predicates,
specializations, and uses of undef. There are various situations where
an if statement can end up further inside a loop nest than strictly
necessary. This PR adds a pass to hoist them.

This results in slightly better codegen for some conv layer schedules on
GPU.

Also reduced the expr count in lots_of_loop_invariants because it spends
a long time inside LLVM


abadams:
Rebased on master and force-pushed in an attempt to get the buildbots to notice it

steven-johnson:
Looks all green. Ready to land?

Issue #5040, opened 2020-06-16
Support cross-compiling Halide


shoaibkamil:
We should support cross-compiling Halide for a different architecture.  Concretely, I'm trying to build Halide for arm64-windows.  After more wrangling than I'd like, I've managed to get a cross-compiled LLVM for arm64-windows, following these steps:

1) Install the appropriate toolchain from the VC installer
2) Build LLVM from source using directions on the Halide webpage, but don't install.  Here, I'm using LLVM 10.
2) `vcpkg env --triplet arm64-windows`
3) `cmake -G Ninja -DCMAKE_CROSSCOMPILING=True -DCMAKE_INSTALL_PREFIX=<install-dir> -DLLVM_TABLEGEN=d:/llvm-x64-build/bin/llvm-tblgen -DCLANG_TABLEGEN=d:/llvm-x64-build/clang-tblgen -DLLVM_DEFAULT_TARGET_TRIPLE=aarch64-pc-windows-mscvc -DLLVM_TARGET=aarch64 -DENABLE_DIA_SDK=off <rest of options from halide webpage>`
Note here that enabling the DIA SDK will cause the build to fail.
4) `cmake --build . --config release --target install --parallel %NUMBER_OF_PROCESSORS%`
5) Build fails because it attempts to use the newly-built llvm-nm.exe for arm64.  Manually edit `build.ninja`, search for `--nm`, and point it to the x64 one built earlier instead
6) `cmake --build . --config release --target install --parallel %NUMBER_OF_PROCESSORS%`

The build should succeed then.  However, trying to build Halide:
1) `cmake -G Ninja -DCMAKE_CROSSCOMPILING=True -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=d:/halide/vcpkg/scripts/buildsystems/vcpkg.cmake -DLLVM_DIR=d:/halide/llvm-arm-install/lib/cmake/llvm -DClang_DIR=d:/halide/llvm-arm-install/lib/cmake/clang -DWITH_PYTHON_BINDINGS=off ..\halide`
2) `cmake.exe --build . --config release --parallel 5` fails with:
```
[9/3549] Generating _initmod_HalidePyTorchCudaHelpers_h.cpp
FAILED: src/runtime/_initmod_HalidePyTorchCudaHelpers_h.cpp
[...]

PR #5042, opened 2020-06-16
Add preliminary AOT Windows ARM64 support

shoaibkamil:
This was tested manually by:
1) Using a generator to generate a static library
2) Compiling the library along with a C++ file using the library using Microsoft's VC++ toolchain for ARM64
3) Copying the resulting file and running it.

Note that in this PR, `arm-64-windows-jit` is explicitly disallowed.

I did this for a few cases in `test/generator`, but would of course like this to be better tested.  The main hurdles are:
- Microsoft does not have an official toolchain that _runs_ on ARM64 devices.  One must use an x64 device Windows machine to compile for ARM64
- I haven't been able to cross-compile Halide for Windows ARM64 (#5040), which potentially could enable JIT (though some issues remain with upstream LLVM cross-compilation, e.g. [vcpkg cannot correctly cross-compile LLVM for Windows ARM64](https://github.com/microsoft/vcpkg/issues/11973) )


steven-johnson:
> Microsoft does not have an official toolchain that runs on ARM64 devices. One must use an x64 device Windows machine to compile for ARM64

I will once again beat the same dead horse and say that we should really have a test suite that is capable of running ~entirely on AOT (aside from tests that are truly JIT-specific, of course).

shoaibkamil:
@steven-johnson Somehow I missed this dead horse, but I would definitely support having an AOT test suite.  It would make testing much easier for iOS and possibly Android.

Presumably this would be the tests from `test/correctness` adapted to run AOT?

steven-johnson:
> Presumably this would be the tests from `test/correctness` adapted to run AOT?

Yes, along with `error` `warning` `performance` etc. Aside from cross-compilation, it would make testing of WebAssembly vastly less painful (and more complete), as we wouldn't have to embed a WASM runtime of some sort into Halide.

shoaibkamil:
Not sure why buildbots didn't run, but this won't be tested anyway.

PR #5043, opened 2020-06-16
Make check_hash() a macro rather than an inline function

steven-johnson:
Apparently some compiler configs can refuse to inline even an always-inline function (or inline it in a way that causes linker errors). This is simple enough for a macro to be fine anyway.

### DIFF
--- a/src/runtime/synchronization_common.h
+++ b/src/runtime/synchronization_common.h
@@ -497,9 +497,12 @@ struct hash_table {
 WEAK char table_storage[sizeof(hash_table)];
 #define table (*(hash_table *)table_storage)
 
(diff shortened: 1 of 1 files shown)


abadams:
Curious it affects this function and not others. I notice HALIDE_ALWAYS_INLINE also declares the linkage class "inline", and in other places we have weak + always inline. Does it fix it to change all instances of __attribute__((always_inline)) to HALIDE_ALWAYS_INLINE in this file?

steven-johnson:
Yeah, making it `inline __attribute__((always_inline))` does fix it -- must be a linkage-type issue. I'll withdraw this change, and instead just do a pass over all the runtime to make sure we declare those in tandem.

PR #5044, opened 2020-06-16
Add ALWAYS_INLINE and WEAK_INLINE to runtime_internal.h

steven-johnson:
Clean up haphazard function attributes in runtime code:

- Runtime code that uses `__attribute__((always_inline))`, but not `__attribute__((weak))`, should always also use `inline` (otherwise, some compilation scenarios can report link errors due to confusion over linkage types). ALWAYS_INLINE is the runtime-specific macro now used uniformly used for these cases.

- Runtime code that uses `__attribute__((always_inline))` and `__attribute__((weak))` should *never* use `inline`. WEAK_INLINE is the runtime-specific macro now used uniformly used for these cases.

### DIFF
--- a/src/runtime/runtime_internal.h
+++ b/src/runtime/runtime_internal.h
@@ -30,7 +30,32 @@ typedef __PTRDIFF_TYPE__ ptrdiff_t;
 typedef ptrdiff_t ssize_t;
 
 #define NULL 0
+
+// --------------
+
+// In Halide runtime code, most functions should just be WEAK, whether or not
+// they're part of the public API.
+//
+// ALWAYS_INLINE is for things that either should be inlined for performance
+// reasons, or for things that go into every compiled pipeline (not just the
+// standalone runtime), as those things have to either disappear entirely by
+// being inlined away, or have "inline" linkage to avoid multiple definition
+// errors on platforms that have no weak linkage.
+//
+// WEAK_INLINED is a special case where we are 'inlining' the bitcode at
+// bitcode-compilation time (rather than C++ compilation time); it's needed
+// for a few places in the runtime where we can't inline in the traditional
+// way.
+
 #define WEAK __attribute__((weak))
+
+// Note that ALWAYS_INLINE should *always* also be `inline`.
+#define ALWAYS_INLINE inline __attribute__((always_inline))
+
+// Note that WEAK_INLINE should *not* also be `inline`
+#define WEAK_INLINE __attribute__((weak, always_inline))
+
+// --------------
 
 #ifdef BITS_64
 #define INT64_C(c) c##L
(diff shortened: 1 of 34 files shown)


abadams:
I'm 99% confident that WEAK_INLINE should just be ALWAYS_INLINE, and the only reason these were weak was to avoid the same linker errors.

steven-johnson:
> I'm 99% confident that WEAK_INLINE should just be ALWAYS_INLINE, and the only reason these were weak was to avoid the same linker errors.

"Trust But Verify" -- I'll open a branch here to test that theory (and also inside google)

abadams:
Looks like it works to just have those two be WEAK_INLINE and everything else be ALWAYS_INLINE

abadams:
oh, plus HALIDE_BUFFER_HELPER_ATTRS needs to be weak apparently

steven-johnson:
Agreed that it seems to work, but is that what we actually want? (ie, does it actually make a difference in codegen in terms of expansion?)

Wondering if alternate names for the macros would make sense, eg

WEAK -> HALIDE_RUNTIME_API
ALWAYS_INLINE -> HALIDE_RUNTIME_INTERNAL

abadams:
weak + always_inline should not be any different to inline + always_inline in the final code. All callers of them just inline them. But because we construct and combine the initial modules in stages there's a risk of losing the inline + always_inline functions somewhere along the way (the buffer helpers), or it not being legal for them to be inline because they're also extern (halide_abort()) in some other modules. We could make everything weak + always_inline, but there are contexts where that's not allowed either.

The always_inline ones are the symbols that go into every compilation (even with no_runtime). The weak things are the symbols that go into a standalone runtime. That's not quite api vs internal, but it's close. We really use the namespaces for API vs internal.

steven-johnson:
> oh, plus HALIDE_BUFFER_HELPER_ATTRS needs to be weak apparently

this has `#ifdef COMPILING_HALIDE_RUNTIME` at the top, but when do we ever compile it *not* for the runtime?

steven-johnson:
PTAL

src/runtime/runtime_internal.h
```diff
@@ -30,8 +30,21 @@ typedef __PTRDIFF_TYPE__ ptrdiff_t;
 typedef ptrdiff_t ssize_t;
 
 #define NULL 0
+
+// In Halide runtime code, WEAK should be used almostly exclusively for
+// functions that are part of the public Halide runtime API.
 #define WEAK __attribute__((weak))
 
+// In Halide runtime code, helper functions that are internal to the runtime
```
abadams, line 38: This comment isn't quite right - most functions should just be weak, whether or not they're part of the public API. always_inline is for things that either should be inlined for performance reasons, or for things that go into every compiled pipeline (not just the standalone runtime). Those things have to either disappear entirely by being inlined away or have "inline" linkage to avoid multiple definition errors on platforms that have no weak linkage.

src/runtime/x86_cpu_features.cpp
```diff
@@ -7,7 +7,7 @@ namespace Internal {
 
 extern "C" void x86_cpuid_halide(int32_t *);
 
-static __attribute__((always_inline)) void cpuid(int32_t fn_id, int32_t *info) {
+static ALWAYS_INLINE void cpuid(int32_t fn_id, int32_t *info) {
```
abadams, line 10: This is now "static inline" linkage. I'm not sure what that means. Probably safe to remove the "static".

src/runtime/runtime_internal.h
steven-johnson: PTAL

src/runtime/x86_cpu_features.cpp
steven-johnson: Fixed

PR #5056, opened 2020-06-19
Add an HL_MEMORY_LIMIT environment var for the autoscheduler

abadams:
This bounds the maximum amount of memory a generated pipeline may
allocate. If you set it too low the autoscheduler's search can run out
of legal states. Increasing HL_BEAM_SIZE may help it get out of the
hole, or there may simply not be any schedule that hits that memory
limit that doesn't also do a ridiculous amount of recompute.

Environment variables aren't great, but we don't have another means to
pass config to the autoscheduler right now.

This was a feature request from the gitter.


steven-johnson:
still getting that 'boundary conditions 17x slower than expected' -- something is up at this point.

abadams:
Yeah, investigating now

abadams:
Still can't repro locally with any target. I'll try updating llvm - maybe it's a regression in trunk. That would make sense as it seems to be confined to the trunk builder.

abadams:
Can't repro at the exact same llvm commit. But also the fast one "unbounded" is 10x slower than it is on my machine, and this is a single-threaded schedule. 10x shouldn't be possible. That test seems to have consistently failed on that buildbot since the gcc7 update ~4 days ago. I can't find any instances of it passing in that builder. Puzzling...

abadams:
Could not repro in a separate PR either. Rerunning that build.

zvookin:
How do you feel about moving toward a naming convention for environment variables that has three cases:

- Affects compilation.
- Affects runtime.
- Affects autoscheduler.

Could also add a case for debug logging, but mostly I want it to be clear which *phase* of stuff it applies to so people don't set an env variable expecting it to do something and then it does not. E.g. ```HL_MEMORY_LIMIT``` really sounds like it should limit how much memory filters are alowed to use for some purpose, or maybe the compiler itself.

abadams:
HL_AUTOSCHEDULE_* ?

steven-johnson:
green, ready to land?

abadams:
Yes. Needs approval though.

PR #5061, opened 2020-06-22
Add is_nan_f32 for metal

shoaibkamil:
Apparently nobody's ever used is_nan_f32 from a metal pipeline yet.


abadams:
This normally goes with is_inf and is_finite

zvookin:
Worth adding a test that makes sure these exist? Or maybe file an issue if that adds too much work.

shoaibkamil:
> Worth adding a test that makes sure these exist? Or maybe file an issue if that adds too much work right now.

Good point.  I modified the existing isnan test to also run on GPU.

steven-johnson:
Since buildmaster is still down, IMHO this is OK to land once you have confident Mac builds/tests in place locally.

PR #5063, opened 2020-06-22
Apply simplify_correlated_differences to comparisons too

abadams:
GuardWithIf conditions can include comparisons with correlated values on
each side, e.g. `((x*8) + y)*2 + z <= x*16`. It's worth cancelling them -
in one case @aekul found it avoids an if statement tree in an unrolled
block.

This would be a fairly trivial change by itself, but I did some light
refactoring while I was there to avoid blowing up the number of template
instantiations.


abadams:
Seems to be breaking on metal. Investigating.

abadams:
Can't repro locally, and didn't repro with the makefile build. Possibly a flake. Rerunning that build

steven-johnson:
Looks good to land.

abadams:
Still needs a review

PR #5065, opened 2020-06-22
better error message if you try to build an app before libHalide

### LINKED ISSUE #5060: 'distrib' directory missing in repo root
This prevents apps from building, see https://github.com/halide/Halide/blob/85218960d4df43622711ed99c4a64e53125d9b74/apps/support/Makefile.inc#L6

abadams:
Fixes #5060

### DIFF
--- a/apps/support/Makefile.inc
+++ b/apps/support/Makefile.inc
@@ -4,7 +4,6 @@ MAKEFLAGS += --no-builtin-rules
 .SUFFIXES:
 
 HALIDE_DISTRIB_PATH ?= $(realpath ../../distrib)
-$(echo $HALIDE_DISTRIB_PATH)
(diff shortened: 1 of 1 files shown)


apps/support/Makefile.inc
```diff
@@ -4,11 +4,15 @@ MAKEFLAGS += --no-builtin-rules
 .SUFFIXES:
 
 HALIDE_DISTRIB_PATH ?= $(realpath ../../distrib)
+$(echo $HALIDE_DISTRIB_PATH)
```
steven-johnson, line 7: Do we want to leave this $echo in here?

abadams: oops, that was debugging code

abadams: Removed.

PR #5066, opened 2020-06-23
Fix squared size error in gpu argument arrays

abadams:
The scalar arguments to GPU calls are put in a stack array. For n args,
this stack array had type foo[n][n] instead of foo[n]. For large number
of arguments this overflows the stack. @aekul found a case with 40MB
worth of stack space reserved for arguments, of which a tiny fraction
was actually used.


steven-johnson:
Green, ready to land

PR #5073, opened 2020-06-25
Add support to build and run on Apple silicon

Developer-Ecosystem-Engineering:
Changes necessary to run successfully on Apple silicon.

If you'd like any adjustments or have feedback please let us know!

```
% xcrun cmake -DCMAKE_BUILD_TYPE=Release \
  -DLLVM_DIR=/Users/deco/Documents/llvm-install/lib/cmake/llvm \
  -DCMAKE_CXX_FLAGS='-arch arm64' -DCMAKE_C_FLAGS='-arch arm64' \
  ../Halide
% xcrun cmake --build . -- -j8

% file src/libHalide.dylib 
src/libHalide.dylib: Mach-O 64-bit dynamically linked shared library arm64

% file src/libHalide.dylib 
src/libHalide.dylib: Mach-O 64-bit dynamically linked shared library arm64
% file test/correctness/correctness_gameoflife 
test/correctness/correctness_gameoflife: Mach-O 64-bit executable arm64
% test/correctness/correctness_gameoflife     
Success!
% test/performance/performance_jit_stress 
Success!
% test/performance/performance_memcpy 
Success!
```

### DIFF
--- a/cmake/HalideTestHelpers.cmake
+++ b/cmake/HalideTestHelpers.cmake
@@ -23,17 +23,6 @@ if (NOT TARGET Halide::Test)
                                INTERFACE
                                ${Halide_SOURCE_DIR}/test/common
                                ${Halide_SOURCE_DIR}/tools)
-
-    # Tests are built with the equivalent of OPTIMIZE_FOR_BUILD_TIME (-O0 or /Od).
-    # Also allow tests, via conditional compilation, to use the entire
-    # capability of the CPU being compiled on via -march=native if the
-    # the compiler supports it. This also presumes tests are run on the
-    # same machine they are compiled on.
-    target_compile_options(Halide_test INTERFACE
-                           $<$<CXX_COMPILER_ID:MSVC>:/Od>
-                           $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-O0>
-                           $<$<CXX_COMPILER_ID:Intel>:-march=native>
-                           $<$<CXX_COMPILER_ID:GNU>:-march=native>)
 endif ()
 
 if (NOT TARGET Halide::ExpectAbort)
--- a/cmake/HalideTestHelpers.cmake
+++ b/cmake/HalideTestHelpers.cmake
@@ -26,12 +26,14 @@ if (NOT TARGET Halide::Test)
 
     # Tests are built with the equivalent of OPTIMIZE_FOR_BUILD_TIME (-O0 or /Od).
     # Also allow tests, via conditional compilation, to use the entire
-    # capability of the CPU being compiled on via -march=native. This
-    # presumes tests are run on the same machine they are compiled on.
+    # capability of the CPU being compiled on via -march=native if the
+    # the compiler supports it. This also presumes tests are run on the
+    # same machine they are compiled on.
     target_compile_options(Halide_test INTERFACE
                            $<$<CXX_COMPILER_ID:MSVC>:/Od>
                            $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-O0>
-                           $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-march=native>)
+                           $<$<CXX_COMPILER_ID:Intel>:-march=native>
+                           $<$<CXX_COMPILER_ID:GNU>:-march=native>)
 endif ()
 
 if (NOT TARGET Halide::ExpectAbort)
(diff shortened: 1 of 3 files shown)


steven-johnson:
Now all we need is appropriate hardware to test this on :-)

abadams:
Thanks! There's zero buildbot coverage of this of course so I don't want to dump it in the buildbot queue. Builds fine for me on linux. @alexreinking does the cmake change seem reasonable to you?

cmake/HalideTestHelpers.cmake
```diff
@@ -26,12 +26,14 @@ if (NOT TARGET Halide::Test)
...
-    # presumes tests are run on the same machine they are compiled on.
+    # capability of the CPU being compiled on via -march=native if the
+    # the compiler supports it. This also presumes tests are run on the
+    # same machine they are compiled on.
     target_compile_options(Halide_test INTERFACE
                            $<$<CXX_COMPILER_ID:MSVC>:/Od>
                            $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-O0>
-                           $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-march=native>)
+                           $<$<CXX_COMPILER_ID:Intel>:-march=native>
+                           $<$<CXX_COMPILER_ID:GNU>:-march=native>)
```
alexreinking, line 36: tbqh we should just remove this because it's going to make cross compiling Halide harder.

alexreinking:
Looks fine to me. That change shouldn't break anything.

cmake/HalideTestHelpers.cmake
```diff
@@ -26,12 +26,14 @@ if (NOT TARGET Halide::Test)
...
     # Also allow tests, via conditional compilation, to use the entire
-    # capability of the CPU being compiled on via -march=native. This
-    # presumes tests are run on the same machine they are compiled on.
+    # capability of the CPU being compiled on via -march=native if the
+    # the compiler supports it. This also presumes tests are run on the
+    # same machine they are compiled on.
     target_compile_options(Halide_test INTERFACE
                            $<$<CXX_COMPILER_ID:MSVC>:/Od>
                            $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-O0>
-                           $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-march=native>)
```
zvookin, line 34: This seems to be turning off -march=native for clang everywhere. Did clang remove the flag or am I misunderstanding the change?

The decision on whether to use this flag has a bit of history and it could go either way, but it seems we should just remove it entirely if we're going to turn it off for clang. (The issue is it chooses to use the architecture of the host compilation system, which may be a higher rev of the instruction set than the one the tests run on. This makes the tests track the hardware they're run on, which is sometimes an advantage in better coverage and sometimes an inconvenience in debugging a test failure between systems.)

alexreinking: It doesn't affect the JIT runtime, though. So having a `host` target is different from `-march=native`. One could add clang back to the list via `$<$<CXX_COMPILER_ID:Clang>:-march=native>`, noting that Clang and AppleClang are _not_ the same compiler as far as CMake is concerned.

alexreinking:
> Looks like it won't break anything, though I'm wondering if the -march=native change affects buildbot testing. Perhaps clang never supported the flag and it is just a bug fix?

Vanilla Clang _does_ support that flag. I suspect the issue is with Apple's distribution of Clang. CMake even recognizes it as a separate compiler.

zvookin:
If it is a difference between Apple's clang and default, then this should probably be changed to exempt MSVC and Apple clang, or at least add clang back in. (This is in the interest of making the PR have minimal changes on anything else. If we want to nuke -march=native, I'm fine with that, but let's do it in a different PR.)

abadams:
IIRC several compilers support march=native for x86 targets only, so it's probably a backend difference rather than an apple clang vs clang difference

steven-johnson:
In any event, seems like the march=native issue should be cleared up in a subsequent PR and not block this one.

alexreinking:
We could also ask @Developer-Ecosystem-Engineering why that change was needed?

Developer-Ecosystem-Engineering:
> We could also ask @Developer-Ecosystem-Engineering why that change was needed?

It was necessary to cross compile at the time; you can build on the DTK without that specific change. We had assumed folks would likely prefer to cross compile so opt'd to leave it to support that case.  I can remove it in the PR if you'd prefer, let me know

alexreinking:
See #5074. I'm just going ahead and fixing the CMake test flags.

abadams:
ok, we have no removed the march=native flag entirely, which should slightly simplify this PR once the conflict is resolved

PR #5074, opened 2020-06-26
Remove march=native from tests build

alexreinking:
This breaks cross compiling on Apple Silicon and probably elsewhere, too. See PR #5073

Also removing disable-optimization flags here. @steven-johnson and I think it's unimportant and it just makes the build have more weird edge cases.


abadams:
Having debug symbols on for the tests is important (otherwise introspection isn't enabled so the .stmt files are much harder to read), but I don't think the optimization level matters all that much. Let's see if it makes a difference for total build time on the bots.

abadams:
Difference in build time seems firmly in the noise, and total build time for the tests is not a huge fraction of buildbot time anyway.

PR #5076, opened 2020-06-27
Add COMDATs to mergable symbols on windows

abadams:
Should fix the build for win-32-trunk

Having a linkonce linkage type on windows without an associated comdat to let the linker know it can pick anything produces linker errors with missing symbols. This was not previously a problem on windows because llvm just emitted linkonce symbols as extern.


abadams:
Failure unrelated

PR #5083, opened 2020-06-30
Work around LLVM 9 horizontal op issue

### LINKED ISSUE #5081: arm-64 test failure with llvm 9
Some vector reductions are triggering an internal llvm error:

https://buildbot.halide-lang.org/master/#/builders/26/builds/2

It may not be worth a workaround, as we're going to drop llvm 9 support soon anyway (when llvm 11 is released).

abadams:
Fixes #5081


abadams:
Sorry, should have marked this as a draft - it didn't actually fix the problem!

abadams:
I won't revert just yet, because it's probably one part of fixing the problem, but I'll reopen the bug

PR #5084, opened 2020-06-30
Check for null in for_each_value

### LINKED ISSUE #5077: Halide::Runtime::Buffer::for_each_value segfaults on an empty buffer
The following code crashes:
```
  Halide::Runtime::Buffer<int> b;
  b.for_each_value([](int v) {});
```

`for_each_value_helper` doesn't check that ptrs is a valid pointer before dereferencing [here](https://github.com/halide/Halide/blob/master/src/runtime/HalideBuffer.h#L1989).

(and a newly-created or empty buffer has data() == nullptr: https://github.com/halide/Halide/blob/master/src/runtime/HalideBuffer.h#L2073)

This behavior is surprising, and maybe it would be better to never call the callback for an empty buffer, e.g. by checking that `ptrs` is non-null in for_each_value_helper or for_each_value_impl?

abadams:
This gives a better error message than a segfault if someone tries to
iterate over each value in an undefined Buffer.

Another option was just making for_each_value a no-op, but that felt
like it might hide a real error.

Fixes #5077

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -2010,6 +2010,19 @@ private:
     template<int N>
     HALIDE_NEVER_INLINE static bool for_each_value_prep(for_each_value_task_dim<N> *t,
                                                         const halide_buffer_t **buffers) {
+        // Check the buffers all have clean host allocations
+        for (int i = 0; i < N; i++) {
+            if (buffers[i]->device) {
+                assert(buffers[i]->host &&
+                       "Buffer passed to for_each_value has device allocation but no host allocation. Call allocate() and copy_to_host() first");
+                assert(!buffers[i]->device_dirty() &&
+                       "Buffer passed to for_each_value is dirty on device. Call copy_to_host() first");
+            } else {
+                assert(buffers[i]->host &&
+                       "Buffer passed to for_each_value has no host or device allocation");
+            }
+        }
+
         const int dimensions = buffers[0]->dimensions;
 
         // Extract the strides in all the dimensions


src/runtime/HalideBuffer.h
```diff
@@ -2060,6 +2060,7 @@ class Buffer {
 
     template<typename Fn, typename... Args, int N = sizeof...(Args) + 1>
     void for_each_value_impl(Fn &&f, Args &&... other_buffers) const {
+        assert(data() != nullptr && "Called for_each_value on an undefined Buffer");
```
dsharletg, line 2063: There is also other_buffers.data()...

abadams:
Added checks for other_buffers.data(). Do you know a cleaner way to do a variadic reduction pre c++17?

abadams:
I think calling for_each_value on an undefined empty buffer is probably a bug elsewhere.

src/runtime/HalideBuffer.h
```diff
@@ -2048,33 +2059,29 @@ class Buffer {
...
         const halide_buffer_t *buffers[] = {&buf, (&other_buffers.buf)...};
-        bool innermost_strides_are_one = Buffer<>::for_each_value_prep(t, buffers);
-
-        Buffer<>::for_each_value_helper(f, dimensions() - 1,
-                                        innermost_strides_are_one,
-                                        t,
-                                        data(), (other_buffers.data())...);
+        bool innermost_strides_are_one, buffer_is_zero_size;
+        Buffer<>::for_each_value_prep(t, buffers, &innermost_strides_are_one, &buffer_is_zero_size);
+        if (!buffer_is_zero_size) {
```
dsharletg, line 2079: This implementation seems OK, but I was thinking of a much simpler change when I suggested it:

    assert(data() != nullptr || number_of_elements() == 0);

(But with some way of checking other_buffers.data()... as well, which your previous commit did.)

dsharletg: Actually reviewed it now. I think you could keep this approach, but not actually return buffer_is_zero_size, and just use it to skip the assert. It's safe to just let it run for_each_value_helper with `t = {0...}`. That would make this a smaller change.

abadams:
It's actually not safe to let it run with number_of_elements zero, which is why the implementation looks like it does, but I've just realized it's still wrong.

Consider the zero-dimensional case.

abadams:
There's no difference in representation between a zero-dimensional buffer with no allocation of size one and a zero-dimensional buffer with no allocation of size zero, but we want one to assert and the other to be a no-op.

abadams:
Rolled it back to the previous commit. I really think making for_each_value on a default-constructed Buffer is a foot-gun

zvookin:
Do you consider it likely a main reason for this to come up is a device only allocation? If so mentioning that possibility in the error message might be useful.

abadams:
Yeah it's quite possible. I'll subdivide the error messages depending on the device state.

src/runtime/HalideBuffer.h
```diff
@@ -2010,6 +2010,19 @@ class Buffer {
     template<int N>
     HALIDE_NEVER_INLINE static bool for_each_value_prep(for_each_value_task_dim<N> *t,
                                                         const halide_buffer_t **buffers) {
+        // Check the buffers all have clean host allocations
+        for (int i = 0; i < N; i++) {
+            if (buffers[i]->device) {
+                assert(!buffers[i]->device_dirty() &&
```
zvookin, line 2016: I think if you flip these two asserts you can eliminate the suggestion to call "allocate" in the device_dirty case.

abadams: Good point. Done

abadams:
squashed and rebased, because the number of commits for such a simple change was getting silly

PR #5087, opened 2020-07-06
Always cast objc_msgSend()

shoaibkamil:
On macOS 10.15 and later, the prototype for `objc_msgSend()` changed to force code to always cast before using, since the implementation is ABI-dependent (see [Apple's docs](https://developer.apple.com/documentation/objectivec/1456712-objc_msgsend), [Mike Ash's explanation](https://www.mikeash.com/pyblog/objc_msgsends-new-prototype.html), [Daniel Jalkut's](https://www.mikeash.com/pyblog/objc_msgsends-new-prototype.html), etc.)

This PR changes all our runtime uses to cast method calls to the right type before using, and fixes a runtime error on macOS 11 beta.  This should have no impact on previous versions of macOS and iOS; I've confirmed current iOS works as well as macOS 10.15.5; I believe the buildbot should confirm compatibility for older macOS.

### DIFF
--- a/src/runtime/metal.cpp
+++ b/src/runtime/metal.cpp
@@ -46,7 +46,7 @@ WEAK mtl_command_queue *new_command_queue(mtl_device *device) {
 
 WEAK mtl_command_buffer *new_command_buffer(mtl_command_queue *queue, const char *label, size_t label_len) {
     objc_id label_str = wrap_string_as_ns_string(label, label_len);
-    
+
     typedef mtl_command_buffer *(*new_command_buffer_method)(objc_id queue, objc_sel sel);
     new_command_buffer_method method = (new_command_buffer_method)&objc_msgSend;
     mtl_command_buffer *command_buffer = (mtl_command_buffer *)(*method)(queue, sel_getUid("commandBuffer"));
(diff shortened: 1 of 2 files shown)


src/runtime/metal.cpp
```diff
@@ -39,13 +39,17 @@ WEAK mtl_buffer *new_buffer(mtl_device *device, size_t length) {
 }
 
 WEAK mtl_command_queue *new_command_queue(mtl_device *device) {
-    return (mtl_command_queue *)objc_msgSend(device, sel_getUid("newCommandQueue"));
+    typedef mtl_command_queue *(*new_command_queue_method)(objc_id dev, objc_sel sel);
+    new_command_queue_method method = (new_command_queue_method)&objc_msgSend;
+    return (mtl_command_queue *)(*method)(device, sel_getUid("newCommandQueue"));
```
abadams, line 44: Do you need to dereference the function pointer? I thought that was a no-op. Also after the function pointer cast, isn't the return type already correct?

```
    return method(device, sel_getUid("newCommandQueue"));
```

Also please add a comment explaining why the cast is necessary.

shoaibkamil: Will add a comment about casting the `objc_msgSend` calls.

For the function pointer deref and the return type cast, I followed what the other functions in `metal.cpp` were doing (I think written by @zvookin?), just to make the changes minimal.  Happy to go back and change them all to not do either of those things, if that's preferable.

abadams: Favoring consistency with existing code seems reasonable to me.

abadams:
Not sure I understand the comment. If OS X changes the signature or calling convention, won't our runtime be wrong? In what sense will things be caught at compile-time?

shoaibkamil:
> Not sure I understand the comment. If OS X changes the signature or calling convention, won't our runtime be wrong? In what sense will things be caught at compile-time?

Let me figure out how to rephrase.  What I was trying to say was: with this signature (which matches what is now the signature in macOS headers) if we don't do the dance of cast-to-correct-function-type-then-call, the compiler will complain.  Previously, since it was declared as a variadic function with at least two parameters, we could call `objc_msgSend(some_objc_id, some_selector)` without the compiler complaining.  But we _want_ it complain if we try to do that now.

abadams:
I see that it forces you to specify the signature at the call-site, but I don't get why that's less error prone than specifying the signature at the declaration. I think I'm missing the way in which this function is special.

shoaibkamil:
Expanded into something hopefully a bit clearer now.

Issue #5089, opened 2020-07-07
cannot generate objects files for multiple target using Halide generator


benzwt:
The tutorial-15 generates 3 functions with different target and pack into a single static library along with halide runtime.
I would like to make the function auto-dispatch by Halide.

@abadams told me to do like this
```
./lesson_15_generate \
    -g my_first_generator \
    -f my_first_generator \
    -e object,c_header\
    -o . \
    target=x86-64-avx-no_runtime,x86-64-sse41-no_runtime,x86-64-no_runtime
```
However, but the program exited with core dump.
```
Error: (at ../tools/GenGen.cpp:4)
Cannot request object for compile_multitarget.
```
my git HEAD is de5428ffc367 and compiled with llvm-1:11~++20200705063810

[how do I dispatch the functions that generated by halide (tutorial lesson 15)?](https://stackoverflow.com/questions/62758966/how-do-i-dispatch-the-functions-that-generated-by-halide-tutorial-lesson-15/62767686#62767686)

shoaibkamil:
It sounds like you might need `-e static_library,c_header` instead of `-e object,c_header`.

benzwt:
After the multiple target was built. I appended the halide_runtime_x86.o to my_first_generator.a 
with ar. Everything works fine.

I'm trying to build the multiple target  on Windows.
The question is how do I append the halide_runtime_x86.o ?
Not so familiar with Windows tools.

```
./lesson_15_generate \
    -g my_first_generator \
    -f my_first_generator \
    -e static_library,c_header\
    -o . \
    target=x86-64-avx-no_runtime,x86-64-sse41-no_runtime,x86-64-no_runtime
./lesson_15_generate \
    -r halide_runtime_x86 \
    -e object,c_header\
    -o . \
    target=host-x86-64

#> ar rcs my_first_generator.a halide_runtime_x86.o
```

PR #5091, opened 2020-07-07
Better error message when trying to compile multitarget to object

abadams:


abadams:
Fixes #5089

alexreinking:
@abadams -- the "Fixes #issue" is only magic when it's in a commit message or in the top-level comment of the PR.

PR #5092, opened 2020-07-07
cmake: python: use system pybind11 if installed

acolinisi:
Fixes builds on offline hosts (common on HPC clusters), which otherwise
fail with failure to fetch pybind11 from the Internet, introduced in
c13efffcc9e282d14c21a5149e4bdee078c85a9c:

	CMake Error at
	pybind11-subbuild/pybind11-populate-prefix/tmp/pybind11-populate-gitclone.cmake:31
	(message):
	  Failed to clone repository: 'https://github.com/pybind/pybind11.git'

Also, make build friendly to distribution packages.  Distribution
package recipe should be in control of (1) dependencies to build against
(e.g. some distros are on pybind11 2.5, and it should be used unless
it's incompatible), and (2) when fetching happens vs when building
happens. So, there is a need for a packager-friendly build mode where
the build system doesn't go out and fetch whatever dependencies it
chooses whenever it chooses. This patch gives priority to the system
pybind11 with fallback to fetching.

An alternative (if you prefer) would be to keep fetch as the default, and add a flag for distributions to use: -DWITH_SYSTEM_PYBIND11 (default OFF).

### DIFF
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -7,16 +7,20 @@ find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 # Adding PyBind11 to the build is the recommended way of integrating with CMake.
 # See: https://pybind11.readthedocs.io/en/stable/compiling.html
 
-# Keep the version in sync with requirements.txt and the Ubuntu 20.04 LTS package (python3-pybind11)
-include(FetchContent)
-FetchContent_Declare(pybind11 GIT_REPOSITORY https://github.com/pybind/pybind11.git GIT_TAG v2.4.3)
-
 # Configure pybind11 to use the same interpreter version as was detected above.
 message(STATUS "Directing pybind11 to Python3 executable ${Python3_EXECUTABLE}")
 set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
 
-# Import PyBind11
-FetchContent_MakeAvailable(pybind11)
+# Keep the version in sync with requirements.txt and the Ubuntu 20.04 LTS package (python3-pybind11)
+set(PYBIND11_VER 2.4.3)
+find_package(pybind11 ${PYBIND11_VER} QUIET)
+if (NOT pybind11_FOUND)
+    include(FetchContent)
+    FetchContent_Declare(pybind11
+                         GIT_REPOSITORY https://github.com/pybind/pybind11.git
+                         GIT_TAG v${PYBIND11_VER})
+    FetchContent_MakeAvailable(pybind11)
+endif ()
 
 ##
 # Add our sources to this sub-tree.


steven-johnson:
LGTM but deferring to @alexreinking; also cloning for testing

python_bindings/CMakeLists.txt
```diff
@@ -7,16 +7,21 @@ find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
...
 message(STATUS "Directing pybind11 to Python3 executable ${Python3_EXECUTABLE}")
 set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
 
-# Import PyBind11
-FetchContent_MakeAvailable(pybind11)
+# Keep the version in sync with requirements.txt and the Ubuntu 20.04 LTS package (python3-pybind11)
+set(PYBIND11_VER 2.4.3)
+find_package(pybind11 ${PYBIND11_VER} QUIET) # not REQUIRED because fallback to fetch
+if (pybind11_INCLUDE_DIR)
+    set(PYBIND11_INCLUDE_DIR ${pybind11_INCLUDE_DIR}) # homogenize with else clause
```
alexreinking, line 18: Where is this variable used in the project? I don't see why setting this variable is needed. If it indeed isn't, then this should be more like

```
find_package(...)
if (NOT pybind11_FOUND)
  ...
endif ()
```

python_bindings/CMakeLists.txt
```diff
@@ -7,16 +7,21 @@ find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
...
+else() # installation not found, try fetching from online
+    include(FetchContent)
+    FetchContent_Declare(pybind11
+        GIT_REPOSITORY https://github.com/pybind/pybind11.git GIT_TAG v${PYBIND11_VER})
```
alexreinking, line 22: If you're going to chop the line, please follow the formatting used elsewhere in the project:

```
FetchContent_Declare(pybind11
                     GIT_REPOSITORY https://github.com/pybind/pybind11.git
                     GIT_TAG v${PYBIND11_VER})
```

python_bindings/CMakeLists.txt
```diff
@@ -7,16 +7,21 @@ find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
...
-
 # Configure pybind11 to use the same interpreter version as was detected above.
 message(STATUS "Directing pybind11 to Python3 executable ${Python3_EXECUTABLE}")
 set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
 
-# Import PyBind11
-FetchContent_MakeAvailable(pybind11)
+# Keep the version in sync with requirements.txt and the Ubuntu 20.04 LTS package (python3-pybind11)
+set(PYBIND11_VER 2.4.3)
+find_package(pybind11 ${PYBIND11_VER} QUIET) # not REQUIRED because fallback to fetch
```
alexreinking, line 16: No need for this comment. The next few lines are obvious.

python_bindings/CMakeLists.txt
acolinisi: Variable was removed in d7034c8df4a0f1662b339d1a6ab9714f2f63541d but I was looking at diff c13efffcc9e282d14c21a5149e4bdee078c85a9c . Fixed.

python_bindings/CMakeLists.txt
acolinisi: Fixed.

python_bindings/CMakeLists.txt
acolinisi: Fixed.

acolinisi:
Btw, d7034c8df4a0f1662b339d1a6ab9714f2f63541d undoes pybind11 2.4.3 -> 2.5.0 version change done in f9d8b7de48be2f232585aafdf898e94f4cdb20b3 . Was that intentional?

alexreinking:
> Btw, [d7034c8](https://github.com/halide/Halide/commit/d7034c8df4a0f1662b339d1a6ab9714f2f63541d) undoes pybind11 2.4.3 -> 2.5.0 version change done in [f9d8b7d](https://github.com/halide/Halide/commit/f9d8b7de48be2f232585aafdf898e94f4cdb20b3) . Was that intentional?

Yes, we want to match the version in Ubuntu 20.04 LTS because that's what the Makefile-based build will use.

alexreinking:
@steven-johnson - if you would, please clone again for testing.

steven-johnson:
Clone looks good to land

PR #5097, opened 2020-07-09
WebAssembly Updates

steven-johnson:
Revamp our WebAssembly support to bring it more up-to-date with current spec:
- Remove the use of V8 for our "jit" support and use the WABT interpreter instead. (It's vastly slower but is correct and is far easier to integrate into Halide.)
- Remove all explicit support for wasm building/testing from the Makefile; move it into CMake instead.
- Update all tests to reflect current status of build/test accuracy.
- Especially update simd_op_check to verify final-spec simd ops.
- Update README.


alexreinking:
re: the GIT_SUBMODULES CMake bug. It was fixed:

https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729

steven-johnson:
> re: the GIT_SUBMODULES CMake bug. It was fixed:
> 
> https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729

...but presumably in 3.18+, so we can't rely on it yet?

alexreinking:
> ...but presumably in 3.18+, so we can't rely on it yet?

Ah, I misunderstood your comment in the code. I thought that the bug was introduced in 3.18.

steven-johnson:
There's more wasm-related work to do in the future (notably, getting benchmarks/performance tests working), but I think this is a good enough stopping point that we should look at reviewing and landing it.

(Note that wasm testing isn't enabled on the buildbots yet -- it never has been! -- but should be possible now thanks to these changes. After this lands, I'll update the buildbots accordingly.)

README_webassembly.md
```diff
@@ -104,108 +90,45 @@ the Halide library itself.
...
-- Set `V8_INCLUDE_PATH` and `V8_LIB_PATH` to point to the paths for V8 include
-  files and library, respectively.
-
-- Set `WITH_V8=1`
-
-- To run the JIT tests, set `HL_JIT_TARGET=wasm-32-wasmrt` (or
-  `HL_JIT_TARGET=wasm-32-wasmrt-wasm_simd128`) and run normally. The test suites
-  which we have vetted to work include correctness, performance, error, and
-  warning. (Some of the others could likely be made to work with modest effort.)
+- To run the JIT tests, set `HL_JIT_TARGET=wasm-32-wasmrt` (possibly adding `wasm_simd128`, `wasm_signext`, and/or `wasm_sat_float_to_int`) and run CMake/CTest normally. Note that wasm testing is only support under CMake (not via Make).
```
abadams, line 101: line wrapping in this file seems necessary. I think @alexreinking used a markdown auto-formatter at some point that might do it?

src/CodeGen_WebAssembly.cpp
```diff
@@ -70,6 +65,18 @@ void CodeGen_WebAssembly::visit(const Cast *op) {
...
         {Target::WasmSimd128, true, Int(16, 8), 0, "llvm.wasm.sub.saturate.signed.v8i16", i16_sat(wild_i32x_ - wild_i32x_)},
         {Target::WasmSimd128, true, UInt(16, 8), 0, "llvm.wasm.sub.saturate.unsigned.v8i16", u16_sat(wild_i32x_ - wild_i32x_)},
+
+        {Target::WasmSimd128, true, UInt(8, 16), 0, "llvm.wasm.avgr.unsigned.v16i8", u8(((wild_u16x_ + wild_u16x_) + 1) / 2)},
+        {Target::WasmSimd128, true, UInt(8, 16), 0, "llvm.wasm.avgr.unsigned.v16i8", u8(((wild_u16x_ + wild_u16x_) + 1) >> 1)},
+        {Target::WasmSimd128, true, UInt(16, 8), 0, "llvm.wasm.avgr.unsigned.v8i16", u16(((wild_u32x_ + wild_u32x_) + 1) / 2)},
+        {Target::WasmSimd128, true, UInt(16, 8), 0, "llvm.wasm.avgr.unsigned.v8i16", u16(((wild_u32x_ + wild_u32x_) + 1) >> 1)},
+
+        // TODO: LLVM should support this directly, but doesn't yet.
+        // To make this work, we need to be able to call the intrinsics with two vecs.
```
abadams, line 75: The way I've had to do this in the past is with force-inlined implementations that accept the wider vec, e.g. see packsswbx16 in src/runtime/x86.ll

src/WasmExecutor.cpp
```diff
@@ -226,75 +247,131 @@ class BDMalloc {
...
-        Local<Object> obj = v.As<Object>();
-        Local<Array> properties = obj->GetPropertyNames(context).ToLocalChecked();
-        int len = properties->Length();
-        wdebug(0) << "Number of properties = " << len << ":\n";
-        for (int i = 0; i < len; ++i) {
-            const v8::Local<v8::Value> key = properties->Get(i);
-            String::Utf8Value str(isolate, key);
-            wdebug(0) << "\t" << i + 1 << ". " << *str << "\n";
-        }
+    // lld will temporarily hijack the signal handlers to ensure that temp files get cleaned up,
```
abadams, line 335: This sounded painful to debug.

src/WasmExecutor.cpp
```diff
@@ -340,208 +420,82 @@ auto dynamic_type_dispatch(const halide_type_t &type, Args &&... args) -> declty
...
-    internal_error << "TODO: 64-bit slots aren't yet supported";
+#undef HANDLE_CASE
 }
+// clang-format on
 
-template<>
-inline void ExtractAndStoreScalar<int64_t>::operator()(const Local<Context> &context, const Local<Value> &val, void *slot) {
-    internal_error << "TODO: 64-bit slots aren't yet supported";
+std::string to_string(const wabt::MemoryStream &m) {
+    // TODO: ugh
```
abadams, line 429: Please elaborate. Are you talking about the weird requirement to const cast?

tutorial/CMakeLists.txt
```diff
@@ -48,39 +48,43 @@ add_tutorial(lesson_08_scheduling_2.cpp WITH_IMAGE_IO WITH_OPENMP)
...
-    # Users of the AOT functions (as opposed to Generators) should link to the relevant
-    # GPU libraries manually.
-    _Halide_target_link_gpu_libs(lesson_10_aot_compilation_run PRIVATE ${HL_TARGET})
-
-    target_include_directories(lesson_10_aot_compilation_run PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
-
-    add_test(NAME tutorial_lesson_10_aot_compilation_run COMMAND lesson_10_aot_compilation_run)
-    set_tests_properties(tutorial_lesson_10_aot_compilation_run PROPERTIES LABELS tutorial)
+    if (TARGET_WEBASSEMBLY AND HL_TARGET MATCHES "wasm")
+        # TODO: Requires custom build rules to work under wasm.
```
abadams, line 52: Should we just error out here?

README_webassembly.md
alexreinking: Yes, it's called prettier.io. I like it quite a bit. I make sure to always pass `--prose-wrap always` at the CLI so that it reflows paragraphs nicely.

tutorial/CMakeLists.txt
alexreinking: No, because you might want to build the _other_ tutorials.

alexreinking: An author warning might be appropriate, however, via `message(AUTHOR_WARNING "Not all tutorials build under WASM")` or similar

src/CodeGen_WebAssembly.cpp
steven-johnson: One of the wasm/LLVM folks has been working on improving these codegen issues in their backend, so I haven't expended much energy on trying to improve it here yet. I'll add this as a note in case circumstances make us need to improve sooner.

src/WasmExecutor.cpp
steven-johnson: it wasn't the worst thing I've ever had to do, but it wasn't much fun, either

src/WasmExecutor.cpp
steven-johnson: Yeah, that comment doesn't belong there. Removing.

tutorial/CMakeLists.txt
steven-johnson: Warnings inserted.

steven-johnson:
Can I get final review/approval for this? (Only failure is unrelated Windows GPU failure)

alexreinking:
I'd like to check that the commands in the README still work even with an LLVM that doesn't have WABT, since the Ubuntu-bundled one doesn't.

steven-johnson:
> I'd like to check that the commands in the README still work even with an LLVM that doesn't have WebAssembly/WABT/whatever, since the Ubuntu-bundled one doesn't.

cool, will wait

PR #5110, opened 2020-07-13
Fix binary2cpp usage in CMake

steven-johnson:
Ensure that all usage has `binary2cpp` in its DEPENDS, and use `$<TARGET_FILE:binary2cpp>` to ensure we expand a full path properly. (In theory, recent CMake builds shouldn't need the explicit expansion, but some cross-compilation configs seem to be skipping the expansion.)

### DIFF
--- a/test/generator/CMakeLists.txt
+++ b/test/generator/CMakeLists.txt
@@ -171,23 +171,23 @@ add_custom_command(OUTPUT "${EC32}.bc"
                    VERBATIM
                    COMMAND clang -O3 -c -m32 -target le32-unknown-nacl-unknown -emit-llvm "$<SHELL_PATH:${EXTERNAL_CPP}>" -o "${EC32}.bc")
 add_custom_command(OUTPUT "${EC32}.cpp"
-                   DEPENDS "${EC32}.bc"
-                   COMMAND binary2cpp external_code_extern_bitcode_32 < "${EC32}.bc" > "${EC32}.cpp")
+                   DEPENDS "${EC32}.bc" binary2cpp
+                   COMMAND $<TARGET_FILE:binary2cpp> external_code_extern_bitcode_32 < "${EC32}.bc" > "${EC32}.cpp")
 
 set(EC64 "external_code_extern_bitcode_64")
 add_custom_command(OUTPUT "${EC64}.bc"
-                   DEPENDS "${EXTERNAL_CPP}"
+                   DEPENDS "${EXTERNAL_CPP}" binary2cpp
                    VERBATIM
                    COMMAND clang -O3 -c -m64 -target le64-unknown-unknown-unknown -emit-llvm "$<SHELL_PATH:${EXTERNAL_CPP}>" -o "${EC64}.bc")
 add_custom_command(OUTPUT "${EC64}.cpp"
-                   DEPENDS "${EC64}.bc"
-                   COMMAND binary2cpp external_code_extern_bitcode_64 < "${EC64}.bc" > "${EC64}.cpp")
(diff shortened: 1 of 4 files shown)


test/generator/CMakeLists.txt
```diff
@@ -171,23 +171,23 @@ add_custom_command(OUTPUT "${EC32}.bc"
...
 add_custom_command(OUTPUT "${EC32}.cpp"
-                   DEPENDS "${EC32}.bc"
-                   COMMAND binary2cpp external_code_extern_bitcode_32 < "${EC32}.bc" > "${EC32}.cpp")
+                   DEPENDS "${EC32}.bc" binary2cpp
+                   COMMAND $<TARGET_FILE:binary2cpp> external_code_extern_bitcode_32 < "${EC32}.bc" > "${EC32}.cpp")
 
 set(EC64 "external_code_extern_bitcode_64")
 add_custom_command(OUTPUT "${EC64}.bc"
-                   DEPENDS "${EXTERNAL_CPP}"
+                   DEPENDS "${EXTERNAL_CPP}" binary2cpp
```
alexreinking, line 179: This command doesn't refer to `binary2cpp`

alexreinking:
I'd like to see how to repro this to better understand why it might not be working / what we might be doing wrong before merging. There should really be no need to do this.

test/generator/CMakeLists.txt
steven-johnson: Fixed

steven-johnson:
> I'd like to see how to repro this to better understand why it might not be working / what we might be doing wrong before merging. There should really be no need to do this.

The only place I've seen it is when running some tests on GHA (while trying to bring that build script up to date with CMake changes). The VMs are running CMake 3.17.0... maybe there was a bug in that specific version?

alexreinking:
> The VMs are running CMake 3.17.0... maybe there was a bug in that specific version?

Possibly? I think we should investigate further before applying a workaround that regresses the clarity of the CMake rules. If/when we get to the bottom of it, there should be a comment explaining why it's necessary because it looks like a misuse of CMake at a glance.

AFAICT, the relevant segments of code in CMake (master branch) are here:

https://github.com/Kitware/CMake/blob/a1f78a481c8cbda1f0c8503c4d661c3c3ddf00a6/Source/cmCustomCommandGenerator.cxx#L132-L143

and

https://github.com/Kitware/CMake/blob/a1f78a481c8cbda1f0c8503c4d661c3c3ddf00a6/Source/cmCustomCommandGenerator.cxx#L211-L227

Maybe there's a bug in the toolchain file? You might need to set `CMAKE_CROSSCOMPILING_EMULATOR` to something (even if it's empty).

steven-johnson:
Nah, something else is weird -- it's not just crosscompiling, I got a dependency error for a native build where `autograd_grad.runtime.a` was trying to be produced before `autograd.generator`, and that's *definitely* messed up, but the deps are correct by inspection.

I'm going to leave this open for now, but for the moment I suspect this is something specific to the GHA VMs (if not the CMake version, then maybe something else); it's not high on my list of things to fix so I'll circle back to it in a bit.

alexreinking:
> Nah, something else is weird -- it's not just crosscompiling, I got a dependency error for a native build where `autograd_grad.runtime.a` was trying to be produced before `autograd.generator`, and that's _definitely_ messed up, but the deps are correct by inspection.

Woah - I have no theories for that one.

steven-johnson:
I'm just gonna close this for now -- the DEPENDS stuff has landed elsewhere, and the other stuff is easy to redo if/when we find it happening elsewhere.

PR #5113, opened 2020-07-14
Try using a different target for generic 32-bit bitcode in our runtime

steven-johnson:
Not sure why, but it appears that using `le32-unknown-nacl-unknown` has an undesired side effect of making the `__atomic` builtins nearly unusable (almost any usage produces warnings about 'large atomic ops'). Experimentally, using `i386-unknown-unknown-unknown` as the 32-bit format for our runtime seems to solve that issue, but may well cause other problems (eg on arm32). This is just a draft run to see what changing the bitcode format does on the buildbots.

attn @abadams

### DIFF
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -151,13 +151,10 @@ set(RUNTIME_CXX_FLAGS -O3
 foreach (i IN LISTS RUNTIME_CPP)
     foreach (j IN ITEMS 32 64)
         if (${j} EQUAL 32)
-            if (${i} MATCHES "windows_.*")
-                # win32 uses the stdcall calling convention, which is x86-specific
-                set(TARGET "i386-unknown-unknown-unknown")
-            else ()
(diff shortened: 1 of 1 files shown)


abadams:
I believe this is why we were using sync instead of atomic. The generic triple doesn't assume much at all about the target, so for example perhaps it doesn't assume that aligned 32-bit writes to memory are already atomic.

steven-johnson:
(Marking as "ready for review" to tickle the buildbots)

> I believe this is why we were using sync instead of atomic. The generic triple doesn't assume much at all about the target, so for example perhaps it doesn't assume that aligned 32-bit writes to memory are already atomic.

Hmm, could be. But then why doesn't `le64-unknown-unknown-unknown` make a similar assumption? (Is it fair to assume that no such 64-bit architecture exists?) And more interestingly, why wouldn't `nacl` make that assumption, since it was basically portable i386?

steven-johnson:
Yeah, this one ain't gonna work. Closing.

PR #5116, opened 2020-07-14
Add Target::WasmSatFloatToInt

steven-johnson:
Cherry-picking a change from #5097 so that reviewing that PR will be simpler; this just adds another feature option for wasm that tooling wasn't ready for before, but is now (saturating float-to-int conversion).


abadams:
Is this worth a feature flag, or is this just something we should turn on unconditionally? I.e. are there going to be targets we care about that don't support this?

steven-johnson:
It's not clear yet, unfortunately -- it's a feature that wasn't in the original wasm spec, but was added not long afterwards, so it's not mandatory for all wasm engines to implement. While it's likely that any real-world-use of halide for wasm will target an engine that includes this (as well as simd), it's hard to say for sure.

steven-johnson:
According to https://webassembly.org/roadmap, Safari doesn't support this yet, and is adding features more slowly than other browsers. Consensus of folks I asked was that making each feature separately opt-in would be wise.

PR #5117, opened 2020-07-15
Add vector tile methods to the Python bindings

Infinoid:
PR #4898 added methods to C++; this makes them usable in Python.

Added a test case.

PR #5118, opened 2020-07-15
Clone of https://github.com/halide/Halide/pull/5117 for testing

steven-johnson:


Infinoid:
https://buildbot.halide-lang.org/master/#/builders/8/builds/6 looks like something my patch may have caused.  Any way to get a backtrace?

steven-johnson:
> https://buildbot.halide-lang.org/master/#/builders/8/builds/6 looks like something my patch may have caused. Any way to get a backtrace?

We're running CTest with `--output-on-failure`, so it *should* be doing that, but... sigh. Running it in super-verbose mode might help?

steven-johnson:
> https://buildbot.halide-lang.org/master/#/builders/8/builds/6 looks like something my patch may have caused. Any way to get a backtrace?

Actually, I don't think so, it's showing up in another (unrelated) pending PR: https://buildbot.halide-lang.org/master/#/builders/8/builds/7

Infinoid:
Ok, I was thinking the addition of new methods could possibly have changed pybind11's method dispatch in a subtle way, I've had troubles with that in the past.  But if the crash happens elsewhere, it makes that a lot less likely.  I'll trust your judgement, thanks.

steven-johnson:
I rebooted the winbot in question and restarted your build. Let's see if it runs OK overnight.

steven-johnson:
It's happening everywhere, so LGTM

PR #5120, opened 2020-07-15
Allow LLVM v12

steven-johnson:


Infinoid:
While you're at it, feel like bumping the number in the error message to match the actual requirement?

https://github.com/halide/Halide/blob/srj-llvm12/Makefile#L2178 says LLVM 7.0, but the check is for 9.0+.

PR #5122, opened 2020-07-15
Upgrade clang-format to v10

steven-johnson:
Upgrade the clang-format checks to clang-format-10, and reformat code accordingly. (Note that there aren't actually any changes necessary for 9 -> 10, but there are some for 10 -> 11 or 10 -> 12.)

Also add a way tp specify the clang-format version for `make format`; it defaults to the version of Clang for the current LLVM, but since clang-format doesn't provide stable formatting across versions, this might be wrong.


steven-johnson:
hmm, the clang-format action claims to support clang-format-10 but fails when attempting to do so

PR #5124, opened 2020-07-15
Rename the CMake var HL_TARGET -> Halide_TARGET, to disambiguate from the env var of the same name

steven-johnson:
It's easy to forget that ${HL_TARGET} and $ENV{HL_TARGET} are very different things in CMake script. (Ask me how I know.) Let's rename the former to HALIDE_TARGET to make the difference more obvious.


alexreinking:
Wouldn't it be better to just not honor the env-var and use "HL_TARGET" as a cache variable only?

abadams:
Could you elaborate on what that means? I'm confused about how the Halide target is set for the aot builds in cmake.

For example, a workflow I don't understand is building an single app with and without the -profile flag and having them live side-by-side in some reasonable way.

steven-johnson:
> Wouldn't it be better to just not honor the env-var and use "HL_TARGET" as a cache variable only?

I thought about that, but casual inspection of the code might mistake one for the other.

steven-johnson:
> Could you elaborate on what that means? I'm confused about how the Halide target is set for the aot builds in cmake.

It's confusing right now (IMHO) and needs some attention. IIUC (@alexreinking please correct me if I'm wrong):

- When you *first* run CMake:
    - if `HL_TARGET` is specified on the command line (via `-DHL_TARGET=whatever`), it uses that
    - otherwise, if `HL_TARGET` is defined *as an env var*, it uses that
- For all subsequent builds, **changes to the HL_TARGET env var are ignored**.
- To change the value used for HL_TARGET in CMake builds, you have to either:
    - re-run CMake again, with a new value for `-DHL_TARGET=whatever2`
    - nuke CMakeCache.txt
    - nuke the entire build dir and rebuild

The fact that we can pay attention the HL_TARGET-the-env-var sometimes, but not always, is confusing and has led me to make mistakes.

In theory, we could make CMake use HL_TARGET-the-env-var all the time, but there apparently isn't an easy way to record that in the build system (as we do with Make), so you risk getting inconsistent builds.

The other options that come to mind:

- Have CMake ignore HL_TARGET-the-env-var entirely and only honor it on the CMake build line. This is less inconvenient than it sounds, but greatly breaks muscle memory for longtime Makefile users. Also, for this to work reliably, the handful of tests that read HL_TARGET-the-env-var directly (eg simd_op_check) could still be out of sync.
- Same as previous, but (somehow) check at each build to verify that HL_TARGET-the-env-var and HL_TARGET-the-cmake-var stay in sync with each other, and warn if not.
- Something else I haven't thought of.

abadams:
I'd be very happy to use the env var as little as possible. For the Makefile-based apps I get into much less trouble using parameterized targets than having targets depend on an env var. (i.e. I prefer `make test-host-cuda` to `HL_TARGET=host-cuda make test`). We currently do this in some ways but not others. E.g. for local laplacian I can say: `make bin/host/process bin/host-profile/process` to get both binaries, but the `make test` target depends on the env var.

Can one parameterize cmake targets without clobbering existing outputs (i.e. the parameters end up in the filename or path)?

alexreinking:
> It's confusing right now (IMHO) and needs some attention. IIUC (@alexreinking please correct me if I'm wrong):
> * When you _first_ run CMake:
>   * if `HL_TARGET` is specified on the command line (via `-DHL_TARGET=whatever`), it uses that
>   * otherwise, if `HL_TARGET` is defined _as an env var_, it uses that
> * For all subsequent builds, **changes to the HL_TARGET env var are ignored**.
>   * To change the value used for HL_TARGET in CMake builds, you have to either:
>     * re-run CMake again, with a new value for `-DHL_TARGET=whatever2`
>     * nuke CMakeCache.txt
>     * nuke the entire build dir and rebuild

This is correct. The terse explanation is that `HL_TARGET` is a _cache variable_ whose _initial_ value is either `$ENV{HL_TARGET}` or `host` (if the former is not set). The build depends on the cache. You can edit the cache by re-running CMake with `-DCACHE_VAR=VAL`, modifying it directly, or (indeed) nuking the build dir and starting from scratch.

steven-johnson:
> I'd be very happy to use the env var as little as possible. 

I'm already on record as wanting to reduce our use of env vars as much as possible :-) I'd love to be able to eradicate HL_TARGET entirely (HL_JIT_TARGET is a different story that I'm not sure about yet).

> For the Makefile-based apps I get into much less trouble using parameterized targets than having targets depend on an env var. (i.e. I prefer `make test-host-cuda` to `HL_TARGET=host-cuda make test`). 

Agreed, though I'd also settle for something that is always explicit (rather than potentially-implicit, like an env var), e.g. `make --config=cuda test`, if that was required.

> Can one parameterize cmake targets without clobbering existing outputs (i.e. the parameters end up in the filename or path)?

I can't think of an obvious way, unfortunately, but I will defer to @alexreinking here.

steven-johnson:
While this PR opened a (necessary) can of worms, are there thoughts on the merit of the change itself? (If we don't want to ever land it, I want to close it.)

alexreinking:
My view is that `HL_TARGET` should be exclusively a cache variable in CMake that never looks at the env-var. AFAIK, no code actually cares about HL_TARGET-the-env-var and if it does, it shouldn't.

steven-johnson:
> AFAIK, no code actually cares about HL_TARGET-the-env-var and if it does, it shouldn't.

There is lots of code that does, at least in tests. Look for `get_target_from_environment()`.

abadams:
Most of the code in Func.h/Pipeline.h that has default arguments get_target_from_environment() is really designed to have a default target of "host", which is what happens if the environment variable is unset.

In retrospect it would be better to just have it default to host, but that would be a breaking change for an unknown number of people who might be controlling a custom aot build with HL_TARGET.

steven-johnson:
> In retrospect it would be better to just have it default to host, but that would be a breaking change for an unknown number of people who might be controlling a custom aot build with HL_TARGET.

At some point we're going to want to make breaking changes in the API and this should absolutely be on the list: calls that have default arguments that are read from env vars are pretty much The Devil Incarnate, no matter how well-intentioned. :-)

steven-johnson:
Anyway, to get back on point, there are (sorta) two issues at hand here:
- This PR just wants to rename the CMake-cache-var to disambiguate from the env-var-of-the-same-name. The motivation for this is that (until recently) we mostly reserved the `HL_` prefix for env vars, which are used in Halide to a (regrettable) extent (especially in autoscheduler code). Taking a second look now, it looks like the CMake rewrite added new CMake-only vars using this pattern (e.g. in `HalideGeneratorHelpers.cmake`); I'd argue that we should probably rename these too, to avoid any suggestion that these are somehow conflated with the env vars.
- There is *also* current a connection between HL_TARGET-the-env-var and HL_TARGET-the-CMake-cache-var, which (as mentioned above) is arguably more confusing than useful and should be removed or at least reconsidered; this PR is *not* seeking to address that, and I had planned to address that in a subsequent PR. (If we don't accept this PR then I'll go ahead and open a separate PR to address that anyway.)

abadams:
Given you'd declared the second point out of scope for this PR, I agree entirely with the first point. It would be less confusing if the cmake cache vars looked different to the env vars, by using a different prefix.

alexreinking:
Okay, then I say we follow the usual CMake naming convention of `${PROJECT_NAME}_<VAR>`. So these variables should be named `Halide_TARGET` instead of `HALIDE_TARGET`.

steven-johnson:
> Okay, then I say we follow the usual CMake naming convention of `${PROJECT_NAME}_<VAR>`. So these variables should be named like `Halide_TARGET` instead of `HALIDE_TARGET`. We can reserve `HL_<VAR>` for environment variables.

SGTM, I will revise accordingly.

steven-johnson:
PTAL

alexreinking:
Should be `Halide`, not `HALIDE` when the variable is exposed to the user / is a cache variable.

steven-johnson:
PTAL

PR #5126, opened 2020-07-17
Update README to suggest cloning a release of LLVM, not a branch

shoaibkamil:
The release tags should be the canonical source for that version; branches can sometimes be broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ works well on OS X and Ubuntu.)
 If you want to build it yourself, first check it out from GitHub:
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -217,7 +217,7 @@ Follow these steps if you want to build LLVM yourself. First, download LLVM's
 sources (these instructions use the latest 10.0 release)
```
steven-johnson, line 217: Nit: these instructions no longer use the 'latest' 10.0 release, but rather, 10.0.0 specifically.

alexreinking: at time of writing :P

PR #5131, opened 2020-07-20
Improve HVX codegen error reporting

steven-johnson:
If you try to compile HVX standalone code with HL_TARGET=hexagon-32-noos, you will die because necessary glue functions are defined in hvx_64 or hvx_128 but not 'baseline' hvx. Add an assertion check with a helpful error meesage to avoid just segfaulting deep inside LLVM.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -50,6 +50,9 @@ CodeGen_Hexagon::CodeGen_Hexagon(Target t)
     user_assert(!target.features_all_of(
         {Halide::Target::HVX_128, Halide::Target::HVX_64}))
         << "Cannot set both HVX_64 and HVX_128 at the same time.\n";
+    user_assert(!target.features_any_of(
+        {Halide::Target::HVX_128, Halide::Target::HVX_64}))
+        << "Must specify either HVX_64 or HVX_128 (but not both).\n";
 }
 
 namespace {
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -1349,6 +1349,8 @@ llvm::Function *CodeGen_Hexagon::define_hvx_intrinsic(llvm::Function *intrin,
                         internal_error
                             << "unhandled broadcast_scalar_word in define_hvx_intrinsic";
                     }
+                    internal_assert(fn)
+                        << "Unable to find scalar broadcast function. Did you forget to specify either hvx_64 or hvx_128?";
                     args[i] = builder->CreateCall(fn, {args[i]});
                 } else if (args[i]->getType()->isIntegerTy()) {
                     args[i] =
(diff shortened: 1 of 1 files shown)


src/CodeGen_Hexagon.cpp
```diff
@@ -1349,6 +1349,8 @@ llvm::Function *CodeGen_Hexagon::define_hvx_intrinsic(llvm::Function *intrin,
                         internal_error
                             << "unhandled broadcast_scalar_word in define_hvx_intrinsic";
                     }
+                    internal_assert(fn)
+                        << "Unable to find scalar broadcast function. Did you forget to specify either hvx_64 or hvx_128?";
```
pranavb-ca, line 1353: This is certainly helpful. Should we instead put a more heavyweight check in the constructor? (https://github.com/halide/Halide/blob/master/src/CodeGen_Hexagon.cpp#L36)
```
user_assert(target.features_any_of({{Halide::Target::HVX_128, Halide::Target::HVX_64}), "Need to specify target feature hvx_64 or hvx_128");
```
after https://github.com/halide/Halide/blob/master/src/CodeGen_Hexagon.cpp#L52

steven-johnson: If it's generally true that one or the other is required for any codegen on HVX, then yes, absolutely.

steven-johnson:
PTAL

src/CodeGen_Hexagon.cpp
```diff
@@ -50,6 +50,9 @@ CodeGen_Hexagon::CodeGen_Hexagon(Target t)
     user_assert(!target.features_all_of(
         {Halide::Target::HVX_128, Halide::Target::HVX_64}))
         << "Cannot set both HVX_64 and HVX_128 at the same time.\n";
+    user_assert(target.features_any_of(
```
dsharletg, line 53: I think this assert should go inside `acquire_hvx_context`, so we only hit this assert when we actually ask for HVX (which doesn't necessarily always happen).

src/CodeGen_Hexagon.cpp
dsharletg: If we assert more generally, I think we should remove the "Did you forget..." here, because that will be unlikely to help.

steven-johnson: PTAL

src/CodeGen_Hexagon.cpp
steven-johnson: OK, but not sure I understand -- how/why would we create a Codegen_Hexagon if we weren't going to actually do HVX codegen?

dsharletg: You can run code on Hexagon without HVX. It's pretty theoretical, but someone might try to do it (e.g. a small glue pipeline between larger pipelines that doesn't need HVX).

I just really don't think asserts should be more strict than necessary.

pranavb-ca: Aah good point. Thanks for catching this @dsharletg

PR #5135, opened 2020-07-23
Flesh out CPack packaging for releases.

### LINKED ISSUE #3971: CMake/CPack 'distrib' target isn't correct for Windows

### LINKED ISSUE #3972: CMake/CPack 'distrib' target omits static libraries

### LINKED ISSUE #4254: Should include doxygen docs in every binary release

alexreinking:
Main changes:

1. Adds CPack scripts for Windows, Mac, and Linux to create release packages easily.
  a. Linux packages include Debug/Release x Static/Shared
  b. macOS packages include Release x Static/Shared
  c. Windows packages include Debug/Release x Shared
2. Starts semantic versioning for Halide at 10.0.0
3. Renames `WITH_EXCEPTIONS` macro to `HALIDE_WITH_EXCEPTIONS`
4. Renamed / introduced CMake variables:
  a. `LLVM_USE_SHARED_LLVM_LIBS` -> `Halide_SHARED_LLVM`. This was never an LLVM option, so we need to stay out of their namespace.
  b. `Halide_BUNDLE_LLVM` -> On all major platforms, unpacks LLVM's static libraries and includes their objects in the Halide static library.

When using `find_package`, users can now request _either_ the static or shared versions of Halide. This works in one of several ways (in order of preference):

1. A user may explicitly request one or the other as a component. Ie. `find_package(Halide REQUIRED static)`. If it cannot load the requested libraries, it dies.
2. A user may set the `Halide_SHARED_LIBS` variable. When true-y, it loads the shared libs, else the static ones. If it cannot load the requested libraries, it dies.
3. When no library-type component is specified and `Halide_SHARED_LIBS` is not defined, it imitates the behavior of FetchContent by examining the `BUILD_SHARED_LIBS` variable. If it is true-y or not defined, it tries to load the shared libraries first. Otherwise it tries to load the static libraries first. Finally, it falls back to the other type.

As an advanced use-case, the user may request _both_ static _and_ shared by specifying both components. In this case, the targets `Halide::shared::{Halide,Generator,RunGenMain}` and `Halide::static::{Halide,Generator,RunGenMain}` are created, but `Halide::Halide` is unavailable. As a consequence, the convenience method `add_halide_library` is also not usable.

Fixes #3971 
Fixes #3972 
Fixes #4254


abadams:
We generally don't break user code without a release with a deprecation warning first, so I'm not sure '0.' is necessary. We'll just bump the major every time we delete some deprecated APIs or backends. It happens infrequently enough.

Makefile
```diff
@@ -199,7 +199,7 @@ RISCV_CXX_FLAGS=$(if $(WITH_RISCV), -DWITH_RISCV, )
 RISCV_LLVM_CONFIG_LIB=$(if $(WITH_RISCV), riscv, )
 
 INTROSPECTION_CXX_FLAGS=$(if $(WITH_INTROSPECTION), -DWITH_INTROSPECTION, )
-EXCEPTIONS_CXX_FLAGS=$(if $(WITH_EXCEPTIONS), -DWITH_EXCEPTIONS -fexceptions, )
+EXCEPTIONS_CXX_FLAGS=$(if $(WITH_EXCEPTIONS), -DHALIDE_WITH_EXCEPTIONS -fexceptions, )
```
steven-johnson, line 202: Why is this renaming necessary?

alexreinking: Because the behavior of Error.cpp changes depending on whether or not Halide was built with exceptions, so our users might need to know whether it was enabled. The old name, `WITH_EXCEPTIONS` is almost guaranteed to conflict with another project. It also now matches `HALIDE_ENABLE_RTTI`.

abadams: I don't think this macro is ever exported though. It's only used in Error.cpp, and no headers. This is now inconsistent with all the other WITH flags in the makefile.

abadams: If we think clients might need to check it dynamically should we add an api like we did for llvm_version?

steven-johnson: The use cases for checking it would seem to need to be compile-time mostly, I would think (e.g. so that you can wrap try/catch or not).

abadams: So perhaps it should be something exported by the Halide package in cmake (not sure of correct terminology here) as the variable HALIDE_WITH_EXCEPTIONS? But this is independent of what the #define is called.

abadams:
Question: What is the expectation of build compatibility across versions? E.g. we're talking about renaming cmake config vars in another thread. Does this require bumping the major because it requires downstream users to adjust their build setup?

Makefile
alexreinking: Linking with varying exception-handling capabilities is generally risky, as is linking with varying RTTI info, which is why packages that allow users to disable them have to propagate that information at build time.

alexreinking:
> Question: What is the expectation of build compatibility across versions? E.g. we're talking about renaming cmake config vars in another thread. Does this require bumping the major because it requires downstream users to adjust their build setup?

That's a good question. Semantic versioning is fundamentally about link-time compatibility. The hard requirement is that you should always be able to load a newer library of the same major version than the one you linked to. I think we should extend that to our build API. However, major version 0 has no such requirements and we should feel free to make breaking changes in the CMake until a numbered release is issued.

abadams:
Suggestion for versioning from offline discussion: Just match the version number of the included LLVM, and try to release at the same cadence (6 months)

steven-johnson:
> Suggestion for versioning from offline discussion: Just match the version number of the included LLVM, and try to release at the same cadence (6 months)

SGTM.

alexreinking:
Question: do we have any need to support the following scenarios:

1. A single project uses both static Halide _and_ shared Halide
2. A single project uses multiple versions of Halide

abadams:
By "Halide" you mean the compiler, right? Not Halide-generated code?

Assuming you mean the compiler, I don't think we should expend any effort in supporting that.

alexreinking:
Yes, I mean the compiler / library, not the generator outputs.

abadams:
Yeah, these multiple version issues normally only come up for large projects with multiple subteams, and large projects right now typically use Halide as an AOT compiler, so multiple Halide versions wouldn't ever end up in the same process.

steven-johnson:
(Monday morning check: is this still pending more work to do, or is it ready for review?)

alexreinking:
> 
> 
> (Monday morning check: is this still pending more work to do, or is it ready for review?)

Still a couple TODOs:

1. Change version number from 0.1.0 -> 10.0.0 (easy)
2. Bundle static libs (hard... won't have time until mid-week, sorry(

alexreinking:
A mac mini just arrived today, so I'll be able to test on mac imminently.

alexreinking:
PR now includes version 10.0 bump and macOS packaging script (using zsh since I did everything on Catalina).

Still TODO: bundling static libs on Linux / macOS. This will likely be a hack.

alexreinking:
All the major things are here. It would be nice to get the `-e o` for multi target feature working and some integration tests written.

alexreinking:
Can I bump this? I think it's in a merge-able state.

steven-johnson:
> All the major things are here. It would be nice to get the `-e o` for multi target feature working and some integration tests written.

Is that on my plate to get done, and is it a blocker for landing this?

alexreinking:
You touched that code most recently, right? But it is not a blocker, no.

steven-johnson:
> You touched that code most recently, right? But it is not a blocker, no.

Yep. Working on it now.

steven-johnson:
https://github.com/halide/Halide/pull/5183 is ready to review if you want `-e o` before landing this. (If not, I'm OK with landing this as-is.)

alexreinking:
> 
> 
> #5183 is ready to review if you want `-e o` before landing this. (If not, I'm OK with landing this as-is.)

We can go ahead and land this now. The `-e o` changes support an orthogonal feature.

PR #5136, opened 2020-07-23
Fix Doxygen warnings in runtime

alexreinking:
Fixes warnings generated by bad Doxygen comments. Normalizes use of `/*!` vs `/**` and `//!` vs `///` to all be the non-`!` versions, which are predominant.

### DIFF
--- a/src/runtime/mini_hexagon_dma.h
+++ b/src/runtime/mini_hexagon_dma.h
@@ -145,16 +145,18 @@ typedef struct stDmaWrapper_DmaTransferSetup {
 } t_StDmaWrapper_DmaTransferSetup;
 
 /**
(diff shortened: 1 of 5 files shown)


src/runtime/mini_hexagon_dma.h
```diff
@@ -22,7 +22,7 @@ __inline static int align(int x, int a) {
```
abadams, line 22: It's possible the contents of this file were copy-pasted from elsewhere and should be kept unchanged to minimize diffs against the source. I'll tag a QC person: @pranavb-ca

pranavb-ca: Thanks @abadams. This change is fine. We'll fold it into our sources to keep this maintainable in-house.

PR #5137, opened 2020-07-24
Set HL_{JIT_}TARGET from CMake Halide_TARGET variable.

### LINKED ISSUE #5132: Building apps/ under CMake needs some documentation
Right now, building apps under CMake has nonobvious bits and needs some documentation and/or cleanup:

- APPS_INTERNAL lets you control whether the apps are built 'internally' or totally separately, but it doesn't appear to be documented anywhere. It should be (at least in the comments), along with description of why you'd want to enable (or disable) this option (ie what are the tradeoffs)

- There doesn't seem to be any predictable way to build and/or test just a single app, and it varies by APPS_INTERNAL:
   - with APPS_INTERNAL=off, building of apps is deferred until CTest is run (I think)
   - with APPS_INTERNAL=on, you can pre-emptively build an individual app, but the names don't always match what you'd expect (eg apps/bilateral_grid is built via the `bilateral_grid_process` target)

(Ideally, we'd have predictable ways to build each app based on the obvious name, regardless of APPS_INTERNAL)

alexreinking:
Also drive-by remove `APPS_INTERNAL` as discussed in #5132 

I'm doing that here because if I hadn't, I would have needed to change the test environment in _every_ app.

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -1,10 +1,4 @@
 ##
-# Option controlling whether apps are built internally or at test time, externally.
-##
-
-option(APPS_INTERNAL "Build apps as part of the internal build" OFF)
-
-##
 # Test apps from the perspective of a consuming project.
 ##
 
@@ -27,14 +21,6 @@ set_tests_properties(create_halide_distrib PROPERTIES
                      DEPENDS delete_halide_distrib)
 
 function(add_app_test NAME)
-    # Either add the apps to the build directly or add them as external tests
-    if (APPS_INTERNAL)
-        # Add a stubbed-out find module for Halide so that the apps can build as part of the parent build.
-        list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
-        add_subdirectory(${NAME})
-        return()
-    endif ()
-
     unset(cmakeToolchainOpts)
     if (NOT "${CMAKE_TOOLCHAIN_FILE}" STREQUAL "")
         list(APPEND cmakeToolchainOpts "-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}")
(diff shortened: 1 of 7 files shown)


abadams:
A common workflow is compiling a test once, and then running it with lots of different targets (e.g. turning on and off the -debug  -profile -disable_llvm_loop_opt flags, or switching from CPU to GPU). What does the cmake workflow look like now for something like:

1) Compile test/performance/boundary_conditions.cpp
2) Run with target host
3) Run with target host-cuda
4) Run with target host-cuda-debug

Can this be done from ctest without recompilation, or is the answer to just run the binary directly and use the env var?

steven-johnson:
LGTM but would like @abadams comments.

alexreinking:
> Can this be done from ctest without recompilation, or is the answer to just run the binary directly and use the env var?

The latter. Though we could drop the `HL_JIT_TARGET` env var from the CTest description and have it work that way.

I notice that simd_op_check{_hvx}, and auto_schedule_no_{parallel,reorder} use `get_target_from_environment` . They should probably use `get_jit_target_from_environment` instead.

abadams:
simd_op_check is a little weird. It's aot-compiling code and then parsing the assembly, so it works for any target, not just JIT-compatible targets. HL_TARGET=arm-64 ./bin/correctness_simd_op_check is well-formed on x86

abadams:
Running the binary is fine for single examples. One thing I worry about is this workflow though:

1) Make a change
2) Build all tests
3) Run all tests for jit target = host
4) Run all tests for jit target = host-cuda
5) git push; open PR

If you reinvoke cmake to change the cache variable Halide_TARGET in between steps 3 and 4 is it going to recompile things?

alexreinking:
> If you reinvoke cmake to change the cache variable Halide_TARGET in between steps 3 and 4 is it going to recompile things?

It will recompile the generator tests, which depend on it.

abadams:
Right, I think I experienced that. The jit tests have no build-time dependency on the cache variable. That model should be fine then.

steven-johnson:
@abadams are you good with this change, or do think it needs more work?

PR #5139, opened 2020-07-25
Don't codegen asserts in blocks when noasserts is on

### LINKED ISSUE #5138: NoAssert Target feature seems to have no effect
I passed a Target object with NoAssert turned on to compile_to_lowered_stmt on a pipeline of mine. Here's what I get:

module name=117_0_t, target=x86-64-linux-avx-avx2-avx512-avx512_skylake-f16c-fma-no_asserts-no_bounds_query-sse41
buffer b0 = {...}
assert(((uint64)reinterpret(((halide_buffer_t *))b0.buffer) != (uint64)0), halide_error_buffer_argument_is_null("b0"))
let b0.min.0 = _halide_buffer_get_min(((halide_buffer_t *))b0.buffer, 0)
let b0.extent.0 = _halide_buffer_get_extent(((halide_buffer_t *))b0.buffer, 0)
let b0.stride.0 = _halide_buffer_get_stride(((halide_buffer_t *))b0.buffer, 0)
assert((b0.stride.0 == 1), 0)
assert((b0.min.0 == 0), 0)
assert((b0.extent.0 == 1000), 0)

I didn't expect to see these assertion. Is there another way to get rid of them ? I have a pipeline with hundreds of inputs, and pages after pages of assertions that validate the shapes of the inputs.

abadams:
Fixes #5138

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -4160,6 +4160,10 @@ void CodeGen_LLVM::visit(const Store *op) {
 }
 
 void CodeGen_LLVM::codegen_asserts(const vector<const AssertStmt *> &asserts) {
+    if (target.has_feature(Target::NoAsserts)) {
+        return;
+    }
+
     if (asserts.size() < 4) {
         for (const auto *a : asserts) {
             codegen(Stmt(a));


steven-johnson:
For reasons I haven't yet figured out, this change causes one pipeline in google to generate bad code: with this change in place, it does an OOB read where it didn't before (verified with ASAN). Investigating. Might ask to revert this in the short-term.

steven-johnson:
...sigh: this might be a case of a latent bug we were just getting lucky on, as the failing pipeline was built with `no_asserts` (for reasons that aren't obvious to me) and it fails with an OOB assert right away.

PR #5140, opened 2020-07-27
Unordered blocks

abadams:
When you unroll a pure var, you get a block of statements that can be executed in any order, including interleaving the executions. This is important for optimizations in llvm like register promotion, but this information is not preserved down through lowering.

This PR adds support for an ordering flag on blocks, and a new loop type "UnorderedUnrolled" that turns into an unordered block. An unordered block is one without a sequence point between the first and rest child stmts.

Most of the changes are specifying the ordering flag on blocks. Deinterleave required a bunch of work because it assumed blocks are linked lists, which is no longer true.


src/LoopCarry.cpp
```diff
@@ -231,17 +231,17 @@ class LoopCarryOverLoop : public IRMutator {
                 stores.push_back(v[i]);
             } else {
                 if (!stores.empty()) {
-                    result.push_back(lift_carried_values_out_of_stmt(Block::make(stores)));
+                    result.push_back(lift_carried_values_out_of_stmt(Block::make(stores, op->ordering)));
```
vksnk, line 234: I wonder if block_to_vector needs to be updated. I think there might be the case when inner block has different ordering, but it will be peeled off to the vector, and then when all the statements are put back into the block, inner block ordering will be overridden with outer block ordering.

src/Deinterleave.cpp
```diff
@@ -578,173 +578,247 @@ class Interleaver : public IRMutator {
...
+            Type type;
+            Expr base;
+        };
+        std::vector<Candidate> candidates;
+        std::vector<Stmt> pending;
+        pending.emplace_back(op);
+        while (!pending.empty()) {
+            Stmt next = std::move(pending.back());
+            pending.pop_back();
+            if (const Block *b = next.as<Block>()) {
```
vksnk, line 595: Similarly to the previous comment about block_to_vector, it seems that block ordering may be lost here.

src/Deinterleave.cpp
```diff
@@ -578,173 +578,247 @@ class Interleaver : public IRMutator {
...
+                    // load of the correct stride on the RHS.  TODO:
+                    // Could we consider mutating the RHS so that we
+                    // can handle more complex Expr's than just loads?
+                    const Load *load = stores[i].as<Store>()->value.as<Load>();
+                    if (!load) {
+                        failed = true;
+                        break;
+                    }
+
+                    // TODO(psuriana): Predicated load is not currently handled.
```
steven-johnson, line 690: Is this really the right assignee for this TODO?

src/LoopCarry.cpp
abadams: ugh yes, preserving it is going to be annoying

src/Deinterleave.cpp
abadams: old code moved around. I'll at least delete the assignee

abadams:
TODO: The current lowering is incorrect in the following case:
```
#include "Halide.h"

using namespace Halide;

    Var x, y, c;

    Func f;
    f(x, y, c) = cast<uint8_t>(x + y + c);
    f(x, y, c) += cast<uint8_t>(1);

    Func g;
    g(x, y, c) = f(x - 1, y, c) + f(x + 1, y, c);

    Var xo, xi;

    g.compute_root()
        .split(x, xo, xi, 4)
        .unroll(xi);
    f.store_at(g, xo)
        .compute_at(g, xi)
        .unroll(x)
        .update()
        .unroll(x);

    g.realize(128, 128, 3);

    return 0;
}

```

store_at outside a loop and compute_at inside the loop requires the loop to be ordered, but the code above lowers it as unordered and has read-modify-write hazards.

vksnk:
Is it possible to add a test to verify that unordered blocks where not generated for the case above? Maybe, adding a custom pass which counts a total number of unordered blocks in the final IR? Also, opposite test which checks that for the cases when unordered blocks are expected they are actually generated?

abadams:
I'm reconsidering this whole thing - it's quite disruptive

abadams:
Closing for now.

PR #5141, opened 2020-07-27
Add matching Free for the Allocate node added in the loop_carry pass

vksnk:


dsharletg:
I thought the free here was supposed to be implicit at the end of the Allocate block, and the Free was just a hint to recover memory earlier? If this is the case, this new Free seems like it would be a no-op.

shoaibkamil:
I think the reason to add this is because IR.h says `It is an error for an allocate node not to contain a free node of the same buffer.`  I'm not sure if there's existing code that relies on this, but I have certainly written code in the past that relied on that invariant.  I think it's fine to codegen to a no-op for Stack memory.

vksnk:
I ran into this when tried to apply the loop_carry pass and then generate code with CodeGen_C, which assumes (or seems to assume) that for every Allocate node, there is a matching Free node. 

However, if that's the case, I'll fix the CodeGen_C instead.

abadams:
Hrm, I don't think I've been respecting that invariant. The compiler is littered with Allocate nodes with no free. I've been treating it like Dillon suggests.

vksnk:
@shoaibkamil thanks for the reference, I read that comment three times before asking in Gitter and somehow always missed that sentence :)

So what would be the right thing to do here? Try to add all missing Free's or remove the sentence about mandatory Free from the comment and fix CodeGen_C?

shoaibkamil:
Either changing the comment+CodeGen_C or changing it so there’s always a matching Free throughout is fine with me. As long as the comment and code are consistent with each other.

abadams:
I'd prefer changing the comment + CodeGen_C

vksnk:
Sounds good, I'll close this one and start a new PR.

PR #5142, opened 2020-07-28
Handle Allocate without matching Free in the CodeGen_C

vksnk:


steven-johnson:
Do we need a test?

vksnk:
> Do we need a test?

I actually "added" a test by removing a line from the existing test :) There was a Free node for the heap allocation, after removing it, it now should test both that a) Free is optional and b) if a Free node is not present, call to the free function for the heap allocations will still be generated.

PR #5143, opened 2020-07-28
Update comment description for the Allocate node

vksnk:


src/IR.h
```diff
@@ -345,8 +345,7 @@ struct Provide : public StmtNode<Provide> {
 
 /** Allocate a scratch area called with the given name, type, and
  * size. The buffer lives for at most the duration of the body
- * statement, within which it is freed. It is an error for an allocate
- * node not to contain a free node of the same buffer. Allocation only
+ * statement, within which it is freed. Allocation only
```
abadams, line 348: How about: within which it may or may not be freed explicitly with a Free node with a matching name

vksnk: Thanks, updated.

PR #5146, opened 2020-07-28
Allow avoiding use of std::atomic in HalideBuffer.h

steven-johnson:
Some target platforms (e.g. some arm32 variants) are unreasonably cranky about the use of `std::atomic` (or, more specifically, the underlying `__atomic` intrinsics); it's not entirely clear whether the fault lies in the libatomic support being missing or incomplete in these environments, vs if there are nonobvious compiler/linker settings that need adapting.

I'm not actually sure if this should actually land or not, but at least one person has a short-term need for it, so I'm putting it out there in case some other people encounter this issue.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -73,13 +73,40 @@ bool any_zero(const Container &c) {
     }
     return false;
 }
+
+#ifdef HALIDE_BUFFER_DO_NOT_USE_STD_ATOMIC
+// Some embedded platforms may not have fully compliant versions of std::atomic<>
+// available (or, more specifically, the underlying __atomic primitives), despite
+// claims of C++11 compliance. This is a workaround to allow those platforms
+// to use this class safely by defining HALIDE_BUFFER_DO_NOT_USE_STD_ATOMIC.
+struct AtomicRefCount {
+    HALIDE_ALWAYS_INLINE explicit AtomicRefCount(int c)
+        : count(c) {
+    }
+    HALIDE_ALWAYS_INLINE int operator++() {
+        return __sync_fetch_and_add(&count, 1);
+    }
+    HALIDE_ALWAYS_INLINE int operator--() {
+        return __sync_fetch_and_add(&count, -1);
+    }
+    HALIDE_ALWAYS_INLINE operator int() const {
+        return count;
+    }
+
+private:
+    int count{1};
+};
+#else
+using AtomicRefCount = std::atomic<int>;
+#endif
+
 }  // namespace Internal
 
 /** A struct acting as a header for allocations owned by the Buffer
  * class itself. */
 struct AllocationHeader {
     void (*deallocate_fn)(void *);
-    std::atomic<int> ref_count;
+    Internal::AtomicRefCount ref_count;
 
     // Note that ref_count always starts at 1
     AllocationHeader(void (*deallocate_fn)(void *))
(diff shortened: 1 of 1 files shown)


mrmarkwell:
+1

src/runtime/HalideBuffer.h
```diff
@@ -73,13 +73,40 @@ bool any_zero(const Container &c) {
     }
     return false;
 }
+
+#ifdef HALIDE_BUFFER_DO_NOT_USE_STD_ATOMIC
```
dsharletg, line 77: Should we just always do this? Either one code path or the other isn't going to get as much test coverage as if we just unconditionally used AtomicRefCount.

abadams:
I was actually thinking the same thing as Dillon. I'd rather have one path that works everywhere than a conditional fallback

zvookin:
The issue with atomics interfaces is that if the entirety of a given interface cannot be mapped to the hardware instruction set, the entire interface has to be mapped to a software fallback. the __sync interface is a smaller interface, thus can be more reliably mapped, but is also a less good interface and not the preferred way to do things. the std::atomic and __atomic* stuff in LLVM is the preferred way, but there is some hardware for which it cannot be fully supported. This is also made tricky because we go through the intermediate LLVM IR via the generic target and that target can force everything through a least common denominator re: instruction set support.

I think the right design is to introduce our own atomics API that is satisfied via LLVM IR on each platforms. For most platforms this would be a generic __atomic* based file. For some it would map to ISA specific intrinsics. This requires some care to make sure LLVM understands the semantics and that the routines are marked appropriately for TSAN.

We could also push for a way to tell LLVM that we really want atomics even if not every operation can be used because we are only using a subset of the operations/types on a given address.

dsharletg:
> I think the right design is to introduce our own atomics API that is satisfied via LLVM IR on each platforms. For most platforms this would be a generic __atomic* based file. For some it would map to ISA specific intrinsics. This requires some care to make sure LLVM understands the semantics and that the routines are marked appropriately for TSAN.

Note this isn't actually in Halide compiled/runtime code, it's just HalideBuffer.h, so I think the equivalent of what you are describing is simply #ifdefs around the code, which is exactly what this PR does. But I think it would be better to just go with the least common denominator to avoid branching test coverage.

steven-johnson:
FYI, the place we actually saw this most recently was on Hexagon, not arm32 (my initial comment above was wrong, though we have seen similar issues on arm-32-linux in the past). @pranavb-ca (and some other QC folk) are investigating to see if this is a legit problem in the Hexagon SDK, or perhaps just a configuration error on our part (ie, missing/wrong flags). 

So either way let's hold off on landing this: if there is a flaw in the Hexagon toolchain, let's just clean this up per suggestions (and avoid the ifdef); if there isn't, this change isn't necessary :-)

zvookin:
I don't think punting to the least common denominator is a good idea. I'd have to go do some digging to enumerate the specifics, but atomics are an area where stuff has to be right and all the work on getting it right has gone into the C++ atomics model. __sync is basically x86 model that is emulated elsewhere. It is a smaller API, but it does not offer certain things that are needed and there is some hardware where it is more costly. The C++ API also offers some ways to automatically detect if it is guaranteed to work so that might be the way to go. It is annoying that we'll need multiple test paths, but I don't think it is avoidable in general.

dsharletg:
If there's a good reason to avoid __sync, then I'm fine with leaving the #ifdef approach in for now.

steven-johnson:
This appears to be a tooling issue in the team that originally reported the issue; they are going to use this workaround locally while investigating, but at this point I don't think we need or want to land it, so I'm closing this issue.

PR #5148, opened 2020-07-29
Fix transposed variable names

abadams:
The ordering of two feature names in the autoscheduler cost model were inconsistent with the names of those variables in the featurization. This makes the cost model confusing.

Fortunately the variables are used in exactly the same way, so just switching the variable names in the cost model clears things up without having to retrain anything. I have verified that it's producing the same schedules before and after.

### DIFF
--- a/apps/autoscheduler/cost_model_generator.cpp
+++ b/apps/autoscheduler/cost_model_generator.cpp
@@ -262,8 +262,8 @@ public:
         Expr native_vector_size = schedule_features(n, idx++, w);
         Expr num_vectors = schedule_features(n, idx++, w);
         Expr num_scalars = schedule_features(n, idx++, w);
-        Expr vector_loads_per_vector = schedule_features(n, idx++, w);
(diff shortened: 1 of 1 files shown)


apps/autoscheduler/cost_model_generator.cpp
```diff
@@ -262,8 +262,8 @@ class CostModel : public Generator<CostModel<training>> {
         Expr native_vector_size = schedule_features(n, idx++, w);
         Expr num_vectors = schedule_features(n, idx++, w);
         Expr num_scalars = schedule_features(n, idx++, w);
-        Expr vector_loads_per_vector = schedule_features(n, idx++, w);
         Expr scalar_loads_per_vector = schedule_features(n, idx++, w);
+        Expr vector_loads_per_vector = schedule_features(n, idx++, w);
```
steven-johnson, line 266: LGTM but this is still intrinsically fragile. Would be nice to follow up by changing all these to use explicit enums defined in (say) Featurization.h.

abadams:
Failures unrelated.

PR #5149, opened 2020-07-29
Fix 'unused variable' warning in the code generated by CodeGen_C with 'no_asserts' feature

vksnk:
Tested with HL_TARGET=host-no_asserts make -C apps/c_backend/ test

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -2524,7 +2524,10 @@ void CodeGen_C::create_assertion(const string &id_cond, const Expr &message) {
     internal_assert(!message.defined() || message.type() == Int(32))
         << "Assertion result is not an int: " << message;
 
-    if (target.has_feature(Target::NoAsserts)) return;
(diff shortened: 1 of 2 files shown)


abadams:
I think the create_assertion that takes two strings is:

1) Going to have the same problem
2) Implies you've evaluated the message and put the result in a variable, even though that would be the wrong semantics
3) Is only used in one place, which could be changed to pass a Call Expr for the message

So it should be deleted.

vksnk:
> I think the create_assertion that takes two strings is:
> 
> 1. Going to have the same problem
> 2. Implies you've evaluated the message and put the result in a variable, even though that would be the wrong semantics
> 3. Is only used in one place, which could be changed to pass a Call Expr for the message
> 
> So it should be deleted.

Thanks for the suggestion! I missed it, because in this specific case assert was checking that the memory allocation is not nullptr, so there was no warning for unused memory, because the pointer was actually used later in the program.

I've removed create_assertion(string, string) and replaced it's only occurrence with Call, PTAL.

PR #5152, opened 2020-07-30
Add a way to show computed Func value bounds

steven-johnson:
We currently dump for debug >=3, but this is incredibly noisy; this allows you define HL_SHOW_BOUNDS_VALUES=1 to dump the computed value bounds for each Func, as a quick way to start debugging some OOB situations.


steven-johnson:
failure is unrelated, PTAL

abadams:
I'm pretty way of introducing more environment variables. Can't one just grep the HL_DEBUG_CODEGEN=2 output?

steven-johnson:
Yeah, fair enough.

PR #5157, opened 2020-08-04
Rename WITH_PTX to WITH_NVPTX so all WITH_* macros match LLVM component names

alexreinking:
I'm in the process of making CMake bundle LLVM as part of Halide and I'm finding that I need to contort my way around these defines because one of them -- WITH_PTX -- is not named after its corresponding LLVM component, NVPTX.

This one small change drops ~60 lines of CMake while touching ~10 lines elsewhere.


abadams:
Probably worth changing the other uses in the makefile for consistency too.

alexreinking:
> Probably worth changing the other uses in the makefile for consistency too.

Done

abadams:
LGTM but also would like an OK from @steven-johnson

alexreinking:
Failure is unrelated (trunk LLVM on HVX)

PR #5158, opened 2020-08-04
Fix incorrect estimates for nl_means autoscheduler

abadams:
We run it on the parrot image, which is 5 megapixels.


steven-johnson:
failures unrelated, merging

PR #5161, opened 2020-08-05
Add a Target to the args of infer_input_bounds()

steven-johnson:
Currently, if the code isn't already jitted (or is jitted with NoBoundsQuery enabled), infer_input_bounds() falls back to whatever is in HL_JIT_TARGET. This is suboptimal, because it hides the fact that infer_input_bounds() relies on the value of that env var. It can also cause unintentional incorrect re-jits; e.g. if simd_op_check is run for HL_TARGET=wasm-32-wasmrt (and HL_JIT_TARGET not set at all), the infer_input_bounds() call can end up re-jitting for native code instead of WebAssembly.

### DIFF
--- a/python_bindings/src/PyFunc.cpp
+++ b/python_bindings/src/PyFunc.cpp
@@ -271,22 +271,22 @@ void define_func(py::module &m) {
             .def("output_buffers", &Func::output_buffers)
 
             .def(
-                "infer_input_bounds", [](Func &f, int x_size, int y_size, int z_size, int w_size) -> void {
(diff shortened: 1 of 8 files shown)


abadams:
This is a breaking change, right? I guess the alternative is another overload plus a deprecation of the existing one?

steven-johnson:
Yep. We can add-and-deprecate if you like, but I suspect this API is used infrequently enough that it's easier to just break things.

steven-johnson:
failure is unrelated

PR #5162, opened 2020-08-05
Update WASM_SHELL_VERSION

steven-johnson:
8.6.325 is the minimum version that fixes some wasm-simd bugs that affect us.


steven-johnson:
failure is unrelated

PR #5165, opened 2020-08-08
Make sure to set_host_dirty in rungen

abadams:
Otherwise synthetic inputs like 'random' end up being 'zero'


abadams:
This is another instance of #4333

abadams:
Failure unrelated.

PR #5172, opened 2020-08-10
Clean up is-jit-compiled checks in Pipeline

steven-johnson:
Because WebAssembly is a special beast, the way it is 'jitted' is special, and the checks to avoid redundant jitting needed extra logic in compile_jit(). Unfortunately there was another place in Pipeline that also needed this special casing. This PR adds a `get_compiled_jit_target()` bottleneck to consolidate this.

(Also added the `Target::defined()` method as a quick way to determine "are any parts of this Target unknown" -- not sure if this is the best name for the method. `is_unknown()`? Something else?)

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -661,6 +661,10 @@ bool Target::supported() const {
     return !bad;
 }
 
+bool Target::defined() const {
+    return os != OSUnknown && arch != ArchUnknown && bits > 0;
+}
+
 void Target::set_feature(Feature f, bool value) {
     if (f == FeatureEnd) return;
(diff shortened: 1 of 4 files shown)


src/Target.cpp
```diff
@@ -661,6 +661,10 @@ bool Target::supported() const {
     return !bad;
 }
 
+bool Target::defined() const {
```
abadams, line 664: defined is very overloaded. Other ideas: fully_specified, has_unknown_fields, has_unknowns, all_fields_set

steven-johnson: I vote for `has_unknowns()` of those.

PR #5174, opened 2020-08-10
Add infer_input_bounds(vector<int>)

steven-johnson:
Add a variant of infer_input_bounds() that takes an explicit vector of int, rather than the up-to-4-int version that is a holdover from the buffer_t days; deprecate the old version; convert all existing code to use the new one.

~~There are two warts here, unfortunately: in theory we should be able to just wrap the ints in {braces}, but in practice that doesn't work for zero- or one-arg lists:~~
~~- zero-arg vectors are ambiguous wrt other overloads of the method, and the compiler will select a deprecated one~~
~~- one-arg vectors are broken in C++ with brace initialization, as a single-integer arg will invoke the "create a vector of this length" ctor rather than "create a vector of length one with this value" ctor.~~

~~These two warts leave me with a bad taste in my mouth; the 'new' approach of always a list of sizes, rather than a sorta-vararg approach, is clearly better, and yet, the ugly code in some places is ugly.~~

Added a temporary initializer-list overload to avoid `{}` and `{1}` from overloading in a way that required the annotations; PTAL

### DIFF
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -2993,7 +2993,6 @@ Realization Func::realize(const Target &target,
 void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
                               const Target &target,
                               const ParamMap &param_map) {
-    user_assert(0) << "do not call";
     vector<int32_t> sizes;
     if (x_size) sizes.push_back(x_size);
     if (y_size) sizes.push_back(y_size);
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -2993,13 +2993,20 @@ Realization Func::realize(const Target &target,
 void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
                               const Target &target,
                               const ParamMap &param_map) {
-    user_assert(defined()) << "Can't infer input bounds on an undefined Func.\n";
-    vector<Buffer<>> outputs(func.outputs());
-    vector<int> sizes;
+    user_assert(0) << "do not call";
+    vector<int32_t> sizes;
     if (x_size) sizes.push_back(x_size);
     if (y_size) sizes.push_back(y_size);
     if (z_size) sizes.push_back(z_size);
     if (w_size) sizes.push_back(w_size);
+    infer_input_bounds(sizes, target, param_map);
+}
+
+void Func::infer_input_bounds(const std::vector<int32_t> &sizes,
+                              const Target &target,
+                              const ParamMap &param_map) {
+    user_assert(defined()) << "Can't infer input bounds on an undefined Func.\n";
+    vector<Buffer<>> outputs(func.outputs());
     for (size_t i = 0; i < outputs.size(); i++) {
         Buffer<> im(func.output_types()[i], nullptr, sizes);
         outputs[i] = std::move(im);
--- a/src/Pipeline.h
+++ b/src/Pipeline.h
@@ -498,9 +498,22 @@ public:
      * of the appropriate size and binding them to the unbound
      * ImageParams. */
     // @{
+    void infer_input_bounds(const std::vector<int32_t> &sizes,
+                            const Target &target = get_jit_target_from_environment(),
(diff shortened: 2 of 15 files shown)


test/correctness/specialize.cpp
```diff
@@ -232,7 +232,7 @@ int main(int argc, char **argv) {
         f.set_custom_trace(&my_trace);
 
         // Check bounds inference is still cool with widths < 8
-        f.infer_input_bounds(5);
+        f.infer_input_bounds(std::vector<int>{5});
```
abadams, line 235: yuck. What goes wrong with {5}? Is it just the wrong type (which we could adapt for), or some deeper error?

abadams: Oh, you explain above. Hrm.

abadams: Looks like an overload that takes a single int would fix it, if we're willing to do that. https://stackoverflow.com/questions/46565165/single-element-vector-initialization-in-a-function-call

steven-johnson: Yeah, that would 'fix' it, but would also defeat the appeal of having a single well-defined calling convention for this. (How this wart ended up in place and apparently remains in place in C++17/C++20 is mind-boggling to me. Principle Of Most WTF apparently applies here.)

steven-johnson:
PTAL (note that I rebased to master)

src/Pipeline.h
```diff
@@ -498,9 +498,22 @@ class Pipeline {
...
      * ImageParams. */
     // @{
+    void infer_input_bounds(const std::vector<int32_t> &sizes,
+                            const Target &target = get_jit_target_from_environment(),
+                            const ParamMap &param_map = ParamMap::empty_map());
+    HALIDE_ATTRIBUTE_DEPRECATED("Call infer_input_bounds() with an explicit vector<int> instead")
     void infer_input_bounds(int x_size = 0, int y_size = 0, int z_size = 0, int w_size = 0,
                             const Target &target = get_jit_target_from_environment(),
                             const ParamMap &param_map = ParamMap::empty_map());
+    // TODO: this is a temporary wrapper used to disambiguate the cases where
```
abadams, line 508: Is this the full story? Or does this serve the single-int-arg case too and so we'll have to keep it?

src/Func.cpp
```diff
@@ -2993,13 +2993,20 @@ Realization Func::realize(const Target &target,
 void Func::infer_input_bounds(int x_size, int y_size, int z_size, int w_size,
                               const Target &target,
                               const ParamMap &param_map) {
-    user_assert(defined()) << "Can't infer input bounds on an undefined Func.\n";
-    vector<Buffer<>> outputs(func.outputs());
-    vector<int> sizes;
+    user_assert(0) << "do not call";
```
abadams, line 2996: temporary debugging code?

src/Pipeline.h
steven-johnson: No: I have verified that code in the new form works correctly both in this PR form, and also if we delete both the deprecated method and this helper.

src/Func.cpp
steven-johnson: gah, scalpel left in patient

src/Pipeline.h
abadams: Wait, what? I thought the {single_int} case was calling the wrong vector constructor?

steven-johnson: Yeah, that's what I thought too, but I apparently was wrong. 

The issue apparently is that `{1}` is a perfectly valid way to initialize an int (eg, `int foo{1};`); as a result, calling `infer_input_bounds({1});` bound to the deprecated method: with default arguments, there is effectively an overload of the form `infer_input_bounds(int x_size);`, and that's preferred since it doesn't have to construct anything. Adding the initializer_list version provides another layer of preference.

(I hope I have got all of this right... this part of C++ is just a nightmare of twisty little passages)

PR #5176, opened 2020-08-11
Remove unnecessary call to halide.hexagon.pack.vh in CG_HVX

steven-johnson:
In the degenerate case of shuffle_vector() calling vlut() to shuffle a vector that is wider than 256 elements, the code was incorrectly using halide.hexagon.pack.vh on a vector-of-bool; this used to be necessary, but hasn't been for a while, so clearly this code path wasn't being exercised.

Remove the halide.hexagon.pack.vh and added a test case to exercise that path.

Also, drive-by removal of #include "EliminateBoolVectors.h" from CGHVX since it is no longer used there.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -9,7 +9,6 @@
 #include "CSE.h"
 #include "CodeGen_Internal.h"
 #include "Debug.h"
-#include "EliminateBoolVectors.h"
 #include "HexagonOptimize.h"
 #include "IREquality.h"
 #include "IRMatch.h"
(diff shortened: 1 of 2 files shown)


dsharletg:
Hmm, I thought we actually did have a test for this: https://github.com/halide/Halide/blob/master/test/correctness/interleave_rgb.cpp

How is that different?

dsharletg:
I guess your test is a deinterleave, but that test is an interleave. And apparently, interleave uses vdelta, while deinterleave uses vlut.

I think it would be best if we could find an existing test to just widen the vectorization of to cover this case if possible...

steven-johnson:
> I think it would be best if we could find an existing test to just widen the vectorization of to cover this case if possible...

Agreed, but I didn't find any by inspection. Let me try inserting some sentinel code into CGHVX::vlut() and see if I can find some that way.

steven-johnson:
Inserted two test cases in simd_op_check_hvx instead of adding vlut. PTAL.

PR #5177, opened 2020-08-11
Schedule last stage of stencil chain on GPU too

abadams:
It was accidentally left on the CPU, which means that benchmarking included a device copy. Oops.


steven-johnson:
ready to merge

PR #5181, opened 2020-08-12
Don't allow Target strings without complete arch-bits-os

### LINKED ISSUE #5009: cmake should halt if HL_TARGET does not contain platform/host
For context:
- Windows
- MSVC build
- `HL_TARGET=d3d12compute` (note: no `host`, or platform/os)

The build process was going its merry way, and all looked fine, until it reached this point:
```
LINK : fatal error LNK1181: cannot open input file 'cost_model.a' [apps\autoscheduler\Halide_Adams2019.vcxproj]
```
(note the '.a' suffix where it should have been '.lib')

We should probably emit some warning during cmake setup if `HL_TARGET` does not have enough context.

steven-johnson:
We previously accepted 'incomplete' Target strings (filling in host attributes for arch-bits-os if unspecified); we thought this would be a convenience, but in practice, this is usually indicative of an error or typo. This changes to make the Target(string) ctor assert-fail if the resulting target has an unspecified arch-bits-os.

Fixes #5009

### DIFF
--- a/python_bindings/correctness/target.py
+++ b/python_bindings/correctness/target.py
@@ -1,21 +1,27 @@
 import halide as hl
 
+
 def test_target():
     # Target("") should be exactly like get_host_target().
     t1 = hl.get_host_target()
     t2 = hl.Target("")
     assert t1 == t2, "Default ctor failure"
-    assert t1.supported();
+    assert t1.supported()
 
     # to_string roundtripping
     t1 = hl.Target()
     ts = t1.to_string()
     assert ts == "arch_unknown-0-os_unknown"
-    assert hl.Target.validate_target_string(ts)
-    t2 = hl.Target(ts)
 
-    # equality
-    assert t2 == t1
+    # Note, this should *not* validate, since validate_target_string
+    # now returns false if any of arch-bits-os are undefined
+    assert not hl.Target.validate_target_string(ts)
+
+    # Don't attempt to roundtrip this: trying to create
+    # a Target with unknown portions will now assert-fail.
+    #
+    # t2 = hl.Target(ts)
+    # assert t2 == t1
 
     # repr() and str()
     assert str(t1) == "arch_unknown-0-os_unknown"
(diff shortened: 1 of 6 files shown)


abadams:
I use this all the time to effectively say host minus something (e.g. x86-64, x86-64-sse41)

Has it been error-prone for you?

alexreinking:
> Has it been error-prone for you?

Yes. It's impossible to know whether you're going to get a `.a` or a `.lib` from that string alone, at least not without duplicating Target.cpp logic in your build system. See the discussion in #5009 and my comment here:

https://github.com/halide/Halide/issues/5009#issuecomment-640765085

Basically, CMake predicts the target name by looking for "host" in the target string. If it finds that, it does the obvious thing. Otherwise it looks for windows and if that's there it picks ".lib". Otherwise, it picks ".a".

This is not a CMake-specific issue. Any build system wanting to predict what the generator will return needs to understand _something_ about the target strings. The goal here is to make that understanding minimal.

abadams:
Slightly annoying, but fair enough. The build system shouldn't have to deal with implicit host.

We seem to be trying to remove all requirements to normalize target strings in the build. I guess the alternative is providing the build system with a way to normalize target strings, but that would be incredibly awkward in most build systems.

steven-johnson:
> We seem to be trying to remove all requirements to normalize target strings in the build. I guess the alternative is providing the build system with a way to normalize target strings, but that would be incredibly awkward in most build systems.

I avoided that even in Blaze/Bazel.

alexreinking:
> We seem to be trying to remove all requirements to normalize target strings in the build. 

That is certainly my goal. Processing target strings is tricky.

> I guess the alternative is providing the build system with a way to normalize target strings, but that would be incredibly awkward in most build systems.

Yeah, you would need to compute the target string(s), pass it (them) to an executable and get the result(s) back. That result would then need to be used to create rules in the build system. Not good.

steven-johnson:
AWESOME, the buildbots themselves break this rule

steven-johnson:
Thought: we could also add `hostbase` (name TBD) that would just infer arch-bits-os and nothing else; that might make some of Andrew's use cases simpler (eg `hostbase-sse2` for SSE2-only even on avx512 machines)

steven-johnson:
Waiting on LGTM from @abadams

PR #5183, opened 2020-08-13
Allow compile_to_multitarget() to emit object files

### LINKED ISSUE #5169: Make generators support object output in multi-config
End users might want to install the static libraries they create using Halide via CMake. Unfortunately, the current HalideGeneratorHelpers has to create IMPORTED STATIC libraries instead of plain old STATIC ones. IMPORTED targets cannot be installed. There are workarounds for this in CMake, but they are quite painful.

If we were to output _objects_ instead, then CMake could call the archiver on them, feel as though it had done something useful, and promote the generated library to plain old STATIC. However, this requires that we know ahead of time which files the generator is going to produce. The CMake helpers also require support for multi-target builds.

We should extend the `-e o[bject]` output mode of the generator in the following way.

1. Require that all targets in a multi-target list specify `no_runtime`. This is already how CMake does it.
2. Output `.o` files in the current directory whose names match the targets _as they appear on the command line_. ie. preserve "host".
3. The naming change should be reflected in single-target, no_runtime builds, too.
4. While we're at it, fix #5009.

steven-johnson:
I think this fits the criteria you need. PTAL.

Fixes #5169


src/Module.cpp
```diff
@@ -752,15 +752,18 @@ class ScopedCompilerLogger {
...
     validate_outputs(output_files);
 
     user_assert(!fn_name.empty()) << "Function name must be specified.\n";
     user_assert(!targets.empty()) << "Must specify at least one target.\n";
+    user_assert(suffixes.empty() || suffixes.size() == targets.size())
+        << "The suffixes list must be empty or the same length as the targets list.\n";
 
     // The final target in the list is considered "baseline", and is used
-    // for (e.g.) the runtime and shared code. It is often just os-arch-bits
+    // for (e.g.) the runtime and shared code. It is often just os-bits-arch
```
abadams, line 766: linux-64-x86? I thought we usually did linux-x86-64

steven-johnson: Canonical form generated by `Target::to_string()` is `arch-bits-os` so both of the above are wrong, will fix

steven-johnson:
Review ping

abadams:
Looks reasonable to me, but Alex's review is more important.

src/Module.cpp
```diff
@@ -830,7 +833,7 @@ void compile_multitarget(const std::string &fn_name,
         }
 
         // Each sub-target has a function name that is the 'real' name plus a suffix
-        std::string suffix = "_" + replace_all(target.to_string(), "-", "_");
+        std::string suffix = "_" + replace_all(target_label, "-", "_");
```
alexreinking, line 836: Why do we change "-" to "_" here? What's wrong with just doing "-" everywhere?

alexreinking:
I suppose it's impossible at this point to make the names consistent between n=1 and n>1, but other than that, my only question is about the extra work of changing "-" to "_". I'm not sure why that's necessary.

alexreinking:
I think I want to get #5186 working before we land this to make sure I don't need any further changes here.

src/Module.cpp
steven-johnson: Reasons likely lost to the mists of time now. But it's been that way for a while and some downstream systems are forced to rely on it (eg Bazel, in which the build system must know exact names for all files produced), so if we changed it here we'd have to change some downstream users, potentially. Not likely a big deal but also not a big deal to keep it this way.

steven-johnson:
> I suppose it's impossible at this point to make the names consistent between n=1 and n>1

No, impossible at all -- it's just a policy decision. 

We could declare that `compile_to_static_library()` maintains its current behavior, but that `compile_to_object_files()` always appends the suffix regardless of n. How useful/necessary would this be to the value of this call?

steven-johnson:
OK, I have changed it so that `compile_to_multitarget()` for object output (but not static_library output) now uniformly adds the proper suffix. Also added missing Python wrappers for `compile_to_multitarget_object_files()` and beefed up the correctness test considerably. 

(I did *not* change the underscore-vs-hyphen code at this point.)

PTAL.

alexreinking:
> OK, I have changed it so that `compile_to_multitarget()` for object output (but not static_library output) now uniformly adds the proper suffix. 

Ok, I will update #5186

alexreinking:
Looks like the `n=1` case is breaking the Makefile?

steven-johnson:
> Looks like the `n=1` case is breaking the Makefile?

whoops, fix pushed

abadams:
Double checking: This doesn't change the filename resulting from compile_to_object, right?

steven-johnson:
> Double checking: This doesn't change the filename resulting from compile_to_object, right?

Should not change that.

steven-johnson:
Hm, this still is a little ticklish: it doesn't change `compile_to_object()`, but it *does* change the behavior of running a Generator with `-e object` and a **single** target. Currently, we always output it as `$FNAME.o`; with this change, it would output `$FNAME_$TARGET.o` -- this breaks (e.g.) apps/linear_algebra, which makes a bunch of .o files, `ar`s them together, and links them together. Options here would be:

- Keep the new behavior and change the Makefile to use `static_library` output instead of `object`
- Keep the new behavior and just fix the Makefile to append the target string to the .o files as necessary
- Tweak this PR so that `-e object` does the same thing that it does now, and add a new option (`-e object_with_target`) that has the new behavior of adding the target string to the end of .o files. This would work but feels weirdly special-case-y and like a wart that we'd have to keep around forever.

I'm inclined to do the first. Anyone have strong feelings otherwise?

abadams:
Yeah, that's the case I was concerned about. I guess I assumed it went through compile_to_object.

I feel strongly that we should preserve the existing behavior of -e o so as to not break peoples' builds, and so that you don't have to think about multitarget if it's not relevant to you.

Can't we just make -e o equivalent to compile_to_object for single-target builds?

steven-johnson:
> Can't we just make -e o equivalent to compile_to_object for single-target builds?

Sure. That's exactly what it was before the latest change I pushed (at Alex's request). Having a symmetry for the object-file outputs (so that they have a uniform naming convention regardless of whether you generate one or more-than one) is appealing. But it is a behavioral change. To be honest, I don't really have a dog in this fight (if I had my druthers I'd require us to always output `.a` files). @alexreinking, how much pain does the special-casing of single-object-file-outputs case in CMake?

(FYI: I updated the apps/linear_algebra Makefile to use static_library instead, which seems like a safe and good change regardless of what we decide here.)

abadams:
Amusingly, people might be using -e o instead of static libraries specifically because build systems treat generated libraries as second class

abadams:
I also like your option 3: Refusing to do multi-target with -e object, and instead requiring -e objects or -e object_bundle or something. Something that conveys that it's multiple things.

steven-johnson:
> Amusingly, people might be using -e o instead of static libraries specifically because build systems treat generated libraries as second class

If, by "amusingly", you mean "tragically", then yes. (Why `ar` has never grown the ability to combine libraries is mind-boggling)

alexreinking:
> @alexreinking, how much pain does the special-casing of single-object-file-outputs case in CMake?

Not very much. I'm currently working around it. However, I _do_ like the `object_bundle` output option idea.

steven-johnson:
Something about the `object_bundle` idea just rubs me the wrong way -- it's an exact synonym for `object` with one different corner case, and the idea that we are essentially adding a new corner case that we will need to support long-term makes my teeth hurt. (It's the option I like least of what I proposed, I really just listed it for completeness.) 

I'm going to leave this PR as-is for tonight and ponder overnight. I welcome any more thoughts people have here.

alexreinking:
> I'm going to leave this PR as-is for tonight and ponder overnight. I welcome any more thoughts people have here.

I'm perfectly fine special casing for n=1 since I have to special case for n>1 _anyway_ on account of the `wrapper` object that only gets generated in that case.

steven-johnson:
> I'm perfectly fine special casing for n=1 since I have to special case for n>1 _anyway_ on account of the `wrapper` object that only gets generated in that case.

Hrm, an excellent point. OK, I'll do some selective reverting to go back to the previous behavior.

steven-johnson:
OK, I have reverted the relevant bits of code, so single-target object files now produce the same (unadorned) filename that they always did. PTAL.

Re: the swapping of _ and - in the filenames, I can be persuaded to make that change if there is a consensus in favor of it, but I'm reluctant to do so otherwise, as it's arguably an API change. @abadams any thoughts?

abadams:
I think the replacement only applies to the suffixes in multi-target, right? I'm not really sure on why those were replaced. I don't have a strong opinion on it. Are there file-systems where hyphens are a no-no?

alexreinking:
> Are there file-systems where hyphens are a no-no?

None that I'm aware of.

steven-johnson:
The question isn't so much "why did we do this in the first place", and more "is there a compelling reason to change it now, and risk breaking downstream build systems that rely on the current naming"

abadams:
I don't consider the names of the object files inside the static library to be part of the interface - wouldn't you have to be doing something pretty weird for that to matter?

steven-johnson:
> I don't consider the names of the object files inside the static library to be part of the interface - wouldn't you have to be doing something pretty weird for that to matter?

Yes, but Hyrum's Law suggests that someone somewhere is basing their entire career on this decision :-)

(I'll try keeping the hyphens and see what breaks.)

steven-johnson:
OK, I think all the issues are addressed. PTAL.

alexreinking:
> 
> 
> OK, I think all the issues are addressed. PTAL.

Merging into #5186 now. Need to figure out how I broke WASM then this can land.

steven-johnson:
@alexreinking you'll need to mark this as 'changed are addressed'

alexreinking:
> @alexreinking you'll need to mark this as 'changed are addressed'

I couldn't find that button, so I just hit 'approve' instead 🙃

PR #5184, opened 2020-08-13
Fix a source of GPU barrier deadlocks

abadams:
Partition loops shouldn't mess with serial loops containing
thread barriers, potentially causing warp divergence and deadlock (seen
in some obscure lens blur schedules).

Also we were generating too many thread barriers in a branch where the
base mutator class was accidentally always mutating something, so
there's a change to FuseGPUThreadLoops to make it more bug-resistant.

Without these additional barriers I have been unable to come up with a
case where a barrier ends up somewhere that would deadlock, so no test.

### DIFF
--- a/src/FuseGPUThreadLoops.cpp
+++ b/src/FuseGPUThreadLoops.cpp
@@ -1050,7 +1050,7 @@ public:
 };
 
 class InjectThreadBarriers : public IRMutator {
-    bool in_threads;
+    bool in_threads, injected_barrier;
 
     using IRMutator::visit;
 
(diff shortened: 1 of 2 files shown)


src/FuseGPUThreadLoops.cpp
```diff
@@ -1151,7 +1153,6 @@ class InjectThreadBarriers : public IRMutator {
         }
 
         return IRMutator::visit(op);
-        return op;
```
steven-johnson, line 1154: How is it possible that none of our compilers complained about this

shoaibkamil: Oh jeez.

PR #5187, opened 2020-08-16
Reschedule BGU to fix performance regression

abadams:
BGU on CUDA had regressed from its stated performance due to the atomic
floating point adds being compiled to CAS loops due to complex indexing
expressions diverging on the LHS and RHS of the +=. Inlining less stuff
into the += operations makes it succeed again, and the schedule was
improved with a few other tweaks.

Longer-term we need a first-class way to represent += so that we're not
sensitive to this sort of divergence.


abadams:
#5189

PR #5191, opened 2020-08-18
Remove PipelineContext

dsharletg:
Remove the PipelineContext worker thread in the Hexagon remote runtime. I think we added this because Halide allocated a lot of memory on the stack, and these threads got a tiny stack by default. However, since then, Halide allocates memory on the "stack", not the actual stack, so I think this is unlikely to be a problem any more.

At the same time, I suspect this code is a cause of deadlocks when trying to run code on Hexagon from multiple threads. It also is just ugly complexity that we should remove if we don't need.

@pranavb-ca @dpalermo please review and test as much as you can. I've tested this on as much code as I can, and I'm now pushing it for the build bots to test.


pranavb-ca:
Thanks for the PR. Will test and let you know.

pranavb-ca:
@shubhamp-ca is testing this.

dsharletg:
Any update on testing?

shubhamp-ca:
@dsharletg  There is a runtime crash observed on one testcase in our testsuite that deals with histograms. The crash is observed on SDM845 and SDM855 devices and not seen on SDM865. I suspect this is due to the pipeline running out stack in older devices but haven't confirmed this yet.

dsharletg:
Maybe there's an easy way to reduce the stack usage in that case?

pranavb-ca:
@dsharletg - When you say "stack" vs actual stack, are you referring to the pseudostack when you say "stack"?

shubhamp-ca:
@dsharletg You can go ahead with the merge for this change. Have created an internal issue for the observed runtime crash in older devices.

dsharletg:
Thanks for testing. Pranav, yes, the pseudostack reduced the amount of memory we place on the stack. There is also `can_allocation_fit_on_stack` that can be used to tweak the usage of the (real) stack, but currently it can't be configured per target. If this fixes the issue you are seeing for Hexagon, we could make it a virtual function of `CodeGen_Posix` so backends can override it.

This does need an approving review before it can be merged however.

pranavb-ca:
@dsharletg  - Did this fix your deadlock issue?

PR #5192, opened 2020-08-18
Remove mmap_dlopen from Hexagon runtime

dsharletg:
Remove support for mmap_dlopen. I'm pretty sure this *can't* work on any supported Hexagon target (requires mmap/mprotect, which doesn't exist any more).

I'd like to get rid of known_symbols.h/cpp too, but it seems necessary on the simulator, and it also has a sketchy comment for the device runtime.


dsharletg:
This passed the Hexagon tests on the buildbots before steven merged master.

Issue #5179, opened 2020-08-11
compute_with() and async() don't work well together


vksnk:
Following example will produce incorrect results:

``` C++
producer1(x, y) = x + y;
producer2(x, y) = 3 * x + 2 * y;
consumer(x, y) = producer1(x, y - 1) + producer1(x, y + 1) + producer2(x, y - 1) + producer2(x, y + 1);
consumer.compute_root();

producer1.compute_root().async();
producer2.compute_root().compute_with(producer1, y);

consumer.bound(x, 0, 16).bound(y, 0, 16);
```
From IR, it looks like the problem that it will have two separate allocations for producer2, one used in producer and one in consumer, so consumer will end up reading the data which was never written.

Similarly, it also will work incorrectly, if we use async() on both producers:
``` C++
producer1.compute_root().async();
producer2.compute_root().compute_with(producer1, y).async();;
```
This one is even more interesting, because it will end up with almost empty fork block for producer2:
``` C++
 fork {                                                                                                                                                                 
  allocate producer2[int32 * 16 * 18]                                                                                                                                   
  produce producer1 {          
   for (producer1.s0.fused.y, -1, 18) {
    for (producer1.s0.x, 0, 16) {
     producer1[((producer1.s0.fused.y*16) + producer1.s0.x) + 16] = producer1.s0.fused.y + producer1.s0.x
    } // for producer1.s0.x
    let t8 = producer1.s0.fused.y*16
    let t7 = producer1.s0.fused.y*2
    for (producer2.s0.x, 0, 16) {
     producer2[(producer2.s0.x + t8) + 16] = (producer2.s0.x*3) + t7
    } // for producer2.s0.x
   } // for producer1.s0.fused.y
   free producer2
   halide_semaphore_release(producer1.semaphore_0, 1)
  }
 } {                       
  allocate producer2[int32 * 16 * 18]                                               
  let producer2.semaphore_0 = (halide_semaphore_t *)alloca(16)
  halide_semaphore_init(producer2.semaphore_0, 0)      
  fork {                                                                            
   produce producer2 {                                                                                                                                                  
    halide_semaphore_release(producer2.semaphore_0, 1)                                                                                                                  
   }                      
  } {
```

The only schedule which works correctly is when async() applied only to producer2 (but I am not sure it will hold if fused group has more than 2 functions):
``` C++
producer1.compute_root();
producer2.compute_root().compute_with(producer1, y).async();;
```

PR #5195, opened 2020-08-19
Extend fork_async_producers to support functions scheduled with compute_with

vksnk:
As described in #5179 compute_with() and async() have issues when used together and this PR extends ForkAsyncProducers to support fused function groups to address these issues. 

The important limitation is that all functions from the same group have to be marked as async and will share fork block/task/thread. I think this is the only reasonable combination, because functions scheduled with compute_with share the loop(s) and launching them in different task would break this assumption.

(interesting side-effect of this is that now it's possible to schedule two completely independent functions to be executed in the same fork block with something like:

```
producer1.async();
producer2.compute_with(producer1, Var::outermost()).async();
```
).

I've added a number of tests which cover some of the possible combinations, but for something like async which depends on things like timing and such, it's quite difficult to be 100% convinced about correctness. For the existing functionality, I've tried to minimize changes to make sure that stuff that worked before still works. Also, I've eyeballed generated IR to make sure that forks and semaphores are generated in proper places and it looked alright.

### DIFF
--- a/src/AsyncProducers.cpp
+++ b/src/AsyncProducers.cpp
@@ -377,6 +377,8 @@ class ForkAsyncProducers : public IRMutator {
             }
         }
         internal_assert(fused_group_index < fused_groups.size());
+        // We want to make sure that for fused function group transformation is
+        // applied to inner Realize node.
         fused_group_func_counter[fused_group_index]++;
         const auto &current_group = fused_groups[fused_group_index];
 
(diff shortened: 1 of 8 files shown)


vksnk:
Monday review ping.

steven-johnson:
Is this just awaiting review?

steven-johnson:
Where does this PR stand?

vksnk:
Yes, waiting for review.

Issue #5201, opened 2020-08-20
Pipeline with async hangs


vksnk:
Following pipeline never completes:

``` C++
        Func producer1, producer2, consumer;
        Var x, y;

        producer1(x, y) = x + y;
        producer2(x, y) = producer1(x, y);
        consumer(x, y) = producer1(x, y - 1) + producer2(x, y + 1);

        consumer.compute_root();

        producer1.compute_at(consumer, y).async();
        producer2.store_root().compute_at(consumer, y).async();

        consumer.bound(x, 0, 16).bound(y, 0, 16);

        Buffer<int> out = consumer.realize(16, 16);
```

vksnk:
It works correctly if producer1 is also scheduled as store_root().

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** The reported schedule no longer hangs; it is now rejected at compile time with a message explaining exactly the constraint vksnk discovered empirically (that it works if `producer1` is also `store_root()`).

**Evidence:** Ran the repro against a current `libHalide` build of `f932856a0`: it now aborts with `Error: The Func f0 is consumed by async Func f1 and has a compute_at location in between the store_at location and the compute_at location of f1. This is only legal when f0 is both async and has a store_at location outside the store_at location of the consumer.` (`src/AsyncProducers.cpp:125`). Added by `674e6cc49` (#7868, "Disallow async nestings that violate read after write dependencies"), with `test/error/bad_async_producer_2.cpp` whose header comment cites this issue.

Closing: fixed by #7868 (commit `674e6cc49`). This schedule is now rejected up front by the check at `src/AsyncProducers.cpp:125` instead of hanging, and the repro is covered by `test/error/bad_async_producer_2.cpp`, which cites this issue.

PR #5202, opened 2020-08-20
Fix #5175: Add an extra check for input dependencies of the async producer

### LINKED ISSUE #5175: async() on only one of the functions in a chain of the functions scheduled with compute_at() will produce incorrect results
Following example, when only one of the producers is scheduled as async(), will produce incorrect results:
``` C++
    {
        Func producer1, producer2, consumer;
        Var x, y;

        producer1(x, y) = x + y;
        producer2(x, y) = producer1(x, y);
        consumer(x, y) = producer2(x, y - 1) + producer2(x, y + 1);
        
        consumer.compute_root();
        
        producer1.store_root().compute_at(consumer, y);//.async();
        producer2.store_root().compute_at(consumer, y).async();
        consumer.bound(x, 0, 16).bound(y, 0, 16);

        Buffer<int> out = consumer.realize(16, 16);

        out.for_each_element([&](int x, int y) {
            int correct = 2 * (x + y);
            if (out(x, y) != correct) {
                printf("out(%d, %d) = %d instead of %d\n",
                       x, y, out(x, y), correct);
                exit(-1);
            }
        });
    }
```

It seems that if producer2 is scheduled with async() then all other functions on the same loop level on which producer depends (producer1 in this case) should be scheduled as async() as well. If producer1 is also scheduled with async() then everything works correctly. 

In the IR, the producer2 is correctly moved to the separate fork block and both producer2 and consumer are protected by semaphores. However, producer1 is in the same block as consumer and is not guarded in any way, so when producer2 (which depends on producer1) is producing asynchronously it may run ahead of producer1 and produce incorrect results.
 
``` C++
 allocate producer1[int32 * 16 * 4]
 let producer2.folding_semaphore._0 = (halide_semaphore_t *)alloca(16)
 halide_semaphore_init(producer2.folding_semaphore._0, 2)                
 allocate producer2[int32 * 16 * 4]
 let producer2.semaphore_0 = (halide_semaphore_t *)alloca(16)
 halide_semaphore_init(producer2.semaphore_0, 0)
 fork {                                 
  for (consumer.s0.y, 0, 16) {
   acquire (producer2.folding_semaphore._0, 1) {
    let producer2.s0.y.min_1.s = select(0 < consumer.s0.y, 1, -1)
    produce producer2 {                                                             
     consume producer1 {          
      let t169 = select(0 < consumer.s0.y, 1, 3)                                                                                                                        
      let t168 = consumer.s0.y + producer2.s0.y.min_1.s
      for (producer2.s0.y, t168, t169) {
       let t170 = (producer2.s0.y % 4)*16
       for (producer2.s0.x, 0, 16) {                                                
        let t167 = producer2.s0.x + t170
        producer2[t167] = producer1[t167]
       } // for producer2.s0.x
      } // for producer2.s0.y
     }         
     halide_semaphore_release(producer2.semaphore_0, 1)
    }
   } // acquire
  } // for consumer.s0.y
 } {
  produce consumer {
   let t171 = (consumer.min.1*consumer.stride.1) + consumer.min.0
   for (consumer.s0.y, 0, 16) {
    let producer1.s0.y.min_1.s = select(0 < consumer.s0.y, 1, -1)
    produce producer1 {
     let t173 = select(0 < consumer.s0.y, 1, 3)
     let t172 = consumer.s0.y + producer1.s0.y.min_1.s
     for (producer1.s0.y, t172, t173) {
      let t174 = (producer1.s0.y % 4)*16
      for (producer1.s0.x, 0, 16) {
       producer1[producer1.s0.x + t174] = producer1.s0.x + producer1.s0.y
      } // for producer1.s0.x
     } // for producer1.s0.y
    }
    acquire (producer2.semaphore_0, 1) {
     consume producer2 {
      let t175 = ((consumer.s0.y + 3) % 4)*16
      let t176 = ((consumer.s0.y + 1) % 4)*16
      let t177 = (consumer.s0.y*consumer.stride.1) - t171
      for (consumer.s0.x, 0, 16) {
       consumer[consumer.s0.x + t177] = producer2[consumer.s0.x + t175] + producer2[consumer.s0.x + t176]
      } // for consumer.s0.x
     }
    } // acquire
    halide_semaphore_release(producer2.folding_semaphore._0, 1)
   } // for consumer.s0.y
  }
 } 
 free producer1
 free producer2
}
```

It seems that either it should be an error to have a schedule like this or async() should be propagated to all dependencies of the given producer in the pipeline.

vksnk:
As described in the #5175, it's possible to write an illegal schedule which successfully goes through the code generation, but generated code produces incorrect results.

This PR adds an extra check to detect such illegal schedules by verifying that if in realize node A there is a producer node B and producer A is async and is consumer of B then B must be also scheduled as async.

### DIFF
--- a/src/AsyncProducers.cpp
+++ b/src/AsyncProducers.cpp
@@ -649,9 +649,65 @@ class TightenForkNodes : public IRMutator {
     bool in_fork = false;
 };
 
+// If in realize node A there is a producer node B and producer A is async and is
+// consumer of B then B must be scheduled as async.
+class CheckAsyncOrder : public IRVisitor {
+    const map<string, Function> &env;
+
+    Scope<> in_realizes;
+    Scope<> in_consumers;
+    map<string, set<string>> realizes_around_producers;
+
+    using IRVisitor::visit;
+
+    void visit(const ProducerConsumer *op) override {
+        if (op->is_producer) {
+            for (auto r = in_realizes.cbegin(); r != in_realizes.cend(); ++r) {
+                realizes_around_producers[op->name].insert(r.name());
+            }
+            auto it = env.find(op->name);
+            internal_assert(it != env.end());
+            Function f = it->second;
+            if (f.schedule().async()) {
+                for (auto r = in_realizes.cbegin(); r != in_realizes.cend(); ++r) {
+                    if (in_consumers.contains(r.name()) && realizes_around_producers[r.name()].count(op->name)) {
+                        auto other_it = env.find(r.name());
+                        internal_assert(other_it != env.end());
+                        Function other_f = other_it->second;
+                        user_assert(other_f.schedule().async()) << "Invalid async: producer " << op->name
+                            << " is a consumer of " << r.name() << ", but " << r.name() << " is inside of the "
+                            << op->name << " Realize node, " << "which is scheduled as async(), so " << r.name()
+                            << " must be scheduled as async() too.";
+                    }
+                }
+            }
+        } else {
+            in_consumers.push(op->name);
+        }
+        IRVisitor::visit(op);
+        if (op->is_producer) {
+        } else {
+            in_consumers.pop(op->name);
+        }
+    }
+
+    void visit(const Realize *op) override {
+        ScopedBinding<> bind(in_realizes, op->name);
+        IRVisitor::visit(op);
+    }
+
+public:
+    CheckAsyncOrder(const map<string, Function> &e)
+        : env(e) {
+    }
+
+};
+
 // TODO: merge semaphores?
 
 Stmt fork_async_producers(Stmt s, const map<string, Function> &env) {
+    CheckAsyncOrder check(env);
+    s.accept(&check);
     s = TightenProducerConsumerNodes(env).mutate(s);
     s = ForkAsyncProducers(env).mutate(s);
     s = ExpandAcquireNodes().mutate(s);
--- a/test/correctness/CMakeLists.txt
+++ b/test/correctness/CMakeLists.txt
@@ -11,6 +11,7 @@ tests(GROUPS correctness
       async.cpp
       async_copy_chain.cpp
       async_device_copy.cpp
+      async_order.cpp
       atomic_tuples.cpp
       atomics.cpp
       autodiff.cpp
--- a/test/correctness/async_order.cpp
+++ b/test/correctness/async_order.cpp
@@ -1,0 +1,95 @@
+#include "Halide.h"
+#include <stdio.h>
+
+using namespace Halide;
+
+int main(int argc, char **argv) {
+    {
+        Func producer1, producer2, consumer;
+        Var x, y;
+
+        producer1(x, y) = x + y;
+        producer2(x, y) = producer1(x, y);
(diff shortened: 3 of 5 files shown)


steven-johnson:
LGTM but @zvookin should probably review

zvookin:
There's something I'm not getting here. Generally the things an async scheduled Func depends on do not need to be async themselves. I'm guessing this has to do with the storage being allocated outside the async invocations and the compute being inside. But then I'm not quite seeing how requiring the dependency to be async always makes it correct. I will sit down and look at it a little more closely, but I think updating the description to say something in terms of which schedules are valid and invalid, rather than describing the IR conditions they result in, might help.

vksnk:
Thank you for reviewing, that's a great question.

> Generally the things an async scheduled Func depends on do not need to be async themselves.

Right, but input dependency need to be computed fully beforehand. If input dependency is still being produced, async_producer needs to have a way to know that data it depends on is ready. 

If I understand correctly, async producers are lifted into separate "thread" based on their store_level, so if store_level is the same as compute_level then there is sort of implicit synchronization point between input dependency and producer:
```
for some_dim:
  compute input_dependency
  --> implicit sync point <--
  fork {
    compute async_producer  (depends on input_dependency)
    semaphores to sync between async_producer and consumer
  } {
    semaphores to sync between async_producer and consumer
    compute consumer of async_producer
  }
```
However, if store level is different from compute level then async producer will be lifted out of the loop and this implicit  synchronization point is gone:

```
fork {
for some_dim:
  compute async_producer  (depends on input_dependency, but can't really know how much of it was produced in another thread) 
  semaphores to sync between async_producer and consumer
} {
for some_dim:
  compute input_dependency
  
  semaphores to sync between async_producer and consumer
  compute consumer of async_producer
}
```
If this logic is correct then solution can be either making a copy of input_producer in both "threads" or inserting explicit synchronization between input_producer and async_producer. The latter looks exactly like what async() already does, so it seemed to me like a more reasonable way to address it. 

(Maybe, it's also possible to do forking differently, so it's not based on store level, but I am not sure and it sounds like a really big change).

vksnk:
Monday review ping.

steven-johnson:
Is this still just awating review?

steven-johnson:
Is this just waiting on review?

vksnk:
Yes, review is still needed. 

Also, I recently realized that #5201 might be just a different type of the same issue, so maybe it's possible/worth trying to generalize check from this PR to detect the case from #5201.

steven-johnson:
Where does this PR stand?

vksnk:
I can try to generalize it to cover the other case, but it would be great to get a feedback first in case there is a flaw in the logic somewhere.

steven-johnson:
This PR is pretty old. Is it still active?

vksnk:
This is still a problem which needs fixing, so yes. That being said, I discovered a test which still fails, so this needs to be figured out first. I am planning to return to async in couple weeks for some internal stuff, so hopefully can get this fixed too then.

abadams:
> This is still a problem which needs fixing, so yes. That being said, I discovered a test which still fails, so this needs to be figured out first. I am planning to return to async in couple weeks for some internal stuff, so hopefully can get this fixed too then.

What was the test that still fails? Is it part of this PR?

PR #5205, opened 2020-08-23
Add target flags for volta, turing, ansel. 

abadams:
Adding these target flags doesn't change much, as the underlying ptx->sass jit compiler in the driver generates code for the card in question already. It's more relevant if you're precompiling to sass, which in master is a commented-out block of code. It does cause a few minor codegen changes though here and there (CAS loops seem to have YIELD in them instead of NOP above capability 6.1).

Also tweaked the cuda schedules for harris, hist, and stencil_chain.


apps/hist/hist_generator.cpp
```diff
@@ -68,29 +69,44 @@ class Hist : public Halide::Generator<Hist> {
...
-                // but otherwise I get a shared memory
-                // blow-up. Something might be overconservative.
-                hist_rows.compute_root()
-                    .gpu_tile(x, y, xi, yi, 32, 8);
                 if (get_target().has_feature(Target::CUDA)) {
-                    hist_rows.update()
-                        .split(y, y, yi, 32)
-                        .split(rx, rxo, rxi, 16)
-                        .reorder(rxi, yi, rxo, y)
+                    // Each thread blow will use atomic integer adds
```
steven-johnson, line 77: s/blow/below/

src/CodeGen_PTX_Dev.cpp
```diff
@@ -742,7 +756,10 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
     }
     if (t.is_float() && t.bits() == 64) {
         // double atomics are supported since CC6.1
-        return target.has_feature(Target::CUDACapability61);
+        return (target.has_feature(Target::CUDACapability61) ||
```
steven-johnson, line 759: At this point, a helper function of `has_cuda_capability_at_least(num)` seems like it would pay for itself

steven-johnson: EDIT: or, alternately, use `get_cuda_capability_lower_bound() >= 61` here instead

src/runtime/HalideRuntime.h
```diff
@@ -1316,8 +1316,11 @@ typedef enum halide_target_feature_t {
     halide_target_feature_sve2,                   ///< Enable ARM Scalable Vector Extensions v2
     halide_target_feature_egl,                    ///< Force use of EGL support.
 
-    halide_target_feature_arm_dot_prod,  ///< Enable ARMv8.2-a dotprod extension (i.e. udot and sdot instructions)
-    halide_target_feature_end            ///< A sentinel. Every target is considered to have this feature, and setting this feature does nothing.
+    halide_target_feature_arm_dot_prod,       ///< Enable ARMv8.2-a dotprod extension (i.e. udot and sdot instructions)
```
steven-johnson, line 1319: nit: since the features list is in an arbitrary order, probably cleaner to group all the cuda features together

abadams: Is it in arbitrary order? I wasn't sure if the target enum values was part of our interface

steven-johnson: Yes, we absolutely do not promise stability of this enum over multiple releases of Halide

abadams:
Failures unrelated.

PR #5208, opened 2020-08-24
Avoid C++ name mangling issues when calling between runtime modules

abadams:
Different runtime modules are initially compiled to llvm bitcode with
different target triples (to support stdcall stuff on windows without
requiring a precompiled windows version of every single runtime module).
These triples are unified before the modules are linked, but before that
happens we need to avoid anything target-triple-sensitive done by clang
when compiling to llvm assembly (e.g. structs that need padding bytes).

One such issue is calling c++ functions across runtime modules. The name
mangling could work out differently on both sides of the call. This PR
removes all instances of this (at least the ones that go via
runtime_internal.h)

The functions declared in Halide::Runtime::Internal didn't seem any more
internal than the extern "C" functions declared immediately above it, so
I promoted those functions (e.g. halide_abort()) to there.


src/runtime/runtime_internal.h
```diff
@@ -217,15 +217,20 @@ ALWAYS_INLINE T reinterpret(const U &x) {
...
+// because it's possible for them to be compiled with different c++
+// name mangling due to mixing and matching target triples.
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+// Empty
+}
 }  // namespace Runtime
 }  // namespace Halide
+using namespace Halide::Runtime::Internal;
```
steven-johnson, line 233: If we can't safely use this to communicate between modules, should we / can we use anonymous namespaces for this purpose instead, to ensure it won't happen by accident?

src/runtime/alignment_128.cpp
```diff
@@ -1,13 +1,5 @@
 #include "runtime_internal.h"
 
-namespace Halide {
-namespace Runtime {
-namespace Internal {
-
-WEAK_INLINE int halide_malloc_alignment() {
+extern "C" WEAK_INLINE int halide_malloc_alignment() {
```
steven-johnson, line 3: suggestion: rename these calls `halide_internal_foo()` to preserve the 'internalness' of them in the name? (no one externally should be overriding them anyway so theoretically this shouldn't break anything.)

shoaibkamil: +1.  I think it'll be better in the long run to easily be able to see at a glance it's internal.

shoaibkamil:
Thanks for doing this-- mostly looks good to me, will pull and test locally late tonight.

src/runtime/alignment_128.cpp
dsharletg: There's another convention used elsewhere to prefix with a leading underscore, is that for signaling internal?

src/runtime/runtime_internal.h
```diff
@@ -183,17 +183,17 @@ struct halide_pseudostack_slot_t {
...
+WEAK void halide_use_jit_module();
+WEAK void halide_release_jit_module();
 
-// A convenient namespace for weak functions that are internal to the
-// halide runtime.
-namespace Halide {
-namespace Runtime {
-namespace Internal {
+WEAK_INLINE int halide_malloc_alignment();
+WEAK_INLINE void halide_abort();
```
slomp, line 190: I think we need to use `extern "C"`here as well

src/runtime/alignment_128.cpp
abadams: Yes, we have three active conventions for internal stuff:

1) Halide::Runtime::Internal
2) _halide
3) Just call it halide_ (most of the existing stuff in runtime_internal.h was in this category)

In this PR I deduped 1 into 3. I could further dedup 3 into 2, but that would be a much larger change and I would prefer to do it in a non-bugfix PR.

src/runtime/runtime_internal.h
abadams: I considered it, but that also implies a linkage class, and we're being very careful with linkage right now. I'm not sure if it would be safe.

src/runtime/alignment_128.cpp
shoaibkamil: SGTM

shoaibkamil:
LGTM but not sure about the failed (possibly unrelated) gcc buildbot.  Verified it fixes it locally.  Thanks Andrew!

PR #5209, opened 2020-08-24
Fix expected files list for lesson 15 test.

alexreinking:
When not cross compiling #5186 allows CMake to bundle generated object files together into a static lib. That means the library will have "lib" prepended to it (on Unix-y systems, at least. See [`CMAKE_STATIC_LIBRARY_PREFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LIBRARY_PREFIX.html)). Unfortunately, that broke the lesson 15 tutorial test, which had the names mostly hard-coded.

Thanks, @Infinoid for reporting this.

@steven-johnson -- are the tutorials not tested on CI, how did we miss this?

### DIFF
--- a/tutorial/CMakeLists.txt
+++ b/tutorial/CMakeLists.txt
@@ -1,8 +1,5 @@
-set(IMAGES images/gray.png images/rgb.png)
-
-foreach (FILE IN LISTS IMAGES)
-    configure_file("${FILE}" "${FILE}" COPYONLY)
-endforeach ()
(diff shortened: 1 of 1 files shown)


steven-johnson:
> @steven-johnson -- are the tutorials not tested on CI, how did we miss this?

¯\_(ツ)_/¯

steven-johnson:
(More seriously: I suspect some tutorials may be failing on some configs and we blacklisted the whole thing. I'll re-enable them and see if we can get more granular.)

PR #5215, opened 2020-08-25
Thoroughly document CMake build

### LINKED ISSUE #5207: windows vcpkg instructions are missing python3
This is required now, I think? The instructions should be updated to include installing it via vcpkg, if that's the correct thing.

alexreinking:
This PR turns `README_cmake.md` into a comprehensive guide to the three main usage stories of the Halide CMake build:

1. Compiling or packaging Halide from source.
2. Building Halide programs using the official CMake package.
3. Contributing to Halide and updating the build files.

**Direct link for easy review**: https://github.com/halide/Halide/blob/docs/cmake/README_cmake.md

Fixes #5207

### DIFF
--- a/README_cmake.md
+++ b/README_cmake.md
@@ -1,22 +1,516 @@
-# CMake Build Rules for Halide
+# Halide and CMake
 
-## Overview
+This is a comprehensive guide to the three main usage stories of the Halide
+CMake build.
+
+1. Compiling or packaging Halide from source.
+2. Building Halide programs using the official CMake package.
+3. Contributing to Halide and updating the build files.
+
+The following sections cover each in detail.
+
+## Table of Contents
+
+- [Getting started](#getting-started)
+  - [Installing CMake](#installing-cmake)
+  - [Installing dependencies](#installing-dependencies)
+- [Building Halide with CMake](#building-halide-with-cmake)
+  - [Basic build](#basic-build)
+  - [Build options](#build-options)
+    - [Find module options](#find-module-options)
+- [Using Halide from your CMake build](#using-halide-from-your-cmake-build)
+  - [A basic CMake project](#a-basic-cmake-project)
+  - [JIT mode](#jit-mode)
+  - [AOT mode](#aot-mode)
+    - [Autoschedulers](#autoschedulers)
+    - [RunGenMain](#rungenmain)
+  - [Halide package documentation](#halide-package-documentation)
+    - [Components](#components)
+    - [Variables](#variables)
+    - [Imported targets](#imported-targets)
+    - [Functions](#functions)
+      - [`add_halide_library`](#add_halide_library)
+- [Contributing CMake code to Halide](#contributing-cmake-code-to-halide)
+  - [General guidelines and best practices](#general-guidelines-and-best-practices)
+    - [Prohibited commands list](#prohibited-commands-list)
+    - [Prohibited variables list](#prohibited-variables-list)
+  - [Adding tests](#adding-tests)
+  - [Adding apps](#adding-apps)
+
+# Getting started
+
+This section covers installing a recent version of CMake and the correct
+dependencies for building and using Halide. If you have not used CMake before,
+we strongly suggest reading through the [CMake documentation][cmake-docs] first.
+
+## Installing CMake
+
+Halide requires at least version 3.16, which was released in November 2019.
+Fortunately, getting a recent version of CMake couldn't be easier, and there are
+multiple good options on any system to do so. Generally, one should always have
+the most recent version of CMake installed system-wide. CMake is committed to
+backwards compatibility and even the most recent release can build projects over
+a decade old.
+
+### Cross-platform
+
+The Python package manager `pip3` has the newest version of CMake at all times.
+This might be the most convenient method since Python 3 is an optional
+dependency for Halide, anyway.
+
+```
+$ pip3 install --upgrade cmake
+```
+
+See the [PyPI website][pypi-cmake] for more details.
+
+### Windows
+
+On Windows, there are three primary methods for installing an up-to-date CMake:
+
+1. If you have Visual Studio 2019 installed, you can get CMake 3.17 through the
+   Visual Studio installer. This is the recommended way of getting CMake if you
+   are able to use Visual Studio 2019. See Microsoft's
+   [documentation][vs2019-cmake-docs] for more details.
+2. If you use [Chocolatey][chocolatey], its [CMake package][choco-cmake] is kept
+   up to date. It should be as simple as `choco install cmake`.
+3. Otherwise, you should install CMake from [Kitware's website][cmake-download].
+
+### macOS
+
+On macOS, the [Homebrew][homebrew] [CMake package][brew-cmake] is kept up to
+date. Simply run:
+
+```
+$ brew update
+$ brew install cmake
+```
+
+to install the newest version of CMake. If your environment prevents you from
+installing Homebrew, the binary release on [Kitware's website][cmake-download]
+is also a viable option.
+
+### Ubuntu Linux
+
+There are a few good ways to install a modern CMake on Ubuntu:
+
+1. If you're on Ubuntu Linux 20.04 (focal), then simply running
+   `sudo apt install cmake` will get you CMake 3.16.
+2. If you are on an older Ubuntu release or would like to use the newest CMake,
+   try installing via the snap store: `snap install cmake`. Be sure you do not
+   already have `cmake` installed via APT. The snap package automatically stays
+   up to date.
+3. For older versions of Debian, Ubuntu, Mint, and derivatives, Kitware provides
+   an [APT repository][cmake-apt] with up-to-date releases. Note that this is
+   still useful for Ubuntu 20.04 because it will remain up to date.
+4. If all else fails, you might need to build CMake from source (eg. on old
+   Ubuntu versions running on ARM). In that case, follow the directions posted
+   on [Kitware's website][cmake-from-source].
+
+For other Linux distributions, check with your distribution's package manager or
+use pip as detailed above. Snap packages might also be available.
+
+**Note:** On WSL 1, the snap service is not available; in this case, prefer to
+use the APT repository. On WSL 2, all methods are available.
+
+## Installing dependencies
+
+We generally recommend using a package manager to fetch Halide's dependencies.
+Except where noted, we recommend using [vcpkg][vcpkg] on Windows,
+[Homebrew][homebrew] on macOS, and APT on Ubuntu 20.04 LTS.
+
+Only LLVM and Clang are _absolutely_ required to build Halide. Halide always
+supports three LLVM versions: the current major version, the previous major
+version, and trunk. The LLVM and Clang versions must match exactly. For most
+users, we recommend using a binary release of LLVM rather than building it
+yourself.
+
+However, to run all of the tests and apps, an extended set is needed. This
+includes [lld][lld], [Python 3][python], [libpng][libpng], [libjpeg][libjpeg],
+[Doxygen][doxygen], [OpenBLAS][openblas], [ATLAS][atlas], and [Eigen3][eigen].
+While not required to build any part of Halide, we find that [Ninja][ninja] is
+the best backend build tool across all platforms.
+
+Note that CMake has many special variables for overriding the locations of
+packages and executables. A partial list can be found in the
+["find module options"](#find-module-options) section below, and more can be
+found in the documentation for the CMake [find_package][find_package] command.
+Normally, you should prefer to make sure your environment is set up so that
+CMake can find dependencies automatically. For instance, if you want CMake to
+use a particular version of Python, create a [virtual environment][venv] and
+activate it _before_ configuring Halide.
+
+### Windows
+
+We assume you have vcpkg installed at `D:\vcpkg`. Follow the instructions in the
+[vcpkg README][vcpkg] to install. Start by installing LLVM.
+
+```
+D:\vcpkg> .\vcpkg install llvm[target-all,enable-assertions,clang-tools-extra]:x64-windows
+D:\vcpkg> .\vcpkg install llvm[target-all,enable-assertions,clang-tools-extra]:x86-windows
+```
+
+This will also install Clang and LLD. The `enable-assertions` option is not
+strictly necessary but will make debugging during development much smoother.
+These builds will take a long time and a lot of disk space. After they are
+built, it is safe to delete the intermediate build files and caches in
+`D:\vcpkg\buildtrees` and `%APPDATA%\local\vcpkg`.
+
+Then install the other libraries:
+
+```
+D:\vcpkg> .\vcpkg install libpng:x64-windows libjpeg-turbo:x64-windows openblas:x64-windows eigen3:x64-windows
+D:\vcpkg> .\vcpkg install libpng:x86-windows libjpeg-turbo:x86-windows openblas:x86-windows eigen3:x86-windows
+```
+
+To build the documentation, you will need to install [Doxygen][doxygen]. This
+can be done either through [Chocolatey][choco-doxygen] or from the [Doxygen
+website][doxygen-download].
+
+```
+> choco install doxygen
+```
+
+To build the Python bindings, you will need to install Python 3. This should be
+done by running the official installer from the [Python website][python]. Be
+sure to download the debugging symbols through the installer. This will require
+using the "Advanced Installation" workflow. Although it is not strictly
+necessary, it is convenient to install Python system-wide on Windows (ie.
+`C:\Program Files`). This makes it easy for CMake to find without needing to
+manually set the `PATH`.
+
+Once Python is installed, you can install the Python module dependencies either
+globally or in a [virtual environment][venv] by running
+
+```
+> pip3 install -r .\python_bindings\requirements.txt
+```
+
+from the root of the repository.
+
+If you would like to use [Ninja][ninja], note that it is installed alongside
+CMake when using the Visual Studio 2019 installer. Alternatively, you can
+install via [Chocolatey][choco-ninja] or place the [pre-built
+binary][ninja-download] from their website in the PATH.
+
+```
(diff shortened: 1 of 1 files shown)


README_cmake.md
```diff
@@ -1,106 +1,377 @@
...
+want CMake to use a particular version of Python, create a virtual environment
+and activate it _before_ configuring Halide.
+
+### Windows
+
+We assume you have vcpkg installed at `D:\vcpkg`. Follow the instructions in the
+vcpkg README to install. First install LLVM via vcpkg.
+
+```
+D:\vcpkg> .\vcpkg install llvm[target-all,clang-tools-extra]:x64-windows
```
abadams, line 141: I'd recommend people turn on enable-assertions for llvm, otherwise it just crashes instead of giving an error message when things go weird.

README_cmake.md
```diff
@@ -1,106 +1,377 @@
...
-### Getting CMake
+```
+> choco install doxygen
+```
 
-Getting a recent version of CMake couldn't be easier, and there are multiple
-good options on any system to do so.
+To build the Python bindings, you will need to install Python 3. This should be
+done by running the official installer from the Python website. Be sure to
+install system-wide (in `C:\Program Files`) and download the debugging symbols
```
abadams, line 166: There's no particular requirement it goes in Program files, right? You just need to check the box that says to add it to the path.

README_cmake.md
```diff
@@ -1,106 +1,377 @@
...
-`sudo apt install cmake` is sufficient.
+### Windows
 
-For older versions of **Debian, Ubuntu, Mint, and derivatives**, Kitware
-provides an [APT repository](https://apt.kitware.com/) with up-to-date releases.
-Note that this is still useful for Ubuntu 20.04 because it will remain up to
-date.
+If you plan to use the Ninja generator, be sure to be in the developer command
+prompt corresponding to your intended environment. In particular, if you wish to
+compile a 32-bit version of Halide, you should use the 64-bit tools. You can
```
abadams, line 220: 32 vs 64 bit mismatch in this sentence?

README_cmake.md
alexreinking: I vaguely recall CMake having trouble finding it installed to a non-standard location

README_cmake.md
alexreinking: Nope. I am about to push a clarification. Using the 32-bit host tools results in a linker crash.

README_cmake.md
```diff
@@ -1,106 +1,377 @@
...
+```
+
+### macOS and Linux
+
+The instructions here are straightforward. Assuming your environment is set up
+correctly, just run:
+
+```
+dev@host:~/Halide$ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S . -B build
+dev@host:~/Halide$ cmake --build ./build
```
abadams, line 260: The section below on using Halide from your cmake build seems to require an install, but these instructions don't produce one, right? I typically need to set CMAKE_INSTALL_PREFIX or whatever it's called too.

alexreinking: Good catch! I'll add those directions

README_cmake.md
```diff
@@ -183,61 +454,141 @@ cases, you must declare that two Halide libraries share a runtime, like so:
...
+  `add_subdirectory` as a Git or FetchContent submodule.
+
+Variables set by the package:
+
+- `Halide_VERSION` -- The full version string of the loaded Halide package
+- `Halide_VERSION_MAJOR` -- The major version of the loaded Halide package
+- `Halide_VERSION_MINOR` -- The minor version of the loaded Halide package
+- `Halide_VERSION_PATCH` -- The patch version of the loaded Halide package
+- `Halide_VERSION_TWEAK` -- The tweak version of the loaded Halide package
+- `Halide_HOST_TARGET` -- The Halide target triple corresponding to "host" for
```
abadams, line 555: I don't understand this entirely. Is this detected at the time you use the package, or at the time the package was built? (host could be different in the two cases, and the one at package creation time is not meaningful for use of Halide)

I ask because the other variables set are properties known at the time the Halide package was created.

alexreinking: These are set when the package is created. The meaning of "host" is actually fixed when Halide is compiled, per our discussion on Halide/build on Gitter. Stamping that down at package creation time is the correct behavior. Detecting the host machine when the package is loaded is wrong for 32-bit Halide on 64-bit Windows

README_cmake.md
```diff
@@ -284,6 +634,16 @@ same platform/bits/os triple (eg. `arm-32-linux`). Features that are in common
 among all targets, including device libraries (like `cuda`) should go in
 `FEATURES`.
 
+Every element of `TARGETS` must begin with the same `arch-bits-os` triple. This
+function understands two _meta-triples_, `host` and `cmake`. The meta-triple
+`host` is equal to the `arch-bits-os` triple used to compile Halide. On
```
abadams, line 639: This bit concerns me. The host architecture used to compile Halide-the-compiler is irrelevant. E.g. you could have built Halide on a machine without avx512 and be using it on a machine with avx512.

README_cmake.md
```diff
@@ -304,17 +664,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+  scenario.
+
+### Prohibited commands list
+
+As mentioned above, using directory properties is brittle and they are therefore
+_not allowed_. The following functions may not appear in any new CMake code.
+
+| Command                             | Alternative                                               |
+| ----------------------------------- | --------------------------------------------------------- |
+| `add_compile_defintions`            | Use `target_compile_definitions`                          |
```
abadams, line 773: typo: defintions

README_cmake.md
abadams: I don't understand how that target triple could be useful then (due to the feature flags being potentially wrong). What is it used for?

README_cmake.md
alexreinking: As I mentioned above, it isn't because it changes the meaning of the "host" meta-target. Of course, the OS and architecture have to remain fixed, but the bitness doesn't.

README_cmake.md
alexreinking: One issue is that just saying "host" doesn't guarantee that you aren't cross-compiling and you can get linker errors re: 32-bit/64-bit mismatch on Windows doing that.

But what's stored in `Halide_HOST_TARGET` is _just_ the triple and doesn't include feature flags. This means that when `host` wouldn't be cross-compiling, it can use "`host`" as the target that's passed to the generator, which lets it take full advantage of the host CPU.

README_cmake.md
steven-johnson: > I vaguely recall CMake having trouble finding it installed to a non-standard location

leave out everything after "CMake having trouble finding", it's cleaner

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
+
+Halide reads and understands several options that can configure the build. The
+following are the most consequential and control how Halide is actually
+compiled.
+
+| Option                       | Default               | Description                                                                                                      |
+| ---------------------------- | --------------------- | ---------------------------------------------------------------------------------------------------------------- |
+| `Halide_BUNDLE_LLVM`         | `OFF`                 | When building Halide as a static library, unpack the LLVM static libraries and add those objects to libHalide.a. |
+| `Halide_SHARED_LLVM`         | `OFF`                 | Link to the shared version of LLVM. Not available on Windows.                                                    |
+| `Halide_ENABLE_RTTI`         | _inherited from LLVM_ | Enable RTTI when building Halide. Recommended to be set to `ON`                                                  |
```
steven-johnson, line 304: Note that RTTI *must* be enabled in order to build/use the Python bindings for Halide.

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
+| Option            | Default | Description                                                |
+| ----------------- | ------- | ---------------------------------------------------------- |
+| `WITH_WABT`       | `ON`    | Include WABT Interpreter for WASM testing                  |
+| `WITH_WASM_SHELL` | `ON`    | Download a wasm shell (e.g. d8) for testing AOT wasm code. |
+
+# Using Halide from your CMake build
+
+This section assumes some basic familiarity with CMake but tries to be explicit
+in all its examples. To learn more about CMake, consult the documentation and
+engage with the community on the CMake Discourse.
```
steven-johnson, line 372: Link to the Discourse site would be appropriate

README_cmake.md
```diff
@@ -304,17 +695,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+is equivalent to setting `-e registration` at the generator command line.
 
 Lastly, each of the `extra-output` arguments directly correspond to an extra
 output (via `-e`) from the generator. The value `OUTVAR` names a variable into
 which a path (relative to `CMAKE_CURRENT_BINARY_DIR`) to the extra file will be
 written.
+
+## Cross compiling
+
+Cross-compiling is a fraught topic in CMake when some binaries must be built for
```
steven-johnson, line 708: nit: 'fraught' is a scary word here and the wording is maybe a bit awkward. Maybe: "Cross-compiling in CMake can be tricky, since CMake doesn't easily support compiling for both the host platform and the cross-platform within the same build"?

README_cmake.md
```diff
@@ -304,17 +695,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+comprehensive, the following serves as a guide for writing quality CMake code
+and outlines the code quality expectations we have as they apply to CMake.
+
+## General guidelines and best practices
+
+The following are some common mistakes that lead to subtly broken builds.
+
+- **Reading the build directory.** While setting up the build, the build
+  directory should be considered _write only_. Using the build directory as a
+  read/write temporary directory is acceptable as long as all temp files are
```
steven-johnson, line 777: Nit: why not just forbid read/write access entirely? Don't have to worry about cleaning up temp files if they aren't ever created...

README_cmake.md
```diff
@@ -304,17 +695,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+  unwanted details about the build environment into packages. Some information
+  is not accurate or available except via generator expressions, eg. the build
+  configuration.
+- **Using the wrong variable.** `CMAKE_SOURCE_DIR` doesn't always point to the
+  Halide source root. When someone uses Halide via FetchContent, it will point
+  to _their_ source root instead. The correct variable is `Halide_SOURCE_DIR`.
+  If you want to know if the compiler is MSVC, check it directly with the `MSVC`
+  variable; don't use `WIN32`. That will be wrong when compiling with clang on
+  Windows.
+- **Using directory properties.** Directory properties have vexing behavior and
```
steven-johnson, line 790: Maybe include examples of 'bad, don't do this' and 'do this instead'

README_cmake.md
```diff
@@ -304,17 +695,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+  are essentially deprecated from CMake 3.0+. Propagating target properties is
+  the way of the future.
+- **Using the wrong visibility.** Target properties can be `PRIVATE`,
+  `INTERFACE`, or both (aka `PUBLIC`). Pick the most conservative one for each
+  scenario.
+
+### Prohibited commands list
+
+As mentioned above, using directory properties is brittle and they are therefore
+_not allowed_. The following functions may not appear in any new CMake code.
```
steven-johnson, line 800: ...are we using any of the forbidden functions in existing code? (if so, maybe let's open a tracking issue to nuke them)

README_cmake.md
```diff
@@ -304,17 +695,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+which must remain in alphabetical order. Generator tests are trickier, but
+following the existing examples is a safe way to go.
+
+## Adding apps
+
+If you're contributing a new app to Halide: great! Thank you! There are a few
+guidelines you should follow when writing a new app.
+
+- Write the app as if it were a top-level project. You should call
+  `find_package(Halide)` and set the C++ version to 14.
```
steven-johnson, line 903: Why set C++ to 14? (Earlier in this doc we show setting the C++ version to 11-or-later so this is a bit contradictory)

README_cmake.md
```diff
@@ -304,17 +695,219 @@ and the `USE_RUNTIME` option is ignored. Other options work as expected.
...
+- Call `enable_testing()` and add a small test that runs the app.
+- Don't assume you have access to a GPU. Write your schedules to be robust to
```
steven-johnson, line 905: - Don't assume you will be run only on specific architecture and/or OS and/or bitness (or something like that)

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
-for dynamic pipelines or an **ahead-of-time (AOT) compiler** for static
-pipelines.
+### Cross-platform
+
+The Python package manager `pip` packages the newest version of CMake at all
+times. This might be the most convenient method since Python 3 is an optional
+dependency for Halide, anyway.
+
+```
+$ pip install --upgrade cmake
```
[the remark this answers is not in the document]
alexreinking: I'll change it to `pip3`. It does matter.

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
+  - [General guidelines and best practices](#general-guidelines-and-best-practices)
+    - [Prohibited commands list](#prohibited-commands-list)
+    - [Prohibited variables list](#prohibited-variables-list)
+  - [Adding tests](#adding-tests)
+  - [Adding apps](#adding-apps)
+
+# Getting started
+
+This section covers installing a recent version of CMake and the correct
+dependencies for building and using Halide.
```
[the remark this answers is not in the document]
alexreinking: Will do.

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
+1. If you're on **Ubuntu Linux 20.04** (focal), then simply running
+   `sudo apt install cmake` is sufficient.
+2. If you are on an older Ubuntu release or would like to use a newer CMake,
+   first try installing via the snap store: `snap install cmake`. Be sure you do
+   not already have `cmake` installed via APT. The snap package automatically
+   stays up to date.
+3. For older versions of **Debian, Ubuntu, Mint, and derivatives**, Kitware
+   provides an [APT repository](https://apt.kitware.com/) with up-to-date
+   releases. Note that this is still useful for Ubuntu 20.04 because it will
+   remain up to date.
```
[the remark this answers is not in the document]
alexreinking: Sure. The snap packages work for arm when you have a new enough Ubuntu release (not sure about the minimum version). But the link would be useful regardless.

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
+built, it is safe to delete the intermediate build files and caches in
+`%VCPKG_ROOT%\buildtrees` and `%APPDATA%\local\vcpkg`.
+
+Then install the other libraries:
 
 ```
-$ mkdir build
-$ cd build
-$ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_MODULE_PATH="/path/to/Halide" ..
+D:\vcpkg> .\vcpkg install libpng:x64-windows libjpeg-turbo:x64-windows openblas:x64-windows eigen3:x64-windows
```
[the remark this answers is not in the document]
alexreinking: I actually don't know. Will look into it and get back to you.

README_cmake.md
```diff
@@ -1,106 +1,408 @@
...
+On macOS, it is possible to install all dependencies via Homebrew:
 
-On Windows, you have two good options for installing CMake. First, if you have
-**Visual Studio 2019** installed, you can get CMake 3.16 through the Visual
-Studio installer. This is the recommended way of getting CMake if you are able
-to use Visual Studio 2019. See the
-[Visual C++ documentation](https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=vs-2019)
-for more details.
+```
+$ brew install llvm libpng libjpeg python@3.8 openblas doxygen ninja
```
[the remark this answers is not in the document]
alexreinking: `llvm` will always be the newest here. I forget why I specified a Python version, but maybe brew doesn't provide a default for that package?

README_cmake.md
alexreinking: Writing is okay if you're doing `configure_file` or the like. It's more about avoiding a bug we used to have that read a file that was later created, so it would work on subsequent runs, but not the first.

README_cmake.md
alexreinking: I tried to do that a bit with the tables in the following sections, but I could definitely flesh this out more.

README_cmake.md
alexreinking: We are not, except in three places where it is _absolutely_ necessary:

1. We use `define_property` to track the Halide targets a runtime library needs to support, which might change throughout the configure step. The property is used in a generator expression to construct a correct command line.
2. We use a `macro` in our package config code to gracefully fail by calling `return()` in the package config scope. This is required by the `find_package` mechanism. The alternative would be to duplicate a 20-line block 8 times.
3. We use `find(GLOB ...)` in our package config code to load dependency scripts that might or might not be there depending on platform/configuration. It's safe here because it's not globbing for source files for a target, but for CMake scripts to be included.

README_cmake.md
alexreinking: Sorry, there is some lingering confusion about which C++ version to use. All of our supported compilers support C++14, so it is safe to request this either way. I will go through and switch things to C++11 until we officially flip the switch to 14.

README_cmake.md
alexreinking: :+1:

README_cmake.md
steven-johnson: I really want us to bypass 14 and go straight to 17. Preferably soon.

alexreinking:
This is ready for a closer look, if you want. Otherwise, we can go ahead and merge and work on it more later.

README_cmake.md
alexreinking: Missed this earlier. Used your wording.

steven-johnson:
LGTM

PR #5216, opened 2020-08-25
Replace large code model build option with target feature.

alexreinking:
The CMake build has an option, `HALIDE_USE_CODEMODEL_LARGE` that I (re)discovered while writing documentation. I am not sure why it exists. It's not mentioned anywhere in the Makefile and it doesn't seem to be used by any of our tests.

This PR turns it into a target feature rather than a build option, but another valid approach might be to remove it outright. This is probably the appropriate venue for this discussion.

### DIFF
--- a/python_bindings/src/PyEnums.cpp
+++ b/python_bindings/src/PyEnums.cpp
@@ -144,6 +144,7 @@ void define_enums(py::module &m) {
         .value("SVE", Target::Feature::SVE)
         .value("SVE2", Target::Feature::SVE2)
         .value("ARMDotProd", Target::Feature::ARMDotProd)
+        .value("LLVMLargeCodeModel", Target::Feature::LLVMLargeCodeModel)
         .value("FeatureEnd", Target::Feature::FeatureEnd);
 
     py::enum_<halide_type_code_t>(m, "TypeCode")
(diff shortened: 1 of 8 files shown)


abadams:
I think the short version is that we never got to the bottom of why it was necessary for some people, so it got flipped back and forth a few times, and eventually ended up with a #define while we debugged. The last person to use it seems to have been @benoitsteiner, so he may know if it's still necessary and would work for him as a target flag.

alexreinking:
Wondering if @benoitsteiner would chime in?

benoitsteiner:
The best explanation why we can't use the default small code model is available at https://stackoverflow.com/questions/40493448/what-does-the-codemodel-in-clang-llvm-refer-to. In a nutshell, we JIT all the pipelines, and since we also have ALSR offsets may not fit in 32 bits. As a result we need the large code model (it's possible the medium model might work as well, I've never tried it).
Having the code model as an option to LLVM should be fine, we'll just need to remember to set it when we upgrade Halide.

alexreinking:
Thanks for the context @benoitsteiner . I've just rebased this onto master, so would appreciate a review.

benoitsteiner (approved):

steven-johnson (approved):

steven-johnson:
Ready to land?

alexreinking:
> 
> 
> Ready to land?

Yup! Just had to rerun one of the buildbots

Issue #5156, opened 2020-08-04
ABI issue with LLVM11 and D3D12 (__stdcall)


slomp:
The following code snippet from `d3d12compute.cpp`:
```
D3D12_CPU_DESCRIPTOR_HANDLE hCPU = binder->CPU[UAV];
hCPU.ptr += i * descriptorSize;
(*device)->CreateUnorderedAccessView(NULL, NULL, &NullDescUAV, hCPU);
```
produces the following assembly
```
LLVM 10                                               LLVM 11
   0x2c619a06bf1:       mov    %r14,-0x1d0(%rbp)         0x19f89f06c01:       mov    %r14,-0x1e0(%rbp)
   ...                                                   ...
   0x2c619a06c09:       mov    %rcx,-0x220(%rbp)         0x19f89f06c19:       mov    %rcx,-0x1d0(%rbp)    <------
                                                         0x19f89f06c20:       lea    -0x1d0(%rbp),%rcx
   0x2c619a06c10:       mov    %rcx,0x20(%rsp)           0x19f89f06c27:       mov    %rcx,0x20(%rsp)
   ...                                                   ...
                        callq  *%rax                                          callq  *%rax
```
On the left, the value of `hCPU.ptr` is in `%rcx`, which is then placed on the stack (`0x20(%rsp)`) to serve as an argument to the subsequent `callq` to `CreateUnorderedAccessView`. This is the correct ABI behavior.

On the right, LLVM11 decided to replace the value of `%rcx` with the **address of that value**  (notice the `lea` instruction).
This is basically turning something that should have been passed by-value into something that is passed by-reference

In short, LLVM 11 is insisting in passing the struct by reference, even though it fits entirely on a 64bit word and should have been passed by-value:
```
typedef struct D3D12_CPU_DESCRIPTOR_HANDLE {
    SIZE_T ptr;
} D3D12_CPU_DESCRIPTOR_HANDLE;
```

abadams:
Could this be a target triple issue?

slomp:
We'll see, it's possible; I'm working on a clean, minimal reproducible code snippet for further investigation.

slomp:
Update:
Yes, it looks like we're muffling a rather important warning:
```
warning: '__stdcall' calling convention is not supported for this target [-Wignored-attributes]
```
because `mini_d3d12.h` is imposing the following:
```
#pragma clang diagnostic ignored "-Wignored-attributes"
```
[...]

PR #5217, opened 2020-08-26
ABI fix for D3D12

slomp:
Currently works for x64, but the general principle should extend to ARM as well.
Issue tracker here: https://github.com/halide/Halide/issues/5156


src/runtime/CMakeLists.txt
```diff
@@ -144,7 +143,8 @@ add_library(Halide_initmod OBJECT)
 
 set(CXX_WARNING_FLAGS -Wall -Werror -Wno-unused-function -Wcast-qual)
 set(RUNTIME_CXX_FLAGS -O3
-    -fno-vectorize -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables -fpic
+    -fno-vectorize -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables
+    #-fpic
```
slomp, line 147: Not entirely sure how to fix this in an elegant way... `-fpic` can't be used with `x64-windows`...

src/runtime/d3d12compute.cpp
```diff
@@ -159,7 +159,7 @@ void *d3d12_get_library_symbol(void *lib, const char *name) {
     return symbol;
 }
 
-#define d3d12_debug_break (*((volatile int8_t *)NULL) = 0)
+#define d3d12_debug_break() (*((volatile int *)NULL) = 0)
```
steven-johnson, line 162: For macros that are intended to be a 'statement', best practice is to wrap in `do{ } while(0)` (with no trailing semicolon):

```
#define d3d12_debug_break() do { (*((volatile int *)NULL) = 0); } while (0)
```

steven-johnson: (Also: holy crap, is dereferencing a nullptr the best way to force a debug break? that seems, uh, brutal...)

src/runtime/windows_d3d12compute_x86.cpp
```diff
@@ -0,0 +1 @@
+#include "d3d12compute.cpp"
```
steven-johnson, line 1: #include'ing a .cpp file is unusual enough that a comment explaining it is appropriate.

src/runtime/CMakeLists.txt
steven-johnson: Alas, I don't know enough about the issues to have a useful comment.

src/runtime/d3d12compute.cpp
abadams: I thought dereferencing NULL was UB, so the compiler is free to assume d3d12_debug_break can never be reached and may just prune any branches (e.g. asserts) that lead to this macro.

src/runtime/CMakeLists.txt
abadams: @alexreinking probably knows the right thing to do

src/runtime/d3d12compute.cpp
slomp: I wouldn't dare saying it's the best way, but it's the only "portable" and reliable ways to get a sig break in gdb and MSVC from this runtime module.
I believe @abadams is right that this is UB, but because of the volatile keyword, the compiler will just emit code and not prune it.
It's ugly, but it works and it's handy to have when it's needed.

alexreinking: This is already always compiled with Clang and Direct3D is inherently non-portable... what's wrong with calling `__builtin_trap()` or `__asm__("int $3")`?

alexreinking: On ARM, it's `brk #0xF000`

src/runtime/CMakeLists.txt
alexreinking: _Why_ can't `-fpic` be used with `x64-windows`? Regardless of how you implement it, that merits a comment. As far as the CMake hacking goes, I would do something like this:

```
foreach (j IN ITEMS 32 64)
  set(fpic -fpic)
  if (${j} EQUAL 32)
  ...
  else ()
    if (${i} MATCHES "windows_.*")
      set(fpic "") # comment explaining why this is
  ...
  foreach (SUFFIX IN ITEMS "" "_debug")
    ...
    add_custom_command(OUTPUT "${LL}"
                       COMMAND clang ${RUNTIME_CXX_FLAGS} ${fpic}
```

src/runtime/d3d12compute.cpp
slomp: When I tried `__builtin_trap()`, it would not break precisely on the stack where I was expecting to. Let's not waste too much energy on this minutia. This #define is pretty much some throw-away code that's there for debugging convenience every now and then.

alexreinking: Since it's not actually used anywhere, I guess it's fine (but then why leave it in?), but if it _were_, I wouldn't call it "minutiae". With as aggressive as C++ compilers generally, and clang in particular, are getting about exploiting UB, it would only be a matter of time...

src/runtime/CMakeLists.txt
slomp: Yeah, it's bizarre that `-fpic` can't be used, but clang wasn't happy:
```
clang : error : unsupported option '-fpic' for target 'x86_64-pc-windows-msvc'
```

alexreinking: Ah, after a little research, it seems that PIC doesn't apply to Windows at all because DLL function addresses are fixed, so there's no point in including it. Then we should set that `fpic` variable to empty when the target is windows and to `-fpic` otherwise.

slomp: Yeah, I think clang actually issues a warning saying that -fpic is not needed and will be ignored, but since we treat warnings as errors, this ends up interrupting the build as an error.

src/runtime/d3d12compute.cpp
slomp: Yeah, I'll remove it. When needed, I'll just type it (or try the other traps again).
I'm pretty much the only person touching the d3d12 backend at the moment, but I can see that becoming a point of confusion and UB as more people contribute to it and inadvertently abuse it.

src/runtime/CMakeLists.txt
```diff
@@ -144,12 +144,13 @@ add_library(Halide_initmod OBJECT)
...
-    #-fpic
     # Note: we don't want static locals to get thread synchronization stuff.
     -fno-threadsafe-statics)
 
 foreach (i IN LISTS RUNTIME_CPP)
     foreach (j IN ITEMS 32 64)
+        # -fpic needs special treatment; see below on windows 64bits
+		set(fpic -fpic)
         if (${j} EQUAL 32)
             if (${i} MATCHES "windows_.*")
```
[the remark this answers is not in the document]
alexreinking: Ah, I guess we don't because we don't supply "windows" as part of the target here.

shoaibkamil:
The buildbot failures don't seem related.  Ok to merge?

steven-johnson:
LGTM to land.

PR #5218, opened 2020-08-26
Cmake/metatargets

alexreinking:
In offline discussion with @abadams re: #5215 we decided that the `cmake` meta-target needs a re-think. It will now work as follows:

1. Manually setting the target triple (either through `add_halide_library(... TARGETS)` or the `Halide_TARGET` cache variable) to `cmake` results in the base target triple for the current CMake target. We document that it does _not_ detect feature flags (because it cannot).
2. Manually setting the target to `host` passes through verbatim (though we of course check to see if matches CMake's active target or not to set up the targets correctly).
3. When no target is specified, `host` is checked and if it doesn't match CMake's active target, it throws a loud AUTHOR_WARNING explaining the problem and defaults to `cmake`
4. Otherwise, it uses `host`

This is to fulfill the following design goals:

1. It should be legal to not specify a target to `add_halide_library`
2. When not specifying a target to `add_halide_library`, it should not cross-compile.
3. Silently giving bad performance is worse than a compiler error

---

Also fixes a couple small CMake oversights I discovered while writing the documentation:

1. `HALIDE_USE_CODEMODEL_LARGE` -> `Halide_USE_CODEMODEL_LARGE`
2. Ensure that cross-compiled `add_halide_library`-ies are globally visible, like normal targets (when not cross compiling)
3. Duplicated compiler flags from `Halide::LanguageOptions`
4. "cmake" -> "host" meta-target promotion was bugged

### DIFF
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -185,7 +185,7 @@ function(add_halide_library TARGET)
     ##
 
     if (crosscompiling)
-        add_library("${TARGET}" STATIC IMPORTED)
+        add_library("${TARGET}" STATIC IMPORTED GLOBAL)
         set_target_properties("${TARGET}" PROPERTIES
                               IMPORTED_LOCATION "${CMAKE_CURRENT_BINARY_DIR}/${GENERATOR_SOURCES}")
     else ()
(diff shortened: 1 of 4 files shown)


alexreinking:
I added it recently, in the cross compiling PR because we need a way to say "use the active target" (ie. "Don't cross compile")

alexreinking:
arm32 failure is due to #5224 - gonna try out the GitHub "squash and merge" button :)

PR #5222, opened 2020-08-27
Fix incremental builds with Ninja

alexreinking:
I noticed that incremental builds on Ninja usually fully rebuilt the runtime. Normally, this is pretty quick, but on Windows/WSL, the filesystem overhead makes this annoying.

CMake has support for scanning C/C++ sources that are processed by custom commands when using Make as a generator via the `IMPLICIT_DEPENDS [C/CXX]` argument. However, for Ninja, a custom depfile must be specified.

I thought "oh, this will be easy! I'll just add the -MD and -MF flags and point CMake to the dep-file". Not so simple. Ninja requires the target in the depfile to match the _exact_ (in this case relative) path to the output of the command. Clang doesn't provide an easy way to override this, and CMake doesn't have any helpers to make this more convenient. Sigh. All three tools conspire to make a fundamental build requirement as obnoxious as possible.

The solution is to run clang from the directory containing the Ninja build (`CMAKE_BINARY_DIR`, _not_ `Halide_BINARY_DIR`). This makes clang put the paths Ninja expects in the depfile. Then CMake has to compute the correct relative paths to everything at configure time.

Big, big kudos to the Ninja `-d explain` flag, which tells you EXACTLY why a command is being run, unlike Make which gives you a ridiculous and incomplete tree that merely hints at why things are running.

This also reveals that incremental builds are broken when using the Visual Studio/msbuild generator. There's no easy fix because msbuild has zero support for depfiles. We'll probably have to mark them all as always out-of-date somehow when neither Make nor Ninja are being used. 🤮

### DIFF
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -165,33 +165,51 @@ foreach (i IN LISTS RUNTIME_CPP)
 
         set(SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/${i}.cpp")
 
-        set(RUNTIME_DEFINES -DCOMPILING_HALIDE_RUNTIME -DLLVM_VERSION=${LLVM_VERSION} -DBITS_${j})
+        set(RUNTIME_DEFINES -DCOMPILING_HALIDE_RUNTIME -DBITS_${j})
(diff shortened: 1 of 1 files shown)


alexreinking:
Failure is due to #5224... merging.

PR #5227, opened 2020-08-30
Replace ANSI Win32 API calls with UTF-16; convert UTF-8 at boundary.

### LINKED ISSUE #5223: Halide Util.cpp uses Win32 functions that are not encoding-safe
Instead of using `CreateDirectoryA()` and other `<foo>A()` functions on Windows, we need to use the corresponding `CreateDirectoryW()` and other `W()` suffixed functions in order to handle directories that may have characters outside the system encoding.  The original reporter at Adobe suggests:

> It is OK to use std::string as their cross platform API. They must decide to use UTF8 and then convert UTF8 to wstring (UTF16) before calling Windows APIs

Currently, trying to use Halide in a user account that has characters outside the system encoding results in an unhandled exception at line 399 of Util.cpp, here:

https://github.com/halide/Halide/blob/d046701e452dcda4fce315a7fe95179d0e43e5bd/src/Util.cpp#L364-L400

alexreinking:
We should rethink the placement of the current directory code in `LLVM_Output.cpp`. That would let me reuse the code in `from_utf8` and `from_utf16`. 

Fixes #5223

### DIFF
--- a/src/LLVM_Output.cpp
+++ b/src/LLVM_Output.cpp
@@ -434,11 +434,24 @@ namespace {
 
 std::string get_current_directory() {
 #ifdef _WIN32
-    std::string dir;
-    char p[MAX_PATH];
-    DWORD ret = GetCurrentDirectoryA(MAX_PATH, p);
-    internal_assert(ret != 0) << "GetCurrentDirectoryA() failed";
-    dir = p;
+    DWORD dir_buf_size = GetCurrentDirectoryW(0, nullptr);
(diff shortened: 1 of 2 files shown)


shoaibkamil:
LGTM, thanks for doing this!  I'll just ping @slomp to take a look as well, since he's much more experienced with Windows.

steven-johnson:
> We should rethink the placement of the current directory code in `LLVM_Output.cpp`. That would let me reuse the code in `from_utf8` and `from_utf16`.

I would endorse moving these (and all other pathname/filesystem stuff) into Util.h, or better yet, creating a new FSUtil.h that owns all this stuff that somehow posix still can't deal with adequately...

PR #5229, opened 2020-08-31
Add unsigned long division to CodeGen_Internal.

aankit-ca:
Useful for HVX where no dedicated instruction for vector division exists.
@dsharletg The PR is not complete but would love to get your initial thoughts on this.
I'll add the bounds analysis check to see if we can skip some iterations of long division.
One con of this implementation is that we are doing den = den + 1 in lower_euclidean_division if den == 0 which we don't need to do for long_div.


test/correctness/long_div.cpp
```diff
@@ -0,0 +1,71 @@
+#include "Halide.h"
```
dsharletg, line 1: We already have https://github.com/halide/Halide/blob/7d836f0f3dbc202a371b8ad55831239c100cb39f/test/correctness/mul_div_mod.cpp

We should retrofit that test to cover this instead of adding a new test.

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,24 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
     }
 }
 
+Expr ulong_div(Expr num, Expr den) {
```
dsharletg, line 391: ulong makes it sound like it works off of the `ulong` type.

I think this should be called `unsigned_long_divide` or something instead.

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,24 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
...
+    internal_assert(num.type() == den.type());
+    Type ty = num.type();
+    internal_assert(ty.is_uint());
+    const int times = ty.bits();
```
dsharletg, line 395: I think this algorithm should have a few comments explaining it.

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,24 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
     }
 }
 
+Expr ulong_div(Expr num, Expr den) {
+    internal_assert(num.type() == den.type());
```
dsharletg, line 392: I think it would be a big optimization to enable the result is a smaller type than the inputs. For example, int32 / int16 -> int16 would be much faster to compute than int32 / int16 -> int32.

src/CodeGen_Hexagon.cpp
```diff
@@ -475,6 +515,11 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
     debug(2) << "Lowering after forwarding stores:\n"
              << body << "\n\n";
 
+    // Replace div with long div if lanes > 1
+    debug(1) << "Replacing div with long_div...\n";
+    body = long_div(body);
+    body = common_subexpression_elimination(body);
```
dsharletg, line 521: I think this is redundant with the common_subexpression_elimination in the long division implementation.

src/CodeGen_Hexagon.cpp
```diff
@@ -421,6 +421,45 @@ Stmt inject_hvx_lock_unlock(Stmt body, const Target &target) {
     return body;
 }
 
+// Replace Div Nodes with long division if lanes > 1.
+class LongDiv : public IRMutator {
```
dsharletg, line 425: Should this just be part of another existing mutator? Or maybe even just `CodeGen_Hexagon::visit(const Div*)`?

aankit-ca: I was initially thinking of making it a part of OptimizePatterns so as to take advantage of the peephole optimizations.

src/CodeGen_Hexagon.cpp
aankit-ca: Yeah. It's just that in the signed int case:
```
                Expr q = cast(a.type(), ulong_div(a_unsigned, b_unsigned));
                Expr a_neg = a >> make_const(a.type(), (a.type().bits() - 1));
                Expr b_neg = b >> make_const(a.type(), (a.type().bits() - 1));
```
a_neg and b_neg here are the same expressions created by lower_int_uint_div. That is why I thought of doing another round of cse.

src/CodeGen_Internal.cpp
aankit-ca: Right. I'll add some comments.

src/CodeGen_Internal.cpp
aankit-ca: Sure. Will change it to the suggested name

test/correctness/long_div.cpp
aankit-ca: Sure. Will do.

src/CodeGen_Internal.cpp
aankit-ca: Right. I'll add some bounds analysis check for ranges of num and den and perform operations in lowest type that num and den can be represented in. Also will add the bounds analysis check to see if some iterations can be skipped.

aankit-ca:
@dsharletg PTAL

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,42 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
     }
 }
 
+Expr unsigned_long_div(Expr num, const Expr &den, const Scope<Interval> &bounds) {
+    internal_assert(num.type() == den.type());
+    internal_assert(num.type().is_uint());
+    Type ty = num.type();
+    Expr leading_zeros, q;
+    q = make_zero(ty);
```
steven-johnson, line 396: Style nit: better to declare-and-init in the same statements, eg

```
    Expr q = make_zero(ty);
    Expr leading_zeros = cast(ty, count_leading_zeros(den));
```

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,42 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
...
+                     << (1 << shift) << ")\n";
+            continue;
+        }
+        Expr new_num = num - (den << shift_expr);
+        // Don't drop any set bits from den after shift. The bit is set if
+        // den << shift is no more than num.
+        Expr bit_set = ((shift_expr <= leading_zeros) && num >= (den << shift_expr));
+        // Update the numerator and the quotient.
+        num = select(bit_set, new_num, num);
+        q = select(bit_set, make_const(ty, 1 << shift) | q, q);
```
steven-johnson, line 422: If `ty` is `UInt(64)`, could this get the wrong answer, since `1 << shift` is going to be int-sized (32 bits)? The `1` may need to be cast to 64-bit.

steven-johnson: (also the similar expression on line 410)

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,59 @@ class OptimizePatterns : public IRMutator {
             // that they generate.
             internal_assert(op->args.size() == 3);
             return mutate(lower_lerp(op->args[0], op->args[1], op->args[2]));
+        } else if (op->is_intrinsic(Call::div_round_to_zero) &&
+                   !op->type.is_float() && op->type.lanes() > 1) {
```
steven-johnson, line 1085: Would it be clearer to use `op->type.is_vector()`?

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,59 @@ class OptimizePatterns : public IRMutator {
...
+            internal_assert(op->args.size() == 2);
+            Expr a = mutate(op->args[0]);
+            Expr b = mutate(op->args[1]);
+            debug(1) << "Using long division: (num: " << a << "); (den: " << b
+                     << ")\n";
+            if (a.type().is_uint()) {
```
steven-johnson, line 1091: Does `Call::div_round_to_zero` require that a and b have identical types? (If not, this call to `unsigned_long_div` could assert-fail)

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,59 @@ class OptimizePatterns : public IRMutator {
...
+    Expr visit(const Let *op) override {
+        return visit_let<Expr>(op);
+    }
+
+    Stmt visit(const LetStmt *op) override {
+        return visit_let<Stmt>(op);
+    }
+
+    Expr visit(const Div *op) override {
+        if (!op->type.is_float() && op->type.lanes() > 1) {
```
steven-johnson, line 1131: Would it be clearer to use `op->type.is_vector()`?

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,42 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
...
+    for (int i = 1; i <= times; i++) {
+        // Check if the bit at 'shift' index should be set in the result.
+        int shift = times - i;
+        Expr shift_expr = make_const(ty, shift);
+        // Skip iteration for long division if possible.
+        if (upper_value && (*upper_value < (1 << shift))) {
+            debug(1) << "Skipping iteration for long div as "
+                     << "(max_quotient: " << *upper_value << ") < "
+                     << (1 << shift) << ")\n";
+            continue;
```
dsharletg, line 414: Break? Or modify the loop bound up front instead of this?

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +388,42 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
     }
 }
 
+Expr unsigned_long_div(Expr num, const Expr &den, const Scope<Interval> &bounds) {
```
dsharletg, line 391: I think it would be better to make the upper bound of the result a parameter and pass that from the calling code, rather than requiring Scope<Interval> as a parameter.

src/CodeGen_Internal.cpp
aankit-ca: Done

src/CodeGen_Internal.cpp
aankit-ca: Done

src/HexagonOptimize.cpp
aankit-ca: Done

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,64 @@ class OptimizePatterns : public IRMutator {
...
+
+            a = mutate(a);
+            b = mutate(b);
+            debug(1) << "Using long div: (num: " << a << "); (den: " << b << ")\n";
+            if (a.type().is_uint()) {
+                return unsigned_long_div(a, b, upper_bound);
+            }
+            // Use unsigned long div for signed integer division.
+            Type ty = a.type().with_code(Type::UInt);
+            Expr a_unsigned = cast(ty, abs(a));
```
dsharletg, line 1101: You shouldn't need this cast, abs returns an unsigned type.

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,64 @@ class OptimizePatterns : public IRMutator {
...
+            Expr a = op->args[0];
+            Expr b = op->args[1];
+            // Run bounds analysis to estimate the range of result.
+            Expr extent_upper = find_constant_bound(a / b, Direction::Upper, bounds);
+            const uint64_t *upper_bound = as_const_uint(extent_upper);
+
+            a = mutate(a);
+            b = mutate(b);
+            debug(1) << "Using long div: (num: " << a << "); (den: " << b << ")\n";
+            if (a.type().is_uint()) {
```
dsharletg, line 1096: How about refactoring this so it just always casts a and b, does an unsigned long division, and then does a sign fixup if needed? That way, signed and unsigned are sharing the same division codepath.

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,64 @@ class OptimizePatterns : public IRMutator {
...
+            Expr b_neg = b >> make_const(a.type(), (a.type().bits() - 1));
+            return q * ((a_neg ^ b_neg) | 1);
         } else {
             return IRMutator::visit(op);
         }
     }
 
+    template<typename NodeType, typename T>
+    NodeType visit_let(const T *op) {
+        // We only care about vector lets.
```
dsharletg, line 1116: Is this true? I think we need the bounds of scalars here, so we can understand the bounds of divisions with broadcasts in them?

src/HexagonOptimize.cpp
dsharletg: Also, I think this logic should be added in a `long_div` function, which calls `unsigned_long_div(abs(a), abs(b), ...)` and does the sign fixup if needed (for signed arguments).

src/HexagonOptimize.cpp
```diff
@@ -1079,11 +1081,64 @@ class OptimizePatterns : public IRMutator {
...
+        return visit_let<Expr>(op);
+    }
+
+    Stmt visit(const LetStmt *op) override {
+        return visit_let<Stmt>(op);
+    }
+
+    Expr visit(const Div *op) override {
+        if (!op->type.is_float() && op->type.is_vector()) {
+            return mutate(lower_int_uint_div(op->a, op->b));
```
dsharletg, line 1137: Doesn't this need to handle signed integers? I think it should call a helper that includes the logic in the `div_round_to_zero` handler.

Also, I think this is going to have to work with `lower_euclidean_div` somehow, in order to get the intended behavior. Or we need the reverse logic for `div_round_to_zero`.

src/CodeGen_Internal.cpp
aankit-ca: Done

src/CodeGen_Internal.cpp
aankit-ca: Done

src/HexagonOptimize.cpp
aankit-ca: Done

src/HexagonOptimize.cpp
aankit-ca: Right now all the call::div_round_to_zero are generated through lower_euclidean_div which also needs a and b to have same types. All calls to div_round_to_zero should have the same types here.

dsharletg: In that case, we should probably assert in the other code path? It's probably not getting tested?

Also, if you do my other suggestions re: adding a helper for signed division, maybe that should clarify e.g. `long_div_round_to_zero` or something.

src/HexagonOptimize.cpp
aankit-ca: We do need to track scalars as well. Corrected it.

src/HexagonOptimize.cpp
aankit-ca: Done

src/HexagonOptimize.cpp
aankit-ca: Removed the cast

src/HexagonOptimize.cpp
aankit-ca: lower_int_uint_div can handle signed as well as unsigned integers. So we don't need to do anything else. Sorry I don't understand the need for the helper function here. It would be helpful if you can provide some more details on it.

Right. This would work with lower_euclidean_div. In cases other than lower_euclidean_div we change the division to mul with shift which is better than long_div. So I guess we only want to use long_div if the lower_int_uint_div generates calls to div_round_to_zero.

aankit-ca:
@dsharletg Ping

dsharletg (commented):
I just noticed, the test also tests mod. How does that get evaluated?

aankit-ca:
@dsharletg I had thought of adding Mod in a separate PR, but added it in this one itself as the changes we minimal.

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +389,57 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
...
+    debug(1) << "Using long div: (num: " << num << "); (den: " << den << ")\n";
+    internal_assert(num.type() == den.type());
+    Expr num_pos = abs(num);
+    Expr den_pos = abs(den);
+    Expr q = cast(num.type(), unsigned_long_div(num_pos, den_pos, max_abs));
+    // Correct the signs for quotient for signed integer division.
+    if (num.type().is_int()) {
+        Expr num_neg = num >> make_const(num.type(), (num.type().bits() - 1));
+        Expr den_neg = den >> make_const(num.type(), (num.type().bits() - 1));
+        q = q * ((num_neg ^ den_neg) | 1);
```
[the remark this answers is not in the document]
aankit-ca: Done

src/CodeGen_Internal.h
```diff
@@ -86,6 +86,11 @@ bool function_takes_user_context(const std::string &name);
  * non-positive. */
 bool can_allocation_fit_on_stack(int64_t size);
 
+/** Does a div_round_to_zero using binary long division for int/uint.
+ *  max_abs is the max value of abs(a/b)
+ *  Returns a when b == 0. */
```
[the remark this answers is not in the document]
aankit-ca: Rewrote it as: max_abs is the maximum absolute value of (a/b)
Not using the word result now as for long_mod_round_to_zero also max_value to be passed is abs(a/b)

src/CodeGen_Internal.cpp
```diff
@@ -440,6 +443,15 @@ Expr long_div(const Expr &num, const Expr &den, const uint64_t *max_abs) {
     return simplify(common_subexpression_elimination(q));
 }
 
+
+Expr long_mod_round_to_zero(const Expr &a, const Expr &b,
```
dsharletg, line 446: I think maybe `unsigned_long_div_round_to_zero` should return a struct/pair of the div (q) and mod (num after the loop) result. That might result in better code when both are used?

aankit-ca: Done

aankit-ca:
@dsharletg  PTAL

dsharletg (approved):
Thanks, this looks good to me! I just have one comment/TODO suggestion.

src/CodeGen_Internal.cpp
```diff
@@ -388,6 +389,64 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
...
+std::pair<Expr, Expr> long_div_mod_round_to_zero(const Expr &num, const Expr &den,
+                                                 const uint64_t *max_abs) {
+    debug(1) << "Using long div: (num: " << num << "); (den: " << den << ")\n";
+    internal_assert(num.type() == den.type());
+    Expr abs_num = (num.type().is_int()) ? abs(num) : num;
+    Expr abs_den = (den.type().is_int()) ? abs(den) : den;
+    std::pair<Expr, Expr> qr = unsigned_long_div_mod_round_to_zero(abs_num, abs_den, max_abs);
+    Expr q = qr.first;
+    Expr r = qr.second;
+    // Correct the signs for quotient and remainder for signed integer division.
```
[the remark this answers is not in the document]
aankit-ca: long_division produces round-to-zero? The only thing is that handling of the case of den == 0;
We do den = den + 1 in lower_euclidean_division if den == 0 which we don't need to do when using long_div.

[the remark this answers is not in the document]
dsharletg: The long division itself is unsigned, so it doesn't matter. But this code produces round-to-zero behavior. Basically, the chain of things that gets produced is:

    lower_euclidean_div(long_div_mod_round_to_zero(...)).

Some of the work here is redundant. It's minor, let's just forget it for now.

dsharletg:
It looks like this is failing on the buildbots: https://buildbot.halide-lang.org/master/#/builders/52/builds/29/steps/21/logs/stdio

It looks like test/correctness/print.cpp does some division...

aankit-ca:
Yeah a mod test was failing (where we didn't even generate long division: x % 2). Adding a call to simplify resolved the issue. I think the buildbots should be green now.

PR #5231, opened 2020-09-01
Teach LICM to not view strict_float as work

### LINKED ISSUE #5230: strict_float causes LICM in strange cases
I randomly noticed this snippet in some IR I was trawling through:

```
  let t39 = (float32)strict_float(0.000000f)
  for (sum_intm.s0.y, 0, 4) {
   let t40 = (f.extent.2*sum_intm.s0.y)*4
   for (sum_intm.s0.x, 0, 4) {
    let t41 = (f.extent.2*sum_intm.s0.x) + t40
    for (sum_intm.s0.z, 0, f.extent.2) {
     sum_intm[sum_intm.s0.z + t41] = t39
    }
   }
  }
```

It seems like we LICMed `strict_float(0.0f)`, when I don't think we would have done this if it were just 0.0f.

abadams:
Fixes #5230


src/LICM.cpp
```diff
@@ -84,6 +84,12 @@ class LiftLoopInvariants : public IRMutator {
                 return false;
             }
         }
+        if (const Call *call = e.as<Call>()) {
+            if (call->is_intrinsic(Call::strict_float)) {
```
steven-johnson, line 88: Are there other intrinsics that also fit the bill here? eg, likely(), size_of_halide_buffer_t()?

abadams: Done. Some of them don't exist at this stage of lowering, but if LICM is ever used earlier it would be good to handle things like likely.

steven-johnson:
Breakage is unrelated, OK to land

PR #5234, opened 2020-09-01
Clean up scatter/gather hvx intrinsics

abadams:
I was very surprised to see scatter/gather intrinsics, because that
would be a substantial IR design change - currently we don't use
intrinsics for anything that mixes vector lanes (i.e. intrinsics should
vectorize trivially, and we should be able to slice up wide vectors into
narrow vectors trivially when using intrinsics).

These are only used in the HVX backend though, so I guess within that
restricted scope it's reasonable, because they're handled specially. I
prefixed them with hvx_ to make this clearer and modified some code to
avoid string comparisons.


pranavb-ca:
@aankit-ca  - FYI.

src/IR.h
```diff
@@ -509,11 +509,14 @@ struct Call : public ExprNode<Call> {
         div_round_to_zero,
         dynamic_shuffle,
         extract_mask_element,
-        gather,
         glsl_texture_load,
         glsl_texture_store,
         glsl_varying,
         gpu_thread_barrier,
+        hvx_gather,
```
dsharletg, line 516: Maybe this shouldn't be in the intrinsics list at all? Most of the HVX specific intrinsics/externs are just strings used in the backend.

abadams: Perhaps not - they could just be const char *hvx_gather = "hvx_gather" somewhere. I didn't want to mess with things that much though - just trying to do a minimal drive-by cleanup in between meetings today

steven-johnson:
There's an hvx128 failure here: https://buildbot.halide-lang.org/master/#/builders/27/builds/29/steps/28/logs/stdio

(Not sure if it's related.)

steven-johnson:
hvx64 failure was definitely unrelated

PR #5235, opened 2020-09-01
Sketch of statically-typed Exprs

steven-johnson:
This is a a rough sketch of an idea I have been pondering: could we usefully add more (optional) static typing to the Halide C++ IR?

This PR just adds a wrapper around `Expr` to allow for static typing (a la `Buffer<>`). The idea here is that an `ExprT<type>` is an Expr that can only a value of the given type (or be undefined).

The main goal here is to allow for easier readability in complex pipelines (eg if an Expr is expected to be of type uint8, this makes it more plain when reading the source code, and failures would happen at compile time, not runtime).

I'm not entirely sure if I like this or not, so I didn't expend a lot of energy trying to complete it -- there may be holes in the implementation.

In the name of science, I modified `cast<T>()` to return `ExprT<T>`, and `Var` to be convertible to `ExprT<int>` (rather than plain `Expr`).

If this seems remotely desirable, perhaps it could lead to similar wrappers for Func, ImageParam, etc.

Thoughts welcome.

### DIFF
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -250,6 +250,9 @@ struct StringImm : public ExprNode<StringImm> {
 
 }  // namespace Internal
 
+template<typename T>
+struct ExprT;
+
 /** A fragment of Halide syntax. It's implemented as reference-counted
  * handle to a concrete expression node, but it's immutable, so you
  * can treat it as a value type. */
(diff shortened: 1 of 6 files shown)


steven-johnson:
I'm not going to mark this as 'ready for review', since it's really just a design experiment at this point, but would like to get opinions from people about the general concept.

PR #5236, opened 2020-09-02
Add support for Func::async in Python bindings

dragly:
The Python bindings were missing a wrapper for `Func::async`. This
change adds a wrapper called `Func.async_` to avoid clashes with the
Python `async` keyword.

### DIFF
--- a/python_bindings/readme.md
+++ b/python_bindings/readme.md
@@ -32,6 +32,7 @@ with some differences where the C++ idiom is either inappropriate or impossible:
 - `Buffer::for_each_value()` is hard to implement well in Python; it's omitted
   entirely for now.
 - `Func::in` becomes `Func.in_` because `in` is a Python keyword.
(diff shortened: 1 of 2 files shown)


dragly:
I am not sure if there is a reason this was missing, but assumed it had just been forgotten since it was not mentioned in python_bindings/readme.md. Or perhaps there are some issues with running things asynchronously in Python?

In any case, I tested it in a small application and it seems to work fine. It generates code containing a `fork` and runs without issues even in Python.

steven-johnson:
Thanks, this is just an oversight on our part, I don't know of any reason that `.async()` shouldn't work just as well via Python.

steven-johnson:
LGTM pending green buildbots at https://github.com/halide/Halide/pull/5238

steven-johnson:
Failures unrelated. Thanks for the fix!

steven-johnson (approved):

PR #5243, opened 2020-09-03
Add code to enable LLVM's internal pass-timing mode

steven-johnson:
By defining the env var HL_PRINT_LLVM_PASS_TIMING=1, we enable LLVM's internal code that dumps timing info for each optimization pass to stderr. This is entirely for internal Halide dev work and not intended for use by any Halide user.


abadams:
There's a comment in LLVM_Output.cpp indicating that the existing way to do this is HL_LLVM_ARGS=-time-passes

steven-johnson:
> There's a comment in LLVM_Output.cpp indicating that the existing way to do this is HL_LLVM_ARGS=-time-passes

Well dang. Nevermind. :-)

PR #5252, opened 2020-09-08
Add check of types for ARM simd_op_check.

dsharletg:
In some cases the operation could be executed with a wider type on arm64 and the test would not catch this.


steven-johnson (approved):

steven-johnson:
Failure is unrelated

PR #5253, opened 2020-09-08
CMake changes to land ahead of releases

alexreinking:
I would still love to get #5228 merged before the CppCon talk so that people can use the autoschedulers, but these changes at least will let us update the buildbots now.

This PR consolidates all the CMake packaging logic into `packaging/CMakeLists.txt`. It had been spread around a bit, but this was getting harder to justify.

It also makes the packaging scripts which invoke the build multiple times suitable for buildbot use.


packaging/Unix.cmake
```diff
@@ -4,8 +4,8 @@ set(CPACK_COMPONENTS_HALIDE_RUNTIME Halide_Runtime)
 set(CPACK_COMPONENTS_HALIDE_DEVELOPMENT Halide_Development)
 
 set(CPACK_INSTALL_CMAKE_PROJECTS
-    static-Debug Halide ALL /
-    shared-Debug Halide ALL /
+#    static-Debug Halide ALL /
```
steven-johnson, line 7: I presume this is intentional, but a comment needs to be added saying why they are disabled

steven-johnson (approved):
LGTM with one nit

steven-johnson:
> I would still love to get #5228 merged before the CppCon talk

What's remaining to be done? (Last I looked it was still marked 'draft')

alexreinking:
> What's remaining to be done? (Last I looked it was still marked 'draft')

I've got it working for CMake, but haven't touched the Makefiles.

tools/package-unix.zsh
```diff
@@ -5,13 +5,15 @@ DIR="$(cd "$(dirname "${(%):-%N}")"/.. >/dev/null 2>&1 && pwd)"
 [ -z "$LLVM_DIR" ] && echo "Must set specific LLVM_DIR for packaging" && exit
```
steven-johnson, line 5: ...It's really weird to have both `package-unix.zsh` and `package-unix.sh`.

steven-johnson: (to be more specific: it would be vastly preferable to have a single .sh that will work under both zsh and bash.)

packaging/Unix.cmake
alexreinking: Done

steven-johnson:
Suggestion: also change `project(Halide VERSION 10.0.0)` -> `12.0.0`, since our default trunk should track default LLVM trunk

alexreinking:
> 
> 
> Suggestion: also change `project(Halide VERSION 10.0.0)` -> `12.0.0`, since our default trunk should track default LLVM trunk

Done

tools/package-unix.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. >/dev/null 2>&1 && pwd)"
+
+[ -z "$LLVM_DIR" ] && echo "Must set specific LLVM_DIR for packaging" && exit
+[ -z "$Clang_DIR" ] && echo "Must set specific Clang_DIR for packaging" && exit
+
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release \
```
steven-johnson, line 8: Another style it: IMHO this would be much easier to read if it was one-arg-per-line, preferably sorted, e.g.

```
cmake \
  -B "$DIR/build/static-Release" \
  -DBUILD_SHARED_LIBS=NO \
  -DClang_DIR="$Clang_DIR" \
  -DCMAKE_BUILD_TYPE=Release \
  -DHalide_BUNDLE_LLVM=YES \
  -DLLVM_DIR="$LLVM_DIR" \
  -DWITH_APPS=NO \
  -DWITH_DOCS=YES \
  -DWITH_PYTHON_BINDINGS=NO \
  -DWITH_TESTS=NO \
  -DWITH_TUTORIALS=NO \
  -DWITH_UTILS=NO \
  -G Ninja \
  -S "$DIR"
```

As it is now, trying to compare these two requires sub-scanning every line.

tools/package-unix.zsh
steven-johnson: This works on my mac for both bash and zsh:

`DIR=$(cd -P -- "$(dirname -- "$0")" && printf '%s\n' "$(pwd -P)/")`

tools/package-unix.sh
alexreinking: Gonna wait for the build bots to finish up, then will just drive-by that since the buildbots don't run it.

steven-johnson (commented):
Yet another suggestion for the packaging scripts: it's probably preferable to have the buildbots use these scripts directly (rather than replicate the logic) so that there's a single point of truth; however, these need to be a bit more configurable for that to work well. In particular, being able to explicitly specify the source and build dirs (rather than inferring them relative to the script) is really gonna be necessary.

tools/package-windows.bat
```diff
@@ -33,6 +35,8 @@ cmake -G "Visual Studio 16 2019" -Thost=x64 -A x64 ^
       -DCMAKE_TOOLCHAIN_FILE=%VCPKG_ROOT%/scripts/buildsystems/vcpkg.cmake ^
       -DWITH_TESTS=NO -DWITH_APPS=NO -DWITH_TUTORIALS=NO ^
       -DWITH_DOCS=YES -DWITH_UTILS=NO -DWITH_PYTHON_BINDINGS=NO ^
+      "-DCMAKE_INSTALL_BINDIR=bin/$<CONFIG>" "-DCMAKE_INSTALL_LIBDIR=lib/$<CONFIG>" ^
+      "-DHALIDE_INSTALL_CMAKEDIR=lib" ^
       -S . -B build/static
 
 if %errorlevel% neq 0 goto return
```
steven-johnson, line 42: `REM LLVM Debug + Halide exceeds the COFF 4GB limit!!`

as you know, I'm fine eliding the debug builds entirely, but buildbot now gets around this by using `/bigobj` for all Windows Debug LLVM builds. Can we just do that here too?

alexreinking: No, because this is an issue with the static library itself, not with any individual object file.

alexreinking:
> Yet another suggestion for the packaging scripts: it's probably preferable to have the buildbots use these scripts directly (rather than replicate the logic) so that there's a single point of truth; however, these need to be a bit more configurable for that to work well. In particular, being able to explicitly specify the source and build dirs (rather than inferring them relative to the script) is really gonna be necessary.

The `Unix.cmake` file needs to know the directory names, too. Maybe we can make the scripts write out the correct file?

tools/package-unix.zsh
alexreinking: The `#!/bin/bash` up top renders that moot. But I guess Apple hasn't deleted bash from their OS _yet_. We can just drop the zsh script.

steven-johnson: In that case, we should probably nuke Unix.cmake too and just toll it into the unix shell script. (Sorry for all the change requests after approval...)

steven-johnson:
The failure for x86-64-linux-testbranch-trunk-cmake is unrelated and should not block landing

alexreinking:
Sounds good.  Is there anything else you need in the package scripts?

steven-johnson:
> Sounds good. Is there anything else you need in the package scripts?

The scripts look good for now. Ready to land once the buildbots get to 18 run.

steven-johnson:
ok, I think this has enough test coverage to land.

PR #5257, opened 2020-09-10
Allow overriding the Halide version via cmake flag

steven-johnson:
This is intended as a temporary trick; in the long run, each version will have its own release branch. This allows us to temporarily try out different 'releases' from the Halide master branch.


alexreinking:
Let me open an issue to track reverting this.

alexreinking (approved):

PR #5275, opened 2020-09-16
Simplify Call::if_then_else

steven-johnson:
According to @abadams, Call::if_then_else promises to evaluate exactly one (but not both) of the true and false exprs, so we should simplify cases in which the condition is provably true or false.

### DIFF
--- a/src/Simplify_Call.cpp
+++ b/src/Simplify_Call.cpp
@@ -506,6 +506,38 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
         } else {
             return Call::make(op->type, op->name, {arg}, op->call_type);
         }
+    } else if (op->is_intrinsic(Call::if_then_else)) {
+        // Note that this call promises to evaluate exactly one of the conditions,
+        // so this optimization should be safe.
+
+        internal_assert(op->args.size() == 3);
+        Expr cond_value = mutate(op->args[0], bounds);
+        Expr true_value = mutate(op->args[1], bounds);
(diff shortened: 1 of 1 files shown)


steven-johnson:
From local testing, it appears that this breaks `correctness_skip_stages_memoize`; investigating

src/Simplify_Call.cpp
```diff
@@ -506,6 +506,38 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
+        Expr cond = cond_value;
+        if (const Call *c = cond.as<Call>()) {
+            if (c->is_intrinsic(Call::likely) ||
+                op->is_intrinsic(Call::likely_if_innermost)) {
+                cond = c->args[0];
+            }
+        }
+
+        if (is_one(cond)) {
+            return true_value;
```
abadams, line 528: I propose not mutating true_value and false_value above, and instead returning mutate(true_value, bounds), so that the bounds can be tracked.

steven-johnson: Will do, but on that note: is the proper usage of the bounds parameter adequately documented for the Simplifier code? It wasn't immediately obvious to me when I should propagate bounds and when I should not.

steven-johnson: PTAL

abadams: bounds is just an out-param passed by pointer, so you should only pass the bounds pointer directly into the mutation of that Expr if you're going to return that mutated Expr.

abadams (approved):

PR #5276, opened 2020-09-16
Missing new line causes OpenCL programs to fail to compile when targetting cl_half/cl_doubles

mirsadm:


steven-johnson:
These changes are unrelated, and should really be offered as two separate PRs (unless there is a compelling reason they must be combined).

mirsadm:
Sorry that is my mistake! I intended to only have that single change in this pull request. The other commit is from a previous pull request which I've rebased to master. I'll remove the other changes.

PR #5277, opened 2020-09-16
Missing new line causes OpenCL programs to fail to compile when targetting cl_half/cl_doubles

mirsadm:
clBuildProgram returns error due to missing \n.


steven-johnson (approved):
Thanks for the fix!

PR #5280, opened 2020-09-16
Ensure that inline_all_trivial_functions() isn't dead-stripped

steven-johnson:
Can break Li2018 in some build environments.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -2,6 +2,7 @@
 #include "AutoScheduleUtils.h"
 #include "Debug.h"
 #include "Error.h"
+#include "Function.h"
 #include "Introspection.h"
 #include <atomic>
 #include <chrono>
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -574,7 +574,7 @@ void load_plugin(const std::string &lib_name) {
     // and this is only called when load_plugin is called anyway.
     // (If there is a better, portable way to ensure this function
     // isn't dead-stripped, then let's do that instead.)
-    (void)Halide::inline_all_trivial_functions({}, {}, {});
+    (void)Halide::Internal::inline_all_trivial_functions({}, {}, {});
 
 #ifdef _WIN32
     std::string lib_path = lib_name;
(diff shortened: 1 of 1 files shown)


src/Util.cpp
```diff
@@ -563,6 +563,18 @@ int get_llvm_version() {
...
+    // isn't used by libHalide, but it is used by the Li2018
+    // autoscheduler plugin, so we need it to be present in the final
+    // libHalide. Unfortunately, build environments that static-link
+    // generators might deadstrip this function, causing Li2018
+    // to fail at load time. This useless call ensures that the function
+    // is kept around; calling it with empty arguments is close to free
+    // and this is only called when load_plugin is called anyway.
+    // (If there is a better, portable way to ensure this function
+    // isn't dead-stripped, then let's do that instead.)
+    (void) Halide::Internal::inline_all_trivial_functions({}, {}, {});
```
abadams, line 576: Is it enough to take the address: internal_assert(&Halide::Internal::inline_all_trivial_functions);

steven-johnson: Not in the configuration I'm using.

alexreinking:
I'm confused. I thought we agreed in #5228 that statically linking generators wasn't a supported scenario. We had a pretty long discussion about why this isn't portable and @abadams seemed to think it wasn't something anyone was doing anyway?

This is _definitely_ a unix-only setup. On Windows, when you statically link a library, it is _gone_. There is no dynamic lookup in the library loader there, which is needed for this sort of thing to work with loadable modules. Otherwise you have to use a shared library (on all platforms).

abadams:
I think statically linking to libHalide is a terrible idea, but I guess I should have realized that Google, with its penchant for statically linking the world, is doing it.

Fortunately in that context it's a single system where Steven has full control over the build rules, so it can be made to work.

alexreinking:
> Fortunately in that context it's a single system where Steven has full control over the build rules, so it can be made to work.

Then why not link with `-Wl,--whole-archive ... -Wl,--no-whole-archive` and avoid this PR?

alexreinking:
> Then why not link with `-Wl,--whole-archive ... -Wl,--no-whole-archive` and avoid this PR?

From offline discussion: one build post-processes linked binaries (the generator executables) to strip unused symbols. Not sure what a good workaround would be. Maybe some hack via `-include`?

steven-johnson:
Dropping, I'll figure out another way.

PR #5281, opened 2020-09-17
Do something like sqrt-free Cholesky for BGU

abadams:
This produces fewer update definitions and less total math in the solve
step, saving some code size. No significant impact on performance for
this app, because the solve step is done at low res, but it's
theoretically more satisfying, and it's nice to have the symmetric
version of the matrix solve available as a reference.

@jiawen


apps/bgu/bgu_generator.cpp
```diff
@@ -126,6 +127,86 @@ Matrix<M, N> solve(Matrix<M, M> A, Matrix<M, N> b, Func compute, Var at, bool sk
...
+
+    // Put the input matrices in a Func to do sqrt-free Cholesky.
+    Var vi, vj;
+    Func f;
+    f(x, y, z, vi, vj) = undef<float>();
+    for (int i = 0; i < M; i++) {
+        for (int j = 0; j < M; j++) {
+            f(x, y, z, i, j) = A(i, j);
+        }
+        for (int j = 0; j < N; j++) {
```
jiawen, line 144: Packing the RHS b into the same func makes this hard to read later.

apps/bgu/bgu_generator.cpp
```diff
@@ -126,6 +127,86 @@ Matrix<M, N> solve(Matrix<M, M> A, Matrix<M, N> b, Func compute, Var at, bool sk
...
+            f(x, y, z, i, j + M) = b(i, j);
+        }
+    }
+
+    for (int j = 0; j < M; j++) {
+        // Normalize the jth column starting at the jth row, storing
+        // the normalization factor on the diagonal. The unnormalized
```
jiawen, line 151: Because A(i, j) is symmetric, the unnormalized version stays in the jth row.

apps/bgu/bgu_generator.cpp
```diff
@@ -126,6 +127,86 @@ Matrix<M, N> solve(Matrix<M, M> A, Matrix<M, N> b, Func compute, Var at, bool sk
     return b;
 };
 
+// Solve Ax = b at each x, y, z exploiting the fact that A is
+// symmetric. Compute the result at the given Func and Var.
+template<int M, int N>
```
jiawen, line 132: Maybe give a reference to how square root free Cholesky works: https://users.wpi.edu/~walker/MA514/HANDOUTS/cholesky.pdf ?

apps/bgu/bgu_generator.cpp
```diff
@@ -126,6 +127,86 @@ Matrix<M, N> solve(Matrix<M, M> A, Matrix<M, N> b, Func compute, Var at, bool sk
...
+        }
+        for (int j = 0; j < N; j++) {
+            f(x, y, z, i, j + M) = b(i, j);
+        }
+    }
+
+    for (int j = 0; j < M; j++) {
+        // Normalize the jth column starting at the jth row, storing
+        // the normalization factor on the diagonal. The unnormalized
+        // version stays in the jth row.
```
jiawen, line 152: Is it faster / cheaper to store everything in place in Halide? It's more readable to have separate L and D Funcs.

apps/bgu/bgu_generator.cpp
```diff
@@ -126,6 +127,86 @@ Matrix<M, N> solve(Matrix<M, M> A, Matrix<M, N> b, Func compute, Var at, bool sk
...
+                    f(x, y, z, i, k) -= f(x, y, z, k, j) * f(x, y, z, j, i);
+                }
+            }
+        }
+    }
+
+    // Back substitute
+    Matrix<M, N> result;
+    for (int k = 0; k < N; k++) {
+        for (int j = 0; j < M; j++) {
```
jiawen, line 176: Explain each step?

LDL^T x = b

This is the forward substitution step that first solves for z such that Lz = b.

The first loop just does the forward substitution step without scaling.
The next for loop does the scaling by D^-1.

apps/bgu/bgu_generator.cpp
jiawen, line 151: Clarify that you're storing 1/D instead of the usual D.

apps/bgu/bgu_generator.cpp
```diff
@@ -126,6 +127,86 @@ Matrix<M, N> solve(Matrix<M, M> A, Matrix<M, N> b, Func compute, Var at, bool sk
...
+            for (int i = 0; i < j; i++) {
+                f(x, y, z, j, M + k) -= f(x, y, z, j, i) * f(x, y, z, i, M + k);
+            }
+        }
+
+        for (int j = 0; j < M; j++) {
+            f(x, y, z, j, M + k) *= f(x, y, z, j, j);
+        }
+
+        for (int j = M - 1; j >= 0; j--) {
```
jiawen, line 186: Now back substitute: find x such that L^T x = z.

Now no need to scale because L^T has a unit diagonal.

apps/bgu/bgu_generator.cpp
abadams: Oh wow, I would have loved that reference when I was writing this. I just made it up as I went along by messing with regular Cholesky.

apps/bgu/bgu_generator.cpp
abadams: Yeah keeping it all in update definitions in the one Func makes the mutual recursion legal.

abadams:
Finally addressed review comments. PTAAL.

jiawen (approved):

PR #5282, opened 2020-09-18
Require C++17 for Halide.

### LINKED ISSUE #4216: When should we consider moving Halide to C++17?
Currently, Halide assumes/requires C++11. When should we consider moving our baseline to C++17? 

(FYI: Google recently moved to allow use of C++17 internally, so we'd be fine with this; I don't know how many other downstream users might be impacted by such a change. Would be nice to start a discussion.)

steven-johnson:
As of Halide 13, we require C++17 to build **or** use Halide. (We previously claimed to require C++11, although Make builds typically ended up using C++14 due to `llvm-config --cxxflags` specifying it.)

Fixes #4216


steven-johnson:
The Linux failures appear to be a latent, pre-existing bug that compiler changes have revealed. Stay tuned...

steven-johnson:
No rush to land this (we probably want to wait until after the `release/11.x` branch is created) but marking 'ready for review' so that people can review/comment at their leisure.

apps/bgu/CMakeLists.txt
```diff
@@ -4,10 +4,13 @@ project(bgu)
 enable_testing()
 
 # Set up language settings
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
 set(CMAKE_CXX_EXTENSIONS NO)
 
+# MSVC won't set __cplusplus correctly unless this flag is also set
```
abadams, line 11: Is there no way to hide this idiocy from our users?

steven-johnson: We reported this to the CMake folks and they insisted they couldn't change it without a 'policy' flag to enable it, because existing (broken) code might rely on this (insanely broken) behavior, so they aren't gonna change it anytime soon.

The alternative that most people seem to end up using is to combine sniffing of `__cplusplus` with sniffing of `_MSC_LANG`, which would work I guess, but is just so f***ing wrong that I didn't want to do it, just on principle. Maybe that's a more pragmatic approach than requiring the build files to insert extra things to work around bad decisions by both MSVC and CMake.

abadams: Can we move the setting onto our side of the fence somehow, instead of in the user CMakeLists?

steven-johnson: Well, yes and no.

We currently don't *need* to check __cplusplus at all (there's a reality check in Codegen_LLVM.cpp but it is just as a backstop).

If we only used C++17 features internal to libHalide, then we wouldn't need this.

If we eventually use a C++17 feature in headers exposed to user code (e.g. in templates), though, then we'd definitely have this problem (ie the user code's build system would have to set this flag).

I guess the better answer is to bite the bullet and accept that checks of that sort will have to check both __cplusplus and _MSC_LANG for the foreseeable future, which is just a sorry state of affairs.

steven-johnson: (fix pushed)

abadams (approved):

steven-johnson:
Failures are unrelated, but I want this one 100% clean before landing.

tutorial/lesson_01_basics.cpp
```diff
@@ -3,11 +3,11 @@
 // This lesson demonstrates basic usage of Halide as a JIT compiler for imaging.
 
 // On linux, you can compile and run it like so:
-// g++ lesson_01*.cpp -g -I <path/to/Halide.h> -L <path/to/libHalide.so> -lHalide -lpthread -ldl -o lesson_01 -std=c++11
+// g++ lesson_01*.cpp -g -I <path/to/Halide.h> -L <path/to/libHalide.so> -lHalide -lpthread -ldl -o lesson_01 -std=c++17
```
BachiLi, line 6: Since "using" Halide doesn't require c++17 -- do we need to change the compile flags in the tutorials? I don't have strong opinions, just asking.

steven-johnson: See the top comment above. Probably should keep some/all at C++11 (I just changed everything in a frenzy before thinking about it), but didn't go back and undo those changes yet to see what folks thought -- in particular, are we ready to declare it requiring C++17 for usage too? (Maybe not, but it would certainly be fine at Google.)

alexreinking:
> NOTE: do we want to consider requiring C++17 to _use_ Halide as well? Is C++17 acceptance widespread enough (esp. among known heavy Halide users) to move in that direction?

For "modern" users, I think this is fine. MSVC 2019 supports C++17 by default. Clang and XCode have since forever. GCC is (oddly) the outlier here, having only settled the standard library ABI in 9.3, but the default Ubuntu LTS compiler is at least that new.

I can't speak for our corporate users, of course.

README.md
```diff
@@ -16,6 +16,9 @@ to an object file, or JIT-compile it and run it in the same process. Halide also
 provides a Python binding that provides full support for writing Halide embedded
 in Python without C++.
 
+(Note that *using* Halide requires C++11 or later, but *building* Halide
+requires C++17.)
+
```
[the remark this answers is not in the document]
steven-johnson: See my top comment re: seeking opinions.

apps/HelloAndroid/build.gradle
```diff
@@ -1,7 +1,7 @@
 import org.apache.tools.ant.taskdefs.condition.Os
 
 // Avoid conflicts with Bazel on case-insensitive filesystems
-buildDir = 'gradle_build' 
+buildDir = 'gradle_build'
```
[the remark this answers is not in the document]
steven-johnson: I would be stunned if it did, but dealing with that is a separate issue.

src/Generator.cpp
```diff
@@ -757,7 +757,7 @@ std::string halide_type_to_c_type(const Type &t) {
 
 int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
     const char kUsage[] =
-        "gengen\n"
+        "gengen \n"
```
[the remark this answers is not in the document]
steven-johnson: Ha, ok :-)

src/Generator.h
```diff
@@ -3107,7 +3107,11 @@ class GeneratorBase : public NamesInterface, public GeneratorContext {
         user_assert(sizeof...(args) == pi.inputs().size())
             << "Expected exactly " << pi.inputs().size()
             << " inputs but got " << sizeof...(args) << "\n";
-        set_inputs_vector(build_inputs(std::forward_as_tuple<const Args &...>(args...), make_index_sequence<sizeof...(Args)>{}));
+        // Deliberately avoiding make_index_sequence and friends here so that we don't have to provide
+        // our own implementations for clients that are using C++11 (which didn't provide them).
+        std::vector<std::vector<StubInput>> inputs(sizeof...(args));
+        build_inputs(inputs, 0, args...);
+        set_inputs_vector(inputs);
```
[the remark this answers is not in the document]
steven-johnson: This file is included by client code, which (as mentioned above) shouldn't require more than C++11.

src/Type.h
```diff
@@ -73,9 +71,7 @@ struct halide_cplusplus_type_name {
 /** A structure to represent the fully scoped name of a C++ composite
  * type for use in generating function signatures that use that type.
  *
- * This is intended to be a constexpr usable type, but we don't depend
- * on C++11 yet. In C++14, it is possible this will be replaced with
- * introspection/reflection facilities.
+ * This is intended to be a constexpr usable type.
```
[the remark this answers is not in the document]
steven-johnson: This file is included by client code, which (as mentioned above) shouldn't require more than C++11.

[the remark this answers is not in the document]
steven-johnson: Uh, I guess so. The real change here was to delete the ultra-stale comment ("We don't depend on C++11 yet").

src/Type.h
```diff
@@ -18,9 +18,7 @@
  * time type checking for both Halide generated functions and calls
  * from Halide to external functions.
  *
- * These are intended to be constexpr producable, but we don't depend
- * on C++11 yet. In C++14, it is possible these will be replaced with
- * introspection/reflection facilities.
+ * These are intended to be constexpr producable.
```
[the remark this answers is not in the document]
steven-johnson: See above.

steven-johnson:
To be a little more clear to reviewers about C++11 restrictions: basically, *anything* that goes into Halide.h still can't require C++14/17; since Halide.h tends to include a fair amount of theoretically-internal-only stuff, this will restrict our usage of C++17 within our internal header files.

src/Generator.h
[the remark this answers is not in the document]
alexreinking: So this means we cannot use C++17 features in any header, then?

alexreinking:
Something to be aware of: GCC 9.3 is the first version to have _non-experimental_ support for C++17. That means you cannot mix C++17 code produced by any version of GCC prior to 9.3 -- all objects / libraries must have been produced by the same compiler version.

For more info, see: https://stackoverflow.com/a/49119902/2137996

abadams:
Perhaps we should restrict the apps (or at least some apps) to c++11 to keep ourselves honest w.r.t. Halide.h

steven-johnson:
OK, so I:
- reverted all the apps, tests, and tutorials to build with C++11 (not C++17)
- reverted cosmetic changes to a few files that got caught in the crossfire (e.g. trailing whitespace)
- fixed a few other typos pointed out.

PTAL.

Note that per @alexreinking's comment, we may want/need to upgrade the GCC version on our bots to at least 9.3. We only use GCC on the linuxbots, and for the x86 ones it's easy (Ubuntu 20 has it preinstalled as default) but for the armbots it might be trickier. I'll investigate.

steven-johnson:
> Note that per @alexreinking's comment, we may want/need to upgrade the GCC version on our bots to at least 9.3. 

See https://github.com/halide/Halide/issues/5645 (

shoaibkamil:
I'm leery of making GCC 9.3+ a requirement for building code that uses Halide (it'd be a different matter if we could build Halide with GCC 9.3+ and client code using a different version of GCC).  Ubuntu 18.04 LTS has quite a large userbase.

More concretely, I'll ask devs here if this could be an issue.  Do we know what the minimum clang version is for clients using Halide with this change?

abadams:
The decision in #4216 was based on gcc 7.1 being adequate. I don't think we should permit anything that requires gcc 9 yet.

steven-johnson:
> The decision in #4216 was based on gcc 7.1 being adequate. I don't think we should permit anything that requires gcc 9 yet.

It's not clear to me just how dire it would be to build C++17 code with GCC 7.x and potentially mix with GCC 9.x (see https://stackoverflow.com/questions/46746878/is-it-safe-to-link-c17-c14-and-c11-objects/49119902#49119902 for a comment from a person who worked on C++17 support in GCC) -- if I read this correctly, the main issue would be if we used C++17 features (yes), compiled with GCC 7.x (yes), and then our library was used by someone using GCC 9.x (potential kablooey).

I supposed we could say "don't do that", and/or "build Halide yourself if you aren't using gcc 7.x".

Or we could back away from moving to C++17 for a while longer.

steven-johnson:
So this is basically ready to go, with the possibly-major-or-not issue of whether GCC7 is a safe and prudent choice. Maybe let's discuss this in the next team meeting to hash out where we want to go?

steven-johnson:
As an update to those following this: we want to do a little more investigation into compiler availability and use (wrt gcc7.x vs gcc9.x) before making a decision; due to work schedules, we're guessing this will delay a decision on this by another couple of weeks or so.

alexreinking:
From our 3/5/21 meeting: we decided to hold off on this change until after Halide 12 is released. Halide 12 will be C++11 everywhere. Halide 13 will be C++17 everywhere (including tests, apps, etc.)

alexreinking:
Should we resurrect this?

steven-johnson:
> Should we resurrect this?

Yep!

README.md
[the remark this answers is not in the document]
alexreinking: This needs to be fixed now that we decided to go C++17 wholesale.

src/Generator.h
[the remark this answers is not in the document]
alexreinking: We decided to go full C++17 so this can be reverted.

src/Generator.h
```diff
@@ -3475,9 +3479,19 @@ class GeneratorBase : public NamesInterface, public GeneratorContext {
...
+
+    template<typename Last>
+    inline void build_inputs(std::vector<std::vector<StubInput>> &inputs, int idx, const Last &last) {
+        inputs[idx] = build_input(idx, last);
+    }
+
+    template<typename First, typename Second, typename... Rest>
+    inline void build_inputs(std::vector<std::vector<StubInput>> &inputs, int idx, const First &first, const Second &second, const Rest &...rest) {
+        build_inputs<First>(inputs, idx, first);
+        build_inputs<Second, Rest...>(inputs, idx + 1, second, rest...);
```
alexreinking, line 3494: Since we're going full C++17, this can be reverted, right?

alexreinking (changes requested):

dependencies/llvm/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO RelWithDebInfo Release MinSizeRel "
 set(CMAKE_MAP_IMPORTED_CONFIG_RELEASE Release MinSizeRel RelWithDebInfo "")
 
 find_package(LLVM ${Halide_REQUIRE_LLVM_VERSION} REQUIRED)
-find_package(Clang REQUIRED CONFIG HINTS "${LLVM_DIR}/../clang")
+find_package(Clang REQUIRED CONFIG HINTS "${LLVM_DIR}/../clang" "${LLVM_DIR}/../lib/cmake/clang")
```
steven-johnson, line 14: @alexreinking I don't remember why this change was made -- is it possible it's just a merge hickey?

src/CodeGen_LLVM.cpp
```diff
@@ -27,12 +27,11 @@
...
-#if !(__cplusplus > 199711L || _MSC_VER >= 1800)
-
-// VS2013 isn't fully C++11 compatible, but it supports enough of what Halide
-// needs for now to be an acceptable minimum for Windows.
-#error "Halide requires C++11 or VS2013+; please upgrade your compiler."
-
+// MSVC won't set __cplusplus correctly unless certain compiler flags are set
+// (and CMake doesn't set those flags for you even if you specify C++17),
+// so we need to check against _MSVC_LANG as well, for completeness.
+#if !(__cplusplus >= 201703L || _MSVC_LANG >= 201703L)
```
steven-johnson, line 33: Question: should we (also) add a similar check in Halide.h? I'm inclined to say yes, just so that downstream users aren't baffled if they get a compile error from trying to use older compilers.

dependencies/llvm/CMakeLists.txt
alexreinking: I have three system wide Ubuntu packages for LLVM installed. Needed this to get it to find the right Clang when I tested the PR.

Notice this is the same thing we do when finding LLD, so it was probably an oversight when I changed that.

alexreinking: i.e. it's a desirable change, just a bit of a rider.

src/CodeGen_LLVM.cpp
alexreinking: I would support that, personally.

steven-johnson:
All green (again). Are we ready to land this?

test/correctness/interpreter.cpp
```diff
@@ -136,6 +136,7 @@ int main(int argc, char **argv) {
```
abadams, line 136: This file doesn't seem related?

abadams:
No objections here.

README_cmake.md
```diff
@@ -558,7 +558,7 @@ immediately after setting the minimum version.
 
 The next three variables set the project-wide C++ standard. The first,
 [`CMAKE_CXX_STANDARD`][cmake_cxx_standard], simply sets the standard version.
-Halide requires at least C++11. The second,
+Halide requires at least c++17. The second,
```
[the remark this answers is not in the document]
alexreinking: @steven-johnson - can you fix this one little thing before merging? No need to wait on buildbots for this, obviously.

alexreinking (approved):

README_cmake.md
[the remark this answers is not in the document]
steven-johnson: done

test/correctness/interpreter.cpp
steven-johnson: ...huh, weird, will revert

abadams:
We have discussed this quite a lot at dev meetings, so I think it's good to just land at this point without waiting for explicit approvals from everyone. It can always be rolled back if someone has a serious complaint.

PR #5283, opened 2020-09-18
Bundle wabt objects into libHalide

alexreinking:
Only affects Halide 11+. No more need to package libwabt.a, just includes its objects into libHalide.a


alexreinking:
This would be another good candidate PR for the `#cmake` label I described in halide/build_bot#92

alexreinking:
We shouldn't unconditionally use `$<TARGET_OBJECTS>`. We should only do that when building static Halide. We should just link normally/privately to wabt when building shared Halide to allow the linker to drop unnecessary symbols.

steven-johnson (approved):
LGTM

alexreinking:
Also, no need to backport since Halide 10 uses LLVM 10 which doesn't support WebAssembly

steven-johnson:
Is this ready to land?

PR #5289, opened 2020-09-21
Set version to 12.0.0.

### LINKED ISSUE #5259: Revert #5257 once older releases land
We're going to be releasing Halide 9 and 10 before bumping the version to 11 to keep pace with the released LLVM versions.

To make life easier on the buildbots, we're landing #5257 temporarily, but it's not a permanent thing.

alexreinking:
Since our next release will be based on LLVM 11.0.0, we should set the master version to 11. Yes, we still support LLVM trunk which is version 12, but we'll just bump our version once that one is officially released.

Fixes #5259

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,5 @@
 cmake_minimum_required(VERSION 3.16)
-
-if (NOT DEFINED Halide_VERSION)
-    set(Halide_VERSION 12.0.0)
-endif ()
-project(Halide VERSION ${Halide_VERSION})
+project(Halide VERSION 12.0.0)
 
 enable_testing()
 


steven-johnson:
...not sure I agree with this. I'd be more inclined to always keep the version of Halide's main branch aligned with the version of LLVM's main branch; when we create `release/11.x` we should edit that version.

alexreinking:
> I'd be more inclined to always keep the version of Halide's main branch aligned with the version of LLVM's main branch

Sure, we can use 12.0.0 instead. The more important change is getting rid of the variable.

steven-johnson:
SGTM

alexreinking:
@steven-johnson -- ready for review. Didn't change the branch name, but the commit message is correct.

steven-johnson (approved):

steven-johnson:
(adding @abadams for discussion)

I approved but then unapproved, because we need to figure out how to handle the buildbots before we do this.

Currently we build Halide's 10, 11, and 12 all from the master branch (via this hack), and produce all packaged versions this way.

What we want is something like:
1) release/10.x branch is built against LLVM10 only, with all Halide_10 packaging coming from here
2) (eventually) release/11.x branch is built against LLVM11 only, yadda
3) Halide main branch is tested against 10/11/12 (to maintain compatibility), but only packaged for Halide_12


The gotchas here:
- If we land this PR, we can do (1) but not (2), because we don't yet have a release/11.x branch. (Any reason we shouldn't go ahead and create one?)
- Buildbot logic will need to be revised some to handle the release-branch builds -- to avoid heavy load we probably want to reduce building on the release branches to something like once/day.
- We probably shouldn't land this change until the buildbots are revised as above (it's a harmless no-op for now unless someone defines Halide_VERSION).

alexreinking:
>     * We probably shouldn't land this change until the buildbots are revised as above (it's a harmless no-op for now unless someone defines Halide_VERSION).

That's perfectly fine. Happy to have the work done and sitting here until the bots are ready.

alexreinking:
Time to merge this!

steven-johnson (approved):

PR #5293, opened 2020-09-22
[d3d12] recycling "frame" resources and removing superfluous sync points

slomp:
In this PR: several performance improvements:
- improved device selection
- removing conservative `wait_unti_*()`
- blocking waits instead of busy-waiting spin-locks
- refactored `TRACELOG` and friends to log elapsed time and report (past a certain threshold)
- compute command execution/lifetime tracking (via monotonically incrementing fence signals)
- recycling expensive API objects (command lists, uniform buffers, descriptor pools, etc)

In an upcoming PR:
- buffer caching (current bottlenecks are in `halide_d3d12compute_device_malloc` and `halide_d3d12compute_device_free`)
- possibly an improved buffer "sub-allocation" system to go along with buffer caching


src/runtime/d3d12compute.cpp
```diff
@@ -102,25 +165,52 @@ struct trace : public Printer<BasicPrinter, sizeof(trace_buf)> {
         }
     }
 };
+
+#define TRACEPRINT(msg) trace() << msg;
+
+#ifdef HALIDE_D3D12_TRACE_TIME
+#define TRACETIME_CHECKPOINT() halide_current_time_ns(user_context)
+//#define TRACETIME_REPORT(t0,t1) TRACEPRINT("Time: " << (t1 - t0) / 1.0e6 << " ms\n")
```
steven-johnson, line 173: Do we need to check in commented-out macros?

src/runtime/d3d12compute.cpp
```diff
@@ -102,25 +165,52 @@ struct trace : public Printer<BasicPrinter, sizeof(trace_buf)> {
...
+#define TRACETIME_REPORT(t0, t1, ...) TRACEPRINT(__VA_ARGS__ << (t1 - t0) / 1000 << " us\n")
+#else
+#define TRACETIME_CHECKPOINT() 0
+#define TRACETIME_REPORT(t0, t1, ...)
+#endif
+
 struct TraceScope {
-    TraceScope() {
+    TraceScope(const char *func) {
```
steven-johnson, line 181: make this ctor `explicit` unless there's a reason not to

src/runtime/d3d12compute.cpp
```diff
@@ -1815,38 +2024,71 @@ WEAK void set_input_buffer(d3d12_binder *binder, d3d12_buffer *input_buffer, uin
...
+    if (current_signal >= signal) {
+        TRACEPRINT("Already synced up!\n");
+        return;
+    }
+
+    HRESULT device_status_before = (*device)->GetDeviceRemovedReason();
 
-    HRESULT result_before = (*device)->GetDeviceRemovedReason();
+    // busy waiting / spin-lock
+    //while (queue_fence->GetCompletedValue() < signal) ;
```
steven-johnson, line 2054: Should this commented-out code still be here?

src/runtime/d3d12compute.cpp
```diff
@@ -3139,6 +3337,11 @@ WEAK int halide_d3d12compute_detach_buffer(void *user_context, struct halide_buf
     d3d12_buffer *dbuffer = peel_buffer(buf);
     unwrap_buffer(buf);
 
+    //uint64_t sig = queue_fence->GetCompletedValue();
```
steven-johnson, line 3340: commented-out code.

steven-johnson (commented):
minor nits

src/runtime/d3d12compute.cpp
```diff
@@ -600,10 +712,67 @@ WEAK d3d12_command_queue *queue = NULL;
...
 WEAK ID3D12RootSignature *rootSignature = NULL;
-
 WEAK d3d12_buffer upload = {};    // staging buffer to transfer data to the device
 WEAK d3d12_buffer readback = {};  // staging buffer to retrieve data from the device
 
+WEAK HANDLE hFenceEvent = NULL;
+
+WEAK d3d12_command_allocator *cmd_allocator_main = NULL;
+
+struct d3d12_frame {
```
shoaibkamil, line 722: Could you add a brief comment here explaining the frame-based approach?

shoaibkamil (approved):
Just needs a comment to describe the overall approach for anyone trying to understand the code later.  Otherwise LGTM!  Thanks for doing this, these are massive improvements relative to the performance beforehand.

steven-johnson:
> these are massive improvements relative to the performance beforehand

Out of curiosity, can you expand on this? (It also might be interesting to add some of that detail in a comment in the code, so that future readers appreciate the importance of that approach)

shoaibkamil:
> > these are massive improvements relative to the performance beforehand
> 
> Out of curiosity, can you expand on this? (It also might be interesting to add some of that detail in a comment in the code, so that future readers appreciate the importance of that approach)

Marcos can probably elaborate more, but previously I believe @abadams observed something like a 100x performance gap between the CUDA and D3D12 backends.  Marcos's improvements have made it so the gap is ~3x, with the difference being attributed to the CUDA backend having a buffer cache.

slomp:
@steven-johnson 
In terms of performance, the following items from the PR description have the greatest impact in performance:
- removing conservative `wait_unti_*()`: we used to sync on every kernel run and buffer copy operation; not anymore.
- recycling expensive API objects: we used to create/destroy "expensive" D3D12 API objects (such as command lists, constant buffers, descriptor pools, etc) all the time on each kernel run; not anymore.

As per @shoaibkamil request, I'll add a description to what I mean by a `d3d12_frame` since I am borrowing the term "frame" from graphics due to the lack of a better term for us in Halide jargon. In our case, every dispatch )kernel run) is a frame in a sense that it requires some API objects to remain alive until t's completed executing on the device, at which point it's safe to recycle them for subsequent kernel dispatches.

I don't have enough information about full Pipelines and Stages in the run-time back-end to possibly group command-lists, descriptor binders, and other expensive API objects together in a more coarse "frame", so I have to treat each kernel dispatch as a "frame" in terms of resource usage and lifetime tracking.

steven-johnson:
(Adding 'backport' tag as this makes D3D12 backend much more usable.)

steven-johnson:
(BTW: please squash/rebase this into a single commit when landing, thanks)

shoaibkamil:
Hmm, didn't run on the buildbots?

steven-johnson:
> Hmm, didn't run on the buildbots?

Buildbots are backed up right now (and the rate of pushes to this PR keeps resetting the requests). They should catch up.

slomp:
I think the buildbot error is unrelated (linux).

steven-johnson:
Yes, error is unrelated, OK to land

Issue #5160, opened 2020-08-05
Naming a Func and an Input<> the same thing generates a user-unfriendly error


dsharletg:
I've run into this a few times lately.

```
class Repro : public Halide::Generator<Repro> {
 public:
  Input<Buffer<uint8_t>> input_{"input", 3};
  Output<Buffer<uint8_t>> output_{"output", 3};

  void generate() {
    Halide::Var x("x"), y("y"), c("c");

    Func input("input");  // Same name as input_
    input(x, y, c) = input_(x, y, c);

    output_(x, y, c) = input(x, y, c);
  }
};
```

This generates an error like:

Internal Error at third_party/halide/halide/src/Simplify_Let.cpp:68 triggered by user code at : Condition failed: !var_info.contains(op->name): Simplify only works on code where every name is unique. Repeated name: input.extent.0
*** SIGABRT received by PID 7218 (TID 7218) on cpu 33 from PID 7218; stack trace: ***
PC: @     0x7f5c7f90d602  (unknown)  raise
    @     0x55e69303ce73       1728  FailureSignalHandler()
    @     0x7f5c7fa8b9a0  748724272  (unknown)
    @     0x55e690462909        112  Halide::Internal::ErrorReport::~ErrorReport()
    @     0x55e690989028        928  Halide::Internal::Simplify::simplify_let<>()
...

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** The reported symptom was an internal error from `Simplify_Let.cpp` ("Repeated name: input.extent.0") when a `Func` and a Generator `Input<Buffer<>>` shared a name. Generator inputs now reserve their name so a later Func gets a distinct one, and the unfixable ordering (Func declared first) produces a clear `user_error`.

[...]

Issue #5292, opened 2020-09-22
An embedded Buffer can 'shadow' a Param with the same name


steven-johnson:
Consider the following snippet:

```
    constexpr int a = 3;
    constexpr int b = 5;

    ImageParam input_a(Int(32), 1, "input");
    Buffer<int> buf_a(1, "input");
    buf_a.fill(a);
    input_a.set(buf_a);

    Buffer<int> input_b(1, "input");
    input_b.fill(b);

    Var x("x");
    Func f("f");
    f(x) = input_a(x) + input_b(x);
    f.compile_jit();

    Buffer<int> result(1);
    f.realize(result);
    if (result(0) != a + b) {
        fprintf(stderr, "Expected to see %d + %d = %d but saw %d\n", a, b, a + b, (int) result(0));
        abort();
    }

    printf("Success!\n");
```

This will fail with `Expected to see 3 + 5 = 8 but saw 6` -- the references to the embedded Buffer are redirected to the ImageParam.

(1) Even when using the same name, how is this redirection happening? i.e. where do we do a lookup-by-name that subverts embedded Buffer?

(2) The Generator infrastructure prevents you from having Params with duplicate names, but it doesn't prevent embedded Buffers from overlapping those names (and the JIT API doesn't do either of these).

PR #5294, opened 2020-09-23
Check for duplicated Parameter/Buffer names in InferArguments (Issue #5292)

steven-johnson:


dsharletg:
Also fixes #5160 ?

steven-johnson:
> Also fixes #5160 ?

Unfortunately -- this only affects things that are visible as 'inputs' to a pipeline (included embedded Buffers); that bug happens for Func that are anywhere inside the pipeline.

steven-johnson:
All green, PTAL

src/InferArguments.cpp
```diff
@@ -32,24 +34,117 @@ class InferArguments : public IRGraphVisitor {
...
+            if (pob.param.is_buffer() && pob.param.buffer().same_as(b)) {
+                // Update this entry to have both the Parameter and Buffer.
+                pob.buffer = b;
+                return true;
+            } else {
+                // A Parameter and Buffer with the same name (but aren't connected) -> illegal
+                return dupe_names_error(name);
+            }
+        } else {
+            internal_error;  // There should be no empty ParamOrBuffers in the map.
```
abadams, line 145: Seems marginally better to make the comment a string argument to internal_error

abadams (approved):

PR #1866, opened 2017-02-22
OpenCL Textures support

bleibig:
Work on this feature started a while ago on the gpu_textures_api_variants branch, but was never fully completed. I am resurrecting this, now being based off of the new halide_buffer_t implementation, which has the necessary type info to allow the call to clCreateImage to always create the right type of image.

In summary, this adds the Target::Textures feature, which when used with Target::OpenCL feature will change the buffers passed into the kernel to be cl_images instead, supporting 1, 2, or 3 dimensions. The standard loads and stores in the kernel are replaced by multidimensional read_image and write_image calls.

The old gpu_textures_api_variants had preliminary support for Metal Textures as well, but it was never completed so I left it out. It can still be added later.

PR #5297, opened 2020-09-24
OpenCL Texture Support

jlaxson:
This PR adds the ability to store Funcs and ImageParams in texture memory for OpenCL and have them accessed via read_image{f,i,ui} in kernels.  This is especially useful in architectures like Snapdragon where normal buffers only use L2 cache, but using textures/images leverages L1 cache - in my workload it's a 2-3x speedup.

Not handled here is non-OpenCL runtimes, and any sort of sampler configuration - textures are treated identically to buffers right now.

To use, a Func or ImageParam can be configured with `.store_in(MemoryType::GPUTexture)`.  InjectHostBufferCopies will detect this and use a separate halide_device_interface_t for managing the clCreateImage allocation and friends.  StorageFlattening will also lower to calls with the image_load_texture and image_store_texture intrinsic rather than strictly buffer offsets, and then CodeGen_OpenCL_Dev will further lower to the appropriate read_image{i,ui,f} or write_image{i,ui_f} calls.

Points for review:
* I added new intrinsics for the texture load because I wasn't totally sure how the existing image_load and glsl_texture_load intrinsics interact.  This code generalizes to 1-3 dimensions whereas the glsl_texture_load codepath is hardwired for three, so I left that alone.
*  I added a MemoryType member to Parameter (and plumbed through to ImageParam and GeneratorInput) as I was scrounging for ways for the user API to convey that an input buffer should be handled through texture memory.  There might be another way to do this?
* There is a test provided that exercises Float and Int 1d, 2d, and 3d textures.
* Handling for buffer .min offsets might not be complete, if someone more versed in the Halide buffer system and OpenCL could point out what should happen here, that would be great.

Thanks to @bleibig in #1866 for the CodeGen and runtime code to follow.


zvookin:
Just did an initial pass over this and it looks like exactly what I had in mind for addressing this problem. So first impression is this is great. Thank you for the PR. I feel like I should do a close line by line read and don't have time to do that now. I would not object to merging this without that as I don't think it has much opportunity to cause problems outside of the new functionality not working as intended. (I.e. my take is low risk of breakage.)

I'd expect that we can go with one set of texture based image load intrinsics, but as covered in the description, it needs a close pass on the details. Moreover, if we did that, it would be better to fold the glsl_ stuff onto the generic names, not the other way around. (Because ideally, this support will be used across multiple backends.)

src/runtime/opencl.cpp
```diff
@@ -1488,3 +1493,386 @@ WEAK halide_device_interface_t opencl_device_interface = {
...
+        } else if (type.bits == 32) {
+            format.image_channel_data_type = CL_FLOAT;
+        } else {
+            halide_assert(user_context, false && "unhandled float bit width for image");
+        }
+    } else {
+        halide_assert(user_context, false && "unhandled data type for image");
+    }
+
+    int last_dim_size = buf->dim[buf->dimensions - 1].extent;
```
steven-johnson, line 1567: `last_dim_size` is unused, which will generate a warning-treated-as-error on some compilers (notable xcode, see https://buildbot.halide-lang.org/master/#/builders/10/builds/41)

src/IR.cpp
```diff
@@ -610,7 +610,9 @@ const char *const intrinsic_op_names[] = {
     "if_then_else",
     "if_then_else_mask",
     "image_load",
+    "image_load_texture",
     "image_store",
+    "image_store_texture",
```
steven-johnson, line 615: These new intrinsics are causing failures when using the C/C++ backend, as it doesn't know how to handle them and just dies. See https://buildbot.halide-lang.org/master/#/builders/35/builds/42

steven-johnson:
Thanks for the CL! For various reasons, we can only run our full buildbot test suite on branches (not forks), so the tests for this PR are being done in https://github.com/halide/Halide/pull/5299 -- when the breakages I mentioned above are addressed here, I'll update that branch and retest.

jlaxson:
Thanks @steven-johnson and @zvookin.  Pushed fixes for those two issues (or at least the symptom of the opencl/cuda one - was overriding target in the test).

If/when you guys are satisfied with this as a minimal start, I have some follow-ons in mind to keep building on it - CUDA, hardware sampling filters, interleaved buffers/multi-channel reads.

While I have you, curious for your thoughts on two design aspects of those followons:

## CUDA
OpenCL device buffer support was done by cloning the existing opencl_device_interface and rewriting to use the clImage functions to create and read/write to the buffer.  I think this worked okay because a memory buffer vs texture image reduce to the same cl_mem and so was reasonably transparent to things like halide_opencl_run.

For CUDA, this is done by still having a memory allocation to manage, (cuMemAlloc), and wrapping it in a cuTextureObject to pass in to the kernel, so there are two things to manage instead of just one.  I can see a few ways to handle it:
1. Separate copy of the cuda_device_interface, buf->device is ptr to a host-side struct with both the actual device pointer and the cuTextureObject.  Lots of code duplication, potential benefit that allocations could be done as cuda arrays which would add the ability to do 3d textures.
2. Modify the existing cuda_device_interface to add storage for the cuTextureObject.  Less code duplication, breaks API/ABI for anyone who's extended the cuda runtime.
3. Add a cuda-specific runtime function like halide_cuda_texture_from_buffer that could be injected before kernel dispatches.  This would prepare the texture object for use in the kernel, and a matching function would free it after.  Doesn't require any changes to the main cuda device interface, and potential benefit that it provides an avenue to convey information about how the kernel intends to use the texture (filtering, normalized values, etc).  But lives as a bit of a one-off.

I think I prefer 3 slightly, but I haven't figured out how OpenCL filtered sampling will play in yet (tough because you have to declare the type at image creation time)

## Intrinsics
Over the course of this first part, have become a bit wiser in how the existing image_load/store intrinsics were used (not very much - perhaps they're relics?). As I get to adding hardware linear sampler support, I think they could be reused in a couple ways:
1. Two pairs of intrinsics - image_load/store would be the current, integer-address functionality I have in image_load_texture.  Then a image_load_sampled intrinsic pair could be that plus information on sampling mode, real-valued coordinates, etc.  They don't inherently have to be separate, but I think there's already enough keying on how many arguments there are to the Call op.
2. Or, extend the Load and Store ops with fields on original coordinates (in addition to flattened as 1d-array accesses) and sampler info.  This would allow the Storage Flattening phase to not care about GPU usage at all, and every codegen backend could have all the information necessary to lower accesses however it wanted - helpful for CUDA where texture reads use the tex2D family of ops, but writes don't change.

Either way, the glsl intrinsics can go away and be replaced by lowering from whichever above option is chosen directly in the GLSL codegen module.

dsharletg:
Regarding CUDA: I'm not sure we need to support images/textures in CUDA at all. In CUDA, it is possible to wrap linear memory with a texture sampler in a very lightweight manner, which enables using the texture cache and interpolation without worrying about how the buffer is allocated and stored. I guess the question is if there is an advantage to the tiled memory layout that textures (might) use?

On Intel GPUs, I did an experiment and found that if you access an image in a relatively coherent manner (e.g. to implement convolution or subtle image warping), tiled memory offered no advantage over linear memory, either theoretically or in practice. There was only an advantage for really sparse/random access, which might not be relevant for most Halide use cases. But maybe CUDA is different?

dsharletg:
But to add to that: this is great and very helpful in OpenCL, where it is harder to just use texture samplers in a lightweight manner on linear memory. However, maybe OpenCL 2.0 makes this better? https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html#cl_khr_image2d_from_buffer

src/Closure.h
```diff
@@ -55,11 +55,14 @@ class Closure : public IRVisitor {
         /** The buffer is written to. */
         bool write;
 
+        /** The buffer is a texture */
+        bool texture;
```
abadams, line 59: This should perhaps be a MemoryType

src/StorageFlattening.cpp
```diff
@@ -245,6 +260,21 @@ class FlattenDimensions : public IRMutator {
             Expr store = Call::make(value.type(), Call::image_store,
                                     args, Call::Intrinsic);
             return Evaluate::make(store);
+        } else if (in_gpu && textures.count(op->name)) {
+            debug(2) << " lower texture store to " << op->name << "\n";
```
abadams, line 264: debug(2) is too noisy for this. Probably can just cut them, or reduce to debug(4) if you want to keep them.

src/runtime/opencl.cpp
```diff
@@ -11,9 +11,8 @@ namespace Runtime {
 namespace Internal {
 namespace OpenCL {
 
-// Define the function pointers for the OpenCL API. OpenCL 1.2
-// currently disabled so we can work on build bots without it.
-//#define HAVE_OPENCL_12
+// Define the function pointers for the OpenCL API.
+#define HAVE_OPENCL_12
```
abadams, line 15: Is it safe for us to require opencl 1.2 now?

src/runtime/opencl.cpp
```diff
@@ -687,16 +687,19 @@ WEAK int halide_opencl_initialize_kernels(void *user_context, void **state_ptr,
         } else {
             debug(user_context) << (void *)program << "\n";
         }
-        (*state)->program = program;
 
+        // halide_print(user_context, "Source: \n");
```
abadams, line 691: Please delete commented-out code, or add a comment explaining why it's commented out

src/runtime/opencl.cpp
```diff
@@ -1488,3 +1493,409 @@ WEAK halide_device_interface_t opencl_device_interface = {
...
+        }
+        if (!from_host && to_host) {
+            int dim = dst->dimensions;
+            size_t offset[] = {0, 0, 0};
+            size_t region[] = {
+                static_cast<size_t>(dst->dim[0].extent),
+                dim >= 2 ? static_cast<size_t>(dst->dim[1].extent) : 1,
+                dim >= 3 ? static_cast<size_t>(dst->dim[2].extent) : 1};
+
+            // int row_pitch = dst->dimensions >= 2 ? dst->dim[1].stride * dst->type.bytes() : 0;
```
abadams, line 1703: Please delete commented-out code throughout this function

abadams (commented):
Thanks! My high-level requests are: 

- I think most places you have an is_texture bool, I'd prefer a MemoryType enum

- You should reuse the existing image_store and image_load instead of making a new one. I believe they are appropriate for this purpose.

The arguments for image_load/store suck. The first arg is a stringimm which is the name of the Func/ImageParam, and the second arg is a Variable called something.buffer. I believe that these should be replaced with just a Variable of type Handle with the name of the Func/ImageParam. However this would be a refactor of the existing code, so I don't think it's necessary or appropriate to do in this PR unless you really feel like it.

jlaxson:
> But to add to that: this is great and very helpful in OpenCL, where it is harder to just use texture samplers in a lightweight manner on linear memory. However, maybe OpenCL 2.0 makes this better? https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html#cl_khr_image2d_from_buffer

It does, although I think some of the structural device interface challenges I mentioned vis a vis cuda remain in that case (have two memory objects to persist/maintain).

src/runtime/opencl.cpp
jlaxson: I'm not sure what was intended before, but there were 1.2-introduced functions (clEnqueueWriteImage and company) defined in mini_cl.h and not guarded by this define.  So I think it de facto already required 1.2 (unless you bypassed halide_default_error)

jlaxson:
> Regarding CUDA: I'm not sure we need to support images/textures in CUDA at all. In CUDA, it is possible to wrap linear memory with a texture sampler in a very lightweight manner, which enables using the texture cache and interpolation without worrying about how the buffer is allocated and stored. I guess the question is if there is an advantage to the tiled memory layout that textures (might) use?

Wrapping linear memory is what I was referring to above.  It looks like the only downside is that you can't do 3D accesses unless you allocate differently, as a cuda array rather than basic allocation.  I think you still need to store the cuTextureObject (regardless of how you allocate) in the buffer_t unless you transiently build one for every kernel call?

> On Intel GPUs, I did an experiment and found that if you access an image in a relatively coherent manner (e.g. to implement convolution or subtle image warping), tiled memory offered no advantage over linear memory, either theoretically or in practice. There was only an advantage for really sparse/random access, which might not be relevant for most Halide use cases. But maybe CUDA is different?

This is interesting and helpful.  Did you include linear sampling in that analysis?  I don't have anything very scientific, but in crude trials it looked like the hardware linear sampler was worth a lot (20% in my workload, though that number would be very specific to this kernel) over four integer image reads and weighting them directly.

src/runtime/opencl.cpp
jlaxson: And thinking about it, if those API functions are not available I think you won't have an issue if you aren't using the GPUTexture code paths

jlaxson:
> Thanks! My high-level requests are:
> 
> * I think most places you have an is_texture bool, I'd prefer a MemoryType enum
> * You should reuse the existing image_store and image_load instead of making a new one. I believe they are appropriate for this purpose.

All these are done

> The arguments for image_load/store suck. The first arg is a stringimm which is the name of the Func/ImageParam, and the second arg is a Variable called something.buffer. I believe that these should be replaced with just a Variable of type Handle with the name of the Func/ImageParam. However this would be a refactor of the existing code, so I don't think it's necessary or appropriate to do in this PR unless you really feel like it.

I agree.  I was thinking of doing this in a followup PR to avoid bloating this one, along with pruning the glsl intrinsics.  Or see above for some thoughts on not using a Call op + intrinsic for these operations at all.

dsharletg:
> This is interesting and helpful. Did you include linear sampling in that analysis? I don't have anything very scientific, but in crude trials it looked like the hardware linear sampler was worth a lot (20% in my workload, though that number would be very specific to this kernel) over four integer image reads and weighting them directly.

Yes, using the hardware linear sampler is great and really hard to beat. What I was referring to is hardware linear interpolation of a linear buffer (a pointer and stride), vs. hardware linear interpolation of a tiled memory buffer (a "texture" or "image", depending on the API). I found little difference between these two (only on older Intel GPUs, but that result matches what you would theoretically expect if you access the memory mostly coherently).

steven-johnson:
opencl.cpp:1559:40: error: comparison of integers of different signs: 'cl_channel_type' (aka 'unsigned int') and 'int' [-Werror,-Wsign-compare]
if (format.image_channel_data_type == -1) {
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~
1 error generated.

steven-johnson:
> opencl.cpp:1559:40: error: comparison of integers of different signs: 'cl_channel_type' (aka 'unsigned int') and 'int' [-Werror,-Wsign-compare]

(BTW: I'm assuming you pushed this fix without building locally, as this *should* have been a warning on all platforms. If you didn't get this warning/error, please let me know so I can figure out why.)

steven-johnson:
(I took the liberty of pushing a simple fix to unblock testing -- feel free to revert and do something different if this isn't appropriate.)

jlaxson:
Thanks @steven-johnson.  No, I didn't get a warning (or error) during build with CMake+Ninja+Apple Clang.

steven-johnson:
> Thanks @steven-johnson. No, I didn't get a warning (or error) during build with CMake+Ninja+Apple Clang.

Hmm, that's weird, because we should always be using the Clang from LLVM when building the runtime files. I'll try reproducing on my mac. (What version of AppleClang/XCode do you have installed?)

jlaxson:
I have `Apple clang version 11.0.0 (clang-1100.0.33.17)` on 10.14.  LLVM is 10.0.1 from homebrew.  The build is indeed using LLVM's clang for runtime:

`/usr/local/opt/llvm/bin/clang-10 -O3 -fno-vectorize -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables -fno-threadsafe-statics -fpic -DCOMPILING_HALIDE_RUNTIME -DBITS_64 -m64 -target le64-unknown-unknown-unknown -emit-llvm -S -MD -MF /Users/jlaxson/Documents/Halide/build/src/runtime/initmod.opencl_64.d -o src/runtime/initmod.opencl_64.ll <root>/src/runtime/opencl.cpp`

I think the CMake build might not set warning flags for runtime files?  I see `set(CXX_WARNING_FLAGS -Wall -Werror -Wno-unused-function -Wcast-qual)` but CXX_WARNING_FLAGS isn't used anywhere.

steven-johnson:
> I think the CMake build might not set warning flags for runtime files?

Yep, that's it: the Makefile builds fail but CMake doesn't. I'll get them in sync.

steven-johnson:
With Metal API Validation enabled (`export METAL_DEVICE_WRAPPER_TYPE=1`), this PR causes `tutorial_lesson_12_using_the_gpu` to fail on OSX (both on the buildbots and on my local Mac). Not sure why yet.

src/InjectHostDevBufferCopies.cpp
```diff
@@ -52,10 +52,12 @@ class FindBufferUsage : public IRVisitor {
     }
 
     void visit(const Call *op) override {
-        if (op->is_intrinsic(Call::image_load)) {
+        if (op->is_intrinsic(Call::image_load) ||
+            op->is_intrinsic(Call::image_load)) {
```
steven-johnson, line 56: ...should this second line be `image_store`?

steven-johnson: ...er, no, I guess not, it's just redundant

src/InjectHostDevBufferCopies.cpp
```diff
@@ -66,6 +68,8 @@ class FindBufferUsage : public IRVisitor {
             if (is_buffer_var(op->args[1])) {
                 devices_touched.insert(current_device_api);
                 devices_writing.insert(current_device_api);
+
```
steven-johnson, line 71: Do not insert extra blank line here

src/InjectHostDevBufferCopies.cpp
```diff
@@ -722,6 +732,7 @@ class InjectBufferCopiesForInputsAndOutputs : public IRMutator {
         void include(const Parameter &p) {
             if (p.defined()) {
                 result.insert(p.name());
+                result_storage[p.name()] = p.memory_type();
```
steven-johnson, line 735: OK, so the cause of the crash for tutorial_lesson_12 is that the `include(Buffer<>)` method also need to fill in entries for `result_storage`, eg

```
        void include(const Buffer<> &b) {
            if (b.defined()) {
                result.insert(b.name());
                result_storage[b.name()] = /* VALUE */;
            }
        }
```

...because as it is now, the code in the `mutate()` method below can try to find an entry that has no value here.

I don't know what the right value is, though; `Buffer<>` doesn't have a MemoryType andI don't know if `Heap` is right. Hmm...

jlaxson: Addressed, I chose MemoryType::Auto, because:

* That's the default I have for Param
* It seems like Buffer<> mainly appears in the AST if you are declaring a Func in the inline/JIT syntax (correct?).  In that case you can always wrap it with ImageParam if you want to be able to override the memory type.

steven-johnson:
Sorry for delay in this, lots of churn on our buildbots lately. The latest test of this is failing on (at least) windows:

https://buildbot.halide-lang.org/master/#/builders/7/builds/2

Unhandled exception: Internal Error at C:\build_bot\worker\x86-64-windows-testbranch-11-cmake\halide-source\src\CodeGen_LLVM.cpp:3259 triggered by user code at : Unknown intrinsic: image_store

jlaxson:
No worries @steven-johnson.  Any hints at how this check in gpu_texture.cpp isn't doing its job?  Works for me locally with HL_JIT_TARGET=host-cuda.
```
    if (!t.has_feature(halide_target_feature_opencl)) {
        printf("[SKIP] No OpenCL target enabled.\n");
        return 0;
    }
```

jlaxson:
Never mind.  Was confused by the naming of build files, it's the generator test.  On it.

jlaxson:
@steven-johnson I addressed the latest round of buildbot fails with an OpenCL version check using the existing compute_compatibility interface.  I'm not sure when the line is crossed to key things on a target feature, but feels  like doing more than bare minimum to support that 10 year old spec is probably not worth it.

steven-johnson:
Still getting a core dump: https://buildbot.halide-lang.org/master/#/builders/74/builds/32/steps/19/logs/stdio

bash: line 1: 30779 Aborted                 (core dumped) /home/halidenightly/build_bot/worker/x86-64-linux-testbranch-11-make/halide-build/bin/host-opencl/generator_aotcpp_gpu_texture
/home/halidenightly/build_bot/worker/x86-64-linux-testbranch-11-make/halide-source/Makefile:1844: recipe for target 'generator_aotcpp_gpu_texture' failed
make: *** [generator_aotcpp_gpu_texture] Error 134


I'll see if I can replicate locally.

steven-johnson:
This is pretty close to all green on the buildbots. Is it ready to land?

steven-johnson:
Only remaining failure is unrelated. Let's review and land.

steven-johnson (approved):

steven-johnson:
Thanks for the changes!

jlaxson:
Thanks @steven-johnson for shepherding

PR #5298, opened 2020-09-28
[d3d12] allocation cache + bugfixes

slomp:
This PR further narrows the performance gap between D3D12 and CUDA; it's almost on par.

Summary of relevant changes:
- critical bugfix on API resource recycling: forgot to reset the descriptor binder state when reusing the object
- rudimentary but effective buffer allocation cache (disabled by default, as it is not feature-complete)

Other changes are cosmetic (refactoring, comments, better trace-logging, etc)


src/runtime/d3d12compute.cpp
```diff
@@ -2617,13 +2640,45 @@ WEAK int halide_d3d12compute_device_malloc(void *user_context, halide_buffer_t *
...
+    TRACEPRINT("d3d12_buffer: " << dbuffer << "\n");
+
+    //wait_until_signaled(dbuffer->signal);
+
+    bool cached = false;
+    if (halide_can_reuse_device_allocations(user_context)) {
+        TRACE_SCOPE("allocation cache");
+        ScopedMutexLock lock(&buffer_pool_lock);
+        for (size_t i = 0; i < MaxBuffersInCache; ++i) {
+            if (buffer_pool[i] != NULL) {
```
abadams, line 2659: Unless I'm missing something, this cache doesn't seem to ever evict anything, so if it fills with small buffers from one part of the pipeline, you won't get any caching at all when you run a second part of the pipeline or another pipeline with larger buffers.

Is it possible to extract the cache logic from cuda.cpp and share that among the backends instead? As it stands we're going to have different logic per GPU backend for allocation caching.

shoaibkamil:
I think this is still incorrect: https://github.com/halide/Halide/blob/4329d9d99a88456e76080cc1082021a407e33359/src/runtime/d3d12compute.cpp#L3270 (and the corresponding one in the `else` branch)

slomp:
> I think this is still incorrect:
> 
> https://github.com/halide/Halide/blob/4329d9d99a88456e76080cc1082021a407e33359/src/runtime/d3d12compute.cpp#L3270
> 
> (and the corresponding one in the `else` branch)

Yeah, that's actually supposed to be `__atomic_load`as I just want to check the value.

src/runtime/d3d12compute.cpp
slomp: > Unless I'm missing something, this cache doesn't seem to ever evict anything, so if it fills with small buffers from one part of the pipeline, you won't get any caching at all when you run a second part of the pipeline or another pipeline with larger buffers.
> 
> Is it possible to extract the cache logic from cuda.cpp and share that among the backends instead? As it stands we're going to have different logic per GPU backend for allocation caching.

Indeed, it's a rudimentary implementation of an allocation cache as there's no LRU/aging/tracking/eviction in place. I made this mostly as a placeholder that just so happened to work relatively well for the cases we have at the moment.

I think we could adopt the one you have in CUDA, but the best thing to do (in d3d12 at least) would be to perform sub-allocations on larger ID3D12Resource heaps, instead of caching the internal buffer wrapping object (d3d12_buffer).

shoaibkamil: Given that there's no eviction here, I think we should, for now, disable the allocation cache and re-enable once we have that.

shoaibkamil:
This looks great, thanks for the quick fixes!  As I said in the reply above, I think we should (for now) disable the allocation cache.  Otherwise, I want to make sure the Windows buildbot runs this under the d3d12compute target.  Locally, everything in `[correctness,tutorial,python]` works for me (for some reason, my local config is unhappy with the cmake setup for apps).

slomp:
> This looks great, thanks for the quick fixes! As I said in the reply above, I think we should (for now) disable the allocation cache. Otherwise, I want to make sure the Windows buildbot runs this under the d3d12compute target. Locally, everything in `[correctness,tutorial,python]` works for me (for some reason, my local config is unhappy with the cmake setup for apps).

Cool, makes sense. I'll refactor the allocation cache to some helper functions
I'll keep them around for performance potential tests, but will not be calling them on master.

shoaibkamil:
Looks like some of the apps are failing on d3d12compute.  Thanks @steven-johnson for whatever you did to tickle the buildbots :)

steven-johnson:
(Pretty please, rebase or squash+merge when landing this PR :-)

shoaibkamil (approved):
Looks good, lets wait til all green

src/runtime/windows_d3d12compute_x86.cpp
```diff
@@ -2,4 +2,12 @@
 // located in "d3d12compute.cpp"; this wrapper is used as a mean to "decorate" the
 // module such that the proper build flags and target triple can be determined based
 // on the prefix (windows_) and suffix (_x86) when building the runtime module.
+
```
steven-johnson, line 5: is there any reason we shouldn't just rename `d3d12compute.cpp` instead of including it? (if so, let's do it as a followup PR)

shoaibkamil: I think the intent is to have `windows_d3d12compute_arm64.cpp` shortly.

src/runtime/d3d12compute.cpp
```diff
@@ -1025,6 +996,54 @@ static void D3D12WaitForPix() {
...
+        return NULL;
+    }
+    char Description[128];
+    for (int i = 0; i < 128; ++i) {
+        Description[i] = desc.Description[i];
+    }
+    Description[127] = '\0';
+    TRACEPRINT("Device selected: " << Description << "\n");
+
+#if 0
```
steven-johnson, line 1014: nit: please don't check in `#if 0` code. Use preprocessor symbols to indicate why it's disabled. e.g.

#define DO_DXGI_OUTPUT 0
#if DO_DXGI_OUTPUT ...

src/runtime/d3d12compute.cpp
```diff
@@ -2541,67 +2525,138 @@ WEAK int halide_d3d12compute_release_context(void *user_context) {
...
 
-    err << "D3D12 Adapter: " << Description << "\n";
+    err << ">>> D3D12 Device Adapter: " << Description << "\n";
+
+    err << ">>>>> End of Halide D3D12 Debug Dump Report <<<<<\n";
 }
 
 using namespace Halide::Runtime::Internal::D3D12Compute;
 
+// NOTE(marcos): purposedly disabling cache on 'master' for now
```
steven-johnson, line 2564: Should probably create an Issue on Github to track this, to lessen the chance we forget to ever re-enable it...

src/runtime/windows_d3d12compute_x86.cpp
slomp: Yeap, it was briefly discussed here:
https://github.com/halide/Halide/issues/5156#issuecomment-674248591

src/runtime/d3d12compute.cpp
```diff
@@ -2711,6 +2778,16 @@ WEAK int halide_d3d12compute_device_release(void *user_context) {
     if (device) {
         d3d12compute_device_sync_internal(device, NULL);
 
+        for (int i = 0; i < MaxFrames; ++i) {
+            d3d12_frame *frame = &frame_pool[i];
+            release_object(frame);
```
steven-johnson, line 2783: Does `release_object()` set the argument to null? If not, shouldn't we?

src/runtime/d3d12compute.cpp
```diff
@@ -3032,6 +3109,19 @@ WEAK int halide_d3d12compute_run(void *user_context,
         }
     }
 
+    {
+        TRACE_SCOPE("pipeline barriers");
+        // TODO(marcos): avoid placing UAV barriers all the time after a dispatch...
+        // in fact, only buffers written to by the dispatch will need barriers, and
+        // only when later bound for read. For now, Halide does not provide enough
+        // context for chosing the right time to place transition barriers.
```
steven-johnson, line 3117: choosing

src/runtime/d3d12compute.cpp
```diff
@@ -3032,6 +3109,19 @@ WEAK int halide_d3d12compute_run(void *user_context,
         }
     }
 
+    {
+        TRACE_SCOPE("pipeline barriers");
+        // TODO(marcos): avoid placing UAV barriers all the time after a dispatch...
```
steven-johnson, line 3114: Create a GitHub issue for this TODO and insert link to it here

steven-johnson:
Build looks clean. Are we ready to squash-and-land?

shoaibkamil:
> Build looks clean. Are we ready to squash-and-land?

I think we want to address your comments and then squash-and-land-- we'll do that today.

src/runtime/d3d12compute.cpp
slomp: In this case, no.
the frame pool is allocated statically when the runtime module is loaded:
```
WEAK d3d12_frame frame_pool[MaxFrames] = {};
```

steven-johnson: No, what I mean is that after the call to `release_object()`, the contents of `frame_pool[i]` are now invalid (the internal pointers refer to a released object). Shouldn't we explicitly clear `frame_pool[i]` to zero to ensure we don't accidentally reuse a stale object? eg

```
template<typename d3d12_T>
static void release_object(d3d12_T *obj) {
    TRACELOG;
    if (!obj) {
        TRACEPRINT("null object -- nothing to be released.\n");
        return;
    }
    release_d3d12_object(obj);
    memset(obj, 0, sizeof(*obj));  // <-- THIS
}
```

src/runtime/d3d12compute.cpp
slomp: Done: https://github.com/halide/Halide/issues/5309

src/runtime/d3d12compute.cpp
slomp: Done: https://github.com/halide/Halide/issues/5309

src/runtime/d3d12compute.cpp
slomp: Ah, you're right, I thought I was resetting the fields in the corresponding `release_d3d12_object<d3d12_frame>` function, but I was only resetting `frame->fence_signal`.

PR #5303, opened 2020-09-29
Fix runtime build rules in Makefile

steven-johnson:
- one of the Windows-specific runtime files had a 32-vs-64 glitch
- CMake now uses `fno-threadsafe-statics` (instead of `-std=gnu++98`) to disable thread-safe static initialization; as a result, this allowed C++11 code requirements to creep in (via d3d12compute.cpp), but we didn't notice because the Makefile wasn't properly building that file due to the 32-vs-64 glitch. Fixed by updating the Makefile to use this flag instead (which was an overdue fix anyway).


alexreinking (approved):

abadams:
oops, thanks for the fix.

abadams (approved):

PR #5305, opened 2020-09-30
Use __builtin_shuffle[vector] for C backend

steven-johnson:
Attempt to use the clang/gcc builtins for doing shuffles.

Notes:
- Apparently *none* of our AOT tests generated any shuffle ops, so I added one.
- I have not yet profiled or disassembled the resulting code to see what is being generated.
- This modification assumes/requires that the generated code has C++11 (at least) available, which was not a stated requirement before.


src/CodeGen_C.cpp
```diff
@@ -526,14 +555,22 @@ class CppVector {
         }
     }
 
-    static Vec shuffle(const Vec &a, const int32_t indices[Lanes]) {
+    template<int N, int First, int Second, int... Rest>
+    static void shuffle_impl(Vec &dst, const Vec &src) {
+        dst.elements[N] = src.elements[First < 0 ? 0 : First];
+        shuffle_impl<N + 1, Second, Rest...>(dst, src);
```
dsharletg, line 561: You don't need `Second`, you can just do Rest...

src/CodeGen_C.cpp
```diff
@@ -526,14 +555,22 @@ class CppVector {
         }
     }
 
-    static Vec shuffle(const Vec &a, const int32_t indices[Lanes]) {
+    template<int N, int First, int Second, int... Rest>
+    static void shuffle_impl(Vec &dst, const Vec &src) {
+        dst.elements[N] = src.elements[First < 0 ? 0 : First];
```
dsharletg, line 560: It might be better to put this in an `if (First >= 0)`, so the compiler is free to ignore the undef elements.

src/CodeGen_C.cpp
```diff
@@ -526,14 +555,22 @@ class CppVector {
         }
     }
 
-    static Vec shuffle(const Vec &a, const int32_t indices[Lanes]) {
+    template<int N, int First, int Second, int... Rest>
+    static void shuffle_impl(Vec &dst, const Vec &src) {
```
dsharletg, line 559: It might be helpful to have a few inlined versions of small numbers of indices, to reduce the recursion depth. I found this improved intermediate code size/compile time in similar code.

src/CodeGen_C.cpp
steven-johnson: I don't understand. Surely that would drop `Second` on the floor?

dsharletg: You don't need the Second template parameter either:

    template<int N, int First, int... Rest>
    static void shuffle_impl(Vec &dst, const Vec &src) {
        dst.elements[N] = src.elements[First < 0 ? 0 : First];
        shuffle_impl<N + 1, Rest...>(dst, src);
    }

Should work.

steven-johnson: No, if I omit that, the code become ambiguous, since `shuffle_impl<1>()` can match either of the expansions, since `...Rest` can be empty.

steven-johnson:
PTAL

src/CodeGen_C.cpp
dsharletg: Ah, the other thing I did differently in similar code was to have the base case be empty, not the last element. That avoids duplicating the body of the recursion.

steven-johnson: ahh, ok, that makes sense, let me try that.

steven-johnson:
PTAL

src/CodeGen_C.cpp
```diff
@@ -2845,9 +2876,7 @@ void CodeGen_C::visit(const Shuffle *op) {
     if (op->type.is_scalar()) {
         rhs << src << "[" << op->indices[0] << "]";
     } else {
-        string indices_name = unique_name('_');
-        stream << get_indent() << "const int32_t " << indices_name << "[" << op->indices.size() << "] = { " << with_commas(op->indices) << " };\n";
-        rhs << print_type(op->type) << "::shuffle(" << src << ", " << indices_name << ")";
+        rhs << print_type(op->type) << "::shuffle<" << with_commas(op->indices) << ">(" << src << ")";
```
dsharletg, line 2879: A thought from my last pass through this: should we just generate code like:

    Vec result = { operand[indices_0], operand[indices_1], ... };

? Or at least in the case we lack builtin_shuffle?

steven-johnson: Is that likely to be a win in terms of the machine code we generate?

dsharletg: I think it's a win in that we generate less code for the compiler to parse and optimize. The final machine code probably wouldn't be very different. 

I would expect the code I proposed to be more likely to get optimized as a shuffle, but I would think they both do.

steven-johnson: Yeah, just thinking about it more though... realistically, I don't think we should ever even see that case. If we have native vector support, we're gonna have __builtin_shuffle or __builtin_shufflevector (for gcc vs clang). Putting `#error` there is probably simpler.

(MSvC supports none of this, of course...)

steven-johnson:
I'm probably going to drop this in favor of a more comprehensive update to this code.

PR #5307, opened 2020-09-30
Fix host_supports_target_device()

shoaibkamil:
`host_supports_target_device()` was returning true if it failed to get a `DeviceAPI` corresponding to the target.

Drive-by fix: in a `-metal-debug` target, `halide_current_time_ns()` was being called in `halide_metal_device_malloc()` before timer initialization, which happens as part of context creation.

### DIFF
--- a/src/DeviceInterface.cpp
+++ b/src/DeviceInterface.cpp
@@ -35,7 +35,9 @@ bool lookup_runtime_routine(const std::string &name,
 bool host_supports_target_device(const Target &t) {
     const DeviceAPI d = t.get_required_device_api();
     if (d == DeviceAPI::None) {
-        return false;
+        // If the target requires no DeviceAPI, then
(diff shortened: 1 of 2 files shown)


steven-johnson (approved):

abadams:
This doesn't make sense to me. If no device api is required by the target, then yeah, the host target supports that vacuously, so you return true.

slomp:
Ah, so that's what it means... Yeah, I get it now.
@shoaibkamil Maybe add Andrew's remark as a comment to that `return true;` statement to avoid confusion in the future.

shoaibkamil:
> 
> 
> This doesn't make sense to me. If no device api is required by the target, then yeah, the host target supports that vacuously, so you return true.

Doh.  Ok, will add a comment and revert that change (metal change is necessary).

PR #5308, opened 2020-09-30
Fix linking to shared LLVM. 

### LINKED ISSUE #5304: Regression of issue #5015
It seems in early August during the cleaning up of ./dependencies/llvm/CMakeLists.txt, the fixes for issue #5015 seem to have been lost.  I was able to get it working with a 3 line patch, but I suspect that this isn't the fix you are ultimately going to want.

```diff
diff --git a/dependencies/llvm/CMakeLists.txt b/dependencies/llvm/CMakeLists.txt
index 7222f0582..4d13f1ac4 100644
--- a/dependencies/llvm/CMakeLists.txt
+++ b/dependencies/llvm/CMakeLists.txt
@@ -71,7 +71,10 @@ endforeach ()
 option(Halide_BUNDLE_LLVM "When built as a static library, include LLVM's objects." OFF)
 option(Halide_SHARED_LLVM "Link against shared LLVM (disables components)." OFF)
 
+if (NOT DEFINED LLVM_LIBNAMES)
 llvm_map_components_to_libnames(LLVM_LIBNAMES ${LLVM_COMPONENTS})
+endif ()
+
 if (TARGET_WEBASSEMBLY)
     find_package(LLD CONFIG REQUIRED HINTS "${LLVM_DIR}/../lld")
     list(APPEND LLVM_LIBNAMES lldWasm)
```

I only have x86 and NVPTX backends on my llvm install, so I couldn't test all of the backends.  With this patch, I was able to build Halide when I passed the following arguments to CMake

```console
cmake -DHalide_BUNDLE_LLVM=OFF -DHalide_SHARED_LLVM=ON -DBUILD_SHARED_LIBS=ON .. -DLLVM_LIBNAMES=$(llvm-config --libs) -G Ninja
```

alexreinking:
Apparently, LLVM's own `llvm_map_components_to_libnames` function is busted if you build only the shared library version.

This PR should fix that up and avoid calling this function when shared LLVM is requested.

We should backport this to 10.0.0 for future Deb packaging, and we probably need to add buildbot testing of shared and bundled scenarios. Maybe just the x64 linux bot can run the generator tests.

Fixes #5304.

### DIFF
--- a/dependencies/llvm/CMakeLists.txt
+++ b/dependencies/llvm/CMakeLists.txt
@@ -3,13 +3,14 @@
 ##
 
 include(CMakeDependentOption)
+include(BundleStatic)
 
 # Fallback configurations for weirdly built LLVMs
 set(CMAKE_MAP_IMPORTED_CONFIG_MINSIZEREL MinSizeRel Release RelWithDebInfo "")
 set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO RelWithDebInfo Release MinSizeRel "")
 set(CMAKE_MAP_IMPORTED_CONFIG_RELEASE Release MinSizeRel RelWithDebInfo "")
 
-find_package(LLVM ${HALIDE_REQUIRE_LLVM_VERSION} REQUIRED)
+find_package(LLVM ${Halide_REQUIRE_LLVM_VERSION} REQUIRED)
 find_package(Clang REQUIRED CONFIG HINTS "${LLVM_DIR}/../clang")
 
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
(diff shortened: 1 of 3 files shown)


dependencies/llvm/CMakeLists.txt
```diff
@@ -25,6 +26,14 @@ endif ()
 
 set(Halide_LLVM_DEFS ${LLVM_DEFINITIONS} $<BUILD_INTERFACE:LLVM_VERSION=${LLVM_VERSION_MAJOR}${LLVM_VERSION_MINOR}>)
 
+##
+# Options for which version of LLVM to use
+##
+
+option(Halide_SHARED_LLVM "Link against shared LLVM (ignores components)." OFF)
+cmake_dependent_option(Halide_BUNDLE_LLVM "When built as a static library, include LLVM's objects." OFF
```
steven-johnson, line 34: A comment about why this is necessary would probably be good

steven-johnson (approved):
LGTM but I think we need explicit comment added somewhere about why this is needed (eg "LLVM's own llvm_map_components_to_libnames function is busted if you build only the shared library version")

alexreinking:
> 
> 
> LGTM but I think we need explicit comment added somewhere about why this is needed (eg "LLVM's own llvm_map_components_to_libnames function is busted if you build only the shared library version")

Done. Added the comment inside the `if(Halide_SHARED_LLVM)` branch.

steven-johnson:
Does this need to run on the buildbots?

alexreinking:
> 
> 
> Does this need to run on the buildbots?

Yes, for sure

alexreinking:
Any idea why the buildbots aren't running on this?

steven-johnson:
Buildbots are very backed up after changes yesterday -- are they pending?

steven-johnson:
Failure is unrelated, landing

PR #5315, opened 2020-09-30
Kill halide_abort()

steven-johnson:
This was added long ago as an attempt to work around issues with the Windows Debug runtime (in which calling `abort()` would produce an "Abort, Retry, Ignore" dialog). We no long do debug builds of any sort on our buildbots, so let's lose all this mess to simplify our world a bit.


abadams:
This would also remove the ability for custom runtimes to override the abort behavior. I don't think we care, because that's just halide_default_error, which is replaceable, and halide_assert, which *should* only be for checking invariants to guard against runtime bugs.

abadams (approved):

alexreinking:
On Windows, we (or our users) can add the following code to our tests to swallow the aforementioned dialog:

```
_CrtSetReportMode(_CRT_ERROR, _CRTDBG_MODE_DEBUG);
_CrtSetReportMode(_CRT_ASSERT, _CRTDBG_MODE_DEBUG);
_CrtSetReportMode(_CRT_WARN, _CRTDBG_MODE_DEBUG);
```

or, perhaps more helpfully,

```
_CrtSetReportMode(_CRT_ERROR, _CRTDBG_MODE_FILE);
_CrtSetReportFile(_CRT_ERROR, _CRTDBG_FILE_STDERR);

_CrtSetReportMode(_CRT_ASSERT, _CRTDBG_MODE_FILE);
_CrtSetReportFile(_CRT_ASSERT, _CRTDBG_FILE_STDERR);

_CrtSetReportMode(_CRT_WARN, _CRTDBG_MODE_FILE);
_CrtSetReportFile(_CRT_WARN, _CRTDBG_FILE_STDERR);
```

steven-johnson:
Failures are unrelated.

PR #5318, opened 2020-10-01
Add a header-only library that handles complex types.

Infinoid:
This adds a headerfile into `tools/`, implementing the ComplexFunc and ComplexExpr classes to provide syntax for doing math on Complex values.  It wraps around a Func, adding an extra inner dimension (of size 2) to store the real and imaginary components.

It is useful for larger applications which use Halide for certain pieces, and want to pass arrays of complex numbers between Halide and other tools.  Tools like `fftw` expect data to be laid out in memory as an array of complex values, like `double complex_array[200][2]` or  `complex<double> array[200]`.

It is inspired by the ComplexFunc implementation in the FFT application, but takes the wrapper approach instead of the Tuple approach.  The Tuple approach lays out the data in memory as two arrays, one for real elements, one for imaginary elements.  This means passing outputs to external tools requires an extra conversion step.  Acting as a wrapper makes that extra step unnecessary.

This is a work in progress.  TODO:
* implement *= and /= if possible
* move halide_complexfunc.h to src/stdlib/

### DIFF
--- a/apps/fft/complex.h
+++ b/apps/fft/complex.h
@@ -106,6 +106,9 @@ inline ComplexExpr operator/(ComplexExpr a, Halide::Expr b) {
     return ComplexExpr(re(a) / b, im(a) / b);
 }
 
+inline ComplexExpr exp(const ComplexExpr &z) {
+    return ComplexExpr(Halide::exp(re(z)) * Halide::cos(im(z)), Halide::exp(re(z)) * Halide::sin(im(z)));
+}
 // Compute exp(j*x)
 inline ComplexExpr expj(Halide::Expr x) {
     return ComplexExpr(Halide::cos(x), Halide::sin(x));
--- a/test/correctness/CMakeLists.txt
+++ b/test/correctness/CMakeLists.txt
@@ -51,6 +51,8 @@ tests(GROUPS correctness
       compile_to_bitcode.cpp
       compile_to_lowered_stmt.cpp
       compile_to_multitarget.cpp
+      complexfunc.cpp
+      complexfunc_tuple.cpp
       compute_at_reordered_update_stage.cpp
       compute_at_split_rvar.cpp
       compute_inside_guard.cpp
@@ -386,4 +388,3 @@ set_target_properties(correctness_async
                       correctness_sliding_window
                       correctness_storage_folding
                       PROPERTIES ENABLE_EXPORTS TRUE)
-
--- a/test/correctness/complexfunc.cpp
+++ b/test/correctness/complexfunc.cpp
@@ -1,0 +1,319 @@
+#include "Halide.h"
+#include "halide_complexfunc.h"
+#include <cmath>
+#include <complex>
+#include <iostream>
+
+using namespace Halide;
+using Halide::Tools::ComplexExpr;
+using Halide::Tools::ComplexFunc;
+using Halide::Tools::exp;
+using Halide::Tools::expj;
+using std::cout;
+using std::endl;
+
+#define N 5
+
+void print_buf(const char *prefix, std::complex<double> *buf, int X, int Y) {
+    for (int y = 0; y < Y; y++) {
+        printf("%s row %d = [", prefix, y);
+        for (int x = 0; x < X; x++) {
+            std::complex<double> value = buf[y * X + x];
+            printf("%c%4.1f+%.1fi ", x ? ',' : ' ', value.real(), value.imag());
+        }
+        printf("]\n");
+    }
+}
+
+Buffer<double> gen_buf() {
+    Buffer<double> input(2, N);
+    std::complex<double> *input_ptr = (std::complex<double> *)input.begin();
+    for (int i = 0; i < N; i++) {
+        std::complex<double> value(1.0 + i, i - 1.0);
+        input_ptr[i] = value;
+    }
+    return input;
+}
+
+void test_io() {
+    printf("test_io\n");
+    // test that complex values can be passed into and out of a Halide kernel.
+    Buffer<double> input = gen_buf();
+
+    Var c("c"), x("x");
+    Func input_clamped = BoundaryConditions::constant_exterior(input, Expr(0.0));
+    ComplexFunc input_complex(c, input_clamped);
+    ComplexFunc result(c, "result");
+    result(x) = input_complex(x);
+
+    Buffer<double> output = result.inner.realize(2, N);
+
+    for (int i = 0; i < N; i++) {
+        int failure_count = 0;
+        std::complex<double> value(1.0 + i, i - 1.0);
+        if (std::abs(output(0, i) - value.real()) > 0.01) {
+            cout << "Wrong real value for element " << i << ". Expected " << value.real() << ", got " << output(0, i) << endl;
+            failure_count++;
+        }
+        if (std::abs(output(1, i) - value.imag()) > 0.01) {
+            cout << "Wrong imaginary value for element " << i << ". Expected " << value.imag() << ", got " << output(1, i) << endl;
+            failure_count++;
+        }
+        if (failure_count)
+            abort();
+    }
+}
+
+void test_ops_complex_complex() {
+    printf("test_ops_complex_complex\n");
+    Buffer<double> input = gen_buf();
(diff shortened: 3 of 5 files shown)


Infinoid:
Ideally this would go into a sort of standard library for Halide app development, so that apps don't have to copy it or roll their own.  @bachili suggested that I put it in `tools/`, so that's where I put it.

BachiLi:
For the context: I think it might be nice that we start to collect these utility functions/classes to build a standard library for Halide.

dsharletg:
Can you replace the ComplexExpr/ComplexFunc in apps/fft with this? I think it would be good to have a concrete use case of this in the PR, and avoid the duplication. Hopefully the Tuple -> Wrapper thing isn't too big of a headache?

Infinoid:
> Can you replace the ComplexExpr/ComplexFunc in apps/fft with this?

I was planning to do this as a local test case / sanity check.  But before I go deleting the Tuple version outright, I'd want some confirmation that it isn't worth keeping around for some other reason.

alexreinking:
> 
> 
> Ideally this would go into a sort of standard library for Halide app development, so that apps don't have to copy it or roll their own. @BachiLi suggested that I put it in `tools/`, so that's where I put it.

Please don't put it in tools... I see tools as a place for implementation-detail code, like GenGen.cpp / RunGenMain.cpp, both of which are referenced only in the build system, not in user code. The other things in there are build scripts and build-time tools that need to be built in order to build Halide (eg. build_halide_h.cpp).

I'm not sure why this shouldn't just be added to libHalide / placed in `src`. Of course, the header should include the relevant internal headers and not all of `Halide.h` in this case.

If we _really_ want to add levels / components to Halide, ie. a "standard library for Halide", we should discuss what ought to be moved out of the main library, and move them to separate directories, eg. `src/{compiler,stdlib}` (maybe even a separate repo?).

BachiLi:
> > Ideally this would go into a sort of standard library for Halide app development, so that apps don't have to copy it or roll their own. @BachiLi suggested that I put it in `tools/`, so that's where I put it.
> 
> **Please don't put it in tools**. Tools is a place for implementation-detail code, like GenGen.cpp / RunGenMain.cpp, both of which are referenced only in the build system, not in user code. The other things in there are build scripts and build-time tools that need to be built in order to build Halide (eg. build_halide_h.cpp). Yes, `halide_image_io.h` and `halide_benchmark.h` are in there, but I think those should be moved out, too.
> 
> I'm not sure why this shouldn't just be added to libHalide / placed in `src`. Of course, the header should include the relevant internal headers and not all of `Halide.h` in this case.
> 
> If we _really_ want to add levels / components to Halide, ie. a "standard library for Halide", we should discuss what ought to be moved out of the main library, and move them to separate directories, eg. `src/{compiler,stdlib}` (maybe even a separate repo?).

Having a header-only stdlib that agglomerate into a separate header file that sits in src/stdlib sounds like a good idea.

alexreinking:
For instance, I could see us installing a directory structure similar to this:

```
include/
|- Halide.h  (mono header)
|- Halide/   (optional sub-components)
   |- ImageIO.h
   |- Benchmark.h
   |- TraceConfig.h
   |- Complex.h
```

Where `include/Halide.h` is rolled up from the headers in `src/compiler` (or `src/main` or whatever we bike-shed), and the headers in `include/Halide/*` are copied verbatim from `src/stdlib`.

Infinoid:
Sure, I don't have strong feelings about `tools/`, we can move it wherever it needs to go.

> like GenGen.cpp / RunGenMain.cpp, both of which are referenced only in the build system, not in user code.

Is it a distinction between source and header files?  Or a distinction between things useful at generation time, vs things useful at runtime?

dsharletg:
> But before we go deleting the Tuple version outright, I'd like some confirmation that it isn't worth keeping around for some other reason.

I don't think it is. If this is better, we should just replace it.

However, I do wonder about this. By "conversion step" in your PR description, do you mean just some `Halide::Runtime::Buffer` setup, or actually memcpying? I think the tuple approach is not incompatible with other APIs (like `std::complex` arrays), you just make two `Halide::Runtime::Buffer`s that address even/odd elements (i.e. the innermost element has stride 2), and pass them as the two tuple elements.

Aside from that, the approach in this PR has some pros and cons:
- Pro: you can more easily write more interesting schedules for interleaved complex data (like std::complex arrays).
- Con: It will require more effort to schedule, even in the 'planar' case.

Infinoid:
> By "conversion step" in your PR description, do you mean just some `Halide::Runtime::Buffer` setup, or actually memcpying?

I mean actual copying.  I'll give an example.

Assume we have a buffer containing N (interleaved) complex numbers that we received from a libfftw call.  We want to pass those to Halide to do some additional math.  Here's how I've been doing that:

```C++
    Buffer<double> input(2, N);
    complex_ptr = (std::complex<double>*)input.begin();
    // call fftw, write complex data to complex_ptr

    Var c("c"), x("x");
    Func input_clamped = BoundaryConditions::constant_exterior(input, Expr(0.0));
    ComplexFunc input_complex("input_complex"), result("result");
    input_complex(x) = ComplexExpr(input_clamped(0,x), input_clamped(1,x)); // INPUT COPY
    // Halide algorithm that does things with input_complex, and writes output to result

    Func interleaved_result("interleaved_result");
    interleaved_result(c, x) = select(c == 0, result(x).re(), result(x).im()); // OUTPUT COPY

    Buffer<double> output = interleaved_result.realize(2, N);
    output_ptr = (std::complex<double>*)input.begin();
    // call fftw again, reading complex data from output_ptr
```

Is there a way to avoid the input/output copies?  Maybe they can be inlined?

Infinoid:
I've been trying to complete the wrapper class, and ran into a problem.  I'm having trouble implementing the wrapper version of the `*=` operator.  That operator would implement `A *= B` like this:

```
A(c,x) = select(c, 
        A(0,x) * B(0,x) - A(1,x) * B(1,x),  // real*real - imaginary*imaginary
        A(0,x) * B(1,x) - A(1,x) * B(0,x)); // real*imaginary - imaginary*real
```

which Halide rejects with the following message:

```
Error: In definition of Func "f12":
All of a function's recursive references to itself must contain the same pure variables in the same places as on the left-hand-side.
```

The `/=` operator will likely run into similar problems.

Halide has good reasons to reject it.  Calculation of either (real or imaginary) component requires the old values of both sides as input.  The schedule may calculate the real and imaginary components separately in time, but Halide can't keep the old values around without separate storage.

The Tuple approach doesn't have this problem, I think because it forces both components to be calculated/stored simultaneously.  So the tuple approach might be the only valid approach for a general complex number class like this.

dsharletg:
> Is there a way to avoid the input/output copies? Maybe they can be inlined?

Yes, there are a few options.

First, you can make some wrappers for an `std::complex<double>* x`:

    Halide::Runtime::Buffer<double> real_part(reinterpret_cast<double*>(x), 1, {{0, N, 2}});
    Halide::Runtime::Buffer<double> imag_part(reinterpret_cast<double*>(x) + 1, 1, {{0, N, 2}});

These are two buffers that have stride 2, offset by one element. This will probably require the pipelines you are passing them to to adjust the expected strides (i.e. `input_tuple.dim(0).set_stride(2)`). However, this will still generate not-quite-optimal code, because the compiler probably won't understand that the two tuple elements are part of a two element structure.

Another option is to make a custom pipeline with a wrapper func that unpacks the struct input to a tuple:

```
    Input<Buffer<double>> input_std_complex{"input_std_complex", 2};
    ...
    input_std_complex.dim(0).set_stride(1).set_min(0).set_extent(2);
    input_std_complex.dim(1).set_stride(2);

    ComplexFunc input;
    input(x) = ComplexExpr(input_std_complex(x, 0), input_std_complex(x, 1));
    ...
    // Call generator helpers that need ComplexFunc, probably let input inline.
```

steven-johnson:
I assume this PR is still on hold?

Infinoid:
This PR is canceled for now.  I still think Halide needs some standard header libraries, including things like complex value handling, but I don't know which implementation is best for that.

For posterity, the most recent version of wrapper-style complexfunc is at https://github.com/ISI-apex/halide-sar-app/blob/master/halide_complexfunc.h .

PR #5322, opened 2020-10-02
Make sure the runtime compiler settings in CMake match those in Make

steven-johnson:
Mainly, we weren't setting any of the warning flags, so CMake builds were more forgiving than Make.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -747,7 +747,7 @@ std::string halide_type_to_c_type(const Type &t) {
 
 int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
     const char kUsage[] =
-        "gengen\n"
+        "gengen \n"
         "  [-g GENERATOR_NAME] [-f FUNCTION_NAME] [-o OUTPUT_DIR] [-r RUNTIME_NAME] [-d 1|0]\n"
         "  [-e EMIT_OPTIONS] [-n FILE_BASE_NAME] [-p PLUGIN_NAME] [-s AUTOSCHEDULER_NAME]\n"
         "       target=target-string[,target-string...] [generator_arg=value [...]]\n"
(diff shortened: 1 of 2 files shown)


src/runtime/CMakeLists.txt
```diff
@@ -145,11 +145,27 @@ set(RUNTIME_HEADER_FILES
...
 add_library(Halide_initmod OBJECT)
 
-set(CXX_WARNING_FLAGS -Wall -Werror -Wno-unused-function -Wcast-qual)
-set(RUNTIME_CXX_FLAGS -O3
-    -fno-vectorize -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables
+# Note: ensure that these flags match the flags in the Makefile.
+# Note: this always uses Clang-from-LLVM for compilation, so none of these flags should need conditionalization.
+set(RUNTIME_CXX_FLAGS
+    -O3
+    -ffreestanding
```
abadams, line 152: This is more than warnings - a lot of these are really important (mostly freestanding)

steven-johnson: The non-warnings were set before. I just reformatted to be one-per-line and alphabetical order.

alexreinking (approved):

src/runtime/CMakeLists.txt
abadams: oops, right you are. I was wondering why anything was working.

src/Generator.cpp
```diff
@@ -747,7 +747,7 @@ std::string halide_type_to_c_type(const Type &t) {
 
 int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
     const char kUsage[] =
-        "gengen\n"
+        "gengen \n"
```
alexreinking, line 750: What's this doing here?

steven-johnson: Eep, I thought I had reverted that. It was a change I pushed to tickle the buildbots. Sorry.

PR #5324, opened 2020-10-03
InjectBufferCopiesForInputsAndOutputs should check for unexpected Call nodes.

steven-johnson:
~~I haven't found a case where this is actually a problem, but I noticed from debugging a different issue that the FindInputsAndOutputs doesn't look at Call nodes at all, so it can miss Buffers/Parameters associated with them.~~

~~Not sure if this is an issue in practice. Maybe we never want buffer copies for these? (If so then we should document it as such.)~~

~~This PR just adds the logic used for this in InferArguments.~~

### DIFF
--- a/src/InjectHostDevBufferCopies.cpp
+++ b/src/InjectHostDevBufferCopies.cpp
@@ -141,9 +141,9 @@ class InjectBufferCopiesForSingleBuffer : public IRMutator {
     using IRMutator::visit;
 
     // The buffer being managed
-    string buffer;
+    const string buffer;
 
-    bool is_external;
+    const bool is_external;
 
     enum FlagState {
         Unknown,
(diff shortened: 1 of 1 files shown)


abadams:
I think the reason is that this is run after storage_flattening, so call nodes aren't used to access buffers.

steven-johnson:
> I think the reason is that this is run after storage_flattening, so call nodes aren't used to access buffers.

Ah, ok. I'll try adding some assertion checks then to verify we aren't overlooking anything.

abadams:
Also, traversing from Calls to the Func definition is a bit weird here, because we're both traversing all of a Stmt and also the Function DAG. I'm not convinced it's correct. It certainly wouldn't be if this mutator were ever used on a sub-stmt.

steven-johnson:
PTAL

abadams (approved):

PR #5325, opened 2020-10-03
Better simplification and codegen for nested vectorization

abadams:
This PR adds some simplifier rules and tweaks some aspects of shuffle codegen to make nested vectorization a great way to do certain dot-producty tasks. For an 8-bit matrix multiply with 32-bit output or an 8-bit large-support blur on arm with the sdot instruction available this is a full four times faster than our best current schedule! Without sdot there's still a ~1.5x speedup on x86 (via better use of pmaddwd) and arm (via better use of horizontal widening adds).

I changed the simplifier to explicitly track lanes for broadcast and ramp. The lane value can be a constant wildcard and can participate in the predicate. This was not the approach @vksnk initially went with because it was more complex than necessary at the time, but it proved necessary for some of the rules I wanted.


src/Simplify_Add.cpp
```diff
@@ -49,14 +49,19 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
         // clang-format off
         if (EVAL_IN_LAMBDA
             (rewrite(x + x, x * 2) ||
-             rewrite(ramp(x, y) + ramp(z, w), ramp(x + z, y + w), is_same_type(x, z)) ||
-             rewrite(ramp(x, y) + broadcast(z), ramp(x + z, y), is_same_type(x, z)) ||
-             rewrite(broadcast(x) + broadcast(y), broadcast(x + y), is_same_type(x, y)) ||
-             rewrite((x + broadcast(y)) + broadcast(z), x + broadcast(y + z), is_same_type(y, z)) ||
-             rewrite((x - broadcast(y)) + broadcast(z), x + broadcast(z - y), is_same_type(y, z)) ||
+             rewrite(ramp(x, y, c0) + ramp(z, w, c0), ramp(x + z, y + w, c0)) ||
```
abadams, line 52: @jn80842 @shoaibkamil Simplifier rules with nested vectors are starting to get non-trivial

src/CodeGen_LLVM.cpp
```diff
@@ -4261,17 +4270,94 @@ void CodeGen_LLVM::visit(const Evaluate *op) {
...
+                }
+                Expr equiv = Shuffle::make(op->vectors, sub_indices);
+                value = codegen(equiv);
+                if (f == 2) {
+                    value = interleave_vectors({value, value});
+                } else if (f == 3) {
+                    value = interleave_vectors({value, value, value});
+                } else {
+                    value = interleave_vectors({value, value});
+                    value = interleave_vectors({value, value});
```
dsharletg, line 4301: I think we should let `interleave_vectors` handle this. Maybe there is some backend where there is a better implementation of `interleave_vectors({value, value, value, value})`? And if not, the implementation should just do this anyways.

And on that note, maybe just write this whole if as `interleave_vectors(std::vector<Value*>(f, value))`

src/CodeGen_LLVM.cpp
```diff
@@ -4261,17 +4270,94 @@ void CodeGen_LLVM::visit(const Evaluate *op) {
...
+
+            if (op->type.bits() * f > 64) {
+                continue;
+            }
+            bool contiguous = true;
+            for (const Expr &vec : op->vectors) {
+                contiguous &= ((vec.type().lanes() % f) == 0);
+            }
+            for (size_t i = 0; i < op->indices.size(); i++) {
+                contiguous &= (op->indices[i] == op->indices[i - (i % f)] + ((int)i % f));
```
dsharletg, line 4341: It might be easier/more readable to just construct the shuffles you are looking for and compare for equality?

src/FlattenNestedRamps.cpp
```diff
@@ -33,8 +37,79 @@ class FlattenRamps : public IRMutator {
...
+                // Compute the number of elements loaded
+                int extent = (int)((max_constant_offset / stride) + 1);
+
+                // If we're gathering from a very large range, it
+                // might be better to just do the gather rather than
+                // doing a big dense load and then shuffling. We
+                // currently do the big-load-and-shuffle if we're
+                // going to use at least a quarter of the values
+                // loaded.
+                const int max_unused_lane_factor = 4;
```
dsharletg, line 101: TODO: Express this in the schedule somehow? Maybe require a wrapper that is `store_in(MemoryType::Register)`?

test/performance/nested_vectorization_gemm.cpp
```diff
@@ -0,0 +1,225 @@
...
+                        .update()
+                        .split(r, ro, ri, vec / 2)
+                        .reorder(ri, x, y, ro)
+                        .vectorize(x)
+                        .unroll(y)
+                        .atomic()
+                        .vectorize(ri, 2)
+                        .unroll(ri);
+                } else {
+                    // ARM schedule. Relies on SDOT
```
dsharletg, line 57: Not always?

dsharletg (approved):
Thanks, this is cool! @vksnk

vksnk (commented):
This is so cool!

src/Simplify_Add.cpp
shoaibkamil: Yeah, we should add support for verifying these.

src/CodeGen_LLVM.cpp
abadams: Improved it by changing to a nested loop (and caught a bug!)

src/CodeGen_LLVM.cpp
```diff
@@ -2683,7 +2683,16 @@ void CodeGen_LLVM::visit(const Call *op) {
             }
 
         } else {
+            if (src.is_scalar() && dst.is_vector()) {
+                value = create_broadcast(value, 1);
```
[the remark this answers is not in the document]
abadams: This was indeed very confusing. After I puzzled it out I added a comment explaining it.

steven-johnson:
Looks like some MSVC-specific template evaluation failure, eg https://buildbot.halide-lang.org/master/#/builders/66/builds/12

abadams:
Oh my, it's trying to use the IRMatcher infrastructure to compare two ints. I'll make a constructor explicit and see if it fixes things.

steven-johnson:
> I'll make a constructor explicit and see if it fixes things.

IMHO we'd be better off with a general code policy of "all ctors must be explicit", with exceptions made on a case-by-case basis. (Ditto for implicit `operator sometype()` operators.)

That said, we have a lot of reliance on implicit conversions in our API, so even if we all agreed that it was a good idea, it would be hard to do without breaking existing usage.

abadams:
This was a code where there was no reason for it to be implicit - just an oversight while writing it. So requiring intentional annotation for everything that is indeed supposed to be implicit would be helpful.

steven-johnson:
I think there is a clang-tidy rule for this. Let me take a look.

abadams:
Meanwhile making the IntLiteral constructor explicit didn't fix things. MSVC is still trying to compare two enum values by making them into IntLiteral objects first. I'm baffled. I inserted explicit casts to int.

dsharletg:
Is this ready to merge? I'd like to start using this.

steven-johnson:
I think we've just been fighting with getting the build to go green.

dsharletg:
It looks like this passed all the build bots, is it ready to merge?

steven-johnson:
> It looks like this passed all the build bots, is it ready to merge?

LGTM but @abadams probably needs to be the one to answer

PR #5331, opened 2020-10-05
fix bounds inference bug for bounded interval / unbounded interval

rootjalex:
Bug in bounds inference engine, discussed with @abadams @shoaibkamil @alexreinking and @jn80842 . 
Maximum value possible from dividing a bounded interval (a0, a1) buy an unbounded interval is max(-a0, a1), which correspond to the cases that a0 is negative but greater in magnitude than a1 (and divided by -1), or where a1 is the maximum possible value (divided by 1).


abadams:
I will clone this to get it to run on the buildbots. Stand by.

abadams (approved):

PR #5334, opened 2020-10-05
Require braces around if/while bodies.

abadams:
We have informally required braces around single-line if statements for a while now. This PR mechanically enforces it with a clang-tidy checker. I did some manual massaging of the result in a few places.


steven-johnson (approved):
Rubberstamping with a hearty agreement

PR #5335, opened 2020-10-06
Reduce GPU Tile Size in Tests

jlaxson:
OpenCL on macOS (Mojave at least) supports a maximum workgroup size of 256.  This causes a few tests to fail that specify larger workgroup sizes when targeting OpenCL on mac.  The sizes don't seem to be substantial to any of these tests, so lower them.

### DIFF
--- a/test/correctness/async_device_copy.cpp
+++ b/test/correctness/async_device_copy.cpp
@@ -45,23 +45,23 @@ int main(int argc, char **argv) {
         if (i == 0) {
             // Synchronously GPU -> CPU
             avg.compute_root().update().reorder(x, y, r).vectorize(x, 8);
-            frames.store_root().compute_at(frames.in(), Var::outermost()).gpu_tile(x, y, xo, yo, x, y, 32, 32);
+            frames.store_root().compute_at(frames.in(), Var::outermost()).gpu_tile(x, y, xo, yo, x, y, 16, 16);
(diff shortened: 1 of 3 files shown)


test/correctness/async_device_copy.cpp
```diff
@@ -45,23 +45,23 @@ int main(int argc, char **argv) {
         if (i == 0) {
             // Synchronously GPU -> CPU
             avg.compute_root().update().reorder(x, y, r).vectorize(x, 8);
-            frames.store_root().compute_at(frames.in(), Var::outermost()).gpu_tile(x, y, xo, yo, x, y, 32, 32);
+            frames.store_root().compute_at(frames.in(), Var::outermost()).gpu_tile(x, y, xo, yo, x, y, 16, 16);
```
steven-johnson, line 48: Nit: if we're going to change these, can we also pull the values out into named constants while we're at it?

steven-johnson (approved):
Would be nice to add comments about tile size for Mac OpenCL to these files to avoid regression.

LGTM with nits.

test/correctness/async_device_copy.cpp
abadams: I disagree. These are numeric parameters tweaked for performance, and unlike most magic constants, have no inherent reason for being what they are. Giving them names just clutters the code.

PR #5339, opened 2020-10-06
Improve NativeVector::convert_from()

steven-johnson:
We should __builtin_convertvector for everything except float->int conversions.


vksnk (approved):

steven-johnson:
I'm probably going to drop this in favor of a more comprehensive update to this code, so please don't land it yet.

PR #5340, opened 2020-10-07
Rewrite internal glue code for vectors in C++ backend.

steven-johnson:
The existing approach (wrapping native vectors in a helper struct) caused too many unnecessary spills. Rewrote so that the underlying Native Vector type is the value being passed around, and improved some of the specializations (with checking of generated code on x64 under clang and gcc). Clang is actually quite good at recognizing patterns and generating appropriate code (at least for x64); gcc is much less so, at least as far as I can tell.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -312,7 +312,7 @@ public:
 
 CodeGen_C::CodeGen_C(ostream &s, Target t, OutputKind output_kind, const std::string &guard)
     : IRPrinter(s), id("$$ BAD ID $$"), target(t), output_kind(output_kind),
-      extern_c_open(false), inside_atomic_mutex_node(false), emit_atomic_stores(false) {
+      extern_c_open(false), inside_atomic_mutex_node(false), emit_atomic_stores(false), using_vector_typedefs(false) {
 
     if (is_header()) {
         // If it's a header, emit an include guard.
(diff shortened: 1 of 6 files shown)


src/CodeGen_C.cpp
```diff
@@ -460,841 +461,737 @@ void CodeGen_C::add_vector_typedefs(const std::set<Type> &vector_types) {
...
     }
 
-    // TODO: this should be improved by taking advantage of native operator support.
     template<size_t InputLanes>
     static Vec concat(size_t count, const NativeVector<ElementType, InputLanes> vecs[]) {
-        Vec r(empty);
+        Vec r;
         for (size_t i = 0; i < Lanes; i++) {
-            r.native_vector[i] = vecs[i / InputLanes][i % InputLanes];
+            r[i] = vecs[i / InputLanes][i % InputLanes];
```
vksnk, line 1064: Same here

vksnk (approved):

dsharletg (approved):

steven-johnson:
Eh, I forgot that there's significant overlap between this and the GPU backends -- the changes to relational operators for vectors are breaking those. I'll circle back and fix that.

steven-johnson:
Only failure is unrelated trunk LLVM breakage. Landing.

PR #5343, opened 2020-10-09
Stronger simplification of fused ramps

dsharletg:
This PR removes the requirement that the denominator in some simplifier rules regarding fused ramps be constant. The new rules just check if the denominator is zero at runtime. If the denominator is constant, this check will simplify away anyways, and if it isn't, this still seems cheaper than computing a mod operation at runtime.

### DIFF
--- a/test/correctness/fuse.cpp
+++ b/test/correctness/fuse.cpp
@@ -64,17 +64,18 @@ int main(int argc, char **argv) {
             .fuse(x, y, xy)
             .vectorize(xy, 16);
 
-        // Compile the program to a stmt.
-        std::string result_file = Internal::get_test_tmp_dir() + "fuse.stmt";
-        Internal::ensure_no_file_exists(result_file);
-        f.compile_to_lowered_stmt(result_file, f.infer_arguments());
+        class CheckForMod : public Internal::IRMutator {
+            using IRMutator::visit;
 
-        // Verify that the compiled stmt does not contain a mod operator.
-        std::vector<char> stmt = Internal::read_entire_file(result_file);
-        if (std::find(stmt.begin(), stmt.end(), '%') != stmt.end()) {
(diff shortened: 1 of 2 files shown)


steven-johnson:
LGTM but deferring to @abadams for Simplifier

abadams:
I have formally verified these rules. The machinery did not suggest any additional variants.

test/correctness/fuse.cpp
```diff
@@ -1,46 +1,81 @@
...
+        p.dim(1).set_stride(f.output_buffer().dim(0).extent());
+
+        // Fuse and vectorize x and y.
+        Var xy("xy");
+        f.compute_root()
+            .fuse(x, y, xy)
+            .vectorize(xy, 16);
+
+        // Compile the program to a stmt.
+        std::string result_file = Internal::get_test_tmp_dir() + "fuse.stmt";
```
abadams, line 68: Would prefer a custom lowering pass to check for const Mod * than a dependence on the file system. E.g:

```
class CheckForMod : public IRMutator {
    using IRMutator::visit;

    Expr visit(const Mod *op) override {
        std::cerr << "Found mod: " << Expr(op) << "\n";
        exit(-1);
        return op;
    }
};

...

    f.add_custom_lowering_pass(new CheckForMod);

```

dsharletg:
Something is wrong, I've pushed a commit per Andrew's comment, but it isn't showing up here on github...

steven-johnson:
> Something is wrong, I've pushed a commit per Andrew's comment, but it isn't showing up here on github...

Yeah, I saw the same thing on another PR. Sounds like a GitHub issue...

dsharletg:
I think this is ready to merge, except for lacking a review... It has passed the build bots and I addressed Andrew's comment.

abadams (approved):

PR #5344, opened 2020-10-09
Mark many single-argument ctors as `explicit`

steven-johnson:
This is some low-hanging fruit found by clang-tidy that seems safe & sensible to land whether or not we decide to enforce requiring explicit in the future.

### DIFF
--- a/src/CPlusPlusMangle.cpp
+++ b/src/CPlusPlusMangle.cpp
@@ -29,10 +29,10 @@ struct MangledNamePart {
     std::string with_substitutions;
 
     MangledNamePart() = default;
-    MangledNamePart(const std::string &mangled)
+    explicit MangledNamePart(const std::string &mangled)
         : full_name(mangled), with_substitutions(mangled) {
     }
-    MangledNamePart(const char *mangled)
+    explicit MangledNamePart(const char *mangled)
         : full_name(mangled), with_substitutions(mangled) {
     }
 };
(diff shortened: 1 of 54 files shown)


abadams:
What does an intentionally-implicit single-arg constructor look like with this checker on?

steven-johnson:
In theory, you'd need to mark it with `// NOLINT` or some similar comment to disable the checking (I forget the exact syntax). 

In practice, this isn't fully ready to enable, because apparently our clang-tidy checker only checks explicitly-listed files, so implicitly-included .h files aren't checked, and that's where most of the interesting exceptions are.

(We could go ahead and land this as-is, but we'd have to agree that explicit-ctors are the correct coding default. Do we think there's any serious pushback on that?)

abadams:
I don't like having to disable the linter as a matter of course. I wish c++ had a keyword for explicitly-implicit. Were there really no intentionally-implicit ones?

steven-johnson:
There are several intentionally-implicit ones, but all are located in the
.h files, not the .cpp files (and thus our current checker skips them).

Yes, it is highly unfortunate that C++ doesn't have a way to mark
'intentionally implicit', but AFAIK, clang-tidy is the only useful tool for
enforcing this.

steven-johnson:
PTAL

src/CPlusPlusMangle.cpp
```diff
@@ -29,10 +29,10 @@ struct MangledNamePart {
     std::string with_substitutions;
 
     MangledNamePart() = default;
-    MangledNamePart(const std::string &mangled)
+    explicit MangledNamePart(const std::string &mangled)
         : full_name(mangled), with_substitutions(mangled) {
     }
-    MangledNamePart(const char *mangled)
+    explicit MangledNamePart(const char *mangled)
```
abadams, line 35: I think this one would be better as implicit

steven-johnson: PTAL. I reverted the changes and added the magic comment that would disable the clang-tidy check if we decided to actually enforce it. (Yes, it's pretty ugly.)

Honestly, we depend a *lot* on implicit conversions in a lot of places, so if decided to turn on this check we'd have a lot of places marked as exempt. It would let us avoid undesired implicit conversions for new code, though.

steven-johnson:
This change isn't exactly critical at this point (or maybe ever), but IMHO marking things as explicit when possible is a good thing, regardless of clang-tidy.

steven-johnson:
Dropping (for now) due to lack of interest.

PR #5345, opened 2020-10-09
Run clang-tidy on src/*.h as well as src/*.cpp

steven-johnson:


abadams (approved):

steven-johnson:
This is still wonky -- and it doesn't help that the `clang-tidy` support in our Makefile is different from what's in the GitHub presubmit test. I'm probably going to withdraw all of these temporarily until I figure out a good way to make it work reliably.

PR #5347, opened 2020-10-12
Move the clang-format and clang-tidy logic into shell scripts

steven-johnson:
This puts the truth for our clang-format and clang-tidy logic into a shell script rather than the Makefile, in hopes of making it slightly easier for CMake users to use. It also adds the crucial `-header-filter=.*` flag that enables actually reporting the tidy errors in our header files, and applies the necessary fixes.


run-clang-format.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -eu
+
+ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+if [ -z "$CLANG_FORMAT" ]; then
+    echo "CLANG_FORMAT not yet, assuming default tool"
```
abadams, line 8: "not yet"?

steven-johnson: s/yet/set/

steven-johnson:
PTAL

steven-johnson:
Finally all green

steven-johnson:
I think this change is safe and sound, PTAL

Makefile
```diff
@@ -2239,51 +2239,19 @@ $(BIN_DIR)/HalideTraceViz: $(ROOT_DIR)/util/HalideTraceViz.cpp $(INCLUDE_DIR)/Ha
...
 	$(CXX) $(OPTIMIZE) -std=c++11 $(filter %.cpp,$^) -I$(INCLUDE_DIR) -I$(ROOT_DIR)/tools -I$(ROOT_DIR)/src/runtime -L$(BIN_DIR) $(IMAGE_IO_CXX_FLAGS) $(IMAGE_IO_LIBS) -o $@
 
-# Run clang-format on most of the source. The tutorials directory is
-# explicitly skipped, as those files are manually formatted to
-# maximize readability. NB: clang-format is *not* stable across versions;
-# we are currently standardized on the formatting from clang-format-10.
-# If CLANG_FORMAT points to a different version, you may get incorrectly-formatted code.
-CLANG_FORMAT ?= ${CLANG}-format
-
+# Note: you must have CLANG_FORMAT_LLVM_INSTALL_DIR set for this rule to work.
```
abadams, line 2242: Can we just set it to $(shell dirname $(CLANG)) here?

steven-johnson: clang-format isn't compatible between versions -- our current formatting is based on clang-format-10 and using other versions will produce slightly different results. So if you building with LLVM12 that would give you clang-format-12 and you'd get skewed formatting. (Yes, this is unfortunate, but the stated position of the maintainers is that they will never promise inter-version compatibility.)

run-clang-tidy.sh
```diff
@@ -0,0 +1,56 @@
...
+echo CLANG_TIDY_LLVM_INSTALL_DIR = ${CLANG_TIDY_LLVM_INSTALL_DIR}
+
+# Use a temp folder for the CMake stuff here, so it's fresh & correct every time
+CLANG_TIDY_BUILD_DIR=`mktemp -d`
+echo CLANG_TIDY_BUILD_DIR = ${CLANG_TIDY_BUILD_DIR}
+
+echo Building compile_commands.json...
+cmake -DCMAKE_BUILD_TYPE=Debug \
+      -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+      -DLLVM_DIR=${CLANG_TIDY_LLVM_INSTALL_DIR} \
```
abadams, line 27: I thought LLVM_DIR needs to be location of the cmake sub-folder, rather than the top-level installation dir?

steven-johnson: um... yeah, it does. How TF is this working?

steven-johnson:
Gentle review ping.

Makefile
abadams: If you have the ubuntu packages clang-format-10 and clang-tidy-10 installed, this should be set to /usr/lib/llvm-10

We could use it as the default. Or maybe just add a comment to that effect?

steven-johnson: Sure, those seem like reasonable defaults.

abadams: Also make variables don't become environment variables of shell scripts that you invoke. You may need CLANG_FORMAT_LLVM_INSTALL_DIR=$(CLANG_FORMAT_LLVM_INSTALL_DIR) before the shell script invocation

steven-johnson: Whoops.

abadams: Seems like it's not finding the .clang-format file and then doing nothing. Investigating...

abadams: That wasn't it - I was confusing myself. The find invocation is discovering emacs backup files and getting confused. Can we replace the find with "git ls-files *.cpp *.c *.h"?

steven-johnson: > Can we replace the find with "git ls-files *.cpp *.c *.h"?

Can we replace emacs with a modern editor? :-)

More seriously, `git ls-files` doesn't seem to support the `**` globstar operator, so the irregular subdir pattern in some places (eg apps) means we'll either miss some files or have to have a complex set of dirs to pass to it (which will get stale).

Does the emacs backup file setup follow a naming pattern that we can exclude from the `find`?

abadams: Yes, but I was wondering if it was good to exclude files not in the repo, so I went with the git ls-files option first. The emacs exclude would be any file starting with a .

These args to find exclude dot files:

`\( -name *.cpp -o -name *.h -o -name *.c \) -and -not -wholename "*/.*"`

steven-johnson: PTAL

Makefile
```diff
@@ -2242,51 +2242,25 @@ $(BIN_DIR)/HalideTraceViz: $(ROOT_DIR)/util/HalideTraceViz.cpp $(INCLUDE_DIR)/Ha
...
-# If CLANG_FORMAT points to a different version, you may get incorrectly-formatted code.
-CLANG_FORMAT ?= ${CLANG}-format
+# Note: you must have CLANG_FORMAT_LLVM_INSTALL_DIR set for this rule to work.
+# Let's default to the Ubuntu install location.
+CLANG_FORMAT_LLVM_INSTALL_DIR ?= /usr/lib/llvm-10
 
 .PHONY: format
 format:
-	find "${ROOT_DIR}/apps" "${ROOT_DIR}/src" "${ROOT_DIR}/tools" "${ROOT_DIR}/test" "${ROOT_DIR}/util" "${ROOT_DIR}/python_bindings" -name *.cpp -o -name *.h -o -name *.c | xargs ${CLANG_FORMAT} -i -style=file
+	CLANG_FORMAT_LLVM_INSTALL_DIR=$(CLANG_FORMAT_LLVM_INSTALL_DIR) @${ROOT_DIR}/run-clang-format.sh
```
abadams, line 2251: The @ symbol doesn't work after the env var

Makefile
```diff
@@ -2242,51 +2242,25 @@ $(BIN_DIR)/HalideTraceViz: $(ROOT_DIR)/util/HalideTraceViz.cpp $(INCLUDE_DIR)/Ha
...
-	echo "  \"file\": \"$$S\" }]" >> $@; \
+# Note: you must have CLANG_TIDY_LLVM_INSTALL_DIR set for these rules to work.
+# Let's default to the Ubuntu install location.
+CLANG_TIDY_LLVM_INSTALL_DIR ?= /usr/lib/llvm-10
 
 .PHONY: clang-tidy
-clang-tidy: $(BUILD_DIR)/compile_commands.json
-	@$(INVOKE_CLANG_TIDY) 2>&1 | grep -v "warnings generated" | grep -v '^$(CLANG)-tidy '
+clang-tidy:
+	CLANG_TIDY_LLVM_INSTALL_DIR=$(CLANG_TIDY_LLVM_INSTALL_DIR) @${ROOT_DIR}/run-clang-tidy.sh
```
abadams, line 2259: same issue here

Makefile
steven-johnson: ack, oops

abadams:
There were two more trivial tweaks I had to make to get it to work, so I just pushed them.

abadams (approved):

PR #5350, opened 2020-10-12
bounds inference (mod) pessimistic on unsigned 0 interval

rootjalex:
Fix discussed with @abadams and @alexreinking . Original code takes `b = [0u, 0u]` and produces an interval of `[0, b.type().max()]` due to wrap-around. Technically correct but pessimistic/not tight. This fix removes that case and simplifies the overall generated expression `interval.max` by removing the `max()`

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -646,9 +646,12 @@ private:
             }
         } else {
             // b is bounded
-            if (b.max.type().is_uint() || (b.max.type().is_int() && is_positive_const(b.min))) {
-                // If the RHS is a positive integer, the result is in [0, max_b-1]
-                interval.max = Max::make(interval.min, b.max - make_one(t));
+            if ((b.max.type().is_int() || b.max.type().is_uint()) && is_positive_const(b.min)) {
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -646,9 +646,12 @@ class Bounds : public IRVisitor {
             }
         } else {
             // b is bounded
-            if (b.max.type().is_uint() || (b.max.type().is_int() && is_positive_const(b.min))) {
-                // If the RHS is a positive integer, the result is in [0, max_b-1]
-                interval.max = Max::make(interval.min, b.max - make_one(t));
+            if ((b.max.type().is_int() || b.max.type().is_uint()) && is_positive_const(b.min)) {
```
steven-johnson, line 649: FYI, `is_int_or_uint()` exists

rootjalex: thanks, just changed!

steven-johnson:
Only failure is unrelated LLVM breakage, good to land

abadams (approved):

PR #5351, opened 2020-10-12
Clone #5350 for buildbots

abadams:


steven-johnson:
Only failure is unrelated LLVM breakage, good to land

PR #5358, opened 2020-10-13
Extend clang-tidy checks into  tools, utils, and python_bindings

steven-johnson:
(and also fix the reported findings)

Note that this also extends the testing to src/autoschedulers, but *not* to src/runtime (yet). It would likely be desirable to do so at some point, but requires a bit more massaging of the build scripts.

Extending to apps/ and test/ might be nice, but would require running all Generators to produce the needed .h files.

### DIFF
--- a/src/autoschedulers/adams2019/cost_model_generator.cpp
+++ b/src/autoschedulers/adams2019/cost_model_generator.cpp
@@ -2,6 +2,8 @@
 // templated such that it can be compiled in either forward or
 // backwards mode, for inference or training respectively.
 
+#include <utility>
+
 #include "Halide.h"
 
 #include "NetworkSize.h"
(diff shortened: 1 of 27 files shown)


src/autoschedulers/adams2019/cost_model_generator.cpp
```diff
@@ -71,7 +80,7 @@ struct ModelWeight<true> : public GeneratorInput<Buffer<float>> {
         Expr smoothed_second_moment_correction = 1 / (1 - pow(0.999f, timestep + 1));
 
         // Update the weights
-        Expr step = learning_rate * smoothed_deriv * smoothed_deriv_correction;
+        Expr step = std::move(learning_rate) * smoothed_deriv * smoothed_deriv_correction;
```
abadams, line 83: I think the changes in this file (particularly the std::moves) are harmful for readability. Several of our clang checks are (very very *very* slightly) good for performance but bad for readability, and they shouldn't be on in non-performance-critical code (which includes just about all of use-of-Halide-as-a-language code)

steven-johnson: ...wtf? yikes. I didn't even notice that (I never thought to check for that sort of craziness). Yeah, that's some weird stuff there. Let me see what I can do.

steven-johnson:
Still fighting with clang-tidy and the nutty way it (doesn't) let you exclude third-party headers properly. Needs more work (hence conversion to draft). Gonna revisit later.

steven-johnson:
Finally beat this into submission. PTAL

abadams (approved):

PR #5359, opened 2020-10-13
Handle more types of nested vectorized += without scalarizing

abadams:
With this change, in a nesting of vectorized vars in an
associative/commutative reduction (e.g. +=), we can now have a reduction
var outermost and a reduction var innermost and get good codegen. This
is still not fully general - there can only be one pure var in the
vectorized stack for it to work. In general is_interleaved_ramp should
be an is_tensor_contraction pass that knows how to do clever codegen for
those.

For the following schedule for a 16-bit large convolution:

```
prod.compute_at(result, x)
  .vectorize(x)
  .update()
  .split(r, ro, ri, 8)
  .split(ri, rio, rii, 2)
  .reorder(rii, x, rio, ro)
  .vectorize(x)
  .atomic()
  .vectorize(rio)
  .vectorize(rii);
```

We get the following IR:

```
let t2262 = (int32x32)vector_reduce(Add, (int32x64(shuffle((int16x15)p10_im_global_wrapper$0[ramp(0, 1, 15)], 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10, 11, 11, 12, 6, 7, 7, 8, 8, 9, 9, 10, 10, 11, 11, 12, 12, 13, 13, 14)) *
                                          int32x64(shuffle((int16x8)p11_im_global_wrapper$0[ramp(0, 1, 8)], 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7))))

f58[ramp(0, 1, 8)] = slice_vectors(t2262, 8, 1, 8) + (slice_vectors(t2262, 0, 1, 8) + (slice_vectors(t2262, 16, 1, 8) + (slice_vectors(t2262, 24, 1, 8) + f58[ramp(0, 1, 8)])))

```

The vector_reduce is from rii, and the sum of slice_vectors is from rio.

The generated asm (for avx512) is:

```
	leal	(%rbx,%rcx), %edx
	movslq	%edx, %rdx
	subq	%r14, %rdx
	vmovdqu	(%r15,%rcx,2), %xmm6
	vbroadcasti64x4	(%r12,%rdx,2), %zmm7 # zmm7 = mem[0,1,2,3,0,1,2,3]
	vpermw	%zmm7, %zmm0, %zmm8
	vpermw	%zmm7, %zmm1, %zmm7
	vpermd	%zmm6, %zmm2, %zmm6
	vpermd	%zmm6, %zmm3, %zmm9
	vpmaddwd	%zmm9, %zmm8, %zmm8
	vpermd	%zmm6, %zmm4, %zmm6
	vpmaddwd	%zmm6, %zmm7, %zmm6
	vextracti64x4	$1, %zmm6, %ymm7
	vextracti64x4	$1, %zmm8, %ymm9
	vpaddd	%ymm5, %ymm8, %ymm5
	vpaddd	%ymm6, %ymm5, %ymm5
	vpaddd	%ymm5, %ymm9, %ymm5
	vpaddd	%ymm7, %ymm5, %ymm5
	addq	$8, %rcx
	cmpq	$128, %rcx
```

The pmaddwds are from rii, and the vpaddds are from rio.

This is 3.5x faster than the best schedule that only vectorizes the pure
var, and about 2x faster than the best schedule that only vectorizes the 
pure var and an unsplit reduction variable.

### DIFF
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -182,11 +182,17 @@ Interval bounds_of_lanes(const Expr &e) {
     }
 };
 
-// A ramp with the lanes repeated (e.g. <0 0 2 2 4 4 6 6>)
+// A ramp with the lanes repeated inner_repetitions times, and then
+// the whole vector repeated outer_repetitions times.
+// E.g: <0 0 2 2 4 4 6 6 0 0 2 2 4 4 6 6>.
 struct InterleavedRamp {
     Expr base, stride;
-    int lanes, repetitions;
+    int lanes, inner_repetitions, outer_repetitions;
 };
+
+bool equal_or_zero(int a, int b) {
+    return a == 0 || b == 0 || a == b;
+}
 
 bool is_interleaved_ramp(const Expr &e, const Scope<Expr> &scope, InterleavedRamp *result) {
     if (const Ramp *r = e.as<Ramp>()) {
@@ -196,13 +202,16 @@ bool is_interleaved_ramp(const Expr &e, const Scope<Expr> &scope, InterleavedRam
             result->base = r->base;
             result->stride = r->stride;
             result->lanes = r->lanes;
-            result->repetitions = 1;
+            result->inner_repetitions = 1;
+            result->outer_repetitions = 1;
             return true;
         } else if (b_base && b_stride && b_base->lanes == b_stride->lanes) {
+            // Ramp of broadcast
             result->base = b_base->value;
             result->stride = b_stride->value;
             result->lanes = r->lanes;
-            result->repetitions = b_base->lanes;
+            result->inner_repetitions = b_base->lanes;
+            result->outer_repetitions = 1;
             return true;
         }
     } else if (const Broadcast *b = e.as<Broadcast>()) {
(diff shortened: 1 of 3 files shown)


src/VectorizeLoops.cpp
```diff
@@ -196,49 +202,53 @@ bool is_interleaved_ramp(const Expr &e, const Scope<Expr> &scope, InterleavedRam
...
-             result->repetitions == 0 ||
-             ra.repetitions == result->repetitions)) {
+            equal_or_zero(ra.inner_repetitions, result->inner_repetitions) &&
+            equal_or_zero(ra.outer_repetitions, result->outer_repetitions)) {
             result->base = simplify(result->base + ra.base);
             result->stride = simplify(result->stride + ra.stride);
-            if (!result->repetitions) {
-                result->repetitions = ra.repetitions;
-            }
+            result->inner_repetitions = std::max(result->inner_repetitions, ra.inner_repetitions);
```
dsharletg, line 238: Don't these need to be equal?

dsharletg (approved):

src/VectorizeLoops.cpp
dsharletg: Oh, I see they are zero if they aren't equal.

test/performance/nested_vectorization_gemm.cpp
```diff
@@ -219,6 +219,85 @@ int main(int argc, char **argv) {
...
+
+            times[use_nested_vectorization] =
+                Tools::benchmark(10, 10, [&]() {
+                    result.realize(out);
+                    out.device_sync();
+                });
+        }
+
+        double speed_up = times[0] / times[1];
+        printf("8-bit blur with reduction dimension outermost vector dim\n"
```
vksnk, line 289: Should be 16-bit instead of 8-bit?

abadams: Thanks, fixed.

abadams:
Updated to use a binary reduction tree for the outer repetitions instead of a linear scan. PTAAL

abadams:
New inner loop (which handles the same number of multiply-adds as the old one, but in an 16x4 tile instead of an 8x8 one:

```
	vmovdqu64	(%rcx,%rdx,8), %zmm5
	vmovq	(%r14,%rdx,8), %xmm6    # xmm6 = mem[0],zero
	vpermw	%zmm5, %zmm0, %zmm7
	vpermw	%zmm5, %zmm1, %zmm5
	vpermd	%zmm6, %zmm2, %zmm8
	vpermd	%zmm8, %zmm3, %zmm8
	vpmaddwd	%zmm8, %zmm7, %zmm7
	vpbroadcastd	%xmm6, %zmm6
	vpmaddwd	%zmm6, %zmm5, %zmm5
	vpaddd	%zmm5, %zmm4, %zmm4
	vpaddd	%zmm7, %zmm4, %zmm4
	incq	%rdx
	cmpq	$32, %rdx
	jne	.LBB0_23
```

steven-johnson:
Looks like there are some performance-test failures on arm64

steven-johnson:
The linux-10 build failure is a CUDA_ERROR_OUT_OF_MEMORY -- not sure if significant or a flake.

The linux-12 failure is the overnight llvm trunk failure and is unrelated (should go away if you sync to head).

abadams:
I double checked - the allocation pattern is the same as master, so I think it's a flake.

steven-johnson:
Looks good to land.

abadams:
I shouldn't be worried about this timeout: https://buildbot.halide-lang.org/master/#/builders/52/builds/43 ?

steven-johnson:
Oh, wait, hm, yeah, I thought that was just the llvm-broken thing -- that is weird. Let me take a look.

steven-johnson:
So I saw a similar failure on a different current PR -- I'm theorizing that the wasm tests were close to the timeout limit before and some upstream change just put them (sometimes) over the limit. Testing that theory now.

steven-johnson:
OK, it looks like `generator_aot_shuffler` is hanging for wasm builds at our current master -- not sure yet whether the bug is ours or LLVMs or Emscriptens or what but it's definitely not your code, so you are good to go here.

Issue #5363, opened 2020-10-15
generator_aot_shuffler is hanging under WASM builds


steven-johnson:

steven-johnson:
Can't repeat at current top-of-tree for LLVM 11, 12, or 13; closing as not repeatable.

PR #5364, opened 2020-10-15
Fix wasm-related glitches in our timing/benchmarking code (Issue #5363)

steven-johnson:


abadams:
The wasm clock can tick backwards?

steven-johnson:
> The wasm clock can tick backwards?

Yep. (So can any of the std::chrono ones, too, unless they are "steady" -- even the high-resolution one.)

abadams (approved):

PR #5365, opened 2020-10-15
Issue #3925 : Remove hvx_64

### LINKED ISSUE #3925: Should HVX_64 support be deprecated?
It's not clear that there is any active use of HVX_64 (as opposed to HVX_128), and any new users wouldn't start there. Should we deprecate-and-remove it? If nothing else, it would allow us to drop buildbot testing of it.

pranavb-ca:
- Removes hvx_64
- Adds a target feature ``hvx`` which defaults to ``Target::HVX_128``

I think we should keep ``hvx_128`` around for a little bit and before ripping it out and simply going by ``hvx``

fixes #3925


apps/hexagon_benchmarks/median3x3_generator.cpp
```diff
@@ -39,8 +39,8 @@ class Median3x3 : public Generator<Median3x3> {
         output.dim(0).set_min(0);
         output.dim(1).set_min(0);
 
-        if (get_target().features_any_of({Target::HVX_64, Target::HVX_128})) {
-            const int vector_size = get_target().has_feature(Target::HVX_128) ? 128 : 64;
+        if (get_target().features_any_of({Target::HVX, Target::HVX_128})) {
+            const int vector_size = get_target().features_any_of({Target::HVX, Target::HVX_128}) ? 128 : 64;
```
dsharletg, line 43: constant 128

dsharletg (changes requested):
Two things that affect a ton of instances:
- Should we just make `Target::HVX_128` == `Target::HVX`, so we can simplify all instances that check these to just check `Target::HVX`?
- There are a lot of instances of `vector_size = get_target()... ? 128 : 64` that should just be `vector_size = 128` now?

pranavb-ca:
> Two things that affect a ton of instances:
> 
> * Should we just make `Target::HVX_128` == `Target::HVX`, so we can simplify all instances that check these to just check `Target::HVX`?
> * There are a lot of instances of `vector_size = get_target()... ? 128 : 64` that should just be `vector_size = 128` now?

dang it - i thought I caught all of those instances. I grepped for HVX_64 and that's why I think I missed out some of those. Will do.

pranavb-ca:
> Two things that affect a ton of instances:
> 
> * Should we just make `Target::HVX_128` == `Target::HVX`, so we can simplify all instances that check these to just check `Target::HVX`?

Actually that is what I have done, but in a sed frenzy I went ahead and changed ``has_feature`` to ``features_any_of``

src/Target.h
```diff
@@ -93,7 +93,7 @@ struct Target {
         CPlusPlusMangling = halide_target_feature_c_plus_plus_mangling,
         LargeBuffers = halide_target_feature_large_buffers,
         HexagonDma = halide_target_feature_hexagon_dma,
-        HVX_64 = halide_target_feature_hvx_64,
+        HVX = halide_target_feature_hvx_128,
```
pranavb-ca, line 96: @dsharlet - ``Target::HVX == Target::HVX_128``

apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -504,26 +500,20 @@ void CameraPipe::generate() {
...
-        //and in HVX 64 we need 4 threads, and on other devices,
-        // we might need many threads.
+        // In HVX 128, we need 4 threads to saturate HVX with work,
+        // on other devices, we might need many threads.
         Expr strip_size;
-        if (get_target().has_feature(Target::HVX_128)) {
-            strip_size = processed.dim(1).extent() / 2;
-        } else if (get_target().has_feature(Target::HVX_64)) {
+        if (get_target().has_feature(Target::HVX)) {
             strip_size = processed.dim(1).extent() / 4;
```
dsharletg, line 507: Keep the HVX_128 branch, not the HVX_64 branch.

src/CodeGen_Hexagon.cpp
```diff
@@ -46,17 +46,15 @@ CodeGen_Hexagon::CodeGen_Hexagon(Target t)
     } else {
         isa_version = 62;
     }
-    user_assert(!target.features_all_of(
-        {Halide::Target::HVX_128, Halide::Target::HVX_64}))
-        << "Cannot set both HVX_64 and HVX_128 at the same time.\n";
+    user_assert(target.has_feature(Target::HVX))
+        << "Creating a Codegen target for Hexagon without the hvx_128 or the hvx target feature.\n";
```
dsharletg, line 50: Remove hvx_128 from this message?

src/Target.cpp
```diff
@@ -871,13 +870,10 @@ int Target::natural_vector_size(const Halide::Type &t) const {
 
     if (arch == Target::Hexagon) {
         if (is_integer) {
-            // HVX is either 64 or 128 *byte* vector size.
-            if (has_feature(Halide::Target::HVX_128)) {
+            if (features_any_of({Halide::Target::HVX, Halide::Target::HVX_128})) {
```
dsharletg, line 873: features(...)

src/Target.cpp
```diff
@@ -966,7 +962,7 @@ bool Target::get_runtime_compatible_target(const Target &other, Target &result)
     }
 
     if ((features & matching_mask) != (other.features & matching_mask)) {
-        Internal::debug(1) << "runtime targets must agree on SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX_64, HVX_128, HexagonDma, and HVX_shared_object\n"
+        Internal::debug(1) << "runtime targets must agree on SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX_128, HexagonDma, and HVX_shared_object\n"
```
dsharletg, line 965: HVX_128 -> HVX

src/Target.cpp
```diff
@@ -1054,8 +1050,8 @@ void target_test() {
         {{"x86-64-linux-cuda", "x86-64-linux-opengl", "x86-64-linux-cuda-opengl"}},
         {{"hexagon-32-qurt-hvx_v65", "hexagon-32-qurt-hvx_v62", "hexagon-32-qurt-hvx_v62"}},
         {{"hexagon-32-qurt-hvx_v62", "hexagon-32-qurt", "hexagon-32-qurt"}},
-        {{"hexagon-32-qurt-hvx_v62-hvx_64", "hexagon-32-qurt", ""}},
-        {{"hexagon-32-qurt-hvx_v62-hvx_64", "hexagon-32-qurt-hvx_64", "hexagon-32-qurt-hvx_64"}},
+        {{"hexagon-32-qurt-hvx_v62-hvx_128", "hexagon-32-qurt", ""}},
```
dsharletg, line 1053: change to hvx?

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -13,7 +13,7 @@
 // simd_op_check into two tests, simd_op_check.cpp and simd_op_check_hvx.cpp
 // so that the latter is free to do its own thing - for simd_op_check_hvx.cpp
 // to run any tests, all that is needed is that HL_TARGET have a HVX related
-// target feature, i.e. one of HVX_64, HVX_128, HVX_v62, HVX_v65 and HVX_v66.
+// target feature, i.e. one of HVX_128, HVX_v62, HVX_v65 and HVX_v66.
```
dsharletg, line 16: HVX_128 -> HVX?

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -663,8 +658,7 @@ int main(int argc, char **argv) {
     printf("HL_TARGET is: %s\n", hl_target.to_string().c_str());
 
     Target t(Target::NoOS, Target::Hexagon, 32);
-    for (const auto &f : {Target::HVX_64,
-                          Target::HVX_128,
+    for (const auto &f : {Target::HVX_128,
```
dsharletg, line 661: HVX_128 -> HVX?

apps/camera_pipe/camera_pipe_generator.cpp
pranavb-ca: I jumped the gun a little bit here. Newer cores have 4 contexts, but presently the lowest we support still has 2 contexts. I'll switch it back to `` / 2 ;``

dsharletg: Oh, I see. Actually, I think having more strips than cores is fine, so maybe keeping this 4 is best.

src/CodeGen_Hexagon.cpp
pranavb-ca: I left it in the message to indicate to users that ``hvx_128`` is still ok to use as a feature.

src/HexagonOffload.cpp
```diff
@@ -972,6 +972,7 @@ Stmt inject_hexagon_rpc(Stmt s, const Target &host_target,
     static const Target::Feature shared_features[] = {
         Target::Profile,
         Target::NoAsserts,
+        Target::HVX,
         Target::HVX_128,
```
dsharletg, line 976: Don't need this?

apps/camera_pipe/camera_pipe_generator.cpp
```diff
@@ -500,11 +500,15 @@ void CameraPipe::generate() {
...
         Expr out_height = processed.height();
 
-        // In HVX 128, we need 4 threads to saturate HVX with work,
-        // on other devices, we might need many threads.
+        // Depending on the HVX generation, we need 2 or 4 threads
+        // to saturate HVX with work.
         Expr strip_size;
         if (get_target().has_feature(Target::HVX)) {
-            strip_size = processed.dim(1).extent() / 4;
+            if (get_target().features_any_of({Target::HVX_v65, Target::HVX_v66})) {
```
dsharletg, line 507: Should this condition be negated, so its future proof?

Also, doesn't this vary between generations? I think maybe we should just unconditionally use 4. It's not a big deal to oversubscribe the contexts, maybe overhead is negligibly higher? But at the same time, we often split more than necessary anyways to get the working sets small enough to fit in the cache.

dsharletg (commented):
I don't know why I can't reply to any of the existing threads on the "conversation" tab... but regarding: 

> I left it in the message to indicate to users that hvx_128 is still ok to use as a feature.

If it's going to be deprecated, wouldn't it make sense to just encourage use of `hvx`?

apps/camera_pipe/camera_pipe_generator.cpp
pranavb-ca: You are right. To keep things simple, I'll switch the code to use 4 threads always.

pranavb-ca:
@dsharletg  - Addressed comments from the last review

dsharletg (approved):
Thanks for the fixes!

steven-johnson (approved):

steven-johnson:
LGTM -- pushed a partial change to the buildbot to avoid the hvx64 build errors, then synced this to master. Let's let it go green before landing.

dsharletg:
This is failing on some of the buildbots:

```
make: *** No rule to make target 'bin/build/initmod.hvx_64_ll.o', needed by 'bin/libHalide.so'.  Stop.
```

Looking at the PR, I don't see any changes to either Makefile or CMakeLists.txt, both should probably change.

pranavb-ca:
https://github.com/halide/build_bot/pull/108

dsharletg:
I think that's a different thing. Yes, the hvx_64 tests will fail to run. But here, libHalide.so is failing to link at all. I think because these two lines need to be removed:

https://github.com/halide/Halide/blob/master/Makefile#L829
https://github.com/halide/Halide/blob/master/src/runtime/CMakeLists.txt#L105

pranavb-ca:
> I think that's a different thing. Yes, the hvx_64 tests will fail to run. But here, libHalide.so is failing to link at all. I think because these two lines need to be removed:
> 
> https://github.com/halide/Halide/blob/master/Makefile#L829
> https://github.com/halide/Halide/blob/master/src/runtime/CMakeLists.txt#L105

Yes, you are right. I was simply reminded of my PR from yesterday because of your comment and so pasted that here. I had taken care of CMakeLists.txt in a previous commit. Just pushed one more change to remove it from Halide/Makefile

alexreinking:
> I think we should keep `hvx_128` around for a little bit and before ripping it out and simply going by `hvx`

If we hold off merging this until v11.0.0 is released, then we don't need to wait at all.

dsharletg:
This has passed all the build bots, @alexreinking is there really a reason to wait to merge this? It isn't a breaking change (unless you are using hvx_64 of course). hvx_128 remains as a useful flag, it's just deprecated (and it is only silently deprecated at this point). I think the steps here are:

1. Remove `hvx_64` (this PR)
2. Deprecate `hvx_128` by warning people to use `hvx` instead.
3. Remove `hvx_128` and only support `hvx`.

I think we should do 1 now that this PR is ready to merge, and avoid generating inevitable merge conflicts for this PR to manage later.

2 and 3 aren't planned for any particular time yet. 2 and 3 should probably be in separate releases of course.

alexreinking:
It's totally fine to put this through a deprecation plan. I thought this PR might not be ready for the v11.0 release, but it looks like I misread the room :)

dsharletg:
OK, I think I'll let @pranavb-ca decide when to merge it. I was also mistaken, some of the build bots are still running.

steven-johnson:
I agree with @dsharlet -- happy to go ahead and land before branching the 11 release. Buildbots are clean, there's just one merge conflict to resolve.

steven-johnson:
UPDATE: the merge conflict was due to my clang-tidy changes, so I went ahead and pushed the fix. Should be good to merge anytime.

steven-johnson:
Green, OK to land

PR #5366, opened 2020-10-16
Add simplifications related to vector reductions and broadcasts

dsharletg:


src/IR.cpp
```diff
@@ -821,6 +821,16 @@ bool Shuffle::is_interleave() const {
     return true;
 }
 
+int Shuffle::is_broadcast() const {
+    int lanes = vectors.front().type().lanes();
+    for (int i = 0; i < (int)indices.size(); i++) {
```
abadams, line 826: Based on the use of is_broadcast, also need to check vectors.size() == 1 and indices.size() % vectors[0].type().lanes() == 0

dsharletg:
It seems like something is broken with the clang-tidy check. I've retried it and it is repeatedly failing on what looks like a tooling issue. And it appears to be blocking the buildbots from running...

dsharletg:
Oh, it looks like the installation of clang-tidy itself is failing. I guess I'll just try it tomorrow.

src/IR.h
```diff
@@ -778,6 +778,11 @@ struct Shuffle : public ExprNode<Shuffle> {
     }
     ///@}
 
+    /** Check if this shuffle is a broadcast of the vector arguments.
+     * returns 0 if the shuffle is not a broadcast, otherwise returns
+     * the broadcast factor. */
+    int is_broadcast() const;
```
vksnk, line 784: I think it's a bit confusing when function called is_* returns something other than bool (I can't suggest a very good alternative name though).

dsharletg: I thought about this too, but I figured it'd be OK because this does actually work as a bool. 0 is false, and anything else is true, and I'm pretty sure it works as part of any larger boolean expression.

src/Simplify_Exprs.cpp
```diff
@@ -53,6 +53,36 @@ Expr Simplify::visit(const Broadcast *op, ExprInfo *bounds) {
 
 Expr Simplify::visit(const VectorReduce *op, ExprInfo *bounds) {
     Expr value = mutate(op->value, bounds);
+    if (const Shuffle *shuffle = value.as<Shuffle>()) {
```
abadams, line 56: I find the control flow here quite confusing. Is the idea just that sometimes you can move the vector reduce inside the shuffle, and sometimes that makes the shuffle go away? If so I think this is more general than is_broadcast, and you can do away with is_broadcast:

Given a reduction factor f, I think you can move the reduce inside the shuffle if indices[i] == (indices[i - i%f]/f) * f + (i % f)

E.g. if you reduce by a factor of four, and the shuffle has indices [4 5 6 7 0 1 2 3 12 13 14 15] then you can move the reduce inside the shuffle, and rewrite the shuffle indices to be every fourth one divided by four: [1 0 3]

In the case where the shuffle indices end up being [0 1 2 3 ... lanes - 1], you don't bother adding the shuffle.

This is different to what we discussed, which was that for min/max/and/or total reductions (output lanes = 1) you can dedup and sort the indices of the inner shuffle, and this is a good idea if it results in a nicer indexing pattern (e.g. a slice)

abadams: Oh I guess the first thing I wrote doesn't handle a total reduction of a shuffle with indices [0 1 0 1 0 1 0 1]. I'm still confused about this though. Can you add a worked example to the comments maybe?

src/Simplify_Exprs.cpp
```diff
@@ -53,6 +53,36 @@ Expr Simplify::visit(const Broadcast *op, ExprInfo *bounds) {
...
+        int broadcast_factor = shuffle->is_broadcast();
+        if (broadcast_factor > 1) {
+            // How many lanes does the smaller vector reduce have?
+            int reduced_result_lanes = std::max(1, op->type.lanes() / broadcast_factor);
+
+            Expr reduced = VectorReduce::make(op->op, shuffle->vectors[0], reduced_result_lanes);
+            if (reduced_result_lanes < op->type.lanes()) {
+                reduced = Shuffle::make_broadcast(reduced, op->type.lanes() / reduced_result_lanes);
+            }
+            if (reduced.type().lanes() != op->type.lanes()) {
```
abadams, line 66: When can these things be equal and not one? Does this handle vector reductions that aren't total?

src/IR.h
steven-johnson: I'd still think the name `broadcast_factor()` might be better

src/Simplify_Exprs.cpp
dsharletg: I think the general idea of what you described would be to check if each of the indices[i/f...i/f + f - 1] is a permutation of indices that are aligned to a multiple of f. If so, you can move the shuffle outside, downsample the indices by f, and divide them by f.

This is more complicated/general, but might actually be easier to understand than my current less general algorithm. Let me try it and see if I can implement it.

dsharletg: Here is an implementation of what you proposed (I think): https://github.com/halide/Halide/compare/simplify-vectorreduce-shuffles2?expand=1

However, this is really a different thing than what this PR implements.

I'm also not sure if we should do this. I'm concerned it will break code that would be faster because the schedule was trying to hit a particular pattern.

I have another different PR that I think does what this PR does better: #5372

dsharletg:
Closing this in favor of #5372, and https://github.com/halide/Halide/compare/simplify-vectorreduce-shuffles2?expand=1 implements one of the ideas discussed in this PR, but is not quite the same thing.

PR #5369, opened 2020-10-16
Backport everything for 10.0.1

alexreinking:
Did a bunch of `git cherry-pick -m 1`'s to merge all the marked PRs. If we're going to do this more, we should be sure to always use squash/merge or rebase.


steven-johnson:
> If we're going to do this more, we should be sure to always use squash/merge or rebase.

I'd be in favor of that regardless, it makes the commit history much saner

steven-johnson (dismissed):

alexreinking:
> I'd be in favor of that regardless, it makes the commit history much saner

alexreinking:
Not sure why there are LLVM 12 failures, but this is Halide 10, so we don't actually need LLVM 12 support. Just need to double check this isn't a more serious bug, or maybe backport some more changes.

abadams:
The llvm 12 failure is because the name of an intrinsic changed in llvm 12, and this branch hasn't been updated to account for it. We could backport that fix just to keep the branch clean, or we could ignore failures from newer llvm versions. I think backporting the fix might actually be pretty easy. I'll try to find it...

abadams:
It was #5341

alexreinking:
> 
> 
> It was #5341

I backported that one too

abadams:
Looks like the next failure is in the same category. I guess we should label all "fix for llvm trunk" commits to be backported if we're going to keep the branch clean against trunk: #5362

alexreinking:
> Looks like the next failure is in the same category. I guess we should label all "fix for llvm trunk" commits to be backported if we're going to keep the branch clean against trunk: #5362

Cherry picked that one too. We should probably not really bother with this for v10 going forward. v10 supports LLVM 9.0-11.0.

abadams (approved):

steven-johnson:
> We could backport that fix just to keep the branch clean, or we could ignore failures from newer llvm versions

I'd vote for ignoring those.

PR #5370, opened 2020-10-17
Make loop partitioning a bit more robust for if statements

dsharletg:
The other visitors in FindSimplifications all use `has_likely_tag`, which searches expressions for `likely`. If did not however, which made loop partitioning brittle, e.g. in the case of `if ((uint1)likely(x))`.

### DIFF
--- a/src/PartitionLoops.cpp
+++ b/src/PartitionLoops.cpp
@@ -361,8 +361,7 @@ class FindSimplifications : public IRVisitor {
         // statement is marked as likely, treat it as likely true and
         // partition accordingly.
         IRVisitor::visit(op);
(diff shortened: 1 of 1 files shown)


dsharletg:
This seems like it might be wrong for things like if (!likely(x))...? Not sure what the right implementation is to avoid the problem I mentioned in the PR description.

abadams:
I think you might want has_uncaptured_likely_tag.

E.g. what do you want to happen for:

```if (min(likely(x), y) < z) {```

abadams (approved):

dsharletg:
The breakage is an LLVM breakage already fixed.

PR #5376, opened 2020-10-19
Change NULL -> nullptr

steven-johnson:
enable the modernize-use-nullptr check in clang-tidy and fix all complaints.

(Note that we used to require the use of NULL in the runtime code since it didn't use C++11, but we now compile the runtime code with C++11)


zvookin (approved):

src/CodeGen_C.cpp
```diff
@@ -2053,7 +2053,7 @@ void CodeGen_C::visit(const Call *op) {
         if (op->args.empty()) {
             internal_assert(op->type.handle_type);
             // Add explicit cast so that different structs can't cache to the same value
-            rhs << "(" << print_type(op->type) << ")(NULL)";
+            rhs << "(" << print_type(op->type) << ")(nullptr)";
```
abadams, line 2056: Do we want this one?

src/PythonExtensionGen.cpp
```diff
@@ -111,7 +111,7 @@ void PythonExtensionGen::convert_buffer(const string &name, const LoweredArgumen
     dest << /*name*/ "\"" << name << "\"";
     dest << ") < 0) {\n";
     release_buffers("        ");
-    dest << "        return NULL;\n";
```
abadams, line 114: These quoted ones also worth considering.

abadams:
Only open question here is if our machine-generated C++ should or already does require c++11

abadams (approved):

src/CodeGen_C.cpp
steven-johnson: Our generated C++ code already requires C++11 (or later), so, sure

src/PythonExtensionGen.cpp
steven-johnson: Since our C++ backend already requires C++11, I think it's fair to assume that the Python extensions should, too. But can revert these if you like.

abadams: Seems reasonable. Just making sure we're doing this intentionally.

steven-johnson:
> Only open question here is if our machine-generated C++ should or already does require c++11

See https://github.com/halide/Halide/pull/5332

src/CodeGen_C.cpp
dsharletg: Isn't this used by e.g. OpenCL and it's variety of compilers of varying quality?

I'd suggest simply using 0 here.

steven-johnson: good point. (though I wouldn't be surprised at a barf at that from dubious OpenCL compilers too.) Maybe swapping back to NULL would be safest since we know it works.

steven-johnson: Swapped back to NULL.

PR #5379, opened 2020-10-20
Enable clang-tidy's modernize-use-default-member-init check 

steven-johnson:
...and fix the resulting code.

### DIFF
--- a/src/AssociativeOpsTable.h
+++ b/src/AssociativeOpsTable.h
@@ -35,13 +35,12 @@ struct AssociativePattern {
     /** Contain the identities for each dimension of the associative op. */
     std::vector<Expr> identities;
     /** Indicate if the associative op is also commutative. */
-    bool is_commutative;
+    bool is_commutative = false;
 
-    AssociativePattern()
-        : is_commutative(false) {
+    AssociativePattern() {
     }
     AssociativePattern(size_t size)
-        : ops(size), identities(size), is_commutative(false) {
+        : ops(size), identities(size) {
     }
     AssociativePattern(const std::vector<Expr> &ops, const std::vector<Expr> &ids, bool is_commutative)
         : ops(ops), identities(ids), is_commutative(is_commutative) {
--- a/src/Associativity.h
+++ b/src/Associativity.h
@@ -83,13 +83,12 @@ struct AssociativeOp {
     AssociativePattern pattern;
     std::vector<Replacement> xs;
     std::vector<Replacement> ys;
-    bool is_associative;
+    bool is_associative = false;
 
-    AssociativeOp()
-        : is_associative(false) {
+    AssociativeOp() {
     }
     AssociativeOp(size_t size)
-        : pattern(size), xs(size), ys(size), is_associative(false) {
+        : pattern(size), xs(size), ys(size) {
     }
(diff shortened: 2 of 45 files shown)


src/runtime/HalideRuntime.h
```diff
@@ -1364,10 +1364,10 @@ extern halide_can_use_target_features_t halide_set_custom_can_use_target_feature
 extern int halide_default_can_use_target_features(int count, const uint64_t *features);
 
 typedef struct halide_dimension_t {
-    int32_t min, extent, stride;
+    int32_t min, extent, stride;  // NOLINT(modernize-use-default-member-init)
```
steven-johnson, line 1367: Since this file has to be usable from C as well as C++, we have to do something here. NOLINT is ugly but works. The alternative would be to make the entire interior of the class #ifdef __cplusplus. Better? Worse?

src/AssociativeOpsTable.h
```diff
@@ -35,13 +35,12 @@ struct AssociativePattern {
     /** Contain the identities for each dimension of the associative op. */
     std::vector<Expr> identities;
     /** Indicate if the associative op is also commutative. */
-    bool is_commutative;
+    bool is_commutative{false};
 
-    AssociativePattern()
-        : is_commutative(false) {
+    AssociativePattern() {
```
abadams, line 40: =default

src/Associativity.h
```diff
@@ -83,13 +83,12 @@ struct AssociativeOp {
     AssociativePattern pattern;
     std::vector<Replacement> xs;
     std::vector<Replacement> ys;
-    bool is_associative;
+    bool is_associative{false};
 
-    AssociativeOp()
-        : is_associative(false) {
+    AssociativeOp() {
```
abadams, line 88: = default. Maybe this change should go in after the default constructors one.

steven-johnson: All those changes are in https://github.com/halide/Halide/pull/5380

abadams: Oh I see, you're merging the PRs into each other. Erm, OK I guess I'll just try to skim over the constructors.

src/Closure.h
```diff
@@ -47,22 +47,21 @@ class Closure : public IRVisitor {
         Type type;
 
         /** The dimensionality of the buffer. */
-        uint8_t dimensions;
+        uint8_t dimensions{0};
```
abadams, line 50: I thought we typically used: member = 0 vs member{0} for PODs

src/runtime/HalideRuntime.h
abadams: Having a default value for the dimension fields seems like a plus when using this struct from C++ code, so I'd vote for the ifdef

src/Closure.h
steven-johnson: Apparently the latter is what clang-tidy prefers? I can check to see if there is a way to change that.

steven-johnson: Yeah, there is a way, but I'll just need to re-do this PR in order to avoid a weird history.

steven-johnson: PTAL

abadams (approved):

PR #5382, opened 2020-10-20
Enable the useful readability-* checks in clang-tidy

steven-johnson:


src/CodeGen_LLVM.cpp
```diff
@@ -3172,7 +3172,7 @@ void CodeGen_LLVM::visit(const Call *op) {
         }
 
         // Create a null-initialized global to track this object.
-        const auto base_fn = sub_fns.back().fn;
```
abadams, line 3175: I suspect the original was a typo and the intent was for it to be `const auto *`

src/Dimension.h
```diff
@@ -83,7 +83,7 @@ class Dimension {
     /** Get a different dimension of the same buffer */
     // @{
     Dimension dim(int i);
-    const Dimension dim(int i) const;
+    Dimension dim(int i) const;
```
abadams, line 86: Wait... why do both these accessors exist if the const one is just as good as the non-const one?

abadams (approved):
lgtm with nits

src/Dimension.h
steven-johnson: They probably don't. I'll just delete the non-const one. The claim is: "readability-const-return-type:
Checks for functions with a const-qualified return type and recommends removal of the const keyword. Such use of const is usually superfluous, and can prevent valuable compiler optimizations."

PR #5384, opened 2020-10-20
Add `logical_not` function for Python

dragly:
This change introduces `logical_not` as a free function and member function
that calls `operator!`.

The reason why a new function is added is because there is no `operator!`
in Python and the `not` keyword cannot be overloaded. Hence, there was
currently no way to call the C++ `operator!` in Python.

### DIFF
--- a/python_bindings/src/PyIROperator.cpp
+++ b/python_bindings/src/PyIROperator.cpp
@@ -185,6 +185,9 @@ void define_operators(py::module &m) {
     m.def("likely_if_innermost", &likely_if_innermost);
     m.def("saturating_cast", (Expr(*)(Type, Expr)) & saturating_cast);
     m.def("strict_float", &strict_float);
+    m.def("logical_not", [](const Expr &expr) -> Expr {
+        return !expr;
+    });
 }
 
 }  // namespace PythonBindings
--- a/python_bindings/correctness/negate_test.py
+++ b/python_bindings/correctness/negate_test.py
@@ -1,0 +1,38 @@
+import halide as hl
+
+def test_free_logical_not_function():
+    x = hl.Var('x')
+
+    f = hl.Func('f')
+    f[x] = x > 5
+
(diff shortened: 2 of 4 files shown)


dragly:
I am not sure if this should be called `negate` or something else. I am also not sure if it should be both a free function and a member, but figured I would add both since it is easier to discuss concrete examples.

The background for this PR is that I have used the bitwise `~` operator in place of the `!` operator since there was no `!` operator in Python. The problem with doing so is that it can inadvertently be applied to a wider type than a bool/uint1, which can make negations of true values stay true (0000 0001 becomes 1111 1110, for instance). I am not sure if it was because I inadvertently turned a `uint1` into a `uint8` or if I hit some platform-specific issue with boolean values in OpenCL. In any case, it would be nice to have a proper way of doing negations in Halide with Python.

python_bindings/src/PyIROperator.cpp
```diff
@@ -185,6 +185,9 @@ void define_operators(py::module &m) {
     m.def("likely_if_innermost", &likely_if_innermost);
     m.def("saturating_cast", (Expr(*)(Type, Expr)) & saturating_cast);
     m.def("strict_float", &strict_float);
+    m.def("negate", [](const Expr &expr) -> Expr {
```
steven-johnson, line 188: `logical_not` might be a better choice here; `negate` could imply a sign-negation.

dragly: Thanks! Changed to `logical_not` now. That is a much better name.

steven-johnson:
Generally looks good, opened a clone for testing. Only comment so far is that it's probably worth adding something to python_bindings/readme.md about this.

dragly:
> Generally looks good, opened a clone for testing. Only comment so far is that it's probably worth adding something to python_bindings/readme.md about this.

I agree. Updated now.

steven-johnson (approved):
Thanks for the fix!

PR #5386, opened 2020-10-21
Fix bounds inference resulting in vector types in some cases

dsharletg:


steven-johnson:
LGTM, running some tests inside google to verify

steven-johnson (approved):

abadams (approved):

PR #5387, opened 2020-10-21
Push lets near their uses (whenever possible) while CSEing +=

pranavb-ca:
The CSE pass tries to do CSE jointly on the index and value of store nodes.
This is to stop:
```
f[x] = f[x] + y
```
from turning into
```
f[x] = f[z] + y
```
This is because the two 'x' indices are not CSE'd together.

However, one problem with jointly CSEing the store index and vaues
is that after CSEing, the pass puts the lets
(the values that were CSE'd) before the new store. That is we get,

```
let(t0...)
let(t1...)
f[t0] = f[t0] + function_of(t1)
```
Suppose however, there was nothing to CSE between the store index and value, that is
the index was unchanged after CSE. In that case, moving lets before the store puts the
lets too far away from their uses. This is ok except, there are passes like LoopCarry
that are beneficial when they are able to see a long continuous block of stores (eg. when
unrolling). But now, they'll see a long sequence of LetStmts.
Instead, if the index was unchanged after CSE, we should build up the stores as

```
f[x] = let(t0.. in (value));
```
This way LoopCarry is given a chance to see a series of stores.
Handling this in CSE, means the LoopCarry pass need not be complicated.

### DIFF
--- a/src/CSE.cpp
+++ b/src/CSE.cpp
@@ -6,6 +6,7 @@
 #include "IROperator.h"
 #include "Scope.h"
 #include "Simplify.h"
+#include "ExprUsesVar.h"
 
 namespace Halide {
 namespace Internal {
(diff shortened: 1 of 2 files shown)


abadams:
I think there might be something more general we could do. In one direction it would be good to only lift out the lets that the index needs as stmts, and leave the rest inside. In the other direction it might be a good idea to jointly CSE entire blocks of Stores, so that they don't get lets in between them. Do you have an example case where it's not doing what you want?

pranavb-ca:
> I think there might be something more general we could do. In one direction it would be good to only lift out the lets that the index needs as stmts, and leave the rest inside. In the other direction it might be a good idea to jointly CSE entire blocks of Stores, so that they don't get lets in between them. Do you have an example case where it's not doing what you want?

I have a reproducer here in https://github.com/halide/Halide/issues/5396. Let me take a stab at the first of the two solutions you have suggested.

pranavb-ca:
@abadams  - PTAL - a more general fix.

pranavb-ca:
@abadams - ping!

abadams:
lgtm. Rerunning the failing performance test in case it's not a flake.

dsharletg (approved):

abadams (approved):

PR #5389, opened 2020-10-22
Fix transitive dependencies

arcbbb:
just came across this loop and found out that `change` was always false.

### DIFF
--- a/src/Associativity.cpp
+++ b/src/Associativity.cpp
@@ -273,7 +273,7 @@ void add_transitive_dependencies(vector<set<int>> &dependencies) {
                     for (const auto &idx : dependencies[j]) {
                         if (dependencies[i].count(idx) == 0) {
                             dependencies[i].insert(idx);
-                            change = false;
+                            change = true;
                         }
                     }
                 }


steven-johnson:
What behavior does this change? Is a test needed to verify the fix?

src/Associativity.cpp
```diff
@@ -273,7 +273,7 @@ void add_transitive_dependencies(vector<set<int>> &dependencies) {
                     for (const auto &idx : dependencies[j]) {
                         if (dependencies[i].count(idx) == 0) {
                             dependencies[i].insert(idx);
-                            change = false;
+                            change = true;
```
vksnk, line 276: I am not familiar with the code, but it does look like a legit bug.

steven-johnson: Agreed, but presumably the change will affect behavior somewhere. If it doesn't affect an existing test, we need a new one added to verify.

abadams: It would have to be some very subtle almost-associative update definition that was previously incorrectly allowed. I think it would be quite challenging to find one.

steven-johnson (approved):
Thanks for the fix!

PR #5390, opened 2020-10-22
Upgrade hexagon_remote/Makefile  to use C++11

steven-johnson:
HalideRuntime.h now requires at least C++11 (for C++ files), so ensure that we pass `-std=c++11` for all those when building the remote.

### DIFF
--- a/src/runtime/hexagon_remote/Makefile
+++ b/src/runtime/hexagon_remote/Makefile
@@ -41,23 +41,24 @@ HEXAGON_SDK_LIBS ?= "${HEXAGON_SDK_ROOT}/libs"
 
 COMMON_FLAGS = -I ${HEXAGON_SDK_INCLUDES}/stddef -I../
 COMMON_CCFLAGS = ${COMMON_FLAGS} -O3 -I ${HEXAGON_SDK_LIBS}/common/remote/ship/android_Release
+COMMON_CXXFLAGS = -std=c++11
 HEXAGON_QAICFLAGS = ${COMMON_FLAGS}
 
 BIN ?= bin
 
 CC-host = ${CXX}
-CXX-host = ${CXX}
+CXX-host = ${CXX} $(COMMON_CXXFLAGS)
 
 CC-v62 = ${HEXAGON_TOOLS_ROOT}/Tools/bin/hexagon-clang
-CXX-v62 = ${HEXAGON_TOOLS_ROOT}/Tools/bin/hexagon-clang++
+CXX-v62 = ${HEXAGON_TOOLS_ROOT}/Tools/bin/hexagon-clang++ $(COMMON_CXXFLAGS)
 LD-v62 = ${HEXAGON_TOOLS_ROOT}/Tools/bin/hexagon-link
 
 CC-arm-64-android = ${ANDROID_NDK_ROOT}/toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/aarch64-linux-android-gcc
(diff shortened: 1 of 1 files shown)


dsharletg:
IIRC there is an issue with enabling this. Maybe adding thread synchronization code around statics? @pranavb-ca?

steven-johnson:
> IIRC there is an issue with enabling this. Maybe adding thread synchronization code around statics? @pranavb-ca?

We should be able to disable that with `-fno-threadsafe-statics`, like we now do for the runtime files.

pranavb-ca:
I am not sure I remember. @dpalermo?

dpalermo:
I seem to remember something as well.  The device side of hexagon_runtime is C not C++ so the issue must have been with either the sim support or host side libs.  Looking at the current Makefile we are only passing -std=c++11 for sim_host.o...avoiding compiling all of the other sim and host side C++ with it.  This one instance of -std=c++11 was committed back in 2016 so it is possible that if it was a clang/lib issue, that it may have since been resolved.  Otherwise, if it was the thread-safe static issue you mentioned, the flag Steven mentioned sounds promising.

steven-johnson:
Adding `-fno-threadsafe-statics` almost certainly won't *hurt* anything (since we were essentially compiling that way before, except for sim_host.cpp), but obviously it would be better if we had confidence there wasn't a lurking bug here. 

FYI, thus currently blocks building hexagon_remote due to the HalideRuntime.h issue above -- we could either back out the changes to HalideRuntime.h to make it no longer assume that __cplusplus >= C++11 (which would be suboptimal) or add a third special case to it to allow for C++-but-not-C++11 (which would be really gross and to be avoided).

pranavb-ca (approved):
I spoke with our engineer who owns C++ library support and in his opinion with the latest 8.1 tools we should be ok with C++11. On the same note, should we update the SDK and Hexagon Tools within (or alternatively, simply drop the latest 8.1.x inside the present 3.3 SDK?)

steven-johnson:
> I spoke with our engineer who owns C++ library support and in his opinion with the latest 8.1 tools we should be ok with C++11. 

To be clear: without `-fno-threadsafe-statics` added?

> On the same note, should we update the SDK and Hexagon Tools within (or alternatively, simply drop the latest 8.1.x inside the present 3.3 SDK?)

Not sure I understand the question. You mean the defaults specified in the Makefile? (3.3.3 / 8.1.05 are what we are building with inside Google.)

steven-johnson:
Monday ping to get final clarification regarding `-fno-threadsafe-statics`

dsharletg:
I think we should just add --fno-threadsafe-statics anyways. Even if it works and builds without it, adding such thread safety would be a significant change in behavior that might affect things.

steven-johnson:
OK, landed that way. We can back it out if there is an issue.

pranavb-ca:
> > I spoke with our engineer who owns C++ library support and in his opinion with the latest 8.1 tools we should be ok with C++11.
> 
> To be clear: without `-fno-threadsafe-statics` added?
>

Sorry I wasn't clear - Yes with ``-fno-threadsafe-statics`` added.
> > On the same note, should we update the SDK and Hexagon Tools within (or alternatively, simply drop the latest 8.1.x inside the present 3.3 SDK?)
> 
> Not sure I understand the question. You mean the defaults specified in the Makefile? (3.3.3 / 8.1.05 are what we are building with inside Google.)

The latter  because moving to a newer SDK and Hexagon tools will require installing a newer SDK. Following that we can change the defaults in the makefile.

PR #5391, opened 2020-10-22
Add Stage.gpu_lanes to Python bindings.

Infinoid:
In the C++ API, the `.gpu_lanes(x)` scheduling method exists for both `Func` and `Stage` objects.  The Python bindings only allow calling it on `Func`; move it over so it can be used on both.


steven-johnson (approved):
Thanks for the fix!

PR #5393, opened 2020-10-22
Add a test case for multiple argument memoize_tag to demonstrate usage.

zvookin:


shoaibkamil:
Thanks for adding this-- we were having a discussion internally about how to use this for memoizing a LUT that depends on a subset of inputs.  I also noticed we don't have a test with a GPU schedule, which might also be good to include.

steven-johnson (approved):

dsharletg (approved):

zvookin:
I'm not sure memoize works for GPU in the way one would like. This should be tested and fixed for sure, but if it hasn't been done yet there are some tricky bits to it.

There is also seems to be a bug where memoizing the top-level Func produces an undefined symbol compiler erro. Will investigate both these things.

PR #5395, opened 2020-10-22
web assembly threads + demo app

### LINKED ISSUE #5397: Runtime GCD strips wasm arch features
Passing the following target string to `-r ... target=`

```
target=wasm-32-wasmrt-wasm_simd128-wasm_threads,wasm-32-wasmrt-wasm_simd128-wasm_threads,wasm-32-wasmrt-wasm_simd128-wasm_threads
```

causes the generator to ignore `wasm_simd128` and `wasm_threads`. This breaks `add_halide_library`'s `USE_RUNTIME` feature.

abadams:
This PR turns on the posix thread pool for wasm, and adds a HelloWasm demo app that does a reaction diffusion thing in a browser window. The output is currently hosted at https://halide-lang.org/wasm_demo

I'm currently trying to figure out how to build the app with cmake, but the rest of it is ready to review.

Fixes #5397


zvookin:
The SIMD modes of the hosted version don't work in the Mac OS Chrome I am using. (Version 86.0.4240.80 (Official Build) (x86_64)) Will debug later, but wondering if there is any failure information that can be surfaced in the UI to help head off issues and frustration.

apps/HelloWasm/Makefile
```diff
@@ -0,0 +1,82 @@
...
+all: js/index.js js/index_simd.js js/index_threads.js js/index_simd_threads.js
+
+$(GENERATOR_BIN)/reaction_diffusion_generator: reaction_diffusion_generator.cpp $(GENERATOR_DEPS)
+	@mkdir -p $(@D)
+	$(CXX) $(CXXFLAGS) $(filter %.cpp,$^) -o $@ $(LIBHALIDE_LDFLAGS)
+
+$(BIN)/wasm/reaction_diffusion_%.a: $(GENERATOR_BIN)/reaction_diffusion_generator
+	@mkdir -p $(@D)
+	$< -g reaction_diffusion_$* -e $(GENERATOR_OUTPUTS) -o $(@D) target=wasm-32-wasmrt-no_runtime threads=false
+$(BIN)/wasm/runtime.a: $(GENERATOR_BIN)/reaction_diffusion_generator
```
steven-johnson, line 19: insert blank line

apps/HelloWasm/core.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <sstream>
+
+#include "HalideBuffer.h"
+#include "reaction_diffusion_init.h"
+#include "reaction_diffusion_render.h"
+#include "reaction_diffusion_update.h"
+
+const int W = 1024, H = 1024;
+
+struct context {
```
steven-johnson, line 16: nit: struct typenames are generally CamelCase in Halide

apps/HelloWasm/core.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if ((ctx->iteration & 15) == 15) {
+        char buf[1024] = {0};
+        snprintf(buf, sizeof(buf),
+                 "Time for Halide update + render: %0.2f ms<br>"
+                 "Time for blit to framebuffer: %0.2f ms<br>"
+                 "Frame rate: %2.0f fps",
+                 ctx->smoothed_runtime, ctx->smoothed_blit_time, ctx->smoothed_fps);
+        // Run some javascript inline to update the web-page
+        EM_ASM({
+            document.getElementById(UTF8ToString($0)).innerHTML = UTF8ToString($1);
```
steven-johnson, line 80: ...the ability to inline javascript into c++ code is just soooooo wrong

src/CodeGen_WebAssembly.cpp
```diff
@@ -136,6 +136,11 @@ string CodeGen_WebAssembly::mattrs() const {
         sep = ",";
     }
 
+    if (target.has_feature(Target::WasmThreads)) {
+        s << sep << ",+atomics,+bulk-memory";
```
steven-johnson, line 140: atomics I get, but why do we need bulk-memory here? (I'm sure there's a good reason, just can't remember)

steven-johnson (commented):
Adding WasmThreads is great, but might be worth splitting into a separate PR, so we can check how it behaves on the rest of the wasm test suite

abadams:
Yeah only the single-threaded scalar version works without turning on experimental features. I'll add a note to index.html.

apps/HelloWasm/core.cpp
abadams: I know, but this is the simplest approach in the emscripten docs. The other approach is more heavy-weight: you can expose javascript functions to the C code via an extern "C" declaration. It's not clear if I can pass strings across this though.

src/CodeGen_WebAssembly.cpp
abadams: I was not able to figure that out. There's an error message telling you to do it. It makes me worry there's some mass copying of memory going on.

abadams:
@alexreinking I'm a bit stumped on how to do the cmake build. The rungen target requires compiling against generated source (the registration files), and also a source file out of the Halide install (RunGenMain.cpp). I can't find any existing examples of this in our cmake builds.

alexreinking:
I'm checking out this branch to just write the CMake stuff quickly, but...

> @alexreinking I'm a bit stumped on how to do the cmake build. The rungen target requires compiling against generated source (the registration files), and also a source file out of the Halide install (RunGenMain.cpp). I can't find any existing examples of this in our cmake builds.

CMake does have RunGenMain support. Anyway, here's an abstract example:

```
add_halide_library(filter FROM gen
                   REGISTRATION reg_cpp)
add_executable(runner ${reg_cpp})
target_link_libraries(runner PRIVATE Halide::RunGenMain)
```

Here it is in use by the Adams2019 demo:

https://github.com/halide/Halide/blob/2d1aee959e70c95600643baff0d028cb9529c0d0/src/autoschedulers/adams2019/CMakeLists.txt#L55-L64

> Edit: Plus the javascript outputs must go in a specific in-tree folder relative to the index.html source, or it won't be able to find them. I guess I could make a build step that copies index.html into the install tree or something?

No need to add a build step if they aren't generated. Just use [`configure_file(<input> <output> COPYONLY)`](https://cmake.org/cmake/help/latest/command/configure_file.html). The input and output are either absolute paths or relative paths to the current source/binary dirs, respectively.

alexreinking:
I built via CMake on WSL, and tested the pipeline on Edge. Everything worked.

apps/HelloWasm/core.cpp
steven-johnson: yeah, this is the least-bad option, I realize. it just feels squicky.

steven-johnson (approved):

abadams:
Summarizing some offline discussion: Performance makes sense on os x and windows (2-3x slower than native code). On linux throwing 16 threads at only makes it 3.5x faster though, with most of the threads sitting idle. On OS X and Windows it's 10x faster with 16 threads. I discovered it's possible to fix this linux performance weirdness by getting the worker threads to spin instead of sleeping, but they have to spin for the whole frame time (1/60s) for that to work, which just burns CPU like crazy. I'm not sure it's necessary to diagnose this now. It seems to be a property of wasm rather than of our thread pool.

PR #5407, opened 2020-10-26
loosen preconditions on div by single point in Bounds.cpp

rootjalex:
Due to Halide division by 0 semantics, the precondition can be loosened for bounds inference on division by a single point, as an interval divided by the zero interval produces the zero interval.
Discussed with @abadams @shoaibkamil and @jn80842 .

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -564,7 +564,7 @@ private:
         } else if (can_prove(b.min == b.max)) {
             Expr e1 = a.has_lower_bound() ? a.min / b.min : a.min;
             Expr e2 = a.has_upper_bound() ? a.max / b.max : a.max;
-            if (can_prove(b.min >= 0) || op->type.is_uint()) {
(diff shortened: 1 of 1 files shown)


src/Bounds.cpp
```diff
@@ -564,9 +564,9 @@ class Bounds : public IRVisitor {
         } else if (can_prove(b.min == b.max)) {
             Expr e1 = a.has_lower_bound() ? a.min / b.min : a.min;
             Expr e2 = a.has_upper_bound() ? a.max / b.max : a.max;
-            if (is_positive_const(b.min) || op->type.is_uint()) {
+            if (can_prove(b.min >= 0) || op->type.is_uint()) {
```
abadams, line 567: op->type.is_uint() is cheaper, so switch the order of the args to this `||`

rootjalex: switched in 03dba70

abadams (approved):

PR #5408, opened 2020-10-26
Add a spin before the cond_wait in the thread pool

abadams:
Opening for discussion. This improves performance on a few apps by 1-3%, and improves the performance of performance_inner_loop_parallel by more like 20%.

@zvookin


abadams:
This was motivated by noticing that our cond_wait in synchronization_common.h doesn't spin before sleeping.

src/runtime/thread_pool_common.h
```diff
@@ -273,24 +276,38 @@ WEAK void worker_thread_already_locked(work *owned_job) {
         if (!job) {
             // There is no runnable job. Go to sleep.
```
dsharletg, line 277: Can you move the declaration/initialization of spin_count here, so it doesn't need to be explicitly reset?

dsharletg: Err, no, you can't...

abadams: The spin_count should only be reset when the thread actually manages to do something useful. In this implementation, if a thread gets woken up for a job and the work is all already claimed, then spins 40 times waiting for a new job to be enqueued before giving up. If the same thing starts to happen with the next parallel loop, then it doesn't keep spinning - it goes to sleep right away because spin_count is still > 40 from the last job.

abadams: Note that this makes it different to just adding a spin inside cond_wait. I believe there are never more than 40 spins per work item completed here.

zvookin:
Does the total number of CPU cycles go up? I would expect it to, but the question is whether it goes up a noticeable amount or just a little bit. Have you looked at how it interacts with the atomic index update change especially for the inner loop case. I'll check it on the benchmarking cases I've been looking at.

The only significant question is whether the experiment is good enough to prove it is a win enough of the time and not a big lose in any case we care about. (Which seems unlikely given 40 spins just isn't that many and it should not be happening particularly often. Like unless there is a bug or my quick analysis is way off.) Then again, the 40 is sort of becoming this magical number that isn't really well grounded so...

Not objection to landing it pending a bit more experimentation, but it strikes me as more a sign that our design is fragile and could be improved than a principled approach. Longer term, I think having the owners sleep on individual condvars and arranging for targeted thread handoffs is probably the better way to get deterministic and closer to optimal scheduling.

abadams:
CPU cycles goes up by about 1%, runtime of good schedules goes down by about 1%. Oddly though I'm seeing the runtime of some schedules that use nested parallelism go up by about 1%, and I haven't been able to figure out why. I've verified that those schedules don't spin anyway (because there are always more jobs on the job stack).

abadams:
Updated to also spin on mutexes even if threads are parked, and performance of the autoschedule (which uses fine-grained parallelism) went from 53ms to 35ms! Similar benefit in interpolate app: 32ms -> 18ms

steven-johnson:
Where do we stand on this? Ready to land? Needs more pondering?

zvookin:
We need to move all the par for experiments into a single branch and do some experiments to evaluate them. My running of the first version of this change is that it barely helps and makes a couple things slightly worse. I need to pickup the latest, but I'd like all of this to end up in the par_for_opt branch and to have spreadsheets showing the measurements of the stuff we're looking at across a chosen set of conditions. Otherwise we're playing whackamole.

steven-johnson:
At this point I assume we aren't landing this anytime soon. Should we leave it open?

abadams:
I think we should just land it. It's a minor change with a pretty good performance boost on some apps (note that Zalman's comment applied to the first version of the change).

steven-johnson:
Just synced to master to re-test since this has been sitting for a while.

Adding @zvookin to reviewers so he can confirm that his concern is resolved.

zvookin:
tl;dr: Feel free to land it but I'm pretty sure it is a lucky thing that won't stand the test of time.

My concern is that in the grand scheme of things, this is more noise than signal. It is another chunk of complexity in a subsystem that needs to be approached more methodically and after doing so, will likely not have this code in place. One can argue that a short term performance improvement is worth it and I don't much disagree. But to make that case reliably, we need performance and resource usage across a broader class of uses. (E.g. it likely depends a lot on the number of cores in the hardware, etc.) It would be really nice if the performance data we do have was in a table in the PR description to help with future work.

Approaching things more methodically involves creating a set of benchmarks that measure at least time and resource usage across a variety of work item and loop length sizes, then picking at least a few points in the hardware space. We should probably also try to have a mechanism to get feedback from users, such as some way to use Google's large scale profiling infrastructure to track the benefit (or cost) of changes like this over time.

My guess is the coding approach will evolve both the threadpool and fast sync mechanisms somewhat. E.g. moving to a conditional critical section design, instead of mutes/cond_wait, which allows pushing the condition evaluation into the fast sync scheduling code and eliminating some amount of spurious wakeups. We've also talked about a signal N threads mechanism to avoid the A/B team logic. (Conditional critical section probably takes care of that too.) The fundamental algorithms of how we do e.g. a parallel for loop also need investigation and likely need to become adaptive.

zvookin (approved):
Approval on the basis that I don't think it will break things and I don't think it is bad. As I said in the comments, I don't think this is the direction we'll ultimately go. Please consider adding actual performance data to the PR.

abadams:
Yeah, I don't think it solves all our problems, but it's sane, and it improve things in some situations. I'll rerun all the apps at least and report back.

abadams:
app | speedup | before | after
-- | -- | -- | --
bilateral manual | 1.002274472 | 5.00152 | 4.99017
bilateral auto | 1.005950422 | 5.16633 | 5.13577
bgu manual | 1.006090277 | 8.85121 | 8.79763
bgu auto | 1.189101281 | 31.4767 | 26.471
conv manual | 0.9863711133 | 4.79113 | 4.85733
conv auto | 0.9998220172 | 12.9203 | 12.9226
hist manual | 1.02685121 | 1.12547 | 1.09604
hist auto | 1.413868213 | 7.45175 | 5.27047
interpolate manual | 1.014333005 | 4.51648 | 4.45266
interpolate auto | 1.24461596 | 16.3031 | 13.0989
lens blur manual | 1.026810895 | 4.42728 | 4.31168
lens blur auto | 1.114805521 | 12.439 | 11.158
local laplacian manual | 0.9972458769 | 24.8033 | 24.8718
local laplacian auto | 1.063665744 | 34.8208 | 32.7366
max filter manual | 1.012758825 | 48.6423 | 48.0295
max filter auto | 1.021298673 | 67.5202 | 66.1121
nl means manual | 1.000094986 | 69.4907 | 69.4841
nl means auto | 1.000902275 | 65.893 | 65.8336
stencil chain manual | 1.038385325 | 4.44377 | 4.2795
stencil chain auto | 0.993515189 | 9.42986 | 9.49141
geomean: | 1.053347848 |   |

abadams:
Something not captured in the stats is that this seems to make runtimes a lot less variable.

abadams:
On the auto schedule for hist, the standard deviation of runtimes goes from 0.85ms down to 0.29ms over 1000 runs (relative to an average runtime of ~6ms)

zvookin:
Please add a brief description of the hardware. (Instruction set, num cores, rough clock speed, whether variable clocks are enabled.)

abadams:
Done (by editing the post above). My system was still configured for siggraph deadline benchmarking jobs.

PR #5410, opened 2020-10-26
doxygen wasn't finding the runtime

abadams:
HalideRuntime.h was missing from the doxygen-generated docs, which is a pretty important header. Turned out that the recursive setting needed turning back on.


alexreinking (approved):

abadams:
Actually I don't think this was affecting the cmake docs. I think they were explicitly enumerating all the relevant folders. I could move the RECURSIVE setting into the Makefile-specific options, but it's probably fine to just leave it on.

alexreinking:
> Actually I don't think this was affecting the cmake docs. I think they were explicitly enumerating all the relevant folders. I could move the RECURSIVE setting into the Makefile-specific options, but it's probably fine to just leave it on.

That is correct. See the documentation:

> `DOXYGEN_RECURSIVE`
> Set to `YES` by this module.

https://cmake.org/cmake/help/v3.19/module/FindDoxygen.html

---

IMHO it's fine to be explicit here as long as setting it doesn't _break_ anything.

PR #5412, opened 2020-10-27
More clang-tidy cleanup

steven-johnson:
Enable all the bugprone-* checks, then disable the ones we don't want (or need code fixes for).

Made a couple of trivial changes to pacify bugprone-unhandled-self-assignment.

Some of the other disabled checks are likely worthwhile but will be done separately to ensure code review isn't confused.

### DIFF
--- a/src/IntrusivePtr.h
+++ b/src/IntrusivePtr.h
@@ -135,7 +135,7 @@ public:
     }
 
     IntrusivePtr<T> &operator=(const IntrusivePtr<T> &other) {
-        if (this == &other) {
+        if (other.ptr == ptr) {  // NOLINT(bugprone-unhandled-self-assignment)
             return *this;
         }
         // Other can be inside of something owned by this, so we
--- a/src/IntrusivePtr.h
+++ b/src/IntrusivePtr.h
@@ -136,6 +136,8 @@ public:
 
     // NOLINTNEXTLINE(bugprone-unhandled-self-assignment)
     IntrusivePtr<T> &operator=(const IntrusivePtr<T> &other) {
+        // Same-ptr but different-this happens frequently enough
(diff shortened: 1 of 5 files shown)


src/IntrusivePtr.h
```diff
@@ -135,7 +135,7 @@ struct IntrusivePtr {
     }
 
     IntrusivePtr<T> &operator=(const IntrusivePtr<T> &other) {
-        if (other.ptr == ptr) {
+        if (this == &other) {
```
abadams, line 138: This is different though. The old code compared by value, and the new code compares by identity. So in the old code, more cases could do the early-out.

steven-johnson: ...yes, this is true, but in practice, is this something that actually happens frequently enough to worry about?

abadams: It happens whenever you assign an Expr to another Expr that happens to already be the same thing. E.g.:

```
Expr my_function(Expr a, Expr b) {
 ...
 a = b;
 ...
}
...
my_function(x, x);
```

This includes uninitialized Exprs:
```
Thing make_a_thing(... const Expr &e ... ) {
  Thing thing; // has default-initialized Expr fields
  ...
  thing.foo = e;
  return thing;
}
...
make_a_thing(..., Expr(), ...)
```

I inserted a check for `other.ptr == ptr && this != &other` and compiled local laplacian. It triggered 51016 times. 43k were the non-null case and 7k were the null case.

steven-johnson: ...that's worth a //NOLINT.

abadams: Probably also worth a link to this discussion to reduce future confusion.

steven-johnson: Done

abadams (approved):

PR #5423, opened 2020-11-02
CUDA Texture Sampler

jlaxson:
This is the next step on my texture sampler journey.  Pretty much nothing works and certainly nothing is tested, but I have some architecture questions, as not much feels very clear to me.

Unlike OpenCL which has a separate device interface for textures, right now I'm passing a bit of information to halide_cuda_run so it can wrap existing buffers in the texture descriptor.  Tradeoff for this is only 1D and 2D samplers can be used, 3D requires allocating a cuArray.  

I'm a bit out over my skis in the PTX and LLVM codegen side (and their rather unholy and certainly undocumented marriage when it comes to the texture intrinsics).  I have the intrinsic call roughly matching the sole example I found, but seem to be talented at crashing inside LLVM code.

[This file](https://smlnj-gitlab.cs.uchicago.edu/manticore/llvm/blob/master/test/CodeGen/NVPTX/tex-read-cuda.ll) is roughly the codegen I'm trying to get.  If I turn off running the modulePassManager, I can see this IR which seems rather close.
```
; Function Attrs: nounwind
declare <4 x i32> @llvm.nvvm.tex.unified.1d.v4s32.s32(i64 %0, i32 %1) #13

; Function Attrs: nounwind
declare <4 x float> @llvm.nvvm.tex.unified.1d.v4f32.s32(i64 %0, i32 %1) #13

  %4 = call <4 x i32> @llvm.nvvm.tex.unified.1d.v4s32.s32(i64 %input, i32 %3) #14
  %5 = extractelement <4 x i32> %4, i32 0
  %6 = mul nsw i32 %t25, 2
  %7 = call reassoc nnan ninf nsz contract afn <4 x float> @llvm.nvvm.tex.unified.1d.v4f32.s32(i64 %f, i32 %6) #14
  %8 = extractelement <4 x float> %7, i32 0
 ```

I'm not sure if explicitly getting `tail call` or the extra presence here of floating point flags is a problem.  When I do run modulePassManager, LLVM fails internally during PTX lowering (line 5030 in [this source](https://llvm.org/doxygen/NVPTXISelLowering_8cpp_source.html)).  Not sure if that's because I'm using the intrinsic wrong or there's something broken/old in LLVM itself.

Any thoughts, feedback welcome, I have a few extra questions inline as well.

### CODE REVIEW

File: src/StorageFlattening.cpp [hunk 1]
```diff
@@ -259,7 +260,7 @@ class FlattenDimensions : public IRMutator {
             Expr store = Call::make(value.type(), Call::image_store,
                                     args, Call::Intrinsic);
             return Evaluate::make(store);
-        } else if (in_gpu && textures.count(op->name)) {
+        } else if (in_gpu && textures.count(op->name) && false && in_device_api != DeviceAPI::CUDA) { // CUDA writes are still directly to memory
```
jlaxson, line 263: I'm not sure what to do here.  In CUDA the texture sampling is read-only, so writes should be flattened as before.  But when the StorageFlattening pass runs, DeviceAPI hasn't been assigned and so there isn't anything to key off to flatten or not flatten at this phase.

There should also be a check somewhere to ensure there are not reads and writes to a buffer-as-texture within any one kernel.

src/CodeGen_GPU_Host.cpp
```diff
@@ -441,7 +441,14 @@ void CodeGen_GPU_Host<CodeGen_CPU>::visit(const For *loop) {
                                          i));
             }
 
-            builder->CreateStore(ConstantInt::get(i8_t, closure_args[i].is_buffer),
+            int8_t buffer_type = 0;
+            if (closure_args[i].is_buffer && closure_args[i].memory_type == MemoryType::GPUTexture) {
+                buffer_type = 2;
```
jlaxson, line 446: This is a little clumsy but it's certainly effective for getting the texture flag to the runtime.  Alternatives welcome

jlaxson:
@zvookin @abadams @steven-johnson @dsharletg

jlaxson:
Also somewhere in codegen there should be a check that you have compute 3.0 if generating a GPUTexture buffer, but not sure where that should go

jlaxson:
Rather unbelievably this works now.  LLVM master has way way better error messages than 11 or whatever I was using before.  Ready for review, beat me up.

src/StorageFlattening.cpp
jlaxson: Moved the DeviceAPI assignment to before StorageFlattening, I need it anyways so allocations can be properly strided if necessary.

steven-johnson:
Opened https://github.com/halide/Halide/pull/5431 for testing

steven-johnson:
Various test failures in the clone build, possibly related

steven-johnson:
see https://buildbot.halide-lang.org/master/#/builders/74/builds/2

steven-johnson:
1-month check-in. (No rush on our side of course, just want to see if this is on the backburner for a while.)

steven-johnson:
I assume this PR is on the backburner. Please ping when the test failures get addressed.

jlaxson:
Thanks @steven-johnson will do, hope to have some more side time come the holiday weeks.

steven-johnson:

steven-johnson:

PR #5427, opened 2020-11-02
Upgrade PyBind11 version in CMake to 2.5

steven-johnson:
v2.4.3 can generate a lot of compiler warnings under C++17; v2.5 fixes these.

Note 1: I am unsure about the issues with keeping in sync with Ubuntu 20.04; tagging @alexreinking for comments

Note 2: the current version of PyBind11 is actually v2.6, but it has many more changes and upgrading looks nontrivial; deliberately holding off on that upgrade for now.


alexreinking:
This is probably fine since the bindings code still compiles against the Ubuntu-supplied version. We should be careful not to rely on any 2.5-specific features, though.

alexreinking (approved):

PR #5428, opened 2020-11-03
Use new flag for clang tidy behavior in cmake

jlaxson:
Every once in a while it seems that CMAKE_EXPORT_COMPILE_COMMANDS gets enabled for me, possibly by VSCode, and the build grinds to a halt. Setting it to false in the cache gets everything working again.  It seems like there are valid applications for a compilation database even when doing normal builds (such as knowing flags for parsing with other tools).  This gives the clang-tidy behavior its own flag.


alexreinking:
Please change the name of the variable to `Halide_CLANG_TIDY_BUILD` and document the change in `README_cmake.md`.

jlaxson:
@alexreinking done

steven-johnson (approved):
Thanks for the fix!

PR #5433, opened 2020-11-04
is_zero is a compile-time check if it's the constant zero

abadams:
I think for this code we want a runtime check if it's zero or not.

@rootjalex

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -653,7 +653,7 @@ private:
                 interval.max = b.max - make_one(t);
             } else if (b.max.type().is_uint()) {
                 // if b.max = 0 then result is [0, 0], else [0, b.max - 1]
-                interval.max = select(is_zero(b.max), make_zero(t), b.max - make_one(t));
+                interval.max = select(b.max == make_zero(t), make_zero(t), b.max - make_one(t));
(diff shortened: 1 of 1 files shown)


vksnk (approved):

src/Bounds.cpp
```diff
@@ -653,7 +653,7 @@ class Bounds : public IRVisitor {
                 interval.max = b.max - make_one(t);
             } else if (b.max.type().is_uint()) {
                 // if b.max = 0 then result is [0, 0], else [0, b.max - 1]
-                interval.max = select(is_zero(b.max), make_zero(t), b.max - make_one(t));
+                interval.max = select(b.max == make_zero(t), make_zero(t), b.max - make_one(t));
```
steven-johnson, line 656: I had to think a bit before realizing the nature of the change. Maybe we should rename `is_zero` and friends to make this more clear?

abadams: Yeah... maybe is_const_zero?

abadams: (and is_const_one)

steven-johnson: SGTM

rootjalex:
Yes, that is what we want. My bad, I should have checked is_zero more closely.

PR #5436, opened 2020-11-05
is_zero -> is_const_zero

abadams:
To prevent future confusion about what these do


steven-johnson (approved):
LGTM pending green.

steven-johnson:
grr, why is buildbot so flaky about refusing to pick up changes

steven-johnson:
(if this is building correctly for you locally then OK to merge I guess)

PR #5438, opened 2020-11-09
add conditions for one-sided bounds to LT/GT, LE/GE, EQ, NEQ

rootjalex:
Discussed with @abadams . Many inequality and equality operators can be bounded even if their operands are not fully bounded. e.g. a = [_, a1] and b = [b0, _] with a1 <= b0 implies a <= b can be true, even when _ means unboundedness on that side of the interval. This also tightens the output of one of the test cases.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -729,14 +729,14 @@ private:
 
         // a.max <(=) b.min implies a <(=) b, so a <(=) b is at least
         // as true as a.max <(=) b.min. This does not depend on a's
-        // lower bound or b's upper bound
+        // lower bound or b's upper bound.
         if (a.has_upper_bound() && b.has_lower_bound()) {
             interval.min = Cmp::make(a.max, b.min);
         }
 
         // a <(=) b implies a.min <(=) b.max, so a <(=) b is at most
         // as true as a.min <(=) b.max. This does not depend on a's
-        // upper bound or b's lower bound
+        // upper bound or b's lower bound.
         if (a.has_lower_bound() && b.has_upper_bound()) {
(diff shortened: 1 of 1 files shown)


abadams:
There are 6 new cases here. They look plausible to me. Have all 6 been formally verified?

rootjalex:
Yes, all were verified using z3.

rootjalex:
Edit to the example given: a = [_, a1] and b = [b0, _] with a1 <= b0 implies a <= b _is_ true, but if !(a1 <= b0) then a <= b can be false or true, therefore a1 <= b0 is sufficient but not necessary.

src/Bounds.cpp
```diff
@@ -701,31 +701,45 @@ class Bounds : public IRVisitor {
...
         // bools, treating false < true, '<=' is in fact
         // implication. So we want conditions min and max such that
         // min implies val implies max.  So min should be a sufficient
         // condition, and max should be a necessary condition.
 
-        interval.min = Cmp::make(a.max, b.min);
-        interval.max = Cmp::make(a.min, b.max);
+        // a.max <(=) b.min implies a <(=) b, so a <(=) b is at least
+        // as true as a.max <(=) b.min. This does not depend on a's
+        // lower bound or b's upper bound
```
steven-johnson, line 732: Nit: add period to end of sentence, here and below

rootjalex: fixed in 2ee5232

abadams:
Cloned for the bots in #5440 but they're taking their sweet time getting to it.

steven-johnson:
CLone is green

abadams (approved):

PR #5441, opened 2020-11-09
Allow creation of IntImm/UIntImm with any number of bits up to 64

vksnk:
Context: I needed to enable the integer types with 24 and 48 bits, but after a discussion with @dsharletg, we thought it might be better to just remove the check for explicit number of bits completely.


dsharletg (commented):
I think it makes sense to remove this, but should we add it in more backend specific places instead? e.g. CodeGen_LLVM and CodeGen_C visitor of IntImm/UIntImm?

abadams:
I'd make it t.bits() >= 1 as well. I think other things are going to break for non-power-of-two integers, but this is a reasonable thing to do. upgrade_type_for_storage and upgrade_type_for_arithmetic in CodeGen_LLVM are probably going to need to account for these as well.

steven-johnson:
Ignore the two pending checks, they no longer exist :-)

vksnk:
Question: initially I was planning to add an assert to the CodeGen_LLVM  which is very similar to what IntImm::make had (i.e. specifically check for allowed bit widths), but it seems like if we make upgrade_type_for_arithmetic to round up number of bits for ints to the next power of two and add a call to it at CodeGen_LLVM::visit(const IntImm *op) then we don't need to assert at all?

abadams:
That seems like it might work, but you probably also need to touch upgrade_type_for_storage

steven-johnson:
What's left to be done on this?

vksnk:
Sorry for the delay on this, I am going to modify upgrade_type_for_storage and upgrade_type_for_arithmetic to account for non-power of two number of bits.

vksnk:
I've modified upgrade_type_for_arithmetic and upgrade_type_for_storage, but it's not enough to make it work. One of the issues is that upgrade_type_for_storage is usually followed by *reinterpret* cast which only works on types which have the same number of bits, so for example,  if we upgrade 48-bit to 64-bit, it will can complain that such cast is not allowed (on the positive side the error message makes more sense now). That being said, I am not sure how does it work for the bool type which is also upgraded from 1-bit to 8-bit? I've tried to add some debug prints into that branch, but they don't seem to trigger, is it handled in some other place?

I think, if desired behavior is to automatically upgrade the types if they are not supported, it might be easier to write a pass which goes over IR and rewrites/upgrades all non-power-of-two-number-of-bits types.

steven-johnson:

vksnk:
There is definitely a lot more work required if we want to properly support non-power-of-two-number-of-bits types on platforms where they are not available (for example, by rewriting/upgrading all non-power-of-two-number-of-bits types). However, this does work on platforms where these types are supported, so in this case this PR is ready for review. 

Thing to note is that it was already possible before to have an expression with non-power-of-two-number-of-bits type (by casting for example) and this PR only extends it by allowing creation of constants with this type (so in a way it won't become more broken than before, maybe a bit more consistent if anything). Also, I think error messages for the codegens which don't support these types are better now.

src/CodeGen_LLVM.cpp
```diff
@@ -1466,10 +1466,28 @@ void CodeGen_LLVM::codegen(const Stmt &s) {
...
 }
+namespace {
+
+bool is_power_of_two(int x) {
+    return (x & (x - 1)) == 0;
+}
+
+int next_power_of_two(int x) {
+    for (int p2 = 1;; p2 *= 2) {
+        if (p2 >= x) {
```
dsharletg, line 1477: I think this could theoretically loop forever... the max value of int is not a power of 2, so it could overflow and wrap around and keep trying.

dsharletg (approved):

vksnk:
Failures are unrelated (lens_blur_filter ran out of memory on GPU), merging.

src/CodeGen_LLVM.cpp
vksnk: Good point, replaced it with the one which ends.

PR #5448, opened 2020-11-13
Revert "Push lets near their uses (whenever possible) while CSEing +="

dsharletg:
Reverts halide/Halide#5387

This change appears to be causing some of our code to crash on device. I'd like to revert it and I'll work on producing a test case to reproduce the issue.

### DIFF
--- a/src/CSE.cpp
+++ b/src/CSE.cpp
@@ -12,7 +12,6 @@ namespace Internal {
 
 using std::map;
 using std::pair;
-using std::set;
 using std::string;
 using std::vector;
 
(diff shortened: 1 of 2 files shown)


steven-johnson (approved):
LGTM pending green

pranavb-ca:
@dsharletg  - will take look if you can give me a reproducer.

dsharletg:
As you could probably guess, I think this change was only uncovering an existing miscompile issue. I think the reproducer I'm generating now is independent of this change (but I'd still like to revert this until we can resolve it).

PR #5452, opened 2020-11-16
Teach halide_host_cpu_count to obey CPU affinity masks on Linux

Infinoid:
`schedtool` can limit an application to only run on a certain set of CPUs/threads.
The Halide runtime thread count should respect that.

### DIFF
--- a/src/runtime/linux_host_cpu_count.cpp
+++ b/src/runtime/linux_host_cpu_count.cpp
@@ -3,8 +3,22 @@
 extern "C" {
 
 extern long sysconf(int);
+extern int sched_getaffinity(int32_t pid, size_t cpusetsize, uint32_t *cpuset);
 
 WEAK int halide_host_cpu_count() {
-    return sysconf(84);
+    uint32_t count = 0, cpuset[32];  // enough bits for 1024 CPUs
+    memset(cpuset, 0, sizeof(cpuset));
+    int rv = sched_getaffinity(0, sizeof(cpuset), cpuset);
+    if (rv == 0) {
+        for (size_t word = 0; word < sizeof(cpuset) / sizeof(cpuset[0]); word++) {
+            count += __builtin_popcount(cpuset[word]);
+        }
+    }
+    if (rv < 0 || count == 0) {
+        return sysconf(84);
+    } else {
+        return count;
+    }
 }
-}
+
+}  // extern "C"


Infinoid:
I tested this by calculating a dummy workload, parallelizing it and turning profiling on.  The thread count reported in the profile should match the number of available threads.

Here is the test program I used: https://gist.github.com/Infinoid/7d3670ba79bc893191ff149d861b1e3b

With no `schedtool` restrictions, I got `threads: 62.429`.
When restricted to 3 CPUs, I got `threads: 2.983`.
When restricted to 7 CPUs, I got `threads: 6.953`.

I was able to build and test it successfully on linux-x86-32 and linux-x86-64, glibc and musl libc.  (The musl builds required commenting out some unrelated `backtrace()` stuff.)

Infinoid:
For compatibility's sake, I should note that the `sched_getaffinity` function in its current form was added in glibc 2.3.4, released in January 2005.  It is linux-specific, but I believe this source file is linux-specific too.

src/runtime/linux_host_cpu_count.cpp
```diff
@@ -3,8 +3,27 @@
...
 
 extern long sysconf(int);
+extern int sched_getaffinity(int32_t pid, size_t cpusetsize, unsigned long *cpuset);
 
 WEAK int halide_host_cpu_count() {
-    return sysconf(84);
-}
+    unsigned long count = 0, cpuset[8];  // enough bits for 256 CPUs on 32 bit, or 512 CPUs on 64 bit
+    memset(cpuset, 0, sizeof(cpuset));
+    int rv = sched_getaffinity(0, sizeof(cpuset), cpuset);
```
dsharletg, line 11: Doesn't this only return the affinity of the calling thread? If someone were to write a program that set the affinity of thread 0 to 0, thread 1 to 1, ..., thread N to N (something I've thought about doing but never done), and then called halide_host_cpu_count on one of those threads, I think it would return 1, which seems wrong.

Infinoid: Possibly.  But I don't think that's a valid use case.  Please correct me if I'm wrong.

I think the Halide runtime uses this function to determine how many worker threads to spawn.  Any other threads you spawn outside of Halide may have their own affinities, but Halide won't ever see or use those.  Halide will only run on the thread that calls it, and any threads spawned by that one, which should inherit the affinity of that one.  Correct?

So let's say you fuss with thread affinities before calling the Halide runtime, setting the affinity of that thread to 1 single CPU.  Then, any additional threads spawned by the Halide runtime will be stuck on that one CPU.  Do you really want to be spawning 64 threads in that case?  I think 1 is a more sensible answer in that case.  Halide was given one CPU to work with, so it should run with one thread.

src/runtime/linux_host_cpu_count.cpp
```diff
@@ -3,8 +3,27 @@
 extern "C" {
 
 extern long sysconf(int);
+extern int sched_getaffinity(int32_t pid, size_t cpusetsize, unsigned long *cpuset);
 
 WEAK int halide_host_cpu_count() {
-    return sysconf(84);
-}
+    unsigned long count = 0, cpuset[8];  // enough bits for 256 CPUs on 32 bit, or 512 CPUs on 64 bit
```
abadams, line 9: I thought a long was always 32-bits. Please just use a uint32_t to make it explicit, and change the declared signature of sched_getaffinity to match. The cpusetsize arg is measured in bytes anyway.

src/runtime/linux_host_cpu_count.cpp
```diff
@@ -3,8 +3,27 @@
...
+    memset(cpuset, 0, sizeof(cpuset));
+    int rv = sched_getaffinity(0, sizeof(cpuset), cpuset);
+    if (rv == 0) {
+        size_t word, bit;
+        for (word = 0; word < sizeof(cpuset) / sizeof(cpuset[0]); word++) {
+            for (bit = 0; bit < sizeof(cpuset[0]) * 8; bit++) {
```
abadams, line 15: Just use __builtin_popcount here to count the number of set bits

src/runtime/linux_host_cpu_count.cpp
Infinoid: Done.

A long is always 32 bits on Windows, but not Linux.  https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models
 (LLP64 vs LP64)

src/runtime/linux_host_cpu_count.cpp
Infinoid: Done.

src/runtime/linux_host_cpu_count.cpp
abadams: Good to know. The runtime module is compiled for a generic OS, not linux, so we should definitely avoid any types that have OS-dependent sizes.

abadams: (despite being linux-specific code)

zvookin:
Cannot happen without docs and tests.

That said, this does not work correctly. The thread that created a worker thread in the thread pool is not necessarily the one that initialized the thread count, nor is it necessarily one that Halide created itself. (A Halide filter can be called by an arbitrary thread in the process and if it involves parallel work, it can add threads to the thread pool.) This PR presumes that all threads in the thread pool descend from a parent with the same affinity mask as the thread that called  ```default_desired_num_threads```.

an alternative would be to add a call that sets the thread affinity for the Halide thread pool. All threads created would get that thread affinity. The number of desired threads would be taken from that mask value. We'd likely also want to make sure the thread pool support handles multiple thread pools a little better than it does via ```user_context``` so one could farm different executions out to different sets of processors.

Infinoid:
> (A Halide filter can be called by an arbitrary thread in the process and if it involves parallel work, it can add threads to the thread pool.)

Oh, can the Halide thread-pool survive past the end of one Halide function, to be reused in another, with possibly different affinity?  If so, I think I see your point, this patch could cause problems with that.

> This PR presumes that all threads in the thread pool descend from a parent with the same affinity mask as the thread that called default_desired_num_threads.

You're right, I have made that assumption, it has been the case for all of my Halide usage so far.  (My apps are following a pretty simple execution model.)

My goal with this PR was quite simple.  I would like be able to launch a Halide app through `schedtool`, `taskset` or `numactl` and have Halide do the right thing.  Similar to how [OpenMP obeys cpu affinities](https://github.com/gcc-mirror/gcc/blob/master/libgomp/config/linux/proc.c#L155) without requiring the user to set `OMP_NUM_THREADS`.

I think the first step is to stop answering the question "how many cpus does the system have?", and instead answer the question "how many cpus are available for Halide's use?".  That's what this patch does.  But ideally it should be determined before any runtime affinity-fanciness happens, so maybe we need to be more stateful about it.

Do you have an example of an app that does this?  (I imagine that such an application would need to be quite careful about it... it would need to cause Halide to spawn the full set of worker threads before it locks down its own thread affinities, it would need to assume Halide keeps those threads around between calls, and it wouldn't have the ability to adjust affinities of Halide's thread pool, so it would have to accept the risk that a Halide worker and an application thread can fight over the same CPU.  Or maybe it can set them up in a `define_extern` call or something.)

src/runtime/linux_host_cpu_count.cpp
```diff
@@ -3,8 +3,22 @@
 extern "C" {
 
 extern long sysconf(int);
+extern int sched_getaffinity(int32_t pid, size_t cpusetsize, uint32_t *cpuset);
 
 WEAK int halide_host_cpu_count() {
-    return sysconf(84);
-}
+    uint32_t count = 0, cpuset[8];  // enough bits for 256 CPUs
```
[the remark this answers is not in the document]
Infinoid: Done.

steven-johnson:

steven-johnson:

Infinoid:
There are clearly some problems with the approach this PR took.  I opened #5614 to track the issue, so the right fix can be determined.

PR #5455, opened 2020-11-18
Remove AndConditionOverDomain and fix Interval::everything() uses in Bounds

rootjalex:
Removal of AndConditionOverDomain mutator discussed with @abadams and @shoaibkamil . Replacement of `Interval::everything()` with `bounds_of_type(op->type)` is relevant for booleans (and Uints/Ints with `bits() <= 16`). Also fixes a bug in `VectorReduce::And/Or` visitor method that produced a non-boolean interval, and replaces it with a more optimal reduction.


src/Bounds.cpp
```diff
@@ -111,6 +111,16 @@ class Bounds : public IRVisitor {
     Bounds(const Scope<Interval> *s, const FuncValueBounds &fb, bool const_bound)
         : func_bounds(fb), const_bound(const_bound) {
         scope.set_containing_scope(s);
+
+        // Find any points that are single_points but fail is_single_point due to
+        // pointer equality checks and replace with single_points.
+        for (auto item = s->cbegin(); item != s->cend(); ++item) {
+            // Updates to Scope<T> do not invalidate iterators.
```
abadams, line 118: But we're not updating s, so we'd be fine anyway.

rootjalex: Ah right - I will remove this comment.

src/Bounds.cpp
```diff
@@ -111,6 +111,15 @@ class Bounds : public IRVisitor {
     Bounds(const Scope<Interval> *s, const FuncValueBounds &fb, bool const_bound)
         : func_bounds(fb), const_bound(const_bound) {
         scope.set_containing_scope(s);
+
+        // Find any points that are single_points but fail is_single_point due to
+        // pointer equality checks and replace with single_points.
+        for (auto item = s->cbegin(); item != s->cend(); ++item) {
+            const Interval &item_interval = item.value();
+            if (!item_interval.is_single_point() && item_interval.is_bounded() && can_prove(item_interval.min == item_interval.max)) {
```
steven-johnson, line 119: nit: we have no line length limit but this one is long enough to deserve a line break, IMHO

steven-johnson, line 119: ...huh, I hadn't realized that `is_single_point()` does a `same_as` check (ie, is the same expr) rather than a value check.

steven-johnson (approved):

src/Bounds.cpp
rootjalex: Yeah, this was added because `Interval(0, 0)` failed `is_single_point()` which produced a really strange output on one of the `and_condition_over_domain` tests. I found it by forgetting to put `simplify(bounds.min)` in my change, and got a weird condition when checking `x == 0` when `x` mapped to `Interval(0, 0)`. It wasn't incorrect, it just was much more complicated than it should have been.

rootjalex: fixed in 7a0cc0c

src/Bounds.cpp
rootjalex: Removed in fc1d4e8

PR #5459, opened 2020-11-19
Bug fix for lossless_cast with minor additions

aankit-ca:
The bug can seen for types where lossless_cast type can represent
cast->value.type() but not cast->type. For eg:

lossless_cast(UInt(16), cast(Int(8), Variable::make(UInt(16), e))) returns
(uint16)e which is incorrect.

The patch also adds lossless_cast of Mod and Ramp expressions.

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -460,6 +460,13 @@ void lossless_cast_test() {
     Type i16 = Int(16);
     Type i32 = Int(32);
     Type u8x = UInt(8, 4);
+    Type u16x = UInt(16, 4);
+    Type u32x = UInt(32, 4);
+    Expr var_i8 = Variable::make(i8, "x");
+    Expr var_i16 = Variable::make(i16, "x");
+    Expr var_u8 = Variable::make(u8, "x");
+    Expr var_u16 = Variable::make(u16, "x");
+    Expr var_u8x = Variable::make(u8x, "x");
 
     Expr e = Ramp::make(0, 1, 4);
     check(u8x, e, cast(u8x, e));
@@ -471,16 +478,16 @@ void lossless_cast_test() {
     e = x % 4;
     check(UInt(8), e, cast(UInt(8), e));
 
-    e = Variable::make(i8, "x") % make_const(i8, -128);
+    e = var_i8 % make_const(i8, -128);
     check(UInt(8), e, cast(UInt(8), e));
 
-    e = Variable::make(i16, "x") % make_const(i16, -256);
+    e = var_i16 % make_const(i16, -256);
     check(UInt(8), e, cast(UInt(8), e));
 
-    e = Variable::make(u16, "x") % make_const(u16, 256);
+    e = var_u16 % make_const(u16, 256);
     check(UInt(8), e, cast(UInt(8), e));
 
-    e = Variable::make(u8, "x") % Variable::make(u8, "y");
+    e = var_u8 % Variable::make(u8, "y");
     check(UInt(8), e, cast(UInt(8), e));
 
     e = cast(u8, x);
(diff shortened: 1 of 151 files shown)


aankit-ca:
@dsharletg @abadams PTAL

src/IROperator.cpp
```diff
@@ -566,6 +618,32 @@ Expr lossless_cast(Type t, Expr e) {
         return Shuffle::make(vecs, shuf->indices);
     }
 
+    if (const Mod *mod = e.as<Mod>()) {
+        if (lossless_cast(t, mod->b).defined()) {
```
aankit-ca, line 622: Making some changes for handling negative numbers here

aankit-ca: Done

dsharletg (commented):
@abadams any opinion on a new internal test vs. a new/modified correctness test? I lean towards a correctness test, especially because lossless_cast is already visible to tests (in IROperator.h).

src/IROperator.cpp
```diff
@@ -438,6 +439,80 @@ Expr const_false(int w) {
     return make_zero(UInt(1, w));
 }
 
+void check(const Type &t, const Expr &in, const Expr &correct) {
```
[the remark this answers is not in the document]
aankit-ca: Done

src/IROperator.cpp
```diff
@@ -438,6 +439,80 @@ Expr const_false(int w) {
...
+        << "Incorrect lossless_cast result:\nlossless_cast("
+        << t << ", "
+        << in
+        << ") gave: "
+        << result
+        << " but expected was: "
+        << correct << "\n";
+}
+
+void lossless_cast_test() {
```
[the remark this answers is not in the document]
aankit-ca: Done

aankit-ca:
Ping for review!

src/IROperator.cpp
```diff
@@ -2365,4 +2404,83 @@ Expr undef(Type t) {
                                 Internal::Call::PureIntrinsic);
 }
 
+namespace Internal {
```
abadams, line 2407: As a general rule, if it *can* be an external correctness test, it *should* be, to keep libHalide smaller. I think this code can be.

src/IROperator.cpp
```diff
@@ -566,6 +569,42 @@ Expr lossless_cast(Type t, Expr e) {
         return Shuffle::make(vecs, shuf->indices);
     }
 
+    if (const Mod *mod = e.as<Mod>()) {
+        Expr val;
+        if (e.type().is_uint()) {
+            val = simplify(mod->b + make_const(e.type(), -1));
```
abadams, line 575: What happens here and below if mod->b is zero?

src/IROperator.cpp
```diff
@@ -566,6 +569,42 @@ Expr lossless_cast(Type t, Expr e) {
...
+        } else if (e.type().is_int()) {
+            // 0 <= a%b < |b|
+            Type wide_ty = e.type().with_bits(e.type().bits() * 2);
```
abadams, line 578: In the cases where you return cast(t, e), could you just do this:
```
Expr x = cast(t, e);
if (can_prove(cast(e.type(), x) == e) {
  return x;
}
```
?

abadams: Or maybe
```
if (e.type().can_represent(t.min()) && 
    e.type().can_represent(t.max()) && 
    can_prove(e >= cast(e.type(), t.min()) && 
    e <= cast(e.type(), t.max())) 
```

Basically I'm wondering if we can do something that appeals to verified code to reason about the bounds of mods and ramps, rather than having to worry about this method being correct. It's hard to get bounds right.

src/IROperator.cpp
```diff
@@ -446,8 +447,9 @@ Expr lossless_cast(Type t, Expr e) {
     }
 
     if (const Cast *c = e.as<Cast>()) {
-        if (t.can_represent(c->value.type())) {
-            // We can recurse into widening casts.
+        // We can recurse into widening casts.
+        if (c->type.can_represent(c->value.type()) &&
+            t.can_represent(c->value.type())) {
```
dsharletg, line 452: I think you can get rid of this second part of the condition, and leave that to the recursion?

aankit-ca: Right

aankit-ca:
I got rid of Ramp and Mod for lossless_cast as per discussion with Dillon.

dsharletg (approved):

PR #5468, opened 2020-11-21
Improvements to apps/Makefile.inc

steven-johnson:
- add boilerplate for host-profile target
- rewrite the RunGen support to use a Make macro (`DEFINE_RUNGEN`) to add more flexibility; the previous approach was limited by Make's inability to have multiple wildcards. (There aren't many clients of this RunGen support in apps/, but I found it useful for an experiment so would like to capture it here for future use.)

### DIFF
--- a/apps/support/Makefile.inc
+++ b/apps/support/Makefile.inc
@@ -97,6 +97,7 @@ ANDROID_NDK_HOST_PLATFORM ?= $(shell echo `uname` | tr '[:upper:]' '[:lower:]')
 ANDROID_API_VERSION ?= 26
 
 CXX-host ?= $(CXX)
+CXX-host-profile ?= $(CXX)
 CXX-host-opencl ?= $(CXX)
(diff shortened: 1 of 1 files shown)


apps/support/Makefile.inc
```diff
@@ -97,6 +97,7 @@ ANDROID_NDK_HOST_PLATFORM ?= $(shell echo `uname` | tr '[:upper:]' '[:lower:]')
 ANDROID_API_VERSION ?= 26
 
 CXX-host ?= $(CXX)
+CXX-host-profile ?= $(CXX)
```
dsharletg, line 100: I don't think we should add -profile versions of these. It's not really a different target, and it's the start of a combinatorial hell hole (e.g. why not CXXFLAGS-host-opencl-profile? And what about CXXFLAGS-host-profile-opencl?)

steven-johnson: Eh, fair point.

PR #5469, opened 2020-11-22
Fix inconsistency between code & documentation.

tjingrant:


abadams:
Thanks. Not running on the buildbots because it's just a comment change.

abadams (approved):

PR #5472, opened 2020-11-23
Fail CMake when LLVM_LINK_LLVM_DYLIB conflicts with wasm

### LINKED ISSUE #5471: LLVM_LINK_LLVM_DYLIB and WebAssembly don't mix
LLVM's build [has an option](https://releases.llvm.org/11.0.0/docs/CMake.html), `LLVM_LINK_LLVM_DYLIB`, which links "tools" to the _shared_ LLVM library. This is ludicrous and no one should use it because it will also link the _static_ lldWasm library to the _shared_ LLVM library. Yet, the Linux version of Homebrew _does_ use this.

Fortunately, the installed CMake files for LLVM tell you if this is the case: they will set `LLVM_LINK_LLVM_DYLIB` to something true-y (`ON` on my system). So the following three things cannot be true at the same time:

1. `TARGET_WEBASSEMBLY` is enabled
2. `LLVM_LINK_LLVM_DYLIB` is enabled
3. `Halide_SHARED_LLVM` is _disabled_

We should check for this and error out if it happens.

alexreinking:
attn @JanHett -- you might want to update your submodule when this is merged.

Fixes #5471

### DIFF
--- a/dependencies/llvm/CMakeLists.txt
+++ b/dependencies/llvm/CMakeLists.txt
@@ -76,10 +76,14 @@ set(wasm_libs "")
 if (TARGET_WEBASSEMBLY)
     find_package(LLD CONFIG REQUIRED HINTS "${LLVM_DIR}/../lld")
 
+    # LLVM has a mis-feature that allows it to build and export both static and shared libraries at the same
+    # time, while inconsistently linking its own static libraries (for lldWasm and others) to the shared library.
+    # Ignoring this causes Halide to link to both the static AND the shared LLVM libs and it breaks at runtime.
+    # From issue: https://github.com/halide/Halide/issues/5471
     if (LLVM_LINK_LLVM_DYLIB AND NOT Halide_SHARED_LLVM)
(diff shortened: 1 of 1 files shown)


dependencies/llvm/CMakeLists.txt
```diff
@@ -75,6 +75,13 @@ endforeach ()
 set(wasm_libs "")
 if (TARGET_WEBASSEMBLY)
     find_package(LLD CONFIG REQUIRED HINTS "${LLVM_DIR}/../lld")
+
+    if (LLVM_LINK_LLVM_DYLIB AND NOT Halide_SHARED_LLVM)
+        message(FATAL_ERROR "LLVM was built with LLVM_LINK_LLVM_DYLIB. "
```
steven-johnson, line 80: It might be useful to reference this PR (and/or issue) in a comment here.

steven-johnson (approved):

dependencies/llvm/CMakeLists.txt
alexreinking: Will do

alexreinking: Done; merging.

PR #5474, opened 2020-11-24
Make context handling in GPU runtimes more consistent and robust.

zvookin:
Move to using common code for kernel compilation caching for CUDA, OpenCL, Metal, and D3D12 GPU runtimes. New caching endeavors to be robust in not using a kernel compiled for one context on another and uses a hash table to avoid small allocations across multiple pages of VM. OpenCL was particularly broken in that code using two contexts was almost guaranteed to fail. This PR also opens the door to allowing better client control of caching, such as setting a size limit or allowing eviction of specific kernels, and is pretty close to allowing runtime overloads of the kernel compilation itself to allow persistent caching across process invocations for GPU APIs that allow this. (The ```compile_kernel``` function in multiple files needs to be promoted to a client visible runtime overload for each GPU API.)

Tests are added to cover many kernels and more than one context. A test using multiple contexts across multiple threads both tests things that didn't necessarily work before and provides an example for a common use case.

Two small fixes to CUDA prevent a crash in a very rare error case and make device release work if the CUDA library is linked directly into the app. (The latter would have shown up as a crash due to allocation caching for static linking as the code to release allocations when freeing a context did not run.)

OpenGL and OpenGLCompute were not addressed in this PR due to both time limitations and because there are more significant issues in these runtimes around this area. OpenGL is basically a Superfund site at this point and should be deleted. OpenGLCompute may or may not be worth preserving, though similar work is needed re: how kernels are communicated to the runtime and compiled.


src/runtime/metal.cpp
```diff
@@ -782,10 +740,11 @@ WEAK int halide_metal_run(void *user_context,
         return -1;
     }
 
-    halide_assert(user_context, state_ptr);
-    module_state *state = (module_state *)state_ptr;
+    mtl_library *library;
+    bool found_libdary = compilation_cache.lookup(metal_context.device, state_ptr, library);
```
shoaibkamil, line 744: Nit: `found_libdary` -> `found_library`

steven-johnson:
> OpenGL is basically a Superfund site at this point and should be deleted. OpenGLCompute may or may not be worth preserving

See https://github.com/halide/Halide/issues/5475

Makefile
```diff
@@ -1622,11 +1628,25 @@ generator_aot_multitarget: $(BIN_DIR)/$(TARGET)/generator_aot_multitarget
 	HL_MULTITARGET_TEST_USE_NOBOUNDSQUERY_FEATURE=1 $(CURDIR)/$<
 	@-echo
 
+# gpu_multi_context_threaded has additional deps to link in
+$(BIN_DIR)/$(TARGET)/generator_aot_gpu_multi_context_threaded: $(ROOT_DIR)/test/generator/gpu_multi_context_threaded_aottest.cpp $(FILTERS_DIR)/gpu_multi_context_threaded_add.a $(FILTERS_DIR)/gpu_multi_context_threaded_mul.a $(RUNTIME_EXPORTED_INCLUDES) $(BIN_DIR)/$(TARGET)/runtime.a
+	@mkdir -p $(@D)
+	$(CXX) $(GEN_AOT_CXX_FLAGS) $(filter %.cpp %.o %.a,$^) $(GEN_AOT_INCLUDES) $(GEN_AOT_LD_FLAGS) $(OPENCL_LD_FLAGS) $(CUDA_LD_FLAGS) -o $@
+
+$(BIN_DIR)/$(TARGET)/generator_aotcpp_gpu_multi_context_threaded: $(ROOT_DIR)/test/generator/gpu_multi_context_threaded_aottest.cpp $(FILTERS_DIR)/gpu_multi_context_threaded_add.halide_generated.cpp $(FILTERS_DIR)/gpu_multi_context_threaded_mul.halide_generated.cpp $(RUNTIME_EXPORTED_INCLUDES) $(BIN_DIR)/$(TARGET)/runtime.a
```
steven-johnson, line 1636: nit: very long line might benefit from splitting to multiple lines

src/runtime/d3d12compute.cpp
```diff
@@ -3023,8 +3005,9 @@ WEAK int halide_d3d12compute_run(void *user_context,
     StartCapturingGPUActivity();
 #endif
 
-    halide_assert(user_context, state_ptr);
-    module_state *state = (module_state *)state_ptr;
+    d3d12_library *library;
```
steven-johnson, line 3008: Does `lookup()` guarantee null-init on failure? If not, init to null here seems prudent.

src/runtime/gpu_context_common.h
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    bool lookup(ContextT context, void *state_ptr, ModuleStateT &module_state) {
+        ScopedMutexLock lock_guard(&mutex);
+        uint32_t id = (uint32_t)(uintptr_t)state_ptr;
+        ModuleStateT *mod_ptr;
+        if (find_internal(context, id, mod_ptr)) {
+            module_state = *mod_ptr;
+            return true;
+        }
```
steven-johnson, line 94: Might be safer to reliably set module_state = null upon failure, so caller can't mistakenly use garbage?

test/correctness/gpu_many_kernels.cpp
```diff
@@ -0,0 +1,92 @@
+#include "Halide.h"
+#include <algorithm>
+
+#include "halide_benchmark.h"
+
+// This test makes sure GPU runtimes can handle many different small
+// kernels and can handle releasing a device context and making a new
+// one and still have many kernels work. This is needed due to kernel
+// compilation caching mechanisms int he GPU runtimes.
```
steven-johnson, line 9: in the

steven-johnson (commented):
Looks good from a quick skim -- gonna wait for buildbots to look clean(er) before reviewing more.

src/runtime/gpu_context_common.h
zvookin: The code is templated in a way where it is not entirely guaranteed a nullptr value fits the type here. Ideally it would use an optional return, but we don't have one handy.

test/correctness/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 tests(GROUPS correctness_multi_gpu
-      SOURCES
+a      SOURCES
```
steven-johnson, line 2: I assume the `a` is a typo

src/runtime/gpu_context_common.h
```diff
@@ -0,0 +1,192 @@
...
+struct GPUCompilationCache {
+    struct CachedCompilation {
+        ContextT context;
+        ModuleStateT module_state;
+        uint32_t kernel_id;
+    };
+
+    halide_mutex mutex;
+
+    static constexpr float kLoadFactor = .5f;
```
steven-johnson, line 17: nit, but seems like we should prefer either `{0}` or `=0` consistently

src/runtime/gpu_context_common.h
```diff
@@ -0,0 +1,192 @@
...
+                return true;
+            }
+        }
+        // This is a logic error that should never occur. It means the table is
+        // full, but it should have been resized.
+        halide_assert(nullptr, false);
+        return false;
+    }
+
+    bool find_internal(ContextT context, uint32_t id, ModuleStateT *&module_state) {
```
steven-johnson, line 67: nit: Do we have a MUST_USE_RESULT macro available in our runtime code? If so, marking `find_internal` and `lookup` that way seems prudent.

Makefile
```diff
@@ -1622,11 +1628,30 @@ generator_aot_multitarget: $(BIN_DIR)/$(TARGET)/generator_aot_multitarget
 	HL_MULTITARGET_TEST_USE_NOBOUNDSQUERY_FEATURE=1 $(CURDIR)/$<
 	@-echo
 
+# gpu_multi_context_threaded has additional deps to link in
+$(BIN_DIR)/$(TARGET)/generator_aot_gpu_multi_context_threaded: $(ROOT_DIR)/test/generator/gpu_multi_context_threaded_aottest.cpp \
```
steven-johnson, line 1632: nit: long lines are loooooong

src/runtime/cuda.cpp
```diff
@@ -1133,9 +1059,9 @@ WEAK int halide_cuda_run(void *user_context,
 #endif
 
     halide_assert(user_context, state_ptr);
-    module_state *loaded_module = find_module_for_context((registered_filters *)state_ptr, ctx.context);
-    halide_assert(user_context, loaded_module != nullptr);
-    CUmodule mod = loaded_module->module;
+    CUmodule mod;
```
steven-johnson, line 1062: = nullptr;

src/runtime/gpu_context_common.h
```diff
@@ -0,0 +1,196 @@
+#include "printer.h"
+#include "scoped_mutex_lock.h"
+
+namespace Halide {
+namespace Internal {
+
+template<typename ContextT, typename ModuleStateT>
+struct GPUCompilationCache {
+    struct CachedCompilation {
+        ContextT context;
```
steven-johnson, line 10: should this struct also have default init values?

steven-johnson (approved):
LGTM with nits

zvookin:
> super-duper-style-nit: seems like `constexpr bool HAS_MULTIPLE_CONTEXTS = true;` (etc) would be preferable?

Possibly will be used to control conditional compilation at some point. Really it should probably nerf the test entirely outside of GPU APIs it can make contexts for. But it does get a little coverage on e.g. Metal and Direct3d so...

PR #5476, opened 2020-11-25
Fixes to bounds inference on shift_left

rootjalex:
One of the `shift_left` rules (for the min of an interval) was incorrect, but can only be replaced by weaker rules. All of the added rules have been formally verified.
@abadams @dsharlet


src/Bounds.cpp
```diff
@@ -1235,8 +1235,27 @@ class Bounds : public IRVisitor {
...
+                                        } else {
+                                            // If b could be positive, then the magnitude might increase.
+                                            interval.min = min(a_interval.min, a_interval.min << b_interval.max);
+                                        }
+                                    }
+                                }
+                                // TODO: Are there any other cases we can handle here?
+                            } else if (a_interval.has_lower_bound() &&
+                                    b_interval.has_upper_bound() &&
+                                    can_prove(a_interval.min >= 0 && b_interval.max <= 0)) {
```
abadams: Why does this require b_interval.max <= 0?

rootjalex: Ah, because I mistakenly thought that it did, and z3 didn't complain that there was an extra + unnecessary condition. Just verified that this works without `b` being bounded.

rootjalex: Fixed in c4b7b8a

PR #5477, opened 2020-11-25
Fixes to bounds inference on shift_left

rootjalex:
One of the shift_left rules (for the min of an interval) was incorrect, but can only be replaced by weaker rules. All of the added rules have been formally verified.
@abadams @dsharlet

Edit: I closed the original draft for this PR (#5476) to get the buildbots to run tests


rootjalex:
Note: when these changes are merged, they should be squashed. Some of the intermediate commits have buggy code as I had some commits that relied on shift behaviors that are not currently expected behavior (dealing with overflow).

src/Bounds.cpp
```diff
@@ -1235,8 +1235,26 @@ class Bounds : public IRVisitor {
                                        !b_interval.min.type().is_uint() &&
                                        can_prove(b_interval.min < 0 &&
                                                  b_interval.min > -t.bits())) {
-                                interval.min = a_interval.min >> abs(b_interval.min);
+                                if (a_interval.min.type().is_uint() || can_prove(a_interval.min >= 0)) {
+                                    interval.min = a_interval.min >> abs(b_interval.min);
+                                } else if (can_prove(a_interval.min < 0)) {
```
abadams, line 1240: Seems weird to prove a_interval.min < 0. That's just that it *could* be negative. Did you mean a_interval.max?

src/Bounds.cpp
```diff
@@ -1235,8 +1235,26 @@ class Bounds : public IRVisitor {
...
+                                    if (b_interval.has_upper_bound()) {
```
abadams, line 1241: I'm a bit puzzled by b_interval.min never shows up in here. We know it has a lower_bound because we're nested inside an if that requires it

src/Bounds.cpp
```diff
@@ -1235,8 +1235,26 @@ class Bounds : public IRVisitor {
                                        !b_interval.min.type().is_uint() &&
                                        can_prove(b_interval.min < 0 &&
                                                  b_interval.min > -t.bits())) {
-                                interval.min = a_interval.min >> abs(b_interval.min);
+                                if (a_interval.min.type().is_uint() || can_prove(a_interval.min >= 0)) {
```
steven-johnson, line 1238: Is it a significant win in terms of execution time to check for is_uint() here? (I'd hope that can_prove(e>=0) would always catch these cases since all uints are >=0)

steven-johnson (commented):
I don't quite follow the logic of the intended fixes, but my brain is tired. I'll look again tomorrow.

src/Bounds.cpp
rootjalex: If the value *can* be negative, then the smallest possible value that *could* be produced by the shift is negative, but you can't produce a negative number from `a << b` if `a >= 0`, so it doesn't matter what `a`'s upper bound is

src/Bounds.cpp
rootjalex: You're totally right, I don't think this optimization requires that `b` is lower bounded at all. Let me verify that to be sure first.

src/Bounds.cpp
```diff
@@ -1235,8 +1235,26 @@ class Bounds : public IRVisitor {
                                        !b_interval.min.type().is_uint() &&
```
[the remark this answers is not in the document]
rootjalex: I believe that this should be the case, but @abadams might contradict me. I have been operating under the assumption that that invariant is maintained.

src/Bounds.cpp
rootjalex: I imagine it would be at least decent - `can_prove` calls `remove_likelies`, `common_subexpression_elimination` and `simplify` which could be quite expensive if `a_interval.min` is a complicated expression, but `is_uint()` should be extremely fast.

rootjalex:
@steven-johnson Any idea what the test failures are caused by?

src/Bounds.cpp
rootjalex: You can think of it as splitting `a_interval` into two ranges, negative and non-negative. The non-negative interval (if it exists), cannot produce a negative result, no matter what the right-hand side is. Only the negative interval could produce a negative value, and in this block, that negative interval would be `[a_interval.min, 0)`, whose most negative value depends on whether or not you can increase the magnitude of any value on that interval or not.

src/Bounds.cpp
rootjalex: Verified. Fixed in e8bcdf2

src/Bounds.cpp
steven-johnson: IMHO this explanation is worth capturing in a comment in the code, since it apparently wasn't obvious to two code reviewers

src/Bounds.cpp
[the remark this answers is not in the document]
steven-johnson: (Would be nice to have that documented somewhere, though that's orthogonal to this PR.)

I ask mainly because code lower down seemed to be optimized based on that assumption, but it wasn't clear to me whether deliberate or not.

src/Bounds.cpp
steven-johnson: in that case, should you also add it on line 1254? 

`a_interval.min.type.is_uint() || can_prove(a_interval.min >= 0)`

steven-johnson: (I also now idly wonder if it would be worth having can_prove check for the expr being `foo >= 0` or `foo < 0` and doing an early-out if the LHS is uint...)

rootjalex: > in that case, should you also add it on line 1254?

sure thing. great catch, thanks!

src/Bounds.cpp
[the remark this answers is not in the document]
rootjalex: I agree, all of the changes I have made to bounds inference assumes that the type of the min, type of the max, and type of the operation are all equal.

src/Bounds.cpp
rootjalex: Fixed in 3055665

src/Bounds.cpp
rootjalex: Sounds good. This code has moved a bit from some of Andrew's other comments, but a comment on this has been added in 1a8930e

steven-johnson:

rootjalex:
None, it can be merged if you and @abadams are happy with it. Sorry for the delay!

steven-johnson:
The failures all appear unrelated, good to merge I think

steven-johnson (approved):

PR #5478, opened 2020-11-25
Add Type::widen and Type::narrow helpers.

dsharletg:
Add helpers for widening and narrowing types. This has a little utility on master, but is mainly peeling off a change from another WIP branch.

### DIFF
--- a/src/Type.h
+++ b/src/Type.h
@@ -357,11 +357,11 @@ public:
     }
 
     /** Return Type with the same type code and number of lanes, but with twice as many bits. */
-    Type widen() const {
+    Type wide() const {
         return with_bits(bits() * 2);
     }
 
-    /** REturn Type with the same type code and number of lanes, but with half as many bits. */
+    /** Return Type with the same type code and number of lanes, but with half as many bits. */
     Type narrow() const {
         return with_bits(bits() / 2);
     }
--- a/src/Type.h
+++ b/src/Type.h
@@ -356,6 +356,16 @@ public:
         return Type(code(), bits(), new_lanes, handle_type);
     }
 
+    /** Return Type with the same type code and number of lanes, but with twice as many bits. */
+    Type widen() const {
+        return with_bits(bits() * 2);
+    }
+
+    /** REturn Type with the same type code and number of lanes, but with half as many bits. */
+    Type narrow() const {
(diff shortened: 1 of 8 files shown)


abadams:
```git grep 'with_bits.*[*/].*2'```

finds a few more opportunities

File: src/Type.h [hunk 1]
```diff
@@ -356,6 +356,16 @@ struct Type {
         return Type(code(), bits(), new_lanes, handle_type);
     }
 
+    /** Return Type with the same type code and number of lanes, but with twice as many bits. */
+    Type widen() const {
+        return with_bits(bits() * 2);
+    }
+
+    /** REturn Type with the same type code and number of lanes, but with half as many bits. */
```
zvookin, line 364: "REturn" -> "Return".

dsharletg:
Thanks, I messed up my grep command :) 

I also renamed `widen` to `wide`, so it doesn't sound like it works in-place.

src/Type.h
```diff
@@ -356,6 +356,16 @@ struct Type {
         return Type(code(), bits(), new_lanes, handle_type);
     }
 
+    /** Return Type with the same type code and number of lanes, but with twice as many bits. */
+    Type wide() const {
```
steven-johnson, line 360: Stylistically, `widen()` seems like a better name than `wide()` here IMHO.

dsharletg: `widen` sounds like it works in-place? When reading the code, I would expect `t.widen()` to modify `t`, not return a wider `t`.

dsharletg: The other thing I considered was `wider` but thought `wide` was better.

steven-johnson: > I would expect t.widen() to modify t, not return a wider t

What can I say, my expectations are the opposite. To me `wide()` reads as a noun, not a verb, which is weird for a method. I can live with `wide()` but might be worth a straw poll on Gitter to see if there is a strong consensus either way.

(Complete threadjack but `Type` is a class that I'd consider just making immutable to avoid this sort of confusion; it's small enough that just making mutated copies seems sensible.)

dsharletg: I renamed it, I think either one works.

steven-johnson (approved):

PR #5480, opened 2020-11-25
Add as_intrinsic helper

dsharletg:
There's a pretty common pattern (especially in a WIP branch) of doing something like:

```
const Call *c = e.as<Call>();
if (c && (c->is_intrinsic(Call::X) || c->is_intrinsic(Call::Y))) {
    ...
}
```

The call declaration can't be made part of the if, because we can't assign it and use it as part of another condition. This PR adds a helper for this:

```
if (const Call *c = Call::as_intrinsic(e, {Call::X, Call::Y}) {
    ...
}
```

### DIFF
--- a/src/ScheduleFunctions.cpp
+++ b/src/ScheduleFunctions.cpp
@@ -1056,8 +1056,8 @@ protected:
         vector<pair<string, Expr>> containers;
         while (1) {
             if (const LetStmt *l = body.as<LetStmt>()) {
-                const Call *call = Call::as_intrinsic(l->value, {Call::promise_clamped});
-                if (!call && !is_pure(l->value)) {
+                const Call *promise_clamped = Call::as_intrinsic(l->value, {Call::promise_clamped});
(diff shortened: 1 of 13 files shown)


src/ScheduleFunctions.cpp
```diff
@@ -1058,9 +1056,8 @@ class InjectFunctionRealization : public IRMutator {
         vector<pair<string, Expr>> containers;
         while (1) {
             if (const LetStmt *l = body.as<LetStmt>()) {
-                const Call *call = l->value.as<Call>();
-                if (!(call && call->is_intrinsic(Call::promise_clamped)) &&
-                    !is_pure(l->value)) {
+                const Call *call = Call::as_intrinsic(l->value, {Call::promise_clamped});
```
abadams, line 1059: Some of these variable names are misleading now when you just read the if condition. How about:

call -> promise_clamped

abadams (approved):

steven-johnson (approved):

PR #5482, opened 2020-11-26
Improve speed of testing apps/

steven-johnson:
- Skip all app tests that are labeled as 'benchmarks'
- Specify `--build-noclean` to avoid unnecessary full rebuilds

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -48,11 +48,15 @@ function(add_app_test NAME)
         "-DLLVM_DIR=${LLVM_DIR}"
         "-DHalide_TARGET=${Halide_TARGET}")
 
-    # Note that we skip all tests that have the 'benchmark' label; some of the apps
-    # have benchmarks that are in the form of tests, and running them via this path
-    # isn't very useful, since we aren't guaranteed that the machine won't be under
-    # load with other tests. (Also, some of the apps have extensive benchmarks that
-    # take quite a while to run; e.g. apps/linear_algebra, which has over 500.)
+    # As a band-aid to keep test times under control, we explicitly skip tests
+    # that are labeled with `slow_tests` -- currently, this label only applies
+    # to a number of benchmarks under apps/linear_algebra, but could be applied
+    # to others as necessary. (The issue here is that --build-and-test will treat
+    # the entire subproject as a single 'test', with a single timeout applied to all
+    # of it, and apparently no way to get useful progress information bubbled out.
+    # It may be that we need to rethink how we handle building/testing the apps
+    # under CMake, in which case the `slow_tests` label will hopefully become
+    # useless and removed.)
     #
     # Note also that we specify `--build-noclean` explicitly; apparently the default
     # for --build-and-test is to 'clean' first, which shouldn't be necessary here.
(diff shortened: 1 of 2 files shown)


apps/CMakeLists.txt
```diff
@@ -48,17 +48,26 @@ function(add_app_test NAME)
...
         "-DHalide_TARGET=${Halide_TARGET}")
 
+    # Note that we skip all tests that have the 'benchmark' label; some of the apps
+    # have benchmarks that are in the form of tests, and running them via this path
+    # isn't very useful, since we aren't guaranteed that the machine won't be under
+    # load with other tests. (Also, some of the apps have extensive benchmarks that
+    # take quite a while to run; e.g. apps/linear_algebra, which has over 500.)
+    #
+    # Note also that we specify `--build-noclean` explicitly; apparently the default
+    # for --build-and-test is to 'clean' first, which shouldn't be necessary here.
```
abadams, line 58: What does this mean for things like local laplacian? It just has one binary, which runs a benchmark. In the cmake build it just has the label "internal_app_tests"

steven-johnson: That test isn't labeled with `benchmark`, so it isn't affected.

abadams:
I was confused because no current app has a test labelled "benchmark", so I wasn't sure what you were excluding.

It looks like you're introducing a new tag specifically for the linear algebra app (for now) to exclude it on the bots. Maybe "benchmark" is a misnomer then, because all the other existing benchmarks aren't excluded. Perhaps "skip_test_on_buildbots"?

alexreinking (approved):

alexreinking:
> It looks like you're introducing a new tag specifically for the linear algebra app (for now) to exclude it on the bots. Maybe "benchmark" is a misnomer then, because all the other existing benchmarks aren't excluded. Perhaps "skip_test_on_buildbots"?

We should set up nightly benchmarking to catch performance regressions, anyway. The name `benchmark` would make sense there.

abadams:
It would, but the way the "benchmark" tag is used in this PR is not for benchmarks. It's just to exclude the single long-running linear algebra thing. That's why I think it's a misnomer. All the other benchmarks are not tagged "benchmark", and given this PR it would be bad if that tag were added, because we do want those running on the bots for the purpose of check-this-doesn't-crash.

alexreinking:
> [...] is not for benchmarks. It's just to exclude the single long-running linear algebra thing. 

Are we talking about the same thing? It is for the 3rd party linear algebra benchmarks (and for the Halide-BLAS benchmarks as well).

> because we do want those running on the bots for the purpose of check-this-doesn't-crash

The ones in `tests/performance` check for relative runtimes, not just doesn't-crash.

abadams:
~All of the apps are benchmarks. They are used as benchmarks in papers. But this new "benchmark" label does not and should not apply to them (given that it turns off testing). That's why I think it's a misnomer.

steven-johnson:
Yeah, this is a band-aid that needs rethinking -- as a band-aid it's not unreasonable but needs to avoid looking like something general. (There are more general issues with apps/ that I'll discuss elsewhere.) I'll revisit this today.

steven-johnson:
PTAL

abadams (approved):

PR #5491, opened 2020-11-30
Simplify signed shifts more strongly

dsharletg:
This PR does two things: simplify shifts that are known to be positive or negative, and avoid creating signed shifts in the first place when the RHS is a constant.

### DIFF
--- a/src/Simplify_Call.cpp
+++ b/src/Simplify_Call.cpp
@@ -176,7 +176,7 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
 
         // If we know the sign of this shift, change it to an unsigned shift.
         if (b_info.min_defined && b_info.min >= 0) {
-            b = cast(b.type().with_code(halide_type_uint), b);
+            b = mutate(cast(b.type().with_code(halide_type_uint), b), nullptr);
         } else if (b_info.max_defined && b_info.max <= 0) {
             result_op = Call::get_intrinsic_name(op->is_intrinsic(Call::shift_right) ? Call::shift_left : Call::shift_right);
             b = mutate(cast(b.type().with_code(halide_type_uint), -b), nullptr);
--- a/src/Simplify_Call.cpp
+++ b/src/Simplify_Call.cpp
@@ -179,7 +179,7 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             b = cast(b.type().with_code(halide_type_uint), b);
         } else if (b_info.max_defined && b_info.max <= 0) {
             result_op = Call::get_intrinsic_name(op->is_intrinsic(Call::shift_right) ? Call::shift_left : Call::shift_right);
-            b = cast(b.type().with_code(halide_type_uint), -b);
+            b = mutate(cast(b.type().with_code(halide_type_uint), -b), nullptr);
         }
 
         // If the shift is by a constant, it should now be unsigned.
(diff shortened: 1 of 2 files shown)


abadams:
I think this looks right but would appreciate @rootjalex also taking a look as he has been thinking about shifts recently.

src/Simplify_Call.cpp
```diff
@@ -170,53 +171,49 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
                     return mutate(Div::make(a, b), bounds);
                 }
             } else {
-                // For signed types, (1 << ub) will overflow into the sign bit while
-                // (-32768 >> ub) propagates the sign bit, making decomposition
+                // For signed types, (1 << (t.bits() - 1)) will overflow into the sign bit while
+                // (-32768 >> (t.bits() - 1)) propagates the sign bit, making decomposition
                 // into mul or div problematic, so just special-case them here.
-                if (shift_left) {
+                if (result_op == Call::get_intrinsic_name(Call::shift_left)) {
```
rootjalex, line 204: `op->is_intrinsic(Call::shift_left)` seems more clear to me, here and in the check above

rootjalex:
> I think this looks right but would appreciate @rootjalex also taking a look as he has been thinking about shifts recently.

changes seem reasonable to me

src/Simplify_Call.cpp
```diff
@@ -170,53 +171,49 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
+            b = mutate(cast(b.type().with_code(halide_type_uint), -b), nullptr);
+        }
 
+        // If the shift is by a constant, it should now be unsigned.
+        uint64_t ub = 0;
+        if (const_uint(b, &ub)) {
             // LLVM shl and shr instructions produce poison for
             // shifts >= typesize, so we will follow suit in our simplifier.
             if (ub >= (uint64_t)(t.bits())) {
                 return make_signed_integer_overflow(t);
```
rootjalex, line 191: Should this be moved down? It looks like an unsigned `a` could trigger a signed_integer_overflow error

rootjalex:
> I think this looks right but would appreciate @rootjalex also taking a look as he has been thinking about shifts recently.

my only comments are nit-picky, the changes themselves seem fine!

src/Simplify_Call.cpp
dsharletg: We can't do that though, because we might have swapped from left to right shifts above?

src/Simplify_Call.cpp
dsharletg: I'm not sure. I think maybe "signed_integer_overflow" might be being used as a more general thing here. I think `unsigned <</>> t.bits()` is still meant to produce a poison value. Maybe `signed_integer_overflow` should be renamed? @abadams

src/Simplify_Call.cpp
dsharletg: The thing that would be best is if we could just use Call::IntrinscOp instead of std::string, but that's a bit messy to do.

rootjalex: Ah sorry, my bad, makes sense!

src/Simplify_Call.cpp
abadams: I think we should possibly just define what it does to be the arm behavior, so that the naming issue is moot.

dsharletg: OK, but until then, the other parts of this PR are useful, so should we just leave this as-is?

abadams (approved):

PR #5492, opened 2020-11-30
Fix case where simplifying interleaves might need a slice of the original vector

dsharletg:


src/Simplify_Shuffle.cpp
```diff
@@ -154,8 +154,11 @@ Expr Simplify::visit(const Shuffle *op, ExprInfo *bounds) {
                         break;
                     }
 
+                    // If the shuffle is a single element, we don't care what the stride is,
+                    // so we can just assume it is what we are looking for.
+                    int i_stride = i_shuffle->indices.size() > 1 ? i_shuffle->slice_stride() : terms;
                     // ... and that it is a slice in the right place...
-                    if (i_shuffle->slice_begin() != (int)i || i_shuffle->slice_stride() != terms) {
+                    if (i_shuffle->slice_begin() != (int)i || i_stride != terms) {
```
abadams, line 161: I find this code pretty confusing. If it doesn't matter when the stride is in the single vector case, where is this handled below? I see a fixed stride of 1.

Also comparing terms to terms as a way of disabling the condition is confusing.

dsharletg: It's not the single vector case, it's when the interleave is of single element slices.

The better fix might be to swap the order of the is_interleave and is_concat cases (is_concat should return true for this case also)?

abadams (approved):

PR #5495, opened 2020-11-30
Simplify a slice of slice

vksnk:


src/Simplify_Shuffle.cpp
```diff
@@ -238,6 +238,20 @@ Expr Simplify::visit(const Shuffle *op, ExprInfo *bounds) {
         }
     }
 
+    // Try to collapse a slice of slice.
+    if (op->is_slice() && (new_vectors.size() == 1)) {
+        if (const Shuffle *inner_shuffle = new_vectors[0].as<Shuffle>()) {
+            if (inner_shuffle->is_slice() && (inner_shuffle->vectors.size() == 1)) {
```
dsharletg, line 244: Do you need the inner shuffles to be one vector? This might be useful for slices of concats too?

vksnk: Hmm, I was going to extend to any number of vectors, but it seems that:

a) make_slice only accepts a single vector.
b) is_slice doesn't actually check that vectors.size() == 1 and only checks that indices form ramp.

I can add an overload for make_slice which accepts multiple vectors, but wanted to check in case there are maybe assumptions somewhere that slice can only take a single vector (in which case, probably might be a good idea to add a check into is_slice).

vksnk: and seems that slice of concats is basically the same as a slice of multiple vectors?

dsharletg: I don't think we need a new helper, but it seems like writing something like `make_slice(make_slice(make_concat(...), ...), ...)` would hit this case.

dsharletg: As discussed offline, the simplifier doesn't do what I thought it did here :)

vksnk:
I've added a rule to simplify slice of concats, PTAL.

src/Simplify_Shuffle.cpp
```diff
@@ -245,6 +245,42 @@ Expr Simplify::visit(const Shuffle *op, ExprInfo *bounds) {
...
+            // Check if we really need to concat all vectors before slicing.
+            if (inner_shuffle->is_concat()) {
+                int slice_min = op->indices.front();
+                int slice_max = op->indices.back();
+                int concat_index = 0;
+                int new_slice_start = -1;
+                vector<Expr> new_concat_vectors;
+                for (const auto &v : inner_shuffle->vectors) {
+                    // Check if current concat vector overlaps with slice.
+                    if ((concat_index >= slice_min && concat_index <= slice_max) || ((concat_index + v.type().lanes() - 1) >= slice_min && (concat_index + v.type().lanes() - 1) <= slice_max)) {
```
dsharletg, line 268: I think adding a new line after the `||` would make this easier to read.

dsharletg (approved):

src/Simplify_Shuffle.cpp
vksnk: Done.

steven-johnson:
Buildbots should be mostly healthy again, so not sure which of these failures might be legit or not.

vksnk:
Failures are unrelated.

PR #5497, opened 2020-11-30
Combine align and slice for the small vectors in align_loads

vksnk:
This improves the case when a smaller vector load is fully contained inside of the aligned native vector load. In the current implementation it will be handled in two passes: firstly, a small vector will be extended to the native vector size (slice from the unaligned native vector load), followed by the second pass where it will be actually aligned, which leads to a) double slice and b) load of vector of double native width. This change combines these two passes when possible to avoid double slice and double load.

### DIFF
--- a/src/AlignLoads.cpp
+++ b/src/AlignLoads.cpp
@@ -115,8 +115,7 @@ private:
 
             // If load is smaller than a native vector and can fully fit inside of it and offset is known,
             // we can simply offset the native load and slice.
-            if (!is_aligned && aligned_offset != 0 && Int(32).can_represent(aligned_offset)
-                 && (aligned_offset + lanes <= native_lanes)) {
+            if (!is_aligned && aligned_offset != 0 && Int(32).can_represent(aligned_offset) && (aligned_offset + lanes <= native_lanes)) {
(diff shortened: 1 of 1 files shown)


dsharletg:
We should also make sure that slices of double vectors that only use one of the vectors get simplified to a slice of just one vector, even if we go ahead with this change.

vksnk:
You would need to know a native vector size in the simplifier to do that. There is a similar case when instead of loading double-width vector, it's a concatenation of two vectors and in that case we can certainly simplify. I'll do it in the other slice-of-slice PR, because it's somewhat related to that (we could also change code in align_loads to do concat of two loads instead of loading double width, but I don't know if it's worth doing).

vksnk:
I've added a rule (#5495) that can simplify a slice of the concat of two vectors into a slice of just one vector when possible, but still I think this PR is useful, because it avoids loading of double-width vector and loads only single-width vector. PTAL.

dsharletg (approved):
I think the whole function needs a refactoring somehow. I can't quite put my finger on it right now though so it looks OK for now.

vksnk:
Failures are unrelated, merging in.

PR #5498, opened 2020-12-01
Restructure interpret_nn

steven-johnson:
- Shuffle stuff into subdirs, remove some dead files, do some Makefile cleanup


dsharletg (approved):
I don't think tflite_parser.h/cpp should be in the interpreter subdir. Otherwise looks good.

steven-johnson:
Done.

PR #5500, opened 2020-12-01
Enable constant folding of broadcasted constants

dsharletg:
Currently, things like `shift_left(x, broadcast(2))` don't get simplified, and things like `shift_right(8, broadcast(1))` don't get constant folded. This PR fixes that.


steven-johnson (commented):
Worthy of adding something to an internal test?

dsharletg:
I added a few vector versions of some tests, but really, this gets a lot more useful coverage from fuzz_simplify.

steven-johnson (approved):

PR #5501, opened 2020-12-01
expr_match for Shuffles

vksnk:
This PR extends expr_match to also support Shuffle nodes.

### DIFF
--- a/src/IRMatch.cpp
+++ b/src/IRMatch.cpp
@@ -259,9 +259,7 @@ public:
 
     void visit(const Shuffle *op) override {
         const Shuffle *e = expr.as<Shuffle>();
-        if (result && e && types_match(op->type, e->type)
-            && op->vectors.size() == e->vectors.size()
-            && op->indices == e->indices) {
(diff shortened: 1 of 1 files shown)


abadams:
No objection, but I would actually like to deprecate expr_match. It's mostly just the same as the IRMatcher patterns, but much much slower.

vksnk:
That's good to know, I am using it similar to how it's used in HexagonOptimize.cpp, i.e. match some pattern to expression and if it matches, then replace it with the call which arguments are some subexpressions (with some optional preprocessing, for example, trying to narrow the arguments) of the original expression. I can't seem to find a direct replacement of the expr_match, which would let me to get matched subexpressions. Is there anything similar?

abadams:
There's a minor extension I could make that would make it a direct replacement. I'll see if I can do it this afternoon.

vksnk:
That's great, thank you!

steven-johnson:
Is this PR defunct?

vksnk:
I only need it in my local branch for some other work, so can keep it there for now. I'll close it.

PR #5502, opened 2020-12-01
Add slow_tests label to resize app

dsharletg:
Similar to #5482. This test builds and tests the cross product of 4 kernels, 3 data types, and upsample/downsample, which can take several minutes on the build bots.


dsharletg:
I'm going to close this, I don't think this app is disproportionally slow after the recent compile time fix.

PR #5505, opened 2020-12-01
Change O3 to O2 everywhere

dsharletg:
I've randomly checked a bunch of code for performance, both in Halide, and a large amount of C++ code recently (in a different project). Here's what I've found:
- In a general large C++ codebase, -O2 was often faster than -O3, and very rarely slower.
- In Halide code (i.e. this PR), everything I checked was noise, except one pretty key large pipeline got 4% faster with -O2.

At the same time:
- O2 compiles faster than O3, sometimes a lot faster (we had several comments noting this).
- It generates less code, sometimes a lot less.
- time make -j4 with -O2: 731.20s user 81.52s system 312% cpu 4:19.73 total
- time make -j4 with -O3: 784.67s user 86.48s system 299% cpu 4:50.80 total


zvookin:
I'd like separate PRs for the Halide compiler and the Makefiles.

apps/bgu/Makefile
```diff
@@ -24,7 +24,7 @@ $(BIN)/%/runtime.a: $(GENERATOR_BIN)/bgu.generator
 
 $(BIN)/%/filter: filter.cpp $(BIN)/%/bgu.a $(BIN)/%/bgu_auto_schedule.a $(BIN)/%/runtime.a
 	@mkdir -p $(@D)
-	$(CXX) $(CXXFLAGS) -I$(BIN)/$* -Wall -O3 $^ -o $@ $(LDFLAGS) $(IMAGE_IO_FLAGS) $(CUDA_LDFLAGS) $(OPENCL_LDFLAGS) $(OPENGL_LDFLAGS)
+	$(CXX) $(CXXFLAGS) -I$(BIN)/$* -Wall -O2 $^ -o $@ $(LDFLAGS) $(IMAGE_IO_FLAGS) $(CUDA_LDFLAGS) $(OPENCL_LDFLAGS) $(OPENGL_LDFLAGS)
```
steven-johnson, line 27: For all the apps/ Makefiles: why not just specify -O2 in the default CXXFLAGS in Makefile.inc and avoid respecifying in the specific Makefile?

CMakeLists.txt
```diff
@@ -47,6 +47,16 @@ set(CMAKE_CXX_STANDARD 14)
...
+# Change CMake default O3 to O2.
+string(REGEX REPLACE "([\\/\\-]O)3" "\\12"
+    CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+string(REGEX REPLACE "([\\/\\-]O)3" "\\12"
+    CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
+string(REGEX REPLACE "([\\/\\-]O)3" "\\12"
+    CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL}")
+string(REGEX REPLACE "([\\/\\-]O)3" "\\12"
+    CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
+
```
alexreinking, line 59: No. These variables are for users to set. This makes it _impossible_ for users to compile with `-O3`.

alexreinking (changes requested):

CMakeLists.txt
alexreinking: If you want to compile with different flags, set `-DCMAKE_CXX_FLAGS_RELEASE="-O2"` at the command line.

dsharletg: How do we change the default?

alexreinking: There's a new feature we can optionally support in CMake 3.19 that will allow us to ship recommended defaults for these flags. But they absolutely do not belong _in_ the CMakeLists.txt.

https://cmake.org/cmake/help/v3.19/manual/cmake-presets.7.html

alexreinking: I opened a PR (#5506) that uses `-O2` in the Release preset for GCC-compatible compilers. This regetx will never match for MSVC (there's no `/O3`) so I left that untouched.

steven-johnson:
So with the current setup, do we default to using `-O2` in CMake? Or does that only work if we use the presets (which requires CMake 3.19+)?

alexreinking:
> 
> 
> So with the current setup, do we default to using `-O2` in CMake? Or does that only work if we use the presets (which requires CMake 3.19+)?

That only happens if we use the presets or specifically set `CMAKE_CXX_FLAGS_RELEASE="-O2"` manually at the command line. We might want to do that in the buildbots. That variable should absolutely not be touched from inside the CMakeLists.txt though.

steven-johnson:
> > So with the current setup, do we default to using `-O2` in CMake? Or does that only work if we use the presets (which requires CMake 3.19+)?
> 
> That only happens if we use the presets or specifically set `CMAKE_CXX_FLAGS_RELEASE="-O2"` at the command line. We might want to do that in the buildbots. That variable should absolutely not be touched from inside the CMakeLists.txt though.

...ugh. I see your point, but that also means that every coder has to remember to specify that as well. Surely there's some way we can change what the default is.

abadams:
I tried compiling libHalide with -O2 and -O3 and benchmarking lowering via: time bin/host/lens_blur.generator -g lens_blur -e stmt -o bin/host -f lens_blur_auto_schedule target=host-no_runtime auto_schedule=true

When compiled with gcc 9.3, -O3 was 3% faster, and the side of libHalide.so increased by 0.2%.
When compiled with clang 10, -O3 was the same speed as -O2 (both of them faster than gcc), and the size of libHalide.so increased by 0.2%.

So I guess with clang O3 is indeed not doing much for us, but it's also not hurting, and O3 helps with gcc. So I think we should leave things at O3.

dsharletg:
Did you find any impact to build time? That was the main reason I thought this was worth doing (when the build bots were so bogged down). I expected impacts to code size/runtime to be negligible, the thing I thought this might do is save 10-20% build time. I think this would be worth doing because I think a few % impact to runtime of Halide itself is not a big deal (GCC's drop in O2 performance doesn't affect our generated code). But I suppose it's possible a few percent regression in runtime performance of Halide when generating code is still possibly going to cancel out some of the reduced build time of Halide itself...

abadams:
At -O3 "make distrib -j16" takes 1:53, and at -O2 it takes 1:48. Those are wall-clock time. In CPU time it's 680s vs 650s.

The tests are all built at -O0, so no change when compiling them.

dsharletg:
OK, between this and the CMake issue this doesn't seem worth it.

PR #5506, opened 2020-12-01
Add CMake presets for 3.19+ users

alexreinking:
CMake 3.19 adds a new feature called [_presets_](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) that make it easier to store, name, and use, common sets of configurations for a CMake project.

This PR adds the necessary file (and documentation!) populated with presets for the following scenarios:

| Name             | Generator          | Build type     | Target                              | Dependencies    |
| ---------------- | ------------------ | -------------- | ----------------------------------- | --------------- |
| `gcc-release`    | Ninja              | Release        | as determined by selected compiler  | system-provided |
| `gcc-debug`      | Ninja              | Debug          | as determined by selected compiler  | system-provided |
| `msvc-release`   | Ninja              | Release        | as determined by vcvars environment | vcpkg           |
| `msvc-debug`     | Ninja              | Debug          | as determined by vcvars environment | vcpkg           |
| `win32`          | Visual Studio 2019 | _multi-config_ | 32-bit Windows, 64-bit host tools   | vcpkg           |
| `win64`          | Visual Studio 2019 | _multi-config_ | 64-bit Windows, 64-bit host tools   | vcpkg           |

The Windows/MSVC presets assume that the environment variable `VCPKG_ROOT` is set to a base vcpkg installation path.

To use them, just run `cmake --preset=gcc-release` from the Halide root and it will create a `./build` directory for you.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,77 @@
+{
+  "version": 1,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 16,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
+      "name": "gcc-debug",
+      "displayName": "GCC (Debug)",
+      "description": "Debug build using Ninja generator and GCC-compatible compiler",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug"
+      }
+    },
+    {
+      "name": "gcc-release",
+      "displayName": "GCC (Release)",
+      "description": "Release build using Ninja generator and GCC-compatible compiler",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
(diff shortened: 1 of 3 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,77 @@
...
+    },
+    {
+      "name": "gcc-release",
+      "displayName": "GCC (Release)",
+      "description": "Release build using Ninja generator and GCC-compatible compiler",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_CXX_FLAGS_RELEASE": "-O2 -DNDEBUG"
```
steven-johnson, line 27: We take pains to ensure that NDEBUG is never defined (e.g. Util.h), so this should really be `-UNDEBUG`

alexreinking: This is to match the existing CMake behavior and to help out with #5505. Just removing `-DNDEBUG` should be enough (plus documenting it).

steven-johnson (approved):

PR #5507, opened 2020-12-02
Restructure apps to be fully external.

alexreinking:
As discussed offline, we're restructuring the apps to be fully external. This change requires changes to the buildbots.

We're ditching the CTest `--build-and-test` mode in favor of a joint build of all the apps (which remain standalone). Now, to build the apps, one would first configure/build/test/install Halide to a local directory and _then_ point the app(s) to that local install during its configure step. Then build/test as any other CMake project.

This looks like the following (eg. on Linux):

```
# Configure, build, install Halide to a local dir
$ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S . -B build
$ cmake --build build
$ (cd build ; ctest -j 20)
$ cmake --install build --prefix build/_Halide_install

# Configure, build, test all apps using the local Halide install
$ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/path/to/build/_Halide_install -S apps -B build/apps
$ cmake --build build/apps
$ (cd build/apps ; ctest -LE slow_tests)
```

We should give the tutorials the same treatment, tbqh.


alexreinking:
Leaving as a draft because I have no idea if this works yet lol

steven-johnson:
> We should give the tutorials the same treatment, tbqh.

SGTM

steven-johnson:
See https://github.com/halide/build_bot/pull/111 for the necessary buildbot changes. 

If you and @abadams have tested this locally and are happy with it, go ahead and land and we'll just tell people to ignore failures in apps/ until a proper fix lands. (Or, alternately, I could quick-push a hack to buildbot to just remove apps testing until the PR above lands.)

alexreinking:
I've tested locally and it works for me 🙂 . One of you should test it, too.

steven-johnson:
Testing this on my Linux box, Halide builds & installs properly, but the apps fail during the cmake command with:

```
CMake Error at /usr/local/google/home/srj/halide-install/lib/cmake/Halide/Halide-Targets-shared.cmake:37 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined:
  Halide::shared::Halide;Halide::shared::Generator;Halide::shared::RunGenMain;Halide::shared::Adams2019;Halide::shared::Li2018;Halide::shared::Mullapudi2016


  Targets not yet defined: Halide::shared::LLVM;Halide::shared::wabt-obj

  /usr/local/google/home/srj/halide-install/lib/cmake/Halide/HalideConfig.cmake:62 (include)
  /usr/local/google/home/srj/halide-install/lib/cmake/Halide/HalideConfig.cmake:100 (_Halide_include)
  bilateral_grid/CMakeLists.txt:12 (find_package)
_
```

steven-johnson:
...but weirdly, works fine on OSX

steven-johnson:
FYI:  my OSX box is running CMake 3.18.4, Linux is 3.16.3 -- what version(s) did you test with?

steven-johnson:
> FYI: my OSX box is running CMake 3.18.4, Linux is 3.16.3 -- what version(s) did you test with?

I built CMake 3.18.4 on my Linux box and that version behaves correctly, so there's something about our setup that is unhappy with 3.16.x -- no idea offhand what's going on. Since 3.16 is our declared minimum we'll need to figure this out before landing.

steven-johnson:
FYI, CMake 3.17.5 also fails

alexreinking:
I know the issue... it's with this code here: 

```
https://github.com/halide/Halide/blob/329a40525f2edb0248db00e01b4d8de9d282e9fd/packaging/HalideConfig.cmake#L69-L74
```

I need to prevent the package script from running multiple times.

steven-johnson (approved):

PR #5508, opened 2020-12-02
Adding CMake build for FFT

alexreinking:
Works, but missing two features:

1. FFTW integration.
2. Small stacks on macOS.

Based on #5507

### DIFF
--- a/apps/fft/CMakeLists.txt
+++ b/apps/fft/CMakeLists.txt
@@ -1,0 +1,56 @@
+cmake_minimum_required(VERSION 3.16)
+project(fft)
+
+enable_testing()
+
+# Set up language settings
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED YES)
+set(CMAKE_CXX_EXTENSIONS NO)
+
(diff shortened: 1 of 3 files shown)


apps/fft/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+find_package(Halide REQUIRED)
+
+# Generator
+add_executable(fft.generator fft_generator.cpp fft.cpp)
+target_link_libraries(fft.generator PRIVATE Halide::Generator)
+
+# Filters
+add_halide_library(fft_forward_r2c FROM fft.generator
+                   GENERATOR fft
+                   PARAMS direction=samples_to_frequency size0=16 size1=16 gain=0.00390625 input_number_type=real output_number_type=complex)
```
steven-johnson, line 21: style nit: might be nice to split the PARAMS into multiple lines rather than one loooong one, if easy to do

steven-johnson (approved):

steven-johnson:
If this builds properly for you locally, go ahead and land it, since the buildbots aren't currently building/testing apps (so 'testing' this PR would just be a waste)

PR #5509, opened 2020-12-02
Fix lossless cast for nested widening casts.

dsharletg:
`lossless_cast` is currently broken for double widening casts, such as `lossless_cast(uint8, i32(i16(u8)))`. I think this change also matches the behavior intended by the comment.


abadams:
Failure looks real. Looks like this makes it try for a pairwise add variant that doesn't exist on arm, so I'm guess something needs an additional condition in CodeGen_ARM

dsharletg:
I just noticed this fix is in #5459, but I think it has bugs like this one. I'm going to close this and comment on that.

Issue #5465, opened 2020-11-20
memoize improvements


dsharletg:
I have a use case for `memoize`, but it doesn't quite work for me as is, at least not without some hackery. I think there are a few changes we should consider making:

- `memoize` should accept a memoise cache key as an argument, so the memoize tag is specified in the schedule rather than the algorithm.
- I'd like more explicit control over the cache. Rather than have my memoized funcs get evicted from the cache, and need to control the cache size, I'd rather explicitly say I'm done with a particular tag. This might only make sense in the context of the first thing, where it would be more reasonable to be confident about the exact value of the cache key.

PR #5510, opened 2020-12-02
Add a feature to name cached memoizations and to evict them by name.

zvookin:
Per discussion with @dsharletg, this allows "naming" memoizations with 64-bits and then evicting entries by using that name. The name is not part of the cache key and is user chosen. A pointer value, a hash of image data, or a monotonic counter are all possibilities for cache eviction names.

I have a small design doc in progress that I will share shortly. It covers the possibility of using multiple typed names of this sort to allow caching CPU allocations as well. (The runtime automatically adds the GPU context as an eviction name and whenever a GPU context is freed, its bit value is used to evict all allocations on that context from the memoization cache.)

Wanted to get the review/buildbot process started here. 

Addresses the second item in issue https://github.com/halide/Halide/issues/5465 .


steven-johnson:
The build failures should go away if you sync to master.

src/Schedule.h
```diff
@@ -529,6 +529,13 @@ class FuncSchedule {
     bool memoized() const;
     // @}
 
+    /** This flag is set to true if the schedule is memoized and has an attached
```
dsharletg, line 532: Update (or remove) comment? It looks stale.

src/runtime/cache.cpp
```diff
@@ -562,6 +571,41 @@ WEAK void halide_memoization_cache_cleanup() {
...
+WEAK void halide_memoization_cache_evict(void *user_context, uint64_t eviction_key) {
+    ScopedMutexLock lock(&memoization_lock);
+
+    for (size_t i = 0; i < kHashTableSize; i++) {
+        CacheEntry *entry = cache_entries[i];
+        if (entry != nullptr) {
+            CacheEntry **prev = &cache_entries[i];
+            while (entry != nullptr) {
+                CacheEntry *next = entry->next;
+                if (entry->has_eviction_key && entry->eviction_key == eviction_key) {
```
dsharletg, line 583: Can this list removal be in a helper function? It looks like all the other potential uses are removing from the end and so are a bit simpler, but I think maybe it would be better if they all used a more general list removal helper anyways?

steven-johnson:
Not sure how to resolve the merge conflict here.

steven-johnson (approved):

steven-johnson:
Looks good, ready to land

PR #5517, opened 2020-12-03
Add possible simplify to GL(Compute) `pow` function

xndcn:
OpenGL(Compute) generates `select` IR for `pow(a, b)` function, which can be simplified when `a` or `b` is const.


steven-johnson:
Opened https://github.com/halide/Halide/pull/5518 for testing

steven-johnson (approved):
Test (https://github.com/halide/Halide/pull/5518) finally looks good, LGTM

PR #5519, opened 2020-12-03
Add simd_op_check tests and a few more patterns

dsharletg:
We care about count_leading_zeros, and this PR fills in some other gaps in coverage. Also add a few patterns, I think we should have non-rounding versions of instructions we have rounding patterns for.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -149,6 +149,18 @@ CodeGen_ARM::CodeGen_ARM(Target target)
         }
     }
 
+    casts.emplace_back("vqdmulh.v4i16", "sqdmulh.v4i16", 4,
+                       i16_sat((wild_i32x4 * wild_i32x4) / (1 << 15)),
+                       Pattern::NarrowArgs);
+    casts.emplace_back("vqdmulh.v8i16", "sqdmulh.v8i16", 8,
+                       i16_sat((wild_i32x_ * wild_i32x_) / (1 << 15)),
+                       Pattern::NarrowArgs);
+    casts.emplace_back("vqdmulh.v2i32", "sqdmulh.v2i32", 2,
+                       i32_sat((wild_i64x2 * wild_i64x2) / Expr(int64_t(1) << 31)),
+                       Pattern::NarrowArgs);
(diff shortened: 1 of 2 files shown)


test/correctness/simd_op_check.cpp
```diff
@@ -1909,6 +1937,10 @@ int main(int argc, char **argv) {
         test.set_num_threads(1);
     }
 
+    if (getenv("FILTER")) {
```
steven-johnson, line 1940: nit: is this really an improvement over the optional command-line arg we already have?

src/CodeGen_ARM.cpp
```diff
@@ -149,6 +149,18 @@ CodeGen_ARM::CodeGen_ARM(Target target)
         }
     }
 
+    casts.emplace_back("vqdmulh.v4i16", "sqdmulh.v4i16", 4,
```
steven-johnson, line 152: So this PR is really "add some cast cases we didn't handle before, plus testing for them in simd_op_check"? Probably worth updating the commit/PR title then, I had assumed the change was just to the test, not the codegen.

steven-johnson (approved):
LGTM with nits.

test/correctness/simd_op_check.cpp
dsharletg: Err, I forgot to remove this. But yes, it is, I add it locally every time I'm doing something with simd_op_check. You can't pass a command line arg to the makefile. It's a few extra steps to work around, and running simd_op_check drops a lot of files wherever you run it from.

abadams: Perhaps simd_op_check shouldn't do that. I've had it bomb my working directory in that way too, and it's super annoying. It should just make a tmp subdir (and print what it is so you can go inspect the files if you really want to)

steven-johnson: +100, output to /tmp dir always would be much better.

re: "FILTER", can we at least make it something slightly less generically named?

steven-johnson:
simd_op_check is failing on arm32: https://buildbot.halide-lang.org/master/#/builders/38/builds/10

dsharletg:
This PR only affects ARM, failures must be unrelated.

PR #5527, opened 2020-12-04
Add overloaded intrinsic mechanism to simplify code generation

dsharletg:
This PR is part 1 of 2 of a significant refactoring of the backend codegen.

This PR adds an "overloaded intrinsic" mechanism to CodeGen_LLVM. The main thing it does is separate types from pattern matching. Currently, patterns are typed, including the number of lanes. This PR changes this, so patterns are independent of the number of lanes (or even the type, but this is currently unused), and generate calls to overloaded intrinsics. Then, the overload resolution handles calling the intrinsic of the appropriate type.

This enables pattern matching to be simpler, and possibly lifted out of the backends into a target independent lowering pass (this is part 2).

There is a lot more than could be done with this mechanism to lift more things out of the runtime .ll files, but leaving this as a separate follow-up as it should be mostly mechanical and easier to review by itself.


src/CodeGen_ARM.cpp
```diff
@@ -462,13 +657,30 @@ void CodeGen_ARM::visit(const Mul *op) {
 }
 
 void CodeGen_ARM::visit(const Div *op) {
+#if 0
+    // TODO: This is probably the easiest thing to convert to IRMatch rewriting, but it
+    // needs work.
```
steven-johnson, line 662: Not a blocker, but what sort of work does it need?

src/CodeGen_X86.cpp
```diff
@@ -250,110 +361,36 @@ void CodeGen_X86::visit(const Cast *op) {
...
-         i8_sat(wild_i16x_)},
-        {Target::AVX2, false, UInt(8, 32), 17, "packuswbx32",
-         u8_sat(wild_i16x_)},
-        {Target::FeatureEnd, false, UInt(8, 16), 0, "packuswbx16",
-         u8_sat(wild_i16x_)},
-        {Target::AVX2, false, UInt(16, 16), 9, "packusdwx16",
-         u16_sat(wild_i32x_)},
-        {Target::SSE41, false, UInt(16, 8), 0, "packusdwx8",
-         u16_sat(wild_i32x_)}};
+        {true, "saturating_add", i8_sat(wild_i16x_ + wild_i16x_)},
```
steven-johnson, line 370: Might want //clang-format-off here to prevent future unhappiness?

test/correctness/simd_op_check.cpp
```diff
@@ -567,12 +567,12 @@ class SimdOpCheck : public SimdOpCheckTest {
             check(arm32 ? "vabd.u32" : "uabd*v*s", 2 * w, u32(abs(i64(u32_2) - u32_3)));
 
             // VABDL    I       -       Absolute Difference Long
-            check(arm32 ? "vabdl.s8" : "sabdl*v*b", 8 * w, i16(absd(i8_2, i8_3)));
-            check(arm32 ? "vabdl.u8" : "uabdl*v*b", 8 * w, u16(absd(u8_2, u8_3)));
-            check(arm32 ? "vabdl.s16" : "sabdl*v*h", 4 * w, i32(absd(i16_2, i16_3)));
-            check(arm32 ? "vabdl.u16" : "uabdl*v*h", 4 * w, u32(absd(u16_2, u16_3)));
-            check(arm32 ? "vabdl.s32" : "sabdl*v*s", 2 * w, i64(absd(i32_2, i32_3)));
-            check(arm32 ? "vabdl.u32" : "uabdl*v*s", 2 * w, u64(absd(u32_2, u32_3)));
+            //check(arm32 ? "vabdl.s8" : "sabdl*v*b", 8 * w, i16(absd(i8_2, i8_3)));
```
steven-johnson, line 570: Why are these commented out?

test/correctness/simd_op_check.cpp
```diff
@@ -1937,6 +1937,10 @@ int main(int argc, char **argv) {
         test.set_num_threads(1);
     }
 
+    if (getenv("FILTER")) {
```
steven-johnson, line 1940: IMHO a less-generic env var name would be preferable here.

steven-johnson (commented):
This is pretty cool! Just some nits. Will wait till it goes green for approval.

src/CodeGen_ARM.cpp
dsharletg: Removed, this was to show @abadams WIP future possibilities

test/correctness/simd_op_check.cpp
dsharletg: Never meant to commit this, reverting

dsharletg:
It's hard to exhaustively check this, but I've tested this for correctness and performance on quite a bit of code now. There are some minor rearrangements of assembly in pmaddwd cases, but other than that, I haven't found any regressions.

src/CodeGen_LLVM.h
```diff
@@ -465,6 +471,26 @@ class CodeGen_LLVM : public IRVisitor {
      * an arbitrary number of vectors.*/
     virtual llvm::Value *interleave_vectors(const std::vector<llvm::Value *> &);
 
+    /** Description of an intrinsic function overload. Overloads are resolved
```
zvookin, line 474: The data structure one likely wants here is an unordered_map indexed by both the name and the types. The key could be a string perhaps even a fixed length type given that we can likely put an upper bound on the number of arguments to an overloaded intrinsic. The goal here is to make sure compile time efficiency is preserved.

dsharletg: I want to consider allowing type promotion though, which means the key lookup isn't just a straightforward check for equality. There's already a bit of this going on which would make this tricky even now.

Also, these vectors are pretty small, I wouldn't be surprised if the benefit of them being in a vector outweighs lookup efficiency.

steven-johnson:
Mac failure in simd_op_check:

Internal Error at /Users/halidenightly/build_bot_new/worker/x86-64-osx-testbranch-11-make/halide-source/src/JITModule.cpp:201 Condition failed: result != 0: HalideJITMemoryManager: unable to find address for _abs_i8x32

src/CodeGen_ARM.cpp
```diff
@@ -149,133 +110,365 @@ CodeGen_ARM::CodeGen_ARM(Target target)
...
+            continue;
+        }
+        std::string full_name = intrin_name;
+        if (!starts_with(full_name, "llvm.")) {
+            full_name = prefix + full_name;
+        }
+
+        Type ret_type = i.ret_type;
+        std::vector<Type> arg_types;
+        arg_types.reserve(4);
```
abadams, line 463: I'd be more comfortable with: 4 -> sizeof(i.arg_types) / sizeof(i.arg_types[0])

Or some other way to synchronize this 4 with the 4 in the struct definition.

src/CodeGen_ARM.cpp
```diff
@@ -328,15 +521,18 @@ void CodeGen_ARM::visit(const Cast *op) {
...
-                    // The arm64 llvm backend wants i32 constants for right shifts.
-                    if (target.bits == 64 && pattern.type == Pattern::RightShift) {
-                        shift = ConstantInt::get(i32_t, shift_amount);
+                    // TODO: It would be nice if call_overloaded_intrin could handle this type promotion.
+                    Expr b;
+                    if (matches[1].type().is_scalar()) {
+                        if (target.bits == 32) {
+                            b = make_const(matches[0].type().with_code(halide_type_int), shift_amount);
+                        } else {
+                            b = make_const(Int(32), shift_amount);
```
abadams, line 530: Would be good to preserve the comment that the arm64 backend wants the type to always be int32

src/CodeGen_LLVM.cpp
```diff
@@ -4722,6 +4706,92 @@ Value *CodeGen_LLVM::get_user_context() const {
...
+            continue;
+        }
+
+        if (i.result_type.with_lanes(1) != result_type.with_lanes(1)) {
+            continue;
+        }
+
+        bool match = true;
+        for (int j = 0; j < (int)i.arg_types.size(); j++) {
+            if (i.arg_types[j].with_lanes(1) != args[j].type().with_lanes(1)) {
```
abadams, line 4744: We have Type::element_of for with_lanes(1). It only saves a single character, so if you think "with_lanes(1)" is clearer perhaps we should just globally replace element_of() with with_lanes(1), but that should be a separate PR.

src/CodeGen_LLVM.cpp
```diff
@@ -4722,6 +4706,92 @@ Value *CodeGen_LLVM::get_user_context() const {
...
+                match = false;
+                break;
+            }
+
+            if (args[j].type().lanes() == 1) {
```
abadams, line 4749: args[j].type().is_scalar()

src/CodeGen_LLVM.cpp
```diff
@@ -4722,6 +4706,92 @@ Value *CodeGen_LLVM::get_user_context() const {
...
+                // We can broadcast the argument.
+                // TODO: Should we prioritize overloads that don't need this?
+            } else if (i.arg_types[j].lanes() == 1) {
+                if (args[j].type().lanes() != 1) {
```
abadams, line 4753: args[j].type().is_vector(), and is_scalar() on the line above

test/correctness/simd_op_check.cpp
abadams: Except you didn't. At least call is SIMD_OP_CHECK_FILTER or something. FILTER is generic enough to already be in someone's environment

abadams (approved):
Just a few nits.

dsharletg:
Failures are unrelated except for one really strange webassembly issue I discussed with Steven.

PR #5530, opened 2020-12-07
Upgrade logging code in interpret_nn

steven-johnson:
To make some integration with the Delegate code easier, I upgraded our logging code to mimic their usage pattern a bit. This is a little gratuitous but will likely pay off in the Delegate code.

- Replaced LOG_FATAL with CHECK(0) everywhere
- Added LOG(INFO/WARN/ERROR), all of which are wrappers that output to stderr (and also to __android_log on android)
- Made the guts of CHECK() a little more purpose-built
- Moved guts of Logger into a cpp file to hopefully reduce code bloat from logging
- Made inclusion of __FILE__ and __LINE__ info contingent on NDEBUG, to reduce internal info of logging that might be desirable to leave in release builds.
- Converted some (but not all) usage of std::cerr to use LOG(whatever).

### DIFF
--- a/apps/interpret_nn/util/error_util.h
+++ b/apps/interpret_nn/util/error_util.h
@@ -1,10 +1,7 @@
 #ifndef ERROR_UTIL_H_
 #define ERROR_UTIL_H_
 
-#include <iostream>
-#include <memory>
 #include <sstream>
-#include <string>
 
 #include "HalideRuntime.h"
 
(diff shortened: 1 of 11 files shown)


apps/interpret_nn/util/error_util.h
```diff
@@ -40,43 +37,53 @@ inline std::ostream &operator<<(std::ostream &s, const std::vector<T> &v) {
...
+    Logger &operator=(Logger &&) = delete;
+};
+
+struct CheckLogger : public Logger {
+    CheckLogger(LogSeverity severity, const char *condition_string);
+    CheckLogger(LogSeverity severity, const char *file, int line, const char *condition_string);
+
+    [[noreturn]] ~CheckLogger() noexcept(false);
+
+    CheckLogger() = delete;
```
dsharletg, line 82: Don't think you need to do this if the base class version is deleted? Not sure though.

dsharletg (approved):

PR #5531, opened 2020-12-08
Pattern match intrinsics in a target independent lowering pass

### LINKED ISSUE #5632: Application compile failure on ppc64le with some (but not all) 'vectorize' directives
Building on a PowerPC (ppc64le) architecture with LLVM 11.0.0 and Halide master, our application [1] builds a generator but then fails when using it to compile the library with our schedule, which previously worked with Halide v10.0.0.  I could not reproduce this on x86.  Specifically, a generator fails unless the `vectorize` directives at lines 276 and 279 are removed [2].  However, the compilation failure is not actually indicative of what/where the problem is.  Instead it fails somewhere in LLVM:

```
$ make backprojection VERBOSE=1
<snipped>
[ 66%] Generating backprojection.h, backprojection.o
./backprojection.generator -n backprojection -d 0 -g backprojection -f backprojection -e c_header,object -o . target=host-c_plus_plus_name_mangling-large_buffers-no_runtime
Scheduling for CPU: target(powerpc-64-linux-c_plus_plus_name_mangling-large_buffers-no_runtime-power_arch_2_07-vsx)
Block size: 64
Vector size: 16
backprojection.generator: /ccs/home/cimes/casper/llvm-project-11/llvm/lib/IR/Instructions.cpp:453: void llvm::CallInst::init(llvm::FunctionType*, llvm::Value*, llvm::ArrayRef<llvm::Value*>, llvm::ArrayRef<llvm::OperandBundleDefT<llvm::Value*> >, const llvm::Twine&): Assertion `(i >= FTy->getNumParams() || FTy->getParamType(i) == Args[i]->getType()) && "Calling a function with a bad signature!"' failed.
make[3]: *** [backprojection.h] Aborted (core dumped)
make[3]: Leaving directory `/autofs/nccs-svm1_home1/cimes/casper/halide-sar-app/build-master'
```

We originally suspected the problem might be with the name mangling or large buffer features, but removing them did not solve the problem, whereas removing just the two aforementioned `vectorize` directives did, even though other `vectorize` directives remain in the schedule.  Halide should probably print out more detailed information when LLVM fails.

I acknowledge that this is a non-trivial pipeline and perhaps difficult to reproduce.  I will work on creating a simpler example if possible, but I wanted to get the problem out there now in case somebody already has ideas about what might be causing this, which also might help create a narrower test case.

Thanks.

[1] https://github.com/ISI-apex/halide-sar-app
[2] https://github.com/ISI-apex/halide-sar-app/blob/f8f6a945906cc782e9f9c98c600ac64755e8b251/backprojection.cpp#L276-L279

dsharletg:
This PR is part 2 of 2 (after #5527) of a major backend refactoring of how we generating calls to intrinsics.

This PR does a few things:
- Moves pattern matching into a target independent lowering pass to produce new intrinsics. This simplifies pattern matching for the backends.
- Because pattern matching is now a separate pass, we can add a new test/correctness/intrinsics.cpp. This has made debugging pattern matching/simd_op_check failures a lot easier (IMO).
- With the new intrinsic matching, we now support rounding shifts with dynamic shift amounts, and they are implemented on ARM. These make a big difference in key use cases. We now support basically all of the shift instructions on ARM, with a few unusual exceptions (e.g. SLI/SRI, though we could support these too I think).
- In general intrinsic optimizations are less logic and more patterns/tables after this PR. However, there are still some bits of logic that are hard to avoid, but some could be further converted to pattern matching.

Fixes #5632


src/IROperator.h
```diff
@@ -327,6 +327,41 @@ Expr requirement_failed_error(Expr condition, const std::vector<Expr> &args);
...
+/** Compute widen(a) * widen(b). a and b may have different signedness. */
+Expr widening_mul(Expr a, Expr b);
+/** Compute widen(a) - widen(b). The result is always signed. */
+Expr widening_sub(Expr a, Expr b);
+/** Compute widen(a) << b. */
+Expr widening_shift_left(Expr a, Expr b);
+/** Compute widen(a) >> b. */
+Expr widening_shift_right(Expr a, Expr b);
+
+/** Compute (a + (1 >> min(b, 0)) / 2) << b */
```
abadams, line 346: Is the comment right? That expression is quite confusing.

dsharletg: I think it is! Note this is a rounding shift *left*, i.e. it's equivalent to `rounding_shift_right(a, -b)`.

If you believe rounding_shift_right is correct, then you can get from that one to this by relatively straightforward simplifications:

`rounding_right_shift(a, b)` = `(a + (1 << max(-b, 0)) / 2) >> -b` = `(a + (1 << -min(b, 0)) / 2) << b` = `(a + (1 >> min(b, 0)) / 2) << b` = `rounding_left_shift(a, b)`

src/IROperator.h
```diff
@@ -327,6 +327,41 @@ Expr requirement_failed_error(Expr condition, const std::vector<Expr> &args);
...
+/** Compute widen(a) * widen(b). a and b may have different signedness. */
+Expr widening_mul(Expr a, Expr b);
+/** Compute widen(a) - widen(b). The result is always signed. */
+Expr widening_sub(Expr a, Expr b);
+/** Compute widen(a) << b. */
+Expr widening_shift_left(Expr a, Expr b);
+/** Compute widen(a) >> b. */
+Expr widening_shift_right(Expr a, Expr b);
+
+/** Compute saturating_add(a, (1 >> min(b, 0)) / 2) << b */
```
abadams, line 346: I think it would be worth noting that the "rounding" part only does something if b is negative, in which case this is equal to rounding_shift_left(a, -b). If you're thinking of positive b the idea of this rounding makes no sense. `1 >> something` is not something you'd see in C.

src/CodeGen_Internal.cpp
```diff
@@ -380,7 +380,7 @@ Expr lower_int_uint_mod(const Expr &a, const Expr &b) {
 
     int bits;
     if (is_const_power_of_two_integer(b, &bits)) {
-        return a & (b - 1);
+        return a & simplify(b - 1);
```
steven-johnson, line 383: If we're gonna call simplify() here, why not do the entire expression?

src/CodeGen_Internal.cpp
```diff
@@ -585,43 +585,41 @@ Expr lower_signed_shift_left(const Expr &a, const Expr &b) {
     internal_assert(b.type().is_int());
     const int64_t *const_int_b = as_const_int(b);
     if (const_int_b) {
-        Type t = UInt(a.type().bits(), a.type().lanes());
         Expr val;
         const uint64_t b_unsigned = std::abs(*const_int_b);
         if (*const_int_b >= 0) {
-            val = a << make_const(t, b_unsigned);
+            val = a << make_const(UInt(a.type().bits()), b_unsigned);
```
steven-johnson, line 591: Would it help readability to add `.as_uint()` for these cases? (We have `with_code()` but it's not really any terser)

src/CodeGen_LLVM.cpp
```diff
@@ -1715,6 +1717,7 @@ void CodeGen_LLVM::visit(const Div *op) {
         Value *b = codegen(op->b);
         value = builder->CreateFDiv(a, b);
     } else {
+        // Should we assert here? Probably should never get here.
```
steven-johnson, line 1720: Add an assert and find out...

src/CodeGen_WebAssembly.cpp
```diff
@@ -26,94 +26,67 @@ CodeGen_WebAssembly::CodeGen_WebAssembly(Target t)
...
+};
+
+// clang-format off
+const WasmIntrinsic intrinsic_defs[] = {
+    {"llvm.sadd.sat.v8i16", Int(16, 8), "saturating_add", {Int(16, 8), Int(16, 8)}, Target::WasmSimd128},
+    {"llvm.uadd.sat.v8i16", UInt(16, 8), "saturating_add", {UInt(16, 8), UInt(16, 8)}, Target::WasmSimd128},
+    {"llvm.sadd.sat.v16i8", Int(8, 16), "saturating_add", {Int(8, 16), Int(8, 16)}, Target::WasmSimd128},
+    {"llvm.uadd.sat.v16i8", UInt(8, 16), "saturating_add", {UInt(8, 16), UInt(8, 16)}, Target::WasmSimd128},
+
+    // TODO: Are these really different than the standard llvm.*sub.sat.*?
```
steven-johnson, line 48: When I first wrote this code, they were necessary in order to get the right instructions emitted. Maybe trunk LLVM has improved since then.

src/CodeGen_X86.cpp
```diff
@@ -167,25 +168,34 @@ bool should_use_pmaddwd(const Expr &a, const Expr &b, vector<Expr> &result) {
     Type t = a.type();
     internal_assert(b.type() == t);
 
-    const Mul *ma = a.as<Mul>();
-    const Mul *mb = b.as<Mul>();
-
-    if (!(ma && mb && t.is_int() && t.bits() == 32 && (t.lanes() >= 4))) {
+    if (!(t.is_int() && t.bits() == 32 && (t.lanes() >= 4))) {
```
steven-johnson, line 171: nit: The extra parens around `t.lanes() >= 4` is unnecessary and hampers readability

src/CodeGen_X86.cpp
```diff
@@ -220,33 +230,6 @@ void CodeGen_X86::visit(const Sub *op) {
     }
 }
 
-void CodeGen_X86::visit(const Mul *op) {
-
-#if LLVM_VERSION < 110
```
steven-johnson, line 225: Do we know that this is no longer necessary? (We haven't yet dropped support for LLVM10)

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,775 @@
...
+            return IRMutator::visit(op);
+        }
+
+        Expr value = mutate(op->value);
+
+        // This mutator can generate redundant casts. We can't use the simplifier because it
+        // undoes some of the intrinsic lowering here, and it causes some problems due to
+        // factoring (instead of distributing) constants.
+        if (const Cast *cast = value.as<Cast>()) {
+            if (cast->type.can_represent(cast->value.type()) || cast->type.can_represent(op->type)) {
```
steven-johnson, line 432: Do we really want `can_represent()` here instead of equality? Are there cases where matching to a *specific* type is important?

src/HexagonOptimize.cpp
```diff
@@ -546,7 +512,12 @@ class OptimizePatterns : public IRMutator {
             std::stable_sort(mpys.begin(), mpys.end(), LoadCompare());
         }
     }
-    Expr visit(const Add *op) override {
+
+    // We need to be able to
```
steven-johnson, line 516: We need to be able to... what?

src/IRMatch.h
```diff
@@ -1339,6 +1339,10 @@ constexpr bool and_reduce(bool first, Args... rest) {
     return first && and_reduce(rest...);
 }
 
+constexpr int const_min(int a, int b) {
```
steven-johnson, line 1342: // TODO: this can be replaced with std::min() once we require C++14 or later

src/IROperator.cpp
```diff
@@ -400,7 +400,7 @@ Expr lossless_cast(Type t, Expr e) {
     }
 
     if (const Cast *c = e.as<Cast>()) {
-        if (t.can_represent(c->value.type())) {
+        if (c->type.can_represent(c->value.type())) {
```
steven-johnson, line 403: I don't think I understand this change.

src/IROperator.h
steven-johnson: > 1 >> something is not something you'd see in C

Well, it's not something you *should* see in C...

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -677,6 +679,11 @@ int main(int argc, char **argv) {
         test_hvx.filter = argv[1];
         test_hvx.set_num_threads(1);
     }
+
+    if (getenv("FILTER")) {
```
steven-johnson, line 683: ...name still needs something less generic

steven-johnson (commented):
Just adding some comments for now, since ~all the buildbots are failing. Will revisit more once they start getting green.

src/CodeGen_Internal.cpp
dsharletg: Excessive simplification is bad, both for performance, and because it replaces intrinsics with "simpler" arithmetic sometimes. In this case, simplify is guaranteed to just be constant folding (b is constant, b - 1 should constant fold).

src/IROperator.cpp
dsharletg: The comment says it wants to allow recursion through widening casts (= casts that don't lose information). The old code didn't actually implement that, the new code does. This is also in a more targeted change #5459 that will hopefully be merged first.

src/CodeGen_X86.cpp
dsharletg: I didn't delete it, it's moved to the Call visitor.

src/FindIntrinsics.cpp
dsharletg: This covers cases like `int32(int16(uint8))` -> `int32(uint8)` or `uint8(uint16(uint32))` -> `uint8(uint32)` It also covers more simple redundant cases like `int32(int32(x))`.

test/correctness/simd_op_check.cpp
```diff
@@ -232,7 +231,7 @@ class SimdOpCheck : public SimdOpCheckTest {
 
         if (use_ssse3) {
             for (int w = 2; w <= 4; w++) {
-                check("pmulhrsw", 4 * w, i16((((i32(i16_1) * i32(i16_2)) + 16384)) / 32768));
+                check("pmulhrsw", 4 * w, i16((i32(i16_1) * i32(i16_2) + 16384) >> 15));
```
steven-johnson, line 234: ...I am genuinely curious as to how you noticed this; do you have a non-sse3 machine somewhere?

dsharletg: I think before this PR (or my last one), we were incorrectly generating pmulhrsw on non-sse3 targets, and there was a simd_op_check for that. When I fixed that (by accident due to refactoring/cleanup), simd_op_check started failing because it was expecting pmulhrsw when it shouldn't be generated.

src/CodeGen_Internal.cpp
dsharletg: Note this is also dropping the number of lanes too.

dsharletg:
I think this is ready for review. It passed has passed all the buildbots except for one test on windows, and that failure is weird: the same test passed on linux, and the test shouldn't be dependent on OS (CUDA 8-bit dot products).

src/CodeGen_LLVM.cpp
```diff
@@ -2956,7 +2944,26 @@ void CodeGen_LLVM::visit(const Call *op) {
                 }
             }
         }
-
+    } else if (op->is_intrinsic(Call::saturating_add) || op->is_intrinsic(Call::saturating_sub)) {
```
abadams, line 2947: This looks like code that will break when we add, e.g. 48-bit integers.

src/CodeGen_LLVM.cpp
```diff
@@ -2956,7 +2944,26 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+            internal_assert(op->type.is_uint());
+            intrin = "llvm.u";
+        }
+        if (op->is_intrinsic(Call::saturating_add)) {
+            intrin += "add.sat.";
+        } else {
+            internal_assert(op->is_intrinsic(Call::saturating_sub));
+            intrin += "sub.sat.";
+        }
+        if (op->type.lanes() > 1) {
```
abadams, line 2962: I don't understand what happens with non-native-width vectors here.

src/CodeGen_LLVM.cpp
```diff
@@ -2956,7 +2944,26 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+            intrin += "v" + std::to_string(op->type.lanes());
+        }
+        intrin += "i" + std::to_string(op->type.bits());
+        value = call_intrin(op->type, op->type.lanes(), intrin, op->args);
```
abadams, line 2966: The lanes arg is supposed to be the native lanes of the intrinsic, so vector width legalization must have already happened somewhere?

src/CodeGen_LLVM.cpp
```diff
@@ -4706,58 +4717,102 @@ Value *CodeGen_LLVM::get_user_context() const {
...
     intrinsics[name].emplace_back(ret_type, std::move(arg_types), intrin);
 }
 
+void CodeGen_LLVM::declare_intrin_overload(const std::string &name, const Type &ret_type, llvm::Function *impl, std::vector<Type> arg_types) {
+    internal_assert(impl);
+    intrinsics[name].emplace_back(ret_type, std::move(arg_types), impl);
+}
+
 Value *CodeGen_LLVM::call_overloaded_intrin(const Type &result_type, const std::string &name, const std::vector<Expr> &args) {
+    constexpr int debug_level = 2;
```
abadams, line 4763: 2 is too verbose for this. The rest of codegen is at 4

src/CodeGen_LLVM.cpp
```diff
@@ -4767,27 +4822,45 @@ Value *CodeGen_LLVM::call_overloaded_intrin(const Type &result_type, const std::
...
+        std::vector<Expr> promoted_args;
+        promoted_args.reserve(args.size());
+        for (size_t i = 0; i < args.size(); i++) {
+            Expr promoted_arg = args[i];
+            if (args[i].type().is_scalar()) {
+                promoted_arg = lossless_cast(resolved->arg_types[i].element_of(), promoted_arg);
+            }
+            if (resolved->arg_types[i].is_vector() && args[i].type().is_scalar() && result_type.lanes() > 1) {
+                // We're passing a scalar to a vector argument, broadcast it.
+                promoted_args.emplace_back(Broadcast::make(promoted_arg, result_type.lanes()));
```
abadams, line 4855: Is there any point in using emplace_back for an rvalue of the vector element type? Doesn't push_back have an overload that takes an rvalue reference and std::moves it into the vector? I'm only been using emplace_back when I want to pass constructor args for the object I'm pushing, and now I'm wondering if I've been doing it wrong.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -385,157 +388,135 @@ void CodeGen_PTX_Dev::visit(const Atomic *op) {
...
-            a = lossless_cast(Int(8, input_lanes), a);
-            b = lossless_cast(Int(8, input_lanes), b);
-            if (!a.defined()) {
-                // try uint
-                a = lossless_cast(UInt(8, input_lanes), mul->a);
-            }
-            if (!b.defined()) {
-                b = lossless_cast(UInt(8, input_lanes), mul->b);
-            }
+        if (!expr_match(p.pattern, op->value, matches)) {
```
abadams, line 422: What stopped you from using IRMatch patterns here?

src/CodeGen_X86.cpp
```diff
@@ -360,61 +345,43 @@ void CodeGen_X86::visit(const Cast *op) {
...
+            if (value) {
+                return;
             }
         }
     }
 
+    if (const Call *mul = Call::as_intrinsic(op->value, {Call::widening_mul})) {
+        if (op->value.type().bits() < op->type.bits() && op->type.bits() <= 32) {
+            // LLVM/x86 really doesn't like 8 -> 16 bit multiplication.
+            value = codegen(simplify(Mul::make(Cast::make(op->type, mul->args[0]), Cast::make(op->type, mul->args[1]))));
```
abadams, line 380: Why the call to simplify?

src/CodeGen_X86.cpp
```diff
@@ -360,61 +345,43 @@ void CodeGen_X86::visit(const Cast *op) {
...
+            value = call_overloaded_intrin(op->type, pattern.intrin, matches);
+            if (value) {
+                return;
             }
         }
     }
 
+    if (const Call *mul = Call::as_intrinsic(op->value, {Call::widening_mul})) {
+        if (op->value.type().bits() < op->type.bits() && op->type.bits() <= 32) {
+            // LLVM/x86 really doesn't like 8 -> 16 bit multiplication.
```
abadams, line 379: This comment could use some more elaboration. The condition looks like it also applies to 16 -> 32 bit widening muls?

src/CodeGen_LLVM.cpp
dsharletg: LLVM's saturating add/sub takes any number of lanes, it must do the legalization internally, just like a regular add/sub/etc. I think we are supposed to treat these like any other arithmetic operator in LLVM.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+bool no_overflow_int(Type t) {
+    return t.is_int() && t.bits() >= 32;
+}
+
+// Returns true iff t does not have a well defined overflow behavior.
+bool no_overflow(Type t) {
+    return t.is_float() || no_overflow_int(t);
+}
+
+// If there's a widening add or subtract in the first e.type().bits() / 2 - 1
```
abadams, line 59: A longer comment or an example would be nice. This comment is confusing.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+    return false;
+}
+
+bool is_safe_for_add(const Expr &e) {
+    return is_safe_for_add(e, e.type().bits() / 2 - 1);
+}
+
+// We want to find and remove an add of 'round' from e.
+Expr find_and_subtract(const Expr &e, const Expr &round) {
+    if (const Add *add = e.as<Add>()) {
```
abadams, line 88: I think it would be more similar to related code if this returned Expr() on failure instead of checking same_as. I worry a branch will get added in future that happens to make a new Expr but doesn't actually successfully subtract something (e.g. someone throws in a call to simplify).

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+        return true;
+    }
+    return false;
+}
+
+bool is_safe_for_add(const Expr &e) {
+    return is_safe_for_add(e, e.type().bits() / 2 - 1);
+}
+
+// We want to find and remove an add of 'round' from e.
```
abadams, line 86: This whole function seems like a long-winded way of saying simplify(e - round). Did you try that? When did it fail?

abadams: Oh I guess you want to know if it was a free simplification in some sense.

abadams: (Maybe extend the comment to explain)

src/CodeGen_LLVM.cpp
dsharletg: LLVM does claim these support any bit width, and even use 4-bit integers as the example: https://llvm.org/docs/LangRef.html#llvm-sadd-sat-intrinsics

I agree it will likely break on a real target with 48-bit integers, but I think the best way to handle that would be to handle saturating_add/sub in that CodeGen class.

src/CodeGen_LLVM.cpp
dsharletg: Let's consolidate on the comment below, I think it's the same point.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                return rounding_shift(cast(add->type, add->args[0]), b);
+            }
+        }
+
+        // If it wasn't a widening or saturating add, we might still
+        // be able to safely accept the rounding.
+        Expr a_less_round = find_and_subtract(a, round);
+        if (!a_less_round.same_as(a)) {
+            // We found and removed the rounding. However, we may have just changed
+            // behavior due to overflow. This is still save if the type is not
```
abadams, line 150: save -> safe

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+            // overflowing, or we can find a widening add or subtract in the tree
```
abadams, line 151: Why does finding the round term make it safe? What if the round term caused overflow in the original and you just removed it?

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+            // of adds/subtracts. This is a common pattern, e.g.
+            // rounding_halving_add(a, b) = shift_round(widening_add(a, b) + 1, 1).
+            // TODO: This could be done with bounds inference instead of this hack
```
abadams, line 154: Would it be better to just do that? This whole thing is pretty gross.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+
+    Expr visit(const Add *op) override {
+        if (!find_intrinsics_for_type(op->type)) {
+            return IRMutator::visit(op);
+        }
+
+        Expr a = mutate(op->a);
+        Expr b = mutate(op->b);
+
+        // Try widening both from the same type as the result, and from uint.
```
abadams, line 181: the same type -> the same signedness

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+            int pow2 = 0;
+            if (is_const_power_of_two_integer(a, &pow2)) {
+                return mutate(b << cast(UInt(b.type().bits()), pow2));
+            } else if (is_const_power_of_two_integer(b, &pow2)) {
+                return mutate(a << cast(UInt(a.type().bits()), pow2));
+            }
+        }
+
+        // We're applying this to float, which seems OK? float16 * float16 -> float32 is a widening multiply?
+        Expr narrow_a = strip_widening_cast(a);
```
abadams, line 276: What's the difference between strip_widening_cast and lossless_narrow?

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+            return IRMutator::visit(op);
+        }
+
+        Expr a = mutate(op->a);
+        Expr b = mutate(op->b);
+
+        if (const Cast *cast_a = a.as<Cast>()) {
+            Expr cast_b = lossless_cast(cast_a->value.type(), b);
+            if (cast_a->type.can_represent(cast_a->value.type()) && cast_b.defined()) {
+                // This cast can be moved outside the min.
```
abadams, line 330: Is this a good idea if it's a narrowing cast? Surely we'd rather do the min in the narrower type?

Oh, the can_represent means that it's a widening cast of some flavor (or uint8 -> float or something). Please update comment.

src/CodeGen_LLVM.cpp
abadams: Ah, I didn't realize. Weird that it's stringly typed.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                rewrite(rounding_shift_right(cast(op_type_wide, widening_sub(x, y)), 1), rounding_halving_sub(x, y), is_x_same_int_or_uint) ||
+
+                false) {
+                return mutate(rewrite.result);
+            }
+            // clang-format on
+
+            // When the argument is a widened rounding shift, we might not need the widening.
+            // When there is saturation, we can only avoid the widening if we know the shift is
+            // a right shift. Without saturation, we can ignore the widneing.
```
abadams, line 404: widneing

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+            auto is_x_wide_int = op->type.is_int() && is_int(x, bits * 2);
+            auto is_x_wide_uint = op->type.is_uint() && is_uint(x, bits * 2);
+            auto is_x_wide_int_or_uint = is_x_wide_int || is_x_wide_uint;
+            // clang-format off
+            if (rewrite(max(min(rounding_shift_right(x, y), upper), lower), rounding_shift_right(x, y), is_x_wide_int_or_uint) ||
```
abadams, line 409: I don't understand. It's a narrowing case (because x is wider than op->type). How can you remove the saturation?

src/CodeGen_LLVM.cpp
dsharletg: I think emplace_back is a superset of push_back, so just using it everywhere seems reasonable? I got some cases of clang-tidy complaining about push_back as well (probably not this one though).

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                rewrite(rounding_shift_right(x, y), rounding_shift_right(x, y), is_x_wide_int_or_uint) ||
```
abadams, line 410: what?

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                rewrite(rounding_shift_left(x, y), rounding_shift_left(x, y), is_x_wide_int_or_uint) ||
+                false) {
+                const Call *shift = Call::as_intrinsic(rewrite.result, {Call::rounding_shift_right, Call::rounding_shift_left});
```
abadams, line 413: Is this all a janky workaround to not having something in IRMatch that just matches and gives you the wildcards instead of doing a full rewrite?

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                rewrite(rounding_shift_right(x - y, 1), rounding_halving_sub(x, y)) ||
+                false) {
+                return mutate(rewrite.result);
+            }
+            // clang-format on
+        }
+
+        if (op->is_intrinsic(Call::widening_mul)) {
+            internal_assert(op->args.size() == 2);
+            Expr narrow_a = strip_widening_cast(op->args[0]);
```
abadams, line 474: Why strip_widening_cast and not lossless_narrow?

src/CodeGen_LLVM.cpp
dsharletg: Going to resolve the related comments, please reopen if needed.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                rewrite(shift_right(x - y, 1), halving_sub(x, y)) ||
+                rewrite(rounding_shift_right(x + y, 1), rounding_halving_add(x, y)) ||
+                rewrite(rounding_shift_right(x - y, 1), rounding_halving_sub(x, y)) ||
+                false) {
+                return mutate(rewrite.result);
+            }
+            // clang-format on
+        }
+
+        if (op->is_intrinsic(Call::widening_mul)) {
```
abadams, line 472: I believe this whole block of code is handling cases where you widen by 4x and instead widening by 2x and adding an outer cast. If that's right, please add a comment.

src/FindIntrinsics.cpp
```diff
@@ -0,0 +1,714 @@
...
+                Type narrow_t = op->type.narrow().narrow().with_code(t);
+                Expr narrow_a = lossless_cast(narrow_t, op->args[0]);
+                Expr narrow_b = lossless_cast(narrow_t, op->args[1]);
+                if (narrow_a.defined() && narrow_b.defined()) {
+                    return mutate(Cast::make(op->type, widening_sub(narrow_a, narrow_b)));
+                }
+            }
+        }
+
+        if (op->is_intrinsic(Call::shift_right) || op->is_intrinsic(Call::shift_left)) {
```
abadams, line 501: Why do we switch from else if to if here?

src/CodeGen_PTX_Dev.cpp
dsharletg: I don't think we can rewrite to calls that aren't intrinsics?

abadams: Ah. Please add a TODO then.

abadams: (to add that feature to IRMatch)

src/FindIntrinsics.cpp
dsharletg: I think there'd be other problems too, AFAIK the simplifier only tracks min/max for int32? We need this for all types.

src/FindIntrinsics.cpp
dsharletg: strip_widening_cast removes all widening casts, ignoring sign and including more than narrowing (e.g. double narrowing).

dsharletg: Oh, I guess it doesn't handle double narrowing. But it does ignore sign (the implementation is just calling lossless_cast but with some logic for signedness).

src/FindIntrinsics.cpp
dsharletg: Consolidating on the comment below

src/FindIntrinsics.cpp
dsharletg: Consolidating on the comment below

src/FindIntrinsics.cpp
dsharletg: I'd say this is a janky workaround for not being able to express lossless-cast like things in rewrite rules. What we're trying to do here is the following rewrites:

    rewrite(saturating_narrow(rounding_shift_right(widen(x), widen(y))),rounding_shift_right(x, y), can_prove(y > 0))
    rewrite(narrow(rounding_shift_right(widen(x), widen(y))),rounding_shift_right(x, y))
    rewrite(narrow(rounding_shift_right(widen(x), widen(y))),rounding_shift_right(x, y))

Doing this requires a combination of pattern matching and logic, which is in the body of the if.

src/FindIntrinsics.cpp
dsharletg: This is what allows widening_mul to produce mixed sign multiplies. If we were to use lossless_narrow, handling widening_mul would require 4 cases of {signed, unsigned} x {signed, unsigned}.

src/FindIntrinsics.cpp
dsharletg: It doesn't matter, just conceptually breaks things up by "category" of operation. I can change it if you want.

src/FindIntrinsics.cpp
dsharletg: Added to the comment.

src/FindIntrinsics.cpp
dsharletg: Adding comment

src/FindIntrinsics.cpp
dsharletg: Added comment.

src/CodeGen_X86.cpp
dsharletg: Added comment

dsharletg:
I think there are no known test failures from this branch at this point. All of the failing build bots look unrelated.

dsharletg:
The vector_reductions failure on trunk LLVM is (surprisingly) an LLVM failure also happening on master.

steven-johnson:
Not sure what's up with the Python failure. I'm sure it's unrelated but investigating.

steven-johnson:
I think all the failures are unrelated and we're probably OK to merge, but since this is ticklish and the LLVM fix is apparently pending, might not be silly to wait

dsharletg:
I think this might still be waiting for more review anyways (please take a look if so).

src/IRMatch.h
abadams: Don't we already? I thought we had OK'd C++17 features.

src/runtime/ptx_dev.ll
```diff
@@ -346,44 +346,68 @@ define weak_odr i32 @halide_ptx_trap() nounwind uwtable alwaysinline {
 }
 
 ; llvm doesn't expose dot product instructions as intrinsics
-define weak_odr i32 @dp4a_s32_s32(i32 %a, i32 %b, i32 %i) nounwind readnone alwaysinline {
-       %d = tail call i32 asm "dp4a.s32.s32    $0, $1, $2, $3;", "=r,r,r,r"(i32 %a, i32 %b, i32 %i) nounwind readnone
+define weak_odr i32 @dp4a_s32_s32(<4 x i8> %a, <4 x i8> %b, i32 %i) nounwind readnone alwaysinline {
+       %a_32 = bitcast <4 x i8> %a to i32
```
abadams, line 350: Does this result in as-good codegen as before? Now we load something as an i32, bit-cast it to 4xi8, then bitcast it back to i32? Seems risky.

test/correctness/intrinsics.cpp
```diff
@@ -0,0 +1,199 @@
...
+    check(widening_add(i32(u8x), i32(u8y)), i64(widening_add(u8x, u8y)));
+
+    // Tricky case.
+    check(i32(u8x) + 1, i32(widening_add(u8x, u8(1))));
+
+    // Check saturating arithmetic
+    check(i8_sat(i16(i8x) + i8y), saturating_add(i8x, i8y));
+    check(u8_sat(u16(u8x) + u8y), saturating_add(u8x, u8y));
+    check(u8(min(u16(u8x) + u16(u8y), 255)), saturating_add(u8x, u8y));
+    //check(u8(min(i16(u8x) + i16(u8y), 255)), saturating_add(u8x, u8y));
```
abadams, line 100: Why is this commented out? Looks like a valid saturating add to me.

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -169,9 +169,8 @@ class SimdOpCheckHVX : public SimdOpCheckTest {
         check("vavg(v*.h,v*.h):rnd", hvx_width / 2, i16((i32(i16_1) + i32(i16_2) + 1) / 2));
         check("vavg(v*.w,v*.w)", hvx_width / 4, i32((i64(i32_1) + i64(i32_2)) / 2));
         check("vavg(v*.w,v*.w):rnd", hvx_width / 4, i32((i64(i32_1) + i64(i32_2) + 1) / 2));
-        check("vnavg(v*.ub,v*.ub)", hvx_width / 1, i8_sat((i16(u8_1) - i16(u8_2)) / 2));
-        check("vnavg(v*.h,v*.h)", hvx_width / 2, i16_sat((i32(i16_1) - i32(i16_2)) / 2));
-        check("vnavg(v*.w,v*.w)", hvx_width / 4, i32_sat((i64(i32_1) - i64(i32_2)) / 2));
+        check("vnavg(v*.h,v*.h)", hvx_width / 2, i16((i32(i16_1) - i32(i16_2)) / 2));
```
abadams, line 172: What happened to the u8 version?

test/correctness/simd_op_check_hvx.cpp
```diff
@@ -649,7 +651,6 @@ class SimdOpCheckHVX : public SimdOpCheckTest {
         int rfac = 4;
         RDom r(0, rfac);
         check("v*.uw += vrmpy(v*.ub,r*.ub)", hvx_width / 4, sum(u32(in_u8(rfac * x + r))));
-        check("v*.uw += vrmpy(v*.ub,r*.ub)", hvx_width / 4, sum(u32(in_u8(rfac * x + r)) * 34));
```
abadams, line 652: Why was this removed? It looks like a special case of the one below.

src/IRMatch.h
steven-johnson: Nope, C++11 is all we count on at the moment (though C++14 can be enabled via LLVM settings in some setups, I think).

We had agreed to move to C++17 baseline for Halide 12 (once we cut the Halide 11 release).

abadams (approved):
A few remaining nits / questions but looks good

src/IRMatch.h
abadams: Ah, I forgot we were waiting till post-release. I told @rootjalex he was good to use constexpr if and was suddenly worried that was a lie. So we can use c++17 in branches that won't merge until after the release.

src/runtime/ptx_dev.ll
dsharletg: I think it's pretty safe. The bitcasts will certainly simplify away, and I'd expect even quite a degree of shuffles to simplify away.

The generated PTX code in at least a few of the test cases in cuda_8_bit_gemm is *identical* to master, including the register names, which suggests to me that the optimized LLVM IR is identical (not just similar).

test/correctness/simd_op_check_hvx.cpp
dsharletg: The simplifier factors the constant out of the vector reduction. The old code had some extra logic to distribute these, but this distribution is somewhat tricky to get right. If it doesn't match to a useful pattern, it could be a significant pessimization (moving a constant multiply inside a reduction loop).

At the same time, these patterns seem only marginally useful. It's basically a box filter but with a non-unit constant factor...? The case above (sum without a constant) probably covers 99% of this case.

However, I did bring back these cases in a way that checks that the reduction generates vrmpy at least for the box filter (with constant 1) part.

test/correctness/simd_op_check_hvx.cpp
dsharletg: I put it back. It requires a special case, and I thought it was a marginal pattern.

test/correctness/intrinsics.cpp
dsharletg: I brought it back. Matching it requires matching a cast in the pattern, which was stumping me earlier.

dsharletg:
@pranavb-ca @aankit-ca Please take a look particularly at the Hexagon pattern changes. This PR reduces the capabilities of pattern matching in a few cases, but I think these changes are OK. I think the old distributing of multiplies through broadcasts is a bit dangerous. If we don't match a pattern, this is potentially a significant pessimization. At the same time, I think this use case is rare. It's a sum multiplied by a constant, which doesn't make a lot of sense. There is almost always an optimization to such an algorithm that removes the constant completely?

I also added a few TODOs that we should address. There's also not a TODO for it, but we should also try to support rounding shifts after this PR, which should be easy.

steven-johnson:
https://buildbot.halide-lang.org/master/#/builders/24/builds/28:

```
The following tests FAILED:
	 87 - correctness_cuda_8_bit_dot_product (Failed)
```

that's a new one.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -447,13 +448,13 @@ void CodeGen_PTX_Dev::codegen_vector_reduce(const VectorReduce *op, const Expr &
         {VectorReduce::Add, 4, i32(widening_mul(wild_u8x, wild_i8x)), "dp4a_u32_s32"},
         {VectorReduce::Add, 4, u32(widening_mul(wild_u8x, wild_u8x)), "dp4a_u32_u32"},
         {VectorReduce::Add, 4, widening_mul(wild_i16x, wild_i16x), "dp2a_s32_s32", Pattern::NarrowOp1},
-        {VectorReduce::Add, 4, widening_mul(wild_i16x, wild_u16x), "dp2a_s32_u32", Pattern::NarrowOp1},
+        {VectorReduce::Add, 4, widening_mul(wild_i16x, wild_i16x), "dp2a_s32_u32", Pattern::NarrowUnsignedOp1},
```
abadams, line 451: Is the second arg here signed or unsigned?

dsharletg: It's unsigned, widened to a signed integer.

abadams: Then why are both args wild_i16x?

dsharletg: When widening a uint8, it can be widened to either a uint16 or an int16. This is a pretty common issue, see e.g. https://github.com/halide/Halide/blob/master/src/HexagonOptimize.cpp#L534-L535 (on master).

`widening_mul` eliminates a lot of this issue, but it doesn't fully eliminate it in the mixed precision case. I suppose we could generalize `widening_mul` further to strip widening casts "unevenly", so this pattern would be `widening_mul(wild_i16x, wild_u8x)`...

dsharletg: If we generalized it to support mixed precision, we'd have to be pretty careful about constants. We'd probably need to always narrow constants to the narrowest type they can represent.

dsharletg: I realized a bit better way to handle this, using the overloaded intrinsics.

src/runtime/ptx_dev.ll
dsharletg: I've now verified that the codegen is identical for all of these tests.

aankit-ca:
@dsharletg We are getting a couple of failures on some internal tests. Could the merge wait for until after the Christmas break?

steven-johnson:
Post-holiday check-in: where does this PR stand? Buildbots are clean. Is there more work to be done?

dsharletg:
@aankit-ca can you just try it again? I made some small fixes that might have fixed the issues you were seeing.

src/HexagonOptimize.cpp
```diff
@@ -1214,135 +1154,180 @@ class VectorReducePatterns : public IRMutator {
...
+            {4, 32, widening_mul(wild_u8x, wild_u8x)},
+            {4, 32, widening_mul(wild_u8x, wild_i8x)},
+            {4, 32, widening_mul(wild_i8x, wild_u8x), Signature::SwapOps},
+            {4, 32, widening_mul(wild_i8x, wild_i8x)},
+            // Sum
+            {4, 32, wild_u8x, Signature::SlidingWindow},
+            {4, 32, wild_u8x, Signature::SlidingWindow},
+            {4, 32, wild_i8x, Signature::SlidingWindow},
+            {4, 32, wild_u8x},
+            {4, 32, wild_u8x},
```
aankit-ca, line 1202: Repeated pattern

src/HexagonOptimize.cpp
```diff
@@ -1214,135 +1154,180 @@ class VectorReducePatterns : public IRMutator {
...
-            {4, UInt(32, 0), UInt(8, 0), UInt(8, 32)},
-            {4, Int(32, 0), UInt(8, 0), Int(8, 32)},
-            {4, Int(32, 0), Int(8, 0), Int(8, 32)},
+            {4, 32, widening_mul(wild_u8x, wild_u8x)},
+            {4, 32, widening_mul(wild_u8x, wild_i8x)},
+            {4, 32, widening_mul(wild_i8x, wild_u8x), Signature::SwapOps},
+            {4, 32, widening_mul(wild_i8x, wild_i8x)},
+            // Sum
+            {4, 32, wild_u8x, Signature::SlidingWindow},
+            {4, 32, wild_u8x, Signature::SlidingWindow},
```
aankit-ca, line 1199: Same pattern in both lines

aankit-ca:
@dsharletg Sorry to keep you waiting. Yeah the recent commits resolved the compile-time failures. There is still a failure in a histogram test but that occurs only with v65 feature flag when we try use VTCM.

dsharletg:
Thanks for looking. 

Can you share more about the failing test? We have a correctness test that sounds like that, but this succeeds for me on this branch:

    HL_JIT_TARGET=host-hvx-hvx_v65 make correctness_histogram

aankit-ca:
@dsharletg We are seeing some significant performance regressions. Even sobel degraded by ~ 27%.

aankit-ca:
> Thanks for looking.
> 
> Can you share more about the failing test? We have a correctness test that sounds like that, but this succeeds for me on this branch:
> 
> ```
> HL_JIT_TARGET=host-hvx-hvx_v65 make correctness_histogram
> ```
> 
> This test uses TCM on v65.

We are trying to reduce the testcase. Will update you once I have more information.

dsharletg:
In the hexagon_benchmarks app, I'm seeing a negligible performance difference between this branch and master... can you clarify which code you are seeing is slower?

dsharletg:
BTW, it might be worth a quick look at what the failure is. It might be easier to debug it than to build a reproducer :) A lot of the failures I fixed on this branch are just small corner cases in pattern matching. But some are hard to debug, and I'm happy to look at a reproducer too.

aankit-ca:
> In the hexagon_benchmarks app, I'm seeing a negligible performance difference between this branch and master... can you clarify which code you are seeing is slower?

The sobel we test has a slightly different from the one in hexagon_benchmarks.

aankit-ca:
> BTW, it might be worth a quick look at what the failure is. It might be easier to debug it than to build a reproducer :) A lot of the failures I fixed on this branch are just small corner cases in pattern matching. But some are hard to debug, and I'm happy to look at a reproducer too.

Absolutely.

aankit-ca:
@dsharletg I’m getting exact same cycles on simulator too. On device I'm getting pretty much the same numbers on master and here as well.
Also, the histogram crash seems to be because of an unrelated problem. The placement of scatter-release op needs to be changed to correct the bug.

src/HexagonOptimize.cpp
```diff
@@ -1169,20 +1123,6 @@ class OptimizePatterns : public IRMutator {
...
-        return IRMutator::visit(op);
-    }
-
-    Expr visit(const Mod *op) override {
-        if (!op->type.is_float() && op->type.is_vector()) {
-            return mutate(lower_int_uint_mod(op->a, op->b));
-        }
-        return IRMutator::visit(op);
-    }
-
```
aankit-ca, line 1185: @dsharletg Do we not want to use long division here? I see degradations in mul_div_mod because of this?

Master

Testing div_mod vector_width: 128
Hexagon simulator executed function 0x91af0 in 546000 cycles
Hexagon simulator executed function 0x98530 in 2262000 cycles
Hexagon simulator executed function 0xaae50 in 15386000 cycles
Hexagon simulator executed function 0x94a60 in 978000 cycles
Hexagon simulator executed function 0xb1590 in 4468000 cycles
Hexagon simulator executed function 0xca5f0 in 24041000 cycles

Here:

Testing div_mod vector_width: 128
Hexagon simulator executed function 0xa1050 in 21911000 cycles
Hexagon simulator executed function 0xaacc0 in 23752000 cycles
Hexagon simulator executed function 0xb42c0 in 24363000 cycles
Hexagon simulator executed function 0xcebf0 in 50166000 cycles
Hexagon simulator executed function 0xef550 in 55319000 cycles
Hexagon simulator executed function 0xd8ee0 in 58044000 cycles

dsharletg: At one point, I had moved this to FindIntrinsics, but reverted that. Fixed.

dsharletg:
The last change only affected Hexagon, which passed. The previous runs on the rest of the build bots did pass.

@aankit-ca, do you see anything else that needs to be addressed before merging?

aankit-ca:
@dsharletg That's it. LGTM.

dsharletg:
This didn't pass all the build bots, but I'm going to merge it anyways. Over the last week+, all of these build bots have passed several times, and the current failures are not related to any changes pushed since then. Two of them are recent LLVM regressions also affecting master. One of the failures is #5634. One looks like a tooling failure.

steven-johnson:
Good call :-)

PR #5532, opened 2020-12-08
Solve the COMDAT in runtime failing on Mac OS X problem once and for all.

zvookin:
This removes Comdat items from LLVM IR when linking a runtime for Mac OS or iOS. This would always be a failure with the current code and it is anticipated that this will either produce one valid symbol or fail with a multiple definition error. I have tested this with the GPU context consistency changes and it does fix the error that code was running into.


abadams (approved):
Might be worth adding to the comment that we don't strip them for all targets because windows explicitly requires them.

steven-johnson (approved):
LGTM pending all green

PR #5535, opened 2020-12-08
Pathnames may or may not be absolute so loosen comparison to allow for this.

zvookin:
Fixes introspection test failure on Mac OS when compiled via Makefile.

### DIFF
--- a/test/correctness/introspection.cpp
+++ b/test/correctness/introspection.cpp
@@ -5,6 +5,21 @@ using namespace Halide;
 
 // The check has to go in the Halide namespace, because get_source_location looks for the first thing outside of it
 namespace Halide {
+bool paths_equal(const std::string &path1, const std::string &path2) {
+    int length_delta = path1.size() - path2.size();
+    if (length_delta > 0) {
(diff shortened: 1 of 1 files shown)


test/correctness/introspection.cpp
```diff
@@ -5,20 +5,34 @@ using namespace Halide;
 
 // The check has to go in the Halide namespace, because get_source_location looks for the first thing outside of it
 namespace Halide {
+bool paths_equal(const std::string &path1, const std::string &path2) {
+    int length_delta = path1.size() - path2.size();
```
steven-johnson, line 9: Will `int = size_t - size_t` do the right thing for negative cases in all compiler environments? I confess I forget this part of the C int conversion spec

abadams (approved):

test/correctness/introspection.cpp
zvookin: We probably should turn on that set of warnings some year. Fixing now.

steven-johnson (approved):

Issue #5523, opened 2020-12-04
Not able to generate blur example for Hexagon DSP


mikbjorklund:
It is not possible to enable the hexagon scheduling directive, (include the hvx target feature in the target.)

```console
~/Halide/apps/blur (master) $ HL_TARGET=arm-64-android-hvx_128 make
bin/host/halide_blur.generator -g halide_blur -e static_library,h,registration,stmt,assembly -o bin/arm-64-android-hvx_128 target=arm-64-android-hvx_128
free(): invalid pointer
Makefile:13: recipe for target 'bin/arm-64-android-hvx_128/halide_blur.a' failed
make: *** [bin/arm-64-android-hvx_128/halide_blur.a] Segmentation fault (core dumped)
```

steven-johnson:
Tagging @pranavb-ca for initial triage

pranavb-ca:
Thanks for bringing to my notice @steven-johnson. Will look into this this afternoon.

pranavb-ca:
I coudln't reproduce this immediately, however @aankit-ca will dig deeper to double-check. BTW, I did have to add `CXX-arm-64-android-hvx_128` and `LDFLAGS-arm-64-android-hvx_128` to `apps/support/Makefile.inc`.

dsharletg:
Hopefully HL_TARGET=arm-64-android-hvx_128 make bin/arm-64-android/... should work? If not we should fix that so it does. Adding a CXX-* and LDFLAGS-* for every combination of optional flags doesn't scale (what about hvx_128-profile? or profile-hvx_128? and so on...)

aankit-ca:
Even I'm not able to reproduce the error and faced the same issue as @pranavb-ca .

HL_TARGET=arm-64-android-hvx_128 make bin/arm-64-android/test
can work with this change: https://github.com/halide/Halide/pull/5536

mikbjorklund:
The actual reason why I raised this ticket is that I am not able to generate the halide_blur lib for hvx_128:
[...]

PR #5536, opened 2020-12-08
Change generator target for using optional target flags for apps/blur

aankit-ca:
Without the change:

HL_TARGET=arm-64-android-hvx_128 make bin/arm-64-android/test

would have target=arm-64-android instead of arm-64-android-hvx_128

@dsharletg PTAL

### DIFF
--- a/apps/blur/Makefile
+++ b/apps/blur/Makefile
@@ -11,7 +11,7 @@ $(GENERATOR_BIN)/halide_blur.generator: halide_blur_generator.cpp $(GENERATOR_DE
 
 $(BIN)/%/halide_blur.a: $(GENERATOR_BIN)/halide_blur.generator
 	@mkdir -p $(@D)
-	$^ -g halide_blur -e $(GENERATOR_OUTPUTS) -o $(@D) target=$*
+	$^ -g halide_blur -e $(GENERATOR_OUTPUTS) -o $(@D) target=$(HL_TARGET)
 
 # g++ on OS X might actually be system clang without openmp
 CXX_VERSION=$(shell $(CXX) --version)


steven-johnson:
I'm not sure I like this fix -- it changes the pattern we use ~everywhere else in Make. 

Why can't the user just do `make bin/arm-64-android-hvx_128/test`?

aankit-ca:
This seemed like a quick fix for https://github.com/halide/Halide/issues/5523.
For `make bin/arm-64-android-hvx_128/test` to work we will need:
CXX-arm-32-android-hvx_128 in apps/support/Makefile and as DIllon pointed out having every combination of optional flag is not feasible.

dsharletg:
I think this is a good change, but I think maybe we should go ahead and do it for all the apps?

The reason I think it is good is because the strategy used for selecting compilers in Makefile.inc doesn't work without it. We shouldn't need a different host compiler/linker flag for each possible combination of optional target flags, only the arch/OS.

Also, I think the current thing is simply broken. `HL_TARGET=arm-64-android-hvx_128 make bin/arm-64-android/blur` ignores `HL_TARGET`, that seems like a bug to me.

steven-johnson:
I'll prep a PR to make this change uniformly across the apps.

abadams:
I don't think this is a good change. It's much better to distinguish between Halide targets using the make target. Messing with environment variables and having the makefile depend on them breaks make's dependency tracking when you're changing targets.

Most makefile rules should entirely ignore HL_TARGET. The only exception is generic targets like "test"

abadams:
Here's how to select compilers for cross-compilation without depending on env vars:

```
FILTER_CXX ?= c++
FILTER_LDFLAGS ?= -lwhatever
FILTER_CXX_FLAGS ?= -I whatever

... the generator rules ...

$(BIN)/%/filter: filter.cpp $(BIN)/%/my_halide_filter.a $(BIN)/%/runtime.a
	@mkdir -p $(@D)
	$(FILTER_CXX) $(FILTER_CXXFLAGS) -I$(BIN)/$* -Wall -O3 $^ -o $@ $(FILTER_LDFLAGS) 

# Tell make to use a different compiler and flags for some targets:
bin/arm-64-android/filter: FILTER_CXX=$(ARM64_CXX)
bin/arm-64-android/filter: FILTER_LDFLAGS=-ldl -llog -DHALIDE_NO_JPEG -DHALIDE_NO_PNG -static-libstdc++

# You can use wildcards too:
bin/arm-64-android-%/filter: FILTER_CXX=$(ARM64_CXX)
bin/arm-64-android-%/filter: FILTER_LDFLAGS=-ldl -llog -DHALIDE_NO_JPEG -DHALIDE_NO_PNG -static-libstdc++ -lEGL
```

steven-johnson:
(See https://github.com/halide/Halide/pull/5539 for a complete suite of this change.)

steven-johnson:
At this point it seems like this change isn't likely to happen, at least not in this form.

PR #5539, opened 2020-12-09
Change apps/ Makefiles handling of HL_TARGET

steven-johnson:
Always specify a Generator's `target` value as `$HL_TARGET`, rather than `$*`; this allows us more flexibility to cross-compile while adding extra features to HL_TARGET that may not be reflected in the base build tooling.

### DIFF
--- a/apps/HelloPyTorch/Makefile
+++ b/apps/HelloPyTorch/Makefile
@@ -84,7 +84,7 @@ $(BIN)/%/add_float32.a: $(GENERATOR_BIN)/add.generator
 		-f add_float32 \
 		-e static_library,c_header,pytorch_wrapper \
 		-o $(@D) \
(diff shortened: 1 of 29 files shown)


steven-johnson:
(skipping buildbots for now since this change is controversial)

abadams:
This breaks any workflow that involves rapidly toggling between two different targets (e.g. adding and removing -profile, or -cuda, or -debug). It makes it really easy to have the actual compiled target not match the target implied by the path. The effect is that you "make clean" and rebuild constantly. Moving away from depending on environment variables in the makefiles was a major usability improvement for me, and I'd hate to backslide. I think there are other ways to deal with the cross-compiler issue (I posted one in the other PR).

steven-johnson:
Closing this for now.

PR #5540, opened 2020-12-09
Fix fragile Makefile for apps/onnx

steven-johnson:
The protoc usage happened to work when building in the app folder but often failed when building from the toplevel Makefile. Also, drive-by silencing of noise from curl, and drive-by fix of "redundant copy" warning in onnx_converter.cc.


steven-johnson:
Review ping

alexreinking (approved):

PR #5537, opened 2020-12-08
Add code to allow running compare_vs_tflite on Android devices

steven-johnson:

PR #5541, opened 2020-12-09
Fix rounding-shift, tweak compare_vs_tflite

steven-johnson:
round_shift_right() now mimics the logic of ARM rounding-shift instructions (and should emit them once some upstream changes land). Tweaked compare_vs_tflite to try to minimize error reporting as a result.

(Note, this is  additive to the change in https://github.com/halide/Halide/pull/5537)

### DIFF
--- a/apps/interpret_nn/halide/common_halide.cpp
+++ b/apps/interpret_nn/halide/common_halide.cpp
@@ -80,9 +80,9 @@ Expr round_shift_right(const Expr &x, const Expr &exponent) {
     Type t_unsigned = t.with_code(halide_type_uint);
     Expr uexponent = cast(t_unsigned, exponent);
     Expr one = cast(x.type(), 1);
-    // This is intended to pattern-match against rounding-shift-right instruction generation
+    // This is intended to pattern-match against rounding-shift-right instruction generation.
     Expr rounding = (one << uexponent) >> 1;
-    return ((x + rounding) >> uexponent);
+    return (saturating_cast(t, x + rounding) >> uexponent);
 }
 
 Expr multiply_quantized(const Expr &x, const Expr &q, const Expr &shift) {
--- a/apps/interpret_nn/halide/common_halide.cpp
+++ b/apps/interpret_nn/halide/common_halide.cpp
@@ -4,6 +4,17 @@ using namespace Halide;
 
 namespace interpret_nn {
 
+namespace {
+
+Expr saturating_add(const Expr &a, const Expr &b) {
+    const Type t = a.type();
+    assert(t == b.type());
+    const Type wt = t.widen();
+    return saturating_cast(t, cast(wt, a) + cast(wt, b));
+}
+
+}  // namespace
+
 void interpret_as_tensor(OutputImageParam p) {
     p.dim(0).set_stride(1).set_min(0);
 }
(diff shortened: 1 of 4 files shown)


apps/interpret_nn/halide/common_halide.cpp
```diff
@@ -78,12 +79,10 @@ Expr round_shift_right(const Expr &x, const Expr &exponent) {
...
     Type t_unsigned = t.with_code(halide_type_uint);
     Expr uexponent = cast(t_unsigned, exponent);
-    // Exponent must satisfy 0 <= exponent <= 31
-    // TODO: Maybe this should be an offset added to x prior to shifting.
-    Expr mask = (cast(x.type(), 1) << uexponent) - 1;
-    Expr remainder = x & mask;
-    Expr threshold = (mask >> 1) + (x < 0);
-    return (x >> uexponent) + (remainder > threshold);
+    Expr one = cast(x.type(), 1);
+    // This is intended to pattern-match against rounding-shift-right instruction generation
```
dsharletg, line 83: Add comment that it assumes the type is int32, otherwise the rounding add would need to be a saturating add.

dsharletg (approved):

apps/interpret_nn/halide/common_halide.cpp
steven-johnson: I hadn't realize that saturating was part of this. I'd better go back and look at the spec. (And will add the comment.)

steven-johnson: ...or better yet, just make it a saturating add now :-)

apps/interpret_nn/halide/common_halide.cpp
```diff
@@ -78,12 +79,10 @@ Expr round_shift_right(const Expr &x, const Expr &exponent) {
...
-    // Exponent must satisfy 0 <= exponent <= 31
-    // TODO: Maybe this should be an offset added to x prior to shifting.
-    Expr mask = (cast(x.type(), 1) << uexponent) - 1;
-    Expr remainder = x & mask;
-    Expr threshold = (mask >> 1) + (x < 0);
-    return (x >> uexponent) + (remainder > threshold);
+    Expr one = cast(x.type(), 1);
+    // This is intended to pattern-match against rounding-shift-right instruction generation.
+    Expr rounding = (one << uexponent) >> 1;
+    return (saturating_cast(t, x + rounding) >> uexponent);
```
dsharletg, line 85: I think this isn't quite a saturating add, you would need to widen x first. Something like:

    saturating_cast(t, cast(x.type().widen(), x) + rounding)

It's a bit messy for generic type code :( It's better if we have a `widen` helper function, and a `saturating_narrow` helper doesn't hurt either:

    return saturating_narrow(widen(x) + rounding) >> uexponent;

steven-johnson: ...why don't we have explicit helpers for these in (eg) IROperators.h, in addition to pattern matching?

steven-johnson: PTAL

steven-johnson: > ...why don't we have explicit helpers for these in (eg) IROperators.h, in addition to pattern matching?

I see that these are being added in https://github.com/halide/Halide/pull/5531 -- woot

dsharletg: They're being added, but not that in that branch, they're still in the "Internal" namespace, and shouldn't be used directly. Promoting them is something we can talk about too, but in a separate change of course.

PR #5544, opened 2020-12-10
Windows on ARM64 support (CPU, and also GPU through D3D12)

slomp:


shoaibkamil:
Tested working on my Surface Pro X for both CPU and D3D12Compute pipelines.

src/runtime/windows_d3d12compute_arm.cpp
```diff
@@ -0,0 +1,13 @@
+// NOTE(marcos): this file is a trivial wrapper around the "core" d3d12 back-end
+// located in "d3d12compute.cpp"; this wrapper is used as a mean to "decorate" the
```
shoaibkamil, line 2: Nit: `s/mean/means`

src/runtime/CMakeLists.txt
```diff
@@ -168,20 +169,52 @@ foreach (i IN LISTS RUNTIME_CPP)
     foreach (j IN ITEMS 32 64)
         # -fpic needs special treatment; see below on windows 64bits
         set(fpic -fpic)
-        if (j EQUAL 32)
-            if (i MATCHES "windows_.*")
-                # win32 uses the stdcall calling convention, which is x86-specific
-                set(TARGET "i386-unknown-unknown-unknown")
-            else ()
+        # for the generic windows 64-bit target, we need -fshort-wchar
```
shoaibkamil, line 172: I believe the Makefile also need changes corresponding to these

src/runtime/windows_d3d12compute_arm.cpp
```diff
@@ -0,0 +1,13 @@
...
+// module such that the proper build flags and target triple can be determined based
+// on the prefix (windows_) and suffix (_arm) when building the runtime module.
+
+#ifndef BITS_64
+// Don't emit compiler messages (via #pragma message) here: some environments might
```
abadams, line 7: So does this mean the windows 32-bit d3d runtime module is built but is empty?

slomp: Yes, pretty much.
Do we have some sort of "null/parrot" runtime module to reroute instead?

src/runtime/CMakeLists.txt
steven-johnson: There is ~no support for building any Windows targets using the Makefiles, so, probably not

EDIT: oh, wait, for crosscompiling from (say) linux -> windows we do need the Makefile to get this right, so, yes, needs attention

src/runtime/CMakeLists.txt
```diff
@@ -168,20 +169,52 @@ foreach (i IN LISTS RUNTIME_CPP)
...
+                    set(fpic "")
+                    set(TARGET "arm-unknown-windows-unknown")
+                else ()
+                    set(fpic "")
+                    set(TARGET "aarch64-unknown-windows-unknown")
+                endif ()
+            # Windows Generic
+            else()
+                if (j EQUAL 32)
+                    # TODO(marcos): generic code won't hold for ARM32... If ARM32 support becomes necessary,
```
steven-johnson, line 200: Do we need to support Windows-for-ARM32 (now or ever)?

src/runtime/CMakeLists.txt
```diff
@@ -200,7 +233,7 @@ foreach (i IN LISTS RUNTIME_CPP)
             set(INITMOD "_initmod_${i}_${j}${SUFFIX}.cpp")
             set(SYMBOL "halide_internal_initmod_${i}_${j}${SUFFIX}")
 
-            set(clang_flags ${RUNTIME_CXX_FLAGS} ${fpic} ${RUNTIME_DEFINES${SUFFIX}} -m${j} -target ${TARGET} -emit-llvm -S)
+            set(clang_flags ${RUNTIME_CXX_FLAGS} ${fpic} ${fshort-wchar} ${RUNTIME_DEFINES${SUFFIX}} -m${j} -target ${TARGET} -emit-llvm -S)
```
alexreinking, line 236: if we add a third one of these (`-fPIC`, `-fshort-wchar`), we should turn this into a list...

src/runtime/CMakeLists.txt
slomp: I don't think so. We can worry about it when (and if) it happens. ARM64 is way more relevant here.
(And I don't think there is any Windows on ARM 32 device out there that supports D3D12)

src/runtime/CMakeLists.txt
```diff
@@ -168,20 +169,52 @@ foreach (i IN LISTS RUNTIME_CPP)
...
-            else ()
+        # for the generic windows 64-bit target, we need -fshort-wchar
+        set(fshort-wchar "")
+        # Windows
+        if (i MATCHES "windows_.*")
+            # Windows x86/x64
+            if (i MATCHES "windows_.*_x86$")
+                if (j EQUAL 32)
+                    # win32 uses the stdcall calling convention, which is x86-specific
+                    set(fpic "")
```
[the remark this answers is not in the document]
slomp: Good point; these conditions changed so much that I missed the fact that they converged to the same settings.

src/runtime/CMakeLists.txt
slomp: I'm all for `${windows_shenanigans_flags}` :)

steven-johnson:
Looks like this still has legit failures on the buildbots?

shoaibkamil:
Yes, we need to fix the Makefile, but @slomp and I are hunting down another bug we discovered with passing narrower types to D3D12Compute.

steven-johnson:
IMHO it would be good to add this target to correctness/cross_compilation

shoaibkamil:
I believe the vector reductions failure is unrelated/known.  PTAL

steven-johnson (approved):
LGTM -- the failures of vector_reductions on LLVM12 builds is a current known bug; if those are the only failures, go ahead and land

steven-johnson:
Failures unrelated; landing.

Issue #4651, opened 2020-02-26
Should we do selective export in Halide?


alexreinking:
(FYI, there is a history here; we used to do selective export, but it proved hard to maintain, for various reasons, and we ended up abandoning it. Opening a new issue for future discussion would be worthwhile.)

_Originally posted by @steven-johnson in https://github.com/halide/Halide/pull/4644_

----

CMake provides a [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) module that makes it somewhat easier to selectively export library symbols in a cross-compiler-compatible way. That, along with testing discipline to cover all the symbols we want to export, should ease the maintenance burden.

The pros of doing this include:

* Smaller binary sizes
* Explicit API surface, which would make it easier to assign semantic version numbers, rather than dates. 
  * If we remove an export declaration, bump the major version; if we add one, bump the minor; if we want to release changes without touching any of these, bump the patch.

Cons:

* Potentially difficult to implement in the Makefile, though it only supports POSIX OSes and GCC-compatible compilers.
* Maintenance was historically a burden.

abadams:
Halide is designed to be extensible (e.g. Func::add_custom_lowering_pass). People extending Halide may want to reuse all sorts internal tools from outside the compiler. It's also nice to be able to test internal components from an external test, and for external plugins like the autoscheduler to have access to all the internal tools. All this made explicitly exporting symbols a game of whack-a-mole where we were slowly creeping towards exporting everything for one reason or another. It also required littering the source with HALIDE_EXPORT macros. I believe the binary size savings were found to be insignificant (IIRC), largely because we were approaching manually exporting everything.

Regarding versioning - the intent is that namespace Halide is our public API, and changing something there must be done with care because it might break users. But for Halide::Internal we promise nothing (and people using those symbols just have to deal with changes).

So because the binary size difference was close to zero, and we already have an explicit API surface via namespaces, I claim that there are no real pros. Just exporting everything eliminates a whole class of failures at minimal cost.

alexreinking:
> Halide is designed to be extensible (e.g. Func::add_custom_lowering_pass). People extending Halide may want to reuse all sorts internal tools from outside the compiler. It's also nice to be able to test internal components from an external test, and for external plugins like the autoscheduler to have access to all the internal tools.

Between this and:

> But for Halide::Internal we promise nothing (and people using those symbols just have to deal with changes).

[...]

PR #5548, opened 2020-12-11
Hide inaccessible symbols in internal linkage

dsharletg:
I had to debug an issue on the arm-32 buildbots recently, and linking libHalide.so was taking ages to run. I don't know if this helps, but it seems like a good thing to do regardless. We already do a lot of this, just not consistently.

Related to #4651.


steven-johnson (approved):

abadams:
+1

I think we should export all symbols to enable crazy experimentation, but only symbols that have a definition in Halide.h. So things inside c++ files that don't appear in a header should definitely be in an anonymous namespace.

dsharletg:
Note that I did move some of the GPU codegen classes to the cpp file, so this PR is hiding some symbols from the header.

abadams:
I noticed that, but those things really should have been internal.

PR #5558, opened 2020-12-14
cmake: detect ppc64le arch

cimes-isi:


cimes-isi:
This simple patch detects the `ppc64le` architecture, as found on the ORNL Summit supercomputer.  I built successfully using this patch with the v10.0.0 tag, though recent changes to Hexagon are preventing me from building to completion on the latest master.  I will open an issue for those problems.  Additionally, I can't complete all the unit tests, which thus far seem to be environmental problems, but I can't guarantee there aren't still correctness issues.  @jrk has indicated that they have used IBM POWER9 in the past though.

steven-johnson:
LGTM but deferring to Alex

alexreinking (approved):

alexreinking:
Thanks for the patch!

PR #5560, opened 2020-12-14
Better document how LLVM_DIR works.

alexreinking:
Hopefully this will make understanding the CMake build easier when using a custom-built LLVM.

This also adjusts `README.md` to use the same directory layout as `README_cmake.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ repository (where this README is).
 
 At any point in time, building Halide requires either the latest stable version
 of LLVM, the previous stable version of LLVM, and trunk. At the time of writing,
-this means versions 10.0 and 9.0 are supported, but 8.0 is not. The commands
+this means versions 11.0 and 10.0 are supported, but 9.0 is not. The commands
 `llvm-config` and `clang` must be somewhere in the path.
 
 If your OS does not have packages for llvm, you can find binaries for it at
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -150,28 +150,27 @@ If you wish to build Halide in a separate directory, you can do that like so:
...
+% cmake -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=$LLVM_ROOT/lib/cmake/llvm -S . -B build
+% cmake --build build
 ```
 
-`LLVM_DIR` should be the folder in the LLVM installation or build tree that
-contains `LLVMConfig.cmake`. It is not required if you have a suitable
-system-wide version installed. If you have multiple system-wide versions
-installed, you can specify the version with `HALIDE_REQUIRE_LLVM_VERSION`. Add
-`-G Ninja` if you prefer to build with the Ninja generator.
+`LLVM_DIR` is the folder in the LLVM installation tree (do not use the build
```
shoaibkamil, line 160: Nit: maybe bold the "do not use the build tree by mistake"

shoaibkamil (approved):

zvookin (approved):

README.md
alexreinking: Good idea. Done.

PR #5561, opened 2020-12-15
Fix some wasm issues

steven-johnson:
- If `wasm_threads` is in the target string, be sure to launch the external shell with `--experimental-wasm-threads`.

- Always use `posix_threads` instead of `fake_threads` for wasm; this used to fail, but apparently recent versions of Emscripten now provide stubs of the calls we need. The goal here is to simplify linkage of the Halide runtime; the current situation means that attempting to mix Halide-generated code with `wasm_threads` both enabled and disabled requires the thread-enabled runtime to work, but getting this to link correctly in some build systems can be tricky, and the failure mode is obscure.

The second change is the trickier one; in theory this shouldn't matter for code that doesn't use `wasm_threads`, but full test coverage is hard to verify. It seems to work fine in all my local testing.

### DIFF
--- a/src/LLVM_Runtime_Linker.cpp
+++ b/src/LLVM_Runtime_Linker.cpp
@@ -809,11 +809,11 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
                 modules.push_back(get_initmod_posix_io(c, bits_64, debug));
                 modules.push_back(get_initmod_linux_host_cpu_count(c, bits_64, debug));
                 modules.push_back(get_initmod_linux_yield(c, bits_64, debug));
-                if (t.has_feature(Target::WasmThreads)) {
-                    modules.push_back(get_initmod_posix_threads(c, bits_64, debug));
-                } else {
-                    modules.push_back(get_initmod_fake_thread_pool(c, bits_64, debug));
-                }
+                // With recent versions of Emscripten, it appears we can use the posix-thread-pool
+                // always, even if `wasm_threads` is not enabled. Will need some more testing
+                // to be sure, but this change greatly simplifies linking together of multiple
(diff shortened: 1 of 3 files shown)


src/LLVM_Runtime_Linker.cpp
```diff
@@ -809,11 +809,11 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
...
-                if (t.has_feature(Target::WasmThreads)) {
-                    modules.push_back(get_initmod_posix_threads(c, bits_64, debug));
-                } else {
-                    modules.push_back(get_initmod_fake_thread_pool(c, bits_64, debug));
-                }
+                // With recent versions of Emscripten, it appears we can use the posix-thread-pool
+                // always, even if `wasm_threads` is not enabled. Will need some more testing
+                // to be sure, but this change greatly simplifies linking together of multiple
+                // wasm modules if they have mixed target strings (some with threads, some without).
+                modules.push_back(get_initmod_posix_threads(c, bits_64, debug));
```
abadams, line 816: So what does wasm_threads even do as a target feature then? If I leave it off and use Func::parallel, what happens?

steven-johnson: It enables extra compiler flags in the Wasm backend. Which...uh... huh, that's a good question, I don't think there's anything else in Halide's code that stops it from trying to run parallel jobs, so, something is wonky. Let me investigate further.

steven-johnson:
The threading issue is frustratingly hard to track down; among other things, there doesn't seem to currently be a command-line wasm executor tool that is both up-to-date with current LLVM backend (due to instruction churn) and also supporting threads properly (eg, d8 will allow you to enable threads but will never execute more than one). Gonna close this and try to apply the other stuff for now separately.

Issue #5559, opened 2020-12-14
Hexagon compilation failure on ppc64le arch


cimes-isi:
With PR #5558, I can build Halide v10.0.0, but recent changes may have broken the build on ppc64le, which I understand is not well-tested.  I'm using a LLVM 11.0.0-rc1 module on a HPC system.  I'm not sure if this is a build issue with the LLVM module that is appearing now due to a new dependency on LLVM features not previously used, or if this is actually an issue with recent Halide changes. Thanks.

```
[ 45%] Building CXX object src/CMakeFiles/Halide.dir/CodeGen_Hexagon.cpp.o
/ccs/home/cimes/casper/Halide-master/src/CodeGen_Hexagon.cpp:544:6: error: no member named 'hexagon_V6_vzb_128B' in namespace 'llvm::Intrinsic'
    {INTRINSIC_128B(vzb), u16v2, "zxt.vub", {u8v1}},
     ^~~~~~~~~~~~~~~~~~~
/ccs/home/cimes/casper/Halide-master/src/CodeGen_Hexagon.cpp:541:45: note: expanded from macro 'INTRINSIC_128B'
#define INTRINSIC_128B(id) llvm::Intrinsic::hexagon_V6_##id##_128B
                           ~~~~~~~~~~~~~~~~~^
<scratch space>:107:1: note: expanded from here
hexagon_V6_vzb_128B
^
/ccs/home/cimes/casper/Halide-master/src/CodeGen_Hexagon.cpp:545:6: error: no member named 'hexagon_V6_vzh_128B' in namespace 'llvm::Intrinsic'
    {INTRINSIC_128B(vzh), u32v2, "zxt.vuh", {u16v1}},
     ^~~~~~~~~~~~~~~~~~~
/ccs/home/cimes/casper/Halide-master/src/CodeGen_Hexagon.cpp:541:45: note: expanded from macro 'INTRINSIC_128B'
#define INTRINSIC_128B(id) llvm::Intrinsic::hexagon_V6_##id##_128B
                           ~~~~~~~~~~~~~~~~~^
<scratch space>:109:1: note: expanded from here
hexagon_V6_vzh_128B
^
/ccs/home/cimes/casper/Halide-master/src/CodeGen_Hexagon.cpp:546:6: error: no member named 'hexagon_V6_vsb_128B' in namespace 'llvm::Intrinsic'
    {INTRINSIC_128B(vsb), i16v2, "sxt.vb", {i8v1}},
[...]

Issue #5475, opened 2020-11-24
Future of OpenGL/OpenGLCompute needs deciding


steven-johnson:
Per various offline discussions (and some online, eg https://github.com/halide/Halide/pull/5474), our OpenGL backend has many limitations/problems and is arguably of dubious value even if implementation issues are fixed. We should really decide whether we want to improve it or officially deprecate/remove them.

Strawman proposal to start a discussion:
- Designate these backends as 'deprecated' as of Halide 11
- Remove them entirely in Halide 12
- Take the work-in-progress Vulkan backend and devote resources to it with the intent of getting it useful for Halide 12

shoaibkamil:
The OpenGL (non-compute) backend is extremely limited and probably not worth keeping.  However, I think OpenGLCompute is worth saving and will definitely help us get Vulkan going (since they are semantically very similar).  The major limitation of OpenGLCompute AFAIK is poor support for anything except 32-bit types; we will need to solve this for Vulkan as well.  So I propose:

- mark OpenGL deprecated in Halide 11 and remove in Halide 12
- move type legalization from D3D12 into a reusable pass that can be used by Vulkan, OpenGLCompute, etc
- get OpenGLCompute working with all GPU tests (except those that do e.g. workgroup operations)
- finish WIP Vulkan backend (@slomp and I are planning on tackling this starting late Dec/early Jan)

slomp:
On that note, I'm almost inclined in creating a general gpu back-end run-time subsystem, and there is a lot of code in the gpu back-ends that could be shared among each other.

alexreinking:
We need to decide on this. I think @shoaibkamil's proposal to deprecate in Halide 11 and remove in Halide 12 is a good one. We need to support this with loud warning messages when using the OpenGL backend.

alexreinking:
I wonder whether #5545 should influence our thinking on this. I think it would be good to get it reviewed and merged for Halide 11 but still move forward with deprecation. The current model of "let OpenGL bit rot until someone fixes it" isn't really sustainable.

steven-johnson:
I'm all in favor of landing #5545 if it passes review.

zvookin:
The issue with OpenGL is not so much bit rot but that it never worked very well in the first place. Some of this is the compiler implementation, but some of it is fundamental limitations in OpenGL itself. (As well as the practical realities of the ecosystem around it.) We can certainly improve the compiler, but if every use case has it dominated by another approach, it seems wasteful to push on this.

[...]

PR #5567, opened 2020-12-16
Move CodeGen_Hexagon to internal linkage and don't include it without WITH_HEXAGON

dsharletg:
This PR is an alternative to #5503 to address #5559. This is similar to #5548.

It moves all of `CodeGen_Hexagon` into an internal linkage namespace, and wraps the whole thing in an `#ifdef WITH_HEXAGON`.

I wanted to do this for all of the `CodeGen_*` classes, but this only works for `CodeGen_Hexagon` because it isn't used by `CodeGen_GPU_Host` as a template parameter.

I think we should move the runtime calls generated by `CodeGen_GPU_Host` to a separate lowering pass (like how HexagonOffload does it), and then we could do this for the rest of the `CodeGen_*` classes as well. I think most of the complexity there is GLSL, which might go away soon (#5475)?

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -8,6 +8,7 @@
 #include "AlignLoads.h"
 #include "CSE.h"
 #include "CodeGen_Internal.h"
+#include "CodeGen_Posix.h"
 #include "Debug.h"
 #include "HexagonOptimize.h"
 #include "IREquality.h"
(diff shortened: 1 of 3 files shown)


steven-johnson:
Is GLSL still needed for OpenGLCompute?

steven-johnson (approved):

dsharletg:
I meant the vertex buffer stuff, which I imagine wouldn't be needed for OpenGLCompute.

PR #5570, opened 2020-12-17
Fix several bugs on Hexagon and some cleanup

dsharletg:
This PR fixes bugs:
- vmpa is not correctly interleaved after refactoring in #5424. This PR reverts the changes associated with this. This *does* have test coverage in simd_op_check, but we aren't running apps/simd_op_check on the build bots currently.
- LetStmt visitor in EliminateInterleaves is incorrect. This has been a bug for a long time, it only appears if there are dead lets in the IR, which doesn't happen in the default lowering configuration (but does happen when attempting to debug by commenting off some of the passes).

It also removes VtmpyGenerator. This has been disabled for a long time, and now we have VectorReduce. Additionally, I have a WIP pass in another branch that finds vector reductions in a target independent lowering pass. If we try to bring back automatic finding of stencil vector reductions, we should build it into that pass instead of fixing VtmpyGenerator.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -547,15 +547,6 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
     debug(2) << "Lowering after optimizing shuffles:\n"
              << body << "\n\n";
 
-// Generating vtmpy before CSE and align_loads makes it easier to match
-// patterns for vtmpy.
-#if 0
-    // TODO(aankit): Re-enable this after fixing complexity issue.
-    debug(1) << "Generating vtmpy...\n";
-    body = vtmpy_generator(body);
-    debug(2) << "Lowering after generating vtmpy:\n" << body << "\n\n";
-#endif
-
     debug(1) << "Aligning loads for HVX....\n";
     body = align_loads(body, target.natural_vector_size(Int(8)));
     body = common_subexpression_elimination(body);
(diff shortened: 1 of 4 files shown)


dsharletg:
@aankit-ca @pranavb-ca please take a look.

steven-johnson:
Windows failure is unrelated

steven-johnson (approved):

aankit-ca:
@dsharletg For vmpa, I think we can simply change a01 to
Shuffle::make_interleave({mpys[0].first, mpys[1].first});
to make it work. This we way we can avoid adding the patterns in hvx_128.ll file.

aankit-ca:
Rest LGTM

dsharletg:
I think we need it, because the interleaving is different. Shuffle::make_interleave interleaves the vectors before splitting to native vector widths, while this implementation interleaves them after splitting to native vector widths. What do you think?

aankit-ca:
I'm sorry. I meant changing a01 to Shuffle::make_concat({mpys[0].first, mpys[1].first});
We will be calling make_interleave only to generate the scalar.

dsharletg:
But I think the same problem applies. By doing Shuffle::make_concat, we concatenate the vectors before slicing to native vector widths. But we need to concatenate the two operands *after* slicing to native vector widths, which is what the runtime hvx_128.ll implementation does.

aankit-ca:
Aah. Right. Yeah we will need to the concatenation in .ll

dsharletg:
Cool, thanks for confirming :) This is a tricky issue, I've had this bug before myself.

PR #5572, opened 2020-12-17
Fix build on FreeBSD/powerpc64

pkubaj:
FreeBSD doesn't use getauxval, but elf_aux_info.

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -10,7 +10,7 @@
 #include "Util.h"
 #include "WasmExecutor.h"
 
-#if defined(__powerpc__)
+#if defined(__powerpc__) && (defined(__FreeBSD__) || defined(__linux__))
 #if defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


src/Target.cpp
```diff
@@ -10,7 +10,11 @@
 #include "Util.h"
 #include "WasmExecutor.h"
 
-#if defined(__powerpc__) && defined(__linux__)
+#if defined(__powerpc__)
+#if defined(__FreeBSD__)
+#include <machine/cpu.h>
+#include <sys/elf_common.h>
+#endif
```
steven-johnson, line 17: The conditional on `defined(__linux__)` got lost here; please leave it in.

steven-johnson:
Thanks! Cloned to https://github.com/halide/Halide/pull/5575 for testing. LGTM if all the buildbots go green.

steven-johnson (approved):
Thanks for the fix!

PR #5580, opened 2020-12-21
replaced indentation in line 20 with spaces to show proper error message

### LINKED ISSUE #5579: Windows Makefile error message misleading
Hello,

I think the line 20 of the Makefile should be spaces to be consistent with the other indentation styles. I know my example show I am running MINGW and I get a further issue, but I wouldn't see that error message initially without attempting it with MINGW.

The error message I received was:
`Makefile:20: *** recipe commences before first target.  Stop.`

The error message I should see is:
`Makefile:20: *** Halide no longer supports the MinGW environment..  Stop.`

My system:
OS: Win10
terminal: MINGW

Also, why would Windows/MINGW not be supported? It's really not MINGW that is not supported, its Windows in general no? The environment variable is set by Windows, if I understand correctly.

Thanks in advance.

paulrdelg:
Hello,

This is referring to my Issue #5579 . All the change consist of is to fix that proper error message.

halide makefile

Image included was of the original cloned repo first attempting to build with Make.

Thanks!

Fixes #5579

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ MAKEFLAGS += --no-builtin-rules
 UNAME = $(shell uname)
 
 ifeq ($(OS), Windows_NT)
-    $(error Halide no longer supports the MinGW environment.)
+    $(error Halide no longer supports the MinGW environment. Please use MSVC through CMake instead.)
 else
     # let's assume "normal" UNIX such as linux
     COMMON_LD_FLAGS=$(LDFLAGS) -ldl -lpthread -lz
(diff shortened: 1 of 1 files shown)


alexreinking:
Since we're here, would you might amending the error message to read:

> Halide no longer supports the MinGW environment. Please use MSVC through CMake instead.

alexreinking (approved):

steven-johnson:
Thanks for the fix!

PR #5583, opened 2020-12-21
Use std::unique_ptr to manage CodeGen classes

dsharletg:


abadams:
I notice you haven't changed any calls to delete on these classes...

Were these only created O(1) times or something? Or were we just leaking them?

abadams:
The destination branch of this and your other PR is not master?

dsharletg:
Yeah I made a bunch of related changes without committing them. Looks like I might have messed up which chunks went into which branch... I couldn't figure out how to test this at the time, but now I realized I could switch back to each branch and test there.

steven-johnson:
std::unique_ptr isn't available until C++14; we currently only require C++11 (though C++14 is often enabled via LLVM config).

IMHO let's hold off on this change until the Halide 11 release branch is cut; for Halide 12 we've already agreed to require C++17 as baseline.

abadams:
I think it's just make_unique that was new in c++14. I've been avoiding it for this reason.

steven-johnson:
Ack, you're right, brain lesion, my bad. LGTM

pranavb-ca:
> std::unique_ptr isn't available until C++14; we currently only require C++11 (though C++14 is often enabled via LLVM config).
> 
> IMHO let's hold off on this change until the Halide 11 release branch is cut; for Halide 12 we've already agreed to require C++17 as baseline.

Sorry, I think i missed this. c++17 as the baseline for Halide/src correct? Or also for the runtime?

steven-johnson:
> Sorry, I think i missed this. c++17 as the baseline for Halide/src correct? Or also for the runtime?

Definitely for src/. Not clear (yet) whether or not attempting to use it for the runtime is possible or practical. (We currently build the runtime with `-std=c++11 -fno-threadsafe-statics` so it may be fine, but we'd have to try and see.)

src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -1373,8 +1373,8 @@ std::string CodeGen_D3D12Compute_Dev::print_gpu_name(const std::string &name) {
 
 }  // namespace
 
-CodeGen_GPU_Dev *new_CodeGen_D3D12Compute_Dev(const Target &target) {
-    return new CodeGen_D3D12Compute_Dev(target);
+std::unique_ptr<CodeGen_GPU_Dev> new_CodeGen_D3D12Compute_Dev(const Target &target) {
```
steven-johnson, line 1376: This appears to build OK everywhere, which is surprising, but maybe we are effectively already using C++14 via the LLVM settings (at least in src/)? In that case, sure

src/CodeGen_GPU_Host.cpp
```diff
@@ -132,21 +132,14 @@ CodeGen_GPU_Host<CodeGen_CPU>::CodeGen_GPU_Host(Target target)
...
-
 template<typename CodeGen_CPU>
 void CodeGen_GPU_Host<CodeGen_CPU>::compile_func(const LoweredFunc &f,
                                                  const std::string &simple_name,
                                                  const std::string &extern_name) {
     function_name = simple_name;
 
     // Create a new module for all of the kernels we find in this function.
-    for (pair<const DeviceAPI, CodeGen_GPU_Dev *> &i : cgdev) {
+    for (pair<const DeviceAPI, std::unique_ptr<CodeGen_GPU_Dev>> &i : cgdev) {
```
steven-johnson, line 142: Thie sure feels like the time for `for (auto &i : cgdev)`

src/CodeGen_GPU_Host.cpp
```diff
@@ -176,9 +169,9 @@ void CodeGen_GPU_Host<CodeGen_CPU>::compile_func(const LoweredFunc &f,
     // Fill out the init kernels block
     builder->SetInsertPoint(init_kernels_bb);
 
-    for (pair<const DeviceAPI, CodeGen_GPU_Dev *> &i : cgdev) {
+    for (pair<const DeviceAPI, std::unique_ptr<CodeGen_GPU_Dev>> &i : cgdev) {
```
steven-johnson, line 172: see above

src/CodeGen_GPU_Host.cpp
```diff
@@ -104,7 +104,7 @@ CodeGen_GPU_Host<CodeGen_CPU>::CodeGen_GPU_Host(Target target)
     // earlier ones.
     if (target.has_feature(Target::OpenGL)) {
         debug(1) << "Constructing OpenGL device codegen\n";
-        cgdev[DeviceAPI::GLSL] = new CodeGen_OpenGL_Dev(target);
+        cgdev[DeviceAPI::GLSL] = std::make_unique<CodeGen_OpenGL_Dev>(target);
```
steven-johnson, line 107: Shouldn't all of the subsequent assignments into `cgdev` in this function also use `make_unique` now too?

steven-johnson (approved):
LGTM with some nits

steven-johnson:
Ready to land, I think

src/CodeGen_GPU_Host.cpp
dsharletg: The others return std::unique_ptr now.

PR #5588, opened 2020-12-22
Add max threads checking for Metal

xndcn:
Originally, this checking will be asserted by Metal API Validation
in Xcode, otherwise the program will crash or output wrong results.

### DIFF
--- a/test/error/metal_threads_too_large.cpp
+++ b/test/error/metal_threads_too_large.cpp
@@ -16,7 +16,8 @@ int main(int argc, char **argv) {
     im.set(input);
 
     Buffer<uint16_t> output(input.width(), input.height());
-    f.realize(output);
+    Target mac_target("x86-64-osx-metal-debug");
+    f.realize(output, mac_target);
     output.copy_to_host();
 
     for (int32_t i = 0; i < output.width(); i++) {
--- a/test/error/metal_threads_too_large.cpp
+++ b/test/error/metal_threads_too_large.cpp
@@ -1,0 +1,33 @@
+#include "Halide.h"
+#include "halide_test_dirs.h"
+
+using namespace Halide;
+
+int main(int argc, char **argv) {
+    ImageParam im(UInt(16), 2, "input");
+    Func f("f");
+    Var x("x"), y("y");
+
+    f(x, y) = im(x, y) + 42;
+    f.gpu_blocks(y).gpu_threads(x, DeviceAPI::Metal);
+
+    // 65536 is larger enough than `maxTotalThreadsPerThreadgroup`
+    Buffer<uint16_t> input = lambda(x, y, cast<uint16_t>(x + y)).realize(65536, 1);
+    im.set(input);
+
+    Buffer<uint16_t> output(input.width(), input.height());
+    f.realize(output);
+    output.copy_to_host();
+
+    for (int32_t i = 0; i < output.width(); i++) {
+        for (int32_t j = 0; j < output.height(); j++) {
+            if (output(i, j) != uint16_t(i + j + 42)) {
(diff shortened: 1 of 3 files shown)


xndcn:
Xcode will do some checking like this:

steven-johnson:
Please look at (and fix) the clang-tidy errors.

xndcn:
> Please look at (and fix) the clang-tidy errors.

Thanks, fixed.

test/error/metal_threads_too_large.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+    // 65536 is larger enough than `maxTotalThreadsPerThreadgroup`
+    Buffer<uint16_t> input = lambda(x, y, cast<uint16_t>(x + y)).realize(65536, 1);
+    im.set(input);
+
+    Buffer<uint16_t> output(input.width(), input.height());
+    f.realize(output);
+    output.copy_to_host();
+
+    for (int32_t i = 0; i < output.width(); i++) {
```
shoaibkamil, line 22: Is this loop necessary if we expect an error?

shoaibkamil:
Thanks for the PR!  In principle, I'm not opposed to adding this check, though we don't do similar checks on any other backend.  It might be I'd like to move towards a standard set of checks we do in non-debug and debug targets across all the GPU runtimes.  Curious to hear what @zvookin thinks.

test/error/metal_threads_too_large.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+using namespace Halide;
+
+int main(int argc, char **argv) {
+    ImageParam im(UInt(16), 2, "input");
+    Func f("f");
+    Var x("x"), y("y");
+
+    f(x, y) = im(x, y) + 42;
+    f.gpu_blocks(y).gpu_threads(x, DeviceAPI::Metal);
```
shoaibkamil, line 12: What happens if the machine doesn't support Metal?  I think this needs a check for whether the JIT target supports Metal, otherwise it should be skipped.

steven-johnson: Yeah, even if the *machine* supports metal, we should skip if the target doesn't support it, eg 

```
    if (!get_jit_target_from_environment().has_feature(Target::Metal)) {
        printf("[SKIP] error/metal_threads_too_large ignored for targets without Metal enabled.\n");
        _halide_user_assert(0);
    }
```

xndcn:
@shoaibkamil Thanks, it's reasonable to only do this check in debug runtime,  so I have moved it.

For Cuda and OpenCL backends, they seem to do this checking in library. "cuLuanchKernel" and "clEnqueueNDRangeKernel" both have a return value to indicate this error: "cudaErrorLaunchOutOfResources" and "CL_INVALID_WORK_ITEM_SIZE".

For OpenGL Compute backend, after "glDispatchCompute" is called, "GL_INVALID_VALUE" is generated to indicate this error, too.

For D3D12Compute backend, since the threadgroups are encoded in the shader source, "D3DCompile" will failed when it exceeds the limit.

So it seems the Metal backend is an exception. Not sure what behavior the Vulkan is.

test/error/metal_threads_too_large.cpp
xndcn: Thank you, since now it only checks in debug runtime, so I add a specific target, like "metal_vector_too_large" did.

test/error/metal_threads_too_large.cpp
xndcn: It's not necessary indeed. Leave it just for code integrity...

steven-johnson (approved):
Clone is green, LGTM

steven-johnson:
@shoaibkamil Do you also want to review before this lands?

PR #5592, opened 2020-12-23
Add initial support for building a TFLite Delegate

steven-johnson:
This builds a TFLite delegate as a shared library, and modifies compare_vs_tflite (and run_compare_on_device) to run *three* ways by default:
- vs "plain" TFLite
- vs TFLite when using our Delegate
- calling our Halide code 'directly'

This is still very rough and has oodles of rough edges and TODOs for things that need improvement, but it does run mobilenetv2 correctly (on desktop and on android). Pushing this out as a PR to preserve work over vacation (and for Dillon to take a look if he likes) but no pressure to land this just yet -- it's still a WIP so it can wait to land if we like (or land now with the understanding that it's still very rough).

### DIFF
--- a/apps/interpret_nn/compare_vs_tflite.cpp
+++ b/apps/interpret_nn/compare_vs_tflite.cpp
@@ -1,10 +1,9 @@
 #include <chrono>
+#include <dlfcn.h>
 #include <fstream>
 #include <iostream>
 #include <random>
-#ifndef _MSC_VER
 #include <unistd.h>
-#endif
 
 #include "halide_benchmark.h"
 #include "interpreter/interpreter.h"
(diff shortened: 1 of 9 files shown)


apps/interpret_nn/compare_vs_tflite.cpp
```diff
@@ -77,25 +76,87 @@ Buffer<void> wrap_tf_lite_tensor_with_halide_buffer(const TfLiteTensor *t) {
...
+            int seed_here = seed++;
+            seeds[t->name()] = seed_here;
+            auto input_buf = t->data<void>();
+            dynamic_type_dispatch<FillWithRandom>(input_buf.type(), input_buf, seed_here);
+            if (verbose) {
+                std::cout << "HALIDE input " << t->name() << " inited with seed = " << seed_here << " type " << input_buf.type() << "\n";
+            }
+        }
+
+        halide_set_num_threads(threads);
```
dsharletg, line 127: I think it would be better to just assert threads == 1, we don't parallelize yet. We won't parallelize within Halide code either.

apps/interpret_nn/delegate/halide_delegate.cpp
```diff
@@ -0,0 +1,881 @@
...
+
+// TODO: this should also allow Int8 once we fix biasing for those
+constexpr int k8BitMask = 1 << kTfLiteUInt8;
+
+bool IsNodeSupported_Add(TfLiteContext *context, TfLiteNode *node, TfLiteRegistration *registration) {
+    if (!(registration->version <= 2)) {
+        return false;
+    }
+    if (!InputsHaveCorrectTypes(
+            node, context,
```
dsharletg, line 639: Super nit, but do you want these extra line breaks here? IMO when modifying code like this, it's a bit annoying if your text editor doesn't do the right thing. And this isn't a very long line.

There are other instances of this, but this one is the most stark (2 extra line breaks on a pretty short line).

apps/interpret_nn/compare_vs_tflite.cpp
steven-johnson: Not gonna assert because we also use `threads` for `TfLiteInterpreterOptionsSetNumThreads()` -- I'll remove this though and add a comment.

apps/interpret_nn/delegate/halide_delegate.cpp
steven-johnson: Nah, this is a case of me being lazy and relying on clang-format and it deciding this is ok, apparently. Fixing.

steven-johnson:
PTAL

dsharletg (approved):

PR #5595, opened 2020-12-27
Better codegen for switch-statement-like if-else chains

abadams:
And added a test that demonstrates writing a little interpreter in
Halide and scheduling it.


abadams:
Tagging @dsharletg because I think he'd be amused by the test.

dsharletg (approved):

steven-johnson:
> writing a little interpreter in Halide and scheduling it.

Now extend it to do all of (say) Java just so that we can run .jar files on GPUs

steven-johnson:
Looks like some legit failures.

abadams:
Yeah, something is going on with OpenCL that I can't repro locally on my desktop. I'll have to try a few other machines.

steven-johnson:
Yeah, this is weird -- the failure for `correctness_mul_div_mod` is only for opencl, and also only seems to fail during automated testing (logging in the the same buildbot and manually running the test under the same conditions doesn't seem to trigger the failure).

steven-johnson:
Also note that it's only the LLVM10 version that is failing; maybe there is a subtle bug in the LLVM code that only turns up there?

dsharletg:
I noticed that same failure in #5584, which is certainly unrelated there.

src/Simplify_And.cpp
```diff
@@ -57,6 +57,7 @@ Expr Simplify::visit(const And *op, ExprInfo *bounds) {
          rewrite(!x && x, false) ||
          rewrite(y <= x && x < y, false) ||
          rewrite(x != c0 && x == c1, b, c0 != c1) ||
+         rewrite(x == c0 && x == c1, false, c0 != c1) ||
```
steven-johnson, line 60: Is this strictly *necessary* or just a drive-by optimization? (Wondering if this could be landed separately to try to flush out the weird failure)

src/Simplify_Stmts.cpp
```diff
@@ -448,6 +447,25 @@ Stmt Simplify::visit(const Block *op) {
             result = Block::make(result, if_rest);
         }
         return result;
+    } else if (if_first &&
```
steven-johnson, line 450: Is this strictly *necessary* or just a drive-by optimization? (Wondering if this could be landed separately to try to flush out the weird failure)

steven-johnson:
The failure cases:
```
(a/b)*b + a%b != a; a, b = -32768, 32767; q, r = 1, 0
Compiled a/b != simplified a/b: -32768/32767 = 1 != (int16)-2
(a/b)*b + a%b != a; a, b = -23048, 28213; q, r = 1, 0
Compiled a/b != simplified a/b: -23048/28213 = 1 != (int16)-1
(a/b)*b + a%b != a; a, b = 23964, -26153; q, r = 1, 0
Compiled a/b != simplified a/b: 23964/-26153 = 1 != (int16)0
(a/b)*b + a%b != a; a, b = -17762, 30484; q, r = 1, 0
Compiled a/b != simplified a/b: -17762/30484 = 1 != (int16)-1
(a/b)*b + a%b != a; a, b = 19307, -7772; q, r = 1, 0
Compiled a/b != simplified a/b: 19307/-7772 = 1 != (int16)-2
Failure!
```

steven-johnson:
Hmm... could it be related to the gcc version somehow? All the linuxbots are running gcc 7.4

src/Simplify_Stmts.cpp
abadams: These two simplifications when combined are what turns:
```
if (x == 0) {
} 
if (x == 1) {
}
```
into
```
if (x == 0) {
} else if (x == 1) {
}
```

which is the pattern that codegen expects. 

It could be landed separately I suppose, but it's part of the same change, not a drive-by

steven-johnson:
`correctness_interpreter` is failing on windows

abadams:
Yeah, and it looks real, as far as I can tell. The thing in common to the two failures is that it doesn't use the llvm backend, where is where the changes mostly are, so it must be something about if-else chains in the C backend. I'll investigate as soon as I get time.

steven-johnson:
Huh, we appear to be completely clean now -- not sure why. Want to keep looking at this or land?

abadams:
That's... unsettling. I'd really like to land this for other reasons, so I'll land and then triage if it rears its head again.

PR #5596, opened 2020-12-28
Delete lane extraction code in vectorization

abadams:
I discovered some code where a full 3/4 of lowering time was pointless checking if scalarized versions of vector let stmts needed constructing. Now that we have robust simplification of horizontal reductions and vector slices I believe this code has become unnecessary and can be deleted.

### DIFF
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -3,7 +3,6 @@
 
 #include "CSE.h"
 #include "CodeGen_GPU_Dev.h"
-#include "Deinterleave.h"
 #include "ExprUsesVar.h"
 #include "IREquality.h"
 #include "IRMutator.h"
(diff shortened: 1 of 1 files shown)


dsharletg (commented):
It doesn't seem like simplification of shuffles is as strong as this code could be. I think we need a simplification to move shuffles (or at least extract_lane) through arithmetic/intrinsics?

abadams:
The only place the shuffles get injected is when a vector let gets used inside a scalarized block of code. vector lets only exist if you did some vector compute in a tuple (e.g. complex multiplication). An llvm shufflevector op is the right thing to emit in this case.

They're also used to extract the lanes in a tracing op, but that's not performance-sensitive.

dsharletg (approved):

vksnk (approved):

PR #5610, opened 2021-01-04
Add CLDoubles feature check for OpenCL double type

xndcn:
Similar to CLHalf feature check for half type.


xndcn:
Since OpenCL 1.2, `cl_khr_fp64` extension is not need any more to support `double` type. But we still need the builtin functions define like `sin_f64`.
If we can get the OpenCL version in codegen, it may be better to trigger the assert when we found a builtin function Call expr with double type.

steven-johnson (approved):

dsharletg:
Steven beat me to it, I was about to merge this also.

Currently, sin_f64 is actually just a macro, and I think using it without hitting this new assert might be pretty difficult, so I think the issue you mentioned might be OK for now at least.

PR #5545, opened 2020-12-10
Fix issues in OpenGL backend

xndcn:
Fix some issues in OpenGL backend.

Fixes #4885
Fixes #4937

PR #5613, opened 2021-01-05
Clone of https://github.com/halide/Halide/pull/5545 for testing

steven-johnson:


alexreinking:
Naturally error code 12288 (0x3000) is `EGL_SUCCESS`. I think it's because EGL can't run without a display (at least by default)

PR #5617, opened 2021-01-06
Don't allow CUDACapability80 on LLVM10

steven-johnson:
LLVM10 can't handle that version of Cuda; we never noticed till now because we didn't have a buildbot with a GPU that could handle it. Modify the sniffers to cap capability at 75 for LLVM10 builds, and fail with user errors if that capability is explicitly requested.


abadams (approved):

steven-johnson:
Failure is unrelated.

PR #5618, opened 2021-01-07
Add fixes to overflow analysis in bounds inference

rootjalex:
Add fixes (and tests) to bounds inference for the following cases:
- `UInt(32)` addition and multiplication overflow was not properly detected.
- Division of `type.min()` by `-1` for `Int(16)` and `Int(8)` was not detected as overflow.
- Ramps of types that cannot represent their `lanes` parameter would produce incorrect bounds (e.g. a ramp of bools)


src/Bounds.cpp
```diff
@@ -402,14 +402,18 @@ class Bounds : public IRVisitor {
             // Assume no overflow for float, int32, and int64
             if (!op->type.is_float() && (!op->type.is_int() || op->type.bits() < 32)) {
                 if (interval.has_upper_bound()) {
-                    Expr no_overflow = (cast<int>(a.max) + cast<int>(b.max) == cast<int>(interval.max));
+                    // TODO(rootjalex): Can't catch overflow of UInt(64) currently.
+                    Type t = op->type.is_uint() ? UInt(64) : Int(32);
```
steven-johnson, line 406: Why do we use 64 bits for the unsigned case but only 32 for signed?

rootjalex: Overflow has defined behavior for all unsigned types, but only for signed integers with bitwidths `<= 16`. Technically the upcasted type for unsigned integers should be `UInt(128)` so we can catch overflow for `UInt(64)` but Halide is not capable of such bitwidths at the moment

src/Bounds.cpp
```diff
@@ -576,24 +581,35 @@ class Bounds : public IRVisitor {
...
-            } else if (a.is_bounded()) {
-                // Sign of b is unknown.
-                Expr cmp = b.min > make_zero(b.min.type().element_of());
-                interval = Interval(select(cmp, e1, e2), select(cmp, e2, e1));
-            } else {
+
+            Type t = op->type.element_of();
+
+            if (op->type.is_int() && op->type.bits() <= 16 &&
+                (!a.has_lower_bound() || !can_prove(a.min != t.min())) &&
```
abadams, line 588: I would find this much easier to understand if the bounds_of_type case was the else case of this if. !can_prove(!badthing) -> general case is just harder to wrap your head around than can_prove(!badthing) -> special case

src/Bounds.cpp
```diff
@@ -1039,7 +1063,14 @@ class Bounds : public IRVisitor {
         string var_name = unique_name('t');
         Expr var = Variable::make(op->base.type().element_of(), var_name);
         Expr lane = op->base + var * op->stride;
-        ScopedBinding<Interval> p(scope, var_name, Interval(make_const(var.type(), 0), make_const(var.type(), op->lanes - 1)));
+        Expr min_value = make_const(var.type(), 0);
+        Expr max_value = make_const(var.type(), op->lanes - 1);
+        if (!var.type().can_represent(static_cast<int64_t>(op->lanes - 1))) {
```
abadams, line 1068: We use C-style casts, not static_cast

src/Bounds.cpp
```diff
@@ -621,13 +637,21 @@ class Bounds : public IRVisitor {
...
             } else {
-                // Divisor is either strictly positive or strictly
-                // negative, so we can just take the extrema.
-                interval = Interval::nothing();
-                interval.include(a.min / b.min);
-                interval.include(a.max / b.min);
-                interval.include(a.min / b.max);
-                interval.include(a.max / b.max);
+                Type t = op->type.element_of();
+                if (t.is_int() && t.bits() <= 16 &&
```
abadams, line 641: Same comment applies here

rootjalex: fixed in b66a82d

rootjalex:
@abadams Fixed your comments and also added `can_overflow` and `can_overflow_int` functions to `Type`, I imagine it'll be easier to change things if overflow behavior for different types is changed.

src/Bounds.cpp
steven-johnson: OK, that makes sense, but what is the eventual TODO action here? (i.e., it implies it's something we can and should fix; what does that fix look like, does it need a tracking issue, etc?)

rootjalex: There are some ways to catch overflow for the highest bitwidth, despite not being capable of representing that type. The TODO is to remind me to test out some of them - but perhaps it's not reasonable, as I expect (and perhaps @abadams can confirm/deny) that overflow with `UInt(64)` is uncommon enough that bounds inference can likely not worry about this case.

I can open an issue to track this, if deemed worth it!

steven-johnson:
> I can open an issue to track this, if deemed worth it!

I'd argue that, as a rule of thumb, all TODOs that are checked in to the codebase should generally have a tracking issue (referenced in the comment). Not always, but usually.

rootjalex:
Sorry for the delay @steven-johnson . I created #5682 for this problem, and updated the TODOs in 87383fb7c .

src/Bounds.cpp
rootjalex: Added in 87383fb7c

steven-johnson (approved):
LGTM with nits.

src/Bounds.cpp
```diff
@@ -576,23 +591,32 @@ class Bounds : public IRVisitor {
...
-                if (e1.same_as(Interval::neg_inf())) {
-                    e1 = Interval::pos_inf();
-                }
-                if (e2.same_as(Interval::pos_inf())) {
-                    e2 = Interval::neg_inf();
+
+            Type t = op->type.element_of();
+
+            if (div_cant_overflow(a, b, t)) {
+                // TODO: handle real numbers with can_prove(b.min > 0) and can_prove(b.min < 0) as well - treating floating point as
```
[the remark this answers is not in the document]
rootjalex: This TODO was actually just moved, it already existed before this PR. I would probably say yes, but will defer to @abadams

src/Bounds.cpp
```diff
@@ -536,6 +541,16 @@ class Bounds : public IRVisitor {
         }
     }
 
+    bool div_cant_overflow(const Interval &a, const Interval &b, Type t) {
```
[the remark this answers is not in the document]
rootjalex: fixed in 59441f1

src/Bounds.cpp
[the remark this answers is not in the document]
abadams: nah, it's just a hint for a possible approach to future readers of this code if they're trying to extend it

steven-johnson:
@abadams is this ready to land?

src/Type.h
```diff
@@ -425,6 +425,18 @@ struct Type {
         return code() == Handle;
     }
 
+    // Returns true iff type is an integral type where overflow is defined.
```
abadams, line 428: change comment from "an integral" to "a signed integral"

src/Type.h
```diff
@@ -425,6 +425,18 @@ struct Type {
...
+    HALIDE_ALWAYS_INLINE
+    bool can_overflow_int() const {
+        return is_int() && bits() <= 16;
+    }
+
+    // Returns true iff type does have a well defined overflow behavior.
```
abadams, line 434: nit: well-defined because well applies to defined, not to behavior.

abadams (approved):
lgtm with two comment nits

src/Type.h
rootjalex: fixed in 4f0a437

src/Type.h
rootjalex: fixed in 4f0a437

PR #5626, opened 2021-01-09
Remove OpenGL support (part 1)

steven-johnson:
Fixes #5475

This removes the OpenGL backend (but *not* the OpenGLCompute backend) from public use:

- Remove Target::OpenGL
- remove DeviceAPI::GLSL
- remove Func::glsl() and Func::shader()
- remove all OpenGL-specific apps and tests
- remove HalideRuntimeOpenGL.h
- remove some internal code that is OpenGL-only

Note that there is still internal code that needs trimming; since the OpenGLCompute backend uses some of the same code, and some of the same build deps, and some of the same runtime shared-library loading, I tried to err on the side of leaving code/buildrules/etc in place for now, with the plan to clean that up in subsequent PRs.

Note also that feature Target::EGL is still present, as I believe it is still useful in conjunction with OpenGLCompute.

PR #5635, opened 2021-01-12
Remove CodeGen_OpenGL_Dev

steven-johnson:
#5626 removed OpenGL support, but didn't remove this no-longer-needed class; removed it here. Moved CodeGen_GLSLBase into CodGen_OpenGLCompute_Dev.cpp (which is now the only subclass), but didn't yet attempt to consolidate them into a single class (not sure if any of that logic might be used/useful in the not-yet-done Vulkan backend).


shoaibkamil (approved):

dsharletg (approved):

steven-johnson:
Failures are unrelated.

PR #5639, opened 2021-01-13
Loosen tolerance for hypot() in vector_math

steven-johnson:
Existing tolerance is too tight for the AMD Ryzen 9 buildbots.

### DIFF
--- a/test/correctness/vector_math.cpp
+++ b/test/correctness/vector_math.cpp
@@ -154,8 +154,8 @@ struct with_unsigned<int64_t> {
 
 template<typename A>
 bool test(int lanes, int seed) {
-    const int W = 16;
-    const int H = 1;
+    const int W = 320;
+    const int H = 16;
 
     const int verbose = true;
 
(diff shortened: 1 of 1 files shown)


zvookin:
Why is this not relative error? If I read the code correctly, the max value is something like ```hypot(1.1f, 640f)``` thus the max hypot result is ~640. I'm okish with .01 error on 640, I'm not ok with .01 error on 1.

dsharletg:
My 2 cents: we should go back to commit d76e215dfede79cad9ff11974bc28e2988b11c09, and just change the close_enough check to be something like `fabs(x - y) < 1e-4 * std::max(x, y) + 1e-6` (not sure exactly what the numbers should be, guessing based on the range of inputs and the constant tolerance).

zvookin:
I suppose the slick way to do it is to use autodiff to propagate a delta in the arguments into a delta in the result. Not asking for that in this PR, but it does seem a promising tool for validating results in something like this.

steven-johnson:
So, interestingly, this doesn't appear to be Intel vs AMD after all; when I run on my local Intel box, but with the same 'natural' host target as the Ryzen bots (`x86-64-linux-avx-avx2-f16c-fma-sse41`), I get results that are ~identical across both machines. Since it only seems to be happening for LLVM10, and only for certain vector types (only uint16 so far), this smells more like a flake that we just never noticed because we didn't have any AVX2 buildbots before now.

steven-johnson:
> If I read the code correctly, the max value is something like `hypot(1.1f, 640f)` thus the max hypot result is ~640.

The max input value is 63.375, not 640. (Still pretty large.)

steven-johnson:
Withdrawn in favor of https://github.com/halide/Halide/pull/5643

PR #5644, opened 2021-01-14
Upgrade pybind11 to 2.6.x

alexreinking:
Users who have 2.6.0 installed system-wide run into an issue with a hack with a workaround etc. etc. in pybind11 that ultimately means it chokes on the `SYSTEM` argument to `pybind11_add_module` on v2.6 but not on v2.5.

Since we're already diverged from Ubuntu, we might as well upgrade to pybind11 2.6, which also happens to fix a bug that lets us remove one of our own awkward workarounds.

### DIFF
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -4,15 +4,7 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-# Adding PyBind11 to the build is the recommended way of integrating with CMake.
-# See: https://pybind11.readthedocs.io/en/stable/compiling.html
-
-# Configure pybind11 to use the same interpreter version as was detected above.
-message(STATUS "Directing pybind11 to Python3 executable ${Python3_EXECUTABLE}")
-set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
-
-# Keep the version in sync with requirements.txt and the Ubuntu 20.04 LTS package (python3-pybind11)
-set(PYBIND11_VER 2.5.0)
+set(PYBIND11_VER 2.6.0)
 find_package(pybind11 ${PYBIND11_VER} QUIET)
 if (NOT pybind11_FOUND)
     include(FetchContent)
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-set(PYBIND11_VER 2.6.0)
+set(PYBIND11_VER 2.6.1)
 find_package(pybind11 ${PYBIND11_VER} QUIET)
 if (NOT pybind11_FOUND)
     include(FetchContent)
--- a/python_bindings/readme.md
+++ b/python_bindings/readme.md
@@ -45,19 +45,24 @@ with some differences where the C++ idiom is either inappropriate or impossible:
 ## Prerequisites
 
 The bindings (and demonstration applications) should work well for Python 3.4
-(or higher), on Linux and OSX platforms. Windows is not yet supported, but could
-be with CMake work. (We have dropped support for Python 2.x and will not accept
-patches to re-enable it.)
+(or higher), on Linux and OSX platforms. Windows support is experimental, and
+available through the CMake build.
 
 #### Python requirements:
 
-See requirements.txt (to be used with `pip`:
-`pip install --user requirements.txt`)
+The best way to get set up is to use a virtual environment:
+
+```console
+$ python3 -m venv venv
+$ . venv/bin/activate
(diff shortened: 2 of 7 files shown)


steven-johnson:
LGTM but will wait to see how the buildbots like it :-)

steven-johnson:
Producing lots of warnings on (at least) the Mac build (https://buildbot.halide-lang.org/master/#/builders/49/builds/123/steps/12/logs/stdio) -- would be good to clean these up.

alexreinking:
> 
> 
> Producing lots of warnings on (at least) the Mac build (https://buildbot.halide-lang.org/master/#/builders/49/builds/123/steps/12/logs/stdio) -- would be good to clean these up.

That's a make-builder. No idea how this change could affect it.

steven-johnson:
(1) there are warnings in the Linux build too (at least on my system)
(2) ack, how to Makefile builds use PyBind11 then? Is it just expected to be present on the system? If so, do we need to manually upgrade the installs on the buildbots?

alexreinking:
> (1) there are warnings in the Linux build too (at least on my system)

The only [warnings I see on `x86-64-linux-testbranch-12-cmake` ](https://buildbot.halide-lang.org/master/#/builders/30/builds/161/steps/7/logs/stdio) are the following:

```
[1874/3367] Building CXX object python_bindings/CMakeFiles/Halide_PyStubs.dir/stub/PyStubImpl.cpp.o
/home/halidenightly/build_bot/worker/x86-64-linux-testbranch-12-cmake/halide-source/python_bindings/stub/PyStubImpl.cpp: In function ‘PyObject* _halide_pystub_impl(const char*, FactoryFunc)’:
/home/halidenightly/build_bot/worker/x86-64-linux-testbranch-12-cmake/halide-source/python_bindings/stub/PyStubImpl.cpp:200:42: warning: ‘pybind11::module_::module_(const char*, const char*)’ is deprecated: Use PYBIND11_MODULE or module_::create_extension_module instead [-Wdeprecated-declarations]
     auto m = pybind11::module(module_name);
                                          ^
In file included from /home/halidenightly/build_bot/worker/x86-64-linux-testbranch-12-cmake/halide-source/python_bindings/stub/PyStubImpl.cpp:6:0:
_deps/pybind11-src/include/pybind11/pybind11.h:897:14: note: declared here
     explicit module_(const char *name, const char *doc = nullptr) {
              ^~~~~~~
```

which, yeah, I should fix.

> (2) ack, how to Makefile builds use PyBind11 then? Is it just expected to be present on the system? If so, do we need to manually upgrade the installs on the buildbots?

Yes. I think they look for a system-wide installation. Ideally, we'd put this in the requirements.txt (also TODO, I guess) and have the buildbot create fresh virtualenvs as needed.

steven-johnson:
Suggestion: add something like `static_assert(PYBIND11_VERSION == 2.6);` somewhere in our PyBind code, so that we can double-verify that we are always building with the version we expect

steven-johnson:
Also need to update requirements.txt:

`pybind11==2.5.0`

steven-johnson:
(nm, I'll push the requirements and version-check myself)

alexreinking:
Can we do `>=` for the minor version?

steven-johnson:
> Can we do `>=` for the minor version?

We *could*, but I'd rather not be surprised by a newer version breaking us somehow. How about we change it if/when it becomes a problem?

steven-johnson:
I updated the buildbots (manually) via `pip3 install pybind11==2.6.0`

alexreinking:
Hrm, I'm generally against inserting hard blocks against newer versions. We're certainly not going to upgrade the buildbots unexpectedly. And now if someone wants to compile Halide against a newer version, they'll have to modify our sources. That's not so nice. Having a min-version check makes sense since it won't compile anyway, but at least this improves the error message.

steven-johnson:
Looks like the extra warnings on OSX+Make are because that defaults to `-Wrange-loop-analysis` enabled and PyBind11 itself is triggering it. I'll disable the warning.

steven-johnson:
> Hrm, I'm generally against inserting hard blocks against newer versions. We're certainly not going to upgrade the buildbots unexpectedly. And now if someone wants to compile Halide against a newer version, they'll have to modify our sources. That's not so nice. Having a min-version check makes sense since it won't compile anyway, but at least this improves the error message.

meh, ok

steven-johnson:
Looking at the PyBind11 repo, it looks like 2.6.1 is the latest. Any reason we shouldn't upgrade to that?

alexreinking:
> 
> 
> Looking at the PyBind11 repo, it looks like 2.6.1 is the latest. Any reason we shouldn't upgrade to that?

No reason I know of

alexreinking:
`correctness_mul_div_mod` failure is unrelated.

steven-johnson (approved):

PR #5647, opened 2021-01-15
Shrink tile size to fit in Mac Mini GPU memory.

alexreinking:
As discussed on Gitter, this schedule uses tile sizes that are too large for some macs. This change shrinks the tile size to 16x16 instead of 32x32 for the Metal API.

### DIFF
--- a/apps/unsharp/unsharp_generator.cpp
+++ b/apps/unsharp/unsharp_generator.cpp
@@ -62,6 +62,10 @@ public:
 
         // Schedule
         if (!auto_schedule) {
+            // Mac Minis have Intel GPUs that don't have enough memory to handle 32x32
+            // tiles. Use a smaller size here to avoid unexpected crashes.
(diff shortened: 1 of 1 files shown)


apps/unsharp/unsharp_generator.cpp
```diff
@@ -62,14 +62,18 @@ class Unsharp : public Halide::Generator<Unsharp> {
 
         // Schedule
         if (!auto_schedule) {
+            // Mac Minis have Intel GPUs that don't have enough memory to handle 32x32
```
shoaibkamil, line 65: Nit: "Intel Mac Minis have GPUs that require tile sizes smaller than 32x32 for this pipeline" or similar.  It's # of registers, not memory that's the issue here.

alexreinking: Fixed.

steven-johnson (approved):

shoaibkamil (approved):

steven-johnson:
failure is unrelated.

PR #5652, opened 2021-01-19
Fix deprecation warning for PyBind 2.6.1 in PyStubImpl.cpp

steven-johnson:


steven-johnson:
Huh, looks like the fix was already there, nm

PR #5659, opened 2021-01-21
Use linker tools on OSX & Linux to limit exports (#4651)

steven-johnson:
As a way to address https://github.com/halide/Halide/issues/4651, but as a hopefully-simpler alternative to https://github.com/halide/Halide/pull/5655, this adds linker-specific scripts for Linux and OSX builds that exports only the `Halide::` and `halide_` symbols in our shared libraries, hiding everything else (most notable, everything LLVM).

It doesn't attempt to do the same for MSVC builds (yet). Even without that, though, this is arguably strictly better than what we had before.

Marking as "draft" for now because the CMake scripting is awful but I didn't want to wait to figure out the right CMake syntax before trying this out. And also, some ugly TODOs and stuff.


src/CMakeLists.txt
```diff
@@ -370,6 +370,18 @@ add_library(Halide
             $<TARGET_OBJECTS:Halide_initmod>)
 add_library(Halide::Halide ALIAS Halide)
 
+# TODO: there is probably a more elegant way to do these conditionals in CMake
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
alexreinking, line 374: This should be `AppleClang` because clang might still use the GNU linker on Linux.

src/CMakeLists.txt
```diff
@@ -370,6 +370,18 @@ add_library(Halide
...
+    # using Clang
+    target_link_options(Halide PRIVATE -Wl,-exported_symbols_list ${CMAKE_CURRENT_LIST_DIR}/exported_symbols.osx)
```
alexreinking, line 376: Should be `"LINKER:-exported_symbols_list,${CMAKE_CURRENT_LIST_DIR}/exported_symbols.osx"`

src/CMakeLists.txt
```diff
@@ -370,6 +370,18 @@ add_library(Halide
...
+elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # using GCC
+    target_link_options(Halide PRIVATE -Wl,--version-script=${CMAKE_CURRENT_LIST_DIR}/exported_symbols.linux)
```
alexreinking, line 379: Should be `"LINKER:--version-script=${CMAKE_CURRENT_LIST_DIR}/exported_symbols.linux"`

alexreinking (changes requested):

steven-johnson:
Feel free to push your corrections directly to this PR

alexreinking:
> Feel free to push your corrections directly to this PR

That's the plan 🙂

abadams:
The linker scripts are much less arcane than I feared

alexreinking:
Okay, I have tested my CMake changes on Linux with GCC and the following linkers: GNU LD, GOLD (via `-fuse-ld=gold`), and LLD (via `-fuse-ld=lld`). _Changing_ the linker is a toolchain-level change and so requires a clean rebuild to re-run the check.

Makefile
```diff
@@ -921,9 +921,15 @@ else
 LIBHALIDE_SONAME_FLAGS=
 endif
 
+ifeq ($(UNAME), Linux)
+LIBHALIDE_EXPORTS=-Wl,--version-script=$(ROOT_DIR)/src/exported_symbols.linux
+else
+LIBHALIDE_EXPORTS=-Wl,-exported_symbols_list $(ROOT_DIR)/src/exported_symbols.osx
```
alexreinking, line 927: Should this not be `-Wl,$(ROOT_DIR)/src/exported_symbols.osx`?

steven-johnson: What's there is what works locally for me. I'll retest this morning to be sure I'm not hallucinating.

src/CMakeLists.txt
```diff
@@ -377,6 +377,11 @@ target_compile_definitions(Halide
                            $<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,STATIC_LIBRARY>:Halide_STATIC_DEFINE>
                            $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:WITH_INTROSPECTION>)
 
+include(TargetExportScript)
+target_export_script(Halide
+                     APPLE_LD "${CMAKE_CURRENT_LIST_DIR}/exported_symbols.osx"
+                     GNU_LD "${CMAKE_CURRENT_LIST_DIR}/exported_symbols.linux")
+
```
steven-johnson, line 384: Redundant, but let's include

```
    ## TODO: implement something similar for Windows/link.exe
    # https://github.com/halide/Halide/issues/4651
```

here too, so that a reader doesn't wonder if special casing for windows was omitted accidentally or not

Makefile
steven-johnson: > What's there is what works locally for me. I'll retest this morning to be sure I'm not hallucinating.

Verified, working properly for me with my local XCode 12.x setup

alexreinking: Huh, okay. I guess the driver is smart enough to pass that on automatically.

src/CMakeLists.txt
alexreinking: Looks like you already did that 🙂

steven-johnson:
Looks good! Just needs reviewer approval.

alexreinking (approved):

abadams (approved):

LebedevRI:
The caveat with this approach alone as opposed to #5655 is that while it resolves the issue for the final library,
during the compile time however, the functions aren't known to be hidden, which _may_ be suboptimal.

alexreinking:
@LebedevRI - IIRC the issue was that we consume standard library types through our API which makes the situation on Windows a lot harder. For instance, I don't think you can export a class that takes in a `std::vector`.

LebedevRI:
> @LebedevRI - IIRC the issue was that we consume standard library types through our API which makes the situation on Windows a lot harder. For instance, I don't think you can export a class that takes in a `std::vector`.

I don't know the windows situation, but i would like to note that even with the current linker script approach,
it is not done for windows, only *nix-like. So i'm a little unsure why an inability to do so on one platform
is a blocker for doing so for the other platforms?

PR #5662, opened 2021-01-22
Lower saturating arithmetic without widening

dsharletg:
Also, we weren't handling it in lower_intrinsic (because CodeGen_LLVM has a good default handler for it, and we don't have coverage of saturating arithmetic in CodeGen_C).

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -632,7 +632,7 @@ Expr lower_saturating_sub(const Expr &a, const Expr &b) {
     internal_assert(a.type() == b.type());
     // Lower saturating add without using widening arithmetic, which may require
     // types that aren't supported.
-    return simplify(clamp(a, a.type().min() - max(b, 0), a.type().max() - min(b, 0))) - b;
+    return simplify(clamp(a, a.type().min() + max(b, 0), a.type().max() + min(b, 0))) - b;
 }
 
 Expr lower_halving_add(const Expr &a, const Expr &b) {
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -623,12 +623,16 @@ Expr lower_rounding_shift_right(const Expr &a, const Expr &b) {
 
 Expr lower_saturating_add(const Expr &a, const Expr &b) {
     internal_assert(a.type() == b.type());
-    return saturating_narrow(widening_add(a, b));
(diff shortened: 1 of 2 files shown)


test/correctness/intrinsics.cpp
```diff
@@ -26,6 +26,41 @@ void check(Expr test, Expr expected) {
...
+        for (int j = 0; j < N; j++) {
+            int64_t b = min_t + ((max_t - min_t) * j) / N;
+            Expr add = saturating_add(make_const(halide_t, a), make_const(halide_t, b));
+            Expr sub = saturating_sub(make_const(halide_t, a), make_const(halide_t, b));
+            Expr result_add = simplify(lower_intrinsic(add.as<Call>()));
+            Expr result_sub = simplify(lower_intrinsic(sub.as<Call>()));
+            int64_t reference_add = std::min(std::max(a + b, min_t), max_t);
+            int64_t reference_sub = std::min(std::max(a - b, min_t), max_t);
+            if (!can_prove(result_add == make_const(halide_t, reference_add))) {
+                std::cout << "failure!\n";
```
steven-johnson, line 47: nit: shouldn't these be `std::cerr`?

steven-johnson (approved):

src/FindIntrinsics.cpp
```diff
@@ -623,12 +623,16 @@ Expr lower_rounding_shift_right(const Expr &a, const Expr &b) {
...
+    // types that aren't supported.
+    return simplify(clamp(a, a.type().min() - min(b, 0), a.type().max() - max(b, 0))) + b;
 }
 
 Expr lower_saturating_sub(const Expr &a, const Expr &b) {
     internal_assert(a.type() == b.type());
-    return saturating_cast(a.type(), widening_sub(a, b));
+    // Lower saturating add without using widening arithmetic, which may require
+    // types that aren't supported.
+    return simplify(clamp(a, a.type().min() - max(b, 0), a.type().max() - min(b, 0))) - b;
```
dsharletg, line 635: Hmm, looking at this, I'm not sure how it works. It seems like these limits should overflow. But the test should be solid here...

dsharletg: I think this is some accidental and magical 2's complement overflow trick... anyways fixed.

dsharletg:
I think all of the failures are unrelated, merging.

PR #5667, opened 2021-01-25
FIx intermittent OSX Python crash

steven-johnson:
The OSX buildbot has been crashing intermittently on some python tests; debugging showed that in some situations, Introspection's calls to `backtrace()` include bogus addresses (eg 0x08), which cause segfaults when you try to inspect memory near them. The reasons for this aren't entirely clear -- for instance, it only seems to repeat reliably when using the Makefile rather than CMake, and only when doing an 'out-of-tree' build. Rather than try to run this to ground further, this PR just checks for address fields that seem obviously unreasonable (first 256 bytes of address space) and ignore them.

### DIFF
--- a/python_bindings/Makefile
+++ b/python_bindings/Makefile
@@ -38,8 +38,10 @@ OPTIMIZE ?= -O3
 # defining DEBUG + undefining NDEBUG gives extra debug info in PyBind11
 # OPTIMIZE ?= -g -DDEBUG=1 -UNDEBUG
 
-# Compiling with -fvisibility=hidden saves ~80k on optimized x64 builds
-CCFLAGS=$(shell $(PYTHON)-config --cflags) $(PYBIND11_CFLAGS) -I $(HALIDE_DISTRIB_PATH)/include -I $(ROOT_DIR) -std=c++11 $(FPIC) -fvisibility=hidden -fvisibility-inlines-hidden $(OPTIMIZE) $(CXXFLAGS)
+# Compiling with -fvisibility=hidden saves ~80k on optimized x64 builds.
+# It's critical to include -fno-omit-frame-pointer, otherwise introspection can
+# break in amusing ways.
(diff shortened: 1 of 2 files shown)


abadams:
I'm trying to figure out how this could happen. A bad stacktrace seems to imply that frame pointers were omitted, but the canary code was supposed to check that. Could it be that the canary code isn't getting called from within python? It's a static constructor injected into user code via Halide.h. Do the python bindings even use Halide.h? Or maybe the python module is compiled with different frame pointer settings to python itself?

We should probably not define WITH_INTROSPECTION when building the python module - introspection isn't going to do anything useful out of python anyway.

steven-johnson:
This isn't even from python code, per se; it's from running Generator(s) that the Python tests use, so we are failing at build time. 

Random guess: on OSX, maybe `backtrace()` needs sidecar information from the .dSYM files; when doing out-of-tree builds, it can't find it for, uh, something, and freaks out in this way.

abadams:
The man page for backtrace says that it breaks when you omit the frame pointer, but should work otherwise. I don't think it needs debug symbols at all. If we have our omit-frame-pointer flags in an inconsistent state between libHalide and the generator binary we're going to get much-harder-to-diagnose crashes than this one, so I'm very hesitant to just slap a band-aid on it.

steven-johnson:
> omit-frame-pointer flags in an inconsistent state

Yeah, that's likely it: the main Makefile uses `-fno-omit-frame-pointer` but the python_bindings Makefile doesn't specify it either way. Adding it to the python bindings seems sensible. (Keeping the newly-added check seems sensible as well, TBH.)

This level of fragility makes my spidey-sense tingle, but enabling introspection by default has been here forever, so I guess we'll keep patching it :-)

abadams (approved):

steven-johnson:
Failure is unrelated.

PR #5671, opened 2021-01-27
Allow Tensor to use an external memory allocation

steven-johnson:
For the TFLite Delegate, Tensors which are inputs or outputs don't need their own allocation (TFLite will do that for us); we should just point at the shared memory, which will save memory allocation, plus the time to copy between TFLite and our memory.

Note that the 'read-only' inputs work basically the same as before, but for normal inputs and outputs, we must update the pointers in the Eval() method of the delegate, as they aren't guaranteed to be valid prior to then. (This is still substantially cheaper than copying the memory entirely.)

I've left the code in the delegate marked with USE_EXTERNAL_TENSORS for now, to make the change a bit clearer. It's not intended to be a long-term flag.

From local benchmarking on my Mac laptop, I don't see a meaningful performance difference for mobilenet_v2. (I haven't measured memory usage but it definitely will be lower.)

### DIFF
--- a/apps/interpret_nn/delegate/halide_delegate.cpp
+++ b/apps/interpret_nn/delegate/halide_delegate.cpp
@@ -11,6 +11,8 @@
 #include "tensorflow/lite/c/c_api.h"
 #include "util/error_util.h"
 
+#define USE_EXTERNAL_TENSORS 1
+
 // TODO: this is likely a worthwhile optimization that we can support without
 // too much effort, but requires some testing harnesses we don't have yet
 // and isn't likely to be our lowest-hanging fruit. Revisit once other optimizations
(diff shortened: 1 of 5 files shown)


steven-johnson:
(ping to @alexreinking -- looks like PRs that are branches-of-branches are no longer triggering builds, one of us needs to take a look)

steven-johnson:
ping to @dsharlet

dsharletg:
Is this needed for correctness or is it an optimization? I think this code will need significant refactoring, and I'm a bit hesitant about the compile time flag adding more code paths to keep alive. Why is it needed?

steven-johnson:
100% an optimization. It avoids redundant memory allocations and unnecessary memory copies.

steven-johnson:
(Converting to draft because we're not ready to land this yet, but want to keep this around to eventually land.)

steven-johnson:
Closing (but keeping the branch alive for reference) as the code has diverged a lot since this was written.

PR #5673, opened 2021-01-27
Remove deprecated variants of infer_input_bounds() in the Python bindings

steven-johnson:
The C++ versions were removed for Halide 12 already, but I missed the Python wrappers.


abadams (approved):

steven-johnson:
Failure is unrelated.

PR #5677, opened 2021-01-28
Add initial support for Sapphire Rapids AVX512 features

jwlawson:
Add an AVX512 target feature for Sapphire Rapids CPUs to make use of the new instructions available. To start with just support the vcvtneps2bf16 conversion instruction, as this is a fairly straight forward proof of concept and allows us to at least test the target feature.

Moving forward we intend to add the VNNI and other BF16 AVX512 instructions that are available in LLVM 12.


src/CodeGen_X86.cpp
```diff
@@ -132,6 +132,12 @@ const x86Intrinsic intrinsic_defs[] = {
     {"llvm.x86.avx512.pmaddw.d.512", Int(32, 16), "pmaddwd", {Int(16, 32), Int(16, 32)}, Target::AVX512_Cannonlake},
     {"llvm.x86.avx2.pmadd.wd", Int(32, 8), "pmaddwd", {Int(16, 16), Int(16, 16)}, Target::AVX2},
     {"llvm.x86.sse2.pmadd.wd", Int(32, 4), "pmaddwd", {Int(16, 8), Int(16, 8)}},
+
+    // Convert FP32 to BF16
+    {"llvm.x86.avx512bf16.cvtneps2bf16.512", BFloat(16, 16), "f32_to_bf16", {Float(32, 16)}, Target::AVX512_SapphireRapids},
+    {"llvm.x86.avx512bf16.cvtneps2bf16.256", BFloat(16, 8), "f32_to_bf16", {Float(32, 8)}, Target::AVX512_SapphireRapids},
+    // LLVM 12 does not support cvtneps2bf16 for 128bit inputs
```
steven-johnson, line 139: LLVM 12 just forked yesterday (LLVM main is now v13); do these changes apply to both

src/CodeGen_X86.cpp
```diff
@@ -509,7 +517,9 @@ void CodeGen_X86::codegen_vector_reduce(const VectorReduce *op, const Expr &init
 }
 
 string CodeGen_X86::mcpu() const {
-    if (target.has_feature(Target::AVX512_Cannonlake)) {
+    if (target.has_feature(Target::AVX512_SapphireRapids)) {
```
steven-johnson, line 520: Do the changes in this method also need `#if LLVM_VERSION >= 120`?

src/runtime/HalideRuntime.h
```diff
@@ -1334,6 +1334,7 @@ typedef enum halide_target_feature_t {
     halide_target_feature_egl,                    ///< Force use of EGL support.
     halide_target_feature_arm_dot_prod,           ///< Enable ARMv8.2-a dotprod extension (i.e. udot and sdot instructions)
     halide_llvm_large_code_model,                 ///< Use the LLVM large code model to compile
+    halide_target_feature_avx512_sapphirerapids,  ///< Enable the AVX512 features supported by Sapphire Rapids processors. This include all of the Cannonlake features, plus AVX512-VNNI and AVX512-BF16.
```
steven-johnson, line 1337: Comment #1: this should probably just be grouped with the other avx512 features earlier in the list. (Note that Halide feature enums are *not* promised to be identical across Halide versions, so there is no need to add them at the end.) Probably ditto for grouping in the other places it appears in this PR.

Comment #2: I realize the wording of "includes all of the $FOO features" appears in the comments for the existing AVX512 features, but is it really true? i.e., does specifying "avx512_sapphirerapids" in a target string really implicitly select all the earlier features? (I don't see any code in Halide proper that seems to enforce this; does the `mcpu()` string we hand to LLVM end up enforcing this?)

src/runtime/x86_cpu_features.cpp
```diff
@@ -79,6 +81,11 @@ WEAK CpuFeatures halide_get_cpu_features() {
             }
             if ((info2[1] & avx512_cannonlake) == avx512_cannonlake) {
                 features.set_available(halide_target_feature_avx512_cannonlake);
+                if ((info2[2] & avx512vnni) == avx512vnni) {
+                    // FIXME: Should also check AVX512-BF16 here, but that needs call to cpuid(eax=7, ecx=1)
```
steven-johnson, line 85: Suggestion: create a GitHub issue for "Implement SapphireLake features" to track the work to be done, and reference it here via `// TODO: (github-issue-link)`

src/CodeGen_X86.cpp
```diff
@@ -132,6 +132,12 @@ const x86Intrinsic intrinsic_defs[] = {
     {"llvm.x86.avx512.pmaddw.d.512", Int(32, 16), "pmaddwd", {Int(16, 32), Int(16, 32)}, Target::AVX512_Cannonlake},
     {"llvm.x86.avx2.pmadd.wd", Int(32, 8), "pmaddwd", {Int(16, 16), Int(16, 16)}, Target::AVX2},
     {"llvm.x86.sse2.pmadd.wd", Int(32, 4), "pmaddwd", {Int(16, 8), Int(16, 8)}},
+
+    // Convert FP32 to BF16
+    {"llvm.x86.avx512bf16.cvtneps2bf16.512", BFloat(16, 16), "f32_to_bf16", {Float(32, 16)}, Target::AVX512_SapphireRapids},
+    {"llvm.x86.avx512bf16.cvtneps2bf16.256", BFloat(16, 8), "f32_to_bf16", {Float(32, 8)}, Target::AVX512_SapphireRapids},
+    // LLVM 12 does not support cvtneps2bf16 for 128bit inputs
+    //{"llvm.x86.avx512bf16.cvtneps2bf16.128", BFloat(16, 4), "f32_to_bf16", {Float(32, 4)}, Target::AVX512_SapphireRapids},
```
steven-johnson, line 140: Suggestion: create a GitHub issue for "Implement SapphireLake features" to track the work to be done, and reference it here via `// TODO: (github-issue-link)`

src/runtime/x86_cpu_features.cpp
```diff
@@ -79,6 +81,11 @@ WEAK CpuFeatures halide_get_cpu_features() {
...
+                    // FIXME: Should this check the LLVM version? Only supported on 12+.
+                    features.set_available(halide_target_feature_avx512_sapphirerapids);
```
steven-johnson, line 87: Hrm, that's a good question. I don't think that `LLVM_VERSION` is (currently) visible to our runtime code. That probably wouldn't be hard to fix (just minor changes in a couple of build scripts).

steven-johnson (commented):
LGTM in general, just a few comments

src/runtime/x86_cpu_features.cpp
dsharletg: I don't think this is necessary. It should just be a no-op if LLVM doesn't support it.

src/CodeGen_X86.cpp
jwlawson: The intrinsic is not available in either version. The masked intrinsic is available, so we can provide a wrapper that uses that, similar to some of the other intrinsics here.

src/CodeGen_X86.cpp
jwlawson: My thinking was to restrict where the feature was set, rather than where the feature is used. I'm not sure which approach would require the least `#if`s or the best errors for users.

Currently these changes will prevent Halide automatically selecting SapphireRapids on LLVM 10 or 11, but will still cause problems if a user provides `HL_TARGET=x86-64-linux-avx512_sapphirerapids`. Would it make sense to add the `#if` to prevent parsing this string for LLVM <12?

src/runtime/HalideRuntime.h
jwlawson: For 2 that's a good point. There is some code in [CodeGen X86](https://github.com/halide/Halide/blob/3defb66236c3cd6812db90a651ff13f6f8170651/src/CodeGen_X86.cpp#L29) that partially addresses this. I should add to that to include SapphireRapids.

src/runtime/x86_cpu_features.cpp
jwlawson: To check my understanding: The code here will be run to check what CPU features are available when running already (AOT) compiled code. This cpu_features function lets the runtime choose which of the available AOT compiled functions to use, so if there is a sapphirerapids version available then that gets used, but if not then the next best target will be used. As this never leads to attempting to compile code it shouldn't matter what the version of LLVM is, only what AOT compiled functions are available.

Does that sound right?

steven-johnson: Yes.

src/CodeGen_X86.cpp
steven-johnson: I guess it depends on whether the changes you are making are entirely in Halide (in which case the LLVM version doesn't matter) or if they rely on LLVM codegen (in which case it does). Which is the case?

abadams (approved):

dsharletg:
Should this new feature be added here: https://github.com/halide/Halide/blob/66205633a5e9ec27fedb9dbf84223bd213e6f5f9/src/Target.cpp#L971 ?

steven-johnson:
> Should this new feature be added here:
> 
> https://github.com/halide/Halide/blob/66205633a5e9ec27fedb9dbf84223bd213e6f5f9/src/Target.cpp#L971
> 
> ?
> (Found by searching the repo for AVX512_CannonLake.)

Yes, probably so

src/CodeGen_X86.cpp
jwlawson: I'm not sure I understand the question. The changes should involve only changing Halide code (not LLVM code) but will rely heavily on the intrinsics being available in LLVM, which they are for 12+.

I have changed this to copy what is being done for CUDA 80. The automatic feature selection is guarded by `LLVM_VERSION`, but if a user deliberately requests it then it will raise an error in `CodeGen_X86::mcpu` and `CodeGen_X86::mattr` for LLVM < 12.

steven-johnson:
LGTM pending a green build at https://github.com/halide/Halide/pull/5687

PR #5678, opened 2021-01-28
Avoid bogus "out-of-memory" error for multiple_scatter under wasm

steven-johnson:
Several errors here:
- the `atomic()` schedule directive ends up calling `halide_mutex_array_create()`; for users of fake_thread_pool, this quietly returned null, leading to misleading error messages. Added an explicit error message if this implementation is called.
- The WebAssembly JIT cannot handle atomics yet (or threads in general) due to WABT limitations; added an assert to fail right away if you request compiling with `wasm_threads`.
- correctness_multiple_scatter uses the `atomic()` directive for one of its cases, so it now is a [SKIP] case for compiling with wasm.

### DIFF
--- a/src/runtime/fake_thread_pool.cpp
+++ b/src/runtime/fake_thread_pool.cpp
@@ -90,6 +90,7 @@ struct halide_mutex_array {
 };
 
 WEAK halide_mutex_array *halide_mutex_array_create(int sz) {
+    halide_error(nullptr, "halide_mutex_array_create not implemented on this platform.");
     return nullptr;
 }
 
--- a/src/WasmExecutor.cpp
+++ b/src/WasmExecutor.cpp
@@ -1349,6 +1349,8 @@ WasmModuleContents::WasmModuleContents(
 #if WITH_WABT
     user_assert(LLVM_VERSION >= 110) << "Using the WebAssembly JIT is only supported under LLVM 11+.";
 
+    user_assert(!target.has_feature(Target::WasmThreads)) << "The Halide WebAssembly JIT doesn't support wasm threads yet.";
+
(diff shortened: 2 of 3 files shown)


abadams (approved):

steven-johnson:
(We missed finding this on our buildbots because of some config issues that missed some tests, followed by explicit disabling of the tests due to a recent incompatibility in wasm-ld between LLVM and EMCC. Hopefully that won't recur once we re-enable them after the next EMSDK release.)

src/runtime/fake_thread_pool.cpp
```diff
@@ -90,6 +90,7 @@ struct halide_mutex_array {
 };
 
 WEAK halide_mutex_array *halide_mutex_array_create(int sz) {
+    halide_error(nullptr, "halide_mutex_array_create not implemented on this platform.");
```
dsharletg, line 93: Should this just return a non-null dummy pointer? Maybe a pointer to a static/global default constructed halide_mutex_array? I'm assuming the design is that halide_mutex_array is an opaque type outside of this file, and all of the implementation functions for it in this file just do nothing.

If so, the rest of the changes in this PR could probably be reverted.

steven-johnson: Hm, that's an interesting thought. I guess having the same pointer returned for multiple calls is fine if you assume that external code doesn't care about the value (and that the other calls are all no-ops). The only point of concern I would have is that we'd have to ensure the semantics of the non-fake-thread versions don't grow a use that is incompatible with this.

steven-johnson:
Withdrawn in favor of https://github.com/halide/Halide/pull/5681

PR #5684, opened 2021-02-01
Add ecx support for runtime X86 cpuid

jwlawson:
Some newer X86 extensions require setting ecx when calling cpuid, for
example AVX512BF16 support is queried using cpuid(eax=7, ecx=1).

### DIFF
--- a/src/runtime/x86_cpu_features.cpp
+++ b/src/runtime/x86_cpu_features.cpp
@@ -9,8 +9,9 @@ extern "C" void x86_cpuid_halide(int32_t *);
 
 namespace {
 
-ALWAYS_INLINE void cpuid(int32_t fn_id, int32_t *info) {
+ALWAYS_INLINE void cpuid(int32_t *info, int32_t fn_id, int32_t extra = 0) {
     info[0] = fn_id;
+    info[1] = extra;
     x86_cpuid_halide(info);
 }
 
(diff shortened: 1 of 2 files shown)


steven-johnson:
There is also an implementation of `cpuid` in `src/Target.cpp`; does it need similar attention?

src/runtime/x86_cpu_features.cpp
```diff
@@ -9,8 +9,9 @@ extern "C" void x86_cpuid_halide(int32_t *);
 
 namespace {
 
-ALWAYS_INLINE void cpuid(int32_t fn_id, int32_t *info) {
+ALWAYS_INLINE void cpuid(int32_t *info, int32_t fn_id, int32_t extra = 0) {
```
steven-johnson, line 12: (1) this changes the argument ordering, so now it's different from the (compile-time) implementation of cpuid in src/Target.cpp; it would be nice to keep them the same to avoid confusion for people reading the code. (I don't really care which ordering is 'correct', just that they match.)
(2) Why is `extra` an argument when it is never set to anything but the default value?

jwlawson: The changes here make the parameter order match with `cpuid` in [src/Target.cpp](https://github.com/halide/Halide/blob/f9413766bf227b05429a3ba5e97251df3a431a74/src/Target.cpp#L45). Is there another version of this function with the opposite ordering?

The `extra` parameter is not used now, but is needed to check AVX512BF16 once #5677 is merged. For that check we need to be able to call `cpuid(info, 7, 1)`.

steven-johnson: ...ack. You're right. I misread the change direction. LGTM.

steven-johnson (approved):

PR #5685, opened 2021-02-01
Delete llvm_builder.yml

steven-johnson:
We don't use the results, it's a waste of resources, and now it's breaking. We can reclaim it from git history if we need it again.


abadams (approved):

alexreinking:
btw - I wrote my own GitHub Actions LLVM builder as an exercise. At some point we might consider using these.

https://github.com/alexreinking/llvm-nightlies

PR #5686, opened 2021-02-01
Lower halving_* intrinsics without widening

vksnk:


dsharletg:
I think halving_* and rounding_halving_* are now tricky enough that we should test for correctness here, like saturating_*: https://github.com/halide/Halide/blob/master/test/correctness/intrinsics.cpp#L30

vksnk:
I've added a test, please take a look.

test/correctness/intrinsics.cpp
```diff
@@ -37,25 +37,23 @@ void check_saturating_add_sub() {
...
-                std::cerr << "expected: " << reference_add << "\n";
-                abort();
-            }
-            if (!can_prove(result_sub == make_const(halide_t, reference_sub))) {
-                std::cerr << "failure!\n";
-                std::cerr << "test: " << sub << "\n";
-                std::cerr << "result: " << result_sub << "\n";
-                std::cerr << "expected: " << reference_sub << "\n";
-                abort();
+            std::vector<std::pair<Expr, int64_t>> intrinsics_with_reference_answer = {
```
dsharletg, line 40: Probably best to just use a regular C array here. Constructing/destroying the vector here might be pretty expensive. I had to tweak N to make the test run in a reasonable amount of time.

dsharletg (approved):

vksnk:
Looks like failures are unrelated, merging in.

PR #5696, opened 2021-02-02
Capture Exprs by ref in IRMatch

abadams:
The bulk of the time in the simplifier is simplifying already-simple expressions, and the bulk of the time there turned out to be the ref counting required to construct the IRMatcher object (which took a copy of the expressions). This PR changes it to capture  expressions by reference instead, which had large effect on lowering time (more than 2x on local laplacian!). 

Before:
```
$ time bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host auto_schedule=false

real	0m1.698s
user	0m1.686s
sys	0m0.012s
```
After:
```
$ time bin/host/local_laplacian.generator -g local_laplacian -e stmt -o bin/host -f local_laplacian target=host auto_schedule=false

real	0m0.765s
user	0m0.749s
sys	0m0.016s
```
Some cleverness was required to stop people from passing rvalue Exprs to IRMatch constructors

### DIFF
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -2167,7 +2167,7 @@ struct IsConst {
         ty.code = halide_type_uint;
         ty.bits = 64;
         ty.lanes = 1;
-        val.u.u64 = is_const(e) ? 1 : 0;
+        val.u.u64 = ::Halide::Internal::is_const(e) ? 1 : 0;
     }
 };
 
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -2427,7 +2427,7 @@ struct IsMinValue {
         // a is almost certainly a very simple pattern (e.g. a wild), so just inline the make method.
         a.make_folded_const(val, ty, state);
         if (ty.code == halide_type_int) {
-            const uint64_t min_bits = (uint64_t)(1) << (ty.bits - 1);
+            const uint64_t min_bits = (uint64_t)(-1) << (ty.bits - 1);
             val.u.u64 = (val.u.u64 == min_bits);
         } else if (ty.code == halide_type_uint) {
             val.u.u64 = (val.u.u64 == 0);
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -207,23 +207,23 @@ struct SpecificExpr {
     constexpr static IRNodeType max_node_type = IRNodeType::Shuffle;
     constexpr static bool canonical = true;
 
-    Expr expr;
+    const BaseExprNode &expr;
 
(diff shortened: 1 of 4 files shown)


steven-johnson:
Note the Windows compilation failure:

https://buildbot.halide-lang.org/master/#/builders/105/builds/36/steps/10/logs/stdio

```
C:\build_bot\worker\halide-testbranch-main-llvm13-x86-32-windows-cmake\halide-source\src\IRMatch.h(2123): error C2440: '?': cannot convert from 'Halide::Internal::IRMatcher::IsConst<Halide::Internal::IRMatcher::SpecificExpr>' to 'bool'
C:\build_bot\worker\halide-testbranch-main-llvm13-x86-32-windows-cmake\halide-source\src\IRMatch.h(2123): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
C:\build_bot\worker\halide-testbranch-main-llvm13-x86-32-windows-cmake\halide-source\src\IRMatch.h(1555): note: see reference to function template instantiation 'void Halide::Internal::IRMatcher::IsConst<T>::make_folded_const<A>(halide_scalar_value_t &,halide_type_t &,Halide::Internal::IRMatcher::MatcherState &) noexcept const' being compiled

```

steven-johnson:
Conceptually this looks great. My only concern is that it's not clear to me (from a quick skim) whether it's impossible (or nearly so) to accidentally end up with a dangling reference (i.e. an IRMatch gets created with a given Expr, then the Expr gets recycled before the IRMatch does).

abadams:
You can indeed end up with a dangling reference if somehow the matcher node outlives the expr. I required that all arguments to matchers are lvalues, but you can still break it with something like:

```
Expr e = ...;
auto node = IRMatcher::add(e, e);
e = ...;
// node now has dangling refs and risks use-after-free
```

I think a 2x speed-up to lowering is worth having to be a little bit careful about how matchers are used.

abadams:
One thing I could do is change matchers to accept BaseExprNode * to make the fact that you're capturing a reference more obvious at the point of use:

```
Expr e = ...;
auto node = IRMatcher::add(e.get(), e.get());
e = ...;
// node now has dangling refs and risks use-after-free
```

It just seemed like boilerplate that didn't necessarily communicate what we were trying to communicate though.

steven-johnson (approved):
LGTM. Not sure if it's worth adding comments warning about the dangling-reference problem or not. Up to you.

steven-johnson:
Failures look unrelated.

PR #5654, opened 2021-01-20
[adams2019] Restructure autoscheduler + add timer

rootjalex:
~~Adds caching of features and schedule enumerations to the adams2019 autoscheduler. Backported from @aekul 's autoscheduler work.~~

~~On my machine, I see a x2 speedup on the time needed to autoschedule lens blur and local laplacian, about a x1.5 speedup for resnet50, and anywhere between x1-1.5 for other pipelines (caching is only useful for larger pipelines, it has little to no effect on smaller pipelines).~~
This PR is now just a restructuring + addition of a timer to the adams2019 autoscheduler, to allow for the addition of caching.

PR #5697, opened 2021-02-03
[adams2019] Add caching to autoscheduler

rootjalex:
Adds caching of features and schedule enumerations to the adams2019 autoscheduler. Backported from @aekul 's autoscheduler work.

On my machine, I see a x2 speedup on the time needed to autoschedule lens blur and local laplacian, about a x1.5 speedup for resnet50, and anywhere between x1-1.5 for other pipelines (caching is only useful for larger pipelines, it has little to no effect on smaller pipelines).

~To enable caching (it is disabled by default), these parameters should be set:
HL_USE_MEMOIZED_FEATURES=1
HL_MEMOIZE_BLOCKS=1~

~Additionally, in order to test feature caching, setting the following value will enable feature caching verification (this will be quite slow):
HL_VERIFY_MEMOIZED_FEATURES=1~

Caching schedule enumerations and features is enabled by default. To disable them, set the following environment variables:
```
HL_DISABLE_MEMOIZED_BLOCKS=1
HL_DISABLE_MEMOIZED_FEATURES=1
```

Tests were also added to verify these caching methods (Note that there likely will be no speed-up on these tests, as the pipelines are not large enough).

This PR was originally #5654 before it was split into two.


steven-johnson:
See failure in https://buildbot.halide-lang.org/master/#/builders/73/builds/60

steven-johnson:
are the environment vars controlling this (HL_USE_MEMOIZED_FEATURES=1, HL_MEMOIZE_BLOCKS=1, etc) meant to be a long-term API, or just a short-term expedient?

rootjalex:
@steven-johnson Short-term, input on thoughts for other APIs would be great. I also am not set on having caching off by default, it could just as easily be turned on by default

steven-johnson:
I have no opinion on on vs off by default.

I am concerned about the very large reliance that our autoschedulers have on env vars as a de facto 'api' for controlling a lot of things -- I don't have an alternate suggestion at this time, but I'd love to eventually have an alternative that doesn't require setting what are (effectively) mutable globals to do this sort of thing.

rootjalex:
@abadams Per the dev meeting, should caching be on by default? I am leaning towards that.

abadams:
Yes, it should be on by default.

aekul (changes requested):

rootjalex:
@abadams  and @aekul - caching is now the default, verification code has been removed, and tests have been added to adams2019/test.cpp.

rootjalex:
Failures appear to be unrelated, just opencl issues

steven-johnson:
Where does this stand? Ready-ish to land? Needs more work?

rootjalex:
I made the requested changes, just waiting on reviews to see if it's ready to land

steven-johnson:
OK. You may want to sync-to-master to see if the buildbots will go clean.

steven-johnson:
Where does this PR stand?

src/autoschedulers/adams2019/test.cpp
```diff
@@ -1,7 +1,58 @@
...
+    set_env_variable("HL_DISABLE_MEMOIZED_BLOCKS", "0", /* overwrite */ 1);
+    auto results_with_caching = p2.auto_schedule(target, params);
+
+    // Reset environment variables to what they were before (memoization variables are reset in main).
+    if (seed_value.empty()) {
+        // Re-empty seed.
+        set_env_variable("HL_SEED", "", /* overwrite */ 1);
+    }
+
+    // Compare produced schedules.
```
aekul, line 41: Very minor: this is comparing the produced features, not schedules.

aekul (commented):
The requested verification test changes look good to me. The rest looks good to me too but I'll let @abadams sign off on it.

src/autoschedulers/adams2019/Featurization.h
```diff
@@ -360,6 +360,32 @@ struct ScheduleFeatures {
...
+These intermediates allow for faster recomputation of such features.
+*/
+struct FeatureIntermediates {
+    double inlined_calls;
+    double num_vectors;
+    double num_scalars;
+    double vector_size;
+    double innermost_pure_loop_extent;
+    double outer_parallelism;
+    double num_warps_per_block;
```
abadams, line 386: This looks GPU-specific. Are some of these struct fields unused?

rootjalex: Oops yes, only 4 of these are used. Will fix.

rootjalex: Fixed in 2da7a71

src/autoschedulers/adams2019/test.cpp
rootjalex: Fixed in 169fb74

rootjalex:
Good to merge? @abadams @aekul

abadams:
I've been reluctant to green-light this because it complicates already-too-complicated code, by making the different aspects of the tree search even more entangled with each other. I worry it locks this code into a state where nobody else will ever be able to update it, due to the way it caches partially-computed bits of the featurization.

Maybe if you include a ridiculous number of explanatory comments it would help? There's no one big comment that explains all the things that caching touches and why.

rootjalex:
Apologies for how delayed I am in updating this - the start-of-the-semester craziness hit pretty hard.

@abadams Please let me know if the comments added to `Cache.h` and `Autoschedule.cpp` are or are not sufficient

rootjalex:
Failure seems to be an un-related build failure

src/autoschedulers/adams2019/Cache.h
```diff
@@ -13,6 +13,45 @@ namespace Halide {
 namespace Internal {
 namespace Autoscheduler {
 
+/*
+  Important changes that caching impacts, outside of this file and Cache.cpp:
+
+  - LoopNest::compute_features
+    If cache_features is enabled (i.e. HL_DISABLE_MEMOIZED_FEATURES!=1) then this function caches
+    the feautizations of its children, and if called again, reuses those cached feauturizations.
```
abadams, line 21: featurizations typo

abadams: (twice)

src/autoschedulers/adams2019/Cache.h
```diff
@@ -13,6 +13,45 @@ namespace Halide {
...
+    The features are saved in a LoopNest's member, std::map<> feature_cache. Some features do not
+    persist, and the FeaturesIntermediates strucct (see Featurization.h) is used to cache useful
```
abadams, line 23: strucct

src/autoschedulers/adams2019/Cache.h
```diff
@@ -13,6 +13,45 @@ namespace Halide {
 namespace Internal {
 namespace Autoscheduler {
```
abadams, line 15: This needs a big-picture overview comment as well as the list of changes below. E.g. there seem to be two types of caching: feature caching and block caching, but I'm still confused about what block caching is from reading the text below. Say what kinds of caching exist, what values are cached, what the key is, and why this saves work.

abadams:
@rootjalex Would be good to get this in for the 12.0 release. Just needs a few more comment tweaks I think.

src/autoschedulers/adams2019/Cache.h
rootjalex: Ah good catch, thanks! Fixed in dccd642

src/autoschedulers/adams2019/Cache.h
rootjalex: Fixed in dccd642

src/autoschedulers/adams2019/Cache.h
rootjalex: Hopefully addressed by dccd642 ? Let me know what you think.

rootjalex:
@abadams I think generally the tilings are faster to save than to re-generate. The speedup that Luke sees on the GPU autoscheduler is much more than we see here though, which I assume is because he's generating more tiling options.

abadams:
The thing I'm still confused about is what is being cached in the blocks case. Reading the code it looks like it's the set of child LoopNest nodes for things scheduled compute_root. So what you're saving is loop nest construction time. Is that correct? The comment made me think it was just saving the tile sizes, which didn't sound useful.

rootjalex:
Those LoopNests are those generated from the tiling options - I think it's a combination of saving LoopNest construction as well as tiling generation.

rootjalex:
I just updated the comment, hopefully it makes the cache description more clear?

abadams (approved):

aekul:
In the blocks case, it's saving all the compute_root level loop nests. Most importantly, this includes their featurizations, which is the main motivation.

aekul (approved):
The changes I requested look good.

rootjalex:
I think the tests that are failing + hanging seem to be due to the issue in #5925 . merging per @abadams

PR #5698, opened 2021-02-03
Don't make new IR nodes if nothing changed when simplifying

abadams:
Some statements and expressions get crafted anew each time when
repeatedly resimplified. This PR fixes all the cases I could find.

No significant changes to most pipelines. Reduces total compile time by 7% on one particularly pathological example.

### DIFF
--- a/src/Simplify_EQ.cpp
+++ b/src/Simplify_EQ.cpp
@@ -106,38 +106,26 @@ Expr Simplify::visit(const EQ *op, ExprInfo *bounds) {
         return mutate(rewrite.result, bounds);
     }
 
-    // Early-out if nothing changed
-    const Add *add = delta.as<Add>();
-    const Sub *sub = delta.as<Sub>();
-    if (sub &&
-        ((sub->a.same_as(op->a) &&
-          sub->b.same_as(op->b)) ||
-         (sub->a.same_as(op->b) &&
-          sub->b.same_as(op->a)))) {
-        return op;
-    } else if (add &&
-               add->a.same_as(op->a) &&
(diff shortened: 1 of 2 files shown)


steven-johnson (approved):

src/Simplify_EQ.cpp
```diff
@@ -106,25 +106,37 @@ Expr Simplify::visit(const EQ *op, ExprInfo *bounds) {
...
+        return op;
+    } else if (add &&
+               add->a.same_as(op->a) &&
+               (op->b.as<IntImm>() ||
+                op->b.as<UIntImm>() ||
+                op->b.as<FloatImm>())) {
+        // a - b became a + c. c must be the negation of b, which
+        // can't possibly get simpler.
+        return op;
+    } else if (delta.same_as(op->a) &&
```
rootjalex, line 126: Having this check before the `rewrite(x - y == 0, x == y)` below means that `a - b == 0` (assuming `a - b` doesn't simplify further) would be returned as simply `a - b == 0`, and never simplified to `a == b`

abadams: Good point. That must be why the tests are failing. I pushed a simpler alternative (though it does construct a another temporary Expr, sadly)

steven-johnson:
LGTM

src/Simplify_EQ.cpp
rootjalex: I think the first two conditions could actually be left in, which would still refrain from constructing temporary Exprs in those cases at least? I think just the last one causes problems.

PR #5702, opened 2021-02-03
Check Sapphire Rapids AVX512BF16 support in runtime

jwlawson:
The Sapphire Rapids target feature controls whether BF16 and VVNI X86
instructions are emitted. Support for both of these is checked in the
Halide compiler, but BF16 is not checked in the runtime as it required
extending the cpuid functionality.

Now we have that cpuid functionality we can add the BF16 check.


steven-johnson:
LGTM pending buildbots.

steven-johnson (approved):

PR #5704, opened 2021-02-03
Provide wrapper around 128 bit cvtneps2bf16

jwlawson:


steven-johnson:
LGTM pending buildbots.

src/LLVM_Runtime_Linker.cpp
```diff
@@ -1029,6 +1031,9 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
             if (t.has_feature(Target::AVX2)) {
                 modules.push_back(get_initmod_x86_avx2_ll(c));
             }
+            if (t.has_feature(Target::AVX512_SapphireRapids)) {
+                modules.push_back(get_initmod_x86_avx512_ll(c));
```
dsharletg, line 1035: Should this LL module be named x86_avx512_sapphirerapids? I am wondering if people might add other things to avx512.ll that aren't intended to be gated on this flag.

dsharletg: Alternatively: maybe we can just add this module if we have Target::AVX512, since usage of the function is already gated on this flag in CodeGen_X86.cpp?

steven-johnson (approved):

PR #5706, opened 2021-02-03
HalideBuffer: cast offset operand to ptrdiff_t to avoid overflow

cimes-isi:
Without this, allocating large buffers from an application fails with a segfault.


cimes-isi:
There may also be some problems elsewhere, like with `offset_of`

abadams (approved):

cimes-isi:
I have similar fixes for some `offset_of` functions, and in `HalideRuntime.h` for `start`, `end`, and `address_of`.  I'll push them shortly to be included in this PR.  I can't claim that these fixes are exhaustive though.

steven-johnson:
Failure is unrelated. Good to land.

PR #5707, opened 2021-02-05
Add mux intrinsic

abadams:
This PR turns mux into an intrinsic, instead of immediately unpacking into a select tree. When you mux a lot of values it produces a deeply nested select tree, and this is pathological for a few different lowering stages (mostly unrolling). Turning it into an intrinsic (lowered to selects during codegen) speeds up lowering time of one big pipeline by about 3x (174s -> 57s)

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -1368,17 +1368,21 @@ Tuple tuple_select(const Expr &condition, const Tuple &true_value, const Tuple &
 }
 
 Expr mux(const Expr &id, const std::vector<Expr> &values) {
-    user_assert(values.size() >= 2) << "mux() only accepts values with size >= 2.\n";
+    user_assert(!values.empty()) << "mux() requires a non-empty vector of values";
+    if (values.size() == 1) {
+        // Useful in generic code where the size of the values vector
+        // might be degenerate.
+        return values[0];
+    }
+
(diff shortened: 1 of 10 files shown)


src/IROperator.cpp
```diff
@@ -1368,17 +1368,21 @@ Tuple tuple_select(const Expr &condition, const Tuple &true_value, const Tuple &
 }
 
 Expr mux(const Expr &id, const std::vector<Expr> &values) {
-    user_assert(values.size() >= 2) << "mux() only accepts values with size >= 2.\n";
+    user_assert(!values.empty()) << "mux() requires a non-empty vector of values";
+    if (values.size() == 1) {
+        // Useful in generic code where the size of the values vector
+        // might be degenerate.
+        return values[0];
```
steven-johnson, line 1375: What does mux() do if the `id` Expression is out of range? (ie is there an assert-fail, or does it just always return the last/first/etc entry)? I presume it's analogous to multi-select and returns the final value for no-match, but I don't see this documented explicitly.

abadams: Same as before this PR. I updated the comment in IROperator.h to be clear.

steven-johnson (approved):

abadams:
Builds were all green before last commit, which was just a comment, so merging.

PR #5708, opened 2021-02-05
[hexagon] Host API for setting stack size

aankit-ca:
@dsharletg @pranavb-ca Users will have to call halide_hexagon_set_thread_params with thread priority and stack size before any RPC call.


src/runtime/hexagon_remote/host_shim.h
```diff
@@ -0,0 +1,39 @@
...
+#endif
+
+/* Set remote session parameters
+ *
+ * @param req, request ID
+ * @param data, address of structure with parameters
+ * @param datalen, length of data
+ * @retval, 0 on success
+ */
+extern "C" __QAIC_REMOTE_EXPORT int __QAIC_REMOTE(remote_session_control)(uint32_t req, void *data, uint32_t datalen) __QAIC_REMOTE_ATTRIBUTE;
```
dsharletg, line 37: Isn't this code just resolving to the definition in libadsprpc_shim.cpp? In that case, I think most of the stuff in this header is unnecessary (and may be wrong, e.g. dllexport stuff), we can just make a libadsprpc_shim.h and declare it there.

aankit-ca: Corrected

src/runtime/hexagon_host.cpp
```diff
@@ -918,6 +921,26 @@ WEAK void halide_hexagon_power_hvx_off_as_destructor(void *user_context, void *
...
 
+WEAK int halide_hexagon_set_thread_params(void *user_context, int priority, int stack_size) {
+    int result = init_hexagon_runtime(user_context);
+    if (result != 0) {
+        return result;
+    }
+
+    debug(user_context) << "halide_hexagon_set_thread_params\n";
+    if (!remote_set_thread_params) {
+        // This runtime doesn't support changing the performance target.
```
dsharletg, line 932: performance target -> thread params?

src/runtime/hexagon_remote/host_shim.cpp
```diff
@@ -3,13 +3,37 @@
...
+    struct remote_rpc_thread_params th;
+    th.domain = CDSP_DOMAIN_ID;
+    th.prio = priority;
+    th.stack_size = stack_size;
+    int retval = remote_session_control(FASTRPC_THREAD_PARAMS, (void *)&th, sizeof(th));
+    if (retval != 0) {
+        __android_log_print(ANDROID_LOG_ERROR, "halide", "Error: Could not set stack_size to %d. remote_session_control returned %d.",
+                            stack_size, retval);
+    }
+    return retval;
```
dsharletg, line 34: I think maybe this should return either halide_error_code_success or halide_error_code_generic_error, I think the current code is returning a Qualcomm error code as something that will probably be interpreted as a Halide error code. The other remote runtime functions are different because we implement them (that said, I've seen qualcomm error codes returned in crash conditions, so this is a bit janky for the other functions too).

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -106,6 +106,10 @@ typedef struct {
 // This is deprecated.
 typedef halide_hexagon_power_t halide_hvx_power_perf_t;
 
+/** Set fastRPC thread priority and staksize. Set thread params before making
```
dsharletg, line 109: stack size (in bytes?)

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -106,6 +106,10 @@ typedef struct {
...
+ * any RPC calls. halide_hexagon_set_thread_params. */
+extern int halide_hexagon_set_thread_params(void *user_context, int priority, int stack_size);
```
dsharletg, line 111: Maybe this and halide_hexagon_set_thread_priority should be documented together so they can share the more detailed comment about priorities?

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -106,6 +106,10 @@ typedef struct {
 // This is deprecated.
 typedef halide_hexagon_power_t halide_hvx_power_perf_t;
 
+/** Set fastRPC thread priority and staksize. Set thread params before making
+ * any RPC calls. halide_hexagon_set_thread_params. */
```
dsharletg, line 110: Remove stray function name here?

src/runtime/HalideRuntimeHexagonHost.h
pranavb-ca: I don't know if ``halide_hexagon_set_thread_priority`` mentions that lower int values mean higher priority. We should mention it in the comments here regardless.

src/runtime/hexagon_host.cpp
```diff
@@ -918,6 +921,26 @@ WEAK void halide_hexagon_power_hvx_off_as_destructor(void *user_context, void *
     halide_hexagon_power_hvx_off(user_context);
 }
 
+WEAK int halide_hexagon_set_thread_params(void *user_context, int priority, int stack_size) {
```
pranavb-ca, line 924: As discussed offline just now, renaming this to ``halide_hexagon_set_remote_thread_params``  makes more sense?

aankit-ca: Done. Thanks

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -17,6 +17,8 @@ extern "C" {
  */
 
 #define HALIDE_RUNTIME_HEXAGON
+#define FASTRPC_THREAD_PARAMS (1)
```
dsharletg, line 20: This (and remote_rpc_thread_params) shouldn't be in this header, they aren't used in the Halide API. I think they should be in hexagon_host.cpp

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -122,15 +121,28 @@ extern int halide_hexagon_set_performance_mode(void *user_context, halide_hexago
...
 
-/** Set the default priority for Halide Hexagon user threads:
+// Used with FASTRPC_THREAD_PARAMS req ID
+struct remote_rpc_thread_params {
+    int domain;      // Remote subsystem domain ID, pass -1 to set params for all domains
+    int prio;        // user thread priority (1 to 255), pass -1 to use default
+    int stack_size;  // user thread stack size, pass -1 to use default
+};
+
+/** Set the default priority/stack_size for Halide Hexagon user threads:
```
dsharletg, line 130: Does set_remote_thread_params apply to all FastRPC threads, or just the calling thread? At first I thought it was the latter, but now the docs make it sound like the former...

src/runtime/HalideRuntimeHexagonHost.h
aankit-ca: Done

src/runtime/HalideRuntimeHexagonHost.h
aankit-ca: halide_hexagon_set_thread_priority does mention about this: "Smaller number for higher priority"

src/runtime/HalideRuntimeHexagonHost.h
aankit-ca: Yes. Updated the comment

src/runtime/hexagon_remote/host_shim.cpp
aankit-ca: Done

src/runtime/hexagon_host.cpp
aankit-ca: Done

src/runtime/hexagon_remote/host_shim.cpp
aankit-ca: I've moved this code to hexagon_host.cpp.

src/runtime/HalideRuntimeHexagonHost.h
aankit-ca: Done. Also I noticed somewhere the default priority value for fastrpc threads was 192 and not 100. I'll check with @pranavb-ca once.

src/runtime/HalideRuntimeHexagonHost.h
aankit-ca: I'm not sure about this. Will confirm with Pranav and get back to you.

src/runtime/HalideRuntimeHexagonHost.h
steven-johnson: s/staksize/stack size/

src/runtime/HalideRuntimeHexagonHost.h
aankit-ca: It will apply to all fastrpc threads spawned by the process and not just he calling thread.

src/runtime/HalideRuntimeHexagonHost.h
```diff
@@ -118,15 +118,21 @@ extern int halide_hexagon_set_performance_mode(void *user_context, halide_hexago
...
 
-/** Set the default priority for Halide Hexagon user threads:
+/** Set the default priority/stack_size for Halide Hexagon user threads:
  *   - Valid priority values range from 1 to 255
  *   - Smaller number for higher priority
  *   - The highest priority for a user thread is 1
  *   - Priority 0 is reserved for OS usage
- * If this routine is not called, the priority will default to 100.
- * This is intended to be called before dispatching any pipeline. */
+ * If halide_hexagon_set_thread_priority routine is not called, the priority
```
pranavb-ca, line 126: @aankit-ca  - With PipelineContext removed ``halide_remote.cpp``, I don't think the plumbing for ``halide_hexagon_set_thread_priority`` to get the priority set for each thread in the parallel thread pool exists any more. We should either completely remove this function or change it to use ``remote_session_control`` where available and fall back to its current approach now.(I don't know if many people use it though). We should also use the new thread_params mechanism to set the thread priority and/ or stack size.

```
// to set only the priority
halide_hexagon_set_remote_thread_params(NULL, 10, -1);
// to set only the stack size
halide_hexagon_set_remote_thread_params(NULL, -1, 1024 * 32);
// to set both
halide_hexagon_set_remote_thread_params(NULL, 10, 1024 * 32);
```

src/runtime/hexagon_host.cpp
```diff
@@ -918,6 +931,34 @@ WEAK void halide_hexagon_power_hvx_off_as_destructor(void *user_context, void *
...
+    th.prio = priority;
+    th.stack_size = stack_size;
+    debug(user_context) << "    set_remote_thread_params("
+                        << "thread_priority: " << priority
+                        << ", thread_stack_size: " << stack_size << ") -> \n";
+    result = set_remote_thread_params(FASTRPC_THREAD_PARAMS, (void *)&th, sizeof(th));
+    debug(user_context) << "        " << result << "\n";
+    if (result != 0) {
+        error(user_context) << "set_remote_thread_params faiiled.\n";
+        return result;
```
pranavb-ca, line 957: Will this fastrcp error code not get conflated with a halide runtime error code?

steven-johnson:
Looks like this PR still needs some cleanup before it can land?

steven-johnson:
Where does this PR stand?

aankit-ca:
@steven-johnson The PR needs some changes. Unfortunately my dev machine is down due to power outage since last week. I'll update the PR soon.

pranavb-ca:
@steven-johnson & @dsharletg  - Due to the winter storm that Texas saw last week, our office where all our hardware for building and testing is saw a massive power outage. As if that wasn't enough, a sprinkler pipe burst caused flooding in our office. As such we  are scrambling to bring up some testing infrastructure for our group. @aankit-ca  will post an update on this PR once we have some capability restored.

steven-johnson:
Thanks for the update!

steven-johnson:
This PR has been sitting around for a few months now; is it still active?

aankit-ca:
@steven-johnson Sorry this PR fell through the cracks. But I think we need this PR. I'll resume working on this.

steven-johnson:
> @steven-johnson Sorry this PR fell through the cracks. But I think we need this PR. I'll resume working on this.

Thanks for the update!

dsharletg:
We now have another stack usage regression causing crashes. Is this PR something that is ready to merge early this week?

src/runtime/hexagon_host.cpp
```diff
@@ -918,6 +931,34 @@ WEAK void halide_hexagon_power_hvx_off_as_destructor(void *user_context, void *
...
+WEAK int halide_hexagon_set_remote_thread_params(void *user_context, int priority, int stack_size) {
+    int result = init_hexagon_runtime(user_context);
+    if (result != 0) {
+        return result;
+    }
+
+    debug(user_context) << "halide_hexagon_set_remote_thread_params\n";
+    if (!set_remote_thread_params) {
+        // This runtime doesn't support changing the thread params.
+        return -1;
```
dsharletg, line 943: In other similar "optional" cases where the runtime function is missing, we return 0.

src/runtime/hexagon_host.cpp
dsharletg: Yes, it will, this should return -1 or some other halide error code.

aankit-ca:
> We now have another stack usage regression causing crashes. Is this PR something that is ready to merge early this week?

Saw this late. I've most of the changes ready. Will put them up for review soon.

aankit-ca:
@dsharletg @pranavb-ca PTAL

alexreinking:
@aankit-ca - is there any path forward to merging this functionality? How were the binary files generated? In a manner equivalent to #7741?

PR #5712, opened 2021-02-05
Add support for AVX512 BF16 dot product

jwlawson:


src/CodeGen_X86.cpp
```diff
@@ -480,14 +486,16 @@ void CodeGen_X86::codegen_vector_reduce(const VectorReduce *op, const Expr &init
         Expr pattern;
         const char *intrin;
         Type narrow_type;
+        bool combine_init;
```
jwlawson, line 489: This may be better as a set of enum flags, similar to [CodeGen_PTX_Dev](https://github.com/halide/Halide/blob/32d5f71cef5ecea883c6fc26305181e555716d01/src/CodeGen_PTX_Dev.cpp#L444)

src/CodeGen_X86.cpp
```diff
@@ -480,14 +486,16 @@ void CodeGen_X86::codegen_vector_reduce(const VectorReduce *op, const Expr &init
...
     };
     // clang-format off
     static const Pattern patterns[] = {
-        {2, i32(widening_mul(wild_i16x_, wild_i16x_)), "pmaddwd", Int(16)},
-        {2, i32(widening_mul(wild_i8x_, wild_i8x_)), "pmaddwd", Int(16)},
-        {2, i32(widening_mul(wild_i8x_, wild_u8x_)), "pmaddwd", Int(16)},
-        {2, i32(widening_mul(wild_u8x_, wild_i8x_)), "pmaddwd", Int(16)},
-        {2, i32(widening_mul(wild_u8x_, wild_u8x_)), "pmaddwd", Int(16)},
+        {2, f32(wild_bf16x_) * f32(wild_bf16x_), "dot_product", BFloat(16), true},
```
dsharletg, line 493: I think this should just be `wild_f32x_ * wild_f32x_`. The operands will be `lossless_cast`ed to BFloat(16), which looks like it should work (https://github.com/halide/Halide/blob/master/src/Type.cpp#L118).

test/correctness/simd_op_check.cpp
```diff
@@ -511,6 +511,17 @@ class SimdOpCheck : public SimdOpCheckTest {
...
+
+            {
+                // 16 bit, 2 element dot product
+                RDom r(0, 2);
+                check("vdpbf16ps*zmm", 16, sum(f32(in_bf16(2 * x + r)) * in_bf16(2 * x + r + 32)));
+                check("vdpbf16ps*ymm", 8, sum(f32(in_bf16(2 * x + r)) * in_bf16(2 * x + r + 32)));
+                check("vdpbf16ps*xmm", 4, sum(f32(in_bf16(2 * x + r)) * in_bf16(2 * x + r + 32)));
+                // We match on `sum(f32 * f32)` for bf16 dot product, so check
+                // that we correctly generate code for the f32 dot product too.
+                check("", 4, sum(in_f32(2 * x + r) * in_f32(2 * x + r + 32)));
```
abadams, line 523: This string is empty? I'm confused.

jwlawson: I don't mind what instructions get generated for the f32 dot product, but want to be sure that it will generate something that behaves correctly. This test catches the potential error if I had not [changed the user_assert](https://github.com/halide/Halide/pull/5712/files#diff-6cd3cc01186f9ce4abaa1a98c9201bd440983ecc5620fbccae383df1116bdce6L507-R518). Is there a better way of checking this?

dsharletg: I think https://github.com/halide/Halide/blob/master/test/correctness/vector_reductions.cpp will already test this, so there's no need to add anything to simd_op_check.

dsharletg (approved):

PR #5717, opened 2021-02-08
Fix symbol names used for outputting schedules into valid C++ source variable names

derek-inteon:
Since Funcs and other classes can have user defined names, the resulting strings may be invalid C++ when used directly for variable names and identifiers during the source code generation for schedules.

This PR adds a 'conform_name' around all variable names being outputted to replace invalid characters with an underscore, to make sure only valid C++ identifiers are outputted for the schedule source.

### DIFF
--- a/src/autoschedulers/adams2019/ASLog.h
+++ b/src/autoschedulers/adams2019/ASLog.h
@@ -31,6 +31,9 @@ public:
     static int aslog_level();
 };
 
+// Conform the given name into a valid C++ identifier (eg for dumping a Func/Var inside a schedule to a header)
+std::string conform_name(const std::string &name, const std::string &prefix="_"); 
+
 }  // namespace Internal
 }  // namespace Halide
 
--- a/src/autoschedulers/adams2019/ASLog.cpp
+++ b/src/autoschedulers/adams2019/ASLog.cpp
@@ -1,5 +1,8 @@
 #include "ASLog.h"
 
+#include <algorithm>
+#include <cctype>
+
 namespace Halide {
 namespace Internal {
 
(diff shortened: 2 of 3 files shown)


src/autoschedulers/adams2019/ASLog.h
```diff
@@ -31,6 +31,9 @@ class aslog {
     static int aslog_level();
 };
 
+// Conform the given name into a valid C++ identifier (eg for dumping a Func/Var inside a schedule to a header)
+std::string conform_name(const std::string &name, const std::string &prefix="_");
```
steven-johnson, line 35: This function doesn't have anything to do with logging, and thus doesn't belong in ASLog.h

derek-inteon: es, that's true. I placed it here since it was a common header, and deals with formatting text, but it's not actually doing any logging.

Would you propose a new header/source file or some other location?

dsharletg: It looks like it's only used in the one source file LoopNest.cpp? I would just put it in there (in an anonymous namespace too).

steven-johnson: SGTM

steven-johnson:
Where does this PR stand?

steven-johnson:
Monday Morning Review ping: Where does this PR stand?

steven-johnson:
Is this PR still active? If not, it will be closed.

steven-johnson:
Is this PR still active?

mcourteaux (changes requested):
As this fix is rather ad-hoc, I believe a custom lowering pass that legalizes all Func names would be more reasonable, as that solves this issue for all auto-schedulers.

Also, there is already some sanitation happening here:
https://github.com/halide/Halide/blob/8bf7827bf08d655f0c842d4fe7271d5bf65600f9/src/autoschedulers/adams2019/State.cpp#L677-L685

So this PR definitely needs changing.

PR #5719, opened 2021-02-08
Minor cleanup to GeneratorStub

steven-johnson:
Cherrypicked from an abandoned CL, this just removes some unnecessary code and de-inlines some methods.


steven-johnson:
Meh, this conflicts with the existing PyStubImpl code in a way that makes it not worth continuing. Withdrawing.

Issue #5634, opened 2021-01-12
intermittent failures of mul_div_mod with host-opencl


steven-johnson:
There have been sporadic failures of correctness_mul_div_mod on the linux buildbots in the past week or two, but only when running with target=host-opencl. So far, no one has been able to reproduce it locally; more strangely, it doesn't seem to repeat even when logging into a buildbot with the failure and manually re-running. Injection point is unclear. Cause is unclear. Opening this issue for now, and will add instances of it occurring in comments.

steven-johnson:
```
Only failure is the rogue mul_div_mod failure:

FAILED TEST: mul_div_mod
Testing mul vector_width: 1
Testing mul vector_width: 2
Testing mul vector_width: 4
Testing mul vector_width: 8
Testing mul vector_width: 16
Testing div_mod vector_width: 1
Testing div_mod vector_width: 2
Testing div_mod vector_width: 4
Testing div_mod vector_width: 8
Testing div_mod vector_width: 16
Failure!
0*-2147483648 -> -2147483648 != 0
Compiled a*b != simplified a*b: 0*-2147483648 = -2147483648 != 0
0*2147483647 -> 2147464527 != 0
Compiled a*b != simplified a*b: 0*2147483647 = 2147464527 != 0
65535*156429877 -> 946612400 != -454946357
23964*-1339975209 -> -1705444715 != -1990403580
47774*672691988 -> 1426553552 != -2053241256
52075*-387718748 -> 419095856 != 187456396
53663*975805134 -> -2138888350 != 389633010
19650*418751201 -> 1851754037 != -696239486
0*-2147483648 -> -2147483648 != 0
Compiled a*b != simplified a*b: 0*-2147483648 = -2147483648 != 0
0*2147483647 -> 2147464527 != 0
Compiled a*b != simplified a*b: 0*2147483647 = 2147464527 != 0
65535*156429877 -> 946612400 != -454946357
23964*-1339975209 -> -1705444715 != -1990403580
47774*672691988 -> 1426553552 != -2053241256
52075*-387718748 -> 419095856 != 187456396
53663*975805134 -> -2138888350 != 389633010
19650*418751201 -> 1851754037 != -696239486
make: *** [/home/halidenightly/build_bot/worker/x86-64-linux-testbranch-11-make/halide-source/Makefile:1820: quiet_correctness_mul_div_mod] Error 1
make: *** Waiting for unfinished jobs....
```

steven-johnson:
[...]

PR #5724, opened 2021-02-09
Add more debug output to mul_div_mod in hopes of tracking down #5634)

steven-johnson:


dsharletg:
I think the thing we really need to add is outputing more of the results, so we can see if the results are getting shifted/corrupted somehow. Like if we printed all the failures, not just the first one, I think that would be helpful. Or maybe we need to limit it to the first N in each dimension to avoid exploding the logs...?

test/correctness/mul_div_mod.cpp
```diff
@@ -569,11 +573,24 @@ int main(int argc, char **argv) {
         }
     }
 
-    Halide::Internal::ThreadPool<bool> pool;
+    size_t num_threads = Halide::Internal::ThreadPool<bool>::num_processors_online();
+    if (target.has_feature(Target::OpenCL)) {
```
abadams, line 577: ooh, could the bug be a race condition somewhere inside the compiler between independent compilations?

abadams (approved):

test/correctness/mul_div_mod.cpp
steven-johnson: yeah, that's one theory. I don't see any obvious issues that make this likely, though. If this seems to clear it up, some targeting investigation of this theory might be worthwhile.

dsharletg: Maybe we should just put a mutex around kernel compilation as a speculative fix? I doubt this would be a performance issue in any real case.

abadams: You mean in case the race condition is inside our opencl runtime at kernel compilation term? I was worried it was inside libHalide, and we don't want to put a mutex around all of compilation.

dsharletg: I guess it could be, but these tests are only failing on OpenCL and CUDA. They're still parallelized for x86, ARM, Hexagon and we haven't seen any flakes on those targets. So I suspect the issue is in the runtime around initialization of the kernels.

PR #5725, opened 2021-02-10
Add support for AV512 VNNI instructions

jwlawson:


dsharletg (approved):
This isn't exactly a code review question, but: is there a u8 * u8 -> u32 dot product? That version seems important from the workloads I've looked at so far...

PR #5730, opened 2021-02-11
Fix EGL bugs

alexreinking:
I'm trying to fix the outstanding EGL bugs on the buildbots before releasing v11. But when trying to repro locally, I ran into a _new_ set of bugs!

1. I noticed that `USE_RUNTIME` is broken because the `egl` backend has to match across all runtime targets. So I added that to the `matching_features` list in Target.cpp
2. This one is more serious. Some EGL implementations can unload their libraries _before_ (or perhaps concurrently with) `__attribute__((destructor))` functions. This causes the EGL device-release to call a function in an unmapped page and segfault. I tried re-querying the location of one of the symbols, but it's apparently cached so the best I could do was explicitly release the device before main exits in the GLSL app.

### DIFF
--- a/apps/glsl/opengl_test.cpp
+++ b/apps/glsl/opengl_test.cpp
@@ -55,4 +55,9 @@ int main(int argc, char *argv[]) {
     test_blur();
     test_ycc();
     test_device_sync();
+
+    // This is supposed to be called as an __attribute__((destructor)), but
+    // some EGL implementations can start to unload their libraries before those
+    // destructors run. On such systems, doing this outside of main() segfaults.
+    halide_device_release(nullptr, halide_opengl_device_interface());
 }
--- a/src/CodeGen_OpenGL_Dev.cpp
+++ b/src/CodeGen_OpenGL_Dev.cpp
@@ -1045,6 +1045,8 @@ void CodeGen_GLSL::add_kernel(const Stmt &stmt, const string &name,
     if (is_opengl_es(target)) {
         stream << "#ifdef GL_FRAGMENT_PRECISION_HIGH\n"
                << "precision highp float;\n"
+               << "#else\n"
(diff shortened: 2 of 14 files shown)


alexreinking:
https://github.com/halide/Halide/blob/df8a5ece9cd23d3e518ee3e5d267a9fc95582076/src/runtime/opengl.cpp#L2049-L2053

From offline discussion: @abadams suggested getting rid of this. Would just have to tell people to manually call `halide_device_release` before unloading a shared library containing our runtime.

alexreinking:
The current state of this PR passes our current buildbot testing, _however_ there are still issues with EGL on the opengl JIT tests (which are not tested). Set `Halide_TARGET=host-opengl-egl` and run `ctest -L opengl` to reproduce. All failures are of the form:

```
terminate called after throwing an instance of 'Halide::RuntimeError'
  what():  Error: Setting up GL framebuffer 1 failed (36054)
```

Where 36054 is `GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT` (according to my system's `GL/glext.h`)

alexreinking:
@xndcn I was wondering if you could take a look at this, since you were the last one to touch the OpenGL code and I do not know very much about this stuff.

Please let me know regardless. Thanks!

alexreinking:
Turns out that OpenGL ES does not allow for 3-channel framebuffers for floating point; only 1, 2, and 4 are allowed (and even then only with an appropriate extension). In plain OpenGL, it automatically converts 3-channel to 4-channel behind the scenes. Rather than introduce our own conversions, I'm just adding a clearer error message when OpenGL ES is active (EGL prefers GLES) and skipping the tests that use 3-channel float buffers.

abadams (approved):

steven-johnson (approved):

PR #5538, opened 2020-12-08
Update images used in apps/ tests

steven-johnson:
Some of them weren't the same as the Make equivalents, which meant that the test diverged between the two build systems (sometimes causing failures due to too-large images).

Issue #5403, opened 2020-10-24
c_source generation brokend on msvc: roundf/round already linked & cannot access private member declared in class


chkone:
When we generate with autoscheduled with:
`Generator.exe -o src\autoschedule\. -g Arithmetic_XXX -f Arithmetic_XXX -e schedule,c_header,c_source -p autoschedule_adams2019.dll -s Adams2019 target=host auto_schedule=true machine_params=32,16777216,40`

Version: tag: Halide 10.0.0
Platform: Windows x64
Compiler: Visual C++ /std:c++lastest
Windows SDK: 10.0.16299.0

On the head of the cpp generated we have:
```
#ifdef _WIN32
     __declspec(dllimport) float __cdecl roundf(float);
     __declspec(dllimport) double __cdecl round(double);
#else
```

Which on my context (== compiler) generate those error:
```
Arithmetic_XXX.halide_generated.cpp(16,37): error C2375: 'roundf': redefinition; different linkage
C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\ucrt\corecrt_math.h(576): message : see declaration of 'roundf'
Arithmetic_XXX.halide_generated.cpp(17,38): error C2375: 'round': redefinition; different linkage
C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\ucrt\corecrt_math.h(532): message : see declaration of 'round'
Arithmetic_XXX.halide_generated.cpp(43,41): error C2065: 'roundf': undeclared identifier
Arithmetic_XXX.halide_generated.cpp(62,50): error C2264: 'round': error in function definition or declaration; function not called
```

math.h contains only:
```
#include <corecrt_math.h>

#ifdef _USE_MATH_DEFINES
    #include <corecrt_math_defines.h>
#endif
```

If I comment the two '__declspec(dllimport)' of roundf & round I got following errors:
```
Arithmetic_XXX.halide_generated.cpp(2747,14): error C2248: 'CppVector<uint8_t,8>::elements': cannot access private member declared in class 'CppVector<uint8_t,8>'
Arithmetic_XXX.halide_generated.cpp(2814): message : see declaration of 'CppVector<uint8_t,8>::elements'
Arithmetic_XXX.halide_generated.cpp(3236): message : see declaration of 'CppVector<uint8_t,8>'
[...]

PR #5734, opened 2021-02-13
Backport outstanding changes to 10.0.1

alexreinking:
This backports #5388, #5403, #5538, #5617, #5640, #5644, and #5675 to release 10.0.1 (we should cut this soon, too).

**This PR should be REBASED, not squashed**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2034,6 +2034,10 @@ ifneq (,$(findstring clang version 11.0,$(CLANG_VERSION)))
 CLANG_OK=yes
 endif
 
+ifneq (,$(findstring clang version 11.1,$(CLANG_VERSION)))
+CLANG_OK=yes
+endif
+
 ifneq (,$(findstring clang version 12.0,$(CLANG_VERSION)))
 CLANG_OK=yes
 endif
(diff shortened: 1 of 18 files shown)


abadams:
Looks like cuda is busted with this release on older llvms, because it tries to use a compute capability that didn't exist back then when the runtime detects it's available. Was the release broken? I don't see the change that added new cuda capabilities in the PR.

alexreinking:
> Looks like cuda is busted with this release on older llvms, because it tries to use a compute capability that didn't exist back then when the runtime detects it's available. Was the release broken? I don't see the change that added new cuda capabilities in the PR.

It might well have been. This is running on the buildbots with newer GPUs, so that's likely why it wasn't caught earlier. Any suggestions?

abadams:
Maybe backport #5617 too?

abadams:
Wait, the description says you did

alexreinking:
> Maybe backport #5617 too?

Well, I _did_ and then forgot to push!

alexreinking:
Now we're getting `Error: CUDA: cuMemAlloc failed: CUDA_ERROR_OUT_OF_MEMORY` on lens_blur on win-worker2

abadams:
If it's the autoschedule faulting, then it's possible backporting #5694 would fix it. Don't worry about it too much though. That's the bot with the GPU from 2014, and the lens blur app is the one that allocates the most GPU memory.

alexreinking:
Up to you. If that fix would make things green, then I don't see why not. Otherwise, I think this is good to land.

alexreinking:
Ah, maybe it's #5538 that I need to fix.

alexreinking:
Great! Now we're all green.

abadams (approved):

Issue #5738, opened 2021-02-16
test/correctness/interpreter failing on opencl and d3d12 on older windows buildbot


abadams:
That buildbot has a gtx 750 in it, and this test isn't failing anywhere else or with cuda on the same machine, so I'm inclined to blame the driver.

steven-johnson:
OK, for now let's blocklist those on that machine. We really need to upgrade the winbots in general. (Maybe one of the decommissioned linuxbots has a less-bad card we could swap in temporarily?)

steven-johnson:
According to lspci -v, one of the old linuxbots has a Quadro 600; the other has a Quadro K600

abadams:
Those both seem to be even older. The gtx 750 was maxwell. Those two cards are fermi and kepler, which came before.

shoaibkamil:
Should this workaround be undone now?

steven-johnson:
> Should this workaround be undone now?

Dunno, what's changed?

shoaibkamil:
Ah, I guess we added a bot, but didn't decommission the GTX750 one.

steven-johnson:
> Ah, I guess we added a bot, but didn't decommission the GTX750 one.

Let's decommission it then. Which one is it?

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_STALE):** The failure was specific to one 2021 Windows buildbot with a GTX 750, and the thread concluded that machine should be decommissioned. The workaround lived in the external buildbot config, not in this repo, and `test/correctness/interpreter.cpp` is enabled unconditionally here with no blocklist to undo.

**Evidence:** test/correctness/CMakeLists.txt:429 lists `interpreter.cpp` with no exclusion; grep for `interpreter` across `.github/workflows/` and the `Makefile` finds no blocklist entry; PR CI now runs under GitHub Actions (`.github/workflows/testing-windows.yml`).

Closing: this tracked a driver-specific failure on a 2021 GTX 750 Windows buildbot that the thread agreed to decommission; there is no blocklist entry left in this repo, and PR testing has since moved to GitHub Actions. Please reopen with fresh logs if `correctness_interpreter` fails on current CI hardware.

PR #5739, opened 2021-02-16
Workaround for issue #5738

abadams:
This disables the test on the specific target for which it is failing. The test is still run under opencl, but only on linux.

### DIFF
--- a/test/correctness/interpreter.cpp
+++ b/test/correctness/interpreter.cpp
@@ -4,6 +4,7 @@ using namespace Halide;
 
 int main(int argc, char **argv) {
 
+    // TODO(#5738): remove after winbots are upgraded
     Target target = get_jit_target_from_environment();
     if (target.os == Target::Windows &&
         (target.has_feature(Target::OpenCL) ||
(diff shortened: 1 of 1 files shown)


test/correctness/interpreter.cpp
```diff
@@ -4,6 +4,14 @@ using namespace Halide;
 
 int main(int argc, char **argv) {
 
+    Target target = get_jit_target_from_environment();
```
steven-johnson, line 7: Add `// TODO(#5738): remove after winbots are upgraded` or similar

steven-johnson (approved):

abadams:
clean builds, woo. It's amazing what you can achieve by simply not running failing tests.

steven-johnson:
for our next trick, we will increase the speed of apps/fft by 100x+ by simply filling in the results to zero and declaring it within acceptable error tolerance :-)

abadams:
Hey, fast-math is on, so it's not a miscompilation

steven-johnson:
more like `-fstupid-fast-math` than `-ffast-math` but sure

PR #5740, opened 2021-02-16
Drop LLVM10 support from master

steven-johnson:
Update build files to require LLVM11+ in master branch. (Since we only regularly test master with 12 and 13 this is conservative.)

Remove all code that is specialized for LLVM < 11.0.

(Rationale is that LLVM10 is well-handled by the release branches.)


abadams:
The most recent release of llvm is still 11, so we should only merge this once they release 12 if we want to continue to support the last two llvm releases.

steven-johnson:
> The most recent release of llvm is still 11, so we should only merge this once they release 12 if we want to continue to support the last two llvm releases.

OK, so to be clear, we still want to ensure that our main branch can compile + work with LLVM 10 thru 13 at this point in time, yes?

In that case, we will need to add a couple more tasks back to the buildbot for PRs, as the current status quo appears to be that PRs against main are only tested against 12 and 13. (cc'ing @alexreinking as FYI)

alexreinking:
> the current status quo appears to be that PRs against main are only tested against 12 and 13. (cc'ing @alexreinking as FYI)

I see LLVMs 11-13 in the checks on this PR. We could add 10, though.

steven-johnson:
> I see LLVMs 11-13 in the checks on this PR. We could add 10, though.

You mean, with this change in place? (1) What checks are you seeing? (2) Dangit, what's wrong with my logic here, LLVM_10 is still in LLVM_BRANCHES.... right?

steven-johnson:
LLVM12 has gone final, so we can (arguably) consider this now.

abadams (approved):

alexreinking (approved):

alexreinking:
Those CUDA capability checks were something else. Those should probably have warned the user.

Issue #4283, opened 2019-09-27
CPU GPU memory transfer bug?


BachiLi:
I might be doing something wrong but the following code doesn't output what I expect:
```
#include "Halide.h"

using namespace Halide;

int main(int argc, char *argv[]) {
    int img_size = 1000;

    Func f, g;
    Var x, y;
    RDom r(0, img_size);

    f(x) = x * x;
    g(x, y) = f(x + y);

    g.compute_root().parallel(y);
    f.store_at(g, y)
     .compute_at(g, x);
    f.gpu_single_thread();
    
    Buffer<int> b = g.realize(10, 10);
    for (int i = 0; i < 10; i++) {
        for (int j = 0; j < 10; j++) {
            std::cout << "b(" << i << "," << j << "):" << b(i, j) << std::endl;
            assert(b(i, j) == (i + j) * (i + j));
        }
    }
    return 0;
}
```
The assertion fails at b(0, 1) where it outputs 0 (expected 1).

abadams:
Yeah, I think there's a missing copy_to_host in between the production and consumption of f due to it being inside the store_at level.

abadams:
The thing is, even if it worked, it would be a terrible schedule, because you would be copying back and forth all of f just to reuse one additional value from it. This is probably why we've never noticed this bug before.

BachiLi:
I know, I was torturing Halide with a variety of tests.

abadams:
Torture tests are always appreciated :)

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the
> reason below. The reasoning was checked against the current source, and a second pass
> tried and failed to refute it, but no human verified this specific text.
> **If the reason is wrong, please reopen the issue and say so** — some fraction of these
> are expected to be wrong.

**Reason (CLOSE_FIXED):** The reported missing copy_to_host between production and consumption of a GPU-computed Func stored at an outer CPU loop level no longer happens. I ran the exact repro from the issue on the current build with HL_JIT_TARGET=host-cuda and every element matched (i+j)^2.

**Evidence:** Live repro prints OK; HL_DEBUG_CODEGEN=1 shows the expected `halide_copy_to_device(f0.buffer, halide_cuda_device_interface())` ... `gpu_thread<CUDA>` ... `halide_copy_to_host(f0.buffer)` sequence inside the store_at region.

[...]

PR #5741, opened 2021-02-16
Allow device copies inside a loop

abadams:
If you have a producer consumer pair that spans two devices, we inject copies between. But we only currently do this at the loop level of the realize node for the producer. This is the wrong place if the compute_at level is distinct from the store_at level. This PR changes things so that the copy injection pass enters all loops for which the buffer is used on multiple devices in the body.


abadams:
@dsharletg here's anothing sliding window test case for you.

abadams:
Also, the intent of this is to fix #4283

dsharletg (approved):

steven-johnson:
Ready to land?

PR #5746, opened 2021-02-16
Fix clang-tidy warnings in src/autoschedulers

steven-johnson:
A handful of minor fixes allows us to remove the custom .clang-tidy for src/autoschedulers entirely. (Note that this PR is additive to https://github.com/halide/Halide/pull/5743, which must land first.)


alexreinking (approved):
No reason to give the autoschedulers special clang-tidy rules when this is all that's changed.

PR #5747, opened 2021-02-17
Remove bugprone-sizeof-expression from src/runtime/.clang-tidy

steven-johnson:
Re-enable `bugprone-sizeof-expression` for clang-tidy in runtime, and annotate the failing cases with NOLINT to silence them. In general, this is a useful check, but it can be way too aggressive (it assumes that sizeof(T) where T is a pointer-to-aggregate is ~always wrong, which is plainly untrue for this code).  IMHO we're better off with keeping the check (to look for future situations) and the NOLINT annotation (which is a reminder to the reader that this is a use that we have vetted and want to keep).

(Note that this PR is additive to #5743, which must land first.)


abadams:
Not sure I buy this one, given that it didn't find anything problematic and requires a bunch of NOLINT tags.

steven-johnson:
Yeah, this one I could arguably take or leave. The reason I bothered to go forward with it is that the thing that it's ostensibly checking (ie, mistakenly doing `sizeof(foo*)` when you wanted `sizeof(foo)`) is a legit error, and in most modern C++ code is a genuinely rare occurrence (we only see it in our runtime code because we are resorting to more C-ish idioms to avoid runtime deps); enabling it and then marking it this way is a way to indicate "I really did doublecheck this and it is correct".

That said, yeah, not a slam dunk. I'll leave this open for a bit and see if anyone else has strong feelings either way.

steven-johnson:
After more pondering, I think I agree this doesn't pay for itself. Withdrawing.

PR #5748, opened 2021-02-17
Eradicate reinterpret() from runtime_internal.h

steven-johnson:
The reinterpret<>() function in runtime_internal.h is dangerous and should not be allowed to exist: it copies from one type to another, copying the minimum of the two sizes, but with no allowance for ensuring that any "extra" bits in the destination are filled in with, well, anything. I have no evidence of a bug specifically caused by this, but a usage such as `reinterpret<uint64>t>(some_ptr)` on a 32-bit runtime will leave the upper 32 bits of the result undefined. Maybe they'll never get used, but this function is a bug waiting to happen.

Fortunately, the only usages of this evil thing are all in hexagon_host.cpp, and all are used to convert between a uint64_t and an ion_device_handle* (which is sometimes represented as a void*). Replacing the handful of usages with bespoke functions to do these conversions seems much safer.


src/runtime/hexagon_host.cpp
```diff
@@ -328,6 +328,20 @@ WEAK void halide_hexagon_finalize_kernels(void *user_context, void *state_ptr) {
 
 namespace {
 
+ALWAYS_INLINE ion_device_handle *uint64_to_ion_device_handle(const uint64_t &dev) {
+    static_assert(sizeof(ion_device_handle *) <= sizeof(uint64_t), "");
+    ion_device_handle *h = nullptr;
+    memcpy(&h, &dev, sizeof(dev));
```
dsharletg, line 334: Isn't this stomping memory if sizeof(void*) < sizeof(dev)?

src/runtime/runtime_internal.h
```diff
@@ -209,12 +209,6 @@ ALWAYS_INLINE T min(const T &a, const T &b) {
     return a < b ? a : b;
 }
 
-template<typename T, typename U>
-ALWAYS_INLINE T reinterpret(const U &x) {
-    T ret;
```
dsharletg, line 214: I think a better fix might be to just initialize `T ret = T()`.

dsharletg (commented):
I think the old code was better. Uninitialized upper 32-bits aren't a big deal if we aren't doing arithmetic on these things, and I think the new code is sketchy too. 

If you feel strongly about it, I proposed an alternative fix. (I have no objection to moving the function to hexagon_host.cpp if it's the only user of the helper.)

src/runtime/hexagon_host.cpp
steven-johnson: sizeof(foo&) == sizeof(foo), so this would only stomp if sizeof(uint64_t) < sizeof(ion_device_handle*), which can't be the case for any existing hexagon device. (see also the static_assert.)

steven-johnson:
> Uninitialized upper 32-bits aren't a big deal

Uninitialized bits are always a big deal; when it's unavoidable and provably harmless, I can live with that, but it's trivially avoidable here.

> and I think the new code is sketchy too.

Could you expand on that a bit? I'm not married to my replacement code but I'd like to hear the concrete concerns.

src/runtime/runtime_internal.h
steven-johnson: That's definitely better than the status quo, but I stand by my assertion that this function is a bad idea, requiring too much care on the part of the caller.

abadams:
To me reinterpret implies src bits == dst bits, so I think it's at least misleadingly named. I think the idiom could just be casting from pointer to uintptr_t and then to uint64_t. That unambiguously widens with zeros.

steven-johnson:
> To me reinterpret implies src bits == dst bits, so I think it's at least misleadingly named. I think the idiom could just be casting from pointer to uintptr_t and then to uint64_t. That unambiguously widens with zeros.

PTAL

src/runtime/hexagon_host.cpp
steven-johnson: ...dammit, yes. (Another reason memcpy for this is bad.)

dsharletg (approved):

PR #5749, opened 2021-02-17
Enable bugprone-macro-parentheses in clang-tidy

steven-johnson:
...adding parens or annotating exceptions with NOLINT where necessary.


src/runtime/mini_cl.h
```diff
@@ -131,65 +131,65 @@ typedef struct _cl_buffer_region {
...
-#define CL_INVALID_GL_OBJECT -60
-#define CL_INVALID_BUFFER_SIZE -61
-#define CL_INVALID_MIP_LEVEL -62
-#define CL_INVALID_GLOBAL_WORK_SIZE -63
-#define CL_INVALID_PROPERTY -64
-#define CL_INVALID_IMAGE_DESCRIPTOR -65
-#define CL_INVALID_COMPILER_OPTIONS -66
-#define CL_INVALID_LINKER_OPTIONS -67
-#define CL_INVALID_DEVICE_PARTITION_COUNT -68
+#define CL_DEVICE_NOT_FOUND (-1)
```
abadams, line 134: Can we not #define negative constants with this checker on? That seems weird.

abadams:
Ratio of genuine fixes to false positives isn't great.... but any of the issues it found would have been a nightmare to debug, so I guess I'll just deal with the boilerplate.

abadams (approved):

PR #5750, opened 2021-02-17
Enable bugprone-incorrect-roundings for clang-tidy

steven-johnson:
...and fix the offending code.


tools/halide_image_io.h
```diff
@@ -263,11 +264,11 @@ inline uint32_t convert(const int64_t &in) {
 }
 template<>
 inline uint32_t convert(const float &in) {
-    return (uint32_t)(in * 4294967295.0 + 0.5);
+    return (uint32_t)std::lround(in * 4294967295.0);
```
abadams, line 267: Can a long represent all uint32_ts? I forget on which platforms a long is still 32-bit signed. This seems like it could wrap in an undefined way. Maybe std::round would be better.

steven-johnson: Ugh. This is actually what clang-tidy recommended. Maybe std::llround()?

alexreinking: For what it's worth, the behavior is not undefined. When the argument to `std::lround` is not representable by `long`, it raises `FE_INVALID` and returns an _implementation defined_ value. Still, `long` is 32-bit on LLP systems (like Windows), so `std::llround()` should be used.

alexreinking: Also, what the heck is that constant? Can `4294967295` be given a name?

steven-johnson: > Also, can that constant be written `0xFFFFFFFF.0`? Or maybe `(float) UINT32_MAX`?

Gotta say that I find it hard to care about this one.

alexreinking: :shrug:

abadams (approved):

PR #5754, opened 2021-02-17
Add Ubuntu package scripts

### LINKED ISSUE #5285: Create Debian packages
It would be nice for users of Debian and/or Ubuntu to simply run:

```
$ sudo apt install libhalide libhalide-dev libhalide-doc
```

However, the rules for getting a package upstreamed are _stringent_ and there's a veritable _book_ of package guidelines for Debian, here:

https://www.debian.org/doc/manuals/debmake-doc/

and Ubuntu extends it here:

https://packaging.ubuntu.com/html/index.html

**However, PPAs are more "anything goes".** I think the right strategy is to figure out deb packaging enough that we get an established user base on the PPA and then try to make the package conform to upstream guidelines, possibly with some help from maintainers.

----

Even with the flexible Ubuntu PPA rules, there are a few things that we _do_ need to keep in mind to be good citizens:

1. We should build and link against Ubuntu's own shared libraries. This just means LLVM.
2. We need to split our library across _three packages_:
  a. `libhalide` which includes the `.so` [1] and the autoscheduler modules.
  b. `libhalide-dev` which includes most other things: `.a`, headers, tools, etc.
  c. `libhalide-doc` which includes the tutorials and the doxygen documentation.
3. We need to list libjpeg-dev, libpng-dev as "recommended" packages of `libhalide-dev`
4. We need to list `libhalide` as a dependency of `libhalide-dev` and `libhalide-doc` as a suggested package for `libhalide-dev`.
5. `libhalide-doc` shouldn't have any dependencies.
6. We need to package from inside a chroot with a minimal environment.

[1] I did an `apt-file` search for every `.a` in my `/usr` directory and of the 500+ static libs there, all but 47 were in the corresponding `-dev` package. The others were from `liblld-10`, which I'm nearly certain is mis-packaged; `ocaml-nox`, which I'm guessing needs them for linking; and `valgrind`, `pristine-tar`, and `python3-numpy` for unknown reasons.

alexreinking:
This PR would add long-awaited Ubuntu package scripts!

## Ubuntu

The Ubuntu package is split into runtime, development, and documentation components. Development naturally depends on runtime. The packages have standard names: `libhalideV`, `libhalideV-dev`, and `libhalideV-doc` where V is the major version (also the SOVERSION). Naturally, they link to system LLVM.

I had created a [PPA we can use](https://launchpad.net/~alexreinking/+archive/ubuntu/halide) but apparently launchpad wants _source debs_ instead of _binary debs_. These have a weird, special format that's poorly documented and---once unpacked---passes through nearly a dozen bespoke debian packaging tools to compile our sources from scratch and feed them through their packaging pipeline. So that's not happening when we can just publish our perfectly good CPack-generated debs on a private APT repo.

### GitHub Actions
There is testing for the Ubuntu packages. We create a GHA VM with developer tools for building and packaging Halide, then save the packages to scratch storage before creating a _second, fresh_ VM with a minimal set of tools. The second VM installs the packages and then tests a very basic pipeline in a wide variety of build scenarios to test that our CMake package works as expected and without _**any**_ special setup. These tests are located in `test/integration`.

## To-dos

- [x] Verify Ubuntu package dependencies
  - [x] [Hear back on the CMake Discourse](https://discourse.cmake.org/t/cpack-debian-dpkg-shlibdeps-with-components/2781) about `dpkg-shlibdeps`: 
- [x] GHA testing
  - [x] Test Ubuntu package
    - [x] Test JIT with shared/static
    - [x] Test AOT with shared/static
    - [x] Test AOT error when attempting to use an autoscheduler and static linking
    - [x] Test JIT with _both_ shared and static
    - [x] Test AOT cross-compile
- [x] Update buildbots to call TGZ / ZIP packaging scripts by their new names.

Fixes #5285


alexreinking:
By the way, I think that GHA would be the _ideal_ place to test these packages if we decide only to support them on common hosts (ie. i686 and/or amd64). It should build Halide in one step, save the binary packages as an artifact, spawn a new, clean VM, and then try installing the package and building / running the apps in a second step.

steven-johnson:
> By the way, I think that GHA would be the _ideal_ place to test these packages if we decide only to support them on common hosts (ie. i686 and/or amd64).

OK for now, though I predict that it will be desirable to have aarch64 support for Nuget (at least) in the medium-term future.

steven-johnson:
> I have created a PPA we can use here: https://launchpad.net/~alexreinking/+archive/ubuntu/halide (we might want a more-official one, but the setup involves creating GPG keys and tying it to an email address, so that was difficult to do without coordination).

Definitely worth looking into a location we'd be happy(er) to support longterm, but that can wait till everything else is done. (Something at `halide-lang.org` seems the obvious option.)

packaging/CMakeLists.txt
```diff
@@ -203,28 +218,38 @@ if (WITH_DOCS)
...
 set(CPACK_PACKAGE_VENDOR Halide)
-set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Halide compiler and libraries")
 set(CPACK_PACKAGE_INSTALL_DIRECTORY ${CPACK_PACKAGE_NAME})
 set(CPACK_PACKAGE_VERSION_MAJOR ${Halide_VERSION_MAJOR})
 set(CPACK_PACKAGE_VERSION_MINOR ${Halide_VERSION_MINOR})
 set(CPACK_PACKAGE_VERSION_PATCH ${Halide_VERSION_PATCH})
 set(CPACK_VERBATIM_VARIABLES YES)
-# set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_LIST_DIR}/Description.txt")
+set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_LIST_DIR}/Description.txt")
 # set(CPACK_RESOURCE_FILE_WELCOME "${CMAKE_CURRENT_LIST_DIR}/Welcome.txt")
```
steven-johnson, line 232: nit: I know this was already commented out, but adding a comment about why is always good

packaging/Description.txt
```diff
@@ -0,0 +1,12 @@
+ Halide is a programming language designed to make it easier to write
+ high-performance image and array processing code on modern machines. Halide
+ currently targets:
+
+ * CPU architectures: X86, ARM, MIPS, Hexagon, PowerPC, RISC-V
```
steven-johnson, line 5: We should probably add WebAssembly to this list.

src/autoschedulers/adams2019/autotune_loop.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
steven-johnson, line 1: Does this specifically require `bash` (vs just `sh`)? OSX machines now default to `zsh`.

steven-johnson (commented):
LGTM.

packaging/Description.txt
alexreinking: Sure. I just copied it from the website.

src/autoschedulers/adams2019/autotune_loop.sh
alexreinking: It definitely _looks_ like bash to me. And it's not a valid executable script without a shebang, anyway, so something has to go up there. macOS still has bash installed and will for a while.

alexreinking:
The changelist was getting crazy, so I completely redid the history so you can understand the different parts of this PR better. Some things&mdash;like the spelling fixes and `#!` in `autotune_loop.sh`&mdash;could be split into a separate PR if that would make reviewing this easier. Those changes were made, however, because `lintian` discovered them.

alexreinking:
Running into this with my cross-compiling efforts. It would be nice if we had some MinGW compatibility again... https://metricpanda.com/rival-fortress-update-45-dealing-with-__chkstk-__chkstk_ms-when-cross-compiling-for-windows/

steven-johnson:
> It would be nice if we had some MinGW compatibility again

Gonna get a hard no on that from me. MinGW is a big pain to support and appears to be an environment rapidly losing mindshare.

alexreinking:
Okay, this is finally ready!

packaging/ubuntu/package.sh
```diff
@@ -0,0 +1,46 @@
...
+
+group "Create Ubuntu packages"
+cd "$halide_build_root"
+rm -rf ./_CPack_Packages ./*.deb lintian.log
+umask 0022
+export LD_LIBRARY_PATH="$halide_build_root/shared-Release/src"
+
+cpack -G DEB -C Release --config "$halide_source/packaging/ubuntu/config.cmake"
+
+# Lintian: some love it, some hate it---but everyone listens.
```
steven-johnson, line 37: This comment seems odd and unnecessary

steven-johnson (approved):

packaging/ubuntu/package.sh
alexreinking: Removed.

packaging/ubuntu/config.cmake
```diff
@@ -0,0 +1,137 @@
...
+# The default here is the host system architecture. It will generally be best
+# to package for ARM on ARM, for x86 on x86, etc. The documentation gets the
+# pseudo-architecture "all" to indicate that it has no binaries (ie. is arch
+# independent).
+unset(CPACK_DEBIAN_PACKAGE_ARCHITECTURE)
+set(CPACK_DEBIAN_HALIDE_DOCUMENTATION_PACKAGE_ARCHITECTURE all)
+
+# Package dependencies.
+# TODO: figure out how to get LLVM major version piped in here.
+set(CPACK_DEBIAN_HALIDE_RUNTIME_PACKAGE_DEPENDS "llvm-11 (>= 11.0.0)")
```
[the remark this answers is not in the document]
alexreinking: Ubuntu 20.04 LTS doesn't package LLVM 12. Maybe it would work to set up LLVM's apt repo, but I think this work can be done in a separate PR.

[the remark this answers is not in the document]
alexreinking: It might also be worth backporting these changes to v11 to provide Halide 11 + LLVM 11 packages on Ubuntu 20.04.

Issue #5204, opened 2020-08-21
Explicit control over loop partitioning


abadams:
Loop partitioning should be controllable with a scheduling directive. It sucks when I want to partition an outer loop over y and have tiles near the edge computed using separate code from tiles near the middle, and instead some inner loop over a convolution kernel gets split. The reverse is annoying too. I should be able to say which loop is going to be partitioned if any.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** The request was for a scheduling directive to say which loop, if any, gets partitioned. `Func`/`Stage::partition(var, Partition)` plus the `never_partition`/`always_partition` conveniences now do exactly that.

**Evidence:** `src/Func.h:354-358` and `src/Func.h:1543-1582` declare `partition(const VarOrRVar &, Partition)`, `never_partition(...)`, `never_partition_all()`, `always_partition(...)`, `always_partition_all()`; the policy is stored per-dim at `src/Schedule.h:463` (`Partition partition_policy`). Added by `186510173` (#7914) and extended by `8c28a73d3` (#7927). Tests: `test/correctness/partition_loops.cpp`, `test/error/bad_partition_always_throws.cpp`.

Closing: implemented in #7914 (commit `186510173`). `Func`/`Stage::partition(var, Partition::Always|Never|Auto)` plus `never_partition`/`always_partition`/`*_all` (`src/Func.h:1543-1582`) give per-loop control over loop partitioning.

Issue #1820, opened 2017-02-08
Sliding window / storage folding can't express rotating through a set of registers


abadams:
The following code should be great, because g's values should rotate through 8 registers instead of ever touching memory. However sliding window can't slide over multiple nested loops at once (x, xi in this case), which is necessary to make this work.

    g(x) = x;
    f(x) = g(x-2) + g(x+2);

    g.fold_storage(x, 8).store_root().compute_at(f, xi);
    f.split(x, x, xi, 8, TailStrategy::RoundUp).unroll(xi);

abadams:
Possible solution: the bulk of the sliding window logic just tries to slide over a var which is known to be monotonically stepping forwards and has a .min and .max symbol in scope. This property holds for many of the original Vars and intermediate Dims in a consumer's splits, not just the final Dims in the dim list. We could generate and order candidate vars to slide over by inspecting the consumer's schedule, rather than by finding serial for loops while inspecting the Stmt.

jrk:
This feels like a more-right way of thinking about sliding window in general. It seems like it also points towards a model which would make it much saner to explicitly specify sliding in the schedule.

(We can discuss in person Monday.)

pranavb-ca:
@abadams - This is a little confusing when it relates to vectorization.
For example, for the following code, I am actually seeing sliding window being able to slide max_y but not fold storage. Am I right in expecting this schedule to trigger storage folding? max_y is computed for every 'x' of output and for every 'x' of output 3 values of max_y are needed that can be computed in a circular buffer of size 3
```
    void generate() {
        max_y(x, y) = max(input(x, y-1), input(x, y), input(x, y+1));
        output(x, y) = max(max_y(x-1, y), max_y(x, y), max_y(x+1, y));
    }
    void schedule() {
        Var xi{"xi"}, yi{"yi"};

        if (get_target().features_any_of({Target::HVX_64, Target::HVX_128})) {
            const int vector_size = get_target().has_feature(Target::HVX_128) ? 128 : 64;
            max_y
              .store_at(Func(output), y)
              .compute_at(Func(output), x)
              .vectorize(x, vector_size, TailStrategy::RoundUp);
            Func(output)
              .vectorize(x, vector_size, TailStrategy::RoundUp);
        } else {
            const int vector_size = natural_vector_size<uint8_t>();
            Func(output)
                .vectorize(x, vector_size)
                .parallel(y, 16);
        }
    }
```

The following is the IR after sliding window
```
[...]

PR #5755, opened 2021-02-18
Change warmup strategy for sliding window

dsharletg:
This branch changes how we handle the "warmup" of sliding windows before running the "steady state". The old strategy was to make the first iteration handle the warmup. This branch changes the strategy to "back up" the loop so the steady state warms up the loop itself, inserting the necessary ifs to guard the producer and consumer to avoid producing or consuming out of bounds values.

This change has the following impacts:
- It is now possible to generate much finer grained sliding window schedules with low overhead. The 'blur' app in this branch is ~30% faster than master.
- Some sliding window schedules appear to get possibly a bit faster in some test cases (local_laplacian, stencil_chain, camera_pipe).
- Folded storage buffers can be smaller, because the folded storage only needs to accommodate the steady state of the loop. The warmups in the old strategy needed larger folded storage windows. See the changes to camera_pipe and local_laplacian for example. This should make storage folding less confusing (the reason for the larger factors was confusing and seemed unnecessary).

This avoids one of the motivations for #5204, and is hopefully helpful for #1820


steven-johnson (approved):
This looks pretty cool. Only a couple of comments from me:
- there are a couple of new TODOs inserted that aren't clear to me regarding they are needs-more-attention-now vs needs-more-attention-later vs ok-for-now-but-possible-future-work. Might be good to expand the comments (and/or add tracking issues if needed.)
- I suspect this should be tagged with the `release_notes` label, so that people can be made aware that their sliding window schedules might warrant attention after this change.

src/Monotonic.cpp
```diff
@@ -30,26 +30,172 @@ using std::string;
...
+
+ConstantInterval unify(const ConstantInterval &a, int64_t b) {
+    ConstantInterval result;
+    result.include(b);
+    return result;
+}
+
+// Helpers for doing arithmetic on ConstantIntervals that avoid generating
+// expressions of pos_inf/neg_inf.
+ConstantInterval add(const ConstantInterval &a, const ConstantInterval &b) {
```
abadams, line 83: @rootjalex could you check these helpers to see if you think they're correct?

src/Monotonic.cpp
```diff
@@ -30,26 +30,172 @@ using std::string;
...
+        result.min *= b;
+    }
+    if (result.has_upper_bound()) {
+        result.max *= b;
+    }
+    return result;
+}
+
+ConstantInterval multiply(const ConstantInterval &a, const ConstantInterval &b) {
+    std::vector<int64_t> bounds;
```
abadams, line 133: I wonder if the heap allocation here is a bottleneck. This seems like it could be faster with a stack array and a pointer to the end element. (e.g. *end++ = a.min * b.max)

src/Monotonic.cpp
```diff
@@ -67,135 +213,112 @@ class MonotonicVisitor : public IRVisitor {
...
+            ConstantInterval ra = result;
+            op->b.accept(this);
+            ConstantInterval rb = result;
+
+            if (const int64_t *b = as_const_int(op->b)) {
+                result = divide(ra, *b);
+            } else if (const uint64_t *b = as_const_uint(op->b)) {
+                result = divide(ra, *b);
+            } else if (const int64_t *a = as_const_int(op->a)) {
+                result = divide(rb, *a);
```
abadams, line 282: I'm confused. Does this trigger for 2/x ?

src/Monotonic.cpp
```diff
@@ -209,10 +332,16 @@ class MonotonicVisitor : public IRVisitor {
...
         a.accept(this);
-        Monotonic ra = result;
+        ConstantInterval ra = result;
         b.accept(this);
-        Monotonic rb = result;
-        result = unify(flip(ra), rb);
+        ConstantInterval rb = result;
+        result = unify(negate(ra), rb);
+        if (result.has_lower_bound()) {
+            result.min = std::max<int64_t>(result.min, -1);
```
abadams, line 340: Comments here would be good. E.g. "It's possible for the value to flip from true to false, which is a step of -1, because ..."

src/Monotonic.cpp
```diff
@@ -233,71 +362,58 @@ class MonotonicVisitor : public IRVisitor {
...
+            ConstantInterval rcond = result;
+
+            op->true_value.accept(this);
+            ConstantInterval ra = result;
+            op->false_value.accept(this);
+            ConstantInterval rb = result;
+            ConstantInterval unified = unify(ra, rb);
+
+            // TODO: How to handle unsigned values?
+            Expr delta = simplify(op->true_value - op->false_value);
```
abadams, line 397: Calling simplify in this visitor is probably a bad idea. Can it be avoided?

src/Monotonic.cpp
```diff
@@ -331,27 +447,27 @@ class MonotonicVisitor : public IRVisitor {
...
             return;
         }
 
         for (size_t i = 0; i < op->args.size(); i++) {
             op->args[i].accept(this);
-            if (result != Monotonic::Constant) {
+            if (!is_constant(result)) {
                 // One of the args is not constant.
-                result = Monotonic::Unknown;
+                result = ConstantInterval();
```
abadams, line 460: ConstantInterval::everything() would be clearer here and elsewhere

src/Prefetch.cpp
```diff
@@ -194,12 +194,18 @@ class InjectPlaceholderPrefetch : public IRMutator {
         Stmt body = mutate(op->body);
 
         if (!prefetch_list.empty() && starts_with(op->name, prefix)) {
+            // Remove ".$n", added by sliding window.
```
abadams, line 197: I still think it would be better for sliding window to rewrite the prefetchdirectives to use the new loop names

src/SimplifyCorrelatedDifferences.cpp
```diff
@@ -118,7 +118,7 @@ class SimplifyCorrelatedDifferences : public IRMutator {
             tmp_lets.swap(lets);
             loop_var = op->name;
             {
-                ScopedBinding<Monotonic> bind(monotonic, loop_var, Monotonic::Increasing);
+                ScopedBinding<ConstantInterval> bind(monotonic, loop_var, ConstantInterval(1, 1));
```
abadams, line 121: ConstantInterval::single_point(1)

src/Simplify_Div.cpp
```diff
@@ -177,9 +177,11 @@ Expr Simplify::visit(const Div *op, ExprInfo *bounds) {
                rewrite((w + (z + (x * c0 + y))) / c1, (y + z + w) / c1 + x * fold(c0 / c1), c0 % c1 == 0 && c1 > 0) ||
                rewrite((w + (z + (y + x * c0))) / c1, (y + z + w) / c1 + x * fold(c0 / c1), c0 % c1 == 0 && c1 > 0) ||
 
-               // Finally, pull out constant additions that are a multiple of the denominator
-               rewrite((x + c0) / c1, x / c1 + fold(c0 / c1), c0 % c1 == 0 && c1 > 0) ||
-               rewrite((c0 - y)/c1, fold(c0 / c1) - y / c1, (c0 + 1) % c1 == 0 && c1 > 0) ||
+               // Finally, pull out additions that are a multiple of the denominator
+               // TODO: I think this rule can be stronger. We should be able to
+               // rewrite (x + 1) / 2 to x / 2 + 1 when x we know x % 2 == 1.
+               rewrite((x + c0) / c1, x / c1 + fold(c0 / c1), c1 > 0 && (c0 % c1 == 0 || can_prove(x % c1 == 0, this))) ||
```
abadams, line 183: Let's not introduce new can_prove uses to simplifier predicates

src/Simplify_LT.cpp
```diff
@@ -68,10 +68,18 @@ Expr Simplify::visit(const LT *op, ExprInfo *bounds) {
 
         // clang-format off
         if (rewrite(broadcast(x, c0) < broadcast(y, c0), broadcast(x < y, c0)) ||
+
+            // We can learn more from equality than less with mod.
+            rewrite(x % y < 1, x % y == 0) ||
+            rewrite(0 < x % y, x % y != 0) ||
+            rewrite(x % c0 < c1, x % c0 != fold(c0 - 1), c1 + 1 == c0) ||
```
abadams, line 75: Not convinced this is correct when c0 < 0, or the next rule is correct when c1 < 0

src/Simplify_LT.cpp
```diff
@@ -68,10 +68,18 @@ Expr Simplify::visit(const LT *op, ExprInfo *bounds) {
...
+            rewrite(c0 < x % c1, x % c1 == fold(c1 - 1), c0 + 2 == c1) ||
+
             (no_overflow(ty) && EVAL_IN_LAMBDA
              (rewrite(ramp(x, y, c0) < ramp(z, y, c0), broadcast(x < z, c0)) ||
               // Move constants to the RHS
               rewrite(x + c0 < y, x < y + fold(-c0)) ||
+              rewrite(c0 < -x, x < fold(-c0)) ||
```
abadams, line 82: -x is represented as 0 - x, so maybe this rule should be rewrite(c0 < c1 - x, x < fold(c1 - c0)) instead, as it's more general.

src/SlidingWindow.cpp
```diff
@@ -78,10 +84,73 @@ class ExpandExpr : public IRMutator {
...
+
+    Stmt visit(const Provide *op) override {
+        if (op->name != func.name()) {
+            return op;
+        }
+        internal_assert(dim_idx < (int)op->args.size());
+        Expr var = op->args[dim_idx];
+        Expr guard = const_true();
+        if (min.defined()) {
+            guard = guard && likely_if_innermost(var >= min);
```
abadams, line 136: The next simplifier pass is going to rewrite this to min <= var, so may as well write it that way.

src/SlidingWindow.cpp
```diff
@@ -245,21 +315,53 @@ class SlidingWindowOnFunctionAndLoop : public IRMutator {
...
+            if (can_slide_up) {
+                new_loop_min_eq =
+                    substitute(loop_var, loop_min, min_required) == substitute(loop_var, new_loop_min_var, prev_max_plus_one);
+            } else {
+                new_loop_min_eq =
+                    substitute(loop_var, loop_min, max_required) == substitute(loop_var, new_loop_min_var, prev_min_minus_one);
+            }
+            // Ignore unsafe promises (intended for the ones generated by
+            // TailStrategy::GuardWithIf, but may be relevant in other cases).
+            new_loop_min_eq = lower_safe_promises(new_loop_min_eq);
```
abadams, line 330: Would it be better to get the solver to look through promises in the same was as it looks through likely and likely_if_innermost?

src/SlidingWindow.cpp
```diff
@@ -292,7 +394,31 @@ class SlidingWindowOnFunctionAndLoop : public IRMutator {
...
+
             return stmt;
+        } else if (!find_produce(op, func.name())) {
+            // The producer might have expanded the loop before the min to warm
+            // up the window. This consumer doesn't contain a producer that might
+            // be part of the warmup, so guard it with an if to only run it on
+            // the original loop bounds.
+            Expr loop_var_expr = Variable::make(Int(32), loop_var);
+            Expr orig_loop_min_expr = Variable::make(Int(32), loop_var + ".loop_min.orig");
+            Expr guard = likely_if_innermost(loop_var_expr >= orig_loop_min_expr);
```
abadams, line 412: Prefer <= to >= for the same reason as above, unless you think it makes it confusing to read to the degree that it's worth a few extra nanoseconds of compile time.

src/SlidingWindow.cpp
```diff
@@ -343,43 +469,83 @@ class SlidingWindowOnFunctionAndLoop : public IRMutator {
...
+    DependsOn check(a, b);
+    s.accept(&check);
+    return check.yes;
+}
+
 // Perform sliding window optimization for all functions
 class SlidingWindow : public IRMutator {
     const map<string, Function> &env;
 
+    list<Function> sliding;
```
abadams, line 547: A comment here on what this holds would be nice. It is the realize nodes currently in scope for things we intend to slide, from innermost to outermost?

src/SlidingWindow.cpp
```diff
@@ -415,16 +579,112 @@ class SlidingWindow : public IRMutator {
...
+                Expr new_loop_min = slider.new_loop_min;
+                if (!sliding_loop_min.same_as(loop_min)) {
+                    // If we didn't start from the loop min, take the union
+                    // of the new loop min and the loop min.
+                    new_loop_min = min(new_loop_min, loop_min);
+                }
+                string new_name = name + ".$n";
+                loop_min = Variable::make(Int(32), new_name + ".loop_min");
+                loop_extent = Variable::make(Int(32), new_name + ".loop_extent");
+                body = substitute({
```
abadams, line 630: Here's where we'd have to rewrite prefetch directives. Should be a simple inline IRMutator. I think it's better than relying on an exact match to the string suffix above.

src/SlidingWindow.cpp
```diff
@@ -415,16 +579,112 @@ class SlidingWindow : public IRMutator {
...
+        }
+    }
+
 public:
     SlidingWindow(const map<string, Function> &e)
         : env(e) {
     }
 };
 
+class AddLoopMinOrig : public IRMutator {
```
abadams, line 666: Comment on what this mutator does and why would be nice

src/StorageFolding.cpp
```diff
@@ -789,22 +773,18 @@ class AttemptStorageFoldingOfFunction : public IRMutator {
...
-
-                    Expr fudge = simplify(substitute(op->name, loop_min, extent - to_acquire));
-                    if (is_const(fudge) && can_prove(fudge <= sema.init)) {
-                        sema.init -= fudge;
-                    } else {
-                        to_acquire = select(loop_var > loop_min, likely(to_acquire), extent);
+                    if (provided.used.defined()) {
+                        to_acquire = select(provided.used, to_acquire, 0);
                     }
+                    if (required.used.defined()) {
```
abadams, line 779: Don't you explicitly undefine required.used above?

src/VectorizeLoops.cpp
```diff
@@ -118,6 +118,25 @@ Interval bounds_of_lanes(const Expr &e) {
         } else if (is_negative_const(r->stride)) {
             return {r->base + last_lane_idx * r->stride, r->base};
         }
+    } else if (const LE *le = e.as<LE>()) {
```
abadams, line 121: @rootjalex would also appreciate you checking this bounds-related logic. I wrote this code so I'm not well positioned to review it.

src/VectorizeLoops.cpp
```diff
@@ -839,7 +858,11 @@ class VectorSubs : public IRMutator {
                 // *every* likely value is true. We can do that by
                 // generating a scalar condition that checks if
                 // the least-true lane is true.
-                Expr all_true = bounds_of_lanes(likely->args[0]).min;
+                Expr all_true = likely->args[0];
+                while (!all_true.type().is_scalar()) {
```
abadams, line 862: Wait, what? bounds_of_lanes only unpeels one layer of vector? We should just fix it instead of having a while loop around it.

abadams:
This very large change to how we lower sliding touches way fewer lines of code than I feared.

src/Monotonic.cpp
dsharletg: I think it might be necessary. FWIW the old logic used 2 can_prove calls on similar expressions, so this seems like an improvement...

src/StorageFolding.cpp
dsharletg: Yes, it's weird, added comment.

src/Monotonic.cpp
abadams: Fair enough, I guess you need it to fold to a constant.

src/StorageFolding.cpp
dsharletg: Actually, I removed it. I was going to comment something to the effect of "this is correct, even though it's always a no-op". I think now it's actually wrong, for the same reason we clear required.used above, we still need to release the folded storage buffer even if we didn't use it.

dsharletg:
Thanks for the review, and thanks for the help (and @vksnk too) for getting this working!

src/Interval.cpp
```diff
@@ -157,5 +157,115 @@ Expr Interval::neg_inf_noinline() {
...
+
+bool ConstantInterval::operator==(const ConstantInterval &other) const {
+    if (min_defined != other.min_defined || max_defined != other.max_defined) {
+        return false;
+    }
+    return (!min_defined || min == other.min) && (!max_defined || max == other.max);
+}
+
+void ConstantInterval::include(const ConstantInterval &i) {
+    if (max_defined && i.max_defined) {
```
rootjalex, line 226: `ConstantInterval::nothing() .include(ConstantInterval(-2, -1));` Would produce the interval `[-2, 0]`, right? That doesn't seem like correct behavior.
Likewise, `ConstantInterval::nothing() .include(ConstantInterval(2, 3));` -> `[1, 3]`

src/Interval.cpp
```diff
@@ -157,5 +157,115 @@ Expr Interval::neg_inf_noinline() {
...
+        max_defined = false;
+    }
+    if (min_defined && i.min_defined) {
+        min = std::min(min, i.min);
+    } else {
+        min_defined = false;
+    }
+}
+
+void ConstantInterval::include(int64_t x) {
```
rootjalex, line 238: Similar issues here, with empty intervals.
`ConstantInterval::nothing().include(5);` -> `[1, 5]`
`ConstantInterval::nothing().include(-5);` -> `[-5, 0]`

src/Interval.cpp
```diff
@@ -157,5 +157,115 @@ Expr Interval::neg_inf_noinline() {
...
+    }
+}
+
+ConstantInterval ConstantInterval::make_union(const ConstantInterval &a, const ConstantInterval &b) {
+    ConstantInterval result = a;
+    result.include(b);
+    return result;
+}
+
+ConstantInterval ConstantInterval::make_intersection(const ConstantInterval &a, const ConstantInterval &b) {
```
rootjalex, line 253: Pretty confident empty intervals would cause issues here too

abadams:
One current issue I should note here is that this doubles the allocation size for a bunch of stages in local laplacian. Also causes a 10% runtime regression on my machine, that seems to vary by machine. The increase in allocation size should be repeatable though.

src/Monotonic.cpp
rootjalex: At first glance, I see a lot of issues if you're using empty intervals with these, because they have defined bounds. i.e. `negate(ConstantInterval::nothing())` -> `[-1, 0]`. I'd like to formally verify some of these, but the problems with empty intervals should probably be fixed first.

src/Monotonic.cpp
```diff
@@ -30,26 +30,173 @@ using std::string;
...
+    return result;
+}
+
+ConstantInterval multiply(const ConstantInterval &a, const ConstantInterval &b) {
+    int64_t bounds[4];
+    int64_t *bounds_begin = &bounds[0];
+    int64_t *bounds_end = &bounds[0];
+    ConstantInterval result;
+    result.min_defined = result.max_defined = true;
+    if (a.has_lower_bound() && b.has_lower_bound()) {
```
rootjalex, line 138: This series of rules seems more confusing than it needs to be - the result never can produce a partially bounded value.

i.e. for `result.min_defined` to be true, we need `a.has_lower_bound() && b.has_upper_bound()` and `a.has_upper_bound() && b.has_lower_bound()`, and similarly `a.has_lower_bound() && b.has_lower_bound()` and `a.has_upper_bound() && b.has_upper_bound()` for `result.max_defined` to be true. At this point, if `a` and `b` are both bounded, then compare all 4 values and choose the mins/maxs for the same result.

src/Interval.cpp
```diff
@@ -157,5 +157,115 @@ Expr Interval::neg_inf_noinline() {
...
+
+ConstantInterval::ConstantInterval(int64_t min, int64_t max)
+    : min(min), max(max), min_defined(true), max_defined(true) {
+}
+
+ConstantInterval ConstantInterval::everything() {
+    return ConstantInterval();
+}
+
+ConstantInterval ConstantInterval::nothing() {
```
rootjalex, line 170: Hmm, I've left a couple of comments talking about issues with this, but I don't think I actually see this function called/used anywhere. Probably is worth removing (along with `ConstantInterval::is_empty()`)

src/Monotonic.cpp
dsharletg: Agreed this was overly complicated. I think with the refactored representation, it's much easier.

Are you sure that the result is never a partially bounded interval? What about `[1, inf) * [0, 1]` ? Shouldn't this be `[0, inf)` ?

dsharletg (commented):
Thanks for the careful look at that interval logic. Empty intervals were definitely broken most of the time.

dsharletg:
Actually this new representation is terrible. I need to rewrite it again...

src/Monotonic.cpp
```diff
@@ -67,135 +214,109 @@ class MonotonicVisitor : public IRVisitor {
...
-            result = flip(rb);
-        } else if (is_negative_const(op->b)) {
-            result = flip(ra);
+        if (op->type.is_scalar()) {
+            op->a.accept(this);
+            ConstantInterval ra = result;
+            op->b.accept(this);
+            ConstantInterval rb = result;
+
+            if (const int64_t *b = as_const_int(op->b)) {
```
rootjalex, line 255: Why isn't `ConstantInterval multiply(const ConstantInterval &a, const ConstantInterval &b)` used here?

src/Interval.cpp
dsharletg: This is what I ended up doing, it avoids a lot of the other problems you pointed out.

src/Monotonic.cpp
dsharletg: b is an Expr, not a ConstantInterval, and DerivativeBounds loses information. I think correct thing to do here is something like the product rule for derivatives:

    a * DerivativeBounds(b) + b * DerivativeBounds(a)

But now that the intervals are ConstantIntervals, that's kind of a pain to implement, and I believe it reduces to the same thing as the logic here (when a or b are constants).

src/Monotonic.cpp
rootjalex: Oh sorry, I should have said that better. The result of a multiply could absolutely be a partially bounded interval, like your example - but I don't think that this function ever produces a partially bounded interval.

rootjalex: If I run:
```
    ConstantInterval a = ConstantInterval(0, 1);
    ConstantInterval b = ConstantInterval::bounded_below(1);
    ConstantInterval c = multiply(a, b);
```
On this branch, I produce an interval that has undefined min and max

rootjalex: Oh oops, I see there was a bit of a refactor, that might no longer be the case.

rootjalex: Now this example no longer works because:
```
ConstantInterval::bounded_below(1);
```
Throws an internal assert, because `bounded_below` seems to assume a negative argument, is this true?

rootjalex: Ok so if I make `bounded_below` use `std::numeric_limits<int64_t>::max()` in the ctor to avoid that error, this example still produces a `c` that has undefined min and max.

src/VectorizeLoops.cpp
rootjalex: Yeah, both of these cases follow the same logic as bounds inference, which has been verified.

@dsharletg is there a reason for the requirement that one of the two operands be a single point that I am missing? For inequalities, there are partial bounds that can be produced relatively easily.

rootjalex: Nevermind, for the exact bounds requirement I think this is probably the best that can be done.

dsharletg: I am not sure, @abadams added this. The logic looks like it would be correct without one side being a broadcast, but it wouldn't be exact, which the comment implies is a requirement.

rootjalex: Yeah sorry, I should have read the top-level comment for `bounds_of_nested_lanes` to realize that this needed to be exact, my bad. I believe this is correct for the specification.

src/Interval.h
```diff
@@ -110,6 +110,69 @@ struct Interval {
...
+
+    /** Construct an interval representing a single point. */
+    static ConstantInterval single_point(int64_t x);
+
+    /** Construct intervals bounded above or below. */
+    static ConstantInterval bounded_below(int64_t min);
+    static ConstantInterval bounded_above(int64_t max);
+
+    /** Is the interval the empty set */
+    bool is_empty() const;
```
rootjalex, line 138: This should be removed

rootjalex: (I think?)

src/Monotonic.cpp
rootjalex: The example now works, and I *think* the casework is correct. I haven't found a prettier way to avoid the series of cases yet either..

dsharletg:
@abadams this is now a (very) slight improvement in memory usage of local_laplacian :)

src/Interval.cpp
```diff
@@ -157,5 +157,108 @@ Expr Interval::neg_inf_noinline() {
...
+    result.min_defined = false;
+    return result;
+}
+
+bool ConstantInterval::is_everything() const {
+    return !min_defined && !max_defined;
+}
+
+bool ConstantInterval::is_single_point() const {
+    return !is_everything() && min == max;
```
rootjalex, line 192: I don't think `!is_everything()` is sufficient, because any partially bounded interval with the same `min` and `max` will be defined as a single point.

i.e.:
```
ConstantInterval c = ConstantInterval::bounded_below(4);
c.is_single_point(); // -> true
```
Pretty sure it should be:
```
min_defined && max_defined && min == max;
```

src/Interval.cpp
```diff
@@ -157,5 +157,108 @@ Expr Interval::neg_inf_noinline() {
...
+}
+
+bool ConstantInterval::is_single_point(int64_t x) const {
+    return !is_everything() && min == x && max == x;
```
rootjalex, line 196: Same issue here

src/Interval.cpp
```diff
@@ -157,5 +157,108 @@ Expr Interval::neg_inf_noinline() {
...
+bool ConstantInterval::has_upper_bound() const {
+    return max_defined;
+}
+
+bool ConstantInterval::has_lower_bound() const {
+    return min_defined;
+}
+
+bool ConstantInterval::is_bounded() const {
+    return !is_everything();
```
rootjalex, line 208: This definition is not the same as `Interval`'s definition, which is `is_bounded -> has_upper_bound() && has_lower_bound()`, which could be confusing. However, I don't think I've found a case where this method is called for a `ConstantInterval`, so maybe it could just be removed

src/Monotonic.cpp
```diff
@@ -233,71 +385,58 @@ class MonotonicVisitor : public IRVisitor {
...
+            // TODO: How to handle unsigned values?
+            Expr delta = simplify(op->true_value - op->false_value);
+            delta.accept(this);
+            ConstantInterval rdelta = result;
+
+            ConstantInterval adjusted_delta;
+            if (const int64_t *const_delta = as_const_int(delta)) {
+                adjusted_delta = multiply(rcond, *const_delta);
+            } else {
+                adjusted_delta = multiply(rcond, rdelta);
```
dsharletg, line 428: I think this is wrong. An example where it produces an incorrect result is: `select(0 < x, max(min(x, 4), 3), 4)`

I was hoping to avoid bringing back the old complicated logic... but I haven't yet figured out a cleaner way to handle this.

dsharletg: I think we don't want rdelta here, I think we actually want bounds_of_expr(delta) instead. i.e. it should be the actual delta, not the derivative of the delta.

rootjalex: I never reviewed this part because I didn't really understand it all that much, but I would like to understand better. Could you explain the comment "// The result is the unified bounds, added to the "bump" that happens when switching from true to false." ? Why is the bump added, and not just a unification like with min and max?

dsharletg: My reasoning is: the bounds of the finite difference is the union of the bounds of the finite difference of the true value and the false value, plus the step function that occurs when switching from false to true.

src/Simplify.cpp
```diff
@@ -278,17 +286,45 @@ void Simplify::ScopedFact::learn_true(const Expr &fact) {
...
         // Both must be true
         learn_true(a->a);
         learn_true(a->b);
+        return;
     } else if (const Not *n = fact.as<Not>()) {
         learn_false(n->a);
-    } else if (simplify->truths.insert(fact).second) {
+        return;
+    }
+    if (simplify->truths.insert(fact).second) {
```
abadams, line 303: Adding compound facts in addition to the leaves looks like it adds quite a lot of redundant facts in the case where you have something like a && b && c && d && e

dsharletg: Note I also added an early return in the And case above, so it shouldn't do that in this case. I made this change so we could learn LE/LT/EQ in addition to the bounds/alignment we learn from those.

src/Simplify_Min.cpp
```diff
@@ -96,6 +98,7 @@ Expr Simplify::visit(const Min *op, ExprInfo *bounds) {
                rewrite(min(x, ((x + c0)/c1)*c1 + c2), a, c1 > 0 && c0 + c2 >= c1 - 1) ||
                rewrite(min(((x + c0)/c1)*c1 + c2, x), a, c1 > 0 && c0 + c2 <= 0) ||
                rewrite(min(x, ((x + c0)/c1)*c1 + c2), b, c1 > 0 && c0 + c2 <= 0) ||
+               rewrite(min((x/c0)*c0, (x/c1)*c1 + c2), a, c2 >= c1 && c1 > 0) ||
```
rootjalex, line 101: Is this true? What if `x` is negative?

rootjalex: I think:  x = -2, c0 = 0, c1 = 1, and c2 = 1 is a counterexample, produced by z3.

dsharletg: I think x negative is fine, but it looks like this needs a condition that c0 > 0

rootjalex: Yeah, that works for z3

abadams: This verified as correct for me... there's an implicit assumption you can make that c0 != 0. Is that enough?

abadams: The LHS is always less than x, even for negative c0, and the RHS is always greater than x

rootjalex: `c0 != 0` also works for my verifier. @abadams are you using the super_simplifier stuff? I have a python translator that's just easy for me to use, but it doesn't seem to understand the implicit assumption c0 != 0, where does that come from?

abadams: https://github.com/halide/Halide/blob/super_simplify_v2/apps/super_simplify/filter_rewrite_rules.cpp#L215

rootjalex: Oops okay maybe I should use that code instead. Is this predicate implicit simply because it should have been removed by an earlier rewrite rule already?

abadams: Yeah, we can assume constant folding and canonicalization rules (e.g. in x + y, x is not a constant) run before more general rules.

src/SlidingWindow.cpp
```diff
@@ -321,7 +504,10 @@ class SlidingWindowOnFunctionAndLoop : public IRMutator {
     }
 
     Stmt visit(const LetStmt *op) override {
-        ScopedBinding<Expr> bind(scope, op->name, simplify(expand_expr(op->value, scope)));
+        Interval bounds_value = bounds_of_expr_in_scope(op->value, bounds, empty_func_value_bounds(), true);
```
abadams, line 507: I think this is just ```find_constant_bounds(op->value, bounds)```

src/SlidingWindow.cpp
```diff
@@ -415,16 +689,133 @@ class SlidingWindow : public IRMutator {
...
+        slid_dimensions = old_slid_dimensions;
+        if (then_case.same_as(op->then_case) && else_case.same_as(op->else_case)) {
+            return op;
+        } else {
+            return IfThenElse::make(op->condition, then_case, else_case);
+        }
+    }
+
+    Stmt visit(const LetStmt *op) override {
+        Interval bounds_value = bounds_of_expr_in_scope(op->value, bounds, empty_func_value_bounds(), true);
```
abadams, line 784: Same here

dsharletg:
After a shocking amount of debugging whack-a-mole, I think this branch is finally fully working on both Halide and our production code and ready to be merged after the new code is reviewed.

I needed to bring back the old sliding window logic (an explicit warmup on the first iteration) because the new logic is not always possible (for example, sliding computed at a func that has a boundary condition on the slid producer). This old logic is used when a safety check for this fails.

Unfortunately, this safety check is quite difficult to prove except in easy cases. So while this new branch does work for things like the blur app, which opens up some interesting possibilities, the new logic currently doesn't pass the safety check in cases I would like it to (e.g. camera_pipe).

I still think this is a good change to make, because we can enable the new logic i n more cases in subsequent work just by making that proof easier, or improving the capabilities of the simplifier to handle the proofs.

dsharletg:
I found a better condition that is easier to prove, so most of the issues I mentioned in my previous comment are now fixed :)

dsharletg:
I'm going to close this and reopen it after peeling off some of these changes into their own separate PRs.

dsharletg:
Reopening now that those two prerequisite PRs have been merged.

steven-johnson:
Looks clean. Ready to land?

dsharletg:
I think this is ready to merge, @abadams, @rootjalex did you want to take a final look at this before merging?

src/FuseGPUThreadLoops.cpp
```diff
@@ -359,19 +363,19 @@ class ExtractSharedAndHeapAllocations : public IRMutator {
...
-                    // somehow unbounded. It's probable that no such
-                    // expression exists. is_monotonic is generally
-                    // less capable than bounds_of_expr_in_scope.
-                    s.size = interval_bounds.max;
+                    break;
+                case Monotonic::Increasing:
+                    s.size = substitute(op->name, simplify(op->min + op->extent - 1), s.size);
+                    break;
+                case Monotonic::Constant:
+                    // The size expression used the variable, but we
```
rootjalex, line 371: In which cases is the dependence false somehow?

rootjalex: Or I guess, if there are cases, is that a bug we should be worried about?

dsharletg: @abadams added this code, I am not sure, Andrew?

src/Interval.cpp
```diff
@@ -157,5 +157,108 @@ Expr Interval::neg_inf_noinline() {
...
+    } else {
+        result.min_defined = false;
+    }
+    if (a.max_defined && b.max_defined) {
+        result.max = std::min(a.max, b.max);
+        result.max_defined = true;
+    } else {
+        result.max_defined = false;
+    }
+    return result;
```
rootjalex, line 260: I don't see any instances of `make_intersection()` being used for `ConstantInterval`s - is it needed? I know it emulates `Interval` but if it's not used I don't see a reason to keep it

src/Monotonic.cpp
```diff
@@ -30,26 +31,204 @@ using std::string;
...
+    }
+    return result;
+}
+
+ConstantInterval add(const ConstantInterval &a, int64_t b) {
+    return add(a, ConstantInterval(b, b));
+}
+
+ConstantInterval negate(const ConstantInterval &r) {
+    ConstantInterval result;
```
rootjalex, line 110: If `ConstantInterval`s should maintain the invariant that `min < max` regardless of `min_defined`/`max_defined`, then this function doesn't necessarily follow that - it could be simplified by replacing:
```
result.min = r.has_upper_bound() ? -r.max : 0;
```
with:
```
result.min = -r.max;
```
And likewise for result.max. Nothing about the ConstantInterval would change, but it would maintain the `min < max` invariant (not sure if this invariant is intended/useful, but it does remove ternary operators that are unnecessary

src/Monotonic.cpp
```diff
@@ -209,10 +364,19 @@ class MonotonicVisitor : public IRVisitor {
...
     void visit_lt(const Expr &a, const Expr &b) {
         a.accept(this);
-        Monotonic ra = result;
+        ConstantInterval ra = result;
         b.accept(this);
-        Monotonic rb = result;
-        result = unify(flip(ra), rb);
+        ConstantInterval rb = result;
+        result = unify(negate(ra), rb);
+        // If the result is bounded, limit it to [-1, 1]. The largest
```
rootjalex, line 371: Is the result limited to this interval? Could `result.min` be less than `-1`, or `result.max` be greater than `1`, which would produce an interval not limited to this range?

src/Monotonic.cpp
```diff
@@ -456,25 +615,47 @@ class MonotonicVisitor : public IRVisitor {
...
+        return Monotonic::Unknown;
+    }
+    Scope<ConstantInterval> intervals_scope;
+    for (Scope<Monotonic>::const_iterator i = scope.cbegin(); i != scope.cend(); ++i) {
+        intervals_scope.push(i.name(), to_interval(i.value()));
+    }
+    return is_monotonic(e, var, intervals_scope);
+}
+
+Monotonic is_monotonic_strong(const Expr &e, const std::string &var) {
```
rootjalex, line 655: Not sure I see the difference between `is_monotonic_strong` and `is_monotonic` with a default empty scope argument, but `is_monotonic_strong` doesn't seem to be used anywhere, so could probably be removed?

src/FuseGPUThreadLoops.cpp
abadams: expr_uses_var occurs before solve/simplify/cse, so the most common reason it might hit this path is just that the var was successfully removed by the simplifier. But I thought it was also possible that is_monotonic might return Constant, even if the var appears in the expr. I didn't want to rule out the possibility that is_monotonic might become more powerful than the simplifier's ability to cancel things in some obscure cases. An example might be something dumb like `x - likely(x)`, where you can replace `likely` with anything that's an identify function. Or perhaps an appearance of x in a place that doesn't affect the derivative, e.g: `print(37, x)`. Or maybe one day is_monotonic might realize that `select(x < 0, 0, 1) + select(0 <= x, 1, 0)` doesn't actually depend on x but the simplifier won't.

src/Monotonic.cpp
dsharletg: I don't think this is safe, because negation can overflow. 

I think we only need min < max when min_defined && max_defined. 

I think the combination of these is why there are so many ternarys where they seem possibly redundant. I think many of them are unnecessary except to avoid possible ubsan issues due to overflow.

src/Monotonic.cpp
dsharletg: the result isn't limited to this, but this logic then does limit it explicitly below.

src/Monotonic.cpp
rootjalex: Ah, that makes sense. Should we be concerned about overflow in other cases? Because I believe that `multiply(ConstantInterval, int64_t)` is passed a `uint64_t` at some point, which could be an issue (I assumed that you were assuming no overflow)

src/Monotonic.cpp
```diff
@@ -30,26 +31,204 @@ using std::string;
...
+        result.max *= b;
+    }
+    return result;
+}
+
+ConstantInterval multiply(const ConstantInterval &a, const Expr &b) {
+    if (const int64_t *bi = as_const_int(b)) {
+        return multiply(a, *bi);
+    } else if (const uint64_t *bi = as_const_uint(b)) {
+        return multiply(a, *bi);
```
rootjalex, line 145: Could be overflow here, converting `uint64_t` to `int64_t`

src/Monotonic.cpp
```diff
@@ -67,135 +246,111 @@ class MonotonicVisitor : public IRVisitor {
...
+            op->a.accept(this);
+            ConstantInterval ra = result;
+            op->b.accept(this);
+            ConstantInterval rb = result;
+
+            // This is essentially the product rule: a*rb + b*ra
+            // but only implemented for the case where a or b is constant.
+            if (const int64_t *b = as_const_int(op->b)) {
+                result = multiply(ra, *b);
+            } else if (const uint64_t *b = as_const_uint(op->b)) {
```
rootjalex, line 291: Same issue here (twice) with `uint64_t` -> `int64_t` conversions

src/Monotonic.cpp
rootjalex: I'm not convinced, what if `result = [-10, -5]` here, then after the following casework, it would be `[-1, -5]`, which is an incorrect interval.

src/FuseGPUThreadLoops.cpp
rootjalex: Ah gotcha, that makes sense.

src/Monotonic.cpp
dsharletg: I was more worried about introducing overflow from earlier implementations of is_monotonic. These ternaries don't protect against generating new overflow (just like basically all integer arithmetic).

I decided to just fix the need for a signed/unsigned case anyways, that was always super annoying.

rootjalex (approved):
The interval code lgtm

src/SlidingWindow.cpp
```diff
@@ -340,46 +488,128 @@ class SlidingWindowOnFunctionAndLoop : public IRMutator {
...
+bool depends_on(const string &a, const string &b, const Stmt &s) {
+    if (a == b) {
+        return true;
+    }
+    Dependencies deps(b);
+    s.accept(&deps);
+    // Recursively search for dependencies. Repeatedly using this on the
+    // same set of Funcs is algorithmically slow, but even an absurd number
+    // of Funcs is still relatively small...
+    for (const string &i : deps.dependencies) {
```
abadams, line 568: I think this is algorithmically slow enough to hit age of the universe for pathological pipelines. Can be fixed by adding a `map<pair<string, string>, bool>` arg and a cache check at the function entry point

abadams (approved):

Issue #5650, opened 2021-01-19
Move CodeGen_GPU_Host to a lowering pass


dsharletg:
Here's what I think we should do on this issue. We can't do what #5567 did to the other targets because `CodeGen_GPU_Host` needs them as CRTP base classes. 

We can fix that by refactoring CodeGen_GPU_Host to be a lowering pass like HexagonOffload.cpp instead. I think this would generally be a great cleanup to do anyways. CodeGen_GPU_Host wouldn't be needed any more. This will be much easier after OpenGL goes away (which is #5626), because OpenGL vertex buffer stuff is the hardest thing to move to a separate lowering pass.

After that is done, we can just do basically the same thing as #5567 to the rest of the CodeGen_* classes.

_Originally posted by @dsharletg in https://github.com/halide/Halide/issues/5503#issuecomment-758271714_

JanHett:
Hello @dsharletg! Sorry for the late reply. To answer your question from #5503, I'd be happy to work on this but I feel like I don't understand the internals of Halide well enough to get started. Code generation et Cie are definitely outside my "comfort zone", but if you can point me at some resources relevant to Halide's implementation I would certainly be interested in giving this refactor a good try.

dsharletg:
Thanks for offering to take a look. The basic thing that needs to happen is that at some point during compilation, we need to inject calls to the runtime functions to implement GPU tasks. Currently, this happens in two places:
- [InjectHostDevBufferCopies.cpp](https://github.com/halide/Halide/blob/master/src/InjectHostDevBufferCopies.cpp): this adds calls to `halide_copy_to_device` and `halide_copy_to_host` as needed.
- [CodeGen_GPU_Host.cpp](https://github.com/halide/Halide/blob/master/src/CodeGen_GPU_Host.cpp): This inserts calls to `halide_*_run`, `halide_*_initialize_kernels` and related runtime functions as needed.

CodeGen_GPU_Host is implemented by directly generating LLVM IR to implement the calls it makes, while InjectHostDevBufferCopies.cpp generates Halide IR first, as part of lowering (in Lower.cpp). 

HexagonOffload.cpp is an example of how we should do things instead. Compare HexagonOffload generating calls to halide_hexagon_run with Halide IR: https://github.com/halide/Halide/blob/1155bdade064444975e752b074f0a8563fd9e910/src/HexagonOffload.cpp#L915
[...]

PR #5757, opened 2021-02-18
Add OpenCL to C/C++ backend

sakehl:
Hi,

I've done some work to add OpenCL to the C/C++ backend.
Most of the work is done in a new file/class `class CodeGen_Simple_OpenCL` and the file `src/CodeGen_Simple_OpenCL.cpp`. It is mostly based on existing implementation that is used in LLVM (in the file `src/CodeGen_GPU_Host.cpp`), although this is completely free of the LLVM implementation.

I've added the function `compile_to_opencl` in the front (to the Func and Pipeline classes), to use this functionality. 

Would you be interested in adding this to the Halide compiler? Or are there any specific changes I'd have to make first?
Somewhat related is this issue: https://github.com/halide/Halide/issues/5650 . If this code cleanup actually happens, a lot of this functionality would move.

Btw, the reason I've done this, is that I wanted to have a good grip on what the Halide compiler actually compiles for GPU code. Since the OpenCL kernel was already present, I wanted to see the exact host code which was generated. A further avenue of mine is to formally verify some of the outputted code of Halide, but this not really relevant further for this pull request.

A working example of what I implemented is:
```c++
#include "Halide.h"
#include <stdio.h>

using namespace Halide;

    Func in("in"), even("even");
    Var x("x");

    in(x) = 2*x*x;
    even(x) = in(x) + in(x+1);

    Var bx("bx"), tx("tx");
    in.compute_root();
    in.gpu_tile(x, bx, tx, 32);
    even.gpu_tile(x, bx, tx, 32);
    even.bound(x,0,100);

    Target target = get_host_target();
    Target new_target = target
      .with_feature(Target::OpenCL)
      ;
    assert(host_supports_target_device(new_target));
    compile_standalone_runtime("opencl_runtime.o",new_target);

    even.compile_to_opencl("even_c.cpp", {}, "even", new_target);
    even.compile_to_header("even_c.h", {}, "even", new_target);
}
```

After which you compile this code, run it, and afterwards compile even_c.c like:
`g++  -Wall  -o even_c_run  -I [Halide-install]/include/ even_c_run.cpp even_c.cpp opencl_runtime.o -lpthread -ldl`
Where `even_c_run.cpp` is a simple file containing:
```c++
#include "HalideBuffer.h"
#include "even_c.h"

#include <stdio.h>

    int n = 100;
    Halide::Runtime::Buffer<int> output(n);

    int error = even(output);
    output.copy_to_host();

    if (error) {
        printf("Halide returned an error: %d\n", error);
        return -1;
    }

    for (int x = 0; x < n; x++) {
        uint8_t output_val = output(x);
        uint8_t correct_val = 2*x*x + 2*(x+1)*(x+1);
        if (output_val != correct_val) {
            printf("output(%d) was %d instead of %d\n",
                   x, output_val, correct_val);
            return -1;
        }
    }

    // Everything worked!
    printf("Success!\n");
    return 0;
}
```


steven-johnson:
Can you expand a bit more on how this differs from our existing OpenCL support? 

(Generally speaking, we always welcome improvements to existing codegen, but it's not clear to me from an initial skim why this is preferable to the existing opencl output we get from adding `opencl` as a target feature, and it's doubtful we'd want to support OpenCL output via two distinct code paths.)

sakehl:
Ah sure of course, in essence, it is not different from what is currently here. The actual device code (GPU code) is exactly the same (Just OpenCL).
* But the host code normally used LLVM as code generator which is nice, but is (at least for me) hard to read and understand.
* The changes I propose here, extend the code generator for the C/C++ backend. They add kernel calls (all the memory calls are already there). It is doing almost the same as the LLVM host code (from which I tried to base this.
*  It is similar to what happens when you use `compile_to_c`, although that now prints "GPU code not supported for C/C++ backend" or something similar. 

Note that this whole functionality could also be added to the "compile_to_c" function, and then the files CodeGen_C.cpp etc. But I thought adding functionality to separate files and functions was nicer.

steven-johnson:
I don't want to reject your work, but it's still not clear to me what advantage this has over our existing OpenCL codegen. Does it provide a clear advantage over the existing OpenCL output? Even if it does, could those advantages be retrofitted to the existing OpenCL backend (rather than creating an entirely new one from scratch)? Does it provide 100% backward compatibility, so that we could delete the existing OpenCL backend and replace it with this?

sakehl:
I'll attach my examples files, so I can try to explain a little bit better what the purpose/goal of these changes are:
[even.zip](https://github.com/halide/Halide/files/6028023/even.zip)

- `even.cpp` is a regular Halide file
  - it makes the file even_c.cpp using the `compile_to_opencl` function. This is normal readable C/C++ file. In this file you can find the string with the OpenCL kernels, and the host code, which calls these kernels.
  - it makes the file "even_c.h" with the function `compile_to_header` which is just the header.
  - It makes the file`opencl_runtime.o` with function `compile_standalone_runtime`. It is the OpenCL runtime, which we need to compile `even_c.cpp`
- `even_c_run.cpp` is just a helper function, to check if the generated `even_c.cpp` behaves as we expect.

So WHY did I make this? To really **see** both the host code (everything executing on the CPU) and the device code (the kernels, that get executed on the GPU). Up to now, you could see what device code was generated by using `HL_DEBUG_CODEGEN`. But the host code, in combination with GPU code, was not easily visible. It was only visible as LLVM, which I cannot easily grasp.

It really has the same purpose as the 'compile_to_c' function. You do not use this to get the fastest code. You use this to see what is happening, what *kind* of code gets generated. To debug/understand the functionality of Halide when generating GPU code. 

Currently, when you use 'compile_to_c' in combination with some gpu functionality (like `gpu_threads`), you get the following error: "C++ Backend does not support gpu_blocks() or gpu_threads() yet, this function will always fail at runtime".  See: https://github.com/halide/Halide/blob/1be92c2cd5ae9f95e37576ffb653cbe4c3458dd0/src/CodeGen_C.cpp#L1590
With these changes, when using `compile_to_opencl`, works exactly like `compile_to_c` but doesn't give you this error.

And lastly, I did this so you could possible formally verify the host code that is used. Which is a project I'm working on. Since I could possible verify C++ code, but not LLVM code.

But to get back to your specific questions:

> Does it provide a clear advantage over the existing OpenCL output? 
The device code (GPU kernels) are exactly the same since it uses the exact same functionality of `src/CodeGen_OpenCL_Dev.cpp`.
The host code (the code running on the CPU, calling the kernels), are now clearly readable in C++ instead of only available as LLVM code. 

> Even if it does, could those advantages be retrofitted to the existing OpenCL backend (rather than creating an entirely new one from scratch)?
The existing backend is primarily used by this functionality. The only thing it does separately is generating the host code, and then more specifically the kernel calls. 
If you really want to combine these things, I think this issue https://github.com/halide/Halide/issues/5650 is the way to go. If you would rather go that route, I can have a look if I can help with it. (Although it seems more difficult).

> Does it provide 100% backward compatibility, so that we could delete the existing OpenCL backend and replace it with this?
So this is not the purpose of these changes. The existing OpenCL backend is what should be kept using. This "new" functionality, is mainly here for understanding/debugging readable code.

steven-johnson:
Ahhh, ok, I see now: this is replacing the *host* code for OpenCL (which ordinarily would be generated by LLVM), not the OpenCL code itself. Yeah, this could be interesting. How hard would it be to generalize for the other GPU backends, I wonder? 

In terms of efficiency, I would suspect that the host code is rarely gonna move the needle much either way; would need profiling/benchmarking to be sure, but I wouldn't be surprised if performance is close to a wash (assuming that the code being generated is functionally equivalent).

I think it would be worth discussing over on #5650 before moving forward with this, though; the refactoring described there really needs to happen (once someone gets the time to do so), and that decision may inform thoughts on this one.

steven-johnson:
So where does this PR stand now that #5650 has landed?

sakehl:
So I took a look, and with #5650 this has become way easier. I made 3 changes with respect to #5650.

- In `src/OffloadGPULoops.cpp` it now only get the module state pointer once (which contains the general OpenCL/CUDA/etc context, e.g. second argument to `halide_opencl_run`) once, and passes a variable when needed again. Previously, it declared multiple variables with the name "_opencl" in the C backend, when you have more than 1 kernel, which made it unable to compile.
- in `src/CodeGen_C.cpp` I've changed the way the GPU kernel source buffers are represented. Normal constant buffers are still represented as an uint8 array. GPU kernel source buffers are first stored as a string literal (making them readable) and are then cast to a uint8 array.
- The function ```void CodeGen_C::visit(const Allocate *op)``` had a strange way of pretty printing, making my code not compiling. With these changes it does. (E.g. previously it gave something like ```_21 = {_22 = 3; _22;};``` , which is not valid).

These are the files used to check the basic functionality, if anyone is interterested. 
[even.zip](https://github.com/halide/Halide/files/6107783/even.zip)

src/CodeGen_C.cpp
```diff
@@ -1653,22 +1657,35 @@ void CodeGen_C::compile(const Buffer<> &buffer) {
...
-    for (size_t i = 0; i < num_elems * b.type.bytes(); i++) {
-        if (i > 0) {
-            stream << ",";
-            if (i % 16 == 0) {
-                stream << "\n";
-                stream << get_indent();
-            } else {
-                stream << " ";
+    // If it is an GPU source kernel, we would like to see the actual output, not the
+    // uint8 representation. We use a string literal for this.
```
dsharletg, line 1661: Do all of the APIs emit text? Maybe some of them emit binary data, where this might do problematic things? I'm not sure about Metal for example.

src/OffloadGPULoops.cpp
```diff
@@ -302,25 +284,26 @@ class InjectGpuOffload : public IRMutator {
...
+            if (!state_needed[api_unique_name]) {
                 continue;
             }
+            Expr state_ptr = make_state_var(api_unique_name);
+            Expr state_ptr_var = Variable::make(type_of<void *>(), api_unique_name);
 
             debug(2) << "Generating init_kernels for " << api_unique_name << "\n";
             vector<char> kernel_src = i.second->compile_to_src();
-            Expr kernel_src_buf = make_buffer_ptr(kernel_src, api_unique_name + "_kernels");
+            Expr kernel_src_buf = make_buffer_ptr(kernel_src, api_unique_name + "gpu_source_kernels");
```
dsharletg, line 295: Should there be an underscore here? This will emit something like "openclgpu_source_kernels" I think.

src/CodeGen_C.cpp
```diff
@@ -35,6 +35,10 @@ extern "C" unsigned char halide_internal_runtime_header_HalideRuntimeD3D12Comput
 
 namespace {
 
+bool endsWith(const string &str, const string &suffix) {
```
dsharletg, line 38: I think we already have an ends_with function: https://github.com/halide/Halide/blob/master/src/Util.h#L158

src/CodeGen_C.cpp
sakehl: Yes, I tested this for OpenCL, Cuda, Metal, D3D12Compute and OpenGLCompute. All print (pretty) GPU kernels.

sakehl:
> Could you attach a sample of what the newly generated output looks like?

Sure, this [even_c.zip](https://github.com/halide/Halide/files/6114369/even_c.zip) file is the output of the `even.cpp` file (from this zip: [even.zip](https://github.com/halide/Halide/files/6107783/even.zip)). The source string starts at line 2396.

Thanks for the reviews, I incorporated the changes you suggested!

src/CodeGen_C.cpp
```diff
@@ -1653,22 +1653,35 @@ void CodeGen_C::compile(const Buffer<> &buffer) {
...
-                stream << "\n";
-                stream << get_indent();
-            } else {
-                stream << " ";
+    // If it is an GPU source kernel, we would like to see the actual output, not the
+    // uint8 representation. We use a string literal for this.
+    if (ends_with(name, "gpu_source_kernels")) {
+        stream << "static const char *" << name << "_string = R\"BUFCHARSOURCE(";
+        for (size_t i = 0; i < num_elems * b.type.bytes(); i++) {
+            stream << (char)(b.host[i]);
```
dsharletg, line 1661: Probably best to use `stream.write(b.host, num_elems)` here. Maybe writing chars one by one is noticeably slow for large kernels?

sakehl:
Done!

dsharletg (approved):

PR #5761, opened 2021-02-19
Constrain calling convention for PyStub

steven-johnson:
There are two calling conventions that are sensible when calling a Generator from Python:
- Inputs are specified via keyword arguments (verbose but more readable)
- Inputs are specified via positional arguments (terse but more like C++)

However, the current PyStub implementation also allows for mixing positional and keyword forms of input (as long as no input is specified multiple times). This change removes that possibility (requiring either all-keyword or all-positional for inputs). I'm not sure why I thought this was a good option originally, but upon reflection (and examination of existing code), calls written in this way tend to be a confusing mess that require too much care when reading, and I think we'd be better off just forbidding this entirely.

Note that this technically is a breaking API change (thus the `release_notes` label); any user code that used this technique previously would now throw an exception and need a (trivial) update. That said, this is (AFAIK) a very rarely used API, and I rather suspect that existing code that relies on being able to mix positional and keyword inputs in this way is likely to be doing so inadvertently rather than deliberately.

### DIFF
--- a/python_bindings/correctness/pystub.py
+++ b/python_bindings/correctness/pystub.py
@@ -36,40 +36,57 @@ def test_simplestub():
     f = simplestub.generate(target, buffer_input=b_in, func_input=f_in, float_arg=3.5)
     _realize_and_check(f)
 
-    # ----------- Inputs w/ mixed by-position and by-name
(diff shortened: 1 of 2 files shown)


steven-johnson:
Monday Morning Review Ping

dsharletg (approved):

PR #5762, opened 2021-02-19
Add TailStrategy::RoundUp to internal stages of local_laplacian and interpolate

dsharletg:
This reduces the size and complexity of the IR we generate, which makes debugging issues easier (and code size/compile time a little smaller).


steven-johnson (approved):

steven-johnson:
Looks ready to land

dsharletg:
@abadams was opposed to this offline and I don't think it's that important.

PR #5766, opened 2021-02-22
Grab-bag of minor Generator-related cleanups

steven-johnson:
Various minor code changes found during refactoring work; cherry-picking here to reduce noise later:
- de-inlining methods in Generator.h
- removing forward decls for nonexistent classes
- avoiding extern declarations that can trivially be avoided by just including Halide.h
- generator_main() should emit to the stream passed in (rather than calling user_error)
- other stuff

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -746,7 +746,7 @@ std::string halide_type_to_c_type(const Type &t) {
     return m.at(encode(t));
 }
 
-int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
+int generate_filter_main_inner(int argc, char **argv, std::ostream &error_output) {
     const char kUsage[] =
         "gengen\n"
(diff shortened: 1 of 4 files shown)


src/Generator.cpp
```diff
@@ -989,10 +990,11 @@ int generate_filter_main_inner(int argc, char **argv, std::ostream &cerr) {
         Target gcd_target = targets[0];
         for (size_t i = 1; i < targets.size(); i++) {
             if (!gcd_target.get_runtime_compatible_target(targets[i], gcd_target)) {
-                user_error << "Failed to find compatible runtime target for "
-                           << gcd_target.to_string()
-                           << " and "
-                           << targets[i].to_string() << "\n";
+                cerr << "Failed to find compatible runtime target for "
```
alexreinking, line 993: Nit: this is easily confused for `std::cerr`. Should probably rename the parameter to `os` or something else conventional, but not clashing with a standard name.

alexreinking (approved):
Approve with

src/Generator.cpp
steven-johnson: Done

PR #5770, opened 2021-02-25
Fix autoscheduler breakage from #5766 (Fix #5769)

### LINKED ISSUE #5769: Linking autoschedulers to generators is broken
Generators don't seem to be using autoschedulers on current master, i.e. if I change the default apps autoscheduler to adams2019 and add changes in the adams2019 autoscheduler, the autoscheduling command seems to complete immediately and doesn't appear to use a schedule different from the non-autoscheduled schedule. 

@abadams  Confirmed this issue for him as well, and says it look like the auto_schedule flag is being dropped, because the IR for autoscheduled and not is the same.

steven-johnson:


steven-johnson:
(Still testing, but I think this is the issue)

abadams (approved):

Issue #2797, opened 2018-03-07
how to generate C for host and OpenCL for device?


ronghongbo:
Hello,

I would like to generate OpenCL code for a device (from a parallel loop in a loop nest), and generate C code for the host (from the rest of the loop nest). By using "compile_jit", I can see the OpenCL code generated, and the host code is automatically executed. How can I dump out the host code as C so that I may play with the C and OpenCL code manually (e.g. modify and compile them in a command shell)? 

thanks!
hongbo

zvookin:
You will need to use the "compile_to_c" method to generate C code. (I
believe this works with GPU backends now, but it may not be very solid.
This is a little used path.)

-Z-

ronghongbo:
Thanks! I tried that, it generates C code, but the host and device code are not separated. The code is like this: 
   halide_copy_to_device(...)
   **#pragma omp parallel for
   for (....) {
        // the code that should be a device kernel 
   }**
   halide_device_free(...)
I hope the above bold lines can be separated out as a device kernel, and the activation of the kernel is explicitly shown.

zvookin:
Halide is not designed to generate separate kernels for GPUs that run
independently of the Halide generated host code (in whatever system the
host code is running).

-Z-

Issue #1971, opened 2017-04-04
External code is not fully implemented for GPU devices and source code backends.


zvookin:
The API design for ExternalCode allows providing source code for various GPU backend compute languages and C++ code for the C++ backend. (And JavaScript eventually.) However this is not fully plumbed through. It is also not covered by tests.

abadams:
This API was removed, so there's nothing left to test.

PR #5775, opened 2021-02-27
Move CodeGen_GPU_Host to a lowering pass

### LINKED ISSUE #5650: Move CodeGen_GPU_Host to a lowering pass

dsharletg:
This PR deletes CodeGen_GPU_Host and replaces it with a new lowering pass. This should enable GPU targets to be used with the C backend.

I think this fixes #5650, #2797, #2084, and #1971 seems relevant but not sure I understand that issue.


abadams:
What in the new file is new code vs code moved from CodeGen_GPU_Host?

dsharletg:
I don't think anything is new. This PR was created with the following steps:
- Remove unused vertex_buffer related arguments to halide_{api}_run
- Copy HexagonOffload.cpp to OffloadGPULoops.cpp, delete irrelevant code.
- Copy CodeGen_GPU_Host visitors and helpers to OffloadGPULoops visitor and helpers
- Port things from LLVM IR to Halide IR

steven-johnson (approved):

src/OffloadGPULoops.cpp
```diff
@@ -0,0 +1,345 @@
...
+            module_state(api_unique_name),
+            kernel_name,
+            Expr(bounds.num_blocks[0]),
+            Expr(bounds.num_blocks[1]),
+            Expr(bounds.num_blocks[2]),
+            Expr(bounds.num_threads[0]),
+            Expr(bounds.num_threads[1]),
+            Expr(bounds.num_threads[2]),
+            Expr(bounds.shared_mem_size),
+            Call::make(Handle(), Call::make_struct, arg_types_or_sizes, Call::Intrinsic),
```
abadams, line 272: I'm confused about this. It looks like it's a pointer to an array of pointers (because make_type_arg returns a pointer), but the runtime expects a pointer to an array of types

src/OffloadGPULoops.cpp
```diff
@@ -0,0 +1,345 @@
...
+        if (allocate->memory_type == MemoryType::GPUShared) {
+            internal_assert(allocate->extents.size() == 1);
+            shared_mem_size += allocate->extents[0] * allocate->type.bytes();
+            found_shared = true;
+        }
+        allocate->body.accept(this);
+    }
+};
+
+Expr make_type_arg(const Type &t) {
```
abadams, line 95: I think you want to reinterpret cast the bits of the halide_type_t to a uint32 instead.

src/OffloadGPULoops.cpp
dsharletg: I changed this, but I don't have any working OpenGL API on my machine that takes types, so I can't test it... we'll have to see what the buildbots say.

src/OffloadGPULoops.cpp
```diff
@@ -229,7 +220,8 @@ class InjectGpuOffload : public IRMutator {
             args.push_back(val);
 
             if (runtime_run_takes_types) {
-                arg_types_or_sizes.push_back(make_type_arg(i.type.with_lanes(1)));
+                internal_assert(sizeof(halide_type_t) == sizeof(uint32_t));
```
steven-johnson, line 223: You could just call the `.as_u32()` method from halide_type_t instead.

PR #5784, opened 2021-03-03
Fix out of bounds reads in strided ARM loads

dsharletg:
This PR fixes long-standing out of bounds reads for strided loads. This combines some of the logic in CodeGen_LLVM (for stride 2 loads) with the logic in CodeGen_ARM (for stride up to 4), and implements it in CodeGen_LLVM.

This will be a performance regression for code that uses strided loads from external buffers without sufficient alignment information to determine the loads are safe. However, measurement on a variety of code shows the performance impact is mostly small or even improvements in some cases.


steven-johnson:
Looks clean. Ready to land?

dsharletg:
We discussed the possibility of emitting a warning when we do not generate vldN when we would have before this change. However, I now think we should not do this, because I found many cases where this warning would be emitted, yet performance is better or unchanged. Overall, the performance impact of this is actually a lot more mixed than I expected. There are a few extreme cases where performance is up to 2x worse, but most of the time, the impact is negligible and often an improvement.

abadams:
It's surprising that there are cases where performance is better. Have you looked at the asm to see why?

dsharletg:
I think those cases are mostly vld2, and the base class generates reasonable code for those (2 loads + shuffle).

dsharletg:
I modified this to implement up to stride 4 in CodeGen_LLVM, using a combination of the old CodeGen_LLVM logic, and the new CodeGen_ARM logic.

abadams:
I'm seeing some good speed-ups on the packed cases in the resize app on x86 due to using dense loads and shuffles instead of gathers.

src/CodeGen_LLVM.cpp
```diff
@@ -1910,63 +1910,76 @@ void CodeGen_LLVM::visit(const Load *op) {
...
-                    shifted_a = true;
-                    base_b -= 1;
-                    align_b = align_b - 1;
-                    shifted_b = true;
-                }
+            // The min moves lower by offset.
+            int load_lanes = ramp->lanes * stride->value;
+            bool alignment_safe = align.remainder >= 0;
+            // The max needs to not cross an alignment boundary.
+            if (align.modulus > 0) {
```
abadams, line 1947: I'm confused about this check. align.modulus could be arbitrarily large and is not necessarily a power of two, so I don't see why this makes it safe. Should this be a check on aligned_stride instead?

dsharletg: I have been unsure about this check in general, especially without #5793, this alignment doesn't include the host alignment.

I'm also checking and performance without the check at all actually seems fine, I think I'll just remove it.

steven-johnson:
`correctness_memoize` is failing on arm32.

dsharletg:
I can't reproduce that and correctness_memoize looks completely unaffected by this change.

dsharletg:
Oh, the same failure occurred on another unrelated PR: https://buildbot.halide-lang.org/master/#/builders/32/builds/16, as well as quite a few other issues.

I think that is surely a flake.

steven-johnson:
> Oh, the same failure occurred on another unrelated PR: https://buildbot.halide-lang.org/master/#/builders/32/builds/16, as well as quite a few other issues.
> 
> I think that is surely a flake.

I wonder if #5780 could be causing the memoize issue?

abadams (approved):

abadams:
Dang, this was actually a major regression for strided loads from input buffers on x86. Now it's a vector gather instead of a load and shuffle. Not sure how we missed it.

Issue #5786, opened 2021-03-03
ARM assembly syntax changed in LLVM for arm-64-osx


dsharletg:
This is breaking simd_op_check, it's not clear how we should fix this.

steven-johnson:
Seems odd that it would be different for arm-64-osx vs arm-64-linux, but I'll take a look

dsharletg:
I misdiagnosed this, this is just because there's a new osx ARM build bot, which I didn't realize. I think Andrew is looking at this already.

abadams:
Yeah this is a difference in assembly printing for M1 somehow. The big issue is that the registers are not type-annotated, but the output type for the instruction *is*. These types are not the same, so there's no obvious way to continue type-checking without adding a third expected format for each instruction.

zvookin:
Is there a flag to request different assembly formatting?

-Z-

PR #5787, opened 2021-03-03
Fixes for macos on arm

abadams:
Now that we actually have a buildbot testing thing, a few minor issues were discovered. The big one is #5786

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ LLVM_LIBDIR = $(shell $(LLVM_CONFIG) --libdir | sed -e 's/\\/\//g' -e 's/\([a-zA
 # Apparently there is no llvm_config flag to get canonical paths to tools,
 # so we'll just construct one relative to --src-root and hope that is stable everywhere.
 LLVM_GIT_LLD_INCLUDE_DIR = $(shell $(LLVM_CONFIG) --src-root | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')/../lld/include
-LLVM_SYSTEM_LIBS=$(shell ${LLVM_CONFIG} --system-libs --link-static | sed -e 's/[\/&]/\\&/g' | sed 's/-llibxml2.tbd//')
+LLVM_SYSTEM_LIBS=$(shell ${LLVM_CONFIG} --system-libs --link-static | sed -e 's/[\/&]/\\&/g' | sed 's/-llibxml2.tbd/-lxml2/')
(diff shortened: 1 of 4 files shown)


test/correctness/simd_op_check.cpp
```diff
@@ -1760,7 +1760,7 @@ class SimdOpCheck : public SimdOpCheckTest {
                 // Saturating integer subtraction
                 check("i8x16.sub_saturate_s", 16 * w, i8_sat(i16(i8_1) - i16(i8_2)));
                 check("i16x8.sub_saturate_s", 8 * w, i16_sat(i32(i16_1) - i32(i16_2)));
-                // N.B. Saturating subtracts are expressed by widening to a *signed* type
+                // N.B. Saturating subtracts are expressed by widening to a signed* type
```
steven-johnson: find-and-replace made a mistake here

steven-johnson (approved):

Makefile
```diff
@@ -60,7 +60,7 @@ LLVM_LIBDIR = $(shell $(LLVM_CONFIG) --libdir | sed -e 's/\\/\//g' -e 's/\([a-zA
 # Apparently there is no llvm_config flag to get canonical paths to tools,
 # so we'll just construct one relative to --src-root and hope that is stable everywhere.
 LLVM_GIT_LLD_INCLUDE_DIR = $(shell $(LLVM_CONFIG) --src-root | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):/\/\1/g')/../lld/include
-LLVM_SYSTEM_LIBS=$(shell ${LLVM_CONFIG} --system-libs --link-static | sed -e 's/[\/&]/\\&/g')
+LLVM_SYSTEM_LIBS=$(shell ${LLVM_CONFIG} --system-libs --link-static | sed -e 's/[\/&]/\\&/g' | sed 's/-llibxml2.tbd//')
```
shoaibkamil, line 63: Huh, I'm surprised this worked.  Regardless, I think the right fix is `sed 's/-llibxml2.tbd/-lxml/'` though, since we probably want to preserve the fact that `llvm-config` claims libxml2 is a required system lib.

shoaibkamil: This problem also occurs on macOS x86-64 for me.

steven-johnson: If you configure LLVM with `-D LLVM_ENABLE_LIBXML2=OFF` at build time then this isn't necessary.

abadams: Would be nice for the makefile to work for people who didn't do that though. I'll take Shoaib's suggestion.

PR #5788, opened 2021-03-04
Add missing min/max/+/- rules

abadams:
These are almost all of the rules in <= four in terms of min/max/add/sub ops leaves that simplify to something with <= three leaves. Found via brute-force enumeration of all such expressions, and then bucketing them by evaluating them on a few random values, then testing the proposed rewrites with the formal verifier.

I left out things of the form: ```max(max(x, -x), 0) -> max(x, -x)```
While correct, that transformation actually hurts our ability to analyze that expression.

### DIFF
--- a/src/Simplify_Max.cpp
+++ b/src/Simplify_Max.cpp
@@ -83,6 +83,20 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(min(y, x), x), b) ||
              rewrite(max(min(x, c0), c1), b, c1 >= c0) ||
 
+             rewrite(max(x, min(y, min(x, z))), a) ||
+             rewrite(max(x, min(y, min(z, x))), a) ||
+             rewrite(max(x, min(min(x, y), z)), a) ||
+             rewrite(max(x, min(min(y, x), z)), a) ||
+             rewrite(max(min(x, min(y, z)), y), b) ||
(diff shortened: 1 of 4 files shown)


src/Simplify_Max.cpp
```diff
@@ -108,6 +122,16 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
                rewrite(max(x, (x/c1)*c1 + c2), a, c1 > 0 && c2 <= 0) ||
                rewrite(max(((x + c0)/c1)*c1, x), b, c1 > 0 && c0 <= 0) ||
                rewrite(max(x, ((x + c0)/c1)*c1), a, c1 > 0 && c0 <= 0) ||
+
+               rewrite(max(x, min(x, y) + c0), a, c0 <= 0) ||
```
rootjalex, line 126: Is there any reason for not including:
```
rewrite(max(x, c0 + min(x, y)), a, c0 <= 0) ||
rewrite(max(x, c0 + min(y, x)), a, c0 <= 0) ||
```
?

abadams: The simplifier rewrites c0 + x to x + c0, so that LHS won't appear.

rootjalex (approved):
Lgtm, the min/max rules should be useful for bounds inference!

steven-johnson:
Looks clean. Ready to land?

PR #5790, opened 2021-03-05
Use wabt-interp for AOT wasm testing

steven-johnson:
We currently use wabt-interp for JIT testing, but v8/d8 for AOT testing. This is not a great idea, since they are radically different runtimes. Since speed isn't an issue for the AOT tests we have, let's just use wabt-interp for that too.

(Opening this as a 'draft' because current top-of-tree WABT won't build properly in the configuration we need; I've opened a fix PR at https://github.com/WebAssembly/wabt/pull/1628)


steven-johnson:
Failures are apparently because wasm-interp doesn't (yet?) support emcc's pthread wrappers, even if you pass --enable-threads. https://github.com/WebAssembly/wabt/issues/1645

PR #5793, opened 2021-03-08
Refactor host alignment to enable specializing

dsharletg:
This PR refactors the implementation of host alignment logic to use the analysis from the simplifier, and bakes it into the Load/Store alignment.

This enables adding an `is_host_aligned` helper function for `OutputImageParam` that allows specializing based on the alignment of the host pointer. This can help avoid performance regressions from #5784 in some cases.

This PR loses the ability of JIT to learn alignment from the actual allocation of internal images.


src/OutputImageParam.cpp
```diff
@@ -107,4 +107,10 @@ OutputImageParam &OutputImageParam::store_in(MemoryType type) {
     return *this;
 }
 
+Expr is_host_aligned(const OutputImageParam &param, const Expr &alignment_bytes) {
```
steven-johnson, line 110: (1) Why a standalone function instead of a method on OutputImageParam?
(2) Either way, this should be added to the Python bindings.

abadams:
I'm a bit worried about this from an IR definition perspective. The alignment of an underlying pointer is a lower-level concept than Load and Store. Load/Store/Allocate could be backed by a hash table, for example, if the allocation is only used sparsely, so while it makes sense to talk about the alignment of the index, which is an integer, I'm not sure it strictly makes sense to talk about the alignment of the memory address it represents. That's a backend-dependent property.

steven-johnson:
Monday morning review ping: where does this PR stand?

PR #5797, opened 2021-03-10
Fix wasm-interp ucon error

steven-johnson:
This is a subtle error that only shows up in builds with assertions enabled in wabt; the issue is that we ask for an int64 value from a wabt `Value` struct, but that struct only has an int32, so an assertion can fail. (Note that there is always storage for both, and the values are constrained and unused inside the JIT anyway, so this is mostly just a cosmetic fix.)

Also added a .gitignore entry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ python_bindings/bin/*
 build-64/*
 build-ios/*
 build-osx/*
+build_wasm/*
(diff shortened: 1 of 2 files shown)


steven-johnson:
arm32 failure is pre-existing and unrelated

abadams (approved):

PR #5803, opened 2021-03-12
Fix deprecation warnings for trunk LLVM

steven-johnson:
(1) Both the 'deprecated' and new, non-deprecated variants existed back to at least LLVM10, and the deprecated variant was commented as deprecated at that point as well; the change in LLVM13 is that they are now annotated with LLVM_ATTRIBUTE_DEPRECATED so we get compiler warnings (and thus errors).

(2) The fixes are simply replicating what the old, deprecated methods did internally.

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -91,20 +91,20 @@ void unpack_closure(const Closure &closure,
     int idx = 0;
     for (const auto &v : closure.vars) {
         Value *ptr = builder->CreateConstInBoundsGEP2_32(type, src, 0, idx++);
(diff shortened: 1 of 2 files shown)


abadams (approved):
Approving to unbreak the build, but I suspect this is just kicking the can. llvm is trying to remove types from pointers in their API still, so we may at some point need to get the llvm type from llvm_type_of(some_halide_type)->getPointerTo()

steven-johnson:
> Approving to unbreak the build, but I suspect this is just kicking the can. llvm is trying to remove types from pointers in their API still, so we may at some point need to get the llvm type from llvm_type_of(some_halide_type)->getPointerTo()

We'll cross that bridge after it's deprecated

PR #5804, opened 2021-03-12
Add Halide_CCACHE_BUILD option for CMake

steven-johnson:
This replicates the approach LLVM already has of baking a use-cache option directly into the CMake options, without having to futz with changing CC/CXX/etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,18 +43,18 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Build Halide with ccache if the package is present
-set(Halide_CCACHE_BUILD OFF CACHE BOOL "Set to ON for a ccache enabled build")
+option(Halide_CCACHE_BUILD "Set to ON for a ccache enabled build" OFF)
 if (Halide_CCACHE_BUILD)
-  find_program(CCACHE_PROGRAM ccache)
-  if (CCACHE_PROGRAM)
-      set(Halide_CCACHE_PARAMS "CCACHE_CPP2=yes CCACHE_HASHDIR=yes"
-          CACHE STRING "Parameters to pass through to ccache")
-      set(CCACHE_PROGRAM "${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM}")
-      set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
-      message(STATUS "Enabling ccache usage for building.")
-  else ()
-    message(FATAL_ERROR "Unable to find the program ccache. Set Halide_CCACHE_BUILD to OFF")
-  endif ()
+    find_program(CCACHE_PROGRAM ccache)
+    if (CCACHE_PROGRAM)
+        set(Halide_CCACHE_PARAMS CCACHE_CPP2=yes CCACHE_HASHDIR=yes
+            CACHE STRING "Parameters to pass through to ccache")
+        set(CMAKE_C_COMPILER_LAUNCHER ${CMAKE_COMMAND} -E env ${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM})
+        set(CMAKE_CXX_COMPILER_LAUNCHER ${CMAKE_COMMAND} -E env ${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM})
+        message(STATUS "Enabling ccache usage for building.")
+    else ()
+        message(FATAL_ERROR "Unable to find the program ccache. Set Halide_CCACHE_BUILD to OFF")
+    endif ()
 endif ()
 
 ##
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,11 @@ option(Halide_CCACHE_BUILD "Set to ON for a ccache enabled build" OFF)
 if (Halide_CCACHE_BUILD)
     find_program(CCACHE_PROGRAM ccache)
     if (CCACHE_PROGRAM)
-        set(Halide_CCACHE_PARAMS CCACHE_CPP2=yes CCACHE_HASHDIR=yes CCACHE_SLOPPINESS=pch_defines,time_macros
+        # TODO: ccache recommends setting CCACHE_SLOPPINESS=pch_defines,time_macros to
+        # enable precompiled header caching. Our timing found it slightly faster with
+        # just CCACHE_SLOPPINESS=pch_defines, so that's what we're using. Maybe revisit
+        # if issues occur (but we don't use any of the time macros so should be irrelevant).
+        set(Halide_CCACHE_PARAMS CCACHE_CPP2=yes CCACHE_HASHDIR=yes CCACHE_SLOPPINESS=pch_defines
             CACHE STRING "Parameters to pass through to ccache")
         set(CMAKE_C_COMPILER_LAUNCHER ${CMAKE_COMMAND} -E env ${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM})
         set(CMAKE_CXX_COMPILER_LAUNCHER ${CMAKE_COMMAND} -E env ${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -42,6 +42,21 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Build Halide with ccache if the package is present
+set(Halide_CCACHE_BUILD OFF CACHE BOOL "Set to ON for a ccache enabled build")
```
alexreinking, line 46:
```
option(Halide_CCACHE_BUILD "Set to ON for a ccache enabled build" OFF)
```
I prefer `option()` for easier grepping and because normal variables can override it (eg. in a build that includes us).

CMakeLists.txt
```diff
@@ -42,6 +42,21 @@ set(CMAKE_CXX_STANDARD 14)
...
+if (Halide_CCACHE_BUILD)
+  find_program(CCACHE_PROGRAM ccache)
+  if (CCACHE_PROGRAM)
+      set(Halide_CCACHE_PARAMS "CCACHE_CPP2=yes CCACHE_HASHDIR=yes"
+          CACHE STRING "Parameters to pass through to ccache")
+      set(CCACHE_PROGRAM "${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM}")
+      set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
```
alexreinking, line 53:
```
set(CMAKE_CXX_COMPILER_LAUNCHER "${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM}")
```
I prefer not to set _global_ properties that would affect an `add_subdirectory` user when a project-local variable would suffice.

alexreinking: Also, this is broken on Windows since it's using bash/ksh syntax. This is portable:

```
set(CMAKE_CXX_COMPILER_LAUNCHER "${CMAKE_COMMAND} -E env ${Halide_CCACHE_PARAMS} ${CCACHE_PROGRAM}")
```

alexreinking (changes requested):

alexreinking:
I'm not sure about the decision to fatal-error if ccache is not found. I don't see the advantage of this feature over putting the following in your `.bashrc`.

```
export CCACHE_CPP2=yes
export CCACHE_HASHDIR=yes
export CCACHE_SLOPPINESS=pch_defines  # Might need to add for the tests

export CMAKE_CXX_COMPILER_LAUNCHER=ccache
```

For the buildbots, we'll still have to set `Halide_CCACHE_BUILD` only on the workers that actually have it installed. Maybe if successfully finding ccache were optional, we could remove some code from the buildbot master script? Though if I have to pick between less CMake code and less Python code, I'll chop down the CMake every time.

CMakeLists.txt
steven-johnson: Done

CMakeLists.txt
steven-johnson: We shouldn't ever be enabling ccache on Windows, since it doesn't exist there. (But I'll make the change.)

steven-johnson: (Also note that the 'set-global-property' approach here is taken directly from LLVM)

steven-johnson:
My main motivation here was to provide a simple ccache configuration for Halide that mimics the technique used for LLVM. IMHO having a single flag to set (rather than having to remember to set CXX or specify a toolchain file) is a nice win.

I really dislike having to customize my .bashrc for stuff like this. (It's nice to have the *option* to, of course).

steven-johnson:
> I'm not sure about the decision to fatal-error if ccache is not found.

I'm not sure either, tbh. That's what LLVM's code does. I guess you could argue that if you explicitly ask for ccache, you'd rather have a hard fail than a whoops-why-are-things-slow. I could go either way on that.

CMakeLists.txt
alexreinking: Just in general: LLVM's build is not exactly a model of good CMake code.

I know ccache isn't a thing on Windows _now_, but it might be in the future. There's an open PR about it. Granted, it doesn't look like it's going to get merged soon.

alexreinking:
Okay, I think this is almost ready to merge, but I would like to test if setting `CCACHE_SLOPPINESS=pch_defines` is a win.

alexreinking:
I tested `CCACHE_SLOPPINESS` on my 10-core workstation on WSL Ubuntu 20.04 LTS. I used the system LLVM.

Here are my results:

| `CCACHE_SLOPPINESS` | First build time | Cache size | Second build time | Hits | Misses | PCH fail |
|---|---|---|---|---|---|---|
| (none) | 3m7.268s | 2567 files (74.7 MB) | 1m30.731s | direct: 851 <br> pp: 7 | 0 | 480 |
| `pch_defines` | 3m9.370s | 2567 files (95.9 MB) | 1m11.791s | direct: 848 <br> pp: 10 | 1 | 479 |

I ran each experiment twice and reported the more favorable times above. I don't know how to reconcile the hits/misses with the 20-second difference in runtime, but it looks like it's worth setting `CCACHE_SLOPPINESS=pch_defines`.

---

Here's the script I used. 

```shell
#!/bin/bash

ccache -C
ccache -p

rm -rf halide-build

ccache -z
ccache -s

cmake -S halide -B halide-build -DCMAKE_BUILD_TYPE=Release \
      -DLLD_DIR=/usr/lib/llvm-11/lib/cmake/lld -DHalide_SHARED_LLVM=YES -DHalide_CCACHE_BUILD=YES
time cmake --build halide-build
ccache -s

rm -rf halide-build

ccache -z
cmake -S halide -B halide-build -DCMAKE_BUILD_TYPE=Release \
      -DLLD_DIR=/usr/lib/llvm-11/lib/cmake/lld -DHalide_SHARED_LLVM=YES -DHalide_CCACHE_BUILD=YES
time cmake --build halide-build
ccache -s
```

CMakeLists.txt
```diff
@@ -42,6 +42,21 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Build Halide with ccache if the package is present
+option(Halide_CCACHE_BUILD "Set to ON for a ccache enabled build" OFF)
+if (Halide_CCACHE_BUILD)
+    find_program(CCACHE_PROGRAM ccache)
+    if (CCACHE_PROGRAM)
+        set(Halide_CCACHE_PARAMS CCACHE_CPP2=yes CCACHE_HASHDIR=yes
+            CACHE STRING "Parameters to pass through to ccache")
```
alexreinking, line 51: Let's add `CCACHE_SLOPPINESS=pch_defines` to this.

alexreinking (changes requested):

CMakeLists.txt
steven-johnson: There is a thing called `clcache` which intends to be a workalike. Haven't investigated.

steven-johnson:
> CCACHE_SLOPPINESS

I am filled with terror at the idea that "sloppiness" is a thing a build system would contemplate; I'd better look up the meaning of it

CMakeLists.txt
steven-johnson: > Let's add `CCACHE_SLOPPINESS=pch_defines` to this.

According to https://ccache.dev/manual/4.2.html#_precompiled_headers, we have to set it to `pch_defines,time_macros` to get GCC precompiled headers to work. It's not clear what (if anything) `pch_defines` does on its own. The help literally just says "be sloppy". I'd really like to know more before defining it.

alexreinking: I just tried both ways and it looks like setting only `pch_defines` and not `time_macros` is faster.

steven-johnson:
Added `CCACHE_SLOPPINESS=pch_defines,time_macros` (but have not yet done before-and-after benchmarking)

CMakeLists.txt
steven-johnson: I like things that are faster, but I hesitate a bit since (from a quick google) literally every example and everything in the docs show them being set in tandem. We never use the __TIME__ or __DATE__ macros so time_macros should be irrelevant to us, of course.

steven-johnson:
(buildbot testing is irrelevant here since the current buildbot doesn't enable this flag. We really just need to decide about thge SLOPPINESS.)

steven-johnson:
I just changed to `CCACHE_SLOPPINESS=pch_defines` and added a comment explaining that this might bear further investigation. Easy enough to change later.

alexreinking (approved):

PR #5805, opened 2021-03-14
Use $<TARGET_FILE> for build_halide_h

steven-johnson:
In theory this shouldn't be necessary, opening as draft

### DIFF
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -224,7 +224,7 @@ function(add_halide_library TARGET)
     endif ()
 
     add_custom_command(OUTPUT ${GENERATOR_OUTPUT_FILES}
-                       COMMAND ${generator_cmd}
+                       COMMAND $<TARGET_FILE:${generator_cmd}>
                        -n "${TARGET}"
                        -d "${GRADIENT_DESCENT}"
                        -g "${ARG_GENERATOR}"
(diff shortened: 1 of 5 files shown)


steven-johnson:
So I know that explicitly expanding a target to full path isn't supposed to be necessary, but it appears to be necessary when crosscompiling for arm32 on an aarch64 machine; for simplicity, we compile *everything* (including our tools like build_halide_h) with the toolchain file, and my assumption here is that because it's a 'foreign' executable, the expansion doesn't happen, because the path isn't expanded and those tools don't get launched. Adding the TARGET_FILE decoration makes it work on those buildbots. (Maybe there's a different setting that is needed, but this seems harmless.)

alexreinking:
Uggghhhh this is CMake at its most brain-dead and obtuse. The real solution is to split off our host-and-build-only tools into a separate project. I haven't had the time to make this change.

As a stopgap, I _think_ modifying the toolchain file (which is needed anyway) to _not_ set `CMAKE_SYSTEM_NAME` will trick CMake into thinking it's not cross-compiling and this should still work for 64-to-32 bit cross-builds.

steven-johnson:
> _**Uggghhhh**_ this is CMake at its most brain-dead and obtuse

That's impressive, actually

steven-johnson:
> 2\. Set `CMAKE_CROSSCOMPILING_EMULATOR` to `/usr/bin/env`, which will just launch the program directly.

This seems like the sounder/safer choice to me -- fudging CMAKE_SYSTEM_NAME seems like a recipe for weird miscompiles.  (Presumably we could also override CMAKE_CROSSCOMPILING_EMULATORon e.g. x86-64 systems to point to qemu?)

alexreinking:
> This seems like the sounder/safer choice to me -- fudging CMAKE_SYSTEM_NAME seems like a recipe for weird miscompiles. (Presumably we could also override CMAKE_CROSSCOMPILING_EMULATOR on e.g. x86-64 systems to point to qemu?)

Yes, qemu ought to work for x86-to-arm scenarios (though obviously with overhead).

alexreinking (changes requested):
Not the right approach. #5808 is better

PR #5808, opened 2021-03-15
Set CMAKE_CROSSCOMPILING_EMULATOR in toolchain.linux-arm32.cmake

steven-johnson:
Also add comments.

(Trying this as an alternative to https://github.com/halide/Halide/pull/5805)

### DIFF
--- a/cmake/toolchain.linux-arm32.cmake
+++ b/cmake/toolchain.linux-arm32.cmake
@@ -1,4 +1,13 @@
-# Toolchain for cross-compiling to Linux-arm32 on a Linux-x86-64 host.
+# Toolchain for cross-compiling to Linux-arm32 on a Linux-x86-64 or Linux-aarch64 host.
+#
+# If you just want to crosscompile, you need to install:
+#
+#   apt-get install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf binutils-arm-linux-gnueabihf
+#
+# On an aarch64 host, you can also run the results if you:
+#
+#   dpkg --add-architecture armhf
+#   apt-get install libc6:armhf libstdc++6:armhf
 
 set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_PROCESSOR arm)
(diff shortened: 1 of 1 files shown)


alexreinking (approved):

alexreinking:
Approved, assuming it works.

steven-johnson:
Testing now, and looks promising, but there is a minor fly in the ointment: we are going to need this toolchain for LLVM builds too, but those can't easily rely on a toolchain file in the Halide repo. (Currently this is working because I made a local copy of the toolchain on the armbots for experimentation.) 

AFAIK, LLVM doesn't provide toolchain files (just instructions on configuring stuff manually). 

Options:
- Use toolchain file for Halide itself; for LLVM, just replicate the necessary CMake stuff directly into buildbot script.
- Add a toolchain file to the buildbot repo and have the workers use that one for LLVM + Halide. (Then we have to keep then in sync, of course.)
- Add a toolchain file locally on the arm32 bots and point to it. Maintain manually. Yuck.
- Don't use the toolchain file at all; instead, put the relevant CMake mojo into the buildbot script. Yuck.
- Something else.

steven-johnson:
Oh, also: do we need to do

    SET(CMAKE_FIND_ROOT_PATH /usr/arm-linux-gnueabihf)

or something similar as well?

alexreinking:
>     * Add a toolchain file to the buildbot repo and have the workers use that one for LLVM + Halide. (Then we have to keep then in sync, of course.)

This is a good idea. I think there's a "push file to worker" step that we could use to keep the workers in sync with the master.

alexreinking:
> Oh, also: do we need to do
> 
> ```
> SET(CMAKE_FIND_ROOT_PATH /usr/arm-linux-gnueabihf)
> ```
> 
> or something similar as well?

I think one typically sets `CMAKE_SYSROOT`. See if things work without setting any of these variables and then try that one first.

steven-johnson:
> This is a good idea. I think there's a "push file to worker" step that we could use to keep the workers in sync with the master.

Yeah, we would still have the potential issue of keeping the toolchain files in sync between the Halide repo and build_bot repo, but that's ok

steven-johnson:
The test above are misleading since some buildbot changes have to land too, but from offline testing, I'm confident enough that it works to land it.

PR #5809, opened 2021-03-15
Avoid printing the statement when it is unchanged

dsharletg:
This PR tweaks lower to avoid printing the IR when a pass does not change it. It also fixes a few cases of reconstructing IR when it has not changed.

### DIFF
--- a/src/Lower.cpp
+++ b/src/Lower.cpp
@@ -84,10 +84,12 @@ namespace {
 
 class LoweringLogger {
     Stmt last_written;
+
 public:
     void operator()(const string &message, const Stmt &s) {
         if (!s.same_as(last_written)) {
-            debug(2) << message << "\n" << s << "\n";
+            debug(2) << message << "\n"
+                     << s << "\n";
             last_written = s;
         } else {
             debug(2) << message << " (unchanged)\n\n";
--- a/src/Lower.cpp
+++ b/src/Lower.cpp
@@ -80,6 +80,18 @@ using std::ostringstream;
 using std::string;
 using std::vector;
 
+namespace {
+
+Stmt new_or_empty(const Stmt &s, Stmt &last_written) {
+    if (s.same_as(last_written)) {
+        return Stmt();
+    }
+    last_written = s;
+    return s;
+}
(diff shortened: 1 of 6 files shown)


src/Lower.cpp
```diff
@@ -131,30 +143,31 @@ Module lower(const vector<Function> &output_funcs,
...
         debug(2) << "Lowering after injecting memoization:\n"
-                 << s << "\n";
+                 << new_or_empty(s, last_written) << "\n";
     } else {
         debug(1) << "Skipping injecting memoization...\n";
     }
-
     debug(1) << "Injecting tracing...\n";
     s = inject_tracing(s, pipeline_name, trace_pipeline, env, outputs, t);
     debug(2) << "Lowering after injecting tracing:\n"
```
abadams, line 164: This is awkward. Would it be better to have a little stateful class with a method:
```
LoweringLogger::operator()(const string &, const Stmt &);
...

log("Lowering after injecting parameter checks", s);
```

src/Lower.cpp
```diff
@@ -80,6 +80,18 @@ using std::ostringstream;
 using std::string;
 using std::vector;
 
+namespace {
+
+Stmt new_or_empty(const Stmt &s, Stmt &last_written) {
+    if (s.same_as(last_written)) {
+        return Stmt();
```
steven-johnson, line 87: Would be nice if we could emit something like `(Unchanged)` for this case, rather than an empty Stmt, but that's just extra-credit.

src/Lower.cpp
```diff
@@ -80,6 +80,18 @@ using std::ostringstream;
...
+    }
+    last_written = s;
+    return s;
+}
+
+}
```
steven-johnson, line 93: }  // namespace

steven-johnson (approved):
Besides making IR simpler to deal with, this could reduce memory churn if the unchanged cases were common. Did you notice any differences in compile time?

abadams (approved):

steven-johnson:
arm32 failure is unrelated

PR #5810, opened 2021-03-15
Conform to CMP0116 when using CMake 3.20+

### LINKED ISSUE #5373: CMake 3.20 will fix Ninja depfiles (but break our workaround).
Some day, we will bump our minimum CMake version. That might not be soon, but when we do, this code will break because it's working around CMake's bad handling of Ninja depfiles that will be fixed with a new policy in 3.20.

Opening this issue so there's a record of it somewhere.

Relevant CMake issue: https://gitlab.kitware.com/cmake/cmake/-/issues/21267
Relevant CMake PR: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5325

https://github.com/halide/Halide/blob/5286a688238ccae8762b24392b192a5c68741daf/src/runtime/CMakeLists.txt#L207-L227

alexreinking:
Use the CMake 3.20+ depfile features when they're available (avoids throwing one warning per initmod...). This will eventually be removed when we're able to upgrade to 3.20 (maybe in a year or two).

Fixes #5373

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.16...3.20)
 project(Halide VERSION 12.0.0)
 
 enable_testing()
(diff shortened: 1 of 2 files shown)


steven-johnson (approved):
Rubberstamping

alexreinking:
@steven-johnson any idea what's going on with llvm13-x86-64-osx-cmake? Does that bot just not have the relevant LLVM nightly?

steven-johnson:
I did various upgrades on the bot, including installing updated XCode, so I removed all the old build products. LLVM12 and 13 are rebuilding now, should be ready in a bit.

alexreinking:
> I did various upgrades on the bot, including installing updated XCode, so I removed all the old build products. LLVM12 and 13 are rebuilding now, should be ready in a bit.

Okay, so failure is unrelated. Merging...

PR #5813, opened 2021-03-16
Grab-bag of minor fixes

steven-johnson:
Fixes from an experimental branch, brought over to simplify irrelevant changes there:

- bash scripts for running TraceViz need to set (at least) `pipefail` to avoid quiet failures in some pathological cases
- we should use `.ldscript` as the suffix for linker version scripts (some build environments are picky about this)


abadams (approved):

alexreinking:
>     * we should use `.ldscript` as the suffix for linker version scripts (some build environments are picky about this)

Curious to know more about this?

steven-johnson:
> > * we should use `.ldscript` as the suffix for linker version scripts (some build environments are picky about this)
> 
> Curious to know more about this?

Some tooling inside Google's build setup *requires* one of .lds, .ldscript to pass this flag in at link time. ¯\_(ツ)_/¯

alexreinking:
Weird. Thanks for the context!

PR #5814, opened 2021-03-16
Add a seed to correctness_async_copy

steven-johnson:
We are seeing a flaky failure on some systems; allowing reproducibility of the PRNG sequence may help track down the cause.

### DIFF
--- a/test/correctness/async_device_copy.cpp
+++ b/test/correctness/async_device_copy.cpp
@@ -4,14 +4,14 @@
 using namespace Halide;
 
 // Return zero, slowly
-Expr expensive_zero(Expr x, Expr y, Expr t, int n) {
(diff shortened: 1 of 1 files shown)


abadams:
This actually makes it less deterministic, for the reasons we discussed offline. So I could see that this increases coverage, but it's not going to reduce flakiness.

steven-johnson:
I don't expect it to reduce flakiness -- I just want to make it easier to repro a failure. (Unless you are saying that currently the random_int() sequence defaults to returning an identical sequence every time, in which case... nevermind I guess?)

abadams:
It does, it returns the same sequence every time by default. It's a pure function of which number call to random_*() it was in the process and the pure vars of the Func that uses it.

Oblig: https://xkcd.com/221/

steven-johnson:
Meh, ok then, withdrawing.

PR #5815, opened 2021-03-16
Enable sliding window in registers

### LINKED ISSUE #1820: Sliding window / storage folding can't express rotating through a set of registers

dsharletg:
This PR enables an alternative strategy for sliding window (used when the producer of a sliding window is `store_in(MemoryType::Register)`). The alternative strategy works by shifting the producer such that the previously computed values are copied from the previous iteration, and the newly computed values are always computed in the same place. This avoids non-constant addresses, which are a prerequisite for storing things in registers (on most targets).

This mostly fixes #1820, though it currently requires bending over backwards in the schedule a bit for sliding vectorized dimensions.

@abadams contributed heavily to the ideas implemented here.

This PR has some other related changes:
- Treat `if_then_else` intrinsics like `select` when partitioning loops.
- After we're done with using `boxes_touched`, rebase some loops to have a min of 0, to enable stronger simplifications.
- More simplifier rules.


dsharletg:
Ping on a code review, please take a look at this when you get a chance @abadams. I've confirmed that this does address the use case we had in mind well.

src/RebaseLoopsToZero.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+using std::string;
+
+namespace {
+
+bool should_rebase(ForType type) {
+    switch (type) {
+    case ForType::Extern:
+    case ForType::GPUBlock:
+    case ForType::GPUThread:
```
abadams, line 16: gpu loops absolutely need to be rebased to zero, and there's a pass inside FuseGPUThreadLoops.cpp that does it. Is that mutator now redundant?

src/Simplify_LT.cpp
```diff
@@ -285,9 +285,43 @@ Expr Simplify::visit(const LT *op, ExprInfo *bounds) {
               rewrite(select(y, z, x + c0) < x + c1, y && (z < x + c1), c0 >= c1) ||
               rewrite(select(y, z, x + c0) < x + c1, !y || (z < x + c1), c0 < c1) ||
 
+              rewrite(c0 < select(x, c1, c2), select(x, fold(c0 < c1), fold(c0 < c2))) ||
```
abadams, line 288: These rules all formally verify ok

src/SlidingWindow.cpp
```diff
@@ -119,6 +119,94 @@ bool find_produce(const Stmt &s, const string &func) {
...
+        }
+        vector<Expr> args = op->args;
+        for (Expr &i : args) {
+            i = mutate(i);
+        }
+        bool sliding_up = old_bounds.max.same_as(new_bounds.max);
+        Expr is_new = sliding_up ? new_bounds.min <= args[dim] : args[dim] <= new_bounds.max;
+        args[dim] -= old_bounds.min;
+        vector<Expr> old_args = args;
+        old_args[dim] = substitute(loop_var, Variable::make(Int(32), loop_var) - 1, old_args[dim]);
```
abadams, line 164: Doesn't this need an equivalent of expand_expr? What if the dependence on the loop var is via a letstmt?

abadams:
Generally lgtm. I think the substitute call is fishy though.

src/RebaseLoopsToZero.cpp
dsharletg: That mutator works a little differently. It substitutes loop_var + min, rather than makes a new let, and that seems like maybe something that might matter for another pass/logic (there's a lot of stuff happening in FuseGPUThreadLoops). I would also be careful about changing when loops get rebased to 0.

Maybe this PR should make it so `rebase_loops_to_zero` accepts a set of `ForType` that get rebased, and use that in FuseGPUThreadLoops? That would at least keep the rebasing happening at the same time, and the only change would be substitute vs. let.

abadams: I think the new mutator is just more correct than the old one. I guess the old one happens earlier though, so we can't just rely on the new one to do the mutation. Your proposal sounds good, but I have no strong feelings one way or the other.

abadams (approved):

steven-johnson:
The OSX failure is unrelated (will be fixed by https://github.com/halide/Halide/pull/5841), should be good to land

src/RebaseLoopsToZero.cpp
dsharletg: I think the duplication here is minimal and there are non-minimal risks in messing with this, so I'll save it for a separate PR.

PR #5818, opened 2021-03-17
Add support for AMX instructions

mcleary:
This pull request continues the work started by @jwlawson in #5780 with the objective of adding initial support for AMX instructions in Halide.

The main addition here is the fix for building Halide using LLVM 11. Support for AMX instructions requires LLVM 12 or newer so when building with LLVM 11 the unsupported instructions are not included.

A new LLVM module was created (x86_amx.ll) to contain all the required intrinsics to enable support for tile operations, this module is only included when LLVM >= 12 is present.


steven-johnson:
(Synced to head to fix some irrelevant LLVM build issues)

mcleary:
> (Synced to head to fix some irrelevant LLVM build issues)

That is a good excuse for me to setup LLVM 13 locally. I will try to see why it is not building with it

steven-johnson:

steven-johnson:
You should sync this to master to force the bots to retry.

mcleary:
> You should sync this to master to force the bots to retry.

I'm not sure if the buildbot is still running since there is a "cancelled" message there.

steven-johnson:
Please try syncing to master once again; hopefully the buildbots will finally be clean.

mcleary:
> Please try syncing to master once again; hopefully the buildbots will finally be clean.

Thanks, I will do that. We are performing some more tests with this code so this will likely be updated soon.

steven-johnson:
Failures are the unrelated cuda-hang failure that we still haven't diagnosed; ok to land

steven-johnson:
Is this PR still active?

mcleary:
> Is this PR still active?

Yes, it is. We are working to add support for the remaining AMX variants, plus some improvements to this PR.

We will update this PR soon.

mcleary:
@steven-johnson we believe this PR is now ready. Please let us know if there are any changes you would like us to make.

steven-johnson:
Thanks! Assigned @dsharletg and @abadams as initial reviewers.

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,322 @@
...
+            }
+        }
+    }
+    return {};
+}
+
+Tile<3> is_3d_tile_index(const Expr &e) {
+    vector<Expr> matches;
+    auto add_sub_pattern = (wild_i32x + wild_i32x) - wild_i32x;
+    if (!expr_match(add_sub_pattern, e, matches)) { return {}; }
```
steven-johnson, line 44: Nit: preferred code style in Halide is to avoid single-line if clauses, so instead,

```
    if (!expr_match(add_sub_pattern, e, matches)) { 
        return {}; 
    }
```

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+struct NewMatmul {
+    bool result = false;
+    Stmt stmt;
+    int tile_x;
+    int tile_y;
+    int tile_r;
+};
+
+NewMatmul
```
steven-johnson, line 94: Nit: put return value on same line as function name except in extreme circumstances.

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,322 @@
...
+    const auto wild_u8x = Variable::make(UInt(8, 0), "*");
+    vector<Expr> matches;
+    const auto pattern1 = wild_i32x + wild_i32x;
+    if (!expr_match(pattern1, op->value, matches)) { return {}; }
+    const auto *reduce = matches[0].as<VectorReduce>();
+    const auto *load = matches[1].as<Load>();
+    if (!reduce || reduce->op != VectorReduce::Add) { return {}; }
+    if (!load || load->name != op->name || !equal(load->index, op->index)) { return {}; }
+
+    // FIXME: Add support for bf16 for LLVM 13+
```
steven-johnson, line 107: Does this need a tracking issue?

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,322 @@
...
+    if (expr_match(pattern2, reduce->value, matches)) {
+        lhs_signed = true;
+    } else if (expr_match(pattern2_unsigned, reduce->value, matches)) {
+        lhs_signed = false;
+    } else {
+        return {};
+    }
+
+    const auto *lhs_load = matches[0].as<Load>();
+    // FIXME: When tile_r is not 4 the broadcast is inside the index, not of the value
```
steven-johnson, line 121: Does this need a tracking issue?

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,322 @@
...
+    } else {
+        return {};
+    }
+
+    const auto *rhs_load = rhs_cast->value.as<Load>();
+    if (!rhs_load) { return {}; }
+
+    const auto lhs_tile = is_3d_tile_index(lhs_load->index);
+    const auto rhs_tile = is_2d_tile_index(rhs_load->index);
+    // FIXME: When tile_r is not 4 the RHS load will be 4D (x, r/4, y, r%4)
```
steven-johnson, line 139: Does this need a tracking issue?

src/ExtractTileOperations.cpp
mcleary: We are internally working on this one. It will likely be a separate pull request

src/ExtractTileOperations.cpp
frengels: I don't think either of these need tracking issues, I'll remove the comment.

src/ExtractTileOperations.cpp
frengels: Done with e3f1ef6

src/ExtractTileOperations.cpp
frengels: Also in e3f1ef6

frengels:
I have updated the PR to add support for AMX-bf16. 
The PR as it is currently is ready from our side.

For future work. Currently the matmul requires a `k` dimension of 4 or 2 for int8 and bf16 respectively. I will create a tracking issue so I can fix this limitation in the future.

frengels:
is there anything I missed preventing this PR from being merged?

steven-johnson:
Let's get this tested on the buildbots :-)

frengels:
It looks to me like all of the checks have passed

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,401 @@
...
+                return {true, std::move(matches[0]), {std::move(matches[2]), std::move(matches[1])}, {r1->lanes, r2->lanes}};
+            }
+        }
+    }
+    return {};
+}
+
+Tile<3> is_3d_tile_index(const Expr &e) {
+    vector<Expr> matches;
+    auto add_sub_pattern = (wild_i32x + wild_i32x) - wild_i32x;
```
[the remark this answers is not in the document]
mmoadeli: The engineer who has implemented above code has left. We are going to investigate it shortly.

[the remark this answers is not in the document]
mmoadeli: > I'm surprised this is an add-sub. Can you comment why this is? Is it perhaps overfit to a particular index calculation?

The add-sub pattern above is not introduced as part of support for AMX in this PR. The pattern is related to the way intermediate index is created for accessing elements, when vectorization is in scheduling and the function input is of types like ImageParam<> or Buffer<>. 

The similar pattern can be seen in tutorials/lesson_10_aot_compilation_generate.cpp when built on a host with NO support for AMX.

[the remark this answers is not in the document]
dsharletg: I see, this is probably subtracting the mins. Does the test still work if we set the buffer min to 0? (In the test, add `A.dim(0).set_min(0).dim(1).set_min(0); B.dim(0).set_min(0).dim(1).set_min(0)`)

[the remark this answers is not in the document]
mmoadeli: It hit the asset at line 362 of the same file in function  *Stmt visit(const Store *op) override*
**Store to AMX tile allocation of a non-tile value**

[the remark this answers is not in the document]
dsharletg: I think this is likely because the index expression doesn't match the pattern being looked for any more (a + b - c), because c is 0. I think we need the test to cover this case (in addition to a case where the mins vary), and fix it so it works.

I think maybe using patterns for looking for the right index expressions is not going to work very well. Maybe what we need to do instead is write some logic to dig through the needed expressions, with more tolerance for variations?

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,401 @@
...
+    int element_width = lhs_load_type.bytes();
+    auto lhs_type = lhs_load_type.with_lanes(1024 / element_width);
+    auto lhs = Call::make(lhs_type, "tile_load", {tile_x, tile_r * element_width, lhs_var, lhs_tile.base * element_width, lhs_tile.stride[0] * element_width}, Call::Intrinsic);
+
+    auto rhs_var = Variable::make(Handle(), rhs_load->name);
+    const auto &rhs_load_type = rhs_load->type;
+    auto rhs_type = rhs_load_type.with_lanes(1024 / element_width);
+    auto rhs = Call::make(rhs_type, "tile_load", {1, tile_y * tile_r * element_width, rhs_var, rhs_tile.base * element_width, rhs_tile.stride[0] * tile_y * element_width}, Call::Intrinsic);
+
+    auto res_type = [&]() {
```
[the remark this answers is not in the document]
frengels: This will cause the `res_type` var to always be fully initialized from the lambdas return value, rather than first partially initializing with the default constructor and assigning afterwards. I can change it if it doesn't fit the Halide style.

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,401 @@
...
+            body = mutate(body);
+            if (found_tile_x < 0 || found_tile_y < 0 || found_tile_r < 0) {
+                return op;
+            }
+            if (!pending_stores.empty()) {
+                // Really only need to go over the pending stores
+                body = mutate(body);
+            }
+
+            auto alloc_type = [&]() {
```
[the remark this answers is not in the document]
frengels: Same as above, to fully initialize right away rather than partially initializing first.

src/runtime/x86_amx.ll
```diff
@@ -0,0 +1,97 @@
...
+  %5 = bitcast x86_amx %4 to <256 x float>
+  ret <256 x float> %5
+}
+declare x86_amx @llvm.x86.tdpbf16ps.internal(i16, i16, i16, x86_amx, x86_amx, x86_amx)
+
+define weak_odr <2 x i1> @tilestored64_i32(i16 %rows, i16 %cols, i8* %ptr, i64 %off, i64 %stride, <256 x i32> %val) nounwind alwaysinline writeonly {
+  %1 = getelementptr i8, i8* %ptr, i64 %off
+  %2 = bitcast <256 x i32> %val to x86_amx
+  tail call void @llvm.x86.tilestored64.internal(i16 %rows, i16 %cols, i8* %1, i64 %stride, x86_amx %2) nounwind writeonly
+  ret <2 x i1> zeroinitializer
```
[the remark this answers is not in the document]
frengels: It should probably be void, it was put there by a colleague who has since left and I never questioned it. I'll check if anything goes wrong when I change this to void.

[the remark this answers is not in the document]
frengels: I couldn't find a Halide equivalent for the void type. I assume it would be fine to return a single `i1` in that case?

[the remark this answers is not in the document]
dsharletg: I think we just use int for runtime functions that need to return something.

src/ExtractTileOperations.cpp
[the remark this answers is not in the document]
dsharletg: Since the same lambda appears in several places, I think making a helper function to convert these types would be better.

src/runtime/x86_amx.ll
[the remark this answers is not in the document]
frengels: I've had a look and it happens to be that when calling `void CodeGen_X86::visit(const Call *op)` it checks whether the return type is a vector, and if not then it won't run the peephole optimization. I think I'd prefer to add a comment to explain why the `<i1 x 2>` is there rather than making peephole optimizations not require vector return types, I'd appreciate any feedback on the direction.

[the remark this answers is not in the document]
dsharletg: I think we can just remove that check for the return type being a vector. That seems like a (very) minor/premature optimization to me.

dependencies/llvm/CMakeLists.txt
```diff
@@ -20,6 +20,9 @@ message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 message(STATUS "Using ClangConfig.cmake in: ${Clang_DIR}")
 
+# LLVM_PACKAGE_VERSION does not propagate to higher scopes
+set(Halide_LLVM_VERSION ${LLVM_PACKAGE_VERSION} CACHE INTERNAL "Provided LLVM version")
+
```
alexreinking, line 25: Why is this here?

src/CodeGen_X86.cpp
```diff
@@ -190,7 +200,7 @@ const x86Intrinsic intrinsic_defs[] = {
 
     {"dpwssdx16", Int(32, 16), "dot_product", {Int(32, 16), Int(16, 32), Int(16, 32)}, Target::AVX512_SapphireRapids},
     {"dpwssdx8", Int(32, 8), "dot_product", {Int(32, 8), Int(16, 16), Int(16, 16)}, Target::AVX512_SapphireRapids},
-    {"dpwssdx4", Int(32, 4), "dot_product", {Int(32, 4), Int(16, 8), Int(16, 8)}, Target::AVX512_SapphireRapids},
+    {"dpwssdx4", Int(32, 4), "dot_product", {Int(32, 4), Int(16, 8), Int(16, 8)}, Target::AVX512_SapphireRapids},
```
alexreinking, line 203: nit: irrelevant whitespace?

dependencies/llvm/CMakeLists.txt
alexreinking: Sorry, GitHub's Ctrl+F didn't show the usage in `runtime/CMakeLists.txt`

src/runtime/CMakeLists.txt
```diff
@@ -115,6 +115,11 @@ set(RUNTIME_LL
     x86_sse41
     )
 
+if (Halide_LLVM_VERSION VERSION_GREATER_EQUAL 12.0)
+  # AMX instructions require LLVM 12 or newer
+  list(APPEND RUNTIME_LL x86_amx)
+endif ()
+
```
alexreinking, line 122: Does including this fail at build time or at runtime only?

frengels:
Just letting you know we haven't lost track of this and the TensorCore PRs. We had some different priorities and annual leave. I look forward to getting this merged soon.

src/ExtractTileOperations.cpp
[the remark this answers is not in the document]
frengels: concerning this I was wondering if it would be a good idea to use the `simplify` operation and then perform the pattern matching on the resulting expression?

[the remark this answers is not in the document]
frengels: I've adjusted the index expression checker to be more robust against those 2 scenarios

src/CodeGen_X86.cpp
frengels: Addressed in b1e1452409ca93b04e90a33681f396a49ec4face

src/runtime/x86_amx.ll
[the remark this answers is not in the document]
frengels: should be addressed with 34557cb48722fadeb03a3febbfbd61883d902144

src/ExtractTileOperations.cpp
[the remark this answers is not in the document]
frengels: I've added the following helper function `Type amx_op_type_result_type(AMXOpType)` to convert each case to the proper `Type`.

frengels:
I don't think the test failures are related to anything in this PR

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+        return Float(32, 256);
+    default:
+        return Type();
+    }
+}
+
+const auto wild_i32 = Variable::make(Int(32), "*");
+const auto wild_i32x = Variable::make(Int(32, 0), "*");
+
+Tile<2> is_2d_tile_index(const Expr &e) {
```
steven-johnson, line 44: Nit: I'd expect a function named "is_whatever" to return bool, but this returns a struct. Something like `get_2d_tile_index` would be better.

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+            auto ramp_2d_pattern = Ramp::make(Ramp::make(wild_i32, wild_i32, r2->lanes), Broadcast::make(wild_i32, r2->lanes), r1->lanes);
+            if (expr_match(ramp_2d_pattern, e, matches)) {
+                return {true, std::move(matches[0]), {std::move(matches[2]), std::move(matches[1])}, {r1->lanes, r2->lanes}};
+            }
+        }
+    }
+    return {};
+}
+
+Tile<3> is_3d_tile_index(const Expr &e) {
```
steven-johnson, line 58: Same nit here.

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+    vector<Expr> matches;
+
+    // there could be a sub node
+    const Sub *sub = e.as<Sub>();
+
```
steven-johnson, line 63: Remove blank line

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+            return {};
+        }
+
+        base -= bcast->value;
+    }
+
+    return {true, base, {x_stride, 0, r_stride}, {x_tile, y_tile, r_tile}};
+}
+
+struct NewMatmul {
```
steven-johnson, line 138: Identifiers with "New" in the name make me nervous, since they usually get stale quickly and refer to a not-really-new thing. Is there a better name?

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+        return {};
+    }
+    const auto *rhs_cast = rhs_broadcast->value.as<Cast>();
+    if (rhs_cast) {
+        if (op_type == AMXOpType::Int8) {
+            if (!(rhs_cast->value.type().element_of() == Int(8) || rhs_cast->value.type().element_of() == UInt(8))) {
+                user_assert(false) << "Expected rhs cast of i8/u8";
+            }
+        } else  // AMXOpType::Bf16
+        {
```
steven-johnson, line 202: Closing brace should be on previous line. (I'm surprised clang-format didn't catch this.)

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+    auto rhs = Call::make(rhs_type, "tile_load", {1, tile_y * tile_r * element_width, rhs_var, rhs_tile.base * element_width, rhs_tile.stride[0] * tile_y * element_width}, Call::Intrinsic);
+    auto res_type = amx_op_type_result_type(op_type);
+
+    // {rows, colbytes, acc, out, lhs, rhs}
+    auto out = Load::make(res_type, new_name, Ramp::make(0, 1, 256), {}, {}, const_true(256), {});
+
+    // 4 bytes for i32, f32
+    auto colbytes = tile_y * 4;
+    auto matmul =
+        Call::make(res_type, "tile_matmul", {tile_x, colbytes, tile_r, out, lhs, rhs}, Call::Intrinsic);
```
steven-johnson, line 256: No need to split this into two lines

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+                (op->type.is_float() && op->type.bits() == 32))
+                << "scheduled tile operations must yield 32-bit integers or 32-bit floats";
+
+            if (op->type.is_int() && op->type.bits() == 32) {
+                op_type = AMXOpType::Int8;
+            } else {
+                op_type = AMXOpType::Bf16;
+            }
+
+            user_assert(!in_allocate) << "Found two possible tile allocations for AMX allocation";
```
steven-johnson, line 322: Would it be helpful to append amx_name or tile_name to the error message, for debugging purposes?

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+            if (found_tile_x < 0 || found_tile_y < 0 || found_tile_r < 0) {
+                return op;
+            }
+            if (!pending_stores.empty()) {
+                // Really only need to go over the pending stores
+                body = mutate(body);
+            }
+
+            auto alloc_type = amx_op_type_result_type(op_type);
+
```
steven-johnson, line 339: No need for this blank line

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
...
+        return Free::make(amx_name);
+    }
+
+    Stmt visit(const ProducerConsumer *op) override {
+        if (op->name != tile_name) {
+            return IRMutator::visit(op);
+        }
+
+        auto body = mutate(op->body);
+        return ProducerConsumer::make(amx_name, op->is_producer, body);
```
steven-johnson, line 358: Nit: `std::move(body)` ?

test/performance/tiled_matmul.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    // schedule the consumer
+    Var mmxi("mmxi"), mmyi("mmyi");
+    mm.in()
+        .tile(x, y, mmxi, mmyi, tile_x, tile_y)
+        .vectorize(mmxi)
+        .vectorize(mmyi);
+
+    Func result = mm.in();
+    //result.print_loop_nest();
```
steven-johnson, line 227: Don't check in commented-out code (unless there is a comment explaining why, as is done elsewhere in this file)

steven-johnson (changes requested):
LGTM so far -- aside from style nits, I think it would be good to split the new test into correctness and performance tests, as Halide does for virtually all other features.

src/ExtractTileOperations.cpp
```diff
@@ -0,0 +1,414 @@
+#include "ExtractTileOperations.h"
+
+#include "IRMatch.h"  // expr_match
```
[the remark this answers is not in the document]
alexreinking: Speaking of which, it might be a good idea to run [IWYU](https://github.com/include-what-you-use/include-what-you-use) on our codebase...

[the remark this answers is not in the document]
steven-johnson: SGTM

frengels:
When converting to correctness tests there's a bit of a change in the pattern for the rhs load when using `Buffer` instead of `ImageParam`, so I'm having to take a bit extra time to make sure it's accepting all possible scenarios and generating correct code.

abadams:
lgtm. The pattern matching seems to be pretty ad-hoc and possibly brittle, but that can always be improved later. The checks for LLVM 12 will be removed pretty soon too.

PR #5822, opened 2021-03-18
Set CMake policy CMP0091 to NEW

mcleary:
This enables the use of CMAKE_MSVC_RUNTIME_LIBRARY to select which
version of the MSVC runtime to link against the application.
This greatly facilitates development on Windows since one can use LLVM
in release mode while keeping Halide in Debug mode by setting
-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded, or any value that matches
the runtime used to build LLVM.

More information on the policy: [CMP0091](https://cmake.org/cmake/help/latest/policy/CMP0091.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,8 @@
 cmake_minimum_required(VERSION 3.16...3.20)
+
+# Enables the use of CMAKE_MSVC_RUNTIME_LIBRARY to select the MSVC runtime
+cmake_policy(SET CMP0091 NEW)
+
 project(Halide VERSION 12.0.0)
 
 enable_testing()


steven-johnson:
LGTM (and also, awesome, I didn't know this was an option!), but will defer to @alexreinking

alexreinking:
This policy is already enabled by virtue of our minimum requirement of 3.16. Closing...

alexreinking:
For reference, the CMP0091 link you posted notes that the policy was introduced in CMake 3.15. The documentation for [`cmake_minimum_required`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings) reads:

> The `cmake_minimum_required(VERSION)` command implicitly invokes the [`cmake_policy(VERSION)`](https://cmake.org/cmake/help/latest/command/cmake_policy.html#setting-policies-by-cmake-version) command to specify that the current project code is written for the given range of CMake versions. All policies known to the running version of CMake and introduced in the `<min>` (or `<max>`, if specified) version or earlier will be set to use `NEW` behavior. All policies introduced in later versions will be unset. This effectively requests behavior preferred as of a given CMake version and tells newer CMake versions to warn about their new policies.

Since #5810, we enable _all policies_ through CMake 3.20 (because of changes to Ninja depfile management).

mcleary:
@alexreinking Thank you very much for the information. For some reason when I enabled CMAKE_MSVC_RUNTIME_LIBRARY for the first time I was still getting linker errors so I thought the policy was using the old behaviour, which is to ignore it. Perhaps my CMakeCache was not up do date of something. After trying again I can confirm the policy is indeed set to new.

PR #5829, opened 2021-03-20
Add align_extent(), to align extent but not min

steven-johnson:
This can be handy when you have an intermediate Func that is being tiled inside an outer Func and you want to ensure that it fits an exact multiple of tiles.

### DIFF
--- a/python_bindings/src/PyFunc.cpp
+++ b/python_bindings/src/PyFunc.cpp
@@ -325,6 +325,7 @@ void define_func(py::module &m) {
             .def("set_estimates", &Func::set_estimates, py::arg("estimates"))
 
             .def("align_bounds", &Func::align_bounds, py::arg("var"), py::arg("modulus"), py::arg("remainder") = 0)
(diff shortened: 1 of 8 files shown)


dsharletg (approved):
This seems OK to me, but @abadams should take a look. I'm not 100% sure that aligning bounds to a modulus without caring about the remainder should mean that the extent is a multiple of the modulus? But it does seem reasonable.

abadams:
The within-compiler representation seems reasonable, but I think it would be better to change the API to be align_extent for this case (to mimic bound_extent)

steven-johnson:
This ended up not being as useful as I thought it would be for my purposes. Adding `align_extent` sounds fine if we want to land this, but without a clear and present need, do we want to?

steven-johnson:
Still not sure if this is genuinely needed, but moved to a separate method as suggested.

abadams:
I'm sure it will come up as useful at some point, and it's simple enough.

abadams (approved):

PR #5830, opened 2021-03-20
Improve performance on some pathological cases

steven-johnson:
Conv2D with small channels or width fell into pathological code. This makes the worst case much more reasonable.

### DIFF
--- a/apps/hannk/halide/convolution_generator.cpp
+++ b/apps/hannk/halide/convolution_generator.cpp
@@ -190,6 +190,9 @@ public:
         // things computed at the tile to have constant size. We can't assume the
         // output is bigger than a minimum size. So, we specialize for decreasing
         // tile sizes, and have a degenerate tile case to handle the rest.
+        //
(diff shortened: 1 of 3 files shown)


apps/hannk/halide/convolution_generator.cpp
```diff
@@ -190,6 +190,9 @@ class Convolution : public Generator<Convolution> {
         // things computed at the tile to have constant size. We can't assume the
         // output is bigger than a minimum size. So, we specialize for decreasing
         // tile sizes, and have a degenerate tile case to handle the rest.
+        //
+        // TODO: this is actually faster in some cases if we just use vector_size=8
+        // (even on AVX2 / Skylake). Should investigate further.
```
dsharletg, line 195: I realized why this is: on Intel, vector_reduction is 2, so vector_size is too big for the edge case. This is fine for now, it will take some more thought/specializations to get this right for the small number of channels output case.

dsharletg (approved):

PR #5833, opened 2021-03-22
Enable targeting broadcasting dot products on ARM

dsharletg:
It's currently possible to write schedules that generate the necessary shuffles for LLVM to generate the broadcasting dot product instructions, but LLVM only recognizes these when the shuffle is the second operand argument. This PR commutes the operands of dot products to make it possible to reliably generate these instructions.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -1162,6 +1162,12 @@ void CodeGen_ARM::codegen_vector_reduce(const VectorReduce *op, const Expr &init
             if (!i.defined()) {
                 i = make_zero(op->type);
             }
+            if (const Shuffle *s = matches[0].as<Shuffle>()) {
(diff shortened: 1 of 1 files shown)


src/CodeGen_ARM.cpp
```diff
@@ -1162,6 +1162,12 @@ void CodeGen_ARM::codegen_vector_reduce(const VectorReduce *op, const Expr &init
             if (!i.defined()) {
                 i = make_zero(op->type);
             }
+            if (const Shuffle *s = matches[0].as<Shuffle>()) {
+                if (s->is_broadcast()) {
+                    // LLVM wants the broadcast as the second operand.
```
steven-johnson, line 1167: Maybe augment this comment with text from the PR description to underscore that this is critical for dot-product generation.

steven-johnson (approved):

PR #5836, opened 2021-03-22
Add simplifier rules helpful for specialization

dsharletg:
This PR makes it easier to simplify the expressions resulting from bounds inference on funcs with specializations.

### DIFF
--- a/src/Simplify_Max.cpp
+++ b/src/Simplify_Max.cpp
@@ -173,6 +173,22 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(x, max(min(x, y), z)), max(x, z)) ||
              rewrite(max(x, max(min(y, x), z)), max(x, z)) ||
 
+             rewrite(max(select(x, y, z), z), select(x, max(y, z), z)) ||
+             rewrite(max(select(x, y, z), y), select(x, y, max(z, y))) ||
+             rewrite(max(z, select(x, y, z)), select(x, max(z, y), z)) ||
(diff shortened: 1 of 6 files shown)


src/Simplify_Max.cpp
```diff
@@ -173,6 +173,22 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(x, max(min(x, y), z)), max(x, z)) ||
              rewrite(max(x, max(min(y, x), z)), max(x, z)) ||
 
+             rewrite(max(select(x, y, z), z), select(x, max(y, z), z)) ||
```
abadams, line 176: @jn80842 I believe this rule is OK because we prefer stronger IR nodes at the root of the true, all else equal. Is that right?

abadams:
The rules all formally verify, but I'd like @jn80842 to double-check the one max/select distribution rule that seems barely-ordered to me.

steven-johnson:

src/Simplify_Max.cpp
abadams: Ok, according to Julie this rule is actually not OK because we prefer weaker nodes at the root, so this may fight with another rule that moves a max inside a select for some reason.

dsharletg: Hmm, I'm going to have to look at this, because I did check that these rules are needed, and they really are. I guess I'll have to try to find a more specific rule.

abadams:
The new rules all formally verify

abadams (approved):

PR #5838, opened 2021-03-23
Allow clean compiles when NDEBUG is defined

steven-johnson:
Mostly just removing unused-variable warnings, to avoid pain in other build systems. Also modify the Makefile to ensure that NDEBUG is actually defined if OPTIMIZE contains `-O[something]`.


slomp (approved):
LGTM

PR #5843, opened 2021-03-23
 Improve wasm_threads documentation

steven-johnson:
The intent of this PR is to make the (considerably) limitations of using threads in wasm more apparent.

Background: wasm_threads  relies specifically on wasm runtime providing an implementation of the `pthreads` API; at this time, this is only (AFAICT) available in wasm when compiling with Emscripten, *and* targeting a web/browser environment. (What's happening under the hood is that Emscripten is taking great pains to use WebWorkers to implement pthreads.)

In other wasm runtime environments (e..g WASI), there is still no story for useful thread usage; in particular, WASI has no provision for creating or joining threads.

### DIFF
--- a/README_webassembly.md
+++ b/README_webassembly.md
@@ -12,8 +12,8 @@ some limitations. Some of the most important:
 -   Sign-extension operations can be enabled via Target::WasmSignExt.
 -   Non-trapping float-to-int conversions can be enabled via
     Target::WasmSatFloatToInt.
--   Threads are not available yet. We'd like to support this in the future but
-    don't yet have a timeline.
+-   Threads have very limited support via Target::WasmThreads; see
+    [below](#using-threads) for more details.
 -   Halide's JIT for Wasm is extremely limited and really useful only for
     internal testing purposes.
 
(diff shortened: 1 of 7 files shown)


steven-johnson:
Gentle review ping

steven-johnson:
Gentle review ping

abadams:
In the future, if we have different ways to support threading in wasm when running on different platforms, should that be more wasm_* flags, or would it be better to group it under a single wasm_threads?

steven-johnson:
> In the future, if we have different ways to support threading in wasm when running on different platforms, should that be more wasm_* flags, or would it be better to group it under a single wasm_threads?

Honestly, I'm not sure. Threading in wasm still seems to be a major work-in-progress, with pthread-emulation-via-emscripten-but-only-in-a-browser the only game in town. Until we see what happens with WASI (ie, will they add proper threading support, and if so, what will it look like and when will it arrive), it's all just guesswork.

It may be that this PR's approach isn't the right one, but at a minimum we need to improve the documentation to make the limitations of threading under wasm more clear.

abadams:
I guess the ideal outcome is that the target flag goes away entirely because .parallel just works on all wasm platforms. Maybe we should just document it better and otherwise wait and see.

steven-johnson:
> I guess the ideal outcome is that the target flag goes away entirely because .parallel just works on all wasm platforms. Maybe we should just document it better and otherwise wait and see.

The target flag isn't gonna go away anytime soon (if ever); threads aren't part of baseline wasm, so we'd have to decide that our wasm generation assumes features not in baseline. That may be a reasonable decision but we're not there yet. In the meantime, I'll just revise the documentation and drop the rest of this.

steven-johnson:
PTAL

abadams (approved):

PR #5847, opened 2021-03-24
Revise hannk::Tensor to use H::R::Buffer<>

steven-johnson:
We originally used a bespoke mix of data and dimensions, wrapping Buffers around them on the fly; this just uses Buffer as the fundamental type, simplifying things. As a bonus, it allows us to avoid copying read-only data (since Buffer can just reference the external pointer directly), making #5671 obsolete.

### DIFF
--- a/apps/hannk/interpreter/model.h
+++ b/apps/hannk/interpreter/model.h
@@ -105,35 +105,22 @@ inline std::ostream &operator<<(std::ostream &s, const QuantizationInfo &q) {
     return s << "{" << q.scale << ", " << q.zero << ", " << q.dimension << "}";
 }
 
-inline Box without_strides(const std::vector<halide_dimension_t> &shape) {
-    Box result;
-    result.reserve(shape.size());
-    for (const halide_dimension_t &i : shape) {
-        result.emplace_back(i);
-    }
-    return result;
-}
-
 class Tensor {
     std::string name_;
     TensorType type_;
-    std::vector<halide_dimension_t> shape_;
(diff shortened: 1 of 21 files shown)


apps/hannk/interpreter/model.h
```diff
@@ -105,129 +105,121 @@ inline std::ostream &operator<<(std::ostream &s, const QuantizationInfo &q) {
...
-    result.reserve(shape.size());
-    for (const halide_dimension_t &i : shape) {
-        result.emplace_back(i);
-    }
-    return result;
-}
-
 class Tensor {
     std::string name_;
     TensorType type_;
```
dsharletg, line 110: Do we still need this? buffer().type() might suffice? It would lose support for string, complex, etc. but I'm not sure we need those anyways.

dsharletg (commented):
This isn't going to run the buildbots (the buildbots don't run this branch).

apps/hannk/interpreter/model.h
steven-johnson: Honestly, not sure. Adding support for those types would range form tricky-but-doable (for complex) to impossible (string). There's only a handful of places that inspect the type after creation. Let me try changing them.

steven-johnson: After looking a bit, I think the answer is that we can probably do away with hannk::TensorType entirely, and just use halide_type_t instead, but I think that would probably be cleaner to do in a subsequent PR.

PR #5849, opened 2021-03-25
Add support for widening_mul in wasm

steven-johnson:
Add support for widening_mul in wasm

### DIFF
--- a/src/runtime/wasm_math.ll
+++ b/src/runtime/wasm_math.ll
@@ -44,16 +44,10 @@ define weak_odr <4 x float> @fast_inverse_sqrt_f32x4(<4 x float> %x) nounwind al
 
 declare <2 x i64> @llvm.wasm.extmul.high.signed.v2i64(<4 x i32>, <4 x i32>);
 declare <2 x i64> @llvm.wasm.extmul.low.signed.v2i64(<4 x i32>, <4 x i32>);
-declare <2 x u64> @llvm.wasm.extmul.high.signed.v2u64(<4 x u32>, <4 x u32>);
-declare <2 x u64> @llvm.wasm.extmul.low.signed.v2u64(<4 x u32>, <4 x u32>);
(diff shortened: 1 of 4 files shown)


src/runtime/wasm_math.ll
```diff
@@ -39,3 +40,110 @@ define weak_odr <4 x float> @fast_inverse_sqrt_f32x4(<4 x float> %x) nounwind al
...
+
+declare <2 x i64> @llvm.wasm.extmul.high.signed.v2i64(<4 x i32>, <4 x i32>);
+declare <2 x i64> @llvm.wasm.extmul.low.signed.v2i64(<4 x i32>, <4 x i32>);
+declare <4 x i32> @llvm.wasm.extmul.high.signed.v4i32(<8 x i16>, <8 x i16>);
+declare <4 x i32> @llvm.wasm.extmul.low.signed.v4i32(<8 x i16>, <8 x i16>);
+declare <8 x i16> @llvm.wasm.extmul.high.signed.v8i16(<16 x i8>, <16 x i8>);
+declare <8 x i16> @llvm.wasm.extmul.low.signed.v8i16(<16 x i8>, <16 x i8>);
+
+; i8 -> i16
+define weak_odr <8 x i16> @widening_mul_i8x8(<8 x i8> %x, <8 x i8> %y) nounwind alwaysinline {
```
dsharletg, line 53: Do we actually need these narrow versions? I would think if we just only had the wider version, things would simplify pretty easily (the extmul.high call would be unused, the output shuffle would simplify with the slicing, and the input shuffles would be generated by our slicing).

steven-johnson: Ah, good point, I hadn't considered that. Let me check.

steven-johnson: Yep, you're right. PTAL.

dsharletg (approved):

PR #5850, opened 2021-03-25
Add support for pairwise_widening_add in wasm

steven-johnson:


test/correctness/simd_op_check.cpp
```diff
@@ -1767,11 +1767,24 @@ class SimdOpCheck : public SimdOpCheckTest {
...
+                    // widening summations into just widening summations.
+                    auto sum_ = [&](Expr e) {
+                        Func f;
+                        f(x) = cast(e.type(), 123);
+                        f(x) += e;
+                        return f(x);
+                    };
+
+                    check("i16x8.extadd_pairwise_i8x16_s", 8 * w, sum_(i16(in_i8(f * x + r))));
+                    check("i16x8.extadd_pairwise_i8x16_u", 8 * w, sum_(u16(in_u8(f * x + r))));
```
dsharletg, line 1784: I think we should also check for signed widening from unsigned operands, e.g.: https://github.com/halide/Halide/blob/master/src/CodeGen_ARM.cpp#L1139

src/CodeGen_WebAssembly.cpp
```diff
@@ -110,6 +117,59 @@ void CodeGen_WebAssembly::init_module() {
     }
 }
 
+void CodeGen_WebAssembly::codegen_vector_reduce(const VectorReduce *op, const Expr &init) {
+    const int factor = op->value.type().lanes() / op->type.lanes();
+
+    const char *intrin = nullptr;
+    std::vector<Expr> intrin_args;
+    Expr accumulator = init;
+    if (op->op == VectorReduce::Add && factor == 2) {
```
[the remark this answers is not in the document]
steven-johnson: OK, but FWIW, this was taken directly from the existing code in Codegen_ARM, which doesn't use a pattern here. (Maybe I should backport it there afterwards...)

src/CodeGen_WebAssembly.cpp
```diff
@@ -110,6 +117,59 @@ void CodeGen_WebAssembly::init_module() {
...
+            if (accumulator.defined()) {
+                // We still have an initial value to take care of
+                string n = unique_name('t');
+                sym_push(n, value);
+                Expr v = Variable::make(accumulator.type(), n);
+                switch (op->op) {
+                case VectorReduce::Add:
+                    accumulator += v;
+                    break;
+                case VectorReduce::Min:
```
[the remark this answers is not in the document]
steven-johnson: Yeah, copypasta from CodegenARM and I was lazy about cutting it down. Will do.

src/CodeGen_WebAssembly.cpp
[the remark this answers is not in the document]
steven-johnson: (the source in Codegen_ARM has the comment: `// TODO: Move this to be patterns? The patterns are pretty trivial, but some of the other logic is tricky`, which got lost in translation.)

test/correctness/simd_op_check.cpp
steven-johnson: There isn't a wasm instruction that does this -- just i8->i16 and u8->u16 (and wider variants). 

I guess we could just use the u8->u16 op in this case and cast the result to i16, on the assumption that the u16 result will always have a zero sign bit for 128-bit vectors....

dsharletg: Yes, that's all the ARM and other targets are doing anyways. I'm not actually sure you need to change the implementation at all, it might just work (and the ARM version might be redundant too), but I think we should test for it.

steven-johnson:
PTAL

steven-johnson:
Monday Morning Review Ping

src/CodeGen_WebAssembly.cpp
```diff
@@ -1,15 +1,19 @@
...
+#include "ConciseCasts.h"
+#include "IRMatch.h"
+#include "IROperator.h"
 #include "LLVM_Headers.h"
-#include <sstream>
 
 namespace Halide {
 namespace Internal {
 
-using std::string;
```
dsharletg, line 9: Did you mean to change this here? Most of the other backends do this (and don't quality string with std::).

src/CodeGen_WebAssembly.cpp
```diff
@@ -110,13 +125,114 @@ void CodeGen_WebAssembly::init_module() {
...
 
-string CodeGen_WebAssembly::mcpu() const {
+void CodeGen_WebAssembly::codegen_vector_reduce(const VectorReduce *op, const Expr &init) {
+#if LLVM_VERSION >= 130
+    struct Pattern {
+        VectorReduce::Operator reduce_op;
+        int factor;
+        Expr pattern;
+        const char *intrin;
+        Type narrow_type;
```
dsharletg, line 135: I don't think you need this, all of the patterns have the narrow type the same as the pattern's type.

src/CodeGen_WebAssembly.cpp
```diff
@@ -110,13 +125,114 @@ void CodeGen_WebAssembly::init_module() {
...
+        uint32_t flags;
```
dsharletg, line 136: You don't need this either, no patterns use any flags.

src/CodeGen_WebAssembly.cpp
steven-johnson: There was a mixture of usage so I normalize to std:: since that is the pattern in most non-backend code. Will normalize the other way if that's the unspoken standard here.

src/CodeGen_WebAssembly.cpp
steven-johnson: At the moment, yes, but I'm not sure if that will be the case for everything, and I wanted to preserve this logic (adapted from Codegen_X86.cpp). That said, easy enough to restore if/when it's needed.

src/CodeGen_WebAssembly.cpp
steven-johnson: See above, ditto.

src/CodeGen_WebAssembly.cpp
dsharletg: I think we should just add it when needed. It adds a fair bit of code that is currently dead. Note that CodeGen_ARM doesn't have either of these.

test/correctness/simd_op_check.h
```diff
@@ -94,27 +94,32 @@ class SimdOpCheckTest {
...
         bool found_it = false;
 
         std::ostringstream msg;
         msg << op << " did not generate for target=" << target.to_string() << " vector_width=" << vector_width << ". Instead we got:\n";
 
-        std::string line;
-        while (getline(asm_file, line)) {
-            msg << line << "\n";
+        {
+            std::ifstream asm_file;
```
dsharletg, line 103: What actually changed here?

steven-johnson: code restructuring that no longer matters. Will revert.

dsharletg (approved):

PR #5853, opened 2021-03-25
Add support for i16x8.q15mulr_sat_s in wasm

steven-johnson:
Also, some drive-by clarifications to other wasm-simd instructions in simd_op_check -- some of the yet-to-be-implemented ones are of dubious use in Halide and may not be worth implementing.


dsharletg (approved):

steven-johnson:
(needs to wait for https://github.com/halide/Halide/pull/5850 to land first)

PR #5854, opened 2021-03-26
Add support for wasm-simd saturating-narrow ops.

steven-johnson:
(Cannot land until https://github.com/halide/Halide/pull/5853 lands.)

### DIFF
--- a/src/runtime/wasm_math.ll
+++ b/src/runtime/wasm_math.ll
@@ -44,16 +44,10 @@ define weak_odr <4 x float> @fast_inverse_sqrt_f32x4(<4 x float> %x) nounwind al
 
 declare <2 x i64> @llvm.wasm.extmul.high.signed.v2i64(<4 x i32>, <4 x i32>);
 declare <2 x i64> @llvm.wasm.extmul.low.signed.v2i64(<4 x i32>, <4 x i32>);
-declare <2 x u64> @llvm.wasm.extmul.high.signed.v2u64(<4 x u32>, <4 x u32>);
-declare <2 x u64> @llvm.wasm.extmul.low.signed.v2u64(<4 x u32>, <4 x u32>);
 declare <4 x i32> @llvm.wasm.extmul.high.signed.v4i32(<8 x i16>, <8 x i16>);
 declare <4 x i32> @llvm.wasm.extmul.low.signed.v4i32(<8 x i16>, <8 x i16>);
-declare <4 x u32> @llvm.wasm.extmul.high.signed.v4u32(<8 x u16>, <8 x u16>);
-declare <4 x u32> @llvm.wasm.extmul.low.signed.v4u32(<8 x u16>, <8 x u16>);
 declare <8 x i16> @llvm.wasm.extmul.high.signed.v8i16(<16 x i8>, <16 x i8>);
 declare <8 x i16> @llvm.wasm.extmul.low.signed.v8i16(<16 x i8>, <16 x i8>);
-declare <8 x u16> @llvm.wasm.extmul.high.signed.v8u16(<16 x u8>, <16 x u8>);
-declare <8 x u16> @llvm.wasm.extmul.low.signed.v8u16(<16 x u8>, <16 x u8>);
 
 ; i8 -> i16
 define weak_odr <8 x i16> @widening_mul_i8x8(<8 x i8> %x, <8 x i8> %y) nounwind alwaysinline {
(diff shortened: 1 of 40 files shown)


test/correctness/simd_op_check.h
```diff
@@ -281,6 +282,10 @@ class SimdOpCheckTest {
                 }
 
                 error_file.close();
+
+                // Occasionally useful for debugging: stop at the first failure.
+                // std::cerr << error_msg.str();
+                // exit(-1);
```
dsharletg, line 288: Intended to submit this?

BTW, this is what I use HL_SIMD_OP_CHECK_FILTER for.

src/runtime/wasm_math.ll
```diff
@@ -104,4 +104,61 @@ define weak_odr <4 x i64> @widening_mul_u32x4(<4 x i32> %x, <4 x i32> %y) nounwi
...
+}
+
+define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwind alwaysinline {
+  %1 = shufflevector <8 x i32> %x, <8 x i32> undef, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
+  %2 = shufflevector <8 x i32> %x, <8 x i32> undef, <4 x i32> <i32 4, i32 5, i32 6, i32 7>
+  %3 = tail call <8 x i16> @llvm.wasm.narrow.unsigned.v8i16.v4i32(<4 x i32> %1, <4 x i32> %2)
+  ret <8 x i16> %3
+}
+
+; The wasm-simd ops always treat inputs as signed; clear the high bit for unsigned inputs to get the right answer.
```
dsharletg, line 142: I don't think this is correct with 2's complement arithmetic?

I'm not sure it's faster than just not using a pattern. It's just a min followed by a narrow, vs. this implementation is a bitwise op followed by a narrow. Unless we know something like this is faster, it's better to just not do anything special for this.

test/correctness/simd_op_check.h
steven-johnson: Yeah, I did intend to -- the issue here is that even with HL_SIMD_OP_CHECK_FILTER you can run several tests (different vectorize widths, different factors, etc.). I'll just remove these if it's a nuisance, but I thought it was a handy reminder of where to insert these checks.

src/runtime/wasm_math.ll
steven-johnson: > I don't think this is correct with 2's complement arithmetic?

Yeah, you're right -- I was thinking that anything with the high bit set should be clamped to the max value (true!) but this would be wrong for the case of 0x8001, which should clamp to 0xff, but would become 0x01 with this logic. 

> I'm not sure it's faster than just not using a pattern. It's just a min followed by a narrow, vs. this implementation is a bitwise op followed by a narrow. Unless we know something like this is faster, it's better to just not do anything special for this.

Leaving it without a specialization causes scalarization of the operation, which is pretty terrible, even if this is a rarely-used operation. Replacing the bitmask with a vector-min operation is perhaps the right thing here.

dsharletg: Why would this scalarize? If we don't pattern match this, it should just be a vector min followed by a (non-saturating) vector narrowing.

steven-johnson: Because the LLVM wasm backend doesn't (yet) have the logic to intelligently vectorize any of these (which is why we're special casing them in the first place). Maybe it will eventually, but we don't need to wait.

dsharletg: Does this mean that simple unsaturated narrowing casts are scalarizing on wasm?

steven-johnson: Last time I checked, yes. (implementing those directly is next on my list; I didn't actually realize that these instructions were saturating until I read the implementation details.)

steven-johnson: Checking this morning, the codegen for `u16(u32_1)` with vectorize = 8 or higher (in recent LLVM13 ) ends up with code with lots of `extract_lane` and `replace_lane` ops, which is, distressingly suboptimal when a plain old `shuffle` will do the trick. I'll offer a PR to do non-saturating narrows before this, since it will make the remaining saturating cases trivial.

steven-johnson: Update: it appears that the current LLVM wasm backend really wants to emit these sort of operations as extract/replace-lane pairs, even if I try to get it to emit (say) shuffle operations directly. Not sure if this is a bug in my code, or a bug in the LLVM backend, or maybe a feature in the LLVM backend (i.e. perhaps current wasm VMs generate better code for those operations via that sequence?) -- I've contacted the relevant folks to get clarity, but until I do, I'm going to just skip the unsigned->unsigned specialization attempt entirely, so we can land the customization for the existing ops.

dsharletg (approved):

PR #5856, opened 2021-03-26
Add TailStrategy::Predicate

dsharletg:
This PR adds a new `TailStrategy::Predicate`. This is very similar to `TailStrategy::GuardWithIf`, except it uses predicated loops unconditionally, instead of heuristically choosing based on the target.

On some targets, we think using predicated loads/stores is always faster than scalarizing (Hexagon). On these targets, we can just use predicated loads/stores for `TailStrategy::GuardWithIf`. On these targets, using `TailStrategy::Predicate` will not change anything.

On most other targets, predicated loads/stores are expensive (emulated with scalar code). In particular, the cost of inserting and extracting the scalars to/from vectors can be high. On these targets, Halide scalarizes the whole body of the loop instead. This is unfortunate in some cases, when the loop is very expensive, it can be worth the added overhead to assemble the scalarized loads/stores into vectors, so we can vectorize the expensive computation. This PR allows programmers to ask for this with the new tail strategy.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1232,9 +1232,7 @@ private:
 
             interval.min = Interval::make_max(interval.min, lower.min);
             interval.max = Interval::make_min(interval.max, upper.max);
-        } else if (op->is_intrinsic(Call::likely) ||
-                   op->is_intrinsic(Call::likely_if_innermost)) {
-            internal_assert(op->args.size() == 1);
+        } else if (Call::as_tag(op)) {
             op->args[0].accept(this);
         } else if (op->is_intrinsic(Call::return_second)) {
             internal_assert(op->args.size() == 2);
@@ -2476,10 +2474,8 @@ private:
             for (const auto &pair : cases) {
                 Expr c = pair.first;
                 Stmt body = pair.second;
-                const Call *call = c.as<Call>();
-                if (call && (call->is_intrinsic(Call::likely) ||
-                             call->is_intrinsic(Call::likely_if_innermost) ||
-                             call->is_intrinsic(Call::strict_float))) {
+                const Call *call = Call::as_tag(c);
+                if (call) {
                     c = call->args[0];
                 }
 
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -1580,6 +1580,7 @@ private:
                 interval.min *= factor;
             }
             break;
+        case VectorReduce::SaturatingAdd:
         case VectorReduce::Mul:
             // Technically there are some things we could say
             // here. E.g. if all the lanes are positive then we're
--- a/src/IROperator.h
+++ b/src/IROperator.h
@@ -307,6 +307,10 @@ Expr remove_likelies(const Expr &e);
  * all calls to likely() and likely_if_innermost() removed. */
 Stmt remove_likelies(const Stmt &s);
 
+/** If the expression is a tag helper call, remove it and return
+ * the tagged expression. If not, returns the expression. */
+Expr strip_tags(const Expr &e);
+
 // Secondary args to print can be Exprs or const char *
 inline HALIDE_NO_USER_CODE_INLINE void collect_print_args(std::vector<Expr> &args) {
 }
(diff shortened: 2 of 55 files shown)


src/IROperator.h
```diff
@@ -1375,6 +1379,12 @@ Expr likely(Expr e);
  * found in an innermost loop. */
 Expr likely_if_innermost(Expr e);
 
+/** Expressions tagged with this intrinsic are suggestions that
+ * vectorization of loops with guard ifs should be implemented with
+ * non-faulting predicated loads and stores, instead of scalarizing
+ * an if statement. */
+Expr predicate(Expr e);
```
abadams, line 1386: Shouldn't this be in the internal namespace?

abadams:
The existence of a static Call::as_tag that takes an Expr sort of implies that Tags should be their own IR nodes rather than a call type. What do you think the pros/cons of making Tag a separate IRNode are? E.g. would it duplicate a bunch of code from Call handlers and methods?

src/IROperator.h
dsharletg: I didn't realize likely_if_innermost wasn't in the internal namespace!

dsharletg:
I don't know that Call::as_tag implies tags should be their own IR node. I do think that might make sense regardless.

Pros:
- The IR node could maintain a list of tags, so the ordering of tags shouldn't matter (in this PR, `predicate(likely(x))` is different than `likely(predicate(x))`...).
- It would be very slightly easier to get at tags (`e.as<Tags>()` vs. `Call::as_tag(e)`), but only *very* slightly.

Cons:
- New IR has a lot of overhead. Calls are transparent to most mutators, new IR is not.

I think that con is pretty big, and the pros are pretty small. At the very least, we should make that change in a separate PR I think.

src/Bounds.cpp
```diff
@@ -2476,10 +2474,8 @@ class BoxesTouched : public IRGraphVisitor {
             for (const auto &pair : cases) {
                 Expr c = pair.first;
                 Stmt body = pair.second;
-                const Call *call = c.as<Call>();
-                if (call && (call->is_intrinsic(Call::likely) ||
-                             call->is_intrinsic(Call::likely_if_innermost) ||
-                             call->is_intrinsic(Call::strict_float))) {
+                const Call *call = Call::as_tag(c);
```
abadams, line 2477: ```if (const Call *call == Call::as_tag(c)) {```

src/IROperator.h
```diff
@@ -307,6 +307,16 @@ Expr remove_likelies(const Expr &e);
  * all calls to likely() and likely_if_innermost() removed. */
 Stmt remove_likelies(const Stmt &s);
 
+/** If the expression is a tag helper call, remove it and return
```
abadams, line 310: remove_likelies is recursive (i.e. it's an IRMutator), but strip tags is only shallowly recursive. This seems like a potential source of error. Should strip_tags just be a full mutator?

src/Bounds.cpp
dsharletg: `call` is used below :) I made this change at first and had to revert it for that.

src/Schedule.h
```diff
@@ -48,6 +48,17 @@ enum class TailStrategy {
...
     GuardWithIf,
 
+    /** Guard the inner loop with an if statement that prevents
+     * evaluation beyond the original extent, with a hint that the
+     * if statement should be implemented with predicated operations.
+     * Always legal. The if statement is treated like a boundary
+     * condition, and factored out into a loop epilogue if possible.
+     * Pros: no redundant re-evaluation; does not constrain input our
+     * output sizes. Cons: increases code size due to separate
+     * tail-case handling; vectorization will scalarize in the tail
```
abadams, line 58: Comment seems incorrect at the end.

src/IROperator.h
dsharletg: This helper is used in place of code that used to do things like:
```
if (const Call *c = as_likely(x)) {
  x = c->args[0];
}
...
```

Where x could be used recursively where the more deeply nested tags might be needed, so I think `strip_tags` needs to have a different meaning than `remove_likelies`.

I have only seen a use case for recursively removing all tags for a specific tag: there's `remove_likelies`, and there is a one-off mutator to remove `predicate`, but these are different than the use cases for `strip_tags`.

dsharletg: (sorry, messages crossed there.)

What do you think of `unwrap_tags`?

abadams (approved):

PR #5858, opened 2021-03-29
Added support for ClangCL

emmenlau:
This PR makes minor changes to the build in order to support ClangCl on VS. The required change is mostly about the stack size. Currently MSVC requires a higher stack size, and this setting was in `Util.h`. However clang does not support the `pragma`. Therefore this PR sets the same flag via CMake compiler options *for the executables built by this project*. The flag is currently not passed transitively to downstream projects. And it is not discriminating between executables in the Halide project - all executables get the higher stack size. This seems to be correct at least for `tutorial`, `test` and `benchmark` executables.

To be discussed:
 - What other build systems (than CMake) need to be supported?
    - Quoting from Alex Reinking (Mar 26 20:15) on Gitter: We only support CMake on Windows and that flag does not affect our non-Windows users (due to _MSC_VER check), so it's ok to move.
    - Therefore it seems the CMake build provides the same functionality as the previous solution *except* it may handle differently the case of passing the stack size to downstream projects. See the next question below.
 - Should the stack size be passed transitively to downstream projects?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,11 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Increase the stack size for MSVC builds:
+if(MSVC)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:8388608")
+endif()
+
 # Build Halide with ccache if the package is present
 option(Halide_CCACHE_BUILD "Set to ON for a ccache enabled build" OFF)
 if (Halide_CCACHE_BUILD)
@@ -60,6 +65,7 @@ if (Halide_CCACHE_BUILD)
         message(FATAL_ERROR "Unable to find the program ccache. Set Halide_CCACHE_BUILD to OFF")
     endif ()
 endif ()
+
 
 ##
 # Import dependencies
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -345,6 +345,9 @@ add_subdirectory(runtime)
 # Build the Halide mono-header.
 ##
 
+set(CXX_COMPILER_IS_GNU_OR_CLANG $<CXX_COMPILER_ID:GNU,Clang,AppleClang>)
+set(CXX_COMPILER_IS_GNU_OR_CLANG_AND_NOT_CLANGCL $<AND:${CXX_COMPILER_IS_GNU_OR_CLANG},$<NOT:$<STREQUAL:${CMAKE_CXX_SIMULATE_ID},MSVC>>>)
+
 set(HALIDE_H "${Halide_BINARY_DIR}/include/Halide.h")
 set(LICENSE_PATH "${Halide_SOURCE_DIR}/LICENSE.txt")
 add_custom_command(OUTPUT "${Halide_BINARY_DIR}/include/Halide.h"
@@ -372,7 +375,7 @@ target_link_libraries(Halide PUBLIC Halide::LanguageOptions)
 target_compile_definitions(Halide
                            PRIVATE
                            $<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,STATIC_LIBRARY>:Halide_STATIC_DEFINE>
-                           $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:WITH_INTROSPECTION>)
+                           $<${CXX_COMPILER_IS_GNU_OR_CLANG_AND_NOT_CLANGCL}:WITH_INTROSPECTION>)
 
 include(TargetExportScript)
 ## TODO: implement something similar for Windows/link.exe
@@ -406,10 +409,10 @@ target_include_directories(Halide INTERFACE ${Halide_INCLUDE_PATH})
 
 target_compile_options(Halide
                        PRIVATE
-                       $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wall>
-                       $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-unused-function>
-                       $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wcast-qual>
-                       $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wignored-qualifiers>
+                       $<${CXX_COMPILER_IS_GNU_OR_CLANG_AND_NOT_CLANGCL}:-Wall>
+                       $<${CXX_COMPILER_IS_GNU_OR_CLANG}:-Wno-unused-function>
+                       $<${CXX_COMPILER_IS_GNU_OR_CLANG}:-Wcast-qual>
+                       $<${CXX_COMPILER_IS_GNU_OR_CLANG}:-Wignored-qualifiers>
 
                        $<$<COMPILE_LANG_AND_ID:CXX,GNU,Clang,AppleClang>:-Woverloaded-virtual>
                        $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wsuggest-override>
--- a/src/Util.h
+++ b/src/Util.h
@@ -44,11 +44,6 @@
 #define HALIDE_NO_USER_CODE_INLINE HALIDE_NEVER_INLINE
 #endif
 
-// On windows, Halide needs a larger stack than the default MSVC provides
-#ifdef _MSC_VER
-#pragma comment(linker, "/STACK:8388608,1048576")
-#endif
-
 namespace Halide {
 
 /** Load a plugin in the form of a dynamic library (e.g. for custom autoschedulers).


CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Increase the stack size for MSVC builds:
+if(MSVC)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:8388608")
+endif()
+
```
alexreinking, line 49: Generally speaking, I think we should keep our hands off these variables. Would you please investigate how MSVC handles multiple `/STACK` arguments? I think that will inform how to proceed. We probably want it to propagate since it is in fact a usage requirement on Windows.

src/CMakeLists.txt
```diff
@@ -345,6 +345,9 @@ add_subdirectory(runtime)
 # Build the Halide mono-header.
 ##
 
+set(CXX_COMPILER_IS_GNU_OR_CLANG $<CXX_COMPILER_ID:GNU,Clang,AppleClang>)
+set(CXX_COMPILER_IS_GNU_OR_CLANG_AND_NOT_CLANGCL $<AND:${CXX_COMPILER_IS_GNU_OR_CLANG},$<NOT:$<STREQUAL:${CMAKE_CXX_SIMULATE_ID},MSVC>>>)
+
```
alexreinking, line 350: I'm very concerned about using `CMAKE_CXX_SIMULATE_ID`, which is undocumented, _especially_ without a comment. I wonder if the right place for ClangCL settings is in a toolchain or preset.

src/Util.h
```diff
@@ -44,11 +44,6 @@
 #define HALIDE_NO_USER_CODE_INLINE HALIDE_NEVER_INLINE
 #endif
 
-// On windows, Halide needs a larger stack than the default MSVC provides
-#ifdef _MSC_VER
-#pragma comment(linker, "/STACK:8388608,1048576")
-#endif
-
```
alexreinking, line 51: Where did `1048576` go? It's the default reservation size, so this might cause extra increases to the stack size. Will that have a meaningful impact on start-up times?

alexreinking (changes requested):

alexreinking:
> * Should the stack size be passed transitively to downstream projects?

If Halide on Windows requires a larger stack size to not crash, then yes, I believe it should because it is therefore a _usage requirement_.

CMakeLists.txt
```diff
@@ -61,6 +66,7 @@ if (Halide_CCACHE_BUILD)
     endif ()
 endif ()
 
+
```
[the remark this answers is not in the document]
emmenlau: Yes, good catch.

src/Util.h
emmenlau: I see that it would be more consistent to port the commit size to cmake, too. After reading the description of the commit size on https://docs.microsoft.com/en-us/cpp/build/reference/stack?view=msvc-160 I did not feel confident that this needs to be set, but I forgot to bring it up in the discussion.

src/CMakeLists.txt
emmenlau: Its actually documented since cmake 3.15, in the form of https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_SIMULATE_ID.html
I agree it could be promoted more and there is https://gitlab.kitware.com/cmake/cmake/-/issues/19724 that has the same complaint. But I doubt the variable will just go away now any time soon, since its used quite a bit with ClangCl and the Intel compiler. But I can add a comment if you like?

alexreinking: Ahhhh `CMAKE_CXX_SIMULATE_ID` does not appear anywhere in the docs. I was mistaken about that.

CMakeLists.txt
alexreinking: Apparently this is not sufficient anyway since we're getting stack overflows with the autoscheduler on Windows now.

src/Util.h
abadams: I wrote this code originally. The thinking was that 1MB is probably the smallest useful stack size, and 8MB is what we frequently need, so may as well allocate things in 1MB chunks. Given that the OS needs to reserve the full 8MB of address space anyway, and lazy allocation of physical memory for pages is the norm now, I doubt the second arg even does anything.

alexreinking: Good to know. Dropping it is probably not an issue, then.

alexreinking:
>It turns out that using the Debug build of LLVM is not such a big issue as long as LLVM is also built with some level of optimization. This required a bit of tweaking in LLVMs cmake instructions, so it may not be the right choice for Halide. But for me it works quite ok so far. Let me know if you're interested in the patch, its a 10-liner only.

Would you mind posting this as a comment so we can see?

steven-johnson:
You should sync to master to retry the bots.

alexreinking:
I've been playing around with a few different ways of communicating our stack size requirement to our users. Here are my take-aways.

1. The removal of the pragma from Halide.h warrants an entry in the release notes since users who are not on CMake will need to increase the `StackReserveSize` in their Visual Studio projects (or whatever the equivalent is in their build system)
2. ClangCL's `lld-link.exe` does not support this pragma right now, but there's no reason it shouldn't. I have patched it locally (one line!) and will open a PR with upstream hopefully in time for LLVM 12, but otherwise in time for LLVM 13.
3. As this is a usage requirement for Halide on Windows, we need to communicate it to our linkees in CMake, which is our One True Build System on Windows. We need to do this in such a way that it can be _raised_ by the user in case they have extra stack space requirements. I will give a solution below, which will need to be documented in `README_cmake.md`

---

To do this effectively in CMake, we will need a custom property:

```
# Inside HalideGeneratorHelpers.cmake
define_property(
    TARGET PROPERTY STACK_RESERVE_SIZE
    BRIEF_DOCS "The minimum stack size required to run."
    FULL_DOCS [=[
Set the stack stack reserved by the operating system on process start. 
Currently Windows-only. Operates by passing the /STACK:N flag to link.exe
]=]
)
```
We will then need to tell CMake how this custom property works when we attach it to the Halide target.
```
# Inside src/CMakeLists.txt
add_library(Halide ...)
...
# ClangCL fails to support setting the /STACK flag via #pragma comment(linker, ...), which
# is how we accomplished this with MSVC previously. CMake doesn't have native support for 
# setting the stack size like this, so we use a custom property.
if (MSVC)
    set_property(TARGET Halide PROPERTY INTERFACE_STACK_RESERVE_SIZE 8388608)
    set_property(TARGET Halide APPEND PROPERTY COMPATIBLE_INTERFACE_NUMBER_MAX STACK_RESERVE_SIZE)
    target_link_options(Halide INTERFACE "/STACK:$<TARGET_PROPERTY:STACK_RESERVE_SIZE>")
endif ()
```

We will need to add an example to `README_cmake.md` to show how to override this:

```
add_executable(my_app main.cpp)
target_link_libraries(my_app PRIVATE Halide::Halide)
# override Halide's 8 MB stack to 16 MB.
if (MSVC)
    set_property(TARGET my_app PROPERTY STACK_RESERVE_SIZE 16777216)
endif ()
```
Previously, I'm not sure users _could_ override this, or maybe they could only with a `#pragma comment(linker, ...)`, which I'm not sure will work.

---

 I have two TODOs here:

1. Validate that all of this exports correctly
2. Check to see if users can override with `#pragma comment(linker, "/STACK:...")`

abadams:
Why delete the pragma (instead of wrapping it in an ifndef that detects clang)?

emmenlau:
> Why delete the pragma (instead of wrapping it in an ifndef that detects clang)?

Because clang does need a higher stack size too (actually pretty identical values to the MSVC compiler). But clang does not accept the same pragma. You would effectively be maintaining two parallel solutions, where the Clang-solution would work for MSVC too, but not vice versa.

emmenlau:
@alexreinking You are progressing really nicely here! I'm terribly busy before easter and will be away afterwards for a week, so please feel free to take over and edit this PR to your liking, or start over in a new PR and drop this one here. I'm not keen on my authorship. I place all my changes from this PR under the public domain, to be used by you or anyone in their own PRs, or to edit this PR in any way you see fit.

abadams:
The reason to keep the pragma (in addition to the cmake changes) would be to support people including Halide.h from vanilla msvc projects, or gyp-generated msvc projects.

emmenlau:
> The reason to keep the pragma (in addition to the cmake changes) would be to support people including Halide.h from vanilla msvc projects, or gyp-generated msvc projects.

That makes perfect sense, I agree. Its hard for me to suggest a best approach, since I can see that @alexreinking has gone to some lengths to support setting the stack size transitively via cmake. What I can say is that the clang compiler has many benefits over the standard MSVC compiler, so maybe its worthwhile to support it (at least in the longer run). But that's just my two cents...

alexreinking:
> The reason to keep the pragma (in addition to the cmake changes) would be to support people including Halide.h from vanilla msvc projects, or gyp-generated msvc projects.

We don't support vanilla VS or Gyp as a first party build system. We don't test those scenarios, so they're liable to break at any point. Those users are welcome to submit and maintain integration tests for those build systems if they expect reliable support from upstream.

Furthermore, I'm not convinced the pragma is the right solution anyway. It adds linker flags in the order they are encountered with later ones taking precedence. For example:

```
#include "my_company_header.h"

#include <Halide.h>
```

would override a (potentially larger) stack size set via pragma in `my_company_header.h`.

abadams:
I'd be willing to bet money that the majority of our windows users don't use cmake. We shouldn't casually and intentionally break non-cmake usage.

emmenlau:
> 2. ClangCL's `lld-link.exe` does not support this pragma right now, but there's no reason it shouldn't. I have patched it and opened a review with upstream. See here: https://reviews.llvm.org/D99680

Awesome!

> 3. As this is a usage requirement for Halide on Windows, we need to communicate it to our linkees in CMake, which is our One True Build System on Windows. We need to do this in such a way that it can be _raised_ by the user in case they have extra stack space requirements. I will give a solution below, which will need to be documented in `README_cmake.md`

This looks like a nice and clean solution and covers the case that users can easily override the setting with higher values (that the current implementation with the `pragma` makes less clear/easy).

> 4. We can check the stack size programatically on library load. We should do so and emit a warning if the stack size is too low.

This sounds very reasonable!

abadams:
Discussion on the gitter is converging to just spawning a thread or fiber with a larger stack to do lowering on, and not touching the build system at all (either via pragmas or cmake stuff).

steven-johnson:
> or gyp-generated msvc projects

...please tell me there aren't still people using gyp :-/

alexreinking:
Stack size requirement obviated by https://github.com/halide/Halide/pull/5873

alexreinking:
Closing in favor of #5876 (no edit access here, unfortunately).

PR #5859, opened 2021-03-29
CMake build: Add more user control

emmenlau:
This PR allows users to override the settings for `CMAKE_INSTALL_RPATH` and `CMAKE_CXX_STANDARD` from outside `CMakeLists.txt`. There should be no negative effects since the defaults are preserved. But users that want or need to override RPATH or the C++ standard get the freedom to do so.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,9 +38,13 @@ endif ()
 set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 
 # Require standard C++14
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+if(NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+elseif(CMAKE_CXX_STANDARD LESS 14)
+    message(FATAL_ERROR "Halide requires C++14 or newer")
+endif()
 
 # Build Halide with ccache if the package is present
 option(Halide_CCACHE_BUILD "Set to ON for a ccache enabled build" OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -38,9 +38,13 @@ endif ()
...
 # Require standard C++14
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+elseif(${CMAKE_CXX_STANDARD} LESS 14)
+    message(FATAL_ERROR "Halide requires C++14 or newer")
```
steven-johnson, line 46: ...this is quite amusing, as we officially require only C++11, not C++14, and yet apparently we were (and still are) requiring 14? Anyone know how we ended up here?

alexreinking: LLVM started requiring C++14. Must have gotten changed around then.

packaging/CMakeLists.txt
```diff
@@ -77,7 +77,9 @@ else ()
     set(rbase $ORIGIN)
 endif ()
 
-set_target_properties(${exes} PROPERTIES INSTALL_RPATH "${rbase};${rbase}/${lib_dir}")
+if ("${CMAKE_INSTALL_RPATH}" STREQUAL "")
+    set_target_properties(${exes} PROPERTIES INSTALL_RPATH "${rbase};${rbase}/${lib_dir}")
+endif ()
```
alexreinking, line 82: `if (NOT CMAKE_INSTALL_RPATH)`

CMakeLists.txt
alexreinking: Also: `if (NOT CMAKE_CXX_STANDARD)`.

alexreinking (approved):

alexreinking (changes requested):
See my if-condition edits.

CMakeLists.txt
alexreinking: Also: `elseif (CMAKE_CXX_STANDARD LESS 14)`

alexreinking (approved):

steven-johnson:
Failures are the unrelated cuda hang that we still haven't diagnosed, ok to land

PR #5863, opened 2021-03-30
Add wasm support for int32->f64 and f32->f64 simd ops

steven-johnson:
At top-of-tree LLVM, the wasm backend never seems to emit the vector version of these ops; pattern-match to target them specifically.


dsharletg (approved):
Should we just be filing bugs on the LLVM WebAssembly backend to address these kinds of things? It seems absurd that LLVM isn't doing this.

steven-johnson:
They are working on it, it's just taking longer than expected. At this point I think I have all of the low-hanging fruit, and the remaining issues have been reported to the relevant wasm/LLVM folk.

PR #5867, opened 2021-03-30
Add pipeline name to assertion messages

gasparitiago:
This commit affects the functions:
- halide_error_bad_type
- halide_error_bad_dimensions
- halide_error_access_out_of_bounds

### DIFF
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -149,6 +149,7 @@ Stmt add_image_checks_inner(Stmt s,
                             const Target &t,
                             const vector<string> &order,
                             const map<string, Function> &env,
+                            const string &pipeline_name,
                             const FuncValueBounds &fb,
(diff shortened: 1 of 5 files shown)


steven-johnson:
`generator_aot_msan` appears to need attention

dsharletg:
Thanks for giving this a try!

I was hoping to think of a way to generate the message for all possible kinds of failures, rather than just this subset of asserts.

I was thinking about things like calling `halide_print("In pipeline ", pipeline_name, ":")` before all assertion failure messages, but it's hard to guarantee that halide_print and halide_error messages go to the same place...

What do you think? The simple thing would be to just add the pipeline name to all assert messages, but that feels like a very blunt solution.

gasparitiago:
Closed due to another idea that maybe is better.

For that, I've created another pull request to keep it more organized: https://github.com/halide/Halide/pull/5869

Issue #5862, opened 2021-03-30
Assert failures from Halide pipelines should report the pipeline name


dsharletg:
When a large application with many pipelines crashes with `halide  : Error: Input buffer input is accessed at 15, which is beyond the max (13) in dimension 1` it would be very helpful if the pipeline that generated the assertion was mentioned somehow.

The current behavior is like a C assert failing without reporting the file or line number.

dsharletg:
I guess another problem here is that we don't get stack traces beyond halide_default_error.

gasparitiago:
Tried to perform the changes here: https://github.com/halide/Halide/pull/5867

gasparitiago:
Another idea: https://github.com/halide/Halide/pull/5869

PR #5869, opened 2021-03-30
Add pipeline name to assertion messages using error_name

gasparitiago:


gasparitiago:
@dsharletg I don't know if this solves the problem mentioned in https://github.com/halide/Halide/issues/5862 but, again, giving a try to, at least, improve the error messages.

dsharletg:
I think this `error_name` is only used in a subset of the assertion messages.

@abadams what do you think? Should we just bite the bullet and add a pipeline name parameter to every `halide_error_*` function?

abadams:
Yes, I think we need to bite that bullet, unfortunately. I don't see a good way around it. We'd have to plumb the pipeline name through in a lot of places in lowering too. Ugh.

test/generator/msan_aottest.cpp
```diff
@@ -194,7 +194,7 @@ extern "C" int halide_msan_annotate_memory_is_initialized(void *user_context, co
         }
     } else if (annotate_stage == AnnotateIntermediateBuffer) {
         if (expect_intermediate_buffer_error) {
-            if (len != 80) {
+            if (len != 99) {
                 fprintf(stderr, "Failure: Expected error message of len=80, saw %d bytes\n", (unsigned int)len);
```
rootjalex, line 198: Error message should be updated to say len=99

steven-johnson:
> Yes, I think we need to bite that bullet, unfortunately.

So should this PR proceed, or get withdrawn?

steven-johnson:
> > Yes, I think we need to bite that bullet, unfortunately.
> 
> So should this PR proceed, or get withdrawn?

Monday morning review ping

steven-johnson:
Is this PR still active?

steven-johnson:
Closing this PR as there has been no activity in months.

PR #5872, opened 2021-03-31
Fix an issue in Halide's float16 compilation support. Add tests.

zvookin:
In EmulateFloat16Math.cpp, conversion from 32-bit float to 16-bit float could produce a NaN value when and infinity is correct. This is because for numbers larger than the exact infinity value, the mantissa could be non zero.

Add tests to cover this case, and float16 infinities in general.

Couple small style/comment cleanups.

### DIFF
--- a/test/correctness/float16_t.cpp
+++ b/test/correctness/float16_t.cpp
@@ -241,7 +241,7 @@ int main(int argc, char **argv) {
         from_f16.compile_to_assembly("/dev/stdout", {}, Target("host-no_asserts-no_bounds_query-no_runtime-disable_llvm_loop_unroll-disable_llvm_loop_vectorize"));
     }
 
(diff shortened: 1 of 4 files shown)


test/correctness/float16_t.cpp
```diff
@@ -228,6 +241,84 @@ int main(int argc, char **argv) {
         from_f16.compile_to_assembly("/dev/stdout", {}, Target("host-no_asserts-no_bounds_query-no_runtime-disable_llvm_loop_unroll-disable_llvm_loop_vectorize"));
     }
 
+    // Check inifinity handling for both float16_t and Halide codegen.
```
steven-johnson, line 244: infinity

steven-johnson (approved):

PR #5873, opened 2021-04-01
Use fibers to guarantee stack size on Windows

alexreinking:
We used to use `#pragma comment(linker, "/STACK:8388608")` to enforce an 8MB stack on Windows. Unfortunately, clang-cl and lld-link don't support this pragma. I have opened a PR with upstream LLVM to correct this<sup>[1]</sup>, but it is clear we need a more robust solution than a transitive linker flag.

This PR adds a helper, `run_with_large_stack` to execute a `std::function<void()>` in a context where it will have at least 8 MB of stack space. Currently, it merely calls the function on Linux, but on Windows it creates a new fiber with the appropriate stack size and runs it there, blocking until the fiber returns.

It is applied in two places: around `lower()` and around the call to `compile_func` in CodeGen LLVM. This was sufficient to build locally on Windows, but there might be more places it's required. We'll let the buildbots sort that out.

<sup>[1]</sup> https://reviews.llvm.org/D99680

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -596,7 +596,7 @@ void generic_fiber_entry_point(void *argument) {
 
 #endif
 
-void call_with_stack_requirement(const std::function<void()> &action) {
+void run_with_large_stack(const std::function<void()> &action) {
 #if _WIN32
     SIZE_T required_stack = 8 * 1024 * 1024;
 
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -577,6 +577,63 @@ int get_llvm_version() {
     return LLVM_VERSION;
 }
 
+#ifdef _WIN32
+
+namespace {
+
+struct DeferredFunction {
+    const std::function<void()> *run;
+    LPVOID fiber;
+};
+
+void generic_fiber_entry_point(void *argument) {
+    auto *action = reinterpret_cast<DeferredFunction *>(argument);
+    (*action->run)();
+    SwitchToFiber(action->fiber);
+}
+
+}  // namespace
+
+#endif
+
+void call_with_stack_requirement(const std::function<void()> &action) {
+#if _WIN32
+    SIZE_T required_stack = 8 * 1024 * 1024;
+
+    ULONG_PTR stack_low, stack_high;
+    GetCurrentThreadStackLimits(&stack_low, &stack_high);
+    ptrdiff_t stack_remaining = (char *)&stack_high - (char *)stack_low;
+
+    if (stack_remaining < required_stack) {
+        debug(1) << "Insufficient stack space (" << stack_remaining << " bytes). Switching to fiber with " << required_stack << "-byte stack.\n";
+
+        auto was_a_fiber = IsThreadAFiber();
+
+        auto *main_fiber = was_a_fiber ? GetCurrentFiber() : ConvertThreadToFiber(nullptr);
+        internal_assert(main_fiber) << "ConvertThreadToFiber failed with code: " << GetLastError() << "\n";
+
+        DeferredFunction func{&action, main_fiber};
+        auto *lower_fiber = CreateFiber(required_stack, generic_fiber_entry_point, &func);
+        internal_assert(lower_fiber) << "CreateFiber failed with code: " << GetLastError() << "\n";
+
+        SwitchToFiber(lower_fiber);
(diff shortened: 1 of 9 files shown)


src/Util.cpp
```diff
@@ -572,6 +577,63 @@ int get_llvm_version() {
...
+
+#endif
+
+void run_with_large_stack(const std::function<void()> &action) {
+#if _WIN32
+    SIZE_T required_stack = 8 * 1024 * 1024;
+
+    ULONG_PTR stack_low, stack_high;
+    GetCurrentThreadStackLimits(&stack_low, &stack_high);
+    ptrdiff_t stack_remaining = (char *)&stack_high - (char *)stack_low;
```
abadams, line 605: Heh, this is slightly too clever for me. Would be better to make a local variable explicitly for the purpose of taking its address instead of reusing an existing one.

abadams (approved):

alexreinking:
Looks like some of the workers were not started in a virtualenv or otherwise numpy is not installed.

alexreinking:
I just got paranoid about what happens if the function throws an exception. Is it valid to catch the exception, move it to the parent fiber and then re-throw it?

abadams:
Looks like Raymond Chen wrote a series of blog posts on our use case. Part 4 seems to cover exceptions. https://devblogs.microsoft.com/oldnewthing/20200605-00/?p=103840

alexreinking:
> 
> 
> Looks like Raymond Chen wrote a series of blog posts on our use case. Part 4 seems to cover exceptions. https://devblogs.microsoft.com/oldnewthing/20200605-00/?p=103840

Sorry I didn't see this before implementing. I'll take a look, but does this look wrong to you?

alexreinking:
Oh, good, looks like I did the right thing. Raymond's implementation supports return values, which is nice, but maybe overkill for what we're doing.

src/Util.cpp
alexreinking: Fixed :)

src/Util.cpp
```diff
@@ -572,6 +577,84 @@ int get_llvm_version() {
     return LLVM_VERSION;
 }
 
+#ifdef _WIN32
+
+namespace {
+
+struct DeferredFunction {
+    const std::function<void()> *run;
```
steven-johnson, line 585: I assume that there is no ownership implied and pointer is assumed to remain valid? (If so, make a comment to that effect might be good)

steven-johnson (approved):
LGTM. Would sure be nice if there was a way to do something similar on Linux/OSX/etc so that we could avoid requiring the linker flags there,too.

alexreinking:
> LGTM. Would sure be nice if there was a way to do something similar on Linux/OSX/etc so that we could avoid requiring the linker flags there,too.

We don't require any stack-size linker flags on Linux, though, do we?

src/Util.cpp
alexreinking: It's really just a parameter pack. Maybe changing the name to GenericFiberArgumentsPack would communicate that better.

alexreinking: But that's right, there's no ownership implied here. All that stuff is pointers to automatic storage

abadams:
> > LGTM. Would sure be nice if there was a way to do something similar on Linux/OSX/etc so that we could avoid requiring the linker flags there,too.
> 
> We don't require any stack-size linker flags on Linux, though, do we?

That's correct. 8MB is the default already.

We do have a custom stack size for the fft app on os x only, but that's to reduce it substantially to act as a canary if we add a new pass that uses too much stack.

src/Util.cpp
alexreinking: I fixed up the names and changed it to a const-ref, so it's more clear there's no ownership.

alexreinking:
@steven-johnson looks like we've got some zombies?

steven-johnson:
> @steven-johnson looks like we've got some zombies?

It's zombies all the way down.

I've commented on an existing bug on buildbot re: the zombies that won't die.
I've open a ticket at Google re: the network that is dropping intermittently.

Unfortunately, the only way to kill the zombies is to stop and restart the master, which I'll do now.

alexreinking:
Thanks! Looks like all the checks are passing now... good to merge?

steven-johnson:
LGTM

PR #5874, opened 2021-04-01
Try to vectorize inner statement of else branch of likely

vksnk:
For the IfThenElse node with 'likely' in the condition, vectorize_loops will create an else branch and scalarize it by wrapping it into serialized version of vectorized loops encountered so far, however, the inner statement of 'likely' still may contain other vectorized loops, which should be safe to vectorize.

### DIFF
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -903,9 +903,6 @@ class VectorSubs : public IRMutator {
                     // but it still may happen that there are vectorized loops inside of the statement
                     // itself which we may want to handle. All the context is invalid though, so
                     // we just start anew for this specific statement.
-                    // NOTE(vksnk): is it possible that we want to scalarize only some of the vectorized loops
-                    // outside of the current IfThenElse(likely()) node instead of all of them (for example,
-                    // only for variables which appear in likely?).
                     Stmt scalarized = scalarize(without_likelies, false);
                     scalarized = vectorize_statement(scalarized, target);
                     Stmt stmt =
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -488,6 +488,8 @@ public:
     }
 };
 
+Stmt vectorize_statement(const Stmt &stmt, const Target &t);
+
 struct VectorizedVar {
     string name;
     Expr min;
@@ -896,10 +898,19 @@ class VectorSubs : public IRMutator {
                     Stmt without_likelies =
                         IfThenElse::make(unwrap_tags(op->condition),
                                          op->then_case, op->else_case);
+
+                    // scalarize() will put back all vectorized loops around the statement as serial,
+                    // but it still may happen that there are vectorized loops inside of the statement
+                    // itself which we may want to handle. All the context is invalid though, so
+                    // we just start anew for this specific statement.
+                    // NOTE(vksnk): is it possible that we want to scalarize only some of the vectorized loops
+                    // instead of all of them (for example, only for variables which appear in likely?).
+                    Stmt scalarized = scalarize(without_likelies, false);
+                    scalarized = vectorize_statement(scalarized, target);
                     Stmt stmt =
                         IfThenElse::make(all_true,
                                          then_case,
-                                         scalarize(without_likelies));
+                                         scalarized);
                     debug(4) << "...With all_true likely: \n"
                              << stmt << "\n";
                     return stmt;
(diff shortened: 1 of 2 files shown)


src/VectorizeLoops.cpp
```diff
@@ -896,10 +898,19 @@ class VectorSubs : public IRMutator {
                     Stmt without_likelies =
                         IfThenElse::make(unwrap_tags(op->condition),
                                          op->then_case, op->else_case);
+
+                    // scalarize() will put back all vectorized loops around the statement as serial,
+                    // but it still may happen that there are vectorized loops inside of the statement
+                    // itself which we may want to handle. All the context is invalid though, so
+                    // we just start anew for this specific statement.
+                    // NOTE(vksnk): is it possible that we want to scalarize only some of the vectorized loops
```
dsharletg, line 906: Isn't this note exactly the issue this fixes? Is this comment stale?

test/correctness/vectorize_nested.cpp
```diff
@@ -192,6 +192,54 @@ int vectorize_all_d() {
...
+        .split(y, yo, y, 8, TailStrategy::GuardWithIf)
+        .vectorize(x)
+        .vectorize(y);
+
+    // We are looking for a specific loop, which shouldn't have been scalarized.
+    class CheckForScalarizedLoop : public Internal::IRMutator {
+        using IRMutator::visit;
+
+        Internal::Stmt visit(const Internal::For *op) override {
+            if (Internal::ends_with(op->name, ".x_inner")) {
```
dsharletg, line 214: Maybe we should also look for the y loop, so we know (a little better at least) the test isn't just not finding a loop even though it's there.

vksnk: I've added a check for y loop as well. Also, I've verified that this test fails prior to this change.

src/VectorizeLoops.cpp
vksnk: I've extended the comment to give a bit more details, but basically this change will try to vectorize loops which are inside of given IfThenElse(likely()) node. All of the vectorized loops outside of it will be scalarized, but I am not sure if it's really necessary, so maybe we can only scalarize some of these loops.

dsharletg: Yeah, I understand, I just don't understand the difference between the NOTE and the comment above? It seems like it's talking about the same thing. I am wondering if we should just delete the NOTE and keep the comment above?

abadams:
A current problem I've noticed is that if an atomic().vectorize() fails to hit one of the recognized patterns, then all loops get scalarized, not just the reduction loop. Does this fix that too?

src/VectorizeLoops.cpp
vksnk: Ah, I see. Yeah, I've removed it now.

vksnk:
> A current problem I've noticed is that if an atomic().vectorize() fails to hit one of the recognized patterns, then all loops get scalarized, not just the reduction loop. Does this fix that too?

No, it doesn't fix it quite yet, but I've been hitting this issue sometimes too and wanted to look at it next. I guess the approach would be to scalarize only vectorized reduction loops and keep other loops as is and try to vectorize them again. Do we need to keep atomic node in that case? (I guess there might be other uses for it, so probably will need to keep it, but I am not sure).

abadams:
I'm not sure if the atomic node has to stay.... if the loop has been serialized then it seems unnecessary, but maybe there's a containing parallel loop? I guess just see what the current logic does.

vksnk:
> I'm not sure if the atomic node has to stay.... if the loop has been serialized then it seems unnecessary, but maybe there's a containing parallel loop? I guess just see what the current logic does.

Thanks, I'll see if it's necessary.

dsharletg:
I guess we need to have an optimization that removes atomic when it isn't necessary? i.e. there is no parallel loop inside the allocation being stored to?

BTW, I think it's OK if that fix goes in a separate PR, this PR is a good change on its own.

vksnk:
> I guess we need to have an optimization that removes atomic when it isn't necessary? i.e. there is no parallel loop inside the allocation being stored to?
> 
> BTW, I think it's OK if that fix goes in a separate PR, this PR is a good change on its own.

Yes, I also think it'd be better as a separate PR.

abadams:
I think that optimization might already exist. See RemoveUnnecessaryAtomics in VectorizeLoops.cpp

dsharletg (approved):

PR #5876, opened 2021-04-02
Improve ClangCL support by disabling, fixing warnings

alexreinking:
Supersedes @emmenlau's #5858, which I guess didn't provide maintainers with write permissions to the PR branch.

Improves ClangCL support by disabling (or fixing) warnings its `-Wall` mode emits.

* Removed a bunch of extra semicolons
* Stopped passing linker script to ClangCL (which throws an unknown flag warning)
* Be smarter about disabling introspection
* Various code fixes (see my comments in review)
* A bit of CMake clean-up near my edits.


src/CMakeLists.txt
```diff
@@ -360,6 +360,9 @@ add_custom_target(HalideIncludes ALL
 # Define the Halide library target.
 ##
 
+set(CXX_COMPILER_IS_GNU_OR_CLANG $<CXX_COMPILER_ID:GNU,Clang,AppleClang>)
+set(CXX_COMPILER_IS_GNU_OR_CLANG_AND_NOT_CLANGCL $<AND:${CXX_COMPILER_IS_GNU_OR_CLANG},$<NOT:$<STREQUAL:${CMAKE_CXX_SIMULATE_ID},MSVC>>>)
```
steven-johnson, line 364: insert sad comment here about why this long identifier actually is necessary

steven-johnson (approved):

steven-johnson:
> Skipping buildbots because the changes have already been tested in the other PR and we don't test ClangCL on CI, anyway.

This will certainly break if we don't test it, but I guess that's ok for now ("best effort"). If we get better Windows testing capability it would be worth adding.

alexreinking:
Still need to fix version script detection (results in `lld-link : warning : ignoring unknown argument '--version-script=D:/Halide/repo/src/exported_symbols.ldscript' [D:\Halide\repo\build\src\Halide.vcxproj]`)

src/CMakeLists.txt
alexreinking: It could perhaps be `CXX_IS_GNULIKE` and `CXX_IS_GNULIKE_NOT_CLANGCL`

steven-johnson: > `CXX_IS_GNULIKE`

did you mean: `CXX_IS_GNU_SLASH_LINUX_LIKE`

:-)

alexreinking:
I'm wondering if a different approach here is appropriate... the only ClangCL-specific changes are to:

1. Disabling `-Wall`, which is smelly to me
2. Disabling `WITH_INTROSPECTION` which could be solved outside of the buildsystem by sniffing `_MSC_VER`.

I wonder if a better approach is to fix/selectively disable warnings and disable introspection on MSVC-like compilers using the usual macros.

@steven-johnson -- any thoughts?

steven-johnson:
Disabling `-Wall` should never be necessary and is definitely a smell. (I overlooked this when I approved it.) Does ClangCL nto support it at all?

alexreinking:
> Disabling `-Wall` should never be necessary and is definitely a smell. (I overlooked this when I approved it.) Does ClangCL nto support it at all?

It does support `-Wall`, but it enables very, very many warnings. This is the new list of warnings to disable (some of these might be real, I don't know):

```
-Wno-c++98-compat-pedantic
-Wno-c++98-compat
-Wno-cast-align
-Wno-comma
-Wno-deprecated-declarations
-Wno-documentation-unknown-command
-Wno-documentation
-Wno-double-promotion
-Wno-exit-time-destructors
-Wno-extra-semi-stmt
-Wno-extra-semi
-Wno-float-conversion
-Wno-float-equal
-Wno-global-constructors
-Wno-implicit-float-conversion
-Wno-implicit-int-conversion
-Wno-implicit-int-float-conversion
-Wno-missing-field-initializers
-Wno-missing-prototypes
-Wno-nonportable-system-include-path
-Wno-old-style-cast
-Wno-reserved-id-macro
-Wno-return-std-move-in-c++11
-Wno-shadow-field-in-constructor
-Wno-shadow-field
-Wno-shadow
-Wno-shorten-64-to-32
-Wno-sign-conversion
-Wno-switch-enum
-Wno-undef
-Wno-undefined-func-template
-Wno-unused-function
-Wno-unused-member-function
-Wno-unused-parameter
-Wno-unused-template
```

I wonder if clang-cl's `-Wall` is more of a `-Weverything` or even `-Weverything -pedantic -Wextra -Weven_more_I_didnt_think_of`.

Another wrinkle: ClangCL doesn't have exceptions enabled by default _and_ it doesn't honor `-fexceptions` as a way of enabling them.

src/CodeGen_Hexagon.cpp
```diff
@@ -283,8 +283,6 @@ class SloppyUnpredicateLoadsAndStores : public IRMutator {
 
             return Call::make(op->type, Call::if_then_else,
                               {condition, load, make_zero(op->type)}, Call::Intrinsic);
-
-            return load;
```
alexreinking, line 287: Ouch. I wonder if this was a bad merge. Probably worth turning on this warning explicitly.

src/LowerWarpShuffles.cpp
```diff
@@ -571,7 +571,7 @@ class LowerWarpShuffles : public IRMutator {
 
         Expr wild = Variable::make(Int(32), "*");
         vector<Expr> result;
-        int bits;
+        int bits = 0;
```
alexreinking, line 574: The reason this isn't needed is _extremely_ subtle: the function `is_const_power_of_two_integer` returns true iff it initializes `bits`. The first read of bits is guarded by short-circuit `&&` below. ClangCL isn't clever enough to see this, and I don't blame it.

There's no downside to this (this is not performance-critical) and disabling uninitialized use warnings seems significantly worse.

src/Util.cpp
```diff
@@ -102,9 +102,8 @@ std::string get_env_variable(char const *env_var_name) {
     if (lvl) {
         return std::string(lvl);
     }
-#endif
-
     return "";
+#endif
 }
```
alexreinking, line 107: This emits an unreachable code warning.

src/Util.cpp
```diff
@@ -529,7 +528,7 @@ struct TickStackEntry {
     int line;
 };
 
-vector<TickStackEntry> tick_stack;
```
alexreinking, line 532: Throws a warning about missing extern and that it should be static if module-local, which it is. No need to export this.

src/WasmExecutor.cpp
```diff
@@ -1586,11 +1586,11 @@ WasmModule WasmModule::compile(
...
     user_error << "Cannot run JITted WebAssembly without configuring a WebAssembly engine.";
     return WasmModule();
-#endif
-
+#else
     WasmModule wasm_module;
     wasm_module.contents = new WasmModuleContents(module, arguments, fn_name, jit_externs, extern_deps);
     return wasm_module;
+#endif
 }
```
alexreinking, line 1594: Similar to above: unreachable code.

src/VectorizeLoops.cpp
```diff
@@ -1672,4 +1672,4 @@ Stmt vectorize_loops(const Stmt &stmt, const map<string, Function> &env, const T
 }
 
 }  // namespace Internal
-}  // namespace Halide
```
alexreinking, line 1675: This was the only file without a trailing newline.

steven-johnson:
> I wonder if clang-cl's `-Wall` is more of a `-Weverything` or even `-Weverything -pedantic -Wextra -Weven_more_I_didnt_think_of`.

Ouch -- yeah, maybe it's intended to be equivalent of MSVC's warning-level-4, which is pretty insanely high IIRC.
 
> Another wrinkle: ClangCL doesn't have exceptions enabled by default _and_ it doesn't honor `-fexceptions` as a way of enabling them.

uh, what? how on earth would a clang-compatible(-ish) compiler omit this?

src/Buffer.h
```diff
@@ -568,12 +568,12 @@ class Buffer {
     Expr operator()(const Expr &first, Args... rest) const {
         std::vector<Expr> args = {first, rest...};
         return (*this)(args);
-    };
+    }
```
steven-johnson, line 571: Ooh, I'd actually love to enable whatever warnings complains about this, it's a minor pet peeve of mine

src/CMakeLists.txt
```diff
@@ -386,57 +383,94 @@ set_target_properties(Halide PROPERTIES
...
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-float-equal>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-global-constructors>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-implicit-float-conversion>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-implicit-int-conversion>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-implicit-int-float-conversion>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-missing-field-initializers>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-missing-prototypes>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-nonportable-system-include-path>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-old-style-cast>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-reserved-id-macro>  # can't have an underscore followed by a capital letter
```
steven-johnson, line 430: Where are we violating this?

src/CodeGen_Hexagon.cpp
steven-johnson: Yow. I'm genuinely surprised this is merely a warning.

steven-johnson (approved):

steven-johnson:
Failures are unrelated (error in nightly LLVM13), ok to land

alexreinking:
Landing. Still need a way to reliably enable exceptions on ClangCL, but that can be done in another PR. This one's scope has crept far enough.

PR #5878, opened 2021-04-05
Simplify asserts always true with Target::LargeBuffers

dsharletg:
This requires tracking bounds through casts, and avoiding abs where the result is the max of an int64 + 1, which the simplifier can't handle.

### DIFF
--- a/src/Simplify_Call.cpp
+++ b/src/Simplify_Call.cpp
@@ -362,6 +362,18 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             return mutate(unbroadcast, bounds);
         }
 
+        if (bounds) {
+            bounds->min = 0;
+            bounds->min_defined = true;
+            if (a_bounds.min_defined && a_bounds.max_defined) {
+                if (a_bounds.min != std::numeric_limits<int64_t>::min() &&
+                    a_bounds.max != std::numeric_limits<int64_t>::min()) {
+                    bounds->max_defined = true;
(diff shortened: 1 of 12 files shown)


src/Simplify_Call.cpp
```diff
@@ -362,6 +362,18 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
             return mutate(unbroadcast, bounds);
         }
 
+        if (bounds) {
+            bounds->min = 0;
+            bounds->min_defined = true;
+            if (a_bounds.min_defined && a_bounds.max_defined) {
+                if (a_bounds.min != std::numeric_limits<int64_t>::min() &&
+                    a_bounds.max != std::numeric_limits<int64_t>::min()) {
```
steven-johnson, line 370: When would a.bounds.max ever equal numeric_limits::min (vs max)?

dsharletg: A constant equal to `t = std::numeric_limits<int64_t>::min()` would have `min = max = t`

steven-johnson: ahh, gotcha, didn't think about the constant case

dsharletg:
It turns out that making the simplifier better at this is having far reaching impacts, it's causing the simplifier to reassociate mins/maxes and casts, which is breaking a lot of pattern matching in the backends.

I think maybe I need to abandon this change because fixing this is a major project I think.

PR #5879, opened 2021-04-05
Add explicit cast to remove ambiguous operator== (Fixes #5329)

### LINKED ISSUE #5329: Ambiguous overloaded operator== in C++20
When compiling a simple halide example, clang-10 presents an error:
```
In file included from /home/developer/CLionProjects/halide-test/main.cpp:20:
/usr/include/Halide.h:27922:27: error: use of overloaded operator '==' is ambiguous (with operand types 'const Halide::Type' and 'halide_type_t')
            return e.type == type && value == val.u.i64;
                   ~~~~~~ ^  ~~~~
/usr/include/Halide.h:562:31: note: candidate function (with reversed parameter order)
    HALIDE_ALWAYS_INLINE bool operator==(const halide_type_t &other) const {
                              ^
/usr/include/Halide.h:3590:10: note: candidate function
    bool operator==(const Type &other) const {
```
This seems to be caused by changing `set(CMAKE_CXX_STANDARD 17)` to `set(CMAKE_CXX_STANDARD 20)`
This can be easily replicated using the first tutorial, and a cmake file as simple as this:
```
cmake_minimum_required(VERSION 3.17)
project(halide_test)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED YES)
set(CMAKE_CXX_EXTENSIONS NO)

find_package(Halide REQUIRED)
add_executable(halide_test lesson_01_basics.cpp)
target_link_libraries(halide_test PRIVATE Halide::Halide)
```

abadams:


steven-johnson (approved):
This cast looks really weird and unnecessary without comment, not sure if worth adding something terse

abadams:
How so? `e.type` is a Halide::Type, and `type` is a halide_type_t

steven-johnson:
> How so? `e.type` is a Halide::Type, and `type` is a halide_type_t

Ah, ok, I thought they were both halide_type_t

steven-johnson:
Failures are unrelated (error in nightly LLVM13), ok to land

PR #5881, opened 2021-04-05
Enable i64x2 comparisons in simd_op_check

steven-johnson:
Top-of-tree LLVM is now emitting comparisons for i64x2 simd operations in wasm, so enable the checks in simd_op_check.

Also, drive-by fixes to ensure some other checks only happen for LLVM13+.

### DIFF
--- a/test/correctness/simd_op_check.cpp
+++ b/test/correctness/simd_op_check.cpp
@@ -1979,8 +1979,9 @@ public:
                 check("i8x16.eq", 16 * w, i8_1 == i8_2);
                 check("i16x8.eq", 8 * w, i16_1 == i16_2);
                 check("i32x4.eq", 4 * w, i32_1 == i32_2);
-                // TODO(https://github.com/halide/Halide/issues/5130): NOT BEING GENERATED AT TRUNK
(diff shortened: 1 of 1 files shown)


test/correctness/simd_op_check.cpp
```diff
@@ -2154,13 +2160,17 @@ class SimdOpCheck : public SimdOpCheckTest {
                 // check("f32x4.demote_f64x2_zero", 4 * w, ???);
 
                 // Single-precision floating point to double-precision
-                check("f64x2.promote_low_f32x4", 2 * w, cast<double>(f32_1));
+                if (Halide::Internal::get_llvm_version() >= 130) {
+                    check("f64x2.promote_low_f32x4", 2 * w, cast<double>(f32_1));
+                }
 
                 // Integer to integer narrowing
-                check("i8x16.narrow_i16x8_s", 16 * w, i8_sat(i16_1));
```
dsharletg, line 2160: If this wasn't failing before, do we need the check?

steven-johnson: It fails if you build with LLVM12.

dsharletg (approved):

PR #5884, opened 2021-04-06
error_run_with_large_stack_throws should compile without exceptions

steven-johnson:
(1) Some downstream environments compile C++ without exceptions by default; this won't compile on those.
(2) We should check that assert-fail also errors out as expected.


alexreinking (approved):

alexreinking:
My bad - we don't test with exceptions off, so I forgot to check this.

PR #5890, opened 2021-04-08
Use guarded versions of vars if they exist in bounds inference

abadams:
Producers scheduled innermost didn't always benefit from .guarded versions of vars (with a promise_clamped on them) because bounds inference just goes and gets the var by name in some circumstances. In those circumstances I made it check for a guarded version of the var in scope.


abadams:
This is to fix #4889

dsharletg (approved):

abadams:
Failures unrelated (llvm 13 timing out on cuda stuff)

PR #5892, opened 2021-04-09
Add debugging code to gpu_dynamic_shared to try to track down hard-to-repro bug on buildbots

steven-johnson:


steven-johnson:
aha: upgrading the nvidia driver on my local linux box to a recent one (460.x) allows me to repro locally - the older driver I was using(440.x-ish) didn't fail. So it's driver-specific to some extent.

steven-johnson:
If I run with cuda-memcheck, I get:

========= Invalid __shared__ write of size 4
=========     at 0x00000210 in _kernel_g_s0_x_x___block_id_x
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x00001900 is out of bounds

steven-johnson:
Bisect is suggesting that #5757 is the injection spot; investigating

steven-johnson:
(1) The mystery hang seems to have resolved itself, or at least is no longer reproducing at recent top-of-tree.
(2) The bug with disable-llvm-opt enabled seems to be an unrelated red herring (this test fails with this flag enabled for as far back as I can find). See https://github.com/halide/Halide/pull/5900 for a workaround there.

PR #5899, opened 2021-04-13
Revert "Display the GPU device code as a string in the C/C++ backend"

steven-johnson:
Testing to see if this is the injection point for recent Cuda-related hangs/errors.


steven-johnson:
(I think that bisect is lying to us here, but I'll let this finish...)

PR #5900, opened 2021-04-13
DisableUnrollLoops should ignore DisableLLVMLoopOpt for PTX

steven-johnson:
It's not entirely clear why, but setting the `DisableUnrollLoops` flag to `true` in the PTX backend can produce Cuda code that will fail at runtime, at least for certain drivers (e.g.: some of the tests in correctness_gpu_dynamic_shared when running under NVidia v460.x).

### DIFF
--- a/src/CodeGen_PTX_Dev.cpp
+++ b/src/CodeGen_PTX_Dev.cpp
@@ -687,7 +687,11 @@ vector<char> CodeGen_PTX_Dev::compile_to_src() {
     b.Inliner = createFunctionInliningPass(b.OptLevel, 0, false);
     b.LoopVectorize = do_loop_opt;
     b.SLPVectorize = true;
-    b.DisableUnrollLoops = !do_loop_opt;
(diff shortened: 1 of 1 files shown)


abadams:
This doesn't seem like the right fix. The code with loop opts disabled is much simpler and clearer, yet it crashes.

You said it was driver-dependent, right? Maybe this is a bug inside the cuda driver?

abadams:
To be clearer, I think the reason this isn't the right fix is that I would like disable_llvm_loop_opt to work for the bodies of cuda kernels, because I think a lot of the time that generates better kernels.

steven-johnson:
> You said it was driver-dependent, right? Maybe this is a bug inside the cuda driver?

Yeah, probably correct -- but this is in the standard nvidia (closed-source) linux driver v460 (which Ubuntu 20 installs by default), and it's been out for a while with no updates. Maybe we should file a bug?

steven-johnson:
Definitely not the right fix, closing

PR #5906, opened 2021-04-14
Add an error for realize with a different number of outputs than defined for pipeline.

zvookin:


src/Pipeline.cpp
```diff
@@ -933,6 +933,8 @@ void Pipeline::prepare_jit_call_arguments(RealizationArg &outputs, const Target
                                           bool is_bounds_inference, JITCallArgs &args_result) {
     user_assert(defined()) << "Can't realize an undefined Pipeline\n";
 
+    user_assert(outputs.size() == this->outputs().size()) << "Realization requests " << outputs.size() << " output(s) for pipeline producing " << this->outputs().size() << " result(s).\n";
```
steven-johnson, line 936: Can I 'nit' here and say that at first read this didn't sound like an error message to me? Maybe wording like 

    The realization requests %x output(s) but the pipeline produces %y result(s).

steven-johnson (approved):

src/Pipeline.cpp
zvookin: Full output is:
```Error: (at /usr/local/google/home/zalman/src/halide/test/error/realization_with_too_many_outputs.cpp:17)
Realization requests 2 output(s) for pipeline producing 1 result(s).```

zvookin: Changed.

steven-johnson:
Failure is unrelated.

PR #5913, opened 2021-04-16
Suppress Metal unused function warning

### LINKED ISSUE #5912: Metal compiler complains about unused functions
The solution is to add the following to the generated shader:

```#pragma clang diagnostic ignored "-Wunused-function"```

shoaibkamil:
Should fix #5912

### DIFF
--- a/src/CodeGen_Metal_Dev.cpp
+++ b/src/CodeGen_Metal_Dev.cpp
@@ -737,7 +737,8 @@ void CodeGen_Metal_Dev::init_module() {
     src_stream.clear();
 
     // Write out the Halide math functions.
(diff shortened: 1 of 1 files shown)


steven-johnson:
LGTM

dsharletg (approved):

zvookin (approved):

zvookin:
BTW, this is one reason I am very dubious about compiler warnings in general and especially from Halide. One idea might be to add a Target flag for "machine generate input" where any warning/error designed to help a human realize they might have made a mistake, rather than to indicate something that cannot correctly and unambiguously be compiled, is turned off.

steven-johnson:
Finally clean, LGTM

PR #5914, opened 2021-04-16
Comment out specializations for f64x2.convert_low_i32x4_s/u

steven-johnson:
LLVM removed the primitives we need (so our code can't be used), but it also doesn't seem to be generating the expected instructions directly (as claimed). Commenting out to un-break tests; issue has been reported to wasm/llvm team.

### DIFF
--- a/src/runtime/wasm_math.ll
+++ b/src/runtime/wasm_math.ll
@@ -141,24 +141,27 @@ define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwi
 
 ; Integer to double-precision floating point
 
-declare <2 x double> @llvm.wasm.convert.low.signed(<4 x i32>)
-declare <2 x double> @llvm.wasm.convert.low.unsigned(<4 x i32>)
-
-define weak_odr <4 x double> @i32_to_double_s(<4 x i32> %x) nounwind alwaysinline {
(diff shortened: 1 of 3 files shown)


src/runtime/wasm_math.ll
```diff
@@ -141,24 +141,27 @@ define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwi
...
-}
-
-define weak_odr <4 x double> @i32_to_double_u(<4 x i32> %x) nounwind alwaysinline {
-  %1 = shufflevector <4 x i32> %x, <4 x i32> undef, <4 x i32> <i32 2, i32 3, i32 undef, i32 undef>
-  %2 = tail call <2 x double> @llvm.wasm.convert.low.unsigned(<4 x i32> %x)
-  %3 = tail call <2 x double> @llvm.wasm.convert.low.unsigned(<4 x i32> %1)
-  %4 = shufflevector <2 x double> %2, <2 x double> %3, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
-  ret <4 x double> %4
-}
+; COMMENTED OUT AT TOP OF TREE; llvm.wasm.convert.low.[un]signed is no longer
```
dsharletg, line 144: You probably don't actually need to comment this out. It won't matter if we don't use it.

dsharletg (approved):

PR #5916, opened 2021-04-19
Add mul_shift_right intrinsic and related improvements

dsharletg:
This PR moves pattern matching of saturating multiplication with shifts to an intrinsic, and includes related improvements:
- Before this PR, only the Hexagon backend had the ability to detect cases where the simplifier had cancelled a power of 2 from one of the operands and the denominator of a `mul_shift_right` operation. Now, all backends have this capability.
- The x86 version of this instruction does not saturate, while the version on other targets does. This meant that code written to target this pattern on x86 did not generate the instruction on ARM/Hexagon or vice-versa. For code that heavily uses this pattern, it makes a huge difference to performance to fix this, I see a ~2x performance improvement on some HANNK ops (#5891) on x86.
- This PR cleans up the types of shift amounts, so pattern matching is easier and more reliable (removed the `bc` helper function used in several backends).
- We already had a `mulhi_shr` intrinsic that was almost the same thing as `mul_shift_right`, but I think `mul_shift_right` is implemented more cleanly and more readily usable.


abadams:
Do you think fixed_point_multiply or multiply_fixed_point might be better? quantized doesn't seem quite like the right word.

abadams:
Also so what should users do now given the saturating vs non saturating behavior? Always use saturation?

dsharletg:
> Do you think fixed_point_multiply or multiply_fixed_point might be better? quantized doesn't seem quite like the right word.

The other name I was considering was simply widening_mul_shift_right (and widening_mul_rounding_shift_right), what do you think?

> Also so what should users do now given the saturating vs non saturating behavior? Always use saturation?

I think always using saturation is probably the best thing. With the saturating_pmulhrs helper in this PR, that means all of (x86, ARM, Hexagon) can do 16-bit saturating multiplies pretty easily, but only x86 can do the non-saturating version.

The cost of the saturation fixup on x86 seems negligible (at least compared to the old implementation of widening multiply + rounding + saturating narrow). Also, if you are doing these multiplications by a constant (a common case e.g. polynomial evaluation), LLVM simplifies the fixup away completely.

abadams:
widening_mul_shift_right sounds like it doesn't then narrow afterwards.

I'm wondering if we're just going to end up with a set of fixed-point math intrinsics, and maybe we should just anticipate that now.

abadams:
fixed point add is the averaging op, assuming a matched precision. Do we already have an intrinsic for that?

dsharletg:
Yes, that is rounding_halving_add, or halving_add. But in my experience, when I'm doing fixed point arithmetic, I let adds gain precision for a while, and then do one shift at the end, or roll the added precision into a multiply.

src/FindIntrinsics.cpp
```diff
@@ -639,33 +642,103 @@ Expr lower_saturating_sub(const Expr &a, const Expr &b) {
...
+        // if it isn't already full precision. This avoids infinite loops despite
+        // "lowering" this to another mul_shift_right operation.
+        Expr missing_q = full_q - q;
+        internal_assert(missing_q.type().bits() == b.type().bits());
+        Expr new_b = simplify(b << missing_q);
+        if (is_const(new_b) && can_prove(new_b >> missing_q == b)) {
+            return mul_shift_right(a, new_b, full_q);
+        }
+        Expr new_a = simplify(a << missing_q);
+        if (is_const(new_a) && can_prove(new_a >> missing_q == a)) {
```
abadams, line 688: What's the can_prove guarding against? Overflow?

src/FindIntrinsics.h
```diff
@@ -28,6 +28,9 @@ Expr lower_halving_sub(const Expr &a, const Expr &b);
 Expr lower_rounding_halving_add(const Expr &a, const Expr &b);
 Expr lower_rounding_halving_sub(const Expr &a, const Expr &b);
 
+Expr lower_multiply_quantized(const Expr &a, const Expr &b, const Expr &q);
```
abadams, line 31: Should this be lower_mul_shift_right now?

src/IROperator.cpp
```diff
@@ -1092,69 +1095,165 @@ Expr widening_sub(Expr a, Expr b) {
...
+    check_representable(qt, q);
+    return mul_shift_right(std::move(a), std::move(b), make_const(qt, q));
+}
+
+Expr rounding_mul_shift_right(Expr a, Expr b, Expr q) {
+    user_assert(a.defined() && b.defined() && q.defined()) << "rounding_mul_shift_right of undefined Expr\n";
+    user_assert(q.type().is_uint()) << "rounding_mul_shift_right shift must be unsigned\n";
+    match_types(a, b);
+    match_lanes(a, q);
+    match_bits(a, q);
```
abadams, line 1243: I see a lot of logic throughout this PR to handle q being a weird type. I'm not sure when or why this would happen though. If it's an unknown Int(32) scalar param? Is this sort of thing tested? I think I only see constant q below.

src/FindIntrinsics.cpp
dsharletg: Yes, this is just computing (a << missing_q) >> missing_q. If that isn't equal to a, then this rewrite isn't safe.

src/IROperator.cpp
abadams: Ah, I'd forgotten how involved the shifts had gotten. If this just matches the shift logic, then fine. But is the non-constant case tested somewhere?

dsharletg: There's testing of (rounding_)lower_mul_shift_right in test/correctness/intrinsics.cpp, which is the same codepath used by codegen for the variable case. There's also tons of usage of the variable shift case in #5891, which I forked this from, and hopefully will be merged soon.

abadams (approved):

PR #5918, opened 2021-04-20
Speed up simd_op_check by only compiling one pipeline per op

dsharletg:
This PR changes simd_op_check to compile just one pipeline, the error checking pipeline. I think the motivation for the original design was to make the assembly to read when the op failed to generate as small as possible. However, the assembly with this isn't that much bigger, and since we compile only one pipeline, there is only one version of a lowered Stmt, so HL_DEBUG_CODEGEN output is easier to read, despite the slightly larger pipeline.

This should also speed up simd_op_check, which is up to ~15 minutes(!) on the ARM buildbots. Unfortunately the speedup is only ~20% (and not ~50%) because this only saves us from doing two Halide lowerings and two compile_to_llvm_module calls, but we still run LLVM twice to generate the assembly and object file.

There's another advantage to this: this PR closes a (small) potential crack for bugs to appear in. If Halide/LLVM generated different code for the error check and the assembly op check, the assembly op check could pass, while the error check could generate different code and incorrectly pass (when it would fail if it generated the same code as the op check).

### DIFF
--- a/test/correctness/compute_with.cpp
+++ b/test/correctness/compute_with.cpp
@@ -143,6 +143,7 @@ int split_test() {
 }
 
 int fuse_test() {
+    const int size = 20;
     Buffer<int> im_ref, im;
     {
         Var x("x"), y("y"), z("z");
@@ -151,7 +152,7 @@ int fuse_test() {
         f(x, y, z) = x + y + z;
         g(x, y, z) = x - y + z;
         h(x, y, z) = f(x + 2, y - 1, z + 3) + g(x - 5, y - 6, z + 2);
-        im_ref = h.realize({100, 100, 100});
+        im_ref = h.realize({size, size, size});
     }
 
     {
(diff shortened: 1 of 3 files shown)


abadams:
The idea was to keep the test version as small as possible to minimize the chance of incorrect passes from other nearby code.

Could the error check hit the instruction that the op was supposed to hit (because llvm autovectorized the scalar version or something)?

dsharletg:
I guess both options have problems:
- Checking the error code for the target op could falsely pass because LLVM autovectorized it. Hopefully this doesn't happen because we turned off LLVM loop optimizations.
- Generating the error check separately from the target op code might not actually test the target op at all, and could silently be failing the correctness test.

The second issue is in fact happening now on arm32: the correctness check generates different code from the op check, and so the op check is failing. But this means that the correctness check was not testing the op we thought we were testing. It seems harmless in this case (dense vector stores...), but it demonstrates the issue I think.

This did also speed up simd_op_check (900s -> 761s on the slow arm32 buildbots).

abadams (approved):

PR #5919, opened 2021-04-20
Add a test and hack fix to illustrate an autoscheduler bug.

zvookin:
This is for Andrew to look at per previous discussion.

### DIFF
--- a/src/autoschedulers/adams2019/FunctionDAG.cpp
+++ b/src/autoschedulers/adams2019/FunctionDAG.cpp
@@ -905,6 +905,45 @@ FunctionDAG::FunctionDAG(const vector<Function> &outputs, const MachineParams &p
             // TODO: peephole the boundary condition call pattern instead of assuming the user used the builtin
             node.is_boundary_condition = node.is_pointwise && starts_with(node.func.name(), "repeat_edge");
 
+            // TODO(abadams|zvookin): This is probably not the way to solve this.
+            // The goal here is to illustrate the bug with a test and show that it
+            // is caused by this particualr behavior of bounds inference.
+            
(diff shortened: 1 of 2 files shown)


abadams:
What should I run to repro? It seems to pass for me.

src/autoschedulers/adams2019/FunctionDAG.cpp
```diff
@@ -549,12 +549,48 @@ void FunctionDAG::Edge::expand_footprint(const Span *consumer_loop, Span *produc
...
+        }
+    } remove_image_loads;
+
     map<string, Function> env;
-    for (const Function &o : outputs) {
+    for (const Function &o : outputs_arg) {
         populate_environment(o, env);
     }
 
+    std::vector<Function> outputs;
```
abadams, line 583: I think this is going to mess up the featurization, which happens at the end of this constructor, because image loads are treated differently to extern calls. Can this mutation not just been done only for the things passed to compute_function_value_bounds and boxes_required? Another option would be making two dags (i.e. keeping the original after the deep copy)

zvookin: Certainly easy enough to change the code. Any suggestions for how to test this sort of thing? Either using existing tests or adding one.

zvookin:
@abadams is going to take this over as there are a few subtleties involved. It may be a week or two due to other commitments.

abadams:
@zvookin Finally got around to this. It did indeed turn out to be subtle. scalar image loads were sneaking in via various scopes (e.g. the rdom bounds can use them), so removing them had to be done as the final step. PTAL

steven-johnson:

PR #5924, opened 2021-04-20
Try to reduce runtime size/build time

dsharletg:
This PR tries to reduce the code size and build time of the runtime:
- Reduce inlining attribute usage.
- Build the runtime modules for size, not speed. The modules are re-optimized during Halide builds, but this at least makes libHalide smaller and maybe it takes less time to re-optimize.
- Don't use -g when building the runtime modules. I think debug info in the runtime is rarely if ever used, the -debug target flag is more about enabling runtime debug output.

This reduces the size of libHalide from ~127 MB to ~120 MB, and it appears to reduce the runtime of correctness_cross_compilation by ~10% or so (very noisy measurement).


abadams:
I'm pretty sure debug symbols in the runtime were in response to a specific need for them in a specific use-case. Does no debug symbols in the debug runtime also mean that we can't get reliable stack traces for code called via the runtime in some circumstances (e.g. inside the thread pool workers)?

abadams:
Nope, they were there from the start: 
https://github.com/halide/Halide/commit/641a9ed230fa5b756b8f4976b17f24f4443e476b

This commit looks like it might be relevant though. However, I think we've fixed comdats in a better way since then:
https://github.com/halide/Halide/commit/a5924f11cafc8225f4096e3a5664ab75e1361987

abadams:
When you said build runtime for size, not speed, I assumed you were talking about tweaking compile_standalone_runtime in Module.cpp and the section that makes a runtime in make_module in JITModule.cpp to pass down flags that change O3 in CodeGen_LLVM::optimize_module to the equivalent of Os, and do something similar at the call to engine_builder.setOptLevel in JITModule.cpp

Is that worth doing too?

dsharletg:
I looked at trying to set per-function optimization levels, but it was difficult to get that to work, and decided it wasn't worth it.

We've looked at changing the optimization level in CodeGen_LLVM from O3 to O2 in the past, but I think we rejected that change (but I still think we should do that...).

dsharletg:
I tried reducing the optimization level of the JIT shared runtime, but it didn't appear to make a significant diffference.

dsharletg:
I think the Hexagon failures in this are a result of optimizing for size. @aankit-ca @pranavb-ca what do you think? It seems like this might be a bug in the linker stripping something that shouldn't be stripped on Hexagon?

pranavb-ca:
> I think the Hexagon failures in this are a result of optimizing for size. @aankit-ca @pranavb-ca what do you think? It seems like this might be a bug in the linker stripping something that shouldn't be stripped on Hexagon?

@dsharletg  - We'll need to add the use of libqcc that defines those functions. Those functions are to save callee-saved registers on the stack and are the threshold for generating calls to them instead of inlining them is lower at Os. We'd need to add the following at https://github.com/halide/Halide/blob/master/src/runtime/hexagon_remote/Makefile#L180 and https://github.com/halide/Halide/blob/master/src/runtime/hexagon_remote/Makefile#L186
```
LIB_QCC        =$(shell $(CXX-v62) -G0 -print-file-name=libqcc.a)

$(BIN)/%/hexagon_sim_remote: $(BIN)/%/sim_remote.o $(BIN)/%/known_symbols.o $(BIN)/%/libsim_qurt.a $(BIN)/%/libsim_qurt_vtcm.a
	mkdir -p $(@D)
	$(LD-$*) -o $@ $(CRT0_STANDALONE) $(CRT0) $(INIT) $(BIN)/$*/sim_remote.o $(BIN)/$*/known_symbols.o $(BIN)/$*/libsim_qurt.a $(LIBDL) \
	--start-group  $(LIB_STANDALONE) --whole-archive $(LIB_C) $(BIN)/$*/libsim_qurt_vtcm.a --no-whole-archive $(LIB_GCC) $(LIB_QCC) --end-group $(FINI) \
	--dynamic-linker= -E --force-dynamic
```

dsharletg:
Can you tell if there is a particular function that is requiring this? If it's just one or two functions, maybe we should just inline those?

But if this is a more general thing affecting a lot of function calls, then we should just link that library.

Alternatively... is this a bad idea on Hexagon? Maybe we just shouldn't make this change? What do you think?

pranavb-ca:
> * Build the runtime modules for size, not speed. The modules are re-optimized during Halide builds, but this at least makes libHalide smaller and maybe it takes less time to re-optimize.
@dsharletg - Is this right? I see functions in the runtime getting the ``optsize`` attribute, which means we may not get the full range of -O3 optimization even when these are compiled into the runtime of a pipeline.

dsharletg:
Interesting, I didn't know there was such a thing. I was looking for exactly such an attribute (or an 'optspeed' equivalent), but maybe it only exists in LLVM and not clang. There are things documented along these lines (`optimize` attribute) but I couldn't get them to do anything noticeable.

This is probably still OK (we shouldn't be spending a lot of time in runtime code), but it probably also needs more testing/benchmarking to be sure.

pranavb-ca:
> Can you tell if there is a particular function that is requiring this? If it's just one or two functions, maybe we should just inline those?
> 
> But if this is a more general thing affecting a lot of function calls, then we should just link that library.
> 
> Alternatively... is this a bad idea on Hexagon? Maybe we just shouldn't make this change? What do you think?

The default spill func threshold is 1(https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/Hexagon/HexagonFrameLowering.cpp#L163). This means that if we are going to need two or more instructions to save callee saved registers, LLVM is going to instead generate a call. Therefore, i think quite a few functions will end up calling out to these spill functions.

PR #5925, opened 2021-04-20
Temporarily disable hanging test

steven-johnson:
LLVM13 is causing vector_reductions to hang (https://reviews.llvm.org/D100099 appears to be the injection point). Disabling this test to unbreak the buildbots.


shoaibkamil (approved):

shoaibkamil:
Should we add a tracking issue here as well?

steven-johnson:
> Should we add a tracking issue here as well?

Done

PR #5927, opened 2021-04-20
Fix count_leading/trailing_zeros

dsharletg:
IROperator.h claims count_leading_zeros/count_trailing_zeros return the number of bits if the argument is 0, which is a good definition, and matches ARM's behavior. However, test/correctness/bit_counting.cpp tested a different behavior, and x86 did not match the behavior defined in IROperator.h

### DIFF
--- a/src/CodeGen_X86.cpp
+++ b/src/CodeGen_X86.cpp
@@ -486,6 +486,22 @@ void CodeGen_X86::visit(const Call *op) {
         internal_assert(shift != nullptr) << "Third argument to mulhi_shr intrinsic must be an unsigned integer immediate.\n";
         value = codegen(cast(op->type, p_wide >> op->type.bits()) >> shift->value);
         return;
+    } else if (op->is_intrinsic(Call::count_leading_zeros) ||
+               op->is_intrinsic(Call::count_trailing_zeros)) {
+        internal_assert(op->args.size() == 1);
+        std::vector<llvm::Type *> arg_type(1);
+        arg_type[0] = llvm_type_of(op->args[0].type());
+        llvm::Function *fn = llvm::Intrinsic::getDeclaration(module.get(),
+                                                             (op->is_intrinsic(Call::count_leading_zeros)) ? llvm::Intrinsic::ctlz : llvm::Intrinsic::cttz,
+                                                             arg_type);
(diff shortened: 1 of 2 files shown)


src/CodeGen_X86.cpp
```diff
@@ -486,6 +486,23 @@ void CodeGen_X86::visit(const Call *op) {
...
+        std::vector<llvm::Type *> arg_type(1);
+        arg_type[0] = llvm_type_of(op->args[0].type());
+        llvm::Function *fn = llvm::Intrinsic::getDeclaration(module.get(),
+                                                             (op->is_intrinsic(Call::count_leading_zeros)) ? llvm::Intrinsic::ctlz : llvm::Intrinsic::cttz,
+                                                             arg_type);
+        llvm::Value *is_const_zero_undef = llvm::ConstantInt::getFalse(*context);
+        llvm::Value *args[2] = {codegen(op->args[0]), is_const_zero_undef};
+        CallInst *call = builder->CreateCall(fn, args);
+
+        // On x86, count_leading_zeros of 0 is undefined behavior, fix up this
```
abadams, line 501: This should not be necessary according to LLVM's documentation. It states that if you pass is_const_zero_undef = false, then it returns the bit width if the arg is zero

abadams: "The ‘llvm.ctlz’ intrinsic counts the leading (most significant) zeros in a variable, or within each element of the vector. If src == 0 then the result is the size in bits of the type of src if is_zero_undef == 0 and undef otherwise. For example, llvm.ctlz(i32 2) = 30."

dsharletg: I saw that, but I was confused because I thought it meant only *const* zero, but I guess that was referring to the is_const_zero_undef argument itself?

Either way, what we have currently is definitely not working on some CPUs. Maybe we need to file a bug on LLVM. I'll have to dig into it more.

dsharletg:
Sent #5928 as a subset of this PR, which only updates the test. Maybe the issue I'm seeing will repro on our AMD build bots.

dsharletg:
I'm just going to close this. The workaround we have landed already (a select around count_leading_zeros) is free, and we'll see if #5928 catches any failures.

PR #5929, opened 2021-04-20
Let the user pass the Func to use to the reduction helpers

abadams:
This makes them schedulable if necessary without having to rewrite parts of someone's pipeline.

### DIFF
--- a/src/InlineReductions.cpp
+++ b/src/InlineReductions.cpp
@@ -109,92 +109,157 @@ private:
 }  // namespace Internal
 
 Expr sum(Expr e, const std::string &name) {
-    return sum(RDom(), std::move(e), name);
+    return sum(RDom(), std::move(e), Func(name));
+}
+
+Expr sum(Expr e, Func f) {
+    return sum(RDom(), std::move(e), std::move(f));
 }
 
 Expr sum(const RDom &r, Expr e, const std::string &name) {
-    Internal::FindFreeVars v(r, name);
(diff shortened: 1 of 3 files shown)


dsharletg:
Aren't the reduction helpers equivalent to:

```
Func reduction;
reduction(x) += ...;

reduction.in();
```

If we're going to be encouraging/making a breaking change that requires access to the inner reduction, maybe we should be encouraging migrating to this instead?

abadams:
I think that pattern is cleaner to write, but the breaking change in question - turning off llvm loop optimizations - may require retrofitting old code, which might be difficult to rewrite in that way. Adding a last arg to the inline reduction helpers makes for an easier patch.

steven-johnson:
> Adding a last arg to the inline reduction helpers makes for an easier patch.

True, but it also makes for a weird (IMHO) syntactical wart that we will be stuck with for a long time (maybe forever). Is retrofitting code to the "proper" style really likely to be that tricky?

abadams:
There are also things like argmax, which aren't so easy to write out manually as a summation

abadams:
Also, this isn't really adding a new last arg. Currently they take a string as the last arg, which is the name of the anonymous Func. This upgrades that interface to accept a Func too. That was the rationale behind this design when we discussed it in the dev meeting.

steven-johnson (commented):
Meh, ok

steven-johnson (approved):

PR #5930, opened 2021-04-21
Remove old multiply codegen for ARM

dsharletg:
This removes CodeGen_ARM's Mul visitor. The mul visitor only did one thing differently from CodeGen_LLVM, which was to rewrite multiplies of 2^n +/- 1 to shifts and adds/subtracts. However, even if this is a helpful optimization, LLVM is rewriting it back to a multiply anyways. This codegen also doesn't use CSE, so it generates quite a lot of extra code.

After removing that, the visitor doesn't do anything differently from the base class.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -52,7 +52,6 @@ protected:
     // @{
     void visit(const Cast *) override;
     void visit(const Sub *) override;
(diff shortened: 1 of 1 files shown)


abadams:
I think this made sense on arm-32 back in 2013 or so.

abadams (approved):

PR #5931, opened 2021-04-21
Skip slow tests for non-host targets that don't depend on target features

dsharletg:


steven-johnson (approved):

dsharletg:
It looks like this is skipping these tests even when HL_JIT_TARGET=host :(. It seems like maybe Target::operator!= is doing something surprising. But also maybe this is not a good way to do this.

steven-johnson:
> It looks like this is skipping these tests even when HL_JIT_TARGET=host :(. It seems like maybe Target::operator!= is doing something surprising. But also maybe this is not a good way to do this.

That's weird. Try logging what each Target is?

test/correctness/strict_float.cpp
```diff
@@ -257,6 +257,11 @@ Buffer<float> block_transposed_by_n(Buffer<float> &buf, int vectorize) {
 }
 
 int main(int argc, char **argv) {
+    if (get_jit_target_from_environment() != get_host_target()) {
```
abadams, line 260: Are you sure about this one? It looks like it tests strict_float on a variety of sse/avx levels where llvm may do different types of fusion.

dsharletg:
This probably isn't a big enough improvement to be worth debugging the issues with it.

PR #5934, opened 2021-04-22
Use arm as our generic target triple

abadams:
le64 / le32 have been removed from LLVM mainline. Let's see if using arm triples with an unknown OS works for our purposes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -971,10 +971,11 @@ $(BIN_DIR)/build_halide_h: $(ROOT_DIR)/tools/build_halide_h.cpp
 -include $(OBJECTS:.o=.d)
 -include $(INITIAL_MODULES:.o=.d)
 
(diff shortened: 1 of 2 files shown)


steven-johnson:
We should probably close this.

abadams:
Yeah. Conclusion was that there's no existing target triple that would work for us.

MaskRay:
Does this mean that le32/le64 in clang can be deleted now?

zvookin:
Our current plan is to have Halide folks take over the le32/le64 LLVM backends, hopefully with simplifications due to not needing to support PNaCl. So hopefully they will not be removed.

MaskRay:
That is also fine, but please add some needed tests. The le32/le64 I restored doesn't have any test.

PR #5935, opened 2021-04-22
[Hexagon] Try vdelta/vrdelta before vlut for some shuffles.

aankit-ca:
The patch tries to generate vdelta/vrdelta instructions for non-ramp
shuffles. Eg:

shuffle(lut_expr,
        < 0,  1,  2,  3,  4,  5,  6,  7,
          9, 10, 11, 12, 13, 14, 15, 16,
         18, 19, 20, 21, 22, 23, 24, 25,
         27, 28, 29, 30, 31, 32, 33, 34,
         36, 37, 38, 39, 40, 41, 42, 43,
         45, 46, 47, 48, 49, 50, 51, 52,
         54, 55, 56, 57, 58, 59, 60, 61,
         63, 64, 65, 66, 67, 68, 69, 70>)

can be generated using vrdelta.

The patch also fixes a bug where we bitcast vdelta/vrdelta with 16/32
bits elements to wrong type. User would see the below error:

llvm-project/llvm/lib/IR/Instructions.cpp:2905: static llvm::CastInst
*llvm::CastInst::Create(Instruction::CastOps, llvm::Value *, llvm::Type *,
                        const llvm::Twine &, llvm::Instruction *):
                        Assertion `castIsValid(op, S, Ty) &&
                        "Invalid cast!"' failed.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -1240,7 +1240,7 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
             // Let LLVM handle concat or slices.
             return CodeGen_Posix::shuffle_vectors(a, b, indices);
         }
-        return vlut(concat_vectors({a, b}), indices);
+        return vdelta(concat_vectors({a, b}), indices);
     }
 
     if (stride == 1) {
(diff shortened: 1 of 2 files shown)


aankit-ca:
@dsharletg @pranavb-ca PTAL

dsharletg (approved):
Very cool change!

abadams:
Is there an existing benchmark where this matters?

dsharletg:
We don't have an existing benchmark in Halide, but this would enable writing e.g. unpacking 10-bit code on Hexagon. Perhaps we should write a test that does that, but the simd_op_check test is covering all of the hard part of this, everything else that happens is pretty trivial.

I did try to unpack 10-bit data a while back on Hexagon and couldn't get it to work because the necessary shuffles scalarized, so this would have fixed that problem.

PR #5937, opened 2021-04-23
[Tutorial] Rename lesson 21 into two parts

xgupta:
There are two chapters with the same name, Auto-Scheduler. renaming them into two chapters seems less confusing.

Naming the title I think done from https://github.com/halide/halide.github.com/blob/master/tutorials/tutorial_introduction.html.
For that, I also send a PR - https://github.com/halide/halide.github.com/pull/11.

Signed-off-by: xgupta <shivam98.tkg@rediffmail.com>


xgupta:
gentle ping for comment, if missed.

PR #5939, opened 2021-04-26
Don't use as_const_int() on temporaries

steven-johnson:
Sometimes we get lucky and it's still valid, but it's always wrong.


rootjalex (approved):
Thanks for the catch

PR #5941, opened 2021-04-26
Collapse Shuffle of strided-loads to shuffle's of non-strided loads

aankit-ca:
@dsharletg I don't think we will access any out of bounds memory with the change. Can you take a look once? If you think this is a good change then I'll add the comments and testcase and clean the patch.

### DIFF
--- a/src/Simplify_Shuffle.cpp
+++ b/src/Simplify_Shuffle.cpp
@@ -282,10 +282,36 @@ Expr Simplify::visit(const Shuffle *op, ExprInfo *bounds) {
         }
     }
 
+    int curr_min = 0; int curr_max = 0;
+    vector<int> new_indices = op->indices;
+    for (size_t i = 0; i < new_vectors.size(); i++) {
+        Expr e = new_vectors[i];
+        curr_max = curr_max + e.type().lanes();
+        if (const Load *load = e.as<Load>()) {
+            if (const Ramp *ramp = load->index.as<Ramp>()) {
+                const int64_t *stride = as_const_int(ramp->stride);
+                if (stride && (*stride != 1) && is_const_one(load->predicate)) {
+                    changed = true;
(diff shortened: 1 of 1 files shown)


src/Simplify_Shuffle.cpp
```diff
@@ -282,10 +282,36 @@ Expr Simplify::visit(const Shuffle *op, ExprInfo *bounds) {
...
+                const int64_t *stride = as_const_int(ramp->stride);
+                if (stride && (*stride != 1) && is_const_one(load->predicate)) {
+                    changed = true;
+                    for (size_t j = 0; j < new_indices.size(); j++) {
+                        int index = new_indices[j];
+                        if ((index >= curr_min) && (index < curr_max)) {
+                            new_indices[j] = index * (*stride);
+                        }
+                    }
+                    Expr new_load_index = Ramp::make(ramp->base, 1, ramp->lanes * (*stride));
```
abadams, line 301: Consider ramp->base = 0, ramp->lanes = 2, ramp->stride = 10

The original load accesses two elements at 0 and 10. This new load accesses all elements from 0 to 19 inclusive.

Isn't that a potentially-new out of bounds load?

aankit-ca: Aah. Got it. I think this expression would prevent this from happening.

`Expr new_load_index = Ramp::make(ramp->base, 1, (ramp->lanes - 1) * (*stride) + 1);`

dsharletg:
I think this might be a harmful change on targets that can do strided loads well, like ARM?

There probably also needs to be some upper bound on the stride that we do this with. Rewriting a load with a large stride into a dense stride might also be harmful.

aankit-ca:
Ok. Closing this for now. Dillon suggested a better way to do this in a way that does not affect other targets. Will try that in the future.

PR #5942, opened 2021-04-27
Add support for dynamic tensors to hannk

steven-johnson:


apps/hannk/interpreter/model.cpp
```diff
@@ -145,7 +151,29 @@ void Tensor::allocate() {
...
+
+    const halide_type_t type = buffer_.type();
+    std::vector<halide_dimension_t> new_dims;
+    new_dims.reserve(new_shape.size());
+    int stride = 1;
+    for (const auto &d : new_shape) {
+        new_dims.emplace_back(d.min, d.extent(), stride);
+        stride *= d.extent();
+    }
+    HalideBuffer<void> new_buffer(type, nullptr, new_dims);
```
dsharletg, line 165: This should check if the new dims are actually different, and skip the expensive reallocation below if they are the same. The ReshapeOp calls this unconditionally, if it is used in a loop, this could be very significant.

steven-johnson: PTAL

dsharletg (approved):

PR #5946, opened 2021-04-29
Replace llvm::sys::fs::F_None with llvm::sys::fs::OF_None

MaskRay:
The former is deprecated.

### DIFF
--- a/src/LLVM_Output.cpp
+++ b/src/LLVM_Output.cpp
@@ -306,7 +306,7 @@ void write_coff_archive(std::ostream &out,
 std::unique_ptr<llvm::raw_fd_ostream> make_raw_fd_ostream(const std::string &filename) {
     std::string error_string;
     std::error_code err;
-    std::unique_ptr<llvm::raw_fd_ostream> raw_out(new llvm::raw_fd_ostream(filename, err, llvm::sys::fs::F_None));
+    std::unique_ptr<llvm::raw_fd_ostream> raw_out(new llvm::raw_fd_ostream(filename, err, llvm::sys::fs::OF_None));
     if (err) {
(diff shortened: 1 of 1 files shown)


src/LLVM_Output.cpp
```diff
@@ -306,7 +306,7 @@ void write_coff_archive(std::ostream &out,
 std::unique_ptr<llvm::raw_fd_ostream> make_raw_fd_ostream(const std::string &filename) {
     std::string error_string;
     std::error_code err;
-    std::unique_ptr<llvm::raw_fd_ostream> raw_out(new llvm::raw_fd_ostream(filename, err, llvm::sys::fs::F_None));
+    std::unique_ptr<llvm::raw_fd_ostream> raw_out(new llvm::raw_fd_ostream(filename, err, llvm::sys::fs::OF_None));
```
steven-johnson, line 309: This probably needs to be guarded with LLVM_VERSION, since we build against LLVM trunk + one or two older versions, eg.
```
#if LLVM_VERSION >= 130  /* 13.0 */
    std::unique_ptr<llvm::raw_fd_ostream> raw_out(new llvm::raw_fd_ostream(filename, err, llvm::sys::fs::OF_None));
#else
    std::unique_ptr<llvm::raw_fd_ostream> raw_out(new llvm::raw_fd_ostream(filename, err, llvm::sys::fs::F_None));
#endif
```

MaskRay: We don't. OF_ is available since 2018 https://reviews.llvm.org/rG1f67a3cba9b09636c56e2109d8a35ae96dc15782

dsharletg (approved):
Thanks for the fix!

PR #5947, opened 2021-04-29
Fix messy bounds due to specializations in inner loops

dsharletg:
When specializing a func in an inner loop and using `specialize_fail`, bounds can be very messy because the bounds are conditionalized. This PR cleans up these bounds by ignoring the condition of the if when one of the branches of an if terminates the program.

I am not completely sure this is safe. I suppose this could change *why* a program terminates early (which assert fails?), but not if it terminates early?


abadams:
I don't understand. If you have if (foo == 3) { access buffer bar} else { exit }, don't you want that condition to appear in box.used, so that buffer bar is only conditionally allocated?

dsharletg:
Here's the problem I'm trying to solve. Here's a snippet of IR:

```
 for (output.s0.x.x, 0, t1005) {
  assert(let t990 = select(t1008, max(t1007, -2), t1007) in (let t991 = select(t1009, max(t990, -3), t990) in (let t992 = select(t1010, max(t991, -4), t991) in ((max(max(select(t1024, max(t992, -5), t992), program.extent.1), 0) + -23) <= t1006))), let t994 = select(t1008, max(t1007, -2), t1007) in (let t995 = select(t1009, max(t994, -3), t994) in (let t996 = select(t1010, max(t995, -4), t995) in halide_error_explicit_bounds_too_small("u", "scratch", t1006 + -1, t1006 + 23, t1006 + -1, max(max(select(t1024, max(t996, -5), t996), program.extent.1), 0)))))
  allocate scratch[int16 * 400] in Register
  produce scratch {
   if (t1026) {
    scratch[ramp(int64((0 - t1006))*(int64)16, (int64)1, 16) aligned(16, 0)] = inputs_0[ramp(int64(((output.s0.x.x*16) + output.0.min.0)) - int64(inputs_0.min.0), (int64)1, 16)]
   } else if (t1025) {
    scratch[ramp(int64((0 - t1006))*(int64)16, (int64)1, 16) aligned(16, 0)] = x16(inputs_0[(int64)0])
   } else {
    assert((uint1)0, halide_error_specialize_fail("Input dimension 0 must have stride 0 or 1."))
   }
   ...
```

The schedule here is specializing scratch (an update definition) based on whether inputs_0 has stride 0 or 1 for dimension 1. Doing this specialization causes that big assert to be generated. The reason is that it is only requiring inputs_0.extent to be big enough if it doesn't hit the assert. But, this seems pointless, because it's going to assert and exit anyways.

Is there a better way to fix this?

steven-johnson:
Where does this PR stand?

steven-johnson:
Is this PR still under active consideration, or should it be retired?

PR #5955, opened 2021-05-03
Add CMake build rules to apps/hannk

steven-johnson:
This is (mostly) correct, I think, with one major issue and a few minor ones:

- The big issue is that I can't figure out the incantation needed to get CMake to cross-compile for Android (pointing it at a built-for-host Halide install fails). I'm assuming for now that there is some oversight or workaround for this (if CMake can't handle this then I'm going to punt on the whole thing).

- The minor-ish issues mainly have to do with noise and junk during the fetched-external projects (e.g., touching anything in one of our CMake files will make the Eigen subproject spew a lot of junk and I haven't figured out how to disable that).

- By default, this fetches & builds the proper version of TFLite; this is nice for seamlessness but adds about 4 minutes to a clean build on my laptop. We could probably add an option to point at an already-built local library (as we do for Make) but I'm not going to bother just yet unless people complain.

### DIFF
--- a/apps/hannk/CMakeLists.txt
+++ b/apps/hannk/CMakeLists.txt
@@ -58,3 +58,11 @@ target_link_libraries(compare_vs_tflite PRIVATE
 target_include_directories(compare_vs_tflite
                            PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
 
+# Tests
+file(GLOB TEST_FILES "test/*/*")
+foreach (t IN LISTS TEST_FILES)
+    add_test(NAME hannk_test_${t} COMMAND compare_vs_tflite ${t} --benchmark 0)
+    set_tests_properties(hannk_test_${t} PROPERTIES
+                         LABELS hannk_tests)
+endforeach()
+
--- a/apps/hannk/CMakeLists.txt
+++ b/apps/hannk/CMakeLists.txt
@@ -5,7 +5,7 @@ enable_testing()
 
 # -fPIC is necessary for .so builds (at least on Linux); not necessary for the non-delegate
 # builds but easier to enable it for everything.
-add_compile_options(-fPIC)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # Set up language settings
 set(CMAKE_CXX_STANDARD 11)
@@ -56,10 +56,10 @@ target_link_libraries(compare_vs_tflite PRIVATE
                       Halide::Tools  # for halide_benchmark.h
                       Halide::Runtime)
 target_include_directories(compare_vs_tflite
-                           PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
+                           PUBLIC $<BUILD_INTERFACE:${hannk_SOURCE_DIR}>)
 
 # Tests
-file(GLOB TEST_FILES "test/*/*")
+file(GLOB TEST_FILES CONFIGURE_DEPENDS "test/*/*")
 foreach (t IN LISTS TEST_FILES)
     add_test(NAME hannk_test_${t} COMMAND compare_vs_tflite ${t} --benchmark 0)
     set_tests_properties(hannk_test_${t} PROPERTIES
--- a/apps/hannk/CMakeLists.txt
+++ b/apps/hannk/CMakeLists.txt
@@ -1,0 +1,60 @@
+cmake_minimum_required(VERSION 3.16)
+project(hannk)
+
+enable_testing()
+
+# -fPIC is necessary for .so builds (at least on Linux); not necessary for the non-delegate
+# builds but easier to enable it for everything.
+add_compile_options(-fPIC)
+
+# Set up language settings
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED YES)
+set(CMAKE_CXX_EXTENSIONS NO)
+
+# Find Halide
+find_package(Halide REQUIRED)
+
+# Set up the version of TFLite we expect
+set(TFLITE_VERSION_MAJOR "2" CACHE STRING "Major version of TFLite to assume")
+set(TFLITE_VERSION_MINOR "5" CACHE STRING "Minor version of TFLite to assume")
+set(TFLITE_VERSION_PATCH "0-rc2" CACHE STRING "Patch version of TFLite to assume")
+
+set(TFLITE_VERSION "${TFLITE_VERSION_MAJOR}.${TFLITE_VERSION_MINOR}.${TFLITE_VERSION_PATCH}")
+
+add_subdirectory(delegate)
+add_subdirectory(halide)
+add_subdirectory(interpreter)
+add_subdirectory(tflite)
+add_subdirectory(util)
+
+# Benchmarking executable
+add_executable(benchmark benchmark.cpp)
+target_link_libraries(benchmark PRIVATE
+                      tflite_parser
+                      interpreter
+                      error_util
+                      file_util
+                      hannk_log_stderr
+                      Halide::Tools  # for halide_benchmark.h
+                      Halide::Runtime)
+
+add_executable(compare_vs_tflite compare_vs_tflite.cpp)
+target_compile_definitions(compare_vs_tflite PRIVATE
+                           -DTFLITE_VERSION_MAJOR=${TFLITE_VERSION_MAJOR}
+                           -DTFLITE_VERSION_MINOR=${TFLITE_VERSION_MINOR}
+                           -DTFLITE_VERSION_PATCH=${TFLITE_VERSION_PATCH}
+                           -DHANNK_USE_REGISTRAR_DELEGATE)
+target_link_libraries(compare_vs_tflite PRIVATE
+                      tensorflowlite_c
+                      tflite_parser
+                      interpreter
+                      error_util
+                      file_util
+                      hannk_log_stderr
+                      hannk_delegate_registrar
+                      Halide::Tools  # for halide_benchmark.h
+                      Halide::Runtime)
+target_include_directories(compare_vs_tflite
+                           PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
+
--- a/apps/hannk/configure_cmake.sh
+++ b/apps/hannk/configure_cmake.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+set -e
+
+HANNK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+BUILD_DIR="${HANNK_DIR}/build"
+
+mkdir -p "${BUILD_DIR}"
+cd "${BUILD_DIR}"
+
+if [ -z ${HALIDE_INSTALL_PATH} ]; then
+HALIDE_INSTALL_PATH=${HOME}/halide-13-install
+fi
+echo Using HalideInstall=${HALIDE_INSTALL_PATH}
+
+if [ -z ${HL_TARGET} ]; then
+HL_TARGET=host
+fi
+echo Using HL_TARGET=${HL_TARGET}
+
+if [ -z ${HALIDE_BUILDER} ]; then
+HALIDE_BUILDER=Ninja
+fi
+echo Using build tool=${HALIDE_BUILDER}
+
+EXTRAS=
+# TODO: this doesn't work (yet); crosscompiling in CMake is painful.
+if [[ "${HL_TARGET}" =~ ^arm-64-android.* ]]; then
+echo Configuring for Android arm64-v8a build...
+echo Using ANDROID_NDK_ROOT=${ANDROID_NDK_ROOT}
+EXTRAS="-DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK_ROOT}/build/cmake/android.toolchain.cmake"
+else
+echo Assuming host build...
+fi
+
+cmake \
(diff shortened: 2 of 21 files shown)


apps/hannk/halide/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+                       NAMESPACE hannk
+                       GENERATOR ${args_GENERATOR_NAME}
+                       FEATURES large_buffers c_plus_plus_name_mangling
+                       PARAMS ${args_GENERATOR_ARGS})
+
+    set(_HANNK_HALIDE_TARGETS ${_HANNK_HALIDE_TARGETS} ${args_TARGET} PARENT_SCOPE)
+endfunction()
+
+_add_hannk_halide_library(
+        TARGET add_uint8_uint8
```
dsharletg, line 29: Is this losing `no_bounds_query` from the elementwise generators?

apps/hannk/halide/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+
+_add_hannk_halide_library(
+        TARGET depthwise_conv_dm1_uint8
+        SRCS depthwise_conv_generator.cpp
+        GENERATOR_NAME DepthwiseConv
+        GENERATOR_ARGS inv_depth_multiplier=1
+        GENERATOR_DEPS)
+
+_add_hannk_halide_library(
+        TARGET fill_uint8
```
dsharletg, line 78: Is this losing `no_asserts`?

dsharletg (approved):
The CMake seems OK, @alexreinking would probably be better able to review.

apps/hannk/halide/CMakeLists.txt
steven-johnson: Those were added after this code is written, so it's less "losing" and more "now we have two build files to maintain"

apps/hannk/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
+cmake_minimum_required(VERSION 3.16)
+project(hannk)
+
+enable_testing()
+
+# -fPIC is necessary for .so builds (at least on Linux); not necessary for the non-delegate
+# builds but easier to enable it for everything.
+add_compile_options(-fPIC)
+
```
alexreinking, line 9: `set(CMAKE_POSITION_INDEPENDENT_CODE ON)`

apps/hannk/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+                      tflite_parser
+                      interpreter
+                      error_util
+                      file_util
+                      hannk_log_stderr
+                      hannk_delegate_registrar
+                      Halide::Tools  # for halide_benchmark.h
+                      Halide::Runtime)
+target_include_directories(compare_vs_tflite
+                           PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
```
alexreinking, line 59: `CMAKE_CURRENT_SOURCE_DIR`

apps/hannk/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+# Tests
+file(GLOB TEST_FILES "test/*/*")
```
alexreinking, line 62: I'm pretty against GLOBs in CMake. I'd _prefer_ this were expanded into a full list, but if we're going to keep the glob, it needs to use `CONFIGURE_DEPENDS` so that re-builds can pick up on new tests added (by automatically re-running CMake).
```
file(GLOB TEST_FILES CONFIGURE_DEPENDS "test/*/*")
```

apps/hannk/delegate/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+add_library(hannk_delegate_registrar STATIC
+            hannk_delegate_provider.cpp
+            hannk_delegate.cpp)
+target_link_libraries(hannk_delegate_registrar PRIVATE
+                      interpreter
+                      error_util
+                      hannk_log_tflite
+                      tensorflowlite_headers)
+target_include_directories(hannk_delegate_registrar PRIVATE
+                           $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
```
alexreinking, line 12: Do you really mean `CMAKE_SOURCE_DIR`, here? Or do you mean `CMAKE_CURRENT_SOURCE_DIR` or `hannk_SOURCE_DIR`?

apps/hannk/delegate/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+                           $<BUILD_INTERFACE:${CMAKE_PREFIX_PATH}/include>)
```
alexreinking, line 13: This is alarming. There's no way this path is right since `CMAKE_PREFIX_PATH` isn't guaranteed to be set.

apps/hannk/delegate/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+
+# --------------- external (shared-library) delegate
+
+# External delegates in TFLite are always .so (never .dylib);
+# on Mac we must declare the library to be MODULE rather than SHARED to get this
+if (APPLE)
+set(SOTYPE MODULE)
+else ()
+set(SOTYPE SHARED)
+endif ()
```
alexreinking, line 24: Why not always use `MODULE`?

apps/hannk/delegate/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+            hannk_delegate_adaptor.cpp
+            hannk_delegate.cpp)
+target_link_libraries(hannk_delegate_external PRIVATE
+                      interpreter
+                      error_util
+                      hannk_log_tflite
+                      tensorflowlite_headers)
+target_include_directories(hannk_delegate_external PRIVATE
+                           $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
+                           $<BUILD_INTERFACE:${CMAKE_PREFIX_PATH}/include>)
```
alexreinking, line 36: Same as above.

apps/hannk/delegate/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+if (APPLE)
+    target_link_options(hannk_delegate_external PRIVATE "-Wl,-exported_symbols_list,${CMAKE_CURRENT_SOURCE_DIR}/exported_symbols.osx")
+else ()
+    target_link_options(hannk_delegate_external PRIVATE "-Wl,--version-script,${CMAKE_CURRENT_SOURCE_DIR}/exported_symbols.ldscript"
+                                                      "-Wl,-soname,libHannkDelegate.so")
```
alexreinking, line 41: Why is the `-soname` flag necessary when `OUTPUT_NAME` is set below?

apps/hannk/halide/CMakeLists.txt
```diff
@@ -0,0 +1,164 @@
...
+# Function to reduce boilerplate
+function(_add_hannk_halide_library)
+    set(options)
+    set(oneValueArgs TARGET GENERATOR_NAME)
+    set(multiValueArgs SRCS GENERATOR_ARGS GENERATOR_DEPS FEATURES)
+    cmake_parse_arguments(args "" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+    add_executable(${args_TARGET}.generator ${args_SRCS})
+    target_link_libraries(${args_TARGET}.generator PRIVATE ${args_GENERATOR_DEPS} common_halide Halide::Generator)
+    target_include_directories(${args_TARGET}.generator PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
```
alexreinking, line 17: Again, `CMAKE_CURRENT_SOURCE_DIR` or `haank_SOURCE_DIR`?

apps/hannk/halide/CMakeLists.txt
```diff
@@ -0,0 +1,164 @@
+
+add_library(common_halide STATIC common_halide.cpp)
+target_link_libraries(common_halide PRIVATE Halide::Halide)
+target_include_directories(common_halide PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
```
alexreinking, line 4: `CMAKE_CURRENT_SOURCE_DIR` or `haank_SOURCE_DIR`?

apps/hannk/halide/CMakeLists.txt
```diff
@@ -0,0 +1,164 @@
...
+_add_hannk_halide_library(
+        TARGET tile_conv_filter_uint8
+        SRCS conv_generator.cpp
+        GENERATOR_NAME TileConvFilter
+        GENERATOR_ARGS
+        GENERATOR_DEPS)
+
+add_library(op_impls INTERFACE)
+target_link_libraries(op_impls INTERFACE ${_HANNK_HALIDE_TARGETS})
+target_include_directories(op_impls INTERFACE $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>)
```
alexreinking, line 164: `CMAKE_CURRENT_BINARY_DIR`.

apps/hannk/interpreter/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+target_link_libraries(elementwise_program PRIVATE Halide::Runtime)
+
+add_library(interpreter STATIC
+            interpreter.cpp
+            interval.cpp
+            lower.cpp
+            model.cpp
+            ops.cpp
+            transforms.cpp)
+target_include_directories(interpreter PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
```
alexreinking, line 14: `CMAKE_CURRENT_SOURCE_DIR` or `haank_SOURCE_DIR`?

apps/hannk/tflite/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+add_library(tensorflowlite_headers INTERFACE)
+target_include_directories(tensorflowlite_headers INTERFACE
+                           $<BUILD_INTERFACE:${tflite_SOURCE_DIR}>)
+
+# ----------------
+
+add_library(tflite_parser STATIC tflite_parser.cpp)
+target_include_directories(tflite_parser
+                           PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
+                           PRIVATE $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
```
alexreinking, line 58: `CMAKE_CURRENT_SOURCE_DIR` or `haank_SOURCE_DIR`? (c.f. `BINARY`)

apps/hannk/util/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+add_library(hannk_log_stderr STATIC
+            hannk_log_stderr.cpp)
+target_include_directories(hannk_log_stderr PUBLIC
+                           $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
+
+add_library(hannk_log_tflite STATIC
+            hannk_log_tflite.cpp)
+target_link_libraries(hannk_log_tflite PRIVATE tensorflowlite_headers)
+target_include_directories(hannk_log_tflite PUBLIC
+                           $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>)
```
alexreinking, line 16: From context, `haank_SOURCE_DIR`.

alexreinking (changes requested):
The biggest thing I'm noticing here is the improper use of `CMAKE_SOURCE_DIR`. If we `add_subdirectory` this app in the global `apps` project, then `CMAKE_SOURCE_DIR` will refer to the _apps_ folder. Not what we want, most likely.

Instead, these references should be changed to `haank_SOURCE_DIR`, which is unambiguously the absolute path to `Halide/apps/haank`, or to `CMAKE_CURRENT_SOURCE_DIR` if that's what's actually intended.

alexreinking:
>     * The big issue is that I can't figure out the incantation needed to get CMake to cross-compile for Android (pointing it at a built-for-host Halide install fails). I'm assuming for now that there is some oversight or workaround for this (if CMake can't handle this then I'm going to punt on the whole thing).

Regarding this, have you tried using `Halide_DIR` instead of `CMAKE_PREFIX_PATH`? CMake tries to use platform-specific library prefixes when searching for a package, so telling it _exactly_ where the version of Halide you want to use is might work.

If you can help me get an Android development environment set up locally, I can try this.

apps/hannk/delegate/CMakeLists.txt
alexreinking: Also, does Hannk not support Windows? If so, that should be checked in the top level. If not, this needs to be adjusted to not add `exported_symbols.ldscript` on Windows.

alexreinking (changes requested):

steven-johnson:
> Also, does Hannk not support Windows?

It may at some point, but we're spending zero time on that at present. If/when demand for hannk-on-windows emerges, we'll revisit this.

apps/hannk/configure_cmake.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [ -z ${HL_TARGET} ]; then
+HL_TARGET=host
+fi
+echo Using HL_TARGET=${HL_TARGET}
+
+if [ -z ${HALIDE_BUILDER} ]; then
+HALIDE_BUILDER=Ninja
+fi
+echo Using build tool=${HALIDE_BUILDER}
```
alexreinking, line 24: CMake respects the `CMAKE_GENERATOR` environment variable already.

alexreinking (approved):

PR #5957, opened 2021-05-04
Small H::R::B cleanups and improvements

dsharletg:
- Reduce duplication of some tricky code with `halide_buffer_t`
- Combine `decref` and `decref_dev`. This might help a little with overhead, and if nothing else, makes it easier to identify it (one less function in the long tail of profiler output).

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -217,8 +217,8 @@ private:
 
     /** Decrement the reference count of any owned allocation and free host
      * and device memory if it hits zero. Sets alloc to nullptr. */
-    void decref() {
-        if (owns_host_memory()) {
+    void decref(bool device_only = false) {
+        if (owns_host_memory() && !device_only) {
             int new_count = --(alloc->ref_count);
             if (new_count == 0) {
                 void (*fn)(void *) = alloc->deallocate_fn;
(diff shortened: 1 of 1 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -229,42 +229,35 @@ class Buffer {
...
-                       "before the device allocation is freed.");
-                if (dev_ref_count && dev_ref_count->ownership == BufferDeviceOwnership::WrappedNative) {
-                    buf.device_interface->detach_native(nullptr, &buf);
-                } else if (dev_ref_count && dev_ref_count->ownership == BufferDeviceOwnership::AllocatedDeviceAndHost) {
-                    buf.device_interface->device_and_host_free(nullptr, &buf);
-                } else if (dev_ref_count && dev_ref_count->ownership == BufferDeviceOwnership::Cropped) {
-                    buf.device_interface->device_release_crop(nullptr, &buf);
-                } else if (dev_ref_count == nullptr || dev_ref_count->ownership == BufferDeviceOwnership::Allocated) {
-                    buf.device_interface->device_free(nullptr, &buf);
+            int new_count = --(dev_ref_count->count);
```
steven-johnson, line 233: I would have been tempted to think that the previous code setup was better, but only on the assumption that `decref()` was usually inlined and that `decref_dev()` is usually not-inlined. Sounds like that's probably not happening based on your observations with the profiler. I wonder if adding always-inline and never-inline directives would be worth experimenting with.

steven-johnson:
Looking at this change (and also some profiler results), the overhead for the dtor for this class is surprisingly high -- which is compounded by the fact that temporaries seem to be common in typical usage, so there are a lot of hidden dtor calls. Would be nice if this could be improved further.

src/runtime/HalideBuffer.h
dsharletg: I think either way, there is one non-inline call, and I think it's useful to consolidate the two, so we can better see what the overhead is in a profiler. And if for some reason neither one inlines, then it's an improvement (from two calls to one).

PR #5959, opened 2021-05-05
Temporarily disable apps/hannk

steven-johnson:
It's failing to config/build properly on the buildbots (but not locally); disabling to avoid build failures until I sort this out


steven-johnson:
Closing in favor of https://github.com/halide/Halide/pull/5960

PR #5962, opened 2021-05-05
Ensure our local flatbuffers.h is included in preference to system variants

steven-johnson:
The local version might be too old for TFLite

### DIFF
--- a/apps/hannk/tflite/CMakeLists.txt
+++ b/apps/hannk/tflite/CMakeLists.txt
@@ -30,7 +30,12 @@ FetchContent_Declare(tflite
 FetchContent_GetProperties(tflite)
 if (NOT tflite_POPULATED)
     FetchContent_Populate(tflite)
+    # Some of the subprojects (e.g. Eigen) are very noisy and emit status messages all the time.
+    # Temporary ignore status messages while adding this to silence it. Ugly but effective.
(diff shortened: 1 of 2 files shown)


alexreinking:
Needs a comment but otherwise LGTM

steven-johnson:
PTAL

apps/hannk/tflite/CMakeLists.txt
```diff
@@ -13,6 +13,11 @@ set(TFLITE_ENABLE_RUY OFF)
 set(TFLITE_ENABLE_XNNPACK OFF)
 
 # Also have to disable some stuff dragged in by TFLite
+if(POLICY CMP0077)
+  # This won't do the trick -- we need to use CMAKE_POLICY_DEFAULT_CMP0077, alas
+  # cmake_policy(SET CMP0077 NEW)
+  set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+endif()
```
alexreinking, line 20: No need to test (`if`) for the policy before setting it because the minimum is 3.16, which means the policy exists. This line should also be moved to the top-level, beneath the `project()` call in `apps/hannk/CMakeLists.txt`.

alexreinking (changes requested):

apps/hannk/tflite/CMakeLists.txt
steven-johnson: done

alexreinking (approved):

PR #5963, opened 2021-05-05
Optimize pooling ops

dsharletg:
- Avoid external padding with boundary conditions, but still avoiding the code explosion from partitioning.
- Implement division in average pooling with a reciprocal.

This makes the pooling ops 2-4x faster.


steven-johnson (approved):

dsharletg:
TFlite build is failing anyways.

Issue #5953, opened 2021-05-03
Dynamically-typed scalar parameters


abadams:
Currently we can specify an `Input<Buffer<>>` to a Generator, which means that the type is given by a GeneratorParam, or perhaps it's a pointer to or vector of `Input<Buffer<>>`, in which case the type is given as an argument to add_input in configure(). 

There's no equivalent way to do that for scalars. The add_input method for scalar parameters requires a static type, and Input<void> doesn't seem to be a legal type for a scalar parameter with type determined by GeneratorParam.

slomp:
I'd also advocate for a more "generic" way to rely even less on encapsulated types.
For example, a generic declaration like `Input<void>` should allow for it to be later transformed into a "Buffer", or "Func', or "Scalar".
Something like: `::add_input(Layout, Type, Dim, Size)` (or add_scalar/add_buffer/add_func).

abadams:
I'm not sure that could work, because different C++ syntax is available for scalar vs non-scalar input members (e.g. operator() exists)

slomp:
Ah... indeed... :/

steven-johnson:
To be clear: is the goal here to
- Allow declaring something like `Input<void>`, so that there is a required input, but whose type must be set in the build system?
- Allow calling `add_input(halide_type_t, ...)` from the configure method, to allow for dynamically-added inputs with dynamic type?
- Both?

abadams:
Both.

PR #5965, opened 2021-05-05
Add dynamically-typed scalar inputs to Generator (#5953)

steven-johnson:


abadams:
I see you went with `Input<Expr>`, which I guess pairs well with `Input<Func>`, but less well with `Input<Buffer<>>` or `Input<float>` (which is also an Expr)

Why `Input<Expr>` rather than `Input<void>`? I guess either could be `Input<>` / `add_input<>` with an appropriate choice of default template parameter. Not sure if we should do that.

steven-johnson:
> Why Input<Expr> rather than Input<void>?

I actually started out wanting to use Input<void>, but too much of the internal infrastructure of the GeneratorInput classes assume that the type is one that can actually be instantiated. Rather than getting dragged into a yak shave adding even more complexity to an already-way-too-complex set of template classes, I opted for Input<Expr> as the easy way out.

steven-johnson:
(I mean, we could probably add some syntactic sugar to allow Input<void> -> Input<Expr>...)

steven-johnson:
All green. Do we like this?

abadams:
I still feel icky about Input<Expr>, but I'm not sure Input<void> is any better. I don't want to argue too long about the color of the bike shed, as this is a minor thing, but this is the sort of decision that might be hard to change later without breaking code, so I would appreciate brief opinions from more people. E.g. @shoaibkamil @dsharletg @zvookin

dsharletg:
I'm not sure about this, because if a parameter is scalar, casting it to some type (including a type depending on a GeneratorParam) should be basically free. That cast could be lifted out of any loops in the pipeline.

So, this behavior can be had basically for free by just making the parameter big enough to hold any types you want to pass, and casting it to the type you want. Do we want to add ~100 lines of code (and more importantly, a new thing to document/read about/understand) to avoid that?

abadams:
There's no one type that can hold either uint64 or double. It's also weird if you're trying to generate a family of overloads:

`f_int(Buffer<int>, int)`, `f_float(Buffer<float>, float)`, etc to have them all use the same large scalar type

shoaibkamil:
> I still feel icky about Input<Expr>, but I'm not sure Input<void> is any better.

Yeah, I agree.  I'm on the `Input<Expr>` side of the fence here though-- I think it is more discoverable than `Input<void>`.  From my side, I think this is good.  @slomp may have an opinion too, since he's closer to the code that needs this.

slomp:
Well, I actually like the `Input<Expr>` better (idiomatically, a meta-type like `Input<Any>` would be more expressive and less ambiguous, but one step at a time). I'm actually of the opposite view of @abadams and @shoaibkamil: `Input<void>` just feels hacky to me; any C++ generic code that specializes on `void` feels like it's just trying to exploit some cracks in the rather poor C++ type system.

In the end, so long as we can express arbitrary Halide types (scalar, buffer and func-like) with `Input<Expr>`, I think it's a first step. I don't expect many people to use this, so we should be OK to further iterate on the design on the future and deprecate/remap `Input<Expr>`.

steven-johnson:
> so long as we can express arbitrary Halide types (scalar, buffer and func-like) with Input<Expr>

`Input<Expr>` is *only* a scalar, and cannot express buffer or func-like.

dsharletg:
> > so long as we can express arbitrary Halide types (scalar, buffer and func-like) with Input

> Input<Expr> is only a scalar, and cannot express buffer or func-like.

If this is what you want, I wonder if what you should be doing here is actually an Input<Buffer<>>, and just use zero strides for scalars? As a bonus, this allows more broadcasting than just scalar or not. The test I added in https://github.com/halide/Halide/pull/5950 demonstrates this.

abadams (approved):
Seems like people are cool with `Input<Expr>`

slomp:
Ok, so, if I understand correctly, we have:
- `Input<Buffer<>>` for buffers
- `Input<Expr>` for scalars
- `Input<Func>` for buffers, funcs, and constants-disguised-as-0-dim-funcs

I think we're good then.

PR #5968, opened 2021-05-05
Add transpose op

dsharletg:
Also relax some unnecessary type constraints, and add a few extra related tests.


apps/hannk/delegate/hannk_delegate.cpp
```diff
@@ -1396,9 +1404,6 @@ class NodeSupport {
         if (!IsVersionOK(1, 2)) {
             return false;
         }
-        if (!InputsHaveCorrectTypes({ANY})) {
```
steven-johnson, line 1399: Removing this means that the shape op can have any number of inputs; this call ensures that there is exactly one input (but it can be of any type). Is that what we want?

dsharletg: Good point, fixed

apps/hannk/delegate/hannk_delegate.cpp
```diff
@@ -1303,11 +1303,9 @@ class NodeSupport {
         if (!IsVersionOK(1, 2)) {
             return false;
         }
-
-        if (!InputHasType(1, I32)) {
+        if (!InputsHaveCorrectTypes({ANY, I32})) {
```
steven-johnson, line 1306: The current helper funcs for detecting "is supported" feel poorly thought out to me (e.g. it was easy for you to make this mistake, which is not good), but not a lot of thought was put into it at the start. At this point though I wonder is switching to a table-driven approach would be worthwhile.

dsharletg: I agree, but we shouldn't change it in this PR.

steven-johnson: Agreed, wasn't suggesting doing it here, just in general

steven-johnson (approved):

PR #5972, opened 2021-05-06
Optimize add generator

dsharletg:
One of the 3 quantized multiplications is unnecessary. This speeds up add quite a bit (close to 50% on ARM).

### DIFF
--- a/apps/hannk/halide/elementwise_generator.cpp
+++ b/apps/hannk/halide/elementwise_generator.cpp
@@ -21,7 +21,6 @@ public:
     Input<uint32_t> input2_shift_{"input2_shift"};
 
     Input<uint8_t> output_zero_{"output_zero"};
-    Input<uint32_t> output_shift_{"output_shift"};
     Input<uint8_t> output_min_{"output_min"};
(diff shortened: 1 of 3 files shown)


vksnk (approved):

apps/hannk/halide/elementwise_generator.cpp
```diff
@@ -32,22 +27,21 @@ class Add : public Generator<Add> {
     void generate() {
         Var x("x"), y("y");
 
-        Expr input1 = i32(i16(input1_(x, y)) - i16(input1_zero_)) << 20;
-        Expr input2 = i32(i16(input2_(x, y)) - i16(input2_zero_)) << 20;
+        Expr input1 = (i16(input1_(x, y)) - i16(input1_zero_)) << 6;
```
steven-johnson, line 30: It's not obvious to me why 6 is the shift amount; is it worth a comment?

steven-johnson (approved):

dsharletg:
Going to merge this because the only change in the last commit was a comment, maybe I can save the build bots some work.

PR #5976, opened 2021-05-06
Rework CMake packaging to improve cross-compiling support

alexreinking:
There are several changes here.

1. The `Halide` CMake package is split into `Halide` and `HalideHelpers`, the latter of which contains all of our architecture-independent code, such as the header-only libraries (buffer, benchmarks, imageio, etc.) and CMake functions. `Halide` depends on `HalideHelpers`, so existing usages are not affected.
2. The various packaging scripts that were previously in `tools` are now in `packaging/${backend}/package.sh`. This allows us to organize backend-specific scripts and settings without running into name clashes or further polluting `tools`.
3. The packaging scripts use [presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) to manage configuring the builds, which require CMake 3.19+. I believe our buildbots are already up to date. The resulting packages are still compatible with CMake 3.16+.

This will require a small buildbot change: it will just need to call `packaging/{tgz,zip}/package.{sh,bat}` instead of `tools/package-{unix,windows}.{sh,bat}`.


steven-johnson (approved):
LGTM. LMK when we want to update the buildbot.

alexreinking:
Looks like I broke `halide_benchmarks.h` somehow. Investigating...

PR #5977, opened 2021-05-06
Optimize copy_from/for_each_value

dsharletg:
- Use a more idiomatic loop pattern that the compiler might be more likely to understand.
- Sort strides by the strides of the buffer read, not the buffer written
- Handle the scalar special case outside `for_each_value_helper`.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1951,55 +1951,42 @@ private:
     // @{
     template<int N>
     struct for_each_value_task_dim {
-        int extent;
-        int stride[N];
+        std::ptrdiff_t extent;
+        std::ptrdiff_t stride[N];
     };
 
     // Given an array of strides, and a bunch of pointers to pointers
     // (all of different types), advance the pointers using the
     // strides.
     template<typename Ptr, typename... Ptrs>
-    HALIDE_ALWAYS_INLINE static void advance_ptrs(const int *stride, Ptr *ptr, Ptrs... ptrs) {
+    HALIDE_ALWAYS_INLINE static void advance_ptrs(const std::ptrdiff_t *stride, Ptr *ptr, Ptrs... ptrs) {
         (*ptr) += *stride;
         advance_ptrs(stride + 1, ptrs...);
     }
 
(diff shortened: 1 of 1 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -1951,55 +1951,42 @@ class Buffer {
...
+        std::ptrdiff_t extent;
+        std::ptrdiff_t stride[N];
     };
 
     // Given an array of strides, and a bunch of pointers to pointers
     // (all of different types), advance the pointers using the
     // strides.
     template<typename Ptr, typename... Ptrs>
-    HALIDE_ALWAYS_INLINE static void advance_ptrs(const int *stride, Ptr *ptr, Ptrs... ptrs) {
+    HALIDE_ALWAYS_INLINE static void advance_ptrs(const std::ptrdiff_t *stride, Ptr *ptr, Ptrs... ptrs) {
```
steven-johnson, line 1962: Idle thought: would using references rather than pointers here enable potentially better codegen?

steven-johnson:
LGTM. Does profiling show a measurable improvement?

src/runtime/HalideBuffer.h
```diff
@@ -1951,55 +1951,42 @@ class Buffer {
...
-                    increment_ptrs((&ptrs)...);
+                Ptr end = ptr + t[0].extent;
+                while (ptr != end) {
+                    f(*ptr++, (*ptrs++)...);
                 }
             } else {
-                for (int i = t[0].extent; i != 0; i--) {
-                    f((*ptrs)...);
-                    advance_ptrs(t[0].stride, (&ptrs)...);
+                for (std::ptrdiff_t i = t[0].extent; i != 0; i--) {
```
abadams, line 1981: Why not ptr != end in this case too?

abadams:
Makes sense to me. Just had one minor comment.

src/runtime/HalideBuffer.h
dsharletg: I doubt it matters, this is getting inlined. I actually would stylistically prefer references but I didn't want to change it in this PR.

abadams (approved):

PR #5978, opened 2021-05-06
Use a VectorReduce to determine if any lanes are true for Hexagon

abadams:
Implementing this TODO seems to work fine and produces better assembly in the hannk Add generator for hvx.

It codegens to a reduction tree of vector max instructions. Not the best thing it could possibly do, but perfectly reasonable and way better than the status quo.


dsharletg (approved):

steven-johnson (approved):

pranavb-ca:
Thanks @abadams. @aankit-ca - FYI

abadams:
Failing check seems unrelated.

PR #5979, opened 2021-05-06
Add missing #include in buffer_util.h

steven-johnson:


dsharletg (approved):

apps/hannk/util/buffer_util.h
```diff
@@ -6,6 +6,7 @@
 
 #include "HalideBuffer.h"
 #include "HalideRuntime.h"
+#include "interpreter/interval.h"  // for max_rank
```
dsharletg, line 9: Not sure about including a comment like this. Fine to leave it but it's weird (we don't have that comment on any other include...).

steven-johnson: Removed

PR #5980, opened 2021-05-06
Avoid OOB reads on AVX512

steven-johnson:
Some depthwise ops (eg test/mobilenet_v2_1.0_224_quant/001.DEPTHWISE_CONV_2D.tflite) will OOB when compiling with AVX512 enabled (the input here is 33 wide and we try to use a 64-wide register load). Not sure if this is the best fix.

### DIFF
--- a/apps/hannk/halide/depthwise_conv_generator.cpp
+++ b/apps/hannk/halide/depthwise_conv_generator.cpp
@@ -53,6 +53,14 @@ public:
         // Apply the c multiplier.
         Func resampled_input("resampled_input");
         Expr c_resampled = inv_depth_multiplier_ >= 0 ? c * inv_depth_multiplier_ : c / depth_multiplier_;
+        if (inv_depth_multiplier_ > 0) {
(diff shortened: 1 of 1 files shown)


dsharletg:
I think this is going to be really slow. I think we should fix this instead: https://github.com/halide/Halide/blob/master/apps/hannk/interpreter/ops.cpp#L821-L825

I wanted to do that with a bounds query call to the pipeline, but couldn't get that to work.

Alternatively, we probably want to support fewer channels without padding on x86. It's probably better to just use AVX2 in this case. That suggests adding specialization(s) for narrower vectors on x86 (and any target that supports a mix of vector sizes).

steven-johnson:
PTAL

dsharletg:
I think this is going to make things quite a bit slower on non-AVX512 targets. This is why I wanted to try to do this with a bounds query instead of a hardcoded constant.

dsharletg:
I don't know why I struggled with this last time, but the straightforward thing worked: #5984

PR #5981, opened 2021-05-07
advance_ptrs() should use refs, not ptrs

steven-johnson:
Examination of compiled output (x86-64, clang w/ optimizer) shows slightly better codegen.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -1959,8 +1959,8 @@ private:
     // (all of different types), advance the pointers using the
     // strides.
     template<typename Ptr, typename... Ptrs>
-    HALIDE_ALWAYS_INLINE static void advance_ptrs(const std::ptrdiff_t *stride, Ptr *ptr, Ptrs... ptrs) {
-        (*ptr) += *stride;
+    HALIDE_ALWAYS_INLINE static void advance_ptrs(const std::ptrdiff_t *stride, Ptr &ptr, Ptrs... ptrs) {
+        ptr += *stride;
         advance_ptrs(stride + 1, ptrs...);
     }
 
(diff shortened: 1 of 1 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -1959,8 +1959,8 @@ class Buffer {
     // (all of different types), advance the pointers using the
     // strides.
     template<typename Ptr, typename... Ptrs>
-    HALIDE_ALWAYS_INLINE static void advance_ptrs(const std::ptrdiff_t *stride, Ptr *ptr, Ptrs... ptrs) {
-        (*ptr) += *stride;
+    HALIDE_ALWAYS_INLINE static void advance_ptrs(const std::ptrdiff_t *stride, Ptr &ptr, Ptrs... ptrs) {
```
dsharletg, line 1962: Shouldn't this be `Ptrs&...` ? I'm surprised it works as-is?

steven-johnson: It wasn't `Ptrs*...` before, but it worked, because it worked as long as the parameter pack expanded into a type that matched. Same deal here.

dsharletg: Yeah, it makes sense when they're pointers, the deduced type would be a pointer. But why would they be deduced to be references here?

Flip this around: if you wanted these to be passed by value, how would you write this function? Note this is exactly what for_each_value_helper is doing, which *needs* the pointers to be passed by value for it to work.

steven-johnson: I'd write it as `Ptr ptr, Ptrs... ptrs`.

FWIW, if this is getting the wrong answer, none of our tests are able to catch it. 

But I'll change it to `Ptr &ptr, Ptrs&... ptrs` to make it clearer, since that works too.

dsharletg: FWIW, make test in apps/hannk fails with dda046da877caa1d6bdcb7701b7f6b44a2d1877b.

My guess is now the codegen is the same as master after fixing it in 57bc0cb287fdfbb78bba3ebcd0c037427fad5dd0 :)

But I still think this is a good change stylistically, even if it doesn't change anything.

dsharletg (approved):

PR #5983, opened 2021-05-07
Refactor quantized multiplications

dsharletg:
I found it hard to reason about the old QuantizedMulAndShift struct. This PR changes this into a more math oriented reasoning, which makes it easier IMO. For example, you can reason about which directions shifts should go, as opposed to memorizing it.

This also changes some float math into integer arithmetic, which should be more precise (though they were powers of two, so it should have been exact anyways).

Also pull some shared constants into a header, and add some tests.


apps/hannk/halide/constants.h
```diff
@@ -0,0 +1,18 @@
+#ifndef HANNK_CONSTANTS_H
+#define HANNK_CONSTANTS_H
+
+namespace hannk {
+
+// After subtracing the zero point, we have 9 bits. We can shift
```
steven-johnson, line 6: might make sense to move `max_rank` here too

apps/hannk/halide/elementwise_generator.cpp
```diff
@@ -27,15 +28,12 @@ class Add : public Generator<Add> {
     void generate() {
         Var x("x"), y("y");
 
-        // After subtracing the zero point, we have 9 bits. We can shift
-        // up by a further 6 bits to 15 bits total to get more precision
-        // for the later operations.
-        Expr input1 = (i16(input1_(x, y)) - i16(input1_zero_)) << 6;
-        Expr input2 = (i16(input2_(x, y)) - i16(input2_zero_)) << 6;
+        Expr input1 = (i16(input1_(x, y)) - i16(input1_zero_)) << add_input_shift;
```
steven-johnson, line 31: yay, named constants :-)

steven-johnson (approved):
Much nicer!

apps/hannk/halide/constants.h
dsharletg: max_rank isn't used anywhere in the Halide code though?

PR #5984, opened 2021-05-07
Don't hardcode depthwise padding

dsharletg:
Alternative to #5980 to fix OOB reads on AVX512.


steven-johnson (approved):
LGTM, fixes it on my system

PR #5985, opened 2021-05-07
Stack input and filter to reduce generated code in FFT app

dsharletg:
This helped with build times less than I hoped (maybe 10 seconds), but it seems like an OK change anyways. If someone used this code as an example, it would be a good thing to do in practice.

### DIFF
--- a/apps/fft/main.cpp
+++ b/apps/fft/main.cpp
@@ -18,21 +18,7 @@
 using namespace Halide;
 using namespace Halide::Tools;
 
-Var x("x"), y("y");
-
-template<typename T>
-Func make_real(const Buffer<T> &re) {
-    Func ret;
-    ret(x, y) = re(x, y);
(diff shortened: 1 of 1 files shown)


zvookin:
If this is targeting halide compilation time, wouldn't an ImageParam or similar be a better way to go? Also, I think the description could be better if it said something like "Take FFT of 2 channels at once rather than two separate FFTs." instead of talking about stacking the input.

dsharletg:
I'm not sure how ImageParam would help with compilation time here? There is no redundant recompilation happening that would be avoided by making the input an ImageParam.

zvookin:
I misunderstood where the compilations were happening. It makes one pipeline that does the whole thing. I still think the comments can be improved. "Stack" is a highly overloaded word in the context.

zvookin (approved):

PR #5990, opened 2021-05-10
Small HVX fixes

dsharletg:
- Fix align_loads for small unaligned loads
- Strengthen eliminate_interleaves by only trying not to add more interleaves/deinterleaves than we remove, instead of never adding any new interleaves/deinterleaves.
- Enable generating vrmpy for unsigned operands accumulating to a signed accumulator (thanks 2's complement).


steven-johnson (approved):

src/AlignLoads.cpp
```diff
@@ -15,6 +15,10 @@ namespace Internal {
 
 namespace {
 
+int align_up(int x, int n) {
+    return (x + (n - 1)) & ~(n - 1);
```
abadams, line 19: `internal_assert((n & (n - 1)) == 0);`

Or change to `((x + n - 1) / (n - 1)) * (n - 1)`

steven-johnson:

src/HexagonOptimize.cpp
```diff
@@ -1362,21 +1379,34 @@ class EliminateInterleaves : public IRMutator {
...
         }
 
+        if (const Load *load = x.as<Load>()) {
+            if (buffers.contains(load->name)) {
+                return buffers.get(load->name) != BufferState::NotInterleaved;
+            }
+        }
+
+        if (const Add *op = x.as<Add>()) {
+            return yields_interleave(op->a) || yields_interleave(op->b);
```
pranavb-ca, line 1389: Shouldn't an add yield an interleave only if both operands do too?

dsharletg: The idea with this change is that as long as one of the operands yields an interleave, we can deinterleave it at ~zero cost, because we're adding one op while removing another. In the case I looked at, it was *highly* profitable to make this trade. In depthwise convolution, it enables the entire convolution to be performed deinterleaved in exchange for only a single extra deinterleave almost at compute_root.

steven-johnson:
Windows failure is irrelevant, bots clean

PR #5993, opened 2021-05-10
Various simplifier improvements

### LINKED ISSUE #5905: Target::LargeBuffers breaks load alignment annotations
When using `Target::LargeBuffers`, loads stop getting annotated with alignment information. For example, building the camera pipe:

Without `-large_buffers`
```
          deinterleaved[ramp((deinterleaved.s0.v0.v0*32) + t3773, 1, 32)] = f1[ramp(((deinterleaved.s0.v0.v0*64) + t3766) + 24, 2, 32) aligned(8, 0)]
          deinterleaved[ramp(((deinterleaved.s0.v0.v0*32) + t3768) + 76, 1, 32)] = f1[ramp(((deinterleaved.s0.v0.v0*64) + t3766) + 25, 2, 32) aligned(8, 1)]
```

With `-large_buffers`
```
          deinterleaved[ramp(t3771 + int64(((deinterleaved.s0.v0.v0*32) + 2)), (int64)1, 32)] = f1[ramp(t3770 + int64((((deinterleaved.s0.v0.v0*64) - t3741) + 28)), (int64)2, 32)]
          deinterleaved[ramp(t3773 + int64(((deinterleaved.s0.v0.v0*32) + 2)), (int64)1, 32)] = f1[ramp(t3772 + int64((((deinterleaved.s0.v0.v0*64) - t3741) + 28)), (int64)2, 32)]
```

dsharletg:
- Redo hoisting loop invariant if statements at the end of lowering
- Track bounds through casts to fix #5905 
- Replace loads completely out of bounds with undef and remove stores completely out of bounds. These can happen inside branches with complicated expressions that don't get simplified, so they are unreachable. However, I've given up on trying to simplify the conditions after weeks of on/off looking at these. Some of the new rules in this PR help get closer, but still not there.


src/Simplify_LT.cpp
```diff
@@ -389,18 +389,14 @@ Expr Simplify::visit(const LT *op, ExprInfo *bounds) {
...
+              rewrite(x/c0 < (x + c2)/c0, x%c0 < c2, c0 > 0) ||
               // c2 == 0
-              rewrite((x + c1)/c0 < x/c0, false, c0 > 0 && c1 >= 0) ||
-              rewrite((x + c1)/c0 < x/c0, true, c0 > 0 && c1 <= 0 - c0) ||
+              rewrite((x + c1)/c0 < x/c0, x%c0 < -c1, c0 > 0) ||
 
               // The addition on the right could be outside
-              rewrite((x + c1)/c0 < x/c0 + c2, false, c0 > 0 && c1 >= c2 * c0) ||
-              rewrite((x + c1)/c0 < x/c0 + c2, true, c0 > 0 && c1 <= c2 * c0 - c0) ||
+              rewrite((x + c1)/c0 < x/c0 + c2, x%c0 < c2 * c0 - c1, c0 > 0) ||
```
abadams, line 399: Synthesized version of these four added an extra one and made some constants non-constant:
```
 rewrite(((x + c0)/c1) < ((x/c1) + c2), (x % c1) < fold((c2*c1) - c0), c1 > 0) ||
 rewrite((x/c0) < ((x + y)/c0), (x % c0) < y, c0 > 0) ||
 rewrite((x/c0) < ((y + x)/c0), (x % c0) < y, c0 > 0) ||
 rewrite(((x + c0)/c1) < (x/c1), (x % c1) < fold(0 - c0), c1 > 0) ||
 rewrite(((x + c0)/c1) < ((x + c2)/c1), (x % c1) < fold(c2 - c0), c1 > 0)

```

abadams: (I think that assumes they're not zero, because that would have constant-folded)

dsharletg:
This PR has a fuzz simplifier failure:
```
Simplify fuzz test seed: 1620756761
a = (int8)-26
b = int8((int0)d)
c = (int8)113
d = (int8)84
e = (int8)52
int8x2((uint8x2)absd(((x2(int8((int0)c))/x2((int8)-71)) % (ramp(int8((int0)e), int8((int0)c), 2) % ramp(int8((int0)e), int8((int0)b), 2))) % ((ramp(int8((int0)b), int8((int0)b), 2)*int8x2(ramp((int16)13966, (int16)7221, 2))) - ((x2(int8((int0)b)) % x2(int8((int0)e))) + (ramp(int8((int0)e), int8((int0)a), 2) % ramp(int8((int0)b), int8((int0)c), 2)))), ramp(select((int8)55 == (max(int8((int0)a), (int8)107) + (int8)81), select((int8)117 == int8((int0)a), int8((int0)b), int8((int0)e)), int8((int0)c)), (int8)120, 2)))
int8x2((uint8x2)absd((x2(int8((int0)c)/(int8)-71) % (ramp(int8((int0)e), int8((int0)c), 2) % ramp(int8((int0)e), int8((int0)b), 2))) % ((ramp(int8((int0)b), int8((int0)b), 2)*int8x2(ramp((int16)13966, (int16)7221, 2))) - ((ramp(int8((int0)e), int8((int0)a), 2) % ramp(int8((int0)b), int8((int0)c), 2)) + x2(int8((int0)b) % int8((int0)e)))), ramp(select(max(int8((int0)a), (int8)107) == (int8)-26, select(int8((int0)a) == (int8)117, int8((int0)b), int8((int0)e)), int8((int0)c)), (int8)120, 2)))
In vector lane 0:
int8((uint8)absd((int8)0, select(max(int8((int0)a), (int8)107) == (int8)-26, select(int8((int0)a) == (int8)117, int8((int0)b), int8((int0)e)), int8((int0)c)))) -> (int8)113
int8((uint8)absd((int8)0, select(int8((int0)a) == (int8)-26, select(int8((int0)a) == (int8)117, int8((int0)b), int8((int0)e)), int8((int0)c)))) -> (int8)52
```

However, this failure also occurs on master! So it is just a crazy coincidence that testing this PR uncovered it.

src/Simplify_Stmts.cpp
```diff
@@ -121,6 +127,13 @@ Stmt Simplify::visit(const IfThenElse *op) {
                equal(unwrapped_condition, 0 < then_for->extent)) {
         // This guard is redundant
         return then_case;
+    } else if (then_if &&
```
abadams, line 130: This is a sort of distributive law for ifs? Would maybe help to add a comment:
```
Before:
if (a) {
  if (b) X
} else {
  if (b) Y
}

After:
if (b) {
  if (a) X
  else Y
}
```

I think it requires the conditions a, b to be pure. If they have side-effects then you're changing which ones get evaluated when.

src/Simplify.cpp
```diff
@@ -247,6 +251,27 @@ void Simplify::ScopedFact::learn_true(const Expr &fact) {
             }
             simplify->bounds_and_alignment_info.push(v->name, expr_info);
             bounds_pop_list.push_back(v);
+        } else if ((div && (v = div->a.as<Variable>())) ||
+                   (div_mul && (v = div_mul->a.as<Variable>()))) {
+            // Learn from expressions of the form (x / a) * b == c
+            // We might know the bounds of the variable.
+            Interval range = solve_for_inner_interval(eq, v->name);
```
abadams, line 258: solve_for_inner_interval is pretty heavy-duty, and might get more expensive in future. Can it be done directly for this form?

dsharletg: It can, but it's a fairly tricky bit of logic. Totally doable, I just didn't want to duplicate it. The other thing I thought about doing was only entering this block if a, b and c are constants. In that case, solve_for_inner_interval should be pretty cheap right?

abadams: I think this is true (z3 can't find counterexamples at least):

```(x / c0) * c1 == c2```
is equivalent to:
``` x >= c0 * (c2/c1) && x < c0 * (c2 / c1 + 1) ```
when
```c0 > 0 && c1 > 0 && c2 % c1 == 0```

**Comment by steven-johnson** [comment]

File: src/Simplify.cpp (hunk 3 above)
Comment (dsharletg) [reply src/Simplify.cpp:304]: I found more small simple equations that needed support. Let me know what you think of the current thing. If we don't want to use solve here, I suspect we either need to add a "solve_constant" helper, or prevent situations like this in the first place by fixing specialize, but I've spent quite a bit of effort on that approach without success.

**Comment by dsharletg** [comment]
All of the recent buildbot failures are cancellations, I'm pretty sure that this has passed every build bot at one point or another.

File: src/IROperator.cpp [hunk 4]
```diff
@@ -2588,6 +2588,14 @@ Expr undef(Type t) {
                                 Internal::Call::PureIntrinsic);
 }
 
+namespace Internal {
+Expr unreachable(Type t) {
+    return Internal::Call::make(t, Internal::Call::unreachable,
+                                std::vector<Expr>(),
+                                Internal::Call::PureIntrinsic);
```
abadams, line 2595: unreachable is not pure, because it can't be safely moved

abadams (approved):
lgtm except that unreachable should be Intrinsic, not PureIntrinsic

src/Simplify_Cast.cpp
```diff
@@ -4,10 +4,16 @@ namespace Halide {
...
 
 Expr Simplify::visit(const Cast *op, ExprInfo *bounds) {
-    // We generally don't track bounds through casts, with the
-    // exception of casts that constant-fold to a signed integer, so
-    // we don't need the bounds of the value.
-    Expr value = mutate(op->value, nullptr);
+    Expr value = mutate(op->value, bounds);
+
+    if (bounds) {
+        if (bounds->min_defined && !op->type.can_represent(bounds->min)) {
```
rootjalex, line 10: if the type can't represent the min, then both the min and the max should be undefined, not just the min

src/Simplify_Cast.cpp
```diff
@@ -4,10 +4,16 @@ namespace Halide {
...
+            bounds->min_defined = false;
+        }
+        if (bounds->max_defined && !op->type.can_represent(bounds->max)) {
```
rootjalex, line 13: Same here for max

rootjalex:
I think that bounds should only be tracked through casts if the min and max are defined and both are representable in the new type, otherwise the resultant bounds don't take into account possible overflow

src/Simplify_Add.cpp
```diff
@@ -188,6 +188,9 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
                rewrite(x + ((c0 - x)/c1 + y)*c1, y * c1 - ((c0 - x) % c1) + c0, c1 > 0) ||
                rewrite(x + (y + (c0 - x)/c1)*c1, y * c1 - ((c0 - x) % c1) + c0, c1 > 0) ||
 
+               rewrite((x/c0) + (y*(x/c1)), (y*(x/c1))*fold((c1/c0) + 1), (c1 % c0) == 0) ||
```
abadams, line 191: These ones fail to verify:
Incorrect rule: rewrite(((x/c0) + (y*(x/c1))), ((y*(x/c1))*((c1/c0) + 1)), ((c1 % c0) == 0))
Counterexample is: c0 = -6908
c1 = 6908
x = -6908
y = 2
For counterexample, LHS = -1 RHS = 0
Incorrect rule: rewrite(((x/c0) + ((x/c1)*y)), (((x/c1)*y)*((c1/c0) + 1)), ((c1 % c0) == 0))
Counterexample is: c0 = -6908
c1 = 6908
x = -6908
y = 2
For counterexample, LHS = -1 RHS = 0

abadams: Doesn't seem to be related to rounding:
```
Incorrect rule: rewrite(((x/c0) + ((x/c1)*y)), (((x/c1)*y)*((c1/c0) + 1)), (((c1 % c0) == 0) && ((c0 > 0) && (c1 > 0))))
Counterexample is: c0 = 2
c1 = 4
x = 7
y = 1
For counterexample, LHS = 4 RHS = 3

Incorrect rule: rewrite(((x/c0) + (y*(x/c1))), ((y*(x/c1))*((c1/c0) + 1)), (((c1 % c0) == 0) && ((c0 > 0) && (c1 > 0))))
Counterexample is: c0 = 2
c1 = 4
x = 7
y = 1
For counterexample, LHS = 4 RHS = 3
```

src/Simplify_Call.cpp
```diff
@@ -612,14 +615,31 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
...
             return mutate(op->args[2], bounds);
         } else {
             Expr true_value = mutate(op->args[1], nullptr);
+            bool true_unreachable = in_unreachable;
+            in_unreachable = false;
             Expr false_value = mutate(op->args[2], nullptr);
+            bool false_unreachable = in_unreachable;
+
+            if (true_unreachable && false_unreachable) {
+                // just let the unreachable flag fall through.
```
abadams, line 634: I don't understand this fall-though. Doesn't in_unreachable get set to false?

PR #5995, opened 2021-05-11
Initial work to add Tensor Cores support in Halide

mcleary:
Tensor Cores are programmable cores that perform warp-level
matrix-multiply and accumulate operations. It can greatly improve
performance of matrix-multiply expressions written in Halide when
the hardware is available.

This works by pattern-matching an expression of the form

```c++
RDom k(0, matrix_size);
C(x, y) += f32(A(k, y)) * f32(B(x, k));
```

and generating code that will realize the function using the GPU.

The current support is quite limited. The only data types currently
support are float16_t for A and B and float for C. The dimentions of
the input matrices must be a multiple of 16 since the only shape
currently supported is m16n16k16. Also note that the generated
PTX code is not the most efficient way to use tensor cores but
it gives good results with a simple schedule.

The following table shows the performance comparison of the modified
cuda_mat_mul app for a matrix size of 2048

| Configuration | Time (ms) | GFlops  |
|----------------------------------|-----------|---------|
| CUDA 5.0                             | 6.01      | 2854.78  |
| CUDA 7.0 (Tensor Cores)     | 5.20      | 3300.91  |

The performance difference is not massive since the current code is not using shared memory

CPU: Intel Core i7-8700
GPU: NVIDIA RTX2060

### DIFF
--- a/apps/cuda_mat_mul/mat_mul_generator.cpp
+++ b/apps/cuda_mat_mul/mat_mul_generator.cpp
@@ -29,7 +29,7 @@ public:
 
         Func prod("prod");
         RDom r(0, size);
-        prod(x, y) += f32(A(r, y)) * f32(B(x, r));
+        prod(x, y) += f32(A(x, r)) * f32(B(r, y));
 
         Var xi, yi, xio, xii, yii, xo, yo, x_pair, xiio, ty;
         RVar rxo, rxi;


src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +809,256 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+                                const Variable *load_a_var_1 = matches_a[0].as<Variable>();
+                                const Variable *load_a_var_2 = matches_a[1].as<Variable>();
+                                const Variable *load_b_var_1 = matches_b[0].as<Variable>();
+                                const Variable *load_b_var_2 = matches_b[1].as<Variable>();
+
+                                // TODO: Can this checks be improved?
+                                const bool load_a_ok = (load_a_var_1 && load_a_var_1->name == k_var_name) || (load_a_var_2 && load_a_var_2->name == k_var_name);
+                                const bool load_b_ok = (load_b_var_1 && load_b_var_1->name == k_var_name) || (load_b_var_2 && load_b_var_2->name == k_var_name);
+
+                                if (load_a_ok && load_b_ok) {
```
rootjalex, line 877: This could be triggered by the pattern A(x, k) * B(x, k), right? There doesn't seem to be a check that the reduction var is used in different dimensions for A and B

test/performance/matrix_multiplication.cpp
```diff
@@ -94,12 +105,12 @@ int main(int argc, char **argv) {
     Buffer<float> output_halide(matrix_size, matrix_size);
 
     simple_version(mat_A.data(), mat_B.data(), output_ref.data(), mat_A.width(), mat_A.stride(1));
-    out.realize(output_halide);
+    output_halide = out.realize(matrix_size, matrix_size);
```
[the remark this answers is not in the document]
mcleary: Thanks for the tip. Fixed in 40026d3eac9483e719544ffa9bfafb892c010ff3

src/CodeGen_PTX_Dev.cpp
mcleary: I accidently marked it as resolved. Sorry, I will check this expression now

mcleary: @rootjalex This particular expression you mentioned won't be matched the first value of `matches_a` won't be a variable.

rootjalex: Hmm but the patterns don't use `k_var_expr`, so I'm not confident about that - but I guess the bug might be in the pattern definitions.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +809,256 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+                    const Expr mul_pattern = wild_f16x * wild_f16x;
+                    if (expr_match(mul_pattern, cast->value, matches)) {
+                        // A[((A.stride.1*t34) - t33) + matmul$1.s1.k$x]
+                        const Load *load_a = matches[0].as<Load>();
+                        // B[(B.stride.1 * matmul$1.s1.k$x) + t35]
+                        const Load *load_b = matches[1].as<Load>();
+
+                        if (load_a && load_b) {
+                            const Expr wild_i32x = Variable::make(Int(32), "*");
+                            const Expr k_var_expr = Variable::make(Int(32), k_var_name);
```
rootjalex, line 858: This appears to be unused, I think it should be part of the patterns?

src/CodeGen_PTX_Dev.cpp
mcleary: When I first wrote this, I had a pattern like this

```c++
const Expr load_a_pattern = wild_i32x + Variable::make(k_var_name, Int(32));
const Expr load_b_pattern = wild_i32x * Variable::make(k_var_name, Int(32)) + wild_i32x;
```

But then I found out that `expr_match` expects the `k_var` to be on the right hand side of the `Add` and `Mul` respectively. This is not always the case, so I replaced it with a wild card expression and I check for the `k_var_name` on lines 874 and 875

src/CodeGen_PTX_Dev.cpp
mcleary: Ah, I left it by mistake. Please see my comment https://github.com/halide/Halide/pull/5995#discussion_r631082738

mcleary: Removed in dae5a5812b42c1e573e439ed1266b9a156d232d6

src/CodeGen_PTX_Dev.cpp
rootjalex: Ah gotcha, thanks for the explanation. I incorrectly assumed that the matcher would catch that case, my apologies.

mcleary: If you have any suggestion to improve this match please let me know

steven-johnson:

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +809,255 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+            if (matmul_load && cast) {
+                // Check if the load is loading from the same place where the store is storing
+                // i.e. if this is an update operation
+                if (matmul_load->name == store->name && equal(matmul_load->index, store->index)) {
+                    // Yes, this is an update operation, now let's check cast to see if it is a matmul that can be converted
+                    // to wmma intrinsic
+
+                    // float32((A[((A.stride.1*t34) - t33) + matmul$1.s1.k$x]*B[(B.stride.1*matmul$1.s1.k$x) + t35]))
+                    const Expr wild_f16x = Variable::make(Float(16), "*");
+                    const Expr mul_pattern = wild_f16x * wild_f16x;
```
dsharletg, line 849: Why not match the add and mul at the same time? It seems like this would simplify this code a fair bit:

    pattern = wild_f32x + cast<float>(wild_f16x) * wild_f16x);

Also, are you sure this is right that the intermediate multiply result is f16 and not f32? i.e. `f32(wild_f16 * wild_f16x)` vs. `f32(wild_f16) * f32(wild_f16)`

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +809,255 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+            // Shape m16n16k16
+            // TODO: Note this won't work unless the extent of the inner loop is constant
+            //       How can the shape be detected if the k is not constant, i.e, if k
+            //       is based on the dimentions of matrix A for example?
+            if (wmma_M == 16 && wmma_N == 16 && wmma_K % 16 == 0) {
+                const int32_t num_tiles_k = wmma_K / 16;
+                wmma_K = 16;
+
+                // Now check the loop body to confirm this is a matrix multiply expression
+                IsMatrixMultiply is_matrix_multiply{loop->name};
```
dsharletg, line 932: I think maybe this should just be a function that digs through the body rather than a visitor? To make sure that this doesn't accidentally find a more deeply nested matrix multiply store? Or, what if the loop contained two separate matrix multiplications?

test/performance/matrix_multiplication.cpp
```diff
@@ -4,15 +4,16 @@
...
-void simple_version(float *A, float *B, float *C, int width, int stride) {
+void simple_version(float16_t *A, float16_t *B, float *C, int width, int stride) {
     for (int iy = 0; iy < width; iy++) {
         for (int ix = 0; ix < width; ix++) {
             float *cc = C + iy * stride + ix;
             *cc = 0.0f;
 
             for (int ik = 0; ik < width; ik++) {
-                *cc = *cc + A[iy * stride + ik] * B[ik * stride + ix];
+                *cc = *cc + static_cast<float>(A[iy * stride + ik] * B[ik * stride + ix]);
```
dsharletg, line 16: Same question regarding the intermediate product above for this. Maybe the test tolerance wouldn't need to be bumped up if this were fixed?

mcleary: You're right, for some reason I thought that only the accumulator was f32. I reverted the change in the tolerance in 152f19d0275cde46f320c8c489a4066984c3222d.

mcleary: Just realised this will require more changes

steven-johnson:
Looks like there are performance regressions on some of the buildbots with this PR in place.

mcleary:
> Looks like there are performance regressions on some of the buildbots with this PR in place.

I made a mistake in my previous commit. Its fixed in 7f9719c749625df8a27ca258fa455d982cd6ea31

src/CodeGen_PTX_Dev.cpp
mcleary: I simplified the matching logic in 7f9719c749625df8a27ca258fa455d982cd6ea31 based on your suggestion, for some reason I was sure that the multiplication was done in lower precision, but that's not the case.

src/CodeGen_PTX_Dev.cpp
mcleary: This makes sense and the visitor is now converted into a a function in d2581cb46fb6e35c40be1d9aa640cdee436682a3

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +810,325 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+
+        if (store) {
+            //  matmul$1[t26] = (float32)matmul$1[t26]
+            //      + (float32((float16)A[((A.stride.1*t34) - t33) + matmul$1.s1.k$x])
+            //      * *float32((float16)B[(B.stride.1*matmul$1.s1.k$x) + t35]))
+            const Expr wild_f32x = Variable::make(Float(32), "*");
+            const Expr wild_f16x = Variable::make(Float(16), "*");
+            const Expr acc_pattern = wild_f32x + f32(wild_f16x) * f32(wild_f16x);
+            vector<Expr> matches;
+            if (expr_match(acc_pattern, store->value, matches)) {
```
dsharletg, line 842: I think this whole set of ifs would be easier to read/understand if they were expressed as early returns:

    if (!expr_match(...)) {
        return MatrixMultiplyInfo();   
    }

    if (!load_c || !load_a || !load_b) {
        return MatrixMultiplyInfo();   
    }
...

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +810,325 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+                        const Expr wild_i32x = Variable::make(Int(32), "*");
+                        // Check if the reduction domain is being used in both A and B to
+                        // validate the matrix multiply
+                        const Expr load_a_pattern = wild_i32x + wild_i32x;
+                        const Expr load_b_pattern = wild_i32x * wild_i32x + wild_i32x;
+
+                        vector<Expr> matches_a, matches_b;
+                        const bool match_a = expr_match(load_a_pattern, load_a->index, matches_a);
+                        const bool match_b = expr_match(load_b_pattern, load_b->index, matches_b);
+                        if (match_a && match_b) {
```
dsharletg, line 865: Should this also try commuting load_a_pattern with load_b_pattern? It seems OK to have them in either order.

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +809,321 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+                        Expr var_c = Variable::make(Handle(), load_c->name);
+
+                        // Calculates the global warp indices used to iterate over the k
+                        // tiles of the matrices
+                        Expr warp_x = (block_id_x * block_dim_x + thread_id_x) / warp_size;
+                        Expr warp_y = block_id_y * block_dim_y + thread_id_y;
+
+                        Expr offset_c_value = i32(global_N * wmma_M * warp_y + wmma_N * warp_x);
+
+#define INLINE_TILE_LOOP 0
```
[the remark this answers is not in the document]
mcleary: Yes definitely. I’m running some benchmarks to see if one has any advantages over the other, but I have a preference for the Halide IR approach as it generates more readable IR, it is easier to modify and doesn’t loose on performance when the loop is unrolled.

I will update the pull request description with my tests.

src/CodeGen_PTX_Dev.cpp
mcleary: Done in 2c83da9ccfbbe54372a6a8b300c54f3d5994aa69

src/CodeGen_PTX_Dev.cpp
```diff
@@ -710,10 +718,20 @@ vector<char> CodeGen_PTX_Dev::compile_to_src() {
     }
     debug(2) << "Done with CodeGen_PTX_Dev::compile_to_src";
 
-    debug(1) << "PTX kernel:\n"
-             << outstr.c_str() << "\n";
+#define DEBUG_PTX 0
+#if DEBUG_PTX
+    // Adding this in the PTX allows the use of Nsight to debug the PTX
+    // TODO: Could this an official Halide feature?
```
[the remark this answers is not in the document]
mcleary: For some reason I didn't consider the `-debug` flag. I've changed it in eb1144fe438bfb415ca6eafd295a4cd3a071e0d3

src/CodeGen_PTX_Dev.cpp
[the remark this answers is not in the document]
mcleary: I changed the code to generate the loop using Halide IR in eb1144fe438bfb415ca6eafd295a4cd3a071e0d3. I've updated the PR description with results from the cuda_mat_mul app

src/CodeGen_PTX_Dev.cpp
mcleary: That is a fair point. I improved the code in c031f495b4af8870aafc97c1137ade51c61ff4d0

apps/cuda_mat_mul/mat_mul_generator.cpp
```diff
@@ -28,30 +29,39 @@ class MatMul : public Halide::Generator<MatMul> {
 
         Func prod("prod");
         RDom r(0, size);
```
dsharletg, line 31: We should do something interesting here, like use `RDom r(A.dim(0).min(), A.dim(0).extent())`

In the schedule, do something interesting like: `RVar ro; prod.update().split(r, ro, r, 64).unroll(r, 16);`

Hopefully that would give an unrolled loop of 4 wmma instructions, and also include a non-wmma tail loop to handle the remainder of the reduction.

dsharletg: Edited that because I realized the proposed schedule was wrong

mcleary: I wasn't sure how to proceed with this but I will try this schedule

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +813,334 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+    }
+
+    //  matmul$1[t26] = (float32)matmul$1[t26]
+    //      + (float32((float16)A[((A.stride.1*t34) - t33) + matmul$1.s1.k$x])
+    //      * *float32((float16)B[(B.stride.1*matmul$1.s1.k$x) + t35]))
+    const Expr wild_f32x = Variable::make(Float(32), "*");
+    const Expr wild_f16x = Variable::make(Float(16), "*");
+    const Expr acc_pattern = wild_f32x + f32(wild_f16x) * f32(wild_f16x);
+    vector<Expr> matches;
+    if (expr_match(acc_pattern, store->value, matches)) {
```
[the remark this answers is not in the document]
mcleary: Changed in 35f8b556e79e3cb4991db5e804affab26f2d4f5f =)

src/CodeGen_PTX_Dev.cpp
```diff
@@ -791,6 +813,334 @@ bool CodeGen_PTX_Dev::supports_atomic_add(const Type &t) const {
...
+            if (ends_with(loop->name, ".__thread_id_y")) {
+                wmma_M = loop_extent_value;
+            } else if (ends_with(loop->name, ".__thread_id_x")) {
+                wmma_N = loop_extent_value;
+            }
+        } else {
+            wmma_K = loop_extent_value;
+
+            // Shape m16n16k16
+            // TODO: Note this won't work unless the extent of the inner loop is constant
```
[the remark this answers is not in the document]
mcleary: I think this is reasonable but it will require more changes in the way that the wmma ops are extracted, i.e. splitting K will generate the wrong kernel with this PR. This is of course something I want to address before merging so I will come back to this soon.

steven-johnson:
Looks like we're getting real failures in some tests.

mmoadeli:
Are there any other issues to address to have this PR merged?

abadams:
It would be good to have some more test coverage for this in test/correctness to at least make sure the intrinsics are actually being generated (e.g. in the style of simd_op_check).

Also, I'm a little concerned that this seems to require a root-level loop that's *just* a matrix multiply. Is there a path to being able to schedule tensor-core-using ops at the blocks level within a larger kernel? E.g. how do I do a matrix-multiply fused with a relu in a single kernel? A tensor-core-using-stage should be able to be schedule compute_at gpu blocks of some other Func.

Finally this seems to be based on pattern matching the source, and if the code changes slightly it'll fall off a performance cliff with no error, which is generally not how Halide does things. I think it should be explicitly scheduled. Here was my original proposal for how this should look from the scheduling language:
https://github.com/halide/Halide/issues/4481

steven-johnson:
Where does this PR stand? Is it going to get more attention?

mmoadeli:
> Where does this PR stand? Is it going to get more attention?

Our plan is to address the issues in this PR and the AMX support PR in September.

steven-johnson:
Any update on this PR?

steven-johnson:

frengels:
sorry about the late reply, yes it's still active

jinderek:
Hi guys, any updates?

frengels:
The work on this PR has been paused for now as it's not been a priority lately, I'm not sure when the work would be resumed.

jinderek:
@frengels I made a test with this PR, the Halide IR seems to be correct. But down to the LLVM IR,  the LLVM seems to opt to a void function:
```

    // .globl   _kernel_matrix_mul_s1_y_blockY___block_id_y // -- Begin function _kernel_matrix_mul_s1_y_blockY___block_id_y
                                        // @_kernel_matrix_mul_s1_y_blockY___block_id_y
.visible .entry _kernel_matrix_mul_s1_y_blockY___block_id_y(
    .param .u64 _kernel_matrix_mul_s1_y_blockY___block_id_y_param_0,
    .param .u64 _kernel_matrix_mul_s1_y_blockY___block_id_y_param_1,
    .param .u64 _kernel_matrix_mul_s1_y_blockY___block_id_y_param_2,
    .param .u32 _kernel_matrix_mul_s1_y_blockY___block_id_y_param_3,
    .param .u32 _kernel_matrix_mul_s1_y_blockY___block_id_y_param_4,
    .param .u32 _kernel_matrix_mul_s1_y_blockY___block_id_y_param_5
)
{


// %bb.0:                               // %entry
    ret;
                                        // -- End function
}
```
Any ideas on this? I'm using LLVM 11.1. And I checked the LLVM passes, it's InstCombine Pass that optimizes the function to void.
Thanks.

PR #6000, opened 2021-05-12
Add -fpic option to debug version on non-windows .ll file

sksarda:
Else, it eliminates reference to global offset table
for symbols to be resolved from remote library causing runtime crash
with -debug option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1047,7 +1047,7 @@ $(BUILD_DIR)/initmod.windows_%_64_debug.ll: $(SRC_DIR)/runtime/windows_%.cpp $(B
 
 $(BUILD_DIR)/initmod.%_64_debug.ll: $(SRC_DIR)/runtime/%.cpp $(BUILD_DIR)/clang_ok
 	@mkdir -p $(@D)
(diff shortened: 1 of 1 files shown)


dsharletg:
Thanks for the PR.

This should probably also change the Makefile (we have redundant builds tooling :( )

steven-johnson:
> This should probably also change the Makefile (we have redundant builds tooling :( )

I think you mean the CMake file :-)

alexreinking:
> This should probably also change the Makefile (we have redundant builds tooling :( )

IIRC the CMake is already correct, but the Makefile was bugged. OP should confirm, though.

dsharletg:
It looks like we need to omit this on windows: https://github.com/halide/Halide/blob/75a079a853e98a7d1d7e2e685fb1242089a41460/src/runtime/CMakeLists.txt#L175-L176

Maybe since this is the Makefile, we shouldn't bother? I assume it's not possible to use the Makefile for that target?

steven-johnson:
> Maybe since this is the Makefile, we shouldn't bother? I assume it's not possible to use the Makefile for that target?

Correct, we don't support Make on Windows.

alexreinking:
> > Maybe since this is the Makefile, we shouldn't bother? I assume it's not possible to use the Makefile for that target?
> 
> Correct, we don't support Make on Windows.

But a Makefile-build Halide should still produce Windows binaries correctly, and that's what this PR seems to be about.

steven-johnson (approved):
Thanks for the fix!

dsharletg:
I think Alex's comment was good, and we probably don't cover the condition Alex mentioned on the buildbots...

PR #6001, opened 2021-05-12
Simplify the new generated expression.

sksarda:
This will simplify the newly generated left
shift expression in case both the operands are constant.

### DIFF
--- a/src/HexagonOptimize.cpp
+++ b/src/HexagonOptimize.cpp
@@ -134,7 +134,7 @@ Expr as_mul(const Expr &a) {
     } else if (const Call *s = Call::as_intrinsic(a, {Call::shift_left, Call::widening_shift_left})) {
         const uint64_t *log2_b = as_const_uint(s->args[1]);
         if (log2_b) {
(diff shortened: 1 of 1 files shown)


src/HexagonOptimize.cpp
```diff
@@ -134,7 +134,7 @@ Expr as_mul(const Expr &a) {
     } else if (const Call *s = Call::as_intrinsic(a, {Call::shift_left, Call::widening_shift_left})) {
         const uint64_t *log2_b = as_const_uint(s->args[1]);
         if (log2_b) {
-            Expr b = make_one(s->type) << cast(UInt(s->type.bits()), (int)*log2_b);
+            Expr b = simplify(make_one(s->type) << cast(UInt(s->type.bits()), (int)*log2_b));
```
dsharletg, line 137: Hmm, why doesn't the simplify call below do this?

sksarda: My bad. You are right, simplify below does what is required. My internal code somehow has slightly different code which missed the simplify. 

I will take back this PR.

PR #6006, opened 2021-05-13
Fix CodeGen_C::print_scalarized_expr

vksnk:
* CppVector/NativeVector object doesn't have .replace() anymore.
* Initialize vector with zero to avoid warning.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -2304,7 +2304,7 @@ string CodeGen_C::print_scalarized_expr(const Expr &e) {
     Type t = e.type();
     internal_assert(t.is_vector());
     string v = unique_name('_');
-    stream << get_indent() << print_type(t, AppendSpace) << v << " = " << print_type(t) + "_ops::broadcast(0);\n";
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):

src/CodeGen_C.cpp
```diff
@@ -2304,12 +2304,12 @@ string CodeGen_C::print_scalarized_expr(const Expr &e) {
     Type t = e.type();
     internal_assert(t.is_vector());
     string v = unique_name('_');
-    stream << get_indent() << print_type(t, AppendSpace) << v << ";\n";
+    stream << get_indent() << print_type(t, AppendSpace) << v << " = " << print_type(t) + "_ops::broadcast(0);\n";
```
dsharletg, line 2307: I wonder if we shouldn't do this, and instead find a way to suppress the warning or just ignore it. It seems like it could hurt performance in some cases?

vksnk: Reverted it back.

steven-johnson:
Ready to land?

PR #6012, opened 2021-05-14
Fix bounds information in ExprInfo for overflow in simplifier

### LINKED ISSUE #6010: Fuzz bounds failure
```
bounds inference fuzz test seed: 1620971360
can't prove upper bound: (1931776 <= -692337496)
a = 113
b = 88
c = 112
d = -14
e = 22
select(x2(ramp(a, e, 3)) > (min(ramp(c, a, 6), x3(x2(a))) % x6(e)), (x2(ramp(d, d, 3))*ramp(ramp(b, 58, 2), ramp(c, 4, 2), 3))*(x3(ramp(c, b, 2))*x3(x2(d))), select(ramp(x3(e), ramp(b, e, 3), 2) <= (x3(ramp(b, e, 2)) % x2(ramp(e, e, 3))), ramp(b, d, 6), ramp(ramp(e, -20, 2), ramp(d, 29, 2), 3) + ramp(x2(9), x2(e), 3)))
[-855300600, -692337496]
In vector lane 0:
select((min(a, c) % e) < a, (b*d)*(c*d), select((b % e) < e, e + 9, b)) -> 1931776
scope {
	a : [113, 117]
	b : [-99, 125]
	c : [111, 113]
	d : [-110, 2]
	e : [22, 102]
}
```

Found on #5990, but I'm pretty sure it's not relevant to that PR.

rootjalex:
Sometimes the simplifier maintains bounds for expressions that are `signed_integer_overflow`. This allows for expressions like:
```
(let t0 = <expression that overflows) in <some expression with t0>)
```
to simplify into values that are not `signed_integer_overflow`. Obviously, that's bad, so this PR prevents bounds information from being tracked for `signed_integer_overflow`.

Edit: another (different, but similar) bug in the simplifier was discovered via #6025, and a fix was added

Fixes #6010 (where it was discovered) and #6025


src/Simplify_Internal.h
```diff
@@ -103,6 +103,14 @@ class Simplify : public VariadicVisitor<Simplify, Expr, Stmt> {
         }
     };
 
+    HALIDE_ALWAYS_INLINE
```
abadams, line 106: I don't think preserving the modulus/remainder is a good idea. I think the body of the if could be `*b = ExprInfo{}`

src/Simplify_Max.cpp
```diff
@@ -54,13 +54,20 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
         int lanes = op->type.lanes();
         auto rewrite = IRMatcher::rewriter(IRMatcher::max(a, b), op->type);
 
+        if (rewrite(max(x, x), x) ||
```
abadams, line 57: I don't think max(x, x) can create a new integer overflow that didn't already exist

src/Simplify_Min.cpp
```diff
@@ -54,13 +54,20 @@ Expr Simplify::visit(const Min *op, ExprInfo *bounds) {
         int lanes = op->type.lanes();
         auto rewrite = IRMatcher::rewriter(IRMatcher::min(a, b), op->type);
 
+        if (rewrite(min(x, x), x) ||
```
abadams, line 57: ditto with min(x, x)

src/Simplify_Max.cpp
rootjalex: I was worried about changing the order of rewrite rule evaluation - can this first rule be moved down to be evaluated with the rest?

src/Simplify_Div.cpp
```diff
@@ -118,8 +119,15 @@ Expr Simplify::visit(const Div *op, ExprInfo *bounds) {
         if (rewrite(IRMatcher::Overflow() / x, a) ||
             rewrite(x / IRMatcher::Overflow(), b) ||
             rewrite(x / 1, x) ||
```
dsharletg, line 121: Should this one go below too?

dsharletg (commented):
Would a less involved change here be to move the rules involving signed integer overflow to the recursively mutating rewrite rules, and then just handle this in the visitor for signed integer overflow?

rootjalex:
@dsharletg Yes - I made that change, and tested a few expressions to make sure the bug no longer exists - I believe this means that only constant folding needs to be moved to the recursively mutating rewrite rules section, so made that change as well. This PR still fixes the two issues it references

dsharletg (approved):

rootjalex:
Meant to comment before merging - failures are build related, not related to this PR

PR #6013, opened 2021-05-14
Add string names to all Vars so pseudo-code output is more human-readable.

kaushik1618:
This change makes the pseudo-code snippets in https://halide-lang.org/tutorials/tutorial_lesson_05_scheduling_1.html match what you get from actually compiling and executing the lesson_05_scheduling_1.cpp source code locally.

### DIFF
--- a/tutorial/lesson_05_scheduling_1.cpp
+++ b/tutorial/lesson_05_scheduling_1.cpp
@@ -116,7 +116,7 @@ int main(int argc, char **argv) {
 
         // The most powerful primitive scheduling operation you can do
         // to a var is to split it into inner and outer sub-variables:
-        Var x_outer, x_inner;
+        Var x_outer("x_outer"), x_inner("x_inner");
         gradient.split(x, x_outer, x_inner, 2);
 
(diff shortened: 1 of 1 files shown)


kaushik1618:
Ping, is there something I need to do to make this easier to review/approve?

abadams:
The pseudocode snippets are supposed to match automatically. Something is going wrong with debugging symbols if they're not. My preferred fix is to figure out what went wrong with those, rather than adding explicit strings.

zvookin:
@abadams @steven-johnson . I can see three ways to go here. First, just merge it. Second, decide the tutorials do not have the names because it makes their source clearer to read. In that event we'd recommend using the introspection support to get better names when needed. Third, modify all tutorials accordingly.

kaushik1618:
I second @zvookin 's thoughts on this, there is a lot of educational value in having the tutorial output snippets match actual outputs. I agree that the right fix is to get to the root of the problem but I don't know when that fix will land.

steven-johnson:

PR #6017, opened 2021-05-17
More simplifier rules

dsharletg:
This PR adds a bunch of simplifier rules that are mostly straightforward new cases of existing rules.


src/Simplify_Select.cpp
```diff
@@ -152,6 +184,9 @@ Expr Simplify::visit(const Select *op, ExprInfo *bounds) {
              rewrite(select(x, y + z, y), y + select(x, z, 0)) ||
              rewrite(select(x, y, y + z), y + select(x, 0, z)) ||
 
+             rewrite(select(x, y - z, y), y + select(x, -z, 0), y != 0) ||
```
abadams, line 187: I think you want !is_const(y) as the predicate for this and the next rule.

dsharletg: That wouldn't work for select(x, 3 - z, 3) though, which is I think what I actually added this rule for?

abadams: Discussed offline, summarized here: that would increase op count, and it would be perfectly legal for another rule to fold the constant right back in, forming a cycle.

src/Simplify_Mul.cpp
```diff
@@ -69,6 +69,8 @@ Expr Simplify::visit(const Mul *op, ExprInfo *bounds) {
 
         if (rewrite((x + c0) * c1, x * c1 + fold(c0 * c1), !overflows(c0 * c1)) ||
             rewrite((c0 - x) * c1, x * fold(-c1) + fold(c0 * c1), !overflows(c0 * c1)) ||
+            rewrite((0 - x) * y, 0 - x * y) ||
+            rewrite(x * (0 - y), 0 - x * y) ||
```
rootjalex, line 73: Do these rules obey the reduction order? @abadams @jn80842

rootjalex (approved):
All of the non-vector rules verify correctly, and the vector rules look reasonable to me. Lgtm as long as those two rules I commented on obey the reduction order

src/Simplify_Mul.cpp
abadams: Yes, promoting an add or sub to the root is one of the ordering terms, and the rule is a tie for all of the ordering terms before it.

PR #6020, opened 2021-05-18
Add TailStrategy::PredicateLoads and TailStrategy::PredicateStores

dsharletg:
This adds two new TailStrategy values, `TailStrategy::PredicateLoads` only predicates the loads and not the stores, and `TailStrategy::PredicateStores` only predicates the stores and not the loads. This effectively makes the tail strategy for the unpredicated stores or loads `TailStrategy::RoundUp`.

This PR also refactors the implementation of predication, from expression tag hints to:
- Representing predicated loads with if_then_else(predicate, load)
- Adding a predicate field to Provide

This PR also now means that TailStrategy::Predicate and TailStrategy::GuardWithIf are identical. Historically, this has caused problems with performance on ARM/x86, but I am struggling to find a case of this to troubleshoot. I think even if it does happen in rare cases, it might be worth the complexity/code size/more consistent behavior across platforms improvements.


src/IROperator.cpp
```diff
@@ -1048,6 +1048,18 @@ Expr predicate(Expr e) {
                                 {std::move(e)}, Internal::Call::PureIntrinsic);
 }
 
+Expr predicate_loads(Expr e) {
+    Type t = e.type();
+    return Internal::Call::make(t, Internal::Call::predicate_loads,
+                                {std::move(e)}, Internal::Call::PureIntrinsic);
```
abadams, line 1054: I don't think this new intrinsic is pure. pure intrinsics can be moved around in this IR without changing the meaning. This one must stay where it is immediately inside the if condition.

src/Schedule.h
```diff
@@ -48,16 +48,34 @@ enum class TailStrategy {
...
      * output sizes. Cons: increases code size due to separate
      * tail-case handling. */
     Predicate,
 
+    /** Guard the loads in the loop with an if statement that
+     * prevents evaluation beyond the original extent. Not legal
+     * for RVars, as it would change the meaning of the algorithm.
+     * The if statement is treated like a boundary condition, and
+     * factored out into a loop epilogue if possible.
+     * Pros: no redundant re-evaluation; does not constrain output
```
abadams, line 65: I don't see how redundant re-evaluation is worse than adding extra computation on the end.

Also I think you mean does not constrain input sizes, not output

src/ScheduleFunctions.cpp
```diff
@@ -79,6 +79,58 @@ bool contains_impure_call(const Expr &expr) {
     return is_not_pure.result;
 }
 
+// A mutator that performs a substitute operation only for a certain kind
```
abadams, line 82: I don't understand how this works when IRNodeType is anything other than Provide. Doesn't it recursively substitute things inside the values?

abadams: Oh it only calls substitute on the args, not the values. That's non-obvious.

abadams: What happens if you predicate stores on:

`hist(im(r.x, r.y)) += 1`

when splitting one of the rvars?

I think this mutator might do the wrong thing in that scenario.

dsharletg: I decided to just disable PredicateStores for reductions, because it's pretty confusing even if it works. (I already didn't enable PredicateLoads).

abadams: For reductions, PredicateStores seems fine to me, even if it implies unpredicated self-loads. No stores = no side-effects.

abadams: I was mostly using that example not to complain about reductions, but to point out that this mutator doesn't really handle nesting. E.g. if a call is found inside the LHS of a Provide. Passing it arbitrary IRNodeTypes is going to break whenever they can nest.

Another issue is if the things that need to be substituted out have already been lifted out of the index of a Call into a let by the CSE pass that happens inside Function::define to protect us from pathological Exprs. Or even worse:

```foo[let y = the_loop_var in bar[y] + y] = 5```

Yuck. OK, maybe disallowing reductions is wise for now.

I think the correct thing (with reductions forbidden) is that for predicate loads, you just substitute in the entire RHS of each provide, and for predicate stores, you substitute in the entire LHS of each provide.

File: src/IROperator.cpp (hunk 1 above)
Comment (dsharletg) [reply src/IROperator.cpp]: I changed this (and the pre-existing Call::predicate), but this seems to break some logic in ScheduleFunctions, which looks for impure calls in conditions.

I changed this back to PureIntrinsic for now. I agree this is an issue, but it's pre-existing with Call::predicate too, and I'm not sure how this should change yet.

**Comment by dsharletg** [comment]
I think this is ready for review again after the last round of changes.

File: src/IR.h [hunk 4]
```diff
@@ -651,6 +651,19 @@ struct Call : public ExprNode<Call> {
         return is_intrinsic() && this->name == get_intrinsic_name(op);
     }
 
+    bool is_intrinsic(std::initializer_list<IntrinsicOp> intrinsics) const {
```
abadams, line 654: What's this about? Isn't this given by the call type?

dsharletg: This is the same as is_intrinsic(IntrinsicOp), except it allows checking for multiple intrinsics at a time a bit more easily

abadams:
I think the nested_tail_strategies test should have some coverage of this (possibly with randomness to avoid it taking too long)

src/IR.h
abadams: Oh right, I was misinterpreting the arg as being the list of all intrinsics that exist

src/ScheduleFunctions.cpp
```diff
@@ -79,6 +79,98 @@ bool contains_impure_call(const Expr &expr) {
...
+            changed = changed || !args.back().same_as(i);
+        }
+        if (changed) {
+            return Provide::make(p->name, p->values, args, p->predicate);
+        } else {
+            return p;
+        }
+    }
+
+    Expr visit(const Call *op) override {
```
abadams, line 109: If this is just mutating the LHS vs RHS, then this call visitor doesn't need to exist. Just call substitute in the Provide visitor.

dsharletg: I think we really want to limit the substitutions to calls. We want only instances of coordinates inside calls to be clamped, not those outside.

dsharletg:
It looks like the LLVM asserts in correctness_vector_reductions are a coincidental recent regression in LLVM on master, unrelated to this PR.

steven-johnson:
> It looks like the LLVM asserts in correctness_vector_reductions are a coincidental recent regression in LLVM on master, unrelated to this PR.

agreed, running a bisect locally to try to track it down

dsharletg:
I think this is ready for review again. I've added `is_var_bounded`, which fixed the remaining test failures from before that I currently know of.

I know we talked about adding this as a new piece of IR, but having it be an intrinsic with a StringImm argument doesn't seem that bad to me. I can of course add a new piece of IR if we still want to, but it's going to be a much bigger change.

abadams:
@rootjalex would be very interested to see how this branch compares to master when tested against your random-tail-strategy variant of the autoscheduler

PR #6030, opened 2021-05-19
Allow hannk-delegate input and output tensors to share memory with tflite

steven-johnson:
We previously allocated duplicate memory buffers for the input and output tensors, and just copied them back and forth as needed, which wastes memory and cycles.

Now, we declare the input and output tensors to be 'external', and update the host pointer before every interpreter run.

Note that dynamic tensors are still a bit of a special case: we still do duplicate allocations there (plus memcpy), because in the most general case we can't know the final size needed until we run the pipeline, but we need to allocate that output to run the pipeline. There are ways we could finesse this -- e.g., give dynamic tensors a lambda callback to allow them to resize the TFLite tensor and then use that storage -- but since dynamic tensors don't seem to be common or large in our test cases, I'm doing it this way for now. (I may circle back and try the lambda approach later.)

Note that this moves the implicit init of `is_constant_` out of the Tensor ctor, and instead always requires an explicit call to `set_constant()`, which I think makes the situation much clearer.

### DIFF
--- a/apps/hannk/interpreter/interval.h
+++ b/apps/hannk/interpreter/interval.h
@@ -154,6 +154,18 @@ public:
     }
 };
 
+template<typename T, size_t Capacity>
+inline std::ostream &operator<<(std::ostream &s, const SmallVector<T, Capacity> &v) {
+    s << "{";
+    for (size_t i = 0; i < v.size(); ++i) {
+        if (i > 0) {
+            s << ", ";
+        }
+        s << v[i];
+    }
+    return s << "}";
+}
+
 // Compute a / b, rounding down.
 inline int floor_div(int a, int b) {
     assert(a >= 0 && b >= 0);
--- a/apps/hannk/interpreter/model.cpp
+++ b/apps/hannk/interpreter/model.cpp
@@ -76,7 +76,6 @@ Tensor::Tensor(std::string name, HalideBuffer<void> buffer, QuantizationInfo qua
     : name_(std::move(name)),
       buffer_(std::move(buffer)),
       quantization_(std::move(quantization)) {
-    is_constant_ = buffer_.data() != nullptr;
 }
 
 Tensor::Tensor(std::string name, halide_type_t type, const Box &bounds, QuantizationInfo quantization)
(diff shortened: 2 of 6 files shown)


apps/hannk/interpreter/interval.h
```diff
@@ -154,6 +154,18 @@ class SmallVector {
     }
 };
 
+template<typename T, size_t Capacity>
+inline std::ostream &operator<<(std::ostream &s, const SmallVector<T, Capacity> &v) {
```
dsharletg, line 158: BTW, I think we should remove this (and std::vector's version) at some point, but this is fine for now.

apps/hannk/interpreter/model.cpp
```diff
@@ -142,7 +154,7 @@ void Tensor::allocate() {
         return;
     }
 
-    if (is_dynamic()) {
+    if (is_dynamic() || is_external()) {
         return;
```
dsharletg, line 158: I was going to suggest asserting buffer_.data() is not null, but that would clearly fail if the if above doesn't return before getting here. Is this case actually hit?

apps/hannk/interpreter/model.h
```diff
@@ -92,9 +92,20 @@ class Tensor {
     std::string name_;
     HalideBuffer<void> buffer_;
     QuantizationInfo quantization_;
+    // If true, this Tensor should be considered constant and should not be mutated.
+    // (It may actually refer to read-only external memory, or it may simply be marked this may as
+    // the result of a transform.)
     bool is_constant_ = false;
+    // If true, this Tensor's storage is externally owned and must not be freed.
+    bool is_external_ = false;
```
dsharletg, line 100: Almost all uses of this are either asserts, or in the delegate, where we could maybe check this from the TFlite side, should we try to do that instead?

If is_external is true, is is_input_ || is_output_ true?

apps/hannk/interpreter/interval.h
steven-johnson: It's extremely useful for debugging.

apps/hannk/interpreter/model.cpp
steven-johnson: Yes; we mark Tensors as external (or not) before we call allocate().

apps/hannk/interpreter/model.h
steven-johnson: > Almost all uses of this are either asserts, or in the delegate, where we could maybe check this from the TFlite side, should we try to do that instead?

I think that using it in asserts is reason enough for it to exist -- the variations in Tensor usage are pretty delicate, and I'd like to keep this validation in place. (I did downgrade most of the HCHECK usage to assert...)

> If is_external is true, is is_input_ || is_output_ true?

For current usage, yes, but there's no intrinsic reason that it has to be.

apps/hannk/interpreter/interval.h
dsharletg: I agree, but I still don't think we should keep it long term. But as I said, for now it's fine.

dsharletg (approved):

PR #6033, opened 2021-05-19
Remove duplicate -e argument in bilateral_grid

### LINKED ISSUE #6008: Duplicate -e argument specified in Makefile for bilateral_grid
https://github.com/halide/Halide/blob/75a079a853e98a7d1d7e2e685fb1242089a41460/apps/bilateral_grid/Makefile#L17

In the Makefile of bilatergal_grid, for the auto_schedule target, -e is specified twice, once as "-e $(GENERATOR_OUTPUTS)" and once as "-e static_library,c_header,schedule". 

In Generator.cpp's generate_filter_main_inner(), the arguments are parsed by creating a map and assigning key value pairs. This has the effect that only the last specified value for an arg will be used if an arg is specified multiple times (i.e. -e $(GENERATOR_OUTPUTS) is not used).
https://github.com/halide/Halide/blob/75a079a853e98a7d1d7e2e685fb1242089a41460/src/Generator.cpp#L810

This is a very minor issue but can give someone inspecting the Makefile the false impression that the $(GENERATOR_OUTPUTS) is used in some way (like that both of the values for the -e's might be appended together) for the bilatergal_grid auto_schedule target when it is not.

steven-johnson:
Fixes #6008

### DIFF
--- a/apps/bilateral_grid/Makefile
+++ b/apps/bilateral_grid/Makefile
@@ -14,7 +14,7 @@ $(BIN)/%/bilateral_grid.a: $(GENERATOR_BIN)/bilateral_grid.generator
 
 $(BIN)/%/bilateral_grid_auto_schedule.a: $(GENERATOR_BIN)/bilateral_grid.generator
 	@mkdir -p $(@D)
-	$^ -g bilateral_grid -e $(GENERATOR_OUTPUTS) -o $(@D) -f bilateral_grid_auto_schedule target=$*-no_runtime auto_schedule=true -e static_library,c_header,schedule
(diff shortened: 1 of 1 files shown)


alexreinking (approved):

alexreinking:
jsyk - the GitHub magic is "Fixes #issue-num" in either the commit message or the PR description. It doesn't check the title and it doesn't work without the magic word "Fixes" (or its synonyms "Close[s/d]", "Resolve[s,d]"). 

See docs: https://docs.github.com/en/github/managing-your-work-on-github/managing-your-work-with-issues-and-pull-requests/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #6034, opened 2021-05-19
Update README for Halide 12 release.

### LINKED ISSUE #4216: When should we consider moving Halide to C++17?

### LINKED ISSUE #5154: Add compiler/platform support matrix to README.md
I was mildly surprised to learn that we don't support building Halide with Homebrew's GCC 10 on macOS. By my understanding, we support the following host compilers and platforms:

| Platform | Compiler | Version |
| --- | --- | --- |
| Windows | MSVC | 14.1+ (VS 2017) |
| Windows | MSVC | 14.2+ (VS 2019) |
| macOS | AppleClang | ?? |
| Linux | GCC | 5.0+ |
| Linux | Clang | 9+ |

Maybe this should just be pasted in to the README?

### LINKED ISSUE #5633: Update README and build files for OpenGLCompute requirements
Support for OpenGL was removed in #5626 for Halide 12, but support for OpenGLCompute is being kept; the requirements for OpenGLCompute (in terms of runtime dependencies, etc) are similar to OpenGL but it's not clear if it's identical or not, so various README and Make/CMake deps may need updating. At the time of this issue being opened, that's not clear, because OpenGLCompute isn't being tested (#5627) due to LLVM symbol exporting (#4651); thus this is a placeholder to ensure that we circle back to address this before Halide 12 is released.

### LINKED ISSUE #5961: Lowest supported Windows version
The stack size expansion routine introduced in 59a04e4c68fae34d3dd59cfe2df586c0311cbaad uses the function `GetCurrentThreadStackLimits`. As this Windows API is only appered after Windows 8, or Windows Server 2012, the resulting binary can't run on Windows 7.

For now, support for Windows 7 can be easily restored by removing the check for remaining stack space and always expand the stack. The question to the dev team is should it be done? Understandably, supporting legacy systems require extra effort and resources. If the decision is to drop support for Windows 7 and earlier, it may a good idea to explicitly mention it somewhere.

alexreinking:
Fixes #5961 
Fixes #5154
Fixes #4216
Fixes #5633  (by not mentioning it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@ Halide is a programming language designed to make it easier to write
 high-performance image and array processing code on modern machines. Halide
 currently targets:
 
-- CPU architectures: X86, ARM, MIPS, Hexagon, PowerPC
-- Operating systems: Linux, Windows, Mac OS X, Android, iOS, Qualcomm QuRT
-- GPU Compute APIs: CUDA, OpenCL, OpenGL Compute Shaders, Apple Metal,
-  Microsoft Direct X 12
+- CPU architectures: X86, ARM, MIPS, Hexagon, PowerPC, RISC-V
+- Operating systems: Linux, Windows, macOS, Android, iOS, Qualcomm QuRT
+- GPU Compute APIs: CUDA, OpenCL, OpenGL Compute Shaders, Apple Metal, Microsoft
+  Direct X 12
 
 Rather than being a standalone programming language, Halide is embedded in C++.
 This means you write C++ code that builds an in-memory representation of a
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):

README.md
```diff
@@ -156,16 +185,16 @@ Follow the above instructions to build LLVM or acquire a suitable binary
 release. Then change directory to the Halide repository and run:
 
 ```
-% cmake -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=$LLVM_ROOT/lib/cmake/llvm -S . -B build
+% cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=$LLVM_ROOT/lib/cmake/llvm -S . -B build
```
abadams, line 188: Doesn't this also require acquiring ninja, would would need to be added to the instructions? I'm not sure it's a good default.

abadams:
LGTM except for a minor concern about making ninja the default.

README.md
alexreinking: I made this change for two reasons, let me know what you think:

1. We only use Ninja with CMake on the buildbots, so if there are issues with the Makefile backend, we won't notice them.
2. The user experience with Ninja is much nicer (parallel by default, cleaner output, etc.)

Also, it's perfectly optional and one can avoid ninja by simply dropping `-G Ninja`. I wrote a note on that below.

abadams: I think that people copy-paste build steps without reading or understanding them, and that this one isn't going to work without first telling them to acquire ninja. Maybe the text below could more explicitly say that if you get the error <exact error string>, try again without -G Ninja

PR #6037, opened 2021-05-19
Improve hannk tensor sharing

steven-johnson:
This builds on top of #6030 to allow sharing memory for dynamic tensors as well, via a callback function to allow for resizing in a flexible way. PTAL.


steven-johnson:
Converted to draft (but keeping open): this improves some corner cases, but could be problematic for use on non-CPU targets due to callbacks. Going to leave here in draft form until I get around to testing on such a setup to see if could be worth the trouble or not.

PR #6038, opened 2021-05-19
Fix tensor aliasing issues

dsharletg:
- Don't try to alias a tensor that already has been allocated
- Fix ridiculous bug

### DIFF
--- a/apps/hannk/interpreter/transforms.cpp
+++ b/apps/hannk/interpreter/transforms.cpp
@@ -64,6 +64,11 @@ void remove_dead_ops(OpGroup *root) {
 
 namespace {
 
+// Check if a tensor already has storage configured.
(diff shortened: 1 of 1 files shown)


apps/hannk/interpreter/transforms.cpp
```diff
@@ -105,18 +110,18 @@ bool maybe_alias_tensors(TensorPtr input, TensorPtr output, SmallVector<int, max
...
+    } else if (output_subset_of_input && !has_storage(output)) {
         for (int &i : offset) {
             i = -i;
         }
         output->set_alias_of(input, offset);
         return true;
     }
 
-    return true;
+    return false;
```
steven-johnson, line 124: yikes

steven-johnson (approved):
May need attention after #6030 and/or #6037 lands.

PR #6041, opened 2021-05-19
Scalar loads/stores shouldn't invalidate predicate vectorization

vksnk:


dsharletg (approved):
Nice find. This explains why it was working in some of my specializations and not others: the specializations were for broadcasting (scalar loads).

steven-johnson:
Does this need a test?

steven-johnson:
Looks like an error in correctness_predicated_store_load: https://buildbot.halide-lang.org/master/#/builders/126/builds/334

vksnk:
Fixed the failure and added a test to verify that the change does something useful (also verified that the new test failed before the change).

test/correctness/predicated_store_load.cpp
```diff
@@ -91,6 +91,31 @@ int predicated_tail_test(const Target &t) {
     return 0;
 }
 
+int predicated_tail_with_scalar_test(const Target &t) {
+    int size = 73;
+    Var x("x"), y("y");
+    Func f("f"), g("g";
```
steven-johnson, line 97: Missing a `)` here

vksnk: Oops, last minute clean-up has gone wrong.

steven-johnson:
Another correctness_predicate_store_load failure: https://buildbot.halide-lang.org/master/#/builders/126/builds/346

vksnk:
> Another correctness_predicate_store_load failure: https://buildbot.halide-lang.org/master/#/builders/126/builds/346

Sorry, having some issues with set up of Hexagon SDK, so couldn't verify locally. 

It seems that at some point for some of the tests predicated loads/stores stopped to appear and this PR fixes that, so some of the test values needed to be adjusted.

vksnk:
Bots are green, submitting.

steven-johnson:
Tagging @alexreinking, this might be needed for a future backport to 12.0.2 (not sure how urgent)

alexreinking:
> Tagging @alexreinking, this might be needed for a future backport to 12.0.2 (not sure how urgent)

We _can_ do this, but maybe it's worth waiting for a few more bugfixes, given how much buildbot time a release takes?

dsharletg:
I'm not sure how necessary backporting this is. It was a long-standing performance bug...

alexreinking:
OTOH - long-standing performance bugs that don't break ABI (like this one) are good candidates for backporting.

PR #6042, opened 2021-05-20
Add DMA-BUF support to host_malloc

dpalermo:
- use DMA-BUF if libdmabufheap.so is present
  - fallback to ION for older devices/OS
  - ION APIs are no longer supported on Android-S/12
  - added context to the HAP_power_get/set calls
  - NULL power context not allowed on newer devices


dpalermo:
Note that the last commit also brought in the libhalide_hexagon_host.so binaries (which I was trying to leave out of the initial review).  Forgot to remove them from my tree after my local test build before the last commit.  I'll leave these here for now, they can be removed later if changes are requested.

dpalermo:
The libhalide_hexagon_remote_skel.so binaries will still need to be rebuilt as well (for the non-NULL power context changes)

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -93,15 +93,32 @@ int halide_hexagon_remote_load_library(const char *soname, int sonameLen,
 }
 
 volatile int power_ref_count = 0;
+static volatile void *power_context = NULL;
+
+static void *get_HAP_power_context() {
+    if (power_context == NULL) {
+        power_context = (void *)malloc(1);
```
dsharletg, line 100: Why not just pass an address of a global value to the function instead of a dynamic allocation?

Does it even need to be a valid allocation? (Currently uninitialized memory is being passed.)

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -93,15 +93,32 @@ int halide_hexagon_remote_load_library(const char *soname, int sonameLen,
...
+        log_printf("get_HAP_power_context: %p\n", power_context);
```
dsharletg, line 101: I don't think we should be doing log_printf in non-error cases.

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -194,15 +214,16 @@ int halide_hexagon_remote_set_performance_mode(int mode) {
...
     unsigned int max_mips = 0;
     uint64 max_bus_bw = 0;
     HAP_power_request_t request;
 
     power_info.type = HAP_power_get_max_mips;
-    int retval = HAP_power_get(NULL, &power_info);
+    int retval = HAP_power_get(pwr_ctx, &power_info);
     if (0 != retval) {
-        log_printf("HAP_power_get(HAP_power_get_max_mips) failed (%d)\n", retval);
+        log_printf("HAP_power_get(%p,HAP_power_get_max_mips) failed (%d)\n", pwr_ctx, retval);
```
dsharletg, line 226: Space

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -134,19 +189,84 @@ extern "C" {
...
+        return;
+    }
+    use_libdmabuf = false;
     use_newer_ioctl = false;
-    ion_fd = open("/dev/ion", O_RDONLY, 0);
-    if (ion_fd < 0) {
-        __android_log_print(ANDROID_LOG_ERROR, "halide", "open('/dev/ion') failed");
+    use_libion = false;
+
+    __android_log_print(ANDROID_LOG_INFO, "halide", "entering halide_hexagon_host_malloc_init");
```
dsharletg, line 202: I really don't think we should do this kind of unconditional logging.

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -52,14 +79,31 @@ struct ion_handle_data {
 #define ION_IOC_FREE _IOWR('I', 1, ion_handle_data)
 #define ION_IOC_MAP _IOWR('I', 2, ion_fd_data)
 
-bool use_newer_ioctl = false;
+extern "C" {
+__attribute__((weak)) extern int ion_open();
```
dsharletg, line 83: Unused? Can you make sure there are no other unused things that I haven't noticed?

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -134,19 +189,84 @@ extern "C" {
...
+        ion_alloc_fd_fn = (rem_ion_alloc_fd_fn)dlsym(lib, "ion_alloc_fd");
+        if (!ion_open_fn || !ion_alloc_fd_fn) {
+            __android_log_print(ANDROID_LOG_ERROR, "halide", "huge problem in libion.so");
+            return;
+        }
+        ion_fd = ion_open_fn();
+        if (ion_fd < 0) {
+            __android_log_print(ANDROID_LOG_ERROR, "halide", "ion_open failed");
+        }
+        __android_log_print(ANDROID_LOG_INFO, "halide",
```
dsharletg, line 242: Remove these unconditional prints

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -173,22 +293,39 @@ void *halide_hexagon_host_malloc(size_t size) {
...
+        }
+    } else if (use_libion) {
+        buf_fd = ion_alloc(ion_fd, size, alignment, 1 << heap_id, ion_flags);
+        if (buf_fd < 0) {
+            __android_log_print(ANDROID_LOG_ERROR, "halide", "ion_alloc(%d, %zd, %zd, %d, %d) failed",
+                                ion_fd, size, alignment, 1 << heap_id, ion_flags);
+            return NULL;
+        }
+    } else {  // !use_libion
+        handle = ion_alloc(ion_fd, size, alignment, 1 << heap_id, ion_flags);
```
dsharletg, line 312: This looks identical to the `if (use_libion)` case?

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -173,22 +293,39 @@ void *halide_hexagon_host_malloc(size_t size) {
...
+        if (handle < 0) {
+            __android_log_print(ANDROID_LOG_ERROR, "halide", "ion_alloc(%d, %zd, %zd, %d, %d) failed",
+                                ion_fd, size, alignment, 1 << heap_id, ion_flags);
             return NULL;
         }
+        // Map the ion handle to a file buffer.
+        if (use_newer_ioctl) {
+            buf_fd = handle;
+        } else {
+            buf_fd = ion_map(ion_fd, handle);
```
dsharletg, line 322: In what conditions is this older codepath used? Maybe we can get rid of it?

dsharletg (changes requested):
Can you make a pass over this code for unused code/declarations, and for unconditional logging (potentially spammy)?

src/runtime/hexagon_remote/halide_remote.cpp
dpalermo: The power team says that the pointer returned from malloc will be unique over all processes.  The value does not need to be initialized.  The context ID is the address returned from malloc.

src/runtime/hexagon_remote/halide_remote.cpp
dpalermo: Done.  Will push up another patch.

src/runtime/hexagon_remote/halide_remote.cpp
dpalermo: Done.

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: Removed.

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: use_newer_ioctl is still used, it was just moved to the top of the file.  I haven't found any unused items.

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: Done.

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: The difference (for !use_libion) is further below in lines 318-328

src/runtime/hexagon_remote/host_malloc.cpp
dsharletg: I'm pretty sure this particular function is unused. Ctrl+F "ion_open()" finds only this line?

dpalermo: Sorry...I thought you meant the deletion of use_newer_ioctl.  All of the "__attribute__((weak)) extern" prototype declarations in this file are actually for functions that are looked up by name using dlsym.  I believe this approach was given to us by the linker team...to make things work correctly with the dynamic loader.
If you really want these prototypes removed, I can try removing them but would need to run it through all of our devices (old & new) to be sure it wasn't needed somewhere.

dpalermo: All attribute((weak)) prototypes == all of the libdmabufheap.so or libion.so symbols
...and not remote_register_buf which is actually called directly in this file.

dsharletg: I see now, all of these weak declarations are unused... I don't see how they could affect the generated code. Also note this is the android linker/loader, not the Hexagon one. We use this approach of dlopen/dlsym in several places and don't forward declare the functions like this. It's quite a bit of redundant specific stuff (argument types) to maintain that is dead code. I think all you need is the typedefs and function pointers.

src/runtime/hexagon_remote/halide_remote.cpp
dsharletg: So the goal here is just to get a unique ID? Can you comment something to that affect?

I also think there's probably a better way to get this. Is there a function to get a process ID on Hexagon, like getpid()?

It would be nice to avoid the need to free this thing.

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: Change builds fine (as expected...unused functions ;-).   The generated libhalide_hexagon_host.so is also exactly the same without these prototypes...so they don't seem to be adding any meta information otherwise to the objects.   I have a feeling we may have been calling theses symbols directly at one point with the libion.so support (which was done a while back)...and switched to dlsym some point later.  The DMA-BUF support just followed the same pattern.  I'll push up another patch.

src/runtime/hexagon_remote/halide_remote.cpp
dsharletg: Also, I find it rather suspicious that malloc is guaranteed to be unique over all processes. That suggests processes are sharing an address space, which I think (I hope...) is not the case?

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: Sorry, again didn't clean out my built objects before pushing that last patch 8-(.

src/runtime/hexagon_remote/halide_remote.cpp
dpalermo: Yes, we were surprised as well by that claim....but haven't tried to prove/disprove it ourselves.  Could easily be that the malloc produces a "unique-enough" context ID over all processes that power/voting works better...as opposed to all processes just using a NULL context ID.  I am not familiar with what goes on in the power/voting with the context ID....just that the newer devices refuse to dispatch is the context ID is NULL...and this is the recommended solution from the power team.

dpalermo: ...and the free isn't explicitly called at the moment....it is just happening on process exit for now.  For a long running process with multiple dispatches, I really didn't see a need to free the power context after a dispatch and get a new one on the next....esp. since this approach get/free isn't sending any information to the power controls.

dpalermo: We could leave the power context change out of this PR if you like....and revisit it separately.  It came up at the same time as getting the DMA-BUF changes working on Android-S/12...and since it was also in the Halide runtime, I figured it would be good to get both of these changes in at once so they were there from day 1 on official release of Android-S/12.   The non-NULL power context is also only needed for devices newer than SM8350 (avoids breaking any currently deployed apps)...another reason we could leave it out for you.

dsharletg: If free_HAP_power_context is dead code, then get rid of it?

In general I think this whole thing is pretty sketchy:
- Why does HAP_power_set need a system wide unique identifier? It should track the process votes come from itself if it cares about that.
- Why is the argument a pointer if it isn't going to be dereferenced?
- Getting a unique ID with malloc is both unlikely to actually work and also kind of sloppy (never freed, calling malloc just to get an identifier is weird).

IMHO the best thing here is to get a proper process unique ID somehow, and cast it to a pointer.

dpalermo: OK, I just went through the latest internal docs from the power team and found:
- it only has to be unique within the process
- they are still recommending the malloc(1)/free for creating/destroying a context
- the context is just an identifier....even though it is a void*, it is never stored into....they even say you could even just use "1" (non-NULL), but that would not be unique within the process

I have a feeling that they used (void *) because they are using malloc to create process-wide unique IDs.

The reason for this change to no longer allow NULL power context is:
"If there are multiple clients running in the same process(e.g., camera and ML clients) and both of them using NULL context for voting purpose. Camera might vote using NULL and ML might remove vote using same NULL context. fastrpc wouldn't know to remove (or) retain vote as both clinets are using same NULL context."

dsharletg: In that case, I suggest just passing the address of a global variable in our runtime. That should give equivalent unique-ness to what you have here (a global variable assigned by malloc).

dpalermo: There is a new answer in the FAQ, however, that wasn't there before that makes me realize that we do need to still find a point at which we can say we are done with a context (specifically see the mention of HAP_power_destroy_client):

```
Is there any limit on number of contexts (or) clients contexts that can be created?

Yes there is limit on number of client context that can be created. This limit is coming from MMPM
to create voting client context. There is a limit of 192 power clients at a time by MMPM on the DSP.
Old client contexts need to be destroyed(HAP_power_destroy_client) to create new client contexts.
```

The HAP_power_destroy_client function doesn't exist in SDK 3.3.3 incs/HAP_power.h....was added in SDK 3.5.1 and is actually supposed to be called through a wrapper HAP_power_destroy which checks (weak) for HAP_power_destroy_client exists before calling it.

dpalermo: Not sure if halide_hexagon_remote_power_hvx_off() is the right place (since we would probably need to revote when we turn hvx back on).   I'm thinking we need to register a function via atexit() to call HAP_power_destroy() (if the power context is non-NULL) as the process exits.  What do you think?

dsharletg: Do we really need to call HAP_power_destroy_client at process exit? It's not going to get cleaned up anyways? What happens if the app crashes?

dpalermo: Good question.   It is possible that the process exiting is sufficient for fastrpc to know we are done.  We may not even need to call HAP_power_destroy_client if fastrpc will itself releases the associated clients/contexts.  I'll ask the power team.

dpalermo: Answer:
"It is ideal to always call HAP_power_destroy In your process tear down process.
Once the process exits(normal or exception) fastrpc teardown will take of releasing all client contexts. It shouldn’t leak once the DSP User PD exits completely."

dpalermo: Pushed up a new change without the malloc/free and not calling HAP_power_destroy on the context ID.

dpalermo: ...and if we aren't ever going to free/change the power_context ID....I just pushed up a simplifying change that just returns the global power_context, doesn't store it.  This should be fine, but I'm rerunning new tests on this locally (on all old & new devices/OSes).  Will report back later.

dpalermo: Testing looks good on this latest change.  Ran fine both on older and newer devices/OSes...including the newest device that requires a non-NULL context.

dpalermo: Would you like me to add some of the above explanation as comments here?

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -94,14 +94,19 @@ int halide_hexagon_remote_load_library(const char *soname, int sonameLen,
 
 volatile int power_ref_count = 0;
 
+static inline void *get_HAP_power_context() {
+    return (void *)(&power_ref_count);
```
dsharletg, line 98: Add some comments explaining this here.

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -94,14 +94,19 @@ int halide_hexagon_remote_load_library(const char *soname, int sonameLen,
...
+}
+
 int halide_hexagon_remote_power_hvx_on() {
     if (power_ref_count == 0) {
+        void *pwr_ctx = get_HAP_power_context();
```
dsharletg, line 103: `pwr_ctx` -> `power_context`

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -52,14 +76,29 @@ struct ion_handle_data {
 #define ION_IOC_FREE _IOWR('I', 1, ion_handle_data)
 #define ION_IOC_MAP _IOWR('I', 2, ion_fd_data)
 
-bool use_newer_ioctl = false;
+extern "C" {
```
dsharletg, line 79: Anonymous namespace, not extern "C"

src/runtime/hexagon_remote/host_malloc.cpp
```diff
@@ -1,17 +1,41 @@
...
+bool use_libdmabuf = false;
+bool use_newer_ioctl = false;
+bool use_libion = false;
+
 namespace {
 
+// DMA-BUF support
+void *dmabufAllocator = NULL;
+
+extern "C" {
```
dsharletg, line 19: These shouldn't be extern "C", only in an anonymous namspace (which they already are).

dsharletg (commented):
Looks mostly good, only minor comments this time.

src/runtime/hexagon_remote/halide_remote.cpp
dpalermo: Done

src/runtime/hexagon_remote/halide_remote.cpp
dpalermo: Done

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: Done

src/runtime/hexagon_remote/host_malloc.cpp
dpalermo: Done

dpalermo:
...and a reason for using the address of something as opposed to a constant would be if someone wrote an app that had two separate components/threads in it...one Halide and one ML....and both were using Hexagon (and calling HAP_power_set)...they would need to be using different context IDs to not interfere with each others settings.   For example, if both threads decided to use (void*)1 as their power context...then the power requests from the two threads couldn't be distinguished as coming from two separate clients.

If you are concerned about returning the address of an existing global, we could instead just create something that can't be used by anything else...like returning a static local:

```
static inline void *get_HAP_power_context() {
    // Halide runtime only needs a single power context
    static int power_context = 0;
    return (void*)(&power_context);
}
```
Let me know what you think.

dpalermo:
Regarding when the different DMA-BUF/ION interfaces came into use, it is close to the Android release cycle, but there is some potential for overlap.

Android S (12):

libdmabufheap.so first appears in Android S GKI 2.0 kernel (from kernel 5.10)...but the Android support page also says:
"Some devices that launch with the Android 12 (2021) platform release using kernel versions v5.10 or higher are required to ship with the GKI kernel. "
https://source.android.com/devices/architecture/kernel/generic-kernel-image
https://source.android.com/devices/architecture/kernel/ion-gki
so this makes it sound like it is also possible that an Android S/12 device could still be using libion.so (if it is not part of "some" devices)

Android P,Q,R (9,10,11):

libion.so first appears in Android P, kernel 4.14 (due to changes from 4.12).  The ioctl support (older/newer) is therefore not needed for Android Q or newer.
https://source.android.com/devices/architecture/kernel/ion_abi_changes
This page also says "Kernel 4.12 heavily refactored the Ion kernel code" which is the reason for use_newer_ioctl.

The use_newer_ioctl code was originally added in 2018-07-23 which is before the official release date of Android P (2018-08-06).  The older ioctl support (added in 2016-07-02) was what was in use up until kernel 4.12.  Android P can actually have kernels as far back as 4.4:
https://source.android.com/devices/architecture/kernel/android-common

So, if we want to continue supporting Android P, we need to keep both the older & newer ioctl support. 

Android O (8):

The older ioctl support (added in 2016-07-02) was what was in use for Android O.

dsharletg (approved):

src/runtime/hexagon_remote/halide_remote.cpp
```diff
@@ -94,14 +94,51 @@ int halide_hexagon_remote_load_library(const char *soname, int sonameLen,
 
 volatile int power_ref_count = 0;
 
+// HAP power_context
+//  - has to be non-NULL (for devices newer than SM8350)
+//  - only has to be unique within the process
+//  - the context is just an identifier
+//  - even though it is a void*, it is never stored into
+//  - for managing multiple clients, could just malloc(1)/free contexts
+//  - could even just use "(void*)1" (non-NULL), but that would not be
```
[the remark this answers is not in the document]
dpalermo: They are asking us not to use a constant as there could be a problem if the app had multiple/different uses of Hexagon (e.g. Halide, ML, ...) and was linked with multiple/different skel files....each of which the power team wants to see using a different power context.  If every skel file just used a constant '1', then they couldn't distinguish the different Hexagon clients requests from each other.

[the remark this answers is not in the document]
dpalermo: Instead of using the address of a writable variable as the power context, we could instead use the read-only address of one of the other functions in halide_remote.cpp: 

```
static inline void *get_HAP_power_context() {
    // Halide runtime only needs a single power context
    return (void*)halide_hexagon_remote_load_library;
}
```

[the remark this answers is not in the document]
dsharletg: I see, using 1 could accidentally collide with another use. What you have in the PR currently is fine. I think this is ready to merge now.

PR #6047, opened 2021-05-20
Don't emit aligned loads to unaligned addresses

### LINKED ISSUE #6046: Halide tries to do aligned loads from unaligned buffers
This happens with strided loads with stride 3 in a complex test case I have. Looking at the relevant portion of CodeGen_LLVM, the alignment tags on the dense load appear to be bunk.

This is a pretty bad bug, because it causes hard-to-diagnose crashes inside compiled Halide pipelines (unless you know what you're looking for).

abadams:
Fixes #6046

### DIFF
--- a/test/correctness/align_bounds.cpp
+++ b/test/correctness/align_bounds.cpp
@@ -214,12 +214,10 @@ int main(int argc, char **argv) {
         g(x) = e;
 
         f.compute_root();
-
         g.bound(x, 0, 1024).vectorize(x, 16, TailStrategy::RoundUp);
 
+        // Just check if it crashes
         g.realize({1024});
-
-        g.compile_to_assembly("/dev/stdout", {}, Target{"x86-64-linux-avx2-no_runtime-no_bounds_query-disable_llvm_loop_opt-no_asserts"});
     }
 
     printf("Success!\n");
--- a/test/correctness/align_bounds.cpp
+++ b/test/correctness/align_bounds.cpp
@@ -197,6 +197,31 @@ int main(int argc, char **argv) {
         }
     }
 
+    // Try a case where aligning a buffer means that strided loads can
+    // do dense aligned loads and then shuffle. This used to trigger a
+    // bug in codegen.
+    {
+        Func f, g;
+        Var x;
+
+        f(x) = x;
+
+        // Do strided loads of every possible alignment
+        Expr e = 0;
+        for (int i = -32; i <= 32; i++) {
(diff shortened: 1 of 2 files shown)


dsharletg (approved):

alexreinking (approved):

abadams:
Currently checking if it solves the more complex case I had...

alexreinking:
Looks like arm64-linux-worker-1 might need to rebuild LLVM 13

abadams:
Yes, it seems to have fixed my real failure cases.

test/correctness/align_bounds.cpp
```diff
@@ -197,6 +197,29 @@ int main(int argc, char **argv) {
...
+    {
+        Func f, g;
+        Var x;
+
+        f(x) = x;
+
+        // Do strided loads of every possible alignment
+        Expr e = 0;
+        for (int i = -32; i <= 32; i++) {
+            e += f(3 * x + i);
```
dsharletg, line 212: I wonder if LLVM might rewrite this to one load plus a bunch of shuffles, which might hide alignment issues? Maybe not now, but it could later?

To avoid this, maybe it would be better to put a loop around the whole thing including the realize, loading one at a time.

alexreinking (approved):

alexreinking:
Not sure what's going on with x86-64-osx-make. @steven-johnson can you take a look?

test/correctness/align_bounds.cpp
steven-johnson: True, but if the test doesn't reliably test the thing we need testing, it doesn't matter how fast it is. (I guess if this works currently then it's ok, but might be worth capturing this thought in a comment if it breaks again.)

abadams: I also think the kind of transformation Dillon is talking about is vanishingly unlikely. We're codegenning three consecutive dense vector loads followed by a big shuffle. LLVM isn't going to change that to different loads followed by a different shuffle. The loads are already optimal.

abadams: If LLVM ever started doing that sort of transformation, we'd need to disable it anyway, because we have scheduling directives to do that and they would stop working.

steven-johnson:
> Not sure what's going on with x86-64-osx-make. @steven-johnson can you take a look?

I think I fixed it, but can't retest this specific build because branch is closed and deleted. Will retry with another.

PR #6055, opened 2021-05-24
Upgrade d8 version for wasm testing

steven-johnson:
We were using a variant of v8.9, but various late-breaking variations in the final spec implementation didn't make it into V8 until v9.1; using top-of-tree LLVM and EMCC require v9.1+ to avoid obscure errors.


shoaibkamil (approved):

steven-johnson:
(BTW: the reason this hasn't been failing on the buildbots is because they were running not-quite-top-of-tree Emscripten; I just updated them, and so we may see breakage until this lands)

alexreinking (approved):

PR #6066, opened 2021-05-28
[Hexagon] Makefile changes for hannk on Hexagon

aankit-ca:
Works on device.
Qurt crash on hexagon-sim.

@dsharlet @pranavb-ca Use this to test:
HL_TARGET=hexagon-32-qurt-hvx_128 make test-hexagon-device


apps/support/Makefile.inc
```diff
@@ -103,6 +106,8 @@ CXX-host-metal ?= $(CXX)
 CXX-host-hvx_128 ?= $(CXX)
 CXX-host-hvx_64 ?= $(CXX)
 CXX-host-hvx ?= $(CXX)
+CXX-hexagon-32-qurt-hvx_128 ?= hexagon-clang++
+CXX-hexagon-32-qurt-hvx_64 ?= $(CXX-hexagon-32-qurt-hvx_128)
```
dsharletg, line 110: I think we've been deprecating hvx_64, do we need this? (And same below)

apps/hannk/Makefile
```diff
@@ -6,26 +6,64 @@ include ../support/Makefile.inc
...
 	$(BIN)/$(HL_TARGET)/compare_vs_tflite
 
 test: compare_vs_tflite
 	$(BIN)/$(HL_TARGET)/compare_vs_tflite test/*/* --benchmark 0
 
+test-hexagon-sim: $(BIN)/$(HL_TARGET)/$(BENCHMARK_OUT)
+	@mkdir -p $@
+	echo "Benchmarking $(TEST)..."
+	echo "$(HEXAGON_SDK_ROOT)/rtos/qurt/computev65/debugger/lnx64/qurt_model.so" > $(BIN)/$(HL_TARGET)/osam.cfg
+	echo "$(DEFAULT_HEXAGON_TOOLS_ROOT)/Tools/lib/iss/qtimer.so --csr_base=0xFC900000 --irq_p=1 --freq=19200000 --cnttid=1" > $(BIN)/$(HL_TARGET)/q6ss.cfg
```
dsharletg, line 45: How stable are these specific files/constants? Are they going to break often?

I have this comment on many things in this PR... It seems like a lot of the config stuff happening in this PR should be built into the tools themselves, so we don't need to bake so many things that could potentially break into our build?

dsharletg (commented):
Thanks for this, very cool!

steven-johnson:
This PR must also include equivalent changes to the CMake build script (which is the canonical build script for this app and the only one that is tested on the buildbots).

apps/support/Makefile.inc
steven-johnson: hvx_64 doesn't exist anymore. hvx_128 is just a synonym for hvx (which is now preferred). Remove hvx_64 and add hvx.

steven-johnson:
> This PR must also include equivalent changes to the CMake build script (which is the canonical build script for this app and the only one that is tested on the buildbots).

(I'm happy to help make those changes once the rest of this PR is approved, BTW)

steven-johnson:
Where does this PR stand?

aankit-ca:
Hi Steven, I was OOO for a few days. I'm back now. I will update the PR soon with the requested changes.

apps/hannk/Makefile
aankit-ca: The constants are stable for the same target (same qurt ‘variant’, like ‘computev65’). The configs are not present in the tools. The sdk examples also do the same thing to run.

apps/support/Makefile.inc
```diff
@@ -103,6 +106,8 @@ CXX-host-metal ?= $(CXX)
 CXX-host-hvx_128 ?= $(CXX)
 CXX-host-hvx_64 ?= $(CXX)
 CXX-host-hvx ?= $(CXX)
+CXX-hexagon-32-qurt-hvx ?= hexagon-clang++
+CXX-hexagon-32-qurt-hvx_128 ?= $(CXX-hexagon-32-qurt-hvx)
```
dsharletg, line 110: I think we should not add this `hvx_128` version of these flags either? I think the only non-deprecated thing is `hvx` right?

apps/hannk/Makefile
```diff
@@ -6,26 +6,64 @@ include ../support/Makefile.inc
...
 CXXFLAGS += -Wno-unused-private-field -fno-exceptions -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -Wunused-variable -I$(MAKEFILE_DIR)
 
+BENCHMARK_OUT = benchmark
+ifeq (hexagon-32-qurt,$(findstring hexagon-32-qurt,$(HL_TARGET)))
+	# Building benchmark application as shared object instead of elf for
+	# hexagon target.
+	BENCHMARK_OUT := $(BENCHMARK_OUT).so
+	BENCHMARK_HEXAGON_FLAGS = -shared -fPIC -G0
+	HEXAGON_STUBS = $(BIN)/$(HL_TARGET)/stubs.o
+	TEST ?= test/misc/MUL.tflite
```
dsharletg, line 16: The benchmark app already runs every benchmark passed to it on the command line. Maybe we could push the whole folder, and run test/*/* like the non-Hexagon tests?

I think all of this logic should be in the test-hexagon-sim/test-hexagon-device target BTW, not stored in environment variables like TEST (that is way too generic).

aankit-ca: Hey Dillon, using `test/*/*` is not working for me. Now I realized that run_main_on_hexagon has a max-limit of 2048 characters.

aankit-ca: We can run all tests one at a time maybe?

dsharletg: I guess we have to. I think we should make all targets work this way though, not just Hexagon, so it is consistent.

apps/hannk/benchmark.cpp
```diff
@@ -41,7 +41,7 @@ void run_benchmark(const std::string &filename, const InterpreterOptions &option
 
 }  // namespace hannk
 
-int main(int argc, char **argv) {
+__attribute__((visibility("default"))) int main(int argc, char **argv) {
```
steven-johnson, line 44: Why is this change needed?

apps/hannk/interpreter/stubs.cpp
```diff
@@ -0,0 +1,19 @@
+#include "stdio.h"
```
steven-johnson, line 1: Please run clang-format on this.

steven-johnson: Also please add comments about what it's for, etc.

src/CPlusPlusMangle.cpp
```diff
@@ -533,7 +533,11 @@ std::string mangle_type(const Type &type, const Target &target, PrevPrefixes &pr
         case 16:
             return "s";
         case 32:
-            return "i";
+            if (target.arch == Target::Hexagon) {
```
steven-johnson, line 536: Can you comment on what's going on here (and why we never noticed the issue before)?

apps/hannk/benchmark.cpp
dsharletg: On Hexagon, this is getting compiled as a shared object, not an executable. So I'm guessing this makes main visible to dlsym.

apps/hannk/Makefile
```diff
@@ -6,25 +6,52 @@ include ../support/Makefile.inc
...
 	$(BIN)/$(HL_TARGET)/compare_vs_tflite
 
 test: compare_vs_tflite
-	$(BIN)/$(HL_TARGET)/compare_vs_tflite test/*/* --benchmark 0
+	$(foreach test_model, $(shell ls -1 test/*/*), $(BIN)/$(HL_TARGET)/compare_vs_tflite $(test_model) --benchmark 0;)
+
+test-hexagon-sim: $(BIN)/$(HL_TARGET)/$(BENCHMARK_OUT)
+	@mkdir -p $@
+	echo "Benchmarking tests..."
+	echo "$(HEXAGON_SDK_ROOT)/rtos/qurt/computev65/debugger/lnx64/qurt_model.so" > $(BIN)/$(HL_TARGET)/osam.cfg
```
dsharletg, line 42: It looks like these constants could go in the run_hexagon_sim.sh script too?

apps/hannk/interpreter/stubs.cpp
aankit-ca: Done

apps/hannk/benchmark.cpp
aankit-ca: Right

apps/hannk/Makefile
aankit-ca: Done

src/CPlusPlusMangle.cpp
aankit-ca: We never tested the name-mangling target feature for Hexagon and hence never saw the problem. I think the changes are needed because of Hexagon being a 32bit proc.

dsharletg: If this is for 32-bit, shouldn't the condition be if (target.bits == 32) ?

But I think we test this code on other 32-bit targets, so I think this might be something else.

dsharletg: @zvookin might know more

dsharletg (approved):
This looks pretty good to me, only some minor comments remaining. However, there are now conflicts in the Makefile.

apps/hannk/Makefile
aankit-ca: Sorry forgot to do these.

src/CPlusPlusMangle.cpp
aankit-ca: @dsharletg  The headers have int32 defined as long and uint32 as unsigned long. That is why this change is needed on Hexagon.

dsharletg (approved):
Thanks for this, and for the iteration on the review!

PR #6069, opened 2021-06-03
Fix dubious "gather" intrinsic for hvx

steven-johnson:
I'm not sure if this is a bug (per se) or not, but: We define an intrinsic for `Call::hvx_gather`, and at several points check for `is_intrinsic(Call::hvx_gather)`, but we never actually create such a Call. Instead, `make_gather` just uses the naked string `"gather"`, which is not the same thing. How is this working? (Is it working?)

Opening this as a PR to gather input (no pun intended) about what's going on here.


abadams:
I think I did the rename from "gather" to Call::hvx_gather. I must have somehow missed this one when that happened (maybe a merge conflict).

abadams (approved):

Issue #6067, opened 2021-06-02
Well-known Call nodes should use well-defined CallTypes (and not provide them via arg to Call::make)


steven-johnson:
Every well-known Call (i.e. everything named under Intrinsic::Call) has a single well-defined value for CallType, so specifying them via Call::make is just an opportunity to get it wrong. We should instead make this typo-proof somehow -- pair them in a table somehow to ensure that the right values are used? Make inline wrappers to construct all the well-known Calls? Something else?

PR #6070, opened 2021-06-03
Convert some Intrinsic calls to PureIntrinsic

steven-johnson:
While investigating https://github.com/halide/Halide/issues/6067, I noticed that a handful of intrinsics that should probably be called as PureIntrinsic are being called as Intrinsic, thus missing out on hoisting opportunities:
- div_round_to_zero
- mod_round_to_zero
- stringify
- mux
- if_then_else (but only sometimes)

(Note that Call::size_of_halide_buffer_t should *not* be called as pure; the only use of it is typically `alloca()`, which already special-cases it efficiently, and hoisting it into a var would just require more fiddling at codegen time, for likely no gain.)

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -58,7 +58,7 @@ Expr stringify(const std::vector<Expr> &args) {
     }
 
     return Internal::Call::make(type_of<const char *>(), Internal::Call::stringify,
-                                args, Internal::Call::Intrinsic);
+                                args, Internal::Call::PureIntrinsic);
 }
 
 Expr combine_strings(const std::vector<Expr> &args) {
@@ -1499,7 +1499,7 @@ Expr mux(const Expr &id, const std::vector<Expr> &values) {
     }
     std::vector<Expr> result{id};
     result.insert(result.end(), values.begin(), values.end());
-    return Internal::Call::make(t, Internal::Call::mux, result, Internal::Call::Intrinsic);
+    return Internal::Call::make(t, Internal::Call::mux, result, Internal::Call::PureIntrinsic);
 }
 
 Expr mux(const Expr &id, const Tuple &tup) {
@@ -1520,7 +1520,7 @@ Expr unsafe_promise_clamped(const Expr &value, const Expr &min, const Expr &max)
     return Internal::Call::make(value.type(),
                                 Internal::Call::unsafe_promise_clamped,
                                 {value, n_min_val, n_max_val},
-                                Internal::Call::Intrinsic);
+                                Internal::Call::PureIntrinsic);
 }
 
 namespace Internal {
@@ -1533,7 +1533,7 @@ Expr promise_clamped(const Expr &value, const Expr &min, const Expr &max) {
     return Internal::Call::make(value.type(),
                                 Internal::Call::promise_clamped,
                                 {value, n_min_val, n_max_val},
-                                Internal::Call::Intrinsic);
+                                Internal::Call::PureIntrinsic);
 }
 }  // namespace Internal
 
@@ -2509,7 +2509,7 @@ Expr div_round_to_zero(Expr x, Expr y) {
     Type t = x.type();
     return Internal::Call::make(t, Internal::Call::div_round_to_zero,
                                 {std::move(x), std::move(y)},
-                                Internal::Call::Intrinsic);
+                                Internal::Call::PureIntrinsic);
 }
 
 Expr mod_round_to_zero(Expr x, Expr y) {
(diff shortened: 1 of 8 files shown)


src/Simplify_Call.cpp
```diff
@@ -649,7 +649,7 @@ Expr Simplify::visit(const Call *op, ExprInfo *bounds) {
                 return Internal::Call::make(op->type,
                                             Call::if_then_else,
                                             {std::move(cond_value), std::move(true_value), std::move(false_value)},
-                                            op->call_type);
```
abadams, line 652: Isn't preserving the existing call type, whatever it is, better than asserting a new one?

It makes for fewer places to update if we ever change the call type again.

src/IROperator.cpp
```diff
@@ -1533,7 +1533,7 @@ Expr promise_clamped(const Expr &value, const Expr &min, const Expr &max) {
     return Internal::Call::make(value.type(),
```
abadams, line 1533: The meaning of promise_clamped is location-sensitive. You can't safely move that piece of IR around, so it's not pure.

src/IROperator.cpp
```diff
@@ -1520,7 +1520,7 @@ Expr unsafe_promise_clamped(const Expr &value, const Expr &min, const Expr &max)
...
```
abadams, line 1520: This one isn't pure, for the same reason as below.

src/Simplify_Call.cpp
steven-johnson: Well, maybe? I made this change on the assumption that all of the calls to each intrinsic should be the same call type (#6067), and as of this PR, all of the values of call_type that we pass on here *should* be PureIntrinsic anyway. Happy to change it (back) but not sure it matters much.

src/IROperator.cpp
steven-johnson: Reverting, but I'm having a hard time envisioning an example where that would be the case. Can you provide an example for posterity here?

dsharletg (commented):
Should we be fixing this by adding helpers (like those in IROperator.h) to make calls, instead of making them directly?

We already have div_round_to_zero and mod_round_to_zero, maybe others too.

A difference is that those helpers validate and adjust types if necessary, but I suspect if those change anything, that is uncovering a bug.

src/IROperator.cpp
abadams:
```
for (int x = 0; x < 10; x++) {
  if (2*x < 10) {
    f(promise_clamped(x, 0, 4));
  }
}
```

steven-johnson:
> Should we be fixing this by adding helpers (like those in IROperator.h) to make calls, instead of making them directly?

Yeah, that's one of the options I suggested in #6067, and for these cases it's probably a better option. But does that suggest we should add specific wrappers for *every* intrinsic? That would be a lot of churn, but maybe the code would be cleaner as a result.

src/IROperator.cpp
abadams: In the example above, I'm not sure if it's safe to move the x outside the if

Then again... hrm. The only users should be inside the if, so how could it matter...

abadams: I guess the implication is that even evaluating a promise clamped, where the arg is not actually clamped, is undefined.

abadams: @dsharletg Maybe promise_clamped is actually pure? What do you think?

steven-johnson:
> We already have div_round_to_zero and mod_round_to_zero, maybe others too.

Yes, but note this counterexample in `lower_euclidean_div`:

```
        // IROperator's div_round_to_zero will replace this with a / b for
        // unsigned ops, so create the intrinsic directly.
```

src/IROperator.cpp
dsharletg: I think it is pure technically, but I think maybe you are thinking of uses of promise_clamped that would break if it were moved (to a let)? So while it is in fact pure, we mark it not pure to avoid this problem?

dsharletg: By uses, I mean consumers of promise_clamped's information, not the code that generates promise_clamped.

dsharletg:
Maybe whatever is sensitive to that (replacing div_round_to_zero with divide for unsigned) should be fixed so it doesn't matter if that happens.

steven-johnson:
Buildbots green, now the question is whether this change is good enough as-is (irrespective of future changes)

steven-johnson:
Monday morning review ping

abadams (approved):

PR #6071, opened 2021-06-03
Support cross-compiling libHalide

alexreinking:
**WIP, doesn't even do what it says in the title yet. Just keeping this PR / branch open for people to discuss.**

In order to cross-compile Halide and the autoschedulers for a different target, a host version of Halide must exist. This requires building Halide twice: once for the host and again for the target.

Thus, users who wish to cross-compile will have to invoke the build twice. This is also how LLVM handles cross compiling: you build `llvm-tblgen` and `clang-tblgen` for the host and then point the target build to the host tools. Naturally, LLVM screws up the CMake implementation (setting environment variables, bleh), but we won't make the same mistake here.

The workflow will be:

```
# Configure the host build
$ cmake -S Halide -B HalideHost

# Run the build up to the HalideHost meta-target (optional; can of course build everything)
$ cmake --build HalideHost --target HalideHost

# Configure the build for the target. Point it to the host build folder.
$ cmake -S Halide -B Halide-target \
        -DCMAKE_TOOLCHAIN_FILE=/path/to/toolchain-target.cmake \
        -DHalideHost_ROOT=$PWD/HalideHost

# Run the target build
$ cmake --build Halide-target
```


steven-johnson:
> The workflow will be:

Would it be useful to add a helper script to do this dance, both for ease of use and also to record the concrete steps needed?

alexreinking:
> Would it be useful to add a helper script to do this dance, both for ease of use and also to record the concrete steps needed?

Only if we're actually going to use it somewhere. Otherwise, I think it should just be added to the README.

steven-johnson:

alexreinking:
I think I need to take another crack at this with a fresh set of eyes. I'll close & delete.

PR #6072, opened 2021-06-03
Changes for building with Hexagon SDK 4.3.0.0 & android-ndk-r19c

dpalermo:


src/runtime/hexagon_remote/known_symbols.cpp
```diff
@@ -20,61 +20,61 @@ void *lookup_symbol(const char *name, const known_symbol *map) {
 extern "C" {
 
 // More symbols we need to support.
-extern int qurt_hvx_lock;
```
dsharletg, line 23: Why does this need to change? It's fine, just wondering why, and if it is unnecessary churn. The overload resolution casting below is also a result of this.

dsharletg (commented):
When I installed the SDK, I got 4.3.0, which is probably what is going to go on the build bots. Would it be possible to make this PR target that SDK version?

src/runtime/hexagon_remote/known_symbols.cpp
dpalermo: The previous definitions caused mismatch errors with the headers in the newer SDKs (likely starting with 4.0.0)...they were updated to match.

dpalermo:
> When I installed the SDK, I got 4.3.0, which is probably what is going to go on the build bots. Would it be possible to make this PR target that SDK version?

That's great...I wasn't sure if 4.3.0 was available externally yet.  I've updated this PR to match.

src/runtime/hexagon_remote/known_symbols.cpp
dsharletg: What is the error you are seeing?

dpalermo: Any of the symbols that are actually functions report errors that look like:

```
known_symbols.cpp:35:12: error: redefinition of 'sqrtf' as different kind of symbol
extern int sqrtf;
           ^
/local/scratch/dpalermo/TSDK/Hexagon_SDK/4.3.0.0/tools/HEXAGON_Tools/8.4.11/Tools/bin/../target/hexagon/include/math.h:484:7: note: previous definition is here
float sqrtf(float) _NO_THROW;
      ^
```
The history on this files shows we had to make this change internally once we started using Hexagon tools 8.4.x.  Looks like improved/tighter error checking in clang++.

dsharletg: This would have been an error at any time in clang I think. I think the problem is a newer SDK started including these symbols transitively from the headers we include here when they weren't included before. But that's reason enough to change it I guess.

src/runtime/hexagon_remote/known_symbols.cpp
```diff
@@ -121,47 +121,47 @@ void *get_known_symbol(const char *name) {
         {"__hexagon_modsi3", (char *)(&__hexagon_modsi3)},
         {"__hexagon_sqrtf", (char *)(&sqrtf)},
         {"sqrtf", (char *)(&sqrtf)},
-        {"sqrt", (char *)(&sqrt)},
+        {"sqrt", (char *)(static_cast<double (*)(double)>(&sqrt))},
```
steven-johnson, line 124: Why are all these casts necessary?

steven-johnson:
(Marked as 'skip_buildbots' for now, because the buildbots only build/test hannk via CMake, not Make, so these changes won't affect much until the CMake build scripts for apps/hannk is updated too)

dsharletg:
There are other hexagon tests too that are affected. I'm actually very surprised this worked because the 4.3.0 SDK isn't installed on the buildbots yet. Maybe it was still using the old runtime/simulator environment vars.

src/runtime/hexagon_remote/known_symbols.cpp
dsharletg: I'm guessing the newer Hexagon SDK headers are apparently including math.h transitively and were not before, and so now sqrt is an overloaded symbol and needs to be resolved to the right thing. sqrtf isn't overloaded so it doesn't need the cast.

dpalermo:
> There are other hexagon tests too that are affected. I'm actually very surprised this worked because the 4.3.0 SDK isn't installed on the buildbots yet. Maybe it was still using the old runtime/simulator environment vars.

All of the Makefile changes that refer to the SDK are in hexagon_remote, which I don't believe your buildbot builds.  The top level Makefile changes are just for the install_qc target (which the buildbots also probably aren't exercising).  The results of these changes still need to be built and checked in by hand at some point.  At this point, I didn't push the results of the build to keep the number of files under review just focused on the initial changes that needed to be made.

src/runtime/hexagon_remote/known_symbols.cpp
dpalermo: Changing the above:
```
-        {"sqrt", (char *)(static_cast<double (*)(double)>(&sqrt))},
+        {"sqrt", (char *)(&sqrt)},
```
and rebuilding, results in:

```
known_symbols.cpp:124:28: error: address of overloaded function 'sqrt' does not match required type 'char'
        {"sqrt", (char *)(&sqrt)},
                           ^~~~
known_symbols.cpp:36:15: note: candidate function
extern double sqrt(double);
              ^
/local/scratch/dpalermo/TSDK/Hexagon_SDK/4.3.0.0/tools/HEXAGON_Tools/8.4.11/Tools/bin/../target/hexagon/include/c++/v1/math.h:1075:46: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long double sqrt(long double __lcpp_x) _NOEXCEPT {return ::sqrtl(__lcpp_x);}
                                             ^
/local/scratch/dpalermo/TSDK/Hexagon_SDK/4.3.0.0/tools/HEXAGON_Tools/8.4.11/Tools/bin/../target/hexagon/include/c++/v1/math.h:1074:46: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY float       sqrt(float __lcpp_x) _NOEXCEPT       {return ::sqrtf(__lcpp_x);}
                                             ^
/local/scratch/dpalermo/TSDK/Hexagon_SDK/4.3.0.0/tools/HEXAGON_Tools/8.4.11/Tools/bin/../target/hexagon/include/c++/v1/math.h:1081:1: note: candidate template ignored: could not match 'typename std::enable_if<std::is_integral<_A1>::value, double>::type (_A1) noexcept' against 'char'
sqrt(_A1 __lcpp_x) _NOEXCEPT {return ::sqrt((double)__lcpp_x);}
^
1 error generated.
```

dsharletg: FWIW transitively including math.h from standard headers might be worth looking into if that is happening. If that's not standard behavior, it could cause problems, and unnecessarily parsing/dead code stripping math.h on *everything* compiled for hexagon is unfortunate.

I don't think any of these headers should be including math.h:
- https://github.com/halide/Halide/blob/master/src/runtime/HalideRuntime.h#L5-L14
- https://github.com/halide/Halide/blob/master/src/runtime/hexagon_remote/known_symbols.cpp#L3-L7

I don't see where else it could be coming from.

dpalermo: Adding -E to the compile line shows that it was included from stdlib.h.

```
/local/scratch/dpalermo/TSDK/Hexagon_SDK/4.3.0.0/tools/HEXAGON_Tools/8.4.11/Tools/bin/../target/hexagon/include/c++/v1/stdlib.h:
...
#ifdef __cplusplus
#include <math.h>
#endif  // __cplusplus

#endif  // _LIBCPP_STDLIB_H
```

dpalermo: This is coming from LLVM libcxx/include/stdlib.h.   LLVM 10 has the include math.h:

     https://github.com/llvm/llvm-project/blob/release/10.x/libcxx/include/stdlib.h

dpalermo: The include math.h goes away in LLVM 11:

    https://github.com/llvm/llvm-project/blob/release/11.x/libcxx/include/stdlib.h

and it looks like they swtiched to builtins for what is needed by stdlib.h:

```
inline _LIBCPP_INLINE_VISIBILITY float abs(float __lcpp_x) _NOEXCEPT {
  return __builtin_fabsf(__lcpp_x); // Use builtins to prevent needing math.h
}
```

dsharletg:
This change looks good to me, but I think we need to try building + running the new runtime on the buildbots

dpalermo:
> This change looks good to me, but I think we need to try building + running the new runtime on the buildbots

Sounds good.  Want me to push up the results of building hexagon_remote?

steven-johnson:
I'm having trouble installing SDK 4.3 on our buildbots (Ubuntu 20.4 LTS). AFAICT I have to install the Qualcomm Package Manager to install the SDK, but when I do so, it fails in the same way as reported [here](https://developer.qualcomm.com/forum/qdn-forums/software/hexagon-dsp-sdk/toolsinstallation/68318). Before I try the workaround suggested at that link, thought I'd check with the QC folks here to see if there will be other issues with installing it on Ubuntu 20 if it's not officially supported.

(Downgrading our buildbots back to Ubuntu 18 is a nonstarter, so we really need some assurance here :-)

dpalermo:
Interesting, I was not aware QPM had issues on 20.04.  I will pass that on to the QPM team to get that resolved.  In the meantime, QPM should only be needed to download the SDK initially, you shouldn't have to run QPM on the 20.04 buildbots.  If you have an 18.04 system that you can run QPM on, you could download the SDK there  (I believe Dillon was able to install it on his system).  Once you have the SDK you should be able to just copy the SDK tree itself to wherever it needs to reside.

dpalermo:
Mail to internal QPM team sent.  I'll let you know what I hear back on that workaround/ability to use QPM on 20.04.

steven-johnson:
> (I believe Dillon was able to install it on his system)

Cool, I'll just do that that will be simpler anyway :-)

steven-johnson:
I have the 4.3 SDK on the buildbots now, but they aren't marked as the current SDK (we still default to 3.3.3 for testing). LMK if/when I should update the bots to actually *use* the new one.

dpalermo:
The QPM team says QPM is supported on Ubuntu 20.04.   They also said that installing libgnome-keyring0 is the recommended fix for the missing dependency.

steven-johnson:
> The QPM team says QPM is supported on Ubuntu 20.04

May I gently suggest that installation failing with a cryptic error message isn't an ideal form of support :-)

dpalermo:
I agree...and have relayed that sentiment ;-)

steven-johnson:
Where does this PR stand?

dpalermo:
The only mentions of bin/v62 I could find were in README.md.  Will watch the buildbot to see if there is any breakage.

steven-johnson:
> @steven-johnson I think we need to enable the new SDK on the build bots for this test to be meaningful on the build bots.

Done. I re-synced this PR to master to force a rebuild.

steven-johnson:
Lots of failures: https://buildbot.halide-lang.org/master/#/builders/126/builds/421

dpalermo:
Looks to me like HL_HEXAGON_SIM_REMOTE is being set in the buildbot script itself:

https://buildbot.halide-lang.org/master/#/builders/126/builds/421
```
 Test correctness, generator Halide_TARGET=host-hvx
 ...
   HL_HEXAGON_SIM_REMOTE=/home/halidenightly/build_bot/worker/halide-testbranch-main-llvm13-x86-64-linux-cmake/halide-source/src/runtime/hexagon_remote/bin/v62/hexagon_sim_remote
```
This could be made conditional on the use of SDK 4.3.0 (which only supports HVX >=v65)
if using SDK 4.3.0, set HL_HEXAGON_SIM_REMOTE to .../bin/v65/hexagon_sim_remote, else use bin/v62.

dpalermo:
...and just to verify before we go too much further, it is true that you are only pushing updates to Halide Hexagon apps out to Pixel 3 (v65) and newer devices at this point?  (as Pixel 2 is v62 and won't be able to use a v65 built Halide Hexagon runtime)

steven-johnson:
> This could be made conditional on the use of SDK 4.3.0 (which only supports HVX >=v65)

For now, I'll just update it to use v65 unconditionally, then restart the failing tests.

steven-johnson:
Looks like the linux builds are passing.

dpalermo:
Resolved conflict (picked up https://github.com/halide/Halide/pull/6085).
Rebuilt affected binaries with SDK 4.3.0 after conflict resolved.

dpalermo:
Failures look unrelated...both in arm-64-osx

steven-johnson:
> Failures look unrelated...both in arm-64-osx

Yeah, though still weird and disturbing. (Maybe an LLVM injection? Someone should bisect)

src/runtime/hexagon_remote/Makefile
```diff
@@ -24,13 +24,17 @@
 #
 
 SDK_ROOT ?= ${HOME}/Qualcomm
-HEXAGON_SDK_ROOT ?= ${SDK_ROOT}/Hexagon_SDK/3.3.3
-HEXAGON_TOOLS_ROOT ?= ${HEXAGON_SDK_ROOT}/tools/HEXAGON_Tools/8.1.05
+HEXAGON_SDK_ROOT ?= ${SDK_ROOT}/Hexagon_SDK/4.3.0.0
+# Need SDK >= 4.2.0.2 for qaic "halide_hexagon_remote_skel_invoke_qaic_version"
```
dsharletg, line 28: What is this referring to? I only see one instance of this string in this PR (here).

dpalermo: That is a symbol that is automatically generated by the newer qaic compiler.  It can be seen in the generated skel source:

[vm-dpalermo-aus Halide/src/runtime/hexagon_remote/bin/src]$ grep ver *skel.c
__attribute__ ((visibility("default"))) uint32_t halide_hexagon_remote_skel_invoke_qaic_version = 10041;

The purpose of it is to allow the DSP dispatch to be able to verify the skel shared object was built with a "new enough" version of qaic.  This was done so it is possible to determine at runtime if a given skel file was generated with a qaic that contained fixes for all critical security fixes (such as the qaic in SDK 4.2.0.2 contains).

Also, the dispatch logic for all devices >sm8350 will only run skel shared objects that have this symbol present and has a value >=10041 (but this won't affect existing Pixel devices).  Older devices (Pixel 3*,4*,5) with the current dispatch obviously won't even be looking for this symbol....but by using the latest qaic, will still have the security fixes...even though they won't be checking for them.

dsharletg: I think this comment might be a bit out of place. I think it should either be removed or elaborated on. I assume this is documented in the Hexagon SDK, perhaps we should just refer to that documentation here?

dpalermo: I don't see anything in the SDK docs that goes into the above detail.
 
How about I change the comment to:
`# Need SDK >= 4.2.0.2 for qaic version 01.00.41 for critical security fixes`

Or I can just remove the comment.

Let me know what you think.

steven-johnson: Adding a comment that `halide_hexagon_remote_skel_invoke_qaic_version` is autogenerated by qaic is essential, I think

dsharletg: Why? How does it help the user? 

From dpalermo's comment above, it seems like the thing that really should be commented here is this:

# Devices >sm8350 will only run skel shared objects that have been built with SDK version 4.2.02 or newer.

But really, this should be documented in more detail in the Hexagon SDK, and we should be referring to that documentation instead.

dpalermo: I think the reason I'm not finding anything specific in the docs about this is that the latest device described in Hexagon_SDK/4.3.0.0/docs is the SM8350.  More likely will be documented in the next SDK release.

dsharletg (approved):

PR #6076, opened 2021-06-04
Mark various HVX intrinsics as "pure"

steven-johnson:
While investigating #6067, I noticed that the four HVX-specific intrinsics are being called as nonpure; I don't know the exact semantics of these, but I suspect that at least some of them could be marked as pure, which would improve hoisting possibilities. Opening this change as a draft PR to get input from folks who have more knowledge of these instructions than I.

### DIFF
--- a/src/HexagonOptimize.cpp
+++ b/src/HexagonOptimize.cpp
@@ -2241,7 +2241,7 @@ class ScatterGatherGenerator : public IRMutator {
         dst_index = mutate(dst_index);
 
         return Call::make(ty, Call::hvx_gather, {std::move(dst_base), dst_index, src, size - 1, new_index},
-                          Call::Intrinsic);
+                          Call::PureIntrinsic);
     }
 
     // Checks if the Store node can be replaced with a scatter_accumulate.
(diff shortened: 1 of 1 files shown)


pranavb-ca:
Thanks @steven-johnson. @aankit-ca  - PTAL. IMHO, all except scatter-release can be marked Pure. But, do take a look once.

aankit-ca:
All scatter, gather and scatter-release have side effects. Scatters and gathers change the program state by writing to memory. Scatter-release have the effect of stalling the pipeline upon next load in the program execution. So I don't think we can mark any of these calls with PureIntrinsic.

steven-johnson:
@aankit-ca Perfect, thanks for letting me know unambigously! I'll drop this PR (and in a future PR I'll add comments to that effect so that no one else makes the same mistake as me)

dsharletg:
@aankit-ca scatter is obviously not pure, but does gather really write to memory?

dsharletg:
Oh, I see, HVX gathers actually write to TCM, they don't gather into a register.

aankit-ca:
> Oh, I see, HVX gathers actually write to TCM, they don't gather into a register.

Right

PR #6077, opened 2021-06-07
Remove large_buffers flag from HANNK

dsharletg:
This improved performance at one point, but it really shouldn't, and it seems like it doesn't help now.

More generally, there is no reason for this to improve performance. So even if it does improve performance, I don't think we should do it. LLVM might change and make the large_buffers version slower, and it would take us a long time to notice.


abadams (approved):

steven-johnson (approved):

steven-johnson (commented):
Ack, I should have pointed out that we need to make this change for CMake too. (After all, CMake is the only one we actually test...)

PR #6078, opened 2021-06-08
Provide bounds of rvars for all functions in the fused group

### LINKED ISSUE #6075: compute_with allocates and computes too much
See https://github.com/halide/Halide/discussions/6064 for detailed description of the issue.

vksnk:
Fixes #6075.


test/correctness/compute_with.cpp
```diff
@@ -2024,6 +2025,91 @@ int store_at_different_levels_test() {
...
+                exit(-1);
+            }
+            return IRMutator::visit(op);
+        }
+    };
+
+    total_sum.add_custom_lowering_pass(new CheckAllocationSize());
+
+    Buffer<int16_t> in(32, 64);
+    in.fill([](int x, int y) { return 1; });
```
steven-johnson, line 2101: nit: you can just do `in.fill(1);` here

steven-johnson:
LGTM but I'm not an expert in this code

test/correctness/compute_with.cpp
vksnk: Done.

vksnk:
Wednesday morning review ping.

src/BoundsInference.cpp
```diff
@@ -1124,15 +1129,12 @@ class BoundsInference : public IRMutator {
                 }
             }
 
-            if (fused_with_f.empty()) {
+            if (fused_with_f.size() == 1) {
```
abadams, line 1132: Could you explain this change?

src/BoundsInference.cpp
```diff
@@ -1200,44 +1202,53 @@ class BoundsInference : public IRMutator {
...
-                    if (in.is_bounded()) {
-                        // bounds_of_inner_var doesn't understand
-                        // GuardWithIf, but we know split rvars never
-                        // have inner bounds that exceed the outer
-                        // ones.
-                        if (!s.rvars.empty()) {
-                            in.min = max(in.min, Variable::make(Int(32), var + ".min"));
-                            in.max = min(in.max, Variable::make(Int(32), var + ".max"));
+                for (const auto &fused : fused_with_f) {
+                    size_t si = 0;
```
abadams, line 1206: Please add a comment what stage index you're getting and why you need to do a search to find it. It's not clear to me what's happening.

src/BoundsInference.cpp
vksnk: Previously, ``` fused_with_f ``` contained a list of stages fused with the current, but not the current stage itself. In this PR, I changed it to contain current stage as well, because it seems to simplify some of the processing. Also, I guess the name of the set is not accurate anymore (unless we consider f to be fused with itself, I guess), so I renamed it to just ``` fused_group ```.

src/BoundsInference.cpp
vksnk: Added a comment, the logic there is the same as in the loop above where we add bounds for every function from the fused  group -- here it's the same except that it's for bounds of reduction variables.

abadams (approved):

Issue #6081, opened 2021-06-08
is_monotonic incorrectly reporting Constant


rootjalex:
Discovered that `is_monotonic` reports `Constant` for the `z` variable of this expression:
```
Expr expr = ((min(x + 2, y) - select(0 < z, min(x, y) + 4, x)) + 4);
```
It should be `Unknown`, because the value is not constant as `z` crosses `0`.

Might be relevant for @dsharletg

rootjalex:
A smaller example that highlights the same issue is:
```
Expr expr = select(0 < z, x, y);
is_monotonic(expr, "z"); -> Constant
```

PR #6083, opened 2021-06-09
Fix is_monotonic issue (#6081)

dsharletg:
This seems like a hack that is likely missing other problematic cases, but I am not sure what the right thing here is...


rootjalex (approved):
Lgtm as a temporary fix, but I'm also not quite sure what the right thing to do here is.

src/Monotonic.cpp
```diff
@@ -443,7 +443,12 @@ class DerivativeBounds : public IRVisitor {
             } else {
                 delta.accept(this);
                 ConstantInterval rdelta = result;
-                adjusted_delta = multiply(rcond, rdelta);
+                // If rdelta is constant, we can't trust it, because if it were truly
+                // constant, find_constant_bounds of the delta would have produced a
+                // bounded interval.
+                if (!is_constant(rdelta)) {
```
abadams, line 449: The obvious thing to check is is_constant(rcond), why not that?

src/Monotonic.cpp
```diff
@@ -443,7 +443,12 @@ class DerivativeBounds : public IRVisitor {
             } else {
                 delta.accept(this);
                 ConstantInterval rdelta = result;
-                adjusted_delta = multiply(rcond, rdelta);
+                // If rdelta is constant, we can't trust it, because if it were truly
```
steven-johnson, line 446: Given the hack/temporary nature of the fix, we should open a tracking issue and add a TODO comment with link to it here

steven-johnson (approved):

src/Monotonic.cpp
steven-johnson: > I don't actually know it is a temporary hack. It might actually be correct.

Well, ok, but "This seems like a hack" and "Lgtm as a temporary fix" were in the comments just above...

src/Monotonic.cpp
dsharletg: Are you saying this is overly conservative in the case that rcond is constant? That this if should be `is_constant(rcond) || !is_constant(rdelta)` ?

abadams: No, it's that I don't understand why !is_constant(rdelta) is correct. I can see that there's a failing case where it's constant, but I don't see why this captures all the failing cases.

dsharletg: Can you propose something specific? Otherwise I am not sure how to go forward here.

I can't think of any cases that fail with this. While this seems like maybe a hack, I think it might actually be correct. In order for something to slip through here, it would need to be unbounded but also not constant, and still produce an incorrect rdelta somehow.

abadams (approved):

PR #6085, opened 2021-06-09
Add more ways for DMA-BUF to fallback to libion.so

dpalermo:
- Try to access libdmabufheap.so, if it succeeds try using DMA-BUF
  - If there are any errors seen with DMA-BUF, fallback to libion.so


dsharletg (approved):

dsharletg:
Going to merge this despite the failures. The changes should not affect the build bots anyways, and I want a commit of this before we update the SDK to 4.3.0.

PR #6086, opened 2021-06-10
Update LoopNest.cpp

NewProggie:
I believe this assert is not necessary anymore. It also prohibits the Adams auto scheduler to be used in more complex scenarios as this is a pretty hard restriction.
However, I am not that confident. Hence, I'd kindly like to ask @benoitsteiner to look over this as he was the original author of this.

### DIFF
--- a/src/autoschedulers/adams2019/LoopNest.cpp
+++ b/src/autoschedulers/adams2019/LoopNest.cpp
@@ -1041,7 +1041,6 @@ const Bound &LoopNest::get_bounds(const FunctionDAG::Node *f) const {
 
     // Compute the region required
     if (f->is_output && is_root()) {
-        internal_assert(f->outgoing_edges.empty()) << "Outputs that access other outputs not yet supported\n";
         // It's an output. Use the bounds estimate.
         for (int i = 0; i < f->dimensions; i++) {
             bound->region_required(i) = f->estimated_region_required[i];


abadams:
I think I added that assert, Benoit may have just moved it in a refactor.

Now that I think about it, it may indeed be fine to remove it. You say it seems to work ok without it?

benoitsteiner:
There was an issue 2 years ago when scheduling outputs that refer to other outputs. The underlying problem may have been fixed since, so it's probably fine to remove the assertion.

NewProggie:
> Now that I think about it, it may indeed be fine to remove it. You say it seems to work ok without it?

I did test this change with my own code and it produced the same output as before. However, there were some failing tests before *and* after applying the fix from this PR (the same ones) so I didn't bother. Namely these were:

	 41 - correctness_boundary_conditions (Bus error)
	229 - correctness_mul_div_mod (Bus error)
	298 - correctness_simd_op_check (SEGFAULT)
	326 - correctness_thread_safety (SEGFAULT)
	351 - correctness_vector_cast (Bus error)
	353 - correctness_vector_math (Bus error)
	489 - performance_thread_safe_jit (Bus error)
	534 - python_apps_bilateral_grid (Failed)
	535 - python_apps_blur (Failed)
	536 - python_apps_erode (Failed)
	537 - python_apps_interpolate (Failed)
	538 - python_apps_local_laplacian (Failed)
	560 - python_tutorial_lesson_02_input_image (Failed)
	565 - python_tutorial_lesson_07_multi_stage_pipelines (Failed)
	567 - python_tutorial_lesson_09_update_definitions (Failed)
	571 - python_tutorial_lesson_12_using_the_gpu (Failed)

> There was an issue 2 years ago when scheduling outputs that refer to other outputs. The underlying problem may have been fixed since, so it's probably fine to remove the assertion.
> We avoid the issue by rewriting a pipeline with outputs O1 and O2 such that O1 = f(x) and O2 = g(O1) as y = f(x), O1 = y, O2 = g(y).

This is interesting as I did try the same before and still triggered the internal_assert. I figured that Halide did look through this temporary copy and optimized that away. Can you point me to a public example where this is working for you?

abadams (approved):

PR #6087, opened 2021-06-11
Handle negative shifts in CodeGen_C

vksnk:


dsharletg (approved):

steven-johnson:
Some opencl failures -- related?

vksnk:
> Some opencl failures -- related?

yeah, it's a real failure. It took me a while to reproduce locally -- for some reason it only happens in cmake build. Looking into it now.

vksnk:
> > Some opencl failures -- related?
> 
> yeah, it's a real failure. It took me a while to reproduce locally -- for some reason it only happens in cmake build. Looking into it now.

Apparently, OpenCL codegen for the case of `int` >> `unsigned int` will cast the second argument to the signed integer and pass it to CodeGen_C which will try to lower signed shift, and pass it back to OpenCL codegen, which will cast again and so on, which is causing infinite recursion.  PTAL.

vksnk:
> Looks sensible, but I don't understand why this would only affect CMake?

Good question, it should affect both. Maybe I messed up something while testing with make.

vksnk:
All tests are green, submitting.

PR #6088, opened 2021-06-11
Added fixes to issues regarding using HALIDE_DEBUG_MATCHED_RULES

evanleenyc:
- Added missing operator<< overloading for IsMaxValue and IsMinValue structs 
- Moved build_replacement(after) to be called before debug matched rules, which was causing result to be undefined in the output


rootjalex (approved):
Lgtm. Buildbot failures are unrelated.

steven-johnson:
six unrelated failures is still too many to be comfy with

PR #6093, opened 2021-06-17
[hannk] More Hygiene

steven-johnson:
- TensorStorage takes a more sensible set of args for ctor
- Tensors don't need to be movable or copyable
- Since we are now using C++17, we can use std::make_unique instead of make_op

### DIFF
--- a/apps/hannk/interpreter/model.h
+++ b/apps/hannk/interpreter/model.h
@@ -20,11 +20,6 @@ namespace hannk {
 class Op;
 using OpPtr = std::unique_ptr<Op>;
 
-template<class T, class... Args>
-std::unique_ptr<T> make_op(Args &&...args) {
-    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
-}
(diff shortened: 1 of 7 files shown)


apps/hannk/interpreter/model.h
```diff
@@ -20,11 +20,6 @@ namespace hannk {
 class Op;
 using OpPtr = std::unique_ptr<Op>;
 
-template<class T, class... Args>
-std::unique_ptr<T> make_op(Args &&...args) {
-    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
```
dsharletg, line 25: I was thinking this was good to keep to go with OpPtr. It enables easily changing to shared_ptr if needed (possible...).

steven-johnson: Fair enough, I'll revert that part but change the body of make_op to use make_unique

dsharletg (approved):

PR #6094, opened 2021-06-17
[hannk] Don't mark Tensors as input or output

steven-johnson:
It's redundant information that can be wrong and requires the builder of the Model to set the right flags on the right Tensors. Instead, build sets of the inputs and/or outputs at the spots we care and check for membership in those. Root-level inputs and outputs should generally be small in count so performance should be a wash.

Also: drive-by change to configure_cmake.sh to allow changing the build type.

### DIFF
--- a/apps/hannk/interpreter/transforms.cpp
+++ b/apps/hannk/interpreter/transforms.cpp
@@ -25,42 +25,59 @@ T *cast_op(Op *x) {
     }
 }
 
+class RemoveDeadOps {
+    const Op *const root_;
+
+    bool is_root_output(const TensorPtr &t) const {
+        return t->is_output();
+    };
+
+public:
+    explicit RemoveDeadOps(Op *root)
+        : root_(root) {
+    }
+
+    void remove_in_group(OpGroup *op_group) {
+        // Find ops with outputs that are unused.
+        // Go in reverse order so removing a dead op
+        // enables earlier ops to be seen as dead.
+        for (int i = op_group->op_count() - 1; i >= 0; --i) {
+            Op *op = op_group->op(i);
+            if (OpGroup *group = cast_op<OpGroup>(op)) {
+                remove_in_group(group);
+            }
+            bool dead = true;
+            for (int j = 0; dead && j < op->input_count(); j++) {
+                if (is_root_output(op->input(j))) {
+                    dead = false;
+                    break;
+                }
+            }
+            for (int j = 0; dead && j < op->output_count(); j++) {
+                // An op isn't dead if its output is an output
+                // of the graph.
+                if (is_root_output(op->output(j))) {
+                    dead = false;
+                    break;
+                }
+
+                if (!op->output(j)->consumers().empty()) {
+                    dead = false;
+                    break;
+                }
+            }
+
+            if (dead) {
+                op_group->remove(op);
+            }
+        }
+    }
+};
+
 }  // namespace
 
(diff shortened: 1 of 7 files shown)


apps/hannk/interpreter/transforms.cpp
```diff
@@ -25,28 +27,28 @@ T *cast_op(Op *x) {
     }
 }
 
-}  // namespace
-
-void remove_dead_ops(OpGroup *root) {
+void remove_dead_ops_impl(OpGroup *root, const std::set<Tensor *> &root_outputs) {
```
dsharletg, line 30: Is root_outputs equivalent to root->output(i) for i = 0...op->output_count()? I think I have a helper function somewhere that checks if a tensor is in the inputs or outputs of an op, maybe we should use that instead of passing this set.

apps/hannk/interpreter/transforms.cpp
```diff
@@ -63,79 +65,89 @@ void remove_dead_ops(OpGroup *root) {
     }
 }
 
+}  // namespace
+
+void remove_dead_ops(OpGroup *root) {
+    std::set<Tensor *> root_outputs;
```
dsharletg, line 71: I think we should just iterate through the outputs whenever we need to check if a tensor is in the set of outputs. It'd be simpler and the performance upside is certainly negligible. It might even be a lot faster for such small sets.

Maybe just add a helper function `is_output(const Op *, const Tensor *)` and call it whenever querying this set. I'm pretty sure I already had that helper function somewhere at least at one time.

apps/hannk/interpreter/transforms.cpp
```diff
@@ -211,12 +223,24 @@ class InPlace : public OpVisitor {
             op->op(i)->accept(this);
         }
     }
+
+    std::set<Tensor *> root_inputs_and_outputs;
+
+public:
+    explicit InPlace(Op *root) {
```
dsharletg, line 230: Same comment here, maybe just use helper functions `is_input` and `is_output`.

steven-johnson:
PTAL

apps/hannk/interpreter/transforms.cpp
```diff
@@ -25,42 +25,59 @@ T *cast_op(Op *x) {
...
-        Op *op = root->op(i);
-        if (OpGroup *group = cast_op<OpGroup>(op)) {
-            remove_dead_ops(group);
-        }
-        bool dead = true;
-        for (int j = 0; dead && j < op->input_count(); j++) {
-            if (op->input(j)->is_output()) {
-                dead = false;
-                break;
+    bool is_root_output(TensorPtr t) const {
```
dsharletg, line 31: const TensorPtr &

apps/hannk/interpreter/transforms.cpp
```diff
@@ -211,12 +228,23 @@ class InPlace : public OpVisitor {
             op->op(i)->accept(this);
         }
     }
+
+    const Op *const root_;
+
+    bool is_root_input_or_output(TensorPtr t) const {
```
dsharletg, line 234: const TensorPtr &

steven-johnson:
I ran `compare_vs_tflite --enable ti --seed 0` against all the tflite models in our reference set (once from this branch, once from master). Showed *zero* differences in output and timing differences in the noise range.

dsharletg (approved):

PR #6097, opened 2021-06-21
[hannk] Rework the InPlace pass to make Arena allocation easier

steven-johnson:
This reworks the way that aliased Tensor storage works to move the smarts out of Tensor itself -- TensorStorage is removed entirely -- and instead, have the InPlace visitor build up a list of aliased-tensor candidates, which then requires the caller (eg Interpreter) to use those to produce shared external buffers for aliasing. This should allow subsequent PRs to introduce further memory allocation improvements in a more straightforward manner.

(I recommend reading the changes on a per-commit basis, which may make the change progression easier to follow.)

### DIFF
--- a/apps/hannk/interpreter/interpreter.cpp
+++ b/apps/hannk/interpreter/interpreter.cpp
@@ -32,13 +32,60 @@ class AllocateAll : public OpVisitor {
     }
 };
 
+size_t size_in_elements(const TensorDimensions &dimensions) {
+    ptrdiff_t begin_offset = 0;
+    ptrdiff_t end_offset = 0;
+    for (int i = 0; i < (int)dimensions.size(); i++) {
+        const int stride = dimensions[i].stride;
+        if (stride < 0) {
+            begin_offset += stride * (ptrdiff_t)(dimensions[i].extent - 1);
+        } else /* stride >= 0 */ {
+            end_offset += stride * (ptrdiff_t)(dimensions[i].extent - 1);
+        }
+    }
+    end_offset += 1;
+    assert(end_offset >= begin_offset);
+    return (size_t)(end_offset - begin_offset);
+}
+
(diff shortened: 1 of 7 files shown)


apps/hannk/interpreter/interpreter.cpp
```diff
@@ -32,17 +32,65 @@ class AllocateAll : public OpVisitor {
...
+    assert(end_offset >= begin_offset);
+    return (size_t)(end_offset - begin_offset);
+}
+
 }  // namespace
 
 void Interpreter::init(InterpreterOptions options) {
     pad_for_ops(model_.get());
-    in_place(model_.get());
+    auto alias_groups = calculate_in_place_aliases(model_.get());
```
dsharletg, line 55: I am wondering if the alias_groups might get invalidated by later passes? What happens if constant folding and dead op removal removes ops/tensors that are used in alias groups?

dsharletg: I think aliasing can make some ops dead (if not now, it could), so we'd at least need to do that twice, and there might be a circular issue there. The old way handled this automatically because `std::shared_ptr<TensorStorage>` would just do the right thing.

steven-johnson: Hm, gotcha. OK, I'll leave this open for now, but look into a slightly different approach.

dsharletg: Maybe the solution here is to have another pass that goes through and removes dead tensors from the alias groups?

steven-johnson: > Maybe the solution here is to have another pass that goes through and removes dead tensors from the alias groups?

Yeah, that could work, but I think I have an alternate approach that may be less dramatic a change.

Issue #5449, opened 2020-11-14
Support float16 arithmetic on ARM


dsharletg:
ARM has what seems like a full set of instructions for float16, but I've been struggling to generate them from Halide: https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?page=11&search=f16

We should be able to generate these, float16 is a major attraction to many developers for other targets like GPUs.

l-oneil:
Hi, is there any update on when FP16 Neon support might get pushed into release? 

If not, is this planned to be worked on? Could somebody perhaps outline what work is required to add support for this? 
I'd potentially be interested in helping here, with some guidance.

abadams:
I don't think anyone is currently planning on working on this. We'd welcome a PR. The way to start would be with some test cases that should generate float16 instructions but don't. E.g. maybe in test/correctness/simd_op_check.cpp. Then you'd see what llvm IR Halide is emitting, then you'd check what llvm IR llvm wants (typically I do this by grepping for the instructions in llvm's test/CodeGen/AArch64 folder), and then try to bridge the gap. E.g. maybe there are intrinsics we're supposed to be calling.

steven-johnson:
#6102 has landed -- is there more work to do, or should we consider this done?

PR #6102, opened 2021-06-22
Float16 support in CodeGen_ARM

stevesuzuki-arm:
This PR is for #5449 .
Armv8-a extension of Half-precision floating point data processing is supported by CodeGen_ARM.

- The target needs to be set as 64-bit with "arm_fp16" feature.
- 32-bit is not supported in this commit.
- Upgrading fp16 to fp32 with emulated conversion is replaced with either
  - fp16 native instruction or
  - fp32 operation with native type conversion of fp16-fp32
- The dedicated unit test for NEON float16 is added.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -1,3 +1,4 @@
+#include <set>
 #include <sstream>
 
 #include "CSE.h"
@@ -56,6 +57,9 @@ protected:
     void visit(const LE *) override;
     void codegen_vector_reduce(const VectorReduce *, const Expr &) override;
     // @}
+    Type upgrade_type_for_arithmetic(const Type &t) const override;
+    Type upgrade_type_for_argument_passing(const Type &t) const override;
+    Type upgrade_type_for_storage(const Type &t) const override;
 
     /** Various patterns to peephole match against */
     struct Pattern {
@@ -77,6 +81,12 @@ protected:
     bool neon_intrinsics_disabled() {
         return target.has_feature(Target::NoNEON);
     }
+
+    bool isFloat16AndHasFeature(const Type &t) const {
+        // NOTE : t.is_float() returns true even in case of BFloat16. We don't include it for now.
+        return t.code() == Type::Float && t.bits() == 16 && target.has_feature(Target::ARMFp16);
+    }
+    bool supports_call_as_float16(const Call *op) const override;
 };
 
 CodeGen_ARM::CodeGen_ARM(const Target &target)
@@ -214,6 +224,7 @@ struct ArmIntrinsic {
         ScalarsAreVectors = 1 << 5,  // Some intrinsics have scalar arguments that are vector parameters :(
         SplitArg0 = 1 << 6,          // This intrinsic requires splitting the argument into the low and high halves.
         NoPrefix = 1 << 7,           // Don't prefix the intrinsic with llvm.*
+        RequireFp16 = 1 << 8,        // Available only if Target has ARMFp16 feature
     };
 };
 
@@ -223,8 +234,9 @@ const ArmIntrinsic intrinsic_defs[] = {
     {"vabs", "abs", UInt(16, 4), "abs", {Int(16, 4)}, ArmIntrinsic::HalfWidth},
     {"vabs", "abs", UInt(32, 2), "abs", {Int(32, 2)}, ArmIntrinsic::HalfWidth},
     {"llvm.fabs", "llvm.fabs", Float(32, 2), "abs", {Float(32, 2)}, ArmIntrinsic::HalfWidth},
+    {"llvm.fabs", "llvm.fabs", Float(16, 4), "abs", {Float(16, 4)}, ArmIntrinsic::HalfWidth | ArmIntrinsic::RequireFp16},
 
-    {"llvm.sqrt", "llvm.sqrt", Float(32, 4), "sqrt_f32", {Float(32, 4)}},
+    {"llvm.sqrt", "llvm.sqrt", Float(32, 2), "sqrt_f32", {Float(32, 2)}, ArmIntrinsic::HalfWidth},
     {"llvm.sqrt", "llvm.sqrt", Float(64, 2), "sqrt_f64", {Float(64, 2)}},
 
     // SABD, UABD - Absolute difference
@@ -301,6 +313,11 @@ const ArmIntrinsic intrinsic_defs[] = {
     {"vmins", "smin", Int(32, 2), "min", {Int(32, 2), Int(32, 2)}, ArmIntrinsic::HalfWidth},
     {"vminu", "umin", UInt(32, 2), "min", {UInt(32, 2), UInt(32, 2)}, ArmIntrinsic::HalfWidth},
     {"vmins", "fmin", Float(32, 2), "min", {Float(32, 2), Float(32, 2)}, ArmIntrinsic::HalfWidth},
+    {"vmins", "fmin", Float(16, 4), "min", {Float(16, 4), Float(16, 4)}, ArmIntrinsic::HalfWidth | ArmIntrinsic::RequireFp16},
+
+    // FCVTZS, FCVTZU
+    {nullptr, "fcvtzs", Int(16, 4), "fp_to_int", {Float(16, 4)}, ArmIntrinsic::HalfWidth | ArmIntrinsic::MangleRetArgs | ArmIntrinsic::RequireFp16},
+    {nullptr, "fcvtzu", UInt(16, 4), "fp_to_int", {Float(16, 4)}, ArmIntrinsic::HalfWidth | ArmIntrinsic::MangleRetArgs | ArmIntrinsic::RequireFp16},
 
     // SMAX, UMAX, FMAX - Max
     {"vmaxs", "smax", Int(8, 8), "max", {Int(8, 8), Int(8, 8)}, ArmIntrinsic::HalfWidth},
@@ -310,6 +327,7 @@ const ArmIntrinsic intrinsic_defs[] = {
     {"vmaxs", "smax", Int(32, 2), "max", {Int(32, 2), Int(32, 2)}, ArmIntrinsic::HalfWidth},
     {"vmaxu", "umax", UInt(32, 2), "max", {UInt(32, 2), UInt(32, 2)}, ArmIntrinsic::HalfWidth},
     {"vmaxs", "fmax", Float(32, 2), "max", {Float(32, 2), Float(32, 2)}, ArmIntrinsic::HalfWidth},
+    {"vmaxs", "fmax", Float(16, 4), "max", {Float(16, 4), Float(16, 4)}, ArmIntrinsic::HalfWidth | ArmIntrinsic::RequireFp16},
 
     // SQNEG, UQNEG - Saturating negation
     {"vqneg", "sqneg", Int(8, 8), "saturating_negate", {Int(8, 8)}, ArmIntrinsic::HalfWidth},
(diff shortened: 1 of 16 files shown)


stevesuzuki-arm:
I appreciate your support! This PR is ready to review. If I'm missing is anything to move forward, please kindly let me know.

src/CodeGen_ARM.cpp
```diff
@@ -77,6 +81,12 @@ class CodeGen_ARM : public CodeGen_Posix {
     bool neon_intrinsics_disabled() {
         return target.has_feature(Target::NoNEON);
     }
+
+    bool isFloat16AndHasFeature(const Type &t) const {
```
dsharletg, line 85: Halide convention would be to name this `is_float16_and_has_feature`

src/IROperator.cpp
```diff
@@ -2179,15 +2179,29 @@ Expr fast_pow(Expr x, Expr y) {
...
-    user_assert(x.type() == Float(32)) << "fast_inverse only takes float arguments\n";
+    user_assert(x.defined()) << "fast_inverse of undefined Expr\n";
     Type t = x.type();
-    return Internal::Call::make(t, "fast_inverse_f32", {std::move(x)}, Internal::Call::PureExtern);
+    if (t == Float(32)) {
+        return Internal::Call::make(t, "fast_inverse_f32", {std::move(x)}, Internal::Call::PureExtern);
+    } else if (t == Float(16)) {
+        return Internal::Call::make(t, "fast_inverse_f16", {std::move(x)}, Internal::Call::PureExtern);
+    } else {
+        user_assert(0) << "fast_inverse only takes float arguments\n";
```
dsharletg, line 2189: We have `user_error` instead of `user_assert(false)` for this (another one below).

test/correctness/float16_t_neon_op_check.cpp
```diff
@@ -0,0 +1,377 @@
...
+    printf("HL_TARGET is: %s\n", hl_target.to_string().c_str());
+    printf("HL_JIT_TARGET is: %s\n", jit_target.to_string().c_str());
+
+    // Only for 64bit target with fp16 feature
+    if (!(hl_target.arch == Target::ARM && hl_target.bits == 64 && hl_target.has_feature(Target::ARMFp16))) {
+        printf("[SKIP] To run this test, set HL_TARGET=arm-64-<os>-arm_fp16. \n");
+        return 0;
+    }
+    // Create Test Object
+    // Use smaller dimension than default(768, 128) to avoid fp16 overflow in reduction test case
```
dsharletg, line 329: I know this is out of scope for this PR (and Halide...), but it would be great if there were fp16 -> fp32 reduction instructions, so we could efficiently reduce fp16 inputs to fp32 outputs.

src/CodeGen_ARM.cpp
steven-johnson, line 85: style nit: prefer snake_case for methods names in Halide, so `is_float16_and_has_feature()`

src/CodeGen_ARM.cpp
```diff
@@ -223,8 +234,9 @@ const ArmIntrinsic intrinsic_defs[] = {
     {"vabs", "abs", UInt(16, 4), "abs", {Int(16, 4)}, ArmIntrinsic::HalfWidth},
     {"vabs", "abs", UInt(32, 2), "abs", {Int(32, 2)}, ArmIntrinsic::HalfWidth},
     {"llvm.fabs", "llvm.fabs", Float(32, 2), "abs", {Float(32, 2)}, ArmIntrinsic::HalfWidth},
+    {"llvm.fabs", "llvm.fabs", Float(16, 4), "abs", {Float(16, 4)}, ArmIntrinsic::HalfWidth | ArmIntrinsic::RequireFp16},
 
-    {"llvm.sqrt", "llvm.sqrt", Float(32, 4), "sqrt_f32", {Float(32, 4)}},
+    {"llvm.sqrt", "llvm.sqrt", Float(32, 2), "sqrt_f32", {Float(32, 2)}, ArmIntrinsic::HalfWidth},
```
steven-johnson, line 239: Is this change related to float16?

src/CodeGen_ARM.cpp
```diff
@@ -567,6 +591,43 @@ const ArmIntrinsic intrinsic_defs[] = {
     {"vabdl_i32x2", "vabdl_i32x2", UInt(64, 2), "widening_absd", {Int(32, 2), Int(32, 2)}, ArmIntrinsic::NoMangle | ArmIntrinsic::NoPrefix},
     {"vabdl_u32x2", "vabdl_u32x2", UInt(64, 2), "widening_absd", {UInt(32, 2), UInt(32, 2)}, ArmIntrinsic::NoMangle | ArmIntrinsic::NoPrefix},
 };
+
+// List of fp16 math functions which we can avoid "emulated" equivalent code generation.
+// Only possible if the target has ARMFp16 feature.
+
+// These can be vectorized as fp16 SIMD instruction
+const std::set<string> float16_native_funcs = {
+    "sqrt_f16",
```
steven-johnson, line 600: Style nit: list the entries alphabetically

src/CodeGen_ARM.cpp
```diff
@@ -567,6 +591,43 @@ const ArmIntrinsic intrinsic_defs[] = {
...
+    "is_nan_f16",
+    "is_inf_f16",
+    "is_finite_f16",
+};
+
+// These end up with fp32 math function call.
+// However, data type conversion of fp16 <-> fp32 is performed natively rather than emulation.
+// SIMD instruction is not available, so scalar based instruction is generated.
+const std::map<string, string> float16_transcendental_remapping = {
+    {"sin_f16", "sin_f32"},
```
steven-johnson, line 614: Style nit: list the entries alphabetically

src/CodeGen_ARM.cpp
```diff
@@ -1316,6 +1433,11 @@ string CodeGen_ARM::mattrs() const {
             separator = ",";
         }
 
+        if (target.has_feature(Target::ARMFp16)) {
+            arch_flags += separator + "+fullfp16";
```
steven-johnson, line 1437: What minimum version of LLVM is required to support this flag?

dsharletg (commented):
Thank you for this PR, great work and I'm very happy to see this support! I just have some minor nits, but I will approve the workflows to run anyways.

The only big comment I have is I wonder about splitting float16 simd_op_check into its own test. I see that the width of the test had to be reduced for reductions to work, but if we were to find a workaround for that issue, would it be better to roll this into the main simd_op_check test? That test is getting rather big and unwieldy, maybe the thing we should do is split all ARM testing into simd_op_check_arm (like we did for HVX). (But I'm fine with leaving that to a separate PR, I'm not asking you to do that split here.)

src/IROperator.cpp
steven-johnson, line 2189: Nit: Float(64) is a float argument but isn't supported, suggest something like `fast_inverse only takes float16 or float32 arguments`

src/IROperator.cpp
```diff
@@ -2179,15 +2179,29 @@ Expr fast_pow(Expr x, Expr y) {
...
-    user_assert(x.type() == Float(32)) << "fast_inverse_sqrt only takes float arguments\n";
+    user_assert(x.defined()) << "fast_inverse_sqrt of undefined Expr\n";
     Type t = x.type();
-    return Internal::Call::make(t, "fast_inverse_sqrt_f32", {std::move(x)}, Internal::Call::PureExtern);
+    if (t == Float(32)) {
+        return Internal::Call::make(t, "fast_inverse_sqrt_f32", {std::move(x)}, Internal::Call::PureExtern);
+    } else if (t == Float(16)) {
+        return Internal::Call::make(t, "fast_inverse_sqrt_f16", {std::move(x)}, Internal::Call::PureExtern);
+    } else {
+        user_assert(0) << "fast_inverse_sqrt only takes float arguments\n";
```
steven-johnson, line 2202: Same nit as above

src/runtime/arm_cpu_features.cpp
```diff
@@ -22,6 +22,8 @@ WEAK CpuFeatures halide_get_cpu_features() {
 
     // TODO: add runtime detection for ARMDotProd extension
     // https://github.com/halide/Halide/issues/4727
+
+    // TODO: add runtime detection for ARMFp16 extension
```
steven-johnson, line 26: Should probably open a tracking issue for this, too

src/CodeGen_ARM.cpp
```diff
@@ -1097,6 +1175,24 @@ void CodeGen_ARM::visit(const Call *op) {
...
+        auto it = float16_transcendental_remapping.find(op->name);
+        if (it != float16_transcendental_remapping.end()) {
+            // This op doesn't have float16 native function.
+            // So we call float32 equivalent func with native type conversion between fp16 and fp32
+            // instead of emulated equivalent code as in EmulatedFloat16Math.cpp
+            std::vector<Expr> new_args(op->args.size());
+            for (size_t i = 0; i < op->args.size(); i++) {
+                new_args[i] = cast(Float(32, op->args[i].type().lanes()), op->args[i]);
+            }
+            auto &fp32_func_name = it->second;
```
steven-johnson, line 1188: clang-tidy complains: 

```
CodeGen_ARM.cpp:1188:13: error: 'auto &fp32_func_name' can be declared as 'const auto &fp32_func_name' [readability-qualified-auto,-warnings-as-errors]
            auto &fp32_func_name = it->second;
```

stevesuzuki-arm:
> The only big comment I have is I wonder about splitting float16 simd_op_check into its own test

I agree with your ideas. Another point is that checking suffix for data type in vector should ideally be performed not only on floating point but also integer type test cases in simd_op_check, which I thought was too much work for this PR. So I must admit current shape is a bit compromised. As you suggested, I agree to leave that to a separate PR.

src/CodeGen_ARM.cpp
stevesuzuki-arm: Done.

src/IROperator.cpp
stevesuzuki-arm: Done.

test/correctness/float16_t_neon_op_check.cpp
stevesuzuki-arm: I agree. Thank you.

src/CodeGen_ARM.cpp
stevesuzuki-arm: Done.

src/CodeGen_ARM.cpp
stevesuzuki-arm: Done.

src/CodeGen_ARM.cpp
stevesuzuki-arm: Done.

src/CodeGen_ARM.cpp
stevesuzuki-arm: I was not able to identify the exact version. But this flag is supported at least LLVM10 which Halide has already dropped support.

src/IROperator.cpp
stevesuzuki-arm: Done.

src/IROperator.cpp
stevesuzuki-arm: Done.

src/runtime/arm_cpu_features.cpp
stevesuzuki-arm: Filed #6106

src/CodeGen_ARM.cpp
stevesuzuki-arm: Done.

dsharletg:
> Another point is that checking suffix for data type in vector should ideally be performed not only on floating point but also integer type test cases in simd_op_check, which I thought was too much work for this PR.

I think we actually used to do this, but Apple's ARM assembly output doesn't have these suffixes, so we removed them.

dsharletg (approved):

test/correctness/float16_t_neon_op_check.cpp
steven-johnson: (Let's open an issue for that so it doesn't get forgotten?)

src/CodeGen_ARM.cpp
steven-johnson: OK, but I guess I'm confused; this looks like we now prefer to use a 2-wide sqrt_f32 instead of a 4-wide version, which seems suboptimal.

dsharletg: The HalfWidth flag generates the full width intrinsic too.

steven-johnson (approved):
LGTM, but see comment about the change to sqrt_f32 -- it was unclear to me about the change there, so I think it probably warrants a comment in the code so that future readers don't also find it unclear

dsharletg:
I think there already is a comment in the place it should be: https://github.com/halide/Halide/blob/master/src/CodeGen_ARM.cpp#L210

That flag is used on 134 intrinsics, I don't think we should add a comment on just one of them about this.

stevesuzuki-arm:
Many thanks for your efforts to review this patch!! Maybe I'll explore what to do for further ideas.

PR #6103, opened 2021-06-22
[hannk] Minor cleanups

steven-johnson:
A few minor things:
- set InterpreterOptions for ModelRunner and Delegate usage, so verbosity works
~~- Remove unused get_tensor() method~~
- rename Tensor::resize() -> resize_dynamic(), because it will only work on dynamic tensors

### DIFF
--- a/apps/hannk/interpreter/interpreter.h
+++ b/apps/hannk/interpreter/interpreter.h
@@ -25,10 +25,6 @@ public:
     explicit Interpreter(std::unique_ptr<OpGroup> m, InterpreterOptions options = InterpreterOptions());
     ~Interpreter();
 
-    // Return the Tensor in the current Model with the given name.
-    // If none with that name, return null. Tensor is still owned by the Model.
-    TensorPtr get_tensor(const std::string &name);
-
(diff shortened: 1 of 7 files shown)


apps/hannk/interpreter/interpreter.h
```diff
@@ -25,10 +25,6 @@ class Interpreter {
     explicit Interpreter(std::unique_ptr<OpGroup> m, InterpreterOptions options = InterpreterOptions());
     ~Interpreter();
 
-    // Return the Tensor in the current Model with the given name.
-    // If none with that name, return null. Tensor is still owned by the Model.
-    TensorPtr get_tensor(const std::string &name);
```
dsharletg, line 30: This is *definitely* going to be needed at some point. AFAICT it's the API production use cases use to get at the inputs/outputs they want.

dsharletg (approved):

apps/hannk/interpreter/interpreter.h
steven-johnson: I'll restore it then.

PR #6105, opened 2021-06-23
Fix performance regression in is_monotonic

dsharletg:
#6083 introduced a performance regression. The old logic worked in the presence of likely and promise_clamped, but the new logic does not. This PR fixes this performance regression.

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -1030,6 +1030,21 @@ struct RemoveLikelies : public IRMutator {
     }
 };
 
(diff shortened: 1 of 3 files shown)


dsharletg:
Hmm, it seems like this code already should be ignoring these. I think maybe the fix for the issue we've found is actually just #6099, though that wouldn't explain why removing promise_clamped is also necessary.

dsharletg:
Unfortunately #6099 is not a fix for this, so this PR (or something else) is still necessary.

abadams (approved):

Issue #6108, opened 2021-06-24
extern_producer (and maybe others) test relies on exact sin/cos behavior


dsharletg:
Recently, we've internally switched to using llvm-libc for sinf, cosf, and similar functions. This slightly changed the bits of the results, which has caused the extern_producer test to start failing.

I think the best thing to do would be to avoid depending on these functions for tests, and floats in general. I recall a number of tests that could run on Hexagon except for they used sin/cos to generate data. Integer functions using mod could also work to generate similar data, and would be more portable and stable.

steven-johnson:
Fixed?

dsharletg:
I fixed the immediately failing test, but there are other tests that should have sin/cos use replaced with integer test data.

PR #6109, opened 2021-06-24
Remove floats from extern_producer

dsharletg:
Small changes to sinf/cosf in libc has caused this test to start failing for us. This PR changes the test to use integers instead. Addresses the immediate issue in #6108, but leaving that open because I think we should remove unnecessary float usage in the rest of the tests as well.

### DIFF
--- a/test/correctness/extern_producer.cpp
+++ b/test/correctness/extern_producer.cpp
@@ -27,7 +27,8 @@ int round_down(int x, int m) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


abadams:
Given that it's arbitrary, I think something like 61 would be a better choice than 64. 64 may hide coordinate system errors because it's a multiple of the tile size used.

abadams (approved):

PR #6115, opened 2021-06-28
Relax overzealous pruning rule

abadams:
We don't allow schedules that fuse to the extent that we can no longer
vectorize. This was implemented incorrectly though. The check assumed
that something was going to be compute_at inside the innermost loop, and
neglected the possibility that we were about to tile that loop.


rootjalex (approved):
Lgtm

steven-johnson:
Is this ready to land?

PR #6117, opened 2021-06-29
[hannk] Implement greedy algorithm in AllocationPlanner

steven-johnson:
This uses a basic greedy approach to doing an allocation plan for tensors in hannk.

Initial testing shows exact result matches between old and new code.

Drive-by changes:
- Change Interpreter's `verbose` -> `verbosity` to allow more output granularity, and update callers as needed.
- Fix two places in ModelRunner that should have called the function hooks rather than the functions directly.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -36,7 +36,9 @@ jobs:
     steps:
       # Specifically use the CMake version that comes with Ubuntu.
       - name: Install dependencies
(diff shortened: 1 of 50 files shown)


steven-johnson:
(Opening as 'draft' for now because I want to do some additional non-buildbot testing before requesting review.)

steven-johnson:
Looks good in non-buildbot testing. We're still ahead of TFLite memory usage, but the gap is much smaller now. I'll do further investigation to be sure I understand where the extra overhead is.

steven-johnson:
Monday Review Ping

dsharletg (approved):

dsharletg (approved):

PR #6121, opened 2021-06-30
Use attribute specifiers for HALIDE_MUST_USE_RESULT and HALIDE_ATTRIBUTE_DEPRECATED

steven-johnson:
As of C++17, we should be able to specify these uniformly via [[attribute]] syntax, rather than a mess of compiler-specific stuff.

(Sadly, there doesn't appear to be an attribute for always-inline that works across all our compilers...)

### DIFF
--- a/src/runtime/HalideRuntime.h
+++ b/src/runtime/HalideRuntime.h
@@ -40,20 +40,13 @@ extern "C" {
 #endif
 
 #ifndef HALIDE_MUST_USE_RESULT
-#ifdef __has_attribute
-#if __has_attribute(nodiscard)
-// C++17 or later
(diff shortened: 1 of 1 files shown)


abadams:
HalideRuntime.h is included by users of generated pipelines. I thought we supported everything all the way back to vanilla C for generated code, because e.g. it might be fed into a crappy cross-compiler. We require c++17 for the compiler itself and users of Halide.h

steven-johnson:
> HalideRuntime.h is included by users of generated pipelines. I thought we supported everything all the way back to vanilla C for generated code, because e.g. it might be fed into a crappy cross-compiler. We require c++17 for the compiler itself and users of Halide.h

Yeah, you're right. Phooey.

PR #6126, opened 2021-07-02
Add TailStrategy::PredicateLoads and TailStrategy::PredicateStores

dsharletg:
This adds two new TailStrategy values, TailStrategy::PredicateLoads only predicates the loads and not the stores, and TailStrategy::PredicateStores only predicates the stores and not the loads. This effectively makes the tail strategy for the unpredicated stores or loads TailStrategy::RoundUp.

This PR also refactors the implementation of predication, from expression tag hints to:

- Representing predicated loads with if_then_else(predicate, load)
- Adding a predicate field to Provide

This PR also now means that TailStrategy::Predicate and TailStrategy::GuardWithIf are identical. Historically, this has caused problems with performance on ARM/x86, but I am struggling to find a case of this to troubleshoot. I think even if it does happen in rare cases, it might be worth the complexity/code size/more consistent behavior across platforms improvements.


dsharletg:
All of the buildbot failures look like flaky network issues. I think this PR is ready to merge, please take a look when you can.

abadams (approved):

PR #6129, opened 2021-07-07
[Hexagon] Use LLVM masked stores.

aankit-ca:
Letting CodeGen_LLVM handle predicated stores for Hexagon allows us
to generate HVX predicated stores instead of scalar predicated stores.

The operators in hannk app should be able to test this patch.

@pranavb-ca @dsharletg


dsharletg (approved):

dsharletg:
You should merge master (contains an LLVM tracking fix) so we can get test coverage of this.

PR #6130, opened 2021-07-07
Track argument change to LLVM's CreateMaskedLoad.

zvookin:


zvookin:
I believe the type should be the vector result type, not the type of a single element of the result.

dsharletg (approved):

alexreinking (approved):

PR #6132, opened 2021-07-12
Fix wasm regression at ToT LLVM

steven-johnson:
llvm.wasm.promote.low was removed. Calling fpext directly is the preferred approach now.

### DIFF
--- a/src/runtime/wasm_math.ll
+++ b/src/runtime/wasm_math.ll
@@ -164,13 +164,7 @@ define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwi
 ;   }
 
 ; single to double-precision floating point
-
-declare <2 x double> @llvm.wasm.promote.low(<4 x float>)
-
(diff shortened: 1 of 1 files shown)


src/runtime/wasm_math.ll
```diff
@@ -164,13 +164,7 @@ define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwi
 ;   }
 
 ; single to double-precision floating point
-
-declare <2 x double> @llvm.wasm.promote.low(<4 x float>)
-
 define weak_odr <4 x double> @float_to_double(<4 x float> %x) nounwind alwaysinline {
```
dsharletg, line 167: Do we need this helper function at all now? This is probably the default CodeGen_LLVM behavior.

steven-johnson: Good question. Stand by.

steven-johnson: Nope: removing this scalarizes the operation.

dsharletg (approved):

PR #6133, opened 2021-07-12
Fix wasm codegen for i32_to_double_s and i32_to_double_u

steven-johnson:
These were also easily fixed by using LLVM builtins.

### DIFF
--- a/src/runtime/wasm_math.ll
+++ b/src/runtime/wasm_math.ll
@@ -164,13 +164,7 @@ define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwi
 ;   }
 
 ; single to double-precision floating point
-
-declare <2 x double> @llvm.wasm.promote.low(<4 x float>)
-
 define weak_odr <4 x double> @float_to_double(<4 x float> %x) nounwind alwaysinline {
-  %1 = shufflevector <4 x float> %x, <4 x float> undef, <4 x i32> <i32 2, i32 3, i32 undef, i32 undef>
-  %2 = tail call <2 x double> @llvm.wasm.promote.low(<4 x float> %x)
-  %3 = tail call <2 x double> @llvm.wasm.promote.low(<4 x float> %1)
-  %4 = shufflevector <2 x double> %2, <2 x double> %3, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
-  ret <4 x double> %4
+  %1 = fpext <4 x float> %x to <4 x double>
+  ret <4 x double> %1
 }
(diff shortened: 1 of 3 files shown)


src/runtime/wasm_math.ll
```diff
@@ -140,28 +140,15 @@ define weak_odr <8 x i16> @saturating_narrow_i32x8_to_u16x8(<8 x i32> %x) nounwi
 }
 
 ; Integer to double-precision floating point
+define weak_odr <4 x double> @i32_to_double_s(<4 x i32> %x) nounwind alwaysinline {
+  %1 = sitofp <4 x i32> %x to <4 x double>
```
dsharletg, line 144: This is almost certainly what CodeGen_LLVM does. Are you sure we need these helpers?

steven-johnson: Without these helpers, everything gets scalarized.

steven-johnson: (But yeah, this is definitely suboptimal. I'll follow up with wasm LLVM folks to see if this can be improved.)

dsharletg: Why does it scalarize? That doesn't make any sense.

steven-johnson: You are correct. It doesn't make any sense. I'm contacting the relevant wasm person right now.

dsharletg (approved):

src/runtime/wasm_math.ll
steven-johnson: Looking more closely at the code, these cases actually *don't* get scalarized, but neither to they generate the opcodes we expect; instead, it generates a sequence of ops that loads two int32s, extends them to int64, then does bit-twiddling to convert them to IEEE754 doubles. Maybe there's a codegen path that predates these instructions being added to the wasmSimd128 spec that hasn't been updated? I'll circle around with the wasm-llvm folks to see what's going on. In the meantime, this is arguably an improvement on the status quo.

steven-johnson: I'm going to leave this open (not landed) pending the LLVM investigation.

dsharletg: Did you also report this for float -> double, from #6132?

steven-johnson: I'm going to provide repro cases for that as well

steven-johnson: Opened https://bugs.llvm.org/show_bug.cgi?id=51098

steven-johnson:
This was fixed in upstream LLVM, so I'm closing this.

PR #6134, opened 2021-07-13
Optimize shallow depthwise convolutions

dsharletg:
For shallow depthwise convolutions (for our purposes, those with a number of channels smaller than and dividing the vector size), it is faster to fuse the c and x dimensions, and handle the stride of the x indices manually, so we can process multiple values of x at a time. In extreme cases, this makes a huge difference (e.g. on HVX with 8 channels, where the vector size is 128).

Even on ARM, this significantly impacts several networks, including one of the mobilenet_v1 networks which we have been tracking the performance of.


apps/hannk/interpreter/ops.cpp
```diff
@@ -810,25 +811,41 @@ namespace {
...
     }
 }
 
+int get_depthwise_conv_channel_alignment() {
+    // Pass minimal sized buffers to learn about the alignment requirements.
+    HalideBuffer<uint8_t> input_buf(nullptr, 1, 1, 1, 1);
+    HalideBuffer<int32_t> bias_buf(nullptr, 1);
+    HalideBuffer<uint8_t> filter_buf(nullptr, 1, 1, 1);
+    HalideBuffer<uint8_t> output_buf(nullptr, 1, 1, 1, 1);
+    depthwise_conv_dm1_uint8(input_buf, 0, filter_buf, 0, bias_buf, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, output_buf);
```
steven-johnson, line 845: Is this expensive enough to be worth caching?

steven-johnson (approved):

apps/hannk/interpreter/ops.cpp
dsharletg: Maybe, but I was going to wait until it shows up in a profiler to do that.

vksnk (approved):

dsharletg:
This has passed the only test coverage on the build bots now.

PR #6138, opened 2021-07-14
[hannk] Allow aliasing of Reshape tensors

steven-johnson:
Previously we didn't allow this because aliased tensors had to have the same rank, which is ~never the case for Reshape. Aliasing for Reshape is a huge win because it essentially becomes a no-op rather than a memcpy. Running against standard set of models shows no regression in differences vs. tflite.

### DIFF
--- a/apps/hannk/interpreter/ops.cpp
+++ b/apps/hannk/interpreter/ops.cpp
@@ -1400,9 +1400,10 @@ void ReshapeOp::execute() {
         assert(new_shape.at(d) == output_buf.dim(d).extent());
     }
 
-    // NOTE: HCHECK, not assert, to ensure a runtime failure to match TFLite
-    // TODO: return error code
-    HCHECK(input_buf.number_of_elements() == output_buf.number_of_elements());
+    // TODO: we must verify these match (and fail at runtime if not).
+    // That said, we should be able to predict this at parse time
+    // (for non-dynamic tensors) and skip the runtime check most of the time.
+    assert(input_buf.number_of_elements() == output_buf.number_of_elements());
 
     assert(in->is_dense());
     assert(out->is_dense());
--- a/apps/hannk/interpreter/ops.cpp
+++ b/apps/hannk/interpreter/ops.cpp
@@ -1400,12 +1400,16 @@ void ReshapeOp::execute() {
         assert(new_shape.at(d) == output_buf.dim(d).extent());
     }
 
-    assert(input_buf.number_of_elements() == output_buf.number_of_elements());
+    // NOTE: HCHECK, not assert, to ensure a runtime failure to match TFLite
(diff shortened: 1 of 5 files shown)


apps/hannk/interpreter/ops.cpp
```diff
@@ -1400,12 +1400,16 @@ void ReshapeOp::execute() {
         assert(new_shape.at(d) == output_buf.dim(d).extent());
     }
 
-    assert(input_buf.number_of_elements() == output_buf.number_of_elements());
+    // NOTE: HCHECK, not assert, to ensure a runtime failure to match TFLite
```
dsharletg, line 1403: We should still be asserting here. This check should be in the parser (i.e. not running every iteration).

steven-johnson:
Ehh, I think this needs more work, actually: it doesn't account for the case of the RHS of the set_alias_of() call having a storage_offset.

apps/hannk/interpreter/tensor.cpp
```diff
@@ -192,13 +222,20 @@ void Tensor::set_alias_of(const TensorPtr &t, const SmallVector<int, max_rank> &
...
     auto &shared_buffer = storage_->buffer;
     assert(shared_buffer.type().bytes() == type().bytes());
-    assert(shared_buffer.dimensions() == (int)offset_bounds.size());
     assert(!shared_buffer.data());
 
     // Check that the storage is big enough for this buffer.
-    for (int i = 0; i < shared_buffer.dimensions(); i++) {
-        assert(offset_bounds[i].min >= shared_buffer.dim(i).min());
-        assert(offset_bounds[i].max <= shared_buffer.dim(i).max());
+    if (this->rank() == t->rank()) {
```
dsharletg, line 228: I think this is a little sketchy. I think reshapes of the same rank are legal, where each dimension has a different extent, but the total number of elements is the same. I think this either needs to be a separate function, or a parameter indicating the aliasing should share "flat" indices or "rank-dimensional" indices.

apps/hannk/interpreter/ops.cpp
steven-johnson: The dynamic-tensor case still needs a runtime check.

apps/hannk/interpreter/tensor.cpp
steven-johnson: Originally I just had a flag inside the Tensor indicating whether the aliasing was for a reshape or not, but that felt clumsy -- I didn't think about the same-rank, different-extent case though. I'll revisit this.

steven-johnson:
PTAL. (Added some drive-by debugging changes to lower_tflite_lstm() and Tensor::dump as well.)

dsharletg (approved):

PR #6144, opened 2021-07-19
[hannk] Rewrite FC in terms of Conv2D

steven-johnson:
FullyConnected is very similar to Conv2D, so rather than maintaining multiple similar implementations, let's translate a FullyConnected node into a Conv2D node (with some Reshape nodes as necessary). Benchmarking before-and-after shows performance to be more or less a wash on x86 OSX laptop and Pixel 4a.

Note that we keep the old FC logic for int16 outputs, as Conv2D doesn't support those yet; if this PR is landed, a followup PR will add that ability to Conv2D, and the existing FC support will be removed entirely.

Also a handful of drive-by fixes.

### DIFF
--- a/apps/hannk/interpreter/lower.cpp
+++ b/apps/hannk/interpreter/lower.cpp
@@ -18,7 +18,7 @@ OpPtr lower_tflite_lstm(TensorPtr data_input, TensorPtr prev_activ_input, Tensor
 
     std::vector<TensorPtr> concat_inputs = {data_input, prev_activ_input};
     ops.push_back(make_op<ConcatenationOp>(concat_inputs, concat_temp, 0));
-    ops.push_back(make_op<FullyConnectedOp>(concat_temp, weights_input, biases_input, activ_temp, activation));
+    ops.push_back(lower_tflite_fullyconnected(concat_temp, weights_input, biases_input, activ_temp, activation));
 
     // Split activ_temp into the 4 ops we need.
     Box elementwise_bounds = activ_temp->bounds();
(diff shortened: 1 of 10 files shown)


apps/hannk/interpreter/lower.cpp
```diff
@@ -74,4 +74,112 @@ OpPtr lower_tflite_lstm(TensorPtr data_input, TensorPtr prev_activ_input, Tensor
...
+        int c_extent = bounds.front().extent();
+        int b_extent = bounds.back().extent();
+        if (bounds.size() == 4) {
+            // Reshape (c, x, y, b) to (cxy, 1, 1, b)
+            c_extent = bounds[0].extent() * bounds[1].extent() * bounds[2].extent();
+        }
+        Box reshaped_bounds = {{0, c_extent - 1}, {0, 0}, {0, 0}, {0, b_extent - 1}};
+        t_reshaped = std::make_shared<Tensor>(name, t->type(), std::move(reshaped_bounds), t->quantization());
+    }
+    t_reshaped->set_constant(t->is_constant());
```
dsharletg, line 143: I don't think this should be set. Constant folding should do it if possible. I think we don't want to mark something constant before it's actually computed like this.

apps/hannk/util/model_runner.cpp
```diff
@@ -571,7 +571,7 @@ int ModelRunner::parse_flags(int argc, char **argv, std::vector<std::string> &fi
 }
 
 void ModelRunner::run(const std::string &filename) {
-    std::cout << "Processing " << filename << " ...\n";
+    std::cout << "\nProcessing " << filename << " ...\n";
```
dsharletg, line 574: Can we not do this? Having one line per output is *dramatically* better than 2 lines per output when running a lot of stuff (100 lines is a lot better than 200 lines). And it means we can copy/paste into spreadsheets without it being a pain.

dsharletg (approved):

apps/hannk/util/model_runner.cpp
```diff
@@ -617,7 +617,13 @@ void ModelRunner::run(const std::string &filename) {
     // ----- Log benchmark times
     if (do_benchmark) {
         for (WhichRun i : active_runs) {
-            std::cout << RunNames[i] << " Time: " << std::chrono::duration_cast<std::chrono::microseconds>(results[i].time).count() << " us\n";
+            const auto t = std::chrono::duration_cast<std::chrono::microseconds>(results[i].time).count();
+            std::cout << RunNames[i] << " Time: " << t << " us";
+            if (i != kTfLite) {
+                const auto t_ref = std::chrono::duration_cast<std::chrono::microseconds>(results[kTfLite].time).count();
+                std::cout << " (Speedup: " << std::fixed << std::setprecision(2) << (double)t_ref / (double)t << ")";
```
dsharletg, line 624: Oh missed this, can we also not do this? Again, the reason being, copy/pasting into a spreadsheet is a lot more annoying, and actually just calculating this in the spreadsheet is easy.

I think in general we should just not change the format of the output in this (unrelated) PR.

apps/hannk/util/model_runner.cpp
steven-johnson: FWIW, running with --benchmark will always output multiple lines per, and this change makes it IMHO much easier to read, but I'll back it out

apps/hannk/interpreter/lower.cpp
steven-johnson: Done.

apps/hannk/util/model_runner.cpp
steven-johnson: Well, sure, but it requires you actually copy/paste stuff into a spreadsheet, which is annoying (to me) for quick checks.

I'll back out these changes here and propose them elsewhere, and we can discuss our different approaches to this so we aren't competing to optimize for different paths...

PR #6145, opened 2021-07-20
Add support for int16 output in Conv2D

steven-johnson:
This allows us to convert all (currently supported) FC ops into Conv2D ops. Remove all the FC-specific Halide and Op code.


dsharletg (approved):
Nice, hooray for deleting code!

PR #6146, opened 2021-07-20
Generalize Conv2D to be a Conv of any dimensionality

dsharletg:
This PR makes it so the ConvOp handles any dimensionality up to 2D. FullyConnected is the same as a 0D Conv in this interpretation.


steven-johnson (approved):
Nice!

PR #6153, opened 2021-07-21
Work around excessively slow build time due to likely scalarization

dsharletg:
@aankit-ca with this, depthwise compiles in just a few seconds.


dsharletg:
Closed in favor of #6154

aankit-ca:
Thank you Dillon for the fix!

PR #6154, opened 2021-07-21
Handle depth_multiplier != 1 in a separate op

dsharletg:
This PR rewrites depthwise conv ops with depth_multiplier != 1 into a separate "upsample channels" op followed by a depthwise conv with depth_multiplier == 1. This simplifies things quite a bit and lifts some messy code out of depthwise conv, where it gets multiple copies (due to specializations).

This should also improve performance when we implement line buffering between ops, because we weren't line buffering this op inside depthwise conv, meaning we redundantly computed a lot of channel upsampling (which can be expensive).

This replaces #6153.


apps/hannk/halide/depthwise_conv_generator.cpp
```diff
@@ -8,8 +8,8 @@ namespace hannk {
 
 class DepthwiseConv : public Generator<DepthwiseConv> {
 public:
-    // If positive, a constant inverse depth multiplier.
-    GeneratorParam<int> inv_depth_multiplier_{"inv_depth_multiplier", -1};
+    // This is used to compute ci = co * inv_depth_multiplier
```
vksnk, line 11: Is it only to support broadcasting version, when inv_depth_multiplier is 0? If yes, it'd be helpful to put this into the comment.

vksnk (approved):

dsharletg:
(Merged because the only changes in the last commit were comments)

Issue #3735, opened 2019-03-15
`prefetch` can't express efficient fine-grained prefetching


dsharletg:
I have some intrinsics code where __builtin_prefetch helps a lot. This code reads small tiles of an input, so hardware prediction of the next address probably struggles with this. The prefetching strategy that works well is this:

    for y:
      for x:
        // Prefetch next line's x value.
        __builtin_prefetch(base + (y + 1) * stride_y) + x * 8)
        // Load current line's x value.
        uint16x8 = load(base + y * stride_y + x * 8)
        // work
        ...

The problem is that Halide's `prefetch` directive can't express this: it takes a buffer, a loop variable, and an offset to describe which iteration of *that* loop to prefetch. This can only describe either prefetching the next value of `x`, or the next whole row of `y`. The issue with the latter option is that it tends to generate a lot of overhead.

One option would be to try to reduce the overhead. I think prefetching the next whole row of y should work in some cases, but the generated code isn't as simplified as the loop itself (for example, a real loop over x gets generated that has a non-static iteration count, even though the actual loop over x does). However, even if that were to be fixed, I suspect there are pipelining advantages to just dropping a single prefetch in the middle of an instruction stream, over generating a bunch of prefetches all at once.

Another option would be to add a parameter to `prefetch`:

    Stage &prefetch(const Func &f, VarOrRVar where, VarOrRVar next, Expr offset = 1,
                    PrefetchBoundStrategy strategy = PrefetchBoundStrategy::GuardWithIf);

- `where` would be the loop to place the prefetch in
- `next` and `offset` would determine the bounds of the thing to prefetch

I think there's some trickiness to doing this. One thing in particular, if `where` is a small inner loop, we don't want any overhead surrounding the prefetch (no loops to traverse to cover a `Box` of bounds).

steven-johnson:
Pretty sure this is addressed by #6155

PR #6155, opened 2021-07-22
Add more fine-grained prefetch() directive (Issue #3735)

steven-johnson:
This adds a new scheduling directive, `prefetch_at()`, that allows you to specify prefetching a value from a different loop variable (See issue #3735 for details). Note that `prefetch_at(f, b, b)` is exactly identical to `prefetch(f, b)` in function.

Opening as draft for now:

~~Defining it as an overload on `prefetch()` isn't possible due to ambiguous overloads that would be created, hence the new name; since the existing `prefetch()` directive is rarely used, do we just change its argument list rather than adding a wrapper?~~

~~Doc comments needed~~

~~Python API wrappers needed~~

- Changing the parameters to __builtin_prefetch are a *big* win, at least on ARM experiments. Do we need to expose these knobs to user code somehow? (Do we know why the previous values where specified?)

- No testing on Hexagon has been done.

- Initial testing on ARM indicates that adding prefetch can add a surprising amount of apparently-unnecessary stack spill from the generated code; this needs to be investigated.


src/runtime/prefetch.cpp
```diff
@@ -6,7 +6,9 @@ extern "C" {
 // parameter breaks a lot of optimizations, but needs to be WEAK
 // so that Codegen_LLVM can find an instance of the Function to insert.
 WEAK_INLINE int _halide_prefetch(const void *ptr) {
-    __builtin_prefetch(ptr, 1, 3);
+    constexpr int rw = 0;        // 1 = write, 0 = read
+    constexpr int locality = 0;  // 0 = no temporal locality, 3 = high temporal locality
```
dsharletg, line 10: I wonder if we need to make these options parameters to the prefetch directive somehow. Experimenting with them does produce different results. However, I suspect these current values (read, no temporal locality) is probably the most useful/common case.

steven-johnson:
The apparently-unnecessary stack spill has been reported to LLVM as https://bugs.llvm.org/show_bug.cgi?id=51172

steven-johnson:
Marked as read for review because it's basically doing what we want -- the issue with LLVM is one that (hopefully) will improve code generation further, but it isn't a blocker, strictly speaking.

src/Func.h
```diff
@@ -1990,13 +2003,101 @@ class Func {
...
-        return prefetch(image.parameter(), var, offset, strategy);
+        return prefetch_at<T>(image, var, var, std::move(offset), strategy);
+    }
+    // @}
+
+    /** prefetch_at() is a more fine-grained version of prefetch(), which allows
+     * specification of different vars for the location of the prefetch() instruction
+     * vs. the location that is being prefetched:
+     *
+     * - the first var specified, 'loop', indicates the loop in which the prefetch will be placed
```
dsharletg, line 2024: I think there must be better names for these, at least `loop`. Both options are loops... I think the first var should be something related to `at`, given the similarity to `compute_at` and `store_at`. `prefetch_at`, and `prefetch_from` for the other variable? Or maybe just `at` and `from` given the context.

src/Prefetch.cpp
```diff
@@ -154,8 +155,10 @@ class InjectPrefetch : public IRMutator {
         } else if (op->bounds.empty()) {
             // Remove the Prefetch IR since it is prefetching an empty region
             user_warning << "Removing prefetch of " << p.name
-                         << " within loop nest of " << p.var << " (offset: "
-                         << p.offset << ") since it is not used at all.\n";
+                         << " within loop nest of " << p.loop_var
+                         << " at location " << p.fetch_var
```
dsharletg, line 159: at -> from?

src/runtime/qurt_hvx.cpp
```diff
@@ -57,7 +57,7 @@ WEAK_INLINE int _halide_prefetch_2d(const void *ptr, int width_bytes, int height
     return 0;
 }
 
-WEAK_INLINE int _halide_prefetch(const void *ptr, int size) {
+WEAK_INLINE int _halide_prefetch_1D(const void *ptr, int size) {
```
dsharletg, line 60: 1D -> 1d (the call site is 1d, and matches 2d convention)

src/Func.h
steven-johnson: Used `at` and `from`, PTAL

src/Prefetch.cpp
steven-johnson: Done

src/runtime/qurt_hvx.cpp
steven-johnson: Done

src/Func.cpp
```diff
@@ -1840,14 +1840,14 @@ Stage &Stage::hexagon(const VarOrRVar &x) {
     return *this;
 }
 
-Stage &Stage::prefetch(const Func &f, const VarOrRVar &var, Expr offset, PrefetchBoundStrategy strategy) {
-    PrefetchDirective prefetch = {f.name(), var.name(), std::move(offset), strategy, Parameter()};
+Stage &Stage::prefetch_at(const Func &f, const VarOrRVar &loop, const VarOrRVar &from, Expr offset, PrefetchBoundStrategy strategy) {
```
dsharletg, line 1843: Rename these too, for consistency with the header?

src/Func.h
```diff
@@ -1990,13 +2003,101 @@ class Func {
...
+     * The following schedule:
+     \code
+     f.compute_root();
+     g.prefetch_at(f, x, x, 2, PrefetchBoundStrategy::NonFaulting);
+     h.prefetch_at(f, x, y, 2, PrefetchBoundStrategy::NonFaulting);
+     \endcode
+     *
+     * will inject prefetch call at the innermost loop of 'g' and 'h' and generate
+     * the following loop nest:
+     * for y = ...
```
dsharletg, line 2049: \code...\endcode

src/Func.h
```diff
@@ -1990,13 +2003,101 @@ class Func {
...
+     \endcode
+     *
+     * The following schedule:
+     \code
+     f.compute_root();
+     g.prefetch_at(f, y, w, 2, PrefetchBoundStrategy::NonFaulting);
+     \endcode
+     *
+     * will produce code that prefetches a tile of data:
+     * for w = ...
```
dsharletg, line 2076: \code...\endcode

src/Func.h
```diff
@@ -1990,13 +2003,101 @@ class Func {
...
+     *   for z = ...
+     *     for y = ...
+     *       for x0 = ...
+     *          prefetch(&f[x0, y, z, w + 2], 1, 16);
+     *       for x = ...
+     *         g(x, y, z, w) = 2 * f(x, y, z, w)
+     *
+     * Note that calling prefetch_at() with the same var for both 'at' and 'from'
+     * is exactly equivalent to calling prefetch() with that var.
```
dsharletg, line 2090: Nit: I'd remove 'exactly' here.

dsharletg (approved):

steven-johnson:
Per discussion at group meeting:
- Existing version of prefetch() had `offset` changed from Expr to int (as a temporary measure to avoid ambiguous call resolution issues), then marked as deprecated
- The new, two-var call is also named prefetch(), but is not deprecated
- All existing calls within the Halide codebase that reference the deprecated call have been updated to the non-deprecated call

steven-johnson:
Do we want to go ahead and land this (in advance of hoped-for LLVM improvements?)

dsharletg:
I think we should, I think the change is working as designed in Halide. And, despite the LLVM codegen issues, this was still a big improvement in a few cases I tried.

PR #6158, opened 2021-07-22
De-predicate loads and stores in Metal/OpenCL/D3D12 backend

### LINKED ISSUE #6157: Nested GuardWithIf generates predicated store on GPU
Recent changes to use predication always seem to have broken the ability on the GPU to have both the `gpu_tile` and `vectorize` tail strategy be `GuardWithIf`.  Here's a small repro:

```
#include "Halide.h"
#include <iostream>
#include <stdio.h>

using namespace Halide;

    Var x("x"), y("y"), xi("xi"), yi("yi");
    Func f("f");

    printf("Defining function...\n");

    f(x, y) = x * y + 2.4f;

    Target target = get_jit_target_from_environment();
    if (target.has_gpu_feature()) {
      f.gpu_tile(x, y, xi, yi, 8, 8, TailStrategy::GuardWithIf).vectorize(xi, 4, TailStrategy::GuardWithIf);
    }

    printf("Realizing function...\n");

    Buffer<float> imf = f.realize({32, 32}, target);

    // Check the result was what we expected
    for (int i = 0; i < 32; i++) {
        for (int j = 0; j < 32; j++) {
            float correct = i * j + 2.4f;
            if (fabs(imf(i, j) - correct) > 0.001f) {
                printf("imf[%d, %d] = %f instead of %f\n", i, j, imf(i, j), correct);
                return -1;
            }
        }
    }

    printf("Success!\n");
    return 0;
}
```

This causes the following error:
```
User error triggered at /Users/kamil/Documents/work/adobe/halide_master/src/CodeGen_Metal_Dev.cpp:422
Condition failed: is_const_one(op->predicate)
Error: (at test/correctness/gpu_vectorize.cpp:23)
Predicated store is not supported inside Metal kernel.
```

Everything seems to work fine if only using `GuardWithIf` for the `vectorize()` call, but not when both `gpu_tile()` and `vectorize()` use that tail strategy.

shoaibkamil:
Because GPU backends do not support predication, this adds a mutator that de-predicates them, and a test to exercise this.  Should fix #6157.

### DIFF
--- a/src/CodeGen_GPU_Dev.cpp
+++ b/src/CodeGen_GPU_Dev.cpp
@@ -115,7 +115,7 @@ protected:
                                 mutate(extract_lane(s->value, ln)),
                                 mutate(extract_lane(s->index, ln)),
                                 s->param,
-                                make_bool(true),
+                                const_true(),
                                 // TODO: alignment needs to be changed
                                 s->alignment)));
             }
--- a/src/CodeGen_GPU_Dev.cpp
+++ b/src/CodeGen_GPU_Dev.cpp
@@ -102,9 +102,10 @@ namespace {
 class ScalarizePredicatedLoadStore : public IRMutator {
 public:
     using IRMutator::mutate;
+    using IRMutator::visit;
 
 protected:
-    Stmt visit(const Store *s) {
+    Stmt visit(const Store *s) override {
         if (!is_const_one(s->predicate)) {
             std::vector<Stmt> scalar_stmts;
             for (int ln = 0; ln < s->value.type().lanes(); ln++) {
@@ -124,7 +125,7 @@ protected:
         }
     }
 
-    Expr visit(const Load *op) {
+    Expr visit(const Load *op) override {
         if (!is_const_one(op->predicate)) {
             Expr load_expr = Load::make(op->type, op->name, op->index, op->image, 
                                         op->param, const_true(op->type.lanes()), op->alignment);
--- a/src/CodeGen_GPU_Dev.cpp
+++ b/src/CodeGen_GPU_Dev.cpp
@@ -1,5 +1,8 @@
 #include "CodeGen_GPU_Dev.h"
+#include "Deinterleave.h"
 #include "Bounds.h"
+#include "IRMutator.h"
+#include "IROperator.h"
 #include "IRVisitor.h"
 
 namespace Halide {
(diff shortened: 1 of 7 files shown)


shoaibkamil:
Draft is to get feedback on the strategy-- I'll fix CMake/clang-format issues and do the other backends if this seems sensical.

src/CodeGen_GPU_Dev.cpp
```diff
@@ -94,5 +97,55 @@ bool CodeGen_GPU_Dev::is_buffer_constant(const Stmt &kernel,
...
+    using IRMutator::mutate;
+    using IRMutator::visit;
+
+protected:
+    Stmt visit(const Store *s) override {
+        if (!is_const_one(s->predicate)) {
+            std::vector<Stmt> scalar_stmts;
+            for (int ln = 0; ln < s->value.type().lanes(); ln++) {
+                scalar_stmts.push_back(IfThenElse::make(
+                                           extract_lane(s->predicate, ln),
```
dsharletg, line 113: It *might* be important to detect predicates of the form ramp < broadcast, and generate a nested if in that case. That is what the old logic did (implicitly/indirectly). I think this is simpler though, so we should see how this performs first.

shoaibkamil: That's a good point.  I'm not sure whether there will be perf differences on other backends, but my limited testing on Metal seems to indicate about the same performance.  Happy to revisit if anyone finds performance issues.

src/CodeGen_GPU_Dev.cpp
```diff
@@ -94,5 +97,55 @@ bool CodeGen_GPU_Dev::is_buffer_constant(const Stmt &kernel,
...
+        if (!is_const_one(s->predicate)) {
+            std::vector<Stmt> scalar_stmts;
+            for (int ln = 0; ln < s->value.type().lanes(); ln++) {
+                scalar_stmts.push_back(IfThenElse::make(
+                    extract_lane(s->predicate, ln),
+                    Store::make(s->name,
+                                mutate(extract_lane(s->value, ln)),
+                                mutate(extract_lane(s->index, ln)),
+                                s->param,
+                                make_bool(true),
```
dsharletg, line 118: `const_true()`

dsharletg (approved):

steven-johnson (approved):

PR #6160, opened 2021-07-23
Add various bounds-related simplifier rules

rootjalex:
All of these simplifier rules came from `find_constant_bounds` failures. Some are oddly specific and I can be convinced to remove them (i.e. in Simplify_Max.cpp and Simplify_Select.cpp), but many of them are useful for `find_constant_bounds`.

### DIFF
--- a/src/Simplify_Add.cpp
+++ b/src/Simplify_Add.cpp
@@ -74,7 +74,7 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
              rewrite(select(x, y, z) + (w + select(x, u, v)), select(x, y + u, z + v) + w) ||
              rewrite(select(x, y, z) + (select(x, u, v) - w), select(x, y + u, z + v) - w) ||
              rewrite(select(x, y, z) + (w - select(x, u, v)), select(x, y - u, z - v) + w) ||
-             rewrite(select(x, c0 - y, c1) + c2, select(x, fold(c0 + c2) - y, fold(c1 + c2)), !overflows(c0 + c2) && !overflows(c1 + c2)) ||
+             rewrite(select(x, c0 - y, c1) + c2, select(x, fold(c0 + c2) - y, fold(c1 + c2))) ||
 
              rewrite(x + y*(-1), x - y) ||
              rewrite(x*(-1) + y, y - x) ||
@@ -150,7 +150,6 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
                rewrite(max(y + c0, x) + c1, max(y, x + c1), c0 + c1 == 0) ||
                rewrite(max(x, y) + min(x, y), x + y) ||
                rewrite(max(x, y) + min(y, x), x + y) ||
-               rewrite(max(x, y*c0 + z) + (u - y)*c0, max(x - y*c0, z) + u*c0) ||
 
                rewrite(min(x, y + (z*c0)) + (z*c1), min(x + (z*c1), y), (c0 + c1) == 0) ||
                rewrite(min(x, (y*c0) + z) + (y*c1), min(x + (y*c1), z), (c0 + c1) == 0) ||
--- a/src/Simplify_Add.cpp
+++ b/src/Simplify_Add.cpp
@@ -74,6 +74,7 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
              rewrite(select(x, y, z) + (w + select(x, u, v)), select(x, y + u, z + v) + w) ||
              rewrite(select(x, y, z) + (select(x, u, v) - w), select(x, y + u, z + v) - w) ||
              rewrite(select(x, y, z) + (w - select(x, u, v)), select(x, y - u, z - v) + w) ||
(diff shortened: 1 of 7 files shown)


rootjalex:
I should have noted: all rules have been formally verified and confirmed to follow the reduction order

src/Simplify_Add.cpp
```diff
@@ -74,6 +74,7 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
              rewrite(select(x, y, z) + (w + select(x, u, v)), select(x, y + u, z + v) + w) ||
              rewrite(select(x, y, z) + (select(x, u, v) - w), select(x, y + u, z + v) - w) ||
              rewrite(select(x, y, z) + (w - select(x, u, v)), select(x, y - u, z - v) + w) ||
+             rewrite(select(x, c0 - y, c1) + c2, select(x, fold(c0 + c2) - y, fold(c1 + c2)), !overflows(c0 + c2) && !overflows(c1 + c2)) ||
```
abadams, line 77: Rules fail to match if a fold on the RHS overflows, so the predicate is unnecessary

src/Simplify_Add.cpp
```diff
@@ -148,6 +150,7 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
                rewrite(max(y + c0, x) + c1, max(y, x + c1), c0 + c1 == 0) ||
                rewrite(max(x, y) + min(x, y), x + y) ||
                rewrite(max(x, y) + min(y, x), x + y) ||
+               rewrite(max(x, y*c0 + z) + (u - y)*c0, max(x - y*c0, z) + u*c0) ||
```
abadams, line 153: Why does this rule show up twice?

src/Simplify_Max.cpp
```diff
@@ -97,6 +97,10 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
              rewrite(max(max(x, y), min(z, x)), a) ||
              rewrite(max(max(x, y), min(z, y)), a) ||
 
+             rewrite(max(select(x, max(z, y), w), z), max(select(x, y, w), z)) ||
```
abadams, line 100: Worth adding the versions where the else case of the select is the redundant max?

src/Simplify_Max.cpp
```diff
@@ -97,6 +97,10 @@ Expr Simplify::visit(const Max *op, ExprInfo *bounds) {
...
+             rewrite(max(select(x, max(z, y), w), y), max(select(x, z, w), y)) ||
+             rewrite(max(min(x*c0, y + c1) - min(select(0 < x, y, min(x*c0, y + c1)), x*c0), 0), 0, c1 < 0) ||
```
abadams, line 102: This one is probably too specific

src/Simplify_Min.cpp
```diff
@@ -97,6 +97,10 @@ Expr Simplify::visit(const Min *op, ExprInfo *bounds) {
              rewrite(min(max(x, y), min(z, x)), b) ||
              rewrite(min(max(x, y), min(z, y)), b) ||
 
+             rewrite(min(max(x, y + c0), y), y, c0 > 0) ||
+             rewrite(min(select(x, min(z, y), w), y), min(select(x, z, w), y)) ||
```
abadams, line 101: Worth adding versions where the min is in the else case?

src/Simplify_Mul.cpp
```diff
@@ -79,6 +79,12 @@ Expr Simplify::visit(const Mul *op, ExprInfo *bounds) {
             rewrite(x * (y * c0), (x * y) * c0) ||
             rewrite(max(x, y) * min(x, y), x * y) ||
             rewrite(max(x, y) * min(y, x), y * x) ||
+
+            rewrite(min(x * c0, c1) * c2, min(x * fold(c0 * c2), fold(c1 * c2)), c2 > 0) ||
```
abadams, line 83: I'm not comfortable with rules that makes constants larger, because it risks triggering signed integer overflow, and we're not completely sound with respect to that.

src/Simplify_Add.cpp
rootjalex: copy-paste error, will fix

src/Simplify_Max.cpp
rootjalex: I don't think I saw these in logging, but it can't hurt. Will add

src/Simplify_Mul.cpp
rootjalex: In one of the above comments, you mentioned that a rule will fail to match if it detects overflow - so is this specifically not wanting to produce larger constants with multiplication? Regardless, I removed these in a827ae5

rootjalex:
@abadams I believe I have addressed all of your comments - I did not see the min-select and max-select variants in my logging but they do seem useful, so I added them as you requested.

src/Simplify_Mul.cpp
rootjalex: It also appears that the second of these rules was causing the fuzz_simplify failure on the buildbots of 0a00af4 , so the lack of comfort with these rules seems plenty reasonable.

abadams (approved):

PR #6161, opened 2021-07-23
Update tutorial todos

abadams:
This is based on our discussion in the dev meeting. Feel free to suggest changes.

### DIFF
--- a/tutorial/todo.txt
+++ b/tutorial/todo.txt
@@ -1,11 +1,24 @@
-- debug_to_file
-- the bounds query interface
+High priority:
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):

alexreinking:
Would be nice to show bounds proxy expressions, too.

Also would be nice to have two CMake chapters (one for JIT and another for AOT)

rootjalex:
++ to Alex's CMake suggestion

abadams:
I don't think we expose bounds proxy expressions in the front-end yet.

PR #6169, opened 2021-07-27
Use generic build command instead of make in run-clang-tidy.sh

### LINKED ISSUE #6163: run-clang-tidy.sh script is broken
At top-of-tree, running it locally (either directly or via make clang-tidy) fails with `*** No rule to make target `HalideIncludes'.  Stop.`. My guess is that some recent CMake change inadvertently broke it?

alexreinking:
Script had previously assumed CMake would generate makefiles. If Ninja is selected via the `CMAKE_GENERATOR` environment variable, the script would fail.

Fixes #6163

### DIFF
--- a/run-clang-tidy.sh
+++ b/run-clang-tidy.sh
@@ -47,7 +47,7 @@ cmake -DCMAKE_BUILD_TYPE=Debug \
 [ -a ${CLANG_TIDY_BUILD_DIR}/compile_commands.json ]
 
 # We must populate the includes directory to check things outside of src/
-cd ${CLANG_TIDY_BUILD_DIR} && make HalideIncludes
+cmake --build ${CLANG_TIDY_BUILD_DIR} --target HalideIncludes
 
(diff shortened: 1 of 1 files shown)


alexreinking:
Buildbot failures are unrelated.

run-clang-tidy.sh
```diff
@@ -47,7 +47,7 @@ cmake -DCMAKE_BUILD_TYPE=Debug \
 [ -a ${CLANG_TIDY_BUILD_DIR}/compile_commands.json ]
 
 # We must populate the includes directory to check things outside of src/
-cd ${CLANG_TIDY_BUILD_DIR} && make HalideIncludes
+cmake --build ${CLANG_TIDY_BUILD_DIR} --target HalideIncludes
```
abadams, line 50: I think the newline ends up with a different working directory than the old one. Everything that follows is absolute paths though, I think....

Did you test this? If so, lgtm.

alexreinking: I did test it and it worked locally.

abadams (approved):

PR #6174, opened 2021-07-29
Rewrite Rules Evaluation Project - Merging Relevant Synthesized Rewrite Rules

evanleenyc:
Conducted experiments to analyze the performance effects of adding 4000+ synthesized rewrite rules to Halide.
Narrowed down the rules to 11 rewrite rules whose associative & commutative variants are added in this PR. 
With these rewrite rules, Halide achieves >10% peak memory reductions in 192 cases in apps including camera_pipe, harris, nl_means, and stencil_chain, which is similar to the results (with all 4000+ rules) from this paper - https://dl.acm.org/doi/pdf/10.1145/3428234


src/Simplify_Mul.cpp
```diff
@@ -80,6 +80,9 @@ Expr Simplify::visit(const Mul *op, ExprInfo *bounds) {
             rewrite(max(x, y) * min(x, y), x * y) ||
             rewrite(max(x, y) * min(y, x), y * x) ||
 
+            rewrite(x * select(y, 1, 0), select(y, x, 0)) ||
+            rewrite(x * select(y, 1, 0), select(y, x, 0)) ||
```
abadams, line 84: These two appear to be the same. Was one supposed to be commuted?

evanleenyc: Yes. They somehow became the same rule when I ran it through filter_rewrite_rules and got me duplicating it..

src/Simplify_Select.cpp
```diff
@@ -191,6 +210,7 @@ Expr Simplify::visit(const Select *op, ExprInfo *bounds) {
              (no_overflow_int(op->type) &&
               (rewrite(select(x, y * c0, c1), select(x, y, fold(c1 / c0)) * c0, c1 % c0 == 0) ||
                rewrite(select(x, c0, y * c1), select(x, fold(c0 / c1), y) * c1, c0 % c1 == 0) ||
+               rewrite(select(x, y + c0, c1), select(x, y, fold(c1 - c0)) + c0) ||
```
rootjalex, line 213: This rule is causing the correctness_simplify failures. We should either remove it or possibly add a `c1 != 0` predicate.

test/correctness/simplify.cpp
```diff
@@ -1713,6 +1713,7 @@ void check_boolean() {
     check(select(x == 3, 5, y) - y, select(x == 3, 5 - y, 0));
     check(select(x == 3, y, 5) - y, select(x == 3, 0, 5 - y));
     check(y - select(x == 3, 5, y), select(x == 3, y + (-5), 0));
+    check(y - select(x == 3, 5, y), select(x == 3, y, 5) + (-5));
```
rootjalex, line 1716: I think you intended to replace the line above this with this one?

evanleenyc: Yes my bad..

rootjalex:
@steven-johnson are the buildbot failures due to an LLVM breakage, or are they due to this PR? I don't see the failure on oher PRs so I am assuming the latter, but just wanting to check

steven-johnson:
> @steven-johnson are the buildbot failures due to an LLVM breakage, or are they due to this PR? I don't see the failure on other PRs so I am assuming the latter, but just wanting to check

LLVM breakage. Should be safe to ignore.

rootjalex (approved):

PR #6175, opened 2021-07-29
Fix Xcode issue that requires at least one source file when building a library from objects.

### LINKED ISSUE #6167: CMake Xcode generator appears broken
```
cmake_minimum_required (VERSION 3.16.0)

project (HalideExample)

set (CMAKE_CXX_STANDARD 17)
set (CMAKE_CXX_STANDARD_REQUIRED YES)
set (CMAKE_CXX_EXTENSIONS NO)

find_package(Halide REQUIRED)

add_executable (my_generators my_generators.cpp)
target_link_libraries (my_generators PRIVATE Halide::Generator)

add_halide_library (my_first_generator FROM my_generators)

add_executable (app main.cpp)
target_link_libraries (app Halide::Halide my_first_generator)
```

This seems to work just fine when using Ninja generator. It will produce `my_first_generator.o` and `my_first_generator.h` as well as `libmy_first_generator.a`. When using Xcode generator all I get are the first two files. Inspecting the generated project tells me that `app` executable depends on `libmy_first_generator.a` as well but this file is not present. I also see that `my_generators` executable is missing in this case...

So it feels like `my_generators` hasn't been built and run.

Any ideas?

alexreinking:
**ATTN:** @popizdeh -- please test this and confirm whether it works for you. 

We don't test Xcode on CI (we use Ninja on all platforms) so adding `skip_buildbots` here.

---

Fixes #6167


popizdeh:
Looking good on my end Alex. Thanks a lot!

abadams (approved):

Issue #2071, opened 2017-05-24
C/C++ backend doesn't emit _argv() wrapper or _metadata() accessor


steven-johnson:
These are rarely used, and thus not necessarily a deal-breaker for many users, but it does make the C/C++ backend incomplete relative to other backends.

zvookin:
Is it possible to generate Halide IR which compiles to these accessors? If
so, it would be cool to change the way we make these to go through IR (e.g.
adding a couple of LoweredFuncs to the Halide::Module) instead of being
specific to LLVM codegen.

-Z-

steven-johnson:
Good question. I'm not sure.

steven-johnson:
Fixed by #6179

PR #6179, opened 2021-08-05
Add argv and metadata support to C++ backend (Issue #2071)

steven-johnson:


src/CodeGen_C.cpp
```diff
@@ -1489,6 +1489,252 @@ void CodeGen_C::forward_declare_type_if_needed(const Type &t) {
...
+        stream << "\nHALIDE_FUNCTION_ATTRS\nconst struct halide_filter_metadata_t *" << function_name << "_metadata();\n";
+        return;
+    }
+
+    auto map_name = [&metadata_name_map](const std::string &from) -> std::string {
+        auto it = metadata_name_map.find(from);
+        return it == metadata_name_map.end() ? from : it->second;
+    };
+
+    auto legalize_name = [](const std::string &from) -> std::string {
```
vksnk, line 1539: There is a `c_print_name` function, which does something similar. Is it possible to use it here instead?

vksnk (approved):
Thanks a lot for adding this!

src/CodeGen_C.cpp
steven-johnson: Done

steven-johnson:
Failures are unrelated.

PR #6181, opened 2021-08-09
RFC: Output universal binary when CMAKE_OSX_ARCHITECTURES is defined.

popizdeh:
@alexreinking this is to start the conversation about supporting universal binaries. This patch allowed me to build Halide as a universal library and to generate filters as universal libraries from CMake.

I'm only vaguely familiar with CMake so please excuse my code :)

There's probably a better way to do this, I saw the `is_crosscompiling` argument but I also read somewhere that arguments to `target=` need to be the same architecture. I'm not sure where this requirement is coming from but this is the main reason I went with the approach of generating per architecture and fusing them together.

Basic idea is pretty simple, CMAKE_OSX_ARCHITECTURES is usually set to "arm64,x86_64" when compiling for Big Sur. Generator is invoked once per architecture and two object files are then merged into a fat file. Separate command is used to generate all the other outputs like header file, registration file, etc. Same thing is done for runtime library.

### DIFF
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -158,7 +158,7 @@ function(add_halide_library TARGET)
         # When cross-compiling, we need to use a static, imported library
         list(APPEND generator_outputs static_library)
         set(generator_sources "${TARGET}${static_library_suffix}")
-    else ()
+    elseif (NOT CMAKE_OSX_ARCHITECTURES)
         # When compiling for the current CMake toolchain, create a native
         list(APPEND generator_outputs object)
         list(LENGTH ARG_TARGETS len)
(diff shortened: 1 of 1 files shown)


alexreinking:
I need access to an M1 Mac to test this. @steven-johnson - can I use the buildbot for a while?

steven-johnson:
> I need access to an M1 Mac to test this. @steven-johnson - can I use the buildbot for a while?

OK with me, but @abadams is the one to check with

alexreinking:
@abadams - cool to use the M1 mac for a while? Maybe over the weekend?

steven-johnson:
Any updates on this PR?

steven-johnson:

alexreinking:
I haven't found a time to sit down with an Apple Silicon machine to myself and test this yet. But I would like to!

PR #6182, opened 2021-08-09
Augment debugging code

steven-johnson:
I upgraded some debugging code in AddImageChecks and Bounds while tracking down a bug, and I think the upgrades are worth keeping for future use.

### DIFF
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -355,9 +355,9 @@ Stmt add_image_checks_inner(Stmt s,
             debug(3) << "In image " << name << " region touched is:\n";
             for (int j = 0; j < dimensions; j++) {
                 debug(3) << "  " << j << ": " << (touched.empty() ? Expr() : touched[j].min)
-                                              << " .. "
-                                              << (touched.empty() ? Expr() : touched[j].max)
-                                              <<"\n";
+                         << " .. "
+                         << (touched.empty() ? Expr() : touched[j].max)
+                         << "\n";
             }
         }
 
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -351,7 +351,15 @@ Stmt add_image_checks_inner(Stmt s,
         }
 
         // Check that the region passed in (after applying constraints) is within the region used
-        debug(3) << "In image " << name << " region touched is:\n";
+        if (debug::debug_level() >= 3) {
+            debug(3) << "In image " << name << " region touched is:\n";
+            for (int j = 0; j < dimensions; j++) {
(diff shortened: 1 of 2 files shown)


src/AddImageChecks.cpp
```diff
@@ -351,7 +351,15 @@ Stmt add_image_checks_inner(Stmt s,
         }
 
         // Check that the region passed in (after applying constraints) is within the region used
-        debug(3) << "In image " << name << " region touched is:\n";
```
abadams, line 354: This old code is confusing... did there use to be a debug(3) inside the loop over j below too?

steven-johnson: It appeared to be a weird dangling thing, presumably left over from some earlier refactoring

abadams (approved):

src/Bounds.cpp
```diff
@@ -142,20 +142,33 @@ class Bounds : public IRVisitor {
...
+public:
     int interval_log_indent = 0;
 
-    void log_interval(const std::string &msg) const {
+private:
+    std::string log_interval_spaces() const {
+        return std::string(interval_log_indent * 2, ' ');
+    }
+
+    void log_interval_exit(const std::string &msg) const {
```
rootjalex, line 153: Is `msg` being used in this method?

src/Bounds.cpp
```diff
@@ -1664,7 +1700,7 @@ Region region_union(const Region &a, const Region &b) {
         Expr max_plus_one = Max::make(max_a, max_b);
         Expr extent = max_plus_one - min;
         result.push_back(Range(simplify(min), simplify(extent)));
-        //result.push_back(Range(min, extent));
+        // result.push_back(Range(min, extent));
```
rootjalex, line 1703: Is there any reason to not remove this line?

src/Bounds.cpp
```diff
@@ -2789,6 +2825,27 @@ map<string, Box> boxes_touched(const Expr &e, Stmt s, bool consider_calls, bool
             s.accept(&provides);
         }
     }
+
+    // for (const auto &it : calls.boxes) {
```
rootjalex, line 2829: Same here, should these be in an `#if` block or just removed?

src/Bounds.cpp
steven-johnson: Not at present. (It was before). Left in because it might be useful for future augmentation.

src/Bounds.cpp
steven-johnson: I didn't touch this, clang-format just tweaked the comment. But yeah, I see no reason for it to stay.

PR #6184, opened 2021-08-10
Add a watchdog timer to Generator

steven-johnson:
In degenerate conditions (eg, bugs in Halide or LLVM, or pathological user code), running a Generator can take arbitrarily long times (we recently found some buildbots that had Generators that had been running for several days).

This adds a simple background thread to generator_main() to ensure that compilations don't take unreasonable lengths of time. It defaults to 15 minutes of wall-clock time, but can be customized by the -t flag.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1,5 +1,7 @@
+#include <atomic>
 #include <cmath>
 #include <fstream>
+#include <thread>
 #include <unordered_map>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


src/Generator.cpp
```diff
@@ -985,6 +990,25 @@ int generate_filter_main_inner(int argc, char **argv, std::ostream &error_output
...
 
+    std::atomic<bool> generator_finished = false;
+    int timeout_in_seconds = std::stoi(flags_info["-t"]);
+    // Use ~1 year for "infinite" timeouts.
+    if (timeout_in_seconds <= 0) {
+        timeout_in_seconds = 60 * 24 * 365;
+    }
+    const auto timeout_time = std::chrono::steady_clock::now() + std::chrono::seconds(timeout_in_seconds);
+    std::thread timeout_monitor([timeout_time, timeout_in_seconds, &generator_finished]() {
+        while (!generator_finished) {
```
alexreinking, line 1001: Why not just put `if (timeout_in_seconds <= 0) { return; }` to the lambda?

alexreinking (approved):
Approved with nit. Setting it to a year is a bit arbitrary... better to not launch the thread at all in these cases, no?

src/Generator.cpp
steven-johnson: Done

PR #6185, opened 2021-08-11
More augmentation of debugging code

steven-johnson:
This expands on #6182 by added tracking for BoxesTouched, and integrating the nesting levels with the previous code. This allows a more complete vision of what's happening during bounds calculation.


src/Bounds.cpp
```diff
@@ -1984,6 +1966,133 @@ class BoxesTouched : public IRGraphVisitor {
...
+        }
+
+        ~BoxesTouchedLogger() {
+            log_box_diffs(this->boxes, self->boxes);
+            self->log_indent--;
+            log_line('}');
+            internal_assert(self->current_logger == this);
+            if (parent_logger) {
+                // Propagate changes to our parent.
+                // This isn't efficient at all, but it's ususally-disabled debugging code.
```
abadams, line 2059: ususally

src/Bounds.cpp
```diff
@@ -2845,7 +2967,7 @@ map<string, Box> boxes_touched(const Expr &e, Stmt s, bool consider_calls, bool
             debug(0) << "  " << j << ": " << it.second[j].min
                      << " .. "
                      << it.second[j].max
-                     << "\n";
+                     << '\n';
```
abadams, line 2970: I dug into benchmarking `<< '\n'` vs `<< "\n"` at one time, and weirdly enough, `<< "\n"` is faster. Sending a single character to a stream goes through a weird slow path.

abadams: (in libstdc++)

src/Bounds.h
```diff
@@ -33,7 +33,8 @@ const FuncValueBounds &empty_func_value_bounds();
 Interval bounds_of_expr_in_scope(const Expr &expr,
                                  const Scope<Interval> &scope,
                                  const FuncValueBounds &func_bounds = empty_func_value_bounds(),
-                                 bool const_bound = false);
+                                 bool const_bound = false,
+                                 int indent = 0);
```
abadams, line 37: Is this indent parameter only used internally? I don't think we should expose it.

src/Bounds.cpp
steven-johnson: Done

src/Bounds.cpp
steven-johnson: Done

src/Bounds.h
steven-johnson: Yep. I just felt like the alternative (making an internal-only wrapper) was even uglier. Let me look again.

steven-johnson: PTAL

abadams (approved):

PR #6189, opened 2021-08-11
Internal::promise_clamped() should be pure (Fixes #6186)

### LINKED ISSUE #6186: Incorrect bounds checking for some schedules
Run the following:

```
    Var x, y, c, chunk;

    ImageParam input(UInt(8), 3);

    Func intermed, output;
    intermed(x, y, c) = input(x, y, c);
    output(x, y, c) = intermed(x, y, c);

    // schedule
    intermed.compute_at(output, y);
    output.split(y, chunk, y, 64, TailStrategy::GuardWithIf);

    Buffer<uint8_t> input_buf(100, 100, 3);
    input_buf.fill(0);
    input.set(input_buf);

    auto result = output.realize({100, 100, 3});

    printf("Success!\n");

    return 0;
}
```

Expected: `Success`
Actual: `Input buffer input is accessed at 127, which is beyond the max (99) in dimension 1`

Initial debugging points to #6126 as the injection point.

steven-johnson:


src/IROperator.h
```diff
@@ -1470,8 +1470,12 @@ namespace Internal {
  * context-dependent, because 'value' might be statically bounded at
  * some point in the IR (e.g. due to a containing if statement), but
  * not elsewhere.
+ *
+ * If pure = true, the call is marked as a PureIntrinsic, allowing hoisting
+ * of the value; the caller must be (even more) vigilant in this case to
+ * ensure that the promise will remain valid in such conditions.
  **/
-Expr promise_clamped(const Expr &value, const Expr &min, const Expr &max);
```
abadams, line 1474: I think promise_clamped is always pure if you use the definition I gave offline, and you don't need the pure flag:

"This intrinsic evaluates to its first argument. If this value is used by a side-effecting operation and it is outside the range specified by its second and third arguments, behavior is undefined. The compiler can therefore assume that the value is within the range given and optimize accordingly. Note that this permits promise_clamped to evaluate to something outside of the range, provided that this value is not used."

abadams: That makes it legal to hoist.

steven-johnson: So are you suggesting that we should change the body of `promise_clamped()` to always use PureIntrinsic?

abadams: Yes

steven-johnson: PTAL

abadams:
I don't think pure_ is required as part of the name. Most instrinsics are pure, after all, and unsafe_promise_clamped is pure too.

steven-johnson:
> unsafe_promise_clamped is pure too.

Err, no it's not. Look at the body of `unsafe_promise_clamped()`.

abadams:
Oh that's weird. unsafe_promise_clamped should really be pure too. The fact it asserts is true for the entire lifetime of the program!

dsharletg (approved):
This looks good to me know and it looks like all of Andrew's feedback has been addressed. I think we should change unsafe_promise_clamped to pure in a separate PR, so we can most safely get this merged and unblock our internal Halide updates.

steven-johnson:
> This looks good to me know and it looks like all of Andrew's feedback has been addressed. 

Running a test inside Google with this change hand-patched in to see if there are new regressions.

> I think we should change unsafe_promise_clamped to pure in a separate PR

Will do.

steven-johnson:
No failures in google that appear to be related in any way; I think we are good to land.

PR #6194, opened 2021-08-11
[Hexagon] Do not pattern match inside if_then_else block

aankit-ca:
Resolves the compilation below compilation error while generating
hannk::upsample_channels_uint8 from hannk/depthwise_conv.generator
for HVX:

Unknown intrinsic dynamic_shuffle

The problem occurs when we pattern match hvx instrinsics inside
if_then_else nodes and try to scalarize them later. In the patch we
prevent matching these intrinsics inside if_then_else blocks.

@dsharletg @pranavb-ca PTAL

### DIFF
--- a/src/HexagonOptimize.cpp
+++ b/src/HexagonOptimize.cpp
@@ -914,6 +914,9 @@ private:
     }
 
     Expr visit(const Call *op) override {
+        if (op->is_intrinsic(Call::if_then_else)) {
+            return op;
+        }
         if (op->is_intrinsic(Call::widening_add)) {
             Expr mpyadds = find_mpyadds(Add::make(cast(op->type, op->args[0]), cast(op->type, op->args[1])));
             if (mpyadds.defined()) {
@@ -1071,6 +1074,13 @@ class VectorReducePatterns : public IRMutator {
             return true;
         }
         return false;
+    }
+
+    Expr visit(const Call *op) override {
+        if (op->is_intrinsic(Call::if_then_else)) {
+            return op;
+        }
+        return IRMutator::visit(op);
     }
 
     Expr visit(const VectorReduce *op) override {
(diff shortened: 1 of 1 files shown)


src/HexagonOptimize.cpp
```diff
@@ -914,6 +914,9 @@ class OptimizePatterns : public IRMutator {
     }
 
     Expr visit(const Call *op) override {
+        if (op->is_intrinsic(Call::if_then_else)) {
```
dsharletg, line 917: I think you should only skip this if the condition is a vector. If the condition is a scalar it should be fine and I think might be an important case. Same for the other 4 instances.

aankit-ca: Done

dsharletg: This is different from what I suggested. This skips it for vector *results*, which are common. The case I am thinking about that we should still match patterns for is if_then_else where the *condition* (the first argument of the call) is a scalar, but the result and operands are vectors.

aankit-ca: Oh understood now. I'm assuming we should pattern match where conditions are Broadcasts as well?

dsharletg: I think so. I think we might see if_then_else with both scalar or broadcasting conditions, I'd have to poke around in the code to be sure one way or the other.

aankit-ca: Yeah. It seems like we can encounter both. Added condition for Broadcasts.

dsharletg (approved):

dsharletg:
The reason for this change has test coverage in hannk, correct? So once we enable testing of that on Hexagon on the buildbots, this will get coverage?

aankit-ca:
> The reason for this change has test coverage in hannk, correct? So once we enable testing of that on Hexagon on the buildbots, this will get coverage?

Yes

PR #6195, opened 2021-08-12
Move parallel/async lowering from LLVM codegen to a standard Halide IR lowering pass.

zvookin:
Time to get code review on this started as it will likely require more effort than average and the sooner this can get solid, the better.

The basic idea here is pretty simple: move codegen for parallel for loops and async chains from being done at the LLVM IR level to doing it at the Halide IR level. The direct goal is to allow using our concurrent runtime via the C++ backend. A secondary goal is to make the closure generation machinery reusable in other places.

This is all plumbed through but there are two issues that need to be addressed before it can be merged:

1) The method for determining whether to use the original ```halide_do_par_for``` depends on an LLVM codegen symbol table lookup that cannot work at the Halide IR level. I believe this should be resolvable by doing a walk of the Halide IR tree, but haven't yet implemented this. (I will do so shortly.)

2) Some tests use the call graph checker to validate a dependency graph of logical calls between Funcs. The checker does not recover the full set of dependencies across lowered functions when closures are made for parallel or async scheduled Funcs. @abadams and I will have to figure out how to handle this. For now the failing tests are disabled with a TODO. (correctness_func_wrapper, correctness_image_wrapper, and correctness_rfactor are the tests in question.)

Making this work required adding some intrinsics to declare and populate typed structures. The main technique here is that declaring a struct returns a properly typed null pointer to that struct and the fully elaborated backed type can be recovered from this null pointer. (In LLVM this is just using the ```getType()``` method on the ```Value``` and in C++ it is recovered using the ```decltype```. This is to work around Halide's type system not being rich enough to handle everything in C++ (or LLVM for that matter). It might be possible to register more detailed type info for Handle types for a given compilation, but that seemed more complicated and likely requires registering dynamically created data to the compiler with an unknown lifetime, thus posing a memory management problem.

New intrinsics are ```make_struct_type```, ```make_typed_struct```, ```load_struct_member```, ```resolve_function_name```, ```get_user_context```, and ```get_pointer_symbol_or_null```. I will not defend this set of additions as a shining example of design, but it also certainly isn't lowering the bar for Halide intrinsics. Happy to consider alternative approaches if folks have ideas.


src/CodeGen_C.cpp
```diff
@@ -2320,7 +2336,7 @@ void CodeGen_C::visit(const Call *op) {
         string arg1 = print_expr(op->args[1]);
         rhs << "return_second(" << arg0 << ", " << arg1 << ")";
     } else if (op->is_intrinsic(Call::if_then_else)) {
-        internal_assert(op->args.size() == 2 || op->args.size() == 3);
+        internal_assert(op->args.size() == 3);
```
dsharletg, line 2339: We need to keep supporting 2-argument versions of if_then_else

src/IR.cpp
```diff
@@ -652,6 +652,12 @@ const char *const intrinsic_op_names[] = {
     "widening_shift_left",
     "widening_shift_right",
     "widening_sub",
+    "make_struct_type",
```
dsharletg, line 655: Should this be renamed? It's odd for an expression to return a type. Maybe make_null_typed_struct?

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,506 @@
...
+namespace Halide {
+namespace Internal {
+
+// TODO(zalman): Find a better place for this code to live.
+LoweredFunc GenerateClosureIR(const std::string &name, const Closure &closure,
+                              std::vector<LoweredArgument> &args, int closure_arg_index,
+                              const Stmt &body, const Target &t) {
+    // Figure out if user_context has to be dealt with here.
+    std::string closure_arg_name = unique_name("closure_arg");
+    args[closure_arg_index] = LoweredArgument(closure_arg_name, Argument::Kind::InputScalar,
```
dsharletg, line 23: I thought we needed to keep some modulus/remainder and other analysis through closures like this. But I don't see that logic in HexagonOffload.cpp... so I'm not sure. I think maybe we should add a test that checks that loads/stores are aligned when expected (if we don't have one already), and make sure those loads/stores are in a parallel loop.

I think we used to propagate this through LoweredArgument, but then when we started adding alignment metadata to loads/stores directly, we stopped doing that.

dsharletg (commented):
Just a quick first pass through this.

alexreinking:
Looks like you need to update `src/CMakeLists.txt` in the same way you updated the Makefile.

src/IR.cpp
zvookin: I renamed it ```declare_struct_type``` which is perhaps more accurate. It returns a value of the type, which can be used to apply the type to a different value later. Whether this is odd or not probably depends on one's viewpoint, but it is a way for the backend type to go from one place to another without Halide IR knowing about it and is also the sort of thing that shows up in template metaprogramming a fair bit.

src/LowerParallelTasks.cpp
zvookin: Is the idea of the test to have an input buffer with host alignment asserted and then make sure a load against that buffer inside a closure is aligned? (E.g. by compiling to a fixed CPU target and pattern matching the asm output, or possibly doing it on LLVM IR.)

dsharletg: We also need to have constraints required to make the load aligned, e.g. a 2D buffer where the stride in y is constrained to be a multiple of the alignment.

I think it would be sufficient to check the Halide IR after lowering, that the load has the correct alignment annotation. No need to check LLVM IR/assembly I don't think.

zvookin: This pass happens at the very end of lowering, so if there was something in the Halide IR, it will already be there. (The place in lowering is sort of dictated by it being moved from backend specific codegen to a lowering pass.) But if there is lookaside or schedule information, that might need to be copied onto a new LoweredArgument or similar. I think if we really want a reliable test, it will need to validate some property of the final output. A Halide IR test may also be useful.

src/CodeGen_C.cpp
```diff
@@ -1300,7 +1311,12 @@ string CodeGen_C::print_reinterpret(Type type, const Expr &e) {
     } else {
         oss << "reinterpret<" << print_type(type) << ">";
     }
-    oss << "(" << print_expr(e) << ")";
+    if (type.is_handle() && !is_const_pointer(type) &&
+        is_const_pointer(e.type())) {
+        oss << "const_cast<std::add_pointer<std::remove_const<std::remove_pointer<decltype(" << print_expr(e) << ")>::type>::type>::type>(" << print_expr(e) << ")";
```
steven-johnson, line 1316: ..."The Aristocrats"!

src/CodeGen_LLVM.cpp
```diff
@@ -2839,6 +2840,119 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+                Value *elem_ptr;
+                size_t indices_index = 0;
+                std::vector<Value *> indices(2);
+                indices[indices_index++] = ConstantInt::get(i32_t, item);
+                indices[indices_index++] = ConstantInt::get(i32_t, i);
+                elem_ptr = CreateInBoundsGEP(builder, value, indices);
+                Value *init_value = args[i + item * initializer_count];
+                // TODO(zalman): Halide IR is not strongly typed re: pointers,
+                // especially to structs. Specific case that is failing is
+                // halide_semaphort_t *.
```
steven-johnson, line 2903: nit: typo

src/CodeGen_C.cpp
zvookin: Had to come up with another answer anyway...

zvookin:
This should be ready to review. I have no idea what the CMaker errors are.

steven-johnson:
> This should be ready to review. I have no idea what the CMaker errors are.

I restarted the failed ones. It appears so far to just be a temporary burp in LLVM.

steven-johnson:
> > This should be ready to review. I have no idea what the CMaker errors are.
> 
> I restarted the failed ones. It appears so far to just be a temporary burp in LLVM.

Correction, nope, and I can repro locally on our master branch. Investigating.

steven-johnson:
It appears that the issue is that building apps/hannk requires us to get the TFLite source, which in turn gets the source to libeigen from GitLab (https://gitlab.com/libeigen/eigen)... which is currently offline (confirmed by other systems that use it). Not sure if related to other current major net outages. I'll temporarily disable testing for apps/hannk until this is resolved.

src/CodeGen_C.cpp
```diff
@@ -2452,6 +2470,110 @@ void CodeGen_C::visit(const Call *op) {
...
+        const Call *decl_call = op->args[0].as<Call>();
+        internal_assert(decl_call != nullptr);
+        std::string c_name = c_print_name(decl_call->name, false);
+        rhs << "(&" << c_name << ")";
+    } else if (op->is_intrinsic(Call::get_user_context)) {
+        internal_assert(op->args.empty());
+        rhs << "_ucon";
+    } else if (op->is_intrinsic(Call::get_pointer_symbol_or_null)) {
+        internal_assert(op->args.size() == 2);
+        // TODO(zalman|abadams): Figure out how to get rid of the mayby foo.buffer exists, maybe it doesn't thing in closures.
```
steven-johnson, line 2575: Nit: "maybe foo.buffer exists, maybe it doesn't thing" in closures

src/CodeGen_LLVM.cpp
```diff
@@ -2840,6 +2841,147 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+                size_t indices_index = 0;
+                std::vector<Value *> indices(2);
+                indices[indices_index++] = ConstantInt::get(i32_t, item);
+                indices[indices_index++] = ConstantInt::get(i32_t, i);
+                elem_ptr = CreateInBoundsGEP(builder, value, indices);
+                Value *init_value = args[i + item * initializer_count];
+                // TODO(zalman): This is a bit of a hack. Halide IR is not strongly typed re: pointers,
+                // especially to structs. Specific case that is failing is halide_semaphore_t *.
+                //
+                // This first case converts an i8_t* (the represenation of void * being used in LLVM) to
```
steven-johnson, line 2905: representation

src/CodeGen_LLVM.cpp
```diff
@@ -2989,11 +3133,15 @@ void CodeGen_LLVM::visit(const Call *op) {
         // restrictions if we recognize the most common types we
         // expect to get alloca'd.
         const Call *call = op->args[0].as<Call>();
+        const int64_t *sz = as_const_int(op->args[0]);
         if (op->type == type_of<struct halide_buffer_t *>() &&
             call && call->is_intrinsic(Call::size_of_halide_buffer_t)) {
             value = create_alloca_at_entry(halide_buffer_t_type, 1);
+        } else if (op->type == type_of<struct halide_semaphore_t *>() &&
+                   semaphore_t_type != nullptr &&
+                   sz && *sz == 16) {
```
steven-johnson, line 3142: Nit: wouldn't it be better to use `sizeof(halide_semaphore_t)` instead of 16?

src/Lower.cpp
```diff
@@ -424,6 +427,24 @@ void lower_impl(const vector<Function> &output_funcs,
         debug(1) << "Skipping GPU offload...\n";
     }
 
+    std::vector<LoweredFunc> closure_implementations;
+    debug(1) << "Lowering Parallel Tasks...\n";
+    s = lower_parallel_tasks(s, closure_implementations, pipeline_name, t);
+    // Process any LoweredFunctions added by other passes. In practice, this
```
steven-johnson, line 433: The "In practice, ..." comment makes me wonder if this should be an error condition instead. Does it at least warrant a TODO and an Issue?

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,519 @@
...
+                bool result = false;
+            };
+
+            // TODO(zvookin|abadams): This makes multiple passes over the
+            // IR to cover each node. (One tree walk produces the min
+            // thread count for all nodes, but we redo each subtree when
+            // compiling a given node.) Ideally we'd move to a lowering pass
+            // that converts our parallelism constructs to Call nodes, or
+            // direct hardware operations in some cases.
+            // Also, this code has to exactly mirror the logic in get_parallel_tasks.
```
steven-johnson, line 226: At the least, we should probably add a similar comment to get_parallel_tasks() to reduce the chances the code diverges.

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,519 @@
...
+        return LetStmt::make(closure_result_name, result, AssertStmt::make(closure_result == 0, closure_result));
+    }
+
+    void get_parallel_tasks(const Stmt &s, std::vector<ParallelTask> &result, std::pair<std::string, int> prefix) {
+        const For *loop = s.as<For>();
+        const Acquire *acquire = loop ? loop->body.as<Acquire>() : s.as<Acquire>();
+        if (const Fork *f = s.as<Fork>()) {
+            add_fork(prefix);
+            get_parallel_tasks(f->first, result, prefix);
+            get_parallel_tasks(f->rest, result, prefix);
```
steven-johnson, line 447: Do Forks ever get nested deeply enough that we should worry about recursion here?

steven-johnson:
Looks like the failures on buildbot/halide-testbranch-main-llvm14-x86-64-linux-cmake are possibly-legit failures related to hvx

steven-johnson:
Where does this PR stand? Awaiting further changes/fixes?

src/CodeGen_LLVM.cpp
zvookin: It's the size in the target, not the compiler.

steven-johnson:
The failure in apps/blur appears to be pre-existing (see https://github.com/halide/Halide/issues/6440); since all the other remaining failures appear to be HVX-specific, we may have a latent memory-corruption error that is just being tickled by these changes. Investigating, but having someone at QC also take a look at this would likely be helpful.

steven-johnson:
I've gotten a pretty spare, cut-down repro case for apps/blur; apply the diff at https://gist.github.com/steven-johnson/db8073dbc0a5d725e0f21e9268e037c2 to get it, but in a nutshell, I nuked the C++ SIMD version, reduced the 'blur' to being a plain copy, removed all scheduling except for `parallel(y)`, and reduced the input/output buffers to a fixed 1x2 size, where each value is equal to the y coordinate:

- Apply the diff
- make distrib
- cd apps/blur
- make sure you have LD_LIBRARY_PATH set so that the HVX simulator, etc can all be found
- `HL_NUM_THREADS=1 OPTIMIZE=-g HL_TARGET=host-hvx make -j72 test`
- Ignore the "Success!" output, and look at the differences listed (spoiler: there should be no differences)

src/CodeGen_C.cpp
```diff
@@ -2458,6 +2476,110 @@ void CodeGen_C::visit(const Call *op) {
...
+    } else if (op->is_intrinsic(Call::resolve_function_name)) {
+        internal_assert(op->args.size() == 1);
+        const Call *decl_call = op->args[0].as<Call>();
+        internal_assert(decl_call != nullptr);
+        std::string c_name = c_print_name(decl_call->name, false);
+        rhs << "(&" << c_name << ")";
+    } else if (op->is_intrinsic(Call::get_user_context)) {
+        internal_assert(op->args.empty());
+        rhs << "_ucon";
+    } else if (op->is_intrinsic(Call::get_pointer_symbol_or_null)) {
```
steven-johnson, line 2579: The LLVM implementation of this symbol attempts a symbol lookup, but this just returns null unconditionally, which is mighty puzzling. Is it a bug, or do I just misunderstand the usage? (Documentation somewhere would be helpful)

src/CodeGen_C.cpp
```diff
@@ -2458,6 +2476,110 @@ void CodeGen_C::visit(const Call *op) {
...
+        internal_assert(op->args.size() >= 2);
+
+        const StringImm *str_imm = op->args[0].as<StringImm>();
+        internal_assert(str_imm != nullptr);
+        std::string name = c_print_name(str_imm->value, false);
+
+        const int64_t *mode = as_const_int(op->args[1]);
+        internal_assert(mode != nullptr);
+
+        if (*mode == 1 || !starts_with(name, "halide_")) {
```
steven-johnson, line 2493: The meaning of `mode` really should be documented somewhere, and the legal values probably deserve named constants. Having to read the implementation to figure out what it's for is not ideal.

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,519 @@
...
+            // TODO(zvookin|abadams): This makes multiple passes over the
+            // IR to cover each node. (One tree walk produces the min
+            // thread count for all nodes, but we redo each subtree when
+            // compiling a given node.) Ideally we'd move to a lowering pass
+            // that converts our parallelism constructs to Call nodes, or
+            // direct hardware operations in some cases.
+            // Also, this code has to exactly mirror the logic in get_parallel_tasks.
+            // It would be better to do one pass on the tree and centralize the task
+            // deduction logic in one place.
+            class MinThreads : public IRVisitor {
```
steven-johnson, line 229: style nit: IMHO this class is a little too complex to be included inside `rewrite_parallel_tasks`; the code would be more readable if the class were pulled out into a file-local class.

steven-johnson:
OK, I think I know what is going on, not sure of the proper fix yet:

Closure only finds buffers via Load/Store/Atomic IR nodes. In current master, that's all good. With this change, though, it means that buffers that are only referenced via the closure_arg to an offloaded Hexagon parallel loop don't get seen as buffers... but they do get seen as ordinary vars, so we do pass a valid symbol, but instead of being the (device, host) pair that the HVX code expects for buffer args, it's just a host ptr, and bad things ensue.

I think that to make this work properly, the call to `lower_parallel_tasks()` from HexagonOffload needs to somehow include the buffer information in the subsequent Closure, but whether this needs additional work inside that call (vs being able to patch it based on just e.g. the lowered_funcs returned) isn't clear to me yet.

steven-johnson:
So I think a reasonable fix might be as simple as moving the Closure gather in HexagonOffload.cpp:765 from after the call to `lower_parallel_tasks()` to just before it. I'm going to take the liberty of pushing such a fix to this PR and see how things look.

steven-johnson:
Looks like correctness_parallel_nested_1 is (still) failing, but only on llvm14-x86-64-linux-cmake?

dsharletg:
It's failing on Hexagon in particular, so it's probably not just a flake.

steven-johnson:
Definitely not a flake, but reverting 2fd4c9f doesn't seem to affect it -- either it was pre-existing or a different issue. Investigating.

steven-johnson:
git bisect seems to think the 'first bad commit' for this test is 0b7bd25 (i.e., the first commit of that PR), so presumably it's something we just haven't found yet -- debugging now

steven-johnson:
I think the real fix for the remaining issue is going to require a little surgery to `lower_parallel_tasks()` for it to calculate & return the proper Closure (on the pre-wrapped results); this may also be the non-hacky solution to the issue that commit 00a0715 was attempting to fix. Too late today for me to get this into shape; assuming my theory is correct, I'll prep this as a PR onto this PR, so it can be reviewed properly.

src/Lower.cpp
```diff
@@ -440,6 +443,33 @@ void lower_impl(const vector<Function> &output_funcs,
...
+    // will likely not work well enough due to ordering issues with
+    // closure generating passes and instead all such passes will need to
+    // be done at once.
+    for (size_t i = initial_lowered_function_count; i < result_module.functions().size(); i++) {
+        result_module.functions()[i].body =
+            lower_parallel_tasks(result_module.functions()[i].body, closure_implementations,
+                                 result_module.functions()[i].name, t);
+    }
+    for (auto &lowered_func : closure_implementations) {
```
steven-johnson, line 467: This is either a bug, or I just don't understand the code: `lower_parallel_tasks()` *replaces* the contents of `closure_implementations()`, rather than adding to it, so at this point, this list is just whatever it happened to be on the final call to `lower_parallel_tasks()` in the loop at line 462-466. Did we want to process the union of all of them? Or just the list from the first call to `lower_parallel_tasks()`?

steven-johnson:
> I think the real fix

I thought I had a good lead on this last night, but now it looks like I was mistaken; still investigating.

I have a simplified version of `correctness_parallel_nested_1` that still repros the failure here: https://gist.github.com/steven-johnson/c90941be99abbea47346ffc6e1097d24

steven-johnson:
Pretty sure that https://github.com/halide/Halide/pull/6457 is the necessary fix

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,508 @@
...
+                                     wrapped_body);
+        struct_index++;
+    }
+    for (const auto &b : closure.buffers) {
+        type_args[struct_index + 2] = make_zero(type_of<void *>());
+        type_args[struct_index + 3] = make_zero(type_of<halide_buffer_t *>());
+        wrapped_body = LetStmt::make(b.first,
+                                     Call::make(type_of<void *>(), Call::load_struct_member,
+                                                {closure_arg, struct_type, make_const(UInt(32), struct_index)},
+                                                Call::PureIntrinsic),
```
steven-johnson, line 50: This call is `PureIntrinsic` but the other two are `Intrinsic`. Is this intentional? If so, we should comment why. If not, we should fix it.

steven-johnson:
Finally green. 

It might be useful to have Google/Adobe users pull and try this out internally before we land it.

src/CodeGen_LLVM.cpp
```diff
@@ -2807,6 +2808,147 @@ void CodeGen_LLVM::visit(const Call *op) {
...
+            vector<llvm::Type *> types(op->args.size() - 2);
+            for (size_t i = 2; i < op->args.size(); i++) {
+                types[i - 2] = codegen(op->args[i])->getType();
+            }
+            struct_type = (llvm::Type *)llvm::StructType::create(*context, types, "struct." + name);
+        } else {
+            struct_type = get_llvm_struct_type_by_name(module.get(), ("struct." + name).c_str());
+            internal_assert(struct_type != nullptr) << "Failed to find structure " << name << "\n";
+        }
+        value = llvm::ConstantPointerNull::get(struct_type->getPointerTo());
```
abadams, line 2833: What are we going to do here if llvm finally removes types from pointers? Have a separate symbol table for types perhaps?

src/CodeGen_C.cpp
```diff
@@ -2458,6 +2476,110 @@ void CodeGen_C::visit(const Call *op) {
...
+                    stream << get_indent() << print_type(op->args[i].type(), AppendSpace) << "f_" << i - 2 << ";\n";
+                }
+                indent--;
+                stream << get_indent() << "};\n";
+            } else {
+                stream << ";\n";
+            }
+        }
+        rhs << "((" << name << " *)nullptr)";
+    } else if (op->is_intrinsic(Call::make_typed_struct)) {
```
abadams, line 2509: What's the difference between make_typed_struct and using a make_struct call with the appropriate return type?

abadams:
Looks like there are quite a few review concerns unaddressed, so I don't think it's ready to land, but FWIW it appears to work for the Adobe unit tests.

I'm totally confused by the need to pass types as values here. I'm not sure why closure generation and passing would require that.

steven-johnson:
> I'm totally confused by the need to pass types as values here. I'm not sure why closure generation and passing would require that.

There's definitely cleanup that could/should be done here, but there is some time pressure to land this for downstream users; assuming testing passes, I'm inclined to clean these issues up in a subsequent PR rather than delay this further.

src/IR.h
```diff
@@ -564,6 +564,12 @@ struct Call : public ExprNode<Call> {
         widening_shift_left,
         widening_shift_right,
         widening_sub,
+        declare_struct_type,
```
steven-johnson, line 567: Nit: the comment at the top of this enum asks to keep them sorted alphabetically.

dsharletg:
I don't think we should merge this before it's ready, it can wait a bit. It would be good to hear from @zvookin on the outstanding comments.

src/CodeGen_LLVM.cpp
zvookin: a) The main reason I'm doing this is to get LLVM IR to type check. I thought it would be safer, but we might just cast to the expected type at the point of use.

b) We can just track the type ourselves. Either by using our own wrapper on llvm::Value and an llvm::Type to carry it around, or via an adjunct table.

c) Do all the work to make sure the full C++ typed pointer is carried in the Halide IR.

It is all internal to our IR and changing it shouldn't cause user visible changes.

zvookin:
> I'm totally confused by the need to pass types as values here. I'm not sure why closure generation and passing would require that.

(The code is not literally passing types as values, rather it creates a value and recovers its type at another point in the generated code. Yes the value is only used to carry the type, but it is perfectly type safe and uses well established mechanisms in LLVM and C++. I want to make sure it is understood there is no weird casting between ```llvm::Type *``` and ```llvm::Value *``` or anything like that.)

Closure creation is effectively:

```
struct closure_struct_unique_name {
    // fields of this specific closure
};

int closure_callback_unique_name(void *closure_data_arg, ...) {
    closure_struct_unique_name *closure_data = (closure_struct_unique_name *)closure_data_arg;
    // unpack fields from closure_data.
    // ...
}

int some_halide_kernel(...) {
    closure_struct_unique_name closure_data = { ... };
    halide_do_par_for_or_something((void *)&closure_data, ...);
}
```

That needs to be modeled in Halide IR and codegened into LLVM IR and C++. I chose to do this by having a call to declare the type and return it as a typed pointer to that type. (One can declare just the type, and either with its fields or as a forward declaration to reference a preexisting type e.g. in the runtime. In that case a nullptr is returned. One can also declare the type as part of constructing a value of that type. Either value will serve to represent the type later.) Later codegen recovers the type from the pointer either using ```getType()``` in LLVM or ```decltype``` in C++. This is typesafe and allows adding other ways to produce such types in the future.

We also need to handle calling routines with pointer to structure type arguments, specifically in the runtime. This is transitive that in some cases, the pointer is inside another structure or array.

Other alternatives would be to binary pack everything and not use struct types at all. Except that we still need struct types when calling runtime routines that take a pointer to a specific structure type. The backend code needs to typecheck. So we have to pass the type somehow or recover it from the reference to the called runtime function. The latter is easy to do in LLVM, but very hard to do in C++.

In this alternative model, we almost certainly end up representing the struct type with a string that is its name. Though it probably is actually a string plus an int because it may be an array. This only works if the names are unique except not so unique that they don't refer back to the thing one wants them to when one wants them to. (E.g. the infamous C++ error messages about struct foo * not being type compatible with struct foo *. )

If LLVM were to completely eliminate types on pointers, it might be possible to just nerf this mechanism into non-existence for that backend and have things compile fine. I'll be a bit surprised if they remove pointer types to that degree, but if so, not a problem.

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+    for (const auto &v : closure.vars) {
+        Expr var = Variable::make(v.second, v.first);
+        closure_elements.emplace_back(var);
+        closure_types.emplace_back(var);
+    }
+    for (const auto &b : closure.buffers) {
+        Expr ptr_var = Variable::make(type_of<void *>(), b.first);
+        closure_elements.emplace_back(ptr_var);
+        closure_elements.emplace_back(Call::make(type_of<halide_buffer_t *>(), Call::get_pointer_symbol_or_null,
```
abadams, line 96: Why isn't this `Variable::make(type_of<halide_buffer_t *>(), b.first + ".buffer")`?

Why might the buffer exist sometimes and not exist other times, and it's OK if it doesn't exist?

If the buffer might not exist and that's OK for some reason, how about keeping a `Scope<Expr>` of all containing letstmts and using `scope.contains(buffer_name) ? scope.get(buffer_name) : reinterpret(Handle(), 0)`

src/LowerParallelTasks.cpp
```diff
@@ -0,0 +1,514 @@
...
+                tasks_array_args.emplace_back(min_threads);
+                tasks_array_args.emplace_back(Cast::make(Bool(), t.serial));
+            }
+        }
+
+        if (tasks_array_args.size() > 2) {
+            // Allocate task list array
+            Expr tasks_list = Call::make(Handle(), Call::make_typed_struct, tasks_array_args, Call::PureIntrinsic);
+            Expr user_context = Call::make(type_of<void *>(), Call::get_user_context, {}, Call::PureIntrinsic);
+            Expr task_parent = Call::make(Handle(), Call::get_pointer_symbol_or_null,
```
abadams, line 421: Same comment applies about keeping a `Scope<Expr>` instead

abadams:
What do you think about having a list of struct types inside the Module object instead of in the Stmt IR? A struct type would have a name and be either a primitive type or a vector of struct types. Then we'd have just the make_typed_struct and load_typed_struct_member instrinsics (or IR nodes) and they could refer to these global struct declarations by name.

The structs themselves would be passed around as Exprs of Handle() type. I think we should assume llvm *will* eventually remove all types from pointer values. That's the plan AFAIK, and the API reflects that. The loads and stores and GEPs are typed, but not the values themselves.

Finally, for function names as Handle()-valued Exprs, I think we should just use a Variable node with that name with type Handle(). These are symbols that are supposed to be in scope, so Variable is appropriate. At the start of codegenning a module we could add declarations for all the loweredfuncs in it and then add the corresponding names to the symbol table.

src/LowerParallelTasks.cpp
dsharletg: Resolved?

steven-johnson:
> What do you think about having a list of struct types inside the Module object instead of in the Stmt IR? 

My first thought is that it would add new management challenges to the IR; currently our IR does a good job of being standalone at any point in the tree, but this would introduce a class of IR nodes that could only be correctly used inside a properly-configured Module.

abadams:
It's not without precedent: A Buffer object, which is referred to by name, can only be used inside a correctly configured Module. Modules are like C translation units. 

This is a struct type that has to be shared between multiple LoweredFunc objects. It exists in the same container that holds those LoweredFuncs, not within the LoweredFuncs themselves. By way of analogy, in C if you have a struct type shared by multiple functions, you put that struct type declaration outside of the functions.

zvookin:
I have no doubt there is a better design to add aggregate types to the IR, but I have trouble seeing how adding structs at the module level is any better than this current approach. E.g. at least this design prints out all the information in debug output. I also don't see a whole lot of cost in changing this later when we have a better design, thus it seems a good first step on the way to something else.

We kind of need this support to deliver certain things and it's been in code review since mid-August so if there are concrete objections to how it works. Moving the struct types out a level could easily be done in a second pass, though if we're going to do that, I'd suggest actually adding scoping as well.

abadams:
My concrete objections to the design are the declaration of types inside imperative Stmt IR, which doesn't mentally type check for me, and the use of StringImms to refer to things in some sort of notional scope, which has caused us considerable pain in the past because passes that track dependencies between bits of IR have to special-case the stringimms.

I want to add the minimum number of new intrinsics. I think we can do this with 2/3, not six.

abadams:
If you don't want to put things in Module, then I woke up with some ideas that would meet my design constraints:

1) 

Make a StructType object, which is either a Type or a list of StructType objects, held behind an Intrusive Ptr. There's also an optional name field, which is empty for anonymous types.

Make new IR nodes MakeTypedStruct and LoadStructMember, both of which have StructType fields that give the type. MakeTypedStruct also has a vector of Expr. It is of type Handle(). LoadStructMember also takes a single int giving the index. It is of the type of that struct field.

2) 

Make a StructType object as above, and add it as an optionally-defined field to Type. There can exist Handle types with the StructType field defined.

Have the existing Intrinsic make_struct have the StructType field set on its Type. The args are just the members.
Add a new Intrinsic load_struct_member which takes something of type Handle, which must have the StructType field set on its Type, and an int that gives the index of the struct member.

---

With both of these options, you would discover the StructTypes while traversing the IR. In the backends you would add a declaration of them to the preamble as you discover them, if you haven't hit that StructType yet. We'd change the existing uses of the make_struct intrinsic accordingly (as a follow-up PR).

Both options avoid the StringImms and avoid an imperative piece of IR used to declare a type.

zvookin:
My position is that this PR is a meaningful improvement in that it moves a significant component of codegen from the LLVM level to the IR level and allows that to be reused across backends. The intrinsics to me are scaffolding and I am in agreement that they are suboptimal, but they are also fine for now. The question is whether the best path forward is to merge the PR and then work on moving the intrinsics to e.g. new IR nodes or if we hold it up for another couple months to do that. The bar for introducing new IR nodes is much higher as it introduces code changes in a lot more places. Removing intrinsics is pretty easy in comparison to adding/modifying/removing an IR node.

E.g. the get_symbol_pointer_or_null thing was to mirror code that was in the LLVM backend and I didn't want to change the behavior because I needed to get the whole thing working and passing tests. (The behavior was also used in a couple other places originally, one for user_context and the other for task_parent. Both of those had other solutions, but the LLVM code really did do a check on the scoped values to implement them.) It turns out the main reason this intrinsic was there is probably dead code anyway and hey, maybe that could have been figured out earlier, but I have enough trouble with not doing things because I'm trying for perfection rather than getting it done. Nuking get_symbol_pointer_or_null is great, but it would have been fine to do that after landing the PR too. (And probably more conservative in that if the .buffer maybe/maybe not code is there for a reason, it would give a place to drop back to without backing out this entire PR.)

I am happy to open issues and add TODOs and even to commit to doing the work to get rid of the intrinsics in some timeframe. But really I do not think they are enough of a problem to create a maintenance burden. Even the LLVM opaque pointer change implies mostly just removing code. (At the cost of losing type checking at the LLVM IR layer.) At worst it requires adding a bit of type info to the scope store.

abadams:
I ran out of time today, but I'm working up one of my StructType proposals. We can take a look at it and see if it's better or worse.

src/Lower.cpp
abadams: Was this resolved?

PR #6198, opened 2021-08-12
[hannk] Remove alignment requirements for shallow DepthwiseConv ops

steven-johnson:


apps/hannk/halide/depthwise_conv_generator.cpp
```diff
@@ -152,8 +152,11 @@ class DepthwiseConv : public Generator<DepthwiseConv> {
             // When we're broadcasting input channels, require that the input has only
             // one channel.
             input_.dim(0).set_extent(1);
-        } else if (inv_depth_multiplier_ == 1) {
+        } else if (inv_depth_multiplier_ == 1 && !shallow_) {
             // Require the input to be aligned.
+            // (Exception: don't require alignment for shallow; that would require
+            // padding on architectures with wide vectors, e.g. AVX512, HVX, which we'd
+            // rather not do. It's better to just pay an alignment penalty.)
```
dsharletg, line 159: I think we do want to do this, we just can't express it currently, because we need to pad the fusing of two dimensions, rather than just padding one dimension.

Super nit: I like to express things like this:

    if (inv_depth_multiplier_ == 0) {
        ... // unchanged code
    } else if (shallow_) {
        // comment...
    } else if (inv_depth_multiplier_ == 1) {
        ... // unchanged code
    }

That way it's more clear what case the comment applies to without the // Exception: ...

steven-johnson: Done

dsharletg (approved):

aankit-ca:
This PR solves the constraint violation errors for shallow depthwise conv for Hexagon. Thank you!

PR #6200, opened 2021-08-13
Fix unroll failures from adams2019 when the Expr depends on estimates

rootjalex:
Adams2019 often tries to unroll loops that are not constant-bounded due to substitution of estimates for input variables. This PR tracks which bounds depend on estimates (`BoundsInfo` now tracks `depend_on_estimate`), and removes the constant_bound tag if the calculation of a constant bound depends on an estimate.

### DIFF
--- a/src/autoschedulers/adams2019/FunctionDAG.cpp
+++ b/src/autoschedulers/adams2019/FunctionDAG.cpp
@@ -691,7 +691,6 @@ FunctionDAG::FunctionDAG(const vector<Function> &outputs, const MachineParams &p
                 for (int j = 0; j < consumer.dimensions(); j++) {
                     const auto &req = node.region_required[j];
                     auto &comp = node.region_computed[j];
-                    // TODO: is this the right thing to do?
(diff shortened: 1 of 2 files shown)


src/autoschedulers/adams2019/FunctionDAG.cpp
```diff
@@ -672,6 +691,8 @@ FunctionDAG::FunctionDAG(const vector<Function> &outputs, const MachineParams &p
                 for (int j = 0; j < consumer.dimensions(); j++) {
                     const auto &req = node.region_required[j];
                     auto &comp = node.region_computed[j];
+                    // TODO: is this the right thing to do?
```
abadams, line 694: Yes it is. Get rid of this TODO

abadams (approved):

src/autoschedulers/adams2019/FunctionDAG.cpp
rootjalex: Whoops, removed

PR #6201, opened 2021-08-14
fix _halide_buffer_crop bug

vangohao:
src/runtime/halide_buffer_t.cpp
has a bug in function "_halide_buffer_crop".
It does not consider the situation when src->type.lanes != 1

### DIFF
--- a/src/runtime/halide_buffer_t.cpp
+++ b/src/runtime/halide_buffer_t.cpp
@@ -151,7 +151,7 @@ halide_buffer_t *_halide_buffer_crop(void *user_context,
         offset += (min[i] - src->dim[i].min) * (int64_t)src->dim[i].stride;
     }
     if (dst->host) {
-        dst->host += offset * src->type.bytes();
+        dst->host += offset * src->type.bytes() * src->type.lanes;
     }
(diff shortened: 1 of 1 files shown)


abadams:
How did you get into a situation with a vector type being passed to halide_buffer_crop? I'm not sure how that could happen. Buffer types are supposed to be always scalar.

vangohao:
I am using define_extern Func to compute complicated stencils in the numerical solution of PDE, in which I construct the halide_buffer_t by myself. I only use Halide to schedule the Loops. (i.e. loop tiling & parallel)
halide_type_t support vector type by setting the lanes, which is perfect to compared to the DOF (degree of freedom) in PDE, so I tried to use it and find this problem.

abadams:
Instead of using vector lanes, add an extra innermost dimension to your buffer. None of the buffer methods are designed to work with vector types.

vangohao:
Thanks for your advise. If none of the buffer methods are designed to work with vector types, then I think an ``assert(src->type.lanes == 1`` or something like that should be inserted, else others who tries to use it like me may get into trouble find wrong results.

PR #6203, opened 2021-08-16
Add IRMutator::mutate_exprs()

steven-johnson:
There's a common pattern in many IRMutators that is "mutate a vector<Expr> and optionally let me know if anything is different".

(Note, this uses C++17 structured-binding syntax, which we previously weren't using in Halide. Objections?)

This adds a shared utility method (well, two, thanks to VariadicVisitor) and plus in the usage in all the places that seemed obvious.

I doubt this moves the needle on speed in either direction, but makes for smaller code.

### DIFF
--- a/src/Simplify.cpp
+++ b/src/Simplify.cpp
@@ -54,7 +54,7 @@ Simplify::Simplify(bool r, const Scope<Interval> *bi, const Scope<ModulusRemaind
     }
 }
 
-std::pair<std::vector<Expr>, bool> Simplify::mutate_exprs(const std::vector<Expr> &old_exprs, ExprInfo *bounds) {
+std::pair<std::vector<Expr>, bool> Simplify::mutate_with_changes(const std::vector<Expr> &old_exprs, ExprInfo *bounds) {
     vector<Expr> new_exprs(old_exprs.size());
     bool changed = false;
 
--- a/src/Simplify.cpp
+++ b/src/Simplify.cpp
@@ -54,6 +54,23 @@ Simplify::Simplify(bool r, const Scope<Interval> *bi, const Scope<ModulusRemaind
     }
 }
 
+std::pair<std::vector<Expr>, bool> Simplify::mutate_exprs(const std::vector<Expr> &old_exprs, ExprInfo *bounds) {
+    vector<Expr> new_exprs(old_exprs.size());
+    bool changed = false;
+
+    // Mutate the args
+    for (size_t i = 0; i < old_exprs.size(); i++) {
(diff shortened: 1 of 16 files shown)


steven-johnson:
Alas, it turns out that the C++ committee deliberately left out an equivalent of `std::ignore` for structured bindings, so the calls that ignore the `changed` return value will emit "unused variable" warnings for some compilers. I've added hamfisted ways to silence these but, yuck.

src/Simplify.cpp
```diff
@@ -54,6 +54,23 @@ Simplify::Simplify(bool r, const Scope<Interval> *bi, const Scope<ModulusRemaind
...
+    for (size_t i = 0; i < old_exprs.size(); i++) {
+        const Expr &old_e = old_exprs[i];
+        Expr new_e = mutate(old_e, bounds);
+        if (!new_e.same_as(old_e)) {
+            changed = true;
+        }
+        new_exprs[i] = std::move(new_e);
+    }
+
+    return {std::move(new_exprs), changed};
```
abadams, line 71: I wonder if it's any faster to exploit NRVO and actually declare the std::pair on the stack and return it as an lvalue. Then the vector should just live in the caller frame instead of needing to be moved. It could well optimize to the same thing though.

abadams:
Oh, and you closed this... I'll look at the new version.

src/Simplify.cpp
steven-johnson: It could work but the syntax makes my teeth hurt.

steven-johnson:
Reopening since there is debate over which approach is better.

abadams (approved):

abadams:
Another possible design would be an optional bool * second arg which gets set to true if there's a change. Not sure it would be cleaner than this though.

steven-johnson:
> Another possible design would be an optional bool * second arg which gets set to true if there's a change. Not sure it would be cleaner than this though.

I have a probably-irrational dislike of that pattern, TBH. Multiple-return-values is vastly preferable, IMHO, and it irks me that C++ seems to keep inching closer to enabling it, yet still failing in ways like this.

steven-johnson:
Given the number of "ignore-the-unused-return-value" that are apparently needed, how about this as a tweak before landing:
- Add `vector<Expr> mutate()`, which just returns the mutated list alone, no change flag (and migrate usage appropriately)
- Rename `mutate_exprs()` to something like `mutate_with_changes()`

abadams:
sgtm. Don't love the name mutate_with_changes, but I can't think of anything better after a few minutes brainstorming.

PR #6204, opened 2021-08-17
Refactor mutate-vector-of-Expr code

steven-johnson:
There's a common pattern in many IRMutators that is "mutate a vector<Expr> and optionally let me know if anything is different".

This just adds `mutate()` and `same_as()` for `vector<Expr>` (and refactors some obvious places to use them).

I doubt this moves the needle in terms of performance, but IMHO it's cleaner code.

(Replaces https://github.com/halide/Halide/pull/6203)

### DIFF
--- a/src/CodeGen_D3D12Compute_Dev.cpp
+++ b/src/CodeGen_D3D12Compute_Dev.cpp
@@ -1047,10 +1047,7 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::add_kernel(Stmt s,
                 std::string new_name = unique_name(op->name);
                 replacements[op->name] = new_name;
 
-                std::vector<Expr> new_extents;
-                for (size_t i = 0; i < op->extents.size(); i++) {
-                    new_extents.push_back(mutate(op->extents[i]));
-                }
+                auto new_extents = mutate(op->extents);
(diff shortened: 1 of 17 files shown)


abadams:
I actually like the other version better. This version is slower because it does a second pass over the list.

I also think that ignoring whether or not the mutation did anything is a code smell in mutators, so I didn't mind the (void)changed. If it didn't do anything, you should probably be returning the existing IR most of the time, so it's worth calling out the places where we don't.

steven-johnson:
> I actually like the other version better. This version is slower because it does a second pass over the list.

Fair, though I made this change on the suspicton that the extra pass is probably in the noise for most uses.
 
> I also think that ignoring whether or not the mutation did anything is a code smell in mutators, so I didn't mind the (void)changed. 

I don't disagree, except that
- None of the existing `mutate()` methods return a did-anything-change flag, so arguably this PR is more similar to the legacy codebase.
- Even with `-Wunused-variable` enabled, neither of my local compilers actually warn about the unused var (the warnings are turning up on g++7 on the linux buildbots).
-

steven-johnson:
Hmm, interestingly, this produces an error during a CSE phase. Clearly I've introduced a subtle error. (Sounds like you dislike this approach enough that it maybe isn't worth debugging...)

PR #6212, opened 2021-08-19
Fix for upcoming LLVM API change

steven-johnson:
(Note that this will break until the buildbots pick up the nightly LLVM change; getting it out there to be ready to land in the morning.)


abadams (approved):

steven-johnson:
Failure is unrelated

PR #6213, opened 2021-08-19
Use C++17 structured binding instead of std::tie

steven-johnson:
Just a minor style thing, but IMHO using C++17 structured bindings is preferable in ~all cases to using std::tie(), so I'm proposing this change.

Also made a drive-by change to populate_environment (now build_environment) since literally every caller was calling it this way.


abadams (approved):

steven-johnson:
Failures are unrelated

PR #6215, opened 2021-08-20
Add docker support for NVIDIA x86_64 envrionments

vladimirmujagic:
Might be useful to have images for various environments to help with learning, testing and benchmarking. Sharing what I have for now, still work in progress but should be usable.


steven-johnson:

PR #6222, opened 2021-08-20
Add modernize-make-shared and modernize-make-unique to .clang-tidy and fix warnings

steven-johnson:


abadams (approved):

zvookin:
Why is using ```make_unique``` an improvement over the ```reset``` pattern?

abadams:
I think just because it's easier to say "don't use new and delete" than "only use new in the argument to reset on a smart ptr"

zvookin:
My general take is that we shouldn't put certain things into the formatting support because there are cases where it is pretty dubious and needs human decisions. If I have to write a worse piece of code to placate a check that isn't measurably helping something, we're not winning. Running them semi-regularly under human control and looking at them as suggestions is fine.

As time goes on, I expect both the pattern recognition and the best practices available in C++ to use for rewriting will both improve, but looking at the four PRs here, the tie thing is good, the for loop thing is "sure, ranged for loops are the new hotness so...", this one is ok but one line of code definitely got uglier, and the reserve thing looks like a productivity loss to me.

steven-johnson:
IIUC, the overall justification for make_unique/make_shared in the language is that they are deemed exception-safe, and thus considered best practices. In Halide, generally any exception being thrown means a fatal error, so worrying about leaks is a nonfactor. That said, I think this is the right way to go, since even Google (aka "exceptions are bad and wrong") is recommending their use in style guides, and IMHO there isn't any stylistic downside to them.

That said, as always, a PR is just a request; if the consensus is that mandating make_unique/make_shared is not in Halide's best interest, I'm fine with that.

steven-johnson:
(PR is clean (osx build is a flake), but obviously not gonna land this unless there's a pretty clear consensus)

src/Pipeline.h
```diff
@@ -121,8 +122,8 @@ class Pipeline {
         }
         template<typename T, typename... Args,
                  typename = typename std::enable_if<Internal::all_are_convertible<Buffer<>, Args...>::value>::type>
-        RealizationArg(Buffer<T> &a, Args &&...args) {
-            buffer_list.reset(new std::vector<Buffer<>>({a, args...}));
+        RealizationArg(Buffer<T> &a, Args &&...args)
+            : buffer_list(std::make_unique<std::vector<Buffer<>>>(std::vector<Buffer<>>{a, std::forward<Args>(args)...})) {
```
abadams, line 126: I thought make_unique forwarded args to a constructor, so can't ```std::vector<Buffer<>>{a, std::forward<Args>(args)...}``` just be ```a, std::forward<Args>(args)...```?

steven-johnson: You'd think so, but that ends up generating oodles of `no matching constructor for initialization of 'std::__1::vector<Halide::Buffer<void>>'` errors. I'll take another look to see if I can figure out why that isn't working.

steven-johnson: Actually, no: std::vector doesn't have a ctor that accepts a variadic list of items, which is what that would match to. (It does have a ctor that accepts an initializer list, and of course a copy-constructor, which is what the code above is using.)

steven-johnson: FYI, the reason that just a plain {wrap-in-braces} here isn't good enough is that a caller passes multiple Buffers of different types (e.g. multiple_scatter.cpp:197) can make type inference problematic here; adding an explicit expectation of the resulting uniform type is necessary. 

That said, using std::initializer_list is likely a better/cheaper way that std::vector here, so I'll do that.

abadams:
lgtm

PR #6223, opened 2021-08-21
Add performance-inefficient-vector-operation to .clang-tidy and fix warnings

steven-johnson:


zvookin:
This pattern is tedious. Is there any evidence this is worth it performance wise? As an optimization for places where it matters, sure. As a general thing to do everywhere, yuck. I was going to suggest constructing the vectors and filling them in by index for most of these, but I guess the objection there is that default constructs all the elements when constructing the vector.

If this is going to be C++ standard practice, then there should be a vector constructor that does the size reservation.

steven-johnson:
> This pattern is tedious. Is there any evidence this is worth it performance wise? As an optimization for places where it matters, sure. As a general thing to do everywhere, yuck. 

No, I didn't do any performance measurement. I suspect any performance increases from these are marginal at best.

But I doubt this is going to be tedious: the tidy check apparently only looks for places that it is 100% confident that the reserve() call can be made completely accurately (ie, the ctor is followed by a loop that pushes exactly N items on the vector).   Adding the reserve() call has no downside that I see. Yeah, I guess there will be the annoyance of "failed clang-tidy, let me fix this" but that strikes me as pretty minor.

> If this is going to be C++ standard practice, then there should be a vector constructor that does the size reservation.

I agree, but no such ctor exists, and (AFAIK) isn't planned for future C++ standards either. 

As always, I'm fine with whatever the Halide consensus thinks here.

dsharletg:
> Adding the reserve() call has no downside that I see. Yeah, I guess there
will be the annoyance of "failed clang-tidy, let me fix this" but that
strikes me as pretty minor.

This is not really that minor to me at least. The clang-tidy script does
bizarre things when I run it locally, so I only fix things after pushing
them and seeing the build bots complain. Maybe the real fix is for me to
troubleshoot why running the clang-tidy script locally rewrites almost
every line of the entire repo.

I think the other recent clang-tidy enablement PRs are OK, but this one I
think we should not enable. I think over half of the instances added in
this PR are reserving dimension lists, which are probably never bigger than
10 (I think the highest dimensional func I know of is 7 or 8 dimensions).
Certainly not big enough for the reserve call to matter. We periodically
profile code from time to time and would fix an issue like this if it were
a problem. And I think basically everyone that reviews code on this project
would mention to add a reserve call if it made sense.

zvookin:
By adding this to the main rule set, we are effectively saying every construction of a ```std::vector```, where the vector is populated in the same function, should be followed by a ```reserve``` call. Otherwise one risks a clang-format failure, which will prevent moving a PR forward. This is for performance, not correctness.

I am saying that the extra line of code is not worth it in most cases on the basis that the minimal or nonexistent performance gain does not justify an extra line of code to read. The cost in writing code is on top of that. To be sure, I may be wrong and if others feel that is the case I'm not going to push on this, but I do wants us to make these sorts of decisions on the basis of increased productivity and code quality, not unexamined dogma.

I would propose having the formatting script take an argument to add extra rules that we should run every now and then to see if there's anything worthwhile. For this one, it will be a small number of warnings and if any are new and look large, we can fix them. (E.g. the 200 pushes in the test should probably get a reserve added.)

Per Dillon's question about reformatting I think the main issue that causes clang-format to make more modifications is the output depends heavily on the version of clang-format one is using. I gave up on getting exactly the right one and just let the buildbots do it.

steven-johnson:
Dropping this one.

Issue #6219, opened 2021-08-20
prefetch ignores offset in some circumstances


abadams:
Consider the following attempt to prefetch:

```

#include "Halide.h"

using namespace Halide;

int main(int argc, char **argv) {
    Func f;
    Var x, y;
    f(x, y) = x + y;
    f.compute_root();

    Func g;

    g(x, y) = f(x, y);

    Var xo, yo, xi, yi;
    g.tile(x, y, xo, yo, xi, yi, 8, 4, TailStrategy::RoundUp)
        .vectorize(xi)
        .unroll(yi);

    f.in().compute_at(g, xo).vectorize(x).unroll(y).prefetch(f, y, x, 123, PrefetchBoundStrategy::NonFaulting);

    g.realize({1024, 1024});

    return 0;
}
```

The prefetch doesn't actually look ahead to x + 123. It just prefetches the address it loads on the next instruction. Changing f.in() to vectorize(x, 8) doesn't seem to change anything, even though x is now the loop over vectors instead of the vectorized loop. Either way I get the IR:

```  
      prefetch(f, ((g.s0.x.xo*2) + t13)*4, 1, 16)
      f_global_wrapper$0[ramp(0, 1, 8)] = f[ramp(((g.s0.x.xo*2) + t13)*4, 1, 8) aligned(8, 0)]
      prefetch(f, (g.s0.x.xo*8) + t14, 1, 16)
      f_global_wrapper$0[ramp(8, 1, 8)] = f[ramp((g.s0.x.xo*8) + t14, 1, 8) aligned(8, 0)]
      prefetch(f, (g.s0.x.xo*8) + t15, 1, 16)
      f_global_wrapper$0[ramp(16, 1, 8)] = f[ramp((g.s0.x.xo*8) + t15, 1, 8) aligned(8, 0)]
      prefetch(f, (g.s0.x.xo*8) + t16, 1, 16)
      f_global_wrapper$0[ramp(24, 1, 8)] = f[ramp((g.s0.x.xo*8) + t16, 1, 8) aligned(8, 0)]
```

steven-johnson:
Any idea if it worked prior to the recent restructuring of the prefetch directive? (ie, is it a new breakage or an always-broken thing?)

abadams:
I am not sure. This was new code. I haven't used prefetch a lot, so I would not be surprised if it were already broken.

abadams:
[...]

PR #6225, opened 2021-08-23
Fix bug in prefetch()

steven-johnson:
In #6155, we incorrectly assume that we can qualify the 'from' prefetch var by just adding 'prefix'; this isn't true if (e.g.) there are any splits involved. Instead, we need to walk through the active loops to find a suitable match. In addition, if no match is found, we now fail with an error (rather than quietly doing something undefined), as the 'from' var is required to be from an active loop.

(Addresses some-but-not-all of #6219)


abadams (approved):

steven-johnson:
Failures are unrelated

PR #6226, opened 2021-08-23
More prefetch fixes 

steven-johnson:
- Arguments to Call::prefetch() must be scalars, not vectors
- Add more testcases to correctness_prefetch

### DIFF
--- a/test/correctness/prefetch.cpp
+++ b/test/correctness/prefetch.cpp
@@ -10,6 +10,11 @@ using std::vector;
 using namespace Halide;
 using namespace Halide::Internal;
 
+template<typename T>
+Expr wild() {
+    return Variable::make(halide_type_of<T>(), "*");
+}
+
 class CollectPrefetches : public IRVisitor {
 private:
     using IRVisitor::visit;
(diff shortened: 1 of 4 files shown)


steven-johnson:
OK, I think this now actually addresses all of the issues in #6219 and fixes it, PTAL

steven-johnson:
PTAL

test/correctness/prefetch.cpp
```diff
@@ -253,13 +260,153 @@ int test8(const Target &t) {
...
+        .unroll(yi);
+
+    f.in()
+        .compute_at(g, xo)
+        .split(x, xo, xi, 4)
+        .vectorize(xi)
+        .unroll(xo)
+        .reorder(xi, y, xo)
+        .unroll(y)
+        .prefetch(f, y, xo, 123 / 4, PrefetchBoundStrategy::NonFaulting);
```
dsharletg, line 325: This is just a regular integer division performed at compile time, why leave it this way?

dsharletg (approved):

test/correctness/prefetch.cpp
abadams: Because it's supposed to be equivalent to the case above that prefetches 123 elements ahead in the x direction. Because this is the xo loop, the correct amount is 123/4

Maybe a comment to that effect...

steven-johnson: Done

PR #6233, opened 2021-09-01
Upgrade clang-format and clang-tidy to LLVM-12

steven-johnson:
Pretty minor -- no formatting changes, only one new warning needing disabling -- just keeping our tooling not too far from top of tree.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -752,7 +752,7 @@ std::string halide_type_to_c_type(const Type &t) {
 
 int generate_filter_main_inner(int argc, char **argv, std::ostream &error_output) {
     const char kUsage[] =
-        "gengen\n"
+        "gengen \n"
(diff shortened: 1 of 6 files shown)


src/Generator.cpp
```diff
@@ -752,7 +752,7 @@ std::string halide_type_to_c_type(const Type &t) {
 
 int generate_filter_main_inner(int argc, char **argv, std::ostream &error_output) {
     const char kUsage[] =
-        "gengen\n"
+        "gengen \n"
```
alexreinking, line 755: If you just need something to trigger the buildbots, you can write `git commit --allow-empty -m "trigger buildbots"`... should cut down on churn on this file.

alexreinking (approved):

src/Generator.cpp
steven-johnson: No, this is because these tests won't run unless at least one cpp/h fie is touched

alexreinking: gah

PR #6234, opened 2021-09-04
Remove torch/extension dependency and fix PyTorch example, adding the correct (GPU) autoscheduler

mgharbi:
1. The runtime helpers for the PyTorch extension include "torch/extension.h". This is unnecessary and creates an unnecessary dependency on "Python.h" for libraries that include PyTorch Halide extensions (e.g. a PyTorch C++ TorchScript extension does not need the Python frontend). This pull request removes this dependency.
2. The HelloPyTorch app was broken: the default autoscheduler [2016] does not work on GPU, for the automatically generated gradient. This pull request fixes it.

### DIFF
--- a/src/CodeGen_PyTorch.cpp
+++ b/src/CodeGen_PyTorch.cpp
@@ -25,13 +25,9 @@ void CodeGen_PyTorch::compile(const Module &module) {
                           "Please add \"-user_context\" to the generator's target options.\n";
         }
         stream << "#include \"ATen/cuda/CUDAContext.h\"\n";
-        stream << "#include \"HalideBuffer.h\"\n";
-        stream << "#include \"HalidePyTorchCudaHelpers.h\"\n";
-        stream << "#include \"HalidePyTorchHelpers.h\"\n";
-    } else {
-        stream << "#include \"HalideBuffer.h\"\n";
-        stream << "#include \"HalidePyTorchHelpers.h\"\n";
     }
+    stream << "#include \"HalideBuffer.h\"\n";
+    stream << "#include \"HalidePyTorchHelpers.h\"\n";
 
     stream << "\n";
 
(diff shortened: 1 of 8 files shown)


mgharbi:
@steven-johnson  #ab8674b fixes an issue with the gradient output for generators, which crashes when a generator has a scalar input because of the `internal_assert`. An easy way to reproduce is to add a line [here](https://github.com/mgharbi/Halide/blob/17f21e40bb4e345485d1bedbda7dad0b6225c012/apps/HelloPyTorch/src/add_generator.cpp#L19):

`Input<float> temp_scalar{"temp_scalar"};`

then try and built the app. Scalar parameters have no `funcs`, so the assert is violated.

src/CodeGen_PyTorch.cpp
```diff
@@ -102,7 +96,11 @@ void CodeGen_PyTorch::compile(const LoweredFunc &f, bool is_cuda) {
         stream << get_indent() << "CUresult res = cuCtxGetCurrent(&ctx);\n";
         stream << get_indent() << "AT_ASSERTM(res == 0, \"Could not acquire CUDA context\");\n";
         stream << get_indent() << "cudaStream_t stream = at::cuda::getCurrentCUDAStream(device_id);\n";
-        stream << get_indent() << "Halide::PyTorch::UserContext user_ctx(device_id, &ctx, &stream);\n";
+        stream << get_indent() << "struct UserContext {int device_id; CUcontext *cuda_context; cudaStream_t *stream; } user_ctx;\n";
+        stream << get_indent() << "user_ctx.device_id = device_id;\n";
+        stream << get_indent() << "user_ctx.cuda_context = &ctx;\n";
+        stream << get_indent() << "user_ctx.stream = &stream;\n";
+        // stream << get_indent() << "Halide::PyTorch::UserContext user_ctx(device_id, &ctx, &stream);\n";
```
steven-johnson, line 103: Don't check in commented-out code.

src/CodeGen_PyTorch.cpp
```diff
@@ -102,7 +96,11 @@ void CodeGen_PyTorch::compile(const LoweredFunc &f, bool is_cuda) {
         stream << get_indent() << "CUresult res = cuCtxGetCurrent(&ctx);\n";
         stream << get_indent() << "AT_ASSERTM(res == 0, \"Could not acquire CUDA context\");\n";
         stream << get_indent() << "cudaStream_t stream = at::cuda::getCurrentCUDAStream(device_id);\n";
-        stream << get_indent() << "Halide::PyTorch::UserContext user_ctx(device_id, &ctx, &stream);\n";
+        stream << get_indent() << "struct UserContext {int device_id; CUcontext *cuda_context; cudaStream_t *stream; } user_ctx;\n";
```
steven-johnson, line 99: nit: space after opening `{`

steven-johnson (approved):
LGTM, and passes on my local Linux box. Let's wait till the buildbots are back up before landing it, though.

steven-johnson:
Thanks for the fixes!

PR #6237, opened 2021-09-09
Fall through to LLVM for unknown shuffles on Hexagon

dsharletg:
I think LLVM handles some unknown shuffles better than we do with vlut. We could also fall through to vdelta instead, but I suspect LLVM does this anyways, and might do something smarter in some special cases.

@aankit-ca @pranavb-ca does this seem like a good thing to do?

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -1268,9 +1268,8 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
         return concat_vectors(ret);
     }
 
-    // Maybe LLVM can do something with this.
-    // Hopefully it uses vdelta in the worst case.
-    return CodeGen_Posix::shuffle_vectors(a, b, indices);
+    // Use a general delta operation.
+    return vdelta(concat_vectors({a, b}), indices);
 }
 
 Value *CodeGen_Hexagon::vlut256(Value *lut, Value *idx, int min_index,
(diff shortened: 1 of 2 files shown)


src/CodeGen_Hexagon.cpp
```diff
@@ -1269,10 +1268,9 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
         return concat_vectors(ret);
     }
 
-    // TODO: There are more HVX permute instructions that could be
-    // implemented here, such as vdelta/vrdelta.
-
-    return vlut(concat_vectors({a, b}), indices);
+    // Maybe LLVM can do something with this.
+    // Hopefully it uses vdelta in the worst case.
+    return CodeGen_Posix::shuffle_vectors(a, b, indices);
```
aankit-ca, line 1273: Should we just replace this with:

vdelta(concat_vectors({a, b}), indices);

aankit-ca:
Hexagon does have good support for using the permute networks. Maybe we should change the fallback for CodeGen::vdelta also to CodeGen_Posix::shuffle_vectors(a, b, indices)  instead of vlut? Do you have some usecases that we could try this on?

dsharletg:
It looks like we need to. Letting this fall through to LLVM generates scalarized code in simd_op_check currently.

aankit-ca:
LGTM

abadams (approved):
Can't vouch for correctness, but @aankit-ca has to I'm rubber-stamping so you can merge.

dsharletg:
The main voucher of correctness is several simd_op_check tests that cover this :)

Speaking of which, I saw that the affected tests passed, but now the buildbot server is down :( Since all affected tests passed, and we need this change merged, I'm going to merge it.

PR #6239, opened 2021-09-10
compiler stack usage improvements

### LINKED ISSUE #2411: Generator executable can be produced but not the corresponding *.a file
Hi, 

I'm working on block matching of x-gradient and y-gradient of two images. The complete codes are pasted at the bottom of the post.
When using only one single gradient, the program compiles and runs correctly.
```
dft(x,y,bx,by) = dft_ref_dx(x,y,bx,by)*dft_src_dx(x,y,bx,by);
```
However when both gradients are used,

```
dft(x,y,bx,by) = dft_ref_dx(x,y,bx,by)*dft_src_dx(x,y,bx,by)+dft_ref_dy(x,y,bx,by)*dft_src_dy(x,y,bx,by)
```
 errors occur when generating the library.
```
g++ -std=c++11 -I ../../include/ -I ../../tools/ -I ../../apps/support/ -fno-rtti block_matching_fft_generator.cpp fft.cpp ../../lib/libHalide.a ../../tools/GenGen.cpp -o bin/block_matching_fft_generator  -lole32 -luuid -Wl,--stack,8388608
bin/block_matching_fft_generator -g block_matching_fft -o ./bin -f block_matching_fft  target=host
make: *** [Makefile:55: bin/block_matching_fft.a] Error -1073741571
```
Why is the executable generated but not the library? 
I couldn't figured out the cause myself, do you have any suggestions?

```
#include "Halide.h"
#include "fft.h"
#include <cstdio>
#include <vector>
#include <cstdlib>
#include <cassert>
using namespace Halide;
class BlockMatchingFFT : public Halide::Generator<BlockMatchingFFT> {
public:
	GeneratorParam<float> gain{"gain", 0.0009765625};
	Input<Buffer<float>> ref{ "ref", 2 };
	Input<Buffer<float>> src{ "src", 2 };
 	Input<int> ts{"ts"};
	Input<int> search_range{"search_range"};
	Output<Func> MxMy{ "MxMy", {Int(32),Int(32)},2 };
 	void generate() {
		Var bx("bx"), by("by"), x("x"), y("y"),bxy("bxy");
		Fft2dDesc fwd_desc;
		Fft2dDesc inv_desc;
		inv_desc.gain = gain;
		
		Func clamped_ref("clamped_ref"), clamped_src("clamped_src");
		Func 
		ref_dx("Rdx"),ref_dy("Rdy"),sub_ref_dx("SRdx"),sub_ref_dy("SRdy"),psub_ref_dx("PSRdx"),psub_ref_dy("PSRdy"),
		src_dx("Sdx"),src_dy("Sdy"),sub_src_dx("SSdx"),sub_src_dy("SSdy"),psub_src_dx("PSSdx"),psub_src_dy("PSSdy");
			 
		clamped_ref = BoundaryConditions::repeat_edge(ref);
		clamped_src = BoundaryConditions::repeat_edge(src);
   		ref_dx(x, y) = clamped_ref(x-1, y) - clamped_ref(x+1, y);
		src_dx(x, y) = clamped_src(x-1, y) - clamped_src(x+1, y);
		ref_dy(x, y) = clamped_ref(x, y-1) - clamped_ref(x, y+1);
		src_dy(x, y) = clamped_src(x, y-1) - clamped_src(x, y+1);
	 
		sub_ref_dx(x,y,bx,by) = ref_dx(bx*ts+ts-1-x,by*ts+ts-1-y);
		sub_src_dx(x,y,bx,by) = src_dx(bx*ts+x-search_range,by*ts+y-search_range);
		psub_ref_dx = BoundaryConditions::constant_exterior(sub_ref_dx, 0,{{0, 16}, {0, 16}});	
		psub_src_dx = BoundaryConditions::constant_exterior(sub_src_dx, 0,{{0, 25}, {0, 25}});	
		
		sub_ref_dy(x,y,bx,by) = ref_dy(bx*ts+ts-1-x,by*ts+ts-1-y);
		sub_src_dy(x,y,bx,by) = src_dy(bx*ts+x-search_range,by*ts+y-search_range);
		psub_ref_dy = BoundaryConditions::constant_exterior(sub_ref_dy, 0,{{0, 16}, {0, 16}});	
		psub_src_dy = BoundaryConditions::constant_exterior(sub_src_dy, 0,{{0, 25}, {0, 25}});	
	

		ComplexFunc dft_ref_dx("dft_ref_dx"),dft_ref_dy("dft_ref_dy"),dft_src_dx("dft_src_dx"),dft_src_dy("dft_src_dy"),dft("dft");
		dft_ref_dx = fft2d_r2c(psub_ref_dx, 32, 32, target, fwd_desc);
		dft_src_dx = fft2d_r2c(psub_src_dx, 32, 32, target, fwd_desc);
		dft_ref_dy = fft2d_r2c(psub_ref_dy, 32, 32, target, fwd_desc);
		dft_src_dy = fft2d_r2c(psub_src_dy, 32, 32, target, fwd_desc);
	
		//dft(x,y,bx,by) = dft_ref_dx(x,y,bx,by)*dft_src_dx(x,y,bx,by);
		dft(x,y,bx,by) = dft_ref_dx(x,y,bx,by)*dft_src_dx(x,y,bx,by)+dft_ref_dy(x,y,bx,by)*dft_src_dy(x,y,bx,by);
		Func ifft("ifft");
		ifft = fft2d_c2r(dft, 32,32, target, inv_desc);	
				
		RDom r_out(0, 2*search_range+1,0,2*search_range+1);
		Func out("out");		
		out(bx,by) = argmin(ifft(r_out.x+ts,r_out.y+ts,bx,by));
		MxMy(bx,by)= Tuple(out(bx,by)[0],out(bx,by)[1]);
			
 		out.compute_root().fuse(bx,by,bxy).parallel(bxy);
		ifft.compute_at(out,bxy);
		dft.compute_at(out, bxy);
		dft_ref_dx.compute_at(out,bxy);
		dft_src_dx.compute_at(out,bxy);
		dft_ref_dy.compute_at(out,bxy);
		dft_src_dy.compute_at(out,bxy);
	}
  	void schedule() { 
  	}
};
HALIDE_REGISTER_GENERATOR(BlockMatchingFFT, "block_matching_fft")
```

abadams:
I found some code in the wild that needs 9mb of stack to lower. It's a
pain to even diagnose the problem definitively, because it requires
plumbing platform-specific linker flags though the build system to 
grant more stack to rule out infinite recursion.

This PR:

- Reduces peak stack usage of similar code in the open source repo
- Increases the stack size for lowering and codegen to 32mb on all
platforms, using stack switching techniques. We started doing this on
Windows a while ago and it hasn't bitten us, so let's try on more
platforms.
- Gives user control over the amount of stack used for lowering and
codegen. It shouldn't be necessary except when diagnosing problems like
this in future.

Using the control I was able to determine that the correctness tests all
pass with 500k of stack, and the apps all pass with 2MB, so 32MB ought
to be enough for anybody.

I found a never-checked-in test for the mux helper which uses 10MB of
stack and really shouldn't need to, so I added that (and opened an
issue) as an example of how to grant more stack when necessary, even
though 10MB is less than our default now.

Also fixed an incorrect comment on the Block node.

Fixes #2411


abadams:
For the pathological case found in the wild, this reduces stack usage to 3-4 MB, so this PR is not just about increasing the stack size.

src/AsyncProducers.cpp
```diff
@@ -457,28 +457,51 @@ class TightenProducerConsumerNodes : public IRMutator {
 
     Stmt make_producer_consumer(const string &name, bool is_producer, Stmt body, const Scope<int> &scope) {
         if (const LetStmt *let = body.as<LetStmt>()) {
-            if (expr_uses_vars(let->value, scope)) {
-                return ProducerConsumer::make(name, is_producer, body);
+            Stmt orig = body;  // Only used to keep a reference to the let chain in scope.
```
alexreinking, line 460: but why is that necessary?

src/Util.cpp
```diff
@@ -648,10 +705,68 @@ void run_with_large_stack(const std::function<void()> &action) {
...
+        arg->run();
+    };
+#endif
 
+    ucontext_t context, calling_context;
+
+    // We'll allocate 16 protected guard pages at the end of the stack
+    // we're making to catch stack overflows when they happen, as
+    // opposed to having them cause silent corruption.
+    const size_t guard_pages = 4096 * 16;
```
alexreinking, line 740: This should use
```
#include <unistd.h>
// ...
const size_t guard_pages = 16 *  sysconf(_SC_PAGESIZE);
```
to be totally POSIX-compliant

alexreinking (changes requested):

alexreinking:
> I found some code in the wild that needs 9mb of stack to lower. It's a
> pain to even diagnose the problem definitively, because it requires
> plumbing platform-specific linker flags though the build system to
> grant more stack to rule out infinite recursion.

Can this not be diagnosed with `ulimit` on Linux?

abadams:
> > I found some code in the wild that needs 9mb of stack to lower. It's a
> > pain to even diagnose the problem definitively, because it requires
> > plumbing platform-specific linker flags though the build system to
> > grant more stack to rule out infinite recursion.
> 
> Can this not be diagnosed with `ulimit` on Linux?

It depends. For the person with the problem it couldn't be done without first messing with system settings to give the initial shell more than 8MB of stack to begin with, because ulimit only goes down.

alexreinking:
> It depends. For the person with the problem it couldn't be done without first messing with system settings to give the initial shell more than 8MB of stack to begin with, because ulimit only goes down.

On my system at least I start with an 8MB stack, but I _can_ increase it via `ulimit -s unlimited`. This was necessary for several of the Haskell benchmarks in the Perceus paper since it would routinely use up too much stack space.

abadams:
IIRC that did not work for the reporter and me (possibly because their build system complicates the issue by having an already-launched persistent daemon). Either way, it's good to have a mechanism to let the compiler use more stack that works on every platform.

src/Util.cpp
abadams: meh, it doesn't actually need to be 16 OS pages, so I'd rather not introduce a dependence on a new function and macro. I just picked that size because it's large but not so large that anyone would notice or object.

alexreinking:
> Either way, it's good to have a mechanism to let the compiler use more stack that works on every platform.

Sorry, I didn't mean to argue against the PR with that question. It was just for clarification.

src/Util.cpp
alexreinking: I mean, we're already including `unistd.h` and `sysconf(_SC_PAGESIZE)` is standard POSIX.1 and has been since before I was born. I honestly don't get the "dependency" objection at all. 

You could also change the comment.

alexreinking (approved):

alexreinking:
Hmm, looks like we're seeing a real failure on the new test with LLVM 14 on Windows.

abadams:
Maybe 10mb is too tight for windows. I bumped it to 12 and we'll see if it passes.

src/AsyncProducers.cpp
abadams: Updated comment. It's because 'body' may be the only reference-counted Stmt keeping the first LetStmt alive, but we're mutating body to point to its innards. We don't want that first LetStmt to have its refcount hit zero and make our pointer dangle.

src/SkipStages.cpp
```diff
@@ -78,37 +78,48 @@ class PredicateFinder : public IRVisitor {
         op->body.accept(this);
         if (should_pop) {
             varying.pop(op->name);
-            //internal_assert(!expr_uses_var(predicate, op->name));
+            // internal_assert(!expr_uses_var(predicate, op->name));
```
steven-johnson, line 81: We should either remove this line or add a comment about why we aren't doing this check.

src/UnrollLoops.cpp
```diff
@@ -83,6 +83,11 @@ class UnrollLoops : public IRMutator {
             Stmt iters;
             for (int i = e->value - 1; i >= 0; i--) {
                 Stmt iter = substitute(for_loop->name, for_loop->min + i, body);
+                // It's necessary to simplify eagerly this iteration
```
steven-johnson, line 86: nit: "eagerly simplify"

src/Util.cpp
```diff
@@ -648,10 +705,70 @@ void run_with_large_stack(const std::function<void()> &action) {
...
+    // stack we're making to catch stack overflows when they happen,
+    // as opposed to having them cause silent corruption. We pick an
+    // amount of memory that should be comfortably larger than most
+    // stack frames - 64k.
+    const size_t guard_band = 64 * 1024;
+
+    void *stack = mmap(nullptr, stack_size.size + guard_band, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+    internal_assert(stack);
+
+    mprotect((char *)stack + stack_size.size, guard_band, PROT_NONE);
```
steven-johnson, line 747: Shouldn't we check the return value?

steven-johnson:
Looks like we're getting segfaults on osx-arm64

abadams:
On Arm Macos the dubious function pointer casting that caused the deprecation of these routines was indeed invalid - only the bottom 32-bits made it through the variadic call. I switched it to pass args in a thread local instead. Should be safer.

steven-johnson:
Looks like correctness_unroll_huge_mux is segfaulting on llvm14-x86-64-windows

alexreinking:
I think it's time to CRD into the windows buildbot and just binary search our way via the env-var to the right stack size for the mux test case.

abadams:
Windows is currently passing.

alexreinking:
> 
> 
> Windows is currently passing.

Oh, good. I saw a Windows failure earlier, but maybe it was re-run?

PR #6240, opened 2021-09-10
Join generator watchdog timer if exception thrown

abadams:
Make sure to join the generator watchdog timer when an exception is
thrown. Otherwise the error message gets swallowed entirely and you just
get "terminate called without an active exception"

Also rejiggered the watchdog to do a timed wait on a condition variable
instead of polling every 100ms.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1,5 +1,6 @@
 #include <atomic>
 #include <cmath>
+#include <condition_variable>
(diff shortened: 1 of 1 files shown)


alexreinking (approved):

alexreinking:
Failure is unrelated... looks like win-worker-1's LLVM trunk isn't building

PR #6248, opened 2021-09-20
Make buffer_get_* calls pure

dsharletg:


abadams:
Only some of these are pure. If you have an extern stage then the extern bounds query returns different results for a buffer_get_foo method before and after the call.

src/AddImageChecks.cpp
```diff
@@ -291,9 +291,9 @@ Stmt add_image_checks_inner(Stmt s,
                                                 param.name() + ".bounds_query." + extern_user);
                 for (int j = 0; j < dimensions; j++) {
                     Expr min = Call::make(Int(32), Call::buffer_get_min,
-                                          {query_buf, j}, Call::Extern);
```
abadams, line 294: This one isn't pure

PR #6251, opened 2021-09-20
Create helper script to run IWYU on Halide

steven-johnson:
My main workstation was offline friday afternoon thru monday morning, so I took the opportunity to try to plug in IWYU, which is infamously ticklish to get right. This PR includes a helper script to run it, along with a 'mapping' file and a handful of changes to source code to enable IWYU to avoid getting confused and making a mess. (There will be a dependent PR off of this one that will actually apply IWYU.)

In theory, this sort of cleanup could improve compile speed; however, I'm not noticing anything meaningful in my local timings (though perhaps the armbots may notice a change). The appeal is more one of appealing to OCD, frankly.

Please note that unlike clang-format or clang-tidy, this is *NOT* intended to be run on a regular basis (and certainly not as a presubmit check); because of the nature of C++ includes, getting IWYU is a bit of a black art and fairly system dependent. I only bothered to get this running correctly on OSX using a Homebrew-installed IWYU 0.16; running on anything else is likely to require manual tweaking and examination every time this is used, which should be infrequent.


steven-johnson:
I'm going to convert this to draft (but leave it open), since it seems unlikely to be of interest any time soon. (I'll harvest a couple of minor changes to land separately).

PR #6253, opened 2021-09-20
wip

steven-johnson:


abadams:
I guess I'm not really convinced that transitive includes are a bad thing. This just seems like it adds lots of lines of code (e.g. including Error.h everywhere).

Definitely worth removing unused includes though. Can we set it up to just detect that?

steven-johnson:
> I guess I'm not really convinced that transitive includes are a bad thing. This just seems like it adds lots of lines of code (e.g. including Error.h everywhere).

At a very general level I disagree; having each file explicitly include what it uses is IMHO a best practice and makes the code more resilient. But I really don't feel strongly about it. I'll leave this sitting open for a while to see if there are other people with strong feelings in any direction.

> Definitely worth removing unused includes though. Can we set it up to just detect that?

Eh, maybe, I'll circle back at some point and see.

PR #6266, opened 2021-09-24
[WIP] Try to reshuffle install footprint

LebedevRI:
I don't know with how much of this upstream will be okay with,
but i think this is basically the last nits i have for #4284.

All this is pretty minor, but i think it moves halide more in-line
with all the standard packages, so that the right stuff is installed
into the right places out-of-the-box.

I don't claim that this change is complete, there may be missing pieces.
In particular i haven't touched build systems other than CMake here.

This is how it looks now:
```
[ 99% 3354/3355][ 99% 0:01:18 + 0:00:00] Install the project...
-- Install configuration: "Release"
-- Installing: /opt/halide-new/lib/libHalide.so.13.0.0
-- Installing: /opt/halide-new/lib/libHalide.so.13
-- Set runtime path of "/opt/halide-new/lib/libHalide.so.13.0.0" to ""
-- Installing: /opt/halide-new/lib/libHalide.so
-- Installing: /opt/halide-new/lib/libHalide_autoschedule_adams2019.so
-- Set runtime path of "/opt/halide-new/lib/libHalide_autoschedule_adams2019.so" to ""
-- Installing: /opt/halide-new/lib/libHalide_autoschedule_li2018.so
-- Set runtime path of "/opt/halide-new/lib/libHalide_autoschedule_li2018.so" to ""
-- Installing: /opt/halide-new/lib/libHalide_autoschedule_mullapudi2016.so
-- Set runtime path of "/opt/halide-new/lib/libHalide_autoschedule_mullapudi2016.so" to ""
-- Installing: /opt/halide-new/include/halide
-- Installing: /opt/halide-new/include/halide/HalideBuffer.h
-- Installing: /opt/halide-new/include/halide/HalidePyTorchCudaHelpers.h
-- Installing: /opt/halide-new/include/halide/HalidePyTorchHelpers.h
-- Installing: /opt/halide-new/include/halide/HalideRuntime.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeCuda.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeD3D12Compute.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeHexagonDma.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeHexagonHost.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeMetal.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeOpenCL.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeOpenGLCompute.h
-- Installing: /opt/halide-new/include/halide/HalideRuntimeQurt.h
-- Installing: /opt/halide-new/include/halide/Halide.h
-- Installing: /opt/halide-new/bin/halide_retrain_cost_model
-- Set runtime path of "/opt/halide-new/bin/halide_retrain_cost_model" to "$ORIGIN:$ORIGIN/../lib"
-- Installing: /opt/halide-new/bin/halide_featurization_to_sample
-- Set runtime path of "/opt/halide-new/bin/halide_featurization_to_sample" to "$ORIGIN:$ORIGIN/../lib"
-- Installing: /opt/halide-new/bin/halide_get_host_target
-- Set runtime path of "/opt/halide-new/bin/halide_get_host_target" to "$ORIGIN:$ORIGIN/../lib"
-- Installing: /opt/halide-new/bin/halide_weightsdir_to_weightsfile
-- Set runtime path of "/opt/halide-new/bin/halide_weightsdir_to_weightsfile" to "$ORIGIN:$ORIGIN/../lib"
-- Installing: /opt/halide-new/share/doc/Halide/README_cmake.md
-- Installing: /opt/halide-new/share/doc/Halide/README.md
-- Installing: /opt/halide-new/share/doc/Halide/README_rungen.md
-- Installing: /opt/halide-new/share/doc/Halide/README_webassembly.md
-- Installing: /opt/halide-new/src/halide
-- Installing: /opt/halide-new/src/halide/GenGen.cpp
-- Installing: /opt/halide-new/src/halide/RunGen.h
-- Installing: /opt/halide-new/src/halide/RunGenMain.cpp
-- Installing: /opt/halide-new/src/halide/halide_benchmark.h
-- Installing: /opt/halide-new/src/halide/halide_image.h
-- Installing: /opt/halide-new/src/halide/halide_image_info.h
-- Installing: /opt/halide-new/src/halide/halide_image_io.h
-- Installing: /opt/halide-new/src/halide/halide_malloc_trace.h
-- Installing: /opt/halide-new/src/halide/halide_trace_config.h
-- Installing: /opt/halide-new/src/halide/mex_halide.m
-- Installing: /opt/halide-new/src/halide/autotune_loop.sh
-- Installing: /opt/halide-new/share/doc/Halide/tutorial
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/clock.h
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_02_input.jpg
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_02_output.jpg
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_col_major.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_fast.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_parallel_tiles.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_row_major.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_split_7_by_3.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_tiled.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_05_vectors.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_08_compute_root.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_08_compute_y.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_08_mixed.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_08_store_root_compute_x.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_08_store_root_compute_y.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_08_tile.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_compute_at_multiple_updates.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_compute_at_pure.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_compute_at_pure_and_update.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_compute_at_rvar.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_compute_at_update.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_inline_reduction.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_update.gif
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_update_rdom.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_09_update_schedule.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_17_rdom_calls_in_predicate.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_17_rdom_circular.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_17_rdom_triangular.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_18_hist_manual_par.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_18_hist_rfactor_par.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_18_hist_rfactor_tile.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_18_hist_rfactor_vec.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_18_hist_serial.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_19_group_updates.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_19_transpose.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_19_wrapper_global.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_19_wrapper_local.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_19_wrapper_unique.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/figures/lesson_19_wrapper_vary_schedule.mp4
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/images
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/images/gray.png
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/images/rgb.png
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_01_basics.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_02_input_image.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_03_debugging_1.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_04_debugging_2.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_05_scheduling_1.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_06_realizing_over_shifted_domains.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_07_multi_stage_pipelines.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_08_scheduling_2.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_09_update_definitions.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_10_aot_compilation_generate.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_10_aot_compilation_run.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_11_cross_compilation.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_12_using_the_gpu.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_13_tuples.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_14_types.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_15_generators.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_16_rgb_generate.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_16_rgb_run.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_17_predicated_rdom.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_18_parallel_associative_reductions.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_19_wrapper_funcs.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_20_cloning_funcs.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_21_auto_scheduler_generate.cpp
-- Installing: /opt/halide-new/share/doc/Halide/tutorial/lesson_21_auto_scheduler_run.cpp
-- Installing: /opt/halide-new/lib/cmake/Halide/Halide-shared-deps.cmake
-- Installing: /opt/halide-new/lib/cmake/Halide/Halide-shared-targets.cmake
-- Installing: /opt/halide-new/lib/cmake/Halide/Halide-shared-targets-release.cmake
-- Installing: /opt/halide-new/lib/cmake/HalideHelpers/Halide-Interfaces.cmake
-- Installing: /opt/halide-new/lib/cmake/HalideHelpers/Halide-Interfaces-release.cmake
-- Installing: /opt/halide-new/lib/cmake/Halide/HalideConfig.cmake
-- Installing: /opt/halide-new/lib/cmake/Halide/HalideConfigVersion.cmake
-- Installing: /opt/halide-new/lib/cmake/HalideHelpers/HalideHelpersConfig.cmake
-- Installing: /opt/halide-new/lib/cmake/HalideHelpers/HalideHelpersConfigVersion.cmake
-- Installing: /opt/halide-new/lib/cmake/HalideHelpers/HalideGeneratorHelpers.cmake
-- Installing: /opt/halide-new/lib/cmake/HalideHelpers/HalideTargetHelpers.cmake
```
which is actually what i would have expected it to be in the first place.

### DIFF
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -14,7 +14,7 @@ set(Halide_INSTALL_HELPERSDIR "${CMAKE_INSTALL_LIBDIR}/cmake/HalideHelpers"
 set(Halide_INSTALL_PLUGINDIR "${CMAKE_INSTALL_LIBDIR}"
     CACHE STRING "Path to Halide plugins folder")
 
-set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/tools"
+set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_PREFIX}/src/halide"
     CACHE STRING "Path to Halide build-time tools and sources")
 
 ##
@@ -61,7 +61,7 @@ install(TARGETS Halide_Runtime
 
 # Captures both the runtime and Halide.h
 install(DIRECTORY ${Halide_BINARY_DIR}/include/
-        TYPE INCLUDE
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/halide
         COMPONENT Halide_Development
         FILES_MATCHING PATTERN "include/*.h")
 
@@ -99,7 +99,7 @@ install(FILES
         ${Halide_SOURCE_DIR}/README_rungen.md
         ${Halide_SOURCE_DIR}/README_webassembly.md
         COMPONENT Halide_Documentation
-        TYPE DATA)
+        DESTINATION ${CMAKE_INSTALL_DOCDIR})
 
 ##
 # Tools
@@ -130,12 +130,12 @@ install(FILES ${Halide_SOURCE_DIR}/src/autoschedulers/adams2019/autotune_loop.sh
 
 if (WITH_TUTORIALS)
     install(DIRECTORY ${Halide_SOURCE_DIR}/tutorial
-            TYPE DATA
+            DESTINATION ${CMAKE_INSTALL_DOCDIR}
             COMPONENT Halide_Documentation
             FILES_MATCHING
             PATTERN "*.cpp"
             PATTERN "*.h"
-            PATTERN "lesson_*.sh"
+            PATTERN "lesson_*/opt/halide-new/share/tools"
             PATTERN "*.gif"
             PATTERN "*.jpg"
             PATTERN "*.mp4"
--- a/src/autoschedulers/adams2019/Makefile
+++ b/src/autoschedulers/adams2019/Makefile
@@ -57,7 +57,7 @@ $(BIN)/cost_model/%.a: $(BIN)/cost_model.generator
 # It's important to use dynamic lookups for undefined symbols here: all of libHalide
 # is expected to be present (in the loading binary), so we explicitly make the symbols
 # undefined rather than dependent on libHalide.so.
-$(BIN)/libautoschedule_adams2019.$(SHARED_EXT): $(SRC)/AutoSchedule.cpp \
+$(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT): $(SRC)/AutoSchedule.cpp \
 				$(SRC)/ASLog.cpp \
 				$(SRC)/Cache.h \
 				$(SRC)/Cache.cpp \
@@ -82,7 +82,7 @@ $(BIN)/libautoschedule_adams2019.$(SHARED_EXT): $(SRC)/AutoSchedule.cpp \
 	@mkdir -p $(@D)
 	$(CXX) -shared $(USE_EXPORT_DYNAMIC) -fPIC -fvisibility=hidden -fvisibility-inlines-hidden $(CXXFLAGS) $(OPTIMIZE) -I $(BIN)/cost_model $(filter-out %.h $(LIBHALIDE_LDFLAGS),$^) -o $@ $(HALIDE_SYSTEM_LIBS) $(HALIDE_RPATH_FOR_LIB)
 
-$(BIN)/retrain_cost_model: $(SRC)/retrain_cost_model.cpp \
+$(BIN)/halide_retrain_cost_model: $(SRC)/halide_retrain_cost_model.cpp \
 				$(SRC)/ASLog.cpp \
 				$(SRC)/DefaultCostModel.h \
 				$(SRC)/DefaultCostModel.cpp \
@@ -96,14 +96,14 @@ $(BIN)/retrain_cost_model: $(SRC)/retrain_cost_model.cpp \
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) -frtti -Wall -I ../support -I $(BIN)/cost_model $(OPTIMIZE) $(filter-out %.h,$^) -o $@ $(LIBHALIDE_LDFLAGS) $(USE_OPEN_MP) $(HALIDE_RPATH_FOR_BIN)
 
-$(BIN)/featurization_to_sample: $(SRC)/featurization_to_sample.cpp
+$(BIN)/halide_featurization_to_sample: $(SRC)/halide_featurization_to_sample.cpp
 	@mkdir -p $(@D)
-	$(CXX) $(CXXFLAGS) $< $(OPTIMIZE) -o $@ 
+	$(CXX) $(CXXFLAGS) $< $(OPTIMIZE) -o $@
 
-$(BIN)/get_host_target: $(SRC)/get_host_target.cpp $(LIB_HALIDE) $(HALIDE_DISTRIB_PATH)/include/Halide.h
+$(BIN)/halide_get_host_target: $(SRC)/halide_get_host_target.cpp $(LIB_HALIDE) $(HALIDE_DISTRIB_PATH)/include/Halide.h
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $(filter %.cpp,$^) $(LIBHALIDE_LDFLAGS) $(OPTIMIZE) -o $@ $(HALIDE_RPATH_FOR_BIN)
-$(BIN)/weightsdir_to_weightsfile: $(SRC)/weightsdir_to_weightsfile.cpp $(SRC)/Weights.cpp
+$(BIN)/halide_weightsdir_to_weightsfile: $(SRC)/halide_weightsdir_to_weightsfile.cpp $(SRC)/Weights.cpp
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $^ $(OPTIMIZE) -o $@
 
@@ -120,22 +120,22 @@ $(GENERATOR_BIN)/demo.generator: $(SRC)/demo_generator.cpp $(GENERATOR_DEPS)
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) -g $(filter %.cpp,$^) -o $@ $(LIBHALIDE_LDFLAGS)
 
 # To use the autoscheduler, set a few environment variables and use the -p flag to the generator to load the autoscheduler as a plugin
-$(BIN)/%/demo.a: $(GENERATOR_BIN)/demo.generator $(BIN)/libautoschedule_adams2019.$(SHARED_EXT)
+$(BIN)/%/demo.a: $(GENERATOR_BIN)/demo.generator $(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT)
 	@mkdir -p $(@D)
 	HL_WEIGHTS_DIR=$(SRC)/baseline.weights \
-	$(GENERATOR_BIN)/demo.generator -g demo -o $(@D) -f demo target=$* auto_schedule=true -p $(BIN)/libautoschedule_adams2019.$(SHARED_EXT) -s Adams2019
+	$(GENERATOR_BIN)/demo.generator -g demo -o $(@D) -f demo target=$* auto_schedule=true -p $(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT) -s Adams2019
 
 $(BIN)/%/demo.rungen: $(BIN)/%/RunGenMain.o $(BIN)/%/demo.registration.cpp $(BIN)/%/demo.a
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) -I$(BIN)/$* $^ -o $@ $(HALIDE_SYSTEM_LIBS) $(IMAGE_IO_FLAGS)
 
 # demonstrates single-shot use of the autoscheduler
-demo: $(BIN)/$(HL_TARGET)/demo.rungen $(BIN)/libautoschedule_adams2019.$(SHARED_EXT)
+demo: $(BIN)/$(HL_TARGET)/demo.rungen $(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT)
 	$< --benchmarks=all --benchmark_min_time=1 --estimate_all
 
 # demonstrates an autotuning loop
 # (using $(BIN) and $(SRC) here seems overkill, but makes copy-n-paste elsewhere easier)
-autotune: $(GENERATOR_BIN)/demo.generator $(BIN)/featurization_to_sample $(BIN)/get_host_target $(BIN)/retrain_cost_model $(BIN)/libautoschedule_adams2019.$(SHARED_EXT) $(SRC)/autotune_loop.sh
+autotune: $(GENERATOR_BIN)/demo.generator $(BIN)/halide_featurization_to_sample $(BIN)/halide_get_host_target $(BIN)/halide_retrain_cost_model $(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT) $(SRC)/autotune_loop.sh
 	@mkdir -p $(@D)
 	bash $(SRC)/autotune_loop.sh \
 		$(GENERATOR_BIN)/demo.generator \
@@ -155,7 +155,7 @@ $(BIN)/test_function_dag: $(SRC)/test_function_dag.cpp $(SRC)/FunctionDAG.h $(SR
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) $(filter-out %.h,$^) -o $@ $(LIBHALIDE_LDFLAGS) $(HALIDE_SYSTEM_LIBS)
 
 # Simple jit-based test
-$(BIN)/%/test: $(SRC)/test.cpp $(BIN)/libautoschedule_adams2019.$(SHARED_EXT)
+$(BIN)/%/test: $(SRC)/test.cpp $(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT)
 	@mkdir -p $(@D)
 	$(CXX) $(CXXFLAGS) $(USE_EXPORT_DYNAMIC) $^ -o $@ $(LIBHALIDE_LDFLAGS) $(HALIDE_SYSTEM_LIBS)
 
@@ -166,7 +166,7 @@ test_function_dag: $(BIN)/test_function_dag
 	$^
 
 run_test: $(BIN)/$(HL_TARGET)/test
-	HL_WEIGHTS_DIR=$(SRC)/baseline.weights LD_LIBRARY_PATH=$(BIN):$(LD_LIBRARY_PATH) $< $(BIN)/libautoschedule_adams2019.$(SHARED_EXT)
+	HL_WEIGHTS_DIR=$(SRC)/baseline.weights LD_LIBRARY_PATH=$(BIN):$(LD_LIBRARY_PATH) $< $(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT)
 
 .PHONY: test clean
 
@@ -178,10 +178,10 @@ build: $(BIN)/$(HL_TARGET)/test \
 	$(BIN)/test_function_dag \
 	$(BIN)/$(HL_TARGET)/included_schedule_file.rungen \
 	$(GENERATOR_BIN)/demo.generator \
-	$(BIN)/featurization_to_sample \
-	$(BIN)/get_host_target \
-	$(BIN)/retrain_cost_model \
-	$(BIN)/libautoschedule_adams2019.$(SHARED_EXT)
+	$(BIN)/halide_featurization_to_sample \
+	$(BIN)/halide_get_host_target \
+	$(BIN)/halide_retrain_cost_model \
+	$(BIN)/libHalide_autoschedule_adams2019.$(SHARED_EXT)
 
 test: run_test test_perfect_hash_map test_function_dag demo test_included_schedule_file autotune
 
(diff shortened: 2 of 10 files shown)


src/autoschedulers/adams2019/Makefile
```diff
@@ -82,7 +82,7 @@ $(BIN)/libautoschedule_adams2019.$(SHARED_EXT): $(SRC)/AutoSchedule.cpp \
 	@mkdir -p $(@D)
 	$(CXX) -shared $(USE_EXPORT_DYNAMIC) -fPIC -fvisibility=hidden -fvisibility-inlines-hidden $(CXXFLAGS) $(OPTIMIZE) -I $(BIN)/cost_model $(filter-out %.h $(LIBHALIDE_LDFLAGS),$^) -o $@ $(HALIDE_SYSTEM_LIBS) $(HALIDE_RPATH_FOR_LIB)
 
-$(BIN)/retrain_cost_model: $(SRC)/retrain_cost_model.cpp \
+$(BIN)/halide_retrain_cost_model: $(SRC)/halide_retrain_cost_model.cpp \
```
abadams, line 85: These are internal utilities, not really things that should be installed in any non-Halide-prefixed bin dir

LebedevRI: Yep, i guessed as much, which is why i'm //trying// to do all this,
because as it stands right now, now na[t]ive `ninja install` puts them into `${CMAKE_INSTALL_PREFIX}/bin`

abadams:
I think for most users most of the new halide-prefixing is just annoying line noise. IMO If these libraries and utilities are installed system-wide, they should be going in a directory with "halide" in the name somewhere. I'm not sure what the convention is for that. /usr/lib/halide/bin/? That's what llvm seems to do.

LebedevRI:
I think i'm trying to do too much in this single patch (which was useful to made sense of what needs to be done).
Let's see if some of these changes can be split off..

src/runtime/CMakeLists.txt
```diff
@@ -334,6 +334,8 @@ endforeach ()
 
 add_library(Halide_Runtime INTERFACE)
 add_library(Halide::Runtime ALIAS Halide_Runtime)
-target_include_directories(Halide_Runtime INTERFACE $<BUILD_INTERFACE:${Halide_BINARY_DIR}/include>)
+target_include_directories(Halide_Runtime INTERFACE
+  $<BUILD_INTERFACE:${Halide_BINARY_DIR}/include>
+  $<INSTALL_INTERFACE:include/halide>)
```
alexreinking, line 339: This is wrong. The `INCLUDES DESTINATION` argument in `packaging/CMakeLists.txt` sets it up correctly already.

packaging/CMakeLists.txt
```diff
@@ -130,12 +130,12 @@ install(FILES ${Halide_SOURCE_DIR}/src/autoschedulers/adams2019/autotune_loop.sh
...
 if (WITH_TUTORIALS)
     install(DIRECTORY ${Halide_SOURCE_DIR}/tutorial
-            TYPE DATA
+            DESTINATION ${CMAKE_INSTALL_DOCDIR}
             COMPONENT Halide_Documentation
             FILES_MATCHING
             PATTERN "*.cpp"
             PATTERN "*.h"
-            PATTERN "lesson_*.sh"
+            PATTERN "lesson_*/opt/halide-new/share/tools"
```
alexreinking, line 138: ...what is happening here?

packaging/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(Halide_INSTALL_HELPERSDIR "${CMAKE_INSTALL_LIBDIR}/cmake/HalideHelpers"
 set(Halide_INSTALL_PLUGINDIR "${CMAKE_INSTALL_LIBDIR}"
     CACHE STRING "Path to Halide plugins folder")
 
-set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/tools"
+set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_PREFIX}/src/halide"
```
alexreinking, line 17: I don't see either name as being particularly more or less standard than the other, and `DATADIR` in most installations will be set to `share/[hH]alide`, anyway, so the suffix is unnecessary. I would say to drop this change for the sake of keeping the PR minimal.

packaging/CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ install(TARGETS Halide_Runtime
 
 # Captures both the runtime and Halide.h
 install(DIRECTORY ${Halide_BINARY_DIR}/include/
-        TYPE INCLUDE
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/halide
```
alexreinking, line 64: I will not accept a patch that unconditionally adds a suffix to a GNUInstallDirs variable, sorry.

packaging/CMakeLists.txt
```diff
@@ -99,7 +99,7 @@ install(FILES
         ${Halide_SOURCE_DIR}/README_rungen.md
         ${Halide_SOURCE_DIR}/README_webassembly.md
         COMPONENT Halide_Documentation
-        TYPE DATA)
+        DESTINATION ${CMAKE_INSTALL_DOCDIR})
```
alexreinking, line 102: Covered by #6267

packaging/CMakeLists.txt
```diff
@@ -130,12 +130,12 @@ install(FILES ${Halide_SOURCE_DIR}/src/autoschedulers/adams2019/autotune_loop.sh
 
 if (WITH_TUTORIALS)
     install(DIRECTORY ${Halide_SOURCE_DIR}/tutorial
-            TYPE DATA
+            DESTINATION ${CMAKE_INSTALL_DOCDIR}
```
alexreinking, line 133: Covered by #6267

alexreinking (changes requested):
I'll leave it to Andrew to address how we want to distribute the autoscheduler utilities going forward. Personally, I'm fine shuffling them into a (totally configurable) subdirectory, but I'm not wild about adding `Halide_` prefixes to their names.

LebedevRI:
Thank you for taking a look! Yes, this is mostly just a proof of concept,
i'm sure we will be able to arrive at the same final results in an acceptable way,
step by step.

packaging/CMakeLists.txt
LebedevRI: @alexreinking this is going to be contentious then.
Can you please explain why must the headers be installed directly into `/usr/include` and not `/usr/include/halide` ?

alexreinking: > Can you please explain why must the headers be installed directly into `/usr/include` and not `/usr/include/halide` ?

First, there is no "must". It is completely possible to set `-DCMAKE_INSTALL_INCLUDEDIR=include/halide` when configuring the build if that's what you need.

Hard-coding a suffix like this makes it impossible for a package maintainer for another system to remove that suffix without patching our build. It's not _always_ right to add. For instance, on Linux, libraries that are installed to `/opt` don't need it. `/opt/halide/include` is completely benign, but `/opt/halide/include/halide` is redundant at best and frustrating at worst. On Windows/Nuget, too, it's common to isolate libraries from each other like this, rather than installing them on top of each other in a common prefix.

Finally, and maybe most importantly, changing the default would be quite unexpected for our users since almost all Halide code in the wild uses `#include <Halide.h>`, not `#include <halide/Halide.h>`. I also don't see any real risk of name clashes since all of our headers are prefixed with either `Halide` or `halide_`.

packaging/CMakeLists.txt
alexreinking: Oh, also, you should _never_ read `CMAKE_INSTALL_PREFIX` in the CMakeLists.txt. Doing this will break `cmake --install build-dir --prefix <new-prefix>` workflows.

src/runtime/CMakeLists.txt
LebedevRI: Aha, that is good to know.

packaging/CMakeLists.txt
LebedevRI: > > Can you please explain why must the headers be installed directly into `/usr/include` and not `/usr/include/halide` ?
> 
> First, there is no "must". It is completely possible to set `-DCMAKE_INSTALL_INCLUDEDIR=include/halide` when configuring the build if that's what you need.
 
> Hard-coding a suffix like this makes it impossible for a package maintainer for another system to remove that suffix without patching our build. It's not _always_ right to add. For instance, on Linux, libraries that are installed to `/opt` don't need it. `/opt/halide/include` is completely benign, but `/opt/halide/include/halide` is redundant at best and frustrating at worst. On Windows/Nuget, too, it's common to isolate libraries from each other like this, rather than installing them on top of each other in a common prefix.

Okay, i see you point.

> Finally, and maybe most importantly, changing the default would be quite unexpected for our users since almost all Halide code in the wild uses `#include <Halide.h>`, not `#include <halide/Halide.h>`. I also don't see any real risk of name clashes since all of our headers are prefixed with `Halide`.

This is a non-issue  `#include <Halide.h>` **are** correct, and will continue to work,
as per your comment: https://github.com/halide/Halide/pull/6266#discussion_r717096810

Let me reframe the question: unless overridden by user (!!!),
would it make sense to default to installing these into subdirectory?

alexreinking: > This is a non-issue #include <Halide.h> are correct, and will continue to work, as per your comment: #6266 (comment)

That's right, _assuming_ the user is consuming Halide via CMake. My understanding is that our corporate users just call CMake to build a Halide distribution and then consume it with another build system (make, blaze, etc.). Changing the default will cause headaches for them. I think @abadams has more context on this.

> Let me reframe the question: unless overridden by user (!!!), would it make sense to default to installing these into subdirectory?

Because of the above, I am quite hesitant to change the default; however, it might make sense to rework our _presets_ and direct users towards those in the README / build documentation. There it might be appropriate to define a preset for a particular system that wants this convention.

abadams: They don't even necessarily do that. E.g. Google famously compiles all third party code from scratch with their own build system. What directories headers land in in a cmake-generated install package isn't going to affect corporate users meaningfully. At most they'll have to add a -I flag somewhere to their build setup.

alexreinking: Thanks, that eases my concerns somewhat. Someone out there might yet be affected, but we won't know about them until they complain, I guess. 

I still don't see the _advantage_ of making this the default, though. If you're installing to a local prefix, then you know what will be there and won't try to, like, install multiple Halide versions on top of each other. If you're creating a package, then you're most likely checking all these, anyway.

I also don't see any _harm_ if a user does put our ten or so `Halide*.h` headers in `/usr/include`. So this change just seems a bit frivolous to me.

I might change my mind if I could see how this new default would be more than aesthetics and even then worth the pain of changing the default. At least, our homebrew and vcpkg packages would need to be updated to continue conforming to the prescribed layouts, which will be a bit of a chore...

LebedevRI: Mostly, the same reason why docs aren't being installed into doc/ but into subdirectory.

While cmake defaults to installing docs into subdirectory
it is expected that a whole directory of headers is being installed,
and it is a common practice to name said directory after the project.

https://cmake.org/cmake/help/latest/command/install.html mentions this:
```
Projects wishing to follow the common practice of installing headers into a project-specific subdirectory will need to provide a destination rather than rely on the above.
```

The harm (at least the one i'm aware of) is subtle, and you have pointed it out already:
If the headers are not located in the directory that is in the default include path,
then unless user either includes them directly, or specifies their location via `-I`,
then they will not be found by the compiler. Having to explicitly specify the headers
that one wants to include leads to less error-prone expirience than
relying on the right ones being picked implicitly.

But as you have said, this will just work with cmake configs.

alexreinking: Let me think about this more while we work on finding the right place for the autoscheduler tools.

I'm going to look through a few sets of packaging guidelines for various platforms and see if there is a common consensus on any of these issues.

LebedevRI:
I've been thinking about this, and i think perhaps i'm approaching this wrong.
I wonder if halide should indeed instead somewhat mimic the LLVM installation footprint, i.e.:
* /usr/share/doc/halide*/*  (as it already does)
* /usr/lib/cmake/halide*/*   (as it already does)
* /usr/include/halide*/*  (change, what we just discussed here)
* /usr/lib/halide*/bin/*   (change, better than prefixing the binaries?)
* /usr/lib/halide*/lib/*    (change, better than prefixing the plugins?)
* /usr/lib/halide*/tools/*    (change)
* symlink for libHalide from /usr/lib into /usr/lib/halide*/lib/

Does this sound worse/better?

alexreinking:
> Does this sound worse/better?

This sounds better for some particular systems and worse for others. I'm pretty sure the layout you propose can already be accomplished by setting cache variables. See our Debian/Ubuntu packaging presets for a sketch...

LLVM is generally a poor model for CMake usage. And the LLVM packages you might see on Ubuntu were developed with the aid of the Ubuntu maintainers. LLVM's first-party binaries are pretty much unusable for library development. 

This is all to say that I want to keep our install rules in the CMakeLists.txt as unopinionated and as minimal as possible, to make the lives of a variety of package maintainers as easy as possible.

LebedevRI:
Thank you everyone.
While i'm not sure about `autotune_loop.sh`, while the defaults are suboptimal,
it looks like it's currently sufficiently configurable.

PR #6267, opened 2021-09-25
CMake: install docs into halide subdirectory of doc dir

LebedevRI:
`DATA` results in those being installed directly into `${CMAKE_INSTALL_PREFIX}/share`,
while they obviously should go into `${CMAKE_INSTALL_PREFIX}/share/doc/Halide`.

Split off of #6266.

### DIFF
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -99,7 +99,7 @@ install(FILES
         ${Halide_SOURCE_DIR}/README_rungen.md
         ${Halide_SOURCE_DIR}/README_webassembly.md
         COMPONENT Halide_Documentation
-        TYPE DATA)
+        TYPE DOC)
 
 ##
 # Tools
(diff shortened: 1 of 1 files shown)


alexreinking:
For style, this PR should use `TYPE DOC` instead of `DESTINATION ${CMAKE_INSTALL_DOCDIR}`

packaging/CMakeLists.txt
```diff
@@ -99,7 +99,7 @@ install(FILES
         ${Halide_SOURCE_DIR}/README_rungen.md
         ${Halide_SOURCE_DIR}/README_webassembly.md
         COMPONENT Halide_Documentation
-        TYPE DATA)
+        DESTINATION ${CMAKE_INSTALL_DOCDIR})
```
alexreinking, line 102: `TYPE DOC`

alexreinking (changes requested):

alexreinking:
Should this be changed, too?

https://github.com/halide/Halide/blob/17ec57a23ccb468369577441ed910ae1a4d5f956/packaging/CMakeLists.txt#L211-L215

alexreinking:
Thanks, I just tested this by hand... it is a clear improvement.

alexreinking (approved):

alexreinking:
Merged! Thanks for the patch!

LebedevRI:
@alexreinking thank you for the review!

PR #6272, opened 2021-09-29
Fix alignment issues in synchronization_common.h

steven-johnson:
To work around old COMDAT issues, we allocated the table as a char array and cast it to what we want; unfortunately this doesn't guarantee the right alignment for the table and in some environments (eg wasm) we can get unaligned-access failures.

We could fix this by forcing the right alignment, but since we fixed COMDAT issues in another way a while back (adding smarts to LLVM_Runtime_Linker), let's just remove the hack and declare it normally.

Also added some drive-by changes to ensure that the hashtable size and HASH_TABLE_BITS were safe (this happened to be the case before but wasn't enforced), and also to init all the fields in hash_bucket.

(Q: do we really need `check_hash()` to exist? With the mods in place above, is it possible for addr_hash() to return a bad index?)

### DIFF
--- a/src/runtime/synchronization_common.h
+++ b/src/runtime/synchronization_common.h
@@ -494,18 +494,18 @@ WEAK void dump_hash() {
 }
 #endif
 
-static ALWAYS_INLINE uintptr_t addr_hash(uintptr_t addr, uint32_t bits) {
+static ALWAYS_INLINE uintptr_t addr_hash(uintptr_t addr) {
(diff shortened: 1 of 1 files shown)


dsharletg (approved):

shoaibkamil:
Ah, I was just about to report `RuntimeError: operation does not support unaligned accesses` on some WASM runs.  It looks like this may be the cause.

PR #6275, opened 2021-09-29
Make parking_control (etc) use vtables

steven-johnson:
This class hierarchy is clearly best modeled with virtual methods (rather than fn ptrs), but was not; we *think* this was due to COMDAT issues that have been resolved by other means. I refactored this to use virtual methods instead (and removed the unused unpark_all function); it seems to work locally.


steven-johnson:
Looks like there's more than meets the eye here, lots of compilation failures. Could be made to work but not essential. Closing but leaving branch in existence for now.

steven-johnson:
From local testing, looks like `-fno-rtti` is the trick needed; let's see what the buildbots think

src/runtime/synchronization_common.h
```diff
@@ -954,120 +875,96 @@ class fast_mutex {
...
-
-    return true;
-}
-
-WEAK void broadcast_parking_control_requeue_callback(void *control, const validate_action &action, bool one_to_wake, bool some_requeued) {
-    broadcast_parking_control *broadcast_control = (broadcast_parking_control *)control;
-
-    if (action.unpark_one && some_requeued) {
-        broadcast_control->mutex->make_parked();
+    bool validate(validate_action &action) override {
```
dsharletg, line 909: Mark these overrides final? Maybe they can be devirtualized sometimes? If they can be devirtualized completely, maybe we should use a different implementation (e.g. non-virtual template CRTP thing)

steven-johnson: Added the `final`. CRTP could be worth a try.

steven-johnson: FWIW, `-O2` enables `-fdevirtualize`; maybe this is happening already.

dsharletg: I think `-fdevirtualize` needs `final` to work, unless the compiler can prove there are no further derived classes in this case, but that seems unlikely...

steven-johnson: Eh, not sure that CRTP would pay for itself here; we'd be able to dispense with an indirect function call (via vtable) but we'd end up with nontrivial code duplication (the bodies of park, unpark_one, and unpark_requeue). Since we were already doing indirect function calls (via function pointer) the performance here should be a wash (just cleaner code IMHO).

alexreinking:
> From local testing, looks like `-fno-rtti` is the trick needed; let's see what the buildbots think

Why? What's the error? This seems a bit strange to me. Plus, disabling RTTI can change the ABI, so unless the runtime's full API surface is `extern "C"` (is it?) then this might have some unintended consequences.

steven-johnson:
> unless the runtime's full API surface is `extern "C"`

It is.

src/runtime/synchronization_common.h
```diff
@@ -571,35 +571,34 @@ struct validate_action {
...
+        return true;
+    }
+    virtual void before_sleep() {
+        // nothing
+    }
+    virtual uintptr_t unpark(int unparked, bool more_waiters) {
+        return 0;
+    }
+    virtual void requeue_callback(const validate_action &action, bool one_to_wake, bool some_requeued) {
+        // nothing
```
alexreinking, line 590: Do we ever want to create a base `parking_control`? If not, it might make sense to disable construction by adding a pure virtual destructor here.
```
    virtual ~parking_control() = 0;

protected:
    // ...
};

inline parking_control::~parking_control() = default;
```

steven-johnson: nah, but I'm not worries about it being misused in this way

alexreinking: Cool :)

alexreinking (approved):

PR #6277, opened 2021-09-29
Remove the runtime/ssp module

steven-johnson:
It doesn't get included via *any* path in the runtime linker, and removing it doesn't seem to affect any tests.

(I haven't looked at the revision history to see when it was added and/or when inclusion of it was removed.)


abadams:
I think this was for mingw

abadams:
Oh I guess the comment says native client. Either way, no longer a target.

abadams (approved):

Issue #6255, opened 2021-09-21
First-class support for execution on user-provided CUDA stream & context


mzient:
Hello,
I've looked at the documentation and the headers, but there's no mention of CUDA contexts or streams except in PyTorch helper. The functions `halide_cuda_acquire_context` and `halide_cuda_get_stream` could theoretically do the trick, but they rely on replacing symbols which is quite brittle and I'd expect this method to fail catastrophically if more than one shared object in the process defines these symbols. Is there a better way to supply these? Perhaps they could become members of `halide_device_interface_t` (with some type-erasing for `CUcontext` / `CUstream`)?

abadams:
The intended way to do it is indeed clobbering those symbols, ideally in the main executable so that contexts and streams are managed centrally rather than in a shared library. 

But yeah, we could also make them replaceable by function pointer, as we do with things like set_custom_do_par_for

mzient:
> ideally in the main executable so that contexts and streams are managed centrally rather than in a shared library

My main executable is Python and the target library is DALI. There will usually be some DL framework alongside with it (PyTorch, TF or MXNet) and possibly other libraries, so the conflict is likely. The fact that there is a helper to do this with PyTorch makes me quite uncomfortable.

mzient:
I think that runtime parameters (such as context, stream, threadpool implementation and the like) should be somehow passed to `realize`. This would allow using one function instance (and importantly - one compilation) for multiple independent realizations.

abadams:
The current way to do that is to make an object that represents your runtime and pass it as a user_context argument. Then you have a single override of the runtime that just casts the void * user context back to your abstract runtime interface and calls a virtual method on it.

abadams:
I'm not sure how well all this plays with JIT though. Most of our runtime-replacement shenanigans are in the context of AOT compilation.

mzient:
> I'm not sure how well all this plays with JIT though

Does the trick with `halide_cuda_acquire_context` and `halide_cuda_get_stream` work with JIT? Not being able to specify execution environment to a JIT-compiled function is a show stopper for me.

abadams:
Symbol replacement with JIT is definitely a different beast. It's done at runtime here: https://github.com/halide/Halide/blob/master/src/JITModule.cpp#L184

I'm not sure what the right way is to get a user-provided symbol into that resolution procedure. @zvookin should know. I suppose we could plumb a ```map<string, void *>``` through with the compile_jit call

abadams:
[...]

PR #6278, opened 2021-10-01
[Draft] Pass JITUserContext::user_context to custom handlers. Add UserContext param. 

mzient:
_The draft intended as a live demonstration of ideas discussed in #6255. It is still not ready for review, let alone merging._

Problem recap: currently, it's impossible to pass per-launch user context and handlers to JIT compiled functions. This becomes problematic when the target is an accelerator (e.g. CUDA device), where it may be desirable o execute the code in user-provided context. It's especially important when running in multi-device systems.
It also decouples the user context passed to custom JIT handlers from JITUserContext.

This draft explores possibilities of passing user context to JIT-compiled code.

Usage:

```C++
struct MyMemPool {
  // ...
  void *allocate(size_t);
  void free(void *);

  static MyMemPool &get(int device_id);
};

void *my_malloc(void *pool, size_t size) {
  return ((MyMemPool*)pool)->allocate(size);
}

void my_free(void *pool, void *mem) {
  return ((MyMemPool*)pool)->free(mem);
}

int main() {
  Func f("my_func");
  Param<float> scale;
  // define the function
  f.set_custom_allocator(my_malloc, my_free);
  f.compile_jit();

  // now the function can run in parallel with 2 different allocators
  
  std::thread t1([&]() {
    UserContextParam uc;
    ParamMap pm1({
      { scale, 2 },
      { uc, &MyMemPool::get(0) }
    });
    Buffer<uint8_t> out_buf1( /* sizes here */);
    f.realize(out_buf1, pm1);
  });

  std::thread t2([&]() {
    ParamMap pm2({
      { scale, 2 },
      { uc, &MyMemPool::get(1) }
    });
    Buffer<uint8_t> out_buf2( /* sizes here */);
    f.realize(out_buf2, pm2);
  });

  t1.join();
  t2.join();
}
```

Signed-off-by: Michał Zientkiewicz <mzient@gmail.com>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -156,6 +156,7 @@ set(HEADER_FILES
     UnpackBuffers.h
     UnrollLoops.h
     UnsafePromises.h
+    UserContextParam.h
     Util.h
     Var.h
     VectorizeLoops.h
--- a/src/JITModule.cpp
+++ b/src/JITModule.cpp
@@ -512,7 +512,7 @@ void merge_handlers(JITHandlers &base, const JITHandlers &addins) {
 void print_handler(void *context, const char *msg) {
     if (context) {
         JITUserContext *jit_user_context = (JITUserContext *)context;
-        (*jit_user_context->handlers.custom_print)(context, msg);
+        (*jit_user_context->handlers.custom_print)(jit_user_context->user_context, msg);
     } else {
         return (*active_handlers.custom_print)(context, msg);
     }
@@ -521,7 +521,7 @@ void print_handler(void *context, const char *msg) {
 void *malloc_handler(void *context, size_t x) {
     if (context) {
         JITUserContext *jit_user_context = (JITUserContext *)context;
-        return (*jit_user_context->handlers.custom_malloc)(context, x);
+        return (*jit_user_context->handlers.custom_malloc)(jit_user_context->user_context, x);
     } else {
         return (*active_handlers.custom_malloc)(context, x);
     }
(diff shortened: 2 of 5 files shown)


zvookin:
I need at least a description of how this is proposed to work. It likely should be in comments in the files extending the APIs. There is no test to show usage. Until this is addressed, I cannot do much with this PR.

I think the basic idea is ok, but the main reason you're having problems with the existing thing is that there wasn't enough thinking about use cases upfront. Even ```ParamMap``` is insufficiently documented that I keep having to justify its existence. (It's purpose is to provide an enduser accessible way to make JIT, ```realize```, calls that take parameters in a threadsafe way.)

mzient:
@zvookin Thanks for the comments.
This code is still very much in draft state - its intended purpose was to serve as an illustration for the discussion of the proposed feature rather than an actual pull request. When it's ready for review, I'll convert it to a regular PR. If this generates too much noise in your systems (I see that there were some CI jobs running, for example), please let me know - I'll close it and re-open at a later stage.

zvookin:
Per "its intended purpose was to serve as an illustration for the discussion of the proposed feature rather than an actual pull request"...

Expecting folks to figure out the issues that need to be discussed from the implementation alone is a lot to ask. Some sentences explaining what you are trying to do, decisions you've made, and perhaps 10 or 20 lines of code showing how this will be used, even if it doesn't compile or work yet, would go a long way toward facilitating the discussion needed. I believe these would also assist you in your work as well.

mzient:
I've added a usage example and linked the issue. We can also move the discussion here, if that's a preferred workflow in your project - or I can even close this draft until it reaches next readiness level.
I'm sorry if I made myself too much at home here, bringing with me the customs from my project, where drafts are rarely proactively looked at.

steven-johnson:

mzient:
> Is this PR still active? Should it be closed?

I think it can be closed - similar functionality is provided by #6313, which has already been merged.

PR #6280, opened 2021-10-02
Remove hopefully dead code.

zvookin:
The external buffer host alignment should always be reflected in the Param attached to a load or store. Remove special case code for buffer parameter arguments.


abadams:
All checks passed. What was that map even for?

abadams (approved):

zvookin:
I believe it predates the host alignment parameter and it was kept as a precaution or something.

PR #6287, opened 2021-10-05
Remove unbound variable in documentation

Elarnon:
In the example for RDom::where, the simplified case contains a free occurence of `r.x`, which should be replaced with `10` since we are in the case `r.x == 10`.


abadams (approved):
Thanks for the fix.

PR #6289, opened 2021-10-05
store_in(MemoryType::Stack) should use alloca if the size is small

abadams:
If an allocation for a Func marked as store_in MemoryType::Stack is just made once (e.g. because it's compute_at just inside a parallel loop), then we currently just call malloc and there's no win from placing it on the stack. 

This PR changes the behavior of MemoryType::Stack to call alloca instead when the size is small. This helps a lot in the single-use case. 

It's the same pseudostack logic as in master. It allocates on first use (even if inside a loop), and reallocates only if the required size grows. Stack is never freed. If the cumulative stack usage for all reallocations for one Func exceeds our can_fit_on_stack threshold, we switch to heap for all future reallocations of this Func.

The following three runtimes are from a test of the single-use case where in the first case we statically know the allocation size, so we can just size the original stack frame appropriately. In the second case we use the new behavior in this PR and subtract from the stack pointer by a computed amount. In the third case we call malloc, as we would on master.

```
Constant-sized stack allocation: 0.165252
Use alloca: 0.184218
Use malloc: 0.215683
```

The runtimes are close together because allocating isn't the only thing this pipeline does. You should view the first of the three figures as the zero, which means the new behavior is about 2.5x faster than the old behavior when it comes to the cost of allocating.

Those numbers are from linux. On macos malloc is slower and the win is 7x.

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -262,6 +262,7 @@ bool function_takes_user_context(const std::string &name) {
 
 bool can_allocation_fit_on_stack(int64_t size) {
     user_assert(size > 0) << "Allocation size should be a positive number\n";
+    // Should match the threshold defined in runtime/pseudostack.cpp
     return (size <= 1024 * 16);
 }
 
--- a/src/CodeGen_Posix.cpp
+++ b/src/CodeGen_Posix.cpp
@@ -221,42 +221,33 @@ CodeGen_Posix::Allocation CodeGen_Posix::create_allocation(const std::string &na
 
         // Even if we're reusing a stack slot, we need to call
         // pseudostack_alloc to potentially reallocate.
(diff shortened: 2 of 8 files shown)


dsharletg (commented):
Thanks for doing this!

src/CodeGen_Internal.cpp
```diff
@@ -262,6 +262,7 @@ bool function_takes_user_context(const std::string &name) {
 
 bool can_allocation_fit_on_stack(int64_t size) {
     user_assert(size > 0) << "Allocation size should be a positive number\n";
+    // Should match the threshold defined in runtime/pseudostack.cpp
```
alexreinking, line 265: I think it would be worthwhile to create a `runtime/constants.h` that's just a bunch of `#define`s or something that records constants that need to be kept consistent between the compiler and the runtime.

abadams:
builds are clean. ptaal

PR #6290, opened 2021-10-06
[hannk] Legalize broadcast binary ops

steven-johnson:
For binary ops that handle broadcasts, only the second input is allowed to be a broadcast. This adds a transforms that legalizes the operand before execution time by swapping the first and second op for those cases. Also adds some handcrafted .tflite files that exercise these cases. (We'd fail with a halide assertion before.)


apps/hannk/interpreter/ops.cpp
```diff
@@ -612,8 +614,13 @@ void BinaryOp::execute() {
 
         switch (op_) {
         case Add:
+            add_uint8(in1_buf, in1->quantization(), 1, in2_buf, in2->quantization(), 1, out_buf, out->quantization(), activation_);
+            return;
         case Sub:
-            add_uint8(in1_buf, in1->quantization(), 1, in2_buf, in2->quantization(), op_ == Add ? 1 : -1, out_buf, out->quantization(), activation_);
+            add_uint8(in1_buf, in1->quantization(), 1, in2_buf, in2->quantization(), -1, out_buf, out->quantization(), activation_);
+            return;
+        case ReverseSub:
```
dsharletg, line 622: Hmm, the need for ReverseSub is making me think maybe we should just bite the bullet and generate the 3 possible specializations for broadcasting instead of requiring swapping broadcasts to the second operand. The extra code size can't be that bad...

I think we don't need the 4th case (both broadcasting) because that shouldn't happen (we should optimize to broadcast after).

PR #6294, opened 2021-10-07
Add ClampUnsafeAccesses pass.

### LINKED ISSUE #6131: Accesses inside indexing expressions should be in compute bounds
This is a bit of a loose thought, but I'm wondering how bounds inference handles cases like this:

```
h(x) = 10;
g(x) = sin(x); // whatever
f(x) = g(h(x));

h.compute_root();
g.compute_root();
f.split(x, xo, xi, 8, TailStrategy::RoundUp);

f.realize(10);
```

It seems to me there might be a hazard with h(x) being uninitialized between x = 10-15, unless accesses inside index expressions are required to be in the compute bounds (so the mem bounds of `f` become the compute bounds of `h`) - is that how BI actually works?

I couldn't find a case (at least during POPL paper crunch) to _actually_ test this because the compiler is a bit too smart about inlining.

alexreinking:
Inject clamps around func calls h(...) when all the following conditions hold:
  1. The call is in an indexing context, such as: f(x) = g(h(x));
  2. The FuncValueBounds of h are smaller than those of its type

Fixes #6131

### DIFF
--- a/src/ClampUnsafeAccesses.cpp
+++ b/src/ClampUnsafeAccesses.cpp
@@ -42,7 +42,7 @@ protected:
         if (is_inside_indexing) {
             auto bounds = func_bounds.at({call->name, call->value_index});
             if (bounds_smaller_than_type(bounds, call->type)) {
-                // TODO: check that the clamped function's allocation bounds might be wider than its compute bounds
+                // TODO(#6297): check that the clamped function's allocation bounds might be wider than its compute bounds
 
                 auto [new_args, changed] = mutate_with_changes(call->args);
                 Expr new_call = changed ? call : Call::make(call->type, call->name, new_args, call->call_type, call->func, call->value_index, call->image, call->param);


src/ClampUnsafeAccesses.cpp
```diff
@@ -0,0 +1,88 @@
...
+            return IRMutator::visit(call);
+        }
+
+        if (is_inside_indexing) {
+            auto bounds = func_bounds.at({call->name, call->value_index});
+            if (bounds_smaller_than_type(bounds, call->type)) {
+                // TODO(#6297): check that the clamped function's allocation bounds might be wider than its compute bounds
+
+                auto [new_args, changed] = mutate_with_changes(call->args);
+                Expr new_call = changed ? call : Call::make(call->type, call->name, new_args, call->call_type, call->func, call->value_index, call->image, call->param);
```
steven-johnson, line 48: (resurrecting a zombie PR...)

Uh, is this right? Shouldn't it be `!changed ? call : ...` ?

alexreinking: Yeeeep, that looks like a typo. Want to try changing it to see if tests fail?

PR #6298, opened 2021-10-08
[CMake] Add an option to specify filename suffix to add to libHalide …

LebedevRI:
…library name

... which is empty by default, but can be helpful to support
side-by-side installation of multiple versions of the library.

I'm open to the exact option name, and the location where the option is introduced.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -366,6 +366,10 @@ add_library(Halide
             $<TARGET_OBJECTS:Halide_initmod>)
 add_library(Halide::Halide ALIAS Halide)
 
+set(Halide_LIBHALIDE_SUFFIX "" CACHE STRING "Suffix to add to libHalide library")
+
+set_target_properties(Halide PROPERTIES OUTPUT_NAME Halide${Halide_LIBHALIDE_SUFFIX})
+
 target_link_libraries(Halide PUBLIC Threads::Threads)
 target_link_libraries(Halide PRIVATE Halide::LLVM)
 target_link_libraries(Halide PUBLIC Halide::LanguageOptions)


LebedevRI:
@alexreinking hmm, i'm sorry, but no, `CMAKE_<CONFIG>_POSTFIX` doesn't work for this.
First, it adds the suffix to all the libraries, even to the autoschedule ones.
Secondly, there's a problem with
https://github.com/halide/Halide/blob/2bfa567a4f49bf8a83e23ae5806aac418d5f485a/src/CMakeLists.txt#L386-L387
That seems go to directly against the stated principle, it hardcodes some vales,
with no way to change them other than patching.

So there are three knobs missing: `Halide_LIBHALIDE_SUFFIX`, `Halide_LIBHALIDE_VERSION`, `Halide_LIBHALIDE_SOVERSION`.
Might you be interested in providing these knobs?

LebedevRI:
@alexreinking thoughts? "no" will be fine, but i think it would be good to know the answer.

LebedevRI:
> > @alexreinking thoughts? "no" will be fine, but i think it would be good to know the answer.
> 
> @LebedevRI - Sorry, I have been busy with other things... I'm not opposed, but I'm curious about your use-case. Can you tell me more about your build scenario? Maybe open a [Discussion](https://github.com/halide/Halide/discussions).

https://salsa.debian.org/pkg-llvm-team/halide/-/commit/e8da6a743af6c9363fa57cede75d4eee2898078d#5d36a055ea360deb30ad5e839ceec5c50c47b94b

alexreinking:
Okay, I think we can add a SOVERSION override in that case. Thanks for explaining.

I'm still not sure why the autoschedulers shouldn't be suffixed... they're very tied up with the particular libHalide they're built against.

LebedevRI:
It does sound interesting, but what about the
```diff
 set_target_properties(Halide PROPERTIES
                       POSITION_INDEPENDENT_CODE ON
-                      VERSION ${Halide_VERSION}
```
part?

alexreinking:
Unless you need to have _both_ `libHalide13.so.0` _and_ `libHalide13.so.1` installed, then `libHalide13.so.13.0.0` would conflict... do you actually anticipate needing this flexibility?

LebedevRI:
> Unless you need to have _both_ `libHalide13.so.0` _and_ `libHalide13.so.1` installed, then `libHalide13.so.13.0.0` would conflict... do you actually anticipate needing this flexibility?

That is a two-fold question.
If such a situation is not anticipated, then there is no point in having proper/any SOVERSION.
But if it happens in reality, handling it post-factum is harder than upfront,
and not handling it will lead to a fun expirience when SOVERSION actually has to be bumped.
Of course, all of this only applies if halide-dependent software ever gets packaged itself.

TLDR: it's perfectly fine if upstream wants to just hardcode one-size-fits-all solution here.

PR #6302, opened 2021-10-08
At some point llvm re-added pavgw intrinsics

abadams:
This is a good thing, because these do not reliably trigger from the
pattern in runtime/x86.ll


LebedevRI:
Thanks!

> Assembly before. God knows what has gone wrong here. I see two pavgw instructions, but also lots of 32-bit integer math.

(on godbolt: https://godbolt.org/z/ros94Gqb6)

abadams:
Also being fixed upstream:
https://reviews.llvm.org/D111571

Thanks @LebedevRI !

I'll leave this PR as-is, so that we have two chances to catch pavgw. If we miss it in our pattern-matching, hopefully llvm will catch it for us.

PR #6305, opened 2021-10-11
Rewrite double/triple narrowing from float on ARM

stevesuzuki-arm:
Add more cases to PR #5896 to support narrowing from floating point.
This works as the workaround of LLVM performance issue that `fptoui <16 x float> %x to <16 x i8>` is compiled into scalarized instructions.

### DIFF
--- a/src/LLVM_Headers.h
+++ b/src/LLVM_Headers.h
@@ -34,7 +34,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/ErrorHandling.h"
 #include <llvm/ADT/StringMap.h>
-#if LLVM_VERSION >= 12
+#if LLVM_VERSION >= 120
 #include <llvm/Analysis/AliasAnalysis.h>
 #endif
 #include <llvm/Analysis/TargetLibraryInfo.h>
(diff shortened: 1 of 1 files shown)


stevesuzuki-arm:
> Would be curious to see before/after assembly? u8_sat(f32) is a fairly common pattern at the end of floating point imaging pipelines that want to write out 8-bit results.

I've attached it as [u8_sat(f32).txt](https://github.com/halide/Halide/files/7324102/u8_sat.f32.txt)

stevesuzuki-arm:
I would suggest to proceed this topic as follows.
This PR handles scalarization issue in Arm only, which means I revert the 2nd commit and we go with the 1st commit only.
For the improvement of float to integer cast,  I'll file an separate Issue item, where we may discuss how to rework [saturating_cast()](https://github.com/halide/Halide/blob/071f5f79f76314ecfb23b87e5f6d4203c2707ae4/src/IROperator.cpp#L1421) and CodeGen_LLVM::[visit(const Cast *op) ](https://github.com/halide/Halide/blob/071f5f79f76314ecfb23b87e5f6d4203c2707ae4/src/CodeGen_LLVM.cpp#L1436)to lower the pattern into saturating LLVM intrinsic.
@abadams, please let me know if you suggest another way.

PR #6310, opened 2021-10-12
Fix default device wrap native function

OmarEmaraDev:
Currently, an attempt to call `device_wrap_native` on a target that uses
the default device wrap native function will result in an error of type
`halide_error_device_interface_no_device`, namely in the OpenGLCompute and
the Hexagon targets. This happens because the default wrap native
function calls `debug_log_and_validate_buf` after the `device_interface` is
set in `halide_device_wrap_native` but before the `device` handle is set,
which is validated as a bad state.

This patch removes the validation call and adds an assert for the handle
much like the other wrap_native implementations in other targets.


steven-johnson:
Update: this is causing some issues downstream, and I wonder at the logic of this change: `halide_assert()` isn't a debug-only assertion (it will just `abort()` if the condition fails, in all build modes)... so the error checking on the next line will never happen. This seems bad and wrong.

steven-johnson:
(side note: we should rename `halide_assert()` to something else, since IMHO it implies "debug mode only or at least not in full opt builds", which is *not* the case)

PR #6313, opened 2021-10-12
Add ability to pass a user context in JIT mode

abadams:
JIT mode has long hijacked user_context for its own purposes. It uses it to store overrides of runtime functions, and it uses it to store an error buffer for accumulating error messages. This means user_context isn't available in JIT mode, which leaves no good way to pass state to custom overrides in a thread-safe manner.

This PR promotes the existing JITUserContext struct to the public namespace, and lets you pass one per call to realize, instead of hiding it inside the realize implementation. By inheriting from this struct and passing a subclass in as the context pointer, you can pass additional state to your runtime overrides! This is demonstrated in the new test. 

Runtime overrides in JIT mode will now be expected to take a JITUserContext * as the first arg instead of a void *. The existing set_custom_foo methods that expect function pointers with void * first args are left in place, but I think we should deprecate them and add new ones that expect a JITUserContext *.

I didn't make any of the base class (JITUserContext) state private or anything, so as a side-effect, this means that handlers or error buffers can be overridden per call to realize now in a thread-safe manner.

### DIFF
--- a/src/Func.h
+++ b/src/Func.h
@@ -831,7 +831,9 @@ public:
 
     /** Same as above, but takes a custom user-provided context to be
      * passed to runtime functions. This can be used to pass state to
-     * runtime overrides in a thread-safe manner. */
+     * runtime overrides in a thread-safe manner. A nullptr context is
+     * legal, and is equivalent to calling the variant of realize
+     * that does not take a context. */
     Realization realize(JITUserContext *context,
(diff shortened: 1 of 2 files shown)


abadams:
`Buffer::copy_to_device` and similar all take a user_context argument as the first arg. Passing something derived from JITUserContext should work. I'll write a quick test.

PR #6317, opened 2021-10-13
[hannk] Ensure Model transforms are valid

steven-johnson:
- Add a Model validity checker to ensure that all the input tensors to a given op are outputs of some previous op (or a constant)
- Revamp OpGroup::add() and remove() to properly handle nested groups (they were previously producing Models that didn't pass the new validity checker)


steven-johnson:
Obsoleted by https://github.com/halide/Halide/pull/6379

PR #6322, opened 2021-10-15
Add a new unsigned division method

abadams:
It uses averages rounding up instead of averages rounding down, to
reduce instruction count on x86.

Division by 7 before:
	vpmulhuw	.LCPI0_1(%rip), %ymm0, %ymm1
	vpsubw	%ymm1, %ymm0, %ymm0
	vpsrlw	$1, %ymm0, %ymm0
	vpaddw	%ymm1, %ymm0, %ymm0
	vpsrlw	$2, %ymm0, %ymm0

Division by 7 after:
        vpmulhuw        .LCPI0_1(%rip), %ymm0, %ymm1
        vpavgw  %ymm0, %ymm1, %ymm0
        vpsrlw  $2, %ymm0, %ymm0

Brute-force verified to be correct

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -348,7 +348,17 @@ Expr lower_int_uint_div(const Expr &a, const Expr &b) {
             // Average with original numerator.
             val = Call::make(val.type(), Call::sorted_avg, {val, num}, Call::PureIntrinsic);
         } else if (method == 3) {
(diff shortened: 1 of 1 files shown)


abadams:
@LebedevRI might be interested in this PR, because it's a use of pavg that llvm could profit from. llvm currently lowers division by 7 and similar constants in the same way as halide master, and this technique should be faster

https://godbolt.org/z/7oeo993an

PR #6327, opened 2021-10-15
Scheduling directive to set an explicit storage bound

vksnk:
This is a proposal to add a scheduling directive which allows to explicitly set a bound for allocation/storage size of the function.

The idea is very simple: it basically allows you to specify an expression which will be used for the function storage allocation extent/size instead of the one inferred by the compiler. 

In order to make it safer it will generate an assert which ensures (at runtime) that the provided size is greater-equal than the one which was inferred by the compiler. Also, it will try to prove at compile time that inferred bound is greater than provided one and if yes will throw a user_error assert.   

This can be useful, for example, when dynamic allocation is very expensive (or not possible at all). Also, some compilers can do a better job optimizing the code when allocations are static.

Questions:
* I started with the name `bound_allocation`, but now think that maybe `bound_storage` would be better (more consistent with existing `align_storage`, `fold_storage`,  `reorder_storage`)? Any other better alternatives?
* Right now the `bound_allocation` function accepts a single expression, however, extents in the Allocate IR node is a vector of expressions. Is there a value of allowing users to provide a vector of extents? I guess it's easy to add overloads for both options just in case.
* I just realized that tuples may require some special care, like a way to provide a separate bound for each of the elements?
* If anyone has any suggestions for interesting and/or tricky test cases - I'll be happy to add.


dsharletg:
> Right now the bound_allocation function accepts a single expression, however, extents in the Allocate IR node is a vector of expressions. Is there a value of allowing users to provide a vector of extents? I guess it's easy to add overloads for both options just in case.

I think bound_extent already does this per-dimension bounding, and it's a little different. I think this should remain a single Expr for the "flat" bound of the allocation.

vksnk:
> > Right now the bound_allocation function accepts a single expression, however, extents in the Allocate IR node is a vector of expressions. Is there a value of allowing users to provide a vector of extents? I guess it's easy to add overloads for both options just in case.
> 
> I think bound_extent already does this per-dimension bounding, and it's a little different. I think this should remain a single Expr for the "flat" bound of the allocation.

`bound_extent` affects extents of both compute and storage, while `bound_storage` is only for storage, so it would allow a more fine-grained control.

PR #6348, opened 2021-10-25
Support vectorization in OpenGLCompute backend

### LINKED ISSUE #1687: OpenGLCompute needs vectorization handling
It was disabled in #1679 because it had when we made closures track buffers by their scalar type. In general the approach to vectorization in OpenGLCompute needs to be revised.

### LINKED ISSUE #4976: openglcompute backend doesn't support vector loads and stores
It should at least scalarize them

OmarEmaraDev:
This patch adds support for vector load and store operations in the
OpenGLCompute backend.
First, a pass identifies the buffers whose loads and stores are all
dense, aligned, and have the same number of lanes. Such buffers are
declared with a vector base type and accessed accordingly. Loads and
stores that do not satisfy those criteria are implemented as gathers and
scatters from buffers whose base type is scalar.

Resolves #4976.
Partially resolves #1687.

### DIFF
--- a/src/CodeGen_OpenGLCompute_Dev.cpp
+++ b/src/CodeGen_OpenGLCompute_Dev.cpp
@@ -765,7 +765,7 @@ public:
     int lanes = -1;
 
     CheckAlignedDenseVectorLoadStore(string buffer)
-        : buffer(buffer) {
+        : buffer(std::move(buffer)) {
     }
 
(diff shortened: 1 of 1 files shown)


dsharletg (approved):

steven-johnson:
Thanks for the fix!

PR #6361, opened 2021-10-28
[hannk] Add support for building/running for wasm

steven-johnson:
Adds limited support for building and running hannk under WebAssembly. (Note that this requires https://github.com/halide/Halide/pull/6360 to land first.)

Requirements:

- You must use CMake to build (Make isn't supported).
- You must have Emscripten v2.0.32 (or later) installed and activated.
- You must have Node.js v16.13 (or later) installed for testing. Note that (as of this writing), EMSDK includes an older version of Node that *will not* work.

Caveats: this uses some *seriously* janky tricks for cross-compiling (all in apps/hannk/halide/CMakeLists.txt); the basic idea is that when crosscompiling, we re-run CMake on the subdirectory, but with an attempt to revert to the default host tooling (rather than what Emscripten has put in place), then set up dependencies to build and run the Generators that way, and then re-connect the generated libraries in the enclosing CMake file.

This works for me locally, and may be acceptable for an app of this sort. (Whether it is robust enough for "real" apps is not clear.) As usual, @alexreinking, I welcome your cmake-fu to let me know whether this is liable to break immediately and/or could be improved. (If this approach actually seems reasonable, it might be good to find a way to generalize it to allow easier wasm builds of other apps...)

I'm going to go ahead and request a review for this, but it may well need more attention before landing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.16...3.20)
 project(Halide
-        VERSION 13.0.0
+        VERSION 14.0.0
         DESCRIPTION "Halide compiler and libraries"
         HOMEPAGE_URL "https://halide-lang.org")
 
--- a/apps/hannk/.gitignore
+++ b/apps/hannk/.gitignore
@@ -1,2 +1,3 @@
 build*/
 bin/
+emsdk
--- a/apps/hannk/.gitignore
+++ b/apps/hannk/.gitignore
@@ -1,2 +1,2 @@
-build/
+build*/
 bin/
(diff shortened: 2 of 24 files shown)


steven-johnson:
With #6365 merged into this branch, this should be good to go once the buildbots go green

alexreinking:
> With #6365 merged into this branch, this should be good to go once the buildbots go green

Sounds good to me. I have another PR queued up that switches HANNK over to using the helpers introduced in #6366. It also reworks cross-compiling detection and makes one call to ninja rebuild the world :+1:

It's ready to go, provided we can get both PRs merged today.

alexreinking (approved):

PR #6364, opened 2021-10-28
[hannk] Clean up aliasing (v2)

steven-johnson:
The code for aliasing tensors was janky. This cleans it up and makes a clear distinction between aliasing done to overlay buffers with crop-and-translate, vs the aliasing done when we reshape tensors. We no longer allow a given tensor to do both of these, and we give preference to Reshape aliasing first.

(Cherry-picked from #6321)

### DIFF
--- a/apps/hannk/interpreter/tensor.h
+++ b/apps/hannk/interpreter/tensor.h
@@ -87,12 +87,26 @@ class Tensor {
     // Currently only used in conjunction with the TFLite delegate.
     bool is_dynamic_ = false;
 
-    // All the Tensors in this list share their TensorStorage with each other.
-    // An arbitrary number of Tensors can alias each other, but at most *one* may be external.
-    //
-    // Note that this list should never have size() of 0 or 1 (the pointer should simply be null).
-    std::shared_ptr<std::vector<std::weak_ptr<Tensor>>> aliases_;
-    AliasType alias_type_ = AliasType::None;
+    struct AliasInfo {
+        // All the Tensors in this list share their TensorStorage with each other.
+        // An arbitrary number of Tensors can alias each other, but at most *one* may be external.
+        //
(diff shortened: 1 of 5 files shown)


steven-johnson:
Monday morning review ping

apps/hannk/interpreter/tensor.h
```diff
@@ -78,12 +86,13 @@ class Tensor {
...
-    // and may have a different rank from the underlying storage.
-    // Only used if is_alias_ = true.
-    bool is_reshape_alias_ = false;
+
+    // All the Tensors in this list share their TensorStorage with each other.
+    // An arbitrary number of Tensors can alias each other, but at most *one* may be external.
+    //
+    // Note that this list should never have size() of 0 or 1 (the pointer should simply be null).
+    std::shared_ptr<std::vector<std::weak_ptr<Tensor>>> aliases_;
+    AliasType alias_type_ = AliasType::None;
```
dsharletg, line 95: Shouldn't this be inside the shared_ptr above? A tuple/struct with the vector?

apps/hannk/interpreter/tensor.h
```diff
@@ -207,10 +219,17 @@ class Tensor {
 
     void resize_dynamic(const Box &new_shape);
 
-    bool is_alias() const {
-        return is_alias_;
+    // Return true iff this Tensor aliases another Tensor.
```
dsharletg, line 222: Comment looks stale

dsharletg (approved):

apps/hannk/interpreter/tensor.h
steven-johnson: Hm, yes, that is probably a better design, will do.

apps/hannk/interpreter/tensor.h
steven-johnson: It is indeed

PR #6366, opened 2021-10-29
Add helper for cross-compiling Halide generators.

alexreinking:
Created a new function, `add_halide_generator`, that helps users write correct cross-compiling builds by establishing the following convention for creating a generator named `TARGET`:

1. Define Halide generators and libraries in the same project
2. Assume two builds: a host build and a cross build.
3. When creating a generator, check to see if we can load a pre-built version of the target.
4. If so, just use it.
5. If not, make sure the full Halide package is loaded and create a target for the generator.
     a. If `CMAKE_CROSSCOMPILING` is set, then _warn_ the user (the variable is unreliable on macOS) that something seems fishy.
     b. Create export rules for the generator. It creates a package `PACKAGE_NAME` and appends to its `EXPORT_FILE`.
     c. Create a custom target also named `PACKAGE_NAME` for building the generators.
     d. Create an alias `${PACKAGE_NAMESPACE}${TARGET}`.
6. Users are expected to use the alias in conjunction `add_halide_library`. Users can test the existence of `TARGET` to determine whether a pre-built one was loaded (and set additional properties if not).
7. Setting `${PACKAGE_NAME}_ROOT` is enough to load pre-built generators.

`PACKAGE_NAME` is `${PROJECT_NAME}-halide_generators` by default.
`PACKAGE_NAMESPACE` is `${PROJECT_NAME}::halide_generators::` by default.
`EXPORT_FILE` is `${PROJECT_BINARY_DIR}/cmake/${PACKAGE_NAME}-config.cmake` by default.

Users are free to avoid the helper if it would not fit their workflow.

This will be useful to integrate into HANNK. It abstracts the approach in #6365 and will hopefully enable cross-compilation for the other apps, too.

Skipping buildbots because only the `test/integration` code uses it for now (and only the GHA tests use this)


steven-johnson (commented):
Generally looks good -- relevant comments are mostly on the other PR right now.

steven-johnson:
What needs to happen for this to land -- do we have enough test coverage? Do we need to update apps/hannk to use this rule?

alexreinking:
> What needs to happen for this to land -- do we have enough test coverage? Do we need to update apps/hannk to use this rule?

I have already updated HANNK to use this in a separate branch... would you like me to group these changes together, then?

alexreinking:
Here, I'm adding the HANNK change so that we can look at how the helper will be used in a real world scenario.

steven-johnson (approved):
LGTM pending green

steven-johnson:
basically all good

Issue #5627, opened 2021-01-09
OpenGLCompute backend isn't being tested


steven-johnson:
In Issue #5475 we decided to remove OpenGL but keep OpenGLCompute; AFAICT, we don't do any testing at all on the OpenGLCompute backend. The only thing in our code that appears to explicitly test it is `apps/openglcompute`, which is Android-specific (and hasn't been built in ~years). If we intend to keep this backend, we need to add some level of testing for it on the buildbots.

shoaibkamil:
Agreed.  Would it be simplest to turn this on for Linux testing?  I realize those buildbots are pretty overtaxed, but I think testing for an additional backend isn't too bad (not like testing against an additional version of LLVM).

steven-johnson:
Agreed. I'll enable that today and see how badly broken we are (as I assume we must be if we haven't been testing).

steven-johnson:
Enabled it and I'm seeing LLVM errors of the form `"Duplicate option categories"` -- IIRC this is related to having multiple LLVM libraries in the open symbol space (i.e., one inside the driver, plus the one inside Halide)? 

(This is on a Linux box with Nvidia hw + drivers)

shoaibkamil:
Yes, that does indeed seem like the same issue.  I worked around it in Makefile builds by setting the environment variable `WITH_LLVM_INSIDE_SHARED_LIBHALIDE=` before building, but of course this is NOT a long-term solution and requires carefully setting `LD_LIBRARY_PATH`s.

steven-johnson:
See: https://github.com/halide/Halide/issues/4651

PR #6371, opened 2021-10-30
Support vectorized Select in OpenGLCompute backend

OmarEmaraDev:
The ternary operator in GLSL does not work with vector types. While the
mix function have overloads to boolean vectors, it is only supported in
version 4.5, so it is not exactly portable. To work around this, we use
the ternary operator on all elements of the vector type.

Necessary for #6348.

### DIFF
--- a/src/CodeGen_OpenGLCompute_Dev.cpp
+++ b/src/CodeGen_OpenGLCompute_Dev.cpp
@@ -615,11 +615,21 @@ void CodeGen_OpenGLCompute_C::visit(const Select *op) {
     string true_val = print_expr(op->true_value);
     string false_val = print_expr(op->false_value);
     string cond = print_expr(op->condition);
-    rhs << print_type(op->type)
-        << "(" << cond
-        << " ? " << true_val
-        << " : " << false_val
-        << ")";
+    if (op->type.is_scalar()) {
+        rhs << cond << " ? " << true_val << " : " << false_val;
+    } else {
+        rhs << print_type(op->type) << "(";
+        for (int i = 0; i < op->type.lanes(); i++) {
+            string index = "[" + std::to_string(i) + "]";
+            rhs << cond << index << " ? "
+                << true_val << index << " : "
+                << false_val << index;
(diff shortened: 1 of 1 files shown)


steven-johnson:
Is there a test case that exercises this? If not, this PR should add one.

OmarEmaraDev:
@steven-johnson Any pointers on how tests for OpenGLCompute should be added? I saw #5627, so I assumed that's is not possible at the moment.
I will be working on advancing and testing the OpenGLCompute backend in the near future, so if testing is something that needs to be addressed first, let me know how we can approach this.

steven-johnson:
> @steven-johnson Any pointers on how tests for OpenGLCompute should be added? I saw #5627, so I assumed that's is not possible at the moment.

Yikes -- I had forgotten we weren't testing these. Not sure if the issues ever got addressed. @shoaibkamil, do you remember the issues here?

shoaibkamil:
Yes, the issues are related to having two different LLVMs included: one within libhalide, and the other within the OpenGL runtime, both (I believe) having public symbol visibility.  @alexreinking had some ideas on how to make this work using cmake, I believe, and may have more info.

I can look into it again later this week, depending on what Alex thinks the right way to do this via cmake is.

alexreinking:
> Yes, the issues are related to having two different LLVMs included: one within libhalide, and the other within the OpenGL runtime, both (I believe) having public symbol visibility. @alexreinking had some ideas on how to make this work using cmake, I believe, and may have more info.

I thought this was solved (on macOS/Linux) by #5659? Is that not the case?

shoaibkamil:
> I thought this was solved (on macOS/Linux) by #5659? Is that not the case?

Thanks for the reminder, I'll check locally to ensure this is fixed and then try to get it working on the buildbot as well.

steven-johnson:
> @steven-johnson Any pointers on how tests for OpenGLCompute should be added?

OK, since OGLC testing is temporarily busted, go ahead and land this without a test, we'll figure it out later.

steven-johnson (approved):

PR #6372, opened 2021-10-30
Vectorize Ramp in OpenGLCompute backend

OmarEmaraDev:
Currently, ramps are generated as a number of independent scalar
expressions that are finally gathered into a vector. For instance,
indexing in vectorized code is filled with ramps like the following:

```
int _11 = int(1) * int(1);
int _12 = _10 + _11;
int _13 = int(2) * int(1);
int _14 = _10 + _13;
int _15 = int(3) * int(1);
int _16 = _10 + _15;
ivec4 _17 = ivec4(_10, _12, _14, _16);
```

This patch simplifies the generated code using a multiply add expression
on a vector containing an arithmetic sequence, such that the code is
as follows:

```
ivec4 _11 = ivec4(0, 1, 2, 3) * int(1) + _10;
```

This is more performant due to vectorization, more compact, and more
readable because the base and the stride are easily identifiable.

### DIFF
--- a/src/CodeGen_OpenGLCompute_Dev.cpp
+++ b/src/CodeGen_OpenGLCompute_Dev.cpp
@@ -550,20 +550,24 @@ void CodeGen_OpenGLCompute_C::visit(const For *loop) {
 }
 
 void CodeGen_OpenGLCompute_C::visit(const Ramp *op) {
-    ostringstream rhs;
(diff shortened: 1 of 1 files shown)


dsharletg (approved):

dsharletg:
Thanks, this makes sense and is a nice change.

Issue #6367, opened 2021-10-29
Unable to schedule two output buffers to be computed simultaneously when they are dependend.


mcourteaux:
I'm trying to boil my problem down to a minimal example. Here is my generator:

```cpp
class GeneratorBug : public Generator<GeneratorBug> {
   public:
    Input<Buffer<float>> input1{"input1", 2};
    Output<Buffer<float>> output1{"output1", 1};
    Output<Buffer<float>> output2{"output2", 1};

    Var k{"kernel"};

    Func intermediate{"intermediate"};

    void generate() {
        intermediate(k) = input1(k, 0) * input1(k, 1);
        output1(k) = intermediate(k) * intermediate(k);
        output2(k) = Halide::sqrt(output1(k) + output1(k));

        Expr num = input1.dim(0).extent();
        input1.dim(0).set_bounds(0, num);
        input1.dim(1).set_bounds(0, 2);
        output1.dim(0).set_bounds(0, num);
        output2.dim(0).set_bounds(0, num);
    }

    void schedule() {
        // clang-format off
        intermediate
            .vectorize(k, 8)
            .compute_at(output1, k)
            ;

        output1
            //.compute_root()
            .vectorize(k, 8)
            .compute_at(output2, k)
            //.compute_with(output2, k, LoopAlignStrategy::Auto)
            ;

        output2
            .compute_root()
            .vectorize(k, 8)
            //.compute_with(output1, k)
            ;
        // clang-format on
    }
};

HALIDE_REGISTER_GENERATOR(GeneratorBug, generator_bug)
```

Clearly, output2 depends on output1. I want everything to be processed in one for loop, with vectors of size 8. However, the statement file looks like this:

![image](https://user-images.githubusercontent.com/845012/139454420-6aeb45dd-4e15-4cd3-ba30-4c14ecc6874f.png)

Clearly the two outputs are produced separately. Uncommenting either of the `compute_with` directives causes an error when running the generator:

```
Unhandled exception: Error: Invalid compute_with: there is dependency between output1 and output2
```

So far, I haven't been able to come up with a workaround. I think the `output1.compute_at(output2, k)` doesn't work because it's an output buffer, which is implicitely like `.compute_root()` if I understand correctly. The `.compute_with()` kinda incorrectly assess that it's impossible to schedule these together as there is a dependency.

abadams:
Here's how to break the dependency to make compute_with legal:

```
        intermediate(k) = input1(k, 0) * input1(k, 1);
[...]

PR #6375, opened 2021-11-02
Fix for the crash from #6367

vksnk:
It would add an empty box for the function even if box_provided didn't return anything for this function, which would later cause a crash when we try to generate min/max lets for the given function.

### DIFF
--- a/src/BoundsInference.cpp
+++ b/src/BoundsInference.cpp
@@ -1133,7 +1133,9 @@ public:
                 auto boxes = boxes_provided(body, empty_scope, func_bounds);
                 for (const auto &fused : fused_group) {
                     string fused_stage_name = fused.first + ".s" + std::to_string(fused.second);
-                    boxes_for_fused_group[fused_stage_name] = boxes[fused.first];
+                    if (boxes.count(fused.first) > 0) {
+                        boxes_for_fused_group[fused_stage_name] = boxes[fused.first];
+                    }
                     for (const auto &fn : funcs) {
                         if (fn.name() == fused.first) {
                             stage_name_to_func[fused_stage_name] = fn;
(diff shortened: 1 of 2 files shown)


src/BoundsInference.cpp
```diff
@@ -1133,7 +1133,9 @@ class BoundsInference : public IRMutator {
                 auto boxes = boxes_provided(body, empty_scope, func_bounds);
                 for (const auto &fused : fused_group) {
                     string fused_stage_name = fused.first + ".s" + std::to_string(fused.second);
-                    boxes_for_fused_group[fused_stage_name] = boxes[fused.first];
+                    if (boxes.count(fused.first) > 0) {
```
abadams, line 1136: Doesn't this do a double-lookup? How about:
```
auto it = boxes.find(fused.first);
if (it != boxes.end()) {
  boxes_for_fused_group[fused_stage_name] = it->second;
}
```

test/correctness/compute_with.cpp
```diff
@@ -2110,6 +2110,93 @@ int rvar_bounds_test() {
     return 0;
 }
 
+int two_compute_at_test() {
+    ImageParam input1(Int(16), 2, "input1");
+    ;
```
abadams, line 2115: stray semicolon

test/correctness/compute_with.cpp
```diff
@@ -2110,6 +2110,93 @@ int rvar_bounds_test() {
     return 0;
 }
 
+int two_compute_at_test() {
```
abadams, line 2113: Worth commenting with the issue number that this test case came from, so that later readers have context

abadams (approved):

vksnk:
Test failures are not related, going to merge.

PR #6381, opened 2021-11-03
Remove halide_assert() from halide_default_device_wrap_native

steven-johnson:
This was inserted in https://github.com/halide/Halide/pull/6310, probably mistakenly, since `halide_assert()` in the Halide runtime is *not* a debug-only assertion). Instead of a controlled runtime failure, we just abort, which is not OK.


steven-johnson:
This may need to be backported to a Halide 13.x bugfix release, depending on whether #6310 was included in that release (@alexreinking)

alexreinking:
> This may need to be backported to a Halide 13.x bugfix release, depending on whether #6310 was included in that release (@alexreinking)

We're in luck! #6310 did _not_ make it in to Halide 13. :slightly_smiling_face:

steven-johnson:
Failures should be fixed by https://github.com/halide/Halide/pull/6386

PR #6382, opened 2021-11-03
Rename halide_assert -> halide_abort_if_false

steven-johnson:
A crashing bug got mistakenly inserted because a new contributor (reasonably) assumed that the `halide_assert()` macro in our runtime code was like a C `assert()` (i.e., something that would vanish in optimized builds).

This is not the case; it is a check that happens in all build modes and always triggers an `abort()` if it fires. We should remove any ambiguity about it, so this proposes to rename it to somethingmore like the Google/Abseil-style CHECK() macro, to make it stand out more.

(We may want to do a followup to verify that all of the uses really are unrecoverable errors that aren't better handled by returning an error.)

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1159,21 +1159,30 @@ void CodeGen_LLVM::optimize_module() {
 #endif
         pb.registerOptimizerLastEPCallback(
             [](ModulePassManager &mpm, OptimizationLevel level) {
+#if LLVM_VERSION >= 140
+                AddressSanitizerOptions asan_options;  // default values are good...
+                asan_options.UseAfterScope = true;     // ... except this one
+                mpm.addPass(createModuleToFunctionPassAdaptor(AddressSanitizerPass(asan_options)));
+#else
                 constexpr bool compile_kernel = false;
                 constexpr bool recover = false;
                 constexpr bool use_after_scope = true;
-#if LLVM_VERSION >= 140
-                // TODO: this is the value that the default ctor uses.
-                // Not sure if it's ideal for Halide.
-                constexpr AsanDetectStackUseAfterReturnMode use_after_return = AsanDetectStackUseAfterReturnMode::Runtime;
-                mpm.addPass(createModuleToFunctionPassAdaptor(AddressSanitizerPass(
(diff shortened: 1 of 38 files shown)


steven-johnson:
Failures unrelated (see https://github.com/halide/Halide/pull/6386)

dsharletg:
Should the original use that caused this problem be an abort even in debug mode? I think maybe it was just wrong to assert at all. In which case, the problem isn't that assert is active in non-debug builds at all.

In general I think the only reason to disable asserts in non-debug builds isn't because we should tolerate assert failures there, but we just don't want the performance hit due to assert overhead.

So, I question whether this PR and related PRs are needed at all?

abadams:
The name is still misleading though, because it evokes "assert", so even if no usage changes +1 to the rename.

steven-johnson:
> Should the original use that caused this problem be an abort even in debug mode? 

Nope. 

I think maybe it was just wrong to assert at all. In which case, the problem isn't that assert is active in non-debug builds at all.
> 
> In general I think the only reason to disable asserts in non-debug builds isn't because we should tolerate assert failures there, but we just don't want the performance hit due to assert overhead.
> 
> So, I question whether this PR and related PRs are needed at all?

steven-johnson:
> In general I think the only reason to disable asserts in non-debug builds isn't because we should tolerate assert failures there, but we just don't want the performance hit due to assert overhead.

I think it's both. An assertion failure / abort() in our runtime should be the absolutely last resort, and only for errors that are truly unrecoverable. The vast majority of existing `halide_assert()` usage looks to be more of the `assert()` usage variety (ie, documenting an invariant that doesn't even need checking at runtime unless in a special build for debugging).

src/CodeGen_LLVM.cpp
```diff
@@ -1159,21 +1159,30 @@ void CodeGen_LLVM::optimize_module() {
 #endif
         pb.registerOptimizerLastEPCallback(
             [](ModulePassManager &mpm, OptimizationLevel level) {
+#if LLVM_VERSION >= 140
```
abadams, line 1162: Not sure what's going on in this file?

abadams:
Oh you merged in a fix for trunk llvm branch. Got it.

abadams (approved):

Issue #6384, opened 2021-11-03
Uses of `halide_assert`/`HALIDE_CHECK` in runtime should be audited


steven-johnson:
See https://github.com/halide/Halide/pull/6382 and https://github.com/halide/Halide/pull/6381 -- the existing `halide_assert` macro in runtime_internal.h is not like the C `assert()` macro, in that `halide_assert()` is never compiled away, but lots of code in our runtime uses it in overly-generous ways that suggest that perhaps this wasn't clear to everyone.

#6382 proposes a mechanical rename to make this behavior more clear, but IMHO we need more work here:
- We should add something like `HALIDE_DEBUG_ASSERT()` to runtime_internal, which would be more like `assert()` in that it would be compiled away entirely unless `DEBUG_RUNTIME` is defined.
- All uses of `halide_assert()`/`HALIDE_CHECK()` should be audited:
    - Many (most?) should probably be converted to use `HALIDE_DEBUG_ASSERT()`, as they are clearly not firing in any reasonable failure mode (otherwise we'd have massive complaints about Halide-using apps aborting in odd conditions), and are just a waste of code space
    - Some should perhaps be converted to a runtime-check-and-return-error (if it is a recoverable error)
    - Some should remain as `HALIDE_CHECK()` (ie, errors which are unrecoverable and aborting is better than continuing)

mcourteaux:
I'd like to add that in my experience, I run into quite a lot of `halide_assert()`s, during execution of generators. These are most likely Halide bugs, so I'd like to suggest to NEVER compile them away when dealing with generators.

steven-johnson:
It would be quite extraordinary to have `halide_assert` (now renamed `halide_abort_if_false`) firing during Generator execution. Are you sure you don't mean `internal_assert` or `user_assert`?

mcourteaux:
You are right, I think! Was running into `internal_asserts()` a few times.

steven-johnson:
Note that the old `halide_assert()` was renamed to `halide_abort_if_false()`, but the underlying issue remains that I think we call it far too aggressively: it should really be reserved for truly unrecoverable errors, in which the host app should be taken down for its own safety. We should prefer to do the usual runtime error handling (i.e.: call `error()` and return a nonzero code) in the vast majority of cases, I suspect.

PR #6385, opened 2021-11-03
Add `halide_debug_assert()` macro (#6384)

steven-johnson:
Also convert usage of `halide_assert()`/`HALIDE_CHECK()` in hashmap.h and gpu_context_common.h to `halide_debug_assert()`, as all the usages looked to be appropriate for debug-mode only.

### DIFF
--- a/src/runtime/gpu_context_common.h
+++ b/src/runtime/gpu_context_common.h
@@ -66,7 +66,7 @@ public:
         }
         // This is a logic error that should never occur. It means the table is
         // full, but it should have been resized.
-        HALIDE_CHECK(nullptr, false);
+        HALIDE_DEBUG_ASSERT(nullptr, false);
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


zvookin:
Do we really need to litter the code long all caps names? We should use lots of asserts and checks and this approach is just horrible from a readability point of view. It draws massive attention to the *name* of the macro, which is just not at all interesting. It's like if every if statement was written "HEY_IM_AN_IF_STATEMENT_PAY_ATTENTION_TO_ME".

abadams (approved):

steven-johnson:
> Do we really need to litter the code long all caps names? 

For HALIDE_CHECK I think that SCREAMING_SNAKE_CASE is appropriate because it really should jump out at you. For debug-assert, though, probably not. would `halide_debug_assert()` be a better choice?

steven-johnson:
(This needs https://github.com/halide/Halide/pull/6382 to land first)

steven-johnson:
Closing in favor of identical-but-rebased https://github.com/halide/Halide/pull/6390

PR #6386, opened 2021-11-03
Fix for top-of-tree LLVM

steven-johnson:


src/CodeGen_LLVM.cpp
```diff
@@ -1159,21 +1159,27 @@ void CodeGen_LLVM::optimize_module() {
...
-                constexpr bool recover = false;
-                constexpr bool use_after_scope = true;
 #if LLVM_VERSION >= 140
-                // TODO: this is the value that the default ctor uses.
-                // Not sure if it's ideal for Halide.
-                constexpr AsanDetectStackUseAfterReturnMode use_after_return = AsanDetectStackUseAfterReturnMode::Runtime;
-                mpm.addPass(createModuleToFunctionPassAdaptor(AddressSanitizerPass(
-                    AddressSanitizerOptions{compile_kernel, recover, use_after_scope, use_after_return})));
+                AddressSanitizerOptions asan_options;  // default values are good...
+                asan_options.UseAfterScope = true;     // ... except this one
```
abadams, line 1164: I assume you're just basing it off the existing code, but do we know why UseAfterScope is true? Would be good to elaborate in the comment if so.

abadams (approved):

src/CodeGen_LLVM.cpp
steven-johnson: I do not know.

PR #6388, opened 2021-11-03
Annotate all runtime functions with HALIDE_MUST_USE_RESULT

steven-johnson:
All the halide runtime functions (and function pointers) that return interesting status results (int or bool) should be annotated with HALIDE_MUST_USE_RESULT (we have evidence that ignoring return codes has led to hard-to-diagnose errors for downstream users). Also update all the internal call sites to report and/or ignore errors as appropriate.

### DIFF
--- a/apps/lens_blur/process.cpp
+++ b/apps/lens_blur/process.cpp
@@ -21,7 +21,7 @@ int main(int argc, char **argv) {
     // Let the Halide runtime hold onto GPU allocations for
     // intermediates and reuse them instead of eagerly freeing
     // them. cuMemAlloc/cuMemFree is slower than the algorithm!
-    halide_reuse_device_allocations(nullptr, true);
+    (void)halide_reuse_device_allocations(nullptr, true);  // ignore errors
 
     Buffer<uint8_t> left_im = load_image(argv[1]);
     Buffer<uint8_t> right_im = load_image(argv[1]);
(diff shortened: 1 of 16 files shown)


abadams:
Things like halide_do_par_for can't actually return non-zero in these tests. They only return non-zero values if the user supplies a halide_error that returns. It's a bit weird to require checking a return value that's always zero for most users.

steven-johnson:
> They only return non-zero values if the user supplies a halide_error that returns.

~All interactive uses of Halide pretty much require this, so I'd argue it's not uncommon at all. (The default halide_error() that simply aborts is useful for learning and testing and not much else.)

dsharletg:
I think this will make it substantially more annoying for people learning, experimenting, or iterating on Halide code. 

Personally, I'm not a fan of this class of compiler enforced hygiene (`-Werror -Wunused-variable` being my biggest pet peeve along these lines). If this kind of thing could be made some kind of presubmit check, that would be a lot better IMO.

steven-johnson:
Based on comments received here and via other forums, I'm going to drop this PR rather than try to argue in favor of it more, because it seems unlikely that we'll get to consensus on it. 

For the record, I feel very strongly that this is the wrong decision, and that making it easy to ignore error conditions is a bad design decision that will come back to bite us in the future.

PR #6393, opened 2021-11-04
[hannk] Upgrade hannk to use TFLite 2.7.0 by default

steven-johnson:


steven-johnson:
(Also, a drive-by change to remove a couple of unused-var warnings)

dsharletg (approved):

PR #6396, opened 2021-11-08
Drop support for LLVM11

steven-johnson:
With Halide 13 released, we should drop support for LLVM11 in Halide trunk, since we only promise to support LLVM trunk + two releases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -840,15 +840,11 @@ RUNTIME_LL_COMPONENTS = \
   wasm_math \
   win32_math \
   x86 \
+  x86_amx \
   x86_avx \
   x86_avx2 \
   x86_avx512 \
   x86_sse41
-
-ifeq (,$(findstring $(LLVM_VERSION_TIMES_10), 110 111))
-# x86_amx.ll won't compile under LLVM11, but we don't need it there, either
-RUNTIME_LL_COMPONENTS += x86_amx
-endif
 
 RUNTIME_EXPORTED_INCLUDES = $(INCLUDE_DIR)/HalideRuntime.h \
                             $(INCLUDE_DIR)/HalideRuntimeD3D12Compute.h \
@@ -2151,7 +2147,7 @@ $(BUILD_DIR)/clang_ok:
 	@exit 1
 endif
 
-ifneq (,$(findstring $(LLVM_VERSION_TIMES_10), 110 111 120 130 140))
+ifneq (,$(findstring $(LLVM_VERSION_TIMES_10), 120 130 140))
 LLVM_OK=yes
 endif
 
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1101,10 +1101,8 @@ void CodeGen_LLVM::optimize_module() {
 
 #if LLVM_VERSION >= 130
     llvm::PassBuilder pb(tm.get(), pto);
-#elif LLVM_VERSION >= 120
-    llvm::PassBuilder pb(/*DebugLogging*/ false, tm.get(), pto);
 #else
-    llvm::PassBuilder pb(tm.get(), pto);
+    llvm::PassBuilder pb(/*DebugLogging*/ false, tm.get(), pto);
 #endif
 
     bool debug_pass_manager = false;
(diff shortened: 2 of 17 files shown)


steven-johnson:
attn @alinas

steven-johnson:
@alexreinking -- not sure about the packaging-for-Ubuntu workflow, I just updated to use LLVM12 instead, but not sure if that's kosher

src/CodeGen_LLVM.cpp
```diff
@@ -1182,20 +1173,9 @@ void CodeGen_LLVM::optimize_module() {
                     asan_options, use_global_gc,
                     use_odr_indicator, destructor_kind));
             });
-#elif LLVM_VERSION >= 120
```
abadams, line 1185: This looks like you deleted an llvm 12 block

test/correctness/simd_op_check.cpp
```diff
@@ -1746,13 +1746,8 @@ class SimdOpCheck : public SimdOpCheckTest {
                     check("v128.const", 8 * w, f32_1 * f32(42));
                     check("v128.const", 4 * w, f64_1 * f64(42));
                 } else {
-                    if (Halide::Internal::get_llvm_version() == 120) {
```
abadams, line 1749: This also looks like an llvm 12 block

abadams (approved):
lgtm except for two bits that confused me

src/CodeGen_LLVM.cpp
steven-johnson: No, I think it's just that the >=120 block was very similar to the else block here (just an extra arg to the lambda passed to registerPipelineStartEPCallback)

test/correctness/simd_op_check.cpp
steven-johnson: it is, but we are keeping it

alexreinking:
@steven-johnson - you need to update `11` to `12` here:

https://github.com/halide/Halide/blob/6071cf644b656b43a6de8b5d2d2d364d30678864/packaging/ubuntu/config.cmake#L63-L67

Makefile
```diff
@@ -840,16 +840,12 @@ RUNTIME_LL_COMPONENTS = \
...
   x86_avx \
   x86_avx2 \
   x86_avx512 \
   x86_sse41
 
-ifeq (,$(findstring $(LLVM_VERSION_TIMES_10), 110 111))
-# x86_amx.ll won't compile under LLVM11, but we don't need it there, either
-RUNTIME_LL_COMPONENTS += x86_amx
-endif
-
```
alexreinking, line 852: You might want to make a similar change here:

https://github.com/halide/Halide/blob/6071cf644b656b43a6de8b5d2d2d364d30678864/src/runtime/CMakeLists.txt#L116-L119

alexreinking: This is the only usage of `Halide_LLVM_VERSION`, which is defined here and I am tempted to remove...

https://github.com/halide/Halide/blob/6071cf644b656b43a6de8b5d2d2d364d30678864/dependencies/llvm/CMakeLists.txt#L23-L24

steven-johnson: > You might want to make a similar change here:
> 
> https://github.com/halide/Halide/blob/6071cf644b656b43a6de8b5d2d2d364d30678864/src/runtime/CMakeLists.txt#L116-L119

done

steven-johnson: > This is the only usage of `Halide_LLVM_VERSION`, which is defined here and I am tempted to remove...
> 
> https://github.com/halide/Halide/blob/6071cf644b656b43a6de8b5d2d2d364d30678864/dependencies/llvm/CMakeLists.txt#L23-L24

I predict we will eventually need it again. Maybe keep with a comment.

test/correctness/simd_op_check.cpp
abadams: But it's an == 120 block, not a >= 120 block, so by keeping it aren't we testing the wrong thing on llvm 13?

Or is the original check wrong?

abadams: bump. I'm confused about why this isn't going to cause simd op check failures on llvm 13+

Or were we only testing wasm on llvm 12 before?

steven-johnson:
> I'm confused about why this isn't going to cause simd op check failures on llvm 13+

We were only testing on llvm14 before, actually, but I still think you are misreading the diff.

I'll test this locally on 12/13/14 to verify, though.

steven-johnson:
> I'll test this locally on 12/13/14 to verify, though.

Passes for all three, so I think we are good.

test/correctness/simd_op_check.cpp
abadams: Old code:
```
                    if (Halide::Internal::get_llvm_version() == 120) {
                        check("i64x2.splat", 8 * w, u16_1 * u16(42));
                        check("i64x2.splat", 4 * w, u32_1 * u32(42));
                    } else {
                        check("i16x8.splat", 8 * w, u16_1 * u16(42));
                        check("i32x4.splat", 4 * w, u32_1 * u32(42));
                   }
```

New code:

```
                    check("i64x2.splat", 8 * w, u16_1 * u16(42));
                    check("i64x2.splat", 4 * w, u32_1 * u32(42));
```

I think the new code is just the llvm_version() == 120 version, so aren't we changing the expected output for llvm 13/14?

I can't figure out what I'm missing.

steven-johnson:
> I can't figure out what I'm missing.

You're missing the fact that the >= 130 case is handled above -- the section you're looking at is in the else case.

abadams:
Ah! Everything is now clear.

PR #6401, opened 2021-11-09
Add defensive checks to halide_buffer_copy_already_locked

steven-johnson:
Found while debugging crashes with performance_async_gpu for OpenGLCompute: the 'if' tree wasn't robust enough for malformed buffers being passed, and could attempt to deref and use a null src->device_interface or dst->device_interface in some cases.

This patch just improves this function to return an error in these cases (rather than crashing); the fact that we are getting malformed buffers passed to us is likely a separate bug.


abadams (approved):

shoaibkamil (approved):

steven-johnson:
Failure is unrelated.

PR #6402, opened 2021-11-09
_halide_buffer_crop() needs to check for runtime failures

steven-johnson:
We currently assume that _halide_buffer_crop() will never fail. This is a bad assumption, as it can call device_crop(), which can fail due to unexpected runtime errors, or from a backend simply leaving the `device_crop` field at the default (unimplemented) case (as is currently the case for the OGLC backend).

When this happens, the `dst` buffer was left in an inconsistent, invalid state (which was what led to the crashes fixed by #6401).

This change modifies _halide_buffer_crop() to return nullptr in the event of an error, and to wrap all callers in a `require()` clause to check the result. (This is not optimal, of course, since the specific error returned by device_crop is getting dropped on the floor, but the existence of an error is no longer ignored.)

This addresses at least some of the failure issues we are seeing in performance_async_gpu with the OpenGLCompute backend.

(Also: drive-by whitespace fix in CodegenC)


src/StorageFolding.cpp
```diff
@@ -103,7 +103,10 @@ class FoldStorageOfFunction : public IRMutator {
                 Expr new_mins = Call::make(type_of<int *>(), Call::make_struct, mins, Call::Intrinsic);
                 vector<Expr> new_args = op->args;
                 new_args[3] = new_mins;
-                expr = Call::make(op->type, op->name, new_args, op->call_type);
+                internal_assert(op->type == type_of<halide_buffer_t *>() &&
```
abadams, line 106: This mutates an existing buffer_crop call, so isn't this going to redundantly wrap it in another require?

steven-johnson: hmm, yes, I suppose it could -- let me see if it will get optimized away or not

steven-johnson: The gotcha here is that this code transmutes `_halide_buffer_crop()` -> `_halide_buffer_set_bounds(_halide_buffer_crop())`; if we leave out the `require()`, we can risk passing nullptr to the enclosing call. I suppose we could just make that call check for an incoming nullptr and bail early...

abadams (commented):
The other runtime calls are checked using assertstmts after the lets, not by wrapping a require. For consistency, I think it would be better to add a non-null assertstmts for each entry in cropped_buffers to the pre_call stmt in the loop on line ~790 (edit: in ScheduleFunctions.cpp)

steven-johnson:
> The other runtime calls are checked using assertstmts after the lets, not by wrapping a require. 

Not so: look at the code you just commented on, `StorageFolding.cpp:132`

abadams:
Fair enough, I guess both idioms are in use.

steven-johnson:
> The other runtime calls are checked using assertstmts after the lets, not by wrapping a require. For consistency, I think it would be better to add a non-null assertstmts for each entry in cropped_buffers to the pre_call stmt in the loop on line ~790 (edit: in ScheduleFunctions.cpp)

Please see https://github.com/halide/Halide/pull/6403

steven-johnson:
Closing in favor of #6403

PR #6405, opened 2021-11-10
Fix obscure bug in widening let substitution

abadams:


shoaibkamil (approved):
Wow, subtle.  Thanks for the quick fix!

abadams:
Failure was unrelated

PR #6412, opened 2021-11-12
[hannk] Pacify clang-tidy

steven-johnson:


apps/hannk/interpreter/model.h
```diff
@@ -306,7 +306,12 @@ class Op {
     // by making the virtual dispatches private, and add inline wrappers that
     // enforce a sane calling convention.
     static inline OpPtr mutate(OpPtr op, OpMutator *m) {
-        return op->mutate_impl(m, std::move(op));
+        // clang-tidy will complain if we just do `op->mutate_impl(m, move(op))`
+        // because the order of evaluation between the move and the invocation is
+        // undefined; while that's true, we know that op will remain valid throughout
```
dsharletg, line 311: I think this might be more than clang-tidy just being fussy, I think it really is sketchy. But the code looks fine.

dsharletg (approved):

apps/hannk/interpreter/model.h
steven-johnson: I can always go revisit this an implement some other sort of dispatch approach. (Or scrap this mess entirely...) But as a practical matter, it's not clear to me how the compiler could generate misbehaving code here: the unique_ptr<Op> needs to keep the pointer alive and valid throughout the life of the call. Since UB makes all things possible, sure, the compiler could legally do anything here, I guess...

dsharletg: I agree the code as written now (after this PR) is safe.

steven-johnson: Just pushed a revision that should be genuinely safe, PTAL

PR #6423, opened 2021-11-17
Handle DWARF debug-abbreviation tables with children lists

steven-johnson:
We were (incorrectly) terminating at the end of the first list-of-children, which led to incomplete tables and thus assertion failures for some tooling setups.

### DIFF
--- a/src/Introspection.cpp
+++ b/src/Introspection.cpp
@@ -17,6 +17,7 @@
 
 #include <cstdio>
 #include <iostream>
+#include <map>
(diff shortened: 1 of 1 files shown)


abadams:
Test failures look real. Some sort of assumption isn't holding.

steven-johnson:
Yeah, there's still something off here, investigating

steven-johnson:
Yeah, this is misguided, closing, look for a different PR

PR #6424, opened 2021-11-17
Fix Introspection issues

steven-johnson:
- DWARF v5 has a slightly different header; this recognizes it so we don't fail immediately
- Add support for the line_strp form
- Allow for a graceful failure if a debug abbreviation is missing; I've only seen this when compiling for TSAN, and I'm honestly not entirely sure if this is a bug in the DWARF generation for those tools vs a subtle flaw in our parsing, but bailing out early and skipping introspection seems kinder than assert-fail.

### DIFF
--- a/src/Introspection.cpp
+++ b/src/Introspection.cpp
@@ -17,6 +17,7 @@
 
 #include <cstdio>
 #include <iostream>
+#include <map>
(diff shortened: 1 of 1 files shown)


abadams (approved):
Makes sense to me.

Issue #23, opened 2012-08-22
Build should be able to bootstrap missing OCaml libraries


jrk:
One of the things which makes the build difficult for first-timers is the need to install OCaml libraries on which we depend. On some platforms this is easy (modern Ubuntu tends to have many up-to-date OCaml packages in apt), while on others it involves a long chain of manual download/configure/make/installs which is an unnecessary distraction.

The build bootstrap process should have functionality which automatically fetches, builds, and installs the relevant libraries in a project-local path, and makes these discoverable to all subsequent build steps. [odb](https://github.com/thelema/odb) is a straightforward option.

The challenge is to do this while also using the system packages when they exist and are sufficient, to avoid too much bloat.

PR #6426, opened 2021-11-17
Rewrite integer lerp using intrinsics

vksnk:
Should be pretty much the same thing as before, but easier to read and maybe may produce better code.

### DIFF
--- a/src/Lerp.cpp
+++ b/src/Lerp.cpp
@@ -134,17 +134,9 @@ Expr lower_lerp(Expr zero_val, Expr one_val, const Expr &weight) {
             case 8:
             case 16:
             case 32: {
-                Expr zero_expand = Cast::make(UInt(2 * bits, computation_type.lanes()),
-                                              zero_val);
-                Expr one_expand = Cast::make(UInt(2 * bits, one_val.type().lanes()),
-                                             one_val);
-
-                Expr rounding = Cast::make(UInt(2 * bits), 1) << Cast::make(UInt(2 * bits), (bits - 1));
-                Expr divisor = Cast::make(UInt(2 * bits), 1) << Cast::make(UInt(2 * bits), bits);
-
-                Expr prod_sum = zero_expand * inverse_typed_weight +
-                                one_expand * typed_weight + rounding;
-                Expr divided = ((prod_sum / divisor) + prod_sum) / divisor;
+                Expr shift = Cast::make(UInt(2 * bits), bits);
+                Expr prod_sum = widening_mul(zero_val, inverse_typed_weight) + widening_mul(one_val, typed_weight);
+                Expr divided = rounding_shift_right(rounding_shift_right(prod_sum, shift) + prod_sum, shift);
 
                 result = Cast::make(UInt(bits, computation_type.lanes()), divided);
                 break;
--- a/src/Lerp.cpp
+++ b/src/Lerp.cpp
@@ -136,6 +136,8 @@ Expr lower_lerp(Expr zero_val, Expr one_val, const Expr &weight) {
             case 32: {
                 Expr shift = Cast::make(UInt(2 * bits), bits);
                 Expr prod_sum = widening_mul(zero_val, inverse_typed_weight) + widening_mul(one_val, typed_weight);
+                // Computes x / (2 ** N - 1) as (x / 2 ** N + x) / 2 ** N.
+                // TODO: on x86 it's actually one instruction cheaper to do the division directly.
                 Expr divided = rounding_shift_right(rounding_shift_right(prod_sum, shift) + prod_sum, shift);
 
                 result = Cast::make(UInt(bits, computation_type.lanes()), divided);


src/Lerp.cpp
```diff
@@ -134,17 +134,9 @@ Expr lower_lerp(Expr zero_val, Expr one_val, const Expr &weight) {
...
-
-                Expr rounding = Cast::make(UInt(2 * bits), 1) << Cast::make(UInt(2 * bits), (bits - 1));
-                Expr divisor = Cast::make(UInt(2 * bits), 1) << Cast::make(UInt(2 * bits), bits);
-
-                Expr prod_sum = zero_expand * inverse_typed_weight +
-                                one_expand * typed_weight + rounding;
-                Expr divided = ((prod_sum / divisor) + prod_sum) / divisor;
+                Expr shift = Cast::make(UInt(2 * bits), bits);
+                Expr prod_sum = widening_mul(zero_val, inverse_typed_weight) + widening_mul(one_val, typed_weight);
+                Expr divided = rounding_shift_right(rounding_shift_right(prod_sum, shift) + prod_sum, shift);
```
abadams, line 139: This is equivalent to the original code, but I'm not convinced it's a good idea on x86. I bet you can do better with an average-round-up instruction.

abadams:
I might spend some time trying to figure out how to do that division by 2^n - 1 more cheaply, but that would be a follow-up PR.

abadams (approved):

src/Lerp.cpp
dsharletg, line 139: This looks odd. I agree it should be the same as the old code, but I think the inner rounding_shift_right expression will be negligible. @zvookin @abadams any idea what this was about? Is it maybe attempting to ensure we actually hit one_val when typed_weight is the max?

abadams: It's a trick to do x/255 as (x/256 + x)/256

dsharletg: OK yeah, division by 2^n-1...

abadams: Note that this is a special case of one of our unsigned division lowering methods. I was about to suggest just dividing and relying on that, but that only covers up to 255

abadams: Actually I could easily extend that table to include 65535 and 2^32 - 1 ...

vksnk: Ah, that's a cool trick, I've been wondering about that too.

abadams: Hrm, no. This method has the benefit that it doesn't require additional widening.

abadams: It's not equal! Testing it, I get the following identity instead:

`x/255 == (x/256 + x + 1) / 256`

abadams: On x86, (x + 127)/255 (which is what I think it should be including rounding), is:

```
	vpaddw	(%rax,%rsi,2), %ymm0, %ymm2
	vpmulhuw	%ymm1, %ymm2, %ymm2
	vpsrlw	$7, %ymm2, %ymm2

```

and ((x + 127)/256 + x + 128) / 256, which is equal, gives:

```
	vpaddw	%ymm0, %ymm2, %ymm3
	vpsrlw	$8, %ymm3, %ymm3
	vpaddw	%ymm3, %ymm2, %ymm2
	vpaddw	%ymm1, %ymm2, %ymm2
	vpsrlw	$8, %ymm2, %ymm2

```

steven-johnson: > It's a trick to do x/255 as (x/256 + x)/256

We should document that trick in a comment.

abadams: So just change that code to something like:
```
Expr divisor = cast(UInt(2 * bits), (1ULL << bits) - 1);
Expr divided = (prod_sum + divisor/2) / divisor;
```

abadams: Oh nuts, on arm it's reversed:
```
	add	v3.8h, v3.8h, v0.8h
	umull	v4.4s, v3.4h, v1.4h
	umull2	v3.4s, v3.8h, v2.8h
	ushr	v4.4s, v4.4s, #23
	ushr	v3.4s, v3.4s, #23
	xtn	v4.4h, v4.4s
	xtn2	v4.8h, v3.4s
	xtn	v3.8b, v4.8h

```

vs

```
	add	v3.8h, v2.8h, v0.8h
	usra	v2.8h, v3.8h, #8
	addhn	v2.8b, v2.8h, v1.8h
```

abadams: btw the nasty thing about this trick is that it fails for the largest uint16_t, so you have to have some way to know that's not possible (as we do here).

abadams: On the other hand, ```x/255 == ((x + 1)/256 + x + 1)/256``` seems to hold for all uint16 x.

So maybe the resolution here is teaching the arm backend this trick for division by 255

steven-johnson: Do any of us have a copy of the book "Hacker's Delight"? IIRC it has quite a long section on this sort of technique, but I last read it at a different employer ~20 years ago...

abadams:
My overall conclusion is that we should merge this PR with a comment explaining the trick and a TODO that notes it's actually one instruction cheaper to do the division directly on x86.

src/Lerp.cpp
abadams: Here's a useful identity for uint16s:

``` x / 255 === let y = saturating_sub(x, 127) in rounding_shift_right(rounding_shift_right(y, 8) + y, 8)```

On ARM we currently lower x/255 as:

```
	umull	v3.4s, v2.4h, v0.4h
	umull2	v2.4s, v2.8h, v1.8h
	ushr	v3.4s, v3.4s, #23
	ushr	v2.4s, v2.4s, #23
	xtn	v3.4h, v3.4s
	xtn2	v3.8h, v2.4s
	xtn	v2.8b, v3.8h
```

but using this identity we get:

```
	uqsub	v1.8h, v1.8h, v0.8h
	urshr	v2.8h, v1.8h, #8
	raddhn	v1.8b, v2.8h, v1.8h
```

Nice.

**Comment by steven-johnson** [comment]
Unrelated, but `correctness_custom_cuda_context` has been failing on Windows for a while now; is anyone looking into it?

**Comment by vksnk** [comment]
I am going to submit this (tests failure is unrelated) and create an issue to track if there are any better ways to do x / 255 (so discussion in the comments is not lost).

**Comment by abadams** [comment]
This actually hurt lerp on x86, because rounding shift right is more expensive than (x + 128) / 256.

Before:
```
	vpaddw	%ymm0, %ymm1, %ymm1
	vpsrlw	$8, %ymm1, %ymm2
	vpaddw	%ymm2, %ymm1, %ymm1
	vpsrlw	$8, %ymm1, %ymm1
	vextracti128	$1, %ymm1, %xmm2
	vpackuswb	%xmm2, %xmm1, %xmm1
```

After:
```
	vpsrlw	$7, %ymm2, %ymm3
	vpand	%ymm0, %ymm3, %ymm3
	vpsrlw	$8, %ymm2, %ymm4
	vpaddw	%ymm2, %ymm4, %ymm2
	vpaddw	%ymm3, %ymm2, %ymm2
	vpsrlw	$7, %ymm2, %ymm3
	vpand	%ymm0, %ymm3, %ymm3
	vpsrlw	$8, %ymm2, %ymm2
	vpaddw	%ymm2, %ymm3, %ymm2
	vpand	%ymm1, %ymm2, %ymm2
	vextracti128	$1, %ymm2, %xmm3
	vpackuswb	%xmm3, %xmm2, %xmm2
```

I'm changing the lowering on x86 anyway, but are there other targets that don't have a rounding_shift_right instruction?

src/Lerp.cpp
alexreinking: > Do any of us have a copy of the book "Hacker's Delight"? IIRC it has quite a long section on this sort of technique, but I last read it at a different employer ~20 years ago...

@steven-johnson - very late, I know, but I keep a copy on my desk. If it ever comes up again that we want to consult this book, let me know.

PR #6429, opened 2021-11-19
Support using V8 as the Wasm JIT interpreter

ngzhian:
This is a partial revert of https://github.com/halide/Halide/pull/5097.
It brings back a bunch of the code in WasmExecutor to set up and use V8
to run Wasm code. All of the code is copy-pasted. There are some small
cleanups to move common code (like BDMalloc, structs, asserts) to a
common area guarded by `if WITH_WABT || WITH_V8`.

Enabling V8 requires setting 2 CMake options:
- V8_INCLUDE_PATH
- V8_LIB_PATH

The first is a path to v8 include folder, to find headers, the second is
the monolithic v8 library. This is because it's pretty difficult to
build v8, and there are various flags you can set. Comments around those
options provide some instructions for building v8.

By default, we still use the wabt for running Wasm code, but we can use
V8 by setting WITH_WABT=OFF WITH_V8=ON. Maybe in the future, with more
testing, we can flip this.

Right now this requires a locally patched build of V8 due to
https://crbug.com/v8/10461, but once that is resolved, the version of V8
that includes the fix will be fine.

Also enable a single test, block_transpose, to run on V8, with these
results:

$ HL_JIT_TARGET=wasm-32-wasmrt-wasm_simd128 \
  ./test/performance/performance_block_transpose
Dummy Func version: Scalar transpose bandwidth 3.45061e+08 byte/s.
Wrapper version: Scalar transpose bandwidth 3.38931e+08 byte/s.
Dummy Func version: Transpose vectorized in y bandwidth 6.74143e+08
byte/s.
Wrapper version: Transpose vectorized in y bandwidth 3.54331e+08 byte/s.
Dummy Func version: Transpose vectorized in x bandwidth 3.50053e+08
byte/s.
Wrapper version: Transpose vectorized in x bandwidth 6.73421e+08 byte/s.
Success!

For comparison, when targeting host:

$ ./test/performance/performance_block_transpose
Dummy Func version: Scalar transpose bandwidth 1.33689e+09 byte/s.
Wrapper version: Scalar transpose bandwidth 1.33583e+09 byte/s.
Dummy Func version: Transpose vectorized in y bandwidth 2.20278e+09
byte/s.
Wrapper version: Transpose vectorized in y bandwidth 1.45959e+09 byte/s.
Dummy Func version: Transpose vectorized in x bandwidth 1.45921e+09
byte/s.
Wrapper version: Transpose vectorized in x bandwidth 2.21746e+09 byte/s.
Success!

For comparison, running with wabt:

Dummy Func version: Scalar transpose bandwidth 828715 byte/s.
Wrapper version: Scalar transpose bandwidth 826204 byte/s.
Dummy Func version: Transpose vectorized in y bandwidth 1.12008e+06
byte/s.
Wrapper version: Transpose vectorized in y bandwidth 874958 byte/s.
Dummy Func version: Transpose vectorized in x bandwidth 879031 byte/s.
Wrapper version: Transpose vectorized in x bandwidth 1.10525e+06 byte/s.
Success!


ngzhian:
@steven-johnson fyi, this is probably not ready to merge, given that there is no version of V8 that can work yet (until https://crbug.com/v8/10461 is resolved), but wanted to upload for you to take a look first.

steven-johnson:
This looks generally promising. Agreed that it's not ready to merge until the bug above is fixed. A few general comments:
- Need to (eventually) add documentation about how to configure and build V8 for this purpose (probably in README_webassembly); my recollection is that there are a number of ways to configure V8 builds, and the necessary settings aren't that obvious (and vary by platform).
- Speaking of platforms: it's OK to target just x86-64-Linux (though OSX support would be nice as well). I wouldn't spend any time getting Windows support to work unless it's really easy.
- It's regrettable that there is so much code duplication between WABT and V8 glue code; there's probably no easy fix here. That said, it would be nice to make a second pass at some point and verify that we are doing what is reasonable to minimize the duplication. (I gather that there is/was a proposal for a Common Embedding API for wasm runtimes but the last time I checked, ~2 years ago, it wasn't well-supported enough by either WABT or V8 to entertain using it.)
- Nit: you need to run the `run-clang-format.sh` script to format the code correctly.

ngzhian:
> * Need to (eventually) add documentation about how to configure and build V8 for this purpose (probably in README_webassembly); my recollection is that there are a number of ways to configure V8 builds, and the necessary settings aren't that obvious (and vary by platform).

Woops, forgot about this, added some documentation to README_webassembly, it also contains pointers to v8.dev, which should be the most accurate one.

> * Speaking of platforms: it's OK to target just x86-64-Linux (though OSX support would be nice as well). I wouldn't spend any time getting Windows support to work unless it's really easy.

Added a note about OS support too.

> * It's regrettable that there is so much code duplication between WABT and V8 glue code; there's probably no easy fix here. That said, it would be nice to make a second pass at some point and verify that we are doing what is reasonable to minimize the duplication. (I gather that there is/was a proposal for a Common Embedding API for wasm runtimes but the last time I checked, ~2 years ago, it wasn't well-supported enough by either WABT or V8 to entertain using it.)

You might be thinking of this https://github.com/WebAssembly/wasm-c-api there hasn't been much progress there. V8 does support that api, but it [doesn't work well](https://bugs.chromium.org/p/v8/issues/detail?id=10461#c3) if you need to work with the rest of the V8 C++ API.

> * Nit: you need to run the `run-clang-format.sh` script to format the code correctly.

Thanks, done!

ngzhian:
https://bugs.chromium.org/p/v8/issues/detail?id=10461#c10 V8 9.8.76 has this change.

steven-johnson:
> https://bugs.chromium.org/p/v8/issues/detail?id=10461#c10 V8 9.8.76 has the compile API we need.

Cool! Let me know when this is ready for a re-review.

ngzhian:
> > https://bugs.chromium.org/p/v8/issues/detail?id=10461#c10 V8 9.8.76 has the compile API we need.
> 
> Cool! Let me know when this is ready for a re-review.

This is ready, ptal.

steven-johnson:
> This is ready, ptal.

OK, I will get to it sometime this week.

steven-johnson:
Hey, I still haven't gotten to this (various crises + required long meetings, etc), but we also need to touch base on issues of integrating this inside Google; in the past, there have been versioning issues (the version(s) of V8 we needed weren't available on a timely basis in google3). Let's discuss this via Chat (rather than here on GitHub) etc at your convenience to be sure this is going to be tractable there as well before moving forward.

src/WasmExecutor.cpp
```diff
@@ -46,7 +54,16 @@ namespace Internal {
 // receive the result value.
 static const char kTrampolineSuffix[] = "_trampoline";
 
-#if WITH_WABT
+#ifdef WITH_V8
+using namespace v8;
+
+#define V8_API_VERSION ((V8_MAJOR_VERSION * 10) + V8_MINOR_VERSION)
+
+static_assert(V8_API_VERSION >= 98,
```
ngzhian, line 62: @steven-johnson required 9.8

steven-johnson (changes requested):
Aside from the documentation nits below this generally looks good. There are some new failures in the tests that need addressing before we can land this, however:

```
$ ctest -L correctness -j72
... much output ...
The following tests FAILED:
         53 - correctness_bounds_of_split (ILLEGAL)
        107 - correctness_exception (Child aborted)
        110 - correctness_extern_consumer (Child aborted)
        112 - correctness_extern_error (Child aborted)
        366 - correctness_vector_reductions (ILLEGAL)
```

These all pass under WABT -- that said, it may well be that these are things that got 'fixed' after V8 was deprecated and WABT added, not sure.

steven-johnson:
See https://github.com/halide/Halide/pull/6478 for a finished PR -- this one can be closed.

PR #6434, opened 2021-11-19
Sketch of AOT-compiled performance tests

steven-johnson:
This is a sketch of a clunky-but-effective way to make some (simple) tests AOT-compiled without having to greatly increase number of source files or add lots of complex build rules. Not sure if it's a win or not -- opinions wanted. Basically, this puts a Generator and a 'main' test harness into the same file, separated by an ifdef; the CMake build rules know how to build the file multiple ways and connect them together.

I'm honestly torn as to whether the icky, please-don't-use-this-in-real-code approach here is outweighed by the convenience factor of having self-contained tests that will AOT-compile properly.

(BTW, this approach produces results that will build-and-run just fine under Wasm! Although the one example I've provided in this PR doesn't do so usefully, because it requires GPU support...)

Also note that I haven't provided build rules for Make (only CMake), and I don't intend to do so, because doing this in Make is a big mess of ugly special cases. Someone else can add new features to Make if they like.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -34,6 +34,11 @@ if (WITH_TEST_PERFORMANCE)
     add_subdirectory(performance)
 endif ()
 
+option(WITH_TEST_PERFORMANCE_AOT "Build performance_aot tests" ON)
+if (WITH_TEST_PERFORMANCE_AOT)
+    add_subdirectory(performance_aot)
+endif ()
+
 option(WITH_TEST_GENERATOR "Build generator tests" ON)
 if (WITH_TEST_GENERATOR)
     add_subdirectory(generator)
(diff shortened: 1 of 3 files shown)


abadams:
Why support both AOT and JIT inside a single test as opposed to porting that test to AOT?

abadams:
Oh I misunderstood what was happening. Disregard previous comment.

abadams:
On the code side, this seems like it just saves having two files. I'm not sure that's worth the ick vs just having the two files open in your editor at the same time.

FWIW I'd support porting the performance tests to AOT and just eating the extra complexity, because it's much easier to profile AOT code with tools like perf.

shoaibkamil:
Thanks @steven-johnson, I was still working out the CMake changes required.  I agree with Andrew: we should just set it up so it's like `test/generators` with one source file for the generator and one for the test.

I can finish whipping up an example :)

steven-johnson:
I'm going to drop this since @shoaibkamil has a better plan :-)

PR #6441, opened 2021-11-22
Avoid needless gather in fast_integer_divide lowering

abadams:
fast_integer_divide did two lookups, one for a multiplier, and one for a
shift. It turns out you can just use count leading zeros to compute a
workable shift instead of having to do a lookup. This PR speeds up use
of fast_integer_divide in cases where the denominator varies across
vector lanes by ~70% or so by avoiding one of the two expensive gathers.


src/FastIntegerDivide.cpp
```diff
@@ -10,17 +10,28 @@ using namespace Halide::Internal::IntegerDivision;
...
+    internal_assert(d.type().element_of() == UInt(8));
+    return 7 - count_leading_zeros(d - 1);
+}
+
 Buffer<uint8_t> integer_divide_table_u8() {
     static std::mutex initialize_lock;
     std::lock_guard<std::mutex> lock_guard(initialize_lock);
     {
-        static Buffer<uint8_t> im(256, 2);
+        static Buffer<uint8_t> im(256);
```
steven-johnson, line 26: Unrelated: since C++11 onwards default to promising thread-safe initialization of statics, we could rewrite this to be a bit terser like so:

```
Buffer<uint8_t> integer_divide_table_u8() {
    static Buffer<uint8_t> im(256) = []() {
        Buffer<uint8_t> im(256);
        for (uint32_t i = 0; i < 256; i++) {
            im(i) = table_runtime_u8[i][2];
            if (i > 1) {
                internal_assert(table_runtime_u8[i][3] == shift_for_denominator(i));
            }
        }
        return im;
    }();

    return im;
}
```

steven-johnson (approved):

src/FastIntegerDivide.cpp
abadams: Good point. Will fix.

shoaibkamil (approved):

abadams:
No, that's compile-time code.

PR #6442, opened 2021-11-23
Codegen_C: buffer compilation needs to special-case scalar buffers

steven-johnson:
The existing code will emit something like `halide_dimension_t foo_buffer_shape[] = {};` for these, which is a zero-length array, which some compilers will (justifiably) say has no effect. We should be able to just use nullptr for the shape in these cases.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -1957,17 +1957,23 @@ void CodeGen_C::compile(const Buffer<> &buffer) {
         stream << "\n};\n";
     }
 
-    // Emit the shape (constant even for scalar buffers)
-    stream << "static const halide_dimension_t " << name << "_buffer_shape[] = {";
-    for (int i = 0; i < buffer.dimensions(); i++) {
-        stream << "halide_dimension_t(" << buffer.dim(i).min()
-               << ", " << buffer.dim(i).extent()
-               << ", " << buffer.dim(i).stride() << ")";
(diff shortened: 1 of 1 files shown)


src/CodeGen_C.cpp
```diff
@@ -1957,17 +1957,23 @@ void CodeGen_C::compile(const Buffer<> &buffer) {
...
+    if (buffer.dimensions()) {
+        // Emit the shape -- note that we can't use this for scalar buffers because
+        // we'd emit a statement of the form "foo_buffer_shape[] = {}", and a zero-length
+        // array will make some compilers unhappy.
+        stream << "static const halide_dimension_t " << name << "_buffer_shape[] = {";
+        for (int i = 0; i < buffer.dimensions(); i++) {
+            stream << "halide_dimension_t(" << buffer.dim(i).min()
+                   << ", " << buffer.dim(i).extent()
+                   << ", " << buffer.dim(i).stride() << ")";
+            if (i < buffer.dimensions() - 1) {
```
dsharletg, line 1970: An alternative (maybe simpler) solution is to just emit a dummy default constructed halide_dimension_t, i.e. {{}}.

dsharletg (approved):

src/CodeGen_C.cpp
steven-johnson: True, but then we need to allocate a (trivial) amount of extra space for dimensions that shouldn't ever be used.

PR #6443, opened 2021-11-23
Add operator<< for Closure

steven-johnson:
Moves the ad-hoc implementation our of HostClosure::arguments() for easier debugging usage. Also, drive-by elimination of the body of HostClosure ctor, which was identical to the one inherited from Closure.


zvookin (approved):

steven-johnson:
Something about this has actually broken OpenCL code, not sure why -- maybe the HostClosure change, somehow?

steven-johnson:
This "works" but should not be landed until we understand how/why the HostClosure change broke something.

steven-johnson:
The failure turned out to be don't-call-virtual-methods-from-a-ctor. Added an explanatory comment so that people won't make the same mistake I did.

Issue #6410, opened 2021-11-12
How to output NVPTX assembly/IR/bytecode?


mcourteaux:
I'm looking to find a way to inspect the NVPTX generated code for my pipelines. The statement files only contain calls to `halide_cuda_run()` to launch kernels. I am looking for the code itself. I found so far that if you use `c_source` as generator output, it will produce a very long C-file in which the code is hiding somewhere. However this fails in case my CPU scheduling part of the pipeline contains stuff that is not supported by the C-backend (like predicated load). I think that there is an output type missing for generators to just output the CUDA kernel assembly.

An example of kernel code hiding in the C-source is this:

```cpp
static const char *_cuda_gpu_source_kernels_string = R"BUFCHARSOURCE(//
// Generated by LLVM NVPTX Back-End
//

.version 3.2
.target sm_20
.address_size 64

	// .globl	_kernel_logit_kernel_offset_s0_kernel_ko___block_id_x // -- Begin function _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x
                                        // @_kernel_logit_kernel_offset_s0_kernel_ko___block_id_x
.visible .entry _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x(
	.param .u64 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_0,
	.param .u64 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_1,
	.param .u64 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_2,
	.param .u32 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_3,
	.param .u32 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_4,
	.param .u32 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_5,
	.param .u32 _kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_6
)
{
	.reg .pred 	%p<6>;
	.reg .f32 	%f<43>;
	.reg .b32 	%r<33>;
	.reg .b64 	%rd<21>;

// %bb.0:                               // %entry
	ld.param.u64 	%rd5, [_kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_0];
	ld.param.u64 	%rd6, [_kernel_logit_kernel_offset_s0_kernel_ko___block_id_x_param_2];
[...]

PR #6444, opened 2021-11-24
Include GPU source kernels in Stmt and StmtHtml file.

mcourteaux:
This works for me nicely with CUDA, OpenCL, OpenGLCompute, Direct3D12Compute, and Metal as backends. I tried collapsing the code for all of them and it works nicely. I put everything in a `<pre>` for the HTML output. Not sure if any of the backends produce non-compatible characters that might break the HTML, but for my kernels, it didn't.

### DIFF
--- a/src/IRPrinter.cpp
+++ b/src/IRPrinter.cpp
@@ -8,6 +8,7 @@
 #include "IROperator.h"
 #include "Module.h"
 #include "Target.h"
+#include "Util.h"
 
 namespace Halide {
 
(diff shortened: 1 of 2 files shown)


mcourteaux:
This mostly fixes #6410.

abadams:
Thanks! To me this seems like a reasonably elegant solution. It might be better if we tracked embedded data buffers and embedded kernel source buffers in separate vectors in the Module class so that we didn't print based on a suffix of the name, but I'm happy with the PR as-is if you don't want to do that.

Also looks like it needs a clang-format run.

mcourteaux:
I don't think it's easy to do, as the Buffers get added to the Module here: https://github.com/halide/Halide/blob/8b68f85ee07814c1896bdbcde57f2927f11cc732/src/Lower.cpp#L467-L471

And the only thing we have left is the fact that it's a Buffer, and not what it's used for, as far as I can tell at least.

For the clang-format, I ran clang-format, and force updated. Not sure if that's the way things go here, but I'd liked to see one commit for this instead of two :yum:

abadams:
Oh right, OffloadGPULoops just adds a reference to it and assumes it'll get picked up. I guess it would have to be a flag on the buffer.

abadams:
Oh, and for clang-format we normally push a second commit rather than force-pushing, and then squash the commits into one when we merge.

abadams (approved):

PR #6447, opened 2021-11-25
Syntax highlighting for embedded PTX code.

mcourteaux:
I put quite some work in this one, but the result is glorious. The string manipulations are not the most rigorous ones, but they work for me. This highlighting is only processed for PTX code, based on the buffer name being `"cuda_gpu_source_kernels"`. Here is an example of the syntax-highlighted PTX code:

image

Some features:

 - comments
 - labels
 - predicates
 - register matching
 - numeric literals
 - collapsible blocks per kernel function
 - shading for vectorized instructions.
 - memory accesses in red
 - different colors per datatype

### DIFF
--- a/src/IRPrinter.cpp
+++ b/src/IRPrinter.cpp
@@ -8,6 +8,7 @@
 #include "IROperator.h"
 #include "Module.h"
 #include "Target.h"
+#include "Util.h"
 
 namespace Halide {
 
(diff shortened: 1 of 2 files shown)


mcourteaux:
Spotted a bug. Will push fix soon.

abadams:
Seems like a lot of work for something minor, but the result does indeed look very nice!

mcourteaux:
Buildbot being unhappy with
 > correctness_partition_max_filter: /home/halidenightly/build_bot/worker/llvm-14-x86-64-linux/llvm-project/llvm/include/llvm/IR/User.h:170: llvm::Value* llvm::User::getOperand(unsigned int) const: Assertion `i < NumUserOperands && "getOperand() out of range!"' failed.

Is not my bad, right? This seems completely unrelated.

abadams:
Yeah that's probably some change in trunk llvm temporarily breaking things. Your code can't possibly be responsible.

abadams (approved):

PR #6455, opened 2021-11-30
Add a fast integer divide that rounds to zero

abadams:
While working on legacy code I discovered a need for this. Performance test shows a good speed-up over native division for vector code:

```
signed division rounding to zero:
type            const-divisor speed-up  runtime-divisor speed-up
 Int(32,  1)     2.416                   1.153
 Int(16,  1)     2.552                   1.457
 Int( 8,  1)     1.782                   0.667
 Int(32,  8)     8.592                   5.908
 Int(16, 16)    53.008                  38.505
 Int( 8, 32)    19.480                   8.197

```


src/FastIntegerDivide.cpp
```diff
@@ -205,6 +245,46 @@ Expr fast_integer_divide(Expr numerator, Expr denominator) {
...
+            break;
+        }
+        }
+
+        // Extract sign bit
+        // Expr xsign = (t.bits() < 32) ? (numerator / (1 << (t.bits()-1))) : (numerator >> (t.bits()-1));
+        Expr xsign = select(numerator > 0, cast(t, 0), cast(t, -1));
+
+        // Multiply-keep-high-half
+        result = (cast(wide, mul) * numerator);
```
dsharletg, line 275: I think this should use `widening_mul` intrinsics, because uses of this are after `find_intrinsics`. Maybe this whole sequence should be `mul_shift_right`.

test/performance/const_division.cpp
```diff
@@ -40,18 +40,33 @@ bool test(int w, bool div) {
...
     if (div) {
-        // Test div
-        f(x, y) = input(x, y) / cast<T>(y + min_val);
+        if (round_to_zero) {
+            // Test div
+            f(x, y) = div_round_to_zero(input(x, y), cast<T>(y + min_val));
 
-        // Reference good version
-        g(x, y) = input(x, y) / cast<T>(y + min_val);
+            // Reference good version
```
steven-johnson, line 55: This looks identical to the case just above, are they supposed to be identical?

tools/find_inverse.cpp
```diff
@@ -127,6 +131,50 @@ bool s_method_1(int den, int64_t mul, int sh_post, int bits) {
     return true;
 }
 
+bool srz_method_0(int den, int sh_post, int bits) {
+    int64_t min = -(1L << (bits - 1)), max = (1L << (bits - 1)) - 1;
+    for (int64_t num = min; num <= max; num++) {
+        // for (int iter = 0; iter < 1000000L; iter++) {
```
steven-johnson, line 137: Why is this commented out? If it's being left in for (eg) debugging purposes, please say so.

test/performance/const_division.cpp
abadams: Yes, they have different schedules which turn the denominator into a constant in one case but not the other.

abadams: (I'll add a comment)

tools/find_inverse.cpp
abadams: Fixed (deleted)

src/FastIntegerDivide.cpp
abadams: This is just moving around the original code, but sure I can update it.

abadams: Actually this code is only called directly by users, so it's before find_intrinsics. The compiler doesn't ever call this.

abadams:
See also related issue #6456

src/FastIntegerDivide.cpp
steven-johnson: Maybe add this as a comment for future readers.

steven-johnson (approved):
LGTM

src/FastIntegerDivide.cpp
dsharletg: I actually think we should change it to intrinsics anyways. But since the code is just moved and pre-existing, maybe it should be a separate PR.

dsharletg (approved):

lordnn:
`buf(x) = fast_integer_divide_round_to_zero(select(x % 2 == 0, 5, -5), 2);`
result is:
2, -3, 2, -3, 2, -3
Not rounded to zero.

abadams:
Looks like there's a bug in the handling of constant denominators (an early-out path that assumes we're rounding to -infinity). Will fix.

abadams:
See #7008

PR #6457, opened 2021-12-01
HexagonOffload needs to call inject_hvx_lock_unlock() before lower_parallel_tasks()

steven-johnson:
Not 100% sure this is the proper fix, but it's definitely the issue we are seeing with correctness_parallel_nested_1; the qurt_hvx_lock stuff only got injected for parallel For loops, but with the changes in factor_parallel_codegen, those are all gone by the time we hit that code. Solution here is to explicitly call inject_hvx_lock_unlock() on blocks that we are lowering as parallel tasks in HexagonOffload, but I need to convince myself we aren't inserting duplicate locks or unlocks now (or just getting lucky).

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -451,7 +451,7 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
     body = sloppy_unpredicate_loads_and_stores(body);
 
     debug(2) << "Lowering after unpredicating loads/stores:\n"
-    //          << body << "\n\n";
+             << body << "\n\n";
 
     if (is_hvx_v65_or_later()) {
         // Generate vscatter-vgathers before optimize_hexagon_shuffles.
@@ -464,7 +464,7 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
     const int lut_alignment = 64;
     body = optimize_hexagon_shuffles(body, lut_alignment);
     debug(2) << "Lowering after optimizing shuffles:\n"
-    //          << body << "\n\n";
+             << body << "\n\n";
 
     debug(1) << "Aligning loads for HVX....\n";
     body = align_loads(body, target.natural_vector_size(Int(8)), 8);
@@ -472,14 +472,14 @@ void CodeGen_Hexagon::compile_func(const LoweredFunc &f,
     // Don't simplify here, otherwise it will re-collapse the loads we
     // want to carry across loop iterations.
     debug(2) << "Lowering after aligning loads:\n"
-    //          << body << "\n\n";
+             << body << "\n\n";
 
     debug(1) << "Carrying values across loop iterations...\n";
     // Use at most 16 vector registers for carrying values.
     body = loop_carry(body, 16);
     body = simplify(body);
     debug(2) << "Lowering after forwarding stores:\n"
-    //          << body << "\n\n";
+             << body << "\n\n";
 
     // Optimize the IR for Hexagon.
     debug(1) << "Optimizing Hexagon instructions...\n";
@@ -2259,7 +2259,6 @@ Stmt inject_hvx_lock_unlock(Stmt body, const Target &target) {
     user_error << "hexagon not enabled for this build of Halide.\n";
     return Stmt();
 }
-
 
 #endif  // WITH_HEXAGON
 
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -3,6 +3,7 @@
 
 #include "AlignLoads.h"
 #include "CSE.h"
+#include "CodeGen_Hexagon.h"
 #include "CodeGen_Internal.h"
 #include "CodeGen_Posix.h"
 #include "Debug.h"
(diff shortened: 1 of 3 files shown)


dsharletg:
I am not sure this is the right fix. I thought the behavior at master was:
- We lock QURT on function entry
- Somewhere in the thread pool, prior to entering a parallel loop, we checked the HVX lock state, and pass that to the worker threads to call `qurt_hvx_lock` (via halide_do_task?)

However, I can't find the implementation of this.

I think after @zvookin's PR, the behavior instead should be simplified, because now we should be locking HVX every time we enter a parallel loop body, i.e. we should be able to simplify the code.

I suspect this may be buggy, because it's trying to re-lock HVX on the same thread where it is already locked. The solution that is probably to *unlock* HVX prior to calling `halide_do_par_for`, then relocking it after.

However, I now suspect that on master, we simply aren't calling qurt_hvx_lock at all on worker threads! I'm wondering if maybe we should just stop calling qurt_hvx_lock entirely? @pranavb-ca @aankit-ca IIRC locking HVX became optional at some point?

dsharletg:
Actually, my suggestion of unlocking + relocking might not be safe, because maybe we can capture vector variables in the parallel loop closure? Maybe these vectors would be lost if we unlocked + relocked HVX?

steven-johnson:
In theory, the behavior with this fix should be identical to what we did before; we're just doing the transformation to insert locks earlier, and the lock/unlock calls get redistributed along with the rest of the code in the loop. That said, this was already a messy thing that was hard to understand, so if this allows us to simplify the situation, I'm all for it -- I'm just not sure what the needed semantics for the lock/unlock are.

steven-johnson:
(converted to draft so I can change the base to master to enable proper testing; don't land it this way)

dsharletg:
OK, here's the logic I was looking for: https://github.com/halide/Halide/blob/ed87acb466f13144be235a33a30f242e30a6a74f/src/CodeGen_Hexagon.cpp#L333-L355

The thing I missed is that it was moved from runtime logic to codegen logic at some point.

steven-johnson:
> OK, here's the logic I was looking for:

Right, and that's (still) what we're doing here, so I think we should be alright. That said, it may be possible to simplify this down further; I will take a look at doing so tomorrow and try to offer an alternate fix.

steven-johnson:
@dsharletg @pranavb-ca @aankit-ca Are the specific rules for when `qurt_hvx_lock()/qurt_hvx_unlock()` need to be called documented? From a quick search through to code, I don't see any explicit docs about what needs to be protected this way;  based on the current implementation of `InjectHVXLocks`, it looks like the rules is that any instruction that uses vector registers in Hexagon code needs to have this lock acquired, but it would be nice to be sure I'm guessing correctly...

pranavb-ca:
Hey @steven-johnson give me some time to get back to you. I'll try this afternoon to see if we can get rid of this call altogether. (Sorry for the delayed response, I have been out of the office for the last two weeks. I am out until tomorrow but might be able to try this this afternoon nonetheless)

dsharletg:
I think we should keep the calls for now and work on removing them separately.

I think what we should do for now is that Zalman's refactor will insert qurt_hvx_lock calls everywhere needed. The problem is we lost the ability to *unlock* HVX prior to calling halide_do_par_for.

The existing logic tries to be clever and avoid unlocking/relocking when possible. I think maybe we should just unconditionally unlock HVX and relock HVX (if needed) prior to any halide_do_par_for call. I doubt the overhead/cost of this will be significant relative to loops that are parallelized.

steven-johnson:
> I think what we should do for now is that Zalman's refactor will insert qurt_hvx_lock calls everywhere needed. The problem is we lost the ability to _unlock_ HVX prior to calling halide_do_par_for.

Not sure I understand -- AFAICT, Zalman's refactor doesn't attempt to do anything like this right now, are you suggesting we modify it to do so? 

> The existing logic (and this PR preserves this) tries to be clever and avoid unlocking/relocking when possible

Yeah, but it seems to have been working for quite a while now. Is there a downside to re-using it that I'm overlooking? Lack of verifiability?

dsharletg:
> Not sure I understand -- AFAICT, Zalman's refactor doesn't attempt to do anything like this right now, are you suggesting we modify it to do so?

Zalman's PR causes the body of parallel loops to be a separate function, which get qurt_hvx_lock calls inserted by CodeGen_Hexagon just like any other function.

> Yeah, but it seems to have been working for quite a while now. Is there a downside to re-using it that I'm overlooking? Lack of verifiability?

Yes, it's complicated, and hard to be sure it is working as intended.

steven-johnson:
> which get qurt_hvx_lock calls inserted by CodeGen_Hexagon just like any other function.

Ah, ok, I see now a case where we do end up with a double-lock getting inserted by this. (Oddly, the test still passes.) I'll withdraw this and try your suggestions.

steven-johnson:
> unconditionally unlock HVX and relock HVX (if needed) prior to any halide_do_par_for call

Speaking of which -- it looks like we currently don't attempt to do any of the locking for HVX code that uses async (i.e. `halide_do_parallel_tasks`); I presume we can continue to just drop these on the floor...

PR #6465, opened 2021-12-02
Minor cleanup of parallel refactor intrinsics

steven-johnson:
- Renamed `load_struct_member` to `load_typed_struct_member` to make it more clear that it is intended for use only with the results of `make_typed_struct`.
- Split `declare_struct_type` into two intrinsics, `define_typed_struct` and `forward_declare_typed_struct`, removing the need for the underdocumented `mode` argument and hopefully making usage clearer
- Added / clarified comments for the intrinsics modified above


dsharletg (approved):

src/CodeGen_C.cpp
```diff
@@ -2550,13 +2562,19 @@ void CodeGen_C::visit(const Call *op) {
             rhs << "(" << print_type(op->type) << ")";
         }
         rhs << "(&" << struct_name << ((count > 0) ? "[0])" : ")");
-    } else if (op->is_intrinsic(Call::load_struct_member)) {
+    } else if (op->is_intrinsic(Call::load_typed_struct_member)) {
+        // Given an instance of a typed_struct, a definition of a typed_struct,
```
zvookin, line 2566: Given all the confusion around a value that solely exists to communicate a type, this comment should probably spend a few words on what an "instance of a typed struct" is. I.e. a Handle type that was previously returned from one of the intrinsics that returns one.

steven-johnson: PTAL

src/CodeGen_C.cpp
```diff
@@ -2550,13 +2562,24 @@ void CodeGen_C::visit(const Call *op) {
...
+        // and the index of a slot, load the value of that slot.
+        //
+        // An instance of a typed_struct is an Expr of Handle() type that has been
+        // created by a call to make_typed_struct.
+        //
+        // An definiton of a typed_struct is an Expr of Handle() type that has been
```
zvookin, line 2572: change "An" to "A".

It is probably worth covering that both are needed because the instance is often a void * by the time it is being used. E.g. it has been passed through an API that takes an opaque pointer as a void *. Thus it needs to be casted back to the correct type so its fields can be accessed.

src/CodeGen_LLVM.cpp
```diff
@@ -2891,18 +2895,28 @@ void CodeGen_LLVM::visit(const Call *op) {
...
         } while (item < count);
-    } else if (op->is_intrinsic(Call::load_struct_member)) {
+    } else if (op->is_intrinsic(Call::load_typed_struct_member)) {
+        // Given an instance of a typed_struct, a definition of a typed_struct,
+        // and the index of a slot, load the value of that slot.
+        //
+        // An instance of a typed_struct is an Expr of Handle() type that has been
+        // created by a call to make_typed_struct.
+        //
+        // An definiton of a typed_struct is an Expr of Handle() type that has been
```
zvookin, line 2905: Ditto comments on version in CodeGen_C.cpp . I find this method of documenting intrinsics highly dubious precisely for the reason that it is duplicated and thus a comment this long is basically unhelpful. Probably worth a TODO and an issue to move all intrinsics documentation into a header file with the intrinsics themselves.

zvookin (approved):

src/CodeGen_C.cpp
steven-johnson: Done

src/CodeGen_LLVM.cpp
steven-johnson: https://github.com/halide/Halide/issues/6468

PR #6467, opened 2021-12-06
Move null check from Printer to halide_string_to_string()

steven-johnson:
The Printer is (currently) usually inlined into every module, so this check is repeated in multiple chunks of code. Since the goal is to avoid crashing when debugging, let's move it to halide_string_to_string() (which will catch all these, and possibly more) and save some code size.

(Further improvements in Printer code size on the way; this change seems worthy of considering separately.)

### DIFF
--- a/src/runtime/printer.h
+++ b/src/runtime/printer.h
@@ -55,12 +55,7 @@ public:
     }
 
     Printer &operator<<(const char *arg) {
-        // Crashing on nullptr here is a big debugging time sink.
-        if (arg == nullptr) {
-            dst = halide_string_to_string(dst, end, "<nullptr>");
-        } else {
-            dst = halide_string_to_string(dst, end, arg);
-        }
+        dst = halide_string_to_string(dst, end, arg);
         return *this;
     }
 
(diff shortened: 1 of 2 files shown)


steven-johnson:
Windows failure appears to be a flake.

steven-johnson:
Gentle review ping

abadams:
string_to_string is used in a lot of places where the arg is known to be non-null (e.g. when you use a print expression in the IR). If this is just about moving the check out of the inlined Printer, can that be handled in #6473 ?

steven-johnson:
> If this is just about moving the check out of the inlined Printer, can that be handled in #6473 ?

It was originally about that, yes, and yes it can be handled elsewhere, but (again) I'd argue that if we want to put in a guard against crashing due to rogue null string ptrs, this is the better place to do it.

steven-johnson:
> > If this is just about moving the check out of the inlined Printer, can that be handled in #6473 ?
> 
> It was originally about that, yes, and yes it can be handled elsewhere, but (again) I'd argue that if we want to put in a guard against crashing due to rogue null string ptrs, this is the better place to do it.

(I'm not 100% married to this change, just think it's worthy of consideration in the name of defensive coding.)

abadams:
I don't think anything performance-critical calls this, so sure why not in the name of defensive programming.

abadams (approved):

Issue #6469, opened 2021-12-06
runtime::Printer has a `scratch` field that is ~never used


steven-johnson:
The runtime Printer has a `scratch` field that (presumably) is meant to avoid using `malloc()` for small print jobs, but in practice, it appears to ~never be used: 
- The length of `scratch` is (theoretically) a template parameter, which defaults to 1024
- ...but `scratch` is actually declared as `char scratch[length <= 256 ? length : 1];`, so any length of > 256 becomes just 1 (which in practice means "always use malloc")
- The places in runtime that override the default are:
    - profiler.cpp (which happens to also end up passing 1024)
    - tracing.cpp (which passes 4096)
    - opencl.cpp (16384)
    - d3d12compute.cpp (64k)
    - d3d12compute.cpp (16)
    - d3d12compute.cpp (256)

So basically, there are only two places (in d3d12compute) that are ever using this field.

The whole thing looks wonky to me -- my guess is that length was a straight passthrough in the past, but got limited this way due to stack-size issues on some architecture, and the ramifications never thought thru.

Anyway, this is an ugly wart that should really be resolved somehow, either by removing scratch entirely (since it's rarely used currently), or by deciding that a 1024-byte sized stack scratch is OK after all, or by deciding that  256-byte scratch is adequate for most things, or... something else.

(Note that d3d12compute.cpp has a `StackPrinter` class that attempts to make use of this, but the class is unused and should be removed.)

shoaibkamil:
Good catch, yikes!  I definitely did not understand this was the behavior.

Pinging @slomp to chime in about the one in `d3d12compute.cpp`.  I think we wanted to avoid `malloc()` and the bigger one is intentional.

steven-johnson:
> Good catch, yikes! I definitely did not understand this was the behavior.

The API is really a bit too clever for its own good

slomp:
When I implemented `StackPrinter`, it was to avoid malloc since I needed larger buffers for tracing / error dumps. The class had it's own internal `char buffer [length]` that was passed to `Printer<type, length>` to "override" the internal `malloc()`.

What seems to have happened is that @abadams did some improvements on Halide's `Printer` and decided to retire my `StackPrinter` (this is the commit for reference 2e5309aed8cfc8fe3a9d182f561e806844b528de)

slomp:
> Good catch, yikes! I definitely did not understand this was the behavior.
> 
[...]

PR #6472, opened 2021-12-06
Deal with Printer::scratch (#6469)

steven-johnson:
Instead of trying to optimize every Printer instance to use stack (and failing), move the StackPrinter concept into printer.h directly and require opt-in at the point of compilation to use stack instead of malloc.

This PR also does a few other drive-by cleanups:
- Ensures that all Printer ctors are explicit
- Makes some template aliases to make using (e.g.) ErrorPrinter with a custom buffer size slightly cleaner syntax
- Have tracing use the `.str()` method, which already deals with MSAN internally
- Make all the Printer data members private
- Fix some evil  code in opencl.cpp that previously used the now-private data members


slomp (approved):
LGTM!

PR #6473, opened 2021-12-06
Trim code size for Printer

steven-johnson:
This builds on top of https://github.com/halide/Halide/pull/6472 to de-inline as much of Printer as possible, moving things into a new 'printer' runtime module using a PrinterBase class.

This is, admittedly, a pretty small improvement: comparing before-and-after on OSX for target=host shows only a ~4k reduction in object size (115k -> 111k for `runtime.o`) but adding ~~targets~~ features with more verbose error reporting and such increases the benefit (eg host-opencl gives a 179k -> 164k reduction for runtime.o).

Since ~all of the usages of Printer in runtime are for error handling, debugging, profiling, or tracing, any possible reduction in performance seems unlikely to be significant.


steven-johnson:
Genuinely puzzled at the Windows-specific link errors here.

PR #6479, opened 2021-12-07
CSE across unrolls and split tuples

abadams:
The PR performs CSE across Tuple components and across the substmts in an unrolled loop. It helps with code size and compile time when you have a large expensive-to-compile expression that gets duplicated across many tuple components or across an unrolled loop. One could argue that you should just compute_at that expression, and yes you should, but it's nice to compile things quickly, even when the schedule is bad.

Reduces code size and compile-time of bgu by a few percent. No measurable effect on performance on anything I benchmarked because our schedules are good. Speeds up compile-time of some pathological code I have by more like 2x.


src/UnrollLoops.cpp
```diff
@@ -80,6 +82,75 @@ class UnrollLoops : public IRMutator {
                 user_warning << "Warning: Unrolling a for loop of extent 1: " << for_loop->name << "\n";
             }
 
+            // Peel lets that don't depend on the loop var to avoid needlessly
```
dsharletg, line 85: Could we just do LICM first, and then not need this code?

test/correctness/cse_across_unroll_and_tuples.cpp
```diff
@@ -0,0 +1,54 @@
...
+public:
+    int result = 0;
+};
+
+int main(int argc, char **argv) {
+    // Check that values shared across tuple elements or unrolled loops only get
+    // computed once in the Halide IR. LLVM will hoist them if we don't, but
+    // compilation can be much faster if we do it earlier, especially if the
+    // unrolled loop or tuple is large.
+    for (int use_tuple = 0; use_tuple < 2; use_tuple++) {
```
dsharletg, line 25: Nit: for this kind of thing, I personally prefer `for (bool use_tuple : {false, true})`

dsharletg (approved):

src/UnrollLoops.cpp
abadams: That was my first attempt. If you unroll an outer loop though it LICMs all inner loops, which is a much larger change than I was comfortable with. LICM also doesn't lift loads, which I want to. LICM needs an overhaul in general to do a proper graph cut instead of the current heuristic. It makes bad decisions about what to lift.

test/correctness/cse_across_unroll_and_tuples.cpp
abadams: done

steven-johnson:
OSX-x86 bot is down for repairs, probably OK to ignore for this

steven-johnson:
The mac-arm-bot failures may be legit injection, or may be uncovering existing latent bug, or may be LLVM injection? Can't repro locally yet.

steven-johnson:
> The mac-arm-bot failures may be legit injection, or may be uncovering existing latent bug, or may be LLVM injection? Can't repro locally yet.

The arm-osx failures definitely appear to be an injection from this PR; I hacked up a cross-compile that would repro the failure locally on my linux box, and it doesn't fail at master, but does fail here: https://gist.github.com/steven-johnson/28295bd146a99cdc844ecd8604ff2624

Running this with this PR fails with:
```
fft test: target=arm-64-osx w=8 h=8
Internal Error at CodeGen_LLVM.cpp:1253 triggered by user code at :
Symbol not found: t4271.widened.c2r_tiledT.s0.x.v82
Aborted
```

steven-johnson (changes requested):
Mac armbot failures require attention before this can land:

```
fft test: target=arm-64-osx w=8 h=8
Internal Error at CodeGen_LLVM.cpp:1253 triggered by user code at :
Symbol not found: t4271.widened.c2r_tiledT.s0.x.v82
Aborted
```

abadams:
I'm just going to abandon this and fix the schedule of the offending code instead. The failure indicates how subtle doing cross-stmt cse correctly is.

PR #6481, opened 2021-12-08
Try removing optional buffer added to closure

abadams:
Hopefully we can nuke this hack. Opening as a PR to see if it's necessary, not to merge it (because it would conflict with the other PR).


steven-johnson:
From local testing on host-metal, host-cuda, and host-hvx (via sim), nothing seems to be breaking.

abadams:
Some performance flakes, but yes! Looks like we can nuke this hack.

steven-johnson (approved):

PR #6482, opened 2021-12-08
Minor tweaks to generate_closure_ir()

steven-johnson:
Cherry-picking some minor changes from an otherwise-failed experimental branch:
- Condense code in generate_closure_ir() via use of a lambda
- Emit the calls to load_typed_struct_member in such a way that they are in ascending order (they currently get emitted in descending order, which is not wrong but is unsatisfying and suboptimal)
- Call simplify() on the final Stmt, so that unused calls to load_typed_struct_member get omitted from the IR (presumably LLVM will strip these out later but it can't fail to do so if we don't emit them at all)

### DIFF
--- a/src/LowerParallelTasks.cpp
+++ b/src/LowerParallelTasks.cpp
@@ -10,6 +10,7 @@
 #include "IROperator.h"
 #include "Module.h"
 #include "Param.h"
+#include "Simplify.h"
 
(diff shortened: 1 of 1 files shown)


steven-johnson:
https://github.com/halide/Halide/pull/6483 contains these changes and more; closing

PR #6483, opened 2021-12-08
Remove 'foo.buffer' from Closure entirely

steven-johnson:
This is a direct adaptation of what #6481 does for the old LLVM-based code, and allows elimination of one use of `get_pointer_or_null()`. PR is meant to be merged into factor_parallel_codegen, not master.


steven-johnson:
Clean, PTAL

PR #6487, opened 2021-12-09
factor parallel codegen with fewer intrinsics

abadams:
This version relies more heavily on the existing make_struct, and puts function names in the Variable namespace as globals.

I ended up not doing a StructType, because it didn't seem to pay for itself compared to using a prototype, as the existing code did. It turns out in that approach it was still possible to remove many of the intrinsics.


src/Closure.cpp
```diff
@@ -106,5 +109,79 @@ void Closure::visit(const Atomic *op) {
...
+    }
+
+    // Sort by decreasing size, to guarantee the struct is densely packed in
+    // memory. We don't actually rely on this, it's just nice to have.
+    std::stable_sort(elements.begin(), elements.end(),
+                     [&](const Expr &a, const Expr &b) {
+                         return a.type().bytes() > b.type().bytes();
+                     });
+
+    for (Expr e : elements) {
```
steven-johnson: style nit: declaring `types` and `name` here (instead of at top) is perhaps better.

src/Closure.cpp
```diff
@@ -106,5 +109,79 @@ void Closure::visit(const Atomic *op) {
...
+        using IRMutator::mutate;
+        Expr mutate(const Expr &e) override {
+            if (!e.as<Call>()) {
+                return make_zero(e.type());
+            } else {
+                return IRMutator::mutate(e);
+            }
+        }
+    } replacer;
+    string prototype_name = unique_name("closure_prototype");
```
steven-johnson, line 153: random thought: any reason not to use single-char shortnames here (e.g. `unique_name('c')`? this is here presumably to improve readability of our IR, but if that's the goal, we should try to also include some name unique to the specific final name of the Closure's loop/lambda/etc; otherwise something like `c$3` is arguably just as good.

src/Closure.cpp
```diff
@@ -106,5 +109,79 @@ void Closure::visit(const Atomic *op) {
...
+        name << e.type() << "_";
+    }
+
+    Expr result = Call::make(Handle(),
+                             Call::make_struct, elements, Call::Intrinsic);
+    return result;
+}
+
+Stmt Closure::unpack_from_struct(const Expr &e, const Stmt &s) const {
+    // Use the struct-packing code just to make sure the order of elements is
```
steven-johnson, line 137: Since this code relies on `Closure` having a deterministic order for vars and buffers, we should add a comment to that effect Closure.h in case someone ever wants to change the usage of `std::map` to (eg) `std::unordered_map`.

src/CodeGen_C.cpp
```diff
@@ -2024,7 +2024,13 @@ string CodeGen_C::print_assignment(Type t, const std::string &rhs) {
     if (cached == cache.end()) {
         id = unique_name('_');
         const char *const_flag = output_kind == CPlusPlusImplementation ? "const " : "";
-        stream << get_indent() << print_type(t, AppendSpace) << const_flag << id << " = " << rhs << ";\n";
+        if (t.is_handle()) {
+            // Don't print void *, which might lose useful type information. just use auto.
+            stream << get_indent() << "auto ";
```
steven-johnson, line 2029: uber-nit: `auto *` might be better, since we always expect some kind of pointer type, yes?

src/CodeGen_C.cpp
```diff
@@ -2051,7 +2057,12 @@ void CodeGen_C::close_scope(const std::string &comment) {
 }
 
 void CodeGen_C::visit(const Variable *op) {
-    id = print_name(op->name);
+    if (starts_with(op->name, "::")) {
+        // This is the name of a global, so we can't modify it.
```
steven-johnson, line 2061: When does this actually happen?

src/CodeGen_C.cpp
```diff
@@ -2429,7 +2440,7 @@ void CodeGen_C::visit(const Call *op) {
             rhs << shape_name;
         } else {
             // Emit a declaration like:
-            // struct {const int f_0, const char f_1, const int f_2} foo = {3, 'c', 4};
+            // struct {int f_0, char f_1, int f_2} foo = {3, 'c', 4};
```
steven-johnson, line 2443: This specific example shouldn't actually be possible (at least for Closures), since `pack_struct` code sorts by size.

src/CodeGen_LLVM.cpp
```diff
@@ -551,41 +584,10 @@ std::unique_ptr<llvm::Module> CodeGen_LLVM::finish_codegen() {
...
-    }
-    set_function_attributes_for_target(function, target);
-
-    // Mark the buffer args as no alias
-    for (size_t i = 0; i < args.size(); i++) {
-        if (args[i].is_buffer()) {
-            function->addParamAttr(i, Attribute::NoAlias);
-        }
+        module->dump();
+        internal_assert(function) << "Could not func a function of name " << extern_name << " in module\n";
```
steven-johnson, line 589: How many funcs would a function func if a function could func... something?

src/CodeGen_LLVM.cpp
```diff
@@ -2817,156 +2819,35 @@ void CodeGen_LLVM::visit(const Call *op) {
...
-            for (const auto &arg : decl_call->args) {
-                mangle_args.emplace_back(arg);
-            }
-            name = cplusplus_function_mangled_name(name, namespaces, decl_call->type, mangle_args, get_target());
+        llvm::Value *struct_instance = codegen(op->args[0]);
+        llvm::Value *struct_prototype = codegen(op->args[1]);
+        llvm::Value *typed_struct_instance = builder->CreatePointerCast(struct_instance, struct_prototype->getType());
+        const int64_t *index = as_const_int(op->args[2]);
+
+        // make_struct can use a fixed-size struct, an array type, or a scalar
```
steven-johnson, line 2836: Not to sidetrack, but: Wait, seriously? Is that pre-existing behavior? If so, why? Presumably there's a benefit that is worth the complication?

src/LowerParallelTasks.cpp
```diff
@@ -379,33 +297,23 @@ struct LowerParallelTasks : public IRMutator {
...
-                // CodeGen adds user_context for us apparently.
-                //                args[0] = Call::make(type_of<void *>(), Call::get_user_context, {}, Call::PureIntrinsic);
-                args[0] = Call::make(type_of<const void *>(), Call::resolve_function_name, {function_decl_call}, Call::PureIntrinsic);
+                // Codegen will add user_context for us
+
+                // Prefix the function name with "::" as we would in C to make
+                // it clear we're talking about something in global scope in
+                // case some joker names an intermediate Func or Var the same
+                // name as the pipeline.
+                args[0] = Variable::make(Handle(), "::" + new_function_name);
```
steven-johnson, line 307: We created `new_function_name` via a call to  `unique_name` above, so shouldn't this be unnecessary?

EDIT: hm, I guess `c_print_name` could transmute it back into a non-unique name via $->_ transforms? Ugh.

steven-johnson (commented):
LGTM so far. Going to pull it locally and see how the output looks.

src/CodeGen_LLVM.cpp
abadams: Yes, it cleans up the LLVM IR and generally makes more optimizations happen, especially when you have a lot of things of the same type in a make_struct call (e.g. due to tracing, where it's all ints)

src/Closure.cpp
abadams: It's the same closure object, so the order would have to change between two iterations over the same map. I don't think even unordered_map is that non-deterministic.

src/Closure.cpp
abadams: It's a confusing thing in the IR. I think it deserves a long name to say why there's an apparently unused struct.

src/Closure.cpp
abadams: This was dead code. I deleted it.

src/CodeGen_C.cpp
abadams: Fixed

src/CodeGen_LLVM.cpp
abadams: Fixed

zvookin:
I'm going to hand the rest of this over to Steven. He and I are talking as needed.

steven-johnson:
`test_internal` is broken because of C++ codegen differences

test/common/check_call_graphs.h
```diff
@@ -14,97 +14,43 @@
...
-
-    void visit(const Halide::Internal::ProducerConsumer *op) override {
+    Halide::Internal::Stmt visit(const Halide::Internal::ProducerConsumer *op) override {
         if (op->is_producer) {
             std::string old_producer = producer;
             producer = op->name;
             calls[producer];  // Make sure each producer is allocated a slot
             // Group the callees of the 'produce' and 'update' together
-            op->body.accept(this);
+            auto new_stmt = mutate(op->body);
```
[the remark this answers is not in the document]
abadams: This is a pattern from other tests. If you want to validate IR at the end of lowering, but before backend stuff starts, you can do it with a custom lowering pass. Those must be mutators, not visitors.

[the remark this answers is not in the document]
abadams: In this case the upside is that it happens before we split things off into closures, which removes the complexity of handling that from this test.

test/correctness/func_clone.cpp
```diff
@@ -64,9 +64,9 @@ int func_clone_test() {
     // Check the call graphs.
     // Expect 'g' to call 'clone', 'clone' to call nothing, and 'f' not
     // in the final IR.
-    Module m = g.compile_to_module({});
     CheckCalls c;
-    c.add_module(m);
+    g.add_custom_lowering_pass(&c, nullptr);
```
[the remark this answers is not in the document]
abadams: Yeah it's 3 lines of code copy-pasted everywhere. This is a holdover from the original structure of these tests (I didn't want to change things too much from master). I guess I should clean it up.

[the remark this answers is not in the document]
abadams: Fixed

test/common/check_call_graphs.h
```diff
@@ -129,7 +75,7 @@ inline int check_call_graphs(CallGraphs &result, CallGraphs &expected) {
                     return a.empty() ? b : a + ", " + b;
                 });
 
-            printf("Expect callees of %s to be (%s); got (%s) instead\n",
+            printf("Expect calless of %s to be (%s); got (%s) instead\n",
```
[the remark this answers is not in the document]
abadams: Fixed

src/CodeGen_C.cpp
```diff
@@ -2024,7 +2024,13 @@ string CodeGen_C::print_assignment(Type t, const std::string &rhs) {
     if (cached == cache.end()) {
         id = unique_name('_');
         const char *const_flag = output_kind == CPlusPlusImplementation ? "const " : "";
-        stream << get_indent() << print_type(t, AppendSpace) << const_flag << id << " = " << rhs << ";\n";
+        if (t.is_handle()) {
+            // Don't print void *, which might lose useful type information. just use auto.
+            stream << get_indent() << "auto *";
```
steven-johnson, line 2029: Looks like my suggestion about `auto *` was bad; I made a clone of this for testing and there are failures because we generated code of the form
```
pipeline_c.halide_generated.cpp:3720:17: error: unable to deduce ‘auto*’ from ‘0’
    auto *_275 = 0;
```

(see https://buildbot.halide-lang.org/master/#/builders/118/builds/541)

...that said, is just using `auto` here ok if the RHS is a plain `0`? If the type is Handle then surely we want to ensure it's at least pointer sized (if not explicitly 64 bits)...

abadams: Ideally we could turn that 0 into 'nullptr'. I'll take a look.

abadams: It was an unused value, so it was actually fine, but it was a place where we were explicitly deciding to use 0 for a pointer, so it was easy to change to nullptr

steven-johnson: now we get

```
pipeline_c.halide_generated.cpp:3720:17: error: unable to deduce ‘auto*’ from ‘nullptr’
    auto *_275 = nullptr;
```

steven-johnson (approved):
LGTM, ok to merge.

PR #6489, opened 2021-12-09
Fail if no_bounds_query specified for HL_JIT_TARGET

steven-johnson:
JIT requires the use of bounds_query; disabling it will almost certainly fail in JIT mode, either with a confusing assert message, or a crash (if you also specify no_asserts). This adds a more useful failure message.

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -431,6 +431,8 @@ Target get_jit_target_from_environment() {
             << "HL_JIT_TARGET must match the host OS, architecture, and bit width.\n"
             << "HL_JIT_TARGET was " << target << ". "
             << "Host is " << host.to_string() << ".\n";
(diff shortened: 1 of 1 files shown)


src/Target.cpp
```diff
@@ -431,6 +431,8 @@ Target get_jit_target_from_environment() {
             << "HL_JIT_TARGET must match the host OS, architecture, and bit width.\n"
             << "HL_JIT_TARGET was " << target << ". "
             << "Host is " << host.to_string() << ".\n";
+        user_assert(!t.has_feature(Target::NoBoundsQuery))
+            << "The Halide JIT requires the use of bounds query, but HL_JIT_TARGET was specified with no_bunds_query: " << target;
```
abadams, line 435: bunds

abadams (approved):

PR #6491, opened 2021-12-10
Document the usage of llvm::legacy::PassManager

steven-johnson:
There is some confusion about whether this usage is acceptable. TL;DR: it's not just acceptable, it's required for the forseeable future. Add comments to capture this to avoid future such questions. (With great thanks to Alina for pointing me at the relevant LLVM discussion links!)

### DIFF
--- a/src/CodeGen_PTX_Dev.cpp
+++ b/src/CodeGen_PTX_Dev.cpp
@@ -639,7 +639,7 @@ vector<char> CodeGen_PTX_Dev::compile_to_src() {
 
     // NOTE: use of the "legacy" PassManager here is still required; it is deprecated
     // for optimization, but is still the only complete API for codegen as of work-in-progress
-    // LLVM14. At the time of this comment, there is no firm plan as to when codegen will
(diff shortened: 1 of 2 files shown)


alexreinking (approved):

src/CodeGen_PTX_Dev.cpp
```diff
@@ -637,6 +637,16 @@ vector<char> CodeGen_PTX_Dev::compile_to_src() {
     raw_svector_ostream ostream(outstr);
     ostream.SetUnbuffered();
 
+    // NOTE: use of the "legacy" PassManager here is still required; it is deprecated
+    // for optimization, but is still the only complete API for codegen as of work-in-progress
+    // LLVM14. At the time of this comment, there is no firm plan as to when codegen will
```
alexreinking, line 642: > At the time of this comment

I know I can check the `git blame`, but it might be worth jotting down (12/10/21) anyway.

steven-johnson: Done

PR #6492, opened 2021-12-10
Add initial support for WebGPU

jrprice:
This adds the core of a new WebGPU backend. There are still many holes in the functionality and rough edges around how this interacts with Emscripten, native WebGPU implementations, and testing. Nevertheless, I'm opening this after offline discussion with @steven-johnson to get some initial feedback on this backend (and to raise awareness that this is happening).

While a lot of the implementation still remains to be done, this PR provides enough to run a 32-bit version of `apps/blur`, the `gpu_only` AOT generator test, and the `correctness_bounds` JIT test. I've also got the HelloWasm app running with the `render` pipeline targeting the GPU (and the other two pipelines still using WASM CPU).

For testing, the AOT generator tests can be made to work for `wasm-32-wasmrt-webgpu` when using Node bindings for [Dawn](https://dawn.googlesource.com/dawn/) (Chromium's native implementation of WebGPU). Both the AOT tests and JIT tests can also work for `host-webgpu` when using Dawn. JIT is not currently supported for `wasm-32-wasmrt-webgpu`.

Unlike the other GPU backends, I'm not employing the `dlopen`/`dlsym` approach in the runtime for getting the API functions. I'm not sure how to make this work when using Emscripten, and since using Dawn directly is only really needed for testing purposes it doesn't seem /too/ onerous to require direct linking, but I'm open to opinions and suggestions here.

Another pain point right now is that the C++ API for WebGPU is not currently stable between different implementations, so there is a build-time switch to toggle between targeting Emscripten vs Dawn. I'm optimistic that this requirement will eventually go away.

There's still some patches that need to land in both Emscripten and Dawn before this backend will work for anyone else, so I'll leave this PR as a draft until those are resolved.

All feedback is very welcome! I've joined the Gitter room as well.


dependencies/wasm/CMakeLists.txt
```diff
@@ -92,8 +92,16 @@ function(add_wasm_executable TARGET)
         -s ASSERTIONS=1
         -s ALLOW_MEMORY_GROWTH=1
         -s ENVIRONMENT=node
+        -s USE_WEBGPU=1
+        -s ASYNCIFY
```
steven-johnson, line 96: ASYNCIFY is a pretty nontrivial code change IIRC -- What in this PR requires it? I don't think we'd want to enable it for most tests.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,864 @@
+#include "HalideRuntimeWebGPU.h"
+#include "device_interface.h"
+#include "gpu_context_common.h"
+#include "printer.h"
+#include "scoped_spin_lock.h"
+
+// TODO: Discover this with CMake, and upstream toggle for stdint.h.
+#include "webgpu/webgpu.h"
```
steven-johnson, line 8: Code in src/runtime isn't allowed to include headers from the wider environment -- this is a hard requirement, to allow for cross-compiling on systems without having to have impossible SDK requirements everywhere. You'll need to pull out the forward declarations you need and put them here (see, e.g., `mini_opengl.h`). If the API for webgpu is too unstable for that to work, it's too unstable for Halide to use :-/

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,864 @@
...
+namespace Runtime {
+namespace Internal {
+namespace WebGPU {
+
+extern WEAK halide_device_interface_t webgpu_device_interface;
+
+WEAK int create_webgpu_context(void *user_context);
+
+// A WebGPU instance/adapter/device defined in this module with weak linkage.
+WGPUInstance WEAK global_instance = nullptr;
```
steven-johnson, line 20: style nit: in Halide runtime, we generally prefer `WEAK type name;` to `name WEAK type;`

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,864 @@
...
+WGPUAdapter WEAK global_adapter = nullptr;
+WGPUDevice WEAK global_device = nullptr;
+// Lock to synchronize access to the global WebGPU context.
+volatile ScopedSpinLock::AtomicFlag WEAK context_lock = 0;
+
+// Size of the staging buffer used for host<->device copies.
+#define WEBGPU_STAGING_BUFFER_SIZE (4 * 1024 * 1024)
```
steven-johnson, line 27: style nit: `constexpr int` would work fine here :-)

src/runtime/CMakeLists.txt
```diff
@@ -163,6 +165,7 @@ set(RUNTIME_CXX_FLAGS
     -Wno-psabi
     -Wno-unknown-warning-option
     -Wno-unused-function
+    -Wno-atomic-alignment
```
steven-johnson, line 168: Why is this required? Where are we getting the warnings? (This sounds like a warning we shouldn't be ignoring!)

src/CodeGen_WebGPU_Dev.cpp
```diff
@@ -0,0 +1,385 @@
...
+}
+
+void CodeGen_WebGPU_Dev::CodeGen_WGSL::visit(const IntImm *op) {
+    internal_assert(op->type.bits() == 32)
+        << "WGSL only supports 32-bit integers";
+    print_assignment(op->type, std::to_string(op->value));
+}
+
+void CodeGen_WebGPU_Dev::CodeGen_WGSL::visit(const UIntImm *op) {
+    internal_assert(op->type.bits() == 32)
```
steven-johnson, line 305: Does it support Boolean? (Halide implements boolean types as `UInt(1)` so this is where you  would get bools.

src/CodeGen_WebGPU_Dev.cpp
```diff
@@ -0,0 +1,385 @@
...
+}
+
+void CodeGen_WebGPU_Dev::CodeGen_WGSL::visit(const Cast *op) {
+    print_assignment(op->type,
+                     print_type(op->type) + "(" + print_expr(op->value) + ")");
+}
+
+void CodeGen_WebGPU_Dev::CodeGen_WGSL::visit(const IntImm *op) {
+    internal_assert(op->type.bits() == 32)
+        << "WGSL only supports 32-bit integers";
```
steven-johnson, line 300: ...well that is disappointing. No other int sizes? or unsigned?

src/CodeGen_WebGPU_Dev.cpp
```diff
@@ -0,0 +1,385 @@
...
+        default:
+            user_error << "Unsupported vector width in WGSL: " << type << "\n";
+        }
+    }
+
+    if (type.is_float()) {
+        user_assert(type.bits() == 32) << "WGSL only supports 32-bit floats";
+        oss << "f32";
+    } else if (type.bits() == 1) {
+        oss << "bool";
```
steven-johnson, line 163: This test should be `type.is_uint() && type.bits() == 1`, or, more simply, `type == UInt(1)`

steven-johnson (commented):
Thanks for this -- I don't have time today for more than a quick skim, but wanted to provide a handful of (minor) bits of feedback, but the important thing to tell you right away is that including webgpu.h (or any other outside-of-Halide header file) in the runtime code is a no-no, at least for anything beyond initial experimentation. Wanted to be sure you were aware of that sooner rather than later.

src/CodeGen_WebGPU_Dev.cpp
jrprice: Done. I actually used `type == Bool()`, which is even simpler again.

src/CodeGen_WebGPU_Dev.cpp
jrprice: Yes. I've implemented this for `Bool()`, although I struggled to construct a Halide function that actually generate `Bool` literals without folding them away.

src/runtime/webgpu.cpp
jrprice: Done. I had just copied that style from the equivalent variables in the CUDA and OpenCL runtimes which for some reason use `name WEAK type`, but I definitely prefer `WEAK type name` too.

src/runtime/webgpu.cpp
jrprice: Done.

dependencies/wasm/CMakeLists.txt
steven-johnson: It may be OK everywhere, I just need to re-read the details of the code transformation to decide if it compromises the value of our (other) wasm tests in any way. If it doesn't, no worries. (Definitely worth adding a comment about this flag being necessary for WebGPU though.)

src/runtime/webgpu.cpp
steven-johnson: OK. If this causes any unusual issues then let us know. (If it's just kinda icky-feeling, well, that's the case in all the other backends too, but as a practical matter, the stuff we assume is the stuff that isn't going to change cavalierly.)

src/runtime/CMakeLists.txt
steven-johnson: Ah, this vaguely rings a bell, but not gonna investigate tonight. Definitely flag it with a `# TODO: explain the issue here` comment so we don't overlook circling back to it.

src/CodeGen_WebGPU_Dev.cpp
steven-johnson: All I can say is that most of the high-performance, well-scheduled chunks of Halide code I've seen in production make *extensive* use of 8- and 16-bit vector types. Their absence would be a big limitation for maximizing performance (and memory usage), not to mention making portable pipelines a lot more painful to write and schedule (since pipelines running on typical SIMD or DSP architectures won't want to leave the benefits of having smaller ints, and thus effectively-wider vectors, on the table...)

abadams: For other type-challenged backends (d3d12) I believe we emulate low bit width types in Halide, so you could take a look at those backends. Feel free to email me for help demonstrating enough demand.

src/runtime/CMakeLists.txt
abadams: See src/runtime/synchronization_common.h. It wraps a lot of these atomic ops to avoid that error. See also src/runtime/scoped_spin_lock.h

dependencies/wasm/CMakeLists.txt
jrprice: Added a comment.

src/runtime/CMakeLists.txt
jrprice: Thanks for the pointers @abadams, I've tweaked my atomic usage and remove `-Wno-atomic-alignment`.

src/runtime/webgpu.cpp
jrprice: I've now done this. We'll need to update it every now and again to track changes to the native API while its still being developed, but this should eventually settle down.

dependencies/wasm/CMakeLists.txt
steven-johnson: OK, after re-reading the details on Asyncify, it seems invasive enough that I'm inclined to suggest we should try to only use it when necessary, rather than across the board (e.g.: "As mentioned earlier, unoptimized builds with Asyncify can be large and slow. Build with optimizations (say, -O3) to get good results.") No need to worry about it at this point (and I'm more than willing to drop this if point if the emcc folk tell me I'm overthinking it).

steven-johnson:
(I took the liberty of pacifying clang-tidy and bringing up to date with top-of-tree; please let me know if you'd prefer I avoid doing changes of this sort here in the future, but these seemed harmless)

steven-johnson:
Maybe I missed it here, but could you add something (either in this PR or in a README, or both) about what is needed to configure this for local testing? e.g. how do I install and configure Dawn for use, what test(s) are expected to work or not work, etc? I'd like to pull this locally and try it out but not sure about the details.

dependencies/wasm/CMakeLists.txt
jrprice: I've made the addition of `-s ASYNCIFY` conditional on the target including the `webgpu` feature for now.

jrprice:
> Maybe I missed it here, but could you add something (either in this PR or in a README, or both) about what is needed to configure this for local testing? e.g. how do I install and configure Dawn for use, what test(s) are expected to work or not work, etc? I'd like to pull this locally and try it out but not sure about the details.

Sorry for the delay - I've added a `README_webgpu.md` file that should cover this. These are the tests that are currently expected to work:
```
correctness_bounds
correctness_func_lifetime
correctness_func_lifetime_2
correctness_gpu_give_input_buffers_device_allocations
correctness_gpu_jit_explicit_copy_to_device
correctness_hello_gpu
correctness_loop_invariant_extern_calls
correctness_lots_of_loop_invariants
correctness_parallel_gpu_nested

generator_aot_gpu_only
```
It's a pretty small set, but I expect that many more will pass without too much additional effort. If you'd like a larger set to pass before landing this PR just let me know.

src/runtime/README_webgpu.md
```diff
@@ -0,0 +1,102 @@
...
+JIT compilation is not supported when using WebGPU with WASM.
+
+## Running natively: `HL_TARGET=host-webgpu`
+
+For testing purposes, Halide can also target native WebGPU libraries, such as
+[Dawn](dawn.googlesource.com/dawn/) or [wgpu](github.com/gfx-rs/wgpu).
+This is currently the only path that can run the JIT correctness tests.
+See [below](#setting-up-dawn) for instructions on building Dawn.
+
+Due to differences between the APIs implemented by native WebGPU libraries and
```
steven-johnson, line 59: Out of curiosity,, what are those differences, and are they likely to remain different in the future, or is this just a work-in-progress mismatch?

src/runtime/README_webgpu.md
```diff
@@ -0,0 +1,102 @@
...
+Emscripten, this currently requires a separate build of Halide.
+Pass `-DWEBGPU_NATIVE_LIB=/path/to/native/library.{so,dylib.dll}` to CMake when
+configuring Halide to enable this path, which will automatically use this
+library for the AOT and JIT tests.
+
+## Setting up Dawn
+
+Building Dawn's Node.js bindings currently requires using CMake.
```
steven-johnson, line 67: Thanks for the detailed description! This sort of setup info can be hard to find / figure out for newcomers.

src/runtime/README_webgpu.md
jrprice: The differences should all eventually disappear, once the native implementations + Emscripten converge on a de facto standard for the native API. 

It's mostly little things, e.g. Emscripten doesn't yet implement `wgpuCreateInstance()` so we pass a `nullptr` instance when requesting a device, whereas Dawn expects a valid instance object. You can see all of the places where the APIs diverge by looking for `WITH_DAWN_NATIVE` in `webgpu.cpp`.

src/CodeGen_WebGPU_Dev.cpp
```diff
@@ -278,6 +303,41 @@ void CodeGen_WebGPU_Dev::CodeGen_WGSL::add_kernel(
     close_scope("shader " + name);
 }
 
+void CodeGen_WebGPU_Dev::CodeGen_WGSL::visit(const Allocate *op) {
+    if (op->memory_type == MemoryType::GPUShared) {
+        // TODO: Implement.
+        internal_assert(false)
```
steven-johnson, line 309: FYI: `internal_error` is equivalent to `internal_assert(false)`

src/CodeGen_WebGPU_Dev.cpp
```diff
@@ -344,9 +404,20 @@ void CodeGen_WebGPU_Dev::CodeGen_WGSL::visit(const For *loop) {
         loop->body.accept(this);
 
     } else {
-        user_assert(loop->for_type != ForType::Parallel)
-            << "Cannot use parallel loops inside WebGPU shaders\n";
-        CodeGen_C::visit(loop);
+        user_assert(loop->for_type == ForType::Serial)
+            << "Cannot only use serial loops inside WebGPU shaders\n";
```
steven-johnson, line 408: The error message doesn't quite match the assert? Or maybe I misunderstand.

jrprice: Oops, fixed :-)

src/CodeGen_WebGPU_Dev.cpp
```diff
@@ -199,6 +213,16 @@ string CodeGen_WebGPU_Dev::CodeGen_WGSL::print_type(Type type,
     return oss.str();
 }
 
+string CodeGen_WebGPU_Dev::CodeGen_WGSL::print_buffer_type(Type type) {
+    if (type.bits() != 32) {
+        internal_assert(type.bits() == 8 || type.bits() == 16);
+        internal_assert(!type.is_float());
+        // Types that are not 32-bits are emulated using atomics.
```
steven-johnson, line 231: What's the performance hit from using atomics in this way?

jrprice: Very substantial. Depends on the platform, the access pattern, and the amount of data, but we've measured anything from 2X to >10X slowdown compared to just using 32-bit types instead. As soon as WGSL gains an extension for these types we can start using that instead, which will help on platforms that can do this natively (macOS, Linux/Android with appropriate Vulkan extensions, and Windows with SM 6+).

If the slowdown is too scary to expose this by default, we could hide this functionality behind a Halide feature flag (e.g. `webgpu-int16`), which could use the extension unconditionally (and produces a runtime error if you try and run those pipelines on a platform that doesn't have it).

Also worth noting: we could avoid the atomics for cases where the accesses are dense loads/stores of whole vectors, meaning we would only incur this penalty for scalar i8/i16 stores or scatters/gathers. Not sure if that helps.

steven-johnson: > As soon as WGSL gains an extension for these types we can start using that instead, which will help on platforms that can do this natively (macOS, Linux/Android with appropriate Vulkan extensions, and Windows with SM 6+).

Any ETA for when that extension might exist?
 
> If the slowdown is too scary to expose this by default

...it looks pretty scary, I gotta admit
 
IIRC @abadams mentioned that other backends have ways to fake smaller types, but I don't know how they are implemented.

jrprice: I've been under the impression that the D3D backend does the same emulation via atomics:
https://github.com/halide/Halide/blob/master/src/CodeGen_D3D12Compute_Dev.cpp#L446-L479

But now that I look closer, it looks like this codepath is actually unused. It appears that the D3D backend can do this without atomics, by just setting the format at the buffer view to something like `DXGI_FORMAT_R16_UINT` on the runtime side whilst declaring the buffer as a 32-bit integer in the shader. This is all news to me, so let me think about whether this changes any of our thoughts about these types on the WebGPU/Chrome side.

steven-johnson: Adding @slomp and @shoaibkamil for the questions about the D3D backend

slomp: Yup, in the case of D3D12, it's possible to define a "buffer view" with a specific format over the buffer memory, then "sample" it from the shader code and the hardware will pack/unpack the values accordingly to either `float` or `int`/`uint` (these are the only native types supported by HLSL, disregarding Shader Model 6.2).

slomp: (In GL-land, you can achieve that with what ARB refers to as "texture buffer".)

jrprice:
> Any updates on this? How close are we getting to complete-enough functionality to do real testing?

This PR is now passing 60/91 of the GPU-enabled correctness tests and there's 9 that I think are n/a for WebGPU. So 22 actual test failures remaining.

One of the main remaining pieces of work here is to handle buffer cropping/slicing and non-contiguous copies. Dynamically-sized GPU tiles and shared memory regions is another source of failures.

I've been working with my colleagues to try to figure out a way to expose functionality that would allow Halide to implement 16-bit and 8-bit integers more efficiently. If the slow 8/16-bit performance isn't considered a blocker for landing this PR though, I can get back to knocking out the remaining test failures in the meantime.

steven-johnson:
> implement 16-bit and 8-bit integers more efficiently

IIRC, @shoaibkamil / @slomp did something like that for the D3D12 backend, but I don't know the details (or am misremembering). Not sure if it's something that could be recycled and/or shared. Pinging them here for response :-)

jrprice:
> IIRC, @shoaibkamil / @slomp did something like that for the D3D12 backend, but I don't know the details (or am misremembering). Not sure if it's something that could be recycled and/or shared. Pinging them here for response :-)

Yes, we've been looking closely at how the D3D12 backend handles this stuff (via typed UAV loads/stores). The issue right now is that WebGPU does not currently expose an equivalent feature, so we either need to add such functionality to WebGPU itself, or do something different. Hoping to have a path forward soon though.

steven-johnson:
Any update on this?

jrprice:
> Any update on this?

Nothing lately, but I expect to back working on this again soon.

steven-johnson:
What's the status on this PR -- is activity likely to be resumed anytime soon?

jrprice:
> What's the status on this PR -- is activity likely to be resumed anytime soon?

This PR should be ready for review and further testing by others, as it implements the baseline functionality that we had previously agreed on a call. 

On my macOS machine using Dawn native, all correctness tests are passing except:
- correctness_gpu_dynamic_shared
- correctness_gpu_non_monotonic_shared_mem_size
- correctness_gpu_param_allocation
- correctness_gpu_reuse_shared_memory
- correctness_gpu_specialize
- correctness_interpreter
- correctness_isnan

These failures are caused by dynamically sized shared memory (supported by WebGPU but not yet implemented in Dawn/Chrome), lack of NaN support (discussed previously), and exceeding maximum allocation sizes for device memory and shared memory.

I'm currently on leave until the end of October, and will be able to address merge conflicts / review comments when I return. I'll also then start looking at ways to improve performance for small integer types.

jrprice:
The WebGPU backend now passes all correctness tests except `correctness_isnan` on my macOS machine using Dawn native on an Intel GPU. There are a few failures on AMD due to a Metal compiler bug. 

This PR should now be ready for review and further testing by others. Everything that was previously agreed for this backend's MVP has now been implemented.

jrprice:
Thanks for the review! I'll address the comments as soon as I can.

> What OS(s) should we target for testing? I'm assuming maybe OSX and Linux-x64?

I've been testing locally on macOS (AMD + Intel). I've just given it a whirl on Linux, but I'm getting some test failures. It looks like we have at least one bug in Dawn's Vulkan backend that is causing issues, which I'll try and get sorted.

Is macOS enough coverage to get started with? If not I'll try and get everything working on Linux as soon as I can.

src/runtime/README_webgpu.md
```diff
@@ -0,0 +1,97 @@
...
+    cp scripts/standalone-with-node.gclient .gclient
+
+    # Fetch external dependencies and toolchains with gclient
+    gclient sync
+
+Finally, build Dawn, enabling both the Node.js bindings and shared libraries:
+
+    mkdir -p <build_dir>
+    cd <build_dir>
+
```
shoaibkamil, line 83: For this step, at least for me, I needed to install a `go` compiler.

src/runtime/README_webgpu.md
```diff
@@ -0,0 +1,97 @@
...
+    cmake <dawn_root_dir> -G Ninja
+        -DCMAKE_BUILD_TYPE=Release \
+        -DDAWN_BUILD_NODE_BINDINGS=1 \
+        -DDAWN_ENABLE_PIC=1 \
+        -DBUILD_SHARED_LIBS=ON
+
+    ninja dawn.node webgpu_dawn
```
shoaibkamil, line 90: @jrprice I believe the manual patch I had to make (for CMake to pick up libdawn) is now merged upstream?

jrprice: Correct, the CMake build should now work without any additional changes.

src/runtime/README_webgpu.md
jrprice: I'll add a note about this dependency.

dependencies/wasm/CMakeLists.txt
```diff
@@ -106,8 +106,22 @@ function(add_wasm_executable TARGET)
...
 
+    if ("${Halide_TARGET}" MATCHES "webgpu")
+        set(EMCC_FLAGS
+            ${EMCC_FLAGS}
+            -s USE_WEBGPU=1
+            -s ASYNCIFY
+        )
+    endif ()
+
+    # TODO: Remove this when Emscripten and Dawn implement the same APIs.
```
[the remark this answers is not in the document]
jrprice: Opened #7248, will add a link in the comment.

src/JITModule.cpp
```diff
@@ -115,6 +115,38 @@ void load_metal() {
...
+        error.clear();
+        llvm::sys::DynamicLibrary::LoadLibraryPermanently(libname, &error);
+        if (error.empty()) {
+            debug(1) << "found!\n";
+            break;
+        } else {
+            debug(1) << "not found.\n";
+        }
+    }
+    user_assert(error.empty()) << "Could not find a native WebGPU library\n";
```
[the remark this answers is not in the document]
jrprice: Done.

src/LLVM_Runtime_Linker.cpp
```diff
@@ -1202,6 +1203,13 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
                 user_error << "Direct3D 12 can only be used on ARM or X86 architectures.\n";
             }
         }
+        if (t.has_feature(Target::WebGPU)) {
+            if (t.os == Target::Windows) {
+                user_error << "WebGPU runtime not yet supported on Windows.\n";
```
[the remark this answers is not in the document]
jrprice: It might "just work", but I don't personally have a Windows machine to test it on. Opened an issue and added a link.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+        .label = nullptr,
+        .requiredFeaturesCount = 0,
+        .requiredFeatures = nullptr,
+        .requiredLimits = &requestedLimits,
+    };
+    wgpuAdapterRequestDevice(adapter, &desc, request_device_callback,
+                             user_context);
+}
+
+size_t roundUpToMultipleOf4(size_t x) {
```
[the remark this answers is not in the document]
jrprice: Done.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+    uint32_t uniform_size = 0;
+    while (args[num_args] != nullptr) {
+        if (arg_is_buffer[num_args]) {
+            num_buffers++;
+        } else {
+            uint32_t arg_size = arg_types[num_args].bytes();
+            halide_abort_if_false(user_context, arg_size <= 4);
+
+            // Round up to 4 bytes.
+            arg_size = (arg_size + 3) & ~3;
```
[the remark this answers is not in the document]
jrprice: Done.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+
+    wgpuBufferRelease(handle->buffer);
+    free(handle);
+    buf->device = 0;
+
+    return halide_error_code_success;
+}
+
+WEAK int halide_webgpu_wrap_native(void *user_context, struct halide_buffer_t *buf, uint64_t mem) {
+    // TODO: Implement this.
```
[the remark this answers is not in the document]
jrprice: Done.

src/WasmExecutor.cpp
```diff
@@ -2306,6 +2306,7 @@ WasmModuleContents::WasmModuleContents(
 
 #if WITH_WABT
     user_assert(!target.has_feature(Target::WasmThreads)) << "wasm_threads requires Emscripten (or a similar compiler); it will never be supported under JIT.";
+    user_assert(!target.has_feature(Target::WebGPU)) << "wasm_webgpu requires Emscripten (or a similar compiler); it will never be supported under JIT.";
```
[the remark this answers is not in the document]
jrprice: Yes, Emscripten provides an implementation of the native WebGPU API that maps to the browser's JavaScript implementation:
https://github.com/emscripten-core/emscripten/blob/main/src/library_webgpu.js

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+        // Write the argument values to the uniform buffer.
+        uint32_t *arg_values =
+            (uint32_t *)wgpuBufferGetMappedRange(arg_buffer, 0, uniform_size);
+        for (uint32_t a = 0, i = 0; a < num_args; a++) {
+            if (arg_is_buffer[a]) {
+                continue;
+            }
+
+            halide_type_t arg_type = arg_types[a];
+            halide_abort_if_false(user_context, arg_type.lanes == 1);
```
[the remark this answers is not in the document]
jrprice: Changed a bunch of these to `halide_debug_assert()`.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+            supportedLimits.limits.maxBufferSize;
+        requestedLimits.limits.maxStorageBufferBindingSize =
+            supportedLimits.limits.maxStorageBufferBindingSize;
+        requestedLimits.limits.maxComputeWorkgroupStorageSize =
+            supportedLimits.limits.maxComputeWorkgroupStorageSize;
+    }
+#endif
+
+    WGPUDeviceDescriptor desc = {
+        .nextInChain = nullptr,
```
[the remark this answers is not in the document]
jrprice: That's very sad. I'll stop using them :-(

steven-johnson:
Hi there -- I'm (finally) looking at getting testing in place, so we can land this, and some of the comments from your original post may or may not be out of date since late 2021:

For testing, the AOT generator tests can be made to work for wasm-32-wasmrt-webgpu when using Node bindings for Dawn. Both the AOT tests and JIT tests can also work for host-webgpu when using Dawn. 

> JIT is not currently supported for wasm-32-wasmrt-webgpu.

I assume this is still the case?

> Unlike the other GPU backends, I'm not employing the dlopen/dlsym approach in the runtime for getting the API functions. I'm not sure how to make this work when using Emscripten, and since using Dawn directly is only really needed for testing purposes it doesn't seem /too/ onerous to require direct linking, but I'm open to opinions and suggestions here.

IIUC, Dawn is both a native library (for the C++ API) but also available integrated into Node.js. Testing with just a native library is easier in Halide's world (and also the only way to test JIT stuff)... I presume we are probably going to need to do at least some testing in the Emscripten/Node world too (unless you say otherwise).

Direct linking to Dawn is indeed something that may be painful to do, for various reasons, but let me actually try it out before I worry any more.

> Another pain point right now is that the C++ API for WebGPU is not currently stable between different implementations, so there is a build-time switch to toggle between targeting Emscripten vs Dawn. I'm optimistic that this requirement will eventually go away.

Over a year later, is this still accurate, or has the C++ API settled down?

> There's still some patches that need to land in both Emscripten and Dawn before this backend will work for anyone else, so I'll leave this PR as a draft until those are resolved.

Any idea if this is still the case?

> When invoking `emcc` to link Halide-generated objects, include these flags: `-s USE_WEBGPU=1 -s ASYNCIFY`.

steven-johnson:
FYI: I took the liberty of adding 'support' for isinf(), isnan(), and isfinite(), so that Halide code that uses these won't fail to compile outright. (I realize the spec explicitly says that nan/inf may or may not be supported, and AFAICT the current Dawn implementation on Mac definitely does not -- even passing in a buffer prepopulated with NaN values will get normalized into zeros -- but use of these functions is caveat emptor already, in that the caller must know that they are on a target that supports them). LMK your thoughts.

steven-johnson:
As of now, running the tests with Dawn-native (ie host-webgpu), the only tests failing are:

```
	234 - correctness_multi_way_select (Subprocess aborted)
	518 - performance_async_gpu (Failed)
	520 - performance_boundary_conditions (Failed)
	588 - generator_aot_gpu_multi_context_threaded (SEGFAULT)
	589 - generator_aotcpp_gpu_multi_context_threaded (Subprocess aborted)
	674 - python_tutorial_lesson_10_aot_compilation_run (Failed)
```

I'm going to take a quick look to see if any of these are things that have obvious fixes, then move on to testing with the Emscripten setup.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+// - A call to halide_webgpu_acquire_context is followed by a matching call to
+//   halide_webgpu_release_context. halide_webgpu_acquire_context should block
+//   while a previous call (if any) has not yet been released via
+//   halide_webgpu_release_context.
+WEAK int halide_webgpu_acquire_context(void *user_context,
+                                       WGPUInstance *instance_ret,
+                                       WGPUAdapter *adapter_ret,
+                                       WGPUDevice *device_ret,
+                                       bool create = true) {
+    halide_abort_if_false(user_context, &context_lock != nullptr);
```
steven-johnson, line 73: There's lots of calls to `halide_abort_if_false()` in existing code, but IMHO we should really avoid it in new code (and ~most of the existing usage should probably be changed) -- we really don't want to `abort()` in runtime code unless we know the error is **unrecoverable**.  In this specific case, someone would have to override the WEAK `context_lock` symbol to make it a null value, which is in the category of "don't do that" -- i.e., I don't see why we should ever check for that sort of pathological condition at runtime.

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+
+namespace {
+
+halide_error_code_t init_error_code = halide_error_code_success;
+
+void device_lost_callback(WGPUDeviceLostReason reason,
+                          char const *message,
+                          void *user_context) {
+    error(user_context) << "WGPU device lost (" << reason << "): "
+                        << message << "\n";
```
steven-johnson, line 229: Error handling in the Halide runtime currently requires that you call `halide_error()` (or use `error()`, which does that for you), **and** return a nonzero value to the (eventual) caller -- in most production code, `halide_error()` is replaced to simply log the message (and not call `abort()`, etc), so this error would get dropped on the floor in those cases.

This appears to be the only place here where we don't return a nonzero value (eventually)... but it's not clear to me how and where we would need to do that. (Set another global that is checked... somewhere?)

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+    int error_code = error_scope.wait();
+    if (error_code != halide_error_code_success) {
+        return error_code;
+    }
+
+    while (__atomic_test_and_set(&result.complete, __ATOMIC_ACQUIRE)) {
+        wgpuDeviceTick(context.device);
+    }
+
+    return result.status == WGPUQueueWorkDoneStatus_Success ?
```
steven-johnson, line 459: You should also call `halide_error()` or `error()` if the result isn't `halide_error_code_success` (yes, this is regrettably hard to get right -- see https://github.com/halide/Halide/issues/7350)

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+}
+
+void request_device_callback(WGPURequestDeviceStatus status,
+                             WGPUDevice device,
+                             char const *message,
+                             void *user_context) {
+    if (status != WGPURequestDeviceStatus_Success) {
+        debug(user_context) << "wgpuAdapterRequestDevice failed ("
+                            << status << "): " << message << "\n";
+        init_error_code = halide_error_code_generic_error;
```
steven-johnson, line 239: You should also call `halide_error()` here

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+            },
+            &result);
+
+        while (__atomic_test_and_set(&result.map_complete, __ATOMIC_ACQUIRE)) {
+            wgpuDeviceTick(context->device);
+        }
+        if (result.map_status != WGPUBufferMapAsyncStatus_Success) {
+            debug(user_context) << "wgpuBufferMapAsync failed: "
+                                << result.map_status << "\n";
+            return halide_error_code_copy_to_host_failed;
```
steven-johnson, line 570: also call `error()` etc

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+}
+
+int do_multidimensional_copy(void *user_context, WgpuContext *context,
+                             const device_copy &c,
+                             int64_t src_idx, int64_t dst_idx,
+                             int d, bool from_host, bool to_host) {
+    if (d > MAX_COPY_DIMS) {
+        error(user_context)
+            << "Buffer has too many dimensions to copy to/from GPU\n";
+        return -1;
```
steven-johnson, line 590: Prefer returning a halide_error_code

steven-johnson:
So it looks to me like this backend isn't currently safe to use from multiple threads -- we store device (etc) in a global, but competing threads can overwrite this value. Am I missing something, or is this just an oversight for a first draft?

steven-johnson:
Please take a look at https://github.com/jrprice/Halide/pull/1 at your convenience.

steven-johnson:
What's the technical reason that we can't support WebGPU under the JIT?

jrprice:
> What's the technical reason that we can't support WebGPU under the JIT?

The WebGPU runtime code requires Emscripten when targeting WASM in order to translate calls to the native WebGPU runtime API into the Javascript API. I think this is a similar requirement to `WasmThreads`.

steven-johnson:
> The WebGPU runtime code requires Emscripten when targeting WASM in order to translate calls to the native WebGPU runtime API into the Javascript API. 

Wait, so the WebGPU runtime code relies on Emscripten? Does this mean that (e.g.) Chrome has to have some of Emscripten baked into it to make this work?

> I think this is a similar requirement to `WasmThreads`.

The WasmThreads stuff is because Wasm didn't have any real threading support at all, but Emscripten added pthreads wrappers to make it work. Halide didn't want to reinvent that wheel for JIT testing.

jrprice:
> Wait, so the WebGPU runtime code relies on Emscripten? Does this mean that (e.g.) Chrome has to have some of Emscripten baked into it to make this work?

No, I mean that the WASM object that Halide generates needs to be compiled with Emscripten, to link the native WebGPU APIs calls against Emscripten's implementation of WebGPU, which forwards those calls to the Javascript API. Sorry if I'm not explaining this right, I'm not hugely familiar with all this WASM/Emscripten stuff in general.

> The WasmThreads stuff is because Wasm didn't have any real threading support at all, but Emscripten added pthreads wrappers to make it work.

I think this is similar to what I'm describing above? You need something to provide an implementation of the WebGPU APIs, and Emscripten is that thing when we target WASM. In theory you could do it manually, as you suggest, though you'd also need to yield control back to the browser for the async stuff (which is what we currently use Emscripten's ASYNCIFY for).

steven-johnson:
OK, now I'm testing the AOT tests using Emscripten/Node (building with HL_TARGET=wasm-32-wasmrt-webgpu and WEBGPU_NODE_BINDINGS set to the right path), and some (but not all) of the tests fail with `ReferenceError: navigator is not defined`. Investigating, but if you've seen this before...

steven-johnson:
So, running the generator tests under Node, most of them pass, but these three fail:

  548 - generator_aot_acquire_release (Failed)
  549 - generator_aot_alias (Failed)
  565 - generator_aot_gpu_only (Failed)

In all cases, we're getting what looks like garbage in the output buffer -- my first guess would be that something in copy_to_host is broken, but not sure how or why. Thoughts or suggestions welcome.

jrprice:
Right, I'm seeing the same thing for `generator_aot_gpu_only` on my machine (didn't test the other two). This definitely used to pass. I will set aside some time to investigate this early next week (nothing obvious jumps out at me right now).

src/runtime/webgpu.cpp
```diff
@@ -0,0 +1,1146 @@
...
+WEAK WGPUInstance global_instance = nullptr;
+WEAK WGPUAdapter global_adapter = nullptr;
+WEAK WGPUDevice global_device = nullptr;
+// Lock to synchronize access to the global WebGPU context.
+volatile ScopedSpinLock::AtomicFlag WEAK context_lock = 0;
+
+// Size of the staging buffer used for host<->device copies.
+constexpr int kWebGpuStagingBufferSize = 4 * 1024 * 1024;
+// A staging buffer used for host<->device copies.
+WGPUBuffer WEAK staging_buffer = nullptr;
```
steven-johnson, line 29: Should access to this be protected with a mutex? Seems like you could have a situation where two host threads were in contention to use this for copy_to_host, etc

jrprice:
The `copy_to_host` bug was in Dawn's Node bindings. I've just landed a fix in Dawn here:
https://dawn-review.googlesource.com/c/dawn/+/121820
With this change, `generator_aot_gpu_only` passes again for me, along with `generator_aot_acquire_release` and `generator_aot_alias`.

src/runtime/README_webgpu.md
```diff
@@ -0,0 +1,100 @@
+# WebGPU support for Halide
```
steven-johnson, line 1: This README should probably me moved to the same location as all the other READMEs

steven-johnson:
> The `copy_to_host` bug was in Dawn's Node bindings. I've just landed a fix in Dawn here: https://dawn-review.googlesource.com/c/dawn/+/121820 With this change, `generator_aot_gpu_only` passes again for me, along with `generator_aot_acquire_release` and `generator_aot_alias`.

So if I rebuild Dawn at top-of-tree, this should work?

(Related: does Dawn have release version(s) at this point? It would be nice if the README here could say "works as of release tag X")

src/runtime/webgpu.cpp
jrprice: I've added a global flag to capture device lost, which is checked on the next `acquire_context`.

jrprice:
> So if I rebuild Dawn at top-of-tree, this should work?

Yes. Please let me know if it doesn't!

> (Related: does Dawn have release version(s) at this point? It would be nice if the README here could say "works as of release tag X")

The closest we have at the moment are release branches that correspond to Chrome releases, e.g. [chromium/5623](https://dawn.googlesource.com/dawn/+/refs/heads/chromium/5623). But for now I've just added a note to the README that says "tested with top-of-tree Dawn as of DATE" (I already had a similar note for Emscripten+Chrome).

src/runtime/webgpu.cpp
steven-johnson: > Should I replace them all with halide_debug_assert()

I'll leave that up to you -- what you have is no worse than other backends currently, it's just that we have developed an antipattern. (I have on my task list "go examine all the halide_abort_if_false and convert to either halide_debug_assert or actual error", but that's gonna be a gradual background task most likely.)

>  should we not abort even in debug mode?

I'd argue that for the Halide runtime, we shouldn't ever call abort() unless we know the system is in a compromised state (e.g. memory corruption). (Note that the default implementation of `halide_error()` *does* call abort(), but it's easy to override and almost always done so for production software.) So, it's a judgement call on your part, but for this backend, I'm guessing there aren't a lot of obvious places where abort() is the right answer.

steven-johnson:
OK, really close now! The only things left I see failing:

	233 - correctness_multi_way_select (Subprocess aborted)

This ends in "WGPU device lost" -- looks like a real (if obscure) bug.

	587 - generator_aot_gpu_multi_context_threaded (Subprocess aborted)
	588 - generator_aotcpp_gpu_multi_context_threaded (Subprocess aborted)

(basically the same test, just one with LLVM backend, one with C++ backend)
this is one that I already looked into and punted on -- there's clearly a threading issue of some sort, but not sure what.

jrprice:
> ```
> 233 - correctness_multi_way_select (Subprocess aborted)
> ```
> 
> This ends in "WGPU device lost" -- looks like a real (if obscure) bug.

Which platform/device is this on? This test is passing for me on macOS on Apple M1, AMD, and Intel GPUs, and also on a Linux/Vulkan device.
 
> ```
> 587 - generator_aot_gpu_multi_context_threaded (Subprocess aborted)
> 588 - generator_aotcpp_gpu_multi_context_threaded (Subprocess aborted)
> ```

I'll take a look at this one shortly.

steven-johnson:
> Which platform/device is this on? This test is passing for me on macOS on Apple M1, AMD, and Intel GPUs, and also on a Linux/Vulkan device.

This is on my x86-64 Mac laptop, running top-of-tree Dawn (and dawn.node) built as of yesterday. Hmm... Let me doublecheck that I was running with `HL_JIT_TARGET=host-webgpu` rather than `HL_JIT_TARGET=wasm-32-wasmrt-webgpu` (though, if this was wrong, ~all of the correcness_xxx tests would be failing, not just this one)

src/runtime/webgpu.cpp
jrprice: OK I see what you're getting at now. Yes, for the `gpu_multi_context_threaded` tests, we need to create a staging buffer for each context to avoid the contention.

jrprice:
I've made the staging buffer per-context, which fixes `generator_aot_gpu_multi_context_threaded` for me. Please take a look at https://github.com/halide/Halide/pull/6492/commits/2e88fe209c5b58cb1a4cae78c33a8de7c17ab454.

steven-johnson:
Looks good. correctness_multi_way_select still fails for me, full output:

```
Entering Pipeline f2
Target: x86-64-osx-avx-avx2-debug-f16c-fma-jit-sse41-user_context-webgpu
 Input (void const *) __user_context: 0x7ff7b596d5f0
 Output Buffer f2: buffer(0, 0x0, 0x0, 0, uint32)
WGPU: halide_webgpu_initialize_kernels (user_context: 0x7ff7b596d5f0, state_ptr: 0x10a76a6e0, program: 0x10a769000, size: 5143)
Caching compiled kernel: 0x7fde16706510 id 2 context 0x7fde4680fe00
Exiting Pipeline f2
WGPU: halide_webgpu_finalize_kernels (user_context: 0x7ff7b596d5f0, state_ptr: 0x2
Entering Pipeline f2
Target: x86-64-osx-avx-avx2-debug-f16c-fma-jit-sse41-user_context-webgpu
 Input (void const *) __user_context: 0x7ff7b596d5f0
 Output Buffer f2: buffer(0, 0x0, 0x7fde16706980, 0, uint32)
WGPU: halide_webgpu_initialize_kernels (user_context: 0x7ff7b596d5f0, state_ptr: 0x10a76a6e0, program: 0x10a769000, size: 5143)
WGPU: halide_webgpu_device_malloc (user_context: 0x7ff7b596d5f0, buf: 0x600002c38910)
      Allocated device buffer 0x600001920900
WGPU: halide_webgpu_run (user_context: 0x7ff7b596d5f0, entry: _kernel_f2_s0___outermost___outermost_v2___block_id_x, groups: 1x1x1, threads: 1x1x1, workgroup_mem: 0
WGPU: halide_webgpu_finalize_kernels (user_context: 0x7ff7b596d5f0, state_ptr: 0x2
Error: WGPU device lost (0): Error creating pipeline state Compiler encountered an internal error
 - While handling unexpected error type Internal when allowed errors are (Validation|DeviceLost).
    at Initialize (/Users/srj/GitHub/dawn/src/dawn/native/metal/ComputePipelineMTL.mm:54)
    at CreateComputePipeline (/Users/srj/GitHub/dawn/src/dawn/native/Device.cpp:1473)
WGPU: unknown error (5): GPU device disconnected
WGPU: unknown error (5): GPU device disconnected
```

steven-johnson:
Also, now some of the Generators tests (running under Node) now fail with:

RuntimeError: Aborted(RuntimeError: unreachable). "unreachable" may be due to ASYNCIFY_STACK_SIZE not being large enough (try increasing it)
    at abort (/Users/srj/GitHub/Halide/build/release/test/generator/generator_aot_cleanup_on_error.js:1:13398)
    at runAndAbortIfError (/Users/srj/GitHub/Halide/build/release/test/generator/generator_aot_cleanup_on_error.js:1:49522)
    at /Users/srj/GitHub/Halide/build/release/test/generator/generator_aot_cleanup_on_error.js:1:53420
    at callUserCallback (/Users/srj/GitHub/Halide/build/release/test/generator/generator_aot_cleanup_on_error.js:1:22442)
    at Timeout._onTimeout (/Users/srj/GitHub/Halide/build/release/test/generator/generator_aot_cleanup_on_error.js:1:22545)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)

jrprice:
> correctness_multi_way_select still fails for me
> ...
> `Error creating pipeline state Compiler encountered an internal error`

Ugh, this is an internal error inside Apple's Metal Shading Language compiler. Are you using an AMD GPU on macOS 12.x or earlier by any chance? We hit a particular bug with this combo a lot, and there's unfortunately not a lot we can do about it. If it's the same one I'm thinking about then it did seem to be fixed in macOS 13.x though.

> Also, now some of the Generators tests (running under Node) now fail with:
> 
> RuntimeError: Aborted(RuntimeError: unreachable). "unreachable" may be due to ASYNCIFY_STACK_SIZE not being large enough (try increasing it)

Should be fixed now. I was doing async stuff inside an async callback 🙃

steven-johnson:
> Ugh, this is an internal error inside Apple's Metal Shading Language compiler. Are you using an AMD GPU on macOS 12.x or earlier by any chance? 

...why, yes, I am. (I didn't even know Apple shipped Intel macs with AMD GPUs...)

If this is fixed in macOS 13.x then we can just ignore it, I guess.

steven-johnson:
> > Ugh, this is an internal error inside Apple's Metal Shading Language compiler. Are you using an AMD GPU on macOS 12.x or earlier by any chance?
> 
> ...why, yes, I am. (I didn't even know Apple shipped Intel macs with AMD GPUs...)
> 
> If this is fixed in macOS 13.x then we can just ignore it, I guess.

Wait, not so fast: Unfortunately, our only Mac x86 buildbot is a Late-2013 Mac Pro... and pre-2017 Macs can't be upgraded to Ventura. Obviously we shouldn't let that prevent us from landing this but we're gonna need to figure out how to get better test coverage. (The gotcha is where to find a fast x86 Mac we can replace it with...)

steven-johnson (approved):
OK, I'm declaring this Ready Enough To Land, at least in a known-beta-quality state.

Other folks on the review line, please either complete your review (or remove yourself from reviewership) at your earliest convenience so we can land this according to protocol :-)

src/runtime/CMakeLists.txt
```diff
@@ -154,6 +156,7 @@ set(RUNTIME_CXX_FLAGS
     # Necessary for using virtual functions in the runtime code.
     -fno-rtti
     -Wall
+    -Wc++20-designator
```
shoaibkamil, line 159: Does this mean C++20 is required for our runtime now?  IIRC, we have so far stuck with C++17 or earlier.

jrprice: No, this is adding the warning to make sure we don't accidentally use C++20 designators (which I had done earlier in the PR).

shoaibkamil (approved):
Thanks for this huge chunk of work!

PR #6494, opened 2021-12-11
rounding shift rights should use rounding halving add

abadams:
On x86 currently we lower cast<uint8_t>((cast<uint16_t>(x) + 8) / 16)
to:

cast<uint8_t>(shift_right(widening_add(x, 8), 4))

This compiles to 8 instructions on x86: Widen each half of the input
vector, add 8 to each half-vector, shift each half-vector, then narrow
each half-vector.

First, this should have been a rounding_shift_right. Some patterns were
missing in FindIntrinsics.

Second, rounding_shift_right had suboptimal codegen in the case where
the second arg is a positive const. On archs without a rounding shift
right instruction you can further rewrite this to:

shift_right(rounding_halving_add(x, 7), 3)

which is just two instructions on x86.

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -174,6 +174,7 @@ protected:
     IRMatcher::Wild<1> y;
     IRMatcher::Wild<2> z;
     IRMatcher::WildConst<0> c0;
+    IRMatcher::WildConst<1> c1;
 
     Expr visit(const Add *op) override {
         if (!find_intrinsics_for_type(op->type)) {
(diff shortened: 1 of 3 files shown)


abadams:
Dillon, I have a question for you in the IRMatch.h code. I'm a bit confused about the types of the second arg to shifts.

src/IRMatch.h
```diff
@@ -1483,7 +1483,48 @@ struct Intrin {
...
 
-    constexpr static bool foldable = false;
+    constexpr static bool foldable = true;
+
+    HALIDE_ALWAYS_INLINE void make_folded_const(halide_scalar_value_t &val, halide_type_t &ty, MatcherState &state) const noexcept {
+        halide_scalar_value_t arg1;
+        // Assuming the args have the same type as the intrinsic is incorrect in
+        // general. But for the intrinsics we can fold (just shifts), the LHS
+        // has the same type as the intrinsic, and we can always treat the RHS
+        // as a signed int, because we're using 64 bits for it. (TODO: Dillon,
```
dsharletg, line 1493: Yes, I think this is OK. Constants will be simplified to unsigned shifts if needed.

test/correctness/simd_op_check.cpp
```diff
@@ -150,6 +150,13 @@ class SimdOpCheck : public SimdOpCheckTest {
             check("pavgb", 8 * w, u8((u16(u8_1) + u16(u8_2) + 1) >> 1));
             check("pavgw", 4 * w, u16((u32(u16_1) + u32(u16_2) + 1) / 2));
             check("pavgw", 4 * w, u16((u32(u16_1) + u32(u16_2) + 1) >> 1));
+
+            // Rounding right shifts should also use pavg
+            check("pavgb", 8 * w, u8((u16(u8_1) + 15) >> 4));
```
dsharletg, line 155: I think these added checks should be in test/correctness/intrinsics.cpp instead (much lighter weight test than simd_op_check and it is target independent).

src/FindIntrinsics.cpp
```diff
@@ -383,43 +384,129 @@ class FindIntrinsics : public IRMutator {
...
-                rewrite(halving_add(widening_sub(x, y), 1), rounding_halving_sub(x, y), is_x_same_int_or_uint) ||
-                rewrite(rounding_shift_right(widening_add(x, y), 1), rounding_halving_add(x, y), is_x_same_int_or_uint) ||
-                rewrite(rounding_shift_right(widening_sub(x, y), 1), rounding_halving_sub(x, y), is_x_same_int_or_uint) ||
-
-                rewrite(max(min(shift_right(widening_mul(x, y), z), upper), lower), mul_shift_right(x, y, cast(unsigned_type, z)), is_x_same_int_or_uint && is_uint(z)) ||
-                rewrite(max(min(rounding_shift_right(widening_mul(x, y), z), upper), lower), rounding_mul_shift_right(x, y, cast(unsigned_type, z)), is_x_same_int_or_uint && is_uint(z)) ||
-                rewrite(min(shift_right(widening_mul(x, y), z), upper), mul_shift_right(x, y, cast(unsigned_type, z)), is_x_same_uint && is_uint(z)) ||
-                rewrite(min(rounding_shift_right(widening_mul(x, y), z), upper), rounding_mul_shift_right(x, y, cast(unsigned_type, z)), is_x_same_uint && is_uint(z)) ||
+            if (
+                // Saturating patterns
```
dsharletg, line 388: Side comment, the reformatting here makes this change a lot harder to review easily.

dsharletg (approved):

test/correctness/simd_op_check.cpp
abadams: It's a test that rounding_shift_right is both caught by pattern matching and also lowers correctly on x86, so I'll move half of it to intrinsics

dsharletg: Aren't the two tests below this the same test after accounting for the pattern matching part?

abadams: Hopefully the latest commit makes it clearer. The first two tests aren't just a rounding shift right. they strength-reduce to an average followed by a shift right. So it tests the new patterns.

The second test checks lower_rounding_shift_right does the right thing.

dsharletg: Oh right, I hadn't noticed the constant isn't a rounding_shift_right. Still, I think the change you just made is good (easier tests to debug if they fail).

src/FindIntrinsics.cpp
steven-johnson: +1, may I suggest restoring original format and putting `clang-format off` back in place? If we want to let these be 'naturally' reformatted then IMHO we should do that in a standalone PR

abadams: I will in future, but here I just manually reformatted a few surrounding lines because they were hard to modify. What I'd done happened to agree with clang-format, so I also removed those comments.

Issue #6474, opened 2021-12-06
`error()` is a large proportion of our runtime size


steven-johnson:
While working on https://github.com/halide/Halide/pull/6473, I wanted to see how much code+constdata space we were using  for error reporting in our runtime, so I tried making `error` an alias for `SinkPrinter`. 

Comparing the object size for `runtime.o` built with `host-opencl` on OSX, I get:
- recent master: 180392 bytes
- with #6473 applied: 164592 bytes
- with error() completely nuked via SinkPrinter: 140984

That's a surprisingly large percentage of our runtime. While I don't think runtime size is (currently) a high priority, I suspect this could be reduced substantially by some combination of:
- Calling existing `halide_error_foo()` helper functions instead of assembling a bespoke error message
- Selectively moving some verbose error messages into DEBUG_RUNTIME only

PR #6502, opened 2021-12-16
Sketch of partial fix for #6474

steven-johnson:
This makes an assumption that ~all of the calls to `error()` in the runtime have descriptive text that is generally only useful for developers, and leaving this text in release builds is of limited-to-no-use. On that assumption:
-Add a new "runtime internal" error code that is a catch-all for these cases
- Add a new `_halide_runtime_error()` error function, which is a smart wrapper that discards all its arguments in non-debug runtimes
- Convert runtime/cuda.cpp as a proof-of-concept of possible code savings. On My OSX box, `bin/host-cuda/runtime.a` is 174128 bytes at current top-of-tree, 163160 bytes with this PR in place. Extending this to the rest of runtime would likely get us down to ~140k if the estimates in #6474 are correct.

Note #1: You could achieve (nearly) the same thing by just changing `error()` to be special-case for DEBUG_RUNTIME, but this formulation is (IMHO) slightly cleaner, since it also allows us to return the error result directly, rather than requiring two statements. It also provides a good excuse to do a once-over of all existing usage, which is probably worthwhile.

- As mentioned above, it basically drops useful text on the floor for release builds, on the assumption that developers can use a debug-runtime build for more details; this may be a terrible assumption. Thoughts?

- This PR makes no attempt to address the really-quite-loose bounds on what can be returned; e.g. there are lots of places we just return a Cuda error where (technically) a halide_error_code_t is expected; this doesn't seem to ever have been a real problem in practice, but it makes my spidey-sense tingle.

### DIFF
--- a/src/runtime/HalideRuntime.h
+++ b/src/runtime/HalideRuntime.h
@@ -1177,6 +1177,9 @@ enum halide_error_code_t {
     /** An explicit storage bound provided is too small to store
      * all the values produced by the function. */
     halide_error_code_storage_bound_too_small = -45,
+
+    /** Catch-all internal error code used by the runtime. */
+    halide_error_code_runtime_internal = -46,
 };
 
 /** Halide calls the functions below on various error conditions. The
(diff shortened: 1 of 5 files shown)


abadams:
I'm thinking that when we get error reports from users, it will help us a lot if they can include the error message. Is it really necessary to make the runtime smaller in this way?

steven-johnson:
> Is it really necessary to make the runtime smaller in this way?

Nope. It was just on my list of stuff to look at and seemed like something simple to look at before taking the holidays off.

PR #6504, opened 2021-12-18
Make random faster by putting the innermost var last

abadams:
40% faster, and I fixed the issue where the low bits of our noise have a low period. This combined with #6506 makes it possible to generate low-bit-width-noise very cheaply for things like dithering.

### DIFF
--- a/src/Random.cpp
+++ b/src/Random.cpp
@@ -82,6 +82,10 @@ Expr random_int(const vector<Expr> &e) {
                                rng32(Variable::make(UInt(32), name)));
         }
     }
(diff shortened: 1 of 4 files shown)


abadams:
The new simplifier rules have been formally verified

steven-johnson (approved):

steven-johnson:
FWIW, the new simplification rules now cause errors of the form `Signed integer overflow occurred during constant-folding. Signed integer overflow for int32 and int64 is undefined behavior in Halide` in code that has never failed in this way before -- I'd like to revert this change pending further investigation.

PR #6505, opened 2021-12-19
Support new warp shuffle intrinsics after CUDA Volta architecture

### LINKED ISSUE #5630: warp shuffle intrinsics no longer work with cuda compute capability 8.0
We need to use the .sync variants starting from that version

jinderek:
This PR fixes #5630 
We need to add ".sync" suffix for warp shuffle intrinsics after Volta architecture:
https://docs.nvidia.com/cuda/volta-tuning-guide/index.html

### DIFF
--- a/src/LowerWarpShuffles.cpp
+++ b/src/LowerWarpShuffles.cpp
@@ -368,6 +368,7 @@ class LowerWarpShuffles : public IRMutator {
     };
     Scope<AllocInfo> allocation_info;
     Scope<Interval> bounds;
+    int cuda_cap;
 
     Stmt visit(const For *op) override {
         ScopedBinding<Interval>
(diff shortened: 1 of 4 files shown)


test/correctness/register_shuffle.cpp
```diff
@@ -185,6 +183,53 @@ int main(int argc, char **argv) {
...
+                printf("out(%d) = %d instead of %d\n",
+                       y, actual, correct);
+                return -1;
+            }
+        }
+    }
+
+
     {
         // Same as above, but in half-warps
```
abadams, line 234: Is this comment wrong now that you've inserted a new test case above?

abadams:
Thanks for this, I just had one very minor comment.

test/correctness/register_shuffle.cpp
jinderek: Yeah. Fixed.

jinderek:
I think the 1 failing of buildbot is not related to this PR. @abadams @jrk

src/LowerWarpShuffles.cpp
```diff
@@ -562,6 +562,22 @@ class LowerWarpShuffles : public IRMutator {
 
         internal_assert(may_use_warp_shuffle) << name << ", " << idx << ", " << lane << "\n";
 
+        // We must add .sync after volta architecture:
+        // https://docs.nvidia.com/cuda/volta-tuning-guide/index.html
+        string sync_suffix = "";
+        Target t = get_jit_target_from_environment();
```
abadams, line 568: Didn't notice this before. This is not the right target to use here - Halide is commonly used as a cross compiler and the jit target is usually some sort of host target for the build machine. I think you'll need to change lower_warp_shuffles to also accept a Target object and plumb that through to here.

jinderek: Oh, I see. Fixed now.

jinderek:
Hi   @abadams , the code is updated, could you help to have a look? (BTW, I think the 1 failing of buildbot is not related to this PR.)

abadams (approved):

abadams:
Thanks for the PR! Merging.

PR #6506, opened 2021-12-19
Make it possible to interpret a wide type as multiple smaller elements

abadams:
This is helpful for things like reinterpreting 32-bit packed rgba values
as individual components for free.


dsharletg (approved):

steven-johnson:
FYI, this appears to have injection massive failures into at least one Hexagon pipeline inside Google. Investigating, may need to rollback pending investigation.

PR #6508, opened 2021-12-21
Fix a missing case in clamp_unsafe_accesses

abadams:


alexreinking (approved):
Thanks for the fix!

alexreinking:
Looks like the crash must be happening here:

``` 
 ​            ​auto​ bounds = func_bounds.​at​({call->​name​, call->​value_index​});
```

Is there some reason an input image would not be included in the FuncValueBounds results?

abadams:
Yeah, we don't know the value bounds of input images. I pushed something that better reflects what I was trying to do.

alexreinking:
Great, yeah, that looks better

Issue #6513, opened 2021-12-23
Improve sanitizer support


LebedevRI:
Hi! I'd like to make use of halide in RawSpeed library (https://github.com/darktable-org/rawspeed/issues/325).

Now that the halide **is** [packaged in debian](https://tracker.debian.org/pkg/halide), it's mostly a no-brainer,
but unfortunately for me RawSpeed is an [avid user of oss-fuzz](https://github.com/google/oss-fuzz/blob/0b499256611c92aa85d86866bca478825fe66945/projects/librawspeed/project.yaml),
and unfortunately i think not loosing that coverage is more important than making use of Halide.

As far as i can tell, there's some support for ASAN and TSAN instrumentation,
which means there are some missing things still: (list)
* MSAN (should be easy?)
* Sanitizer coverage
* UBSan (it's frontend-based, i'm not sure how this would look, only for the actual user-written C/C++ code)
* UX improvement to just make use of the standard clang `-fsanitize=` flags?

Thoughts?

Blocks https://github.com/darktable-org/rawspeed/issues/325.

steven-johnson:
- MSAN: we have support for MSAN as well, but AOT-only (MSAN is documented as incompatibile with JIT implementations in LLVM)
- What is "Sanitizer Coverage"?
- UBSan is basically all clang at this point, and thus irrelevant for Halide-generated code. We could add such a mode to Halide but first we'd need to explicitly enumerate everything that is known to be UB in the Halide language; I'm not certain if that's well-defined at this point. (FWIW, UBSan sounds great but in practice it's almost impossible to use it with any C++ codebase unless it's built that way from the beginning; even inside Google it's basically admitted to be of limited use because you'll hit so much UB in well-behaved C++ programs, which really tells you something about the state of C++, alas)
- UX improvements I'm all in favor of. 

But yeah, PRs for any and all of this very welcome, sanitizers are great tools and we'd love more support.

See also https://github.com/halide/Halide/issues/6436

LebedevRI:
Ok, sanitizer-wise i think things are actually good now. Thanks!
Minimally, i still miss `-march=` handling of #6515/#6514, will look into that next.

More generally, i think there are some more things potentially missing,
with not all being feasible, but just to list:
* [XRay](https://llvm.org/docs/XRay.html) (trivial, will likely look into that)
* (LLVM-specific) LTO (uh)
* (LLVM-specific) PGO (uh)
* Source-based coverage (probably quite complicated, aside from most basic form)
[...]

PR #6516, opened 2021-12-24
[NFC-ish] Finish MSAN handling (Refs. #6513)

LebedevRI:
Somehow, initially i missed that there was MSan support,
so it might be good to actually mention that we don't need to run
any MSan passes here, and that we didn't forget to run them.

Secondly, it seems inconsistent not annotate the functions
with `Attribute::SanitizeMemory`, like we do for others.

I suppose it isn't strictly required, since they are used
to actually drive the instrumentation passes, and we don't
run MSan pass, but they are also used to disable some LLVM optimizations,
and that //might// be important. Or not, but then i suppose
there should be a comment about it?

Let me know if i should replace the second change with a comment too.

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1171,6 +1171,9 @@ void CodeGen_LLVM::optimize_module() {
         });
     }
 
+    // Target::MSAN handling is sprinkled throughout the codebase,
+    // there is no need to run MemorySanitizerPass here.
+
     if (get_target().has_feature(Target::TSAN)) {
         pb.registerOptimizerLastEPCallback(
             [](ModulePassManager &mpm, OptimizationLevel level) {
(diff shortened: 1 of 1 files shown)


alexreinking (approved):

alexreinking:
Failures are again unrelated... same arm64/llvm14 failure.

alexreinking:
> Secondly, it seems inconsistent not annotate the functions with Attribute::SanitizeMemory, like we do for others.
>
> I suppose it isn't strictly required, since they are used to actually drive the instrumentation passes, and we don't run MSan pass, but they are also used to disable some LLVM optimizations, and that //might// be important. Or not, but then i suppose there should be a comment about it?
>
> Let me know if i should replace the second change with a comment too.

Requesting a review from @steven-johnson (whenever he's back) to address this. IIRC he implemented MSAN support and I'm not sure why this wouldn't be annotated.

steven-johnson (approved):
LGTM pending green buildbots -- I think the arm failure are just stale from bad LLVM revisions, rerunning them now

PR #6517, opened 2021-12-25
Implement SanitizerCoverage support (Refs. #6513)

LebedevRI:
Please refer to https://clang.llvm.org/docs/SanitizerCoverage.html

TLDR: `ModuleSanitizerCoveragePass` instruments the IR by inserting
calls to callbacks at certain constructs. What the callbacks should do
is up to the implementation. They are effectively required for fuzzing
to be effective, and are provided by e.g. libfuzzer.

One huge caveat is `SanitizerCoverageOptions` which controls
which which callbacks should actually be inserted.
I just don't know what to do about it. Right now i have hardcoded
the set that would have been enabled by `-fsanitize=fuzzer-no-link`,
because the alternative, due to halide unflexibility,
would be to introduce ~16 suboptions to control each one.

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -846,7 +846,7 @@ void compile_multitarget(const std::string &fn_name,
             Target::MSAN,
             Target::NoRuntime,
             Target::TSAN,
-            Target::SANCOV,
+            Target::SanitizerCoverage,
             Target::UserContext,
         }};
         for (auto f : must_match_features) {
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -837,7 +837,7 @@ void compile_multitarget(const std::string &fn_name,
             user_error << "All Targets must have matching arch-bits-os for compile_multitarget.\n";
         }
         // Some features must match across all targets.
-        static const std::array<Target::Feature, 9> must_match_features = {{
+        static const std::array<Target::Feature, 10> must_match_features = {{
             Target::ASAN,
             Target::CPlusPlusMangling,
             Target::Debug,
@@ -846,6 +846,7 @@ void compile_multitarget(const std::string &fn_name,
             Target::MSAN,
             Target::NoRuntime,
             Target::TSAN,
+            Target::SANCOV,
             Target::UserContext,
         }};
         for (auto f : must_match_features) {
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -384,7 +384,7 @@ const std::map<std::string, Target::Feature> feature_name_map = {
     {"llvm_large_code_model", Target::LLVMLargeCodeModel},
     {"rvv", Target::RVV},
     {"armv81a", Target::ARMv81a},
-    {"sancov", Target::SANCOV},
+    {"sanitizercoverage", Target::SanitizerCoverage},
     // NOTE: When adding features to this map, be sure to update PyEnums.cpp as well.
 };
 
@@ -422,7 +422,7 @@ Target get_jit_target_from_environment() {
     host.set_feature(Target::TSAN);
 #endif
 #if __has_feature(coverage_sanitizer)
-    host.set_feature(Target::SANCOV);
+    host.set_feature(Target::SANITIZERCOVERAGE);
 #endif
 #endif
     string target = Internal::get_env_variable("HL_JIT_TARGET");
@@ -1008,7 +1008,7 @@ bool Target::get_runtime_compatible_target(const Target &other, Target &result) 
         SoftFloatABI,
         TSAN,
         WasmThreads,
-        SANCOV,
+        SanitizerCoverage,
     }};
     // clang-format on
 
@@ -1037,7 +1037,7 @@ bool Target::get_runtime_compatible_target(const Target &other, Target &result) 
     }
 
     if ((features & matching_mask) != (other.features & matching_mask)) {
-        Internal::debug(1) << "runtime targets must agree on SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX, HexagonDma, HVX_shared_object, SANCOV\n"
+        Internal::debug(1) << "runtime targets must agree on SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX, HexagonDma, HVX_shared_object, SANITIZERCOVERAGE\n"
                            << "  this:  " << *this << "\n"
                            << "  other: " << other << "\n";
         return false;
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -384,7 +384,7 @@ const std::map<std::string, Target::Feature> feature_name_map = {
     {"llvm_large_code_model", Target::LLVMLargeCodeModel},
     {"rvv", Target::RVV},
     {"armv81a", Target::ARMv81a},
-    {"sanitizercoverage", Target::SanitizerCoverage},
+    {"sanitizer_coverage", Target::SanitizerCoverage},
     // NOTE: When adding features to this map, be sure to update PyEnums.cpp as well.
 };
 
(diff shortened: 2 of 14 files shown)


LebedevRI:
Hmm, looks like the test needs to be less fragile. Will look into that.

alexreinking:
Just so you know, there's no need to force push... we always squash-merge our PRs and the history can be useful for review.

alexreinking:
Requesting reviews from @steven-johnson since he's familiar with the sanitizer stuff and from @zvookin since he will have opinions on how to deal with passing options to this.

LebedevRI:
> Requesting reviews from @steven-johnson since he's familiar with the sanitizer stuff and from @zvookin since he will have opinions on how to deal with passing options to this.

Thanks!
Re options: we basically only have two choices:
* just add a single best-guess meta-feature, like done here, that hardcodes the set of coverage to enable,
* or add ~16 features to toggle each option in `SanitizerCoverageOptions` struct.

test/generator/sancov_aottest.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+extern "C" void __sanitizer_cov_trace_pc_indir(uintptr_t Callee) {
+    if (!enable_callbacks) return;
+    printf("Hit __sanitizer_cov_trace_pc_indir. Success!\n");
+}
+
+template<typename T>
+void clear_out(T &image) {
+    image.for_each_element([&](int x, int y, int c) {
```
abadams, line 78: image.fill(-42);

src/Module.cpp
```diff
@@ -846,6 +846,7 @@ void compile_multitarget(const std::string &fn_name,
             Target::MSAN,
             Target::NoRuntime,
             Target::TSAN,
+            Target::SANCOV,
```
steven-johnson, line 849: annoying style nit: all-caps `SANCOV` seems out of place here as an all-caps abbreviation, despite MSAN, etc, since the latter are commonly referred to that way in LLVM docs but the former seems to be `SanitizerCoverage`. If we are using this only for testing purposes then I think we should use either `SanitizerCoverage` or `sanitizer_coverage` to fit in with existing patterns. (Tagging @abadams here for his opinion as well.)

test/generator/sancov_aottest.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+extern "C" void __sanitizer_cov_trace_pc_indir(uintptr_t Callee) {
+    if (!enable_callbacks) return;
+    printf("Hit __sanitizer_cov_trace_pc_indir. Success!\n");
+}
+
+template<typename T>
+void clear_out(T &image) {
+    image.fill(-42);
```
steven-johnson, line 78: Nit: if `T` is an unsigned type, some compilers can complain about assigning a signed literal to it. Either avoid using a negative value or (probably more simply) just always use a signed type here.

steven-johnson:
> One huge caveat is `SanitizerCoverageOptions` which controls which which callbacks should actually be inserted. I just don't know what to do about it. Right now i have hardcoded the set that would have been enabled by `-fsanitize=fuzzer-no-link`, because the alternative, due to halide unflexibility, would be to introduce ~16 suboptions to control each one.

I think that's probably fine for the time being -- we can expand and add flexibility in the future as desired. That said, we should add a tracking issue here ("Target::SanitizerCoverage only supports equivalent of fuzzer-no-link") and add a link to it in comments at the relevant sites in the code.

steven-johnson (changes requested):
Some nits re: style and comments, otherwise LGTM pending green

src/Module.cpp
abadams: IMO SanitizerCoverage to match the llvm docs, but I don't feel strongly about it

LebedevRI: Do you want me to change all instances of SANCOV to SanitizerCoverage, or just some?

> If we are using this only for testing purposes

I'm not sure what you mean by that? Roughly, the feature will be requested the way it is requested in the test.

steven-johnson: > change all instances of SANCOV to SanitizerCoverage, or just some?

I'd vote for all (except the ones in HalideRuntime.h, which should be `sanitizer_coverage` to match the conventions there)

LebedevRI: Ok, let me know if i did it wrong.

src/Target.cpp
```diff
@@ -384,7 +384,7 @@ const std::map<std::string, Target::Feature> feature_name_map = {
     {"llvm_large_code_model", Target::LLVMLargeCodeModel},
     {"rvv", Target::RVV},
     {"armv81a", Target::ARMv81a},
-    {"sancov", Target::SANCOV},
+    {"sanitizercoverage", Target::SanitizerCoverage},
```
steven-johnson, line 387: "sanitizer_coverage", so that it matches the name in HalideRuntime.h

src/Target.cpp
```diff
@@ -422,7 +422,7 @@ Target get_jit_target_from_environment() {
     host.set_feature(Target::TSAN);
 #endif
 #if __has_feature(coverage_sanitizer)
-    host.set_feature(Target::SANCOV);
+    host.set_feature(Target::SANITIZERCOVERAGE);
```
steven-johnson, line 425: This looks like it won't compile, did you mean Target::SanitizerCoverage

src/Target.cpp
```diff
@@ -1037,7 +1037,7 @@ bool Target::get_runtime_compatible_target(const Target &other, Target &result)
     }
 
     if ((features & matching_mask) != (other.features & matching_mask)) {
-        Internal::debug(1) << "runtime targets must agree on SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX, HexagonDma, HVX_shared_object, SANCOV\n"
+        Internal::debug(1) << "runtime targets must agree on SoftFloatABI, Debug, TSAN, ASAN, MSAN, HVX, HexagonDma, HVX_shared_object, SANITIZERCOVERAGE\n"
```
steven-johnson, line 1040: SanitizerCoverage

test/generator/sanitizercoverage_generator.cpp
```diff
@@ -2,12 +2,12 @@
 
 namespace {
 
-class SANCOV : public Halide::Generator<SANCOV> {
+class SANITIZERCOVERAGE : public Halide::Generator<SANITIZERCOVERAGE> {
```
steven-johnson, line 5: s/SANITIZERCOVERAGE/SanitizerCoverage/ or really anything that isn't ALLCAPS

test/generator/sanitizercoverage_generator.cpp
```diff
@@ -2,12 +2,12 @@
...
 public:
     Output<Buffer<int8_t>> output{"output", 3};
 
     void generate() {
-        // Currently the test just exercises Target::SANCOV
+        // Currently the test just exercises Target::SANITIZERCOVERAGE
```
steven-johnson, line 10: Target::SanitizerCoverage

test/generator/sanitizercoverage_generator.cpp
```diff
@@ -16,10 +16,10 @@ class SANCOV : public Halide::Generator<SANCOV> {
     }
 
 private:
-    // Currently the test just exercises Target::SANCOV
+    // Currently the test just exercises Target::SANITIZERCOVERAGE
```
steven-johnson, line 19: Target::SanitizerCoverage

steven-johnson (changes requested):

src/Target.cpp
LebedevRI: I did verify that this builds for me, starting with clean build dir.

Issue #6518, opened 2021-12-27
Add support for not building autoschedulers


LebedevRI:
Lack of such functionality results in e.g. https://buildd.debian.org/status/fetch.php?pkg=halide&arch=riscv64&ver=13.0.2-2&stamp=1640095586&raw=0

Quoting @alexreinking from chat:

> I would gladly accept a PR that adds a `Halide_ENABLE_AUTOSCHEDULERS` build option. It should be dependent on `BUILD_SHARED_LIBS` being true and on by default, as it currently is.
> Note that it's not trivial since several of the tests assume the autoschedulers exist

alexreinking:
> Lack of such functionality results in [...]

Can you explain what is happening here, a bit? Scanning through thousands of log lines to determine the relevant parts is not quick.

LebedevRI:
```
Unhandled exception: Error: The Mullapudi2016 autoscheduler is not supported for the target: riscv-64-linux-no_runtime

```

alexreinking:
I think we should also evaluate why the Mullapudi autoscheduler assert-fails on RISC-V. I'm guessing it wants a CPU target and is missing an option here.

This might be resolved by adding a helper on the target class to distinguish CPUs from accelerators/other targets.

abadams:
IMO the assert is bogus and should just be deleted.

alexreinking:
> IMO the assert is bogus and should just be deleted.

This might actually be a good case for a warning... Mullapudi2016 is not expected to generate good schedules for some targets, but that's not a reason to error out. But we also don't want to leave the opposite impression, right?

zvookin:
Warnings are not a thing.

I can't think of any reason RISC V would be a problem here unless there is some architecture specific data involved, which I don't see on a quick look. (MachineParams is not arch specific, but actual implementation of processor/etc. specific.) If there is some arch specific logic, making RISC V do the same thing as ARM or MIPS is probably fine.

alexreinking:
> Warnings are not a thing.

I don't know what you mean by this. We have dozens of `user_warning`s in our codebase. Is there an ongoing effort to remove all of them?

abadams:
Each one represents a failure to either make something a hard error or not a problem to begin with.

abadams:
[...]

PR #6520, opened 2021-12-28
Mullapudi2016: don't hardcode the list of supported targets

LebedevRI:
As discussed in https://github.com/halide/Halide/issues/6518,
this is a bit dubious, and e.g. prevents building on RISC-V,
because there is no way to not build autoschedulers currently.


zvookin (approved):

alexreinking (approved):
Thanks!

alexreinking:
Failures are unrelated -- same upstream failure (arm64/llvm14)

LebedevRI:
@alexreinking thank you!

PR #6523, opened 2021-12-31
If cmake built a python module, teach cmake to install the python module

Infinoid:
CMake can build the python module, but it's ignored during `make install`.  This teaches cmake to install it, if it was built.

The cmake `FindPython3` module makes some distutils calls to figure out where modules should go, and exposes those as cmake vars.  Specifically, [Python3_STDLIB, Python3_STDARCH, Python3_SITELIB, Python3_SITEARCH](https://cmake.org/cmake/help/v3.15/module/FindPython3.html#result-variables) give us the various cases of "python3/" vs "python3.x", and "dist-modules" vs "site-modules".

This patch uses `Python3_SITEARCH` by default, which maps to `/usr/lib/python3.9/site-packages/` for me on debian linux.  The var `Halide_INSTALL_PYTHONDIR` can be used to override that.  The intention is for distro packaging to point that wherever it needs to go.

### DIFF
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -46,6 +46,18 @@ foreach (dep IN ITEMS Halide_LLVM Halide_wabt)
 endforeach ()
 
 ##
+# Python bindings
+##
+
+if (WITH_PYTHON_BINDINGS)
+    set(Halide_INSTALL_PYTHONDIR "${CMAKE_INSTALL_LIBDIR}/python3/site-packages"
+        CACHE STRING "Path to Halide Python bindings folder")
+    install(TARGETS Halide_Python
+            LIBRARY DESTINATION ${Halide_INSTALL_PYTHONDIR} COMPONENT Halide_Python
+            NAMELINK_COMPONENT Halide_Python)
+endif ()
+
+##
 # Library-type-agnostic interface targets
 ##
 


Infinoid:
I don't know what the right values for `EXPORT`, `COMPONENT` and `NAMELINK_COMPONENT` should be.

I had initially tried using `Halide_Runtime`, similar to other install rules in that file.  But that caused missing target errors when configuring the `apps/` folder.

Setting them to `Halide_Python` seems to work, but I don't know the full implications of this.

LebedevRI:
You need at least this additional change: https://github.com/LebedevRI/Halide/commit/988224d03a95196b9589a6fc4d4163b8ab0725c1
Some python tests fail for me:
```
30/42 Test #579: python_tutorial_lesson_04_debugging_2 ......................Subprocess aborted***Exception:   0.33 sec
terminate called after throwing an instance of 'std::runtime_error'
  what():  pybind11::object_api<>::operator() PyGILState_Check() failure.

      Start 580: python_tutorial_lesson_05_scheduling_1
31/42 Test #580: python_tutorial_lesson_05_scheduling_1 .....................Subprocess aborted***Exception:   0.55 sec
<...>
Evaluating gradient over a 5x4 box with x split by two 
Begin pipeline gradient_split_5x4.0()
produce gradient_split_5x4:
  for y:
    for x.x_outer:
      for x.x_inner in [0, 1]:
        gradient_split_5x4(...) = ...
terminate called after throwing an instance of 'std::runtime_error'
  what():  pybind11::object_api<>::operator() PyGILState_Check() failure.
```
Until they pass i can't proceed further with debian side of this (i have initial debian scaffoling changes for that though.)

Infinoid:
I incorporated @LebedevRI's CACHE PATH fix.  (Thanks!)
Apparently cmake 3.22.1 needs it, but cmake 3.18.4 worked fine without it.

I reported the `PyGILState_Check` test failures separately as #6524, as they fail with or without this PR.

alexreinking:
> Apparently cmake 3.22.1 needs it, but cmake 3.18.4 worked fine without it.

This is because CMake 3.19+ (I think) prefers to not cache the Python package result variables. This is to support loading multiple versions in different directories.

LebedevRI:
> > Apparently cmake 3.22.1 needs it, but cmake 3.18.4 worked fine without it.
> 
> This is because CMake 3.19+ (I think) prefers to not cache the Python package result variables. This is to support loading multiple versions in different directories.

FWIW, i'm not sure that my fix is right

LebedevRI:
> > Apparently cmake 3.22.1 needs it, but cmake 3.18.4 worked fine without it.
> 
> This is because CMake 3.19+ (I think) prefers to not cache the Python package result variables. This is to support loading multiple versions in different directories.

So i guess the right fix would be to rerun python discovery instead?

alexreinking:
> So i guess the right fix would be to rerun python discovery instead?

Either that or cache the relevant variables upon first discovery with a `Halide_` prefix

LebedevRI:
@Infinoid just so this becomes most uncontroversial, could the following diff be applied?
```diff
diff --git a/packaging/CMakeLists.txt b/packaging/CMakeLists.txt
index c7082828f..c1c632f36 100644
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -17,6 +17,8 @@ set(Halide_INSTALL_PLUGINDIR "${CMAKE_INSTALL_LIBDIR}"
 set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/tools"
     CACHE STRING "Path to Halide build-time tools and sources")
 
+find_package(Python3 REQUIRED)
+
 set(Halide_INSTALL_PYTHONDIR "${Python3_SITEARCH}"
     CACHE STRING "Path to Halide Python bindings folder")
 
diff --git a/python_bindings/CMakeLists.txt b/python_bindings/CMakeLists.txt
index bab669ab3..75fc03d63 100644
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -4,8 +4,6 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-set(Python3_SITEARCH "${Python3_SITEARCH}" CACHE PATH "Third-party platform dependent installation directory." FORCE)
-
 set(PYBIND11_VER 2.6.2)
 find_package(pybind11 ${PYBIND11_VER} QUIET)
 if (NOT pybind11_FOUND)

```

Infinoid:
I'm happy to include any requested changes, but

```diff
+find_package(Python3 REQUIRED)
```

I think this one would cause failures when a user didn't ask for Python, and doesn't have `python3-dev` installed.

What do you think of doing this, instead?

* Leave the cached variable in `python_bindings`, give it a `Halide_` prefix, and set it to `INTERNAL`
* Move the consumer down into the `if (WITH_PYTHON_BINDINGS)` block, so it's only referenced in cases when it's present

Like this:

```diff
diff --git a/packaging/CMakeLists.txt b/packaging/CMakeLists.txt
index c7082828f..c678e4b0c 100644
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -17,9 +17,6 @@ set(Halide_INSTALL_PLUGINDIR "${CMAKE_INSTALL_LIBDIR}"
 set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/tools"
     CACHE STRING "Path to Halide build-time tools and sources")
 
-set(Halide_INSTALL_PYTHONDIR "${Python3_SITEARCH}"
-    CACHE STRING "Path to Halide Python bindings folder")
-
 ##
 # Main library exports
 ##
@@ -53,6 +50,8 @@ endforeach ()
 ##
 
 if (WITH_PYTHON_BINDINGS)
+    set(Halide_INSTALL_PYTHONDIR "${Halide_Python3_SITEARCH}"
+        CACHE STRING "Path to Halide Python bindings folder")
     install(TARGETS Halide_Python
             EXPORT Halide_Python
             LIBRARY DESTINATION ${Halide_INSTALL_PYTHONDIR} COMPONENT Halide_Python
diff --git a/python_bindings/CMakeLists.txt b/python_bindings/CMakeLists.txt
index bab669ab3..4b4779568 100644
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-set(Python3_SITEARCH "${Python3_SITEARCH}" CACHE PATH "Third-party platform dependent installation directory." FORCE)
+set(Halide_Python3_SITEARCH "${Python3_SITEARCH}" CACHE INTERNAL "Third-party platform dependent installation directory.")
 
 set(PYBIND11_VER 2.6.2)
 find_package(pybind11 ${PYBIND11_VER} QUIET)
```

LebedevRI:
> I think this one would cause failures when a user didn't ask for Python, and doesn't have `python3-dev` installed.

Right. I guess i'm at loss of words as to why halide's repo structure is **so** interesting in all regards.

That will work, although i would have guessed not having an extra variable may be better.

Infinoid:
> Either that or cache the relevant variables upon first discovery with a `Halide_` prefix

Okay, I did that.

I also tested it with a bunch of cmake versions, with and without specifying a `Halide_PYTHON_INSTALLDIR`.  It seems to behave consistently across cmake versions.

One qualm I have is that `$Python3_SITEARCH` is completely independent of `$CMAKE_INSTALL_PREFIX`.  If you are installing to `/usr/local` (the default on Linux), it would try to put the Python module in `/usr/lib/python3` instead.  If you are installing to `$HOME/local`, it would still try to put it in `/usr/lib/python3`, and probably fail due to write permissions.  I'm guessing this won't work very well on other OSes either, but I haven't tried.

So I'm starting to think `$Python3_SITEARCH` isn't a good default.  It tells us where Python was installed, but that's not necessarily where Halide will be installed, and it seems like you would need to override it in most cases.

Would `$CMAKE_INSTALL_LIBDIR` or `$CMAKE_INSTALL_LIBDIR/python3` be a better default?  I think people will need to point `$PYTHONPATH` to it either way, so I'd rather keep it simple.

LebedevRI:
Oh, i haven't even thought of that. Yes. i think that is a problem.

> Would `$CMAKE_INSTALL_LIBDIR` or `$CMAKE_INSTALL_LIBDIR/python3` be a better default? I think people will need to point `$PYTHONPATH` to it either way, so I'd rather keep it simple.

Well, `$CMAKE_INSTALL_LIBDIR/python3/site-packages/` then though.

Infinoid:
> Well, `$CMAKE_INSTALL_LIBDIR/python3/site-packages/` then though.

Done.  Patch is super simple now.

packaging/CMakeLists.txt
```diff
@@ -45,6 +45,19 @@ foreach (dep IN ITEMS Halide_LLVM Halide_wabt)
...
+# Python bindings
+##
+
+if (WITH_PYTHON_BINDINGS)
+    set(Halide_INSTALL_PYTHONDIR "${CMAKE_INSTALL_LIBDIR}/python3/site-packages"
+        CACHE STRING "Path to Halide Python bindings folder")
+    install(TARGETS Halide_Python
+            EXPORT Halide_Python
+            LIBRARY DESTINATION ${Halide_INSTALL_PYTHONDIR} COMPONENT Halide_Python
+            NAMELINK_COMPONENT Halide_Python)
```
alexreinking, line 58: Nothing is done with this `EXPORT` set... either:

1. It should not be exported, or
2. it needs to be added to the set of CMake targets available via `find_package`. This will involve editing `packaging/common/HalideConfig.cmake` to load it when the `Python` component is requested. This will also include calling `find_dependency` on the correct version of Python (which will need to be recorded).

In the case of (2), testing should be added. In the case of (1), an issue should be opened.

alexreinking (changes requested):

packaging/CMakeLists.txt
LebedevRI: I'd say just (1) is more than enough for a start :)

Infinoid: Not exported, issue opened.

alexreinking (approved):

LebedevRI:
@alexreinking Thank you!

PR #6525, opened 2022-01-02
Fix Python GIL lock handling (Fixes #6524, Fixes #5631)

### LINKED ISSUE #5631: Release GIL in the generated Python extension source code
It would be very nice to have an option for releasing GIL in the generated Python extension source code (or maybe just always do this). I assume it should be OK since NumPy releases the lock too, so this ought to be the expected behaviour.

The change itself should be very simple: just wrap [call to the actual function](
https://github.com/halide/Halide/blob/9b99acb0b9033e7c90d369e14154332737db3a6b/src/PythonExtensionGen.cpp#L350) into [`Py_BEGIN_ALLOW_THREADS/Py_END_ALLOW_THREADS`](https://docs.python.org/3/c-api/init.html#c.Py_BEGIN_ALLOW_THREADS) macros.

### LINKED ISSUE #6524: Python test failures with pybind11 v2.8.1
Debian bookworm, by default, has llvm 13, python 3.9.9, and pybind11 2.8.1.  Two python tests fail:

```
30/42 Test #579: python_tutorial_lesson_04_debugging_2 ......................Subprocess aborted***Exception:   0.33 sec
terminate called after throwing an instance of 'std::runtime_error'
  what():  pybind11::object_api<>::operator() PyGILState_Check() failure.

      Start 580: python_tutorial_lesson_05_scheduling_1
31/42 Test #580: python_tutorial_lesson_05_scheduling_1 .....................Subprocess aborted***Exception:   0.55 sec
<...>
Evaluating gradient over a 5x4 box with x split by two 
Begin pipeline gradient_split_5x4.0()
produce gradient_split_5x4:
  for y:
    for x.x_outer:
      for x.x_inner in [0, 1]:
        gradient_split_5x4(...) = ...
terminate called after throwing an instance of 'std::runtime_error'
  what():  pybind11::object_api<>::operator() PyGILState_Check() failure.
```

When I revert pybind11 to version 2.6.2 (using packages from the previous debian release), reconfigure and rebuild, the python tests pass.

```
1/2 Test #571: python_tutorial_lesson_04_debugging_2 ....   Passed    0.45 sec
2/2 Test #572: python_tutorial_lesson_05_scheduling_1 ...   Passed    6.89 sec
```

Here is a minimal reproducer, based on tutorial lesson 04:

```python3
#!/usr/bin/env python3

import halide as hl

x = hl.Var("x")

test = hl.Func("test")
test[x] = 0
test.trace_stores()
test.parallel(x)
test.realize([8])
```

If I comment out the call to `trace_stores()` or the call to `parallel()`, the test passes.  Both are necessary to trigger the bug.

LebedevRI:
As disscussed in https://github.com/halide/Halide/pull/6523#issuecomment-1003545664
and later in https://github.com/halide/Halide/issues/6524,
pybind11 v2.8.1 added some defensive checks that fail for halide,
namely in `python_tutorial_lesson_04_debugging_2`
and `python_tutorial_lesson_05_scheduling_1`.

https://github.com/halide/Halide/issues/6524#issuecomment-1003569810 notes:
> * Python calls a Halide-JIT-generated function , which runs with the GIL held.
> * Halide runtime spawns worker threads.
> * The worker threads try to call pybind11's py::print function to emit traces.
> * Pybind11 complains, correctly, that the worker thread doesn't hold the GIL.
>
> Trying to acquire the GIL hangs, because the main thread is still holding it. I tried teaching the main thread to release the GIL (as suggested in #5631), but I still saw hangs when I tried this.

I have tried, and just dropping the lock before calling into halide,
or just acquiring it in `halide_python_print` doesn't work,
we need to do both.

I have verified that the two tests fail without this fix,
and pass with it.

### DIFF
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -4,13 +4,14 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-set(PYBIND11_VER 2.6.2)
-find_package(pybind11 ${PYBIND11_VER} QUIET)
+set(PYBIND11_MIN_SUPPORTED_VER 2.6.2)
+find_package(pybind11 ${PYBIND11_MIN_SUPPORTED_VER} QUIET)
 if (NOT pybind11_FOUND)
+    set(PYBIND11_BUNDLED_VER 2.9.0)
     include(FetchContent)
     FetchContent_Declare(pybind11
                          GIT_REPOSITORY https://github.com/pybind/pybind11.git
-                         GIT_TAG v${PYBIND11_VER})
+                         GIT_TAG v${PYBIND11_BUNDLED_VER})
     FetchContent_MakeAvailable(pybind11)
 endif ()
 
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -4,14 +4,13 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-set(PYBIND11_MIN_SUPPORTED_VER 2.6.2)
-find_package(pybind11 ${PYBIND11_MIN_SUPPORTED_VER} QUIET)
+set(PYBIND11_VER 2.6.2)
+find_package(pybind11 ${PYBIND11_VER} QUIET)
 if (NOT pybind11_FOUND)
-    set(PYBIND11_BUNDLED_VER 2.9.0)
     include(FetchContent)
     FetchContent_Declare(pybind11
                          GIT_REPOSITORY https://github.com/pybind/pybind11.git
-                         GIT_TAG v${PYBIND11_BUNDLED_VER})
+                         GIT_TAG v${PYBIND11_VER})
     FetchContent_MakeAvailable(pybind11)
 endif ()
 
--- a/python_bindings/src/PyError.cpp
+++ b/python_bindings/src/PyError.cpp
@@ -10,6 +10,7 @@ void halide_python_error(JITUserContext *, const char *msg) {
 }
 
 void halide_python_print(JITUserContext *, const char *msg) {
+    py::gil_scoped_acquire acquire;
     py::print(msg, py::arg("end") = "");
 }
 
--- a/python_bindings/src/PyFunc.cpp
+++ b/python_bindings/src/PyFunc.cpp
@@ -117,6 +117,7 @@ void define_func(py::module &m) {
             .def(
                 "realize",
                 [](Func &f, Buffer<> buffer, const Target &target) -> void {
+                    py::gil_scoped_release release;
                     f.realize(buffer, target);
                 },
                 py::arg("dst"), py::arg("target") = Target())
@@ -125,6 +126,7 @@ void define_func(py::module &m) {
             .def(
                 "realize",
                 [](Func &f, std::vector<Buffer<>> buffers, const Target &t) -> void {
+                    py::gil_scoped_release release;
                     f.realize(Realization(buffers), t);
                 },
                 py::arg("dst"), py::arg("target") = Target())
@@ -132,7 +134,12 @@ void define_func(py::module &m) {
             .def(
                 "realize",
                 [](Func &f, const std::vector<int32_t> &sizes, const Target &target) -> py::object {
-                    return realization_to_object(f.realize(sizes, target));
+                    std::optional<Realization> r;
+                    {
+                        py::gil_scoped_release release;
+                        r = std::move(f.realize(sizes, target));
+                    }
+                    return realization_to_object(*r);
                 },
                 py::arg("sizes") = std::vector<int32_t>{}, py::arg("target") = Target())
 
--- a/python_bindings/src/PyPipeline.cpp
+++ b/python_bindings/src/PyPipeline.cpp
@@ -96,6 +96,7 @@ void define_pipeline(py::module &m) {
 
             .def(
                 "realize", [](Pipeline &p, Buffer<> buffer, const Target &target) -> void {
+                    py::gil_scoped_release release;
                     p.realize(Realization(buffer), target);
                 },
                 py::arg("dst"), py::arg("target") = Target())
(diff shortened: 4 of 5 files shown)


Infinoid:
I verified this patch fixes the failures I was seeing, and doesn't add any new ones.

```
100% tests passed, 0 tests failed out of 614
```

Thanks!

LebedevRI:
> I verified this patch fixes the failures I was seeing, and doesn't add any new ones.
> 
> ```
> 100% tests passed, 0 tests failed out of 614
> ```
> 
> Thanks!

Theoretically, we need to drop the lock *every* time we call halide API from python,
and take the lock when calling python api from halide.
I only handled `realize` since that is the entry that caused the given crashes,
and only dealt with `halide_python_print`, while e.g. `HalidePythonCompileTimeErrorReporter`
is still "wrong". So i very much expect that this adds more brokenness.

Infinoid:
> Theoretically, we need to drop the lock _every_ time we call halide API from python,
> and take the lock every time we call python api from halide.

For calls from Python into Halide that don't spawn threads, the easiest thing to do is leave the lock alone.  Then it's fine if Halide calls back into Python, because it still holds the lock.

Other cases can be reviewed on a case-by-case basis.

LebedevRI:
> Other cases can be reviewed on a case-by-case basis.

That's why i posted this yes, clearly if no one cared to fix this problem,
they won't mind having more potential problems.

Infinoid:
Well, my point was, no threads == no problem.  I don't think that pipeline definitions, scheduling, AOT compilation, target definitions or buffer accesses spawn threads.

The only entry points I can think of that do spawn threads are `Func.realize()`, `Pipeline.realize()`, and calling an AOT-compiled function, and I think this patch covers those cases.  But I'm not an expert.

steven-johnson (changes requested):
I don't know enough about the GIL to be the ideal reviewer here, but I *do* think that we should update `python_bindings/CMakeLists.txt` to specify (at least) v2.8.1 for the version we pull for building. (Maybe 2.9, since that is apparently the most recent release?). If you update this PR the buildbots should test using the version specified.

LebedevRI:
> I don't know enough about the GIL to be the ideal reviewer here, but I _do_ think that we should update `python_bindings/CMakeLists.txt` to specify (at least) v2.8.1 for the version we pull for building. (Maybe 2.9, since that is apparently the most recent release?). If you update this PR the buildbots should test using the version specified.

Err, i do not follow. Why do you want to bump the required pybind version?
Just because version A is required, doesn't mean that buildbots can't test version A+1.

steven-johnson:
> Err, i do not follow. Why do you want to bump the required pybind version? 

Because our version is out of date and we like to keep close to the latest release unless theres a reason not t.

> Just because version A is required, doesn't mean that buildbots can't test version A+1.

True, but there isn't any mechanism for doing so right now -- we'd have to add one.

LebedevRI:
> > Err, i do not follow. Why do you want to bump the required pybind version?
> 
> Because our version is out of date and we like to keep close to the latest release unless theres a reason not t.
I see, but then this is a separate concern from what we have at hand.

Roughly, with caveats, depending on the most newest version of everything makes it hard while not impossible
to use ${PROJ} on anything but the bleeding edge distributions.

Case in point, if you bump it to v2.9, then i'm simply going to abandon this patch,
because i will no longer be able to build it in debian unstable,
and if you push just to v2.8, then it will become unbuildable in debian stable,
which will create additional problems for the original reporter, @Infinoid.

> > Just because version A is required, doesn't mean that buildbots can't test version A+1.
> 
> True, but there isn't any mechanism for doing so right now -- we'd have to add one.

steven-johnson:
> Case in point, if you bump it to v2.9, then i'm simply going to abandon this patch, because i will no longer be able to build it in debian unstable

Not true for the CMake build, which *always* pulls and builds a specific, captive version, ignoring what is present on the system. (If this isn't what's happening, it's a bug in our CMake rules.) Are you using Make?

LebedevRI:
> > Case in point, if you bump it to v2.9, then i'm simply going to abandon this patch, because i will no longer be able to build it in debian unstable
> 
> Not true for the CMake build, which _always_ pulls and builds a specific, captive version, ignoring what is present on the system. (If this isn't what's happening, it's a bug in our CMake rules.) Are you using Make?

Hang on, we are talking about different things here, aren't we :)
I'm talking about the `find_package()` stuff, *NOT* `FetchContent`.
I don't have an opinion on what version of `pybind` you should bundle,
and i can bump that to v2.9 if you want,
but i do have an opinion on the *required* pybind version,
which i don't want to bump.

steven-johnson:
ah, ok, in that case I defer entirely too @alexreinking :-)

LebedevRI:
> ah, ok, in that case I defer entirely too @alexreinking :-)

Bumped the bundled version, CI seems to be happy with it :)

python_bindings/CMakeLists.txt
```diff
@@ -4,13 +4,14 @@
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-set(PYBIND11_VER 2.6.2)
-find_package(pybind11 ${PYBIND11_VER} QUIET)
+set(PYBIND11_MIN_SUPPORTED_VER 2.6.2)
```
steven-johnson, line 7: Nit: I wonder why we don't always FetchContent for this, so we always get a known version; IIUC, PyBind11 is 100% a header-only library, with zero runtime dependencies, and it isn't that large.

steven-johnson (approved):

python_bindings/CMakeLists.txt
LebedevRI: Downloading stuff during cmake/make time is a deal-breaker for packaging, for the record.

alexreinking (changes requested):

python_bindings/CMakeLists.txt
```diff
@@ -4,13 +4,14 @@
...
+find_package(pybind11 ${PYBIND11_MIN_SUPPORTED_VER} QUIET)
 if (NOT pybind11_FOUND)
+    set(PYBIND11_BUNDLED_VER 2.9.0)
```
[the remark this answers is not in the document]
alexreinking: @steven-johnson - we might need to add Python-only testing for "minimum version" and "trunk" pybind11

alexreinking (approved):
I've opened #6531 to record that we need to test newer pybind11, too. With that backed out, this LGTM.

alexreinking:
Failure is a performance test on llvm14 that does not use Python... cannot possibly be related.

LebedevRI:
@alexreinking thank you!

PR #6526, opened 2022-01-02
Handle mixed-width args to mul-shift-right

abadams:
and codegen it to pmulhuw on x86

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -384,6 +384,7 @@ protected:
             auto is_x_same_int = op->type.is_int() && is_int(x, bits);
             auto is_x_same_uint = op->type.is_uint() && is_uint(x, bits);
             auto is_x_same_int_or_uint = is_x_same_int || is_x_same_uint;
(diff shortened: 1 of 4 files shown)


src/FindIntrinsics.cpp
```diff
@@ -493,6 +494,25 @@ class FindIntrinsics : public IRMutator {
                         rounding_mul_shift_right(x, y, cast(unsigned_type, c0)),
                         is_x_same_int_or_uint && c0 >= bits) ||
 
+                // We can also match on smaller shifts if one of the args is
+                // narrow. We don't do this for signed (yet), because the
+                // saturation issue is tricky.
```
steven-johnson, line 499: Do we think we will ever do so, tricky or not? (If so, tracking issue, please)

steven-johnson (approved):
LGTM pending green

src/FindIntrinsics.cpp
abadams: Not sure. It depends on if it comes up in practice. It's part of the unbounded frontier of ways we could make instruction selection more clever. I don't want to add a tracking issue for each possible new way to use an instruction.

dsharletg (approved):

PR #6529, opened 2022-01-03
Fix simd-op-check for top-of-tree LLVM

steven-johnson:


abadams:
Hrm, that seems like a regression in llvm. Even if it's still a single instruction, it occupies an register with the constant. I'll take a closer look.

steven-johnson:
> Hrm, that seems like a regression in llvm. Even if it's still a single instruction, it occupies an register with the constant. I'll take a closer look.

See https://reviews.llvm.org/D116166

abadams:
Ah, I was just looking for that, thanks. I guess twice the throughput is worth the extra register used nearly all of the time. Maybe add that link as a comment in the PR to explain why it's doing that?

abadams (approved):

PR #6530, opened 2022-01-04
Install Python tutorials

LebedevRI:
I know we have previously discussed that `TYPE DOC` should be used,
but unfortunately i'm not sure that will work here,
because doc/tutorial directory is already occupied by C++ tutorials,
and i don't think they should be mixed.
I'm open to alternative suggestions.

### DIFF
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -141,6 +141,14 @@ if (WITH_TUTORIALS)
             PATTERN "*.jpg"
             PATTERN "*.mp4"
             PATTERN "*.png")
+
+    if (WITH_PYTHON_BINDINGS)
+      install(DIRECTORY ${Halide_SOURCE_DIR}/python_bindings/tutorial/
+              DESTINATION ${CMAKE_INSTALL_DOCDIR}/tutorial-python
+              COMPONENT Halide_Documentation
+              FILES_MATCHING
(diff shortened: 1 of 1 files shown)


packaging/CMakeLists.txt
```diff
@@ -141,6 +141,14 @@ if (WITH_TUTORIALS)
             PATTERN "*.jpg"
             PATTERN "*.mp4"
             PATTERN "*.png")
+
+    if (WITH_PYTHON_BINDINGS)
+      install(DIRECTORY ${Halide_SOURCE_DIR}/python_bindings/tutorial/
+              DESTINATION ${CMAKE_INSTALL_DOCDIR}/tutorial-python
+              COMPONENT Halide_Documentation
+              FILES_MATCHING
+              PATTERN "*.py")
```
alexreinking, line 150: With only one `PATTERN`, prefer to place on the same line as `FILES_MATCHING`.

alexreinking (approved):
LGTM with a nitpick

alexreinking:
> I know we have previously discussed that TYPE DOC should be used, but unfortunately i'm not sure that will work here,

Yeah, I think this is fine for now. We'd need to adjust our directory structure to be able to use `TYPE DOC` here, which... meh. It's low priority.

LebedevRI:
@alexreinking thank you!

PR #6534, opened 2022-01-05
Allow third parties to externally override SOVERSION

alexreinking:
Debian and other third-party packagers might need or want to patch our sources for a variety of reasons. In those cases, they might also need to override the SOVERSION.

See here for a practical example:

https://salsa.debian.org/pkg-llvm-team/halide/-/blob/f881de70cd83095053e13047b63f61faf6bc7a36/debian/patches/0006-Fixup-libhalide-version-soversion-for-debian-package.patch

ATTN @LebedevRI

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -385,10 +385,14 @@ target_export_script(Halide
                      APPLE_LD "${CMAKE_CURRENT_LIST_DIR}/exported_symbols.osx"
                      GNU_LD "${CMAKE_CURRENT_LIST_DIR}/exported_symbols.ldscript")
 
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):

LebedevRI:
Thanks, this will simplify things a bit!
Note the second part of that patch - in order to make the libHalide co-installable, we also override it's `OUTPUT_NAME`.

alexreinking:
> Thanks, this will simplify things a bit! Note the second part of that patch - in order to make the libHalide co-installable, we also override it's `OUTPUT_NAME`.

My impression was that `CMAKE_<CONFIG>_POSTFIX` was sufficient for this.

LebedevRI:
@alexreinking thank you!

PR #6537, opened 2022-01-05
Remove unnecessary `std::move` calls

steven-johnson:
Compilers with `-Werror` will fail with `error: moving a temporary object prevents copy elision`


LebedevRI:
Hm, i guess i thought `-Werror` was enabled in CI and something like that would have been caught.
Sorry.

dsharletg (approved):

alexreinking (approved):

alexreinking:
I'm surprised clang tidy didn't complain about this

PR #6539, opened 2022-01-05
Attempted redo of faster noise

abadams:
This restricts the new simplifier rules to unsigned only.


steven-johnson:
Failure is unrelated

steven-johnson (approved):

PR #6542, opened 2022-01-06
Tweak type-comparing code for Xtensa

steven-johnson:
This is an untested, just-fyi PR, which is full of premature optimization, but still, FYI:

`halide_type_t::as_u32()` is a constexpr method that returns a unique value for a given type (including lanes). This allows us to do a couple of things:

(1) by forcing `is_native_xtensa_vector()` to  use this operator, we can (in theory) get more efficient code for this frequently-used helper by inlining it as essentially a single compare-two-u32 instruction. (I haven't actually tested this. Maybe it's less efficient.)

(2) since it's a constexpr function, you can also use it to construct `switch` statements as I've done here (both with and without the lanes); the syntax is suboptimal (I'm open for suggestions) but may produce cleaner code. (Again, untested?)

(2a) aside from switch, we could also use this approach to build tables of static+constant data (since the method is `constexpr`).

### DIFF
--- a/src/CodeGen_Xtensa.cpp
+++ b/src/CodeGen_Xtensa.cpp
@@ -1909,19 +1909,19 @@ class ScopedDmaInitializer {
         stream << std::flush;
 
         std::set<Type> native_vector_types = {
-            Type(Type::Int, 8, 64),
-            Type(Type::UInt, 8, 64),
-            Type(Type::Int, 16, 32),
-            Type(Type::UInt, 16, 32),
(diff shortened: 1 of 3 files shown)


vksnk (approved):
Thanks a lot, this definitely looks better.

I'll test it locally.

vksnk:
> Thanks a lot, this definitely looks better.
> 
> I'll test it locally

Tested locally - looks good.

PR #6549, opened 2022-01-07
Fix description of rounding_shift_left/rounding_shift_right

dsharletg:
The comment refers to an older understanding which is incorrect for saturating cases, e.g. `rounding_shift_right(255, 1)` = 128 (new definition) and not 127 (old definition).


abadams (approved):

dsharletg:
This PR only affects comments, I will merge (ignoring failures).

PR #6554, opened 2022-01-12
Use add_halide_generator() everywhere in apps/

steven-johnson:
Existing CMake code uses add_executable + target_link_libraries to create a Generator, but the somewhat-recently-added add_halide_generator() helper is a cleaner way to do this. Move to using it everywhere in apps/ instead.

(Note that add_halide_generator()) doesn't yet work for in-tree builds, unfortunately.)

### DIFF
--- a/apps/bgu/CMakeLists.txt
+++ b/apps/bgu/CMakeLists.txt
@@ -12,8 +12,7 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(bgu.generator bgu_generator.cpp)
-target_link_libraries(bgu.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(bgu.generator SOURCES bgu_generator.cpp)
 
 # Filters
 add_halide_library(bgu FROM bgu.generator)
--- a/apps/bgu/CMakeLists.txt
+++ b/apps/bgu/CMakeLists.txt
@@ -15,8 +15,8 @@ find_package(Halide REQUIRED)
 add_halide_generator(bgu.generator SOURCES bgu_generator.cpp)
 
 # Filters
-add_halide_library(bgu FROM bgu::halide_generators::bgu.generator)
-add_halide_library(bgu_auto_schedule FROM bgu::halide_generators::bgu.generator
+add_halide_library(bgu FROM bgu.generator)
+add_halide_library(bgu_auto_schedule FROM bgu.generator
                    GENERATOR bgu
                    AUTOSCHEDULER Halide::Mullapudi2016)
 
--- a/apps/bgu/CMakeLists.txt
+++ b/apps/bgu/CMakeLists.txt
@@ -15,8 +15,8 @@ find_package(Halide REQUIRED)
 add_halide_generator(bgu.generator SOURCES bgu_generator.cpp)
 
 # Filters
-add_halide_library(bgu FROM bgu.generator)
-add_halide_library(bgu_auto_schedule FROM bgu.generator
+add_halide_library(bgu FROM bgu::halide_generators::bgu.generator)
+add_halide_library(bgu_auto_schedule FROM bgu::halide_generators::bgu.generator
                    GENERATOR bgu
                    AUTOSCHEDULER Halide::Mullapudi2016)
 
--- a/apps/bilateral_grid/CMakeLists.txt
+++ b/apps/bilateral_grid/CMakeLists.txt
@@ -16,11 +16,11 @@ add_halide_generator(bilateral_grid.generator SOURCES bilateral_grid_generator.c
 target_link_libraries(bilateral_grid.generator PRIVATE Halide::Tools)
 
 # Filters
-add_halide_library(bilateral_grid FROM bilateral_grid.generator
+add_halide_library(bilateral_grid FROM bilateral_grid::halide_generators::bilateral_grid.generator
                    STMT bilateral_grid_STMT
                    SCHEDULE bilateral_grid_SCHEDULE)
 
-add_halide_library(bilateral_grid_auto_schedule FROM bilateral_grid.generator
+add_halide_library(bilateral_grid_auto_schedule FROM bilateral_grid::halide_generators::bilateral_grid.generator
                    GENERATOR bilateral_grid
                    STMT bilateral_grid_auto_schedule_STMT
                    SCHEDULE bilateral_grid_auto_schedule_SCHEDULE
--- a/apps/bilateral_grid/CMakeLists.txt
+++ b/apps/bilateral_grid/CMakeLists.txt
@@ -12,8 +12,9 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_halide_generator(bilateral_grid.generator SOURCES bilateral_grid_generator.cpp)
-target_link_libraries(bilateral_grid::halide_generators::bilateral_grid.generator PRIVATE Halide::Tools)
+add_halide_generator(bilateral_grid.generator
+                     SOURCES bilateral_grid_generator.cpp
+                     LINK_LIBRARIES Halide::Tools)
 
 # Filters
 add_halide_library(bilateral_grid FROM bilateral_grid::halide_generators::bilateral_grid.generator
--- a/apps/bilateral_grid/CMakeLists.txt
+++ b/apps/bilateral_grid/CMakeLists.txt
@@ -17,11 +17,11 @@ add_halide_generator(bilateral_grid.generator
                      LINK_LIBRARIES Halide::Tools)
 
 # Filters
-add_halide_library(bilateral_grid FROM bilateral_grid::halide_generators::bilateral_grid.generator
+add_halide_library(bilateral_grid FROM bilateral_grid.generator
                    STMT bilateral_grid_STMT
                    SCHEDULE bilateral_grid_SCHEDULE)
 
-add_halide_library(bilateral_grid_auto_schedule FROM bilateral_grid::halide_generators::bilateral_grid.generator
+add_halide_library(bilateral_grid_auto_schedule FROM bilateral_grid.generator
                    GENERATOR bilateral_grid
                    STMT bilateral_grid_auto_schedule_STMT
                    SCHEDULE bilateral_grid_auto_schedule_SCHEDULE
--- a/apps/bilateral_grid/CMakeLists.txt
+++ b/apps/bilateral_grid/CMakeLists.txt
@@ -13,6 +13,7 @@ find_package(Halide REQUIRED)
 
 # Generator
 add_halide_generator(bilateral_grid.generator SOURCES bilateral_grid_generator.cpp)
+target_link_libraries(bilateral_grid.generator PRIVATE Halide::Tools)
 
 # Filters
 add_halide_library(bilateral_grid FROM bilateral_grid.generator
--- a/apps/bilateral_grid/CMakeLists.txt
+++ b/apps/bilateral_grid/CMakeLists.txt
@@ -13,7 +13,7 @@ find_package(Halide REQUIRED)
 
 # Generator
 add_halide_generator(bilateral_grid.generator SOURCES bilateral_grid_generator.cpp)
-target_link_libraries(bilateral_grid.generator PRIVATE Halide::Tools)
+target_link_libraries(bilateral_grid::halide_generators::bilateral_grid.generator PRIVATE Halide::Tools)
 
 # Filters
 add_halide_library(bilateral_grid FROM bilateral_grid::halide_generators::bilateral_grid.generator
--- a/apps/bilateral_grid/CMakeLists.txt
+++ b/apps/bilateral_grid/CMakeLists.txt
@@ -12,8 +12,7 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(bilateral_grid.generator bilateral_grid_generator.cpp)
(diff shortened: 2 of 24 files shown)


apps/bgu/CMakeLists.txt
```diff
@@ -12,8 +12,7 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(bgu.generator bgu_generator.cpp)
-target_link_libraries(bgu.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(bgu.generator SOURCES bgu_generator.cpp)
 
 # Filters
 add_halide_library(bgu FROM bgu.generator)
```
alexreinking, line 18: When using `add_halide_generator`, you need to use the namespaced name to get correct cross-compiling behavior. This would need to be:

```
add_halide_library(bgu FROM bgu::halide_generators::bgu.generator)
```

This needs to be adjusted everywhere.

alexreinking (changes requested):

apps/bgu/CMakeLists.txt
steven-johnson: I hear you, and yet, it seemed to work properly for me locally.

alexreinking: That is because you aren't cross-compiling. When you try to cross-compile these, it will fail.

alexreinking: The plain-named target exists in the host build to allow for build customization... the pattern is:

```
add_halide_generator(gen SOURCES ...)
if (TARGET gen)
  # set gen's properties, etc.
endif ()

# project::halide_generators::gen either comes from a host build import or is an alias to plain "gen"
add_halide_library(lib FROM project::halide_generators::gen)
```

steven-johnson: (1) Fair enough, but I don't understand why the previous code wouldn't also fail in those cases. 
(2) Can we / should we try to make it so that it will fail reliably in all cases?

alexreinking: (1) the old code (as in, pre-PR) _would_ fail to cross-compile, yes.

(2) I am super open to ideas. I think we're pretty free to change `add_halide_generator` since it's quite new.

steven-johnson: Gotcha.

Couldn't we have `add_halide_library()` check the `FROM` and if it's not present, look in `PACKAGE::halide_generators::FROM`?

alexreinking: > Couldn't we have `add_halide_library()` check the `FROM` and if it's not present, look in `PACKAGE::halide_generators::FROM`?

We absolutely could do that.

apps/camera_pipe/CMakeLists.txt
```diff
@@ -12,15 +12,12 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(camera_pipe.generator camera_pipe_generator.cpp)
-target_link_libraries(camera_pipe.generator
-                      PRIVATE
-                      Halide::Generator
-                      Halide::Tools)
+add_halide_generator(camera_pipe.generator SOURCES camera_pipe_generator.cpp)
+target_link_libraries(camera_pipe.generator PRIVATE Halide::Tools)
```
alexreinking, line 16:
```
if (TARGET camera_pipe.generator)
   target_link_libraries(camera_pipe.generator PRIVATE Halide::Tools)
endif ()
```
Though it wouldn't be unreasonable to link to `Halide::Tools` by default or to include an argument to `add_halide_generator` that calls `target_link_libraries` appropriately behind the scenes.

apps/bilateral_grid/CMakeLists.txt
```diff
@@ -12,15 +12,15 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(bilateral_grid.generator bilateral_grid_generator.cpp)
-target_link_libraries(bilateral_grid.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(bilateral_grid.generator SOURCES bilateral_grid_generator.cpp)
+target_link_libraries(bilateral_grid.generator PRIVATE Halide::Tools)
```
alexreinking, line 16: Same issue here

alexreinking (changes requested):

apps/cuda_mat_mul/CMakeLists.txt
```diff
@@ -25,11 +25,10 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(mat_mul.generator mat_mul_generator.cpp)
-target_link_libraries(mat_mul.generator PRIVATE Halide::Generator)
+add_halide_generator(mat_mul.generator SOURCES mat_mul_generator.cpp)
 
 # Filters
-add_halide_library(mat_mul FROM mat_mul.generator
+add_halide_library(mat_mul FROM mat_mul::halide_generators::mat_mul.generator
```
alexreinking, line 31: Project name is `cuda_mat_mul`, so this should be `cuda_mat_mul::...`

apps/camera_pipe/CMakeLists.txt
steven-johnson: Why do we need to check to see if it's a target... didn't we just add it as a target?

steven-johnson: also, this doesn't seem like the right fix -- if it somehow isn't a target, it won't build properly without this

alexreinking: When cross-compiling, the fully qualified name will exist as an imported target. When doing a host build, the unqualified name will _also_ exist as a build target.

apps/bilateral_grid/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ find_package(Halide REQUIRED)
 
 # Generator
 add_halide_generator(bilateral_grid.generator SOURCES bilateral_grid_generator.cpp)
-target_link_libraries(bilateral_grid.generator PRIVATE Halide::Tools)
+target_link_libraries(bilateral_grid::halide_generators::bilateral_grid.generator PRIVATE Halide::Tools)
```
alexreinking, line 16: Does this actually work? I thought you couldn't set properties through aliases (and for imported targets, `PRIVATE` makes no sense)

steven-johnson: ...I thought that's what you wanted me to do. 

I thought the whole point of this was to make it work for crosscompiling. The `if TARGET` approach won't because there apparently won't be a non-decorated version of the library in that case.

alexreinking: I'm sorry, I must not have explained well. `if (TARGET ${unqualified_name})` is the correct way to check if you're in the host-build scenario, which is the one that matters if you're going to set build options on `${unqualified_name}`, like which libraries to link to the generator executable.

* In the host build scenario, these targets exist:
  * `${unqualified_name}` is a normal CMake target that will create an executable. It is linked to `Halide::Generator` by default.
  * `${qualified_name}` is an `ALIAS` target to `${unqualified_name}`.
* In the cross-build scenario, this target exists:
  * `${qualified_name}` is an `IMPORTED` target that points to the executable in the host build.
  * `${unqualified_name}` does not exist.

In the cross-build, you are specifically importing the host-built tools. Since they have already been built, it does not make sense to set build properties on them.

apps/hist/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(hist.generator hist_generator.cpp)
-target_link_libraries(hist.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(hist.generator SOURCES hist_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/harris/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(harris.generator harris_generator.cpp)
-target_link_libraries(harris.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(harris.generator SOURCES harris_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/interpolate/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(interpolate.generator interpolate_generator.cpp)
-target_link_libraries(interpolate.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(interpolate.generator SOURCES interpolate_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/lens_blur/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(lens_blur.generator lens_blur_generator.cpp)
-target_link_libraries(lens_blur.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(lens_blur.generator SOURCES lens_blur_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/max_filter/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(max_filter.generator max_filter_generator.cpp)
-target_link_libraries(max_filter.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(max_filter.generator SOURCES max_filter_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/nl_means/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(nl_means.generator nl_means_generator.cpp)
-target_link_libraries(nl_means.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(nl_means.generator SOURCES nl_means_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/stencil_chain/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(stencil_chain.generator stencil_chain_generator.cpp)
-target_link_libraries(stencil_chain.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(stencil_chain.generator SOURCES stencil_chain_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

apps/unsharp/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(unsharp.generator unsharp_generator.cpp)
-target_link_libraries(unsharp.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(unsharp.generator SOURCES unsharp_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

alexreinking (approved):

apps/bgu/CMakeLists.txt
```diff
@@ -12,8 +12,7 @@ set(CMAKE_CXX_EXTENSIONS NO)
 find_package(Halide REQUIRED)
 
 # Generator
-add_executable(bgu.generator bgu_generator.cpp)
-target_link_libraries(bgu.generator PRIVATE Halide::Generator Halide::Tools)
+add_halide_generator(bgu.generator SOURCES bgu_generator.cpp)
```
alexreinking, line 15: No need for `Halide::Tools`?

steven-johnson:
I didn't re-add Halide::Tools to anything that didn't require it for compilation. My guess is that we had a lot that were adding it from copy-paste codegen.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -148,6 +148,16 @@ function(add_halide_library TARGET)
         message(FATAL_ERROR "Missing FROM argument specifying a Halide generator target")
     endif ()
 
+    if (NOT TARGET ${ARG_FROM})
+        # FROM is usually an unqualified name; if we are crosscompiling, we might need a
+        # fully-qualified name, so add the default package name and retry
+        set(FQ_ARG_FROM "${PROJECT_NAME}::halide_generators::${ARG_FROM}")
+        if (NOT TARGET ${FQ_ARG_FROM})
+            message(FATAL_ERROR "Unable to local FROM as either ${ARG_FROM} or ${FQ_ARG_FROM}")
```
alexreinking, line 156: locate*

alexreinking (approved):

alexreinking:
Now that this is merged, we should actually try cross-compiling some of these.

PR #6555, opened 2022-01-12
Enable deprecations warnings

steven-johnson:
We currently disable deprecation warnings inside Halide. This re-enables them there, and also inside add_halide_generator().

(Note, additive to PR #6554)


cmake/HalideGeneratorHelpers.cmake
```diff
@@ -57,6 +57,16 @@ function(add_halide_generator TARGET)
...
+        # For crosscompiling builds, the Halide headers will be included using -isystem,
+        # which will cause all warnings to be ignored. This is not helpful, since
+        # we *want* deprecation warnings to be propagated. So we must set
+        # NO_SYSTEM_FROM_IMPORTED in order for it to be seen.
+        set_target_properties(${TARGET} PROPERTIES NO_SYSTEM_FROM_IMPORTED YES)
+        target_compile_options(${TARGET} PRIVATE
+            $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wdeprecated-declarations>
+            $<$<CXX_COMPILER_ID:MSVC>:/w14996>  # 4996: compiler encountered deprecated declaration
+        )
+
```
alexreinking, line 69: I would like these to be opt-out. There should be a bit of code like this:

```
if (NOT ARG_NO_DEFAULT_FLAGS AND NOT Halide_NO_DEFAULT_FLAGS)
    # code from PR
endif ()
```

There are several reasons for this, but the biggest is that `NO_SYSTEM_FROM_IMPORTED` is _very_ coarse, and users who link to another library in addition to Halide (say they want to use Boost or Abseil or fmt or something) for a generator will start getting warnings from those libraries.

src/CMakeLists.txt
```diff
@@ -476,7 +476,9 @@ target_compile_options(
         $<$<CXX_COMPILER_ID:MSVC>:/wd4291>  # 4291: No matching operator delete found
         $<$<CXX_COMPILER_ID:MSVC>:/wd4503>  # 4503: disable "decorated name length exceeded, name was truncated"
         $<$<CXX_COMPILER_ID:MSVC>:/wd4800>  # 4800: forcing value to bool 'true' or 'false' (performance warning)
-        $<$<CXX_COMPILER_ID:MSVC>:/wd4996>  # 4996: compiler encountered deprecated declaration
+
+        # No: enable deprecation warnings
+        # $<$<CXX_COMPILER_ID:MSVC>:/wd4996>  # 4996: compiler encountered deprecated declaration
```
alexreinking, line 481: While we're here, can we remove `Clang` from `$<$<CXX_COMPILER_ID:Clang,AppleClang>:-Wno-return-std-move-in-c++11>` above? GitHub won't let me comment outside the diff.

alexreinking (changes requested):

steven-johnson:
PTAL

alexreinking:
What's the deal with the Windows failure here?

steven-johnson:
> What's the deal with the Windows failure here?

No idea, investigating

alexreinking:
I'm waiting to chat with the Clang developers about this before approving.

steven-johnson:
> I'm waiting to chat with the Clang developers about this before approving.

I presume we're still pending a good resolution here?

alexreinking:
> I presume we're still pending a good resolution here?

I think so... I would be interested to find ways to trip the depreciation warning without requiring any special build configuration.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -57,6 +59,18 @@ function(add_halide_generator TARGET)
         add_executable(${gen} ALIAS ${TARGET})
         target_link_libraries(${TARGET} PRIVATE Halide::Generator ${ARG_LINK_LIBRARIES})
 
+        if (NOT ARG_NO_DEFAULT_FLAGS AND NOT Halide_NO_DEFAULT_FLAGS)
+            # For crosscompiling builds, the Halide headers will be included using -isystem,
+            # which will cause all warnings to be ignored. This is not helpful, since
+            # we *want* deprecation warnings to be propagated. So we must set
+            # NO_SYSTEM_FROM_IMPORTED in order for it to be seen.
+            set_target_properties(${TARGET} PROPERTIES NO_SYSTEM_FROM_IMPORTED YES)
```
alexreinking, line 67: It seems that Clang has a special flag for this... https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers

The other approach is to try to make the deprecation warning fire on the class that is directly instantiated by the user, so that the warning appears to originate from user code.

steven-johnson:
Rebased and updated to main -- not sure where we stand on this. IIRC there were issues with deprecation warnings on templated methods not getting triggered, but that's moot (the warnings in question never got added and might not ever be added). Maybe this is better than nothing?

steven-johnson:
PTAL

steven-johnson:
Monday Morning Review Ping

alexreinking (approved):
Looks good.

PR #6561, opened 2022-01-13
Fix for top-of-tree LLVM

steven-johnson:


alexreinking (approved):
I still think it would be good to leave a comment explaining what the difference is here and why it matters to LLVM

PR #6563, opened 2022-01-15
[APP] Fix `hexagon_benchmarks` build (use two-var prefetch)

masahi:
Fix the following error when compiling `hexgaon_benchmarks` app:
```
Halide/distrib/include/Halide.h:24829:77: error: ‘Halide::Func& Halide::Func::prefetch(const T&, Halide::VarOrRVar, int, Halide::PrefetchBoundStrategy) [with T = Halide::GeneratorInput<Halide::Buffer<unsigned char> >]’ is deprecated: Call prefetch() with the two-var form instead. [-Werror=deprecated-declarations]
24829 |         return this->template as<Class>().Method(std::forward<Args>(args)...)
```

### DIFF
--- a/apps/hexagon_benchmarks/conv3x3_generator.cpp
+++ b/apps/hexagon_benchmarks/conv3x3_generator.cpp
@@ -52,7 +52,7 @@ public:
                 .vectorize(xi)
                 .unroll(yi);
             if (use_prefetch_sched) {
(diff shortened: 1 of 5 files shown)


pranavb-ca (approved):

steven-johnson (approved):
LGTM pending green

steven-johnson:
Failure is unrelated, landing

PR #6568, opened 2022-01-19
Rename Output to OutputFileType and deprecated Output

abadams:
It would be nice in future if we could promote Generator::Input and
Generator::Output to the Halide namespace. For that to happen, we need
to rename the enum Output first. This PR adds a new name for it, and
deprecates the old name. 

Not married to the name OutputFile. Other suggestions welcome.

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -24,7 +24,7 @@ class Buffer;
 struct Target;
 
 /** Enums specifying various kinds of outputs that can be produced from a Halide Pipeline. */
-enum class Output {
+enum class OutputFile {
     assembly,
     bitcode,
     c_header,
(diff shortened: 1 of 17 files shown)


src/Module.h
```diff
@@ -43,6 +43,26 @@ enum class Output {
     stmt_html,
 };
 
+class [[deprecated("Use OutputFile instead of Output")]] Output {
```
steven-johnson, line 46: IMHO we should still prefer to use `HALIDE_ATTRIBUTE_DEPRECATED()` here, as it allows downstream folks to disable it via `HALIDE_ALLOW_DEPRECATED`

steven-johnson (changes requested):
Should update the Python versions of these enums to use the same nomenclature, see PyEnums.cpp

Re the name, OutputFile is ok with me, but also maybe OutputFileType or OutputType?

zvookin:
I hate to ask this, and feel free to ignore if it doesn't make sense... The word "type" is heavily overloaded in a compiler project to begin with and I've come to think being careful around its use is a good idea. Specifically trying to use other words when one is not actually talking about a type in either the source or implementation language. E.g. an "output type" would be the type (buffer of ints or something) of an output.

To that end, maybe "OutputKind" is better. We're already using Kind in some places.

This is not a thing I'd object to the PR over, but figured I'd mention it in case improving our words and consistent use of them seems like a thing to do here.

steven-johnson:
> To that end, maybe "OutputKind" is better. We're already using Kind in some places.

SGTM, +1

steven-johnson (approved):

steven-johnson:
LGTM

steven-johnson:
Win32 is apparently unrelated (see https://github.com/halide/Halide/pull/6581 for hopeful fix), LGTM to land once green

PR #6574, opened 2022-01-20
Expand optional static-typing for Buffer to include dimensionality

steven-johnson:
Currently, `Halide::Runtime::Buffer<>`* allows you to provide a static ElementType, which allows you to enforce compile-time type checking to ensure incompatible Buffers aren't intermixed. You can also defeat this by using an ElementType of `void`, which defers all type checking to runtime.

Unfortunately, there isn't a way to do the same for the *dimensionality* of a Buffer; all checking for matching dimensionality is done at runtime. This PR proposes to extend `Halide::Runtime::Buffer<>`* to allow for optional specification of dimensionality as well, with only a slight incompatibility with existing code.

Currently, `H::R::B<>` is a template with two arguments:
- `H::R::B<typename T = void, int D = 4>`
    - `T`, which is the element type; if omitted, defaults to `void`.
    - `D`, which specifies the amount of storage to reserve for dimensionality (but *not* the actual dimensionality); if omitted, defaults to `4`. Note that I've often seen this argument specified on the (mistaken) assumption that it is the dimensionality! In practice, almost no code ever needs to specify this (with apps/hannk being the only notable counterexample that comes to mind).

This PR proposes to change H::R::B<> to a template with three arguments:
- `H::R::B<typename T = void, int Dims = -1, int InClassDimStorage = ...>`
    - `T`, which is the element type; if omitted, defaults to `void`. **This is unchanged from previous behavior.**
    - `Dims`, which specifies the dimensionality of the Buffer:
        - If `Dims` >= 0, the Buffer must always have `dimensions() == Dims`; attempts to (e.g.) construct or access such a buffer with the 'wrong' number of arguments will fail at compile time, assignment between Buffers with different dimensionality will fail at compile time, `slice()` and `embed()` will return properly-typed results, etc.
        - If `Dims` == -1, the Buffer can have any dimensionality, with compatibility being checked at runtime. (Basically, the same behavior as before.)
    - `InClassDimStorage`, which is the same as the old `D` argument. Its default value is determined like so:
       - If `Dims` is >= 1, its default value is `Dims`, since we shouldn't ever see any more or less.
       - If `Dims` is 0, we use a default value of 1 (just to be defensive against rogue code that assumes a valid pointer for dims in all cases)
       - Otherwise, it defaults to 4 (i.e., the same value as before)

TL;DR:
- Code that uses `Buffer<>`, `Buffer<void>`, or `Buffer<SomeType>` should work as-is, unchanged, with this change.
- Code that uses `Buffer<void, SomeInt>` or `Buffer<void, SomeInt>` will need attention to work after this change, but most uses I've seen fail at compile time and are easy to spot.

Note that the Generator infrastructure was expanded to add appropriate overloads for `Input<>` and `Output<>` to infer dimensionality from the Buffer specification (and metadata_tester_generator.cpp was modified to use this approach, for testing purposes); if this PR is approved, it is my intent to follow up with a PR that deprecates the ability to specify dimensionality in the old way.

*all this also applies to Halide::Buffer<> as well, just omitting it above for simplicitly

### DIFF
--- a/apps/fft/fft_aot_test.cpp
+++ b/apps/fft/fft_aot_test.cpp
@@ -61,7 +61,7 @@ int main(int argc, char **argv) {
         auto in = real_buffer();
         for (int j = 0; j < kSize; j++) {
             for (int i = 0; i < kSize; i++) {
-                in(i, j) = signal_1d[i] + signal_1d[j];
+                in(i, j, 0) = signal_1d[i] + signal_1d[j];
             }
         }
 
@@ -155,7 +155,7 @@ int main(int argc, char **argv) {
 
         for (size_t j = 0; j < kSize; j++) {
             for (size_t i = 0; i < kSize; i++) {
-                float sample = out(i, j);
+                float sample = out(i, j, 0);
                 float expected = cos(2 * kPi * (i / 16.0f + .125f));
                 if (fabs(sample - expected) > .001) {
                     std::cerr << "fft_inverse_c2r mismatch at (" << i << ", " << j << ") " << sample << " vs. " << expected << "\n";
--- a/apps/fft/fft_aot_test.cpp
+++ b/apps/fft/fft_aot_test.cpp
@@ -19,6 +19,9 @@ const int32_t kSize = 16;
 
 using Halide::Runtime::Buffer;
 
+// Note that real_buffer() is 3D (with the 3rd dimension having extent 0)
+// because the fft is written generically to require 3D inputs, even when they are real.
+// Hence, the resulting buffer must be accessed with buf(i, j, 0).
 Buffer<float, 3> real_buffer(int32_t y_size = kSize) {
     return Buffer<float, 3>::make_interleaved(kSize, y_size, 1);
 }
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -297,7 +297,12 @@ private:
                       "Number of arguments to Buffer() does not match static dimensionality");
         buf.dimensions = DimsSpecified;
         if constexpr (Dims == BufferDimsUnconstrained) {
-            buf.dim = (DimsSpecified <= InClassDimStorage) ? shape : new halide_dimension_t[DimsSpecified];
+            if constexpr (DimsSpecified <= InClassDimStorage) {
+                buf.dim = shape;
+            } else {
+                static_assert(DimsSpecified >= 1);
+                buf.dim = new halide_dimension_t[DimsSpecified];
+            }
         } else {
             static_assert(InClassDimStorage >= Dims);
             buf.dim = shape;
@@ -597,7 +602,10 @@ public:
     Buffer()
         : shape() {
         buf.type = static_halide_type();
-        make_static_shape_storage<0>();
+        // If Dims are statically known, must create storage that many.
+        // otherwise, make a zero-dimensional buffer.
+        constexpr int buf_dimensions = (Dims == BufferDimsUnconstrained) ? 0 : Dims;
+        make_static_shape_storage<buf_dimensions>();
     }
 
     /** Make a Buffer from a halide_buffer_t */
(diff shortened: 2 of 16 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -116,24 +116,29 @@ struct DeviceRefCount {
...
+ * InClassDimStorage defaults to 4, which should cover nearly all usage.
  *
  * The class optionally allocates and owns memory for the image using
  * a shared pointer allocated with the provided allocator. If they are
  * null, malloc and free are used.  Any device-side allocation is
  * considered as owned if and only if the host-side allocation is
  * owned. */
-template<typename T = void, int D = 4>
+template<typename T = void,
+         int Dims = -1,
```
zvookin, line 134: This probably ought to introduce a constant, probably declared in class, for the -1. The name is something like "DimsUnconstrained".

steven-johnson: `DynamicDims` is defined (line 188) but it isn't visible to the template parameters; AFAICT I'd need to declare it outside the class. Would that be preferable?

steven-johnson: Done.

apps/fft/fft_aot_test.cpp
```diff
@@ -61,7 +61,7 @@ int main(int argc, char **argv) {
         auto in = real_buffer();
         for (int j = 0; j < kSize; j++) {
             for (int i = 0; i < kSize; i++) {
-                in(i, j) = signal_1d[i] + signal_1d[j];
+                in(i, j, 0) = signal_1d[i] + signal_1d[j];
```
abadams, line 64: Add a comment explaining why real_buffer() is not 2D (The fft requires is written generically to require 3D inputs, even when they are real).

src/runtime/HalideBuffer.h
```diff
@@ -521,7 +556,9 @@ class Buffer {
     }
 
     /** Get the dimensionality of the buffer. */
+    // TODO: make constexpr, optimize for const case
```
abadams, line 559: Why not handle this now. I think it's enough to just have an if (has_static_dimensions)

src/runtime/HalideBuffer.h
```diff
@@ -885,6 +932,7 @@ class Buffer {
      * take ownership of the data, and does not set the host_dirty flag. */
     template<typename Array, size_t N>
     explicit Buffer(Array (&vals)[N]) {
+        // TODO: this could probably be made constexpr
```
abadams, line 935: IMO either handle the TODO now or remove it. If the compiler can constant-fold it, it will. There shouldn't be any benefit in explicitly marking it as constexpr.

abadams:
Updating the apps to use static dimensionality in their generator inputs is left for a future PR? If you don't intend to do it right away, perhaps open a tracking issue.

abadams (approved):

steven-johnson:
> Updating the apps to use static dimensionality in their generator inputs is left for a future PR? If you don't intend to do it right away, perhaps open a tracking issue.

https://github.com/halide/Halide/issues/6584

A lot of that is done in https://github.com/halide/Halide/pull/6575; it would be easy enough to pull out those changes (without the deprecation part) and merge them.

apps/fft/fft_aot_test.cpp
steven-johnson: Done

src/runtime/HalideBuffer.h
steven-johnson: Done.

src/runtime/HalideBuffer.h
steven-johnson: Done (comment removed)

abadams:
Oh right, I forgot about that other PR.

steven-johnson:
Working on https://github.com/halide/Halide/pull/6585 found a few oversights I missed, PTAL if you like

PR #6575, opened 2022-01-21
Deprecate GeneratorInput/Output ctors that are obsolete after #6574

steven-johnson:
This deprecates the ctors to Generator Input/Output that are "redundant" after #6574 lands (i.e.: type and dimensions should be specified in the Buffer<T, D> declaration instead of as arguments). Also updates all our existing Generators to conform. Additive to #6574.


steven-johnson:
(converting to draft because this is going to landed in smaller pieces, but I don't want it to get lost yet)

Issue #6054, opened 2021-05-24
AMX not able to scale larger than a k dimension of 4/2


frengels:
When multiplying and MxK matrix with a KxM matrix the current implementation in #5818 only allows K dimensions of 4 or 2 for int8 and bf16 respectively.
AMX's tiled operations should be able to scale to higher K values, as long as they're a multiple of 4 bytes (4 elements for int8, 2 for bf16).

frengels:
I've been looking into this issue, it looks like when splitting the `k` dimension it creates a `For` within the tile allocation. When there's already an outer `k` loop this is fine because there's 2 loops. But if the inner loop covers the entire matrix we want to process then there's again only a single loop present, so as a result it's impossible to discern whether the aim is to process in `k` dimensions of 4 bytes or whether to process the entire matrix.

The full matrix in the following example has a `k` dimension of 16

case where `k` == 4, rro indicates the outer accumulating loop.
```
produce matmul {
 matmul[ramp(0, 1, 48)] = x48(0)
 for (matmul.s1.r9$x.rro, 0, 4) {
  matmul[ramp(0, 1, 48)] = (int32x48)vector_reduce(Add, (int32x192((int8x192)lhs[(ramp(x32((matmul_global_wrapper$0.min.1 + matmul_global_wrapper$0.s0.x.mmxi.base.s)*lhs.stride.1), x32(lhs.stride.1), 6) + x48(ramp(matmul.s1.r9$x.rro*4, 1, 4))) - x192((lhs.min.1*lhs.stride.1) + lhs.min.0)])*x6(int32x32((int8x32)rhs[ramp(ramp((((matmul_global_wrapper$0.min.0 + matmul_global_wrapper$0.s0.y.mmyi.base.s)*rhs.stride.1) + (matmul.s1.r9$x.rro*rhs.stride.2)) - ((rhs.min.2*rhs.stride.2) + ((rhs.min.1*rhs.stride.1) + rhs.min.0)), 1, 4), x4(rhs.stride.1), 8)])))) + matmul[ramp(0, 1, 48)]
 }
}
```

where `k` == 8, you can see 2 loops. The inner can be seen as the increments of 4 bytes.
```
produce matmul {
 matmul[ramp(0, 1, 48)] = x48(0)
 for (matmul.s1.r9$x.rro, 0, 2) {
  for (matmul.s1.r9$x.rri.rri, 0, 2) {
[...]

PR #6582, opened 2022-01-26
Allow AMX instructions with K dimension larger than 4 bytes

frengels:
Addresses #6054 
The current version of AMX can only handle K dimensions which are 4 bytes in size. This is because the way the right hand side (rhs) matrix is laid out is different from the left hand side (lhs). Where the lhs matrix is a regular MxK matrix the rhs is instead K/4xNx4 for i8/u8 and K/2xNx2 for bf16. To handle this an extra `get_3d_rhs_tile_index` function was added.
The correctness tests have been further expanded to handle a variety of cases.
* tile K dimension of 4 bytes
This tests the previous code, using 2d/1d tile index pattern matcher
* tile K dimension larger than 4 bytes but equal to the matrix's K dimension
This handles one of the 2 patterns checked for
* tile K dimension larger than 4 bytes and not equal to the matrix's K dimension
This handles the other pattern which is important for getting the correct offset

### DIFF
--- a/src/ExtractTileOperations.cpp
+++ b/src/ExtractTileOperations.cpp
@@ -261,6 +261,43 @@ Tile<3> get_3d_rhs_tile_index(const Expr &e, int element_width) {
 
     return {true, base, {stride, 0, 0}, {tile_x, tile_y, tile_r}};
 }
+
+struct BaseStride {
+    bool result{false};
+    Expr base{};
+    Expr stride{};
+};
+
+BaseStride get_rhs_tile_index(const Expr &index, int element_width, int tile_x, int tile_y, int tile_r) {
+    const auto rhs_tile2 = get_2d_tile_index(index);
+
+    if (!rhs_tile2.result) {
+        const auto rhs_tile1 = get_1d_tile_index(index);
+
+        if (!rhs_tile1.result) {
+            auto rhs_tile3 = get_3d_rhs_tile_index(index, element_width);
+            if (rhs_tile3.extent[0] != tile_x || rhs_tile3.extent[1] != tile_y || rhs_tile3.extent[2] != tile_r) {
+                return {};
+            }
+
+            return {true, rhs_tile3.base, rhs_tile3.stride[0] * element_width};
+        } else {
+            if (rhs_tile1.extent[0] != tile_y * tile_r) {
+                return {};
+            }
+
+            // times 4 because of the rhs layout
+            return {true, rhs_tile1.base, rhs_tile1.stride[0] * (4 / element_width)};
+        }
+    } else {
+        if (tile_y != rhs_tile2.extent[0] || tile_r != rhs_tile2.extent[1]) {
+            return {};
+        }
+
+        return {true, rhs_tile2.base, rhs_tile2.stride[0]};
+    }
+}
+
 struct Matmul {
     bool result = false;
     Stmt stmt;
@@ -315,14 +352,24 @@ Matmul convert_to_matmul(const Store *op, const string &new_name, AMXOpType op_t
 
     const auto *lhs_load = matches[0].as<Load>();
     const auto *rhs_broadcast = matches[1].as<Broadcast>();
-    if (!lhs_load || !rhs_broadcast) {
+
+    const Cast *rhs_cast = nullptr;
+
+    if (lhs_load && !rhs_broadcast) {
+        // now working on a larger k dimension
+        rhs_cast = matches[1].as<Cast>();
+    } else {
+        rhs_cast = rhs_broadcast->value.as<Cast>();
+    }
+
+    if (!lhs_load || !rhs_cast) {
         return {};
     }
-    const auto *rhs_cast = rhs_broadcast->value.as<Cast>();
+
     if (rhs_cast) {
         if (op_type == AMXOpType::Int8) {
             if (!(rhs_cast->value.type().element_of() == Int(8) || rhs_cast->value.type().element_of() == UInt(8))) {
-                user_assert(false) << "Expected rhs cast of i8/u8";
+                user_error << "Expected rhs cast of i8/u8, got " << rhs_cast->value.type();
             }
         } else {  // AMXOpType::Bfloat16
             user_assert(rhs_cast->value.type().element_of() == BFloat(16)) << "Expected rhs cast of bf16";
@@ -350,28 +397,14 @@ Matmul convert_to_matmul(const Store *op, const string &new_name, AMXOpType op_t
     Expr rhs_base;
     Expr rhs_stride;
 
-    const auto rhs_tile2 = get_2d_tile_index(rhs_load->index);
-    if (!rhs_tile2.result) {
-        const auto rhs_tile1 = get_1d_tile_index(rhs_load->index);
+    auto opt_base_stride = get_rhs_tile_index(rhs_load->index, amx_op_type_size(op_type), tile_x, tile_y, tile_r);
 
-        if (!rhs_tile1.result) {
-            return {};
-        }
-
-        if (rhs_tile1.extent[0] != tile_y * tile_r) {
-            return {};
-        }
-
-        rhs_base = rhs_tile1.base;
-        rhs_stride = rhs_tile1.stride[0];
-    } else {
-        if (tile_y != rhs_tile2.extent[0] || tile_r != rhs_tile2.extent[1]) {
-            return {};
-        }
-
-        rhs_base = rhs_tile2.base;
-        rhs_stride = rhs_tile2.stride[0];
+    if (!opt_base_stride.result) {
+        return {};
     }
+
+    rhs_base = opt_base_stride.base;
+    rhs_stride = opt_base_stride.stride;
 
     if (op->index.type().lanes() != tile_x * tile_y ||
         factor != tile_r) {
@@ -388,7 +421,8 @@ Matmul convert_to_matmul(const Store *op, const string &new_name, AMXOpType op_t
     auto rhs_var = Variable::make(Handle(), rhs_load->name);
     const auto &rhs_load_type = rhs_load->type;
     auto rhs_type = rhs_load_type.with_lanes(1024 / element_width);
-    auto rhs = Call::make(rhs_type, "tile_load", {1, tile_y * tile_r * element_width, rhs_var, rhs_base * element_width, rhs_stride * tile_y * element_width}, Call::Intrinsic);
+
+    auto rhs = Call::make(rhs_type, "tile_load", {tile_r / (4 / element_width), tile_y * 4, rhs_var, rhs_base * element_width, rhs_stride}, Call::Intrinsic);
     auto res_type = amx_op_type_result_type(op_type);
 
     // {rows, colbytes, acc, out, lhs, rhs}
@@ -528,6 +562,7 @@ class ExtractTileOperations : public IRMutator {
             found_tile_x = matmul.tile_x;
             found_tile_y = matmul.tile_y;
             found_tile_r = matmul.tile_r;
+
             return matmul.stmt;
         }
 
@@ -542,7 +577,7 @@ class ExtractTileOperations : public IRMutator {
         }
 
         // Otherwise there is some other operation using the allocation, so we cannot use the AMX instructions
-        user_assert(false) << "Found non-tile operations for AMX tile allocation";
+        user_error << "Found non-tile operations for AMX tile allocation";
         return op;
     }
 };
--- a/src/ExtractTileOperations.cpp
+++ b/src/ExtractTileOperations.cpp
@@ -56,16 +56,26 @@ const auto wild_i32x = Variable::make(Int(32, 0), "*");
 
 Tile<1> get_1d_tile_index(const Expr &e) {
     if (const auto *r1 = e.as<Ramp>()) {
-        Expr pattern = ((wild_i32 * wild_i32) + wild_i32) * wild_i32;
 
-        std::vector<Expr> results;
-        if (!expr_match(pattern, r1->base, results)) {
-            return {};
+        const auto stride_var = Variable::make(Int(32), "stride");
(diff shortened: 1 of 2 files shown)


src/ExtractTileOperations.cpp
```diff
@@ -143,6 +164,126 @@ Tile<3> get_3d_tile_index(const Expr &e) {
...
+    const Ramp *r_ramp = bc_ramp->value.as<Ramp>();
+
+    if (!r_ramp) {
+        return {};
+    }
+
+    int tile_r = r_ramp->lanes;
+
+    // get the base and stride
+    const Broadcast *base_stride_bc = add_lhs->b.as<Add>()->b.as<Broadcast>();
```
rootjalex, line 209: This should be:
```
const Broadcast *base_stride_bc = dim_expr->b.as<Broadcast>();
```

rootjalex: Actually, it seems that `base_stride_bc` is the exact same as `y_bc`. Seems to be a bit of duplicated logic here, should one of these series of checks be checking the `a` side of `add_lhs`?

src/ExtractTileOperations.cpp
```diff
@@ -143,6 +164,126 @@ Tile<3> get_3d_tile_index(const Expr &e) {
...
+        return {};
+    }
+
+    const Ramp *base_stride_ramp = base_stride_bc->value.as<Ramp>();
+
+    if (!base_stride_ramp) {
+        return {};
+    }
+
+    Expr base = base_stride_ramp->base.as<Broadcast>()->value;
```
rootjalex, line 221: This is unsafe, `base_stride_ramp->base.as<Broadcast>()` could return `nullptr`

steven-johnson:
You may need to sync this to the master branch to fix the build breakages.

src/ExtractTileOperations.cpp
frengels: resolved with 20b0c3aec52bd776a68bd5f9631d21c3b7685166

src/ExtractTileOperations.cpp
```diff
@@ -39,12 +39,33 @@ Type amx_op_type_result_type(AMXOpType op_ty) {
...
+        return -1;
+    }
+}
+
 const auto wild_i32 = Variable::make(Int(32), "*");
 const auto wild_i32x = Variable::make(Int(32, 0), "*");
 
 Tile<1> get_1d_tile_index(const Expr &e) {
     if (const auto *r1 = e.as<Ramp>()) {
-        return {true, r1->base, {r1->stride}, {r1->lanes}};
```
rootjalex, line 47: Do we want some logic akin to HexagonOptimize's `apply_commutative_patterns`, here and below?

rootjalex, line 47: Do we want some logic akin to HexagonOptimize's `apply_commutative_patterns`, here and below?

src/ExtractTileOperations.cpp
```diff
@@ -143,6 +164,138 @@ Tile<3> get_3d_tile_index(const Expr &e) {
     return {true, base, {x_stride, 0, r_stride}, {x_tile, y_tile, r_tile}};
 }
 
+Tile<3> get_3d_rhs_tile_index(const Expr &e, int element_width) {
```
rootjalex, line 167: Perhaps not necessary for this PR, but we really should express these as pattern-based rules, they would be far easier to parse and debug

rootjalex, line 167: Perhaps not necessary for this PR, but we really should express these as pattern-based rules, they would be far easier to parse and debug

src/ExtractTileOperations.cpp
frengels: would that be something like an `expr_match` that also checks the commutative?

rootjalex: Yes, althought I believe `apply_commutative_patterns` only checks one level of commutation. I'm just wondering if we should be checking for all of these:
```
((wild_i32 * wild_i32) + wild_i32) * wild_i32
wild_i32 * ((wild_i32 * wild_i32) + wild_i32)
(wild_i32 + (wild_i32 * wild_i32)) * wild_i32
wild_i32 * (wild_i32 + (wild_i32 * wild_i32))
```

frengels: That would make sense, to do that I feel like it requires to implement a new `expr_match` which applies those rules?

src/ExtractTileOperations.cpp
```diff
@@ -39,12 +39,33 @@ Type amx_op_type_result_type(AMXOpType op_ty) {
...
     if (const auto *r1 = e.as<Ramp>()) {
-        return {true, r1->base, {r1->stride}, {r1->lanes}};
+        Expr pattern = ((wild_i32 * wild_i32) + wild_i32) * wild_i32;
+
+        std::vector<Expr> results;
+        if (!expr_match(pattern, r1->base, results)) {
+            return {};
+        }
+
+        auto stride = std::move(results[1]);
```
rootjalex, line 66: Why is the stride the second arg to the multiplication? It could easily be the first, no?

src/ExtractTileOperations.cpp
frengels: I've played around a bit with extending the `expr_match` to support commutative operations where relevant, but I think for now I've also settled on the map approach.

test/correctness/tiled_matmul.cpp
```diff
@@ -225,38 +259,31 @@ int main(int argc, char **argv) {
...
-        printf("Success!\n");
     }
 
-    printf("Running AMX matmul (signed/unsigned)\n");
-    if (!matmul_su()) {
-        return -1;
-    } else {
-        printf("Success!\n");
-    }
+    // llvm >= 13.0 is required for unsigned and float AMX instructions
```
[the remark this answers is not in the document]
frengels: Should be resolved now

src/ExtractTileOperations.cpp
frengels: I have added a check for the stride being an `IntImm`, that should prevent the other term being mistakenly matched.

src/ExtractTileOperations.cpp
frengels: Yes, some of it is a bit messy. It would also be nice to have the ability to wildcard on the number of lanes so that arbitrary `Ramp`s and `Broadcast`s could be matched and have their number of lanes retrieved.

src/ExtractTileOperations.cpp
```diff
@@ -39,12 +39,46 @@ Type amx_op_type_result_type(AMXOpType op_ty) {
...
-        return {true, r1->base, {r1->stride}, {r1->lanes}};
+
+        const auto stride_var = Variable::make(Int(32), "stride");
+        const auto v1 = Variable::make(Int(32), "v1");
+        const auto v2 = Variable::make(Int(32), "v2");
+        const auto v3 = Variable::make(Int(32), "v3");
+
+        Expr patterns[] = {
+            ((v1 * stride_var) + v2) * v3,
+            v3 * ((v1 * stride_var) + wild_i32),
```
rootjalex, line 67: nit: `wild_i32` -> `v2`

src/ExtractTileOperations.cpp
```diff
@@ -39,12 +39,46 @@ Type amx_op_type_result_type(AMXOpType op_ty) {
...
 Tile<1> get_1d_tile_index(const Expr &e) {
     if (const auto *r1 = e.as<Ramp>()) {
-        return {true, r1->base, {r1->stride}, {r1->lanes}};
+
+        const auto stride_var = Variable::make(Int(32), "stride");
+        const auto v1 = Variable::make(Int(32), "v1");
+        const auto v2 = Variable::make(Int(32), "v2");
+        const auto v3 = Variable::make(Int(32), "v3");
+
+        Expr patterns[] = {
```
rootjalex, line 65: For completeness, should we also look for the following:
```
((stride_var * v1) + v2) * v3,
v3 * ((stride_var * v1) + wild_i32),
(v2 + (stride_var * v1)) * v3,
v3 * (v2 + (stride_var * v1)),
```
Alternatively the check that `stride` is constant could also check if `v1` is constant.

src/ExtractTileOperations.cpp
```diff
@@ -39,12 +39,46 @@ Type amx_op_type_result_type(AMXOpType op_ty) {
...
+            (v2 + (v1 * stride_var)) * v3,
+            v3 * (v2 + (v1 * stride_var)),
+        };
+
+        std::map<std::string, Expr> matches;
+        for (const auto &pattern : patterns) {
+            if (expr_match(pattern, r1->base, matches)) {
+                auto stride = std::move(matches["stride"]);
+                if (!stride.as<IntImm>()) {
+                    return {};
```
rootjalex, line 77: should this be a `continue`?

src/ExtractTileOperations.cpp
rootjalex: Ah true, that would be quite convenient, perhaps such support could be added to the pattern matcher in a future PR.

src/ExtractTileOperations.cpp
frengels: addressed in 672d86ff37c27b685b647c3216fb4c1c426a3575

src/ExtractTileOperations.cpp
frengels: addressed in 33408424eaca3472e043701c1f83acb14e819f28

src/ExtractTileOperations.cpp
frengels: same as above 33408424eaca3472e043701c1f83acb14e819f28, I inverted the condition of the if statement

rootjalex:
Lgtm. @abadams ?

src/ExtractTileOperations.cpp
abadams: In comment form, could you add a fairly complete list of example Exprs that would match? That will help whoever changes it to pattern matching in the future.

src/ExtractTileOperations.cpp
```diff
@@ -197,14 +369,24 @@ Matmul convert_to_matmul(const Store *op, const string &new_name, AMXOpType op_t
...
+    if (!lhs_load || !rhs_cast) {
         return {};
     }
-    const auto *rhs_cast = rhs_broadcast->value.as<Cast>();
+
     if (rhs_cast) {
         if (op_type == AMXOpType::Int8) {
             if (!(rhs_cast->value.type().element_of() == Int(8) || rhs_cast->value.type().element_of() == UInt(8))) {
-                user_assert(false) << "Expected rhs cast of i8/u8";
+                user_error << "Expected rhs cast of i8/u8, got " << rhs_cast->value.type();
```
abadams, line 389: This is likely not enough context for a user_error. Is there any identifying information as to which Func is involved here? Maybe printing the full statement will be more of a clue, along with the message that AMX doesn't support casts to other than i8/u8 on the RHS (if that's accurate).

src/ExtractTileOperations.cpp
```diff
@@ -197,14 +369,24 @@ Matmul convert_to_matmul(const Store *op, const string &new_name, AMXOpType op_t
 
     const auto *lhs_load = matches[0].as<Load>();
     const auto *rhs_broadcast = matches[1].as<Broadcast>();
-    if (!lhs_load || !rhs_broadcast) {
+
+    const Cast *rhs_cast = nullptr;
+
+    if (lhs_load && !rhs_broadcast) {
+        // now working on a larger k dimension
```
abadams, line 376: I don't understand this comment. It seems to describe the PR rather than the code here.

src/ExtractTileOperations.cpp
```diff
@@ -143,6 +183,138 @@ Tile<3> get_3d_tile_index(const Expr &e) {
...
+                return {};
+            }
+
+            return {true, rhs_tile3.base, rhs_tile3.stride[0] * element_width};
+        } else {
+            if (rhs_tile1.extent[0] != tile_y * tile_r) {
+                return {};
+            }
+
+            // times 4 because of the rhs layout
```
abadams, line 306: Could you expand this comment a bit? I assume it's some artifact of how AMX lays things out in memory?

src/ExtractTileOperations.cpp
frengels: added some more detail to the error here b56aad1eff11e7ec75ba58280e406f568b136949.

src/ExtractTileOperations.cpp
frengels: I've added a slightly larger description here 7be1b5b6867ce22ae36bd4f15e0f07545f3c5c81.
And with further explanation for the layout of the matrices a80fc766ff63a2002a4fc459c04b17ef0127d784.

src/ExtractTileOperations.cpp
frengels: This should be addressed by 81bacbb680ef74ab47ec30700a4868b272e8f281.

src/ExtractTileOperations.cpp
abadams: I think this is the last remaining comment request. Once you push a new commit, the buildbots should get themselves unconfused.

frengels: oops, almost didn't see this, should be in 314656cbb1084923ace01b727623124f2dc5ee3b.
I remember the reason why I didn't use patterns for this is because I can't match on a wildcarded number of lanes. And any attempt to use `.with_lanes(0)` would make asserts in the `make` functions for the nodes fail since they often require at least 1 lane, but 0 is required to make `expr_match` ignore the amount of lanes when matching.

abadams: Sure, but please add example patterns *as a comment* to help future readers

abadams:
Thanks! Looks like it just needs a clang-format

abadams (approved):

PR #6588, opened 2022-01-26
Add static-dimension annotations to Buffer usage in python_bindings and tutorials

steven-johnson:
Plus more fixes to corner cases in src/ and halide_image_io that I missed earlier

### DIFF
--- a/python_bindings/correctness/addconstant_generator.cpp
+++ b/python_bindings/correctness/addconstant_generator.cpp
@@ -16,31 +16,31 @@ public:
     Input<float> constant_float{"constant_float"};
     Input<double> constant_double{"constant_double"};
 
-    Input<Buffer<uint8_t>> input_uint8{"input_uint8", 1};
-    Input<Buffer<uint16_t>> input_uint16{"input_uint16", 1};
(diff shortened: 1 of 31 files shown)


abadams:
I'm not convinced about the changes to the tutorials. Adding an explicit dimensionality on the left hand side of a realize call is redundant with the number of arguments passed on the right hand side. I don't think it adds any clarity.

steven-johnson:
> I'm not convinced about the changes to the tutorials. Adding an explicit dimensionality on the left hand side of a realize call is redundant with the number of arguments passed on the right hand side. I don't think it adds any clarity.

If we want to avoid redundancy there, we could just use `auto`...

Not gonna spend time arguing this, but IMHO putting thus into the tutorials is one of the most important things to do; the idioms we show new users there is what they will use as they write new code. If we want them to use static dimensions when appropriate (and I do want them to), we should show them how to do so here.

abadams:
auto doesn't quite work because you need a cast to an explicit type of Buffer for operator() to work on the resulting object. I'm not convinced it's appropriate to use explicit static dimensionality in a casual JIT context in which so much other stuff is implicit or automatic.

steven-johnson:
(going to land this in pieces instead)

PR #6589, opened 2022-01-27
Various fixes to static-dimensioned Buffer

steven-johnson:
More details that should have been in #6574:
- Buffer needed to make some methods constexpr, and also had a broken return value for `as<>()`
- Various templated methods needed to add `int Dims` as a parameter
- Generator::add_input() and add_output() needed specializations for static-buffer types
- sliced() and embedded() should use the default values for InClassDimStorage
- halide_image_io should use a named constant

### DIFF
--- a/src/Func.h
+++ b/src/Func.h
@@ -740,8 +740,8 @@ public:
     explicit Func(Internal::Function f);
 
     /** Construct a new Func to wrap a Buffer. */
-    template<typename T>
-    HALIDE_NO_USER_CODE_INLINE explicit Func(Buffer<T> &im)
+    template<typename T, int Dims>
+    HALIDE_NO_USER_CODE_INLINE explicit Func(Buffer<T, Dims> &im)
         : Func() {
         (*this)(_) = im(_);
     }
@@ -2560,7 +2560,7 @@ HALIDE_NO_USER_CODE_INLINE T evaluate(JITUserContext *ctx, const Expr &e) {
         << " as a scalar of type " << type_of<T>() << "\n";
     Func f;
     f() = e;
-    Buffer<T> im = f.realize(ctx);
+    Buffer<T, 0> im = f.realize(ctx);
     return im();
 }
 
@@ -2615,7 +2615,7 @@ HALIDE_NO_USER_CODE_INLINE T evaluate_may_gpu(const Expr &e) {
     Func f;
     f() = e;
     Internal::schedule_scalar(f);
-    Buffer<T> im = f.realize();
+    Buffer<T, 0> im = f.realize();
     return im();
 }
 
--- a/src/Generator.h
+++ b/src/Generator.h
@@ -1272,7 +1272,7 @@ enum class IOKind { Scalar,
  * -- Assignment of a Buffer<>, with compatible type and dimensions,
  * causing the Input<Buffer<>> to become a precompiled buffer in the generated code.
  */
-template<typename T = void>
+template<typename T = void, int Dims = Buffer<>::BufferDimsUnconstrained>
 class StubInputBuffer {
     friend class StubInput;
     template<typename T2>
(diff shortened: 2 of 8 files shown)


src/Generator.h
```diff
@@ -3271,10 +3271,26 @@ class GeneratorBase : public NamesInterface, public GeneratorContext {
         return p;
     }
 
-    // Create a Input<Buffer> or Input<Func> with compile-time type
+    // Create Input<Buffer> with dynamic type & dimensions
     template<typename T,
-             typename std::enable_if<T::has_static_halide_type>::type * = nullptr>
+             typename std::enable_if<!std::is_arithmetic<T>::value && !std::is_same<T, Halide::Func>::value>::type * = nullptr>
+    GeneratorInput<T> *add_input(const std::string &name, const Type &t, int dimensions) {
+        static_assert(!T::has_static_dimensions, "You can only call this version of add_input() for a Buffer<T, D> where T is void or omitted .");
```
abadams, line 3278: These static assert conditions seem to be the same on both lines

src/Generator.h
```diff
@@ -3283,6 +3299,19 @@ class GeneratorBase : public NamesInterface, public GeneratorContext {
         return p;
     }
 
+    // Create Input<Buffer> with compile-time type & dimensions
+    template<typename T,
+             typename std::enable_if<!std::is_arithmetic<T>::value && !std::is_same<T, Halide::Func>::value>::type * = nullptr>
+    GeneratorInput<T> *add_input(const std::string &name) {
+        static_assert(T::has_static_dimensions, "You can only call this version of add_input() for a Buffer<T, D> where T is not void.");
```
abadams, line 3306: Same comment here

src/Generator.h
```diff
@@ -3308,10 +3336,24 @@ class GeneratorBase : public NamesInterface, public GeneratorContext {
...
+        param_info_ptr->filter_outputs.push_back(p);
+        return p;
+    }
+
+    // Create Output<Buffer> with dynamic type & dimensions
     template<typename T,
-             typename std::enable_if<!std::is_arithmetic<T>::value>::type * = nullptr>
+             typename std::enable_if<!std::is_arithmetic<T>::value && !std::is_same<T, Halide::Func>::value>::type * = nullptr>
     GeneratorOutput<T> *add_output(const std::string &name, const Type &t, int dimensions) {
+        static_assert(!T::has_static_dimensions, "You can only call this version of add_output() for a Buffer<T, D> where T is void or omitted .");
```
abadams, line 3355: and here

src/Generator.h
```diff
@@ -3332,6 +3376,20 @@ class GeneratorBase : public NamesInterface, public GeneratorContext {
         return p;
     }
 
+    // Create Output<Buffer> with compile-time type & dimensions
+    template<typename T,
+             typename std::enable_if<!std::is_arithmetic<T>::value && !std::is_same<T, Halide::Func>::value>::type * = nullptr>
+    GeneratorOutput<T> *add_output(const std::string &name) {
+        static_assert(T::has_static_dimensions, "You can only call this version of add_output() for a Buffer<T, D> where T is not void.");
```
abadams, line 3383: and here, unless I'm really not understanding something

tools/halide_image_io.h
```diff
@@ -2088,31 +2090,32 @@ struct ImageTypeConversion {
         // as documentation and a backstop against breakage.
         static_assert(!ImageType::has_static_halide_type,
                       "This variant of convert_image() requires a dynamically-typed image");
+        constexpr int DimsUnconstrained = Internal::DimsUnconstrained;
```
abadams, line 2093: Would "AnyDims" be a better name than "DimsUnconstrained"?

src/Generator.h
steven-johnson: whoops, first should be `has_static_halide_type`

src/Generator.h
steven-johnson: ditto

src/Generator.h
steven-johnson: ditto

src/Generator.h
steven-johnson: ditto

tools/halide_image_io.h
steven-johnson: `Unconstrained` is the term we use in HalideBuffer.h so that's what I used here.

steven-johnson:
PTAL

tools/halide_image_io.h
abadams: I know, I was reconsidering the name in general. Seeing it in usage makes it seem clunky.

src/Func.h
```diff
@@ -2615,7 +2615,7 @@ HALIDE_NO_USER_CODE_INLINE T evaluate_may_gpu(const Expr &e) {
     Func f;
     f() = e;
     Internal::schedule_scalar(f);
-    Buffer<T> im = f.realize();
+    Buffer<T, 0> im = f.realize();
```
abadams, line 2618: This is just a minor stack space optimization, right? I don't see that this zero makes much of a difference because the dims array is never accessed. I have no objection to adding it, I just want to make sure I understand what's happening.

tools/halide_image_io.h
steven-johnson: Ah, gotcha. I think I originally called it "DynamicDims" and @zvookin suggested "DimsUnconstrained". "AnyDims" is fine with me too and definitely terser. I'll make the change everywhere if you like.

src/Func.h
steven-johnson: Correct, it's just a tiny optimization. 

(On that note: the code for Buffer currently always allocates space for at least one halide_dimension_t, even for zero-dim buffers, on the assumption that there is probably sloppy code out there that assumes that buffer.dim always points to something valid...)

abadams (approved):

PR #6590, opened 2022-01-27
Fixes the Pytorch Wrapper Codegen for CPU-only machines.

mgharbi:
This PR does two things:
1. split the helper function that wraps PyTorch tensors into Halide buffers into 2: one for CPU tensors, one for GPU tensors. Before the PR, the wrapper may fail on CPU-only machine because `halide_cuda_device_interface` is missing.
2. add a default `__user_context = nullptr` in CPU-only ops. The auto-scheduler can create intermediate functions that have a __user_context input. This case was not handled by the CodeGen, so compilation would fail. We now create a default null context instead.

### DIFF
--- a/src/runtime/HalidePyTorchHelpers.h
+++ b/src/runtime/HalidePyTorchHelpers.h
@@ -15,7 +15,8 @@
 
 #include "HalideBuffer.h"
 
-#include "HalideRuntimeCuda.h"
+// Forward declare the cuda_device_interface, for tensor wrapper.
+const halide_device_interface_t *halide_cuda_device_interface();
 
 #define HLPT_CHECK_CONTIGUOUS(x) AT_ASSERTM(x.is_contiguous(), #x " must be contiguous")
 #define HLPT_CHECK_CUDA(x) AT_ASSERTM(x.type().is_cuda(), #x " must be a CUDA tensor")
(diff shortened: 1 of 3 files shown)


src/runtime/HalidePyTorchHelpers.h
```diff
@@ -90,16 +91,29 @@ inline Buffer<scalar_t> wrap(at::Tensor &tensor) {
 #endif
     Buffer<scalar_t> buffer;
```
steven-johnson, line 92: ...why not just

    return Buffer<scalar_t>(pData, dims);

?

src/runtime/HalidePyTorchHelpers.h
```diff
@@ -90,16 +91,29 @@ inline Buffer<scalar_t> wrap(at::Tensor &tensor) {
...
+template<class scalar_t>
+inline Buffer<scalar_t> wrap_cuda(at::Tensor &tensor) {
+    check_type<scalar_t>(tensor);
+    std::vector<int> dims = get_dims(tensor);
+#if HL_PYTORCH_API_VERSION >= 13
+    scalar_t *pData = tensor.data_ptr<scalar_t>();
+#else
+    scalar_t *pData = tensor.data<scalar_t>();
+#endif
+    Buffer<scalar_t> buffer;
```
steven-johnson, line 109: ...why not

    Buffer<scalar_t> buffer(dims);

and then delete line 112?

steven-johnson (approved):
LGTM pending green, with nits

src/runtime/HalidePyTorchHelpers.h
mgharbi: fair!

src/runtime/HalidePyTorchHelpers.h
mgharbi: done

steven-johnson:
Please also use the `run-clang-format.sh` script, or manually fix errors if you prefer

PR #6591, opened 2022-01-27
Convert Buffer<> usage in python_bindings/ to use static dimensions

steven-johnson:
...where it seems sensible to do so, that is.


steven-johnson:
Monday Morning Review Ping

abadams (approved):

PR #6592, opened 2022-01-27
Convert Buffer<> usage in test/generators to use static dimensions

steven-johnson:
...where it seems sensible to do so, that is.

(also test/integration since it's small)


steven-johnson:
Monday Morning Review Ping

abadams (approved):

PR #6602, opened 2022-02-01
Reenable warning about unscheduled update definitions

abadams:
and fix associated issues in the tests and apps.

This is an old warning that stopped triggering because it wasn't tested.
We should either remove it, fix the trigger conditions, or perhaps make
it an error. This PR fixes the trigger conditions and fixes all
instances of the warning in our tests and apps.

The warning triggers if you schedule some but not all of the update
definitions of a Func. It's to protect against the common error of only
scheduling the pure definition of something like a summation. The
warning can be suppressed by inserting a call to func.update(idx).

### DIFF
--- a/test/correctness/tuple_reduction.cpp
+++ b/test/correctness/tuple_reduction.cpp
@@ -61,14 +61,13 @@ int main(int argc, char **argv) {
             f.hexagon(y).vectorize(x, 32);
         }
         for (int i = 0; i < 10; i++) {
+            f.update(i);
             if (i & 1) {
                 if (target.has_gpu_feature()) {
                     f.update(i).gpu_tile(x, y, xo, yo, xi, yi, 16, 16);
                 } else if (target.has_feature(Target::HVX)) {
                     f.update(i).hexagon(y).vectorize(x, 32);
                 }
-            } else {
-                f.update(i);
             }
         }
 
@@ -103,14 +102,13 @@ int main(int argc, char **argv) {
 
         // Schedule the even update steps on the gpu
         for (int i = 0; i < 10; i++) {
+            f.update(i);
             if (i & 1) {
                 if (target.has_gpu_feature()) {
                     f.update(i).gpu_tile(x, y, xo, yo, xi, yi, 16, 16);
                 } else if (target.has_feature(Target::HVX)) {
                     f.update(i).hexagon(y).vectorize(x, 32);
                 }
-            } else {
-                f.update(i);
             }
         }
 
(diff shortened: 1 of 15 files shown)


steven-johnson (approved):

steven-johnson (changes requested):
Lots of broken stuff, removing approval

abadams:
Looks like an llvm master breakage, and an uncovered bug in atomics. I'll investigate.

steven-johnson:
> Looks like an llvm master breakage, and an uncovered bug in atomics. I'll investigate.

Could bit be related to https://reviews.llvm.org/D115955 ? (It was reverted last night, I'm working on getting a repro case for them)

steven-johnson (approved):

alexreinking:
> The warning can be suppressed by inserting a call to func.update(idx).

Seeing `func.update(idx);` as a single line in code _looks_ like a discarded expression that shouldn't have a side-effect. I think it would be nicer to read something more like `func.update(idx).schedule_default();` or some other verb (:bike: :derelict_house:) that clearly indicates what is happening.

test/correctness/tuple_reduction.cpp
```diff
@@ -103,14 +102,13 @@ int main(int argc, char **argv) {
 
         // Schedule the even update steps on the gpu
         for (int i = 0; i < 10; i++) {
+            f.update(i);
```
alexreinking, line 105: Here's an example of a usage that trips me up... wouldn't it be clearer here to see `f.update(i).schedule_default();`?

steven-johnson:
So where do we stand with this? Do we want to land it as-is and hope to do a proper linter tool in the future? Or what?

alexreinking:
Up to Andrew, I guess. My suggestion was just a suggestion.

test/correctness/tuple_reduction.cpp
abadams: schedule_default sounds like it reverts any existing schedule. I could try to implement that, but it would be different.

alexreinking: Like I said, I don't care about the name and I'm not interested in bikeshedding it... pick whatever sounds right to you.

steven-johnson: suggestion:
- call it `unscheduled()`
- if the update stage already has any schedule, fail with an error
- if the update stage has no schedule, suppress the warning

abadams: I was just about to type the same idea, so I guess it's a winner.

alexreinking: Another benefit... one can `git grep '\.unscheduled()`.

alexreinking (approved):

alexreinking:
This looks great! Thanks for indulging me, @abadams :slightly_smiling_face:

steven-johnson:
The Windows failure appears to be a flake, although possibly a real bug (opened issue as https://github.com/halide/Halide/issues/6621). OK to land.

PR #6605, opened 2022-02-04
WIP for Vulkan runtime -- memory allocation & cleanups

derek-gerstmann:
This is a WIP pull request onto my own branch to allow feedback for the current design and direction of the Vulkan backend, and associated utility classes that have been added.

Primarily, I'd like some eyes on the current direction of the Vulkan backend, the src/runtime/internal classes, the clang-built unit tests for tests/runtime.

The goal was to provide minimal reusable classes that can easily be integrated into the runtime, with zero dependencies, no STL (including std::move, std::forward, etc), custom memory allocators, and interfaces that fit within current conventions.

Please let me know if anyone has any general commments or feedback thus far, as I work through the remaining issues for the runtime and code generation.

Thanks!


dependencies/spirv/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+include(FetchContent)
```
LebedevRI, line 1: Would it please be possible to make use of the packaged version of the headers instead of always fetching them?
I would imagine that would require having a `cmake/modules/FindSPIRVHeaders.cmake`,
which would look for them in a configurable set of directories.
https://packages.debian.org/sid/all/spirv-headers/filelist

derek-gerstmann: @LebedevRI Hi! Yes, that does seem to be a better solution for integrating into existing build environments.  I'll add that to my list!  Thanks for the feedback!

src/runtime/internal/block_storage.h
```diff
@@ -0,0 +1,366 @@
...
+
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+
+// Dynamically resizable array for block storage (eg plain old data)
+// -- No usage of constructors/destructors for value type (T)
+// -- Implementation uses memcpy/memmove for copying
+// -- Customizable allocator ... default uses HalideSystemAllocator
+template<typename T>
```
steven-johnson, line 14: We'll need to carefully examine how the use of shared template classes like this affects codegen -- in particular, we should verify that we don't end up with redundant copies of the template methods instantiated into every runtime module (and not deduped by the linker). Ordinarily I wouldn't worry, but our runtime build/link model is unusual enough that we should probably verify that it's really happening.

derek-gerstmann: Understood. If template explosion is a big concern, I could just remove the type safety and make these classes just allocate raw bytes, and leave it up to the client code to do the typecasting.

steven-johnson:
Any updates on this?

derek-gerstmann:
@steven-johnson Yes!  Previous comments about templates and SPIR-V headers have been addressed, and I recently merged with main. Added base functionality. Currently working on correctness tests ... 280 pass, 68 fail.  A few missing pieces in CodeGen and buffer dataflow, but its coming along!

steven-johnson:
> [@derek-gerstmann](https://github.com/derek-gerstmann) pushed 110 commits.

whoa

steven-johnson:
Any update?

derek-gerstmann:
Still making progress! Currently working on CodeGen.

derek-gerstmann:
Closing this WIP PR and refactoring changes into standalone commits to make things easier for uptake.

PR #6608, opened 2022-02-06
Fixes to support LLVM with opaque pointers.

zvookin:
First set of changes to remove code that recovers the target type from an LLVM ```PointerType``` as this functionality is going away with opaque pointers. I think this is the bulk of the work for this, but there is probably some needed to our .ll and it needs to be tested with LLVM's -opaque-pointers flag everywhere. (Which is not working for me yet. Not sure how much of the issue is on our side, but I get the impression this is happening right now and it isn't done yet. Ultimately the flag will be unnecessary as LLVM will just have opaque pointers.)

Couple TODOs to remove debug code before commit. It is still there to get extra info during CI runs if something fails. (The debug validation has to use the API we are removing, which is why it needs to be removed.) Maybe one other TODO raising an issue for code review.

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -2937,8 +2937,7 @@ void CodeGen_LLVM::visit(const Call *op) {
                                                   ConstantInt::get(i32_t, (int)*index)});
             // TODO(zvookin|abadams): Two ways to go here: traverse the LLVM type
             // or just assume op-type is correct. Going ahead and traversing the type.
-            llvm::Type *result_type = struct_type ? struct_type->getElementType(*index)
-                                                  : array_type->getArrayElementType();
+            llvm::Type *result_type = struct_type ? struct_type->getElementType(*index) : array_type->getArrayElementType();
             value = builder->CreateLoad(result_type, gep);
         } else {
             // The struct is actually just a scalar
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -145,17 +145,42 @@ using std::vector;
 
 namespace {
 
-llvm::Value *CreateConstGEP1_32(IRBuilderBase *builder, Value *ptr, unsigned index) {
+void print_type_pair(const char *label, llvm::Type *a, llvm::Type *b) {
+    debug(0) << label;
+    a->print(llvm::dbgs(), true);
+    debug(0) << " ptr type: ";
+    b->print(dbgs(), true);
+    debug(0) << "\n";
+}
+
+llvm::Value *CreateConstGEP1_32(IRBuilderBase *builder, llvm::Type *gep_type,
+                                Value *ptr, unsigned index) {
 #if LLVM_VERSION >= 130
-    return builder->CreateConstGEP1_32(ptr->getType()->getScalarType()->getPointerElementType(), ptr, index);
+    // TODO(zvookin): To be removed before merging. Code is here to get debugging info
+    // during CI test runs if the assert below fails. Assert also must be removed
+    // as it relies on non-opaque pointers.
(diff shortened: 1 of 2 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -270,6 +295,7 @@ void CodeGen_LLVM::initialize_llvm() {
             for (const std::string &s : arg_vec) {
                 c_arg_vec.push_back(s.c_str());
             }
+            //            c_arg_vec.push_back("-opaque-pointers");
```
steven-johnson, line 298: Left in accidentally?

src/CodeGen_LLVM.cpp
```diff
@@ -1830,17 +1872,21 @@ Value *CodeGen_LLVM::codegen_buffer_pointer(const string &buffer, Halide::Type t
...
+    llvm::Type *pointer_load_type = load_type->getPointerTo(address_space);
 
-    llvm::Type *load_type = llvm_type_of(type)->getPointerTo(address_space);
+#if LLVM_VERSION <= 150
+    llvm::Type *base_address_type = base_address->getType();
 
     // If the type doesn't match the expected type, we need to pointer cast
-    if (load_type != base_address_type) {
-        base_address = builder->CreatePointerCast(base_address, load_type);
+    if (pointer_load_type != base_address_type) {
```
steven-johnson, line 1884: Does this apply to the branched-but-not-final LLVM14 as well? (As of right now we're not testing that but I will set up the buildbots to add that shortly)

steven-johnson (commented):
LGTM with nit about version 14 vs 15

src/CodeGen_LLVM.cpp
zvookin: Should not be committed, but good to have it appear here briefly as documentation on how to test with LLVM in opaque pointer mode during its transition. Will remove or comment.

steven-johnson: +1 to leave-in-with-comment

src/CodeGen_LLVM.cpp
```diff
@@ -1830,17 +1874,14 @@ Value *CodeGen_LLVM::codegen_buffer_pointer(const string &buffer, Halide::Type t
...
+    unsigned address_space = base_address->getType()->getPointerAddressSpace();
+    llvm::Type *pointer_load_type = load_type->getPointerTo(address_space);
 
-    llvm::Type *load_type = llvm_type_of(type)->getPointerTo(address_space);
-
-    // If the type doesn't match the expected type, we need to pointer cast
-    if (load_type != base_address_type) {
-        base_address = builder->CreatePointerCast(base_address, load_type);
-    }
+    // TODO: This can likely be removed once aopaque pointers are default
```
abadams, line 1882: typo: aopaque -> opaque

src/CodeGen_LLVM.cpp
```diff
@@ -2873,17 +2916,28 @@ void CodeGen_LLVM::visit(const Call *op) {
...
                                                  {ConstantInt::get(i32_t, 0),
                                                   ConstantInt::get(i32_t, (int)*index)});
-            value = builder->CreateLoad(gep->getType()->getPointerElementType(), gep);
+            // TODO(zvookin|abadams): Two ways to go here: traverse the LLVM type
+            // or just assume op-type is correct. Going ahead and traversing the type.
+            llvm::Type *result_type = struct_type ? struct_type->getElementType(*index) : array_type->getArrayElementType();
+            value = builder->CreateLoad(result_type, gep);
         } else {
             // The struct is actually just a scalar
+            // TODO(zvookin|abadams): Is adding this assert sensible?
```
abadams, line 2939: Yes, I think this is sensible.

abadams (approved):
lgtm. Don't forget to remove the print_type_pair stuff before merging.

PR #6610, opened 2022-02-08
Fix bug in mul_shift_right matching

abadams:
widening_mul handles mixed-signedness args, but mul_shift_right doesn't.
We shouldn't map a mixed-signedness widening_mul to a mul_shift_right.

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -380,6 +380,8 @@ protected:
             Type signed_type_wide = op_type_wide.with_code(halide_type_int);
             Type unsigned_type = op->type.with_code(halide_type_uint);
 
+            // Give concise names to various predicates we want to use in
+            // rewrite rules below.
             int bits = op->type.bits();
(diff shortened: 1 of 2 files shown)


steven-johnson (approved):

src/FindIntrinsics.cpp
```diff
@@ -384,6 +384,7 @@ class FindIntrinsics : public IRMutator {
             auto is_x_same_int = op->type.is_int() && is_int(x, bits);
             auto is_x_same_uint = op->type.is_uint() && is_uint(x, bits);
             auto is_x_same_int_or_uint = is_x_same_int || is_x_same_uint;
+            auto x_y_same_sign = (is_int(x) && is_int(y)) || (is_uint(x) && is_uint(y));
```
shoaibkamil, line 387: I think a comment here would help-- it's not necessarily clear at first glance what `is_int()` means compared to `Type::is_int()`

shoaibkamil (approved):

PR #6612, opened 2022-02-11
Update WasmExecutor for WABT API changes

steven-johnson:
The API for WABT has changed in top-of-tree. This fixes the incompatibilities. It also (temporarily) changes to pulling WABT from top-of-tree (rather than a known release) as the most recent labeled release doesn't have these changes. (We'll change back to a fixed WABT release when the next one is released.)

### DIFF
--- a/dependencies/wasm/CMakeLists.txt
+++ b/dependencies/wasm/CMakeLists.txt
@@ -16,7 +16,7 @@ if ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
 endif ()
 
 if (WITH_WABT)
-    set(WABT_VER 1.0.25)
+    set(WABT_VER main)
 
     message(STATUS "Fetching WABT ${WABT_VER}...")
     FetchContent_Declare(wabt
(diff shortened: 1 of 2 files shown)


abadams (approved):

dependencies/wasm/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
 endif ()
 
 if (WITH_WABT)
-    set(WABT_VER 1.0.25)
+    set(WABT_VER main)
```
alexreinking, line 19: I am worried about this making our builds unpredictable. I think better would be to hard-code an override-able git commit hash, like so:
```suggestion
    set(WABT_VER "09c40635207d42dd30ffaca22477fd3491dd9e7d"
        CACHE STRING "WABT commit to download (tag, branch name, or hash)")
```
and then in CI we can pull nightlies by setting the `WABT_VER` cache variable to `main`.

alexreinking (changes requested):

dependencies/wasm/CMakeLists.txt
alexreinking: (`09c40635207d42dd30ffaca22477fd3491dd9e7d` is the current top-of-tree hash)

steven-johnson: Fair, keep in mind this should be very short term

alexreinking (approved):

PR #6614, opened 2022-02-14
Unbreak WABT again by using main instead of a commit

steven-johnson:
Commit 09c40635207d42dd30ffaca22477fd3491dd9e7d is no longer being recognized for WABT; rather than whack-a-mole, let's just use `main` temporarily (I've requested the WABT team make a new tag ASAP so we can standardize on the new API)


alexreinking:
Ugh, who force-pushes to main/master?!

alexreinking (approved):
Still wish this were a cache variable, but I guess this is just a sharp edge no matter what until WABT tags a release

PR #6620, opened 2022-02-17
Convert all remaining Generators to prefer statically-dimensioned Inputs and Output where possible

steven-johnson:


apps/HelloPyTorch/src/add_generator.cpp
```diff
@@ -15,9 +15,9 @@ Func add_(const Input &input_a, const Input &input_b) {
 
 class AddGenerator : public Generator<AddGenerator> {
 public:
-    Input<Buffer<>> input_a{"input_a", 4};
-    Input<Buffer<>> input_b{"input_b", 4};
-    Output<Buffer<>> output{"output", 4};
+    Input<Buffer<void, 4>> input_a{"input_a"};
```
abadams, line 18: This is a little unfortunate. Is it possible to make Buffer<4> work? I don't think that's doable in c++, right?

src/autoschedulers/adams2019/cost_model_generator.cpp
```diff
@@ -15,38 +15,38 @@ using Halide::Derivative;
 // A model weight is either just an input, or an input and an output
 // (the updated weights and the ADAM state) depending on whether we're
 // doing inference or training.
-template<bool training>
```
abadams, line 18: In this file, runtime dimensionality seems much cleaner (e.g. there's value in having all the types be the same down below)

apps/HelloPyTorch/src/add_generator.cpp
steven-johnson: No, definitely not. IMHO this is a minor price to pay.

src/autoschedulers/adams2019/cost_model_generator.cpp
steven-johnson: We could selectively revert this one. (I did the conversion by deleting all the old ctors temporarily and then fixing everything.)

steven-johnson:
Tuesday Morning Review Ping

src/autoschedulers/adams2019/cost_model_generator.cpp
abadams: +1 to selectively reverting this one

steven-johnson: Before I revert this, can I ask what you mean by "having all the types be the same down below"? Looking at this again, I don't know which types you mean, and so I'm confused about the specific nature of your concern. (I'm trying to understand why you feel the old approach is cleaner, so I can factor different views into future work.)

src/autoschedulers/adams2019/cost_model_generator.cpp
```diff
@@ -448,13 +450,12 @@ class CostModel : public Generator<CostModel<training>> {
             // to the model weights.
             Derivative d_loss_d = propagate_adjoints(loss_output);
 
-            Weight *weights[] = {&head1_filter, &head1_bias,
-                                 &head2_filter, &head2_bias,
-                                 &filter1, &bias1};
-
-            for (Weight *w : weights) {
```
abadams, line 455: It's losing the ability to iterate over the weights here that I was referring to.

steven-johnson: That could be fixed pretty trivially by making `backprop()` a virtual method and adding a `ModelWeightBase` class.

steven-johnson: (Just pushed a revision to do this.)

PR #6623, opened 2022-02-19
Wild match object should not be foldable

rootjalex:
`Wild` objects should not be foldable, `WildConst` should be used when matching against constants. There are no use cases of a `Wild` object being folded, so this should be removed unless there is a relevant use-case.


abadams (approved):

steven-johnson:
Looks like there are some failures

abadams:
This PR just deletes some dead code, so I think those failures are a fresh llvm injection.

rootjalex:
Good to merge then?

PR #6625, opened 2022-02-21
Remove incorrect docs on widening_add

rootjalex:
The result of `widening_add` is not always signed.


abadams (approved):

rootjalex:
PR only changes comment, failure is unrelated

PR #6626, opened 2022-02-21
Make IRComparer consider nans to be less than non-nans.

### LINKED ISSUE #6624: select statement that contains NaN value always returns NaN
Hi.

I am trying to make a function that reads an input buffer (passed in as a Halide ImageParam, where the data is interleaved), and that returns NaN or 1, based on whether or not xyz[col, row, 0] is NaN. (Full reproducer below.)

```
xyz = ImageParam(
    type=Float(32),
    dimensions=3,
    name="xyz",
)
xyz.dim(0).set_stride(3)
xyz.dim(2).set_stride(1)
```
```
nan_or_one = Func("nan_or_one")
nan_or_one[col, row] = select(
    is_nan(xyz[col, row, 0]),
    f32(np.nan),  # This select always returns NaN. If this line is changed to f32(2.0) then it works as one expected, ie. 2.0 is returned when xyz[col, row, 0] is NaN, else 1.0 is returned.
    f32(1.0),
)
```
This select statement always returns NaN. If I change NaN to fex 2.0, then I get the expected output, ie. 2.0 when xyz[col, row, 0] is NaN, else 1.0. But for some reason the select statement does not handle NaN the same way as another float value.

Reproducer: https://gist.github.com/apartridge/2e652b7aceb80dfee67a4ca903e6bf90
Output with HL_DEBUG_CODEGEN=2: https://gist.github.com/apartridge/a3838c6a5ee43e7139bc246d7a97cc4b
Output with HL_DEBUG_CODEGEN=10000: https://gist.github.com/apartridge/902f0d1d3e75f230977cc3817719857b

I am using Halide 13 on Ubuntu 20 using the Python bindings. (I also see the problem on Halide 10.) I am using StrictFloat. 

Note that this is simplified a bit compared to my original use case. What I want is to convert an input xyz buffer to an xyzw buffer, where I set w=1 if xyz is non-NaN, else set w=NaN.

This issue looks quite similar to https://github.com/halide/Halide/issues/4213.

Also discussed on Gitter:
https://matrix.to/#/!cSNWpLyImSoNFJJFZE:gitter.im/$mZznBCmUdGov8_3Vsy05N2Fnc3-7AtZaeQUp_2mxJwg?via=matrix.org&via=gitter.im

Thanks for any assistance :)

alexreinking:
Our IRComparer's `compare_scalar` believed that all floating point values are totally ordered, which is not true. NaNs always compare false in ordering comparisons.

Fixes #6624

### DIFF
--- a/src/IREquality.cpp
+++ b/src/IREquality.cpp
@@ -106,6 +106,22 @@ IRComparer::CmpResult IRComparer::compare_scalar(T a, T b) {
         return result;
     }
 
+    if constexpr (std::is_floating_point_v<T>) {
+        // NaNs are equal to each other and less than non-nans
+        if (std::isnan(a) && std::isnan(b)) {
+            result = Equal;
+            return result;
+        }
+        if (std::isnan(a)) {
+            result = LessThan;
+            return result;
+        }
+        if (std::isnan(b)) {
+            result = GreaterThan;
+            return result;
+        }
+    }
+
     if (a < b) {
         result = LessThan;
     } else if (a > b) {
@@ -125,6 +141,7 @@ IRComparer::CmpResult IRComparer::compare_expr(const Expr &a, const Expr &b) {
         return result;
     }
 
+    // Undefined values are equal to each other and less than defined values
     if (!a.defined() && !b.defined()) {
         result = Equal;
         return result;
--- a/test/correctness/cse_nan.cpp
+++ b/test/correctness/cse_nan.cpp
@@ -33,6 +33,6 @@ int main() {
     } else {
         fprintf(stderr, "ERROR: T = %f ; TR = %f ; F = %f ; FR = %f\n",
                 true_buf(0, 0, 0), true_result(0, 0), false_buf(0, 0, 0), false_result(0, 0));
-        return 1;
+        return -1;
     }
 }
--- a/test/correctness/cse_nan.cpp
+++ b/test/correctness/cse_nan.cpp
@@ -28,9 +28,11 @@ int main() {
     nan_or_one.realize({false_result}, t);
 
     if (std::isnan(true_result(0, 0)) && false_result(0, 0) == 1.0f) {
-        puts("Success!");
+        printf("Success!\n");
+        return 0;
     } else {
         fprintf(stderr, "ERROR: T = %f ; TR = %f ; F = %f ; FR = %f\n",
                 true_buf(0, 0, 0), true_result(0, 0), false_buf(0, 0, 0), false_result(0, 0));
+        return 1;
     }
 }
--- a/test/correctness/cse_nan.cpp
+++ b/test/correctness/cse_nan.cpp
@@ -6,8 +6,7 @@ using namespace Halide;
 
 int main() {
     ImageParam xyz{Float(32), 3, "xyz"};
-    Target t = get_host_target();
-    t.set_feature(Target::StrictFloat);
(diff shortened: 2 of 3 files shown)


LebedevRI:
Does this need any "is in strict fp mode" check?

alexreinking:
> Does this need any "is in strict fp mode" check?

I don't believe so, no. I don't think we ever want to consider `nan` equal to anything but `nan` for the purposes of comparing IR nodes.

alexreinking:
Sorry, I just realized I forgot to actually _use_ the updated target in the test... going to fix that now.

test/correctness/cse_nan.cpp
```diff
@@ -0,0 +1,36 @@
...
+    Buffer<float> false_result{1, 1};
+
+    xyz.set(true_buf);
+    nan_or_one.realize({true_result}, t);
+
+    xyz.set(false_buf);
+    nan_or_one.realize({false_result}, t);
+
+    if (std::isnan(true_result(0, 0)) && false_result(0, 0) == 1.0f) {
+        puts("Success!");
```
abadams, line 31: printf would be more similar to the other tests

test/correctness/cse_nan.cpp
```diff
@@ -0,0 +1,36 @@
...
+    } else {
+        fprintf(stderr, "ERROR: T = %f ; TR = %f ; F = %f ; FR = %f\n",
+                true_buf(0, 0, 0), true_result(0, 0), false_buf(0, 0, 0), false_result(0, 0));
+    }
```
abadams, line 35: (and a return -1 in this case)

test/correctness/cse_nan.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
```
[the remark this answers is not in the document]
alexreinking: Done

alexreinking:
Failing test looks like a performance flake to me... what are your thoughts, @abadams?

src/IREquality.cpp
```diff
@@ -106,6 +106,22 @@ IRComparer::CmpResult IRComparer::compare_scalar(T a, T b) {
         return result;
     }
 
+    if constexpr (std::is_floating_point_v<T>) {
+        // NaNs are equal to each other and less than non-nans
```
steven-johnson, line 110: This makes me very nervous. In IEEE754, NaN != everything, including NaN.

alexreinking: This is just for matching the identity of IR nodes... in particular the previous revision considered NaN **is** equal to everything

alexreinking: We could discuss --- in a future PR, perhaps --- extending the set of results for `IRComparer` to include `NotEqual`, which would allow it to express partial orders (like the one that exists in IEEE 754). I have no idea how far-reaching that change would be.

abadams: For the purposes of this function we could also just bit-cast the float to a uint64 and compare the bits. The important thing here is that things that are not equal compare as not equal.

alexreinking:
Leaving this for someone else to merge or patch further.

apartridge:
Thanks for fixing this issue! Would it be possible to make a 13.0.5 release with this fix?

alexreinking:
@apartridge - that would be possible, but my bandwidth is somewhat limited until mid-week next week.

PR #6622, opened 2022-02-18
Disallow `Type::narrow()` and `Type::widen()` from producing bitwidths between 1 and 8 bits

rootjalex:
Allowing these methods to create bitwidths of `2` and `4` can cause issues with `find_intrinsics`. Specifically, this fixes #6601 .

Discussed with @abadams .

PR #6629, opened 2022-02-23
Avoid double narrowing in widening_add/widening_sub if type is 8-bit

vksnk:
#6622, specifically making narrowing of uint1 an error broke some of our internal tests. The simplest reproducer would be an expression like `u8(x > 0) + u8(y > 0)`, which would get replaced by `widening_add(x > 0 + y > 0)` where type of `widened_add` is `uint8`, which later at https://github.com/halide/Halide/blob/main/src/FindIntrinsics.cpp#L620 will get narrowed twice and cause u8->u1->error. 

Here I just looked for all places where narrow().narrow() happens and added a check to make sure that type has at least 16-bit, which seems to fix all failing tests I was seeing.

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -614,7 +614,7 @@ protected:
             if (narrow_a.defined() && narrow_b.defined()) {
                 return mutate(Cast::make(op->type, widening_mul(narrow_a, narrow_b)));
             }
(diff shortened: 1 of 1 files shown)


rootjalex (approved):
Yikes, sorry about that. Thanks for fixing this!

vksnk:
The test failure seems to be unrelated, so merging in.

PR #6633, opened 2022-03-03
Python Bindings didn't allow for zero-D Funcs, ImageParams, Buffers

steven-johnson:
There were no overloads or tests for accessing the element of any of these in the zero-D case, and the obvious syntax (`[]`, to mirror C++ `()` in these cases) isn't legal in Python. To support this uncommon-but-necessary case, I'm proposing that we use the syntax ~~`[None]`~~, `[()]`, which isn't pretty, but is less bad than other options I've considered so far. (Suggestions welcome.)

### DIFF
--- a/python_bindings/correctness/basics.py
+++ b/python_bindings/correctness/basics.py
@@ -284,6 +284,22 @@ def test_vector_tile():
     p = hl.Pipeline([f, g])
     p.compile_jit()
 
+def test_scalar_funcs():
+    input = hl.ImageParam(hl.UInt(16), 0, 'input')
+
+    f = hl.Func('f')
+    g = hl.Func('g')
+
+    input[None]
+
+    (input[None]+input[None]) / 2
+    f[None]
+    g[None]
+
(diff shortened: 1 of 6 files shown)


abadams:
An alternative to None might be the empty tuple: [()]

alexreinking:
`None` is fine. It's the most obvious thing besides the obvious, but illegal, syntax.

steven-johnson:
FWIW, `[()]` works even without this PR (we'd want to add the test cases, of course). I can live with either syntax.

steven-johnson:
Final votes on syntax?

alexreinking:
> Final votes on syntax?

On second thought, we _shouldn't_ use `None` for this... Numpy uses it in its indexing logic to introduce a _new_ axis of extent 1. That's not what's happening here, so to allow `None` in this case might well cause confusion. Let's go with the empty tuple and just make sure it's tested.

steven-johnson:
PTAL

alexreinking (approved):

PR #6637, opened 2022-03-04
Define an AbstractGenerator interface

steven-johnson:
This PR refactors the Generator infrastructure to introduce the AbstractGenerator interface; this provides a relatively clean API that can be used to introduce alternate "Generator" syntaxes/implementations that still participate in the build infrastructure in the same way as existing Generators. 

The impetus for this was originally some experimental work to allow a more functional-style Generator binding in C++; that work is on hold indefinitely, but this work has proved useful in the ongoing work to add Generators to the Python bindings. I've maintained this as a private branch for a year or so (!) but at this point I think landing it will make further development simpler. PTAL.

### DIFF
--- a/src/AbstractGenerator.h
+++ b/src/AbstractGenerator.h
@@ -109,7 +109,7 @@ public:
      */
     virtual std::vector<ArgInfo> get_output_arginfos() = 0;
 
-    /** Return the names for all known Constants (aka GeneratorParams) in this Generator.
+    /** Return the names for all known GeneratorParams in this Generator.
      * (Synthetic params are excluded and will never be returned here.)
      * Always legal to call on any AbstractGenerator instance, regardless of what other methods
      * have been called. Note that while the results are returned in a vector, the order of
@@ -180,9 +180,6 @@ public:
      * This call is infrequently used, and is only useful in JIT or Stub situations;
      * this allows you to bind the inputs to be Funcs, Buffers, etc from another Halide code
      * fragment, allowing inlining of multiple fragments together.
-     *
-     * Note that there is no way to rebind individual inputs. (This could be added but
-     * is unlikely to be useful, as all existing usage is all-or-none.)
      *
      * CALL-AFTER: n/a
      * CALL-BEFORE: build_pipeline
--- a/src/AbstractGenerator.h
+++ b/src/AbstractGenerator.h
@@ -1,0 +1,217 @@
+#ifndef HALIDE_ABSTRACT_GENERATOR_H_
+#define HALIDE_ABSTRACT_GENERATOR_H_
+
+#include <functional>
+#include <map>
+#include <string>
+#include <vector>
+
+#include "Expr.h"
+#include "Func.h"
+#include "Parameter.h"
+#include "Pipeline.h"
+#include "Schedule.h"
+#include "Target.h"
+#include "Type.h"
+
+namespace Halide {
+
+class GeneratorContext;
+
+namespace Internal {
+
+enum class IOKind { Scalar,
+                    Function,
+                    Buffer };
+
+using ExternsMap = std::map<std::string, ExternalCode>;
+
+/**
+ * AbstractGenerator is an ABC that defines the API a Generator must provide
+ * to work with the existing Generator infrastructure (GenGen, RunGen,
+ * Generator Stubs). The existing Generator<>-based instances all implement
+ * this API, but any other code that implements this (and uses RegisterGenerator
+ * to register itself) should be indistinguishable from a user perspective.
+ *
+ * An AbstractGenerator is meant to be "single-use"; typically lifetimes will be
+ * something like:
+ * - create an instance (with a specific Target)
+ * - optionally set GeneratorParam values
+ * - optionally re-bind inputs (if using in JIT or Stub modes)
+ * - call build_pipeline()
+ * - optionally call get_funcs_for_output() to get the output(s) (if using in JIT or Stub modes)
+ * - discard the instance
+ *
+ * AbstractGenerators should be fairly cheap to instantiate! Don't try to re-use
+ * one by re-setting inputs and calling build_pipeline() multiple times.
+ *
+ * Note that an AbstractGenerator instance is (generally) stateful in terms of the order
+ * that methods should be called; calling the methods out of order may cause
+ * assert-fails or other undesirable behavior. Read the method notes carefully!
+ */
+class AbstractGenerator {
+public:
+    virtual ~AbstractGenerator() = default;
+
+    /** ArgInfo is a struct to contain name-and-type information for the inputs and outputs to
+     * the Pipeline that build_pipeline() will return.
+     *
+     * Note that this looks quite similar to Halide::Argument, but unfortunately
+     * that is not a good fit here, as it cannot represent Func inputs (only
+     * Buffer and Scalar). Rather than extend Argument to handle that (and risk
+     * confusing the public API), we'll create this (which is only for internal
+     * use).
+     *
+     * TODO: name is suboptimal. Is there a better one?
+     */
+    struct ArgInfo {
+        std::string name;
+        IOKind kind = IOKind::Scalar;
+        // Note that this can have multiple entries for Tuple-valued Inputs or Outputs
+        std::vector<Type> types;
+        int dimensions = 0;
+    };
+
+    /** Return the name of this Generator. (This should always be the name
+     * used to register it.) */
+    virtual std::string get_name() = 0;
+
+    /** Return the Target, autoscheduler flag, and MachineParams that this Generator
+     * was created with. Always legal to call on any AbstractGenerator instance,
+     * regardless of what other methods have been called. (All AbstractGenerator instances
+     * are expected to be created with immutable values for these, which can't be
+     * changed for a given instance.)
+     *
+     * CALL-AFTER: any
+     * CALL-BEFORE: any
+     */
+    virtual GeneratorContext context() const = 0;
+
+    /** Return a list of the names for inputs, in the correct order.
+     * All input and output names will be unique within a given Generator instance.
+     * If this is called after add_input(), the added inputs will be returned.
+     * Always legal to call on any AbstractGenerator instance, regardless of what other methods
+     * have been called.
+     *
+     * CALL-AFTER: any
+     * CALL-BEFORE: any
+     */
+    virtual std::vector<ArgInfo> get_input_arginfos() = 0;
+
+    /** Return a list of the names for outputs, in the correct order.
+     * All input and output names will be unique within a given Generator instance.
+     * If this is called after add_output(), the added outputs will be returned.
+     * Always legal to call on any AbstractGenerator instance, regardless of what other methods
+     * have been called.
+     *
+     * CALL-AFTER: any
+     * CALL-BEFORE: any
+     */
+    virtual std::vector<ArgInfo> get_output_arginfos() = 0;
+
+    /** Return the names for all known Constants (aka GeneratorParams) in this Generator.
+     * (Synthetic params are excluded and will never be returned here.)
+     * Always legal to call on any AbstractGenerator instance, regardless of what other methods
+     * have been called. Note that while the results are returned in a vector, the order of
+     * names in the result aren't important (unlike for inputs and outputs, where order matters).
+     *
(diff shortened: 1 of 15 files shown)


src/AbstractGenerator.h
```diff
@@ -0,0 +1,214 @@
...
+ * this API, but any other code that implements this (and uses RegisterGenerator
+ * to register itself) should be indistinguishable from a user perspective.
+ *
+ * An AbstractGenerator is meant to be "single-use"; typically lifetimes will be
+ * something like:
+ * - create an instance (with a specific Target)
+ * - optionally set GeneratorParam values
+ * - optionally re-bind inputs (if using in JIT or Stub modes)
+ * - call build_pipeline()
+ * - optionally call get_funcs_for_output() to get the output(s) (if using in JIT or Stub modes)
```
zvookin, line 42: How are the outputs specified if in AOT mode?

src/AbstractGenerator.h
```diff
@@ -0,0 +1,194 @@
...
+    // @{
+    virtual void bind_input(const std::string &name, const std::vector<Parameter> &v) = 0;
+    virtual void bind_input(const std::string &name, const std::vector<Func> &v) = 0;
+    virtual void bind_input(const std::string &name, const std::vector<Expr> &v) = 0;
+    // @}
+
+    /** Emit a Generator Stub (.stub.h) file to the given path. Not all Generators support this.
+     *
+     * If you call this method, you should not call any other AbstractGenerator methods
+     * on this instance, before or after this call.
```
antonysigma, line 176: Thank you @steven-johnson for providing the `AbstractGenerator` base class. I see it as an opportunity to port any custom AOT implementations to the Generator interface. Here's one example to emit FPGA accelerator code for HW synthesis, which is incompatible to the `Generator<>` interface:

https://github.com/jingpu/Halide-HLS/blob/HLS/apps/hls_examples/conv_hls/pipeline.cpp#L131-L135

Quick question: if I were to implement `emit_cpp_stub` function
```c++
class MyGenerator final : public AbstractGenerator {
...
bool MyGenerator::emit_cpp_stub(const string& path) final {

        output.compile_to_lowered_stmt(path + "pipeline_host.ir.html", args, HTML);
        output.compile_to_header(path + "pipeline_host.h", args, "pipeline_native");
        output.compile_to_object(path + "pipeline_host.o", args, "pipeline_native");

Target hls_target = get_target_from_environment();
        hls_target.set_feature(Target::CPlusPlusMangling);
        output.compile_to_lowered_stmt(path + "pipeline_target.ir.html", args, HTML, hls_target);
        output.compile_to_hls(path + "pipeline_target.cpp", args, "pipeline_hls", hls_target);
        output.compile_to_header(path + "pipeline_target.h", args, "pipeline_hls", hls_target);
}
};
```

Can I call define the following Makefile target to emit the AOT binaries? How about the cpp stub?
```Make
pipeline_target.cpp: pipeine.generator
    ./$< -g pipeline_name target=zynq -e 'html,h,cpp' -o . && echo "$@ is generated"

pipeline_host.o: pipeline.generator
    ./$< -g pipeline_name target=host -e 'html,h,o' -o . && echo "$@ is generated"
```

steven-johnson: > Thank you @steven-johnson

Well, it hasn't landed yet, but I hope it will soon :-)

> which is incompatible to the `Generator<>` interface

Interesting, can you share a bit more about how and why your work is incompatible? (It's always good to learn about how and why and existing design isn't adequate.)

> if I were to implement emit_cpp_stub function

The implementation you have there isn't actually emitting a cpp_stub -- a cpp_stub is a very weird chunk of code that is intended to make a Generator more usable in JIT mode, or when composing Generators together. (It's very rarely used and frankly should be removed if downstream users ever stop using it.) What are you trying to accomplish here?

BTW #2: I *highly* recommend you *don't* examine any environment variables in your Generator; that includes `get_target_from_environment()`; you can see the Target that was specified for your generator simply by calling its `get_target()` method.

antonysigma: > Interesting, can you share a bit more about how and why your work is incompatible? (It's always good to learn about how and why and existing design isn't adequate.)

Sorry, I take it back. I realized the original author of the forked project implemented `Pipeline::compile_to_hls`, but never got that far to integrate it into `Generator<>`. So, if I were to schedule an image processing pipeline with `Generator<>::schedule()`, I can schedule `target=host` but not `target=zynq`.

https://github.com/jingpu/Halide-HLS/blob/905d2f2ad560246673ba3a84b8a6d8be308e481f/src/Pipeline.h#L148-L163

> not emitting a cpp_stub.

Got it. Thanks for the reminder.

> don't examine any environment variables.

My bad. when I drafted the pseudo code, I copied it out of the JIT code from the original author. As I mentioned, it would be ideal if I can organize the JIT code (link below) into either `Generator<>` or `AbstractGenerator`. I simply don't know how it is done, starting from the developer provided `Pipeline::compile_to_*` API in `src/pipeline.h`.

https://github.com/jingpu/Halide-HLS/blob/HLS/apps/hls_examples/conv_hls/pipeline.cpp
https://github.com/jingpu/Halide-HLS/blob/HLS/apps/hls_examples/unsharp_hls/pipeline.cpp#L113-L153

I can create a new issue for a more in depth discussion. Again, I appreciate your work on `AbstractGenerator` class.

steven-johnson: > I can create a new issue for a more in depth discussion.

Yes, please do.

steven-johnson:
Gentle Review Ping

src/AbstractGenerator.h
```diff
@@ -0,0 +1,190 @@
...
+    virtual ~AbstractGenerator() = default;
+
+    /** ArgInfo is a struct to contain name-and-type information for the inputs and outputs to
+     * the Pipeline that build_pipeline() will return.
+     *
+     * Note that this looks rather similar to Halide::Argument, but unfortunately
+     * that is not a good fit here, as it cannot represent Func inputs (only
+     * Buffer and Scalar), nor can it really handle Outputs.
+     *
+     * TODO: name is suboptimal. Is there a better one?
```
abadams, line 66: As this is a nested class, I don't think agonizing over the name matters much.

src/AbstractGenerator.h
```diff
@@ -0,0 +1,190 @@
...
+        ArgInfoDirection dir = ArgInfoDirection::Input;
+        ArgInfoKind kind = ArgInfoKind::Scalar;
+        // Note that this can have multiple entries for Tuple-valued Inputs or Outputs
+        std::vector<Type> types;
+        int dimensions = 0;
+    };
+
+    /** Return the name of this Generator. (This should always be the name
+     * used to register it.) */
+    virtual std::string get_name() = 0;
```
abadams, line 79: Maybe we're not entirely consistent, but I think most of the time we omit get_ on getters.

src/AbstractGenerator.h
```diff
@@ -0,0 +1,190 @@
...
+
+    /** Return the Target, autoscheduler flag, and MachineParams that this Generator
+     * was created with. Always legal to call on any AbstractGenerator instance,
+     * regardless of what other methods have been called. (All AbstractGenerator instances
+     * are expected to be created with immutable values for these, which can't be
```
abadams, line 84: The immutable bit troubles me. There should be some mechanism to set or modify these in the Generator itself (or in the registration of the generator?) for those of us build systems incapable of supporting ad-hoc per-generator configuration. Currently this can be done in the configure() method.

src/AbstractGenerator.h
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * Names that aren't known generator names should assert-fail.
+     *
+     * Values that can't be parsed for the specific GeneratorParam (e.g. passing "foo" where
+     * an integer is expected) should assert-fail at some point (either immediately, or when
+     * build_pipeline() is called)
+     *
+     * This can be called multiple times, but only prior to build_pipeline().
+     *
+     * CALL-AFTER: nona
```
abadams, line 113: nona -> none

src/AbstractGenerator.h
```diff
@@ -0,0 +1,190 @@
...
+     * (Most inputs will have exactly one, but inputs that are declared as arrays
+     * will have multiple.)
+     *
+     * CALL-AFTER: build_pipeline
+     * CALL-BEFORE: none
+     */
+    virtual std::vector<Parameter> get_input_parameter(const std::string &name) = 0;
+
+    /** Given the name of an output, return the Func(s) for that output.
+     * (Most outputs will have exactly one, but outputs that are declared as arrays, or that return Tuples,
```
abadams, line 139: A Func that returns a Tuple is still just one Func

src/Generator.cpp
```diff
@@ -1164,14 +1381,6 @@ GeneratorParamBase::~GeneratorParamBase() {
 }
 
 void GeneratorParamBase::check_value_readable() const {
```
abadams, line 1383: Why did the assert go away?

src/Generator.cpp
```diff
@@ -1551,211 +1728,160 @@ Pipeline GeneratorBase::get_pipeline() {
...
+    get_arguments(args, ArgInfoDirection::Input, param_info().inputs());
+    get_arguments(args, ArgInfoDirection::Output, param_info().outputs());
+    return args;
+}
+
+std::vector<Parameter> GeneratorBase::get_input_parameter(const std::string &name) {
+    auto *input = find_input_by_name(name);
+
+    std::vector<Parameter> params;
+    // TODO: replicated code, yuck
```
abadams, line 1818: You could put the common bit in a lambda and call it in both if branches

src/AbstractGenerator.h
steven-johnson: Fair, will delete comment.

src/AbstractGenerator.h
steven-johnson: Hrm, you're probably right -- I think I originally used this convention to distinguish from existing methods during development. I'll (probably) fix this, pending a reality check.

src/Generator.cpp
steven-johnson: ...I don't remember. Let me figure it out.

src/Generator.cpp
```diff
@@ -1453,13 +1633,13 @@ void GeneratorBase::check_exact_phase(Phase expected_phase) const {
 void GeneratorBase::advance_phase(Phase new_phase) {
     switch (new_phase) {
     case Created:
-        internal_error << "Impossible";
+        internal_error;
```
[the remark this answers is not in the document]
steven-johnson: Eh -- I figured that "Impossible" isn't appreciably better than empty and saves a few bytes. Will restore if you prefer.

src/AbstractGenerator.h
steven-johnson: (FWIW, it's definitely not universally true, e.g. the `get_` methods on GeneratorContext.)

steven-johnson: Done, PTAL

steven-johnson:
This branch is getting harder to maintain as time goes on. Can we consider landing it soon?

src/AbstractGenerator.h
abadams: I think this comment will need updating depending on how #6760 resolves.

steven-johnson: Agreed.

steven-johnson: I've updated the comments here to reflect the landing of #6760.

steven-johnson:
Monday morning review ping -- I've made some minor changes (mostly to comments) and I think this is ready to land, with the caveat that there is some followup work that is desirable but more sensible to do separately:
- `generate_filter_main` and `execute_generator` should probably be moved from Generator.h/cpp to AbstractGenerator.h/.cpp since that's all they need; since it would be pure code motion, doing it here would just confuse the PR a bit.
- It's become clear from discussion on #6729 that the nature of GeneratorContext needs to be re-examined (i.e. does it actually need to exist or are we better off going back to its contents being more like plain old GeneratorParams); that is an important discussion to have but I don't think that needs to block this from landing. The most important point of contention (IMHO) has been mostly settled by #6760 landing. 
- It's highly likely that `AbstractGenerator` will continue to mutate slightly in response to ongoing work on #6764, but that's to be expected

abadams (approved):

PR #6644, opened 2022-03-09
Fix TSAN-reported data races in profiler shutdown code

steven-johnson:
TSAN reports data races for halide_profiler_pipeline_end() and  halide_profiler_shutdown() when reading/writing current_func; we need to acquire the lock.

### DIFF
--- a/src/runtime/profiler.cpp
+++ b/src/runtime/profiler.cpp
@@ -429,20 +429,33 @@ __attribute__((destructor))
 WEAK void
 halide_profiler_shutdown() {
     halide_profiler_state *s = halide_profiler_get_state();
-    if (!s->sampling_thread) {
-        return;
+    {
+        // Acquiring the mutex here is essential, as the sampling thread
+        // could be contending for current_func.
+        ScopedMutexLock lock(&s->lock);
+        if (!s->sampling_thread) {
+            return;
+        }
+        s->current_func = halide_profiler_please_stop;
     }
 
-    s->current_func = halide_profiler_please_stop;
     halide_join_thread(s->sampling_thread);
-    s->sampling_thread = nullptr;
-    s->current_func = halide_profiler_outside_of_halide;
 
-    // Print results. No need to lock anything because we just shut
-    // down the thread.
-    halide_profiler_report_unlocked(nullptr, s);
+    {
+        // Note that locking here may not be essential (since we just shut
+        // down the thread), but it's good practice and I don't know of
+        // a good reason not to.
+        ScopedMutexLock lock(&s->lock);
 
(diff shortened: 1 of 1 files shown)


src/runtime/profiler.cpp
```diff
@@ -429,20 +429,33 @@ __attribute__((destructor))
 WEAK void
 halide_profiler_shutdown() {
     halide_profiler_state *s = halide_profiler_get_state();
-    if (!s->sampling_thread) {
-        return;
+    {
+        // Acquiring the mutex here is essential, as the sampling thread
```
abadams, line 433: The sampling thread only reads current_func, and this thread only writes current_func. I take it TSAN doesn't like this. Non-sampling threads also write current_func via a volatile store, which apparently TSAN is cool with? It's all intentionally racing. Perhaps this just needs to be a call to halide_profiler_set_current_func so that all the stores are at least done in the same way.

abadams:
All the places that directly store to current_func should be replaced with calls to halide_profiler_set_current_func. It makes no sense to lock some writes to it when we're mostly just doing non-locked volatile writes to it.

src/runtime/profiler.cpp
steven-johnson: TSAN assumes that all data races are bad. There are ways to annotate a data race as "tsan should ignore this" but it's hard to do in our runtime environment for various reasons.

FWIW, I found this while trying to track down an intermittent hang when using the `profile` feature with WebAssembly (when `wasm_threads` is also enabled); some cases work just fine but some hang, and so far I haven't found any rhyme or reason to why. TSAN doesn't work under wasm and likely never will (wasm is 32-bit only and TSAN requires 64-bit apparently), but running the hanging case under "native" compilation reported these data races -- sadly, fixing them didn't have any effect on the hang when compiling to wasm, but clean code is good code.

Given that the current design relies on being able to have ignorable data races, I guess I'll just drop this PR.

zvookin: The solution is likely to make the variable atomic. (The issue is if we don't use a mutex or atomic, the reads and writes aren't guaranteed to get barriers, etc.)

zvookin:
I will look at this later. I'm not sure mutexing is the right answer, but given that TSan is complaining and that there is an observable bug for wasm, we very likely need to do something.

steven-johnson:
> there is an observable bug for wasm

There is, but I think it's not directly related to this; rather, it's related to the peculiarities of the "pthreads" that Emscripten offers, and restrictions on which threads can block (hint: not main) and whether you should pre-create apool of pthreads (apparently: yes).

PR #6646, opened 2022-03-12
[CMake] Deduplicate `Halide_LLVM_VERSION` and `LLVM_PACKAGE_VERSION`

LebedevRI:
As far as i can tell, they are always exactly identical,
and i don't believe the claim that `LLVM_PACKAGE_VERSION`
doesn't exist in upper scopes is valid.

### DIFF
--- a/dependencies/llvm/CMakeLists.txt
+++ b/dependencies/llvm/CMakeLists.txt
@@ -20,9 +20,6 @@ message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 message(STATUS "Using ClangConfig.cmake in: ${Clang_DIR}")
 
-# LLVM_PACKAGE_VERSION does not propagate to higher scopes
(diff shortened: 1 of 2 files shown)


alexreinking:
Do all versions set it as a cache variable?

LebedevRI:
> Do all versions set it as a cache variable?
>
> Even if they do, is that a guarantee going forward? Generally, modern CMake packages are moving towards _not_ caching such things so that sibling folders can load different versions. Seems like the kind of thing we'd just need to add back later.

I don't follow at all, but you already rely on `LLVM_PACKAGE_VERSION` being set into cache variable:
https://github.com/halide/Halide/search?q=LLVM_PACKAGE_VERSION

> Also, is this _every_ use of `Halide_LLVM_VERSION`? IIRC that variable gets `configure_file`'d into our package.

I mean, can you point me at the other uses?
https://github.com/halide/Halide/search?q=Halide_LLVM_VERSION

alexreinking:
Ah, I was thinking of `packaging/CMakeLists.txt` which uses `LLVM_PACKAGE_VERSION`. Thanks for the clean-up!

alexreinking (approved):

Issue #6515, opened 2021-12-24
Missing AVX2_znver3 target


LebedevRI:
It seems like the `-march=<>` is mimicked via target features,
in which case there's a large number of cpu's/features missing.
At the very least, i'd like to have one for znver3.

LebedevRI:
Ok, so. Trying to think of the generalization here, and i have a general philosophy/design question.

As everyone knows, clang/gcc have three ways to control the target CPU:
* `-mattr=` - specify which features (ISA sets) to enable/disable
* `-mtune=` - without changing the enabled ISA sets, target this CPU
* `-march=` - both of the above

Now, there is only a single configuration method in halide, the features string, and features can only be enabled,
so there isn't much room to model all that flexibility, so i guess if the particular cpu is specified in the features,
it should mean both "target this cpu" and "tune for this cpu"? Or is there desire to be able to separate the two?

LebedevRI:
Correction: in string form, we could have `target_cpu=ABC,tune_cpu=DEF`, that would be transparently stored as strings
in `Halide::Target`, IFF we are okay with not being able to configure them via `Target::set_feature()`/`Target::has_feature()`.

abadams:
I'd rather just add lots more flags, honestly. The current flags are equivalent(ish) to mattrs. I have no objection to adding flags like Target::Zen3, or Target::IceLake. These flags would be used primarily by CodeGen_LLVM::mcpu() and would act like mtune.

If we want something like march instead, we could extend complete_x86_target in CodeGen_X86.cpp and do similar in other backends. That's where implied features get turned on (e.g. avx512 implies avx2). Should the target flag "SkylakeX" imply "avx512"? Quite possibly no, because it's a clock speed hit, so I guess that suggests the flags would have to be equivalent to mtune. But perhaps that's surprising... thoughts?

LebedevRI:
Would be considered too disruptive to add `Target::{set,get}_{target,tune}_cpu(string)`?
I think my main worry is the fact that fixing #6514 will result in the CPU always being present in `Target`,
and going the full-flexibility route from the start might be less issue-prone.

abadams:
It would be less disruptive to add more target flags, and we generally prefer enums to strings. I don't think the cpu always being present in a resolved "host" target is a bad thing. If anything it makes it clearer what's going on. We'd have to get a bit more clever with cpuid though.

LebedevRI:
IF we can only add features, and presumably we don't want add two features per CPU,
[...]

PR #6648, opened 2022-03-15
`-mtune=native` CPU autodetection for AMD Zen 3 CPU

LebedevRI:
Based on the base support for `-mtune=` implemented in #6655,
as discussed in https://github.com/halide/Halide/issues/6515,
this implements bare-bones CPU detection using vendor/family/model derived from CPUID,
and adds autodetection for `znver3` via that framework.

The local result is:
```
$ ./src/autoschedulers/adams2019/get_host_target
x86-64-linux-tune_znver3-avx-avx2-f16c-fma-sse41
```

Usage caveat: if you do not intend to run the code you are compiling on the host,
but `host` is part of the target string, you may want to ensure that you override the autodetected tune CPU.

Fixes https://github.com/halide/Halide/issues/6515

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -62,6 +62,63 @@ static void cpuid(int info[4], int infoType, int extra) {
 #endif
 #endif
 
+#if defined(__x86_64__) || defined(__i386__) || defined(_MSC_VER)
+
+enum class VendorSignatures {
+    UNKNOWN,
+    GENUINE_INTEL,
+    AUTHENTIC_AMD,
+};
+
+VendorSignatures get_vendor_signature() {
+    int info[4];
+    cpuid(info, 0, 0);
+
+    if (info[0] < 1) {
+        return VendorSignatures::UNKNOWN;
+    }
+
+    // "Genu ineI ntel"
+    if (info[1] == 0x756e6547 && info[3] == 0x49656e69 && info[2] == 0x6c65746e) {
+        return VendorSignatures::GENUINE_INTEL;
+    }
+
+    // "Auth enti cAMD"
+    if (info[1] == 0x68747541 && info[3] == 0x69746e65 && info[2] == 0x444d4163) {
+        return VendorSignatures::AUTHENTIC_AMD;
+    }
+
+    return VendorSignatures::UNKNOWN;
+}
+
+void detect_family_and_model(unsigned info0, unsigned &family,
+                             unsigned &model) {
+    family = (info0 >> 8) & 0xF;  // Bits 8..11
+    model = (info0 >> 4) & 0xF;   // Bits 4..7
+    if (family == 0x6 || family == 0xF) {
+        if (family == 0xF) {
+            // Examine extended family ID if family ID is 0xF.
+            family += (info0 >> 20) & 0xFf;  // Bits 20..27
+        }
+        // Examine extended model ID if family ID is 0x6 or 0xF.
+        model += ((info0 >> 16) & 0xF) << 4;  // Bits 16..19
+    }
+}
+
+Target::Processor get_amd_processor(unsigned family, unsigned model) {
+    switch (family) {
+    case 0x19:
+        // FIXME: do we need to check model number?
+        return Target::Processor::ZnVer3;
+    default:
+        break;  // Unknown AMD CPU.
+    }
+
+    return Target::Processor::ProcessorGeneric;
+}
+
+#endif  // defined(__x86_64__) || defined(__i386__) || defined(_MSC_VER)
+
 Target calculate_host_target() {
     Target::OS os = Target::OSUnknown;
 #ifdef __linux__
@@ -111,8 +168,18 @@ Target calculate_host_target() {
 #else
     Target::Arch arch = Target::X86;
 
+    VendorSignatures vendor_signature = get_vendor_signature();
+
     int info[4];
     cpuid(info, 1, 0);
+
+    unsigned family = 0, model = 0;
+    detect_family_and_model(info[0], family, model);
+
+    if (vendor_signature == VendorSignatures::AUTHENTIC_AMD) {
+        processor = get_amd_processor(family, model);
+    }
+
     bool have_sse41 = (info[2] & (1 << 19)) != 0;
     bool have_sse2 = (info[3] & (1 << 26)) != 0;
     bool have_avx = (info[2] & (1 << 28)) != 0;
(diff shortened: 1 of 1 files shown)


src/Target.cpp
```diff
@@ -164,19 +234,23 @@ Target calculate_host_target() {
...
             }
+            // FIXME: port to family/model -based detection.
             if ((info2[1] & avx512_cannonlake) == avx512_cannonlake) {
                 initial_features.push_back(Target::AVX512_Cannonlake);
 
                 const uint32_t avx512vnni = 1U << 11;  // vnni result in ecx
                 const uint32_t avx512bf16 = 1U << 5;   // bf16 result in eax, with cpuid(eax=7, ecx=1)
                 int info3[4];
                 cpuid(info3, 7, 1);
+                // FIXME: port to family/model -based detection.
```
abadams, line 253: Could you elaborate on this comment? Why should we detect vnni instructions using the family/model?

src/Target.cpp
```diff
@@ -164,19 +234,23 @@ Target calculate_host_target() {
         }
         if ((info2[1] & avx512) == avx512) {
             initial_features.push_back(Target::AVX512);
+            // FIXME: port to family/model -based detection.
             if ((info2[1] & avx512_knl) == avx512_knl) {
```
abadams, line 238: These avx512_foo features don't mean "tune for processor foo, which happens to have avx512", they mean "enable the avx512 instructions that exist on this processor". It was our crude attempt at making sense of the many overlapping avx512 feature sets.

So I think it's a bit of a misnomer to call the new feature AVX2_znver3. I'd just make it Znver3

src/Target.cpp
LebedevRI: These comment are about things like
```
initial_features.push_back(Target::AVX512_SapphireRapids);
```
... which is an inherently fragile approach, it relies on the fact
that there is _currently_ no other CPU that supports these features.

For example, if you know that the CPU only supports AVX2,
that doesn't tell you which CPU that is,
it could be any number of CPU's (haswell,znver1,znver2,znver3).

src/Target.cpp
LebedevRI: Aha, that is an interesting context, thanks!

But still, `avx512_foo` does result in `-mcpu=foo`, so i would say my comment isn't entirely wrong.

> So I think it's a bit of a misnomer to call the new feature AVX2_znver3. I'd just make it Znver3

Yay :)

src/CodeGen_X86.cpp
```diff
@@ -33,7 +33,8 @@ Target complete_x86_target(Target t) {
     }
     if (t.has_feature(Target::AVX512_Cannonlake) ||
         t.has_feature(Target::AVX512_Skylake) ||
-        t.has_feature(Target::AVX512_KNL)) {
+        t.has_feature(Target::AVX512_KNL) ||
+        t.has_feature(Target::Znver3)) {
```
LebedevRI, line 37: Should it also enable other features that would have been autodetected for `znver3` otherwise?
(currently, `f16c` and `fma`)

LebedevRI:
Buildbot issues seem to not be caused by this PR.
Right now i'm looking for guidance re the thoughts i've raised in the PR's description.

LebedevRI:
Ok, the more i think about this the more i believe this needs to be done in two steps.
Let me split this up.

LebedevRI:
@abadams @zvookin @steven-johnson

Ok, now that the baseline #6655 is here, i've rebased this.
I think it's the most bare-bones variant of this that is possible.
I'm not sure if it would be better to implement the autodetection for the other tune CPU's at the same time?

After this the only missing piece is run-time multi-version selection.

src/Target.cpp
```diff
@@ -62,6 +62,63 @@ static void cpuid(int info[4], int infoType, int extra) {
 #endif
 #endif
 
+#if defined(__x86_64__) || defined(__i386__) || defined(_MSC_VER)
```
steven-johnson, line 65: Not necessarily as part of this PR, but... this file is getting a lot of processor-specific ifdef stuff in it, which is getting to be a code smell at this point IMHO (I'm a big believer in trying to keep "core" code mostly-core and segregating processor/platform/etc specific code into separate units where possible). We should think about whether and/or how to reorganize this (after these changes) to keep this from exploding into a mess of messiness.

LebedevRI: Makes sense to me.

LebedevRI:
Those two windows builds are still flaky  I'm open to suggestions whether it would instead be best/fastest
to implement autodetection for all CPU's added in #6655 in one go, or not.

src/Target.cpp
```diff
@@ -62,6 +62,63 @@ static void cpuid(int info[4], int infoType, int extra) {
...
+
+enum class VendorSignatures {
+    UNKNOWN,
```
steven-johnson, line 68: Style nit: most enums in Halide use InitialCapCamelCase rather than ALL_CAPS

src/Target.cpp
```diff
@@ -62,6 +62,63 @@ static void cpuid(int info[4], int infoType, int extra) {
...
+
+    // "Auth enti cAMD"
+    if (info[1] == 0x68747541 && info[3] == 0x69746e65 && info[2] == 0x444d4163) {
+        return VendorSignatures::AUTHENTIC_AMD;
+    }
+
+    return VendorSignatures::UNKNOWN;
+}
+
+void detect_family_and_model(unsigned info0, unsigned &family,
```
steven-johnson, line 94: `info0` is gathered as an `int` but passed as an `unsigned`. Seems more sensible to just continue to use `int` for all of these here, unless there's a compelling to use `unsigned`.

src/Target.cpp
```diff
@@ -62,6 +62,63 @@ static void cpuid(int info[4], int infoType, int extra) {
...
+        }
+        // Examine extended model ID if family ID is 0x6 or 0xF.
+        model += ((info0 >> 16) & 0xF) << 4;  // Bits 16..19
+    }
+}
+
+Target::Processor get_amd_processor(unsigned family, unsigned model) {
+    switch (family) {
+    case 0x19:
+        // FIXME: do we need to check model number?
```
steven-johnson, line 111: Does this need an Issue to be opened to track it?

src/Target.cpp
```diff
@@ -165,19 +232,23 @@ Target calculate_host_target() {
         }
         if ((info2[1] & avx512) == avx512) {
             initial_features.push_back(Target::AVX512);
+            // FIXME: port to family/model -based detection.
```
steven-johnson, line 235: Nit: there are existing `FIXME` in Halide, but generally we prefer `TODO`, I think

steven-johnson (commented):
LGTM with nits

src/Target.cpp
LebedevRI: Oh. But we are really interested in (unsigned) bit fields, so sign really isn't a thing.
How about we canonicalize in opposite direction instead?

src/Target.cpp
LebedevRI: Let me just add the model check i guess :)

src/Target.cpp
LebedevRI: Hm, so i guess it was signed because that is what required for MSVC `__cpuidex()`.
I took liberty of changing to unsigned and adding a wrapper to hopefully keep `__cpuidex()` happy,
but i guess i could go to `int` instead if you insist...

steven-johnson: It's `int` everywhere else. Let's just use that.

LebedevRI: Well, okay then.

LebedevRI:
> LGTM with nits

Thank you for the review, the bots seems to be green.

steven-johnson (approved):

LebedevRI:
Thank you!

PR #6649, opened 2022-03-15
Update initialization of WABT `store` field to work with top-of-tree

steven-johnson:
The copy and move assignment operators for Store are going away; initialize ours in the traditional way to work with both old and new code.

attn: @sbc100


alexreinking (approved):
Lgtm

PR #6650, opened 2022-03-15
Innopeaktech/hexagon instrumentation

vladl-innopeaktech:
This is a set of utilities for deterministic profiling of Halide code generated for the Hexagon DSP. It works by injecting code during Halide Generator invocation which calls a Qualcomm DSP library which emits the profiling data on logcat. The logcat output is then converted to JSON-formatted profiles with another `logcat2json`, and these JSON files can then be visualized with some python scripts that create text reports and a Graphviz dotfile.


pranavb-ca:
@vladl-innopeaktech  - thank you for this PR. @aankit-ca  - FYI. Can you PTAL?

aankit-ca:
@vladl-innopeaktech Can you please document the code?

vl-01:
@aankit-ca I have added comments to the c++ code. Is this what you had in mind?

vladl-innopeaktech:
@aankit-ca I've added comments to the code, is this what you had in mind?

aankit-ca:
Hey @vladl-innopeaktech, I'll take a look at the patch today. Can you add a use-case to show how it can be used.

vl-01:
@aankit-ca I've updated the code as per your comments, use case coming soon.

util/Hexagon/reports/cfgdot
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/python
```
alexreinking, line 1: Please give Python files a `.py` extension

util/Hexagon/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  json
+  URL https://github.com/nlohmann/json/releases/download/v3.10.5/json.tar.xz
+)
+FetchContent_MakeAvailable(json)
+
+FetchContent_Declare(
+  cxxopts
+  URL https://github.com/jarro2783/cxxopts/archive/refs/tags/v3.0.0.tar.gz
+)
+FetchContent_MakeAvailable(cxxopts)
```
alexreinking, line 22: FetchContent is a liability for a number of reasons... please replace these with calls to `find_package`.

util/Hexagon/README.md
```diff
@@ -0,0 +1,57 @@
...
+Then link `profiling.o` with `lib${MY_LIB}_skel.so`.
+
+To build `logcat2json`, from the source root directory:
+```
+    mkdir build
+    cmake -B ./build -S . && cmake --build ./build
+```
+
+## Profiling
+1. Include the header `instrumented_generator.h` in your Halide generator source.
```
alexreinking, line 26: This isn't getting installed, so how are users supposed to get this header?

util/Hexagon/instrumented_generator.h
```diff
@@ -0,0 +1,246 @@
+#pragma once
+
+/** \file
+ * Defines a Generator interface which injects timing code around loops and
+ * emits a report to logcat via adsprpc (Hexagon only).
+ */
+
+#include "Halide.h"
```
alexreinking, line 8: There needs to be an `INTERFACE` target for this that depends on `Halide::Halide`.

util/Hexagon/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+project(hexagon-instrument
+    VERSION 0.1
+    LANGUAGES CXX)
+
+include(FetchContent)
+
+add_executable(logcat2json logcat2json.cpp)
+set_target_properties(logcat2json PROPERTIES
+    CXX_STANDARD 17
+  )
```
alexreinking, line 11: Use `target_compile_features(logcat2json PRIVATE cxx_std_17)`.

util/Hexagon/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.16)
+project(hexagon-instrument
+    VERSION 0.1
+    LANGUAGES CXX)
```
alexreinking, line 4: Should this really be its own project?

alexreinking (changes requested):
Some more thought needs to be put in to the build system and integrations for this PR. In particular, the use of FetchContent is inappropriate (see, for example, #5092) and this feature isn't packaged at all.

util/Hexagon/reports/cfgdot
alexreinking: Furthermore, these should all be `#!/usr/bin/env python` to allow virtual environments to work.

util/Hexagon/CMakeLists.txt
vl-01: Not necessarily. What is a better place for it?

vladl-innopeaktech: Not necessarily. What is a better place for it?

util/Hexagon/README.md
vladl-innopeaktech: I'm not sure. I have tried adding an instrumented_generator target, I think this should do it.

vladl-innopeaktech:
@alexreinking @aankit-ca Updated according to requests

util/Hexagon/README.md
alexreinking: This is going to require a change in `/packaging/CMakeLists.txt` to install the `logcat2json` tool and the `InstrumentedGenerator` target. The latter should be grouped with the other platform-agnostic (i.e. source-only) targets.

util/Hexagon/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+find_package(nlohmann_json)
+find_package(cxxopts)
```
alexreinking, line 2: Is it actually useful have the `InstrumentedGenerator` target _without_ the `logcat2json` tool? If not, then I think these should both be `REQUIRED` and the `add_subdirectory(Hexagon)` call above should be guarded by an off-by-default `option(WITH_UTILS_HEXAGON ...)` instead.

util/Hexagon/README.md
alexreinking: The `reports/*.py` scripts should also be installed somewhere beneath `share/`, maybe near the autoscheduling helper scripts? When installed, we can add symlinks without the `.py` extension, if you like.

util/Hexagon/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Adds timing and reporting to Hexagon code generated by a Halide Generator.
+
+## Input Files
+`instrumented_generator.h`
+`profiling.cpp`
+`logcat2json.cpp`
+
+## Building
+For now, build `profiling.cpp` with your DSP code. Later we will likely include this precompiled binary.
```
alexreinking, line 11: We should also get this built and packaged as a library. Or, if it's small, we can inject it via `target_sources(INTERFACE)`, as we do with GenGen (`Halide::Generator`).

alexreinking (changes requested):
I think we're getting closer, but we need to get the packaging stuff solved now, or (in my experience) it won't get solved later.

alexreinking:
Actually, I'm wondering if the overall structure of this makes the most sense for upstreaming... it looks to me like the custom lowering pass should be controlled by a feature flag, rather than by interposing a `Generator` subclass between the existing first-party generator infra and the user's actual generator.

As for the `profiling.cpp` stuff... maybe some of it can be a runtime module that's only included when the feature flag is present. I think there's some precedent for this.

The command-line tools for interpreting the output seem fine as they are.

This way enabling/disabling this instrumentation can happen at the command line or in the build system, rather than by necessarily changing sources. I think this feature will be better for everyone if this is the case. New users won't have to change their code, for one.

vladl-innopeaktech:
That sounds good. The flag would need to be applied at the build configuration level, right? 
Where could I look for an example of this runtime module design?

vladl-innopeaktech:
@alexreinking I have put the build behind a feature flag (I think) and am building the DSP library based on whether or not the Hexagon SDK environment is present. I should put a prebuilt copy somewhere in the distribution as well, do you have any suggestions on where?
I'm not sure how to set the InstrumentedGenerator using a feature flag. This is something that would be controlled by a generator runtime argument? I can imagine pulling this off by setting a define like `#define Generator InstrumentedGenerator` before compilation based on the presence of an argument. Not sure that is gonna be the right way.
I haven't really used cmake's `install` functionality before so I could be missing a lot of important practices here.

vladl-innopeaktech:
@alexreinking I've made an attempt at enabling the instrumentation without the user having to include. I wound up moving the header into `src` as I needed visibility there, although if we need better separation, I could declare an external function pointer instead. Is this similar to what you had in mind?

alexreinking:
@vladl-innopeaktech - I can take a look, but I don't feel comfortable being the sole approver for a change as large as this.

vladl-innopeaktech:
@alexreinking It's definitely not ready to merge yet, I just need a few pointers on how to proceed. Once I've got everything in place and then test it, we could bring more reviewers in.

Issue #6651, opened 2022-03-17
Halide feature names for AVX512 don't match commonly-used abbreviations


steven-johnson:
We use names that are (presumably) based on the specific microarchitecture where a feature was first introduced (e.g. avx512_sapphirerapids), but the apparently-standard extensions are much terser and more granular, e.g. from https://en.wikichip.org/wiki/x86/avx-512:

```
AVX512F - [AVX-512 Foundation](https://en.wikichip.org/w/index.php?title=x86/avx512f&action=edit&redlink=1) is base of the 512-bit SIMD instruction extensions which is a comprehensive list of features for most HPC and enterprise applications. AVX-512 Foundation is the natural extensions to AVX/AVX2 which is extended using the [EVEX](https://en.wikichip.org/w/index.php?title=x86/evex&action=edit&redlink=1) prefix which builds on the existing [VEX](https://en.wikichip.org/w/index.php?title=x86/vex&action=edit&redlink=1) prefix. Any processor that implements any portion of the AVX-512 extensions MUST implement AVX512F.
AVX512CD - [AVX-512 Conflict Detection](https://en.wikichip.org/w/index.php?title=x86/avx512cd&action=edit&redlink=1) Instructions offer additional vectorization of loops with possible address conflict.
AVX512DQ - [AVX-512 Doubleword and Quadword](https://en.wikichip.org/w/index.php?title=x86/avx512dq&action=edit&redlink=1) Instructions add new 32-bit and 64-bit AVX-512 instructions (conversion, transcendental support, etc..).
AVX512PF - [AVX-512 Prefetch](https://en.wikichip.org/w/index.php?title=x86/avx512pf&action=edit&redlink=1) Instructions add new prefetch instructions for gather/scatter and PREFETCHWT1 .
AVX512ER - [AVX-512 Exponential and Reciprocal](https://en.wikichip.org/w/index.php?title=x86/avx512er&action=edit&redlink=1) Instructions (ERI) offer 28-bit precision RCP, RSQRT and EXP transcendentals for various scientific applications.
[...]

PR #6655, opened 2022-03-19
`-mtune=`/`-mcpu=` support for x86 AMD CPU's

LebedevRI:
As per https://github.com/halide/Halide/issues/6651#issuecomment-1072789066,
implements `tune` features, that only drive `-mcpu=` for LLVM,
and don't imply any further halide features.

Caveats/missing from this patch:
* Autodetection
* What about picking optimal multiversion at runtime?
* This only handles AMD processors. Intel part can be done, but it is a bigger list. Though, perhaps, that would allow requiring the tune on X86?
* Do we want to deduplucate this by moving it to a header with a macro?
* Are tests wanted?


zvookin:
These don't affect anything in Halide lowering do they? They're just passthroughs to LLVM. In this case, I'd propose we add a string list to Target that carries the LLVM string. It can be parsed/serialized in the targets string as e.g. "llvm_tune_<string>". The selection between differently compiled kernels using halide_can_use_target_features won't handle thing, but I don't think it will work right with these tuning flags anyway. (Effectively one will need some way of picking a closest tuning for a given processor if there isn't an exact match. And even for an exact match, that requires CPUID sniffing to do something with the feature flags, which I'm not sure makes sense.)

LebedevRI:
Thank you for taking a look. I believe there are three questions in total here, so let me split them up.

> These don't affect anything in Halide lowering do they? They're just passthroughs to LLVM.

This is a great question. I believe, when you are asking whether or not that affects anything in Halide lowering,
you aren't asking about *this* pr (because clearly, it doesn't), but about the larger, greater picture.

I believe the answer is: i disagree, i don't believe that to be the reality.
Each CPU is different, and expecting that no useful information can be extracted
from knowing the exact target CPU may be, well, setting the stage for suboptimal results.

For example, only znver2 finally has single-op AVX256 operations,
before that they were 2op, including loads/stores. I suspect that does affect the optimal schedule:
https://github.com/halide/Halide/blob/be1269b15f4ba8b83df5fa0ef1ae507017fe1a69/apps/conv_layer/conv_layer_generator.cpp#L142-L146

> In this case, I'd propose we add a string list to Target that carries the LLVM string. It can be parsed/serialized in the targets string as e.g. "llvm_tune_".

This won't really work as long as we have to be able to serialize them into halide feature string.
The CPU name is not guaranteed to e.g. not have `-` (the feature separator).

> The selection between differently compiled kernels using halide_can_use_target_features won't handle thing, but I don't think it will work right with these tuning flags anyway. (Effectively one will need some way of picking a closest tuning for a given processor if there isn't an exact match. And even for an exact match, that requires CPUID sniffing to do something with the feature flags, which I'm not sure makes sense.)

There are two questions here - picking the exact match, and picking the fallback in case of no exact match.

I don't see why picking the exact match would be anything other than trivial,
in fact i have already implemented that in [pull request #6648](https://github.com/halide/Halide/pull/6648/files#diff-1ec9cb9c532c21501d9579aa801ada182bf1138ed052a271a0c429dbcb9f5de0),
we'll simply need to perform the same CPU detection we'll have to do
when expanding `host` pseudo-feature. In fact, i will argue that doing so is required
to achieve optimal results, otherwise multi-versioning will end up penalizing performance.

As for picking the fallback, i'm not sure, but picking *any* compatible variant will be fine in principle.
I guess it depends as to how much custom logic (to do that) can be added to run-time.

zvookin:
Are tuning flags ever used together or is it a "pick one and only one" thing? My general take is the idea is to identify a single microarchitecture, at least for a broad swath of the functionality. (E.g. I could imagine two tuning specifications, one for the scalar part and one for the vector part. Especially in a more decoupled design. But generally everything is connected and one wants to identify the entire microarchitecture model at once.) If so, using bit flags to identify them is not a great fit both in terms of encoding and that it does not implicitly enforce the required "one hot" invariant.

Per the "do tuning flags ever affect lowering" my impression is this needs to be "no, they do not," at least if the point Alex made about ISA enables being separate from tuning is strictly followed. One may indeed choose to used different schedules for different microarchitectures, but that comes from the user or autoscheduler and is not a change in the execution paths in lowering.

This does mean choosing optimized kernels may need to pick between options based solely on the tuning info, not ISA support. I'm not convinced ```halide_can_use_target_features``` is the right thing for that. Mainly because it returns a boolean and any prioritization is done by choosing the ordering it is called in. That is not possible when there isn't mostly a superset relationship between the options. (E.g. SSE3 dominates SSE2, and SSE3 code doesn't run on SSE2 only CPUs. However if AMD and Intel introduce processors with exactly the same instruction set, but different performance characteristics, ```halide_can_use_target_features``` strictly return the same thing for both for all inputs because the code can run. If I choose to only tune for one of them, that code needs to be used on the other.)

Note my point here isn't that ```halide_can_use_target_features``` should constrain our adding support for tuning, but rather I think we're going to need a new runtime mechanism or to tell people to implement their own kernel selection for tunings.

For encoding strings to pass to LLVM where dash ("-") may occur in the flag name, we'll need an escape mechanism or some change to syntax of target strings. An example would be to have a new separator that starts an uninterpreted (string typed) list of key value pairs. (There's a bit of design there. Or we could just use JSON or something.) Some of the keys would just have their values passed through to LLVM. The canonicalization would sort and do a few things to ensure equal targets produce the same string, but generally the new information just be carried in Target.

Adding a map to ```Target``` might be a bit costly in in argument passing, but the first order change there would be to make sure ```Target``` is passed by reference where possible. Another approach would be to move to pooled storage for certain things and have a compilation context the pool is attached to. That way there'd only be one instance of a given target value. This could be used for other things too, but is a larger code change.

zvookin:
I guess I was thinking we'd add Target flags to control lowering for the cases were there is actual code that pays attention to tuning. Of course one would likely want that flag to also turn on the LLVM tuning flag, which means the features would interact with the string typed stuff, which means Halide has to know something about the strings.

An enum and an added field is a win over feature flags if it is indeed single valued. The shortcoming here is one still cannot access a given backend tuning unless Halide has added support for it. E.g. right now when I want to assess the difference targeting a specific microarchitecture makes in LLVM, I hand edit Halide code. The enum would at least give a place to put that and maybe make it into public Halide, if it isn't just a one off, but there are a lot of these and in some sense they may not be entirely enumerable. Could have an "Other" value in the enum I guess that is backed by a string...

LebedevRI:
Moved per-CPU tunes into their own enum, out of features.
Is this closer to what you had in mind?

src/Target.cpp
```diff
@@ -555,10 +599,11 @@ void bad_target_string(const std::string &target) {
         }
     }
     user_error << "Did not understand Halide target " << target << "\n"
-               << "Expected format is arch-bits-os-feature1-feature2-...\n"
+               << "Expected format is arch-bits-os-processor-feature1-feature2-...\n"
```
abadams, line 602: This implies that "processor" is a mandatory field. I'd just leave it out here. To a user it acts like a feature flag, except that you can only set one of them.

abadams:
Works for me. @zvookin ?

src/Target.h
```diff
@@ -50,6 +50,24 @@ struct Target {
     /** The bit-width of the target machine. Must be 0 for unknown, or 32 or 64. */
     int bits = 0;
 
+    /** The specific processor to be targeted, tuned for.
```
zvookin, line 53: I would like a comment after each processor giving a longer name, such as the manufacturer and "Zen" expanded, possibly the year they were introduced, etc. These may make sense to the person who wrote them today, but they will not a year from now or necessarily to anyone else. Basically enough to give one an idea if the chip one has handy matches something in the list and to help Google searches narrow it down.

src/Target.cpp
zvookin: I think it is pretty important for this to not invalidate existing target strings. We definitely need to accept strings with no processor specified on input. (I'm guessing that is already the case but have not read the entire PR.) Open question whether output should omit the processor if it is generic. Strictly speaking, always inserting it will likely break something so we would need to phase that in if we are serious about backward compatibility.

src/Target.cpp
```diff
@@ -484,7 +510,19 @@ bool merge_string(Target &t, const std::string &target) {
                 return false;
             }
             os_specified = true;
+        } else if (lookup_processor(tok, t.processor)) {
+            if (processor_specified) {
+                return false;
+            }
+            processor_specified = true;
         } else if (lookup_feature(tok, feature)) {
+            if (tok.substr(0, std::strlen("tune_")) == "tune_") {
```
LebedevRI, line 519: Hm, dead code, forgot to delete.

src/Target.cpp
LebedevRI: > I think it is pretty important for this to not invalidate existing target strings. We definitely need to accept strings with no processor specified on input. (I'm guessing that is already the case but have not read the entire PR.)

Right, (at least currently?) it's not required, and defaults to generic,
which results in feature-driven tune CPU selection.

> Open question whether output should omit the processor if it is generic. Strictly speaking, always inserting it will likely break something so we would need to phase that in if we are serious about backward compatibility.

Well, should we omit or not? :)

src/Target.h
LebedevRI: Should be done, do tell if you want it to be even more detailed.

src/Target.cpp
abadams: I vote we omit it printing it when it's generic, and we also omit it from this error string, because as it stands it implies processor is a required field.

LebedevRI: Right, since `host` feature is expanded, i guess not printing `tune_generic` is safe to do.

Though, i think we should still list the known/supported CPU's, no?
How about just explicitly saying here that CPU is optional?

src/Target.cpp
```diff
@@ -307,6 +308,31 @@ bool lookup_arch(const std::string &tok, Target::Arch &result) {
...
+    {"tune_amdfam10", Target::Processor::AMDFam10},
+    {"tune_btver1", Target::Processor::BtVer1},
+    {"tune_bdver1", Target::Processor::BdVer1},
+    {"tune_bdver2", Target::Processor::BdVer2},
+    {"tune_bdver3", Target::Processor::BdVer3},
+    {"tune_bdver4", Target::Processor::BdVer4},
+    {"tune_btver2", Target::Processor::BtVer2},
+    {"tune_znver1", Target::Processor::ZnVer1},
+    {"tune_znver2", Target::Processor::ZnVer2},
+    {"tune_znver3", Target::Processor::ZnVer3},
```
LebedevRI, line 324: Important design consideration: currently, the string key is effectively identical to the LLVM CPU string,
and i would really really like to keep it that way, so the proper `tune_*` can be autogenerated easily
from the LLVM CPU string.

LebedevRI:
@zvookin @abadams thank you for taking a look!
I believe i have addressed all current review notes, feel free to point out if i missed any!

src/Target.cpp
zvookin: Want to go ahead and put that in a comment here or on the enum?

zvookin:
Looks solid to me. Let me chat with Steven before approval to see whether we want to pull this for a spin first. Either way, should be able to land in the next couple days I think.

python_bindings/src/PyEnums.cpp
```diff
@@ -82,6 +82,21 @@ void define_enums(py::module &m) {
         .value("RISCV", Target::Arch::RISCV)
         .value("WebAssembly", Target::Arch::WebAssembly);
 
+    py::enum_<Target::Processor>(m, "TargetProcessor")
+        .value("TuneGeneric", Target::Processor::ProcessorGeneric)
```
steven-johnson, line 86: Suggestion: here and elsewhere, consider always listing all these enums in lexicographical order, so that there is an obvious and consistent order and thus additions will go into an obvious place.

src/Target.cpp
steven-johnson: SGTM on both counts

src/Target.cpp
```diff
@@ -555,13 +592,16 @@ void bad_target_string(const std::string &target) {
...
-               << "Expected format is arch-bits-os-feature1-feature2-...\n"
+               << "Expected format is arch-bits-os-processor-feature1-feature2-...\n"
                << "Where arch is: " << architectures << ".\n"
                << "bits is either 32 or 64.\n"
                << "os is: " << oses << ".\n"
+               << "processor is: " << processors << ".\n"
                << "\n"
                << "If arch, bits, or os are omitted, they default to the host.\n"
                << "\n"
+               << "If processor is omitted, it default to the tune_generic.\n"
```
steven-johnson, line 603: `               << "If processor is omitted, it defaults to tune_generic.\n"`

src/Target.h
```diff
@@ -50,6 +50,26 @@ struct Target {
...
+     * Corresponds to processor_name_map in Target.cpp. */
+    enum Processor {
```
steven-johnson, line 55: Important Concern: Surely all these values need to echoed into an enum in HalideRuntime.h, just as we do for Features?

Nit #1: Add a comment reminding coders that these values are *not* going to have the same numeric values over time (as is the case for Features already)

Nit #2: Per previously suggestion, seems like it would be simpler to always list these lexicographically, to make keeping all future changes/additions consistent a simpler task. Is there a good reason to order them in a different way?

LebedevRI: > Important Concern: Surely all these values need to echoed into an enum in HalideRuntime.h, just as we do for Features?

Ok, done.
After that, this enum becomes sorted/sortable, but that new one can't be sorted.

steven-johnson:
No red flags from a test Google integration. LGTM.

src/Target.h
zvookin: I'm not sure putting it in the runtime is necessary to land this.

LebedevRI: > I'm not sure putting it in the runtime is necessary to land this. Specifically, there is no use at runtime for them right now, though it could be surfaced somehow to help runtime dispatch. But that isn't plumbed through and can be added later.

And removed again :)

zvookin (approved):
I think it is down to at most a couple comment changes or such. Approved.

LebedevRI:
Not really sure what is wrong with those two windows CI checks.

steven-johnson:
> Not really sure what is wrong with those two windows CI checks.

Looks like a flake, I restarted them. (Our Windows buildbots aren't as robust as they really should be...)

LebedevRI:
> > Not really sure what is wrong with those two windows CI checks.
> 
> Looks like a flake, I restarted them. (Our Windows buildbots aren't as robust as they really should be...)

Almost there... :)

LebedevRI:
Thanks everyone!

Looks like the last bot finally went green.
Any further comments, @abadams @zvookin @steven-johnson ?
Anything for me left to do here?

abadams (approved):

LebedevRI:
Thank you!

PR #6659, opened 2022-03-25
Allow `make test_apps` to work with ASAN

steven-johnson:
With asan or tsan in the target, there is a space in the OPTIMIZE var, so it needs to be quoted.


steven-johnson:
Monday morning review ping

abadams (approved):

PR #6661, opened 2022-03-29
Add synthetic GeneratorParams to stubs (both C++ and Python)

steven-johnson:
Currently, there is no way to set 'synthetic' GeneratorParams via a GeneratorStub. ('synthetic' == things like "fieldname.type" for an input Buffer with an undefined static type.)

C++ GeneratorStubs are rarely used, but the Python stubs are useful, and (more importantly) revising this behavior to be possible will make some future work on Python generators more consistent.

Note that this should (in theory) have no effect on any existing C++ stub users; the new fields added to the stubs all have unique names, and aren't added to the default signatures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1549,6 +1549,7 @@ STUBTEST_GENERATOR_ARGS=\
 	array_input.type=float32 array_input.size=2 \
 	int_arg.size=2 \
 	tuple_output.type=float32,float32 \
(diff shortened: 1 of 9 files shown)


steven-johnson:
After consideration overnight, I think this PR is probably overkill for what I need; I'm probably going to cherry-pick the most important parts from this into a new PR and then ponder the rest.

PR #6664, opened 2022-03-29
Upgrade the PyStub code to handle synthetic GeneratorParams

steven-johnson:
PyStub is the code that lets Python code call a (C++) Generator; this upgrades the code to allow passing in of 'synthetic' GeneratorParams (e.g. buffer.type), which it previously didn't handle.

At the moment this doesn't make a big difference (I'm not aware of ~any meaningful amount of code outside of the Python tests that use PyStub), but this will streamline some possible changes in an experimental branch and make future diffs less painful.

### DIFF
--- a/python_bindings/correctness/pystub.py
+++ b/python_bindings/correctness/pystub.py
@@ -174,10 +174,27 @@ def test_complexstub():
                     simple_input=input,
                     array_input=[ input, input ],
                     float_arg=float_arg,
-                    int_arg=[ int_arg, int_arg ],
-                    untyped_buffer_output_type="uint8",
+                    int_arg=[int_arg, int_arg],
+                    # We can put GeneratorParams anywhere in the list we want;
+                    # they will be examined and applied before any inputs are
+                    # processed, and thus, any type/dimension constraints will be
+                    # enforced on inputs.
+                    #
+                    # Note that in Python, synthetic GeneratorParams
+                    # replace the '.' with a double underscore ('__'), so setting
+                    # untyped_buffer_output.type is accomplished like so:
+                    untyped_buffer_output__type="uint8",
                     extra_func_input=func_input,
-                    vectorize=True)
+                    vectorize=True,
+                    # Let's set some others, even though we don't need to here:
+                    # We can specify a halide Type via string or object here:
+                    simple_input__type=hl.UInt(8),
+                    untyped_buffer_input__type="uint8",
+                    # Can specify a list-of-types for Tuple output
+                    # tuple_output__type=[hl.Float(32), hl.Float(32)],
+                    # Alternately, we could specify comma-delimited string:
+                    tuple_output__type="float32,float32",
+                 )
 
     # return value is a tuple; unpack separately to avoid
     # making the callsite above unreadable
--- a/python_bindings/correctness/complexstub_generator.cpp
+++ b/python_bindings/correctness/complexstub_generator.cpp
@@ -17,7 +17,6 @@ Halide::Buffer<Type, 3> make_image(int extra) {
 
 class ComplexStub : public Halide::Generator<ComplexStub> {
 public:
-    GeneratorParam<Type> untyped_buffer_output_type{"untyped_buffer_output_type", Float(32)};
     GeneratorParam<bool> vectorize{"vectorize", true};
     GeneratorParam<LoopLevel> intermediate_level{"intermediate_level", LoopLevel::root()};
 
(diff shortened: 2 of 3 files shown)


python_bindings/correctness/pystub.py
```diff
@@ -174,10 +174,27 @@ def test_complexstub():
...
-                    untyped_buffer_output_type="uint8",
+                    int_arg=[int_arg, int_arg],
+                    # We can put GeneratorParams anywhere in the list we want;
+                    # they will be examined and applied before any inputs are
+                    # processed, and thus, any type/dimension constraints will be
+                    # enforced on inputs.
+                    #
+                    # Note that in Python, synthetic GeneratorParams
+                    # replace the '.' with a double underscore ('__'), so setting
+                    # untyped_buffer_output.type is accomplished like so:
```
alexreinking, line 185: Why is this the case? Can't it be something like `untyped_buffer_output={"type": "uint8"}`?

steven-johnson: eh... I guess it could be? I avoided that mainly because when I wrote this I thought I'd want to implement the same thing for C++ stubs, in which case having a common syntax would be desirable, and the __ approach is more straightforward to manage for C++. I'll give your suggestion a try and see how it feels.

steven-johnson: Actually, there *is* a very good reason that syntax won't work as-is: inputs that also need their type (etc) set. Currently:

    untyped_buffer_input=constant_image,  // line 173 in this PR
    untyped_buffer_input__type="uint8",  // line 186 in this PR

With this proposal, we'd have the obviously-illegal:

    untyped_buffer_input=constant_image, 
    untyped_buffer_input={"type": "uint8"},

We could finesse this in various ways, of course, e.g.

    # require a dict that includes value as well as type, dim, etc for these cases
    untyped_buffer_input={"value": constant_image, "type": "uint8"},

...but that feels weird and awkward IMHO.

alexreinking: I see your point and I keep going back and forth on whether using `__` or a dictionary like this is _more_ weird and awkward. Double underscores generally indicate language-level magic in Python, which isn't really the case here. On the other hand, the `"value"` key isn't really mirrored anywhere else in the API.

Why can't we get the type out of the `constant_image` object? When it's created on line 150, it's typed with `hl.UInt(8)`.

steven-johnson: > Why can't we get the type out of the constant_image object?

(sorta-topic-change to other work going on elsewhere...)

We *can*, the question is whether we *should*: there is a distinction between "is passed-in-argument well-typed" (yes, it is), vs "is the type required by the receiver well-defined or not" (in this case, not).

Currently, in C++ Generators (and by extension, the almost-never-used C++-based stubs), we take the approach that we don't require the receiver to be explicitly typed iff we can infer a concrete type from the passed-in-argument. This seems appealing, and works, but has several drawbacks that I'd like to avoid in the still-work-in-progress Python generator design:
- It complicates the internal code for Generators, since we may no longer know the type constraints on Inputs/Outputs until you actually execute the Generator to produce the IR.
- It impairs the readability for Generators, since you may not be able to look at the code (or even the code + the build file containing GeneratorParam values) and know what the expected types are. (For 'types', also include 'dimensions' and 'array-size', of course.)

IMHO a superior design would have been to require that all 'untyped' things be explicitly specified via GeneratorParams, so that no Generator code must be executed to deduce the signature of the resulting function. Trying to change that design point for C++ Generators would be a fool's errand at this point, but I'd really like to make this a design point to enforce in Python.

steven-johnson: > I see your point and I keep going back and forth on whether using `__` or a dictionary like this is _more_ weird and awkward. Double underscores generally indicate language-level magic in Python, which isn't really the case here. 

Yeah, I see your point and don't disagree. 

Arguably, the 'bug' here is that 'synthetic' params like this were created with a '.' in their names as a way to ensure they didn't overlap with any other param, since '.' was not and is not a valid user-provided name -- since there wasn't (and still isn't) a way to specify these for C++ stubs (which are rarely used, a nightmare to implement and use, and probably will never be enhanced), this isn't a limitation, but for Python-based stubs (which are potentially *much* more useful, for various reasons), being able to support these would be quite useful.

So the real question here is: is there a syntax that fits reasonably with Python named-argument convention which is reasonable?

steven-johnson: Any further thoughts on this? I'm not especially happy with the `__` approach, but lack a better idea at the moment...

alexreinking (approved):
I don't love the `__`, but now that we're on the other side of the Halide 14 window, I feel like we're safe to change this if inspiration strikes and we find a better solution. Better not to let this go stale.

steven-johnson:
> Better not to let this go stale

This will keep for a while. I'd rather get it right before landing it, and some of the related design is still in flux, so I'm gonna hold off until there's something we actually like, or we have a pressing need for it.

alexreinking:
Sounds good. I re-requested a review from myself so it doesn't get merged accidentally.

steven-johnson:
After further thought, I think this is actually moot for this use case -- when calling the Stub, we can always infer the type(s) (etc) that we need from the actual inputs. Closing.

PR #6669, opened 2022-03-30
Fix GPU depredication/scalarization

shoaibkamil:
When vectorizing GPU tiled code, we were not correctly handling predicated loads/stores.  This adds a test demonstrating the problem and fixes it for OpenCL, Metal, and D3D12Compute.


src/CodeGen_OpenCL_Dev.cpp
```diff
@@ -858,8 +858,50 @@ void CodeGen_OpenCL_Dev::CodeGen_OpenCL_C::visit(const Shuffle *op) {
...
+        // OpenCL requires using .s<n> format for extracting an element
+      ostringstream rhs;
+      rhs << print_expr(op->vectors[0]);
+      rhs << ".s" << op->indices[0];
+      print_assignment(op->type, rhs.str());
+    } else if (op->type.is_scalar()) {
+        CodeGen_C::visit(op);
     } else {
-        internal_error << "Shuffle not implemented.\n";
+        internal_assert(!op->vectors.empty());
```
abadams, line 870: This logic seems to be repeated three times. Is it really so different across the GPU backends that none of it can usefully be in the base class?

src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -957,6 +958,43 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::visit(const Atomic *op) {
...
+        // This code has always assumed/required that all the vectors
+        // have identical types, so let's verify
+        const Type t0 = op->vectors[0].type();
+        for (const auto &v : op->vectors) {
+            internal_assert(t0 == v.type());
+        }
+        string storage_name = unique_name('_');
+        rhs << "{";
+        for (int i : op->indices) {
+            rhs << vecs[i] << ",";
```
slomp, line 991: This will end up producing a "trailing comma" after the last vector and before the closing `}`; is this acceptable syntax?

src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -957,6 +958,43 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::visit(const Atomic *op) {
...
+        CodeGen_C::visit(op);
+    } else {
+        internal_assert(!op->vectors.empty());
+        for (size_t i = 1; i < op->vectors.size(); i++) {
+            internal_assert(op->vectors[0].type() == op->vectors[i].type());
+        }
+        internal_assert(op->type.lanes() == (int)op->indices.size());
+        const int max_index = (int)(op->vectors[0].type().lanes() * op->vectors.size());
+        for (int i : op->indices) {
+            internal_assert(i >= -1 && i < max_index);
```
slomp, line 972: Is `-1` a valid index?

slomp (approved):
Approving, with some comments to review

src/CodeGen_OpenCL_Dev.cpp
shoaibkamil: I considered adding a base class for the various `CodeGen_<x>::CodeGen_<x>_C` classes.  Right now each of them is derived from `CodeGen_C`.  However, I decided not to do that because it felt like a larger refactoring was necessary-- we probably could consolidate a lot of the logic into a superclass, since HLSL/OpenCL C/Metal have a lot of overlap.

I'm happy to this specific functionality into a superclass if you think that's a good idea.

src/CodeGen_D3D12Compute_Dev.cpp
shoaibkamil: Fixed!

src/CodeGen_D3D12Compute_Dev.cpp
shoaibkamil: It is acceptable, but I cleaned it up anyway to make it more consistent.

src/CodeGen_OpenCL_Dev.cpp
abadams: I was wondering if it could go into codegen_c, but I see that the shuffle handling there relies on a helper class.

I think it's time to add a CodeGen_GPU_C intermediate class and we can slowly move stuff into it over time.

src/CodeGen_D3D12Compute_Dev.cpp
```diff
@@ -957,6 +958,46 @@ void CodeGen_D3D12Compute_Dev::CodeGen_D3D12Compute_C::visit(const Atomic *op) {
...
+        }
+
+        string src = vecs[0];
+        ostringstream rhs;
+        // This code has always assumed/required that all the vectors
+        // have identical types, so let's verify
+        const Type t0 = op->vectors[0].type();
+        for (const auto &v : op->vectors) {
+            internal_assert(t0 == v.type());
+        }
```
rootjalex, line 987: This logic for verifying that the vector types match appears to be duplicated above (and likewise in the Metal and OpenCL backends)

steven-johnson:
Ready to land?

PR #6670, opened 2022-03-31
Clean up Python extensions in python_bindings

steven-johnson:
This was a giant mess of unspecified ugliness:
- We had two distinct kinds of Extensions that were sharing the same namespace, but weren't compatible. The difference between the two wasn't really documented.
- Generators for the python tests were mixed in willy-nilly; moved to their own subdir.
- We still weren't reliably using a linker script to limit exports everywhere.
- The target_export_script() function in CMake quietly did nothing (!) for Module libraries, which was wrong.
- Ensured that the PyStub.cpp and PyStubImpl.cpp files were compiled with both forms of visibility=hidden in CMake
- Cleaned up stuff related to the 'complexstub' and 'simplestub' generators (removing the now-redundant stub part of the name too)
- Probably other stuff?

There are probably errors and/or suboptimal bits in the CMake code, please review carefully.


steven-johnson:
Gentle review ping

steven-johnson:
Monday review ping

cmake/PythonExtensionHelpers.cmake
```diff
@@ -0,0 +1,115 @@
...
+    endif ()
+
+    # Create the Halide generator executable.
+    add_executable(${TARGET}.generator ${ARG_SOURCES})
+    target_link_libraries(${TARGET}.generator PRIVATE Halide::Generator ${ARG_LINK_LIBRARIES})
+
+    # TODO: this should work (and would be preferred to the code above)
+    # but CMake fails with "targets not yet defined"; investigate.
+    # add_halide_generator(${TARGET}.generator
+    #                      SOURCES ${ARG_SOURCES})
```
alexreinking, line 62: How can I repro this locally? Would like to get this working...

cmake/PythonExtensionHelpers.cmake
```diff
@@ -0,0 +1,115 @@
...
+                       PYTHON_EXTENSION ${TARGET}.py.cpp
+                       FEATURES ${ARG_FEATURES}
+                       PARAMS ${ARG_PARAMS}
+                       TARGETS cmake)
+
+    # Take the native-code output of the Generator, add the Python-Extension
+    # code (to make it callable from Python), and build it into the AOT Extension we need.
+    Python3_add_library(${TARGET} MODULE ${${TARGET}.py.cpp})
+    target_link_libraries(${TARGET} PRIVATE aot_${TARGET})
+    set_target_properties(${TARGET} PROPERTIES OUTPUT_NAME ${ARG_GENERATOR}) # Python3_add_library adds target info to name.
```
alexreinking, line 80: If we end up releasing this module to users, we will probably want a way to keep that info in the output name. IIUC, it's there to prevent the Python interpreter from loading an ABI-incompatible extension.

steven-johnson: Oops! I have to admit that I just cargo-culted this from previous code. Let's try removing it and seeing if things still work?

alexreinking: I probably added it in there earlier to work around something I didn't know I was doing wrong :upside_down_face:

cmake/PythonExtensionHelpers.cmake
alexreinking: Will take a look...

cmake/PythonExtensionHelpers.cmake
steven-johnson: ...it may be that actually using `distutils`/`setuptools` instead is the right answer, but it felt kinda heavyweight for the needs here. That said, maybe we'd be better off using it as a way to Set A Good Example? I dunno.

steven-johnson: FYI, simply removing line 80 makes things fail to load. What (exactly) is Python3_add_library doing to the name? It seems to be poorly documented...

alexreinking: https://github.com/Kitware/CMake/blob/eea53ea0a919b3c3f444433106ef578fcc429e5f/Modules/FindPython/Support.cmake#L3311-L3356

alexreinking: It runs the equivalent of `python3-config --extension-suffix` to determine the SOABI version (on my system this is `cpython-38-x86_64-linux-gnu`) and then prepends `.${SOABI}` to the usual suffix (e.g. `.so`, `.pyd`).

alexreinking: Also relevant: https://peps.python.org/pep-3149/

cmake/PythonExtensionHelpers.cmake
alexreinking: This is due to the way that `add_halide_generator` tries to ensure that our targets are available... it runs `find_package(Halide REQUIRED)`, which makes sense for third-party projects, but doesn't work so well for in-tree. Need to open an issue for this...

cmake/PythonExtensionHelpers.cmake
alexreinking: Turns out this comment is misleading... the code itself is fine. It's getting rid of the `py_aot_` prefix from the target name. But CMake isn't actually even adding the soabi info in the first place.

alexreinking (approved):

PR #6673, opened 2022-04-03
Future-proof 'processor` to `tune processor`

LebedevRI:
This is mainly a future proofing.

#6655 added certain `processor` fields, which specify the CPU
for which we should tune, without implying features,
but the field name is awfully generic.

Once we want to add a second processor field,
that will mean features, such a field should have a name
of `processor`/`processor_target`, yet `processor` already means `tune`...

I'm not renaming the enum, i don't see why we'd need a second one.

### DIFF
--- a/python_bindings/src/PyEnums.cpp
+++ b/python_bindings/src/PyEnums.cpp
@@ -83,7 +83,7 @@ void define_enums(py::module &m) {
         .value("WebAssembly", Target::Arch::WebAssembly);
 
     // Please keep sorted.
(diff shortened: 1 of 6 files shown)


steven-johnson (approved):
LGTM pending green buildbots

LebedevRI:
Thank you! This **is** a breaking change in a way, and my main hope
is that not many uses of that field were introduced just yet :)

LebedevRI:
Thank you!

PR #6683, opened 2022-04-07
Fix "set but not used" warnings/errors

steven-johnson:
Apparently XCode 13.3 has smarter warnings about unused code and emits warnings/errors for these, so let's clean them up.


steven-johnson:
(Note that right now, the x86 macbot is running 13.3 but the armbot is not; the latter needs an OS upgrade first)

abadams (approved):

PR #6691, opened 2022-04-08
Always mark _ucon as 'unused' in Codegen_C

steven-johnson:
even if asserts are enabled, since generated closure functions may not use it


abadams:
Should we perhaps rename it halide_maybe_unused, to avoid confusing readers?

steven-johnson:
> halide_maybe_unused

Done

vksnk (approved):

PR #6692, opened 2022-04-08
Enable `bugprone-implicit-widening-of-multiplication-result` and fix …

steven-johnson:
…errors

clang-tidy-13 has a new check, `bugprone-implicit-widening-of-multiplication-result`:

"The check diagnoses instances where a result of a multiplication is implicitly widened, and suggests (with fix-it) to either silence the code by making widening explicit, or to perform the multiplication in a wider type, to avoid the widening afterwards."

This seems like a reasonable thing, in theory, so I enabled it and manually fixed everything it complained about (the `-fix` option didn't work well here). Most of the actual changes are probably inconsequential, but checks of this sort seem appropriate for clang-tidy. What do other folks think?


src/HexagonOptimize.cpp
```diff
@@ -1254,7 +1254,7 @@ class VectorReducePatterns : public IRMutator {
                 // may not recognize.
                 if (const Shuffle *shuff = a.as<Shuffle>()) {
                     vector<int> a0_indices(out_lanes), a1_indices(out_lanes);
-                    for (int i = 0; i < out_lanes; i++) {
```
abadams, line 1257: This sort of change just seems bug prone in a different way. Casting an integer loop extent to unsigned makes it run ~forever if that integer is negative.

I guess we could use ptrdiff_t, but that's starting to get verbose. I think you ought to be able to index an vector at the product of two ints, and allocate a vector to be sized according to the product of two ints, and this check seems to flag that as bug prone. I don't think it's a net win.

src/Util.cpp
```diff
@@ -751,7 +751,7 @@ void run_with_large_stack(const std::function<void()> &action) {
     // as opposed to having them cause silent corruption. We pick an
     // amount of memory that should be comfortably larger than most
     // stack frames - 64k.
-    const size_t guard_band = 64 * 1024;
+    const size_t guard_band = (size_t)(64 * 1024);
```
abadams, line 754: I also think you ought to be able to a size_t as the product of two int literals.

src/HexagonOptimize.cpp
steven-johnson: > if that integer is negative

true, but if a vector-lane-count is negative in Halide, I'm gonna bet you a dollar there are a hundred other catastrophic results. (I'll change this instance, just sayin')

steven-johnson: > I think you ought to be able to index an vector at the product of two ints

Tell that to the C/C++ standards committee -- `int` is int32 here, and the promotion rules won't promote to int64 without casting, so large values would overflow. (In this particular case, we happen to know that's wildly unlikely, as lane counts just aren't going to get that large without serious misbehavior, but then, guarding against serious misbehavior is the point of this check.)

> I don't think it's a net win.

Yeah, I am hesitant about this check in general -- the concept sounds like a good one, but some of the checks seem poorly thought out. (e.g. it will warn for a constant expr like `32 * 1024 * 1024` where it could statically verify that the result doesn't exceed the type.) Unfortunately, the (current) state of this check is all-or-nothing; if we want it to catch the actual risks, we have to put up with dopey ones too. Maybe not worth enabling (yet), maybe would be worthwhile in future versions of clang-tidy? Hence the request for thoughts.

src/Util.cpp
steven-johnson: Yep, this is clearly just a dumb mistake on the part of this version of clang-tidy. One would think they'd have caught this.

PR #6694, opened 2022-04-08
Add `break` to avoid 'possible unintentional fallthru' warning

steven-johnson:


rootjalex (approved):

LebedevRI:
Wouldn't it be wise to enable `-Werror` in CI, even if not always-on-by-default?

steven-johnson:
> Wouldn't it be wise to enable `-Werror` at least in the CI, even if not always-on-by-default?

I'd say yes, absolutely, but that's a different issue, please open one :-)

PR #6695, opened 2022-04-11
Deprecate variadic-template version of Realization ctor

steven-johnson:
The variadic-template approach was useful before C++11 (!) added brace initialization, but preferring an explicit vector-of-Buffer is arguably better, and provides better symmetry with the Python bindings. 

Also, some drive-by tweaks to other Realization methods.


steven-johnson:
Gentle review ping

abadams (approved):

Issue #2082, opened 2017-05-25
C/C++ backend doesn't handle matlab feature at all


steven-johnson:
Resulting code is broken.

steven-johnson:
Note that matlab support currently requires _argv support, thus this is (also) blocked on https://github.com/halide/Halide/issues/2071

abadams:
Matlab support was removed

PR #6696, opened 2022-04-11
Drop support for Matlab extensions

steven-johnson:
Anecdotally, this hasn't been used in ~years, and the original author (@dsharletg) had suggested dropping it a while back. I'm going to propose we go ahead and drop it for Halide 15 and see who complains. (See also #2082)


abadams:
SGTM, but it would be nice to save the app and convert it to a regular C++ app, because it's our example of how to schedule IIR filters.

steven-johnson:
> it would be nice to save the app and convert it to a regular C++ app,

It's unlikely I'll have time to do that soon. I'll just leave this as a draft until someone (including) me has time to get around to it.

steven-johnson:
> SGTM, but it would be nice to save the app and convert it to a regular C++ app, because it's our example of how to schedule IIR filters.

So, actually, `apps/iir_blur` appears to be literally the same code, except with more-complete scheduling (e.g. for Cuda), so I think this really is 100% redundant.

abadams:
Oh yeah, now I recall we ported it to a regular app to use in autoscheduling research.

PR #6697, opened 2022-04-11
Fixes for top-of-tree LLVM

steven-johnson:
Defaulting to opaque pointers broke some runtime stuff. I think this should be what we need.


steven-johnson:
Please note that this should get us building again, but there appears to be a separate codegen injection of some sort that has broken a couple of tests -- I'd like to land this to get us less-broken and then fix that subseqeuently.

alexreinking (approved):

Issue #6702, opened 2022-04-13
Python-binding: Add method to convert from Func to pure stage?


apartridge:
In the [C++ Func API](https://halide-lang.org/docs/class_halide_1_1_func.html#a3f623d418efe24a783d8609d2718ea06) there is an operator to convert from Func to its pure stage:

> `Halide::Func::operator Stage	(		)	const`
> You can cast a [Func](https://halide-lang.org/docs/class_halide_1_1_func.html) to its pure stage for the purposes of scheduling it.

I could not find the same conversion in the Python wrapper. I was looking for it because I wanted to use `compute_with`, in order to calculate 3 functions in a `Pipeline` in the same loop. I was able to work around the issue by making a `LoopLevel` object.

Reproducer:

```
from halide import Pipeline, Func, Var, LoopLevel


def gpu_tile_in_place(func):
    TILE_SIZE = 16
    if func.dimensions() == 3:
        func.reorder(img, col, row)
    func.compute_root()
    func.gpu_tile(col, row, col_outer, row_outer, col_inner, row_inner, TILE_SIZE, TILE_SIZE)


col = Var("col")
row = Var("row")
img = Var("img")
col_inner = Var("col_inner")
col_outer = Var("col_outer")
row_inner = Var("row_inner")
row_outer = Var("row_outer")

output_A = Func("output_A")
output_A[col, row, img] = img * 2

output_B = Func("output_B")
output_B[col, row, img] = img + 1

output_C = Func("output_C")
output_C[col, row] = (col + row) / 2

gpu_tile_in_place(output_A)
gpu_tile_in_place(output_B)
gpu_tile_in_place(output_C)

output_pipeline = Pipeline([output_A, output_B, output_C])

# These don't work:
#output_B.compute_with(output_A, img)
#output_C.compute_with(output_A, col_inner)
# They give the following error:
"""
TypeError: compute_with(): incompatible function arguments. The following argument types are supported:
    1. (self: halide.Func, loop_level: halide.LoopLevel, align: List[Tuple[halide.VarOrRVar, halide.LoopAlignStrategy]]) -> halide.Func
    2. (self: halide.Func, loop_level: halide.LoopLevel, align: halide.LoopAlignStrategy = <LoopAlignStrategy.Auto: 3>) -> halide.Func
[...]

PR #6704, opened 2022-04-13
Python: make Func implicitly convertible to Stage (#6702)

steven-johnson:
This allows for `compute_with` and `rfactor` to work more seamlessly in Python.

Also:
- Move two compute_with() variant bindings from PyFunc and PyStage to PyScheduleMethods, as they are identical between the two
- drive-by removal of redundant `py::implicitly_convertible<ImageParam, Func>();` call

attn: @apartridge


abadams (approved):

steven-johnson:
(Adding 'backport me' because this might be easy & worth doing to backport to earlier versions, but definitely not essential)

steven-johnson:
arm32 failure is being addressed separately

PR #6711, opened 2022-04-17
`get_amd_processor()`: implement detection for the rest of supported AMD CPU's

LebedevRI:
I have *not* personally tested that these are detected correctly,

Cross-reference between
* https://github.com/llvm/llvm-project/blob/955cff803e081640e149fed0742f57ae1b84db7d/llvm/lib/Support/Host.cpp#L968-L1041
* https://github.com/llvm/llvm-project/blob/955cff803e081640e149fed0742f57ae1b84db7d/compiler-rt/lib/builtins/cpu_model.c#L520-L586
* https://github.com/gcc-mirror/gcc/blob/000c1b89d259fadb466e1f2e63c79da45fd17372/gcc/common/config/i386/cpuinfo.h#L111-L264

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -104,8 +104,41 @@ void detect_family_and_model(int info0, unsigned &family, unsigned &model) {
     }
 }
 
-Target::Processor get_amd_processor(unsigned family, unsigned model) {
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):

LebedevRI:
Thank you!

Issue #4916, opened 2020-05-12
C++ backend doesn't handle float16/bfloat16 correctly


steven-johnson:
It emits code that assumes that `float16_t` and `bfloat16_t` already exist, and can interoperate directly with other C++ numeric types (via cast, operators, etc), which isn't true, thus the resulting code won't compile.

(Note we have code to do most of this in Float16.h/.cpp but that is compile-time only, and not currently visible to runtime code.)

Issue #3967, opened 2019-06-27
There is no way to declare Halide::Runtime::Buffer<float16>


steven-johnson:
`Halide::float16_t` is a compiletime type, but there is no equivalent runtime type, and thus no overload for `halide_type_of<>()` which produces `Float(16)`. You can of course work around this by using a dynamically-typed buffer, but static typing is good, so we should remedy this somehow.

steven-johnson:
Did this get addressed as part of https://github.com/halide/Halide/pull/3667?

abadams:
Hrm, no I don't think so, because float16_t is still defined in libHalide, not in the runtime. So you can use it as long as you link libHalide, (and type_of is defined there), which is not super useful unless you're jitting. We could consider making the float16 implementation in Float16.h/Float16.cpp part of the runtime too.

steven-johnson:
Could we just add types to `HalideRuntime.h` that are strictly proxies for halide_type_of<>() for float16 and bfloat16?

alsrgv:
That would be great!

ArunabhaDeb2018:
Hello All,
Is it possible to do this now for fp16 data type
Halide::Runtime::Buffer<float16>
I have tried something like below
_Float16* temp=new _Float16[1000]
Buffer<_Float16>o(temp,1,1000);
I get the below error
But I get an expected a pointer type error. I am doing AOT compile wigh halide 16 for android ndk build. 
Kindly let me know if there is a way to use fp16 for halide buffers.

PR #6712, opened 2022-04-17
Finish various missing bits of float16/bfloat16 support

steven-johnson:
Fixes issues #3709, #3967, #4916. I think this is the meaningful remaining bits (outside of Python) that have been left unanswered for a while.

Note the addition of `halide_float16_t` and `halide_bfloat16_t` in HalideRuntime.h as runtime placeholders that are equivalent to the language-only variants in Float16.h.

Note that there are similar issues in the Python bindings, but I'm going to land those in a subsequent PR to minimize confusion.


src/CodeGen_C.cpp
```diff
@@ -315,6 +315,29 @@ class TypeInfoGatherer : public IRGraphVisitor {
     std::set<Type> vector_types_used;
 };
 
+std::string maybe_emit_numeric_constant(const Expr &e) {
```
zvookin, line 318: Should this use the word "emit" if it is just returning the text, not emitting it to the output?

steven-johnson: Fair, will fix

steven-johnson:
Something is wonky on ARM systems -- I'm guessing that the addition of the f16/bf16 arguments has somehow confused parameter passing for ARM but haven't nailed it down yet. Investigating.

src/CodeGen_C.cpp
```diff
@@ -315,7 +315,7 @@ class TypeInfoGatherer : public IRGraphVisitor {
     std::set<Type> vector_types_used;
 };
 
-std::string maybe_emit_numeric_constant(const Expr &e) {
+std::string maybe_print_numeric_constant(const Expr &e) {
```
zvookin, line 318: I would call this "stringify_if_constant" or "text_if_constant". "Maybe" is nice if one wants to give the routine a sense of mystery... One line comment "// Returns text of value if constant, empty string otherwise" would also be helpful. (Yes, it's a total nitpick comment at the top of a code review. My apologies, I find the "maybe" pattern questionable, but maybe it's just that I'm allergic to that Carly Rae Jepsen song :-))

zvookin: Maybe a bit more constructively, what happens if we move the 16-bit float handling logic here into the visitor for ```FloatImm```?

steven-johnson: > Maybe a bit more constructively, what happens if we move the 16-bit float handling logic here into the visitor for FloatImm?

This case doesn't use that visitor for emitting anything, so, we'd need to restructure code to ensure that visitation happened when emitting all FloatImm values.

steven-johnson:
Note that this now deliberately *disallows* Input<float16_t> and Input<bfloat16_t>, mainly for expediency; it's likely to be rare that using Input<float> instead isn't a simpler and better choice.

(The motivation here is that getting the codegen right on ARM is slightly tricky, in that float16-as-uint would be passed in an integer register, while float16-as-native-arm-type will be passed in an fp register; if all parts of codegen don't agree on the right type, amusing and hard-to-debug crashes can occur. While it's certainly desirable to work out all such flakiness in our code, just on general principle, it's not clear to me that this is likely to be worth the effort at the moment, at least not in comparison to other work to be done.)

steven-johnson:
Gentle review ping

abadams:
Real usage of float16 in Halide that I've seen generally involves a user float16 type, e.g. from some library, and thus requires adding a template instantion for halide_type_of to get it to play nice with things like Halide::cast and Halide::Runtime::Buffer. This is highly non-obvious. The addition of this new runtime type seems to confuse matters further, because you can't actually do anything with it, right? It's just something that has a halide_type_of instantiation? I've been stalling on approving this because I'm wondering if there's something different we can do to help here.

steven-johnson:
Good and fair points. I think I'll drop the PR in its current form and pull out the changes in smaller pieces, probably omitting the new runtime type entirely for now. I do think we need a 'standard' story for runtime types, but I'm not sure what that should look like.

steven-johnson:
...I'd some point I'd really like to circle back and finish this. (Or have someone else finish it.) It's not at all critical, but it bugs me.

steven-johnson:
@zvookin recently pointed me at https://clang.llvm.org/docs/LanguageExtensions.html#half-precision-floating-point, which points out that recent Clang versions have builtin types for these; presumably we should rely on those, falling back to a custom type on other compilers.

PR #6713, opened 2022-04-18
Codegen_C: check the result of snprintf()

steven-johnson:
Technically, we should check the return code of sprintf() for failure conditions. This is a minimal fix designed to quietly ensure that encoding errors aren't overlooked.

(Since our "C" output actually requires C++11 at this point, I was tempted to replace it with `ostringstream`, but decided to avoid bringing that in...)

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -64,6 +64,7 @@ const string headers = R"INLINE_CODE(
 #include <float.h>
 #include <limits.h>
 #include <math.h>
+#include <stdarg.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


zvookin:
Was an error actually encountered in practice?

This looks to be introducing a Halide runtime routine via the naming convention, which I don't think is the intention. I would make this strictly internal to C++ codegen and changing the naming to indicate that.

If the error actually occurred in the wild, I would add some sort of logging or fast fail if the target contains the Debug flag.  I recently spent too long debugging an issue with an embedded runtime dropping debug output on the floor. Debugging may seem like it's not mission critical, but decisions to fail silently in debug paths can cause significant productivity loss and increase frustration in during already stressful work.

zvookin:
Reading the man page for vsnprintf, it is not clear to me the page identifies a case where it can return an error. (As opposed to routines which output to a FILE *, which can fail on I/O errors.)

steven-johnson:
> it is not clear to me the page identifies a case where it can return an error

https://www.cplusplus.com/reference/cstdio/vsnprintf/ claims: 

```
The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character.
If an encoding error occurs, a negative number is returned.
Notice that only when this returned value is non-negative and less than n, the string has been completely written.
```

steven-johnson:
> This looks to be introducing a Halide runtime routine via the naming convention

Lots of other existing helpers in Codegen_C use this same convention (eg `halide_cpp_max`, `halide_cpp_min`, so really it's just following existing precedent. (Arguably we should change all of them, probably into a namespace since our generated "C" code now actually requires C++11 or later, per line 60...)

PR #6717, opened 2022-04-19
Combine string constants in combine_strings()

steven-johnson:
This is a pretty trivial optimization, but when printing (or enabling `debug`), it cuts the number of `halide_string_to_string()` calls we generate by ~half.

### DIFF
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -62,6 +62,10 @@ Expr stringify(const std::vector<Expr> &args) {
 }
 
 Expr combine_strings(const std::vector<Expr> &args) {
+    if (args.empty()) {
+        return Expr("");
+    }
+
     // Insert spaces between each expr.
     std::vector<Expr> strings(args.size() * 2);
     for (size_t i = 0; i < args.size(); i++) {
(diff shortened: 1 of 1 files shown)


src/IROperator.cpp
```diff
@@ -73,6 +73,20 @@ Expr combine_strings(const std::vector<Expr> &args) {
         }
     }
 
+    // Now combine all adjacent string literals, which is
+    // useful to reduce emitted code size when printing
+    size_t i = 0;
+    while (i < strings.size() - 1) {
```
abadams, line 79: Isn't this a for loop?

steven-johnson: Yes, except that in the 'continue' case we'd need to decrement i, since we want to stay on the same element and the for loop would increment it, and for a size_t that could underflow on the first element, and even if the increment 'fixes' it, it still feels icky and dangerous.

abadams: Oh right, I missed the continue.

abadams (approved):

src/IROperator.cpp
steven-johnson: ...we both missed the 'oops I will crash if passed an empty vector' bug :-(

Issue #6710, opened 2022-04-16
Migrate src/CodeGen_PTX_Dev.cpp to new pass manager


MaskRay:
Using LLVM's legacy PM for optimization pipeline was [deprecated in 13.0.0](https://releases.llvm.org/13.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir) and the relevant functionality was scheduled to be removed after 14.0.0 (got delayed). The legacy PM for code generation (assembly, object file generation; mainly addPassesToEmitFile) is still supported.

Clang dropped legacy PM support in [D123609](https://reviews.llvm.org/D123609) (milestone: 15.0.0). I plan to remove instrumentation passes like gcov, PGO, and sanitizers (milestone: 15.0.0).

LLVM's legacy PM for optimization pipeline may be completely removed in 16.0.0 (`llvm::create*Pass`, `populateModulePassManager`, etc). llvm-project `clang/lib/CodeGen/BackendUtil.cpp` can serve as an example using the new pass manager.

abadams:
This has been raised a few times before. We do not use the legacy PM for the optimization pipeline. We only use it for code generation.

MaskRay:
In `src/CodeGen_PTX_Dev.cpp`, these functions use the legacy PM for the optimization pipeline.
```cpp
    PassManagerBuilder b;
    b.OptLevel = 3;
    b.Inliner = createFunctionInliningPass(b.OptLevel, 0, false);
    b.LoopVectorize = do_loop_opt;
    b.SLPVectorize = true;
    b.DisableUnrollLoops = !do_loop_opt;

    target_machine->adjustPassManager(b);
 
    b.populateFunctionPassManager(function_pass_manager);
    b.populateModulePassManager(module_pass_manager);
```

abadams:
Ah, I didn't realize the PTX backend had lagged behind. We upgraded the main CodeGen_LLVM a while ago. Thanks for the heads up.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

[...]

PR #6718, opened 2022-04-19
Update CodeGen_PTX_Dev to use new PassManager

steven-johnson:
This was still using the LegacyPassManager for optimization, which will be going away at some point. (Code changes by @alinas; I'm just opening this PR on her behalf)

Fixes #[6710](https://github.com/halide/Halide/issues/6710)

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1115,9 +1115,6 @@ void CodeGen_LLVM::optimize_module() {
     llvm::CGSCCAnalysisManager cgam;
     llvm::ModuleAnalysisManager mam;
 
-    llvm::AAManager aa = pb.buildDefaultAAPipeline();
-    fam.registerPass([&] { return std::move(aa); });
-
     // Register all the basic analyses with the managers.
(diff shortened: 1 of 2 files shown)


steven-johnson:
attn @[MaskRay](https://github.com/MaskRay)

abadams (approved):

abadams:
@steven-johnson @alinas Do you recall the reason behind the change to CodeGen_LLVM in this PR? It doesn't seem related to PTX and it seems likely that it's the cause of the large performance regression mentioned in the issue.

steven-johnson:
I will ping @alinas directly.

LebedevRI:
I highly suspect that is depended on the LLVM version, it's not needed for fresh LLVM (14+ i would say),
but *is* needed for earlier ones: https://reviews.llvm.org/D113210

alinas:
IIUC correctly, the regression is for this PR which is including both commits: switch to new pass manager and cleaning up adding the default AA pipeline. Assuming removing the default AA pipeline is the issue, Roman's change will fix this.
If it doesn't, we'll need a reproducer to understand why the switch to the new pass manager caused the regression.

PR #6719, opened 2022-04-19
Smarten type_of<> for fn ptrs; fix async_parallel for C backend

### LINKED ISSUE #2093: C++ backend should be able to use Halide runtime for parallel.
Currently, the C++ backend uses "#pragma omp parallel" to emit parallel for loops. We should make this an option under a Target flag and also put in code to call halide_do_par_for with appropriate closures.

steven-johnson:
(Fixes #2093)

This basically just adds the right type annotations to make the parallel code produced by the C backend compile properly. This could have been fixed by inserted some brute-force void* casting into the C backend, but this felt a lot cleaner.

The one thing here I'm a little unsure about is how I extended the Type code to be able to handle function-pointer types correctly; it works but doesn't feel very elegant.

### DIFF
--- a/src/LowerParallelTasks.cpp
+++ b/src/LowerParallelTasks.cpp
@@ -240,6 +240,8 @@ struct LowerParallelTasks : public IRMutator {
             const std::string closure_arg_name = unique_name("closure_arg");
             auto closure_arg = make_scalar_arg<uint8_t *>(closure_arg_name);
 
+            Type closure_function_type;
+
             std::vector<LoweredArgument> closure_args(use_parallel_for ? 3 : 5);
             closure_args[0] = make_scalar_arg<void *>("__user_context");
             if (use_parallel_for) {
@@ -250,6 +252,7 @@ struct LowerParallelTasks : public IRMutator {
                 closure_args[1] = make_scalar_arg<int32_t>(t.loop_var);
                 closure_args[2] = closure_arg;
                 // closure_task_parent remains undefined here.
+                closure_function_type = type_of<halide_task_t>();
             } else {
                 // The closure will be a halide_loop_task_t, with arguments like:
                 //
@@ -275,6 +278,8 @@ struct LowerParallelTasks : public IRMutator {
                 closure_args[2] = make_scalar_arg<int32_t>(loop_extent_name);
                 closure_args[3] = closure_arg;
                 closure_args[4] = make_scalar_arg<void *>(closure_task_parent_name);
+
+                closure_function_type = type_of<halide_loop_task_t>();
             }
 
             {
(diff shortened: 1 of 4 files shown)


abadams (approved):

src/LowerParallelTasks.cpp
```diff
@@ -275,6 +278,8 @@ struct LowerParallelTasks : public IRMutator {
                 closure_args[2] = make_scalar_arg<int32_t>(loop_extent_name);
                 closure_args[3] = closure_arg;
                 closure_args[4] = make_scalar_arg<void *>(closure_task_parent_name);
+
```
zvookin, line 281: Seems a little odd to have a blank line here but the comment line up above.

src/Type.h
```diff
@@ -84,10 +84,11 @@ struct halide_handle_cplusplus_type {
...
     enum Modifier : uint8_t {
-        Const = 1 << 0,     ///< Bitmask flag for "const"
-        Volatile = 1 << 1,  ///< Bitmask flag for "volatile"
-        Restrict = 1 << 2,  ///< Bitmask flag for "restrict"
-        Pointer = 1 << 3,   ///< Bitmask flag for a pointer "*"
+        Const = 1 << 0,            ///< Bitmask flag for "const"
+        Volatile = 1 << 1,         ///< Bitmask flag for "volatile"
+        Restrict = 1 << 2,         ///< Bitmask flag for "restrict"
+        Pointer = 1 << 3,          ///< Bitmask flag for a pointer "*"
+        FunctionPointer = 1 << 4,  ///< Bitmask flag for a function pointer (if set, Pointer will also be set)
```
zvookin, line 91: If Pointer is also going to be set, I'm not sure that word should be repeated in the name here. On the one hand, this seems fine, on the other, I'm not sure if the designation for a function pointer is really right as we don't have anything like all the information about the type. So far as I can tell, we're using this to avoid putting the pointer annotation (*) in because the function pointer is actually a typedef that already contains the indirection, as they always do, and the type printing just happens to work in this way. To this end, I think the right name for this might be ```FunctionTypedef``` and the comment can indicate it should always have the ```Pointer``` flag set as well.

src/LowerParallelTasks.cpp
steven-johnson: Done

zvookin (approved):
Approving, pending comments on naming, etc. There was a fair bit of subtlety with getting all this to match up across all the platforms lat I worked on it, but Andrew simplified it somewhat and I don't think this affects LLVM backends at all so it should be fairly safe.

src/Type.h
steven-johnson: Done

PR #6721, opened 2022-04-20
Make CodegenC stop pretending it can generate plain C code

steven-johnson:
The public interface seems to claim you can produce C output (in addition to C++), but this is a lie; our output is C++11 (!) and has been for a while. Remove this pretense and clean up a bit.

Note that the only practical difference I see in the C-vs-C++ code was:
- whether to emit functions as extern "C" vs name-mangled; this was (and is) actually handled by the Target::CPlusPlusMangling feature directly.
- the C++ version emits all temporaries with `const` declarations, which is arguably good.

(No, I'm not going to bother renaming the class or file unless someone insists.)

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -3077,7 +3077,7 @@ void CodeGen_C::visit(const Allocate *op) {
             stream << op_name
                    << "[" << size_id << "];\n";
         } else {
-            stream << "*"
+            stream << "* const "
                    << op_name
                    << " = ("
                    << op_type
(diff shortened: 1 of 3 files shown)


steven-johnson:
(Opening as "draft" so that the buildbots can verify whether this has broken any GPU targets that use parts of this code.)

abadams:
Our generated headers use C++ features? I thought we were keeping them straight C.

steven-johnson:
> Our generated headers use C++ features? I thought we were keeping them straight C.

No, the headers a C-safe, but it's still identical code no matter how you set the flag.

abadams:
I was confused because the enum value CHeader disappeared, and now there's only CPlusPlusHeader, which seems to imply C++ features are kosher to use there.

zvookin:
I'm worried that we're losing any indication that the output in a header must be compilable with a plain C compiler. (E.g. any C++ features must be under #ifdef's.) At the very least this should be documented and we should have a test that ensures it is so. E.g. generate and compile something using plain C.

I get that this is not a fun thing but it is pretty important to maintain this promise for various pieces of interoperability.

steven-johnson:
The insertion of `const` broke various GPU backends that rely on this for codegen, so the PR is moot at this point. 

I may drop this PR as it's not particularly critical (it was just something I noticed while working on unrelated changes), but I do think there is cleanup and/or clarification to do in this area.

PR #6723, opened 2022-04-21
Remove `rounding_halving_sub` and non-existent arm rhsub instructions

rootjalex:
These instructions don't seem to exist, they don't get produced when compiling for arm. https://github.com/gcc-mirror/gcc/commit/2e828dfe91a39a7a898ca6d8827048760e08b236 suggests that they were supposed to exist at some point, but were never implemented.

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -297,14 +297,6 @@ const ArmIntrinsic intrinsic_defs[] = {
     {"vrhadds", "srhadd", Int(32, 2), "rounding_halving_add", {Int(32, 2), Int(32, 2)}, ArmIntrinsic::HalfWidth},
     {"vrhaddu", "urhadd", UInt(32, 2), "rounding_halving_add", {UInt(32, 2), UInt(32, 2)}, ArmIntrinsic::HalfWidth},
 
-    // SRHSUB, URHSUB - Halving sub with rounding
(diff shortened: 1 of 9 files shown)


steven-johnson:
LGTM

steven-johnson:
Ready to land?

rootjalex:
Discussed offline with @abadams , arm was the only backend to target `rounding_halving_sub`, and the semantics don't actually make sense, there would still be overflow at the extremes. i.e. for `int8_t`:
```
 (127 - (-128) + 1) / 2 = 128 (overflow)
```
Therefore, this PR now removes the intrinsic entirely.

abadams:
The one platform that seems to have a rounding_halving_sub-like-instruction is hexagon, but that one saturates instead of overflowing.

abadams (approved):
lgtm pending green

abadams:
Interesting. Looks like that's an artifact of risc-v doing a cross product of operations, rounding modes, and saturation. I don't think it makes sense to use that instruction unless you also set vxsat, making it a saturating_rounding_halving_sub (which we don't have, but could feasibly add). The saturating version actually makes sense. HVX has it too. A non-saturating version just seems like a footgun to me.

rootjalex:
The RISC-V spec has defined overflow behavior, which is interesting.
```
For vasub and vasubu, overflow is ignored and the result wraps around.
```
This behavior seems far less useful than the saturating variants, so it seems safe to still remove this.

rootjalex:
Only failing test appears unrelated

PR #6726, opened 2022-04-22
Rust binding doc updated

r7shekhawat:
pulling to updated the rust binding docs.


language_bindings/rust/src/runtime_tests.rs
```diff
@@ -0,0 +1,156 @@
+//todo re-add tests
```
alexreinking, line 1: This file is completely commented out, is that intentional?

language_bindings/rust/test_files/iir_blur.cpp
```diff
@@ -0,0 +1,179 @@
+// This file defines a generator for a first order IIR low pass filter
```
alexreinking, line 1: This appears to be a duplicate of `language_bindings/rust/examples/iir_blue/src/gens/iir_blur.cpp`. Is that so? Why can't there be a single file? Also, how is this file built? I see neither Make nor CMake rules.

language_bindings/rust/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "halide"
+description ="Rust build script and runtime libaries for conviently useing Halide code in rust projects"
+version = "0.1.0"
```
alexreinking, line 4: This will need to be kept in sync with the repository version somehow. Can this be programmed to parse the top-level CMakeLists.txt?

language_bindings/rust/README.md
```diff
@@ -0,0 +1,62 @@
+# halide rust bindings
+
+
+These instructions also assume a fresh install of Ubntu LTS 21.04.
```
alexreinking, line 4: Ubuntu* and 21.04 is not an LTS release

language_bindings/rust/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+First, update your system.
```
alexreinking, line 6: These instructions should duplicate the top-level README, but rather link to the appropriate section.

language_bindings/rust/examples/iir_blur/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use halide::runtime::{HalideBuffer};
+use halide::runtime::runtime_bindings::{halide_buffer_t, halide_type_t};//todo rust enum type to runtime mod and possibly escape from useing buffer_t
+use image::io::Reader;
+
+fn main(){
+
+    println!("halide mainish thing");
+
+    let img = Reader::open("images/cat.png")
```
alexreinking, line 15: I can't comment on the image itself, so I'll comment here. Where is that image from? What is its license?

language_bindings/rust/src/Runtime_bindings.rs
```diff
@@ -0,0 +1,300 @@
+
```
alexreinking, line 1: Has this code passed through [rustfmt](https://github.com/rust-lang/rustfmt) or another similar tool? Is there a `clang-tidy` equivalent for Rust that's worth using? [Clippy](https://github.com/rust-lang/rust-clippy), maybe?

Any such tools should run in a GitHub Actions workflows. There's already one for clang-format that can be extended to this (and generally to other languages down the line). I'm not a Rustacean myself, but I just want to make sure that we're using the proper tools for Rust code quality at the outset.

language_bindings/rust/src/build.rs
```diff
@@ -0,0 +1,330 @@
...
+            gen.arg(temp);
+        }
+        gen.output().expect("failed to run")
+    }
+
+    /// Make the Halide runtime
+    ///
+    /// The Halide runtime is required for using halide and contains the buffer_t and other useful functions
+    pub fn make_runtime(&self) -> Result<()> {
+        println!("Cmd to runt: {:?}", self.gen_exe.to_str().unwrap());
```
alexreinking, line 205: "Command to run" (not "runt")

alexreinking: These also kind of look like debug prints. Can we add some `debug() <<` equivalent that also honors the existing environment variables?

Mayor-Jellies:
Hey @alexreinking thanks for the fast comments and review. We've updated the branch and broken the code into a different stucture. We will update again once we go through your comments.

alexreinking:
I'm converting this to a draft PR since the code is still changing rather rapidly. When it's ready for broader review, feel free to mark it as such.

Mayor-Jellies:
@alexreinking We believe the code is ready for review, unable to mark it as ready for review due to permissions.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    /// This complies-> runs-> renames-> binds the generator
+    ///
+    /// panics if any step fails if it panics it is useful to call the functions separately and print there results or outputs
+    pub fn build_bind(&self) {
+        assert!(self.compile().status.success());
+
+        assert!(self.run_gen().status.success());
+
+        assert!(self.bind().is_ok());
```
Ralith, line 353: When a panic is desired, prefer `some_result.unwrap();` to `assert!(some_result.is_ok())`. That allows the panic message to convey some information about the failure.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+use std::{env, io};
+
+/// This builder must have a path to a complied Halide folder. It is used to build multiple generators quickly
+///
+///
+pub struct GenBuilder {
+    halide_path: PathBuf,
+    gen_path: PathBuf,
+
+    rs_output: PathBuf,
```
Ralith, line 28: The code below frequently calls `to_str().unwrap()` on these, and in some cases they are constructed directly from strings. If paths must be strings, it might be more robust and efficient to store a `String` from the start, avoiding all the conversions and possible panics.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                    .join(gen_name.as_str())
+                    .with_extension("cpp")
+                    .to_str()
+                    .unwrap()
+            );
+            temp_gen_exe.push(gen_name.as_str());
+        }
+        temp_gen_exe.with_extension("generator");
+
+        println!("temp: {:?}", temp_gen_exe.as_os_str());
```
Ralith, line 119: Avoid performing console I/O directly in library code, as it may interact poorly with downstream requirements. If you want to emit diagnostics, the `log` crate might be more appropriate, as it allows downstream code to filter and route diagnostics as desired. Alternatively, if this is debugging output for your own use and not of interest downstream at all, it should be removed.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+        println!(
+            "cargo:rustc-link-search=native={}",
+            self.rs_output.to_str().unwrap()
+        );
+
+        let mut gen = Command::new(self.gen_exe.to_str().unwrap());
+        gen.args(["-g", self.gen_name.as_str()]);
+        gen.args(["-f", self.gen_name.as_str()]);
+
+        //Todo change to build dir
```
Ralith, line 206: Is this still TODO?

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                .with_extension("a")
+                .to_str()
+                .unwrap(),
+        )
+    }
+
+    /// Renames the gen outputs to be what rust and bindgen expect on linux distros
+    ///
+    /// IE halide.a -> libhalide.a
+    pub fn rename(&self) -> Result<()> {
```
Ralith, line 288: This seems like an implementation detail. Does it really need to be public? Further, is there an underlying issue that should be fixed in Halide's build system rather than hacked around here?

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                self.rs_output
+                    .join(self.gen_name.as_str())
+                    .with_extension("h")
+                    .to_str()
+                    .unwrap(),
+            )
+            .allowlist_function(self.gen_name.as_str())
+            .blocklist_item("halide_buffer_t")
+            .generate()
+            .expect("unable to gen bindings")
```
Ralith, line 334: Prefer to propagate errors rather than panicing, or document why this panic is believed to be impossible.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+use std::fs;
+use std::io::{ErrorKind, Result};
+use std::ops::Add;
+use std::path::PathBuf;
+use std::process::{Command, Output};
+///Required module for use in build.rs scripts
+///
+/// This Module will compile and bind a generator and create a Halide runtime.
+///
+//TODO update rust docs
```
Ralith, line 18: Is this still TODO?

language_bindings/rust/rust_runtime/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+            device_interface: std::ptr::null(),
+            dimensions: 3,
+            host: self.data,
+            flags: self.flags,
+            padding: std::ptr::null_mut(),
+            type_: self.t,
+            dim: dim.as_mut_ptr(),
+        };
+
+        std::mem::forget(dim);
```
Ralith, line 57: This is a memory leak.

language_bindings/rust/rust_runtime/src/runtime_bindings.rs
```diff
@@ -0,0 +1,300 @@
+
+//! runtime_bindings is a manually generated using bindgen on halide_runtime.h.
+//!
+//! Eventually automatically generating this file may be a better approach, but may introduce stability issues.
```
Ralith, line 4: Incorrect indentation on this file.

Ralith, line 4: FWIW, pre-generated bindings are generally preferred because `bindgen` is a heavy dependency and its process is somewhat fragile. It might be nice to provide a shell script to update it, however.

language_bindings/rust/rust_runtime/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub width: i32,
+    pub height: i32,
+    pub channels: i32,
+    pub t: halide_type_t,
+    pub data: *mut f32,
+    pub flags: u64,
+}
+
+impl HalideBuffer {
+    pub fn create_buffer(&mut self) -> halide_buffer_t {
```
Ralith, line 20: Why does this need `&mut self`?

language_bindings/rust/rust_tests/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//!DO NOT PUBLISH
```
Ralith, line 1: Specify this explicitly in the Cargo.toml instead: https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

language_bindings/rust/rust_tests/src/runtime_tests.rs
```diff
@@ -0,0 +1,81 @@
+use std::ffi::c_void;
+use std::os::raw::c_int;
+
+use halide_runtime::*;
+use halide_runtime::runtime_bindings::*;
+
+#[test]
+fn i_got_here() {
+    assert_eq!(2 + 2, 4);
+}
```
Ralith, line 10: This can probably be dropped.

language_bindings/rust/rust_tests/src/runtime_tests.rs
```diff
@@ -0,0 +1,81 @@
...
+#[test]
+fn memalloc() {
```
Ralith, line 12: This test doesn't have any asserts. What is it verifying?

language_bindings/rust/rust_runtime/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use crate::runtime_bindings::*;
+
+/// Is a helper struct to more easily create buffer_t objects in Rust.
+pub struct HalideBuffer {
+    pub width: i32,
+    pub height: i32,
+    pub channels: i32,
+    pub t: halide_type_t,
+    pub data: *mut f32,
```
Ralith, line 15: Should this be `&mut [f32]` for safety?

language_bindings/rust/rust_runtime/src/runtime_bindings.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    #![allow(non_camel_case_types)]
+    #![allow(non_upper_case_globals)]
+
+    extern "C" {
+        pub fn halide_malloc(
+            user_context: *mut ::std::os::raw::c_void,
+            x: isize,
+        ) -> *mut ::std::os::raw::c_void;
+    }
```
Ralith, line 15: If this definition is provided by libhalide.a, then this crate must have a build script which instructs cargo to link libhalide.a.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+        assert!(execution_result.is_ok());
+
+        gen.rename_runtime();
+    }
+}
+
+impl Generator<'static> {
+    ///Make the generator executable using Halide GenGen and g++
+    pub fn compile(&self) -> Output {
+        let mut cmd_compile = Command::new("g++");
```
Ralith, line 158: For portability, this should use the `cc` crate rather than hard-coding a dependency on g++ specifically.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+///
+pub struct Generator<'a> {
+    gen_name: String,
+    gen_exe: PathBuf,
+    halide_path: PathBuf,
+    gen_path: PathBuf,
+
+    rs_output: PathBuf,
+
+    gcc_flags: Vec<&'a str>,
```
Ralith, line 46: Prefer using owned rather than borrowed values in structs unless there's a very good reason otherwise. Dealing with borrowed values inside opaque structs tends to complicate and constrain downstream code considerably.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+/// This builder must have a path to a complied Halide folder. It is used to build multiple generators quickly
+///
+///
+pub struct GenBuilder {
+    halide_path: PathBuf,
+    gen_path: PathBuf,
+
+    rs_output: PathBuf,
+
+    //generators: Vec<Generator<'static>>,
```
Ralith, line 30: Dead comment?

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+            temp.push_str("-debug");
+            gen.arg(temp);
+        }
+        gen.output().expect("failed to run")
+    }
+
+    /// Make the Halide runtime
+    ///
+    /// The Halide runtime is required for using Halide and contains the buffer_t and other useful functions
+    pub fn make_runtime(&self) -> Result<()> {
```
Ralith, line 224: Is the purpose of this crate to compile Halide code, or to compile the Halide implementation itself, or both? These are different tasks that should be decoupled. In particular, tooling to compile Halide code should support a pre-compiled external Halide implementation, installed in the environment as documented at https://github.com/halide/Halide/#getting-halide, without any special effort from the caller. Only if none are available should this infrastructure attempt to compile the Halide implementation itself, and it should do so transparently to the caller (i.e. without requiring a path to source code be passed in).

Ralith (changes requested):
Reviewed for rust style/design/idioms. I recommend running `cargo fmt` and `cargo clippy` to catch some general issues. There's also quite a lot of commented-out code that should be removed before this is merged.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                .to_str()
+                .unwrap(),
+        )
+    }
+
+    /// Run bindgen on a generator
+    ///
+    /// This bind only creates the genname funtion binding and specificly blocklists the buffer_t
+    pub fn bind(&self) -> Result<()> {
+        let bindings = bindgen::Builder::default()
```
Ralith, line 323: In the example, this seems to be used only to produce a single, simple `iir_blur` function declaration. `bindgen` is a very heavy dependency as it requires LLVM, and rather fragile as it's sensitive to the C/C++ build environment. If the minimal requirements in the example are representative, could `bindgen` be avoided by just having downstream code declare the interface in question themselves? This would also make downstream code more readable, as right now it's unclear what the signature of `iir_blur` is, exactly.

alexreinking:
> I recommend running cargo fmt and cargo clippy to catch some general issues.

Then I really think we should go a step further and run these tools in CI so that future Rust code continues to conform. It'd be consistent with our clang-format and clang-tidy usage/policy.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
+#![warn(missing_docs)]
+//#![warn(missing_doc_code_examples)]
+
+//!This crate is an example of calling Halide generated code from Rust.
+//!This crate also has an example of IIR-blur Halide app that get called from rust.
+//! IIR blur takes a image input, goes through the full Halide pipeline, and outputs a blurred image all using rust.
+//!
```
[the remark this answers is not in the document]
alexreinking: My opinion of this crate is also modulated by whether or not it is an example. If it's an example it's fine, but if not, it's quite far from being a general way to invoke a C compiler. It certainly won't work with MSVC, whose ABI Rust targets on Windows.

language_bindings/rust/apps/iir_blur/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+use halide_runtime::runtime_bindings::{halide_buffer_t, halide_type_t}; //todo rust enum type to runtime mod and possibly escape from useing buffer_t
+use halide_runtime::HalideBuffer;
+use image::io::Reader;
+
+fn main() {
+    println!("halide mainish thing");
+
+    let img = Reader::open("images/Hummingbird.jpg")
+        //Img source: https://commons.wikimedia.org/wiki/File:Hummingbird.jpg#filelinks
+        //Image released into public domain by Jon Sullivan PDPhoto.org
```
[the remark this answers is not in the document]
alexreinking: I would put it in a `LICENSE` file in `language_bindings/rust` using the [Debian format](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/):

```
Files: apps/iir_blur/images/Hummingbird.jpg
Copyright: Public Domain
Comment: https://commons.wikimedia.org/wiki/File:Hummingbird.jpg#filelinks
License: public-domain
 This work has been released into the public domain by its author, Jon Sullivan
 PD Photo.org. This applies worldwide. In some countries this may not be legally
 possible; if so: Jon Sullivan PDPhoto.org grants anyone the right to use this
 work for any purpose, without any conditions, unless such conditions are
 required by law.
```

language_bindings/rust/rust_build/src/lib.rs
alexreinking: No, static libraries don't get a `lib` prefix on Linux/Unix.

alexreinking: Not with the CMake build system, but maybe with GenGen.cpp's output? Our releases include `libHalide.a`, for example.

language_bindings/rust/rust_runtime/src/runtime_bindings.rs
alexreinking: > pre-generated bindings are generally preferred 

That's quite surprising to me. Not the way other binding tools I've worked with, work. It runs the risk of very rapidly getting the bindings out of sync with the main library. But trusting that this is the case...

> It might be nice to provide a shell script to update it, however.

...I think this is mandatory. At the very least, the command used to create this file needs to be documented _somewhere_.

language_bindings/rust/rust_build/src/lib.rs
alexreinking: > Only if none are available should this infrastructure attempt to compile the Halide implementation itself, and it should do so transparently to the caller (i.e. without requiring a path to source code be passed in).

I would sooner see an error in this case, rather than attempting to build Halide from scratch. Which LLVM to use? Does the user have 10+ minutes to wait for Halide to build?

Ralith: That's reasonable; e.g. the Rust `openssl` crate will panic in its build script if OpenSSL can't be found in the environment. It can also be useful to allow a custom search path to be passed in via an environment variable (e.g. `HALIDE_PATH` perhaps?).

language_bindings/rust/rust_build/src/lib.rs
Ralith: I'm a bit confused about whether this code is dealing with the Halide runtime or with compiled Halide code. If the latter, we should avoid creating a name collision with the runtime library.

language_bindings/rust/rust_runtime/src/runtime_bindings.rs
Ralith: Life sucks for a user who wants to compile something with FFI bindings, only to find that they need to install a suitable version of libclang. Particularly on Windows this can be a high bar, but even on Linux it can be very painful if e.g. a suitable version is not packaged. Given that the output doesn't vary between user machines for most sensible C APIs anyway, it makes more sense for it to be the maintainer's responsibility to keep it up to date.

It can be very helpful to use CI to assert that the generated bindings are up to date by rerunning the script and checking for a diff in the output. Particularly apt in this case, where there may not be a dedicated maintainer for the Rust bindings.

alexreinking: > It can be very helpful to use CI to assert that the generated bindings are up to date by rerunning the script and checking for a diff in the output. Particularly apt in this case, where there may not be a dedicated maintainer for the Rust bindings.

Yeah, I think this is sufficient.

language_bindings/rust/rust_runtime/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "halide_runtime"
```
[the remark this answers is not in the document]
Mayor-Jellies: we choose to name the dir rust_runtime and rust_build as if someone is searching through the whole halide repository we wanted it instantly clear this was for rust. But it certainly is a debatable point and the directory to crate inconsistency may be worse.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        let mut gen = Command::new(self.gen_exe.to_str().unwrap());
+        gen.args(["-g", self.gen_name.as_str()]);
+        gen.args(["-f", self.gen_name.as_str()]);
+
+        //Todo change to build dir
+        gen.args(["-o", self.rs_output.to_str().unwrap()]);
+        gen.env("LD_LIBRARY_PATH", self.halide_path.join("lib"));
+
+        //Todo add debug
```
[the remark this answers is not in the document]
Mayor-Jellies: dead comment will remove.

language_bindings/rust/rust_build/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                "-g",
+                self.gen_path.to_str().unwrap(),
+                temp.to_str().unwrap(),
+            ]);
+        }
+        // Output flag
+        cmd_compile.args(["-o", self.gen_exe.to_str().unwrap()]);
+
+        cmd_compile.args(self.gcc_flags.clone());
+        //compile.args(["-Wl,-rpath","-Wl,/home/rootbutcher2/CLionProjects/Halide-Rusts-tests/Halide/distrib/lib/"]);
```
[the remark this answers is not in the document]
Mayor-Jellies: This is how make builds a halide generator with gengen, this is left to draw attention to the difference. Should an in-place comment be made to highlight this difference? or simply remove the line?

language_bindings/rust/rust_build/src/lib.rs
Mayor-Jellies: We attempted to use CC crate but found that it was automatically adding lib and .a to the compilation, could not find a workaround here. To be clear this cmd_complie makes an executable that is run later. If you are aware of a workaround here to use CC we would prefer that as well.

language_bindings/rust/rust_build/src/lib.rs
Mayor-Jellies: We were under the impression that LLVM would already be a requirement for Halide. bindgen is used here to create an iir_blur.rs file with a single iir_blur(buffer_t, float, buffer_t) extern function. This matches the function in iir_blur.h while we could write a function that takes a name and generates an extern function .rs in the same format it would not be able to deal with other function signatures very well. If for example, the generator required an int, 2 floats, or an additional buffer.

language_bindings/rust/rust_build/test_files/iir_blur.cpp
```diff
@@ -0,0 +1,178 @@
+// This file defines a generator for a first order IIR low pass filter
+// for a 2D image.
```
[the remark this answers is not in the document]
Mayor-Jellies: It is a modified version accepting interleaved data. We could modify the example code to reformat to planner then the iir_blur generators would be the same. It may be better for demonstration to use another generator for this purpose IE a "convert_to_planner" the example app would then use 3 generators to_planner->iir_blur->to_interleaved?

language_bindings/rust/rust_build/src/lib.rs
Ralith: I'm not familiar with the Halide workflow. Can you clarify for me why we're trying to compile an executable binary here? I thought the goal was to compile a static library that downstream Rust code can link to. Could the binary be instead be obtained from a precompiled Halide installation?

language_bindings/rust/rust_build/test_files/iir_blur.cpp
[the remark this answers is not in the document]
Ralith: So these are intended to be independent examples, then? Is it useful to have multiple examples for this Rust layer at all? I'd expect the general Halide documentation/examples to be the primary resource a user will draw upon, with a Rust-specific example only necessary to illustrate the Rust portions of the workflow, which I wouldn't expect to change significantly between examples.

language_bindings/rust/rust_build/src/lib.rs
Ralith: > We were under the impression that LLVM would already be a requirement for Halide

Even if LLVM is a dependency of Halide, that does not necessarily imply that the user has a compatible version of libclang installed somewhere that bindgen can find it. LLVM versions aren't all compatible, LLVM doesn't necessarily imply libclang, and It may even be possible to have Halide available with no independent LLVM on the system at all, if Halide statically links it.

> while we could write a function that takes a name and generates an extern function .rs in the same format it would not be able to deal with other function signatures very well. If for example, the generator required an int, 2 floats, or an additional buffer.

Because this is a single function with a very predictable signature, it may be reasonable to ask users to write the FFI declaration themselves. It would take about as much space as the `include!` invocation, and make the process overall much simpler and lighter. You could even provide a declarative macro that generates both the FFI declaration and a safe wrapper.

steven-johnson:
This PR hasn't gotten much attention lately, is it still active?

PR #6729, opened 2022-04-25
Prefer accessing Generator's target, auto_schedule, and machine_params...

steven-johnson:
...via the existing getter methods, rather than the direct fieldnames.

The underlying GeneratorParams implementing them are going to (eventually) go away, in favor of the fields in GeneratorContext; this is a first step in cleaning up the code.

### DIFF
--- a/apps/HelloPyTorch/src/add_generator.cpp
+++ b/apps/HelloPyTorch/src/add_generator.cpp
@@ -30,7 +30,7 @@ public:
         output.set_estimates({{0, kEdge}, {0, kEdge}, {0, kEdge}, {0, kEdge}});
 
         // Schedule
-        if (!auto_schedule) {
+        if (!get_auto_schedule()) {
             Var tx("tx"), xy("xy"), cn("cn"), allvars("allvars");
             if (get_target().has_gpu_feature()) {
                 output
@@ -84,7 +84,7 @@ public:
         d_input_b.set_estimates({{0, kEdge}, {0, kEdge}, {0, kEdge}, {0, kEdge}});
 
         // Schedule
-        if (!auto_schedule) {
+        if (!get_auto_schedule()) {
             Var tx("tx"), xy("xy"), cn("cn"), allvars("allvars");
 
             if (get_target().has_gpu_feature()) {
--- a/apps/bgu/bgu_generator.cpp
+++ b/apps/bgu/bgu_generator.cpp
@@ -430,7 +430,7 @@ public:
             b(2, 2) += weighted_lambda * gain;
 
             // Now solve Ax = b
-            Matrix<3, 4> result = transpose(solve_symmetric(A, b, line, x, auto_schedule, get_target()));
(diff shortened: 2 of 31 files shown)


steven-johnson:
Monday morning review ping. I'd like to consider getting this reviewed (even if we don't move forward with #6731) because it is likely to mirror the way Python Generators work and thus make for a bit more symmetry between the two.

steven-johnson:
PTAL

abadams:
My current understanding of this is that there are implementation reasons that these aren't just plain old generator params in the generator base class. IIRC the separate context object exists so that it can be chained through to other generators when doing forms of generator composition. But to the user, it sure seems like a cleaner abstraction if they were just plain old generator params instead of something special. This explains my hesitation. Putting them behind getters doesn't seem like an improvement to the code that uses them. 

Something else I'm unclear on is the status of the "get_" prefix here. I thought that was removed from at least one of these classes.

steven-johnson:
> My current understanding of this is that there are implementation reasons that these aren't just plain old generator params in the generator base class. IIRC the separate context object exists so that it can be chained through to other generators when doing forms of generator composition. 

Yep.

> But to the user, it sure seems like a cleaner abstraction if they were just plain old generator params instead of something special. 

You have a fair point -- this is an unfortunate side-effect of code changing over time as we add more abilities. The concept of `GeneratorContext` was basically "the GeneratorParam-like things that really need to be set to the same value across all code being generated at a given time" -- i.e., if you compile a Generator that uses sub-Generators (e.g. via Stub or yet-to-land approaches), you normally want to be pretty sure that things like Target and MachineParam are in sync (or, if not, you are making that change *very* deliberately and not at all by accident). Making a new class to encapsulate this seemed wise (especially since it means that if we need to add more info about what gets passed around, it can put there without having to change calling convention anywhere). 

Maybe this concept was misguided and should never have been adopted in the first place -- we could presumably back it out and get rid of GeneratorContext entirely, and make all of these Plain Old GeneratorParams -- though I'd have to do some reality checking to ensure there aren't any hidden subtleties in (e.g.) GeneratorStubs that would bite us.

In any event, we clearly need to have a broad consensus on the right way to handle all this going forward -- the original impetus here was to try to clean up various ill-defined things in Generator that came to the forefront as part of the work on Python Generators (e.g., should GeneratorParams be mutable after Generator creation). So that's probably job #1.
 
> Something else I'm unclear on is the status of the "get_" prefix here. I thought that was removed from at least one of these classes.

They were removed from `GeneratorContext`. They aren't removed here (yet) because we'd have a name conflict with `GeneratorParam target` and `Target target()` both wanting to be member names -- the former would have to go to get rid of the latter, so it would be a staged removal.

steven-johnson:
(BTW, the impetus for doing this is that it affects the design of AbstractGenerator, which, in turn, affects the design of Python Generators. I'd really like to land the former, as it's getting more painful to maintain as a fork as time goes by; if we want to backtrack on GeneratorContext then I could live with it, but we should make a decision.)

PR #6731, opened 2022-04-25
Deprecate the direct use of `target`, `auto_schedule`, `machine_params`

steven-johnson:
These GeneratorParams have been special-cased ~forever, and really need to be a part of GeneratorContext rather than their own GeneratorParams. This defaults to making them vanish entirely; code can opt-in to continue using it (with a deprecation warning) by defining `HALIDE_ALLOW_LEGACY_GENERATOR_PARAMS` at compiletime (similar to how we handled deprecation of the `build()` method last version).

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1261,6 +1261,19 @@ std::vector<std::string> GeneratorRegistry::enumerate() {
     return result;
 }
 
+#ifdef HALIDE_ALLOW_LEGACY_GENERATOR_PARAMS
+
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
+#endif  // HALIDE_ALLOW_LEGACY_GENERATOR_PARAMS
+
 GeneratorBase::GeneratorBase(size_t size, const void *introspection_helper)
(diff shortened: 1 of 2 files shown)


abadams:
Could you expand on why it's good for these to not be GeneratorParams? E.g. I find it very useful to be able to modify the target in a configure() method, as if it were any other GeneratorParam (e.g. to turn on Profile temporarily for a single generator without having to mess with the build target, which can have unwanted side-effects depending on the build system setup, like messing with paths).

abadams:
It works because they're GeneratorParams, and you can modify GeneratorParams using other GeneratorParams in configure(). Surely that's WAI. Why do we want them to be anything other than GeneratorParams?

steven-johnson:
> you can modify GeneratorParams using other GeneratorParams in configure()

...I guess that happens to work, but it's not supposed to, and is explicitly documented otherwise. From Generator.h:

```
You can dynamically add Inputs and Outputs to your Generator via adding a
 configure() method; if present, it will be called before generate(). It can
 examine GeneratorParams but it may not examine predeclared Inputs or Outputs;
 the only thing it should do is call add_input<>() and/or add_output<>(), or call
 set_type()/set_dimensions()/set_array_size() on an Input or Output with an unspecified type.
 ```

steven-johnson:
The goal of *all* GeneratorParams was to be immutable from the point of view of the Generator.

steven-johnson:
It looks like Hyrum's Law has bitten me here, and the way these GeneratorParams is actually being used is, uh, different from what I had anticipated or intended. Guess I need to think on this some more.

PR #6735, opened 2022-04-27
More typed-Func work

steven-johnson:
- Allow Func output_type(s)(), outputs(), dimensions(), and output_buffer(s)() to be called on undefined Funcs if the Func has required_type and required_dimensions specified. This allows for greater flexibility in defining pipelines in which you may want to set or examine constraints on a Func that hasn't been defined yet; previously this required restructuring code or other awkwardness.
- Ensure that the Funcs that are defined for ImageParams and Generator fields define the types-and-dims when known.
- Add some tests.

### DIFF
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -219,11 +219,9 @@ const std::vector<Type> &Func::output_types() const {
 /** Get the number of outputs this function has. */
 int Func::outputs() const {
     const auto &types = defined() ? func.output_types() : func.required_types();
-    // For backwards compatibility, we want to return 0 if the Func
-    // has no type requirements.
-    if (types.empty()) {
-        return 0;
-    }
+    user_assert(!types.empty())
+        << "Can't call Func::outputs on Func \"" << name()
+        << "\" because it is undefined or has no type requirements.\n";
     return (int)types.size();
 }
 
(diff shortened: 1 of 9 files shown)


src/Func.cpp
```diff
@@ -197,23 +197,34 @@ void Func::define_extern(const std::string &function_name,
...
+        << "Can't call Func::output_type on Func \"" << name()
+        << "\" because it is undefined or has no type requirements.\n";
+    return types;
 }
 
 /** Get the number of outputs this function has. */
 int Func::outputs() const {
-    return func.outputs();
+    const auto &types = defined() ? func.output_types() : func.required_types();
+    // For backwards compatibility, we want to return 0 if the Func
```
abadams, line 222: Is there existing code that depends on this?

src/Function.h
```diff
@@ -312,6 +316,10 @@ class Function {
     /** If the Function has dimension requirements, check that the given argument
      * is compatible with them. If not, assert-fail. (If there are no dimension requirements, do nothing.) */
     void check_dims(int dims) const;
+
+    /** Define the output buffers. If the Function has types specified, this can be called at
+     * any time. If not, it can only be called for a Function with a pure definition. */
+    void define_output_buffers(const std::vector<Type> &types, int dims) const;
```
abadams, line 322: "define" is a weird verb here. Would "create" be better?

src/Func.cpp
steven-johnson: Not that I know of, but Hyrum's Law exists

src/Function.h
steven-johnson: done

src/Func.cpp
steven-johnson: (I mean, we can make it an error case and see what breaks...)

abadams: I'd make it an error. Checking the number of outputs of an undefined untyped Func seems like something that should throw an error.

steven-johnson: ok. Should the same be true for `dimensions()`?

abadams: I think so, yes.

steven-johnson: done and done

abadams (approved):

PR #6736, opened 2022-04-27
Add a CTAD for PrintTypeList

steven-johnson:
This smells like a compiler bug to me, because I don't see any ambiguity, but this is needed to eliminate a downstream issue, and should be harmless to conforming C++17 compilers.

### DIFF
--- a/src/Function.cpp
+++ b/src/Function.cpp
@@ -362,8 +362,13 @@ struct PrintTypeList {
     }
 };
 
-// Add an explicit CTAD here just in case we are compiling in an environment with a
-// warning like `-Wctad-maybe-unsupported` enabled.
+// Add an explicit (user-defined) deduction guide here, just in case we are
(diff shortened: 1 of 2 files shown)


src/Function.cpp
```diff
@@ -362,6 +362,11 @@ struct PrintTypeList {
     }
 };
 
+// Add an explicit CTAD here just in case we are compiling in an environment with a
```
zvookin, line 365: Can ```CTAD``` be spelled out? I have no idea what this means.

steven-johnson: Done

steven-johnson: (I'm glad I'm not the only one who had no idea what it meant...)

abadams (approved):

steven-johnson:
I'm closing this because this warning is stupid and I'm just gonna disable it downstream where I first saw it

PR #6742, opened 2022-04-30
Revise PyStub calling convention for GeneratorParams

steven-johnson:
This is a rethink of https://github.com/halide/Halide/pull/6661, trying to make it saner in anticipation of the ongoing Python Generator work.

TL;DR: instead of mixing GeneratorParams in with the rest of the keywords, segregate them into an optional `generator_params` keyword argument, which is a plain Python dict. This neatly solves a couple of problems:
- synthetic params with funky names aren't a problem anymore.
- error reporting is simpler because before an unknown keyword could have been intended to be a GP or an Input.
- GP values are now clear and distinct from Inputs, which is IMHO a good thing.

This is technically a breaking change, but I doubt anyone will notice; this is mainly here to get a sane convention in place for use with Python Generators as well.

Also, a drive-by change to Func::output_types() to fix the assertion error message.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -79,6 +79,11 @@ bool is_valid_name(const std::string &n) {
             return false;
         }
     }
+    // prohibit this specific string so that we can use it for
+    // passing GeneratorParams in Python.
+    if (n == "generator_params") {
+        return false;
+    }
     return true;
 }
 
--- a/python_bindings/Makefile
+++ b/python_bindings/Makefile
@@ -151,7 +151,8 @@ GENPARAMS_complex=\
 	array_input.type=uint8 \
 	int_arg.size=2 \
 	simple_input.type=uint8 \
-	untyped_buffer_input.type=uint8
+	untyped_buffer_input.type=uint8 \
+	untyped_buffer_output.type=uint8
 
 GENPARAMS_simple=\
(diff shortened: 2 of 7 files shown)


steven-johnson:
Monday Morning Review Ping

src/Generator.cpp
```diff
@@ -79,6 +79,11 @@ bool is_valid_name(const std::string &n) {
             return false;
         }
     }
+    // prohibit this specific string so that we can use it for
+    // passing GeneratorParams in Python.
+    if (n == "generator_params") {
```
steven-johnson, line 84: Alternately: instead of forbidding this name, we could make the optional argument to the call be `_generator_params` (ie, start with an underscore), since it has always been forbidden to start a Generator param, input, or output with an underscore (and thus no possibility of name collision). Not sure if this is considered to be a good thing in Python or not though?

alexreinking: I prefer forbidding this name. Named arguments in Python rarely (never?) start with an underscore.

steven-johnson: Agreed, though `namedtuple` does make liberal use of starts-with-underscore names for various methods and fields that are definitely intended to be public -- wasn't sure if a similar exception might be appropriate here.

abadams (approved):

PR #6746, opened 2022-05-04
Define a Generator framework in Python

steven-johnson:
Work in progress, opening as draft for testing + sharing. More details + documentation to come soon.


python_bindings/correctness/pystub.py
```diff
@@ -14,172 +17,165 @@ def _realize_and_check(f, offset = 0):
...
-            vectorize=True)
+            int_arg=int_arg,
+            extra_input=constant_image_u16,
+            vectorize=True,
+            # We can put GeneratorParams anywhere in the list we want --
+            # they will be examined and applied before anything else --
+            # but it's usually better form to put them all at the end.
+            #
+            # We can specify a halide Type via string or object here
+            simple_input__type=hl.UInt(8),
```
steven-johnson, line 165: Note, the stuff regarding "__" here is out of date and will get updated after https://github.com/halide/Halide/pull/6742 lands (assuming it gets full approval)

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -148,15 +148,24 @@ function(add_halide_library TARGET)
...
-        # FROM is usually an unqualified name; if we are crosscompiling, we might need a
-        # fully-qualified name, so add the default package name and retry
-        set(FQ_ARG_FROM "${PROJECT_NAME}::halide_generators::${ARG_FROM}")
-        if (NOT TARGET ${FQ_ARG_FROM})
-            message(FATAL_ERROR "Unable to locate FROM as either ${ARG_FROM} or ${FQ_ARG_FROM}")
-        endif ()
-        set(ARG_FROM "${FQ_ARG_FROM}")
-    endif ()
+    if(ARG_FROM MATCHES ".py$")
+        make_shell_path(PYTHONPATH "$<TARGET_FILE_DIR:Halide::Python>")
```
alexreinking, line 152: I'm not sure if we're installing our `make_shell_path` helper. If not, this will be broken after install. It's probably broken anyway if this module doesn't `include(MakeShellPath)`.

alexreinking: I think we should add an integration test here that makes sure the installed package works for a simple Python generator.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -148,15 +148,24 @@ function(add_halide_library TARGET)
...
+        if (NOT TARGET ${ARG_FROM})
+            # FROM is usually an unqualified name; if we are crosscompiling, we might need a
+            # fully-qualified name, so add the default package name and retry
+            set(FQ_ARG_FROM "${PROJECT_NAME}::halide_generators::${ARG_FROM}")
+            if (NOT TARGET ${FQ_ARG_FROM})
+                message(FATAL_ERROR "Unable to locate FROM as either ${ARG_FROM} or ${FQ_ARG_FROM}")
+            endif ()
+            set(ARG_FROM "${FQ_ARG_FROM}")
+        endif()
+        set(GENERATOR_CMD "$<TARGET_FILE:${ARG_FROM}>")
```
alexreinking, line 166: `$<TARGET_FILE>` shouldn't be necessary here since the target itself will be the first argument to `add_custom_command`'s `COMMAND` argument. It already handles CMake targets correctly in this case.

steven-johnson: I'm pretty sure I tried that and it didn't work, but... I'll try again

cmake/HalideGeneratorHelpers.cmake
steven-johnson: > our make_shell_path helper

...I didn't realize that was a helper, I assumed it was a CMake builtin :-/

> an integration test

not sure I understand; how is that different from what we're already doing in the Python tests?

python_bindings/src/builtin_helpers_src.py
```diff
@@ -0,0 +1,787 @@
...
+def _make_freezable(cls: type):
+
+    # TODO: kinda janky, is this the best way to implement this?
+    def _freeze(self):
+        setattr(self, "_frozen", True)
+
+    def _freezable_setattr(self, name, value):
+        if getattr(self, "_frozen", False):
+            raise AttributeError("Invalid write to field '%s'" % name)
+        self.__class__._old_halide_setattr(self, name, value)
```
alexreinking, line 354: Are we concerned about users setting their own `__setattr__` hooks?

python_bindings/src/builtin_helpers_src.py
```diff
@@ -0,0 +1,787 @@
...
+    cls = _python_generators.get(name, None)
+    if not isclass(cls):
+        cls = None
+    return cls
+
+
+def alias(**kwargs):
+
+    def alias_impl(cls):
+        nonlocal kwargs
```
alexreinking, line 744: This isn't necessary since you aren't writing to `kwargs`. It's already captured read-only, enough for calling `.items()`.

python_bindings/src/builtin_helpers_src.py
steven-johnson: Uh, to be honest... I'm not sure. I just assumed that was the appropriate thing to do, since we're injecting a change into a possibly-arbitrary class. Is it more common to just overwrite?

python_bindings/src/builtin_helpers_src.py
```diff
@@ -0,0 +1,787 @@
...
+
+
+def generator(name=""):
+    # This code relies on dicts preserving key-insertion order, which is only
+    # guaranteed for all Python implementations as of v3.7.
+    assert sys.version_info >= (
+        3, 7), "Halide Generators require Python 3.7 or later."
+
+    def generator_impl(cls):
+        nonlocal name
```
alexreinking, line 772: Again, I don't think you need this. It just keeps the call frame for `generator` alive longer.

python_bindings/src/builtin_helpers_src.py
```diff
@@ -0,0 +1,787 @@
...
+        if not name:
+            name = _fqname(cls)
```
alexreinking, line 774: Because of nonlocal, this writes to generator's frame, not to generator_impl's. I don't see a good reason for this.

python_bindings/src/builtin_helpers_src.py
steven-johnson: At one point this code *did* need it, though maybe it doesn't now. (We do write to name on the next line.)

(The nonlocal usage in `alias` was just a copy-paste mistake.)

python_bindings/src/builtin_helpers_src.py
steven-johnson: how about replacing with `n = name if name else _fqname(cls)`

python_bindings/src/builtin_helpers_src.py
```diff
@@ -0,0 +1,787 @@
...
+#         if i:
+#             write(" ")
+#         write(arg)
+#     write('\n')
+
+# Everything below here is implicitly in the `halide` package
+
+
+def _fail(msg: str):
+    raise RuntimeError(msg)
```
alexreinking, line 23: It might be polite to use a custom exception type like `GeneratorError` for the sake of metaprogramming systems built on top of this.

cmake/HalideGeneratorHelpers.cmake
alexreinking: > I assumed it was a CMake builtin :-/

IIRC it works around a bug in `$<SHELL_PATH>` that was fixed in 3.18 or something.

alexreinking: > not sure I understand; how is that different from what we're already doing in the Python tests?

The Python tests are run in-tree, not the way the apps work, where they `find_package(Halide)`.

python_bindings/src/builtin_helpers_src.py
alexreinking: `n = name or _fqname(cls)` is equivalent, but it's also totally kosher to re-use `name` as in `name = name or _fqname(cls)`.

python_bindings/src/builtin_helpers_src.py
alexreinking: It's fine, it just looked defensive. I would have just overwritten and used `object.__setattr__`.

python_bindings/src/builtin_helpers_src.py
steven-johnson: Agreed, but I took a quick stab at this and it makes some of the tests hard to get right, since errors thrown by the pybind11 layer (e.g. "cannot cast foo to bad") are always RuntimeError. I'll make this a TODO for now.

steven-johnson: Update: I decided this was fairly doable after more examination -- ended up adding a base `hl.Error` class for the generic bindings, with `hl.GeneratorError` as a subclass of it used by the Generator-specific code.

steven-johnson:
Closing this in favor of https://github.com/halide/Halide/pull/6746

alexreinking:
> Closing this in favor of #6746

I think you mean #6764

steven-johnson:
fixed :-/

PR #6750, opened 2022-05-05
Add a `HalideError` base class to Python bindings

steven-johnson:
Per suggestion from @alexreinking, this remaps all exceptions thrown by the Halide Python bindings to be `halide.HalideError` (or a subclass thereof), rather than plain old `RuntimeError`.

### DIFF
--- a/python_bindings/stub/PyStubImpl.cpp
+++ b/python_bindings/stub/PyStubImpl.cpp
@@ -40,12 +40,14 @@ void halide_python_error(JITUserContext *, const char *msg) {
 }
 
 void halide_python_print(JITUserContext *, const char *msg) {
+    py::gil_scoped_acquire acquire;
     py::print(msg, py::arg("end") = "");
 }
 
 class HalidePythonCompileTimeErrorReporter : public CompileTimeErrorReporter {
 public:
     void warning(const char *msg) override {
+        py::gil_scoped_acquire acquire;
         py::print(msg, py::arg("end") = "");
     }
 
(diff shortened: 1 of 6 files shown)


python_bindings/stub/PyStubImpl.cpp
```diff
@@ -63,6 +65,23 @@ void install_error_handlers(py::module &m) {
     handlers.custom_error = halide_python_error;
     handlers.custom_print = halide_python_print;
     Halide::Internal::JITSharedRuntime::set_default_handlers(handlers);
+
+    static py::object halide_error = py::module_::import("halide").attr("HalideError");
+    if (halide_error.is(py::none())) {
+        throw std::runtime_error("Could not find halide.HalideError");
+    }
+
+    // Make a subclass of HalideError for this stub to throw.
```
alexreinking, line 74: Why? Is it even possible to catch this exception specifically in user code if it's dynamically generated?

alexreinking (approved):
Approved with a question

python_bindings/stub/PyStubImpl.cpp
steven-johnson: Nah, it's mainly because I can't figure out how to get PyBind to actually get and throw an instance of the base class. I'll revisit this and try to actually figure in out...

alexreinking (approved):

python_bindings/stub/PyStubImpl.cpp
alexreinking: Looks good now, thanks!

PR #6757, opened 2022-05-10
AOT `_metadata()` getters should be multitarget-aware

steven-johnson:
Multitarget codegen has worked correctly for years, but calling the generated `_metadata()` function (returning `halide_filter_metadata_t*`) has been subtly wrong: the `target` field always contains the baseline target. This PR proposes to smarten the codegen by using the same `can_use_target` logic for the generated metadata getters, so that the `target` field you inspect always matches the one that was used for the actual filter that will be executed.

(This is, admittedly, a rather niche use case, but if nothing else, it is helpful for some debug and test scenarios, as it gives us a simpler way to verify at runtime that the version we are executing is the one we expect.)

### DIFF
--- a/test/correctness/compile_to_multitarget.cpp
+++ b/test/correctness/compile_to_multitarget.cpp
@@ -5,6 +5,13 @@
 
 using namespace Halide;
 
+// Given a path like /path/to/some/file.ext, return file.ext
+// If the path contains no separators (/ or \), just return it as-is
+std::string leaf_name(const std::string &path) {
+    size_t sep = std::min(path.rfind('/'), path.rfind('\\'));
+    return path.substr(sep == std::string::npos ? 0 : sep + 1);
+}
+
 std::string get_fname(const std::string &base) {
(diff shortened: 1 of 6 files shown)


test/correctness/compile_to_multitarget.cpp
```diff
@@ -205,7 +212,8 @@ void test_compile_to_everything(Func j, bool do_object) {
         // it exists or not --  so just fill in with arbitrary strings.
         return std::unique_ptr<Internal::CompilerLogger>(new Internal::JSONCompilerLogger("generator_name", "function_name", "autoscheduler_name", Target(), "generator_args", false));
     };
-    compile_multitarget(fname, outputs, targets, target_strings, module_producer, compiler_logger_factory);
+    // The first argument to compile_multitarget is *function* name, not filename
+    compile_multitarget(leaf_name(fname), outputs, targets, target_strings, module_producer, compiler_logger_factory);
```
abadams, line 216: If it's a function name, why call leaf_name on it? What's going on here?

PR #6760, opened 2022-05-11
Allow overriding of `Generator::init_from_context()` for debug purposes

steven-johnson:


src/Generator.h
```diff
@@ -3385,7 +3390,27 @@ class GeneratorBase : public NamesInterface {
     GeneratorBase(size_t size, const void *introspection_helper);
     void set_generator_names(const std::string &registered_name, const std::string &stub_name);
 
-    void init_from_context(const Halide::GeneratorContext &context);
+    // Note that it is explicitly legal to override init_from_context(), so that you can (say)
+    // create a modified context with a different Target (eg with features enabled or disabled), but...
+    //
+    // *** WARNING ***
+    //
+    // Modifying the context here can be fraught with subtle hazards, especially when used
```
abadams, line 3398: This is a pretty scary comment. Is there no clean mental model for how this interacts with multi-target? Doesn't multi-target just make N instances of the generator? Ideally any changes to the target here would not affect naming of generated anything. Is that the case?

steven-johnson: Well, no, there isn't, because once we allow mutating, anything is possible. e.g., the multitarget code could carefully be checking to be sure that we only call (say) AVX2 code after we have checked that AVX2 is indeed present on the host... but if you mutate Target to set the AVX2 feature flag unconditionally, bad things are gonna happen.

steven-johnson: Probably worth adding that this scary comment applies 100% to the current status quo for doing this (i.e.: directly mutating the `target` GeneratorParam). The issue (in both cases) is that the multitarget code is implicitly promising that it will check each subtarget for validity at runtime; of course, it can't do so reliably if it can't predict what kind of code is being generated.

If one never uses multitarget codegen, then this concern doesn't apply, of course. But there still remains the issue of where the decisions about where codegen details should be specified. The approach that I wanted with Generator was that the external build system should make ~all of the decisions about what to target, with the Generator only able to read-and-react to this: e.g., adjusting the schedule based on the Target/Features it sees, or (in some cases) simply assert-failing that it can't produce anything useful for the given Target.

For dev/debugging purposes, of course, none of the above is particularly important, and sometimes you just need to say "build this thing with `profile` and `debug` enabled so I can see what the heck is going on", and that's the use case I was attempting to target with this PR and comment -- this is really intended for use in dev/debug situations, when working with build systems in which doing temporary overrides is infeasible without editing the source somehow.

And for the record, my insistence on immutability here isn't theoretical: in the early days of Generator usage at Google, I had to track down misbehavior that turned out to be caused by a Generator mutating its Target. (Sadly I can't remember the exact details of the bug, only that it hadn't occurred to me that someone would do such a thing. In hindsight, I should have tried to insist on immutability back then.)

abadams: There should be some sort of contract here that says "You can do these sorts of mutations here and this is what will happen if you do", rather than just saying "here be dragons". I imagining a sole generator out of hundreds that benefits from enable_llvm_loop_opt. You might want to leave that on in production and it would be good to state how that sort of thing interacts with multi-target in the comment.

steven-johnson: Fair, will update

steven-johnson:
Monday Morning Review Ping

abadams (approved):

PR #6762, opened 2022-05-11
Expand the x86 SIMD variants tested in correctness_vector_reductions

steven-johnson:
A recent bug in LLVM codegen was missed because it only affected x86 architectures with earlier-than-AVX2 SIMD enabled; it didn't show up for AVX2 or later. This revamps correctness_vector_reductions to re-run multiple times when multiple SIMD architectures are available on x86 systems.

(correctness_vector_reductions was chosen here because it reliably demonstrated the specific failures in this case.)

### DIFF
--- a/test/correctness/vector_reductions.cpp
+++ b/test/correctness/vector_reductions.cpp
@@ -2,7 +2,11 @@
 
 using namespace Halide;
 
-int main(int argc, char **argv) {
+namespace {
+
+void run_test(const Target &target) {
+    std::cout << "vector_reductions: Testing with " << target << "\n";
+
     for (int dst_lanes : {1, 3}) {
         for (int reduce_factor : {2, 3, 4}) {
             std::vector<Type> types =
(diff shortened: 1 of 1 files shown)


abadams:
We used to run the full correctness tests with HL_JIT_TARGET=x86-64-linux-sse41 or whatever in addition to the host target. I guess we stopped doing that? Should we revive it?

steven-johnson:
> We used to run the full correctness tests with HL_JIT_TARGET=x86-64-linux-sse41 or whatever in addition to the host target.

No, we are still doing that -- in this particular case, we didn't catch the error because our nightly LLVM builds on the x86 machines were made just prior to the commit that injected the erroneous behavior (and so this change actually wouldn't have caught it in our tests).

That said, this change still has value: partly because it adds more robustness to this test, but (mostly) because I want this behavior when running this test inside Google: the tests are run on a heterogenous set of machines (mostly AVX2-class, some less, some more), and LLVM changes are pulled in on a hard-to-predict schedule; if this version of this test had landed inside Google it would have caught this failure much earlier.

(Yes, I thought about making a specific-to-Google version of this test, but then I decided that the changes were modest and strictly a win everywhere.)

steven-johnson:
Gentle review ping

alinas (approved):
LGTM.

abadams:
I'm still a bit confused about this - does it actually add any test coverage to our CI, or is this redundant with running the test under multiple values of HL_JIT_TARGET?

steven-johnson:
It adds coverage in a way that doesn't require customizing multiple different test scripts.

PR #6764, opened 2022-05-12
Define a Generator framework in Python

steven-johnson:
Work in progress, opening as draft for testing + sharing. More details + documentation to come soon.


python_bindings/src/PyGenerator.cpp
```diff
@@ -0,0 +1,193 @@
...
+            .def("machine_params", &GeneratorContext::machine_params)
+            // TODO: handle get_externs_map() someday?
+            .def("__repr__", [](const GeneratorContext &context) -> std::string {
+                std::ostringstream o;
+                o << "<halide.GeneratorContext " << context.target() << ">";
+                return o.str();
+            });
+
+    py::object scope = m.attr("__dict__");
+    py::exec(py::str((const char *)builtin_helpers_src, builtin_helpers_src_length), scope);
```
alexreinking, line 168: After #6815 and the `src/` rearrangement, I don't think we'll need or want to do this binary2cpp / py::exec dance.

steven-johnson: Cool, I eagerly await :-)

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -400,11 +419,12 @@ function(_Halide_add_halide_runtime RT)
     endif ()
 
     add_custom_command(OUTPUT ${GEN_OUTS}
-                       COMMAND ${ARG_FROM} -r "${TARGET}.runtime" -o . ${GEN_ARGS}
+                       COMMAND ${ARG_GENERATOR_CMD} -r "${TARGET}.runtime" -o . ${GEN_ARGS}
                        # Defers reading the list of targets for which to generate a common runtime to CMake _generation_ time.
                        # This prevents issues where a lower GCD is required by a later Halide library linking to this runtime.
                        target=$<JOIN:$<TARGET_PROPERTY:${TARGET}.runtime,Halide_RT_TARGETS>,$<COMMA>>
                        DEPENDS "${ARG_FROM}"
+                       DEPENDS "${ARG_DEPENDS}"
```
alexreinking, line 427:
```suggestion
                       DEPENDS "${ARG_FROM}" ${ARG_DEPENDS}
```

Does having two `DEPENDS` clauses work? That makes me nervous.

python_bindings/src/halide/__init__.py
```diff
@@ -22,3 +22,4 @@
 
 from halide_ import *
 from halide_ import _, _1, _2, _3, _4, _5, _6, _7, _8, _9
+from _generator_helpers import *
```
alexreinking, line 25: Do we need this here? This line dumps everything from `_generator_helpers` into our public API. Is that intentional? It makes sense for `halide_`, which is designed to be outward-facing.

alexreinking: It should probably be `from ._generator_helpers` too (note the `.`).

cmake/HalideGeneratorHelpers.cmake
steven-johnson: Oops. Maybe? Willfix.

python_bindings/src/halide/__init__.py
steven-johnson: We need some of it. E.g. `halide.GeneratorParam` which is defined in _generator_helpers but we want as part of our public API. Let me see if I can selectively import instead.

steven-johnson: Using `from ._generator_helpers` didn't work for me -- stuff didn't get imported

steven-johnson: After merge with latest stuff, the `from ._generator_helpers` works properly now

steven-johnson:
At this point, the main thing lacking here is adequate documentation (which I'm still working on); therefore I'm gonna go ahead and mark this as 'ready for review' (with the understanding that the docs will be added before landing) so that proper review can go ahead and start.

steven-johnson:
Lots of new stuff added to `README_python.md`, PTAL

README_python.md
```diff
@@ -1,17 +1,46 @@
...
+<!-- /MarkdownTOC -->
+
 Halide provides Python bindings for most of its public API. Only Python 3.x is
 supported; at this time, Python 3.8 (or higher) is recommended. The Python
 bindings are supported on 64-bit Linux, OSX, and Windows systems.
 
 In addition to the ability to write just-in-time Halide code using Python, you
-can also use existing Halide Generators (written in C++) to produce Python
-extensions that can be used within Python code.
+can write Generators using the Python bindings, which can simplify build-system
```
zvookin, line 37: Make a link from Generators to the appropriate section of this document.

README_python.md
```diff
@@ -170,68 +201,294 @@ essentially zero cost, with storage being shared. Thus, we can usually pass it
...
+
+Originally, Halide only supported writing Generators in C++. In this document,
+we'll use the term "C++ Generator" to mean "Generator written in C++ using the
+classic API", the term "Python Generator" to mean "Generator written in Halide's
+Python bindings", and just plain "Generator" when the discussion is relatively
+neutral with respect to the implementation language/API.
+
+### Writing a Generator in Python
+
+A Python Generator is a class that: - has the `@hl.generator` decorator applied
```
zvookin, line 225: Consider using a bullet list here for readability.

README_python.md
```diff
@@ -170,68 +201,294 @@ essentially zero cost, with storage being shared. Thus, we can usually pass it
...
+expected output(s) of the resulting `Pipeline`. - declares zero or more member
+fields that are initialized with values of `hl.GeneratorParam`, which can be
+used to pass arbitrary information from the build system to the Generator. A
+GeneratorParam can carry a value of type `bool`, `int`, `float`, `str`, or
+`hl.Type`. - declares a `generate()` method that fill in the Halide IR needed to
+define all of the Outputs - optionally declares a `configure()` method to
+dynamically add Inputs or Outputs to the pipeline, based on (e.g.) the values of
+`GeneratorParam` values or other external inputs
+
+Let's look at a fairly simple example:
```
zvookin, line 239: The main thing missing here is a standard guideline for documentation. I'm not sure if this should be a longer comment above the class that describes all the parameters, comments on the parameter lines themselves, or a docstring in the definition of the class, but I am certain one should never write a generator without documentation on what it does and what each of its parameters, inputs, and outputs are.

README_python.md
```diff
@@ -1,17 +1,46 @@
 # Halide Bindings for Python
```
zvookin, line 1: We should consider changing "you" to "one" throughout this document.

README_python.md
steven-johnson: Ouch, it was one once -- the Markdown formatter scrambled it, fixing

README_python.md
steven-johnson: Done

README_python.md
steven-johnson: Good point! I'll prep that as a PR-onto-this-PR, as it may need separate comment.

README_python.md
steven-johnson: I don't disagree, but it's not clear to me what the appropriate phrasing becomes in all such cases -- sometimes it becomes awkwardly passive-voice, whereas the 'you/your' phrasing doesn't (at least to my ear). Do you see this as simply a matter of preferred style, or are there other correctness issues I'm missing?

steven-johnson:
A handful of changes today, including:
-  There wasn't an implementation of `Generator::compile_to_callable()` in Python -- oops! Added it along with a test.
-  Reflecting on feedback from the last Halide meeting, I removed the public `set_generatorparam_value()` methods from Python, and instead require GeneratorParam values to be specified at Generator construction time (via an optional dict argument to the ctor). I think this will better reflect actual usage an instinct; PTAL.
-  Renamed some of the Generators under test/generator to ensure there is clarity as to C++ vs Python in the name of the Generator (when the 'same' one is implemented both ways)

src/Pipeline.cpp
```diff
@@ -665,7 +674,7 @@ Callable Pipeline::compile_to_callable(const std::vector<Argument> &args_in, con
 
     for (const auto &out : outputs) {
         for (Type t : out.output_types()) {
-            args.emplace_back(out.name(), Argument::OutputBuffer, t, out.dimensions(), ArgumentEstimates{});
+            args.emplace_back(strip_uniquifier(out.name()), Argument::OutputBuffer, t, out.dimensions(), ArgumentEstimates{});
```
abadams, line 677: What does this stripped name get used for? It's currently legal to write nonsense like:

```
#include "Halide.h"

using namespace Halide;

    Func f1("f"), f2("f");
    Var x;
    f1(x) = x;
    f2(x) = x + 1;

    Pipeline p({f1, f2});

    p.compile_jit();

    return 0;
}
```

but that still seems to work in this branch

steven-johnson: OK, revisiting this: the issue here is that we are building up a list of Arguments to use for the Callable; for most uses, the name is just "decoration" (e.g. for annotating the function prototype in the header). In this case, though, we're going to use them to populate the list of `kwargs` in the Python wrappers, so that we can do something like `simple(buffer_input=b_in, float_arg=float_in, simple_output=b_out)`. 

The gotcha, of course, is that Outputs in a Generator are essentially Funcs, and the names get uniquified as a result, so if you the same Generator multiple times (or different Generators with matching output names), you'll get outputs with names like `simple_output$4`, which are unusable as kwargs.

This change was really just a quick hack, on the assumption that any Argument we see that ends in `$NUM` is just a uniquified name and that we can trim off that bit to get the 'real' name.

I can revisit the code paths and see if there is way to ensure that the 'original' Output names get transmitted from the Generator all the way down this path, but it would likely take the addition of some new data channel to make that happen reliably, which is why I took the quick way out.

Given the use case here, I suspect that just doing the stripping is safe, with the caveat that you *could* construct a pipeline like this that would be ambiguous to call via kwargs in Python -- but not via a Generator, since they enforce that all inputs and outputs have unique names -- so it would be a case of "don't do that".

README_python.md
```diff
@@ -1,17 +1,47 @@
...
+        - [Calling a Generator Directly](#calling-a-generator-directly)
+        - [The Lifecycle Of A Generator](#the-lifecycle-of-a-generator)
+        - [Notable Differences Between C++ and Python Generators](#notable-differences-between-c-and-python-generators)
+- [Keeping Up To Date](#keeping-up-to-date)
+- [License](#license)
+
+<!-- /MarkdownTOC -->
+
 Halide provides Python bindings for most of its public API. Only Python 3.x is
 supported; at this time, Python 3.8 (or higher) is recommended. The Python
```
alexreinking, line 33: We should reword this into a requirement, I think.

README_python.md
```diff
@@ -1,17 +1,47 @@
...
 In addition to the ability to write just-in-time Halide code using Python, you
-can also use existing Halide Generators (written in C++) to produce Python
-extensions that can be used within Python code.
+can write [Generators](#halide-generators-in-python) using the Python bindings,
+which can simplify build-system integration (since no C++ metacompilation step
+is required).
 
-(Note that the ability to write Halide Generators in Python is *not* supported
-at present, but is expected to be supported soon.)
+You can also use existing Halide Generators (written in C++) to produce Python
```
alexreinking, line 41: > (written in C++)

but not the new ones in Python?

README_python.md
```diff
@@ -62,7 +92,7 @@ with some differences where the C++ idiom is either inappropriate or impossible:
     Buffer<>(Type t, vector<int> extents, string name = "");
     ```
 
-    in Python, only the second variant is provided.
```
alexreinking, line 65: Above, is written "variadic argumentlist" (missing space). GitHub's interface doesn't let me comment too far away from a change, ugh.

alexreinking: Some of the entries below might need revision. For instance, "No mechanism is provided for overriding any runtime functions from Python." probably doesn't apply to the Python generators.

README_python.md
```diff
@@ -170,35 +200,240 @@ essentially zero cost, with storage being shared. Thus, we can usually pass it
 directly to existing Python APIs (like `imsave()`) that expect 'image-like'
 objects without any explicit conversion necessary.
 
-## Using Halide Generators from Python
+## Halide Generators In Python
+
+In Halide, a "Generator" is a unit of encapsulation for Halide code. It is
```
alexreinking, line 205: It is **a**

README_python.md
```diff
@@ -170,35 +200,240 @@ essentially zero cost, with storage being shared. Thus, we can usually pass it
...
+self-contained piece of code that can: - Produce a chunk of Halide IR (in the
```
alexreinking, line 206: Should this be a real list?

README_python.md
alexreinking: I'm happy with this document using the second person. "One" sounds excessively formal to my ear. Comparable documentation from other languages uses similar second person wording, see, e.g. the Rust Book https://doc.rust-lang.org/book/ch03-04-comments.html

README_python.md
```diff
@@ -289,6 +552,201 @@ But if you construct the numpy arrays yourself (like above), you can pass
...
+class LogicalOpGenerator:
+    op = hl.GeneratorParam("xor")
+    with_offset = hl.GeneratorParam(False)
+
+    input = hl.InputBuffer(hl.UInt(8), 2)
+    mask = hl.InputScalar(hl.UInt(8))
+
+    output = hl.OutputBuffer(hl.UInt(8), 2)
+
+    def configure(g):
```
alexreinking, line 606: Do we want to use the `g = self` idiom here? Maybe I missed it, but if we want to advocate for using `g` in place of `self` for that argument name, we should call it out and indicate that it's just for terseness and isn't magic.

README_python.md
```diff
@@ -170,35 +200,240 @@ essentially zero cost, with storage being shared. Thus, we can usually pass it
...
+This will be called by the Generator machinery to build the Pipeline. As with
+C++ Generators, the only required task is to ensure that all Output fields are
+fully defined, in a way that matches the type-and-dimension constraints
+specified.
+
+It is required that the `generate()` method be defined by the Generator.
+
+(Note that, by convention, Halide Generators use `g` instead of `self` in their
+`generate()` method to make the expression language terser; this is not in any
+way required, but is recommended.)
```
alexreinking, line 376: Whoops, I missed this earlier. This needs to be called out more; somewhere higher up than four heading levels deep.

README_python.md
```diff
@@ -170,35 +200,240 @@ essentially zero cost, with storage being shared. Thus, we can usually pass it
...
 `add_halide_library` and `add_halide_python_extension_library()`:
 
 ```
 # Build a Halide library as you usually would, but be sure to include `PYTHON_EXTENSION`
-add_halide_library(my_filter
-                   GENERATOR my_filter_generator
-                   SOURCES my_filter_generator.cpp
+add_halide_library(xor_filter
+                   GENERATOR logical_op_generator
+                   SOURCES logical_op_generator.py
```
alexreinking, line 435: I'm confused by this... shouldn't this just be `xor_filter FROM logical_op_generator.py`?

README_python.md
```diff
@@ -289,6 +552,201 @@ But if you construct the numpy arrays yourself (like above), you can pass
...
+-   `Input<Func>` and `Output<Func>`: these were deliberately left out in order
+    to simplify Python Generators. It's possible that something similar might be
+    added in the future.
+-   GeneratorParams with LoopLevel types: these aren't useful without
+    `Input<Func>`/`Output<Func>`.
+-   GeneratorParams with Enum types: using a plain `str` type in Python is
+    arguably just as easy, if not easier.
+-   `get_externs_map()`: this allows registering ExternalCode objects to be
+    appended to the Generator's code. In our experience, this feature is very
+    rarely used. We consider adding this in the future if necessary.
```
alexreinking, line 738: We will

alexreinking (changes requested):
I still have to look at the main event, `_generator_helpers.py`, but there's plenty to act on now.

src/Pipeline.cpp
steven-johnson: (Gentle ping to @abadams for feedback on my feedback)

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -68,26 +72,47 @@ function(add_halide_generator TARGET)
...
+            if (NOT len EQUAL 1)
+                message(FATAL_ERROR "Python Generators must specify exactly one source file.")
+            endif()
+
+            # Make a fake target here that we can attach the Python source to,
+            # so that we can extract 'em in add_halide_library()
+            add_custom_target(${TARGET} ALL)
+            set_property(TARGET ${TARGET} PROPERTY Halide_PYTHON_GENERATOR_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/${ARG_SOURCES}")
+
+            # TODO: what do we need to do for PACKAGE_NAME PACKAGE_NAMESPACE EXPORT_FILE in this case?
```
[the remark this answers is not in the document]
steven-johnson: Let be more clear: I'm not entirely clear on what these are for when using C++ Generators. So I definitely don't know how to set up for python. Leaving this one open for now, I guess.

python_bindings/src/halide/halide_/PyGenerator.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include <pybind11/embed.h>
+
+namespace Halide {
+namespace PythonBindings {
+
+namespace {
+
+using Halide::Internal::AbstractGenerator;
+using Halide::Internal::AbstractGeneratorPtr;
+#ifdef HALIDE_ALLOW_GENERATOR_EXTERNAL_CODE
```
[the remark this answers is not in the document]
steven-johnson: It's not; see https://github.com/halide/Halide/pull/6844

python_bindings/test/generators/complexpy_generator.py
```diff
@@ -0,0 +1,66 @@
+import halide as hl
+
+x = hl.Var('x')
+y = hl.Var('y')
+c = hl.Var('c')
+
+@hl.generator(name = "complexpy")
+class ComplexPy:
+    vectorize = hl.GeneratorParam(True)
```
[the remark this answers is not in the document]
steven-johnson: Apples and oranges; this call expects a plain Python object. But I see your point.

python_bindings/src/halide/halide_/PyType.h
```diff
@@ -3,6 +3,11 @@
 
 #include "PyHalide.h"
 
+// This is an expedient: a unique Handle type that we use as a placeholder
+// in Python generators. It should never be visible to the end user.
+struct NoneType {};
+HALIDE_DECLARE_EXTERN_STRUCT_TYPE(NoneType);
```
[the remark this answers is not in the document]
steven-johnson: Never visible to the user, but yeah, possible confusing; I'll rename.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -200,17 +224,25 @@ function(add_halide_library TARGET)
...
-            message(FATAL_ERROR "Unable to locate FROM as either ${ARG_FROM} or ${FQ_ARG_FROM}")
-        endif ()
-        set(ARG_FROM "${FQ_ARG_FROM}")
-    endif ()
-
-    _Halide_place_dll(${ARG_FROM})
+    get_property(py_src TARGET ${ARG_FROM} PROPERTY Halide_PYTHON_GENERATOR_SOURCE)
+    if (py_src)
+        set(PYTHONPATH "$<TARGET_FILE_DIR:Halide::Python>/..")
+        set(GENERATOR_CMD ${CMAKE_COMMAND} -E env PYTHONPATH=${PYTHONPATH} ${Python3_EXECUTABLE} $<SHELL_PATH:${py_src}>)
```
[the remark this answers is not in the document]
steven-johnson: (1) How could it be anything else?
(2) How would we detect and/or ensure?

src/Pipeline.cpp
abadams: I'm mostly concerned about the non-python non-Generator case, because as you say name collisions are checked in a Generator. I'm worried that we'll have two sets of out-of-sync names floating around that produce hard-to-debug errors later due to the mismatch if someone does something creative with a Callable, e.g. by using names from Callable::arguments() to go look up into a map of Funcs by name. 

Maybe it would be safer to strip the $* at the point where the kwargs are populated.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -100,11 +125,10 @@ function(add_halide_generator TARGET)
...
                "HALIDE_GENERATOR_PYSTUB(${GEN_NAME}, ${MODULE_NAME})\n")
 
-        file(WRITE
-             "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.py_stub_generated.cpp"
-             "${stub_text}")
+        set(generated_cpp "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.py_stub_generated.cpp")
+        file(WRITE "${generated_cpp}" "${stub_text}")
 
-        Python3_add_library(${TARGET}_pystub MODULE WITH_SOABI "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.py_stub_generated.cpp" ${ARG_SOURCES})
+        Python3_add_library(${TARGET}_pystub MODULE WITH_SOABI "${generated_cpp}" ${ARG_SOURCES})
```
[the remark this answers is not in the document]
steven-johnson: Hm, if building WITH_PYTHON_BINDINGS=OFF then yes, I suppose. But this PR builds just fine that way (several of the buildbots only test that way), probably because we never attempt to use any of the python-related stuff when this is disabled. Not sure if we should wrap this (and other) sections in WITH_PYTHON_BINDINGS checks, vs just let people's code fail to build when they try to use Python with a non-Python Halide.

README_python.md
steven-johnson: Fixed.

README_python.md
steven-johnson: Updated to clarify that this applies to JIT code, and added a section on the state of runtime functions for AOT code.

README_python.md
steven-johnson: Done

README_python.md
steven-johnson: jfc, the markdown formatter I use keeps un-formatting this from a real list, fixed

README_python.md
steven-johnson: Yep, will call this out earlier and more emphatically

README_python.md
steven-johnson: See above

README_python.md
steven-johnson: Edit: after looking at this some I wonder if this is a mistake in our style thoughts -- if you didn't notice the docs on a read-through, other folks won't either, and will wonder WTF is going on. Maybe it would be clearer to have the convention (in our code at least) be

```
    def generate(self):
        g = self  # for terseness in expressions
       ...
```

? 

(I won't make any changes here till I hear your thoughts)

README_python.md
steven-johnson: Yep, fixed

python_bindings/test/generators/complexpy_generator.py
```diff
@@ -0,0 +1,66 @@
...
+    extra_input_name = hl.GeneratorParam("")
+
+    typed_buffer_input = hl.InputBuffer(hl.UInt(8), 3)
```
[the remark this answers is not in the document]
steven-johnson: I think I'm going to try out this (and/or the native-python-type allowances) as a PR-to-this-PR, so we can see what it feels like as an additive change.

python_bindings/test/generators/bitpy_generator.py
```diff
@@ -0,0 +1,19 @@
+import halide as hl
+
+x = hl.Var('x')
+y = hl.Var('y')
+
+@hl.generator(name = "bitpy")
+class BitGenerator:
+    bit_input = hl.InputBuffer(hl.Bool(), 1)
```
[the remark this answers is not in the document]
steven-johnson: I think I'm going to try out this as a PR-to-this-PR, so we can see what it feels like as an additive change.

cmake/HalideGeneratorHelpers.cmake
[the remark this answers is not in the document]
steven-johnson: Actually, yeah, it's definitely possible for out-of-tree builds (eg apps): if they don't explicitly call it, it won't have been called. So the question is: should we do so here lazily? Or (alternately) do something like

`if (NOT TARGET Halide::Python) message(FATAL_ERROR "You must call find_package(Python3) in order to use this rule with Python Generators"`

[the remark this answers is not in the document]
steven-johnson: Resolved (I think) by b789450926c2ca64bad5700bed4faae11be8baa3

src/Pipeline.cpp
steven-johnson: Good call. f613271835688244830b68379de2543fb9bfc59b has the fix, I think. PTAL.

cmake/HalideGeneratorHelpers.cmake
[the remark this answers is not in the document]
alexreinking: Let's not worry about this for _this_ PR, but please create an issue assigned to the two of us along the lines of "Make Python generators work in cross compiling scenarios"

alexreinking (changes requested):
Still reviewing `_generator_helpers.py`, just publishing a couple comments now.

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+            gp = self._gp_dict[name]
+            assert isinstance(gp, GeneratorParam)
+            old_value = gp._value
+            new_value = GeneratorParam._parse_value(name, type(old_value), value)
+            gp._value = new_value
+        elif name == "autoscheduler":
+            _check(not self.autoscheduler().name, "The GeneratorParam %s cannot be set more than once" % name)
+            self.autoscheduler().name = value
+        elif name.startswith("autoscheduler."):
+            sub_key = name[14:]
```
alexreinking, line 687: magic number here...

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+            _python_generators[k] = new_cls
+        return cls
+
+    return alias_impl
+
+
+def generator(name=""):
+    # This code relies on dicts preserving key-insertion order, which is only
+    # guaranteed for all Python implementations as of v3.7.
+    assert sys.version_info >= (3, 7), "Halide Generators require Python 3.7 or later."
```
alexreinking, line 792: Aha. this should not be an `assert`, but a `_check`.

alexreinking (changes requested):
Mostly quibbly stuff. Ping when when resolved and I'll be quick about re-reviewing.

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+from .halide_ import _unique_name, _UnspecifiedType
+from inspect import isclass
+from typing import Any, Optional
+import sys
+
+# Formatted with:
+# yapf --style="{based_on_style: google, column_limit: 120, indent_width: 4}"
+
+
+# 'print()' is consumed by `halide.print()` which is implicitly present;
```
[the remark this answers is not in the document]
steven-johnson: It's part of the Halide API. `Expr print(Expr, ...);`

steven-johnson:
Opened https://github.com/halide/Halide/issues/7014

cmake/HalideGeneratorHelpers.cmake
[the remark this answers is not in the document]
steven-johnson: https://github.com/halide/Halide/issues/7014

python_bindings/src/halide/_generator_helpers.py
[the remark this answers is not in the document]
steven-johnson: (So, no, it can't really be renamed.)

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+def _print(*args):
+
+    def write(data):
+        sys.stdout.write(str(data))
+
+    for i, arg in enumerate(args):
+        if i:
+            write(" ")
+        write(arg)
+    write('\n')
```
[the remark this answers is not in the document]
steven-johnson: I left it in because I kept needing to re-add it for debugging. I'll try the builtins.print approach and if that work I'll use that instead.

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+    # TODO: use regex instead?
+    s = str(name)
+    if "__" in s:
+        return False
+    if not s[0].isalpha():
+        return False
+    for c in s[1:]:
+        if not (c.isalnum() or c == "_"):
+            return False
+    return True
```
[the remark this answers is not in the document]
steven-johnson: Nah, that makes it even more permissive. The goal here is to *exactly* match the rules that C++ Generators use. I'll revise to ensure it doesn't include other Unicode areas (yet).

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+        "The name '%s' is not valid for a GeneratorParam, Input, or Output." % name,
+    )
+    return name
+
+
+# Transmute 'None' into our internal "UnspecifiedType" as a placeholder;
+# also add some syntactic sugar, to allow users to alias
+# bool -> UInt(1), int -> Int(32), float -> Float(32)
+def _sanitize_type(t: object) -> Type:
+    if t is None:
```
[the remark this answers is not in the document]
steven-johnson: Done

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -0,0 +1,806 @@
...
+        return Float(32)
+    else:
+        _check(isinstance(t, Type), "Expected a Halide Type, but saw: %s" % t)
+        return t
+
+def _normalize_type_list(types: object) -> list[Type]:
+    # Always treat _UnspecifiedType as a non-type
+    if types is None:
+        types = []
+    elif isinstance(types, Type) and types == _UnspecifiedType():
```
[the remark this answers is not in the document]
steven-johnson: Because `_UnspecifiedType()` returns a `hl.Type`, not a Python `Type`

python_bindings/src/halide/_generator_helpers.py
steven-johnson: Done

README_python.md
steven-johnson: Any thoughts on this one?

steven-johnson:
PTAL

alexreinking (approved):
This LGTM now, though I understand there are some `pytype` issues?

steven-johnson:
> This LGTM now, though I understand there are some `pytype` issues?

Yeah... the issue is that pytype is the preferred type checker at Google, and it's a *static* type checker. Currently we use the `@generator` decorator to inject `hl.Generator` as a base class, but there is (apparently) no way to indicate this to pytype, with the result being that the type checker doesn't know that it's a subclass, and complains when any inherited methods are called (e.g. `natural_vector_size()`). 

This could be fixed by requiring that user Generators explicitly descend from `hl.Generator` (and having the decorator check that is the case), which is disappointing, as it adds unnecessary redundancy in the Generator declarations, eg..

```
@hl.generator()
class MyGen(hl.Generator):
```

For the sake of argument, we *could* have it both ways: have the decorator verify that it is a subclass, and inject the baseclass only if it isn't, but that has a subtle downside: people coding without pytype would likely omit the explicit base class, but then they'd produce code that would need to be modified when getting imported and used in a pytype environment. This would be a huge nuisance in the long run, with the (likely) result that people end up adding the explicit inheritance always just to dodge complaints.

At this point I'm inclined to just require the explicit inheritance in order to get around this problem, but I'm not happy about it (I find it especially unsatisfying that apparently PyType doesn't have a way to annotate dynamically-generated classes, because that's A Thing That Happens in dynamic languages like Python).

PR #6765, opened 2022-05-13
Fix fundamental confusion about target/tune CPU

LebedevRI:
Sooo. Uh, remember when in https://github.com/halide/Halide/pull/6655
we've agreed that we want to add support to precisely specify
the CPU for which the code should be *tuned* for,
but not *targeted* for. Aka, similar to clang's `-mtune=` option,
that does not affect the ISA set selection?

So guess what, that's not what we did, apparently.
`CodeGen_LLVM::mcpu()` / `halide_mcpu` actually do specify
the *target* CPU. It was obvious in retrospect, because e.g.
`CodeGen_X86::mattrs()` does not, in fact, ever specify `+avx2`,
yet we get AVX2 :) So we've unintentionally added `-march=` support.
Oops.

While i'd like to add `-march=` support, that was not the goal here.

Fixing this is complicated by the fact that
`llvm::Target::createTargetMachine()` only takes `CPU Target` string,
you can't specify `CPU Tune`.

But this is actually a blessing in disguise,
because it allows us to fix another bug at the same time:

There is a problem with halide "compile to llvm ir assembly",
a lot of information from Halide Target is not /really/ lowered
into LLVM Module, but is embedded as a metadata,
that is then extracted by halide `make_target_machine()`.

While that is not a problem in itself, it makes it *impossible*
to dump the LLVM IR, and manually play with it,
because e.g. the CPU [Target] and Attributes (ISA set)
are not actually lowered into the form LLVM understands,
but are in some halide-specific metadata.

So, to fix the first bug, we must lower the CPU Tune
into per-function `"tune-cpu"` metadata,
and while there we might as well lower `"target-cpu"`
and `"target-features"` similarly, which is what clang does:
https://godbolt.org/z/fxnEb6Exr

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -598,6 +598,7 @@ void get_target_options(const llvm::Module &module, llvm::TargetOptions &options
     bool use_pic = true;
     get_md_bool(module.getModuleFlag("halide_use_pic"), use_pic);
 
+    // FIXME: can this be migrated into `set_function_attributes_from_halide_target_options()`?
     bool per_instruction_fast_math_flags = false;
     get_md_bool(module.getModuleFlag("halide_per_instruction_fast_math_flags"), per_instruction_fast_math_flags);
 
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -590,11 +590,9 @@ bool get_md_string(llvm::Metadata *value, std::string &result) {
     return false;
 }
 
-void get_target_options(const llvm::Module &module, llvm::TargetOptions &options, std::string &mcpu, std::string &mattrs) {
+void get_target_options(const llvm::Module &module, llvm::TargetOptions &options) {
     bool use_soft_float_abi = false;
     get_md_bool(module.getModuleFlag("halide_use_soft_float_abi"), use_soft_float_abi);
-    get_md_string(module.getModuleFlag("halide_mcpu"), mcpu);
-    get_md_string(module.getModuleFlag("halide_mattrs"), mattrs);
     std::string mabi;
     get_md_string(module.getModuleFlag("halide_mabi"), mabi);
     bool use_pic = true;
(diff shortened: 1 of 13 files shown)


rootjalex:
I'm very much in support of this - I've run into multiple issues with not being able to actually use the LLVM IR output without first fixing target features (i.e. https://github.com/llvm/llvm-project/issues/55127 )

src/CodeGen_Internal.cpp
```diff
@@ -590,11 +590,9 @@ bool get_md_string(llvm::Metadata *value, std::string &result) {
     return false;
 }
 
-void get_target_options(const llvm::Module &module, llvm::TargetOptions &options, std::string &mcpu, std::string &mattrs) {
+void get_target_options(const llvm::Module &module, llvm::TargetOptions &options) {
     bool use_soft_float_abi = false;
     get_md_bool(module.getModuleFlag("halide_use_soft_float_abi"), use_soft_float_abi);
```
LebedevRI, line 595: To explicitly call this out: yes, none of this should be done here, ideally,
as much as possibly should be ported into `set_function_attributes_from_halide_target_options()`.

src/CodeGen_LLVM.h
```diff
@@ -106,11 +106,12 @@ class CodeGen_LLVM : public IRVisitor {
...
 
-    /** What should be passed as -mcpu, -mattrs, and related for
-     * compilation. The architecture-specific code generator should
-     * define these. */
+    /** What should be passed as -mcpu (warning: implies attrs!), -mattrs,
+     *  and related for compilation. The architecture-specific code generator
+     *  should define these. */
     // @{
-    virtual std::string mcpu() const = 0;
+    virtual std::string mcpu_target() const = 0;
```
steven-johnson, line 113: Possibly-dumb question: would it make sense to refactor all of these into a single virtual method returning a struct? How common is it to override only one of target/tune/attrs without affecting the others?

LebedevRI: No idea, you are asking wrong person :)
The obvious problem is that when a new field is added,
instead of loudly failing to compile at every place that needs to be updated,
you get some default garbage. So i wouldn't do that.

src/CodeGen_LLVM.h
```diff
@@ -106,11 +106,12 @@ class CodeGen_LLVM : public IRVisitor {
     virtual void end_func(const std::vector<LoweredArgument> &args);
     // @}
 
-    /** What should be passed as -mcpu, -mattrs, and related for
-     * compilation. The architecture-specific code generator should
-     * define these. */
+    /** What should be passed as -mcpu (warning: implies attrs!), -mattrs,
```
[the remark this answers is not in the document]
LebedevRI: Done, i think.

src/CodeGen_Internal.cpp
LebedevRI: I've filed https://github.com/halide/Halide/issues/6769 for now.

LebedevRI:
Looks like bots are happy now :)
If you want me to make other changes discussed here, please just state so.

src/CodeGen_LLVM.h
```diff
@@ -106,11 +106,21 @@ class CodeGen_LLVM : public IRVisitor {
...
+     *  and related for compilation. The architecture-specific code generator
+     *  should define these.
+     *
+     *  `mcpu_target()` - target this specific CPU, in the sense of the allowed
+     *  ISA sets *and* the CPU-specific tuning/assembly instruction scheduling.
+     *
+     *  `mcpu_tune()` - expect that we will be running on this specific CPU,
+     *  so perform CPU-specific tuning/assembly instruction scheduling, *but*
+     *  DON'T sacrifice the portability, support running on other CPU's, only
```
steven-johnson, line 118: grammar nit: "CPUs" (no apostrophe)

steven-johnson:
LGTM. Adding @abadams for completeness.

abadams (approved):

LebedevRI:
Thank you!

Issue #6763, opened 2022-05-12
Fusing a specialization drops compute_withs from generated code


JoelLinn:
The following pseudo code

```c++
Func out0, out1;
out0(x,y) = 1 * in(x,y);
out1(x,y) = 2 * in(x,y);

out0.vectorize(x, vs);
out1.vectorize(x, vs).compute_with(out0, x);

out0.specialize(in.dim(1).stride() == 128).fuse(x,y,f);
```

will generate code where out1 is never written for the specialized case, only out0 is.
This should instead either give a generation time error or ideally fuse the compute_with function(s) as well.

JoelLinn:
Thanks for quick fix.
But how would I fuse out0 and out1 for the specialization only?
If I am not mistaken, the individual specialization will fail for compute_with functions.
Can I retrieve the out1 Stage that belongs to the specialization from the out0 specialized staged somehow?

PR #6770, opened 2022-05-17
Update the list of fused_pairs and run validate_fused_group for specalization definitions too

vksnk:
Fixes https://github.com/halide/Halide/issues/6763.

From the issue above, the following program will produce incorrect code inside of the specializaiton branch:

```
Func out0, out1;
out0(x,y) = 1 * in(x,y);
out1(x,y) = 2 * in(x,y);

out0.vectorize(x, vs);
out1.vectorize(x, vs).compute_with(out0, x);

out0.specialize(in.dim(1).stride() == 128).fuse(x,y,f);
```

instead this should be an error just like in the case of the same code, but without specialization: 

```
Func out0, out1;
out0(x,y) = 1 * in(x,y);
out1(x,y) = 2 * in(x,y);

out0.vectorize(x, vs);
out1.vectorize(x, vs).compute_with(out0, x);

out0.fuse(x,y,f);
```
which will produce a following error:

```
terminate called after throwing an instance of 'Halide::CompileError'
  what():  Error: Invalid compute_with: cannot find x in out0.s0
```

After this PR the same error will be produced for the case with specialization branch.

### DIFF
--- a/src/RealizationOrder.cpp
+++ b/src/RealizationOrder.cpp
@@ -163,14 +163,15 @@ void populate_fused_pairs_list(const string &func, const Definition &def,
                    func, stage_index, fuse_level.var().name());
     if (fuse_level.stage_index() == 0) {
         parent.definition().schedule().fused_pairs().push_back(pair);
-        for (int i = 0; i < parent.definition().specializations().size(); i++) {
-            parent.definition().specializations()[i].definition.schedule().fused_pairs().push_back(pair);
+        for (auto &s : parent.definition().specializations()) {
+            s.definition.schedule().fused_pairs().push_back(pair);
         }
     } else {
         internal_assert(fuse_level.stage_index() > 0);
-        parent.update(fuse_level.stage_index() - 1).schedule().fused_pairs().push_back(pair);
-        for (int i = 0; i < parent.update(fuse_level.stage_index() - 1).specializations().size(); i++) {
-            parent.update(fuse_level.stage_index() - 1).specializations()[i].definition.schedule().fused_pairs().push_back(pair);
+        auto &fuse_stage = parent.update(fuse_level.stage_index() - 1);
+        fuse_stage.schedule().fused_pairs().push_back(pair);
+        for (auto &s : fuse_stage.specializations()) {
+            s.definition.schedule().fused_pairs().push_back(pair);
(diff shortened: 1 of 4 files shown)


src/RealizationOrder.cpp
```diff
@@ -163,9 +163,15 @@ void populate_fused_pairs_list(const string &func, const Definition &def,
                    func, stage_index, fuse_level.var().name());
     if (fuse_level.stage_index() == 0) {
         parent.definition().schedule().fused_pairs().push_back(pair);
+        for (int i = 0; i < parent.definition().specializations().size(); i++) {
```
steven-johnson, line 166: style nit: `for (const auto s : parent.definition().specializations())` might be a better choice here

src/ScheduleFunctions.cpp
```diff
@@ -2369,9 +2369,17 @@ void validate_fused_groups_schedule(const vector<vector<string>> &fused_groups,
 
             validate_fused_group_schedule_helper(
                 iter->first, 0, iter->second.definition(), env);
+            for (size_t i = 0; i < iter->second.definition().specializations().size(); i++) {
```
steven-johnson, line 2372: See similar nits above

steven-johnson (approved):
LGTM with nits

src/RealizationOrder.cpp
vksnk: Done

src/RealizationOrder.cpp
```diff
@@ -163,9 +163,15 @@ void populate_fused_pairs_list(const string &func, const Definition &def,
...
+            parent.definition().specializations()[i].definition.schedule().fused_pairs().push_back(pair);
+        }
     } else {
         internal_assert(fuse_level.stage_index() > 0);
         parent.update(fuse_level.stage_index() - 1).schedule().fused_pairs().push_back(pair);
+        for (int i = 0; i < parent.update(fuse_level.stage_index() - 1).specializations().size(); i++) {
```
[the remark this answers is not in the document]
vksnk: Done

src/ScheduleFunctions.cpp
vksnk: Done

PR #6771, opened 2022-05-18
Add execute_generator() API

steven-johnson:
This refactors the existing `generate_filter_main()` call in two, moving the interesting implementation of how to drive AOT into the new `execute_generator()` call (reducing `generate_filter_main()` to parsing argc/argv and error reporting).

The new `execute_generator()` is intended to be used (eventually) from Python, as a way to drive Generator compilation from a Python script more easily. The PR doesn't provide a Python wrapper for this call yet (that will come in a subsequent PR).

Also, a drive-by removal of the "error_output" arg to generate_filter_main() -- AFAICT, no one has ever used it for anything but stderr, and the refactoring now just directs all errors to `user_error` uniformly.

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -87,15 +87,6 @@ bool is_valid_name(const std::string &n) {
     return true;
 }
 
-std::string compute_base_path(const std::string &output_dir,
-                              const std::string &function_name,
-                              const std::string &file_base_name) {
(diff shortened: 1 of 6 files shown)


steven-johnson:
Monday Morning Review Ping

abadams (approved):

PR #6772, opened 2022-05-19
Add Func::type()/types(), deprecate Func::output_type()/output_types()

steven-johnson:
Almost everything in Halide IR that has a getter for "what is your Halide::Type" uses `type()` for the method name; Func is the lone exception, using `output_type()` and `output_types()` (since Tuple-valued Funcs can have multiple types.

This PR proposes to normalize this for Func by adding `type()` and `types()` and deprecating the old names; in addition to updating existing usages inside Halide, there is more drive-by change than usual:
- The base class for Generator's Input and Output classes already defined methods named `type()` and `types()`; these got renamed to `gio_type()` and `gio_types()` to avoid possible confusion or ambiguity. Additionally, they got moved into `protected` since nothing outside of Generator-related friend classes was intended to have access to them.
- All the Generator Input and Output classes now have `type()` methods that are wrappers for the underlying implementation (e.g Func, Expr, ImageParam, etc).

### DIFF
--- a/apps/fft/fft.cpp
+++ b/apps/fft/fft.cpp
@@ -107,7 +107,7 @@ ComplexExpr mul(ComplexExpr a, float re_b, float im_b) {
 // Specializations for some small DFTs of the first dimension of a
 // Func f.
 ComplexFunc dft2(ComplexFunc f, const string &prefix) {
-    Type type = f.output_types()[0];
+    Type type = f.types()[0];
 
     ComplexFunc F(prefix + "X2");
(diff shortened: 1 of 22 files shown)


zvookin:
Pretty sure the reason this is explicitly "output_types" is because "types" in the context could be confused for the argument types.

steven-johnson:
> Pretty sure the reason this is explicitly "output_types" is because "types" in the context could be confused for the argument types.

True, but I think the similarity and terseness make it worth doing. Do you object?

PR #6773, opened 2022-05-19
Add forwarders for 'name()' method for Funcs, ImageParam

steven-johnson:
(followup to #6772)

### DIFF
--- a/src/Generator.h
+++ b/src/Generator.h
@@ -1440,7 +1440,7 @@ protected:
     size_t array_size() const;
     virtual bool is_array() const;
 
-    const std::string &name() const;
+    const std::string &gio_name() const;
     IOKind kind() const;
 
     bool gio_types_defined() const;
@@ -1709,7 +1709,7 @@ public:
 
     template<typename T2>
     operator StubInputBuffer<T2>() const {
-        user_assert(!this->is_array()) << "Cannot assign an array type to a non-array type for Input " << this->name();
+        user_assert(!this->is_array()) << "Cannot assign an array type to a non-array type for Input " << this->gio_name();
         return StubInputBuffer<T2>(this->parameters_.at(0));
     }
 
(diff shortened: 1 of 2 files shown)


src/Generator.h
```diff
@@ -2334,9 +2337,6 @@ class GeneratorOutputBase : public GIOBase {
 
     // }@
 
-#undef HALIDE_OUTPUT_FORWARD
```
abadams, line 2337: Why this change? Is the macro used later?

steven-johnson: Oh, sorry, drive-by change -- apparently a holdover from way back when there were multiple forwarding macros, just dead code now

steven-johnson:
OK, I obviously didn't even test this locally because I thought it was a no-brainer, silly me

steven-johnson:
Had to rename the inherited `GIOBase::name()` method (as I did for `type()` yesterday) to avoid ambiguity with the new forwards, PTAL

abadams:
Hrm, I guess the name of an Output<Buffer> doesn't have to match the name of the Func you assign to it. As things currently stand, does calling name() return the Output name (i.e. gio_name)? Does this change it to return the Func name?

abadams:
That seems surprising. My expectation would have been that if you declare an Output<Buffer> with a name, calling name() on it gives you that name back, and you have to do something like my_output_buffer.func().name() or Func(my_output_buffer).name() to get the name of a Func.

steven-johnson:
> My expectation would have been that if you declare an Output with a name, calling name() on it gives you that name back

OK, but keep in mind that the (current) Python generator design has no special `name()` method on inputs or outputs -- the only such method available is the one on the underlying Func, ImageParam, etc. -- so this change brings the C++ behavior closer to the (proposed) Python design.

Also keep in mind that even in C++, the `gio_name` of all Inputs and Outputs is 100% fixed at compiletime; there is no way to replace those names via GeneratorParam or whatnot, so having a `name()` getter is something that would rarely be used.

steven-johnson:
Monday Morning Review Ping

steven-johnson:
I'm just going to drop this PR entirely: the only usage of the GIOBase::name() function that I'm aware of was in one piece of Google code, and we changed it to use a hardcoded name instead (because that's essentially what it was using). I presume that if/when other folks realize that the `name()` method is no longer public, they will let us know whether it's a problem for them.

Issue #6756, opened 2022-05-09
Loading wider type than the type of the Buffer


LebedevRI:
Problem: given buffer of `uint8_t`'s, how do to perform a load of `uint64_t` from it?
It's possible i'm completely missing it, but so far i have not found a reasonable solution.

The obvious way is: (for little-endian loads, that is)
```c++

  ImageParam input(type_of<uint8_t>(), 1, "input");

  Var elt("elt");

  Func wide_load("wide_load");
  wide_load(elt) = cast(CACHE_TYPE, input(elt));
  RDom r(0, CACHE_BYTES);
  wide_load(elt) = wide_load(elt) | (cast(CACHE_TYPE, input(elt + r)) << (8 * r));

  wide_load.update(0).unroll(r);

  Func cache("cache");
  cache(elt) = wide_load(elt);
```
(full code and dumps: https://godbolt.org/z/fhs9ajqhK)
But that doesn't really work, halide performs u8 loads,
and LLVM happily misses the point, has no middle-end load widening,
pulls apart the sequence during middle-end optimizations,
and by the back-end ISEL, there is no real way to recover.

What does work is:
```c++
  std::vector<Expr> input_bytes_le;
  for (int i = 0; i != CACHE_BYTES; ++i)
    input_bytes_le.push_back(input(input_offset + i));

  Func cache("cache");
  cache(elt) =
      reinterpret(CACHE_TYPE, Internal::Shuffle::make_concat(input_bytes_le));
```
(https://godbolt.org/z/hM844f37E)
... but something tells me i'm not supposed to use stuff from `Internal` directly...

Thoughts?

zvookin:
The general case of accessing data with unusual layouts comes up a fair bit. The simplest one is just interleaved vs. planar, which we support with strides and constraints on inputs/outputs, but this causes consistent difficulty for every new Halide user. Sometimes one wants to access data that is in arrays of structs and in an extreme case, those can have bitfields. Non-uniform sampling in YUV is another case. The similarity to the issue here is there is a lot of need for encapsulation of these patterns, including accessing integer data at different bit widths than it is declared with.

It seems introducing a set of encapsulations along the lines of BoundaryConditions would be useful.

Ideally these encapsulations would simply provide a new Func that is naturally indexable but that can also be scheduled in a meaningful way. The scheduling might be specified as an input to the encapsulation that makes the Func, or it could be that the encapsulation returns all the handles necessary to schedule the internal logic. (Or we could use Generators, maybe...)

[...]

PR #6775, opened 2022-05-19
Narrowing/widening the Func along a dimension (#6756)

LebedevRI:
Narrowing/widening the Func along a dimension

As discussed, this implements an utility to make it easier
to write code that wants to essentially operate either on
smaller chunks than the whole type, or operate on several
consecutive elements as a large element.

The bigger picture is that it is desirable to perform load
widening in such situations, and while this doesn't do that,
at least having a common interface should be a step in that direction.

I've took liberty to add/expose some QoL variable-bit-lenth
operations in IROperator, while there.

I believe, this has sufficient test coverage now.
While implmeneting, i stumbled&fixed
https://github.com/halide/Halide/pull/6782

Refs. #6756

### DIFF
--- a/src/FuncTypeChanging.cpp
+++ b/src/FuncTypeChanging.cpp
@@ -1,0 +1,130 @@
+#include "FuncTypeChanging.h"
+
+namespace Halide {
+
+static bool operator==(const Var &a, const Var &b) {
+    return a.same_as(b);
+}
+
+namespace FuncTypeChanging {
+
+// NOTE: Precondition: `chunk_idx u< num_chunks`!
+static Expr get_nth_chunk(const Expr &value, const Type &chunk_type,
+                          const Expr &chunk_idx, ChunkOrder chunk_order) {
+    int num_bits_total = value.type().bits();
+    int num_bits_per_chunk = chunk_type.bits();
+    int num_chunks = num_bits_total / num_bits_per_chunk;
+    user_assert(num_bits_total > num_bits_per_chunk &&
+                num_bits_total % num_bits_per_chunk == 0 && num_chunks > 1)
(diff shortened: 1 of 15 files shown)


src/FuncTypeChanging.cpp
```diff
@@ -0,0 +1,131 @@
...
+static bool operator==(const Var &a, const Var &b) {
+    return a.same_as(b);
+}
+
+namespace FuncTypeChanging {
+
+// NOTE: Precondition: `chunk_idx u< num_chunks`!
+static Expr get_nth_chunk(const Expr &value, const Type &chunk_type,
+                          const Expr &chunk_idx, ChunkOrder chunk_order) {
+    unsigned num_bits_total = value.type().bits();
```
steven-johnson, line 14: (here and elsewhere) `bits` is a signed type (`int`); even though the value is always expected to be positive, convention in Halide is to do the math in matching types

LebedevRI:
Weak weekly ping, thanks.

LebedevRI:
This is taking too long.
Must the rest of the infra be there, or can this proceed as-is?

LebedevRI:
Can we move forward with this as-is?

abadams:
I think we need the intrinsics and good codegen in place first, then we can see what the syntactic sugar can be to hit them. I'm away at a conference this week but I do have some down-time to code. I know the amount of time this is taking is frustrating, so I'll use it to try to get the intrinsics working over the next few days.

LebedevRI:
I think that sounds eerily similar to premature optimizations.

abadams:
The goal is to hit the minimum viable solution first, and worry about user-facing sugar for it second.

Issue #2450, opened 2017-10-18
Reuse compiled pipeline on multiple threads.


mcourteaux:
I'm building a scientific academic research application, that uses Halide. I have access to a beefy 40-core machine. Is it possible to reuse a compiled Halide function/pipeline on multiple threads in parallel? In terms of: setting pipeline parameters, setting input buffers, setting output buffers, making sure internal temporary buffers are not shared.

I would like to save on compilation time, as I'm now literally JIT compiling the same pipeline 40 times before the application starts it's actual work.

zvookin:
tl;dr We need to add some support to do this, though it isn't too difficult.

While multithreaded calls are fully supported for AOT compilation, there are two obstacles to making them work for JIT. (Note I'm going to use` Pipeline` instead of `Func` below as the actual implementation of functionality is at the Pipeline level. There should be pass through methods from `Func` to `Pipeline` for everything mentioned and thus the advice should apply to both.)

First, the `Pipeline` has to be compiled before it is called. The easiest way to do this is to arrange to call `compile_jit()` on the `Pipeline` on a single thread before doing realizations on other threads.

Second, JIT calls typically take their arguments from global `Param` and `ImageParam` objects and these are not threadsafe. The solution is to call the argv function directly with values taken from some passed in data structure, such as a map that has named entries corresponding to the names of the `Param` and `ImageParam` objects. If you need it in a hurry and are willing to make an array of pointers to the raw arguments, all this would require is making an accessor on `Pipeline` to the `argv_function` on the compiled module. But we should be able to make a threadsafe method that ensures the compilation has happened and takes all the args from passed in values...

mcourteaux:
I am most definitely not familiar with the internals of Halide. The `argv_function` you are referring to, is that the function that is returned by `compile_jit()`? If so, does it take the same parameters as if I would generate a AOT header file?

The return value of `compile_jit()` could be documented better, I think.

edubois:
That would be a very helpful functionality, are there any plans on doing this? Just curious.

zvookin:
Yes, it is on my TODO list. Probably not tomorrow, but probably within a month.

eldakms:
Hi, any update on this? I would also be very interested. Thanks!

edubois:
Hi,
Any news on this topic? Many thanks,

zvookin:
[...]

PR #6777, opened 2022-05-24
Add a new, alternate JIT-call convention

steven-johnson:
Add a way to call JIT code in C++ using the same calling conventions as AOT code. As a feature, this needs more work (e.g. more examples, more documentation), but it's arguably complete enough to land as-is to let people try it out.

Basically, you can now call `Func::compile_to_callable()`, which:
- compiles the Func via JIT
- returns a Callable, which is basically like a std::function, with a calling convention that ~exactly matches that of an AOT-compiled version of the same Func (even returning an `int` error result).
- the Callable returned is immutable and can be called repeatedly, and is unaffected by any subsequent changes made to the Func

This isn't intended to replace the `realize()` API, but it has a number of useful advantages:
- since the calling convention is basically identical to the AOT calling convention, it makes it much easier to write code that could be reused for both JIT and AOT environments
- the arguments you pass are threadsafe; with `realize()` you have to use `ParamMap` to achieve this
- since each Callable is fully jitted and isolated upon creation, it is easier to re-use the same Func with (say) different Target features within the same JIT space, or different JITHandlers, etc.

Known caveats:
- calling a Callable requires *exact* type matching for scalar types; e.g., if an argument is `uint8` and you pass `int32`, you will fail at runtime. Inserting an explicit cast is the easy remedy, but this is an area where the calling convention diverges from AOT. (this could perhaps be fixed, but at a cost of more overhead for the calling "thunk" code, and I wanted to keep it as lean as possible.)
- infer_input_bounds() isn't implemented for this API ~~I'm not sure if we need to do so, or if we can just tell people to use the same approach they do for AOT code~~ -- it's not necessary, as the same approach can (and should) be used for bounds inference as in AOT mode.

This is especially useful when used in conjunction with the new `create_callable_from_generator()` API, which lets you create a Callable from any Generator that is registered in the current Process, without the need for any of the Generator to be exposed at compiletime via .h file (either human-written or halide-generated-.stub.h file).


abadams:
I like it. I think it would be worth checking any real-world uses of ParamMap to see if they'd be cleaner this way.

steven-johnson:
> I like it. I think it would be worth checking any real-world uses of ParamMap to see if they'd be cleaner this way.

I've pinged the commenters on https://github.com/halide/Halide/issues/2450 to see if any of them have thoughts.

steven-johnson:
As an example of the overhead of the code as of right now: Given something like

```
Halide::Callable::call<1ul, Halide::Buffer<unsigned char, -1>&, int, float, Halide::Buffer<unsigned char, -1>&>:
    0x10000a2c0 <+0>:   pushq  %rbp
    0x10000a2c1 <+1>:   movq   %rsp, %rbp
    0x10000a2c4 <+4>:   pushq  %r15
    0x10000a2c6 <+6>:   pushq  %r14
    0x10000a2c8 <+8>:   pushq  %r13
    0x10000a2ca <+10>:  pushq  %r12
    0x10000a2cc <+12>:  pushq  %rbx
    0x10000a2cd <+13>:  subq   $0x58, %rsp
    0x10000a2d1 <+17>:  movq   %r9, -0x30(%rbp)
    0x10000a2d5 <+21>:  movq   %r8, %r15
    0x10000a2d8 <+24>:  movq   %rcx, %r12
    0x10000a2db <+27>:  movq   %rdx, %r13
    0x10000a2de <+30>:  movq   %rsi, %r14
    0x10000a2e1 <+33>:  movq   %rdi, %rbx
    0x10000a2e4 <+36>:  callq  0x10000a9f6               ; symbol stub for: Halide::Callable::arg_count() const
    0x10000a2e9 <+41>:  movl   $0xffffffff, %esi         ; imm = 0xFFFFFFFF
    0x10000a2ee <+46>:  cmpq   $0x5, %rax
    0x10000a2f2 <+50>:  jne    0x10000a396               ; <+214> at Halide.h:15316:13
    0x10000a2f8 <+56>:  movq   %rbx, %rdi
    0x10000a2fb <+59>:  callq  0x10000a9ea               ; symbol stub for: Halide::Callable::call_check_info() const
    0x10000a300 <+64>:  xorl   %ecx, %ecx
    0x10000a302 <+66>:  cmpl   $0x44003, (%rax)          ; imm = 0x44003
    0x10000a308 <+72>:  setne  %cl
    0x10000a30b <+75>:  leaq   -0x58(%rbp), %rdx
    0x10000a30f <+79>:  movq   %r14, -0x58(%rbp)
    0x10000a313 <+83>:  movq   %rdx, -0x80(%rbp)
    0x10000a317 <+87>:  xorl   %edx, %edx
    0x10000a319 <+89>:  cmpl   $0x34003, 0x4(%rax)       ; imm = 0x34003
    0x10000a320 <+96>:  setne  %dl
    0x10000a323 <+99>:  leal   (%rcx,%rdx,2), %ecx
    0x10000a326 <+102>: movq   0x8(%r13), %rdx
    0x10000a32a <+106>: addq   $0x20, %rdx
    0x10000a32e <+110>: movq   %rdx, -0x78(%rbp)
    0x10000a332 <+114>: xorl   %edx, %edx
    0x10000a334 <+116>: cmpl   $0x22000, 0x8(%rax)       ; imm = 0x22000
    0x10000a33b <+123>: setne  %dl
    0x10000a33e <+126>: leaq   -0x48(%rbp), %rsi
    0x10000a342 <+130>: movl   (%r12), %edi
    0x10000a346 <+134>: movl   %edi, -0x48(%rbp)
    0x10000a349 <+137>: movq   %rsi, -0x70(%rbp)
    0x10000a34d <+141>: xorl   %esi, %esi
    0x10000a34f <+143>: cmpl   $0x22002, 0xc(%rax)       ; imm = 0x22002
    0x10000a356 <+150>: leal   (%rcx,%rdx,4), %ecx
    0x10000a359 <+153>: setne  %sil
    0x10000a35d <+157>: leal   (%rcx,%rsi,8), %ecx
    0x10000a360 <+160>: leaq   -0x40(%rbp), %rdx
    0x10000a364 <+164>: movss  (%r15), %xmm0             ; xmm0 = mem[0],zero,zero,zero
    0x10000a369 <+169>: movss  %xmm0, -0x40(%rbp)
    0x10000a36e <+174>: movq   %rdx, -0x68(%rbp)
    0x10000a372 <+178>: xorl   %esi, %esi
    0x10000a374 <+180>: cmpl   $0x34003, 0x10(%rax)      ; imm = 0x34003
    0x10000a37b <+187>: setne  %sil
    0x10000a37f <+191>: shll   $0x4, %esi
    0x10000a382 <+194>: movq   -0x30(%rbp), %rax
    0x10000a386 <+198>: movq   0x8(%rax), %rax
    0x10000a38a <+202>: addq   $0x20, %rax
    0x10000a38e <+206>: orl    %ecx, %esi
    0x10000a390 <+208>: movq   %rax, -0x60(%rbp)
    0x10000a394 <+212>: je     0x10000a39e               ; <+222> at Halide.h
    0x10000a396 <+214>: movq   %rbx, %rdi
    0x10000a399 <+217>: callq  0x10000a3c0               ; Halide::Callable::fail_bad_call<1> at Halide.h:15293
    0x10000a39e <+222>: leaq   -0x80(%rbp), %rdx
    0x10000a3a2 <+226>: movl   $0x5, %esi
    0x10000a3a7 <+231>: movq   %rbx, %rdi
    0x10000a3aa <+234>: callq  0x10000a9fc               ; symbol stub for: Halide::Callable::call_argv(unsigned long, void const**) const
    0x10000a3af <+239>: addq   $0x58, %rsp
    0x10000a3b3 <+243>: popq   %rbx
    0x10000a3b4 <+244>: popq   %r12
    0x10000a3b6 <+246>: popq   %r13
    0x10000a3b8 <+248>: popq   %r14
    0x10000a3ba <+250>: popq   %r15
    0x10000a3bc <+252>: popq   %rbp
    0x10000a3bd <+253>: retq
```

So... it's not tiny, but it's not huge, and is mostly straight-line code. Definitely open for suggestions as to how it could be made smaller/faster.

mcourteaux:
I like it too. I checked what the new API looks like, and I like it that the style of calling is now much more like the AOT approach. I was using for 99% of my stuff in Halide AOT pipelines, but had a few unit tests that are JITing code. The ParamMap there is kinda odd.

For me it checks the most important two boxes:

 - It can compile pipelines that produce multiple outputs (which was not possible with `realize()`, I believe).
 - Gets rid of the need for the unpleasant ParamMap structure, which made it unattractive to switch from JIT to AOT or the other way around.

In the light of shrinking the gap between jitted and AOT code, I can point out that I got really confused with all the old Param naming conventions (ImageParam, Param, ParamMap) from the days when Halide was more targetted to image-only stuff. I like that the generators use more general naming (Buffer, Input, Output), which is missing from the jitting interface, IMO. 

Regarding the overhead, wouldn't it be possible to get this whole system of jit compiling to literally return the function pointer, as if it was an AOT function? The only tricky part would be to get the correct function pointer type, which I think might be possible to automatically produce with some C++ templating efforts.

All and all, I'm positive about these efforts, I really like where it's going. :smile: Cheers!

steven-johnson:
Added support for Callable in the Python bindings too (though without kwargs support as of yet, we could probably add that if we like though)

steven-johnson:
> * It can compile pipelines that produce multiple outputs (which was not possible with `realize()`, I believe).

This actually is possible to do using `Pipeline::realize` instead of `Func::realize`, but it can be a bit awkward.

> In the light of shrinking the gap between jitted and AOT code, I can point out that I got really confused with all the old Param naming conventions (ImageParam, Param, ParamMap) from the days when Halide was more targetted to image-only stuff. I like that the generators use more general naming (Buffer, Input, Output), which is missing from the jitting interface, IMO.

I think there is a general consensus in that regard -- e.g. I think there are multiple people who would be happy to have `Input<Buffer>` become a type alias for `ImageParam`, etc. -- it's just a little hard to untangle those from the entrails of Generator code, but I've been trying to work in that direction.

> Regarding the overhead, wouldn't it be possible to get this whole system of jit compiling to literally return the function pointer, as if it was an AOT function? The only tricky part would be to get the correct function pointer type, which I think might be possible to automatically produce with some C++ templating efforts.

Yeah, I thought about that -- in theory we could (also) have a templated version that returns a `std::function<>`, but the gotcha of course is that you'd have to recapitulate the arguments in the call, e.g.

      auto c = f.compile_to_stdfunction<int(halide_buffer_t*, int, double, int, halide_buffer_t*)>({...list of Arguments still needed here as we can't reliably infer everything we need from the template signature});

Naturally, this is tedious and error prone (though I'm sure we could check that you specified template arguments that matched the Arguments you specified).

...now, that said, adding support for this to a *Generator* would indeed remove the need to recapitulate the Arguments, since the Generator has that information internally.

steven-johnson:
Added wrappers to allow creating a Callable from any registered Generator, with (trivial) example in `example_jittest.cpp` -- note that the Generator doesn't need any .h file exposed to be called this way, it just has to be registered. PTAL. More example usage coming soon.

steven-johnson:
Cleanup up and added more examples (see stubtest_jittest for an interesting one) -- as mentioned in the revised PR description there is more work to be done, but I'd like to ask for a review of this PR as-is:
- What work *must* be done prior to landing?
- What work would likely be desirable afterwards, as necessary or desirable enhancement? (e.g., the `std::function` support described in a comment above)

(Disclaimer: I go on a 2.5 week vacation beginning June 8, so if we don't land it before then, either someone else will need to nurse it to landing, or we wait till end of June to land it. If it's not ready to land by then, of course we won't, but if it's close...)

test/generator/stubtest_jittest.cpp
```diff
@@ -60,56 +59,136 @@ int main(int argc, char **argv) {
...
+            {"untyped_buffer_input.dim", "3"},
+            {"simple_input.type", "float32"},
+            {"array_input.type", "float32"},
+            {"array_input.size", "2"},
+            {"int_arg.size", "2"},
+            {"tuple_output.type", "float32,float32"},
+            {"vectorize", "true"},
+        };
+
+        Callable stubtest = create_callable_from_generator(context, "stubtest", gp);
```
abadams, line 156: Referring to the registered generator by string name seems a bit weird here. Given a Generator class (not a stub), is it possible to say something akin to MyGeneratorClass::create(target).compile_to_callable()? Looking through Generator.h it seems like the awkward bit is setting the generator params

steven-johnson: You have a point -- I did it this way because for C++ Generators, there usually is no in-scope reference to the Generator class itself (it's often in a .cpp file with no .h, in an anon namespace to boot), so doing a lookup by name is simpler, and since it's still resolved at Halide Compilation time, an assertion-failure from name lookup is tantamount to any other compilation error.

That said, it would be easy enough to add such a method so that C++ Generators that *are* declared in-scope could be used directly. 

Also, maybe it would be feasible to just allow an `extern` declaration of the C++ Generator (or its factory or whatnot) to avoid needing .h files and just relying on name-mangling to keep the build honest -- though extern declaration of a CRTP template is kinda ugly. (Yeah, any sort of extern-declaration in a .cpp file is pretty gross, but this is C++ we're talking about, it's not like we have non-gross options for this sort of thing.)

steven-johnson: > maybe it would be feasible to just allow an extern declaration of the C++ Generator

So the answer here is "No, not really": forward declaration of a template class defined via CRTP is, well, honestly, I haven't figured out how to do that. But I'm not sure that would help much, since we'd still just have a class forward declaration without any "guts" to it (ie no methods to call). 

So I guess we're really looking at three possibilities here:
- Your Generator is defined in a such a way that its class definition is fully visible at the call site (either it's in a .h file or you have multiple Generators in a single .cpp). Using your approach here would indeed work and arguably be simplest, but AFAIK literally zero of the Generators inside Google fit this pattern, and only a tiny handful in existing public Halide code. If we want to encourage people to put Generators into .h files (instead of .cpp files) then this makes sense for future code (and we have no good reason not to support it in addition to other stuff), but as of right now it would seem to be of extremely limited use.
- The existing looking-by-string approach I have in place now. This is unsatisfying for the reasons you list, but works (well) in all the existing use cases I know of.
- The halfway approach... this is kinda yucky but here goes:
    - We rely on the fact that `HALIDE_REGISTER_GENERATOR()` is a macro and can define hidden helpers (and that this macro is *always* in the global namespace
    - We add a macro like `HALIDE_EXTERN_GENERATOR_FACTORY()` which is just the extern-declaration for those hidden helper funcs
    - Then we can use it something like
    ```
    // file: example_generator.cpp
    class ExampleClassName : Generator<ExampleClassName> {...};
    HALIDE_REGISTER_GENERATOR(ExampleClassName, example_registered_name)

    // example_jittest.cpp
    // (Note that we have to use the 'registration' name here, not the class name,
    // since the class name may well be in an anonymous namespace that is literally
    // impossible to specify here.)
    auto factory = HALIDE_EXTERN_GENERATOR_FACTORY(example_registered_name);

    auto callable = factory(context)->compile_to_callable();

    ```
    Stylistically, this isn't really any better than the current "look up by string" approach in this PR, but it would have the effect of resolving issues at link time, rather than compilation time. What do you think?

mcourteaux:
> Yeah, I thought about that -- in theory we could (also) have a templated version that returns a `std::function<>`, but the gotcha of course is that you'd have to recapitulate the arguments in the call, e.g.
> 
> ```
>   auto c = f.compile_to_stdfunction<int(halide_buffer_t*, int, double, int, halide_buffer_t*)>({...list of Arguments still needed here as we can't reliably infer everything we need from the template signature});
> ```
> 
> Naturally, this is tedious and error prone (though I'm sure we could check that you specified template arguments that matched the Arguments you specified).

@steven-johnson Hmm, I'm not sure repeating the argument types is needed. What about this?

    auto function = func.compile_to_stdfunc({p_img, p_int, p_float});

Where `compile_to_stdfunc` takes in a templated parameter pack tuple for all the Input and Outpus.
Then using some type traits, we map them to their resulting primitive:
```cpp
template<typename T> struct to_func_arg {};
template<template T> struct to_func_arg<Param<T>> { typedef type T; };
template<> struct to_func_arg<ImageParam> { typedef type halide_buffer*; };
```
I'm no template wizard, but I think you should be able to pass every tuple parameter through this type trait `to_func_arg<Args...>` to get the types needed to put inside the `std::function<...>`? We might have to separate inputs and outputs in a separate tuple?

I have no clue how to exactly do this in templating, but putting it together, might look something like this:

```cpp
template<class Args...>
std::function<void(to_func_arg<Args...>::type)> compile_to_stdfunc(Args... args) {

}
```
Maybe this is not even possible (I have never worked with `...` before so probably syntax is off, but see it used in many places).

steven-johnson:
> Then using some type traits, we map them to their resulting primitive:

The problem is that we can't infer all of the traits at compile-time: e.g. it's legal to use `Param<void>` which takes a `Halide::Type` as an argument (similar to `Buffer<void>`). In that case we can't know ahead of time what the right concrete scalar value is. (I guess we could just disallow using that method in that case...)

abadams:
I don't think we should take the current code structure too seriously as likely JIT usage. If a Generator is being used in a JIT context, I imagine the structure could be something like:

Foo.h:
```
int do_the_thing(const Buffer &in, Buffer &out);
```

Foo.cpp:
```
class MyGenerator ... 

int do_the_thing(const Buffer &in, Buffer &out) {
  static Callable c = MyGenerator::something(...);
  return c(in, out);
}
```

steven-johnson:
Revised the code to move the type-checking-code stuff into a constexpr table -- we can compile this into constant data at compiletime and just use that table for type checking inside `call_argv`, so the amount of code we inject at the call site is much smaller now. The example given previously (with `-O3`) now compiles to something like:

```
Halide::Callable::call<Halide::Buffer<unsigned char, -1>&, double, float, Halide::Buffer<unsigned char, -1>&>:
    0x100006890 <+0>:   pushq  %rbp
    0x100006891 <+1>:   movq   %rsp, %rbp
    0x100006894 <+4>:   subq   $0x50, %rsp
->  0x100006898 <+8>:   leaq   -0x28(%rbp), %rax
    0x10000689c <+12>:  movq   %rsi, -0x28(%rbp)
    0x1000068a0 <+16>:  movq   %rax, -0x50(%rbp)
    0x1000068a4 <+20>:  movq   0x8(%rdx), %rax
    0x1000068a8 <+24>:  addq   $0x20, %rax
    0x1000068ac <+28>:  movq   %rax, -0x48(%rbp)
    0x1000068b0 <+32>:  leaq   -0x18(%rbp), %rax
    0x1000068b4 <+36>:  movsd  (%rcx), %xmm0             ; xmm0 = mem[0],zero
    0x1000068b8 <+40>:  movsd  %xmm0, -0x18(%rbp)
    0x1000068bd <+45>:  movq   %rax, -0x40(%rbp)
    0x1000068c1 <+49>:  leaq   -0x10(%rbp), %rax
    0x1000068c5 <+53>:  movss  (%r8), %xmm0              ; xmm0 = mem[0],zero,zero,zero
    0x1000068ca <+58>:  movss  %xmm0, -0x10(%rbp)
    0x1000068cf <+63>:  movq   %rax, -0x38(%rbp)
    0x1000068d3 <+67>:  movq   0x8(%r9), %rax
    0x1000068d7 <+71>:  addq   $0x20, %rax
    0x1000068db <+75>:  movq   %rax, -0x30(%rbp)
    0x1000068df <+79>:  leaq   0xa2e(%rip), %rcx         ; int Halide::Callable::call<Halide::Buffer<unsigned char, -1>&, double, float, Halide::Buffer<unsigned char, -1>&>(Halide::JITUserContext*, Halide::Buffer<unsigned char, -1>&, double&&, float&&, Halide::Buffer<unsigned char, -1>&)::actual_arg_types
    0x1000068e6 <+86>:  leaq   -0x50(%rbp), %rdx
    0x1000068ea <+90>:  movl   $0x5, %esi
    0x1000068ef <+95>:  callq  0x100006ec0               ; symbol stub for: Halide::Callable::call_argv(unsigned long, void const**, unsigned int const*) const
    0x1000068f4 <+100>: addq   $0x50, %rsp
    0x1000068f8 <+104>: popq   %rbp
    0x1000068f9 <+105>: retq
```

Also, some drive-by encapsulation of some code from JITCache and Callable.

steven-johnson:
> class MyGenerator ... {
>   ...
> }
> 
> int do_the_thing(const Buffer &in, Buffer &out) {
>   static Callable c = MyGenerator::something(...);
>   return c(in, out);
> }
> ```

For this usage, it's already possible (with this PR) to do this:

```
GeneratorContext c() {
  return GeneratorContext(get_jit_target_from_environment();
}
int do_the_thing(const Buffer &in, Buffer &out) {
  static Callable c = MyGenerator::create(c())->compile_to_callable();
  return c(in, out);
}
```

(only semi-related: stashing a Callable in a static var makes my spidey-sense tingle, for reasons I can't quite put my finger on...)

steven-johnson:
Added code to produce a well-typed `std::function`, as discussed in comments earlier.

This is probably gonna be my pre-vacation stopping point.

steven-johnson:
>  I'll update this PR to ensure both are possible shortly.

Done, PTAL

src/Callable.cpp
```diff
@@ -0,0 +1,178 @@
...
+    } else {
+        do_check_fail(-1, argc, "defining");
+    }
+}
+
+// Entry point used from the std::function<> variant; we can skip the check_qcci() stuff
+// since we verified the signature when we created the std::function, so incorrect types or counts
+// should be impossible.
+/*static*/ int Callable::call_argv_fast(size_t argc, const void *const *argv) const {
+    // Callable should enforce these, so we can use assert() instead of internal_assert()
```
abadams, line 143: I'm confused by this. Don't we use internal_assert for checking invariants? Or is the claim that these are already checked elsewhere and this is effectively just documentation?

src/Callable.cpp
```diff
@@ -0,0 +1,178 @@
...
+    assert(contents->jit_cache.jit_target.has_feature(Target::UserContext));
+    assert(contents->jit_cache.arguments[0].name == "__user_context");
+
+    JITUserContext *context = *(JITUserContext **)const_cast<void *>(argv[0]);
+    assert(context != nullptr);
+
+    JITFuncCallContext jit_call_context(context, contents->saved_jit_handlers);
+
+    // debug(2) << "Calling jitted function\n";
```
abadams, line 152: Probably should remove this commented-out code, or raise the debug level

src/Callable.cpp
```diff
@@ -0,0 +1,178 @@
...
+    int exit_status = contents->jit_cache.call_jit_code(contents->jit_cache.jit_target, argv);
+    // debug(2) << "Back from jitted function. Exit status was " << exit_status << "\n";
+
+    // If we're profiling, report runtimes and reset profiler stats.
+    contents->jit_cache.finish_profiling(context);
+
+    // Don't call jit_call_context.finalize(): if no error handler was installed,
```
abadams, line 159: How does the error message make it out to the user in the case where they don't install a custom error handler?

src/Callable.cpp
```diff
@@ -0,0 +1,178 @@
...
+    // it will call halide_error_handler for us, which is fine for realize()
+    // and friends because there is no other way to report an error. For this code
+    // path, though, we just return the error code (if any); if a custom error
+    // handler was installed, it presumably will get the first shot at handling it,
+    // and if not, the caller must handle it.
+    //
+    // No: jit_call_context.finalize(exit_status);
```
abadams, line 166: Not sure what these two lines are about. They seem redundant.

src/Callable.h
```diff
@@ -0,0 +1,357 @@
...
+    static constexpr halide_type_t type() {
+        return halide_type_t();
+    }
+    static constexpr int dims() {
+        return -1;
+    }
+};
+
+template<typename T, int Dims>
+struct HalideBufferStaticTypeAndDims<::Halide::Buffer<T, Dims>> {
```
abadams, line 50: Should these be static methods on Halide::Buffer and Halide::Runtime::Buffer instead?

src/Callable.h
```diff
@@ -0,0 +1,357 @@
...
+    static constexpr std::array<QuickCallCheckInfo, sizeof...(Args)> make_qcci_array() {
+        return std::array<QuickCallCheckInfo, sizeof...(Args)>{make_qcci<Args>()...};
+    }
+
+    // ---------------------------------
+
+    // This value is constructed so we can do a complete type-and-dim check
+    // of Buffers, and is used for the make_std_function() method, to ensure
+    // that if we specify static type-and-dims for Buffers, the ones we specify
+    // actually match the underlying code. We take horrible liberties with halide_type_t
```
abadams, line 143: please describe the liberties in the comment. I see some repurposing of the lanes field below?

src/Callable.h
```diff
@@ -0,0 +1,357 @@
...
+
+        template<typename... Args>
+        explicit ArgvStorage(Args &&...args) {
+            fill_slots(0, std::forward<Args>(args)...);
+        }
+
+    private:
+        template<typename T, int Dims>
+        HALIDE_ALWAYS_INLINE void fill_slot(size_t idx, const ::Halide::Buffer<T, Dims> &value) {
+            // Don't call ::Halide::Buffer::raw_buffer(): it includes "user_assert(defined())"
```
abadams, line 224: So what happens if you pass an undefined buffer?

I don't think we should worry too much about the cost of easily-predicted branches. They're often free.

src/Callable.h
```diff
@@ -0,0 +1,357 @@
...
+        return call_argv_checked(count, &argv.argv[0], actual_arg_types.data());
+    }
+
+    /** Return the expected Arguments for this Callable, in the order they must be specified, including all outputs.
+     * Note that the first entry will *always* specify a JITUserContext. */
+    const std::vector<Argument> &arguments() const;
+
+public:
+    template<typename... Args>
+    HALIDE_MUST_USE_RESULT int
```
abadams, line 295: If the user has installed an error handler that throws an exception or aborts, must they really use the result here?

src/Callable.h
```diff
@@ -0,0 +1,357 @@
...
+        } else {
+            // Explicitly prepend JITUserContext* as first actual-arg-type.
+            constexpr auto actual_arg_types = make_fcci_array<JITUserContext *, First, Rest...>();
+            check_fcci(actual_arg_types.size(), actual_arg_types.data());
+
+            // Capture *this to ensure that the CallableContents stay valid as long as the std::function does
+            return [*this](auto &&first, auto &&...rest) -> int {
+                // Explicitly prepend an (empty) JITUserContext to the args.
+                JITUserContext empty;
+                constexpr size_t count = 1 + 1 + sizeof...(rest) + 1;
```
abadams, line 347: What do the three 1s represent? The first is going to be JITUserContext, then there's I guess the argument "first", and the last one is?

abadams:
Generally looks good (and is a nice step forward for Halide!), but I'm concerned/confused about error handling.

src/Callable.cpp
steven-johnson: The latter -- I'll add a comment to that effect. The idea here is to avoid the cost of internal_assert() in this case since it's theoretically redundant here.

src/Callable.cpp
steven-johnson: Exactly as we do for AOT calls: we return an error code, which the caller must check.

src/Callable.cpp
steven-johnson: Scalpel left in patient, removed.

src/Callable.cpp
abadams: The default for AOT calls is the pipeline calls halide_error, which prints a message then calls abort()

abadams: It's common for people to override this to log the error string in some other way and then return. The default JIT behavior is that the error string is thrown as an exception when the pipeline returns. This Callable design seems to just drop the error string though and return a non-zero value.

src/Callable.h
steven-johnson: ...maybe? I figured they were specialized enough not to warrant adding to both of those classes. Easy enough to generalize if we need them elsewhere someday (but so far, we haven't).

src/Callable.h
steven-johnson: Done.

src/Callable.h
steven-johnson: You crash on a null pointer access. Don't do that :-)

But seriously, I'll include the user_assert() if you like, but it included a surprising amount of code (including an out-of-line call IIRC), enough to make me want to elide it here -- I was really trying hard to keep the amount of hidden overhead to a bare minimum. Let me know if I should re-add it.

src/Callable.h
steven-johnson: I think it represents a dumb typo on my part, fixing.

steven-johnson:
PTAL

src/Callable.h
abadams: If we could somehow convert that case into passing a nullptr halide_buffer_t as the argument, then the result would be the Halide pipeline throwing an error. E.g:
```
auto ptr = value.get();
ptr ? ptr->raw_buffer : nullptr;
```

steven-johnson: Nice compromise, done.

steven-johnson:
OK, I have revised the error handling code per our discussion and added more testing -- PTAL

test/correctness/callable_generator.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+                out_(x) = reinterpret<uint64_t>(handle_);
+                if (vectorize_) {
+                    out_.vectorize(x, 4);
+                }
+            }
+        };
+
+        auto gen_1 = TestGen3::create(context);
+        gen_1->set_generatorparam_values({{"vectorize", "false"}});
```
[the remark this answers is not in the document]
abadams: This is a nit, but IMO it would good to change it to avoid the unnecessary string-typing.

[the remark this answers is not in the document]
steven-johnson: I'm gonna push back on the string-typing concern a bit:
- This uses the same syntax that build systems would use, which IMHO is a good thing
- Setting the generator param here via C++ would work (because we know the complete type), but it wouldn't work for the look-up-by-registered-name version because that returns an AbstractGenerator. (This syntax works with both).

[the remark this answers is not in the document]
abadams: IMO This should serve as an example of how to get a callable out of a generator declared in scope. If you look up by registered string name it's reasonable to set the generator params by string, but if you create the generator by actual C++ name, you should also set the generator params by actual c++ name. There's no reason to make yourself vulnerable to typos that could have been caught at C++ compile time here.

I think the build system isn't that relevant because we're talking about jit usage of an in-scope generator, which isn't going to show up in the build system at all. Even if it did, in the build system it would look like `vectorize=false`, not `{{"vectorize", "false"}}`. It's debatable whether or not `vectorize.set(false)` is closer to one or the other.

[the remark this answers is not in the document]
steven-johnson: Fine, will change

src/Callable.h
abadams: Unless you've overriding error handling, the only value a halide pipeline can possibly return is zero. Returning non-zero values is non-standard behavior. Our AOT apps (e.g. local_laplacian) don't check the return code for this reason.

steven-johnson: Making it easy to ignore possible error results is... not a good idea. 

When we added the `HALIDE_MUST_USE_RESULT` to AOT results inside Google, it revealed a *lot* of places where errors were being ignored (and most of those places were using a runtime that replaced the error handler).

We are introducing an effectively entirely new API here; it would be a serious design mistake to have the default recommended style be "ignore potential errors".

steven-johnson: Updated per chat discussion, PTAL

mcourteaux:
Congrats on this massive amount of work! :partying_face:

PR #6780, opened 2022-05-24
slow tests should support sharding

steven-johnson:
The simd_op_check test suite is pretty slow (especially for wasm, where it is interpreted); at one point we tried to use ThreadPool to speed it up, but too many pieces of Halide IR aren't threadsafe and we disabled it long ago.

This removes the ThreadPool usage entirely, and instead adds support for the GoogleTest 'sharded test' protocol, which uses certain env vars to allow a test to opt in for splitting its test into smaller pieces.

At present our buildbot isn't attempting to make use of this feature, but it will be a big win for downstream usage in Google, where tests that run "too long" are problematic and splitting them into multiple shards makes various day to day activiites much more pleasant.

### DIFF
--- a/test/common/test_sharding.h
+++ b/test/common/test_sharding.h
@@ -26,14 +26,19 @@ namespace Test {
 // The Halide buildbots don't (yet) make use of these, but some downstream consumers do.
 
 class Sharder {
+    // returns empty string (not null) if env var not found
     static std::string get_env(const char *v) {
         const char *r = getenv(v);
         if (!r) r = "";
         return r;
     }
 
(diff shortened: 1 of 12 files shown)


steven-johnson:
(If this change seems acceptable, I will probably refactor the sharding code into a small utility for tests and apply it to a few other extra-slow tests, e.g. vector_reductions)

abadams:
Factoring it into a small utility header or something sounds good to me

steven-johnson:
Refactored and added changes to other slow tests. PTAL.

vksnk (approved):

PR #6782, opened 2022-05-26
[miscompile] Don't de-negate and change direction of shifts-by-unsigned

LebedevRI:
I know this does not have a test, will add one later.

I'm afraid the problem is really obvious:
https://github.com/halide/Halide/blob/b5f024fa83b6f1cfe5e83a459c9378b7c5bf096d/src/CodeGen_LLVM.cpp#L2628-L2649
^ the shift direction is treated as flippable by the codegen
iff the shift amount is signed :)

I've hit this when writing tests for https://github.com/halide/Halide/pull/6775


LebedevRI:
Now with a test :)

abadams (approved):

steven-johnson (approved):

LebedevRI:
Hm, looks like a fluke?

LebedevRI:
Thank you!

Issue #6783, opened 2022-05-26
18% performance regression on x86 introduced by Update CodeGen_PTX_Dev to use new PassManager (#6718)


froody:
460c77eca Update CodeGen_PTX_Dev to use new PassManager (#6718)

I updated to main (ad1e7f6a6) and noticed that my pipeline took was much slower compared to Halide-v13.0.4. I ran git bisect and identified the above commit as the cause of a 18% regression in my pipeline on x86, i.e. my benchmark results:

Before: 0.5851s
After: 0.6914s

I'm not sure if I can share the details of my pipeline but if you have trouble reproducing the problem then let me know and I'll see what I can do.

llvm-version: llvmorg-13.0.0
cmake invocation: 
```
cmake -G Ninja \
                 -DCMAKE_BUILD_TYPE=Release \
                 -DWITH_PYTHON_BINDINGS=ON \
                 -DPYTHON_EXECUTABLE=$CONDA_ENV/bin/python3.9 \
                 -DPython3_ROOT=$CONDA_ENV \
                 -DLLVM_TARGETS_TO_BUILD="host;JSBackend;WebAssembly" \
                 -DLLVM_INCLUDE_EXAMPLES=OFF -DLLVM_INCLUDE_TESTS=OFF -DCLANG_INCLUDE_TESTS=OFF \
                 ..
```
CPU: AMD EPYC 7742

abadams:
Thanks for the regression report. Which LLVM version are you compiling against? Or are you using a prebuilt binary?

abadams:
Hopefully fixed by #6785

PR #6785, opened 2022-05-26
LLVM codegen: register AA pipeline if LLVM is older than 14

LebedevRI:
It's the default after https://reviews.llvm.org/D113210 / https://github.com/llvm/llvm-project/commit/13317286f8298eb3bafa9ddebd1c03bef4918948,
but still needs to be done for earlier LLVM's.

Refs. https://github.com/halide/Halide/issues/6783
Refs. https://github.com/halide/Halide/pull/6718

Partially reverts https://github.com/halide/Halide/pull/6718

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1119,6 +1119,13 @@ void CodeGen_LLVM::optimize_module() {
     llvm::CGSCCAnalysisManager cgam;
     llvm::ModuleAnalysisManager mam;
 
+#if LLVM_VERSION < 140
+    // If building against LLVM older than 14, explicitly specify AA pipeline.
+    // Not needed with LLVM14 or later, already the default.
+    llvm::AAManager aa = pb.buildDefaultAAPipeline();
+    fam.registerPass([&] { return std::move(aa); });
+#endif
+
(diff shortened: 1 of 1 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -1119,6 +1119,13 @@ void CodeGen_LLVM::optimize_module() {
     llvm::CGSCCAnalysisManager cgam;
     llvm::ModuleAnalysisManager mam;
 
+#if LLVM_VERSION < 140
```
LebedevRI, line 1122: Note that pedantically, this check is wrong for the first half of commits between 14-init and 14.0.0,
since i believe it says that it's not needed starting with 14-init, while it's not needed starting with 14-init+~2k.

LebedevRI, line 1122: Note that pedantically, this check is wrong for the first half of commits between 14-init and 14.0.0,
since i believe it says that it's not needed starting with 14-init, while it's not needed starting with 14-init+~2k.

LebedevRI:
Looks like yet another traditional fluke.

abadams (approved):

LebedevRI:
Thank you!

PR #6787, opened 2022-05-26
Move some options from execute_generator back to generate_filter_main

steven-johnson:
Loading plugins and setting the default autoscheduler name both change global state, which isn't a desirable fit for execute_generator(), since it's not intended to mutate global state. (Mutating the state from a main() function is of course a reasonable thing to do.)

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -867,6 +867,19 @@ gengen
         }
     }
 
+    // It's possible that in the future loaded plugins might change
+    // how arguments are parsed, so we handle those first.
+    for (const auto &lib_path : split_string(flags_info["-p"], ",")) {
+        if (!lib_path.empty()) {
+            load_plugin(lib_path);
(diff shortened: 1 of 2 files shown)


abadams:
I get the plugins, but isn't the autoscheduler name per-generator execution?

steven-johnson:
> autoscheduler name per-generator execution

well, yeah, but (unfortunately) it's currently controlled by a global -- if you tried to call execute_generator() from multiple threads, you'd be unhappy. (to be fair, there are probably plenty of other places in our compiler that also aren't threadsafe, but no point in adding to them.

really, the right fix here is to change the `auto_schedule` GeneratorParam from a bool into a string and (eventually) nuke the 'default' setting entirely, but I don't want to mess with that until we resolve the discussion about GeneratorContext and friends.

abadams (approved):

PR #6792, opened 2022-05-31
Improved find_constant_bound(s)

rootjalex:
This PR provides a series of methods for removing/simplifying correlated expressions for `find_constant_bounds`:

- Bounded let-substitutions (~~n=100~~ edit: n=16). We don't want to always substitute all lets, but some constant bounds can be calculated just by a small number of substitutions.
- Removing unbounded terms from mins/maxs. A (simplistic) example is below:
```
Find lower bound on:
max(x, y) - (z + y)
With z : [0, 8]
```
  This method would note that `x` is unbounded, and therefore the lhs of the `max` can be stripped, producing:
```
y - (z - y) -> 0 - z -> lower bounded by -8
```
- Affine term reordering. Halide’s TRS-based simplification can only cancel terms in sums up to a certain depth, this method uses a linear-time algorithm for canceling like-terms.
- Pushing rationals inwards. This technique pushes multiplications inwards to allow stronger simplification. More importantly, it pushes divisions inwards via a safe approximation, most encapsulate by the following equations:
```
// Addition:
(a / n) + (b / n) <= (a + b) / n <= (a / n) + (b / n) + 1
// Subtraction:
(a / n) - (b / n) - 1 <= (a - b) / n <= (a / n) - (b / n)
```
  This allows us to push divisions inside additions/subtractions which can improve the ability to cancel like terms in a lot of generated equations.

@abadams  ran a series of experiments with randomly-generated schedules (n=256) on a series of apps (bgu, camera_pipe, conv_layer, depthwise_separable_conv, harris, hist, iir_blur, lens_blur, max_filter, stencil_chain, unsharp), and here is a summary of the results (percentages are total across the benchmarks):

Less failed unrolls: bgu (5 -> 3), camera_pipe (69 -> 58), harris (197 -> 160), lens_blur (158 -> 12), max_filter (338 -> 242), unsharp (110 -> 32)
Less memory: camera_pipe (0.6%), depthwise_separable_conv (0.3%), hist (0.1%), lens_blur (0.6%), unsharp (0.2%)
Less malloc calls: camera_pipe (173592 -> 171612), harris (608943  -> 608655), lens_blur (144324 -> 141888), stencil_chain (701085 -> 698712), unsharp (127428 -> 127284)
Some small runtime improvements (0.05% to 0.6%) : bgu, camera_pipe, harris, hist, iir_blur, lens_blur, max_filter, stencil_chain

More memory: harris (0.06%), iir_blur (0.002%), stencil_chain (0.1%)

The runtime improvements might not be statistically significant, but I think better loop unrolling and improved stack allocations are important contributions.

For apps with no improved unrolling, compilation times increase by a small amount (~3%). With improved unrolling, there are large increases but are mostly due to the fact that generating the unrolled code takes longer in both our codegen and LLVM codegen.

This work was part of a project with @abadams and @shoaibkamil .


src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1144 @@
...
+#include "SimplifyCorrelatedDifferences.h"
+#include "UniquifyVariableNames.h"
+
+#include <vector>
+
+namespace Halide {
+namespace Internal {
+
+static bool disable_approximate_methods() {
+    return get_env_variable("HL_DISABLE_APPROX_CBOUNDS") == "1";
```
steven-johnson, line 19: Do we want to pay this cost for every call to `find_constant_bound[s]`? Is this just for debugging or for "real" cases? Maybe we should instead do something like

```
static bool disable_approximate_methods() {
    static bool b = get_env_variable("HL_DISABLE_APPROX_CBOUNDS") == "1";
    return b;
}
```

rootjalex: Oops, good catch, thanks! Will fix.

rootjalex: Actually, I'm not even entirely sure this should remain here, I considered allowing the new methods to be toggled on or off, but I'm not sure we really need that. @abadams what do you think?

steven-johnson: If you leave this in then someone, somewhere absolutely will end up basing an entire industry on this one little env var, so unless there is a crystal clear use for it now and in the future, please remove it, or at least leave commented-out by default.

rootjalex: Removed in 6577f30

abadams:
Torture testing inside Google would be pretty helpful, thanks.

steven-johnson:
Testing in Google, I find only one new failure, but... it appears to be a hang (or near-infinite loop) inside `Bounding small realizations...` when compiling one specific Generator. Adding to the fun, it's in some proprietary stuff that might be hard to share publicly. Let me see if I can narrow things down further...

steven-johnson:
Yeah, we definitely get stuck ~forever in `bound_small_allocations()`, which was only changed to include the new header, so something about the change in definition has injected something here. Let me see if I can come up with a repro case I can share.

steven-johnson:
So far, what I'm finding is that we have a fairly complex Expr that is the input to remove_unbounded_terms:

```
(((let t418 = min(max(min(max(min(max(min(max(min(max(min(max(min(max(min((foo$13.extent.0 + foo$13.min.0) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0) in (max(min(max(min(max(min(max(min(max(min(max(min(max(t418, 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), max(min(max(min(max(min(max(min(max(max(min(max(t418, 1) + 4, input.extent.0), t418), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 4, input.extent.0), 1) + 1) + 3, input.extent.0), 1) + -1)) - (let t407 = min(max(min(min(min(max(foo$13.min.0, -1), min(max(min(input.extent.0 + -1, foo$13.min.0), 0), max(min(foo$13.min.0 + 2, input.extent.0), 1) + -1)) + 2, input.extent.0) + -1, foo$13.min.0), 0), max(min(min(max(min(min(max(foo$13.min.0, -1) + 2, input.extent.0) + -1, foo$13.min.0), 0), max(min(min(max(min(input.extent.0 + -1, foo$13.min.0), 0), max(min(foo$13.min.0 + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) in (let t411 = min(max(min(min(min(min(max(min(input.extent.0 + -1, t407), 0), max(min(t407 + 2, input.extent.0), 1) + -1), t407) + 2, input.extent.0) + -1, t407), 0), max(min(min(max(min(min(t407 + 2, input.extent.0) + -1, t407), 0), max(min(min(max(min(input.extent.0 + -1, t407), 0), max(min(t407 + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) in (let t423 = min(max(min(min(min(min(max(min(input.extent.0 + -1, t411), 0), max(min(t411 + 2, input.extent.0), 1) + -1), t411) + 2, input.extent.0) + -1, t411), 0), max(min(min(max(min(min(t411 + 2, input.extent.0) + -1, t411), 0), max(min(min(max(min(input.extent.0 + -1, t411), 0), max(min(t411 + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) in (let t427 = min(max(min(min(min(min(max(min(input.extent.0 + -1, t423), 0), max(min(t423 + 2, input.extent.0), 1) + -1), t423) + 2, input.extent.0) + -1, t423), 0), max(min(min(max(min(min(t423 + 2, input.extent.0) + -1, t423), 0), max(min(min(max(min(input.extent.0 + -1, t423), 0), max(min(t423 + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1) in min(max(min(min(t427 + 2, input.extent.0) + -1, t427), 0), max(min(min(min(max(min(input.extent.0 + -1, t427), 0), min(min(max(min(t427 + 2, input.extent.0), 1), min(min(min(max(min(input.extent.0 + -1, t427), 0), min(min(max(min(t427 + 2, input.extent.0), 1), min(min(min(max(min(input.extent.0 + -1, t427), 0), min(min(max(min(t427 + 2, input.extent.0), 1), min(min(min(max(min(input.extent.0 + -1, t427), 0), min(min(max(min(t427 + 2, input.extent.0), 1), min(min(min(max(min(input.extent.0 + -1, t427), 0), min(max(min(t427 + 2, input.extent.0), 1), min(max(min(input.extent.0 + -1, t427), 0), min(max(min(t427 + 2, input.extent.0), 1), min(max(min(input.extent.0 + -1, t427), 0), min(max(min(t427 + 2, input.extent.0), 1), max(min(input.extent.0 + -1, t427), 0) + 1)) + 1)) + 1)), max(min(t427 + 2, input.extent.0), 1) + -1), min(max(min(input.extent.0 + -1, t427), 0), max(min(t427 + 2, input.extent.0), 1) + -1) + 2)), max(min(input.extent.0 + -1, t427), 0) + 1)), max(min(t427 + 2, input.extent.0), 1) + -1), min(max(min(input.extent.0 + -1, t427), 0), max(min(t427 + 2, input.extent.0), 1) + -1) + 2)), max(min(input.extent.0 + -1, t427), 0) + 1)), max(min(t427 + 2, input.extent.0), 1) + -1), min(max(min(input.extent.0 + -1, t427), 0), max(min(t427 + 2, input.extent.0), 1) + -1) + 2)), max(min(input.extent.0 + -1, t427), 0) + 1)), max(min(t427 + 2, input.extent.0), 1) + -1), min(max(min(input.extent.0 + -1, t427), 0), max(min(t427 + 2, input.extent.0), 1) + -1) + 2)), max(min(input.extent.0 + -1, t427), 0) + 1)), max(min(t427 + 2, input.extent.0), 1) + -1) + 2, input.extent.0), 1) + -1)))))) + 1)
```

which becomes insanely huge afterwards (too large to bother pasting here -- something like 7MB of text when the Expr is printed), and that's *after* the call to simplify().

rootjalex:
Is it possible to know if the scope has any values actually set? Sorry, I didn't realize that printing scope only prints the names, I need the corresponding intervals as well.

rootjalex:
Definitely seems like the issue here is `substitute_some_lets`. Not sure exactly what the count should be, but 100 is too high

steven-johnson:
> Is it possible to know if the scope has any values actually set? Sorry, I didn't realize that printing scope only prints the names, I need the corresponding intervals as well.

```
{
  output$1.s0.x.max: 0, (void *)pos_inf
  output$1.s0.x.min: 0, (void *)pos_inf
  output$1.s0.y.max: 0, (void *)pos_inf
  output$1.s0.y.min: 0, (void *)pos_inf
  output$1.s1.r8$x.max: 3, 3
  output$1.s1.r8$x.min: 0, 0
  output$1.s1.x.max: 0, (void *)pos_inf
  output$1.s1.x.min: 0, (void *)pos_inf
  output$1.s1.y.max: 0, (void *)pos_inf
  output$1.s1.y.min: 0, (void *)pos_inf
  output$10.s0.x.max: 0, (void *)pos_inf
  output$10.s0.x.min: 0, (void *)pos_inf
  output$10.s0.y.max: 0, (void *)pos_inf
  output$10.s0.y.min: 0, (void *)pos_inf
  output$10.s1.r125$x.max: 3, 3
  output$10.s1.r125$x.min: 0, 0
  output$10.s1.x.max: 0, (void *)pos_inf
  output$10.s1.x.min: 0, (void *)pos_inf
  output$10.s1.y.max: 0, (void *)pos_inf
  output$10.s1.y.min: 0, (void *)pos_inf
  output$11.s0.x.max: 0, (void *)pos_inf
  output$11.s0.x.min: 0, (void *)pos_inf
  output$11.s0.y.max: 0, (void *)pos_inf
  output$11.s0.y.min: 0, (void *)pos_inf
  output$11.s1.r138$x.max: 3, 3
  output$11.s1.r138$x.min: 0, 0
  output$11.s1.x.max: 0, (void *)pos_inf
  output$11.s1.x.min: 0, (void *)pos_inf
  output$11.s1.y.max: 0, (void *)pos_inf
  output$11.s1.y.min: 0, (void *)pos_inf
  output$12.s0.x.max: 0, (void *)pos_inf
  output$12.s0.x.min: 0, (void *)pos_inf
  output$12.s0.y.max: 0, (void *)pos_inf
  output$12.s0.y.min: 0, (void *)pos_inf
  output$12.s1.r151$x.max: 3, 3
  output$12.s1.r151$x.min: 0, 0
  output$12.s1.x.max: 0, (void *)pos_inf
  output$12.s1.x.min: 0, (void *)pos_inf
  output$12.s1.y.max: 0, (void *)pos_inf
  output$12.s1.y.min: 0, (void *)pos_inf
  output$13.s0.x.max: 0, (void *)pos_inf
  output$13.s0.x.min: 0, (void *)pos_inf
  output$13.s0.y.max: 0, (void *)pos_inf
  output$13.s0.y.min: 0, (void *)pos_inf
  output$13.s1.r164$x.max: 3, 3
  output$13.s1.r164$x.min: 0, 0
  output$13.s1.x.max: 0, (void *)pos_inf
  output$13.s1.x.min: 0, (void *)pos_inf
  output$13.s1.y.max: 0, (void *)pos_inf
  output$13.s1.y.min: 0, (void *)pos_inf
  output$14.s0.x.max: (void *)neg_inf, (void *)pos_inf
  output$14.s0.x.min: (void *)neg_inf, (void *)pos_inf
  output$14.s0.y.max: (void *)neg_inf, (void *)pos_inf
  output$14.s0.y.min: (void *)neg_inf, (void *)pos_inf
  output$14.s1.r177$x.max: 3, 3
  output$14.s1.r177$x.min: 0, 0
  output$14.s1.x.max: (void *)neg_inf, (void *)pos_inf
  output$14.s1.x.min: (void *)neg_inf, (void *)pos_inf
  output$14.s1.y.max: (void *)neg_inf, (void *)pos_inf
  output$14.s1.y.min: (void *)neg_inf, (void *)pos_inf
  output$2.s0.x.max: 0, (void *)pos_inf
  output$2.s0.x.min: 0, (void *)pos_inf
  output$2.s0.y.max: 0, (void *)pos_inf
  output$2.s0.y.min: 0, (void *)pos_inf
  output$2.s1.r21$x.max: 3, 3
  output$2.s1.r21$x.min: 0, 0
  output$2.s1.x.max: 0, (void *)pos_inf
  output$2.s1.x.min: 0, (void *)pos_inf
  output$2.s1.y.max: 0, (void *)pos_inf
  output$2.s1.y.min: 0, (void *)pos_inf
  output$3.s0.x.max: 0, (void *)pos_inf
  output$3.s0.x.min: 0, (void *)pos_inf
  output$3.s0.y.max: 0, (void *)pos_inf
  output$3.s0.y.min: 0, (void *)pos_inf
  output$3.s1.r34$x.max: 3, 3
  output$3.s1.r34$x.min: 0, 0
  output$3.s1.x.max: 0, (void *)pos_inf
  output$3.s1.x.min: 0, (void *)pos_inf
  output$3.s1.y.max: 0, (void *)pos_inf
  output$3.s1.y.min: 0, (void *)pos_inf
  output$4.s0.x.max: 0, (void *)pos_inf
  output$4.s0.x.min: 0, (void *)pos_inf
  output$4.s0.y.max: 0, (void *)pos_inf
  output$4.s0.y.min: 0, (void *)pos_inf
  output$4.s1.r47$x.max: 3, 3
  output$4.s1.r47$x.min: 0, 0
  output$4.s1.x.max: 0, (void *)pos_inf
  output$4.s1.x.min: 0, (void *)pos_inf
  output$4.s1.y.max: 0, (void *)pos_inf
  output$4.s1.y.min: 0, (void *)pos_inf
  output$5.s0.x.max: 0, (void *)pos_inf
  output$5.s0.x.min: 0, (void *)pos_inf
  output$5.s0.y.max: 0, (void *)pos_inf
  output$5.s0.y.min: 0, (void *)pos_inf
  output$5.s1.r60$x.max: 3, 3
  output$5.s1.r60$x.min: 0, 0
  output$5.s1.x.max: 0, (void *)pos_inf
  output$5.s1.x.min: 0, (void *)pos_inf
  output$5.s1.y.max: 0, (void *)pos_inf
  output$5.s1.y.min: 0, (void *)pos_inf
  output$6.s0.x.max: 0, (void *)pos_inf
  output$6.s0.x.min: 0, (void *)pos_inf
  output$6.s0.y.max: 0, (void *)pos_inf
  output$6.s0.y.min: 0, (void *)pos_inf
  output$6.s1.r73$x.max: 3, 3
  output$6.s1.r73$x.min: 0, 0
  output$6.s1.x.max: 0, (void *)pos_inf
  output$6.s1.x.min: 0, (void *)pos_inf
  output$6.s1.y.max: 0, (void *)pos_inf
  output$6.s1.y.min: 0, (void *)pos_inf
  output$7.s0.x.max: 0, (void *)pos_inf
  output$7.s0.x.min: 0, (void *)pos_inf
  output$7.s0.y.max: 0, (void *)pos_inf
  output$7.s0.y.min: 0, (void *)pos_inf
  output$7.s1.r86$x.max: 3, 3
  output$7.s1.r86$x.min: 0, 0
  output$7.s1.x.max: 0, (void *)pos_inf
  output$7.s1.x.min: 0, (void *)pos_inf
  output$7.s1.y.max: 0, (void *)pos_inf
  output$7.s1.y.min: 0, (void *)pos_inf
  output$8.s0.x.max: 0, (void *)pos_inf
  output$8.s0.x.min: 0, (void *)pos_inf
  output$8.s0.y.max: 0, (void *)pos_inf
  output$8.s0.y.min: 0, (void *)pos_inf
  output$8.s1.r99$x.max: 3, 3
  output$8.s1.r99$x.min: 0, 0
  output$8.s1.x.max: 0, (void *)pos_inf
  output$8.s1.x.min: 0, (void *)pos_inf
  output$8.s1.y.max: 0, (void *)pos_inf
  output$8.s1.y.min: 0, (void *)pos_inf
  output$9.s0.x.max: 0, (void *)pos_inf
  output$9.s0.x.min: 0, (void *)pos_inf
  output$9.s0.y.max: 0, (void *)pos_inf
  output$9.s0.y.min: 0, (void *)pos_inf
  output$9.s1.r112$x.max: 3, 3
  output$9.s1.r112$x.min: 0, 0
  output$9.s1.x.max: 0, (void *)pos_inf
  output$9.s1.x.min: 0, (void *)pos_inf
  output$9.s1.y.max: 0, (void *)pos_inf
  output$9.s1.y.min: 0, (void *)pos_inf
  foo$1.s0.x.max: 0, (void *)pos_inf
  foo$1.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$1.s0.x.min: 0, (void *)pos_inf
  foo$1.s0.y.max: 0, (void *)pos_inf
  foo$1.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$1.s0.y.min: 0, (void *)pos_inf
  foo$10.s0.x.max: 0, (void *)pos_inf
  foo$10.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$10.s0.x.min: 0, (void *)pos_inf
  foo$10.s0.y.max: 0, (void *)pos_inf
  foo$10.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$10.s0.y.min: 0, (void *)pos_inf
  foo$11.s0.x.max: 0, (void *)pos_inf
  foo$11.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$11.s0.x.min: 0, (void *)pos_inf
  foo$11.s0.y.max: 0, (void *)pos_inf
  foo$11.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$11.s0.y.min: 0, (void *)pos_inf
  foo$12.s0.x.max: 0, (void *)pos_inf
  foo$12.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$12.s0.x.min: 0, (void *)pos_inf
  foo$12.s0.y.max: 0, (void *)pos_inf
  foo$12.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$12.s0.y.min: 0, (void *)pos_inf
  foo$13.s0.x.max: (void *)neg_inf, (void *)pos_inf
  foo$13.s0.x.min: (void *)neg_inf, (void *)pos_inf
  foo$13.s0.y.max: (void *)neg_inf, (void *)pos_inf
  foo$13.s0.y.min: (void *)neg_inf, (void *)pos_inf
  foo$2.s0.x.max: 0, (void *)pos_inf
  foo$2.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$2.s0.x.min: 0, (void *)pos_inf
  foo$2.s0.y.max: 0, (void *)pos_inf
  foo$2.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$2.s0.y.min: 0, (void *)pos_inf
  foo$3.s0.x.max: 0, (void *)pos_inf
  foo$3.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$3.s0.x.min: 0, (void *)pos_inf
  foo$3.s0.y.max: 0, (void *)pos_inf
  foo$3.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$3.s0.y.min: 0, (void *)pos_inf
  foo$4.s0.x.max: 0, (void *)pos_inf
  foo$4.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$4.s0.x.min: 0, (void *)pos_inf
  foo$4.s0.y.max: 0, (void *)pos_inf
  foo$4.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$4.s0.y.min: 0, (void *)pos_inf
  foo$5.s0.x.max: 0, (void *)pos_inf
  foo$5.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$5.s0.x.min: 0, (void *)pos_inf
  foo$5.s0.y.max: 0, (void *)pos_inf
  foo$5.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$5.s0.y.min: 0, (void *)pos_inf
  foo$6.s0.x.max: 0, (void *)pos_inf
  foo$6.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$6.s0.x.min: 0, (void *)pos_inf
  foo$6.s0.y.max: 0, (void *)pos_inf
  foo$6.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$6.s0.y.min: 0, (void *)pos_inf
  foo$7.s0.x.max: 0, (void *)pos_inf
  foo$7.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$7.s0.x.min: 0, (void *)pos_inf
  foo$7.s0.y.max: 0, (void *)pos_inf
  foo$7.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$7.s0.y.min: 0, (void *)pos_inf
  foo$8.s0.x.max: 0, (void *)pos_inf
  foo$8.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$8.s0.x.min: 0, (void *)pos_inf
  foo$8.s0.y.max: 0, (void *)pos_inf
  foo$8.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$8.s0.y.min: 0, (void *)pos_inf
  foo$9.s0.x.max: 0, (void *)pos_inf
  foo$9.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo$9.s0.x.min: 0, (void *)pos_inf
  foo$9.s0.y.max: 0, (void *)pos_inf
  foo$9.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo$9.s0.y.min: 0, (void *)pos_inf
  foo.s0.x.max: 0, (void *)pos_inf
  foo.s0.x.max.s: (void *)neg_inf, (void *)pos_inf
  foo.s0.x.min: 0, (void *)pos_inf
  foo.s0.y.max: 0, (void *)pos_inf
  foo.s0.y.max.s: (void *)neg_inf, (void *)pos_inf
  foo.s0.y.min: 0, (void *)pos_inf
}
```

rootjalex:
@steven-johnson Do you think you could run Google testing again? I think my tests just never had such enormous expressions, the example you provided should end reasonably fast now.

steven-johnson:
> @steven-johnson Do you think you could run Google testing again? I think my tests just never had such enormous expressions, the example you provided should end reasonably fast now.

Testing now, but hiding an apparently-critical constant (the `count` arg to `substitute_some_lets`) as a default-value argument seems suboptimal. If 16 is a good value for everything, make it internal to the function and name and comment on it. If it's not a good value for everything, don't give it a default value.

steven-johnson:
(Tests look good so far, stand by)

rootjalex:
@steven-johnson Thanks for the feedback, I added documentation explaining the chosen value.

steven-johnson:
I don't see any regressions in Google now, LGTM, land with approval

steven-johnson (approved):

rootjalex:
Thanks @steven-johnson ! @abadams good to go?

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+    if (direction == Direction::Lower) {
+        return Direction::Upper;
+    } else {
+        return Direction::Lower;
+    }
+}
+
+/*
+ * Methods that push divisions or multiplications by constants inwards
+ * These are approximation techniques that approximate in a give direction.
```
abadams, line 40: give -> given

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+            } else {
+                Expr recurse = handle_push_none(op->a, flip(direction));
+                return Div::make(std::move(recurse), op->b);
+            }
+        }
+    }
+    return expr;
+}
+
+// Taken from Simplify_Let.cpp
```
abadams, line 346: Move it into a common file, maybe ExprUsesVar.h/cpp

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+class StripUnboundedTerms : public IRMutator {
+public:
+    // A count of the number of unbounded vars in a given sub expression.
+    int32_t unbounded_vars = 0;
+
+private:
+    // Which direction we are approximating.
+    Direction direction;
+    // Pointer to scope for checking if a variable is bounded or not.
+    const Scope<Interval> *scope_ptr;
```
abadams, line 385: Can this be a const ref instead? Doesn't look like it can be null

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+                    return Div::make(std::move(a), op->b);
+                }
+            }
+            // We could handle the negative case, but that is very uncommon.
+        }
+        return op;
+    }
+
+    Expr visit(const Min *op) override {
+        // If we are trying to Lower bound a Min,
```
abadams, line 498: Lower -> lower

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+        // we merge the lower bounds of the two sides.
+        if (direction == Direction::Lower) {
+            return IRMutator::visit(op);
+        }
+
+        // If we are trying to Upper bound a Min, then we can take either of the two sides of a Min,
```
abadams, line 504: Upper -> upper

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+    CountVarUses(std::map<std::string, int> &_var_uses)
+        : var_uses(_var_uses) {
+    }
+};
+
+/*
+ * Visitor for removing terms that are completely unbounded from
+ * a min or a max.
+ */
+class StripUnboundedTerms : public IRMutator {
```
abadams, line 376: It looks like this class can only correctly handle the listed IR node types, and everything it would make a mess on is explicitly listed below. It seems like this would be safer as a recursive function - when someone adds a new IR node type this class will be implicitly broken and they won't know to update it.

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+    AffineTerm negate(const AffineTerm &term) {
+        return AffineTerm{term.expr, -term.coefficient};
+    }
+
+    // Useful for extracting an affine term from an (x * c) term
+    bool is_mul_by_int_const(const Expr &expr, AffineTerm &term) {
+        const Mul *mul = expr.as<Mul>();
+        if (mul) {
+            if (const IntImm *a = mul->a.as<IntImm>()) {
+                term.coefficient = a->value;
```
abadams, line 662: This should only match is a->value and -(a->value) are representable as int, because you negate it above.

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+
+                if (next.expr.as<Add>() || next.expr.as<Sub>()) {
+                    // After mutation it became an add or sub, throw it back on the pending queue.
+                    pending.push_back(next);
+                } else {
+                    // If this term is a mul by a constant, we can extract out the constant and keep trying.
+                    AffineTerm term;
+                    if (is_mul_by_int_const(next.expr, term)) {
+                        internal_assert(term.expr.defined());
+                        term.coefficient *= next.coefficient;
```
abadams, line 709: Should check for overflow here

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+    inline std::vector<AffineTerm> simplify_summation(const std::vector<AffineTerm> &terms) {
+        // TODO: for now, only do linear-term simplification.
+        // There are other simplifications that can be done
+        // (i.e. pushing terms into a Min or Max to cancel terms),
+        // but we need a good polynomial-time algorithm for it.
+        return simplify_linear_summation(terms);
+    }
+
+    // Two-finger O(n) algorithm for simplifying sums.
+    std::vector<AffineTerm> simplify_linear_summation(const std::vector<AffineTerm> &terms) {
```
abadams, line 736: I don't understand how this linear-time algorithm works. An explanatory comment would help.

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+                return Select::make(op->condition, a, b) + like_terms;
+            }
+        } else {
+            return IRGraphMutator::visit(op);
+        }
+    }
+};
+
+// Used to bound the number of substitutions.
+class SubstituteSomeLets : public IRMutator {
```
abadams, line 925: What does this do on something like: let b = a + a in let c = b + b in let d = c + c in ....

Does it produce an Expr of size 2 ^ count?

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1141 @@
...
+    }
+    return false;
+}
+
+bool should_use_approximate_methods(const Expr &expr) {
+    return !is_const(expr) && expr.type() == Int(32) && possibly_correlated(expr);
+}
+
+Interval approximate_constant_bounds(const Expr &expr, const Scope<Interval> &scope) {
+    Interval interval;
```
abadams, line 1037: Would be more readable if you initialized this to everything or nothing. I don't remember what the default constructor does.

src/ConstantBounds.cpp
rootjalex: Yes, I believe it does, the count only bounds the number of substtitutions, not the accumulated size. Would you prefer different behavior?

src/ConstantBounds.cpp
rootjalex: fixed in cf591f763

src/ConstantBounds.cpp
rootjalex: fixed in cf591f763

src/ConstantBounds.cpp
rootjalex: fixed in cf591f763

src/ConstantBounds.cpp
rootjalex: added in cf591f763 , let me know if it is satisfactory

src/ConstantBounds.cpp
rootjalex: fixed in cf591f763

src/ConstantBounds.cpp
rootjalex: done in bf131e4d0

src/ConstantBounds.cpp
rootjalex: I agree. I will do this, but probably won't be able to today, hopefully tomorrow.

src/ConstantBounds.cpp
abadams: Wait, something looks fishy here. You put values into the scope after mutating them, but then when you pull them out of the scope you mutate them a second time. I can't figure out what this would do with shadowed lets.

IMO we need to bound the actual size of the generated expression. 2^16 is still pretty damn big.

src/ConstantBounds.cpp
abadams: It's sorted by deep equality comparison? I just see a should_commute. should_commute is only a partial order.

rootjalex: No, I think we had a previous discussion about deep_equality being too expensive and should_commute being good enough but we could make it stronger. Should we change that?

src/ConstantBounds.cpp
rootjalex: I think the second mutation shouldn't happen, thanks for the catch.

Do you have any recommendations on how to bound overall size? I guess we could track multiplicative substitutions, i.e. in the example you gave, substituting `c` counts for 3 substitutions, as it has 2 `b` substitutions plus the actual `c` substutiton

rootjalex: Are shadowed lets allowed in Halide syntax? `count_var_uses` definitely assumes that there are no shadowed lets. I thought #6583 was explicitly trying to prevent shadowed lets because we don't allow them.

src/ConstantBounds.cpp
rootjalex: changed in 7a2615444. Would love your thoughts on the `Let` case, that's the one I'm a little unsure on.

src/ConstantBounds.cpp
abadams: Let's change to deep equality and see what the impact is.

src/ConstantBounds.cpp
abadams: Conclusion from offline discussion: 

- This may be called in sliding window which is before removing shadowed lets. 
- Should be feasible to track total # substitutions by including an int in the Scope that counts the number of substitutions that occurred when mutating the RHS of that let

src/ConstantBounds.cpp
```diff
@@ -0,0 +1,1076 @@
...
+            // However, if we are trying to lower bound a Max,
+            // then we can take either of the two sides of a Max,
+            // so remove a side with unbounded variables.
+            return strip_unbounded_minmax(op, direction, scope, var_uses);
+        }
+    } else if (const Let *op = expr.as<Let>()) {
+        // We want to recurse through Lets.
+        auto [v_count, v_new] = strip_unbounded(op->value, direction, scope, var_uses);
+        auto [b_count, b_new] = strip_unbounded(op->body, direction, scope, var_uses);
+        // We might want to only count b_count.
```
abadams, line 527: Yes, this should just be b_count

src/ConstantBounds.cpp
rootjalex: Should we only substitute in the final expression?
i.e. suppose we have: `(let x = w + w in (let y = x + x in (let z = y + y in z + z)))` and we have substition count = 1. Should the resultant Expr be `(let x = w + w in (let y = (w + w) + x in (let z = y + y in z + z)))` or should it be `(let x = w + w in (let y = x + x in (let z = y + y in (y + y) + z)))` or neither? I assume (and prefer) the latter, but I don't think there's a good scheme to make that happen.

Also, if we have the count to do full substitutions, should the resultant Expr be:
`(let x = w + w in (let y = x + x in (let z = y + y in (((w + w) + (w + w)) + ((w + w) + (w + w))) + (((w + w) + (w + w)) + ((w + w) + (w + w))))))` or should the substitutions also happen inside of the let values? i.e. the Expr should become `(let x = w + w in (let y = (w + w) + (w + w) in ....`? I think this also ties into the above question - if we have a count of 1, and the scope for Exprs says that `z = ((w + w) + (w + w)) + ((w + w) + (w + w))` then we can't substitute that in because the tracked cost is too expensive, so perhaps the resultant Expr should be unchanged in this case?

steven-johnson:
Are we hoping that this will allow us to remove the `HL_PERMIT_FAILED_UNROLL` hack?

steven-johnson:
Hey, just a periodic status check on this one.

rootjalex:
Sorry - I'm getting a tad behind, and this PR has been on the back burner for a bit. I will try to get to it in the next few weeks.

steven-johnson:

rootjalex:
It still has a bit of work to be done, and I have not managed to get to it yet. I haven't forgotten it, and will aim to address it by the end of September (I know that's far away and I apologize, but I am currently in paper-writing mode + am about to move across the country)

alexreinking:
@rootjalex—what, conceptually, is missing from this PR? What would it take to bring it up to date?

PR #6806, opened 2022-06-10
Rewrite strided loads of 4 in AlignLoads

vksnk:
This is in addition to the current 1-, 2-, 3- strides. 

This change is motivated by the 4x downsample use-case where I see a significant performance improvement after enabling these.

### DIFF
--- a/src/AlignLoads.cpp
+++ b/src/AlignLoads.cpp
@@ -78,8 +78,8 @@ private:
             // non-constant strides.
             return IRMutator::visit(op);
         }
-        if (!(*const_stride == 1 || *const_stride == 2 || *const_stride == 3)) {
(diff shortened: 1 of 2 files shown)


abadams (approved):

zvookin:
Do you feel this is already adequately covered in tests? If so, all good.

zvookin (approved):

vksnk:
> Do you feel this is already adequately covered in tests? If so, all good.

This pass is only enabled for hvx in the main branch (I'm using it for something else in a different) and I added a check now into simd_op_check_hvx (similar to what was there for stride-3).

vksnk:
Test failure is unrelated (correctness_custom_cuda_context), so will submit.

PR #6807, opened 2022-06-13
Fix two minor bugs triggered by an or reduction with early-out

abadams:
I was trying out using a where clause in a big or reduction to get early-out behavior and found two bugs that this PR fixes.

First, if there's a param that's only used in a specialization of a wrapper func, it isn't picked up by InferArguments.

Second, a variable-free condition that feeds into bounds inference is assumed to depend only on parameters, which is a bad assumption because it could also be a scalar Func. 

Both of these manifested as compilation failures.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -2640,7 +2640,7 @@ private:
     void visit(const IfThenElse *op) override {
         TRACK_BOXES_TOUCHED;
         op->condition.accept(this);
-        if (expr_uses_vars(op->condition, scope)) {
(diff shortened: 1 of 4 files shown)


vksnk (approved):

alexreinking:
clang-tidy complains:
```
InferArguments.cpp:187:19: error: loop variable is copied but only used as const reference; consider making it a const reference [performance-for-range-copy,-warnings-as-errors]
        for (auto p : func.wrappers()) {
                  ^
             const  &
```

abadams:
Failure unrelated.

Issue #6801, opened 2022-06-04
Internal representation for misaligned loads


LebedevRI:
Now that the syntactic sugar is mostly a solved problem (https://github.com/halide/Halide/pull/6775),
i'm starting to look at the second part of the problem - actually noticing the wide load.

And now that i think about it, that's where the big problem is. The `Load` expr
takes **element index**, so given a load of `i8` at an 'opaque' index `i`,
e.g. we can't actually represent that as low 8 bits of a load of `i16` at index `i/2`,
because `i` might not be a multiple of 2...

Now, as i see it there are 3 solutions:
1. Add a new `MisalignedLoad` expr
2. Change `Internal::Load` to store the byte offset instead of element index.
3. give up completely

The 1. is easy to prototype, but then basically nothing (that supports `Internal::Load`) will support it,
or we end up with huge code duplication and poor test coverage;
while 2. is rather invasive (which should be fine since it's in `Internal` namespace, right? right? :))
but is certainly a more general solution.

I can do either, and would prefer 2. if there's buy-in, but i'd like to hear some thoughts first.
CC @zvookin @abadams @alexreinking @steven-johnson 

Refs. https://github.com/halide/Halide/issues/6756

abadams:
Changing the load node to take byte offsets instead of element offsets would break far too much to be viable. Don't we only want to do this for element-aligned loads anyway? We have pretty good modulus/remainder infrastructure to check if something is a multiple of 2.

LebedevRI:
> Changing the load node to take byte offsets instead of element offsets would break far too much to be viable.

Right. Too much code in this repo, or outside of this repo?
I'm still trying to understand how different is the treatment of the downstreams as compared to what LLVM does.
I'm somewhat fearful of adding another representation for the same concept, see LLVM's intrinsic's zoo.

I agree that "fix everything at once" forward migration strategy is problematic,
and _without thinking much_, it seems like a mitigation strategy to stage this piece-wise is possible:
1. Step 1 (single commit)
   1. rename `index` field to `impl_index`
   2. add an `Expr index() { return impl_index; }`
   3. migrate all now-compile-time-broken uses of `index` field to calls to `index()` function.
2. Step 2 (another single commit)
   1. rename `impl_index` to `offset`
   2. adjust constructor to multiply index by the element size
   3. add another constructor from `offset` (disambiguate to avoid confusion, e.g. https://godbolt.org/z/34r6nGTbn)
[...]

PR #6809, opened 2022-06-14
[Codegen_LLVM] Directly load scalar that we'd load as vector and rein…

LebedevRI:
…terpret

Second (of three) pieces of the load widening puzzle.
Here, the codegen is taught to directly emit scalar loads,
instead of doing vector load and `bitcast`ing it.

Refs. https://github.com/halide/Halide/issues/6801
Refs. https://github.com/halide/Halide/issues/6756
Refs. https://github.com/halide/Halide/pull/6775

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -2526,6 +2526,50 @@ void CodeGen_LLVM::codegen_atomic_rmw(const Store *op) {
     }
 }
 
+// Given a `Call::reinterpret` to \p dst type of the `Load` \p l,
+// can we instead perform a load of \p dst type?
+// If so, returns (scalar!) index expression.
+Expr CodeGen_LLVM::should_load_scalar(const Load *l, Type dst) {
+    // For now, simply do not deal with predicated loads at all.
+    // FIXME: we could support scalar/splat predicate though.
+    if (!is_const_one(l->predicate)) {
+        return Expr();
+    }
+
+    Type src = l->type;
+
+    // We are interested in the case where we load a vector, and cast to scalar.
+    if (src.lanes() == 1 || dst.lanes() != 1) {
+        return Expr();
+    }
+
+    // The index must be produced by a `Ramp`.
+    const Ramp *r = l->index.as<Ramp>();
+    if (!r) {
+        return Expr();
+    }
+
+    // The `Ramp` must have scalar base pointer, and be dense (stride=1).
+    // FIXME: if we are loading a vector of i8, stride=-1 is fine (need bswap).
+    if (!is_const_one(r->stride) || r->base.type().lanes() != 1) {
+        return Expr();
+    }
+
+    // The `Call::reinterpret` must effectively be a no-op.
+    if (dst.bits() * dst.lanes() != src.bits() * src.lanes()) {
+        return Expr();
+    }
+
+    // Neither of of the types must need an upgrade.
+    if (upgrade_type_for_storage(src) != src ||
+        upgrade_type_for_storage(dst) != dst) {
+        return Expr();
+    }
+
+    // Awesome! Note that the index is for the `Load`'s current element type!
+    return r->base;
+}
+
 void CodeGen_LLVM::visit(const Call *op) {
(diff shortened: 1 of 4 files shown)


test/correctness/bitcast_vector_load_into_scalar.cpp
```diff
@@ -0,0 +1,140 @@
...
+        success &= test_with_n_times_chunk_type<CHUNK_TYPE>(num_chunks, t);
+    }
+
+    return success;
+}
+
+bool test_all(Target t) {
+    bool success = true;
+
+    success &= test_with_chunk_type<uint8_t>(t);
```
LebedevRI, line 121: There is some kind of failure to simplify for the i8 chunk type somewhere in halide:
```
   let t5 = b0[(store.min.0 + store.s0.x.rebased)*2]
   store[store.s0.x.rebased] = uint64((uint16)reinterpret(ramp(t5, b0[((store.min.0 + store.s0.x.rebased)*2) + 1] - t5, 2)))
```
but suddenly it's fine for larger types:
```
   store$1[store$1.s0.x.rebased] = uint64((uint32)reinterpret(b3[ramp((store$1.min.0 + store$1.s0.x.rebased)*4, 1, 4) aligned(4, 0)]))
```

LebedevRI (commented):
This only has basic (stride=+1/-1) correctness test coverage,
but no tests that would ensure that the load actually got widened.
If we'd do that at halide ir level, it would be obvious to check,
but here i'm not really sure how to do that.

src/CodeGen_LLVM.cpp
```diff
@@ -2582,6 +2626,19 @@ void CodeGen_LLVM::visit(const Call *op) {
...
         llvm::Type *llvm_dst = llvm_type_of(dst);
+        if (const Load *l = op->args[0].as<Load>()) {
+            if (Expr index = should_load_scalar(l, dst); index.defined()) {
+                Value *ptr =
+                    codegen_buffer_pointer(l->name, src.element_of(), index);
+                // Pessimistically assume only scalar element alignment.
+                LoadInst *load = builder->CreateAlignedLoad(
+                    llvm_dst, ptr, llvm::Align(l->type.bytes()));
+                // FIXME: can we emit better TBAA for constant indexes here?
+                add_tbaa_metadata(load, l->name, /*index=*/Expr());
```
abadams, line 2637: I think it's correct to use the original load index here, because the tbaa metadata is all in terms of the allocated type before the reinterpret.

abadams:
The implementation makes sense, aside from my concerns about using the reinterpret intrinsic to change vector lanes, but the test is not great - We don't use vector types in front-end code, and I have no idea what might break if we did. I guess it's OK as a temporary measure, though the weird failure with uint8s is alarming and might indicate that some assumption is being violated somewhere.

src/CodeGen_LLVM.cpp
LebedevRI: Ah no, wait, of course it is, because we really don't change which bytes we load.

LebedevRI:
> The implementation makes sense, aside from my concerns about using the reinterpret intrinsic to change vector lanes, but the test is not great - We don't use vector types in front-end code, and I have no idea what might break if we did.

Right. Well, i'm not sure how else to write that test given what's currently available :)

> the weird failure with uint8s is alarming and might indicate that some assumption is being violated somewhere.

It's some missed simplification. In good case, it is simplified during `Lowering after second simplifcation:`,
but clearly not in 2xi8 case...

LebedevRI:
I guess, `reinterpret` intrinsic first needs to be promoted into a first-class IR node.

steven-johnson:
Is this PR still active?

LebedevRI:
> Is this PR still active?

Yes. The rough queue is:
* fixing Deinterleave (PR almost ready)
* Implementing Reinterpret support in Deinterleave
* Finally changing Reinterpret semantics
* ???
* Adding concat_bits/extract_bits intrinsics
* Teaching vectorizer to handle concat_bits/extract_bits intrinsics nicely
* #6775
* #6809

PR #6810, opened 2022-06-20
[CMake] Mark multi-threaded tests as such

LebedevRI:
When i originally run halide tests for the first few times during debian packaging,
it caused hard computer reset, i guess due to the CPU current safety cut-off.
That never happened before/elsewhere, and i think i resolved it on my end,
but ever since, i've essentially been running halide tests one at the time (`-j1`):
https://salsa.debian.org/pkg-llvm-team/halide/-/blob/4fba9ca9622b9c4b268d9bdbfd1cae47354b9ea7/debian/rules#L133-135

This is rather not great. I see that performance tests are sequential already,
but i would like to go a bit further, and be able to differentiate which tests
are simple in the sense that they do not spawn any internal threads,
and which do spawn threads.

This does just that. I've added an `abort();` into `halide_spawn_thread()`,
and marked all tests that now failed as `multithreaded`.
I'm not *sure* this caught everything, but it seems like that's the most of it.

Note that i have completely ignored GPU tests, since that story on AMD is rather sad :/

Is this something that is acceptable?


src/autoschedulers/adams2019/CMakeLists.txt
```diff
@@ -117,7 +117,7 @@ if (BUILD_SHARED_LIBS)
              COMMAND test_apps_autoscheduler $<TARGET_FILE:Halide_Adams2019>)
 
     set_tests_properties(test_apps_autoscheduler PROPERTIES
-                         LABELS Adams2019
```
alexreinking, line 120: These should be semicolon-delimited.

cmake/HalideTestHelpers.cmake
```diff
@@ -49,9 +49,15 @@ function(add_halide_test TARGET)
              COMMAND ${args_COMMAND} ${args_ARGS}
              WORKING_DIRECTORY "${args_WORKING_DIRECTORY}")
 
+    # If the test is not marked as being multi-threaded,
+    # limit it to one thread, as requested.
+    if (NOT "multithreaded" IN_LIST args_GROUPS)
+        set(MAYBE_FORCE_NO_THREADING "HL_NUM_THREADS=1")
+    endif()
+
```
alexreinking, line 57: I think we don't want this here. With the labels in place, you can always run:

```
$ HL_NUM_THREADS=1 ctest -LE multithreaded 
$ ctest -L multithreaded
```

But with this here, you _can't_ run the unmarked tests with multiple threads.

LebedevRI: > With the labels in place, you can always run:
> 
> ```
> $ HL_NUM_THREADS=1 ctest -LE multithreaded 
> $ ctest -L multithreaded
> ```

Hm, that's a good point, thanks!

> you can't run the unmarked tests with multiple threads.

That's actually The point, to ensure that there are no false-negatives ("uses threads but not marked as such"),
by essentially requiring to explicitly opt-in into using threads.

I can drop this of course, but i think this is an important part to ensure lack of regressions.

alexreinking (approved):

LebedevRI:
Heh, *all* bots being green is a rare sight.

alexreinking:
Hmm, are the Windows bots not running right now?

steven-johnson:
They seem to have broken over the weekend, back up now

steven-johnson:
looks green

LebedevRI:
Thank you!
If further 14.x point releases are planned, it might be nice to backport this.

PR #6815, opened 2022-06-24
Halide pip package for python bindings

lukastruemper:
@alexreinking Ahh, I had a way too long day. Cleaned up some repos after work and deleted the halide fork with original PR https://github.com/halide/Halide/pull/6794


python_bindings/src/PyHalide.cpp
```diff
@@ -31,7 +31,7 @@ static_assert(PY_VERSION_HEX >= 0x03000000,
               "We appear to be compiling against Python 2.x rather than 3.x, which is not supported.");
 
 #ifndef HALIDE_PYBIND_MODULE_NAME
-#define HALIDE_PYBIND_MODULE_NAME halide
+#define HALIDE_PYBIND_MODULE_NAME halide_
```
steven-johnson, line 34: Why this change?

python_bindings/src/halide/__init__.py
```diff
@@ -0,0 +1 @@
+from .halide_ import *
```
steven-johnson, line 1: apologies, but I don't recognize the lead-with-a-dot syntax, what does that indicate?

steven-johnson:
As a meta-comment here: we will (eventually) need to either augment the Makefile to do this as well, or (preferably) nuke Make support for the Python bindings. (I already did the latter once but it got restored). The build rules for (waves hands, looks around) are just way too complex to have to maintain in tandem unless there is a *compelling* need for them, and I argue that there isn't.

alexreinking:
> As a meta-comment here: we will (eventually) need to either augment the Makefile to do this as well, or (preferably) nuke Make support for the Python bindings. (I already did the latter once but it got restored). The build rules for (waves hands, looks around) are just way too complex to have to maintain in tandem unless there is a _compelling_ need for them, and I argue that there isn't.

I buy this argument and favor dropping Make support for this. I have no inclination to maintain the Makefiles for building the Python package. Does anyone else on the team?

lukastruemper:
> I will need to make some edits to this that are easier to just do than to explain. Once I make those changes, I'll have you review the PR with my implicit sign-off @steven-johnson

Oh wow, I didn't know what I started by adding a setup.py haha :-D

cmake/FindHalide.cmake
```diff
@@ -0,0 +1,2 @@
+message(VERBOSE " Spoofing find_package(Halide) since in-tree builds already have Halide available.")
```
LebedevRI, line 1: Why not instead only look for halide conditionally iff the python_bindings/ is the root cmake dir?

alexreinking:
Still trying to figure out how best to restore the Python tests/stubs now that there's an actual Python module in the way. I can think of two options:

1. Change the test code to run directly against the native `halide_` module.
2. Run the tests in tox, which will fully build the 
3. Find some way to recreate the final package structure in the CMake _build_ tree (easier said than done).

(1) is more expedient and (2) is more robust. However (1) plays nicer with CTest/existing infrastructure and the best (2) can do is be called from a shim in CTest. On the other hand, (3) would be a good compromise, but I've had a couple false starts thanks to multi-configuration generators. Might need to wait until after I move into my new house on Friday.

steven-johnson:
Just touching base here, roughly how much more work is needed for this?

alexreinking:
Closing in favor of #6886 since `main` has diverged substantially

PR #6823, opened 2022-06-28
Tweak python apps for better Blaze/Bazel compatibility

steven-johnson:
- Don't write to current directory (rely on an env var to say where to write)
- Don't read from arbitrary absolute paths (again, rely on an env var)
- Drive-by removal of unnecessary #include in Codegen_LLVM.cpp inside a lambda (!)


python_bindings/apps/CMakeLists.txt
```diff
@@ -5,11 +5,13 @@ set(SCRIPTS
     interpolate.py
     local_laplacian.py)
 
+make_shell_path(TEST_IMAGES_DIR $<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/../../apps/images>)
```
alexreinking, line 8: For a single directory, you don't need to use `make_shell_path`. Once we bump the minimum CMake version to 3.22, we can get rid of the helper altogether! 😄

alexreinking: Also this list of environment variables is getting long enough to split onto multiple lines.

```suggestion
set(test_env
    "PYTHONPATH=$<SHELL_PATH:$<TARGET_FILE_DIR:Halide::Python>>"
    "HL_TARGET=${Halide_TARGET}"
    "TEST_TMPDIR=$<SHELL_PATH:${CMAKE_CURRENT_BINARY_DIR}>"
    "TEST_IMAGES_DIR=$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/../../apps/images>"
)
```

python_bindings/apps/CMakeLists.txt
```diff
@@ -5,11 +5,13 @@ set(SCRIPTS
...
+
 foreach (SCRIPT IN LISTS SCRIPTS)
     get_filename_component(BASE ${SCRIPT} NAME_WE)
     add_test(NAME python_apps_${BASE}
              COMMAND Python3::Interpreter "$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/${SCRIPT}>")
     set_tests_properties(python_apps_${BASE} PROPERTIES
                          LABELS python
-                         ENVIRONMENT "PYTHONPATH=$<SHELL_PATH:$<TARGET_FILE_DIR:Halide::Python>>;HL_TARGET=${Halide_TARGET}")
+                         ENVIRONMENT "PYTHONPATH=$<SHELL_PATH:$<TARGET_FILE_DIR:Halide::Python>>;HL_TARGET=${Halide_TARGET};TEST_TMPDIR=${CMAKE_CURRENT_BINARY_DIR};TEST_IMAGES_DIR=${TEST_IMAGES_DIR}")
```
alexreinking, line 16:
```suggestion
                         ENVIRONMENT "${test_env}")
```

python_bindings/apps/bilateral_grid.py
```diff
@@ -102,7 +102,7 @@ def generate_compiled_file(bilateral_grid):
 
 
 def get_input_data():
-    image_path = os.path.join(os.path.dirname(__file__), "../../apps/images/rgb.png")
+    image_path = os.path.join(os.environ["TEST_IMAGES_DIR"], "rgb.png")
```
alexreinking, line 105: Should this really throw when not set? Why shouldn't it default to the old hardcoded value? Same with everywhere else that now reads an env-var.

At the very least, checking for a `KeyError` early on and printing a helpful error message would be nicer than "BZZT uncaught KeyError!"

python_bindings/tutorial/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set(TESTS
     )
 
 make_shell_path(PYTHONPATH "$<TARGET_FILE_DIR:lesson_10_halide>" "$<TARGET_FILE_DIR:Halide::Python>")
+make_shell_path(TEST_IMAGES_DIR $<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/../../tutorial/images>)
```
alexreinking, line 20:
```suggestion
set(test_env
    "HL_TARGET=${Halide_TARGET}"
    "TEST_TMPDIR=$<SHELL_PATH:${CMAKE_CURRENT_BINARY_DIR}>"
    "TEST_IMAGES_DIR=$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/../../tutorial/images>"
)
```

python_bindings/tutorial/CMakeLists.txt
```diff
@@ -25,7 +26,7 @@ foreach (TEST IN LISTS TESTS)
 
     set_tests_properties(python_tutorial_${TEST_NAME} PROPERTIES
                          LABELS python
-                         ENVIRONMENT "PYTHONPATH=${PYTHONPATH};HL_TARGET=${Halide_TARGET}")
+                         ENVIRONMENT "PYTHONPATH=${PYTHONPATH};HL_TARGET=${Halide_TARGET};TEST_TMPDIR=${CMAKE_CURRENT_BINARY_DIR};TEST_IMAGES_DIR=${TEST_IMAGES_DIR}")
```
alexreinking, line 29:
```suggestion
                         ENVIRONMENT "${test_env}")
```

python_bindings/tutorial/lesson_11_cross_compilation.py
```diff
@@ -18,6 +19,8 @@ def main():
     brighter = hl.Func("brighter")
     x, y = hl.Var("x"), hl.Var("y")
 
+    tmpdir = os.environ["TEST_TMPDIR"]
```
alexreinking, line 22: For example:
```suggestion
    if not (tmpdir := os.getenv('TEST_TMPDIR')):
        sys.exit('Must set environment variable TEST_TMPDIR to a suitable temporary directory')
```
Although I forget when Python added the `:=` walrus operator...

python_bindings/tutorial/lesson_11_cross_compilation.py
```diff
@@ -93,7 +96,7 @@ def main():
                                 1]       # Current version of elf
 
         length = len(arm_32_android_magic)
-        f = open("lesson_11_arm_32_android.o", "rb")
+        f = open(os.path.join(tmpdir, "lesson_11_arm_32_android.o"), "rb")
```
alexreinking, line 99: If tmpdir were a `Path` variable (from the standard `pathlib` module) then you could just write: `tmpdir / "lesson_11_arm_32_android.o"`

alexreinking (changes requested):
Not gonna die on any of these hills, but thought they were worth a little discussion, anyway.

python_bindings/apps/bilateral_grid.py
steven-johnson: Because it should always be set.

python_bindings/apps/CMakeLists.txt
steven-johnson: Done

python_bindings/tutorial/lesson_11_cross_compilation.py
steven-johnson: goo goo goo joob, I think the exception itself will be adequate for anyone to figure it out.

python_bindings/tutorial/lesson_11_cross_compilation.py
steven-johnson: there is zero other usage of that in our current python; if we want to convert, let's do it all at once, not piecewise.

steven-johnson:
PTAL

abadams:
These sorts of changes make the tutorials incrementally more robust, but incrementally less useful as tutorials. Code in the tutorial that isn't the point of the tutorial needs to be brutally cut to the bare minimum, and any appeasement of build systems to make the tutorials easily testable should be done in the build system.

python_bindings/tutorial/lesson_02_input_image.py
```diff
@@ -18,7 +18,7 @@ def main():
     # brightens an image.
 
     # First we'll load the input image we wish to brighten.
-    image_path = os.path.join(os.path.dirname(__file__), "../../tutorial/images/rgb.png")
+    image_path = os.path.join(os.environ["TEST_IMAGES_DIR"], "rgb.png")
```
abadams, line 21: The intended usage for these tutorials is someone typing "python lesson_02_input_image.py" at the shell. Won't this break that?

python_bindings/apps/bilateral_grid.py
alexreinking: @abadams - I raised your same concern here.

steven-johnson:
> These sorts of changes make the tutorials incrementally more robust

Good point. Let me see what we did for the C++ tutorials for appeasement.

steven-johnson:
> Good point. Let me see what we did for the C++ tutorials for appeasement.

So rhe short answer is: nothing, apparently. We don't even build them in the Blaze build, much less test them. This saddens me, but I can live with it, so I'll back out the changes to tutorials here and just stop trying to make the Python tutorials work too.

The apps are a different story. We build/test relatively few of the C++ apps in Blaze (mainly because rewriting the build rules hasn't been deemed important); for Python, the build rules are often much simpler so it's easier to add the rules, and more coverage is good.

I'll update this PR to:
- remove changes to tutorials entirely
- make the apps actually conditionalize using the env vars, so that they can work robustly either way

 Stay tuned.

alexreinking (approved):

abadams (approved):

PR #6824, opened 2022-06-28
Update presets to format version 3

alexreinking:
This lets VSCode work out of the box, at least in my environment. Before, it would complain about not supporting version-1 presets. It should be enough to run the GHA bots for this change, since the only thing using presets is the Ubuntu packaging (which should really be removed).


steven-johnson:
I have yet to ever try VSCode, and don't have time to install and try out this week. Happy to LGTM as-is, but if anyone else uses it, they might be a better approver.

alexreinking:
> Happy to LGTM as-is, but if anyone else uses it, they might be a better approver.

I'm sure I'm the first since someone would have opened a bug report about this, lol

PR #6827, opened 2022-06-28
Better lowering of halving_sub and rounding_halving_add

abadams:
Previously, lower_halving_sub and lower_rounding_halving_add both used 9
ops. This change redirects halving_sub to use rounding_halving_add, and
redirects rounding_halving_add to use halving_add. In the case that none
of these instructions exist natively, this reduces it to 7/8 ops for
signed/unsigned halving sub and 6 ops for rounding halving add.

More importantly, this lets halving_sub make use of pavgw/b on x86 to
reduce it to 3 ops for u8 and u16 inputs.

Edit: You can make use of pavgw/pavgb for signed rounding averages too, 
but this is very x86-specific.


rootjalex (commented):
I think need to be careful about wrapping semantics, probably should have a special case for int32

src/FindIntrinsics.cpp
```diff
@@ -894,14 +894,17 @@ Expr lower_halving_add(const Expr &a, const Expr &b) {
...
+        return e;
+    }
 }
 
-// TODO: These should using rounding_shift_right, but lowering that
-// results in double widening and the simplifier doesn't fix it.
 Expr lower_rounding_halving_add(const Expr &a, const Expr &b) {
     internal_assert(a.type() == b.type());
-    return (a >> 1) + (b >> 1) + (((a & 1) + (b & 1) + 1) >> 1);
+    return halving_add(a, b) + ((a ^ b) & 1);
```
abadams, line 907: This add can't overflow, because a and b can't simultaneously be the max value if (a^b) is high

src/FindIntrinsics.cpp
```diff
@@ -894,14 +894,17 @@ Expr lower_halving_add(const Expr &a, const Expr &b) {
 
 Expr lower_halving_sub(const Expr &a, const Expr &b) {
     internal_assert(a.type() == b.type());
-    return (a >> 1) - (b >> 1) - (((b & 1) - (a & 1) + 1) >> 1);
+    Expr e = rounding_halving_add(a, ~b);
+    if (a.type().is_uint()) {
+        return e + make_const(e.type(), (uint64_t)1 << (a.type().bits() - 1));
```
abadams, line 899: This add can overflow, but that's fine because it's in a uint branch

src/CodeGen_X86.cpp
```diff
@@ -523,6 +523,18 @@ void CodeGen_X86::visit(const Call *op) {
...
+               op->type.bits() <= 16 &&
+               op->is_intrinsic(Call::rounding_halving_add)) {
+        // We can redirect signed rounding halving add to unsigned rounding
+        // halving add by adding 128 / 32768 to the result if the sign of the
+        // args differs.
+        internal_assert(op->args.size() == 2);
+        Type t = op->type.with_code(halide_type_uint);
+        Expr a = cast(t, op->args[0]);
+        Expr b = cast(t, op->args[1]);
+        codegen(rounding_halving_add(a, b) + ((a ^ b) & (1 << (t.bits() - 1))));
```
abadams, line 536: This add is of a uint too, but I forgot to cast the result back to signed. Will fix.

src/FindIntrinsics.cpp
vksnk, line 899: It's a bit hard to figure out why this works (took me a while), might be useful to have a comment to explain it (I got something like this, but maybe there is an easier way:

Assuming type is uint:

 ~b = 2 ** max_bits - b - 1

then 

rounding_halving_add(a, ~b) + 2 ** (max_bits -  1) = (a + (2 ** max_bits - b - 1) + 1) / 2 +  2 ** (max_bits -  1) = (a  - b + (2 ** max_bits)) / 2  +  2 ** (max_bits -  1) = (a - b) / 2  +  2 ** (max_bits -  1)   +  2 ** (max_bits -  1)  = (a - b) / 2  +  2 ** max_bits = (a - b) / 2  

).

abadams: Good point. I added a comment.

How I actually discovered it was wondering if rounding_halving_add(x, ~y) was related to halving_sub(x, y), and noticing it was just off by 128 exactly.

vksnk (approved):

rootjalex (approved):
Lgtm

PR #6829, opened 2022-06-29
[vulkan phase0] Add adts for containers and memory allocation to runtime

derek-gerstmann:
This set of changes is Phase0 for adding Vulkan support. Additional PRs will include Phase1 (adding SPIRV-IR), Phase2 (Vulkan RT updates), and Phase3 (tests and CI changes).

This Phase0 is a standalone set of commits for adding a few basic data structures to the Halide runtime internals that I found useful when bringing up the Vulkan runtime.  Specifically, it adds the following header only class interfaces:

src/runtime/internal/block_allocator.h
src/runtime/internal/block_storage.h
src/runtime/internal/linked_list.h
src/runtime/internal/memory_arena.h
src/runtime/internal/memory_resources.h
src/runtime/internal/pointer_table.h
src/runtime/internal/region_allocator.h
src/runtime/internal/string_storage.h
src/runtime/internal/string_table.h

These are intently written to be minimal, and only depend on cstdlib headers so they can be used inside the Halide runtime library.  To avoid template usage, they manage byte allocations and coerce types as necessary. 

Unit tests have been added under:

test/runtime/*.cpp

Both CMake and Make project files have been updated to build and run these tests as requested.


test/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    target_link_libraries(runtime_internal_${NAME} PRIVATE Halide::Test)
+    target_include_directories(runtime_internal_${NAME} PRIVATE "${Halide_SOURCE_DIR}/src")
+
+    target_link_libraries(runtime_internal_${NAME} PRIVATE Halide::Runtime)
+    set_target_properties(runtime_internal_${NAME} PROPERTIES ENABLE_EXPORTS TRUE)
+    target_compile_options(runtime_internal_${NAME} PUBLIC -Wno-builtin-declaration-mismatch )
+    target_compile_definitions(runtime_internal_${NAME} PUBLIC HALIDE_VERSION=${Halide_VERSION}  )
+    target_compile_definitions(runtime_internal_${NAME} PUBLIC HALIDE_VERSION_MAJOR=${Halide_VERSION_MAJOR} )
+    target_compile_definitions(runtime_internal_${NAME} PUBLIC HALIDE_VERSION_MINOR=${Halide_VERSION_MINOR} )
+    target_compile_definitions(runtime_internal_${NAME} PUBLIC HALIDE_VERSION_PATCH=${Halide_VERSION_MINOR} )
```
alexreinking, line 12:
```suggestion
    target_compile_definitions(runtime_internal_${NAME} PUBLIC HALIDE_VERSION_PATCH=${Halide_VERSION_PATCH} )
```

alexreinking: Also, you can place multiple definitions in `target_compile_definitions`:
```
target_compile_definitions(
    runtime_internal_${NAME}
    PRIVATE
    HALIDE_VERSION=${Halide_VERSION}
    HALIDE_VERSION_MAJOR=${Halide_VERSION_MAJOR}
    HALIDE_VERSION_MINOR=${Halide_VERSION_MINOR}
    HALIDE_VERSION_PATCH=${Halide_VERSION_PATCH}
    COMPILING_HALIDE_RUNTIME
    COMPILING_HALIDE_RUNTIME_TESTS
)
```
Also, the visibility here should be `PRIVATE`. Since there are executables, it's just a matter of style, but setting them to `PUBLIC` is kind of a head-scratcher.

test/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+function(halide_define_runtime_internal_test NAME)
+    add_executable(runtime_internal_${NAME} ${NAME}.cpp)
+    target_link_libraries(runtime_internal_${NAME} PRIVATE Halide::Test)
+    target_include_directories(runtime_internal_${NAME} PRIVATE "${Halide_SOURCE_DIR}/src")
+
+    target_link_libraries(runtime_internal_${NAME} PRIVATE Halide::Runtime)
+    set_target_properties(runtime_internal_${NAME} PROPERTIES ENABLE_EXPORTS TRUE)
+    target_compile_options(runtime_internal_${NAME} PUBLIC -Wno-builtin-declaration-mismatch )
```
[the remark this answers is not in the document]
alexreinking: Again, `PUBLIC` visibility here is odd. Should all be `PRIVATE`.

test/runtime/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@ function(halide_define_runtime_internal_test NAME)
 
     target_link_libraries(runtime_internal_${NAME} PRIVATE Halide::Runtime)
     set_target_properties(runtime_internal_${NAME} PROPERTIES ENABLE_EXPORTS TRUE)
-    target_compile_options(runtime_internal_${NAME} PUBLIC -Wno-builtin-declaration-mismatch )
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        target_compile_options(runtime_internal_${NAME} PUBLIC -Wno-builtin-declaration-mismatch )
+    endif()
```
alexreinking, line 10:
```suggestion
```
[This is not a valid Clang flag](https://gcc.godbolt.org/z/nnrvs6E19) and will print its own "warning: unknown warning option" warnings. We should really not be in the business of maintaining warning flag compatibility tables. Why are these warnings spurious, anyway?

test/runtime/CMakeLists.txt
derek-gerstmann: Ah, good catch! Yeah there's no reason they should be PUBLIC ... I'll set everything to PRIVATE.

derek-gerstmann:
Hrmm ... looks like halide-testbranch-main-llvm15-x86-64-windows-cmake fails on "correctness_atomics" ... that's unrelated to anything in this commit. =/

steven-johnson:
Windows builds have numerous flakes, unfortunately; some are due to outdated build machines (winbot1) and others are, well, flakes. We should really run them all to the ground, but resources are limited.

src/runtime/internal/block_allocator.h
```diff
@@ -0,0 +1,477 @@
...
+    allocators = ma;
+    block_list.initialize(user_context,
+                          sizeof(BlockResource),
+                          config.initial_capacity,
+                          allocators.system);
+}
+
+MemoryRegion *BlockAllocator::reserve(void *user_context, const MemoryRequest &request) {
+
+    debug(nullptr) << "BlockAllocator: Reserve ("
```
steven-johnson, line 146: Here and elsewhere: when we actually have `user_context` available (as we do here), `debug(user_context)` is what you should do; `debug(nullptr)` is only for places that no user_context could possibly be available, e.g. static ctors or dtors

src/runtime/internal/block_allocator.h
```diff
@@ -0,0 +1,477 @@
...
+    return block_list.size();
+}
+
+// --
+
+}  // namespace Internal
+}  // namespace Runtime
+}  // namespace Halide
+
+#endif  // HALIDE_RUNTIME_BLOCK_ALLOCATOR_H
```
steven-johnson, line 477: nit, here and elsewhere: all source files should end with a newline

src/runtime/internal/memory_arena.h
```diff
@@ -0,0 +1,312 @@
...
+MemoryArena::MemoryArena(void *user_context,
+                         const Config &cfg,
+                         const SystemMemoryAllocatorFns &alloc)
+    : config(cfg),
+      blocks(user_context, {sizeof(MemoryArena::Block), 32, 32}, alloc) {
+    halide_debug_assert(user_context, config.minimum_block_capacity > 1);
+}
+
+MemoryArena::~MemoryArena() {
+    destroy(nullptr);
```
steven-johnson, line 97: This troubles me; either the code needs user_context to be safe (in which case we can't pass nullptr here), or it doesn't (in which case, why is passing it even an option)?

test/CMakeLists.txt
```diff
@@ -39,4 +39,15 @@ if (WITH_TEST_GENERATOR)
     add_subdirectory(generator)
 endif ()
 
+if(MSVC)
+    set(ENABLE_RUNTIME_TESTS OFF)
```
steven-johnson, line 43: Why can't we do the runtime tests under MSVC?

src/runtime/internal/block_allocator.h
derek-gerstmann: Good catch!  Thanks ... I'll fix these occurrences!

src/runtime/internal/block_allocator.h
```diff
@@ -0,0 +1,477 @@
...
+
+    BlockResource *block = static_cast<BlockResource *>(block_entry->value);
+    halide_abort_if_false(user_context, block != nullptr);
+    halide_abort_if_false(user_context, block->allocator != nullptr);
+
+    MemoryRegion *result = reserve_memory_region(user_context, block->allocator, request);
+    if (result == nullptr) {
+
+        size_t actual_size = constrain_requested_size(request.size);
+        //        debug(0) << "BlockAllocator: No free blocks found! Allocating new empty block of size ("
```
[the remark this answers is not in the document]
derek-gerstmann: Agreed!  I'll remove these commented out statements ....

src/runtime/internal/block_allocator.h
```diff
@@ -0,0 +1,477 @@
...
+        if (allocator->collect(user_context)) {
+            result = allocator->reserve(user_context, request);
+        }
+    }
+    return result;
+}
+
+BlockAllocator::BlockEntry *
+BlockAllocator::find_block_entry(void *user_context, const MemoryProperties &properties, size_t size, bool dedicated) {
+
```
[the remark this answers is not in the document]
derek-gerstmann: Ooops!

test/CMakeLists.txt
derek-gerstmann: Because the runtime tests include src/runtime/runtime_internal.h which was written to only support clang (GCC's front end is close enough it works fine as well).  I originally setup the tests to compile with clang (in the same way as the actual runtime bitcode files), but that wasn't very clean and didn't integrate well with the other tests, so we switched to just using the native system compiler.  We could certainly add MSVC compatibility to runtime_internal.h and ifdef the attributes and types that are causing compile errors, but it seemed outside the scope of this PR.

src/runtime/internal/memory_resources.h
```diff
@@ -0,0 +1,280 @@
...
+
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+
+// --
+
+// Hint for allocation usage indicating whether or not the resource
+// is in use, available, or dedicated (and can't be split or shared)
+enum AllocationStatus {
```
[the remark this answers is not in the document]
derek-gerstmann: Good point ... I'll change these.

test/CMakeLists.txt
steven-johnson: Fair, maybe worth capturing this in a comment there for posterity

alexreinking: I would actually print something to the configure log via `message(STATUS)`

Also, a user can still write `-DWITH_TEST_RUNTIME=YES` at the command line with MSVC and enable these tests. Is that intentional? If not, you might want to write something like:

```
include(CMakeDependentOption)  # if necessary, might be included in a file above this and so would be redundant
cmake_dependent_option(WITH_TEST_RUNTIME "Build runtime tests" ON
                       "NOT MSVC" OFF)
if (WITH_TEST_RUNTIME)
  message(STATUS "Halide internal runtime tests enabled")
  add_subdirectory(runtime)
else ()
  message(STATUS "Halide internal runtime tests disabled")
endif ()
```

steven-johnson:
linux-x86-32 breakage should be fixed by https://github.com/halide/Halide/pull/6836

test/CMakeLists.txt
alexreinking: @derek-gerstmann - That's what I'm here for :slightly_smiling_face:

derek-gerstmann:
Looks like there's just the flaky windows machine that is failing on unrelated tests ...

steven-johnson (approved):
LGTM. 

I must admit that I have some misgivings about the amount of code we apparently need to support Vulkan. I assume that the additional code will only get linked in when Vulkan is enabled in features?

derek-gerstmann:
> I must admit that I have some misgivings about the amount of code we apparently need to support Vulkan. > I assume that the additional code will only get linked in when Vulkan is enabled in features?

Cool.  Yeah, I hear you ... and there's quite a bit more coming too for CodeGen and the Runtime, but those are wrapped with an ifdef guard and only built at compile time if the CMake option is for TARGET_VULKAN/SPIRV is enabled.

PR #6835, opened 2022-06-30
Rearrange subdirectories in python_bindings

steven-johnson:
This is intended to facilitate a few things:
- Move all Generators used in tests, apps, etc to a single directory to simplify the build rules (this is especially useful for the work in https://github.com/halide/Halide/pull/6764)
- Put all the test and apps stuff under a single directory to facilitate adding some Python packaging that can make integration into Bazel/Blaze builds a bit less painful

@alexreinking, does this look like the layout we discussed before?


alexreinking:
> does this look like the layout we discussed before?

Yes, it's a step towards it... I want to explore rearranging the `src/` directory to look more like the intended package layout. But that can wait for #6815

steven-johnson:
linux-x86-32 case should be fixed by https://github.com/halide/Halide/pull/6836 and shouldn't block us here

alexreinking (approved):

PR #6842, opened 2022-07-10
Check RDom::where predicates for race conditions

### LINKED ISSUE #6808: The check for race conditions doesn't consider where clauses
Halide should require allow_race_conditions() to be willing to compile the following code, but it doesn't. The where clause introduces a race.

```
    Func f;
    Var x;
    RDom r(0, 10);
    f(x) = 1;
    r.where(f(0) == 1);
    f(r) = 2;

    f.update().parallel(r);
```

ATTN @alexreinking

alexreinking:
Fixes #6808

### DIFF
--- a/src/Func.cpp
+++ b/src/Func.cpp
@@ -390,14 +390,16 @@ void Stage::set_dim_type(const VarOrRVar &var, ForType t) {
             // This check is fairly conservative.
             if (var.is_rvar && is_parallel(t) && !definition.schedule().allow_race_conditions()) {
                 const string &func_name = function.name();
-                const auto &pred = var.rvar.domain().predicate();
-                user_assert(!expr_uses_var(pred, func_name))
-                    << "In scheudle for " << name() << ", marking var "
-                    << var.name() << " as parallel may introduce a race "
-                    << "condition, resulting in incorrect output. This is "
-                    << "because the RDom::where predictate reads from "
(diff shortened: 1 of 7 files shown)


src/Func.cpp
```diff
@@ -386,6 +386,22 @@ void Stage::set_dim_type(const VarOrRVar &var, ForType t) {
...
             dim.for_type = t;
 
+            // Check that RDom::where is not data-dependent on the current func.
+            // This check is fairly conservative.
+            if (var.is_rvar && is_parallel(t) && !definition.schedule().allow_race_conditions()) {
+                const string &func_name = function.name();
+                const auto &rdom = var.rvar.domain();
+                if (rdom.defined()) {
+                    user_assert(!expr_uses_var(rdom.predicate(), func_name))
+                        << "In scheudle for " << name() << ", marking var "
```
abadams, line 396: scheudle

abadams:
It might be simpler and more precise to visit the rdom predicate when sniffing for loads here: https://github.com/halide/Halide/blob/main/src/ParallelRVar.cpp#L93

src/Func.cpp
alexreinking: Fixed

abadams (approved):

alexreinking:
Windows GPU atomics failure is unrelated.

PR #6844, opened 2022-07-11
Deprecate/remove Generator::get_externs_map() and friends

steven-johnson:
This is a feature of Generator that was added years ago to allow adding external code libraries in LLVM bitcode form (rather than simply as extern "C" or similar). In theory it allow for better codegen for external code modules (since LLVM has access to all the bitcode for optimization); in practice, we only know of one project that ever used it, and that project no longer exists. Additionally, it tended to be fairly flaky in terms of actual use -- e.g., missing symbols tended to crop up unpredictably.

The issues with this feature are likely fixable, but since it hasn't (AFAICT) been used in ~years, we're better off deprecating it for Halide 15 and removing for Halide 16.

(If anyone out there is still relying on this feature, obviously you should speak up ASAP.)


abadams (approved):
code lgtm - let's see if anyone complains

zvookin:
I'm not going to object to removing it, but the concept will likely have to reappear at some point. It was supposed to be the start of being able to include backend specific code to support the output of a generated. E.g. shader code that Halide GPU codegen could call out to. There was a specific use case that showed up a bunch of issues using it for LLVM bitcode which basically comes down to "linking is hard." IIRC, the use case was motivated by reducing compilation time rather than increasing optimization opportunities. (The external modules were Halide generated code, but precompiled.)

For shaders and whatnot, it is not possible to just provide the extra stuff in an object file handed to a later linking operation. We'll have to have some mechanism inside our GPU runtime to support this.

steven-johnson:
> the concept will likely have to reappear at some point

Yeah, I'm not opposed to the concept -- the underlying goal here is really to continue pruning the Generator infrastructure (to get C++ aligned better with the upcoming Python work) and this stuck out as something that is almost certainly unused outside of its one test. When we re-add this in the future we will hopefully have a better idea of how to design a more flexible system.

zvookin (approved):

steven-johnson:
Also removed `ExternalCode` and related stuff, PTAL perhaps

PR #6853, opened 2022-07-13
Promote Reinterpret Intrinsic into an Reinterpret IR Node

LebedevRI:
As discussed in https://github.com/halide/Halide/issues/6801#issuecomment-1152731683

I don't think this is complete, there are likely a few more places
that need to be taught about it still, altough i think this is mostly it.

Note that this only promotes the intrinsic,
this does not adjust it's handling, as hinted in:
https://github.com/halide/Halide/issues/6801#issuecomment-1155603752

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1476,6 +1476,62 @@ void CodeGen_LLVM::visit(const Cast *op) {
     }
 }
 
+void CodeGen_LLVM::visit(const Reinterpret *op) {
+    Type dst = op->type;
+    Type src = op->value.type();
+    llvm::Type *llvm_dst = llvm_type_of(dst);
+    value = codegen(op->value);
+    if (src.is_handle() && !dst.is_handle()) {
+        internal_assert(dst.is_uint() && dst.bits() == 64);
+
+        // Handle -> UInt64
+        llvm::DataLayout d(module.get());
+        if (d.getPointerSize() == 4) {
+            llvm::Type *intermediate = llvm_type_of(UInt(32, dst.lanes()));
+            value = builder->CreatePtrToInt(value, intermediate);
+            value = builder->CreateZExt(value, llvm_dst);
+        } else if (d.getPointerSize() == 8) {
+            value = builder->CreatePtrToInt(value, llvm_dst);
+        } else {
+            internal_error << "Pointer size is neither 4 nor 8 bytes\n";
+        }
+
+    } else if (dst.is_handle() && !src.is_handle()) {
+        internal_assert(src.is_uint() && src.bits() == 64);
+
+        // UInt64 -> Handle
+        llvm::DataLayout d(module.get());
+        if (d.getPointerSize() == 4) {
+            llvm::Type *intermediate = llvm_type_of(UInt(32, src.lanes()));
+            value = builder->CreateTrunc(value, intermediate);
+            value = builder->CreateIntToPtr(value, llvm_dst);
+        } else if (d.getPointerSize() == 8) {
+            value = builder->CreateIntToPtr(value, llvm_dst);
+        } else {
+            internal_error << "Pointer size is neither 4 nor 8 bytes\n";
+        }
+
+    } else {
+        if (src.is_scalar() && dst.is_vector()) {
+            // If the source type is a scalar, we promote it to an
+            // equivalent vector of width one before doing the
+            // bitcast, because llvm's bitcast operator doesn't
+            // want to convert between scalars and vectors.
+            value = create_broadcast(value, 1);
+        }
+        if (src.is_vector() && dst.is_scalar()) {
+            // Similarly, if we're converting from a vector to a
+            // scalar, convert to a vector of width 1 first, and
+            // then extract the first lane.
+            llvm_dst = get_vector_type(llvm_dst, 1);
+        }
+        value = builder->CreateBitCast(value, llvm_dst);
+        if (src.is_vector() && dst.is_scalar()) {
+            value = builder->CreateExtractElement(value, (uint64_t)0);
+        }
+    }
+}
+
 void CodeGen_LLVM::visit(const Variable *op) {
     value = sym_get(op->name);
 }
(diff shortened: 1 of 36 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -2593,61 +2648,6 @@ void CodeGen_LLVM::visit(const Call *op) {
...
-            if (src.is_vector() && dst.is_scalar()) {
-                // Similarly, if we're converting from a vector to a
-                // scalar, convert to a vector of width 1 first, and
-                // then extract the first lane.
-                llvm_dst = get_vector_type(llvm_dst, 1);
-            }
-            value = builder->CreateBitCast(value, llvm_dst);
-            if (src.is_vector() && dst.is_scalar()) {
-                value = builder->CreateExtractElement(value, (uint64_t)0);
-            }
```
LebedevRI, line 2649: FWIW, i do not understand at all why this never exploded. This is just completely broken.

abadams: Which part is broken? The idea is that if we want to reinterpret a vector float x 1 as a scalar uint32, it first bitcasts to vector uint32 x 1, and then extracts lane zero to get the uint32.

LebedevRI: The entirety of this logic. LLVM IR bitcast does not change the type size,
you can't bitcast i8 to i32, or i8 to <4 x i8>, or i32 to i8, or <4 x i8> to i8,
you can only change the scalar element size (including casting between a scalar and a vector)

abadams: I don't see any code here that changes the type size. We know the reinterpret call has already asserted the type size is the same (see reinterpret in IROperator.cpp).

LebedevRI: Aha! So what happened was, i thought that i had to change forementioned legality check
because the node is now being created in more places, but in the reality, it was the opposite,
i didn't port all the places that used to handle the reinterpret intrinsic. :)
After porting a few more, the original logic is good again, and all the tests finally pass :)

LebedevRI:
Ok, i've gone over everything that used to override the `visit(const Call*)`, and i think this should now be pretty complete.
Note that this is a no-op promotion, it did not alter it's handling as hinted in https://github.com/halide/Halide/issues/6801#issuecomment-1155603752

LebedevRI:
Looks like bots are happy now?

@abadams i think it would be cleaner to change the handling in a separate follow-up?
Right now i'm not really sure how that would look, or which parts should be affected.

src/IROperator.h
```diff
@@ -1054,12 +1054,12 @@ Expr is_finite(Expr x);
  *  same sign as the original expression. Vectorizes cleanly. */
 Expr fract(const Expr &x);
 
-/** Reinterpret the bits of one value as another type. */
+// Thin wrapper for `reinterpret()`.
```
rootjalex, line 1057: nit: I would leave the original comment here, this is the user-facing function.

src/IRPrinter.cpp
```diff
@@ -519,6 +519,13 @@ void IRPrinter::visit(const Cast *op) {
     stream << ")";
 }
 
+void IRPrinter::visit(const Reinterpret *op) {
+    // FIXME: should it be printed differently from the Cast?
```
abadams, line 523: I think you can delete this FIXME

src/LICM.cpp
```diff
@@ -89,6 +89,10 @@ class LiftLoopInvariants : public IRMutator {
                 return false;
             }
         }
+        if (const Reinterpret *reinterpret = e.as<Reinterpret>()) {
+            // Don't lift these intrinsics. They're free.
```
abadams, line 93: Not an intrinsic, so maybe alter comment to say "Reinterpret nodes" instead of "these intrinsics"

LebedevRI: Whoops, i forgot.

src/RegionCosts.cpp
```diff
@@ -80,6 +80,11 @@ class ExprCost : public IRVisitor {
         arith += 1;
     }
 
+    void visit(const Reinterpret *op) override {
+        op->value.accept(this);
+        arith += 1;
```
abadams, line 85: Remove arith += 1. This is a change to existing behavior, but it's a good one.

abadams:
LGTM except where I commented otherwise. Approving.

abadams (approved):

src/EliminateBoolVectors.cpp
```diff
@@ -136,6 +136,8 @@ class EliminateBoolVectors : public IRMutator {
         }
     }
 
+    // FIXME: what about Reinterpret?
```
LebedevRI, line 139: @abadams what about this one?
I do believe this is a correct translation, but is that the correct handling in reality?

LebedevRI:
@rootjalex thanks for taking a look!

@abadams thank you for the review!
Can you give me a rough list of the 'passes' that will need to be adjusted next?
In https://github.com/halide/Halide/issues/6801#issuecomment-1155603752, `Deinterleaver` was mentioned. Anything else obvious?

src/LICM.cpp
LebedevRI: Ugh, i hate github UI.

abadams:
Nothing else obvious comes to mind in terms of passes to modify to do interesting things with reinterprets that change the number of vector lanes. The next step is probably to add extract_bits and concat_bits intrinsics and then make it so that the vectorization pass handles them cleverly. e.g. a vectorized concat_bits equivalent to `(u16(a) << 8) | b` could codegen to a shuffle vector that interleaves the two u8 input vectors followed by a vector reinterpret. A vectorized extract_bits could codegen to a reinterpret to a narrower type with more lanes followed by a shuffle.

For the use-cases we care about, the idea is that the shuffles would optimize away later, e.g. in the first case because we're interleaving strided loads. In the second case things get complicated... I tried to write down the sequence of transformations that gets me to the code I want, but it's not simple.

LebedevRI:
> Nothing else obvious comes to mind in terms of passes to modify to do interesting things with reinterprets that change the number of vector lanes.

Ok, that's reassuring i suppose.

> The next step is probably to add extract_bits and concat_bits intrinsics and then make it so that the vectorization pass handles them cleverly.

Hm, so we *do* want to introduce proper intrinsics for these? Okay.
I've implemented them (& a few more related QoL helpers) in https://github.com/halide/Halide/pull/6775/files#diff-d1915f767d2fe08376341061a7e10cdd4e11728211ffb6b399675108b98b5aca in native halide,
i guess i will have to reimplement them all :)

> e.g. a vectorized concat_bits equivalent to `(u16(a) << 8) | b` could codegen to a shuffle vector that interleaves the two u8 input vectors followed by a vector reinterpret. A vectorized extract_bits could codegen to a reinterpret to a narrower type with more lanes followed by a shuffle.

Right, that's obvious.

> For the use-cases we care about, the idea is that the shuffles would optimize away later, e.g. in the first case because we're interleaving strided loads. In the second case things get complicated... I tried to write down the sequence of transformations that gets me to the code I want, but it's not simple.

LebedevRI:
Bots seem happy, ship it? :)

LebedevRI:
Thank you!
I'll look into the mentioned follow-ups.

PR #6854, opened 2022-07-14
Silence Adams2019 Autoscheduler

steven-johnson:
The overall goal of this PR is that when *using* the Adams2019 autoscheduler, the default behavior should be no output to stdout or stderr, period, unless a fatal error occurs: no progress, no warnings, nothing. (This can be overridden by setting `HL_DEBUG_AUTOSCHEDULE` or `HL_DEBUG_CODEGEN` as you see fit, for progressively more logging.)

This change happened in multiple steps, though:
- ~~reworked the `aslog()` class to be a proper subclass of `std::ostream` rather than a one-off templated thing~~
- reworked all the `dump()` method to output solely to an `ostream` argument rather than assuming an output level
- tweaked `aslog()` levels in various places; in particular, the *only* place right now that uses `aslog(0)` is the ScopedTimer class (which isn't in use right now and generally shouldn't be).

Note that the "no output at all" rule doesn't apply to training scripts, tests, etc... just the "normal" usage of the autoscheduler itself.

I'm sure that some of the new aslog levels I've assigned will need tweaking, but I think this change is fairly reasonable as-is.

~~(Feedback on the change to make `aslog()` a proper ostream is welcome; I'm tempted to do something similar to the `debug()` class in Halide proper but want more feedback first...)~~

### DIFF
--- a/src/autoschedulers/adams2019/ASLog.cpp
+++ b/src/autoschedulers/adams2019/ASLog.cpp
@@ -41,7 +41,7 @@ std::array<aslog_stream, kMaxLevel + 1> aslog_streams = {
     aslog_stream(0),
     aslog_stream(1),
     aslog_stream(2),
-    aslog_stream(3)
+    aslog_stream(3),
 };
 
 }  // namespace
--- a/src/autoschedulers/adams2019/ASLog.cpp
+++ b/src/autoschedulers/adams2019/ASLog.cpp
@@ -1,8 +1,5 @@
 #include "ASLog.h"
 
-#include <array>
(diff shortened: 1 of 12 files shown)


steven-johnson:
Just realized this morning that the way I've reworked `aslog()` means that we may be evaluating (and discarding) all the RHS terms even when not logging them (while the old code probably skipped that evaluation). Let me re-examine further.

steven-johnson:
ok, I think this is less weird, PTAL

abadams (approved):

PR #6861, opened 2022-07-20
Remove (most) of the env var usage from Adams2019

steven-johnson:
The Adams2019 autoscheduler used a huuuuge number of env vars to tweak things, which isn't a good long-term design. I've moved most of these into the extensible AutoschedulerParams that was recently introduced, and updated tests etc accordingly. (A couple were simply removed entirely since they were long-deprecated.)

A couple remain: 
- `HL_DEBUG_AUTOSCHEDULE`, which is the similar to HL_DEBUG_CODEGEN, and will stay ~forever
- `HL_CYOS`, which allows for interactive (!) scheduling, remains present, but disabled via `#ifdef HALIDE_AUTOSCHEDULER_ALLOW_CYOS` by default -- since it's unlikely that anyone other than people actualyl working on the autoscheduler itself are likely to want this feature, it seems like a bad idea to compile it in by default.
- `HL_PERMIT_FAILED_UNROLL`, which controls something in libHalide (!) and can't directly be removed here. I'm *hoping* that https://github.com/halide/Halide/pull/6792 will improve things enough that we can simply remove this bit of hackery; if it doesn't, we may need to entertain adding a `Feature` flag for this purpose (though that seems pretty hackish too).


steven-johnson:
Gentle review ping

src/autoschedulers/adams2019/LoopNest.h
```diff
@@ -34,6 +32,10 @@ bool may_subtile();
 std::vector<std::vector<int64_t>> generate_tilings(const vector<int64_t> &s, int d, int factor, bool allow_splits);
 
 struct LoopNest {
+    explicit LoopNest(bool may_subtile)
```
abadams, line 35: may_subtile is essentially a flag for a comparison to prior work we wanted to do in the paper. It's pretty tedious to have to plumb it everywhere and wasteful to allocate space in the LoopNest struct to store it. I see that the params struct is already plumbed through into the child-generation routine, so why is this even necessary?

src/autoschedulers/adams2019/AutoSchedule.cpp
```diff
@@ -305,45 +266,47 @@ IntrusivePtr<State> optimal_schedule_pass(FunctionDAG &dag,
             q.emplace(std::move(s));
         };
 
+#ifdef HALIDE_AUTOSCHEDULER_ALLOW_CYOS
```
abadams, line 269: Why a separate ifdef for this one? CYOS is for debugging why a schedule that you want doesn't seem to be in the search space, or is being rejected based on cost. It'd be pretty annoying to have to recompile the autoscheduler to use it.

steven-johnson: I argue that literally no one other than a developer/maintainer of this autoscheduler will ever use this feature; having an interactive mode in what is otherwise a command-driven tool seems like an odd fit at best, and a security risk at worse. But maybe I'm wrong about usage? I'd love to hear your experience regarding who has used this feature (and how and why) in the past.

src/autoschedulers/adams2019/LoopNest.h
steven-johnson: Let me recheck -- when I did this I don't think that struct was plumbed everywhere it needed to go.

src/autoschedulers/adams2019/AutoSchedule.cpp
```diff
@@ -156,28 +115,32 @@ struct ProgressBar {
...
 };
 
-// Get the HL_RANDOM_DROPOUT environment variable. Purpose of this is described above.
-uint32_t get_dropout_threshold() {
-    string random_dropout_str = get_env_variable("HL_RANDOM_DROPOUT");
-    if (!random_dropout_str.empty()) {
-        return atoi(random_dropout_str.c_str());
-    } else {
-        return 100;
+#ifdef HALIDE_ALLOW_LEGACY_AUTOSCHEDULER_API
```
abadams, line 118: This is a bit hidden. In the environment variables list about, we should keep the descriptions, and say that if you want to use the following environment variables to control more features of the autoscheduler, you should #define HALIDE_ALLOW_LEGACY_AUTOSCHEDULER_API here, but this is deprecated and better done using the parameters struct

src/autoschedulers/adams2019/AutoSchedule.cpp
abadams: Actually CYOS should possibly just be moved into the params struct

abadams: (Race condition on our previous comments)

Hrm, you're probably right.

steven-johnson: Does that mean it's ok as-is, or do you want revisions?

abadams: Maybe just put it behind the legacy api flag instead of its own separate thing

steven-johnson: Hmm, well my plan was to remove the legacy API flag entirely sometime in Halide 16, but I suspect you'd like to keep this for future experimentation?

src/autoschedulers/adams2019/AutoSchedule.cpp
steven-johnson: Good point, will do.

src/autoschedulers/adams2019/AutoSchedule.cpp
abadams: Fair enough, I guess it's fine as-is

src/autoschedulers/adams2019/AutoSchedule.cpp
steven-johnson: Done, PTAL

src/autoschedulers/adams2019/AutoSchedule.cpp
```diff
@@ -20,24 +20,32 @@
...
+
+#ifdef HALIDE_ALLOW_LEGACY_AUTOSCHEDULER_API
+
+  Most of the settings in this Autoscheduler are controlled by the values specified via
+  an `autoscheduler.fieldname` GeneratorParam, as listed in the Adams2019Params struct;
+  this is the preferred way to set these.
+
+  For now, however, you can (instead) control these settings via env vars;
+  doing so requires that you compile all of Halide with HALIDE_ALLOW_LEGACY_AUTOSCHEDULER_API
+  defined. )Note that this ability is deprecated, and likely to be removed in Halide 16.)
```
abadams, line 38: nit: ) -> (

abadams (approved):

LebedevRI:
@steven-johnson this broke one test, not sure why this wasn't caught:
```
$ ctest -LE performance -L multithreaded -j1 -R test_apps_autoscheduler --output-on-failure
Test project /builddirs/halide/build-Clang-Release-LLVM14
    Start 3: test_apps_autoscheduler
1/1 Test #3: test_apps_autoscheduler ..........***Failed    0.01 sec
Usage: /builddirs/halide/build-Clang-Release-LLVM14/src/autoschedulers/adams2019/test_apps_autoscheduler <autoscheduler-lib> <weights-path>


0% tests passed, 1 tests failed out of 1

Label Time Summary:
Adams2019        =   0.01 sec*proc (1 test)
multithreaded    =   0.01 sec*proc (1 test)

Total Test time (real) =   0.02 sec

The following tests FAILED:
          3 - test_apps_autoscheduler (Failed)
Errors while running CTest

```

steven-johnson:
Looks like an easy fix, but scary that we missed it -- apparently buildbots aren't running these tests? I'll fix that.

PR #6865, opened 2022-07-20
[Codegen_LLVM] Radically simplify `visit(const Reinterpret *op)`

LebedevRI:
1. LLVM IR `bitcast` happily bitcasts between vectors and scalars: https://godbolt.org/z/9zqx11rna
2. `ptrtoint` already implicitly truncates/zero-extends if the int is larger than the pointer type: https://llvm.org/docs/LangRef.html#ptrtoint-to-instruction
3. `inttoptr` already implicitly truncates/zero-extends if the int is larger than the pointer type: https://llvm.org/docs/LangRef.html#inttoptr-to-instruction

So we don't need to do any of that 'special' handling.

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1478,58 +1478,14 @@ void CodeGen_LLVM::visit(const Cast *op) {
 
 void CodeGen_LLVM::visit(const Reinterpret *op) {
     Type dst = op->type;
-    Type src = op->value.type();
     llvm::Type *llvm_dst = llvm_type_of(dst);
     value = codegen(op->value);
-    if (src.is_handle() && !dst.is_handle()) {
-        internal_assert(dst.is_uint() && dst.bits() == 64);
-
-        // Handle -> UInt64
-        llvm::DataLayout d(module.get());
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):
:happydance:

steven-johnson:
> 1. LLVM IR `bitcast` happily bitcasts between vectors and scalars: https://godbolt.org/z/9zqx11rna
> 2. `ptrtoint` already implicitly truncates/zero-extends if the int is larger than the pointer type: https://llvm.org/docs/LangRef.html#ptrtoint-to-instruction
> 3. `inttoptr` already implicitly truncates/zero-extends if the int is larger than the pointer type: https://llvm.org/docs/LangRef.html#inttoptr-to-instruction

(Might be worth adding these points to a comment in the file, for future readers)

LebedevRI:
I don't think i've seen `generator_aot_async_parallel` failure before, i guess it's a fluke?

LebedevRI:
Looks like bots are okay with this? The failures are the same as in #6866.

LebedevRI:
@steven-johnson thank you!

PR #6866, opened 2022-07-20
[Codegen_LLVM] Define all the things

LebedevRI:
Long-term plan for LLVM is to get rid of `undef`,
and replace it with zero-initialization,
err, `poison`, because it has nicer semantics.

Everywhere where we use `undef` as a placeholder in shuffle
(be it either for a second operand, or undef shuffle mask element),
or as a base 'empty' vector we are about to fully override
via insertelement, we can just switch those to poison nowadays.

The scary part is the `Call::undef` semantics/lowering,
perhaps it will need to be `freeze poison`.
For this reason, it might not be unreasonable to give this a spin downstream first.

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -998,8 +998,8 @@ llvm::Function *CodeGen_Hexagon::define_hvx_intrinsic(llvm::Function *intrin,
 Value *CodeGen_Hexagon::create_bitcast(Value *v, llvm::Type *ty) {
     if (BitCastInst *c = dyn_cast<BitCastInst>(v)) {
         return create_bitcast(c->getOperand(0), ty);
-    } else if (isa<UndefValue>(v)) {
-        return UndefValue::get(ty);
+    } else if (isa<PoisonValue>(v)) {
+        return PoisonValue::get(ty);
     } else if (v->getType() != ty) {
         v = builder->CreateBitCast(v, ty);
     }
@@ -1178,7 +1178,7 @@ Value *CodeGen_Hexagon::shuffle_vectors(Value *a, Value *b,
                 i -= a_elements;
             }
         }
-        return shuffle_vectors(b, UndefValue::get(b->getType()), shifted_indices);
+        return shuffle_vectors(b, shifted_indices);
     }
 
     // Try to rewrite shuffles that only access the elements of a.
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -2021,15 +2021,11 @@ void CodeGen_LLVM::visit(const Load *op) {
                 Value *load_i = codegen_dense_vector_load(op->type.with_lanes(load_lanes_i), op->name, slice_base,
                                                           op->image, op->param, align, nullptr, false);
 
-                SmallVector<Constant *, 256> constants;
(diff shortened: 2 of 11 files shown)


abadams (approved):

LebedevRI:
> Reading up on it, yeah I suspect what we want for Call::undef is frozen poison. It's a fixed-but-indeterminate value.

The problem with that, and why i didn't go for that immediately, is that would be less strict than the current lowering:
https://alive2.llvm.org/ce/z/YW5xBQ

Given that it is documented as really dangerous
https://github.com/halide/Halide/blob/main/src/IROperator.h#L1339-L1340
i'm wondering if we could first go with the proposed lowering,
and then depending on the fallout, soften it via freeze?

cc @nunoplopes

nunoplopes:
Give what's written here:
https://github.com/halide/Halide/blob/359026a87d8d83d5be4485f3fd063775bbce7351/src/IROperator.h#L1324-L1327

The only possible value with that semantics in LLVM is poison. You can't skip storing an undef or freeze poison.
(note that I'm clueless about Halide, just giving some advice from the LLVM perspective)

LebedevRI:
> The only possible value with that semantics in LLVM is poison. You can't skip storing an undef or freeze poison.

Thank you for confirming my understanding :)
With that we have established that if something breaks,
it was broken already and it should be fixed, not halide.

abadams:
The problem is that Call::undef is used for two purposes in Halide. In front-end code the definition of Call::undef is that storing it has no effect, and it can't arithmetically combine with values other than being a select branch. It's not related to llvm's undef or poison. It's something different.

But there may also be places where we use Call::undef inside lowering or backends to mean something more like llvm's undef/poison.

LebedevRI:
Ack.

Data point: as of all the tests + apps (excl. hannk), we never actually LLVM IR-gen the `Call::undef`.
Do we have any examples where we //do// irgen it? It would be good to have at least one in the tree.
Why might we irgen Call::undef, because halide failed to eliminate it? If not, when might it legitimately appear?

Since we are fine with the preexisting codegen (`undef`), and aren't changing it to `freeze undef`,
because no bugs have been rootcaused to it, i'm not confident that switching to `poison` would regress things.
But indeed it might, so maybe it would be good for downstreams to try this as-is.

abadams:
If that's the case, we should probably just fail an assertion inside CodeGen_LLVM if we see Call::undef. I thought it might have been introduced during codegen itself for certain constructs, but if it isn't, let's forbid it.

LebedevRI:
> If that's the case, we should probably just fail an assertion inside CodeGen_LLVM if we see Call::undef. I thought it might have been introduced during codegen itself for certain constructs, but if it isn't, let's forbid it.

Posted https://github.com/halide/Halide/pull/6871

LebedevRI:
Looks like bots are happy with this, but i suppose #6871 will need to go in first, and this is rebased afterwards.

LebedevRI:
(rebased over landed #6871)

LebedevRI:
Not sure what to make of `performance_inner_loop_parallel` failure (https://buildbot.halide-lang.org/master/#/builders/19/builds/371), but the rest are flukes.

steven-johnson:
Clearly a flake, landing

LebedevRI:
@steven-johnson thank you!

PR #6867, opened 2022-07-20
Python source reorg

alexreinking:
This is to make #6815 smaller and easier to review. This PR does the following within the `python_bindings` directory:

1. It moves the C++ sources into `src/halide/halide_`
2. It renames the native module to `halide_`. Putting the `_` at the beginning would lead to a symbol being named `PyInit__halide`, which is _technically_ reserved by C++ (double underscores).
3. The `halide` folder is now a valid Python package thanks to `__init__.py`. This just imports and re-exports everything inside `halide_`, which is fully qualified as `halide.halide_`.
4. The test paths and install rules are adjusted accordingly.

This will make it simpler to integrate plain Python code into our package. For instance, #6764 should be adjusted accordingly.


steven-johnson:
> * It renames the native module to `halide_`. Putting the `_` at the beginning would lead to a symbol being named `PyInit__halide`, which is _technically_ reserved by C++ (double underscores).

`halide_` is just fine with me, but I'm just gonna point out that as a practical matter, C++ is ~never gonna be able to break code that uses __ in this way. (I'm pretty sure that CodegenC violates this already.)

python_bindings/src/halide/__init__.py
```diff
@@ -0,0 +1,7 @@
+import sys
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).parent.resolve(strict=True)))
```
steven-johnson, line 4: A comment on this line would be nice.

alexreinking: Yeah, this is a stop-gap.

I would really like this file to be just the following:
```
from .halide_ import *
from .halide_ import _, _1, ..., _9
```
However,
1. relative imports cannot be overloaded with `sys.path`
2. for a variety of reasons, copying the python sources next to the `halide_.*.so` module is difficult to make work in 100% of cases in CMake
3. even if we could copy them reliably, copying these files into the build folder seems inelegant

Fortunately, there are apparently other hooks besides `sys.path` that we could use to redirect a failing relative import. I haven't looked into them yet, but didn't feel like holding up this PR for it.

alexreinking: Will do

Issue #6868, opened 2022-07-20
Python Bindings need a way for AOT code to manage GPU buffers


steven-johnson:
From https://github.com/halide/Halide/discussions/6846, a number of issues:

- The `.py.cpp` code that we produce for AOT-compiled Python extensions operates solely on the Python Buffer protocol, which is CPU-memory only (it has no provision for a buffer's memory to live on device). As such it really should always call `copy_to_host()` on all output buffers, to ensure results are flushed properly, but it does not. (This should be an easy fix).

- More problematically, there isn't currently a way to avoid needless copy-to-host calls right now, since Python Buffer doesn't support anything but host, and we don't have the equivalent of `Halide::Runtime::Buffer` in our Python bindings. We could add such an equivalent, but it would be vastly preferable to adopt an existing solution already in use by other GPU-accelerated Python libraries. `dlpack` (https://dmlc.github.io/dlpack/latest/index.html) appears to be a likely candidate but investigation is needed.

PR #6869, opened 2022-07-20
Add set-host-dirty/copy-to-host to PythonExtensionGen

steven-johnson:
See https://github.com/halide/Halide/issues/6868: Python Buffers are host-memory-only, so if the AOT-compiled halide code runs on (say) GPU, it may fail to copy the inputs to device and/or the results back to host. This fixes that. (We still need a solution that allows for lazy copies, but that will require adding another protocol that supports it.)

attn @dragly


steven-johnson:
Monday Morning Review Ping

abadams (approved):

PR #6871, opened 2022-07-20
[Codegen] Fail to codegen `Call::undef`, just like `Call::signed_integer_overflow`

LebedevRI:
See discussion in https://github.com/halide/Halide/pull/6866.
It's not obvious if that codepath is ever hit,
let's optimistically assume that it is not.

If this turns out to be not true, we'll have to deal with
a more complicated question of the proper lowering for it,
can it be `poison`, or must it be a `freeze poison`.


LebedevRI:
Looks like the bots are indeed happy with this.
What's the plan here, speculatively merge and revert upon reported problems?

steven-johnson:
> What's the plan here, speculatively merge and revert upon reported problems?

SGTM

LebedevRI:
@steven-johnson Thank you!

PR #6677, opened 2022-04-05
Faster `widening_mul(int16x, int16x) -> int32x` for x86 (AVX2 and SSE2)

rootjalex:
Halide/LLVM currently lowers `widening_mul(int16x, int16x) -> int32x` to `vpmovsxwd` + `vpmulld`, this PR changes to `vpmovzxwd` + `vpmaddwd`.

Fixes #6676

Benchmark courtesy of @abadams is compute-limited:
```
Before: 0.0475274s
Now: 0.0344593s
```

```
#include "Halide.h"
#include "halide_benchmark.h"

using namespace Halide;

double run_test(const int N) {
    Func f("f");
    Var x("x"), y("y");
    ImageParam im0(Int(16), 2);
    ImageParam im1(Int(16), 2);

    RDom r(0, 10);
    f(x, y) += cast<int32_t>(im0(x, y)) * cast<int32_t>(im1(r, y));
    Target t = get_target_from_environment();
    const int vec_size = t.natural_vector_size<int16_t>();
    f.update().vectorize(x, vec_size).unroll(r);
    f.compile_jit();

    Buffer<int16_t> in0(N, 1);
    Buffer<int16_t> in1(N, 1);
    Buffer<int32_t> out(N, 1);
    in0.raw_buffer()->dim[1].extent = N;
    in1.raw_buffer()->dim[1].extent = N;
    out.raw_buffer()->dim[1].extent = N;
    in0.raw_buffer()->dim[1].stride = 0;
    in1.raw_buffer()->dim[1].stride = 0;
    out.raw_buffer()->dim[1].stride = 0;

    for (int i = 0; i < N; i++) {
        in0(i, 0) = rand() & 0xff;
        in1(i, 0) = rand() & 0xff;
        out(i, 0) = 0;
    }

    im0.set(in0);
    im1.set(in1);

    double timed = Halide::Tools::benchmark(10, 10, [&]() {
        f.realize(out);
        out.device_sync();
    });
    return timed;
}

int main(int argc, char **argv) {
    const int N = 1024;
    double ta = 0;
    for (int i = 0; i < 100; i++) {
        ta += run_test(N);
    }
    std::cerr << ta << "\n";
    return 0;
}
```

PR #6878, opened 2022-07-21
[x86] Codegen phaddw, phaddd, and pmaddwd

rootjalex:
This PR adds support for `int16 -> int32` horizontal widening adds to use `pmaddwd`, and pattern matches on horizontal adds to use `phadd(w | d)`, which is faster than the permute + padd that we currently generate for such reductions (cite @abadams + llvm-mca because my x86 machine is currently broken). 

Also fly-by move of some code that I incorrectly placed in the wrong runtime file in #6677 .

### DIFF
--- a/src/CodeGen_X86.cpp
+++ b/src/CodeGen_X86.cpp
@@ -185,9 +185,15 @@ const x86Intrinsic intrinsic_defs[] = {
     // LLVM does not provide an unmasked 128bit cvtneps2bf16 intrinsic, so provide a wrapper around the masked version.
     {"vcvtneps2bf16x4", BFloat(16, 4), "f32_to_bf16", {Float(32, 4)}, Target::AVX512_SapphireRapids},
 
-    // 2-way dot products
-    {"llvm.x86.avx2.pmadd.ub.sw", Int(16, 16), "saturating_dot_product", {UInt(8, 32), Int(8, 32)}, Target::AVX2},
-    {"llvm.x86.ssse3.pmadd.ub.sw.128", Int(16, 8), "saturating_dot_product", {UInt(8, 16), Int(8, 16)}, Target::SSE41},
+    // Horizontal adds that use (v)phadd(w | d).
+    {"phaddw_sse3", UInt(16, 8), "horizontal_add", {UInt(16, 16)}, Target::SSE41},
+    {"phaddw_sse3", Int(16, 8), "horizontal_add", {Int(16, 16)}, Target::SSE41},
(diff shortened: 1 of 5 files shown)


steven-johnson (approved):
LGTM pending green -- I assume these changes are good against non-trunk LLVM versions too.

rootjalex:
Yes, it should be fine - the phadd intrinsics have been in LLVM for over a decade [sse3](https://github.com/llvm/llvm-project/commit/2377206923c1b77296899b44a77ae814806ff2f2) [avx2](https://github.com/llvm/llvm-project/commit/682b850602d3380528c5ecafe352c48422e19724).

steven-johnson:
`The vector and scalar versions of op_phaddw_285 disagree. Maximum error: 60123`

rootjalex:
Failures were due to a shuffling mistake, I forgot that AVX2 has the 128 bit boundaries for instructions - fixed in 0acae70, and confirmed that we still produce better codegen for reductions.

abadams:
There's a trade-off between adding new llvm ir implementations of things, and rewriting Halide Exprs into forms that would match existing patterns. E.g. for the widening add, you could rewrite the Expr so that it matches should_use_dot_product.

My question is: Did you consider that, and does this PR fall on the right side of that line?

rootjalex:
We don't have any existing patterns for `phadd` instructions, so I'll only respond to your question in regards to the use of `pmaddwd` here.
I did not consider rewriting within Halide IR, but your suggestion raises a good point - I realize that I did not provide LLVM implementations for the AVX512 dot_product instructions, which could be similarly used. I suspect that your suggest would benefit from greater generality, and would be better suited for extensibility, so I will change this PR to transform all of the `horizontal_widening_add`s into the pattern that will match against the existing dot_product patterns

rootjalex:
I will say that I don't think these particular patterns are suitable for rewriting to enable `should_use_dot_product`, but instead I will aim for producing the patterns that will match the VectorReduce dot_product instructions.

steven-johnson:
buildbots are green -- is this ready to land?

rootjalex:
I might close this PR in favor of #6884 , which will make the `horizontal_widening_add` patterns useful on any architecture that has `pmaddwd` variants. Just waiting to get feedback on whether that PR is indeed a better way of doing things.

steven-johnson:
Monday Morning Review Ping -- where does this PR stand? Is it still likely to be dropped in favor of #6884?

rootjalex:
Yes, I am closing this in favor of #6884 . If we end up not merging #6884 for some reason, I will re-open this PR.

PR #6879, opened 2022-07-21
Ensure $CMAKE_{lang}_OUTPUT_EXTENSION is set before using it

### LINKED ISSUE #6876: Bundling LLVM libs includes `__.SYMDEF` members
Observed on macOS with Xcode 13.4 (and Xcode 14 beta), CMake build.

If using `-DBUILD_SHARED_LIBS=OFF` and `-DHalide_BUNDLE_LLVM=ON`, the static library generated via CMake includes at least one extra `__.SYMDEF` member.  Newer toolchains (e.g. Apple clang 14) error out during link if multiple of these are in the generated library.

shoaibkamil:
Fixes #6876.

This adds a check to `BundleStatic.cmake` to ensure the extension is non-empty, so we don't pick up files we shouldn't when bundling objects into libHalide.

### DIFF
--- a/cmake/BundleStatic.cmake
+++ b/cmake/BundleStatic.cmake
@@ -184,7 +184,9 @@ function(transfer_locations)
 
             set(globs "")
             foreach (lang IN LISTS languages)
-                list(APPEND globs "${stage}/*${CMAKE_${lang}_OUTPUT_EXTENSION}")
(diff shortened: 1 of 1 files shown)


alexreinking (approved):
looks good, as discussed

alexreinking:
Performance flake is unrelated.

Issue #6863, opened 2022-07-20
runtime_internal_string_storage fails under ASAN


LebedevRI:
@derek-gerstmann

There is a number of issues:
*
```
        StringStorage ss;
        halide_abort_if_false(user_context, ss.length() == 0);
```
is UB because `ss` is garbage and not zero-initialized
* `memcpy` in `StringStorage::prepend()` over overlapping memory ranges
* ???

derek-gerstmann:
@LebedevRI Ooh, thanks for the report! Yes, I just built these against address sanitizer, and there's some obvious defects.  Very much in favor of adding an ASAN pass to the build to catch these types of issues. I'll clean things up and add a pull request.

PR #6880, opened 2022-07-21
#6863 - Fixes to make address sanitizer happy for internal runtime classes

derek-gerstmann:
This PR addresses some obvious defects detected when built under ASAN as reported by @LebedevRI in Issue #6863 :

Fixed initialization defects in StringStorage that could cause buffer overruns.
Fixed memory leaks within RegionAllocator and BlockAllocator.
Added system memory allocation tracking to all internal runtime tests.

### DIFF
--- a/src/runtime/internal/string_storage.h
+++ b/src/runtime/internal/string_storage.h
@@ -31,9 +31,26 @@ struct StringUtils {
         return count;
     }
 
-    static size_t count_length(const char *str) {
+    // retuns true if s1 contains s2 (within n characters)
+    static bool contains(const char *s1, const char *s2, size_t n) {
+        if(is_empty(s2)) { return true; } // s2 is empty ... return true to match strstr 
+        char starts_with = *s2;
+        for (size_t length = strlen(s2); length <= n; n--, s1++) {
+            if (*s1 == starts_with) {
+                for (size_t i = 1; i <= length; i++) {
+                    if (i == length)
+                        return true;
+                    if (s1[i] != s2[i])
+                        break;
+                }
+            }
+        }
+        return false;
+    }
+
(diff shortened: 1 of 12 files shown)


steven-johnson:
With no disrespect intended *at all* towards Derek for his excellent work, this is the sort of thing that made me worry when I saw that we needed to add all this support to make Vulkan happen -- support classes of this stuff are easy to make these sorts of mistakes on and it pains me that we have to reinvent them here. I'm now idly wondering if there is an "STL-lite" sort of thing we could use for this instead of continuing to re-invent this wheel...

src/runtime/internal/string_storage.h
```diff
@@ -31,9 +31,26 @@ struct StringUtils {
...
 
-    static size_t count_length(const char *str) {
+    // retuns true if s1 contains s2 (within n characters)
+    static bool contains(const char *s1, const char *s2, size_t n) {
+        if (is_empty(s2)) { return true; }  // s2 is empty ... return true to match strstr
+        char starts_with = *s2;
+        for (size_t length = strlen(s2); length <= n; n--, s1++) {
+            if (*s1 == starts_with) {
+                for (size_t i = 1; i <= length; i++) {
+                    if (i == length)
```
steven-johnson, line 41: All if statements should be {braced} (I'm surprised that clang-format didn't complain, maybe we aren't running it on the runtime stuff? Perhaps please just manually format all these files)

steven-johnson (approved):

src/runtime/internal/string_storage.h
derek-gerstmann: Oops ... yeah strange that clang-format didn't catch these ... it's certainly running on all the internal classes I added.  I'll go ahead and manually reformat this section!

PR #6881, opened 2022-07-22
Add ASAN support to CMake build files

steven-johnson:
Opening as draft for now to get some buildbot coverage. Caveats:
- passes locally on OSX and Linux (x86), except for stuff known to be broken by https://github.com/halide/Halide/pull/6880, not tested on Windows or Arm
- python tests are disabled under asan for now; getting those working requires a bit more care
- the CMake file organization etc is probably suboptimal


cmake/HalideSanitizerHelpers.cmake
```diff
@@ -2,23 +2,29 @@
...
-    # - detect_leaks, because circular Expr chains in Halide can indeed leak,
-    #   but we don't care here
-    # - detect_container_overflow, because this is a known false-positive
-    #   if compiling with a non-ASAN build of LLVM (which is usually the case)
-    set(SANITIZER_ENV_VARS "ASAN_OPTIONS=detect_leaks=0:detect_container_overflow=0")
-    # -fPIC required for some ASAN build configurations (generally .so), so just enable it unversally here
-    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+if (Halide_CLANG_TIDY_BUILD)
+    message(STATUS "Ignoring all sanitizers because Clang-Tidy is in use.")
 else()
```
LebedevRI, line 7: Here and elsewhere: can't you just `return()`?
(In general, i find halide's style of else after-return to be confusing..)

steven-johnson:
...unfortunately, I'm probably going to abandon this work, at least temporarily, as getting it right is depressingly tricky for Halide.

The problem here is that trying to mix ASAN-compiled code from different compilers is basically a nonstarter, and that's exactly what we're trying to do here: Halide code from LLVM (either AOT or JIT), combined with 'shell' code from either AppleClang (OSX) or GCC (most Linux). Getting this working means using LLVMClang to compile ~everything (not just our bitcode), and that means getting all the platform-specific libraries, include paths, etc correct (including guessing at where CMake is helpfully injecting those for us and overriding them). None of this is rocket science, but it's stuff I don't have to finish at the moment due to other stuff taking priority.

(ASAN has been enabled and tested for Halide inside Google for a long time, but the nature of the Google codebase makes it refreshingly easy to enable there.)

I'm gonna leave this open as a draft in case anyone else wants to build on it in hopes of getting it working.

PR #6882, opened 2022-07-22
[vulkan phase1] Add SPIR-V IR 

derek-gerstmann:
This set of changes is Phase1 for adding Vulkan support, which adds a standalone class hierarchy for building and encoding a subset of SPIR-V (which is utilized in the forthcoming Vulkan CodeGen as part of Phase2). Specifically, it adds the following files with new class interfaces:

```
src/SpirvIR.h
src/SpirvIR.cpp
```

Basic internal test coverage has been placed in the `spirv_ir_test()` method and integrated into:

`test/internal.cpp`

Additionally, the CMake infrastructure has been updated to pull in the official SPIR-V headers from the Khronos repo (or optionally use an existing checkout in a specific folder).

There is a CMake option for enabling TARGET_SPIRV (which is set to ON by default with this commit to allow tests to run).  We can disable this prior to merging in main if that's preferred.


steven-johnson:
Is this ready to review? There are apparently lots of broken things on the bots.

derek-gerstmann:
> Is this ready to review? There are apparently lots of broken things on the bots.

Sorry … I’m still working through these failures!

steven-johnson:
ok, no worries :-)

test/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 include(HalideTestHelpers)
 
+if (TARGET_SPIRV)
+    add_compile_definitions(WITH_SPIRV)
+endif ()
+
```
alexreinking, line 6: We avoid directory commands like `add_compile_definitions` in this codebase. This define should probably be attached to the `Halide` library via `target_compile_definitions(Halide INTERFACE WITH_SPIRV)`

src/CMakeLists.txt
```diff
@@ -523,6 +530,11 @@ if (TARGET_OPENGLCOMPUTE)
     target_compile_definitions(Halide PRIVATE WITH_OPENGLCOMPUTE)
 endif ()
 
+option(TARGET_SPIRV "Include SPIR-V target" ON)
+if (TARGET_SPIRV)
+    target_compile_definitions(Halide PRIVATE WITH_SPIRV)
```
alexreinking, line 535: This is duplicated

src/CMakeLists.txt
```diff
@@ -414,6 +416,11 @@ if (TARGET Halide_V8)
     target_compile_definitions(Halide PRIVATE WITH_V8)
 endif ()
 
+if (TARGET Halide_SPIRV)
+    target_link_libraries(Halide PRIVATE Halide_SPIRV)
+    target_compile_definitions(Halide PRIVATE WITH_SPIRV)
```
alexreinking, line 421: This is duplicated

dependencies/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ add_subdirectory(llvm)
 
 add_subdirectory(jpeg)
 add_subdirectory(png)
+add_subdirectory(spirv)
```
alexreinking, line 28: This is where `if (TARGET_SPIRV)` should go. Then unindent `spirv/CMakeLists.txt`.

dependencies/spirv/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+    if (DEFINED HALIDE_SPIRV_INCLUDE_DIR)
+        add_library(Halide_SPIRV INTERFACE)
+        target_include_directories(Halide_SPIRV SYSTEM INTERFACE ${HALIDE_SPIRV_INCLUDE_DIR})
+        message(STATUS "Using SPIR-V headers from ${HALIDE_SPIRV_INCLUDE_DIR}...")
+    else()
+        message(FATAL_ERROR "SPIRV-Headers were not found - please enable FETCH_SPIRV_HEADERS or specify a valid path with 'SPIRV-Headers_SOURCE_DIR'!")
+    endif()
+
+endif()
```
alexreinking, line 37: Here is a correct implementation:

```suggestion
cmake_dependent_option(
    FETCH_SPIRV_HEADERS "Fetch SPIR-V headers from remote repo (disable and use SPIRV-Headers_SOURCE_DIR to use an existing checkout)." OFF
    TARGET_SPIRV OFF
)

if (FETCH_SPIRV_HEADERS)
    set(SPIRV_VERSION 1.5.4)
    message(STATUS "Fetching SPIR-V ${SPIRV_VERSION}...")
    FetchContent_Declare(
        SPIRV-Headers
        GIT_REPOSITORY https://github.com/KhronosGroup/SPIRV-Headers.git
        GIT_TAG ${SPIRV_VERSION}
        GIT_SHALLOW TRUE
        GIT_SUBMODULES ""
    )

    FetchContent_MakeAvailable(SPIRV-Headers)

    set(SPIRV_INCLUDE_DIR ${SPIRV-Headers_SOURCE_DIR}/include)    
else ()
    find_path(SPIRV_INCLUDE_DIR spirv/1.0/spirv.h)
endif ()

if (NOT SPIRV_INCLUDE_DIR)
    message(FATAL_ERROR "SPIRV-Headers were not found - please enable FETCH_SPIRV_HEADERS or specify a valid path with 'SPIRV_INCLUDE_DIR'!")
endif ()

message(STATUS "Using SPIR-V headers from ${SPIRV_INCLUDE_DIR}...")

add_library(Halide_SPIRV INTERFACE IMPORTED)
target_include_directories(Halide_SPIRV INTERFACE "${SPIRV_INCLUDE_DIR}")
```

alexreinking: I would prefer to move Halide away from using FetchContent for anything by default. It's okay (until we reach CMake 3.24) to include it as an opt-in convenience, but we've already had issues raised with people trying to build us in air-gapped environments. Better to get an error message we control than one that some internal CMake module controls.

In the case when FetchContent is not used, we should at least _try_ to look for the header we're using. That's the purpose of the `find_path` command.

Int he case when FetchContent is used, since we are not using their CMake targets (and are instead making our own), we shouldn't use `MakeAvailable` since that will run all of their CMake code inside our project. Instead, we'll just `Populate` it to trigger the download, but nothing else.

src/CMakeLists.txt
```diff
@@ -414,6 +416,11 @@ if (TARGET Halide_V8)
     target_compile_definitions(Halide PRIVATE WITH_V8)
 endif ()
 
+if (TARGET Halide_SPIRV)
+    target_link_libraries(Halide PRIVATE Halide_SPIRV)
```
alexreinking, line 420:
```suggestion
    target_link_libraries(Halide PRIVATE Halide::SPIRV)
```

src/SpirvIR.h
```diff
@@ -0,0 +1,557 @@
...
+#include <map>
+#include <set>
+#include <stack>
+#include <unordered_map>
+#include <vector>
+
+#include "IntrusivePtr.h"
+#include "Type.h"
+
+#include <spirv/1.0/spirv.h>  // Use v1.0 spec as the minimal viable version (for maximum compatiblity)
```
alexreinking, line 22: This will end up in the user-supplied `Halide.h` which means consumers will need to have the SPIR-V headers available. Is this _necessary_? If so, new code will need to be written in our package config file to handle creating the `Halide::SPIRV` target. That likely means writing a Find module that calls `find_path` the same way I do below.

If this `#include` can be confined to `SpirvIR.cpp`, then the situation is simpler.

steven-johnson: I'll go further and say that it **cannot** be necessary. We must not require these headers for Halide.h.

derek-gerstmann: Hrmmm ... we only include a single stand-alone header file which is the SPIR-V typedefs and enums, so there's no additional dependencies.

test/CMakeLists.txt
derek-gerstmann: So, the issue I faced here was the precompiled headers don't grab target specific compile definitions, only project level compile definitions.  Also, `$<BUILD_INTERFACE:>` doesn't appear to pass the definitions down for compiling object files.

derek-gerstmann: I worked around the precompiled headers by always including a stub function for the internal test.

src/SpirvIR.h
derek-gerstmann: Specifically, this file: [spirv.h](https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/1.0/spirv.h)

test/CMakeLists.txt
alexreinking: > the precompiled headers don't grab target specific compile definitions, only project level compile definitions

I don't think that's _quite_ true. They grab the target-specific compile definitions for `_test_internal` and then the same header is reused everywhere. See `test/CMakeLists.txt` and `cmake/HalideTestHelpers.cmake` for the relevant uses of `target_precompile_headers`.

derek-gerstmann: Hrmm ... I'm not entirely sure why, but the precompiled headers for Halide.h resulted in an error saying that they were invalid if I defined a target_compile_definitions(_internal_test PRIVATE WITH_SPIRV), even though it was also defined in the Halide compile definitions.

dependencies/spirv/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+endif ()
+
+if (NOT SPIRV_INCLUDE_DIR)
+    message(FATAL_ERROR "SPIRV-Headers were not found - please enable FETCH_SPIRV_HEADERS or specify a valid path with 'SPIRV_INCLUDE_DIR'!")
+endif ()
+
+message(STATUS "Using SPIR-V headers from ${SPIRV_INCLUDE_DIR}...")
+add_library(Halide_SPIRV INTERFACE IMPORTED)
+target_include_directories(Halide_SPIRV 
+                           SYSTEM # Use -isystem instead of -I; this is a trick so that clang-tidy won't analyze these includes
```
alexreinking, line 35: [I think](https://cmake.org/cmake/help/latest/prop_tgt/NO_SYSTEM_FROM_IMPORTED.html) `SYSTEM` is unnecessary here since `Halide_SPIRV` is `IMPORTED`. It should also be renamed to `Halide::SPIRV` to ensure that CMake always treats this name as a target (and to therefore be defensive against future typos).

derek-gerstmann: I renamed the library target, but left the SYSTEM declaration so that the <> notation for the include file path still worked.

derek-gerstmann:
Hrmm ... I don't understand why the packaging script fails and I can't see the log output.  Any tips?

dependencies/spirv/CMakeLists.txt
derek-gerstmann: Actually, this didn't work, and I reverted back to Halide_SPIRV target names since the dependencies need to exist before Halide itself, and Halide::SPIRV was getting silently ignored for linkage.

src/SpirvIR.h
zvookin: This seems require an external header from a thirdparty to be present to compile Halide. The obvious solution to that is to add the file to this PR. It also has a different license, which should not be a problem, but... it kinda always is... I'd also look at putting this code in a subdirectory. I know we have not done so for other stuff but it is getting to be time to do that. It may be enough to just add the Khronos header to a thirdparty directory and include it from there. (We have done some things like this for the BitWriter support and the NVidia ptx libs.)

I do agree that maybe its time to have some stuff in the compiler that isn't in Halide.h...

dependencies/spirv/CMakeLists.txt
alexreinking: I would like to repro that since it's not at all what I would expect. Do you have some steps?

src/SpirvIR.h
derek-gerstmann: This would certainly simplify things. I can add the header file to the dependencies/spirv folder along with the Khronos license if that works.

abadams: Why is this going to end up in Halide.h though? Can we just keep SpirvIR.h out of Halide.h?

dependencies/spirv/CMakeLists.txt
derek-gerstmann: Just changing the target name from `Halide_SPIRV` to `Halide::SPIRV` like this:

`add_library(Halide::SPIRV INTERFACE)`
`target_include_directories(Halide::SPIRV SYSTEM INTERFACE "${SPIRV_INCLUDE_DIR}")`

Causes errors with cmake v3.23.2:

`CMake Error at dependencies/spirv/CMakeLists.txt:35 (add_library):
  The target name "Halide::SPIRV" is reserved or not valid for certain CMake
  features, such as generator expressions, and may result in undefined
  behavior.

CMake Error at dependencies/spirv/CMakeLists.txt:36 (target_include_directories):
  Cannot specify include directories for target "Halide::SPIRV" which is not
  built by this project.
`

derek-gerstmann: Using the old target name and adding an ALIAS for Halide::SPIRV like so:

````
add_library(Halide_SPIRV INTERFACE)
add_library(Halide::SPIRV ALIAS Halide_SPIRV)
target_include_directories(Halide_SPIRV SYSTEM INTERFACE "${SPIRV_INCLUDE_DIR}")
````

And used like before:

````
if (TARGET Halide::SPIRV)
    target_link_libraries(Halide PRIVATE Halide::SPIRV)
endif()

````
DOES add the correct -isystem path to the compiler definitions ... but then we incur a dependency on Halide_TARGETS for the interface library Halide::SPIRV which isn't installed.

File: src/SpirvIR.h (hunk 7 above)
Comment (derek-gerstmann) [reply src/SpirvIR.h:22]: Absolutely ... the current use of SpirvIR.h is only going to be inside of CodeGen_Vulkan.cpp's implementation (forthcoming).  If there's a better way to organize this to make this clear, I'm more than happy to restructure things!

Comment (abadams) [reply src/SpirvIR.h:22]: ok, so we just need to take it out of the list of headers that makes up Halide.h, and it'll be fine. Looking at the change to the Makefile, it's currently in the list. I didn't check the cmake build.

File: src/SpirvIR.cpp [hunk 9]
```diff
@@ -0,0 +1,1710 @@
...
+    }
+
+    const std::string key = hash_constant(type, data);
+    ConstantMap::const_iterator it = constant_map.find(key);
+    if (it != constant_map.end()) {
+        return it->second;
+    }
+
+    Type scalar_type = type.with_lanes(1);
+    std::vector<SpvId> components;
```
steven-johnson, line 1064: nit: `components.reserve(type.lanes());` seems worthwhile here

src/SpirvIR.cpp
```diff
@@ -0,0 +1,1710 @@
...
+    uint32_t word_count = 1;
+    word_count += has_type() ? 1 : 0;
+    word_count += has_result() ? 1 : 0;
+    word_count += length();
+
+    // Preface the instruction with the format
+    // - high 16-bits indicate instruction length (number of 32-bit words)
+    // - low 16-bits indicate op code
+    binary.push_back(((word_count) << SpvWordCountShift) | contents->op_code);
+    if (has_type()) { binary.push_back(contents->type_id); }
```
steven-johnson, line 131: Style nit, here and elsewhere: if statements should never be on the same line like this (I'm suprised that clang-format didn't flag this)
```
    if (has_type()) { 
        binary.push_back(contents->type_id); 
    }
```

steven-johnson (approved):
LGTM with nits

dependencies/spirv/CMakeLists.txt
alexreinking: You would have to change from:
```
add_library(Halide_SPIRV INTERFACE)
```
to
```
add_library(Halide::SPIRV INTERFACE IMPORTED)
```
since names with colons `::` can only be used in `IMPORTED` and `ALIAS` targets. When a target is `IMPORTED` its `INTERFACE_INCLUDE_DIRECTORIES` property (which is set by `target_include_directories(... INTERFACE ...)`) is treated as `SYSTEM` by default.

> but then we incur a dependency on Halide_TARGETS for the interface library Halide::SPIRV which isn't installed

Since `Halide::SPIRV` is meant to be _totally_ private, just write
```
target_link_libraries(Halide PRIVATE "$<BUILD_INTERFACE:Halide::SPIRV>")
```
instead.

test/CMakeLists.txt
alexreinking: You should look at the verbose output from a Ninja build to diagnose this. I wrote the following sample build to probe the behavior of `target_precompile_headers` and what I said was correct re: which flags it sees.

```
cmake_minimum_required(VERSION 3.23)
project(test)

string(APPEND CMAKE_CXX_FLAGS " -DGLOBAL")

add_compile_definitions(DIRECTORY)

add_library(impl impl.cpp)
target_compile_definitions(impl PUBLIC $<BUILD_INTERFACE:PROPAGATED>)
target_precompile_headers(impl PRIVATE impl.hpp)

add_executable(app main.cpp)
target_compile_definitions(app PRIVATE LOCAL)
target_link_libraries(app PRIVATE impl)
target_precompile_headers(app REUSE_FROM impl)
```
Create empty files:
```
$ touch main.cpp impl.{hpp,cpp}
```
Configure the build with Ninja:
```
$ cmake -G Ninja -S . -B build 
```
Run a dry-run:
```
$ cmake --build build/ -- -nv
[1/5] /usr/bin/c++ -DDIRECTORY -DPROPAGATED  -DGLOBAL -Winvalid-pch -x c++-header -include /home/reinking/test/build/CMakeFiles/impl.dir/cmake_pch.hxx -MD -MT CMakeFiles/impl.dir/cmake_pch.hxx.gch -MF CMakeFiles/impl.dir/cmake_pch.hxx.gch.d -o CMakeFiles/impl.dir/cmake_pch.hxx.gch -c /home/reinking/test/build/CMakeFiles/impl.dir/cmake_pch.hxx.cxx
[2/5] /usr/bin/c++ -DDIRECTORY -DPROPAGATED  -DGLOBAL -Winvalid-pch -include /home/reinking/test/build/CMakeFiles/impl.dir/cmake_pch.hxx -MD -MT CMakeFiles/impl.dir/impl.cpp.o -MF CMakeFiles/impl.dir/impl.cpp.o.d -o CMakeFiles/impl.dir/impl.cpp.o -c /home/reinking/test/impl.cpp
[3/5] /usr/bin/c++ -DDIRECTORY -DLOCAL -DPROPAGATED  -DGLOBAL -Winvalid-pch -include /home/reinking/test/build/CMakeFiles/impl.dir/cmake_pch.hxx -MD -MT CMakeFiles/app.dir/main.cpp.o -MF CMakeFiles/app.dir/main.cpp.o.d -o CMakeFiles/app.dir/main.cpp.o -c /home/reinking/test/main.cpp
[4/5] : && /usr/bin/cmake -E rm -f libimpl.a && /usr/bin/ar qc libimpl.a  CMakeFiles/impl.dir/impl.cpp.o && /usr/bin/ranlib libimpl.a && :
[5/5] : && /usr/bin/c++ -DGLOBAL  CMakeFiles/app.dir/main.cpp.o -o app  libimpl.a && :
```

src/SpirvIR.cpp
```diff
@@ -0,0 +1,1710 @@
...
+
+    SpvId pointer_type_id = declare_id(SpvPointerTypeId);
+    SpvInstruction inst = SpvFactory::pointer_type(pointer_type_id, storage_class, base_type_id);
+    module.add_type(inst);
+    pointer_type_map[key] = pointer_type_id;
+    return pointer_type_id;
+}
+
+SpvBuilder::ConstantKey SpvBuilder::hash_constant(const Type &type, const void *data) const {
+    ConstantKey key(type.bytes() + 4, ' ');
```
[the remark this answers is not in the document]
derek-gerstmann: Yes, you're entirely correct ... the function name is misleading, since the hashing actually occurs in unordered_map. Unfortunately, none of the SPIR-V types have a fixed size (SPIR-V allows arbitrary vector widths and even arbitrary precision for integers and floats). We could adopt our own hashing scheme and produce a uint64_t value, but I assumed that std::hash would be more sophisticated.  A quick search reveals that libstdc++ uses MurmurHashUnaligned2 for its string hashing, which seems like a good choice.

dependencies/spirv/CMakeLists.txt
derek-gerstmann: Cool ... thanks for the detailed explanation!

src/SpirvIR.cpp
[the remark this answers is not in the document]
derek-gerstmann: Perhaps renaming the fake hash methods into make_key methods would be better?  E.g.

`SpvBuilder::hash_constant() => SpvBuilder::make_constant_key()`

derek-gerstmann:
Looks like the only failure was the unrelated correctness_atomics test on the flaky Windows box. I'll change the default for the cmake option TARGET_SPIRV to OFF to disable this before merging to main.

steven-johnson:
> correctness_atomics test on the flaky Windows box

we really need to figure out what the deal with that test is.

PR #6872, opened 2022-07-20
Add support for generating x86 sum-of-absolute-difference reductions

abadams:
x86 has an instruction that computes a horizontal sum of groups of 8 pairwise absolute differences of uint8s. We can represent this with a VectorReduce node easily enough. This PR adds support for it.

PR #6884, opened 2022-07-25
[x86] Separate vector instruction selection and CodeGen passes

rootjalex:
Left as a draft for now because this work is unfinished, but I was hoping to get feedback.

This PR is intended to create a separate vector-instruction-selection pass for x86, akin to HexagonOptimize. That pass is defined in X86Optimize.(h | cpp), and as many rules as possible are written via the IRMatch.h templating. In order to successfully create this TRS, this PR also adds:
- a VectorIntrinsic IR Node, which is not restricted to element-wise operations (i.e. it supports `dot_product`s)
- updates to IRMatch.h, i.e. supporting bitwise operations and type hints via a new `typed()` method

Work that still needs to be done (possibly):
- [] Implement VectorReduce logic similar to `CodeGen_LLVM::codegen_vector_reduce` to split up VectorReduce nodes
- [] Implement a base class for ^ and possibly for division + fixed-point intrinsics lowering (?)
- [] Add the SapphireRapids float32 + bfloat16 accumulating `dot_product` pattern, and the `psadbw` from #6872 

This TRS addresses the issue that @abadams raised on #6878, namely that rewriting `horizontal_widening_add` directly into Halide IR that corresponds to the `dot_product` instructions is cleaner and more extensible than the implementation in that PR.

I am tagging people that might be interested, but would love feedback from anyone (especially on the points above, and anywhere in the code that says TODO / FIXME).

### DIFF
--- a/src/CodeGen_LLVM.h
+++ b/src/CodeGen_LLVM.h
@@ -362,6 +362,7 @@ protected:
     void visit(const IfThenElse *) override;
     void visit(const Evaluate *) override;
     void visit(const Shuffle *) override;
+    void visit(const VectorIntrinsic *) override;
     void visit(const VectorReduce *) override;
     void visit(const Prefetch *) override;
     void visit(const Atomic *) override;
--- a/src/CodeGen_X86.cpp
+++ b/src/CodeGen_X86.cpp
@@ -7,6 +7,7 @@
 #include "LLVM_Headers.h"
 #include "Simplify.h"
 #include "Util.h"
+#include "X86Optimize.h"
 
 namespace Halide {
 namespace Internal {
@@ -20,30 +21,6 @@ using namespace llvm;
 #if defined(WITH_X86)
 
 namespace {
-
-// Populate feature flags in a target according to those implied by
-// existing flags, so that instruction patterns can just check for the
-// oldest feature flag that supports an instruction.
-Target complete_x86_target(Target t) {
-    if (t.has_feature(Target::AVX512_SapphireRapids)) {
-        t.set_feature(Target::AVX512_Cannonlake);
-    }
-    if (t.has_feature(Target::AVX512_Cannonlake)) {
-        t.set_feature(Target::AVX512_Skylake);
-    }
-    if (t.has_feature(Target::AVX512_Cannonlake) ||
-        t.has_feature(Target::AVX512_Skylake) ||
-        t.has_feature(Target::AVX512_KNL)) {
-        t.set_feature(Target::AVX2);
-    }
-    if (t.has_feature(Target::AVX2)) {
-        t.set_feature(Target::AVX);
-    }
-    if (t.has_feature(Target::AVX)) {
-        t.set_feature(Target::SSE41);
-    }
-    return t;
-}
 
 /** A code generator that emits x86 code from a given Halide stmt. */
 class CodeGen_X86 : public CodeGen_Posix {
@@ -69,10 +46,7 @@ protected:
 
     /** Nodes for which we want to emit specific sse/avx intrinsics */
     // @{
-    void visit(const Add *) override;
-    void visit(const Sub *) override;
     void visit(const Cast *) override;
-    void visit(const Call *) override;
     void visit(const GT *) override;
     void visit(const LT *) override;
     void visit(const LE *) override;
@@ -83,7 +57,7 @@ protected:
     void visit(const Allocate *) override;
     void visit(const Load *) override;
     void visit(const Store *) override;
-    void codegen_vector_reduce(const VectorReduce *, const Expr &init) override;
+    void visit(const VectorIntrinsic *) override;
     // @}
 
 private:
@@ -265,85 +239,6 @@ void CodeGen_X86::init_module() {
     }
 }
 
-// i32(i16_a)*i32(i16_b) +/- i32(i16_c)*i32(i16_d) can be done by
-// interleaving a, c, and b, d, and then using dot_product.
-bool should_use_dot_product(const Expr &a, const Expr &b, vector<Expr> &result) {
-    Type t = a.type();
-    internal_assert(b.type() == t);
-
-    if (!(t.is_int() && t.bits() == 32 && t.lanes() >= 4)) {
-        return false;
-    }
-
-    const Call *ma = Call::as_intrinsic(a, {Call::widening_mul});
-    const Call *mb = Call::as_intrinsic(b, {Call::widening_mul});
-    // dot_product can't handle mixed type widening muls.
-    if (ma && ma->args[0].type() != ma->args[1].type()) {
-        return false;
-    }
-    if (mb && mb->args[0].type() != mb->args[1].type()) {
-        return false;
-    }
-    // If the operands are widening shifts, we might be able to treat these as
-    // multiplies.
-    const Call *sa = Call::as_intrinsic(a, {Call::widening_shift_left});
-    const Call *sb = Call::as_intrinsic(b, {Call::widening_shift_left});
-    if (sa && !is_const(sa->args[1])) {
-        sa = nullptr;
-    }
-    if (sb && !is_const(sb->args[1])) {
-        sb = nullptr;
-    }
-    if ((ma || sa) && (mb || sb)) {
-        Expr a0 = ma ? ma->args[0] : sa->args[0];
-        Expr a1 = ma ? ma->args[1] : lossless_cast(sa->args[0].type(), simplify(make_const(sa->type, 1) << sa->args[1]));
-        Expr b0 = mb ? mb->args[0] : sb->args[0];
-        Expr b1 = mb ? mb->args[1] : lossless_cast(sb->args[0].type(), simplify(make_const(sb->type, 1) << sb->args[1]));
-        if (a1.defined() && b1.defined()) {
-            std::vector<Expr> args = {a0, a1, b0, b1};
-            result.swap(args);
-            return true;
-        }
-    }
-    return false;
-}
-
-void CodeGen_X86::visit(const Add *op) {
-    vector<Expr> matches;
-    if (should_use_dot_product(op->a, op->b, matches)) {
-        Expr ac = Shuffle::make_interleave({matches[0], matches[2]});
-        Expr bd = Shuffle::make_interleave({matches[1], matches[3]});
-        value = call_overloaded_intrin(op->type, "dot_product", {ac, bd});
-        if (value) {
-            return;
-        }
-    }
-    CodeGen_Posix::visit(op);
-}
-
-void CodeGen_X86::visit(const Sub *op) {
-    vector<Expr> matches;
-    if (should_use_dot_product(op->a, op->b, matches)) {
-        // Negate one of the factors in the second expression
-        Expr negative_2 = lossless_negate(matches[2]);
-        Expr negative_3 = lossless_negate(matches[3]);
-        if (negative_2.defined() || negative_3.defined()) {
-            if (negative_2.defined()) {
-                matches[2] = negative_2;
-            } else {
-                matches[3] = negative_3;
-            }
-            Expr ac = Shuffle::make_interleave({matches[0], matches[2]});
-            Expr bd = Shuffle::make_interleave({matches[1], matches[3]});
-            value = call_overloaded_intrin(op->type, "dot_product", {ac, bd});
-            if (value) {
-                return;
-            }
-        }
-    }
-    CodeGen_Posix::visit(op);
-}
-
 void CodeGen_X86::visit(const GT *op) {
     Type t = op->a.type();
 
@@ -450,41 +345,10 @@ void CodeGen_X86::visit(const Select *op) {
 }
 
 void CodeGen_X86::visit(const Cast *op) {
-
     if (!op->type.is_vector()) {
         // We only have peephole optimizations for vectors in here.
         CodeGen_Posix::visit(op);
         return;
-    }
-
-    struct Pattern {
-        string intrin;
-        Expr pattern;
-    };
-
-    // clang-format off
-    static Pattern patterns[] = {
-        // This isn't rounding_multiply_quantzied(i16, i16, 15) because it doesn't
-        // saturate the result.
-        {"pmulhrs", i16(rounding_shift_right(widening_mul(wild_i16x_, wild_i16x_), 15))},
-
-        {"saturating_narrow", i16_sat(wild_i32x_)},
-        {"saturating_narrow", u16_sat(wild_i32x_)},
-        {"saturating_narrow", i8_sat(wild_i16x_)},
-        {"saturating_narrow", u8_sat(wild_i16x_)},
-
-        {"f32_to_bf16", bf16(wild_f32x_)},
-    };
-    // clang-format on
-
-    vector<Expr> matches;
-    for (const Pattern &p : patterns) {
-        if (expr_match(p.pattern, op, matches)) {
-            value = call_overloaded_intrin(op->type, p.intrin, matches);
-            if (value) {
-                return;
-            }
-        }
     }
 
     if (const Call *mul = Call::as_intrinsic(op->value, {Call::widening_mul})) {
@@ -499,216 +363,6 @@ void CodeGen_X86::visit(const Cast *op) {
     }
 
     CodeGen_Posix::visit(op);
-}
-
-void CodeGen_X86::visit(const Call *op) {
-    if (!op->type.is_vector()) {
-        // We only have peephole optimizations for vectors in here.
-        CodeGen_Posix::visit(op);
-        return;
-    }
-
-    // A 16-bit mul-shift-right of less than 16 can sometimes be rounded up to a
-    // full 16 to use pmulh(u)w by left-shifting one of the operands. This is
-    // handled here instead of in the lowering of mul_shift_right because it's
-    // unlikely to be a good idea on platforms other than x86, as it adds an
-    // extra shift in the fully-lowered case.
-    if ((op->type.element_of() == UInt(16) ||
-         op->type.element_of() == Int(16)) &&
-        op->is_intrinsic(Call::mul_shift_right)) {
-        internal_assert(op->args.size() == 3);
-        const uint64_t *shift = as_const_uint(op->args[2]);
-        if (shift && *shift < 16 && *shift >= 8) {
-            Type narrow = op->type.with_bits(8);
-            Expr narrow_a = lossless_cast(narrow, op->args[0]);
-            Expr narrow_b = narrow_a.defined() ? Expr() : lossless_cast(narrow, op->args[1]);
-            int shift_left = 16 - (int)(*shift);
-            if (narrow_a.defined()) {
-                codegen(mul_shift_right(op->args[0] << shift_left, op->args[1], 16));
-                return;
-            } else if (narrow_b.defined()) {
-                codegen(mul_shift_right(op->args[0], op->args[1] << shift_left, 16));
-                return;
-            }
-        }
-    } else if (op->type.is_int() &&
-               op->type.bits() <= 16 &&
-               op->is_intrinsic(Call::rounding_halving_add)) {
-        // We can redirect signed rounding halving add to unsigned rounding
-        // halving add by adding 128 / 32768 to the result if the sign of the
-        // args differs.
-        internal_assert(op->args.size() == 2);
-        Type t = op->type.with_code(halide_type_uint);
-        Expr a = cast(t, op->args[0]);
-        Expr b = cast(t, op->args[1]);
-        codegen(cast(op->type, rounding_halving_add(a, b) + ((a ^ b) & (1 << (t.bits() - 1)))));
-        return;
-    } else if (op->is_intrinsic(Call::absd)) {
-        internal_assert(op->args.size() == 2);
-        if (op->args[0].type().is_uint()) {
-            // On x86, there are many 3-instruction sequences to compute absd of
-            // unsigned integers. This one consists solely of instructions with
-            // throughput of 3 ops per cycle on Cannon Lake.
-            //
-            // Solution due to Wojciech Mula:
-            // http://0x80.pl/notesen/2018-03-11-sse-abs-unsigned.html
-            codegen(saturating_sub(op->args[0], op->args[1]) | saturating_sub(op->args[1], op->args[0]));
-            return;
-        } else if (op->args[0].type().is_int()) {
-            codegen(Max::make(op->args[0], op->args[1]) - Min::make(op->args[0], op->args[1]));
-            return;
-        }
-    }
-
-    struct Pattern {
-        string intrin;
-        Expr pattern;
-    };
-
-    // clang-format off
-    static Pattern patterns[] = {
-        {"pmulh", mul_shift_right(wild_i16x_, wild_i16x_, 16)},
-        {"pmulh", mul_shift_right(wild_u16x_, wild_u16x_, 16)},
-        {"saturating_pmulhrs", rounding_mul_shift_right(wild_i16x_, wild_i16x_, 15)},
-    };
-    // clang-format on
-
-    vector<Expr> matches;
-    for (const auto &pattern : patterns) {
-        if (expr_match(pattern.pattern, op, matches)) {
-            value = call_overloaded_intrin(op->type, pattern.intrin, matches);
-            if (value) {
-                return;
-            }
-        }
-    }
-
-    CodeGen_Posix::visit(op);
-}
-
-void CodeGen_X86::codegen_vector_reduce(const VectorReduce *op, const Expr &init) {
-    if (op->op != VectorReduce::Add && op->op != VectorReduce::SaturatingAdd) {
-        CodeGen_Posix::codegen_vector_reduce(op, init);
-        return;
-    }
-    const int factor = op->value.type().lanes() / op->type.lanes();
-
-    struct Pattern {
-        VectorReduce::Operator reduce_op;
-        int factor;
-        Expr pattern;
-        const char *intrin;
-        Type narrow_type;
-        uint32_t flags = 0;
-        enum {
-            CombineInit = 1 << 0,
-            SwapOperands = 1 << 1,
-            SingleArg = 1 << 2,
-        };
-    };
-    // clang-format off
-    // These patterns are roughly sorted "best to worst", in case there are two
-    // patterns that match the expression.
-    static const Pattern patterns[] = {
-        // 4-way dot products
-        {VectorReduce::Add, 4, i32(widening_mul(wild_u8x_, wild_i8x_)), "dot_product", {}, Pattern::CombineInit},
-        {VectorReduce::Add, 4, i32(widening_mul(wild_i8x_, wild_u8x_)), "dot_product", {}, Pattern::CombineInit | Pattern::SwapOperands},
-        {VectorReduce::SaturatingAdd, 4, i32(widening_mul(wild_u8x_, wild_i8x_)), "saturating_dot_product", {}, Pattern::CombineInit},
-        {VectorReduce::SaturatingAdd, 4, i32(widening_mul(wild_i8x_, wild_u8x_)), "saturating_dot_product", {}, Pattern::CombineInit | Pattern::SwapOperands},
-
-        // 2-way dot products
-        {VectorReduce::Add, 2, i32(widening_mul(wild_i8x_, wild_i8x_)), "dot_product", Int(16)},
-        {VectorReduce::Add, 2, i32(widening_mul(wild_i8x_, wild_u8x_)), "dot_product", Int(16)},
-        {VectorReduce::Add, 2, i32(widening_mul(wild_u8x_, wild_i8x_)), "dot_product", Int(16)},
-        {VectorReduce::Add, 2, i32(widening_mul(wild_u8x_, wild_u8x_)), "dot_product", Int(16)},
-        {VectorReduce::SaturatingAdd, 2, i32(widening_mul(wild_u8x_, wild_i8x_)), "saturating_dot_product", {}, Pattern::CombineInit},
-        {VectorReduce::SaturatingAdd, 2, i32(widening_mul(wild_i8x_, wild_u8x_)), "saturating_dot_product", {}, Pattern::CombineInit | Pattern::SwapOperands},
-        {VectorReduce::SaturatingAdd, 2, widening_mul(wild_u8x_, wild_i8x_), "saturating_dot_product"},
-        {VectorReduce::SaturatingAdd, 2, widening_mul(wild_i8x_, wild_u8x_), "saturating_dot_product", {}, Pattern::SwapOperands},
-
-        {VectorReduce::Add, 2, i32(widening_mul(wild_i16x_, wild_i16x_)), "dot_product", {}, Pattern::CombineInit},
-        {VectorReduce::Add, 2, i32(widening_mul(wild_i16x_, wild_i16x_)), "dot_product", Int(16)},
-        {VectorReduce::SaturatingAdd, 2, i32(widening_mul(wild_i16x_, wild_i16x_)), "saturating_dot_product", {}, Pattern::CombineInit},
-
-        {VectorReduce::Add, 2, wild_f32x_ * wild_f32x_, "dot_product", BFloat(16), Pattern::CombineInit},
-
-        // One could do a horizontal widening addition with
-        // other dot_products against a vector of ones. Currently disabled
-        // because I haven't found other cases where it's clearly better.
-
-        {VectorReduce::Add, 2, u16(wild_u8x_), "horizontal_widening_add", {}, Pattern::SingleArg},
-        {VectorReduce::Add, 2, i16(wild_u8x_), "horizontal_widening_add", {}, Pattern::SingleArg},
-        {VectorReduce::Add, 2, i16(wild_i8x_), "horizontal_widening_add", {}, Pattern::SingleArg},
-    };
-    // clang-format on
-
-    std::vector<Expr> matches;
-    for (const Pattern &p : patterns) {
-        if (op->op != p.reduce_op || p.factor != factor) {
-            continue;
-        }
-        if (expr_match(p.pattern, op->value, matches)) {
-            if (p.flags & Pattern::SingleArg) {
-                Expr a = matches[0];
-
-                if (p.narrow_type.bits() > 0) {
-                    a = lossless_cast(p.narrow_type.with_lanes(a.type().lanes()), a);
-                }
-                if (!a.defined()) {
-                    continue;
-                }
-
-                if (init.defined() && (p.flags & Pattern::CombineInit)) {
-                    value = call_overloaded_intrin(op->type, p.intrin, {init, a});
-                    if (value) {
-                        return;
-                    }
-                } else {
-                    value = call_overloaded_intrin(op->type, p.intrin, {a});
-                    if (value) {
-                        if (init.defined()) {
-                            Value *x = value;
-                            Value *y = codegen(init);
-                            value = builder->CreateAdd(x, y);
-                        }
-                        return;
-                    }
-                }
-            } else {
-                Expr a = matches[0];
-                Expr b = matches[1];
-                if (p.flags & Pattern::SwapOperands) {
-                    std::swap(a, b);
-                }
-                if (p.narrow_type.bits() > 0) {
-                    a = lossless_cast(p.narrow_type.with_lanes(a.type().lanes()), a);
-                    b = lossless_cast(p.narrow_type.with_lanes(b.type().lanes()), b);
-                }
-                if (!a.defined() || !b.defined()) {
-                    continue;
-                }
-
-                if (init.defined() && (p.flags & Pattern::CombineInit)) {
-                    value = call_overloaded_intrin(op->type, p.intrin, {init, a, b});
-                    if (value) {
-                        return;
-                    }
-                } else {
-                    value = call_overloaded_intrin(op->type, p.intrin, {a, b});
-                    if (value) {
-                        if (init.defined()) {
-                            Value *x = value;
-                            Value *y = codegen(init);
-                            value = builder->CreateAdd(x, y);
-                        }
-                        return;
-                    }
-                }
-            }
-        }
-    }
-
-    CodeGen_Posix::codegen_vector_reduce(op, init);
 }
 
 void CodeGen_X86::visit(const Allocate *op) {
@@ -741,6 +395,11 @@ void CodeGen_X86::visit(const Store *op) {
         return;
     }
     CodeGen_Posix::visit(op);
+}
+
+void CodeGen_X86::visit(const VectorIntrinsic *op) {
+    value = call_overloaded_intrin(op->type, op->name, op->args);
+    internal_assert(value) << "CodeGen_X86 failed on " << Expr(op) << "\n";
 }
 
 string CodeGen_X86::mcpu_target() const {
--- a/src/IR.cpp
+++ b/src/IR.cpp
@@ -901,6 +901,17 @@ Stmt Atomic::make(const std::string &producer_name,
     return node;
 }
 
+Expr VectorIntrinsic::make(Type type, const std::string &name, const std::vector<Expr> &args) {
+    user_assert(!name.empty()) << "VectorIntrinsic without a name\n";
+    user_assert(!args.empty()) << "VectorInrinsic without arguments\n";
+
+    VectorIntrinsic *node = new VectorIntrinsic;
+    node->type = type;
+    node->name = name;
+    node->args = args;
+    return node;
+}
+
 Expr VectorReduce::make(VectorReduce::Operator op,
                         Expr vec,
                         int lanes) {
@@ -1079,6 +1090,10 @@ void ExprNode<Call>::accept(IRVisitor *v) const {
 template<>
 void ExprNode<Shuffle>::accept(IRVisitor *v) const {
     v->visit((const Shuffle *)this);
+}
+template<>
+void ExprNode<VectorIntrinsic>::accept(IRVisitor *v) const {
+    v->visit((const VectorIntrinsic *)this);
 }
 template<>
 void ExprNode<VectorReduce>::accept(IRVisitor *v) const {
@@ -1268,6 +1283,10 @@ Expr ExprNode<Call>::mutate_expr(IRMutator *v) const {
 template<>
 Expr ExprNode<Shuffle>::mutate_expr(IRMutator *v) const {
     return v->visit((const Shuffle *)this);
+}
+template<>
+Expr ExprNode<VectorIntrinsic>::mutate_expr(IRMutator *v) const {
+    return v->visit((const VectorIntrinsic *)this);
 }
 template<>
 Expr ExprNode<VectorReduce>::mutate_expr(IRMutator *v) const {
--- a/src/IR.h
+++ b/src/IR.h
@@ -886,6 +886,15 @@ struct Atomic : public StmtNode<Atomic> {
     static const IRNodeType _node_type = IRNodeType::Atomic;
 };
 
+struct VectorIntrinsic : public ExprNode<VectorIntrinsic> {
+    std::string name;
+    std::vector<Expr> args;
+
+    static Expr make(Type type, const std::string &name, const std::vector<Expr> &args);
+
+    static const IRNodeType _node_type = IRNodeType::VectorIntrinsic;
+};
+
 /** Horizontally reduce a vector to a scalar or narrower vector using
  * the given commutative and associative binary operator. The reduction
  * factor is dictated by the number of lanes in the input and output
--- a/src/IRMatch.cpp
+++ b/src/IRMatch.cpp
@@ -296,6 +296,22 @@ public:
         }
     }
 
+    void visit(const VectorIntrinsic *op) override {
+        const VectorIntrinsic *e = expr.as<VectorIntrinsic>();
+        if (result && e &&
+            types_match(op->type, e->type) &&
+            e->name == op->name &&
+            e->args.size() == op->args.size()) {
+            for (size_t i = 0; result && (i < e->args.size()); i++) {
+                // FIXME: should we early-out? Here and in Call*
+                expr = e->args[i];
+                op->args[i].accept(this);
+            }
+        } else {
+            result = false;
+        }
+    }
+
     void visit(const VectorReduce *op) override {
         const VectorReduce *e = expr.as<VectorReduce>();
         if (result && e && op->op == e->op && types_match(op->type, e->type)) {
@@ -505,6 +521,9 @@ bool equal_helper(const BaseExprNode &a, const BaseExprNode &b) noexcept {
     case IRNodeType::Shuffle:
         return (equal_helper(((const Shuffle &)a).vectors, ((const Shuffle &)b).vectors) &&
                 equal_helper(((const Shuffle &)a).indices, ((const Shuffle &)b).indices));
+    case IRNodeType::VectorIntrinsic:
+        return (((const VectorIntrinsic &)a).name == ((const VectorIntrinsic &)b).name &&
+                equal_helper(((const VectorIntrinsic &)a).args, ((const VectorIntrinsic &)b).args));
     case IRNodeType::VectorReduce:
         // As with Cast above, we use equal instead of equal_helper
         // here, because while we know a.type == b.type, we don't know
--- a/src/IRMatch.h
+++ b/src/IRMatch.h
@@ -2198,6 +2198,39 @@ HALIDE_ALWAYS_INLINE auto cast(halide_type_t t, A &&a) noexcept -> CastOp<declty
     return {t, pattern_arg(a)};
 }
 
+// A node for expressing type hints, when rules are ambiguously typed.
+template<typename A>
+struct TypeHint {
+    struct pattern_tag {};
+    Type type;
+    A a;
+
+    constexpr static uint32_t binds = bindings<A>::mask;
+
+    constexpr static IRNodeType min_node_type = IRNodeType::Cast;
+    constexpr static IRNodeType max_node_type = IRNodeType::Cast;
+    constexpr static bool canonical = A::canonical;
+
+    HALIDE_ALWAYS_INLINE
+    Expr make(MatcherState &state, halide_type_t type_hint) const {
+        return a.make(state, type);
+    }
+
+    constexpr static bool foldable = false;
+};
+
+template<typename A>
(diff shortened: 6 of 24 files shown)


src/IRMatch.h
```diff
@@ -1839,6 +1881,109 @@ HALIDE_ALWAYS_INLINE auto ramp(A &&a, B &&b, C &&c) noexcept -> RampOp<decltype(
...
+        print_args<0>(0, s);
+    }
+
+    HALIDE_ALWAYS_INLINE
+    Expr make(MatcherState &state, halide_type_t type_hint) const {
+        std::vector<Expr> r_args(sizeof...(Args));
+        // TODO(rootjalex): How do we do type hints for the args?
+        // TODO(rootjalex): Is there a way to do basically an unrolled
+        // loop of the below? this is ugly.
+        // Supposedly C++20 will have constexpr std::transform, perhaps
```
steven-johnson, line 1944: It will likely be a *long* while before Halide can rely on C++20 being available.

rootjalex: Gotcha. Is there a reason why? I thought we just upgraded to C++17, so I figured C++20 wouldn't be that far off.

rootjalex: And thanks for taking a look at this!

steven-johnson: Anecdotally, C++20 is getting adopted at a *much* slower pace than C++11 or C++17 did. (It's not clear that there's even a timetable to adopt it inside Google.) We'd have to be confident that ~all significant Halide customers are ready to make that commitment.

rootjalex: Gotcha, thanks for the explanation! I guess I'll stick to the ugly `if constexpr` expressions :')

rootjalex:
@steven-johnson Thanks for the review! I believe I addressed all of your comments in ec2cd4e, other than fixing the weird IRVisitor / IRMutator behavior on VectorInstruction nodes - I'm still debugging a piece of that

rootjalex:
It appears that the HVX failures here are actually due to a bug in CSE that is removing the definition of `hexagon_module_state` before it's used - I am struggling to make a minimal example, so I will try to make a separate PR fixing that issue.

src/IR.h
```diff
@@ -886,16 +886,44 @@ struct Atomic : public StmtNode<Atomic> {
     static const IRNodeType _node_type = IRNodeType::Atomic;
 };
 
-/** Represent a target-specific vector instruction.
- *  Intrinsic may not be element-wise operation, i.e.
- *  dot_products. */
-struct VectorIntrinsic : public ExprNode<VectorIntrinsic> {
-    std::string name;
+/** Represent a length-agnostic and target-specific
```
zvookin, line 889: I gather the difference from the existing approach (```Call``` node) is that the result type and the input types do not have to have the same shape. Is this the defining characteristic of when one would use this vs. ```Call```?

The use of "target specific" needs to be clarified. E.g. some of the options are not target specific. Moreover, if the operation differs in some way from the exact definition of the machine instruction names, such as providing normalization of length, the differences from the machine instruction need to be documented.

LLVM is adding the vector predicated instructions and my guess is this is going to become the way to generate LLVM vector IR. These take predication masks and a vector length. One possibility is to add those parameters as field of this node.

Finally, how this node interacts with various code generation options needs to be documented. Specifically, I would expect actually target specific instructions -- e.g. those documented as for x86 -- to only be allowed when targeting the specific target. It may or may not be the case that it has to be via LLVM, but is asking for an x86 instruction on another architecture supposed to work? (I.e. are we planning to provide target independent lowering for all of these or only some of them?)

rootjalex: Thanks for the feedback! These are excellent questions/comments.

Yes, the main difference between `Call` nodes and this node is that `Call` nodes are element-wise operations, so they cannot express operations with a different output vector length than the inputs.

The intent behind "target specific" is that these nodes should not be created for backends that do not implement the operation. However, there are some operations that may exist on multiple backends (i.e. `abs`), that are expected to have the same semantics.

> One possibility is to add those parameters as field of this node.

That could absolutely be a possibility for the future, however this node is not intended to be used for every single vector operation (i.e. we see no reason to use it for a simple vector addition), only for target-specific instructions.

> Specifically, I would expect actually target specific instructions -- e.g. those documented as for x86 -- to only be allowed when targeting the specific target.

Yes, that is required for these nodes. I have yet to come up with a good abstraction for enforcing this though, especially because many of the x86-specific instructions actually require SSE41 (i.e. the predicated patterns in X86Optimize.cpp ensure that the target has the required feature before generating a `VectorInstruction` node).

> It may or may not be the case that it has to be via LLVM, but is asking for an x86 instruction on another architecture supposed to work? (I.e. are we planning to provide target independent lowering for all of these or only some of them?)

No, we absolutely do not intend to implement lowering for instructions on other architectures. x86 instruction nodes should only be generated when the target is x86 and that instruction is supported on the target. This is very different from `Call` intrinsics, which must all be supported on any architecture.
For example, x86 does not support `absd`, and patterns in X86Optimize.cpp contains patterns for efficiently lowering this instruction onto x86.

zvookin, line 889: Also, for the ones targeting machine instructions, is it expected the argument types exactly match the hardware specification? I expect the plan is to support some instructions that can be used on any of a number of types with a single name here. Also the order of the arguments may be difficult to infer from just the name even if one has the hardware spec handy. This will need to be documented somewhere and likely a comment on each name (enum member) is the way to do it.

zvookin, line 889: Another question is whether it is worth adding information, possibly optional, that describes the shape change for each instruction as a function of its argument shapes. Some are standard coherent shapes, some are pure reductions, etc.

rootjalex: The argument types are expected to be multiples of the hardware specification, so they can be split up into series of the target instructions. 
We already do support instructions with many different possible input types (i.e. `dot_product`), and we expect the backend to have a matching implementation.

> Also the order of the arguments may be difficult to infer from just the name even if one has the hardware spec handy.

The order of the arguments is expected to match the order of the arguments in the intrinsic definition, but I see your point. I'm not sure I have a good solution for this though.

> This will need to be documented somewhere and likely a comment on each name (enum member) is the way to do it.

Is that better than documenting in the intrinsics tables for each backend?

rootjalex: That could be interesting, though I'm not sure I know of a use case where that information is needed. Do you have an idea of some sort of analysis where this information is useful?

For x86, I would probably have `elementwise`, `reduce2`, `reduce4`, `accumulate_reduce2`, and `accumulate_reduce4`. I suspect that this list would grow quite quickly, so I am hesitant to attempt to track this information without concrete use cases

rootjalex:
Hexagon failures should be fixed by #6894. I don't know what the windows failures are though, and I don't have a windows machine to debug on yet.

steven-johnson:

rootjalex:
I need to get access to a Windows machine to debug the winbot failures. I will try to get access to one sometime this week - other than debugging that, this PR should be good to review.

steven-johnson:
Huh, the windows failures are indeed odd -- and the error messages unhelpful. Could it be an illegal instruction on those machines? WinBot1 and 2 are pretty old.

rootjalex:
> Huh, the windows failures are indeed odd -- and the error messages unhelpful.

Indeed :')

> Could it be an illegal instruction on those machines?

That is quite possible - I thought I had safely guarded all instructions, but it could be that I missed some.

> WinBot1 and 2 are pretty old.

Do you know which features they should have enabled? Are they just SSE2?

steven-johnson:
This is weird: if I run `correctness_simd_op_check` in `cmd.exe`, I get normal output and no failures. If, otoh, I run it in the "git bash" shell, I get no output at all... but I eventually get a segfault.

rootjalex:
.....That is super strange. I don't even know how to begin debugging that.

steven-johnson:
I did a litting debugging with WinDbg and eventually got a failure mode for simd_op_check that said:

```
HEAP[correctness_simd_op_check.exe]: HEAP: Free Heap block 00000144984FAFE0 modified at 00000144984FB058 after it was freed
(2954.1e4): Break instruction exception - code 80000003 (first chance)
```

with a stacktrace of

```
[0x6]   ucrtbase!_malloc_base + 0x36   
[0x7]   Halide!operator new + 0x1f   
[0x8]   Halide!Halide::Internal::VectorInstruction::make + 0x87   
[0x9]   Halide!Halide::Internal::IRMatcher::VectorInstructionOp<Halide::Internal::IRMatcher::Wild<0>,Halide::Internal::IRMatcher::SpecificExpr>::make + 0xdd   
[0xa]   Halide!Halide::Internal::IRMatcher::TypeHint<Halide::Internal::IRMatcher::VectorInstructionOp<Halide::Internal::IRMatcher::Wild<0>,Halide::Internal::IRMatcher::SpecificExpr> >::make + 0xe4   
[0xb]   Halide!Halide::Internal::IRMatcher::CastOp<Halide::Internal::IRMatcher::TypeHint<Halide::Internal::IRMatcher::VectorInstructionOp<Halide::Internal::IRMatcher::Wild<0>,Halide::Internal::IRMatcher::SpecificExpr> > >::make + 0xe4   
[0xc]   Halide!Halide::Internal::IRMatcher::Rewriter<Halide::Internal::IRMatcher::VectorReduceOp<Halide::Internal::IRMatcher::SpecificExpr,Halide::Internal::IRMatcher::IntLiteral,0> >::build_replacement<Halide::Internal::IRMatcher::CastOp<Halide::Internal::IRMatcher::TypeHint<Halide::Internal::IRMatcher::VectorInstructionOp<Halide::Internal::IRMatcher::Wild<0>,Halide::Internal::IRMatcher::SpecificExpr> > > > + 0x104   
[0xd]   Halide!Halide::Internal::`anonymous namespace'::Optimize_X86::visit + 0x164f   
[0xe]   Halide!Halide::Internal::ExprNode<Halide::Internal::VectorReduce>::mutate_expr + 0x1b   
[0xf]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x10]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x11]   Halide!Halide::Internal::`anonymous namespace'::mutate_binary_operator<Halide::Internal::Add> + 0x29   
[0x12]   Halide!Halide::Internal::IRMutator::visit + 0x14   
[0x13]   Halide!Halide::Internal::`anonymous namespace'::Optimize_X86::visit + 0x1de9   
[0x14]   Halide!Halide::Internal::ExprNode<Halide::Internal::Add>::mutate_expr + 0x1b   
[0x15]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x16]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x17]   Halide!Halide::Internal::IRMutator::visit + 0x46   
[0x18]   Halide!Halide::Internal::StmtNode<Halide::Internal::Store>::mutate_stmt + 0x18   
[0x19]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x1a]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x1b]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x1c]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x1d]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x1e]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x1f]   Halide!Halide::Internal::IRMutator::visit + 0x1d   
[0x20]   Halide!Halide::Internal::StmtNode<Halide::Internal::ProducerConsumer>::mutate_stmt + 0x1b   
[0x21]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x22]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x23]   Halide!Halide::Internal::IRMutator::visit + 0x26   
[0x24]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x25]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x26]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x27]   Halide!Halide::Internal::IRMutator::visit + 0x47   
[0x28]   Halide!Halide::Internal::StmtNode<Halide::Internal::Allocate>::mutate_stmt + 0x18   
[0x29]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x2a]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x2b]   Halide!Halide::Internal::IRMutator::visit + 0x55   
[0x2c]   Halide!Halide::Internal::StmtNode<Halide::Internal::For>::mutate_stmt + 0x18   
[0x2d]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x2e]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x2f]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x30]   Halide!Halide::Internal::StmtNode<Halide::Internal::LetStmt>::mutate_stmt + 0x1b   
[0x31]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x32]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x33]   Halide!Halide::Internal::IRMutator::visit + 0x55   
[0x34]   Halide!Halide::Internal::StmtNode<Halide::Internal::For>::mutate_stmt + 0x18   
[0x35]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x36]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x37]   Halide!Halide::Internal::IRMutator::visit + 0x1d   
[0x38]   Halide!Halide::Internal::StmtNode<Halide::Internal::ProducerConsumer>::mutate_stmt + 0x1b   
[0x39]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x3a]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x3b]   Halide!Halide::Internal::IRMutator::visit + 0x26   
[0x3c]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x3d]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x3e]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x3f]   Halide!Halide::Internal::IRMutator::visit + 0x47   
[0x40]   Halide!Halide::Internal::StmtNode<Halide::Internal::Allocate>::mutate_stmt + 0x18   
[0x41]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x42]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x43]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x44]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x45]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x46]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x47]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x48]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x49]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x4a]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x4b]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x4c]   Halide!Halide::Internal::StmtNode<Halide::Internal::LetStmt>::mutate_stmt + 0x1b   
[0x4d]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x4e]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x4f]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x50]   Halide!Halide::Internal::StmtNode<Halide::Internal::LetStmt>::mutate_stmt + 0x1b   
[0x51]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x52]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x53]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x54]   Halide!Halide::Internal::StmtNode<Halide::Internal::LetStmt>::mutate_stmt + 0x1b   
[0x55]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x56]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x57]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x58]   Halide!Halide::Internal::StmtNode<Halide::Internal::LetStmt>::mutate_stmt + 0x1b   
[0x59]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x5a]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x5b]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x5c]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x5d]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x5e]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x5f]   Halide!Halide::Internal::IRMutator::visit + 0x3c   
[0x60]   Halide!Halide::Internal::StmtNode<Halide::Internal::Block>::mutate_stmt + 0x18   
[0x61]   Halide!Halide::Internal::IRMutator::mutate + 0x28   
[0x62]   Halide!Halide::Internal::IRGraphMutator::mutate + 0x10f   
[0x63]   Halide!Halide::Internal::optimize_x86_instructions + 0x45   
```

steven-johnson:
FYI: the crash above happens reliably on Windows, but I haven't been able to trigger anything similar on OSX or Linux builds. I'm gonna try pulling this into an experimental branch at Google to run with more sanitizers.

steven-johnson:
So I'm at a bit of a loss to explain the apparent use-after-free on Windows; I pulled these changes into Google and ran this test on Linux-x64 under ASAN, TSAN, and MSAN, and none of them reported *any* issues. I suppose it's possible that this is  a Windows-specific (or MSVC-specific) issue, of course.

(One thing that would still be worth trying that I haven't done would be to build a Debug version of libHalide + simd_op_check and see if the MSVC Debug Runtime can give us any more info; unfortunately, that apparently requires building a Debug version of LLVM too, because trying to mix the different runtimes refuses at link time... and building a Debug version of LLVM is, uh, *big* and *slow* and I'm literally unsure if any of our WinBots have enough RAM to do so, period.)

rootjalex:
Thanks for taking the time to try to debug this!

I suspect that this is an issue with `SpecificExpr`, which just grabs a reference. I think I know how to solve this, but I just got access to a (very old) Windows machine that has literally *nothing* useful set-up on it, so it will take me a while to confirm

rootjalex:
Actually, I am pretty confident I found the bug, I am just going to use the buildbots to confirm instead of waiting for LLVM to install on this old laptop

rootjalex:
The only failure appears unrelated, so I am marking this as ready for review - note that this PR relies on #6900 , so if we end up reverting that, this PR will need to change (it uses `saturating_cast` intrinsic pattern matchings).

steven-johnson:
> I suspect that this is an issue with SpecificExpr, which just grabs a reference

If it is taking a reference to an Expr-that-can-go-stale, that would indeed fit the pattern -- unfortunately the template code in IRMatch.h is too complex for me to immediately see how this could be happening. Did you trace it all the way thru or just suspect it?

I like your proposed fix, but if this is in fact what is happening, we need to upgrade SpecificExpr to use `Expr` instead of `const Expr &` -- if the reference can go stale from the callsites you fixed, I will bet you a dollar that there are other such sites as well, either now or in the future. (Yes, this will cost a bit more reference counting, but if the alternative is crashing, performance is irrelevant).

steven-johnson:
> note that this PR relies on https://github.com/halide/Halide/pull/6900

In that case we shouldn't land this yet -- if we can't track down the failure in 6900 in the next day or two, I think we need to (temporarily) revert it pending a fix, since we know it is breaking downstream code, and it's not good policy to keep known-broken code in the main branch.

rootjalex:
> Did you trace it all the way thru or just suspect it?

I ran into this issue last year at some point, and it caused random crashes - once I saw the `SpecificExpr` in your debug script I immediately suspected it as the source of the issue.

> we need to upgrade SpecificExpr to use Expr instead of const Expr & -- if the reference can go stale from the callsites you fixed, I will bet you a dollar that there are other such sites as well, either now or in the future. (Yes, this will cost a bit more reference counting, but if the alternative is crashing, performance is irrelevant).

I agree, these bugs are very annoying to hunt down. @abadams any thoughts? I'm happy to change `SpecificExpr` in this PR or another one.

> In that case we shouldn't land this yet -- if we can't track down the failure in 6900 in the next day or two, I think we need to (temporarily) revert it pending a fix, since we know it is breaking downstream code, and it's not good policy to keep known-broken code in the main branch.

Absolutely. Any chance the changes I pushed to #6961 last night fixed the issues in Google?

steven-johnson:
> Any chance the changes I pushed to https://github.com/halide/Halide/pull/6961 last night fixed the issues in Google? 

I'll try that now.

rootjalex:
> I'll try that now.

Thank you!

abadams:
IIRC, having SpecificExpr hold an Expr instead of a BaseExprNode reference is catastrophic for performance and code size, because now constructing a pattern requires atomic operations, and we have stack frames where thousands of patterns are constructed but not used.

So just changing it to an Expr, while appealing, is probably not a viable fix. The IRMatch code doesn't use Exprs for a good reason.

abadams:
I suggest changing it to Expr to see if that's the issue, and if it is, seeing if there's some way we can force everything that might ingest an Expr in IRMatch.h to only accept a "const Expr &" and not an "Expr" or an "Expr &&". Right now it's all universal references.

abadams:
Oh, we already guard against it:

```

template<typename T>
HALIDE_ALWAYS_INLINE void assert_is_lvalue_if_expr() {
    static_assert(!std::is_same<typename std::decay<T>::type, Expr>::value || std::is_lvalue_reference<T>::value,
                  "Exprs are captured by reference by IRMatcher objects and so must be lvalues");
}

```

src/IRMatch.h
```diff
@@ -1529,6 +1535,14 @@ HALIDE_ALWAYS_INLINE auto intrin(Call::IntrinsicOp intrinsic_op, Args... args) n
     return {intrinsic_op, pattern_arg(args)...};
 }
 
+template<typename A>
+auto abs(A &&a) noexcept -> Intrin<decltype(pattern_arg(a))> {
+    return {Call::abs, pattern_arg(a)};
```
abadams, line 1540: Here and in other helpers try adding `assert_is_lvalue_if_expr<A>();`

rootjalex: Done in [9a5327c](https://github.com/halide/Halide/pull/6884/commits/9a5327c6243d8c53b1214150a2583eaceba44b98)

rootjalex:
`v_intrin` and the Call intrinsic objects should now all statically fail if we use an Expr as an arg, and I confirmed that it does indeed catch the cases that were causing the Windows failures.

steven-johnson:
Nice work finding and fixing the dangling-reference stuff -- it didn't occur to me that you could insert `is_lvalue...` checks as a means of verification.

> having SpecificExpr hold an Expr instead of a BaseExprNode reference is catastrophic for performance and stack space usage

We should capture this in a code comment, so someone (e.g. me) won't make the same suggestion in the future...

steven-johnson:
(LGTM pending green)

rootjalex:
> Nice work finding and fixing the dangling-reference stuff

Thanks!

> it didn't occur to me that you could insert is_lvalue... checks as a means of verification.

Didn't occur to me either, I'm glad Andrew pointed me to the right method :)

> We should capture this in a code comment, so someone (e.g. me) won't make the same suggestion in the future...

Done!

steven-johnson:
Pulling this into Google for testing, I get many failures of the form:

```
Internal Error at [third_party/halide/halide/src/IR.cpp:42] Condition failed: a.type() == b.type(): Add of mismatched types
*** SIGABRT received by PID 3966 (TID 3966) on cpu 210 from PID 3966; stack trace: ***
PC: @     0x7ff6fddcf347  (unknown)  gsignal
    @     0x560160be7b34        976  FailureSignalHandler()
    @     0x7ff6fdf2a1c0  (unknown)  (unknown)
    @     0x7ff6fddcf347        136  gsignal
    @     0x7ff6fddd0797        304  abort
    @     0x56015d5ac033        112  Halide::Internal::ErrorReport::~ErrorReport()
    @     0x56015d72eb05        352  Halide::Internal::Add::make()
    @     0x56015d7446df         96  Halide::Internal::IRMutator::visit()
    @     0x56015dd1bc33        432  Halide::Internal::(anonymous namespace)::Optimize_X86::visit()
```

I'll try to narrow it down...

rootjalex:
@steven-johnson I have a suspicion that it could be the result of an issue with the top-down type inference in IRMatch.h. Any chance you could try out the small change I made to this branch on rootjalex/x86-optimize-test (d5f507e) ? Sorry about this

steven-johnson:
> @steven-johnson I have a suspicion that it could be the result of an issue with the top-down type inference in IRMatch.h. Any chance you could try out the small change I made to this branch on rootjalex/x86-optimize-test ([d5f507e](https://github.com/halide/Halide/commit/d5f507ebe7b1eb6f0ce2bafb78f247ba39d6ef56)) ? Sorry about this

First of many that popped up:

```
Condition failed: expr.type() == e.type(): Found type mismatch: (uint16x8)absd((int16x8)t21816, (int16x8)rounding_halving_add((int16x8)t21817, (int16x8)t21818)) -> (max((int16x8)t21816, int16x8(((uint16x8)vector_instruction("rounding_halving_add", uint16x8((int16x8)t21817), uint16x8((int16x8)t21818)) + (uint16x8)bitwise_and((uint16x8)bitwise_xor(uint16x8((int16x8)t21817), uint16x8((int16x8)t21818)), x8((uint16)32768))))) - min((int16x8)t21816, int16x8(((uint16x8)vector_instruction("rounding_halving_add", uint16x8((int16x8)t21817), uint16x8((int16x8)t21818)) + (uint16x8)bitwise_and((uint16x8)bitwise_xor(uint16x8((int16x8)t21817), uint16x8((int16x8)t21818)), x8((uint16)32768))))))
```

steven-johnson:
So far, they *all* are an `absd()` expr

rootjalex:
Ah, that makes perfect sense - [this code](https://github.com/halide/Halide/blob/69b50af793d6eb850eea3dccb16426479edbff9d/src/CodeGen_X86.cpp#L558) isn't type-checked even though it *technically* changes the sign, and just directly re-using that sequence broke `absd` lowering on signed integers. Will push a fix

steven-johnson:
(Leaving in that type-check assert sounds good too)

rootjalex:
> (Leaving in that type-check assert sounds good too)

Oh sure, that's probably a good idea. I'll add it to this branch then.

rootjalex:
And thanks for helping debug this @steven-johnson !

steven-johnson:
WIth the latest fix, I now get:

```
Deinterleaver should never receive VectorInstruction node, received:
(uint16x32)vector_instruction("rounding_halving_add", (uint16x32)t20677, x32((uint16)63))
```

rootjalex:
Ah shoot, I suspect that's coming from the `extract_lane()` calls I see in CodeGen_LLVM, in `scalarize` and `codegen_atomic_rmw`. I didn't realize `deinterleave` could be called that late in codegen. I will have to think about a fix, I don't have a good idea how to do so right now, because this essentially means we need to lower vector instructions and I really did not want to have to do that.

rootjalex:
Actually, thanks to `give_up_and_shuffle`, the easy fix I just pushed should work (fingers crossed)

steven-johnson:
With give-up-and-shuffle in place, I now get failures like:

```
*** SIGSEGV (@0x7f2332b36ff8), see [go/stacktraces#s15](http://goto.google.com/stacktraces#s15) received by PID 1107 (TID 1107) on cpu 72; stack trace: ***
PC: @     0x5608a5a9b644  (unknown)  Halide::Internal::(anonymous namespace)::RemoveLets::visit()
    @     0x5608a906eb74        976  FailureSignalHandler()
    @     0x7f2334d5a1c0  (unknown)  (unknown)
    @     0x5608a5a9b644         16  Halide::Internal::(anonymous namespace)::RemoveLets::visit()
    @     0x5608a5bb5478         32  Halide::Internal::ExprNode<>::mutate_expr()
    @     0x5608a5bc6fd3         80  Halide::Internal::IRGraphMutator::mutate()
    @     0x5608a5bc4bbb        112  Halide::Internal::IRMutator::mutate_with_changes()
    @     0x5608a5bc6800         80  Halide::Internal::IRMutator::visit()
    @     0x5608a5bb5a3b         32  Halide::Internal::ExprNode<>::mutate_expr()
    @     0x5608a5bc6fd3         80  Halide::Internal::IRGraphMutator::mutate()
    @     0x5608a5bc4bbb        112  Halide::Internal::IRMutator::mutate_with_changes()
    ```

**Comment by rootjalex** [comment]
Hmm the handler for an extract_element in Simplify_Shuffle does look awfully suspicious, I think it needs a check for whether or not there are multiple vectors in the shuffle.

**Comment by rootjalex** [comment]
The issue is that `simplify` calls `extract_lane` which calls `deinterleave` which calls `simplify` and onwards - for any case that hits `give_up_and_shuffle`, both the case I've added and the existing cases, this can cause an infinite loop

**Comment by steven-johnson** [comment]
OK, with that in place, I now get failures in `test/correctness_compute_with`, of the form

```
Internal Error at third_party/halide/halide/src/CodeGen_LLVM.cpp:1318 triggered by user code at :
Symbol not found: t19054
```

**Comment by rootjalex** [comment]
I see that on the buildbots - I will try to debug it later today. Luckily I also see this test failure on my machine so it should be far easier to debug.

**Comment by steven-johnson** [comment]
OK, with that last commit, google presubmits look good now.

That said, given the amount of whack-a-mole here, we probably want to review the last spate of changes pretty carefully to be confident there aren't more spiders lurking underneath.

**Comment by rootjalex** [comment]
> OK, with that last commit, google presubmits look good now.

Awesome, thanks!

> That said, given the amount of whack-a-mole here, we probably want to review the last spate of changes pretty carefully to be confident there aren't more spiders lurking underneath.

Definitely. I am a bit concerned about the issues we were seeing with 15e1a8c in place, which is why I reverted that change and went for an alternative route. I don't know the invariant hoisting code well enough to debug how that change broke it.

**Comment by steven-johnson** [comment]
So where do we stand on this... did the revert of #7009 cause a problem for this?

**Comment by rootjalex** [comment]
No, #7009 is completely unrelated (other than it also being instruction-selection oriented). This PR is only focused on x86, and is just waiting for review(s)

File: src/CodeGen_LLVM.h [hunk 4]
```diff
@@ -514,6 +517,11 @@ class CodeGen_LLVM : public IRVisitor {
      * across backends. */
     virtual void codegen_vector_reduce(const VectorReduce *op, const Expr &init);
 
+    /** Split up a VectorReduce node if possible, or generate LLVM
```
abadams, line 520: Please elaborate on this comment, also nit: lines in the block comment should start with *

src/CodeGen_X86.cpp
```diff
@@ -270,83 +279,40 @@ void CodeGen_X86::init_module() {
...
+// FIXME: This is nearly identical to CodeGen_LLVM, should re-factor this somehow.
+// Only difference is the call to `optimize_x86_instructions()`
+void CodeGen_X86::compile_func(const LoweredFunc &f, const std::string &simple_name,
+                               const std::string &extern_name) {
+    // Generate the function declaration and argument unpacking code.
+    begin_func(f.linkage, simple_name, extern_name, f.args);
+
+    // If building with MSAN, ensure that calls to halide_msan_annotate_buffer_is_initialized()
+    // happen for every output buffer if the function succeeds.
+    if (f.linkage != LinkageType::Internal &&
```
abadams, line 291: Possibly this block of code should be in begin_func?

src/CodeGen_X86.cpp
```diff
@@ -455,38 +421,12 @@ void CodeGen_X86::visit(const Select *op) {
...
-        if (expr_match(p.pattern, op, matches)) {
-            value = call_overloaded_intrin(op->type, p.intrin, matches);
-            if (value) {
-                return;
-            }
-        }
-    }
-
     if (const Call *mul = Call::as_intrinsic(op->value, {Call::widening_mul})) {
         if (op->value.type().bits() < op->type.bits() && op->type.bits() <= 32) {
```
abadams, line 431: Add op->type.is_vector() to this condition

src/CodeGen_X86.cpp
```diff
@@ -455,38 +421,12 @@ void CodeGen_X86::visit(const Select *op) {
 }
 
 void CodeGen_X86::visit(const Cast *op) {
-
     if (!op->type.is_vector()) {
```
abadams, line 424: delete this block

src/Deinterleave.cpp
```diff
@@ -402,6 +407,12 @@ Expr deinterleave(Expr e, int starting_lane, int lane_stride, int new_lanes, con
     Deinterleaver d(starting_lane, lane_stride, new_lanes, lets);
     e = d.mutate(e);
     e = common_subexpression_elimination(e);
+    if (const Shuffle *shuffle = e.as<Shuffle>()) {
+        if (shuffle->is_extract_element() && shuffle->vectors.size() == 1) {
```
abadams, line 411: Remove this if, and just don't simplify if the root node is a shuffle. Add a comment about simplify calling this function on shuffles to explain why.

src/IR.cpp
```diff
@@ -907,6 +907,46 @@ Stmt Atomic::make(const std::string &producer_name,
     return node;
 }
 
+namespace {
+
+const char *const instruction_op_names[] = {
+    // Shared:
```
abadams, line 913: Shared -> Use by multiple backends

src/IR.cpp
```diff
@@ -907,6 +907,46 @@ Stmt Atomic::make(const std::string &producer_name,
     return node;
 }
 
+namespace {
+
+const char *const instruction_op_names[] = {
```
abadams, line 912: Add comment like "names of instructions used in VectorInstruction nodes"

src/IR.cpp
abadams: Justify the overlap between this and intrinsics expressed using call nodes in a comment.

The important difference is that if you see one of these nodes, there's guaranteed to be an instruction on the target arch that implements it. No further lowering or emulation is required.

src/IR.cpp
```diff
@@ -907,6 +907,46 @@ Stmt Atomic::make(const std::string &producer_name,
...
+    // Shared:
+    "abs",
+    "dot_product",
+    "rounding_halving_add",
+    "saturating_add",
```
abadams, line 917: Summary from offline discussion: We disagree whether or not this overlap with Call intrinsics should exist, but are tabling it for now.

src/IR.cpp
abadams: Comments about comments should actually apply to header

src/IR.h
```diff
@@ -910,6 +910,47 @@ struct Atomic : public StmtNode<Atomic> {
...
+    };
+
+    InstructionOp op;
+    std::vector<Expr> args;
+
+    static Expr make(Type type, InstructionOp op, const std::vector<Expr> &args);
+
+    static const IRNodeType _node_type = IRNodeType::VectorInstruction;
+
+    const char *get_instruction_name() const;
```
abadams, line 951: This should be static and accept the enum to mirror Call::get_intrinsic_name

src/IRMatch.cpp
```diff
@@ -296,6 +296,22 @@ class IRMatch : public IRVisitor {
...
     }
 
+    void visit(const VectorInstruction *op) override {
+        const VectorInstruction *e = expr.as<VectorInstruction>();
+        if (result && e &&
+            types_match(op->type, e->type) &&
+            e->op == op->op &&
+            e->args.size() == op->args.size()) {
+            for (size_t i = 0; result && (i < e->args.size()); i++) {
+                // FIXME: should we early-out? Here and in Call*
```
abadams, line 306: It does early out (result is in the for loop condition)

src/IRMatch.h
```diff
@@ -1803,6 +1894,7 @@ inline std::ostream &operator<<(std::ostream &s, const BroadcastOp<A, B> &op) {
 template<typename A, typename B>
 HALIDE_ALWAYS_INLINE auto broadcast(A &&a, B lanes) noexcept -> BroadcastOp<decltype(pattern_arg(a)), decltype(pattern_arg(lanes))> {
     assert_is_lvalue_if_expr<A>();
+    assert_is_lvalue_if_expr<B>();
```
abadams, line 1897: Can lanes really be a concrete Expr? Maybe assert not concrete expr instead.

src/IRMatch.h
```diff
@@ -1872,6 +1964,110 @@ HALIDE_ALWAYS_INLINE auto ramp(A &&a, B &&b, C &&c) noexcept -> RampOp<decltype(
...
+
+    HALIDE_ALWAYS_INLINE
+    void print_args(std::ostream &s) const {
+        print_args<0>(0, s);
+    }
+
+    HALIDE_ALWAYS_INLINE
+    Expr make(MatcherState &state, halide_type_t type_hint) const {
+        std::vector<Expr> r_args(sizeof...(Args));
+        // TODO(rootjalex): How do we do type hints for the args?
```
abadams, line 2024: I believe this is done

src/IRMatch.h
```diff
@@ -1872,6 +1964,110 @@ HALIDE_ALWAYS_INLINE auto ramp(A &&a, B &&b, C &&c) noexcept -> RampOp<decltype(
...
+
+        r_args[0] = std::get<0>(args).make(state, {});
+        if constexpr (sizeof...(Args) > 1) {
+            r_args[1] = std::get<const_min(1, sizeof...(Args) - 1)>(args).make(state, {});
+        }
+        if constexpr (sizeof...(Args) > 2) {
+            r_args[2] = std::get<const_min(2, sizeof...(Args) - 1)>(args).make(state, {});
+        }
+
+        // for (int i = 0; i < sizeof...(Args); i++) {
```
abadams, line 2038: Remove commented-out code

src/IRMatch.h
```diff
@@ -1872,6 +1964,110 @@ HALIDE_ALWAYS_INLINE auto ramp(A &&a, B &&b, C &&c) noexcept -> RampOp<decltype(
...
+        : op(_op), args(args...) {
+        static_assert(sizeof...(Args) > 0 && sizeof...(Args) <= 3,
+                      "VectorInstructionOp must have non-zero arguments, and update make() if more than 3 arguments.");
+    }
+};
+
+template<typename... Args>
+std::ostream &operator<<(std::ostream &s, const VectorInstructionOp<Args...> &op) {
+    // TODO(rootjalex): Should we print the type?
+    s << "vector_instr(\"";
```
abadams, line 2058: vector_instr -> v_instr

src/IRMatch.h
```diff
@@ -2080,6 +2282,39 @@ HALIDE_ALWAYS_INLINE auto cast(halide_type_t t, A &&a) noexcept -> CastOp<declty
     return {t, pattern_arg(a)};
 }
 
+// A node for expressing type hints, when rules are ambiguously typed.
+template<typename A>
+struct TypeHint {
```
abadams, line 2287: If possible, make the type a template parameter instead of a member (either a C++ type or a halide_type_t), because each pattern consumes stack space in methods that use rewriters.

src/IRMatch.h
```diff
@@ -2080,6 +2282,39 @@ HALIDE_ALWAYS_INLINE auto cast(halide_type_t t, A &&a) noexcept -> CastOp<declty
...
+
+    constexpr static uint32_t binds = bindings<A>::mask;
+
+    constexpr static IRNodeType min_node_type = IRNodeType::Cast;
+    constexpr static IRNodeType max_node_type = IRNodeType::Cast;
+    constexpr static bool canonical = A::canonical;
+
+    HALIDE_ALWAYS_INLINE
+    Expr make(MatcherState &state, halide_type_t type_hint) const {
+        return a.make(state, type);
```
abadams, line 2300: Update this to use the template type with the lanes of the type_hint field

src/IRMatch.h
```diff
@@ -2306,6 +2541,8 @@ template<typename A>
 struct IsFloat {
     struct pattern_tag {};
     A a;
+    int bits;
```
abadams, line 2544: bits can be a template parameter instead (and also in is_int and is_uint)

src/IRMatch.h
```diff
@@ -2320,30 +2557,84 @@ struct IsFloat {
...
+    }
+    s << ")";
+    return s;
+}
+
+template<typename A>
+struct IsBFloat {
+    struct pattern_tag {};
+    A a;
+    int bits;
```
abadams, line 2590: bits can be template parameter

src/InstructionSelector.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+Expr InstructionSelector::visit(const Div *op) {
+    if (op->type.is_vector() && op->type.is_int_or_uint()) {
+        // Lower division here in order to do pattern-matching on intrinsics.
+        return mutate(lower_int_uint_div(op->a, op->b));
+    }
+    return IRGraphMutator::visit(op);
+}
+
+Expr InstructionSelector::visit(const Mod *op) {
```
abadams, line 21: consider lowering lerp here too

src/InstructionSelector.cpp
```diff
@@ -0,0 +1,34 @@
...
+    if (op->type.is_vector() && op->type.is_int_or_uint()) {
+        // Lower mod here in order to do pattern-matching on intrinsics.
+        return mutate(lower_int_uint_mod(op->a, op->b));
+    }
+    return IRGraphMutator::visit(op);
+}
+
+Expr InstructionSelector::visit(const VectorReduce *op) {
+    return mutate(codegen->split_vector_reduce(op, Expr()));
```
abadams, line 30: We should pull split_vector_reduce out of CodeGen_LLVM into the InstructionSelector and have it emit VectorInstruction nodes with enum values that correspond to the things that LLVM natively supports.

This might mean the instruction selector needs to know the native vector width, and maybe how to upgrade types for arithmetic. This could be pulled out of LLVM into common support code, e.g. in CodeGen_Internal.cpp

abadams: (because currently we believe that this will cause InstructionSelector to emit call nodes to llvm intrinsics)

src/ModulusRemainder.cpp
```diff
@@ -213,6 +214,12 @@ void ComputeModulusRemainder::visit(const Shuffle *op) {
     result = ModulusRemainder{};
 }
 
+void ComputeModulusRemainder::visit(const VectorInstruction *op) {
+    internal_error << "modulus_remainder of VectorInstruction:\n"
```
abadams, line 218: This seems likely to trigger because ModulusRemainder is used inside codegen for alignment queries.

abadams: Copy what the VectorReduce visitor does

src/Monotonic.cpp
```diff
@@ -535,6 +535,11 @@ class DerivativeBounds : public IRVisitor {
         result = ConstantInterval::single_point(0);
     }
 
+    void visit(const VectorInstruction *op) override {
+        // TODO(rootjalex): Should this be an error?
```
abadams, line 539: No, just remove this todo

src/Simplify_Exprs.cpp
```diff
@@ -59,6 +59,11 @@ Expr Simplify::visit(const Broadcast *op, ExprInfo *bounds) {
     }
 }
 
+Expr Simplify::visit(const VectorInstruction *op, ExprInfo *bounds) {
+    clear_bounds_info(bounds);
```
abadams, line 63: This should recursively simplify the args

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+#include "IRMatch.h"
+#include "IRMutator.h"
+#include "IROperator.h"
+#include "InstructionSelector.h"
+#include "Simplify.h"
+
+namespace Halide {
+namespace Internal {
+
+#if defined(WITH_X86)
```
abadams, line 15: I don't think this is necessary. It's only necessary if the file is going to refer to llvm x86-specific stuff

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+            std::vector<Expr> args = {a0, a1, b0, b1};
+            result.swap(args);
+            return true;
+        }
+    }
+    return false;
+}
+
+/** A top-down code optimizer that replaces Halide IR with VectorInstructions specific to x86. */
+class Optimize_X86 : public InstructionSelector {
```
abadams, line 63: InstructionSelector_X86 for the class and file?

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+public:
+    /** Create an x86 code optimizer. Processor features can be
+     * enabled using the appropriate flags in the target struct. */
+    Optimize_X86(const Target &target, const CodeGen_LLVM *codegen)
+        : InstructionSelector(target, codegen) {
+    }
+
+    using IRGraphMutator::mutate;
+    Expr mutate(const Expr &e) override {
+        Expr expr = IRGraphMutator::mutate(e);
```
abadams, line 73: Everywhere you have IRGraphMutator:: here should probably be InstructionSelector::

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+        internal_assert(expr.type() == e.type()) << "(X86Optimize) Found type mismatch: " << e << " -> " << expr << "\n";
+        return expr;
+    }
+
+protected:
+    bool should_peephole_optimize(const Type &type) {
```
abadams, line 79: Not convinced this is necessary or helpful. It may well be that we want to apply these rewrites to scalars too.

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+        if (!should_peephole_optimize(op->type)) {
+            return IRGraphMutator::visit(op);
+        }
+
+        std::vector<Expr> matches;
+        // TODO(rootjalex): is it possible to rewrite should_use_dot_product
+        // as a series of rewrite-rules? lossless_cast is the hardest part.
+        const int lanes = op->type.lanes();
+
+        // FIXME: should we check for accumulating dot_products first?
```
abadams, line 99: remove this fixme

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+             rewrite(
+                 cast(BFloat(16, lanes), x),
+                 v_instr(VectorInstruction::f32_to_bf16, x),
+                 is_float(x, 32))) ||
+
+            false) {
+            return mutate(rewrite.result);
+        }
+
+        // TODO: should we handle CodeGen_X86's weird 8 -> 16 bit issue here?
```
abadams, line 225: Should resolve this TODO one way or the other

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+
+              false)) ||
+
+            false) {
+            return mutate(rewrite.result);
+        }
+
+        // Fixed-point intrinsics should be lowered here.
+        // This is safe because this mutator is top-down.
+        // FIXME: Should this be default behavior of the base InstructionSelector class?
```
abadams, line 405: Address FIXME

src/X86Optimize.cpp
```diff
@@ -0,0 +1,633 @@
...
+                Call::widening_sub,
+            })) {
+            return mutate(lower_intrinsic(op));
+        }
+
+        return IRGraphMutator::visit(op);
+    }
+
+    Expr visit(const VectorReduce *op) override {
+        // FIXME: We need to split up VectorReduce nodes in the same way that
```
abadams, line 430: Remove FIXME

src/runtime/x86_avx2.ll
```diff
@@ -72,3 +62,34 @@ define weak_odr <16 x i16> @hadd_pmadd_i8_avx2(<32 x i8> %a) nounwind alwaysinli
...
+   ret <8 x i32> %3
+ }
+ declare <8 x i32> @llvm.x86.avx2.phadd.d(<8 x i32>, <8 x i32>) nounwind readnone
+
+ define weak_odr <8 x i32> @hadd_pmadd_i16_avx2(<16 x i16> %a) nounwind alwaysinline {
+   %1 = tail call <8 x i32> @llvm.x86.avx2.pmadd.wd(<16 x i16> %a, <16 x i16> <i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1, i16 1>)
+   ret <8 x i32> %1
+ }
+
+ define weak_odr <8 x i32> @wmul_pmaddwd_avx2(<8 x i16> %a, <8 x i16> %b) nounwind alwaysinline {
```
abadams, line 87: This should be removed from x86.ll if it's still there

rootjalex:
Just to update any watchers of this PR: I fully intend to address Andrew's review points, but there will be a delay in getting it done, due to moving / general business with the start of the semester.

abadams:
I have discovered some suboptimal codegen on x86 which I believe would be best fixed in this PR. The following generates a single pmaddwd:
```
    Var x;

    Func f1, f2;
    f1(x) = cast<uint8_t>(x);
    f2(x) = cast<uint8_t>(x);

    Func g;
    g(x) = cast<int32_t>(0);
    RDom r(0, 2);
    g(x) += cast<int32_t>(f1(2 * x + r)) * f2(2 * x + r);

    g.update().atomic().vectorize(r).vectorize(x, 8);
    f1.compute_root();
    f2.compute_root();
```

But if you change the int32_t to a uint32_t, it generates much worse code, even though the high bit is known to be zero because it's widening from a u8. If you try to work around it by adding a uint32_t cast outside the widening mul, then the outer cast is just removed.

rootjalex:
> But if you change the int32_t to a uint32_t, it generates much worse code, even though the high bit is known to be zero because it's widening from a u8. If you try to work around it by adding a uint32_t cast outside the widening mul, then the outer cast is just removed.

Yep, definitely seems like a useful rule to add to this PR - it would be a bit annoying to add to the current rule set in CodeGen_X86 due to the reinterprets that would be used. I will add this rule when I am addressing the rest of the comments!

PR #6885, opened 2022-07-26
Minor code cleanups in PythonExtensionGen

steven-johnson:
This is a bit gratuitous -- the old code wasn't *wrong* in any way I know of -- but this makes the emitted code a bit cleaner and terser, esp. wrt error handling (the old code could emit lots of redundant code). Also brings the codegen style a bit more in line with (eg) Codegen_C.


steven-johnson:
See also https://github.com/halide/Halide/pull/6888

PR #6886, opened 2022-07-26
Pip packaging new attempt

alexreinking:
A newly rebased version of #6815 to work on with a fresh history.

Still needs more thorough testing, both on CI and locally.

**Do not merge without including the following line in the commit message:**

Co-authored-by: Lukas Trümper <lukas.truemper@outlook.de>

See: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors


alexreinking:
Looks like the buildbots were relying on FetchContent to grab pybind11. We should just install the appropriate version on the machines or, better yet, create per-run virtual environments that have the appropriate packages installed.

steven-johnson:
We have a *lot* of stuff that relies on FetchContent. If it is (apparently) going to be a problem, we really need to surface that as a separate issue first and figure out a good plan for replacing it.

steven-johnson:
FWIW, the Linux bots already have pybind11 2.6.2 installed in the venv we use -- not sure why it isn't being found

steven-johnson:
I'm now installing the 'global' version in our venv on the buildbots:

`python3 -m pip install pybind11[global]==2.6.2`

We'll see if that moves the needle.

steven-johnson:
so, `python3 -m pip install pybind11[global]==2.6.2` works fine on the posixy setups, but apparently not on Windows. 

We could use `vcpkg` to install on Windows, but... apparently if you want to use vcpkg to install a specific version, you can't just do it from the command line; you need to create a JSON "manifest" file (aka "just use the IDE"). While this shouldn't be hard, it irks me greatly, just on principle. Gonna see if I can brutalize pip into working somehow.

alexreinking:
> FWIW, the Linux bots already have pybind11 2.6.2 installed in the venv we use -- not sure why it isn't being found

We might just need to set `pybind11_ROOT` to the pybind11 installation directory inside the venv

steven-johnson:
> We might just need to set pybind11_ROOT to the pybind11 installation directory inside the venv

See above, installing *in the venv* but with `[global]` seems to work on posixy. Why would we need to set ROOT?

alexreinking:
Maybe I'm not clear on what exactly [global] is doing there?

steven-johnson:
> Maybe I'm not clear on what exactly [global] is doing there?

See https://pybind11.readthedocs.io/en/stable/installing.html

> You can download the sources and CMake files as a Python package from PyPI using Pip. Just use:
> 
> `pip install pybind11`
>
> This will provide pybind11 in a standard Python package format. If you want pybind11 available directly in your environment root, you can use:
> 
> `pip install "pybind11[global]"`
> This is not recommended if you are installing with your system Python, as it will add files to /usr/local/include/pybind11 and /usr/local/share/cmake/pybind11, so unless that is what you want, it is recommended only for use in virtual environments or your pyproject.toml file (see [Build systems](https://pybind11.readthedocs.io/en/stable/compiling.html#compiling)).

python_bindings/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = [
+  "setuptools>=42",
+  "wheel",
+  "scikit-build",
+  "pybind11==2.6.2",
```
steven-johnson, line 6: orthogonal to this PR, but we should probably document why we use this version of pybind11; IIRC it's because it's a version that we believe to be readily available on multiple distros?

steven-johnson:
LGTM (so far) pending green, but one thing that seems lacking is documentation: how do I build a pip package, how do I use it locally (ie do I need to 'pip install' it or is there another mechanism), etc... I realize this is all well-documented elsewhere but a paragraph or three of pointers would be useful for quick reference.

steven-johnson:
With https://github.com/halide/build_bot/pull/208 landed, you may want to change the default of fetchcontent back to OFF?

python_bindings/setup.py
```diff
@@ -0,0 +1,53 @@
+from skbuild import setup, cmaker, utils
```
steven-johnson, line 1: Don't we need to have `skbuild` in pyproject.toml or requirements.txt or, uh, something?

python_bindings/pyproject.toml
steven-johnson: ah, nm, it's `scikit-build` in pip

alexreinking: >IIRC

I no longer recall, so to-do in another PR. Maybe create an issue about upgrading it? Maybe there were breaking API changes?

alexreinking:
This needs more work on Windows to accurately track down dependencies. We'll need some magic with `os.add_dll_directory()` and we might even want to wait to support Windows pip packaging until we upgrade to CMake 3.22, which comes with important features for tracking down dynamic dependencies.

steven-johnson (approved):

PR #6887, opened 2022-07-26
Deinterleave: generalize to be lane mask-based

LebedevRI:
Currently, it was implemented to only handle "deinterleaving"
(really, removal of not-demanded lanes) of lanes that can be
described by a linear polynominal. While this clearly worked,
there already were some FIXME comments, and it was already
clearly not generic-enough, e.g. it could not recurse into shuffles.

For proper handling of Reinterpret, which either merges adjacent lanes.
or splits lanes, this means that no support is possible at all.

### DIFF
--- a/src/Deinterleave.cpp
+++ b/src/Deinterleave.cpp
@@ -11,6 +11,8 @@
 #include "Scope.h"
 #include "Simplify.h"
 #include "Substitute.h"
+#include <numeric>
+#include <optional>
 
 namespace Halide {
 namespace Internal {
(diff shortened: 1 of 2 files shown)


src/Deinterleave.cpp
```diff
@@ -198,202 +384,345 @@ class Deinterleaver : public IRGraphMutator {
...
+            in = Broadcast::make(in, lanes);
         }
-        if (op->value.type().lanes() > 1) {
-            // There is probably a more efficient way to this.
-            return mutate(flatten_nested_ramps(op));
+
+        // But, it is possible that we *still* don't demand all of the produced
+        // lanes, in which case we need a Shuffle after all.
+        if (in.type().lanes() != new_lanes.size()) {
+            in = Shuffle::make({in}, new_lanes);
```
LebedevRI, line 462: (We only get here if the source had more than one lane)
Conceptual question: do we really want to keep the `broadcast`
even if we will then immediately drop some of the produced lanes?

src/Deinterleave.cpp
```diff
@@ -176,19 +178,203 @@ Stmt collect_strided_stores(const Stmt &stmt, const std::string &name, int strid
...
+        if (mask[src_lane]) {
+            indices.emplace_back(src_lane);
+        }
+    }
+    return indices;
+}
+
+template<typename T, typename UnaryOperation, typename UnaryPredicate>
+std::vector<char> indices_to_mask(const std::vector<T> &data, int indice_end,
+                                  UnaryOperation get_indice,
```
abadams, line 204: indice should be index throughout this PR

src/Deinterleave.cpp
```diff
@@ -176,19 +178,203 @@ Stmt collect_strided_stores(const Stmt &stmt, const std::string &name, int strid
...
+    template<typename T, typename UnaryOperation, typename UnaryPredicate>
+    static IndiceXFormMapping get(const std::vector<T> &data, int indice_end,
+                                  UnaryOperation get_indice,
+                                  UnaryPredicate pred) {
+        // Which indices are demanded at all?
+        std::vector<char> indice_mask =
+            indices_to_mask(data, indice_end, get_indice, pred);
+
+        // What are the (in-order) demanded indices?
+        decltype(IndiceXFormMapping::indices) indices =
```
abadams, line 232: Why decltypes here instead of auto?

src/Deinterleave.cpp
```diff
@@ -198,202 +384,345 @@ class Deinterleaver : public IRGraphMutator {
...
-        Expr in = Shuffle::make({op->value}, input_lanes);
-        return VectorReduce::make(op->op, in, new_lanes);
+        ScopedValue<DemandedLanes> old_lanes(lanes, input_lanes);
+        return VectorReduce::make(op->op, mutate(op->value), lanes.size());
     }
 
+    // This is essentially the same as `Ramp`/`Shuffle` handling,
+    // just specialized.
+    //
+    // FIXME: do we really prefer/want to retain the Broadcast
```
abadams, line 400: The simplifier turns a shuffle of a broadcast into a smaller shuffle. Can't we just do the same here?

LebedevRI: > The simplifier turns a shuffle of a broadcast into a smaller broadcast. Can't we just do the same here?

Well, that was the question i was asking :)
I assume it's the same story for `Ramp`.
Thanks!

LebedevRI:
Note that this is not quite ready for full review.

src/Deinterleave.cpp
abadams: I think the complexity here is due to deinterleave running before flattening nested vectors. So the question is if you ask for the even lanes of:

```
ramp(ramp(0, 1, 10), broadcast(100), 10)
```

Should you get:

```
ramp(ramp(0, 1, 10), broadcast(200), 5)
```

Or:

```
ramp(ramp(0, 2, 5), broadcast(100), 10)
```

And I guess the correct answer is the latter, which is why things are complicated.

So in the case you're talking about, I guess it's a shuffle of a broadcast of vectors, e.g. taking the even lanes of:

```
broadcast(ramp(0, 1, 3), 3)
```

i.e. [0, 1, 2, 0, 1, 2, 0, 1, 2], which is I guess [0, 2, 1, 0, 2]

yuck.

Maybe we can just do the correct thing here, even if it produces ugly IR, and rely on the simplifier to clean up any cases that can be cleaned up

src/Deinterleave.cpp
```diff
@@ -198,202 +384,345 @@ class Deinterleaver : public IRGraphMutator {
...
-                return op->vectors[starting_lane];
-            } else if ((int)op->vectors.size() % lane_stride == 0) {
-                // Pick up every lane-stride vector.
-                std::vector<Expr> new_vectors(op->vectors.size() / lane_stride);
-                for (size_t i = 0; i < new_vectors.size(); i++) {
-                    new_vectors[i] = op->vectors[i * lane_stride + starting_lane];
-                }
-                return Shuffle::make_interleave(new_vectors);
-            }
+        // FIXME: are `op->vectors` guaranteed to not have duplicates?
```
abadams, line 639: There's no such guarantee

abadams:
This whole pass was written before we had nested vectors, and I don't believe it was ever updated to handle them, so it may not have behaved as expected when run during lowering (though I don't believe it was explicitly wrong)

When used in codegen, vectors have been flattened.

LebedevRI:
On a second though..
@abadams thank you for taking a look, but this is a bit too much feedback, so i'm failing to parse it.
Let me focus on a single specific question:
given `broadcast({1, 2, 3}, 3)`, which produces `{1, 2, 3, 1, 2, 3, 1, 2, 3}`, if we only demand `i%2==0` lanes,
i.e. `{1, 3, 2, 1, 3}`, do we want to represent it as `shuffle({broadcast({1, 2, 3}, 3)}, {0, 2, 4, 6, 8)`,
or as `shuffle({1, 2, 3}, {0, 2, 1, 0, 2})`?
Note that this is indeed nested vector specific. For non-nested vectors we don't have this issue.

abadams:
Oh I see. I think we want the latter. I think there's never a reason to produce a shuffle of a broadcast. It should either be a single shuffle node, or a single broadcast node.

LebedevRI:
> Oh I see. I think we want the latter. I think there's never a reason to produce a shuffle of a broadcast. It should either be a single shuffle node, or a single broadcast node.

Thank you! Aha, that makes sense.

Now, the same question about the ramp.
Given `a`, `b`, `c`, `d`, which are scalars (lanes=1),
given `ramp({a, b}, {c, d}, 3)`, which produces `{a+0*c, b+0*d, a+1*c, b+1*d, a+2*c, b+2*d}`
if we only want lanes `0, 1, 3` (!!! not a 1'st order polynominal, can not fold into ramp),
i.e. `{a+0*c, b+0*d, b+1*d}`, do we want to represent it as `shuffle({ramp({a, b}, {c, d}, 3)}, {0, 1, 3)`,
or it's expanded form that does not involve computing the lanes that we will discard?

abadams:
We want to be able to do things like load(shuffle(ramp)) -> shuffle(load(ramp)), so I think we should keep the ramps

LebedevRI:
Okay, understood. Thank you.

src/Deinterleave.cpp
LebedevRI: ... the question being, should this code here be trying to deduplicate the `op->vectors`?

abadams: I don't think so, no.

LebedevRI:
Ok, i think i finally got the ramp handling.
This could use some tests i suppose, but this should be basically it.

LebedevRI:
@abadams all yours! :)
Will follow-up with `Reinterpret` handling after this is merged.

src/Simplify_Shuffle.cpp
```diff
@@ -96,11 +96,13 @@ Expr Simplify::visit(const Shuffle *op, ExprInfo *bounds) {
     // Try to collapse a shuffle of broadcasts into a single
     // broadcast. Note that it doesn't matter what the indices
     // are.
+    // FIXME: broadcast-of-vector is fine, but needs shuffle indice check.
     const Broadcast *b1 = new_vectors[0].as<Broadcast>();
-    if (b1) {
+    if (b1 && b1->value.type().lanes() == 1) {
```
LebedevRI, line 101: This is a miscompile fix.
Since github is so hilarisouly atrocious at handling (or rather, not handling at all)
stacked/dependent PRs, i'm not sure if submitting this separately won't result in more mess.
Without this, one newly added test fails.

src/Deinterleave.cpp
```diff
@@ -176,224 +178,637 @@ Stmt collect_strided_stores(const Stmt &stmt, const std::string &name, int strid
...
+    mask.reserve(scale * src_mask.size());
+    for (char src_mask_elt : src_mask) {
+        std::fill_n(std::back_inserter(mask), /*count=*/scale,
+                    /*value=*/src_mask_elt);
+    }
+    return mask;
+}
+
+// Compute the elements of the specified ramp/linear space.
+std::vector<int> ramp_to_indexes(int starting_lane, int lane_stride,
```
abadams, line 195: Change indexes to indices. The singular is "index" and the plural is "indices" because for some god damn reason English sometimes uses latin plurals and it's a 3rd declension noun and why oh why did I take four years of Latin.

LebedevRI: Oh, index, but indices, got it.

steven-johnson: <A "well, actually..." emerges>

[Both "indexes" and "indices" are acceptable English usages](https://www.nasdaq.com/articles/indexes-or-indices-whats-the-deal-2016-05-12)

steven-johnson: > Though both are still widely used, they take on different usage in their senses. "Indices" is used when referring to mathematical, scientific and statistical contexts.

src/Deinterleave.cpp
```diff
@@ -176,224 +178,637 @@ Stmt collect_strided_stores(const Stmt &stmt, const std::string &name, int strid
...
+    std::vector<int> indices;
+    indices.reserve(new_lanes);
+    for (int i = 0; i < new_lanes; i++) {
+        indices.emplace_back(starting_lane + lane_stride * i);
+    }
+    return indices;
+}
+
+// What are the (in-order) demanded indices?
+std::vector<int> mask_to_indices(const std::vector<char> &mask) {
```
abadams, line 206: I'd a little bit confused about why this PR uses lists of demanded lanes but also masks. Why are both representations necessary?

LebedevRI: Neither is optimal for the problem at hand.
For example, it is `O(1)` to compute the number of demanded lanes given the list (vector) of demanded lanes,
but if you have a mask of demanded lanes, it's `O(N)`.
Likewise, given a mask, it is trivial to upscale/downscale it (repeat/or-reduce bits),
but given indices it becomes somewhat less straight-forward.
Etc, etc.

We //could// settle for one, but code like this is exceptionally easy to be suddenly broken,
which is why i've intentionally written the diff as it is now, very verbose.

abadams: I think upscaling indices is:
```
vector<int> upscale(const vector<int> &v, int scale) {
  vector<int> result(v.size() * scale);
  for (size_t i = 0; i < v.size(); i++) {
    for (int s = 0; s < scale; s++) {
      result[i*scale + s] = v[i] * scale + s;
    }
  }
  return result;
}
```
Which doesn't seem that much more complex than upscaling a mask. Can we switch to just indices?

LebedevRI: Okay, i've rewritten `upscale_indices()`.

Though, that doesn't help much, because `IndexXFormMapping::get()` still does that roundtrip,
because the alternative is using a set/sorting a vector, and i can't possibly imagine that being better...

abadams: If that all it's for now, can it be made a private method in IndexXFormMapping called, e.g. sort_and_dedup_indices?

abadams: I was proposing reducing the scope in which it's defined, and renaming it to make clear what its sole usage is, so that people reading the file understand more clearly what it's for. But if you're saying there are still other uses, then I guess don't do that.

Another minor comment on naming: I think "affine" is more common than "1st_order_polynomial" in compilers, and it's a little terser.

LebedevRI: (I ended up not needing it in another place, all hail #6908!)

My main concern is that said function has nothing specific to do with it's (only) user.
By that logic, we should also drop other single-use functions.
But to me that only hurts - they are there because they have obvious semantics,
and are separateable from users. Abstractions all the way, it's the only way to try to not sink.

LebedevRI:
@abadams poke :) does my contrived reasoning sound convincing?

src/Deinterleave.cpp
```diff
@@ -176,224 +178,647 @@ Stmt collect_strided_stores(const Stmt &stmt, const std::string &name, int strid
...
-          new_lanes(new_lanes),
-          external_lets(lets) {
+    Deinterleaver(DemandedLanes lanes_, const Scope<> &lets)
+        : lanes(std::move(lanes_)), external_lets(lets) {
+    }
+
+    using IRGraphMutator::mutate;
+
+    Expr mutate(const Expr &e) override {
+        // FIXME: is there really only always-on assertions?
```
abadams, line 468: Yes, I'd much rather have compiler bugs cause internal assertion failures than silent problems for users. We get a more useful bug report that way.

What are you checking here? Is the concern that the check is expensive?

src/Deinterleave.cpp
```diff
@@ -176,224 +178,647 @@ Stmt collect_strided_stores(const Stmt &stmt, const std::string &name, int strid
...
-            ModulusRemainder align = op->alignment;
-            // TODO: Figure out the alignment of every nth lane
-            if (starting_lane != 0) {
-                align = ModulusRemainder();
-            }
-            return Load::make(t, op->name, mutate(op->index), op->image, op->param, mutate(op->predicate), align);
+        Type t = op->type.with_lanes(lanes.size());
+        ModulusRemainder align = op->alignment;
+        // TODO: Figure out the alignment of every nth lane
+        std::optional<std::pair<int /*base*/, int /*step*/>> lanes_affine =
```
abadams, line 526: For legibility, it might be worth an Affine struct with two ints in it instead of a std::pair.

src/Deinterleave.cpp
```diff
@@ -794,6 +1219,23 @@ Stmt rewrite_interleavings(const Stmt &s) {
 }
 
 namespace {
+
+template<typename UnaryFunction>
+void for_each_mask(UnaryFunction fun) {
```
abadams, line 1224: Please add a comment. It took me a while to figure out what this function does.

src/Deinterleave.cpp
```diff
@@ -827,8 +1284,257 @@ void deinterleave_vector_test() {
     Expr vec_x = Variable::make(Int(32, 4), "vec_x");
     Expr vec_y = Variable::make(Int(32, 4), "vec_y");
     check(Shuffle::make({vec_x, vec_y}, {0, 4, 2, 6, 4, 2, 3, 7, 1, 2, 3, 4}),
-          Shuffle::make({vec_x, vec_y}, {0, 2, 4, 3, 1, 3}),
-          Shuffle::make({vec_x, vec_y}, {4, 6, 2, 7, 2, 4}));
+          Shuffle::make({vec_x, Shuffle::make_extract_element(vec_y, 0)},
```
abadams, line 1287: I'm don't think these outputs are better than the old ones. The contain nested shuffles and look like they might codegen to unnecessary shuffle instructions.

abadams:
I worry this PR is overkill. The main purpose of the deinterleave pass is to simplify things like f[ramp(0, 1, 16) / 2] to interleave(f[ramp(0, 1, 8)], f[ramp(0, 1, 8)]). If the demanded indices aren't affine, I don't think we can do anything much better than a single shuffle op, and doing more might just make it harder to get the compiler to generate the asm you want. Are there actual use-cases for the non-affine cases here where this implementation is better than the baseline of inserting a Shuffle node whenever you hit a non-affine case (e.g. a VectorReduce or Reinterpret node).

src/Deinterleave.cpp
LebedevRI: My concern is that this check is expensive and statically dead, it's not going to ever fire.

src/Deinterleave.cpp
LebedevRI: 1. This is a subjective judgement. Are there any p
2. It really doesn't matter how we represent the shuffles, like or not, during codegen legalization, at least for X86, all shuffles are expanded and then new shuffles constructed from the basic element flow. https://godbolt.org/z/qW7x17aKs

abadams: On some platforms (Hexagon) we emit specific shuffle intrinsics that llvm preserves, and we're moving in the direction of handling more of instruction selection ourselves on all platforms (and finding ways to stop llvm from undoing our work).

To avoid subjective judgements, I'd really changes to be motivated by before/after assembly snippets that show this codegens better for some plausible pieces of input code. Otherwise I don't see what this complexity is buying us.

LebedevRI:
> I worry this PR is overkill. The main purpose of the deinterleave pass is to simplify things like f[ramp(0, 1, 16) / 2] to interleave(f[ramp(0, 1, 8)], f[ramp(0, 1, 8)]).

This is the common theme i'm encountering in this project. _**Zero**_ documentation...
I think i'm rapidly plummeting back into "why do i even care" territory.

> <...>

I think i'm going to go with the following reply:
are there any relevant benchmarks that could be used to objectively estimate the impact?

Affine or not is a red herring here, the actual problem, which seems to exist already,
is the fact that recursion can cause introduction of the shuffles to extract needed lanes,
once we stop being able to recurse further. Oh, and there's also the fact that
in doing so we either created irreparable mess of partially-full vectors,
or happen to split too-wide vectors into nice full vectors that result in optimal ILP.

src/Deinterleave.cpp
LebedevRI: > we're moving in the direction of handling more of instruction selection ourselves on all platforms (and finding ways to stop llvm from undoing our work).

FWIW, this seems like extremely misplaced work. Has any thought been/being put into actually benefiting everyone and working on fixing the final source of truth? :)

abadams: Yes, we've thought about whether or not we should just be improving llvm's instruction selection, but there are a few problems with that approach. First, not all of our backends are llvm. E.g. the xtensa branch needs to do pretty involved fixed-point instruction selection and is not llvm-based. Hexagon is technically llvm, but in practice it's all opaque intrinsics that llvm just passes through. 

Second, instruction selection is practically easier to do at the Halide level because it's a more restricted IR, for which we've developed a lot of relevant analysis tools (e.g. good bounds inference machinery). It's also practically easier because this is the code we understand and have control over.

The third more philosophical issue is that llvm's main performance goal is making naively written c++ code fast. When it comes to vector instruction selection, llvm is *intentionally* undoing our work, because it correctly believes that on average, it knows more about performance than some random C++ dev. It has been doing this by destructuring intrinsics in C++, deleting them entirely in LLVM IR, and replacing them with pattern matching. This can be brittle. Sometimes part of the pattern gets lifted as a loop invariant, or gets folded with a nearby shuffle, and instruction selection goes off the rails and you get bad code. Opening bugs for each case is a game of whack-a-mole.

Halide has very different users, so we're taking a different approach. Our approach has been to steadily add intrinsics inside the IR, first portable fixed-point ones (see FindIntrinsics.cpp), and now machine-specific vector instructions (See #6884). This means we can do instruction selection as IR->IR passes, and expose increasing amounts of control over instruction selection upwards to expert users. I don't think this approach would be welcome inside llvm.

abadams:
We normally proceed by having some piece of code that's generating poor assembly, making a change to the compiler, and demonstrating that it's producing better assembly now and not regressing anything in the apps. 

In the case here, I'm not sure if it's even possible to write front-end code that would trigger the new code paths, and that's why I'm wary of the PR. This is complex code that I will have to maintain five years from now, and I don't understand what the generality and complexity is buying us in terms of performance on some real use-case. If you don't have a use-case in mind, let's just do the simplest thing possible to make this pass correct for the new reinterpret semantics and move on to the next problem.

I'm sorry I'm being frustrating. We really appreciate your recent contributions.

LebedevRI:
<details>
<summary>ban button is over there</summary>
I've contemplated not continuing this quickly derailed discussion, but.

> We really appreciate your recent contributions.

The problem is that it appears i do not, since, well, clearly,
i'm still nudging this thread, which may have _subj_.

The problem is explicitly *NOT* that the diff in question is effectively dead.
All code is bad and broken, we just don't know about it yet.

_My_ problem here is with, essentially, that i'm seemingly wasting time
seemingly without getting any closer to the virtual endgoal.
Yet things still go downhill from there.

My problem is with communication/miscommunication/quality of communication.

It was said that deinterleave needed to handle reinterpret,
so i've gone ahead, and due to the lack of any, even high-level, documentation,
interpreted that it's state was not the explicit design (give up early, give up often),
but an implementation artifact. So i've **really** gone full in, and under that assumption,
rewrote it in full generality without the affine restriction.

And yet, personally, i can only describe the long-awaited feedback
as total and complete lack of positive affirmation, lack of appreciation for the design,
taking it for granted, and effectively immediately discarding it.

So no, frustration does not quite describe the seething, irradiating, happiness that this has initially brought to me.

Again, and once again, i'm only describing my view on this. I don't know if it matches reality.
</details>

PR #6888, opened 2022-07-26
Rewrite PythonExtensionGen to be C++ based

steven-johnson:
This is intended as an alternative to #6885 -- this is even *more* gratuitous, but:
- We have ~always compiled Python extensions using C++ anyway
- This code is arguably terser, cleaner, and safer (the cleanups happen via dtors)
- The code size difference is negligible (~300 bytes out of 160k for addconstant.cpython-39-darwin.so)

### DIFF
--- a/src/PythonExtensionGen.cpp
+++ b/src/PythonExtensionGen.cpp
@@ -99,36 +99,8 @@ std::pair<string, string> print_type(const LoweredArgument *arg) {
 
 }  // namespace
 
-void PythonExtensionGen::convert_buffer(const string &name, const LoweredArgument *arg) {
-    internal_assert(arg->is_buffer());
-    const int dims_to_use = arg->dimensions;
-    // Always allocate at least 1 halide_dimension_t, even for zero-dimensional buffers
-    const int dims_to_allocate = std::max(1, dims_to_use);
-    dest << "    halide_buffer_t buffer_" << name << ";\n";
-    dest << "    halide_dimension_t dimensions_" << name << "[" << dims_to_allocate << "];\n";
-    dest << "    Py_buffer view_" << name << ";\n";
-    dest << "    if (_convert_py_buffer_to_halide(";
-    dest << /*pyobj*/ "py_" << name << ", ";
-    dest << /*dimensions*/ (int)dims_to_use << ", ";
-    dest << /*flags*/ (arg->is_output() ? "PyBUF_WRITABLE" : "0") << ", ";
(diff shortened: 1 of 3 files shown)


src/PythonExtensionGen.cpp
```diff
@@ -151,112 +123,125 @@ void PythonExtensionGen::compile(const Module &module) {
...
-            // anyway, since we specified PyBUF_STRIDED.
-            PyErr_Format(PyExc_ValueError, "Invalid buffer: suboffsets not supported");
-            PyBuffer_Release(&buf);
-            return -1;
+template<int dimensions>
+struct PyHalideBuffer {
+    static constexpr int dims_to_allocate = (dimensions < 1) ? 1 : dimensions;
+
+    Py_buffer py_buf;
+    halide_dimension_t halide_dim[dims_to_allocate];  // Must allocate at least 1, even if d=0
```
shoaibkamil, line 138: Nit: this comment seems like it should be on line 135

src/PythonExtensionGen.cpp
```diff
@@ -295,103 +288,137 @@ HALIDE_PYTHON_EXPORT PyObject* PyInit_)INLINE_CODE";
...
+    dest << indent << "if (result != 0) {\n";
+    indent.indent += 2;
+    dest << indent << "PyErr_Format(PyExc_ValueError, \"Halide error %d\", result);\n";
+    dest << indent << "return nullptr;\n";
+    indent.indent -= 2;
+    dest << indent << "}\n";
+    dest << "\n";
+
+    dest << indent << "Py_INCREF(Py_None);\n";
+    dest << indent << "return Py_None;\n";
```
shoaibkamil, line 417: Is there a reason we return `None` now, as opposed to `True`?

shoaibkamil (approved):
I think this is an overall win, given how much cleaner the code is. I slightly prefer this version to the other one.

src/PythonExtensionGen.cpp
steven-johnson: Done

src/PythonExtensionGen.cpp
steven-johnson: Eh, maybe overthinking it? Conceptually, if it were written in Python, `None` would be the 'right' return value (since in C++ the return value is simply an error code, which is handled by exceptions here). Returning `True` isn't wrong per se -- it will likely be ignored by the caller -- but `None` seems right-er.

PR #6891, opened 2022-07-27
Upgrade clang-format and clang-tidy to v14

steven-johnson:
With LLVM16 being branched off, that means we'll be dropping LLVM13 in the months ahead; this PR proposes to upgrade our required versions of clang-tidy and clang-format to those for LLVM14 (so that no one will have to build or install LLVM13 just for those).

The changes for clang-format are insignificant (it appears to be a bit more aggressive about fixing single-line if statements, which is good).

The changes for clang-tidy include an update to the `modernize-use-default-member-init` which I think is arguably a good thing, so I elected to leave it enabled and have clang-tidy fix it. Basically, it prefers that member variables that are inited to constant values do so using the `member = value;` style, rather than the old `ctor() : member(value){}` style, e.g.

```
struct A {
  A() : i(5) {}
  A(int i) : i(i) {}
  int i;
};

// becomes

struct A {
  A() {}
  A(int i) : i(i) {}
  int i = 5;
};
```

### DIFF
--- a/src/CodeGen_LLVM.h
+++ b/src/CodeGen_LLVM.h
@@ -157,13 +157,13 @@ protected:
     virtual Type upgrade_type_for_argument_passing(const Type &) const;
 
     std::unique_ptr<llvm::Module> module;
-    llvm::Function *function;
-    llvm::LLVMContext *context;
-    llvm::IRBuilder<llvm::ConstantFolder, llvm::IRBuilderDefaultInserter> *builder;
-    llvm::Value *value;
-    llvm::MDNode *very_likely_branch;
-    llvm::MDNode *default_fp_math_md;
-    llvm::MDNode *strict_fp_math_md;
+    llvm::Function *function = nullptr;
+    llvm::LLVMContext *context = nullptr;
+    llvm::IRBuilder<llvm::ConstantFolder, llvm::IRBuilderDefaultInserter> *builder = nullptr;
+    llvm::Value *value = nullptr;
+    llvm::MDNode *very_likely_branch = nullptr;
+    llvm::MDNode *default_fp_math_md = nullptr;
+    llvm::MDNode *strict_fp_math_md = nullptr;
     std::vector<LoweredArgument> current_function_args;
     //@}
 
@@ -208,16 +208,16 @@ protected:
 
     /** Some useful llvm types */
     // @{
-    llvm::Type *void_t, *i1_t, *i8_t, *i16_t, *i32_t, *i64_t, *f16_t, *f32_t, *f64_t;
-    llvm::StructType *halide_buffer_t_type,
-        *type_t_type,
-        *dimension_t_type,
-        *metadata_t_type,
-        *argument_t_type,
-        *scalar_value_t_type,
-        *device_interface_t_type,
-        *pseudostack_slot_t_type,
-        *semaphore_t_type;
+    llvm::Type *void_t = nullptr, *i1_t = nullptr, *i8_t = nullptr, *i16_t = nullptr, *i32_t = nullptr, *i64_t = nullptr, *f16_t = nullptr, *f32_t = nullptr, *f64_t = nullptr;
+    llvm::StructType *halide_buffer_t_type = nullptr,
+                     *type_t_type,
+                     *dimension_t_type,
+                     *metadata_t_type = nullptr,
+                     *argument_t_type = nullptr,
+                     *scalar_value_t_type = nullptr,
+                     *device_interface_t_type = nullptr,
+                     *pseudostack_slot_t_type = nullptr,
+                     *semaphore_t_type;
 
     // @}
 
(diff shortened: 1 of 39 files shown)


src/CodeGen_LLVM.h
```diff
@@ -208,16 +208,16 @@ class CodeGen_LLVM : public IRVisitor {
...
-        *dimension_t_type,
-        *metadata_t_type,
-        *argument_t_type,
-        *scalar_value_t_type,
-        *device_interface_t_type,
-        *pseudostack_slot_t_type,
-        *semaphore_t_type;
+    llvm::Type *void_t = nullptr, *i1_t = nullptr, *i8_t = nullptr, *i16_t = nullptr, *i32_t = nullptr, *i64_t = nullptr, *f16_t = nullptr, *f32_t = nullptr, *f64_t = nullptr;
+    llvm::StructType *halide_buffer_t_type = nullptr,
+                     *type_t_type,
```
abadams, line 213: I think it's currently benign, but it would be good to initialize all of these to null instead of just some of them.

src/Parameter.cpp
```diff
@@ -19,15 +19,15 @@ struct ParameterContents {
...
     Buffer<> buffer;
-    uint64_t data;
+    uint64_t data = 0;
     int host_alignment;
     std::vector<BufferConstraint> buffer_constraints;
     Expr scalar_default, scalar_min, scalar_max, scalar_estimate;
     const bool is_buffer;
     MemoryType memory_type = MemoryType::Auto;
 
     ParameterContents(Type t, bool b, int d, const std::string &n)
```
abadams, line 29: Reading through this PR, I'm not so sure this new check is a good thing. It splits up initialization of fields by a not-very-useful criterion so that you have to look in two places to discover what the initial values are for the members.

tools/halide_image_io.h
```diff
@@ -1755,7 +1755,7 @@ struct halide_tiff_header {
 template<typename ElemType, int BUFFER_SIZE = 1024>
 struct ElemWriter {
     ElemWriter(FileOpener *f)
-        : f(f), next(&buf[0]), ok(true) {
```
abadams, line 1758: Here for example I think it's much clearer if the initial state for ElemWriter is written down in a single place, instead of being split between the constructor and the members.

steven-johnson: Interesting! I see it a different way: with this change, all of the members that are default-inited (either to a constant value or default-ctor for nontrivial types) are done so in the class declaration, leaving the classical "ctor-init-list" used only for things that are inited by an argument to the ctor itself (or by expressions that aren't otherwise constant). IMHO that is more clear.

abadams: This particular case was what tipped me over the edge to not liking it to the point of commenting. next(&buf[0]) is not a function of the arguments, it's just the initial value for that field. I don't see it as relevant that those bits aren't a compile-time constant.

src/StmtToHtml.cpp
```diff
@@ -1000,7 +1000,7 @@ class StmtToHtml : public IRVisitor {
     }
 
     StmtToHtml(const string &filename)
-        : id_count(0), context_stack(1, 0) {
+        : context_stack(1, 0) {
```
abadams, line 1003: Here's another case where it doesn't depend on the args - it's just not the sort of constant that's blessed with member initialization

src/autoschedulers/common/cmdline.h
```diff
@@ -779,7 +779,7 @@ class parser {
         std::string nam;
         char snam;
         std::string desc;
-        bool has;
+        bool has = false;
```
abadams, line 782: I also find it weird to read a list of members like this that vary between initialized and not, depending on whether their initial value is a compile-time constant or not.

steven-johnson: This particular example isn't a good one for your point, IMHO (the strings were always initialized 'here' and not in the ctor, and `snam` was left uninitialized, period)

That said, I see your point here -- it is unfortunate that C++ has multiple ways to do this, often located in separate files, and so regardless of which convention we use here, we can't know for sure whether (and how) `snam` is initialized at all. 

I'm not married to this change, but I must admit that I do prefer the idea (and readability) of member vars being explicitly initialized at the declaration point -- it strikes me as putting more critical information about a members initial value along with its name and type. Unless you are violently allergic to this, though, why don't I leave it open here for a while and see if anyone else has thoughts pro or con. (And we can easily disable `modernize-use-default-member-init` entirely and revert those parts of the change, so if opinion is a "nay", I still think upgrading the tool versions is worthwhile)

(It would be interesting to imagine a world in which C++ required *all* variables to be explicitly initialized... with the ability to explicitly leave a field uninitialized via `= poison;` or something similar.)

tools/halide_image_io.h
steven-johnson: I think this case is probably a limitation of clang-tidy being conservative (ie, being unable to prove to itself that this was a valid construct) -- `ElemType *next = &buf[0];` is perfectly valid here AFAICT (and I'll push a manual edit to this PR to fix that).

That said, this isn't an issue I want to spend any nontrivial amount of time on; I'll just open a new PR with the new clang-tidy behavior disabled and see how that looks. (The main desire here is just to keep our tooling relatively new, not to enforce arbitrary new checks.)

steven-johnson:
Please see https://github.com/halide/Halide/pull/6902 instead.

Issue #6893, opened 2022-07-28
[HVX] InjectHexagonRpc breaks CSE


rootjalex:
The InjectHexagonRpc mutator inserts `Let` nodes that label `Load` names instead of `Variable` names, specifically in the definition of `state_var()`. This breaks let substitutions (i.e. `RemoveLets()` in CSE.cpp), as `Let` is defined to only relate to `Variable` nodes:
https://github.com/halide/Halide/blob/b9a33561d38a4a59b9d4c08eb09253892b418a38/src/IR.h#L259-#L262

There are two solutions that I see:
- Allow `Let` nodes to label *any* name (i.e. load names). This makes substitution messier though, because you can't substitute `Let` values into `Load`s.
- InjectHexagonRpc should create `state_var`s in some other way - is there another way to represent pointer dereference?

This bug was discovered via #6884, which called CSE quite late in lowering.

PR #6894, opened 2022-07-28
[HVX] Fix state_var issue

rootjalex:
The issue in #6893 is the result of an attempt to dereference a pointer to the module_state pointer. This PR simply adds a runtime method to dereference the pointer, in order to fix that issue. I'm not necessarily sure if this is the best approach, would love feedback from anyone who knows the HVX runtime better than I do.

### DIFF
--- a/src/runtime/hexagon_host.cpp
+++ b/src/runtime/hexagon_host.cpp
@@ -755,10 +755,8 @@ WEAK uint64_t halide_hexagon_get_device_size(void *user_context, struct halide_b
     return handle->size;
 }
 
-WEAK void *halide_hexagon_get_module_state(void **host) {
-    if (host == nullptr) {
-        return nullptr;
-    }
+WEAK void *halide_hexagon_get_module_state(void *user_context, void **host) {
+    halide_abort_if_false(user_context, host != nullptr);
     return host[0];
 }
 
--- a/src/runtime/hexagon_host.cpp
+++ b/src/runtime/hexagon_host.cpp
@@ -755,6 +755,13 @@ WEAK uint64_t halide_hexagon_get_device_size(void *user_context, struct halide_b
     return handle->size;
 }
 
+WEAK void *halide_hexagon_get_module_state(void **host) {
+    if (host == nullptr) {
+        return nullptr;
+    }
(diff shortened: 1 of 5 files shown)


abadams:
@pranavb-ca might be able to weigh in

src/runtime/hexagon_host.cpp
```diff
@@ -755,6 +755,13 @@ WEAK uint64_t halide_hexagon_get_device_size(void *user_context, struct halide_b
     return handle->size;
 }
 
+WEAK void *halide_hexagon_get_module_state(void **host) {
+    if (host == nullptr) {
+        return nullptr;
```
steven-johnson, line 760: Is `nullptr` a value that the caller can deal with intelligently? (i.e., is this an error condition?)

rootjalex: I'm not quite sure. This check doesn't exist in the current implementation, so perhaps I should remove it. I wanted to avoid null pointer dereferencing.

steven-johnson: Always a good idea. Might be worth throwing an `abort()` or similar in the null path to see?

rootjalex: Sure, just tried it out in 2df1b2cdc

pranavb-ca:
LGTM. @aankit-ca can help you test this if you need. (sorry, i am on vacation with limited access to my work machine). I have added Ankit as a reviewer too.

rootjalex:
Thanks for taking a look @pranavb-ca - sorry for bothering you on your vacation!

steven-johnson:
So how confident are we in this change? Do we have a way to verify that the null-check code is actually being exercised?

rootjalex:
I don't think we have a way to verify that it is, however this change is actually safer - the current version in main will just load without the nullptr check.

steven-johnson:
Monday review ping to @aankit-ca

rootjalex:
bump

steven-johnson:
If @aankit-ca isn't available for this review, we should either find someone else or just land it as-is.

abadams (approved):

PR #6896, opened 2022-07-29
Don't try to fold saturating_sub of VectorReduce

### LINKED ISSUE #6883: `try_to_fold_vector_reduce<Call>` has incorrect behavior
In CodeGen_LLVM, we call `try_to_fold_vector_reduce<Call>` on `saturating_add` or `saturating_sub` calls, while not providing information as to whether or not the accumulation is an addition or a subtraction:

https://github.com/halide/Halide/blob/11a049c3967a277173e288ffd802f08ce1a1b78e/src/CodeGen_LLVM.cpp#L2835-#L2839

This seems like incorrect behavior - I noticed this while restructuring CodeGen_X86 into separate optimization and code generation passes, because it appears that the accumulating saturating dot product instructions should trigger on both of these patterns:
```
saturating_sub(wild_i32x, VectorReduce(SaturatingAdd, factor=4, widening_mul(wild_i16x, wild_i16x)))
```
```
saturating_add(wild_i32x, VectorReduce(SaturatingAdd, factor=4, widening_mul(wild_i16x, wild_i16x)))
```

rootjalex:
We should only try to fold the pattern:
```
saturating_add(x, VectorReduce(SaturatingAdd, y))
```
Currently, we also try to fold this pattern (incorrectly):
```
saturating_sub(x, VectorReduce(SaturatingAdd, y))
```

Fixes #6883

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -2832,7 +2832,7 @@ void CodeGen_LLVM::visit(const Call *op) {
         internal_assert(op->args.size() == 2);
 
         // Try to fold the vector reduce for a call to saturating_add
-        const bool folded = try_to_fold_vector_reduce<Call>(op->args[0], op->args[1]);
+        const bool folded = op->is_intrinsic(Call::saturating_add) && try_to_fold_vector_reduce<Call>(op->args[0], op->args[1]);
 
         if (!folded) {
             std::string intrin;


abadams (approved):

rootjalex:
@steven-johnson Do the failing buildbots have Sapphire Rapids processors? If not, the failures are likely unrelated, I believe that is the only processor with accumulating saturating dot product instructions

steven-johnson:
> Do the failing buildbots have Sapphire Rapids processors

almost certainly not. but all of these look like flakes to me. re-running now.

rootjalex:
Thank you!

steven-johnson:
Remaining failure is definitely a flake, landing

PR #6897, opened 2022-07-29
[Codegen_LLVM] Annotate LLVM IR functions with `nounwind`/`mustprogress` attributes

LebedevRI:
My reasoning is as follows, please correct me if i'm wrong:
1. Halide-generated code never throws exceptions
2. Halide-generated code always `call`s (as opposed to `invoke`s) the functions, there is no exception-safety RAII
3. Halide loops are meant to have finite number of iterations, they aren't meant to be endless and side-effect free
4. Halide (IR) assertions *might* abort.
5. Likewise, external callees *might* abort. (???)

Therefore, when not in presence of external calls,
it is obvious that (1) no exception will be unwinded
out of the halide-generated function, (2) none of the loops
will end up being endless with no observable side-effects.
... which is the semantics that is being stated by the
    LLVM IR function attributes `nounwind`+`mustprogress`.

I'm less clear as to what are the prerequisites on the behavior of
the external callees, but i do believe that they must also
at least not unwind. I guess they are also at least required to
either return or abort eventually.

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -701,6 +701,16 @@ void set_function_attributes_from_halide_target_options(llvm::Function &fn) {
     fn.addFnAttr("tune-cpu", mcpu_tune);
     fn.addFnAttr("target-features", mattrs);
 
+    // Halide-generated IR is not exception-safe.
+    // No exception should unwind out of Halide functions.
+    // No exception should be thrown within Halide functions.
+    // All functions called by the Halide function must not unwind.
+    fn.setDoesNotThrow();
+
+    // Side-effect-free loops are undefined.
+    // But asserts and external calls *might* abort.
+    fn.setMustProgress();
+
     // Turn off approximate reciprocals for division. It's too
     // inaccurate even for us.
     fn.addFnAttr("reciprocal-estimates", "none");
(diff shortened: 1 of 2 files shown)


abadams:
nounwind should be correct, but asserts call halide_error, which might abort, depending on the overload in place, or even do something horrible like a longjmp. The default one aborts.

LebedevRI:
> nounwind should be correct, but asserts call halide_error, which might abort, depending on the overload in place, or even do something horrible like a longjmp. The default one aborts.

Hmm, interesting, then when asserts are present, it can't be `willreturn`, but only `mustprogress`.
I'm guessing it's the same story about external calls?

abadams:
Yes I think mustprogress is valid. One of the benefits of not being Turing-complete I guess.

LebedevRI:
Cool, updated the PR.
As a follow-up, we could deduce `willreturn` when `no_asserts` + no external calls.

abadams (approved):

LebedevRI:
@abadams while there, can you trigger the bots on this, please? :)

LebedevRI:
Looks like all good here?
https://buildbot.halide-lang.org/master/#/builders/123/builds/39 is red regardless,
and https://buildbot.halide-lang.org/master/#/builders/81/builds/35 needs to be retriggered.

LebedevRI:
Thanks!

PR #6900, opened 2022-07-30
Make saturating_cast an intrinsic

rootjalex:
This PR should produce no change in any code behavior, it simply replaces `saturating_cast` with an intrinsic, which delays the lowering a bit, but makes writing pattern-matching code (i.e. #6884 and the follow-on ARM PR that I am working on) far easier to write and more concise.

The only possible change is that constant folding of `saturating_cast` may be delayed as a result, though I am unsure how common of a case that would be.


abadams:
The only possible difference I can imagine is something like f(saturating_cast<uint8_t>(clamp(x, 5, 10))) is going to have worse bounds than it currently does. Maybe saturating_cast needs to be handled in Bounds.cpp?

rootjalex:
Makes sense, I'll see what I can do.

rootjalex:
This PR has become a bit more extensive than I expected, so here are the main updates:

- Handle `saturating_cast` in bounds inference - this is a very conservative handler, but essentially bounds information is only kept across a call to `saturating_cast` if `can_prove` can prove that the cast does not saturate, otherwise `bounds_of_type` of the casted-to type are used. Multiple tests have been added to confirm the expected behavior.
- Multiple CodeGen backends (WebAssembly, ARM, HexagonOptimize, and X86) that have special handling for saturating casts have been updated to look for the patterns in `Call` visitors instead of `Cast` visitors. ARM has the most extensive changes.
- FindIntrinsics.cpp has been updated both with patterns that lift to saturating casts and patterns that incorporate saturating casts in the LHS (i.e. patterns for `saturating_add`. I've also added code that performs simplifications, i.e. `i8_sat(int32(wild_i16x)) -> i8_sat(wild_i16x)` related to `saturating_cast` - these simplifications were necessary in order to pass simd_op_check for some ARM instructions.

@abadams, I believe this PR is ready for review again, thanks for the patience!

src/CodeGen_ARM.cpp
```diff
@@ -1177,12 +1145,55 @@ void CodeGen_ARM::visit(const Call *op) {
...
                 value = call_overloaded_intrin(op->type, pattern.intrin, matches);
                 if (value) {
                     return;
                 }
             }
         }
+
+        // If we didn't find a pattern, try rewriting any saturating casts.
+        static const vector<pair<Expr, Expr>> cast_rewrites = {
+            // Double or triple narrowing saturating casts are better expressed as
```
abadams, line 1168: I know this is just code motion, but is it me or does this comment not match the code?

src/FindIntrinsics.cpp
```diff
@@ -575,6 +583,57 @@ class FindIntrinsics : public IRMutator {
...
+        if (
+            // Saturating patterns.
+            rewrite(saturating_cast(op->type, widening_add(x, y)),
+                    saturating_add(x, y),
+                    is_x_same_int_or_uint) ||
+            rewrite(saturating_cast(op->type, widening_sub(x, y)),
+                    saturating_sub(x, y),
+                    is_x_same_int_or_uint) ||
+            rewrite(saturating_cast(op->type, shift_right(widening_mul(x, y), z)),
+                    mul_shift_right(x, y, cast(unsigned_type, z)),
```
abadams, line 605: Why is there a cast to the unsigned type of z if the predicate says it's already an unsigned type?

src/IRMatch.h
```diff
@@ -1358,6 +1358,7 @@ struct Intrin {
     struct pattern_tag {};
     Call::IntrinsicOp intrin;
     std::tuple<Args...> args;
+    Type optional_type_hint;
```
abadams, line 1361: At first I thought this was a type hint for the args and got confused. A comment like this would have helped:
// The type of the output of the intrinsic node. Only necessary in cases where it can't be inferred from the input types (e.g. saturating_cast).

abadams (approved):
Makes sense. Just had some minor comments on the comments.

src/CodeGen_ARM.cpp
rootjalex: I think the comment does match the code? Its better to express double-narrowing saturating casts as two single-narrowing saturating casts

src/FindIntrinsics.cpp
rootjalex: I think a combination of copy-paste and a lack of double-checking. I will fix it

src/IRMatch.h
rootjalex: Will do!

src/CodeGen_ARM.cpp
```diff
@@ -1177,12 +1145,55 @@ void CodeGen_ARM::visit(const Call *op) {
...
+            // regular narrowing casts.
```
abadams, line 1169: "regular narrowing casts" sounds like they're not saturating casts

rootjalex: Ah gotcha - I will fix it.

src/FindIntrinsics.cpp
rootjalex: Oh, no. The cast is actually a narrowing cast, because `unsigned_type` is the output type bitwidth. This matches the original patterns here: https://github.com/halide/Halide/blob/9ca75606a0a9c8fa43c1dee82db734cc3d0f7e39/src/FindIntrinsics.cpp#L454-#L469
I am now convinced that this is actually a bug, we should be checking if `z` can be losslessly casted to the unsigned type.
On current main, I see this buggy behavior:
```
Type t = Int(16, 16);
Expr x = Variable::make(t, "x");
Expr y = Variable::make(t, "y");
Expr z = Variable::make(UInt(32, 16), "z");
Expr expr = cast(t, clamp(widening_mul(x, y) >> z, t.min(), t.max()));
std::cout << expr << "\n";
std::cout << find_intrinsics(expr) << "\n";
```
```
int16x16(max(min((int32x16)shift_right((int32x16)widening_mul((int16x16)x, (int16x16)y), (uint32x16)z), x16(32767)), x16(-32768)))
(int16x16)mul_shift_right((int16x16)x, (int16x16)y, uint16x16((uint32x16)z))
```
I believe this is a bug, depending on our definition of `shift_right`. Is it defined for all values of the rhs or only `[-bits, +bits]`?

rootjalex: If the latter is true, then this is a safe lifting rule to make, otherwise we may need to fix multiple rules

src/IRMatch.h
rootjalex: I also updated the `match` method to check for matching the `type_hint` if it is defined - because `saturating_cast(Uint(8), x)` could conceivably match any `saturating_cast` call before this change.

rootjalex:
I don't want to merge this without confirmation from Andrew about shift semantics - technically we wouldn't be introducing anything new, as the existing rules already perform the narrowing cast on the rhs operand, but I just want to be sure before adding more rules that perform this operation.

src/FindIntrinsics.cpp
rootjalex: Discussed with @abadams  offline, the assumption here is that we assume no out-of-range shifts can be present in the input, therefore these rules are perfectly valid.

steven-johnson:
> The only possible change is that constant folding of saturating_cast may be delayed as a result, though I am unsure how common of a case that would be.

I suspect this is causing many constant-folding-overflow failures in Google code, investigating

steven-johnson:
Here's a repro case that breaks with this PR in place, but worked before:

```
    Var x, y, c;

    const int kLutSize = 256;
    Func lut("lut");
    lut(x) = u8(x);
    lut.compute_root();

    Func input;
    input(x, y, c) = u16(x + y + c);
    input.compute_root();

    Func a;
    a(x, y, c) = u16_sat((i32(input(x, y, 0)) * i16(c) +
                          i32(input(x, y, 1)) * i16(c) +
                          i32(input(x, y, 2)) * i16(c)));
    a.compute_root();

    Expr b = a(x, y, c);
    const int kLutResampleRate = 1024 / kLutSize;
    Expr x0 = u16(b / kLutResampleRate);

    Func output;
    output(x, y, c) = lut(x0);

    auto foo = output.realize({16, 16, 3});
```

rootjalex:
Thanks for the repro example! It appears the overflow is being introduced in `compute_function_value_bounds` for `a`, during the simplification here: https://github.com/halide/Halide/blob/f7a30e06e4aa1331eb060f18c80eeced5431ddbd/src/Bounds.cpp#L1220-#L1225 . I am not quite sure what changed here, but I will keep digging.

steven-johnson:
Also note, the OpenGLCompute backend isn't handling this intrinsic and will fail with 'unknown intrinsic', I'll prep a PR

rootjalex:
I expected all non-CPU backends to call `lower_intrinsic` on unknown intrinsics, instead of panicking - but I am also very *not* familiar with any non-CPU backend.

PR #6901, opened 2022-07-30
[Simplify] Drop no-op single-input identity shuffles

LebedevRI:
This came up in #6887, happens sometimes even in `apps/`.

(and in fact i initially forgot to recurse, so it was micompiling stuff, took way to look to pinpoint..)


rootjalex (approved):

LebedevRI:
Thanks!

Issue #6899, opened 2022-07-29
AutoSchedule - NewAPI with Adams2019 - Condition failed: pipeline_feat_queue.data()


soufianekhiat:
Commit:
11a049c3967a277173e288ffd802f08ce1a1b78e
#6863 [- Fixes to make address sanitizer happy for internal runtime cl…

Calling:
```cpp
Target oTargetCPU = ...
Func oFunc = ...
AutoSchedulerResults oResult =
	oFunc.apply_autoscheduler( oTargetCPU,
	{ "Adams2019",
	//{ "Mullapudi2016",
		{ { "parallelism", "16" } }
	} );
```
Work fine with "Mullapudi2016" and assert with "Adams2019".
Assert:
Internal Error at F:\git\Halide\src\autoschedulers\adams2019\DefaultCostModel.cpp:139 triggered by user code at : Condition failed: pipeline_feat_queue.data() && "Call set_schedule_features before calling enqueue\n":

Did I miss something on the code convention of the new API?

steven-johnson:
I've never seen that assert fire before, locally or elsewhere, so offhand I don't know what could be causing it. If you can post a more-complete piece of code that demonstrates the problem, I'd be happy to take a look.

soufianekhiat:
Sure,
Here the full code that trigger this assert:
```cpp
#include <Halide.h>
#include <vector>

int main()
{
	namespace H = Halide;

	std::vector<float> src;
	src.resize( 256 );
	float xi = 1.0f;
	for ( float& x : src )
	{
		x = xi;
		xi += 1.0f;
	}

	int const iSize = static_cast< int >( src.size() );

	H::Buffer<float, 1> hBuffer( &src[ 0 ], { iSize } );

	H::Func f( "buffer_func" );
	H::Var x( "x ");

	f( x ) = hBuffer( x );

	H::load_plugin( "autoschedule_adams2019.dll" );
	H::load_plugin( "autoschedule_li2018.dll" );
	H::load_plugin( "autoschedule_mullapudi2016.dll" );

	H::Target targetCPU = H::get_jit_target_from_environment();

	f.set_estimates( { { 0, iSize } } );

	H::Pipeline p( f );
	H::AutoSchedulerResults result =
		p.apply_autoscheduler(	targetCPU,
								{ "Adams2019",
								//{ "Mullapudi2016",
								   { { "parallelism", "16" } }
								} );

	return 0;
}
```
Thanks

steven-johnson:
[...]

PR #6903, opened 2022-08-01
Ensure autoscheduler plugins are built with -rdynamic or equiv

steven-johnson:
Autoscheduler plugins are supposed to be built with `-rdynamic` or the equivalent, so that they never attempt to load any libraries to resolve symbols. However:
- Our Make builds used the right flags, but also included libHalide in the linker path, so we could end up with it loaded if you had things configured the right (wrong?) way; this could end up with the plugin registering itself in a 'captive' libHalide with its own globals. Oops.
- Our CMake build never attempted to set these flags at all.

Note that (AFAICT) this isn't an issue on Windows, where `-rdynamic` behavior is more of the default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1903,6 +1903,14 @@ correctness_opencl_runtime: $(BIN_DIR)/$(TARGET)/correctness_opencl_runtime
 	cd $(TMP_DIR) ; $(CURDIR)/$<
 	@-echo
 
+correctness_load_plugin: $(BIN_DIR)/correctness_load_plugin autoschedulers
+	@-mkdir -p $(TMP_DIR)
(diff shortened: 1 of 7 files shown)


steven-johnson:
(Found as a result of investigation info https://github.com/halide/Halide/issues/6899; this issue masked my ability to replicate the autoscheduler assertion on non-Windows machines)

alexreinking:
From out-of-band conversation: `-rdynamic` isn't relevant on Linux, here. @steven-johnson is finding a more targeted fix.

steven-johnson:
Yeah, pretty sure this happens to fix the bug I saw but in a misleading way, I'm gonna close this, stand by for a new one

PR #6904, opened 2022-08-01
[Codegen_LLVM] Arguments/Loads/Returns are well-defined

LebedevRI:
https://llvm.org/docs/LangRef.html#well-defined-values:
```
Given a program execution, a value is well defined if the value
does not have an undef bit and is not poison in the execution.
An aggregate value or vector is well defined if its elements
are well defined. The padding of an aggregate isn’t considered,
since it isn’t visible without storing it into memory and
loading it with a different type.
```

Same story as with https://github.com/halide/Halide/pull/6897,
this matches //my// understanding of Halide's semantics,
but this may very well not be what the reality is.
(@alexreinking/@abadams ?)

To be noted, even //if// this is correct, i do believe
this has a very high chance to cause issues.
I'm also not sure how this plays with MSan.

Notably, while annotating function arguments/return is trivial,
for load's this only annotates the `load` instructions that we
actually create, i.e. it misses loads in the IR that we precompiled
and then link into our module.

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -711,6 +711,16 @@ void set_function_attributes_from_halide_target_options(llvm::Function &fn) {
     // But asserts and external calls *might* abort.
     fn.setMustProgress();
 
+    // All parameters and return values are well-defined,
+    // do not have any undef bits, and are not poison in execution.
+    if (!fn.getReturnType()->isVoidTy())
+        fn.addRetAttr(Attribute::NoUndef);
+    for (Argument &A : fn.args()) {
+        if (!A.hasAttribute(Attribute::ImmArg)) {
+            A.addAttr(Attribute::NoUndef);
+        }
+    }
+
     // Turn off approximate reciprocals for division. It's too
     // inaccurate even for us.
     fn.addFnAttr("reciprocal-estimates", "none");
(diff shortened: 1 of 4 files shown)


abadams:
Some Halide schedules intentionally load values that were never stored to. E.g. it's possible to generate this IR:

allocate tmp1[16], tmp2[16]
tmp1[0] = input[0]
tmp2[ramp(0, 1, 16)] = tmp1[ramp(0, 1, 16)]
output[0] = tmp2[0]

Does that mean it's not OK to annotate load nodes in this way?

LebedevRI:
Eeek, so as per the msan instrumentation, we do not require *at all* that even the input bufferes
are anywhat initialized, and simply mark the output buffers as fully initialized regardless?

LebedevRI:
To spell this out explicitly: the problem with sometimes loading junk is that if we knowingly intentionally sometimes do it,
then in those cases where we didn't mean to do that, we can't actually tell whether or not we intentionally loaded junk,
or we were passed junk input (which we should diagnose). So _ideally_ yeah, we shouldn't be intentionally loading junk...

steven-johnson:
> So ideally yeah, we shouldn't be intentionally loading junk

No argument here. I'd be in favor of revisiting a 'natural' MSAN implementation (as a prerequisite to this) and seeing just how performant (or not) it is.

abadams:
We never intentionally load junk from input buffers, only from allocations within a pipeline. The reason is because it helps generate tighter vector code in many cases, and provably can't affect any values stored to output buffers.

PR #6906, opened 2022-08-01
Fix broken Makefile rules for autoschedulers on OSX

steven-johnson:
A few issues here:
- Make was building the plugins as .dylib on OSX, but they should have been .so to match Linux (and just on general principles)
- On OSX, explicitly linking libHalide.dylib into a plugin means that it will load its own copy of libHalide, which is bad, because it means the plugin doesn't share the same set of globals. We need to omit that explicit dependency and allow it to just find the exported symbols at load time.
- Add a test to verify the fix; run it everywhere even though it should only have been failing for Make-build OSX builds.

Finally, let me add that we really need to set a sunset date for supporting Make in Halide. The Makefiles aren't really maintained properly anymore, and when something subtle goes wrong, it takes an unreasonable amount of time to debug for something that is no longer our canonical build tool.


alexreinking:
> Finally, let me add that we really need to set a sunset date for supporting Make in Halide. The Makefiles aren't really maintained properly anymore, and when something subtle goes wrong, it takes an unreasonable amount of time to debug for something that is no longer our canonical build tool.

amen to that

steven-johnson:
> > Finally, let me add that we really need to set a sunset date for supporting Make in Halide. The Makefiles aren't really maintained properly anymore, and when something subtle goes wrong, it takes an unreasonable amount of time to debug for something that is no longer our canonical build tool.
> 
> amen to that

Opened https://github.com/halide/Halide/issues/6907 to hopefully start a serious discussion about it

src/autoschedulers/mullapudi2016/Makefile
```diff
@@ -15,6 +15,8 @@ endif
 
 CXXFLAGS += -I$(COMMON_DIR)
 
-$(BIN)/libautoschedule_mullapudi2016.$(SHARED_EXT): $(SRC)/AutoSchedule.cpp $(LIB_HALIDE)
```
alexreinking, line 18: Should this be order-only instead? If `AutoSchedule.cpp` includes Halide.h, it needs a dependency on that somehow.

steven-johnson: ...yeah, probably

alexreinking (approved):

PR #6909, opened 2022-08-03
[AMX] syscall to enable AMX instructions on real hardware

frengels:
The current work on AMX with #6582 has made it possible to generate AMX instructions for various tile allocations on the emulator. However for the real hardware there's a `syscall` required to enable the feature in the kernel, I've used the [following](https://github.com/torvalds/linux/blob/v5.16/tools/testing/selftests/x86/amx.c) (5.16 is the lowest kernel version supported) as inspiration and tested the result on intel's devcloud.

A few concerns I have:
- Currently injecting the `syscall` will enable the instructions on real hardware but trigger a fault in the emulator.
  - Would it advisable to add another target feature for the emulator (sde), so it's possible to disable this `syscall`?
- I'm not sure if I handled the AMX feature detection correctly
  - Could it be that this should be part of `src/runtime/x86_cpu_features.cpp`?
- Theoretically a user could disable AMX support at runtime making the CpuFeatures no longer a static truth like assumed in `halide_default_can_use_target_features`.
  - We could assume nobody will realistically enable/disable the feature during runtime of a program.
  - Disable caching of the `CpuFeatures` if the CPU could possibly support AMX.

Looking forward to any feedback.


abadams:
This seems to be a syscall that you make once and it enables AMX usage for the entire process? I'm not sure if Halide should be responsible for that if so. If it were scoped and you're supposed to enable and disable AMX access around every use of it, then it would make sense to inject that into our generated code, but without scoping, shouldn't the user be responsible for making sure their process has access to AMX? We could perhaps provide a user-callable function for it in the runtime module.

The reason I'm wary is that it looks like this will only work on linux x86-64 (e.g. it hardcodes a syscall number), and it will actively fail on windows, macos, etc. Aren't people going to use AMX on Windows boxes at some point?

src/LLVM_Runtime_Linker.cpp
```diff
@@ -1087,6 +1089,7 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
                 modules.push_back(get_initmod_x86_avx512_ll(c));
             }
             if (t.has_feature(Target::AVX512_SapphireRapids)) {
+                modules.push_back(get_initmod_amx(c, bits_64, debug));
```
zvookin, line 1092: This appears to have the possibility of injecting Linux specific code on non-Linux paltforms.

zvookin:
I totally agree with Andrew's request for more info on how this works. To clarify the Linux specific nature of this, the syscall being Linux only is not a show stopper, but the code does need to indicate that this is the case and if other systems require similar system calls, we would likely want to mirror the support for those platforms. (Or at least setup the filenaming and such to allow for it.) E.g. see how ```halide_host_cpu_count``` works.

frengels:
It is correct that only need to perform the syscall once to enable AMX for the entire process. I agree that it should be called by the user then. I'll work on scoping the code correctly to x86_64 linux.

src/runtime/runtime_internal.h
```diff
@@ -186,6 +186,8 @@ struct halide_pseudostack_slot_t {
 WEAK void halide_use_jit_module();
 WEAK void halide_release_jit_module();
 
+WEAK int halide_enable_amx();
```
abadams, line 189: If we expect users to call this, I think you want the declaration in HalideRuntime.h instead.

frengels: I'm having some trouble manually calling the function from outside of a kernel. I think it's because the runtime is only linked into the kernel's object file, I'm guessing I'd have to explicitly link the runtime into the hosting binary as well?

abadams:
You might need to add a reference to it in the list in src/runtime/runtime_api.cpp to stop it from getting dead-stripped when not linking a standalone runtime.

Makefile
```diff
@@ -1008,12 +1009,23 @@ RUNTIME_TRIPLE_WIN_ARM_32 = "arm-unknown-windows-unknown"
...
+RUNTIME_TRIPLE_AMX_X86_32 = "i386-unknown-unknown-unknown"
+RUNTIME_TRIPLE_AMX_X86_64 = "x86_64-unknown-unknown-unknown"
+
 # `-fno-threadsafe-statics` is very important here (note that it allows us to use a 'modern' C++
 # standard but still skip threadsafe guards for static initialization in our runtime code)
 #
 # `-fno-rtti` is necessary to allow us to use classes with virtual functions in the runtime code
 RUNTIME_CXX_FLAGS = -std=c++17 -O3 -fno-vectorize -ffreestanding -fno-blocks -fno-exceptions -fno-unwind-tables -fno-threadsafe-statics -fno-rtti
 
+$(BUILD_DIR)/initmod.amx_x86_32.ll: $(SRC_DIR)/runtime/amx.cpp $(BUILD_DIR)/clang_ok
```
steven-johnson, line 1021: FYI, these changes/additions will also need to be made in src/runtime/CMakeLists.txt -- we support CMake on all our platforms, whereas Make only on a small percentage.

PR #6911, opened 2022-08-03
Add sub-licenses to toplevel LICENSE.txt

steven-johnson:
For compliance reasons, all licensed code in this repo needs to have the license text replicated in the toplevel LICENSE file, as is already the case for (e.g.) apps/bgu and apps/cmdline.h

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -63,3 +63,31 @@ LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
 ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+----
+
+dependencies/spirv is Copyright (c) 2015-2018 The Khronos Group Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and/or associated documentation files (the
+"Materials"), to deal in the Materials without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Materials, and to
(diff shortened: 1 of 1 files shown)


LebedevRI:
Sounds reasonable to me given it's explicitly stated to which files that applies.

zvookin:
If we put a license title in the separators, will the diff be cleaner? Also, is the text reformatting just done for looks? Generally one wants the text to be as close to the original file as possible, including whitespace.

derek-gerstmann:
Thanks for merging all of these into the top-level license file.  Should we add the ONNX app too (Apache 2.0)?  The app requires the .proto file in order to run, which is distributed in binary form.

steven-johnson:
> If we put a license title in the separators, will the diff be cleaner? 

Not sure I understand.

> Also, is the text reformatting just done for looks? Generally one wants the text to be as close to the original file as possible, including whitespace.

It was a mix of text wrapped at different widths (and the BLAS was not wrapped at all in the original source) so I rewrapped to 80 chars for uniformity. Revering is easy if the we want to keep it closer to original.

> Should we add the ONNX app too (Apache 2.0)? The app requires the .proto file in order to run, which is distributed in binary form.

Probably, yeah, stand by

zvookin:
It is different wrapping because that is what the original license files do. Generally leaving this is preferred, but if we want to change it, I'd really prefer a separate commit so the addition of a license is completely separate form changes to other licenses.

Per the diff, when reading this sort of thing, it's very important to know what license changed. If a change that is only to one license spills over to another due to diff getting confused, that's an issue. But it is probably not what is happening here, rather there are more things changing than I expected. I'd say separate PRs for each license change are a good idea.

steven-johnson:
> > Should we add the ONNX app too (Apache 2.0)? The app requires the .proto file in order to run, which is distributed in binary form.
> 
> Probably, yeah, stand by

Actually, no, not in this case: we aren't including any ONNX-related source or binary that is Apache-licensed. You must get the Apache-licensed thing yourself.

steven-johnson:
PTAL

derek-gerstmann:
> PTAL

Ahh ... okay, it wasn't clear if the use of the reliance on the binaries required a license or not.  LGTM.

steven-johnson:
> I'd say separate PRs for each license change are a good idea.

Fair, I'll close this and offer several separate ones instead.

PR #6916, opened 2022-08-03
Upgrade CMake minimum version to 3.22

### LINKED ISSUE #6910: Upgrade CMake to 3.22
We've long had a policy that we tie our minimum CMake version to the one present in the most recent Ubuntu LTS release. Well, 22.04 has been out for a while and it bundles CMake 3.22, so it's time we moved on this.

There are several places where this will make our build simpler and make upcoming projects easier to execute. An inexhaustive list in no particular order:

* `$<TARGET_RUNTIME_DLLS:...>` will help us produce viable `pip` packages on Windows.
* CMP0116 will make our runtime modules' build a bit simpler
* CMP0125 and CMP0126 fix CMake's ludicrous cache variable-normal variable interactions and will make it viable for us to hard-code fewer things
* `cmake_dependent_option` is less wonky and easier to use
* We'll be able to use newer CUDA-related CMake features in our apps' and tests' builds.
* VSCode requires CMake 3.21 as a minimum for first-tier support.

For what it's worth, I think we should try to "live at head" with respect to CMake. I think we should upgrade our minimum any time we find that we need to write special cases based on policy or CMake version, or we find ourselves hacking in a feature that exists in a newer version.

steven-johnson:
Fixes #6910.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16...3.23)
+cmake_minimum_required(VERSION 3.22...3.23)
 project(Halide
         VERSION 15.0.0
         DESCRIPTION "Halide compiler and libraries"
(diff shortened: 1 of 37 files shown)


alexreinking (approved):
LGTM. I'll actually make use of the new min version in a follow-up PR

PR #6917, opened 2022-08-03
Remove unused GHA and packaging workflows.

alexreinking:
Now that @LebedevRI implemented upstream Debian packaging, there's no need to keep our own around. Remove that, along with unused GHA workflows


steven-johnson (approved):
I loooooove all-red PRs

LebedevRI:
Note that nightly builds are presently still not a thing.

PR #6920, opened 2022-08-04
Add ASAN support to CMake via toolchain file

steven-johnson:
Perhaps a cleaner approach for ASAN in CMake, this uses a toolchain file to drive clang to run asan -- the point here being that everything participating in the ASAN Party has to use the same version of the same compiler (including the bitcode in our runtime), so this means we pretty much have to use the LLVM-provided clang. 

Note that (as written) it will only work on Linux, though it could probably be adapted to work on any posixy system without much work.  Since we'd probably only run sanitizer coverage on our Linux bots anyway, this is likely fine. 

Note also that it requires building some C++ runtime code in LLVM that we currently don't do by default.

Note also that it currently doesn't test any Python code under ASAN; to make that work, we'd need to use `-shared-asan` (vs Clang's default of `-static-asan`); in theory that can be made to work, but it requires setting an LD_PRELOAD env var in lots of places across CMake, which is uglier than I'd like, so I have just skipped it for now. (Maybe there's a cleaner way?)

Anyway, if this approach works, we'd probably add more toolchain files for other sanitizers.


apps/CMakeLists.txt
```diff
@@ -19,7 +19,11 @@ enable_testing()
 # add_subdirectory(HelloWasm)  # TODO(#5374): missing CMake build
 # add_subdirectory(HelloiOS)  # TODO(#5374): missing CMake build
 # add_subdirectory(auto_viz)  # TODO(#5374): missing CMake build
-add_subdirectory(bgu)
+if (Halide_ENABLE_ASAN)
```
alexreinking, line 22: I think you meant `Halide_ASAN_ENABLED` here, but I think the approach is off.

A third-party user might add the asan flags themselves without using our toolchain file. We can detect whether this is the case by using `check_symbol_exists(__SANITIZE_ADDRESS__ "" ASAN_ENABLED)` and then branching off `ASAN_ENABLED`. All implementations that I'm aware of define `__SANITIZE_ADDRESS__` iff ASAN is enabled.

src/CMakeLists.txt
```diff
@@ -360,7 +360,7 @@ set(HALIDE_H "${Halide_BINARY_DIR}/include/Halide.h")
 set(LICENSE_PATH "${Halide_SOURCE_DIR}/LICENSE.txt")
 add_custom_command(OUTPUT "${Halide_BINARY_DIR}/include/Halide.h"
                    COMMAND ${CMAKE_COMMAND} -E make_directory "$<SHELL_PATH:${Halide_BINARY_DIR}/include>"
-                   COMMAND build_halide_h "$<SHELL_PATH:${LICENSE_PATH}>" ${HEADER_FILES} > "$<SHELL_PATH:${HALIDE_H}>"
+                   COMMAND $<TARGET_FILE:build_halide_h> "$<SHELL_PATH:${LICENSE_PATH}>" ${HEADER_FILES} > "$<SHELL_PATH:${HALIDE_H}>"
```
alexreinking, line 363: I don't think this is necessary with the crosscompiling emulator change above.

cmake/toolchain.linux-x64-asan.cmake
```diff
@@ -0,0 +1,55 @@
...
+set(CMAKE_SHARED_LINKER_FLAGS_INIT ${_ASAN_FLAGS})
+set(CMAKE_MODULE_LINKER_FLAGS_INIT ${_ASAN_FLAGS})
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+
+# Set to empty string to indicate the resulting binaries can be natively executed
+set(CMAKE_CROSSCOMPILING_EMULATOR)
```
alexreinking, line 37: I think this should be `/usr/bin/env` to be a no-op. Empty makes CMake think "there is no emulator", rather than "no emulator is necessary".

cmake/toolchain.linux-x64-asan.cmake
```diff
@@ -0,0 +1,55 @@
...
+
+# A few downstream build rules need to be able to sniff these:
+# - ASAN means we can't run with large stacks, so some things will overflow
+set(Halide_ASAN_ENABLED ON)
+# - Pretty much every sanitizer affects performance, so skip benchmarks
+set(Halide_ANY_SANITIZERS_ENABLED ON)
```
alexreinking, line 43: See my other comment about detection.

python_bindings/CMakeLists.txt
```diff
@@ -56,6 +56,15 @@ endif ()
 add_subdirectory(src/halide)
 add_subdirectory(stub)
 
+if (Halide_ASAN_ENABLED)
+    # Getting asan to work in shared mode requires some LD_PRELOAD hackery
+    # in a lot of places; just skipping them for now.
+    if (WITH_TEST_PYTHON OR WITH_TUTORIALS)
+        message(STATUS "Python tests/tutorials are not yet supported under ASAN.")
```
alexreinking, line 63: Should probably be a `WARNING`

src/autoschedulers/adams2019/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 set(WF_CPP baseline.cpp)
 configure_file(baseline.weights baseline.weights COPYONLY)
 add_custom_command(OUTPUT ${WF_CPP}
-                   COMMAND binary2cpp baseline_weights < baseline.weights > ${WF_CPP}
+                   COMMAND $<TARGET_FILE:binary2cpp> baseline_weights < baseline.weights > ${WF_CPP}
```
alexreinking, line 9: Shouldn't be needed

src/runtime/CMakeLists.txt
```diff
@@ -279,7 +279,7 @@ foreach (i IN LISTS RUNTIME_CPP)
                                VERBATIM)
 
             add_custom_command(OUTPUT "${INITMOD}"
-                               COMMAND binary2cpp ${SYMBOL} < "${BC}" > "${INITMOD}"
+                               COMMAND $<TARGET_FILE:binary2cpp> ${SYMBOL} < "${BC}" > "${INITMOD}"
```
alexreinking, line 282: Shouldn't be needed

cmake/toolchain.linux-x64-asan.cmake
```diff
@@ -0,0 +1,55 @@
...
+#
+# Note: requires LLVM to be built with -DLLVM_ENABLE_RUNTIMES="compiler-rt;libcxx;libcxxabi;libunwind"
+#
+# Note: only tested with LLVM/Clang 16 as of this comment. Earlier versions
+# may likely work but are not tested.
+
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR i686)
+
+set(LLVM_INSTALL_DIR ${LLVM_DIR}/../../../)
```
alexreinking, line 16: Should `FATAL_ERROR` if `LLVM_DIR` is not set

cmake/toolchain.linux-x64-asan.cmake
```diff
@@ -0,0 +1,55 @@
...
+set(Halide_ASAN_ENABLED ON)
+# - Pretty much every sanitizer affects performance, so skip benchmarks
+set(Halide_ANY_SANITIZERS_ENABLED ON)
+
+# If running under ASAN, we need to suppress some errors:
+# - detect_leaks, because circular Expr chains in Halide can indeed leak,
+#   but we don't care here
+# - detect_container_overflow, because this is a known false-positive
+#   if compiling with a non-ASAN build of LLVM (which is usually the case)
+set(SANITIZER_ENV_VARS "ASAN_OPTIONS=detect_leaks=0:detect_container_overflow=0")
```
alexreinking, line 50: Might want this as a CACHE variable for easier ad-hoc overriding

src/runtime/CMakeLists.txt
```diff
@@ -322,7 +322,7 @@ foreach (i IN LISTS RUNTIME_LL)
                        DEPENDS "${LL_TRANSFORMED}"
                        VERBATIM)
     add_custom_command(OUTPUT "${INITMOD}"
-                       COMMAND binary2cpp "halide_internal_initmod_${i}_ll" < "${BC}" > "${INITMOD}"
+                       COMMAND $<TARGET_FILE:binary2cpp> "halide_internal_initmod_${i}_ll" < "${BC}" > "${INITMOD}"
```
alexreinking, line 325: Shouldn't be needed

src/runtime/CMakeLists.txt
```diff
@@ -333,22 +333,22 @@ foreach (i IN LISTS RUNTIME_BC)
     set(RT_BC "${CMAKE_CURRENT_SOURCE_DIR}/nvidia_libdevice_bitcode/libdevice.${i}.10.bc")
 
     add_custom_command(OUTPUT "${INITMOD}"
-                       COMMAND binary2cpp "halide_internal_initmod_ptx_${i}_ll" < "$<SHELL_PATH:${RT_BC}>" > "${INITMOD}"
+                       COMMAND $<TARGET_FILE:binary2cpp> "halide_internal_initmod_ptx_${i}_ll" < "$<SHELL_PATH:${RT_BC}>" > "${INITMOD}"
```
alexreinking, line 336: Shouldn't be needed

src/runtime/CMakeLists.txt
```diff
@@ -333,22 +333,22 @@ foreach (i IN LISTS RUNTIME_BC)
...
                        DEPENDS binary2cpp "${RT_BC}"
                        VERBATIM)
     target_sources(Halide_initmod PRIVATE ${INITMOD})
 endforeach ()
 
 add_custom_command(OUTPUT "_initmod_inlined_c.cpp"
-                   COMMAND binary2cpp "halide_internal_initmod_inlined_c" < "$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/halide_buffer_t.cpp>" > "_initmod_inlined_c.cpp"
+                   COMMAND $<TARGET_FILE:binary2cpp> "halide_internal_initmod_inlined_c" < "$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/halide_buffer_t.cpp>" > "_initmod_inlined_c.cpp"
```
alexreinking, line 343: Shouldn't be needed

src/runtime/CMakeLists.txt
```diff
@@ -333,22 +333,22 @@ foreach (i IN LISTS RUNTIME_BC)
...
                    DEPENDS "halide_buffer_t.cpp" binary2cpp
                    VERBATIM)
 target_sources(Halide_initmod PRIVATE "_initmod_inlined_c.cpp")
 
 foreach (i IN LISTS RUNTIME_HEADER_FILES)
     string(REPLACE "." "_" SYM_NAME "${i}")
     add_custom_command(OUTPUT "_initmod_${SYM_NAME}.cpp"
-                       COMMAND binary2cpp "halide_internal_runtime_header_${SYM_NAME}" < "$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/${i}>" > "_initmod_${SYM_NAME}.cpp"
+                       COMMAND $<TARGET_FILE:binary2cpp> "halide_internal_runtime_header_${SYM_NAME}" < "$<SHELL_PATH:${CMAKE_CURRENT_SOURCE_DIR}/${i}>" > "_initmod_${SYM_NAME}.cpp"
```
alexreinking, line 351: Shouldn't be needed

test/performance/CMakeLists.txt
```diff
@@ -1,44 +1,49 @@
...
-      packed_planar_fusion.cpp
-      realize_overhead.cpp
-      rgb_interleaved.cpp
-      tiled_matmul.cpp
-      vectorize.cpp
-      wrap.cpp
-      )
+if (Halide_ANY_SANITIZERS_ENABLED)
+    # All sanitizers impact performance, so don't even bother with this test suite
+    message(STATUS "Skipping all performance testing because at least one Sanitizer is enabled.")
```
alexreinking, line 3: Just `return ()` rather than `else ()`

tutorial/CMakeLists.txt
```diff
@@ -84,7 +81,7 @@ if (TARGET_NVPTX)
 
         target_include_directories(lesson_10_aot_compilation_run PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
 
-        add_test(NAME tutorial_lesson_10_aot_compilation_run COMMAND lesson_10_aot_compilation_run)
+        add_test(NAME tutorial_lesson_10_aot_compilation_run COMMAND $<TARGET_FILE:lesson_10_aot_compilation_run>)
```
alexreinking, line 84: Shouldn't be needed

tutorial/CMakeLists.txt
```diff
@@ -177,7 +174,7 @@ else ()
                           Halide::ImageIO
                           Halide::Tools)
 
-    add_test(NAME tutorial_lesson_16_rgb_run COMMAND lesson_16_rgb_run)
+    add_test(NAME tutorial_lesson_16_rgb_run COMMAND $<TARGET_FILE:lesson_16_rgb_run>)
```
alexreinking, line 177: Shouldn't be needed

tutorial/CMakeLists.txt
```diff
@@ -207,6 +204,6 @@ if (TARGET Halide::Mullapudi2016)
     target_link_libraries(lesson_21_auto_scheduler_run PRIVATE
                           auto_schedule_false auto_schedule_true Halide::Tools)
 
-    add_test(NAME tutorial_lesson_21_auto_scheduler_run COMMAND lesson_21_auto_scheduler_run)
+    add_test(NAME tutorial_lesson_21_auto_scheduler_run COMMAND $<TARGET_FILE:lesson_21_auto_scheduler_run>)
```
alexreinking, line 207: Shouldn't be needed

alexreinking (changes requested):
See comments for immediately actionable feedback. I'm not sure about the plumbing for the sanitizer env-vars. Are those _exact_ settings _necessary_ or can we leave it up to the environment?

steven-johnson:
> Are those _exact_ settings _necessary_

It didn't work at all until I added those, so, apparently :-)

apps/CMakeLists.txt
steven-johnson: > All implementations that I'm aware of define __SANITIZE_ADDRESS__ iff ASAN is enabled.

That may be, but it's not the way Clang documents. See Util.h for that.

steven-johnson: > A third-party user might add the asan flags themselves without using our toolchain file.

Not sure I want to care about that much, because I don't think that approach is gonna work properly (unless they add their own toolchain to ensure everything uses the same compiler). If it was that easy, that's what I would have done in the first place :-)

src/CMakeLists.txt
steven-johnson: Alas, maybe it shouldn't be, but it still is :-/

cmake/toolchain.linux-x64-asan.cmake
steven-johnson: Interestingly, that will make things fail: apparently we execute this code again during the 'Detecting C compiler ABI info' phase, but without the value set, so we just have to hope for the best. (If it's not set, the compilation will fail anyway, of course)

cmake/toolchain.linux-x64-asan.cmake
steven-johnson: Done, but it didn't make any difference.

test/performance/CMakeLists.txt
steven-johnson: done

apps/CMakeLists.txt
alexreinking: I am actually concerned about divergence between CMake's and Halide's sources' view of whether or not ASAN is enabled. The following check should keep them synced:

```
include(CheckCXXSymbolExists)
check_cxx_symbol_exists(HALIDE_INTERNAL_USING_ASAN "${Halide_SOURCE_DIR}/src/Util.h" ASAN_ENABLED)
```

src/CMakeLists.txt
alexreinking: Are you sure about this? I just tested it locally with removing the `$<TARGET_FILE:...>` around `binary2cpp` and it worked fine. Maybe you changed the toolchain file without first deleting the build directory?

alexreinking: Here's my repro script:

```
#!/bin/bash

set -ex

rm -rf build
git diff
cmake -G Ninja -S . -B build --toolchain cmake/toolchain.linux-x64-asan.cmake -DLLVM_DIR=/usr/lib/llvm-15/lib/cmake/llvm -DHalide_SHARED_LLVM=YES -DCMAKE_BUILD_TYPE=Release
cmake --build build --target HalideIncludes --verbose
```

Here's its output, note the git diff:

```
(default) reinking@penguin:~/Halide$ ./repro.sh 
+ rm -rf build
+ git diff
diff --git a/cmake/toolchain.linux-x64-asan.cmake b/cmake/toolchain.linux-x64-asan.cmake
index 0c6c75120..c14ad3b14 100644
--- a/cmake/toolchain.linux-x64-asan.cmake
+++ b/cmake/toolchain.linux-x64-asan.cmake
@@ -34,7 +34,7 @@ set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
 
 # Set to empty string to indicate the resulting binaries can be natively executed
-set(CMAKE_CROSSCOMPILING_EMULATOR)
+set(CMAKE_CROSSCOMPILING_EMULATOR /usr/bin/env)
 
 # A few downstream build rules need to be able to sniff these:
 # - ASAN means we can't run with large stacks, so some things will overflow
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index dd86add7a..c31e37c32 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -360,7 +360,7 @@ set(HALIDE_H "${Halide_BINARY_DIR}/include/Halide.h")
 set(LICENSE_PATH "${Halide_SOURCE_DIR}/LICENSE.txt")
 add_custom_command(OUTPUT "${Halide_BINARY_DIR}/include/Halide.h"
                    COMMAND ${CMAKE_COMMAND} -E make_directory "$<SHELL_PATH:${Halide_BINARY_DIR}/include>"
-                   COMMAND $<TARGET_FILE:build_halide_h> "$<SHELL_PATH:${LICENSE_PATH}>" ${HEADER_FILES} > "$<SHELL_PATH:${HALIDE_H}>"
+                   COMMAND build_halide_h "$<SHELL_PATH:${LICENSE_PATH}>" ${HEADER_FILES} > "$<SHELL_PATH:${HALIDE_H}>"
                    DEPENDS build_halide_h "${LICENSE_PATH}" ${HEADER_FILES}
                    WORKING_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}"
                    VERBATIM)
+ cmake -G Ninja -S . -B build --toolchain cmake/toolchain.linux-x64-asan.cmake -DLLVM_DIR=/usr/lib/llvm-15/lib/cmake/llvm -DHalide_SHARED_LLVM=YES -DCMAKE_BUILD_TYPE=Release
-- The C compiler identification is Clang 15.0.0
-- The CXX compiler identification is Clang 15.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib/llvm-15/lib/cmake/llvm/../../..//bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib/llvm-15/lib/cmake/llvm/../../..//bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Halide 'host' platform triple:   x86-64-linux
-- Halide 'cmake' platform triple:  x86-64-linux
-- Halide default AOT target:       host
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Could NOT find OpenGL (missing: OPENGL_opengl_LIBRARY OPENGL_glx_LIBRARY OPENGL_INCLUDE_DIR) 
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
-- Performing Test HAVE_FFI_CALL
-- Performing Test HAVE_FFI_CALL - Success
-- Found FFI: /usr/lib/x86_64-linux-gnu/libffi.so  
-- Performing Test Terminfo_LINKABLE
-- Performing Test Terminfo_LINKABLE - Success
-- Found Terminfo: /usr/lib/x86_64-linux-gnu/libtinfo.so  
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11") 
-- Found LibXml2: /usr/lib/x86_64-linux-gnu/libxml2.so (found version "2.9.10") 
-- Found LLVM 15.0.0
-- Using LLVMConfig.cmake in: /usr/lib/llvm-15/lib/cmake/llvm
-- Using ClangConfig.cmake in: /usr/lib/llvm-15/lib/cmake/clang
-- Enabling AArch64 backend
-- Enabling AMDGPU backend
-- Enabling ARM backend
-- Enabling Hexagon backend
-- Enabling Mips backend
-- Enabling NVPTX backend
-- Enabling PowerPC backend
-- Enabling RISCV backend
-- Enabling WebAssembly backend
-- Enabling X86 backend
-- Using LLDConfig.cmake in: /usr/lib/llvm-15/lib/cmake/lld
-- Compiling Halide WITH RTTI.
-- Compiling Halide WITH exceptions.
-- Found JPEG: /usr/lib/x86_64-linux-gnu/libjpeg.so (found version "62") 
-- Found PNG: /usr/lib/x86_64-linux-gnu/libpng.so (found version "1.6.37") 
-- Fetching WABT 1.0.29...
-- Looking for alloca.h
-- Looking for alloca.h - found
-- Looking for unistd.h
-- Looking for unistd.h - found
-- Looking for snprintf
-- Looking for snprintf - found
-- Looking for strcasecmp
-- Looking for strcasecmp - found
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of ssize_t
-- Check size of ssize_t - done
-- Check size of size_t
-- Check size of size_t - done
CMake Warning at build/_deps/wabt-src/CMakeLists.txt:214 (message):
  Error compiling dummy.c file


-- Found PythonInterp: /home/reinking/.venv/default/bin/python3 (found suitable version "3.9.2", minimum required is "3.5") 
-- Performing Test LINKER_HAS_FLAG_VERSION_SCRIPT
-- Performing Test LINKER_HAS_FLAG_VERSION_SCRIPT - Success
-- Building autoschedulers enabled
-- Building tests enabled
-- Skipping all performance testing because at least one Sanitizer is enabled.
-- Building internal runtime tests enabled
-- Building Python bindings enabled
-- Found Python3: /home/reinking/.venv/default/bin/python3.9 (found version "3.9.2") found components: Interpreter Development Development.Module Development.Embed 
-- pybind11 v2.6.2 
CMake Warning (dev) at /home/reinking/.venv/default/lib/python3.9/site-packages/cmake/data/share/cmake-3.23/Modules/CMakeDependentOption.cmake:89 (message):
  Policy CMP0127 is not set: cmake_dependent_option() supports full Condition
  Syntax.  Run "cmake --help-policy CMP0127" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
  build/_deps/pybind11-src/CMakeLists.txt:98 (cmake_dependent_option)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Performing Test HAS_FLTO_THIN
-- Performing Test HAS_FLTO_THIN - Success
CMake Warning at python_bindings/CMakeLists.txt:63 (message):
  Python tests/tutorials are not yet supported under ASAN.


-- Building tutorials enabled
-- Building docs disabled
-- Building utils enabled
-- Configuring done
-- Generating done
-- Build files have been written to: /home/reinking/Halide/build
+ cmake --build build --target HalideIncludes --verbose
[1/3] /usr/lib/llvm-15/lib/cmake/llvm/../../..//bin/clang++   -fsanitize=address -O3 -DNDEBUG -std=c++17 -MD -MT tools/CMakeFiles/build_halide_h.dir/build_halide_h.cpp.o -MF tools/CMakeFiles/build_halide_h.dir/build_halide_h.cpp.o.d -o tools/CMakeFiles/build_halide_h.dir/build_halide_h.cpp.o -c /home/reinking/Halide/tools/build_halide_h.cpp
[2/3] : && /usr/lib/llvm-15/lib/cmake/llvm/../../..//bin/clang++ -fsanitize=address -O3 -DNDEBUG -fsanitize=address tools/CMakeFiles/build_halide_h.dir/build_halide_h.cpp.o -o tools/build_halide_h   && :
[3/3] cd /home/reinking/Halide/src && /home/reinking/.venv/default/lib/python3.9/site-packages/cmake/data/bin/cmake -E make_directory /home/reinking/Halide/build/include && /usr/bin/env /home/reinking/Halide/build/tools/build_halide_h /home/reinking/Halide/LICENSE.txt AbstractGenerator.h AddAtomicMutex.h AddImageChecks.h AddParameterChecks.h AlignLoads.h AllocationBoundsInference.h ApplySplit.h Argument.h AssociativeOpsTable.h Associativity.h AsyncProducers.h AutoScheduleUtils.h BoundaryConditions.h Bounds.h BoundsInference.h BoundSmallAllocations.h Buffer.h Callable.h CanonicalizeGPUVars.h ClampUnsafeAccesses.h Closure.h CodeGen_C.h CodeGen_D3D12Compute_Dev.h CodeGen_GPU_Dev.h CodeGen_Internal.h CodeGen_LLVM.h CodeGen_Metal_Dev.h CodeGen_OpenCL_Dev.h CodeGen_OpenGLCompute_Dev.h CodeGen_Posix.h CodeGen_PTX_Dev.h CodeGen_PyTorch.h CodeGen_Targets.h CompilerLogger.h ConciseCasts.h CPlusPlusMangle.h CSE.h Debug.h DebugArguments.h DebugToFile.h Definition.h Deinterleave.h Derivative.h DerivativeUtils.h DeviceAPI.h DeviceArgument.h DeviceInterface.h Dimension.h EarlyFree.h Elf.h EliminateBoolVectors.h EmulateFloat16Math.h Error.h Expr.h ExprUsesVar.h Extern.h ExternFuncArgument.h ExtractTileOperations.h FastIntegerDivide.h FindCalls.h FindIntrinsics.h FlattenNestedRamps.h Float16.h Func.h Function.h FunctionPtr.h FuseGPUThreadLoops.h FuzzFloatStores.h Generator.h HexagonOffload.h HexagonOptimize.h ImageParam.h InferArguments.h InjectHostDevBufferCopies.h Inline.h InlineReductions.h IntegerDivisionTable.h Interval.h Introspection.h IntrusivePtr.h IR.h IREquality.h IRMatch.h IRMutator.h IROperator.h IRPrinter.h IRVisitor.h JITModule.h Lambda.h Lerp.h LICM.h LLVM_Output.h LLVM_Runtime_Linker.h LoopCarry.h Lower.h LowerParallelTasks.h LowerWarpShuffles.h MainPage.h Memoization.h Module.h ModulusRemainder.h Monotonic.h ObjectInstanceRegistry.h OffloadGPULoops.h OutputImageParam.h ParallelRVar.h Param.h Parameter.h ParamMap.h PartitionLoops.h Pipeline.h Prefetch.h Profiling.h PurifyIndexMath.h PythonExtensionGen.h Qualify.h Random.h RDom.h Realization.h RealizationOrder.h RebaseLoopsToZero.h Reduction.h RegionCosts.h RemoveDeadAllocations.h RemoveExternLoops.h RemoveUndef.h runtime/HalideBuffer.h runtime/HalideRuntime.h Schedule.h ScheduleFunctions.h Scope.h SelectGPUAPI.h Simplify.h SimplifyCorrelatedDifferences.h SimplifySpecializations.h SkipStages.h SlidingWindow.h Solve.h SplitTuples.h StmtToHtml.h StorageFlattening.h StorageFolding.h StrictifyFloat.h Substitute.h Target.h ThreadPool.h Tracing.h TrimNoOps.h Tuple.h Type.h UnifyDuplicateLets.h UniquifyVariableNames.h UnpackBuffers.h UnrollLoops.h UnsafePromises.h Util.h Var.h VectorizeLoops.h WasmExecutor.h WrapCalls.h > /home/reinking/Halide/build/include/Halide.h
```

cmake/toolchain.linux-x64-asan.cmake
alexreinking: Sorry, but can you show me repro steps that contradict those I gave above?

alexreinking:
I have moved the sanitizer environment variables to presets for a few reasons:

1. They were complicating our build rules considerably and necessitated hacks in a few places.
2. As environment variables, they can be set externally.
3. The build itself succeeds without those variables set.
4. We should probably know if the build _starts_ to fail without those set.

Using the preset is easy and should be how we implement this in the buildbots:

```
$ cmake --preset linux-x64-asan
$ cmake --build --preset linux-x64-asan
$ ctest --preset linux-x64-asan
```

This sequence of commands works for me on Google infra.

steven-johnson:
LGTM

alexreinking:
With the newest commit, the apps have a corresponding preset. From the apps directory, run:

```
$ cmake --preset linux-x64-asan -DLLVM_ROOT=/path/to/llvm -DHalide_ROOT=/path/to/halide-install
$ cmake --build --preset linux-x64-asan
$ ctest --preset linux-x64-asan
```

I note that the apps require the ASAN options set during the build, too. This is also working for me on gLinux.

steven-johnson:
> $ cmake --build --preset linux-x64-asan

grumble, it really doesn't matter, but I am grumpy about not being able to just use `ninja` directly here

steven-johnson:
> $ cmake --preset linux-x64-asan -DLLVM_ROOT=/path/to/llvm

Why do we need to set LLVM_ROOT here? The buildbots have set only LLVM_DIR for ~ever.

alexreinking:
> I am grumpy about not being able to just use `ninja` directly here

For compiling Halide, which doesn't require any env vars to be set, you can just call `ninja`. For the apps, you'd need to `ASAN_OPTIONS` some other way (have you ever used [direnv](https://direnv.net/)?)

steven-johnson:
> (have you ever used [direnv](https://direnv.net/)?)

Yikes, no, magic hidden things that set env vars based on my location is (to me) an indication of a system gone wrong

alexreinking:
> Yikes, no, magic hidden things that set env vars based on my location is (to me) an indication of a system gone wrong

Eh, I like it for setting some common options locally, but to each their own.

steven-johnson:
> Why do we need to set LLVM_ROOT here? The buildbots have set only LLVM_DIR for ~ever.

To be more on point: it seems suboptimal to have to set two LLVM related config vars here (esp when one can generally be inferred from the other); since LLVM_DIR is already set, is there a compelling reason we can't assume LLVM_ROOT is something like ${LLVM_DIR}/../../.. ?

steven-johnson:
> cmake --build --preset linux-x64-asan

On gLinux, this fails for me with `CMake Error: Could not read presets from /usr/local/google/home/srj/GitHub/Halide/build-asan: File not found`

alexreinking:
> is there a compelling reason we can't assume LLVM_ROOT is something like ${LLVM_DIR}/../../.. ?

On Ubuntu, there are two copies of `LLVMConfig.cmake`, one in `LLVM_ROOT/cmake` and another in `LLVM_ROOT/lib/cmake/llvm`. You don't need to set `LLVM_DIR` if you set `LLVM_ROOT`.

alexreinking:
> On gLinux, this fails for me with `CMake Error: Could not read presets from /usr/local/google/home/srj/GitHub/Halide/build-asan: File not found`

Don't change directory, run all three commands from the root Halide dir.

steven-johnson:
LGTM

alexreinking (approved):

alexreinking:
Windows failure is unrelated

PR #6921, opened 2022-08-04
Fix wrong install path for *.py files

steven-johnson:
We were looking in a nonexistent dir, so we never copied `__init__.py` as we should have.

### DIFF
--- a/python_bindings/src/halide/CMakeLists.txt
+++ b/python_bindings/src/halide/CMakeLists.txt
@@ -50,9 +50,6 @@ if (WIN32 AND BUILD_SHARED_LIBS)
                        VERBATIM)
 endif ()
 
-# HALIDE_PYTHON_PATH is what you need to add to PYTHONPATH to use our bindings
-set(HALIDE_PYTHON_PATH "${HALIDE_PYTHON_PATH}")
-
 ##
 # Packaging
 ##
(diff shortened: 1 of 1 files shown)


python_bindings/src/halide/CMakeLists.txt
```diff
@@ -50,6 +50,9 @@ if (WIN32 AND BUILD_SHARED_LIBS)
                        VERBATIM)
 endif ()
 
+# HALIDE_PYTHON_PATH is what you need to add to PYTHONPATH to use our bindings
+set(HALIDE_PYTHON_PATH "${HALIDE_PYTHON_PATH}")
+
```
alexreinking, line 55: This doesn't do anything?

python_bindings/src/halide/CMakeLists.txt
```diff
@@ -60,7 +63,7 @@ include(GNUInstallDirs)
 set(Halide_INSTALL_PYTHONDIR "${CMAKE_INSTALL_LIBDIR}/python3/site-packages"
     CACHE STRING "Path to the Python site-packages folder")
 
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/halide"
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
```
alexreinking, line 66: Whoops! 🤦

python_bindings/src/halide/CMakeLists.txt
steven-johnson: oops, scalpel left in patient, removing

alexreinking (approved):

PR #6922, opened 2022-08-04
Rework internal PYTHONPATH maintenance

### LINKED ISSUE #6870: Improve __init__.py import logic
In #6867 the `halide` Python module is split into a pure-Python parent package and a native-code sub-package. During the build, however, the native code exists in the build tree while the pure-Python code stays in the source tree.

The current hack is to edit `sys.path`, but then this code winds up in the distributed package. Not great.

The solution appears to be to create a test launcher script that adds a custom _[finder](https://docs.python.org/3/reference/import.html#finders-and-loaders)_ to the Python import system that knows where the native code module is.

steven-johnson:
Not sure if this is the right way to do this, but: Currently, we have to set PYTHONPATH to include both `halide_.so` and `__init__.py` separately; this is ugly, and will get uglier once #6764 lands. This attempts to remedy this by copying `__init__.py` into the build folder in the same place as `halide_.so`, then creates a fake target to use to point at the containing folder.

The use of a fake imported target to get this directory is kinda squicky, though.

Also, @alexreinking said this approach might not work reliably for build systems with mutliple configs (eg MSVC). Opening as draft for comment.

Fixes #6870


steven-johnson:
Lots of OSX failures :-/  e.g.

https://buildbot.halide-lang.org/master/#/builders/86/builds/57/steps/12/logs/python_tutorial_lesson_10_aot_compilation_generate

steven-johnson:
https://buildbot.halide-lang.org/master/#/builders/42/builds/47/steps/13/logs/python_tutorial_lesson_10_aot_compilation_generate

> ImportError: libHalide.so.14: cannot open shared object file: No such file or directory

... why is it looking for Halide 14 when we are building trunk with LLVM 16?

steven-johnson:
I took the liberty of pushing a change that I think will fix at least some of the breakage. It's still ugly. PTAL.

steven-johnson:
Bah -- that fixed Posixy systems but broke Windows.

steven-johnson:
Finally clean, but unsure of hackiness. PTAL.

steven-johnson:
LGTM once green

python_bindings/CMakeLists.txt
```diff
@@ -49,11 +49,40 @@ if (NOT Halide_ENABLE_RTTI OR NOT Halide_ENABLE_EXCEPTIONS)
...
 endif ()
 
+##
+# A helper for creating tests with correct PYTHONPATH
+##
+
+function(add_python_test)
+    cmake_parse_arguments(ARG "" "FILE;LABEL" "PYTHONPATH;ENVIRONMENT" ${ARGN})
+
+    list(PREPEND ARG_PYTHONPATH "$<TARGET_FILE_DIR:Halide::Python>/..")
```
steven-johnson, line 59: We'll also need some similar logic (in the `pygen` branch) to construct the proper PYTHONPATH for use when driving Python generators (which aren't tests), but we can refactor that after this merges.

alexreinking (approved):

PR #6923, opened 2022-08-05
Minimal approach to making Deinterleave correct for Reinterpret

abadams:
A minimal baseline for the purpose of discussion and comparison with #6887

### DIFF
--- a/src/Deinterleave.cpp
+++ b/src/Deinterleave.cpp
@@ -275,6 +275,16 @@ private:
         return expr;
     }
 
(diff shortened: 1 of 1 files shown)


rootjalex (approved):
Lgtm in principle, but would it be possible to add a test for the specific case of calling deinterleave on a reinterpret where to.bits() != from.bits()?

abadams:
Good point. Will do.

abadams:
Actually I don't think I can trigger that without #6928 , and without this, that PR is slightly busted. I'll just combine the two.

PR #6924, opened 2022-08-05
[vulkan phase2] Vulkan Runtime  

derek-gerstmann:
This set of changes is Phase2 for adding Vulkan support, which adds the runtime class interfaces for interacting with the Vulkan API. Specifically, it adds the following files:

```
src/CodeGen_Vulkan_Dev.cpp
src/runtime/HalideRuntimeVulkan.h
src/runtime/mini_vulkan.h
src/runtime/vulkan.cpp
src/runtime/vulkan_context.h
src/runtime/vulkan_extensions.h
src/runtime/vulkan_functions.h
src/runtime/vulkan_interface.h
src/runtime/vulkan_internal.h
src/runtime/vulkan_memory.h
src/runtime/vulkan_resources.h
src/runtime/windows_vulkan.cpp
```

The forthcoming Phase3 will address testing and CI modifications.


steven-johnson:
C:\build_bot\worker\halide-testbranch-main-llvm16-x86-64-windows-cmake\halide-source\src\CodeGen_Vulkan_Dev.cpp(1082): error C2398: Element '3': conversion from 'SpvBuiltIn_' to '_Ty' requires a narrowing conversion
        with
        [
            _Ty=uint32_t
        ]

Makefile
```diff
@@ -9,6 +9,12 @@
 #     For correctness and performance tests this include halide build time and run time. For
 #     the tests in test/generator/ this times only the halide build time.
 
+# Halide project version
+HALIDE_VERSION_MAJOR ?= 15
```
steven-johnson, line 13: Per earlier discussions elsewhere, you shouldn't be under any obligation to add Vulkan support to Make (CMake-only is fine)

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,1304 @@
...
+        void visit(const Prefetch *) override;
+        void visit(const Fork *) override;
+        void visit(const Acquire *) override;
+
+        void visit_binop(Type t, const Expr &a, const Expr &b, uint32_t opcode);
+
+        // ID of last generated Expr.
+        uint32_t id;
+        // IDs are allocated in numerical order of use.
+        uint32_t next_id{0};
```
steven-johnson, line 117: nit: we don't configure clang-tidy to enforce this, but I generally we prefer `next_id = 0;` style for in-class-member-init, rather than `next_id{0};`

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,1304 @@
...
+    } emitter;
+
+    std::string current_kernel_name;
+};
+
+// --
+
+void CodeGen_Vulkan_Dev::SPIRVEmitter::add_instruction(std::vector<uint32_t> &region, uint32_t opcode,
+                                                       std::initializer_list<uint32_t> words) {
+    region.push_back(((1 + words.size()) << 16) | opcode);
```
steven-johnson, line 186: > ((1 + words.size()) << 16) | opcode

nit: This formula is repeated several times and seems like it could be error-prone, probably worth an inline wrapper?

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,1304 @@
...
+    region.insert(region.end(), words.begin(), words.end());
+}
+
+void CodeGen_Vulkan_Dev::SPIRVEmitter::add_instruction(uint32_t opcode, std::vector<uint32_t> words) {
+    spir_v_kernels.push_back(((1 + words.size()) << 16) | opcode);
+    spir_v_kernels.insert(spir_v_kernels.end(), words.begin(), words.end());
+}
+
+uint32_t CodeGen_Vulkan_Dev::SPIRVEmitter::emit_constant(const Type &t, const void *data) {
+    // TODO: this needs to emit OpConstantComposite for constants with lane > 1
```
steven-johnson, line 206: (Here and any/all remaining TODO comments:) Does this need a tracking issue? (If already planned to be fixed in subsequent phase, no; if open-ended someday-fix, then probably yes)

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,1304 @@
...
+    uint32_t binding_counter = 1;
+    uint32_t scalar_index = 0;
+    for (const DeviceArgument &arg : args) {
+        uint32_t param_id = next_id++;
+        if (arg.is_buffer) {
+            uint32_t element_type = map_type(arg.type);
+            uint32_t runtime_arr_type = next_id++;
+            uint32_t struct_type = next_id++;
+            uint32_t ptr_struct_type = next_id++;
+            add_instruction(spir_v_types, SpvOpTypeRuntimeArray, {runtime_arr_type, element_type});
```
steven-johnson, line 1133: Idle thought: would it be clearer code if you added helper methods like `add_type()`, `add_annotation()` etc which  were shorthand for calling `add_instruction()` with the relevant vector as the first arg?

src/runtime/vulkan.cpp
```diff
@@ -0,0 +1,847 @@
...
+                                       bool create) {
+
+    halide_abort_if_false(user_context, instance != nullptr);
+    halide_abort_if_false(user_context, device != nullptr);
+    halide_abort_if_false(user_context, queue != nullptr);
+    halide_abort_if_false(user_context, &thread_lock != nullptr);
+    while (__atomic_test_and_set(&thread_lock, __ATOMIC_ACQUIRE)) {}
+
+    // If the context has not been initialized, initialize it now.
+    halide_abort_if_false(user_context, &cached_instance != nullptr);
```
steven-johnson, line 49: ...how is it possible for these to be null? This seems genuinely impossible.

src/runtime/vulkan.cpp
```diff
@@ -0,0 +1,847 @@
...
+    VulkanContext ctx(user_context);
+    if (ctx.error != VK_SUCCESS) {
+        return ctx.error;
+    }
+
+#ifdef DEBUG_RUNTIME
+    uint64_t t_before = halide_current_time_ns(user_context);
+#endif
+
+    // Running a Vulkan pipeline requires a large number of steps
```
steven-johnson, line 593: Thanks for this comment, it's helpful :-)

src/runtime/vulkan_context.h
```diff
@@ -0,0 +1,329 @@
...
+          command_pool(0),
+          physical_device(nullptr),
+          queue(nullptr),
+          queue_family_index(0),
+          error(VK_SUCCESS) {
+
+        int result = halide_vulkan_acquire_context(user_context,
+                                                   reinterpret_cast<halide_vulkan_memory_allocator **>(&allocator),
+                                                   &instance, &device, &physical_device, &command_pool, &queue, &queue_family_index);
+        halide_abort_if_false(user_context, result == 0);
```
steven-johnson, line 62: (1) is `halide_abort_if_false()` really the best way to deal with errors here? (ie, is it unrecoverable and `abort()` the only option?) 
(2) do we really need to check all the values being output (in addition to the function result)? for a debug-runtime-only build I'd be fine with this, but this feels like overkill for a normal runtime.

src/runtime/vulkan_extensions.h
```diff
@@ -0,0 +1,298 @@
...
+WEAK ScopedSpinLock::AtomicFlag build_options_lock = 0;
+WEAK bool build_options_initialized = false;
+
+// --------------------------------------------------------------------------
+namespace {
+
+void vk_set_layer_names_internal(const char *n) {
+    if (n) {
+        size_t buffer_size = sizeof(layer_names) / sizeof(layer_names[0]);
+        strncpy(layer_names, n, buffer_size);
```
steven-johnson, line 37: ...I know that our opencl runtime already uses strncpy, but we should probably avoid adding new uses of it, since it requires the "add a null terminator just in case" nonsense that is easy to forget.

src/runtime/vulkan_memory.h
```diff
@@ -0,0 +1,721 @@
...
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+namespace Vulkan {
+
+// --------------------------------------------------------------------------
+
+// Enable external client to override Vulkan allocation callbacks (if they so desire)
+WEAK ScopedSpinLock::AtomicFlag custom_allocation_callbacks_lock = 0;
+static const VkAllocationCallbacks *custom_allocation_callbacks = nullptr;  // nullptr => use Vulkan runtime implementation
```
steven-johnson, line 16: It's generally not a good idea to declare something as `static` in a .h file, since you'll end up with a separate static instance in every .cpp file you include this in -- if that is actually what you *want* here you really need to add comments here as to why that is desirable and intentional (but even then it seems dubious0

src/runtime/vulkan_memory.h
```diff
@@ -0,0 +1,721 @@
...
+    static void deallocate_region(void *user_context, MemoryRegion *region);
+
+    size_t bytes_allocated_for_blocks() const;
+    size_t blocks_allocated() const;
+
+    size_t bytes_allocated_for_regions() const;
+    size_t regions_allocated() const;
+
+private:
+    static const uint32_t invalid_usage_flags = uint32_t(-1);
```
steven-johnson, line 89: nit: could probably be `static constexpr`

steven-johnson (changes requested):
Generally looks very nice; my comments are mostly nits but a few actionable items. My main concern here is that there are a lot of TODO comments (and also some `if 0` unimplemented sections) in Codegen_Dev, and it's not clear to me which/how-many of them are "need to complete before this will work" vs "this should be improved someday but isn't blocking a functional initial release" -- clarifying those would IMHO make this easier for other folks to read and grasp the current state.

derek-gerstmann:
> C:\build_bot\worker\halide-testbranch-main-llvm16-x86-64-windows-cmake\halide-source\src\CodeGen_Vulkan_Dev.cpp(1082): error C2398: Element '3': conversion from 'SpvBuiltIn_' to '_Ty' requires a narrowing conversion with [ _Ty=uint32_t ]

Oooh, thanks!  I'll fix this!

Makefile
derek-gerstmann: Ah, okay!  It wasn't clear if we were maintaining both ...

derek-gerstmann:
> Generally looks very nice; my comments are mostly nits but a few actionable items. My main concern here is that there are a lot of TODO comments (and also some `if 0` unimplemented sections) in Codegen_Dev, and it's not clear to me which/how-many of them are "need to complete before this will work" vs "this should be improved someday but isn't blocking a functional initial release" -- clarifying those would IMHO make this easier for other folks to read and grasp the current state.

Thanks!  Okay, cool! Yes, the CodeGen_Dev is in the process of being refactored to use the SPIRV_IR classes and to address most of the TODO comments.  Once this is in place, I'll update the docs and status for what's working, etc!

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Agreed ... this will be going away.

src/runtime/mini_vulkan.h
```diff
@@ -0,0 +1,6038 @@
+#ifndef HALIDE_MINI_VULKAN_H
+#define HALIDE_MINI_VULKAN_H
+
+/*
+** Copyright (c) 2014-2017 The Khronos Group Inc.
```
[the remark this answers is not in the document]
derek-gerstmann: Ah, good catch!  I'll add this!!

src/runtime/vulkan_context.h
derek-gerstmann: Agreed ... this was just a quick check that I added for testing ... it definitely needs to be more robust. I'll fix this.

steven-johnson:

derek-gerstmann:
> Monday Morning Review Ping -- where does this PR stand?

Apologies … I’m on vacation but will be back later this week! Most of the requested changes are pushed … remaining items are on the codegen, and docs!

src/runtime/vulkan_functions.h
```diff
@@ -0,0 +1,41 @@
+// VULKAN_FN(vkCreateInstance)
```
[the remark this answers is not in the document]
derek-gerstmann: It turns out the the Vulkan loader library exports this directly, so don't need to resolve the function pointer.  I've added a comment to clarify this and mention its declaration is in mini_vulkan.h

derek-gerstmann:
Ready for review!

steven-johnson:
Linux failures appear to be real, PTAL

derek-gerstmann:
> Linux failures appear to be real, PTAL

Strange. The arm-64-linux failure seems to come from a typo in tests/correctness/target.cpp which isn't part of my changes.  I'll merge with main again.

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,1758 @@
...
+    // Declare the kernel function
+    SpvId void_type_id = builder.declare_void_type();
+    SpvId kernel_func_id = builder.add_function(name, void_type_id);
+    SpvFunction kernel_func = builder.lookup_function(kernel_func_id);
+    builder.enter_function(kernel_func);
+
+    // TODO: only add the SIMT intrinsics used
+    SpvFactory::Variables entry_point_variables;
+    auto intrinsics = {"WorkgroupId", "LocalInvocationId"};
+    for (const std::string &intrinsic_name : intrinsics) {
```
shoaibkamil, line 1525: Nit: `intrinsics` is `const char*[]`, so this does an implicit copy.  Probably better to either make `intrinsics` `std::string []` or to use `const char* const&` in the loop

README_vulkan.md
```diff
@@ -0,0 +1,569 @@
...
+
+For NVIDIA:
+```
+$ sudo add-apt-repository ppa:graphics-drivers/ppa
+$ sudo apt update
+$ sudo apt upgrade
+# - replace ### with latest driver release (eg 515)
+$ sudo apt install nvidia-driver-### nvidia-settings vulkan vulkan-utils
+```
+
```
shoaibkamil, line 72: For Ubuntu 22.04, these steps didn't quite work for me.  The last step I needed to do was:

```
sudo apt install nvidia-driver-515 nvidia-settings libvulkan vulkan-tools
```

After that, I can run tutorials or a test with `HL_JIT_TARGET=host-vulkan` successfully.  However, building with `-DHalide_TARGET=host-vulkan` doesn't work due to link errors:
```
[3094/3631] Linking CXX executable test/generator/generator_aot_can_use_target
FAILED: test/generator/generator_aot_can_use_target
: && /usr/bin/c++   test/generator/CMakeFiles/generator_aot_can_use_target.dir/can_use_target_aottest.cpp.o -o test/generator/generator_aot_can_use_target  test/generator/libcan_use_target.a  test/generator/libcan_use_target.runtime.a  -ldl && :
/usr/bin/ld: test/generator/libcan_use_target.runtime.a(can_use_target.runtime.o): in function `halide_vulkan_acquire_context':
posix_allocator.cpp:(.text.halide_vulkan_acquire_context+0x1d8): undefined reference to `vkGetInstanceProcAddr'
/usr/bin/ld: posix_allocator.cpp:(.text.halide_vulkan_acquire_context+0x520): undefined reference to `vkCreateInstance'
/usr/bin/ld: posix_allocator.cpp:(.text.halide_vulkan_acquire_context+0x103a): undefined reference to `vkGetInstanceProcAddr'
/usr/bin/ld: posix_allocator.cpp:(.text.halide_vulkan_acquire_context+0x104e): undefined reference to `vkGetInstanceProcAddr'
/usr/bin/ld: test/generator/libcan_use_target.runtime.a(can_use_target.runtime.o): in function `Halide::Runtime::Internal::Vulkan::vk_load_vulkan_functions(VkInstance_T*)':
posix_allocator.cpp:(.text._ZN6Halide7Runtime8Internal6Vulkan24vk_load_vulkan_functionsEP12VkInstance_T+0x11): undefined reference to `vkGetInstanceProcAddr'
/usr/bin/ld: posix_allocator.cpp:(.text._ZN6Halide7Runtime8Internal6Vulkan24vk_load_vulkan_functionsEP12VkInstance_T+0x2a): undefined reference to `vkGetInstanceProcAddr'
/usr/bin/ld: posix_allocator.cpp:(.text._ZN6Halide7Runtime8Internal6Vulkan24vk_load_vulkan_functionsEP12VkInstance_T+0x43): undefined reference to `vkGetInstanceProcAddr'
/usr/bin/ld: test/generator/libcan_use_target.runtime.a(can_use_target.runtime.o):posix_allocator.cpp:(.text._ZN6Halide7Runtime8Internal6Vulkan24vk_load_vulkan_functionsEP12VkInstance_T+0x5c): more undefined references to `vkGetInstanceProcAddr' follow
collect2: error: ld returned 1 exit status
```

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,1758 @@
...
+        void visit(const Not *) override;
+        void visit(const Select *) override;
+        void visit(const Load *) override;
+        void visit(const Ramp *) override;
+        void visit(const Broadcast *) override;
+        void visit(const Call *) override;
+        void visit(const Let *) override;
+        void visit(const LetStmt *) override;
+        void visit(const AssertStmt *) override;
+        //        void visit(const ProducerConsumer *) override;
```
shoaibkamil, line 103: Delete?

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Nice catch!  I'll add this to my next batch of fixes.

README_vulkan.md
derek-gerstmann: Ahhh ... thanks!  I'll add vulkan-tools to the docs!

Hrmm ... I'll look into the linker issues ... looks like the vulkan loader isn't getting added to the linkage for AOT generators. =/

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Yep.  Not needed.

src/SpirvIR.cpp
```diff
@@ -596,120 +703,343 @@ void SpvModule::encode(SpvBinary &binary) const {
...
     if (type_id == SpvInvalidId) {
-        type_id = declare_function_type(return_type, param_types);
+        type_id = add_function_type(return_type, param_types);
     }
     return type_id;
 }
 
-SpvId SpvBuilder::map_constant(const Type &type, const void *data) {
+SpvId SpvBuilder::declare_function(const std::string &name, SpvId function_type) {
+    SpvId existing_id = lookup_id(name);
```
shoaibkamil, line 1027: Can you add a comment as to why re-declaring a function isn't an error?  I'm a bit confused why we need this logic here.

derek-gerstmann: Sure thing ... I'll add doc-strings for the builder class.  Yeah, these declaration methods are really convenience methods for "if this thing isn't defined already, then define it, otherwise give me its existing id", which avoids making the caller do all these checks, and avoids inadvertently creating duplicates.

src/SpirvIR.cpp
```diff
@@ -1478,18 +2026,47 @@ SpvInstruction SpvFactory::exec_mode_local_size(SpvId function_id, uint32_t wg_s
...
     return unary_op(SpvOpNot, type_id, result_id, src_id);
 }
 
+SpvInstruction SpvFactory::logical_not(SpvId type_id, SpvId result_id, SpvId src_id) {
+    return unary_op(SpvOpLogicalNot, type_id, result_id, src_id);
+}
+
+SpvInstruction SpvFactory::shift_right_logical(SpvId type_id, SpvId result_id, SpvId src_id, SpvId shift_id) {
+    SpvInstruction inst = SpvInstruction::make(SpvOpShiftRightLogical);
+    inst.add_operand(type_id);
```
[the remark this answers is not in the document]
derek-gerstmann: Good point ... that'd certainly clean up some of the redundancy in the factory methods for the ops!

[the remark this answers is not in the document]
derek-gerstmann: Added!

src/SpirvIR.cpp
derek-gerstmann: Added!

derek-gerstmann:
Ready to merge! All tests are now passing on my tested configs for Linux & Windows using the target `host-vulkan-vk_int8-vk_int16-vk_int64-vk_float16-vk_float64-vk_v13` on LLVM v14.x. 

More details in the VULKAN_readme.md regarding current limitations and caveats.

steven-johnson (changes requested):
Generally looks great! Lots of nits (not surprising given the size, of course), and a few substantive things pointed out (e.g. a possible div-by-zero in Vulkan_Dev), but there are two general issues I want to reiterate here:

- I think we need to audit the uses of `halide_abort_if_false()` and reality-check which ones are actually necessary. IMHO the use of that call is far too generous in the rest of our runtime: the only reason we should ever call `abort()` in our runtime is if we encounter an **unrecoverable** error, which could compromise the integrity/safety/security of the host app. (In most cases we should simply be returning a runtime error.) My suspicion is that that unrecoverable cases are rare enough that the `halide_abort_if_false()` macro wouldn't even pay for itself anymore (ie, it would be explicit calls to abort()).

- Relying on env vars to configure the settings. This is (IMHO) a code small that we have relied on far too long; it is admittedly super-handy for command-line experimentation, but it's not at all a good fit for embedding in (say) a mobile app, since env vars are basically global variables that may be set to values you don't control when your app is launched... so even if you are fine with the "default" values, so still need to explicitly reset those default values to be sure a bad actor hasn't set them to something weird or evil. If we really want to provide the use of env vars for convenience, let me suggest a compromise: in addition to an API that sets all the config values explicitly (eg via passed struct), we could also provide an API that slurps the current env var values into a struct (so you could get the always-from-env-var behavior by adding a single line of code... but it's opt-in, rather than opt-out).

src/Target.cpp
```diff
@@ -374,6 +374,31 @@ Target::Feature get_host_cuda_capability(Target t) {
     return cap;
 }
 
+Target::Feature calculate_host_vulkan_capability(Target t) {
+    const auto *interface = get_device_interface_for_device_api(DeviceAPI::Vulkan, t);
+    internal_assert(interface->compute_capability);
+    int major, minor;
+    int err = interface->compute_capability(nullptr, &major, &minor);
+    internal_assert(err == 0) << "Failed to query cuda compute capability\n";
```
shoaibkamil, line 382: `cuda` -> `vulkan`

src/Target.cpp
```diff
@@ -374,6 +374,31 @@ Target::Feature get_host_cuda_capability(Target t) {
...
+    int ver = major * 10 + minor;
+    if (ver < 10) {
+        return Target::FeatureEnd;
+    } else if (ver < 10) {
```
shoaibkamil, line 386: The `if` and `else if` clauses have the same condition

src/runtime/internal/block_allocator.h
```diff
@@ -5,6 +5,8 @@
 #include "memory_resources.h"
 #include "region_allocator.h"
 
+// #define DEBUG_INTERNAL
+
```
shoaibkamil, line 9: accidental debug code

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2919 @@
...
+
+    uint32_t workgroup_size[3] = {0, 0, 0};
+};
+
+CodeGen_Vulkan_Dev::SPIRV_Emitter::SPIRV_Emitter(Target t)
+    : IRVisitor(), target(t) {
+    // Empty
+}
+
+void CodeGen_Vulkan_Dev::SPIRV_Emitter::scalarize(const Expr &e) {
```
shoaibkamil, line 428: Really feels like this should be an IR->IR pass somewhere earlier in the pipeline.  Can we create an issue for that?  I imagine this or similar code appears in multiple GPU backends.

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2919 @@
...
+    if (!value_type.is_bool()) {
+        value_id = cast_type(Bool(), value_type, value_id);
+    }
+    std::vector<uint8_t> true_data(target_type.bytes(), uint8_t(0));
+    std::vector<uint8_t> false_data(target_type.bytes(), uint8_t(0));
+    for (int i = 0; i < target_type.lanes(); ++i) {
+        if (target_type.is_int_or_uint() && target_type.bits() == 8) {
+            int8_t *td = reinterpret_cast<int8_t *>(&true_data[0]);
+            int8_t *fd = reinterpret_cast<int8_t *>(&false_data[0]);
+            td[i] = int8_t(1);
```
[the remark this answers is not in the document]
shoaibkamil: I'd advocate for changing these to the more common-in-Halide-code style of `(type)value` just for consistency.

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2919 @@
...
+        internal_assert(op->args.size() == 1);
+        visit_unary_op(SpvOpNot, op->type, op->args[0]);
+    } else if (op->is_intrinsic(Call::if_then_else)) {
+        Expr cond = op->args[0];
+        if (const Broadcast *b = cond.as<Broadcast>()) {
+            cond = b->value;
+        }
+        if (cond.type().is_vector()) {
+            scalarize(op);
+
```
shoaibkamil, line 1106: Nit: blank line

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2919 @@
...
+    } else if (op->is_intrinsic(Call::shift_left)) {
+        internal_assert(op->args.size() == 2);
+        if (op->type.is_uint() || (op->args[1].type().is_uint())) {
+            visit_binary_op(SpvOpShiftLeftLogical, op->type, op->args[0], op->args[1]);
+        } else {
+            Expr e = lower_signed_shift_left(op->args[0], op->args[1]);
+            e.accept(this);
+        }
+    } else if (op->is_intrinsic(Call::strict_float)) {
+        // TODO: Enable/Disable RelaxedPrecision flags?
```
shoaibkamil, line 1171: Do we need a github issue for this todo?

Makefile
```diff
@@ -133,6 +141,8 @@ WITH_LLVM_INSIDE_SHARED_LIBHALIDE ?= not-empty
 HL_TARGET ?= host
 HL_JIT_TARGET ?= host
 
+HL_VERSION_FLAGS=-DHALIDE_VERSION="$(HALIDE_VERSION)" -DHALIDE_VERSION_MAJOR=$(HALIDE_VERSION_MAJOR) -DHALIDE_VERSION_MINOR=$(HALIDE_VERSION_MINOR) -DHALIDE_VERSION_PATCH=$(HALIDE_VERSION_PATCH)
```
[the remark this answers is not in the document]
derek-gerstmann: Yes, that's much cleaner!  I'll change the formatting.

README_vulkan.md
```diff
@@ -0,0 +1,229 @@
...
+-   A vendor specific Vulkan device driver
+-   The generic Vulkan loader library
+
+For AMD & NVIDIA & Intel devices, download and install the latest graphics driver 
+for your platform. Vulkan support should be included.
+
+## Windows 
+
+To build Halide AOT generators, you'll need the Vulkan SDK (specifically the Vulkan loader library and headers):
```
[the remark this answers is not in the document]
derek-gerstmann: This is largely an unfortunate packaging issue on Windows where the CMake infrastructure for Vulkan expects to find the Vulkan headers (which we don't need), as well as the Vulkan loader library (which is needed for linking), all of which are packaged up in the SDK. So installing the SDK is the easiest solution.

src/runtime/vulkan_memory.h
```diff
@@ -0,0 +1,939 @@
...
+VulkanMemoryAllocator *vk_create_memory_allocator(void *user_context,
+                                                  VkDevice device,
+                                                  VkPhysicalDevice physical_device,
+                                                  const VkAllocationCallbacks *alloc_callbacks) {
+
+    SystemMemoryAllocatorFns system_allocator = {vk_system_malloc, vk_system_free};
+
+    VulkanMemoryConfig config = memory_allocator_config;
+
+    const char *min_block_size_env = getenv("HL_VK_MIN_BLOCK_SIZE");
```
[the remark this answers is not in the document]
derek-gerstmann: Agreed.  I'll add additional config functions for the vulkan API (like get/set device type) for these runtime config variables.

src/runtime/windows_vulkan.cpp
```diff
@@ -0,0 +1,2 @@
+#define WINDOWS
```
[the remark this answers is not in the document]
derek-gerstmann: This was primarily to conform to platform conventions ... all the other runtime targets appear to do something similar to allow a separate target to enable different binary linkage for each platform.

src/Target.cpp
derek-gerstmann: Ooops ... thanks!  I'll fix.

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Completely agree!  I'll add an issue for this!

README_vulkan.md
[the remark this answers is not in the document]
steven-johnson: We'll probably need to revisit this further, later on. May I suggest you open an Issue on it and refer to it here.

src/runtime/vulkan_memory.h
```diff
@@ -0,0 +1,939 @@
...
+    return result;
+}
+
+// --------------------------------------------------------------------------
+
+namespace {
+
+// --------------------------------------------------------------------------
+// Halide System allocator for host allocations
+void *vk_system_malloc(void *user_context, size_t size) {
```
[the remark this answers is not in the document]
derek-gerstmann: Vulkan has alignment constraints for resources allocated by the runtime, which are handled by the VulkanMemoryAllocator when allocating buffers and regions.  This vk_system_malloc callback is for host allocations which are used for allocating internal memory used within Halide.

README_vulkan.md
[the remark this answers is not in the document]
derek-gerstmann: Issue added .... tracking it under: 

[vulkan] Evaluate SDK dependency for AOT generators #7235

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Added issue ... tracking this as:

Refactor and consolidate common GPU backend IR operations #7238

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Issue added ... tracking this under:

[vulkan] Add support for strict_float / relaxed_precision in SPIR-V #7239

README_vulkan.md
```diff
@@ -0,0 +1,229 @@
...
+validated and disassembled via the SPIR-V tools:
+
+https://github.com/KhronosGroup/SPIRV-Tools
+
+
+# Current Status
+
+All correctness tests are now passing on tested configs for Linux & Windows using the target `host-vulkan-vk_int8-vk_int16-vk_int64-vk_float16-vk_float64-vk_v13` on LLVM v14.x. 
+
+MacOS passes most tests but encounters internal MoltenVK code translation issues for wide vectors, and ambiguously function calls.
```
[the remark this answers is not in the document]
derek-gerstmann: ambiguously called functions :smile:

README_vulkan.md
```diff
@@ -0,0 +1,229 @@
...
+
+Python apps, tutorials and correctness tests are now passing, but the AOT cases are skipped since the runtime environment needs to be customized to locate the platform specific Vulkan loader library.
+
+Android platform support is currently being worked on.
+
+# Caveats:
+
+-   Other than 32-bit floats and integers, every other data type is optional per the Vulkan spec
+-   Float 64-bit types can be enabled, but there aren't any native math functions available in SPIR-V
```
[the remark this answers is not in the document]
derek-gerstmann: During code-gen, if an optional feature isn't enabled and would be required to generate correct code, an error is issued.

For float-64 math functions, a "missing pure extern" error will occur since there is nothing appropriate to call.

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2919 @@
...
+        SpvId src_b_id = builder.current_id();
+
+        SpvId type_id = builder.declare_type(op->type);
+        SpvId result_id = builder.reserve_id(SpvResultId);
+        builder.append(SpvFactory::binary_op(SpvOpBitwiseAnd, type_id, result_id, src_a_id, src_b_id));
+        builder.update_id(result_id);
+    } else if (op->type.is_int() || op->type.is_uint()) {
+        // Just exploit the Euclidean identity
+        Expr zero = make_zero(op->type);
+        Expr equiv = select(op->a == zero, zero,
```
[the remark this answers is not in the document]
derek-gerstmann: This is the same as the GL-Compute backend.  The Vulkan spec indicates "... division by zero produces an unspecified result but must not lead to Vulkan interruption or termination".

src/SpirvIR.cpp
```diff
@@ -596,153 +970,384 @@ void SpvModule::encode(SpvBinary &binary) const {
...
+    reset();
+}
+
+void SpvBuilder::reset() {
+
+    active_id = SpvInvalidId;
+    active_function = SpvFunction();
+    active_block = SpvBlock();
+
+    kind_map.clear();
```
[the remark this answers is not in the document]
derek-gerstmann: It's called via the `void CodeGen_Vulkan_Dev::SPIRV_Emitter::init_module()` method, which gets called multiple times in certain test cases that use the JIT within separate scope brackets.

src/runtime/vulkan_context.h
```diff
@@ -0,0 +1,427 @@
...
+    vk_get_required_device_extensions(user_context, required_device_extensions);
+
+    StringTable optional_device_extensions;
+    vk_get_optional_device_extensions(user_context, optional_device_extensions);
+
+    StringTable supported_device_extensions;
+    vk_get_supported_device_extensions(user_context, *physical_device, supported_device_extensions);
+
+    bool valid_device = vk_validate_required_extension_support(user_context, required_device_extensions, supported_device_extensions);
+    halide_abort_if_false(user_context, valid_device);
```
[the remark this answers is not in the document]
derek-gerstmann: Good catch.  This is recoverable ... I'll return an appropriate error code.

src/runtime/vulkan_context.h
```diff
@@ -0,0 +1,427 @@
...
+    vkDestroyInstance(instance, alloc_callbacks);
+    return halide_error_code_success;
+}
+
+int vk_select_device_for_context(void *user_context,
+                                 VkInstance *instance, VkDevice *device,
+                                 VkPhysicalDevice *physical_device,
+                                 uint32_t *queue_family_index) {
+
+    // For now handle more than 16 devices by just looking at the first 16.
```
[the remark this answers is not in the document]
derek-gerstmann: :laughing: Me neither!  It's nonsensical ... I'll fix this.

steven-johnson:
Question for schedule estimation: after this PR lands, do we expect this to be a functional (although perhaps alpha- or beta- quality) backend? Or are there more things we know need to land for it to be usable?

derek-gerstmann:
> Question for schedule estimation: after this PR lands, do we expect this to be a functional (although perhaps alpha- or beta- quality) backend? Or are there more things we know need to land for it to be usable?

I'd consider this beta-quality as it is ... it's passing all tests on the platforms and devices I have access to, and it's more feature complete than the OpenGL compute backend.  However, it needs more testing on other platforms/devices, and performance tuning before I'd call it production ready.

steven-johnson:
You will need to sync this PR to `main` in order to build with top-of-tree LLVM -- ordinarily I'd just do that instead of suggesting it but there are enough conflicts that I suspect it would be better handled by you

steven-johnson:
(I took the liberty of syncing this to main, since the conflicts in test/generator were a little subtle)

derek-gerstmann:
> (I took the liberty of syncing this to main, since the conflicts in test/generator were a little subtle)

Ahh, you beat me to it! Yea, I noticed the generator conflicts were a bit more involved. Thank you!

steven-johnson:
There's a subtle but critical mistake in a lot of the runtime code (which is unfortunately easy to make):

When an error occurs in any Halide runtime code, you must do *BOTH* of these things:
- call `halide_error()`, either directly, or indirectly (typically via the `error()` stream output)
- AND, return a nonzero error code from the original entry point of the runtime call.

It's easy to leave one of these out, but failing to return a nonzero error code means that anything that replaces the error handler with something that doesn't abort won't see the errors (and may crash).

The specific case that caused me trouble was vulkan_memory.h:746, but from a quick glance, it looks like failing to return a nonzero error code is fairly pervasive in this PR (again, it's an easy mistake to make).

I'm surprised there doesn't seem to be any open Issue on this, so I'm opening one: https://github.com/halide/Halide/issues/7350

derek-gerstmann:
> When an error occurs in any Halide runtime code, you must do BOTH of these things:
> call halide_error(), either directly, or indirectly (typically via the error() stream output)
> AND, return a nonzero error code from the original entry point of the runtime call.

Ooooh!  Thanks for clarifying … I definitely did not know that!  I’ll make another pass and clean up the return codes so these propagate!

derek-gerstmann:
@steven-johnson Please take a look at the error handling fixes when you get a chance, and let me know if I missed anything!

steven-johnson:
> @steven-johnson Please take a look at the error handling fixes when you get a chance, and let me know if I missed anything!

I am on vacation next week -- I probably won't be able to review properly until March 27. 

(In the meantime, might make sense to merge this with main and resolve the conflicts)

derek-gerstmann:
> > @steven-johnson Please take a look at the error handling fixes when you get a chance, and let me know if I missed anything!
> 
> I am on vacation next week -- I probably won't be able to review properly until March 27.
> 
> (In the meantime, might make sense to merge this with main and resolve the conflicts)

Okay, no worries! Okay will do … I’ll merge this with main and resolve the conflicts

mirsadm:
Apologies if this is not the appropriate place to mention this. I've been keeping an eye out for this PR since I'm interested in migrating some code to use Vulkan on Android. Unfortunately, I can see that the following extensions are being used:
```
            if (arg.type.bits() == 8) {
                builder.require_extension("SPV_KHR_8bit_storage");
            } else if (arg.type.bits() == 16) {
                builder.require_extension("SPV_KHR_16bit_storage");
            }
```

In my limited testing they are not widely available on Android devices. The only safe option has been to use **uint32_t** and read 2 (or 4) pixels at a time.

derek-gerstmann:
> Apologies if this is not the appropriate place to mention this. I've been keeping an eye out for this PR since I'm interested in migrating some code to use Vulkan on Android. Unfortunately, I can see that the following extensions are being used:
> 
> ```
>             if (arg.type.bits() == 8) {
>                 builder.require_extension("SPV_KHR_8bit_storage");
>             } else if (arg.type.bits() == 16) {
>                 builder.require_extension("SPV_KHR_16bit_storage");
>             }
> ```
> 
> In my limited testing they are not widely available on Android devices. The only safe option has been to use **uint32_t** and read 2 (or 4) pixels at a time.

Hi!  Thanks very much for the feedback!  The current implementation requires that the host target flag being used when compiling turns on the features for vk_uint8 and vk_uint16 to enable this codegen path, and then the runtime validates that the required extensions are available when attempting to execute the code.  So, if you know that you're destined to run on a machine config that only has 32bit type support (which is the default assumption for vulkan), you shouldn't try and use 8bit or 16bit types.  I agree we could do more work to try and pack smaller datatypes into larger ones for configurations that don't support 8-bit or 16-bit storage, but this would impose a performance hit to do the implicit packing / unpacking with masking and shifts.  Is it acceptable for your use case to write your Halide algorithm such that it only uses 32-bit types for storage?

mirsadm:
> Hi! Thanks very much for the feedback! The current implementation requires that the host target flag being used when compiling turns on the features for vk_uint8 and vk_uint16 to enable this codegen path, and then the runtime validates that the required extensions are available when attempting to execute the code. So, if you know that you're destined to run on a machine config that only has 32bit type support (which is the default assumption for vulkan), you shouldn't try and use 8bit or 16bit types. I agree we could do more work to try and pack smaller datatypes into larger ones for configurations that don't support 8-bit or 16-bit storage, but this would impose a performance hit to do the implicit packing / unpacking with masking and shifts. Is it acceptable for your use case to write your Halide algorithm such that it only uses 32-bit types for storage?

Hi! Thanks for the reply. Yes it is of course possible to use 32-bit types but in my brief testing the performance penalty due to the extra memory bandwidth far outweighs the extra compute to unpack the pixels. I'll give it test once it is merged to master :)

steven-johnson:
> in my brief testing the performance penalty due to the extra memory bandwidth far outweighs the extra compute to unpack the pixels

Understandable, but... if a device doesn't have the uint8/uint16 extensions, then what other alternative is there?

derek-gerstmann:
> > in my brief testing the performance penalty due to the extra memory bandwidth far outweighs the extra compute to unpack the pixels
> 
> Understandable, but... if a device doesn't have the uint8/uint16 extensions, then what other alternative is there?

For what it’s worth, this Android developers guide seems to indicate that most shipping devices should have adequate support:

https://source.android.com/docs/core/graphics/implement-vulkan#versions

Note that v1.2 supports 8bit-storage

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2924 @@
...
+
+namespace {  // anonymous
+
+// --
+
+template<typename CodeGenT, typename ValueT>
+ValueT lower_int_uint_div(CodeGenT *cg, Expr a, Expr b);
+
+template<typename CodeGenT, typename ValueT>
+ValueT lower_int_uint_mod(CodeGenT *cg, Expr a, Expr b);
```
rootjalex, line 33: Are these ever used or defined? I don't see them anywhere else.

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2924 @@
...
+            uint32_t storage_buffer_count = 0;
+            SpecializationConstants specialization_constants;
+            SharedMemoryUsage shared_memory_usage;
+            WorkgroupSizeBinding workgroup_size_binding;
+        };
+        using DescriptorSetTable = std::vector<DescriptorSet>;
+        DescriptorSetTable descriptor_set_table;
+
+        // The workgroup size.  May vary between kernels.
+        uint32_t workgroup_size[3];
```
rootjalex, line 270: Why 3? As a non-GPU person this seems like a random magic number.

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2924 @@
...
+        builder.update_id(constant_id);
+    } else if (imm->type.bits() == 32) {
+        const uint32_t value = (uint32_t)(imm->value);
+        SpvId constant_id = builder.declare_constant(imm->type, &value);
+        builder.update_id(constant_id);
+    } else if (imm->type.bits() == 64) {
+        const uint64_t value = (uint64_t)(imm->value);
+        SpvId constant_id = builder.declare_constant(imm->type, &value);
+        builder.update_id(constant_id);
+    } else {
```
rootjalex, line 501: Seems like these could be simplified / shortened through a simple template. (and float too?)

mirsadm:
> Understandable, but... if a device doesn't have the uint8/uint16 extensions, then what other alternative is there?

With the devices I have at hand (Snapdragon 855/865), they support the feature shaderInt16 which allows them to use 16 bit integers in shader code. They are not able to use 16 bit buffers for input/output. Additionally, they support the extension ```VK_KHR_shader_float16_int8``` which allows me to read 32 bit ints into 2 float16 and get better throughput. Vulkan and Android is a bit of minefield but this has been the only way I have managed to get real time performance on Android with broader support.

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Ah, yes ... workgroup size is the extents for the three dimensional index space used as part of the kernel dispatch. It can also be used to adjust the layout for work items (aka GPU threads), based on logical groupings.

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Ah, good catch.  These declarations inside the anonymous namespace are no longer used ... the real definitions are inside the Internal namespace and declared in CodeGen_Internal.  I'll remove these! Thx!

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Ill add a comment clarifying its usage! Thx!

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Agreed.  I'll clean this up a bit. Thx!

derek-gerstmann:
> > Understandable, but... if a device doesn't have the uint8/uint16 extensions, then what other alternative is there?
> 
> With the devices I have at hand (Snapdragon 855/865), they support the feature `shaderInt16` which allows them to use 16 bit integers in shader code. They are not able to use 16 bit buffers for input/output. Additionally, they support the extension `VK_KHR_shader_float16_int8` which allows me to read 32 bit ints into 2 float16 and get better throughput. 

Hmmm. I'm not sure how this might be added as a feature ... we'd need to know that you intend to pass a buffer of 32bit integers with the intent of indexing it as two float16s?  And we'd have to know at CodeGen time, which would imply that you'd need some set of feature flags to trigger this. Feel free to add additional suggestions and use cases to the ticket I created:
https://github.com/halide/Halide/issues/7478

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2928 @@
...
+            declare_constant_float<bfloat16_t>(imm->type, imm->value);
+        } else {
+            declare_constant_float<float16_t>(imm->type, imm->value);
+        }
+    } else if (imm->type.bits() == 32) {
+        declare_constant_float<float>(imm->type, imm->value);
+    } else if (imm->type.bits() == 64) {
+        declare_constant_float<double>(imm->type, imm->value);
+    } else {
+        internal_error << "Vulkan backend currently only supports 32-bit or 64-bit floats\n";
```
rootjalex, line 538: This method seems to also handle bfloat16 and float16, is the error message incorrect?

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2928 @@
...
+        e.accept(this);
+    } else if (op->is_intrinsic(Call::saturating_cast)) {
+        Expr e = lower_intrinsic(op);
+        e.accept(this);
+
+    } else if (op->is_intrinsic()) {
+        Expr lowered = lower_intrinsic(op);
+        if (lowered.defined()) {
+            Expr e = lower_intrinsic(op);
+            e.accept(this);
```
rootjalex, line 1221: I think this should be:
```
lowered.accept(this);
```

src/CodeGen_Vulkan_Dev.cpp
```diff
@@ -0,0 +1,2928 @@
...
+    module.insert(module.end(), (const char *)spirv_binary.data(), (const char *)(spirv_binary.data() + spirv_binary.size()));
+}
+
+void CodeGen_Vulkan_Dev::SPIRV_Emitter::add_kernel(const Stmt &s,
+                                                   const std::string &kernel_name,
+                                                   const std::vector<DeviceArgument> &args) {
+    debug(2) << "Adding Vulkan kernel " << kernel_name << "\n";
+
+    // Add function definition
+    // TODO: can we use one of the function control annotations?
```
rootjalex, line 2797: Should this be an open issue, or removed?

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Good catch ... yes comment is out of date.  I'll fix this.

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Oh geez ... yeah, definitely not what I intended! Thanks for catching this.  I'll fix this!

src/CodeGen_Vulkan_Dev.cpp
derek-gerstmann: Added an open issue!

derek-gerstmann:
@steven-johnson I don't have write access to do the merge ... could you approve this PR when you get a chance?

steven-johnson:
> @steven-johnson I don't have write access to do the merge ... could you approve this PR when you get a chance?

Sure, but did the other reviewers approve it? It still says "needs 1 approving review"

derek-gerstmann:
> > @steven-johnson I don't have write access to do the merge ... could you approve this PR when you get a chance?
> 
> Sure, but did the other reviewers approve it? It still says "needs 1 approving review"

Nope. Nobody that has made comments has hit the approval button, even though all comments were addressed so far.

steven-johnson:
@rootjalex @shoaibkamil One of you, please approve this so we can land it :-)

rootjalex (approved):
Thanks for addressing the feedback!

steven-johnson (approved):

PR #6927, opened 2022-08-08
Halide::Error should not extend std::runtime_error

steven-johnson:
Unfortunately, the std error/exception classes aren't marked for DLLEXPORT under MSVC; we need our Error classes to be DLLEXPORT for libHalide (and python bindings). The current situation basically causes MSVC to generator another version of `std::runtime_error` marked for DLLEXPORT, which can lead to ODR violations, which are bad. AFAICT we don't really rely on this inheritance anywhere, so this just eliminates the inheritance entirely.

(Note that I can't point to a specific malfunction resulting from this, but casual googling based on the many warnings MSVC emits about the current situation has me convinced that it needs addressing.)

### DIFF
--- a/src/Error.cpp
+++ b/src/Error.cpp
@@ -55,12 +55,12 @@ Error &Error::operator=(const Error &that) {
     return *this;
 }
 
-Error::Error(Error &&that) {
+Error::Error(Error &&that) noexcept {
     this->what_ = that.what_;
     that.what_ = nullptr;
 }
 
-Error &Error::operator=(Error &&that) {
+Error &Error::operator=(Error &&that) noexcept {
     if (this != &that) {
         delete[] this->what_;
         this->what_ = that.what_;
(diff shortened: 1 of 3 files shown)


alexreinking:
Just a note: we ran into this issue before when investigating #4651 ages ago.

alexreinking (approved):

alexreinking:
Adding the release notes tag as this is an API break. Anyone catching `std::runtime_error` will no longer catch this error.

steven-johnson:
This is (finally) clean (rerunnning a Windows flake)... TL;DR: thrown objects need copy constructors (and failure mode for this on MSVC is obscure). PTAL.

alexreinking:
Is std::string also not DLLEXPORT?

steven-johnson:
> Is std::string also not DLLEXPORT?

Nope. Neither is std::unique_ptr, or pretty much anything in std, apparently.

PR #6928, opened 2022-08-08
Add minimal useful implementation of extracting and concatenating bits

abadams:
This is a minimum viable zero-cost way to load vectors of one type as larger or smaller vectors of a narrower or wider type respectively. It does this using two new intrinsics: concat_bits, and extract_bits.

Interleaves of extract_bits calls can simplify to simply loading a wider vector and then reinterpreting it, and concatenating the bits of a narrow strided load can simplify to doing a larger dense load and reinterpreting it. There are probably other cases where we can do some clever simplification. More test-cases welcome.

This PR doesn't contain any syntactic sugar for reinterpreting Func types. For the user there's just a fixed idiom one can use to get the desired effect.


src/IROperator.h
```diff
@@ -1561,6 +1561,54 @@ Expr gather(const Expr &e, Args &&...args) {
...
+
+template<typename T>
+Expr extract_bits(const Expr &e, const Expr &lsb) {
+    return extract_bits(type_of<T>(), e, lsb);
+}
+// @}
+
+/** Given a number of Exprs of the same type, concatenate their bits producing a
+ * single Expr of the same type code of the input but with more bits. The
+ * number of arguments must be a power of two.
```
rootjalex, line 1596: This constraint doesn’t seem to be enforced, might be useful to add a user_assert

src/FlattenNestedRamps.cpp
```diff
@@ -118,16 +118,55 @@ class FlattenRamps : public IRMutator {
...
 }  // namespace
 
 Stmt flatten_nested_ramps(const Stmt &s) {
-    FlattenRamps flatten_ramps;
-    return flatten_ramps.mutate(s);
+    return SimplifyConcatBits().mutate(FlattenRamps().mutate(s));
 }
 
 Expr flatten_nested_ramps(const Expr &e) {
     FlattenRamps flatten_ramps;
```
rootjalex, line 168: This object looks unused, should be removed

src/IROperator.cpp
```diff
@@ -2654,4 +2654,18 @@ Expr gather(const std::vector<Expr> &args) {
     return make_scatter_gather(args);
 }
 
+Expr extract_bits(Type t, const Expr &e, const Expr &lsb) {
+    return Internal::Call::make(t, Internal::Call::extract_bits, {e, lsb}, Internal::Call::Intrinsic);
+}
+
+Expr concat_bits(const std::vector<Expr> &e) {
+    user_assert(!e.empty()) << "concat_bits requires at least one argument\n";
```
rootjalex, line 2662: Should this method require at least two arguments? Either that, or a call with one argument should just simplify to that argument (in Simplify_Call.cpp perhaps?)

test/correctness/simplify.cpp
```diff
@@ -2348,6 +2348,10 @@ int main(int argc, char **argv) {
               Evaluate::make(0));
     }
 
+    {
+        check(concat_bits(x), x);
+    }
```
rootjalex, line 2353: concat_bits({x})

rootjalex (approved):

PR #6932, opened 2022-08-11
Tutorial 10 needs to be skipped for Python when targeting Wasm (just as non-Python does)

steven-johnson:


python_bindings/tutorial/CMakeLists.txt
```diff
@@ -16,52 +16,64 @@ set(tests
...
-        PYTHONPATH "$<TARGET_FILE_DIR:lesson_10_halide>"
+        PYTHONPATH "${PYPATH_${test_name}}"
     )
 endforeach ()
 
-## Add some hacks for getting CMake to delay compiling lesson_10_halide until after the test has run. The "better" way
-## of doing this might be to treat lesson 10 like an app and give it its own CMakeLists.txt, but since this is a one-off
-## it is probably less maintenance work to do it like this.
+if (TARGET_WEBASSEMBLY AND Halide_TARGET MATCHES "wasm")
+    message(WARNING "Not all tutorials build under WASM.")
```
alexreinking, line 36: return()

steven-johnson: If I do that, the packaging stuff below won't be processed

steven-johnson: (This particular tutorial really just needs to be rewritten to be way less janky; using PyGen when it lands is arguably the right thing, but for now, yuck)

alexreinking: ahhh GitHub's interface was confusing me. I thought this was the end of the file.

alexreinking (approved):

PR #6935, opened 2022-08-12
Revise PyGen API to have implicit GeneratorContext

steven-johnson:
This proposes to revise the Python Generator API to have GeneratorContext *never* be an explicit argument; rather, to have an *implicit* current GeneratorContext that is active. This reduces boilerplate when composing code (e.g. via Generator.call). To temporarily push a new GeneratorContext on the active stack, `with hl.set_context(ctx):` is the appropriate thing to do.

### DIFF
--- a/python_bindings/src/halide/_generator_helpers.py
+++ b/python_bindings/src/halide/_generator_helpers.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
-from contextlib import contextmanager
 from contextvars import ContextVar
 from enum import Enum
 from functools import total_ordering
@@ -373,19 +372,19 @@ def _unsorted_cls_dir(cls):
 
 _halide_generator_context = ContextVar('halide_generator_context', default=None)
 
-@contextmanager
-def set_context(context:GeneratorContext):
-    token = _halide_generator_context.set(context)
-    try:
-        yield context
-    finally:
-        _halide_generator_context.reset(token)
-
 def _get_generator_context() -> GeneratorContext:
     context = _halide_generator_context.get()
     _check(isinstance(context, GeneratorContext), "There is no active GeneratorContext")
     return context
 
+def _generatorcontext_enter(self: GeneratorContext) -> GeneratorContext:
+    if not hasattr(self, "_tokens"):
+        self._tokens = []
+    self._tokens.append(_halide_generator_context.set(self))
+    return self
+
+def _generatorcontext_exit(self: GeneratorContext) -> None:
+    _halide_generator_context.reset(self._tokens.pop())
 
 class Generator(ABC):
     """Base class for Halide Generators in Python"""
@@ -681,7 +680,7 @@ class Generator(ABC):
         # For most Generators this won't matter, but if the Generator
         # invokes SomeOtherGenerator.call(), it would be nice to have this
         # be the default, so that the end user doesn't have to mess with it.
-        with set_context(self.context()):
+        with self.context():
             self.generate()
 
         self._input_parameters = {n: getattr(self, n).parameter() for n in self._inputs_dict}
--- a/python_bindings/src/halide/_generator_helpers.py
+++ b/python_bindings/src/halide/_generator_helpers.py
@@ -377,9 +377,7 @@ _halide_generator_context = ContextVar('halide_generator_context', default=None)
 def set_context(context:GeneratorContext):
     token = _halide_generator_context.set(context)
     try:
-        _print("activating context: ", context)
         yield context
-        _print("deactivating context: ", context)
     finally:
         _halide_generator_context.reset(token)
 
--- a/python_bindings/src/halide/_generator_helpers.py
+++ b/python_bindings/src/halide/_generator_helpers.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
+from contextvars import ContextVar
 from enum import Enum
 from functools import total_ordering
 from .halide_ import *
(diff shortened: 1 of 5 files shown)


python_bindings/src/halide/_generator_helpers.py
```diff
@@ -366,6 +371,24 @@ def _unsorted_cls_dir(cls):
         yield (k, v)
 
 
+_halide_generator_context = ContextVar('halide_generator_context', default=None)
+
+@contextmanager
+def set_context(context:GeneratorContext):
+    token = _halide_generator_context.set(context)
+    try:
+        _print("activating context: ", context)
```
alexreinking, line 380: Is this a stray debug print?

alexreinking (changes requested):
Looks pretty dang good! But what's with the prints?

python_bindings/src/halide/_generator_helpers.py
steven-johnson: oops, yes

steven-johnson:
(Using ContextVar felt like maybe overkill, but it seemed like the prudent thing to do)

python_bindings/test/correctness/pystub.py
```diff
@@ -270,9 +264,12 @@ def test_complex(cls, extra_input_name = ""):
...
 if __name__ == "__main__":
-    test_simple(simple_pystub)
-    test_complex(complex_pystub)
-    test_complex(complex_pystub, extra_input_name = "foozz_input")
-    test_simple(SimplePy)
-    test_complex(ComplexPy)
-    test_complex(ComplexPy, extra_input_name = "foo_input")
+    target = hl.get_jit_target_from_environment()
+    ctx = hl.GeneratorContext(target)
+    with hl.set_context(ctx):
```
alexreinking, line 269: Yeah, this is a bit awkward... is it not possible to implement this stuff on `GeneratorContext` such that it adjusts the context and resets it later? Something like this might work:

```python
class GeneratorContext:
    # ... stuff ...
    def __enter__(self):
        token = _halide_generator_context.set(self)
        class _ResetToken:
            def __exit__(self):
                _halide_generator_context.reset(token)
        return _ResetToken()
```
Then you could write:
```python
with hl.GeneratorContext(target):
    # ...
```

alexreinking: An alternative approach could be to write:
```python
@contextmanager
def set_context(*args, **kwargs):
    context = hl.GeneratorContext(*args, **kwargs)
    token = _halide_generator_context.set(context)
    try:
        yield context
    finally:
        _halide_generator_context.reset(token)
```

and then the idiom would be:
```python
with hl.set_context(target) as context:
    # ...
```

alexreinking (changes requested):

python_bindings/test/correctness/pystub.py
alexreinking: Ack, no, my first suggestion doesn't work. Python calls __exit__ on the original object, not the one that is returned by __enter__.

steven-johnson: > is it not possible to implement this stuff on GeneratorContext

I tried a lot of things yesterday and couldn't see an obvious way to make it work, but yes, I would greatly prefer that syntax

> Something like this might work

Not sure I understand how that could work: for one, we're returning an instance of _ResetToken (so the `with ctx as c` variant would have c as a _ResetToken rather than a GeneratorContext). But more problematically, I don't see how __exit__ would ever get called; if I read PEP 343 correctly, the __exit__ method is always called on the original object in the with statement, not on the value returned.

> An alternative approach could be to write:

I'm not sure I understand how this is different from what I already have here? (aside from the fact that it implicitly converts a Target into a GeneratorContext, which would be handy to do but trivial to slot into what is already here)

steven-johnson: > You could still do this, though (tested):

Yeah... I guess that would work, though I must admit it makes me feel kinda squicky for reasons I can't put my finger on. I'll give it a try.

alexreinking: > makes me feel kinda squicky for reasons I can't put my finger on

Me too... I'm trying to find a case where "normal" code calls `__enter__` and `__exit__` out of sync, but this might just be paranoid. I feel like API ergonomics should trump sharp corners that are only exposed if you peel off the warranty sticker.

python_bindings/src/halide/_generator_helpers.py
```diff
@@ -366,6 +371,22 @@ def _unsorted_cls_dir(cls):
...
+def set_context(context:GeneratorContext):
+    token = _halide_generator_context.set(context)
+    try:
+        yield context
+    finally:
+        _halide_generator_context.reset(token)
+
+def _get_generator_context() -> GeneratorContext:
+    context = _halide_generator_context.get()
+    _check(isinstance(context, GeneratorContext), "There is no active GeneratorContext")
```
alexreinking, line 386: `.get()` above already throws is there is no active context. This check is really only confirming that we didn't internally mess up by calling `set_context` on something wrong.

steven-johnson: IIUC it should return None:

`_halide_generator_context = ContextVar('halide_generator_context', default=None)`

alexreinking: ah you put a default

python_bindings/test/correctness/pystub.py
steven-johnson: Revised, PTAL

alexreinking (approved):

alexreinking:
LGTM

PR #6938, opened 2022-08-15
Add pip packaging workflow to GHA

alexreinking:
_tldr_: 
```shell
$ pip install halide
```
---

This PR adds automated Python packaging to the Halide repo. It runs:

* When a PR is _merged_, this workflow will run on `main` and push to the _test_ PyPI index. This gives us sort-of nightlies for free.
* When a release is tagged, this workflow will run on the tagged commit and push to the _official_ PyPI index.

As this workflow is very expensive (Linux aarch64 takes several hours to complete) I don't want to spend our CI minutes on pull requests. We can fix Python packaging asynchronously until we find a cross-compiling solution that works for Linux aarch64.

In service of this, a few other changes had to be made:

* cibuildwheel is a semi-official helper action for building a wheel against multiple versions of Python. It assumes that setup.py and supporting files are in the root of your repo. Rather than reinvent the wheel, I moved the files. C'est la vie.
* The Adams2019 autoscheduler includes a hack for building "fat" (multiarch) objects on macOS. Our generators should probably learn that `mac-universal2-64` is a thing.
* I removed the code that detects the version number of the bindings because it was janky and breaking while I tried to test this. Maybe it can be restored one day so that our CMakeLists.txt (or perhaps some other file) can be the single source of truth for our version number. Alternatively, we could use something like [bumpversion](https://github.com/c4urself/bump2version) to make this process easier, but still manual.


.github/workflows/pip.yml
```diff
@@ -21,10 +21,28 @@ jobs:
...
+          path: zlib
+          ref: v1.2.12
       - name: Install dependencies
         run: |
           yum makecache
-          yum install -y ninja-build zlib-static
+          yum install -y ninja-build
+      - name: Configure ZLIB
+        run: >
+          cmake
```
steven-johnson, line 35: ...how the heck does a Linux Docker image not have zlib already installed

alexreinking: It has shared zlib only, but we need static

alexreinking:
Getting undiagnosable permissions errors... asking on GitHub Actions Discussions here: https://github.com/orgs/community/discussions/30173

steven-johnson (approved):

PR #6939, opened 2022-08-15
Add/update Python Readme

steven-johnson:
This moves the Python README to the toplevel and reworks it considerably, adding details and updating various bits.

Note that the Python documentation here is still incomplete; this is intended as a prelude to adding documentation for Python Generators in a future PR.


steven-johnson:
See also: https://github.com/halide/Halide/issues/6940

README_python.md
```diff
@@ -0,0 +1,273 @@
...
+    some point):
+
+    -   `DeviceInterface`
+    -   `evaluate()`
+
+-   The following parts of the Halide public API are currently only partially
+    supported by the Python bindings (but are all expected to be supported at
+    some point):
+
+    -   `ParamMap`
```
alexreinking, line 140: Merge these together?

README_python.md
```diff
@@ -0,0 +1,273 @@
...
+passing in an array constructed somewhere else, the easiest thing to do is to
+`.transpose()` it before passing it to your Halide code.)
+
+## Keeping Up To Date
+
+If you use the Halide Bindings for Python inside Google, you are *strongly*
+encouraged to
+[subscribe to announcements for new releases of Halide](https://github.blog/changelog/2018-11-27-watch-releases/),
+as it is likely that enhancements and tweaks to our Python support will be made
+in future releases.
```
alexreinking, line 264: Is this still relevant?

steven-johnson: Definitely, as I suspect that we'll find missing pieces & sharp corners that we'll want to fix sooner rather than later

README_python.md
steven-johnson: Actually, the ParamMap stuff should just go away entirely in Python, since it's there only for thread safety, and Callable provides a better way to do that. Willfix.

steven-johnson: Done

alexreinking (approved):

PR #6941, opened 2022-08-15
Export HalidePythonExtensionHelpers.cmake for installs

steven-johnson:
The functions in this will be helpful to end-users and should be exported, installed, and documented. (Note that both are actually related to Generators *and* Python; it might make more sense just to move them to the existing HalideGeneratorHelpers file?)


cmake/HalidePythonExtensionHelpers.cmake
```diff
@@ -28,7 +28,7 @@ set(_STUB_DIR "${Halide_SOURCE_DIR}/python_bindings/stub")
 # suffixed with "_stub" for Stub extensions. (TODO: this is unsatisfyingly hackish; better suggestions
 # would be welcome.)
 
-function(target_export_single_symbol TARGET SYMBOL)
+function(_target_export_single_symbol TARGET SYMBOL)
     configure_file("${_STUB_DIR}/ext.ldscript.apple.in" "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.ldscript.apple")
     configure_file("${_STUB_DIR}/ext.ldscript.linux.in" "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.ldscript")
```
alexreinking, line 33: The contents of these files need to be exported too,

cmake/HalidePythonExtensionHelpers.cmake
```diff
@@ -28,7 +28,7 @@ set(_STUB_DIR "${Halide_SOURCE_DIR}/python_bindings/stub")
...
     target_export_script(
```
alexreinking, line 34: This function is not built into CMake, but is rather one of our helpers, which is not already installed. That will need to be added to the package as well.

packaging/CMakeLists.txt
```diff
@@ -198,6 +198,7 @@ install(FILES
         ${CMAKE_CURRENT_BINARY_DIR}/HalideHelpersConfigVersion.cmake
         ${Halide_SOURCE_DIR}/cmake/HalideGeneratorHelpers.cmake
         ${Halide_SOURCE_DIR}/cmake/HalideTargetHelpers.cmake
+        ${Halide_SOURCE_DIR}/cmake/PythonExtensionHelpers.cmake
```
alexreinking, line 201: HalidePythonExtensionHelpers?

alexreinking (changes requested):
We need to build up some more comprehensive integration testing with our CMake package to catch packaging mistakes. At the very least, it shouldn't all be in my head.

cmake/HalidePythonExtensionHelpers.cmake
steven-johnson: Done

cmake/HalidePythonExtensionHelpers.cmake
steven-johnson: Moved contents of the template files into literal strings to avoid deps.

steven-johnson:
PTAL

steven-johnson:
PTAL

alexreinking (approved):

alexreinking:
> We need to build up some more comprehensive integration testing with our CMake package to catch packaging mistakes. At the very least, it shouldn't all be in my head.

Looks good, but I really feel the need to prioritize this now.

Issue #6944, opened 2022-08-16
compute_expr_cost() doesn't handle saturating_cast


steven-johnson:
Building apps/camera_pipe issues many dozens of the warning `Warning: Unhandled intrinsic call saturating_cast` -- seems like we should handle it.

PR #6947, opened 2022-08-17
Handle saturating_cast in compute_expr_cost()

rootjalex:
Addresses #6944


abadams (approved):

steven-johnson:
...lotsa failures in `performance_inner_loop_parallel` on Linux machines?

rootjalex:
That's super weird, those tests don't even use saturating casts at all. It also appears to be failures both on linux-worker-1, is there an issue with that buildbot?

rootjalex:
The piece of code that was changed isn't even touched by the execution of that test (on my machine), so this seems like a weird flake

rootjalex:
I just confirmed that this code is also not touched by that test on an x86 machine, so this seems like a builedbot flake, I'm going to assume it is fine to merge.

PR #6949, opened 2022-08-17
Remove add_python_aot_extension() rule in CMake

steven-johnson:
Move it into `add_halide_library` instead, as another output option.

(add_python_stub_extension will likely be moved as well, in a subsequent PR)


alexreinking (approved):

alexreinking:
Needs documentation in `README_cmake.md`, too

PR #6952, opened 2022-08-17
Remove add_python_stub_extension(), adding the functionality to add_halide_generator() instead

steven-johnson:
This is a bit more involved than the previous one:
- ~~To simplify the world, I moved the generic implementation of the PyStub call from its own bespoke library and into the Halide Python library; my reasoning here is that any code that can *usefully* use a PyStub (ie, to generate Halide IR in Python) is gonna have to have this dependency anyway. To accomplish this I had to change the pybind11_add_module type from MODULE to SHARED (otherwise I can't add it as a link dependency)... not sure if there is a downside to this that I don't realize~~   I moved this into a static library instead, the changes to Halide Python were just too weird
- The functionality of `add_python_stub_extension()` is now added to `add_halide_generator()` (note, *not* add_halide_library); if you specify `PYSTUB <generator-name>` then you get your stub generated. Note that we lost the ability to customize the stubname (it's now always `<generator-name>_pystub`); if this turns out to be unpopular then we can add the ability back in (I just wanted to keep options to a minimum)


README_cmake.md
```diff
@@ -909,32 +909,45 @@ add_halide_generator(
...
-This function guarantees that a Halide generator target named `<namespace><target>` is available. It will first search
-for a package named `<package-name>` using `find_package`; if it is found, it is assumed that it provides the target.
-Otherwise, it will create an executable target named `target` and an `ALIAS` target `<namespace><target>`. This function
-also creates a custom target named `<package-name>` if it does not exist and `<target>` would exist. In this case,
-`<package-name>` will depend on `<target>`, this enables easy building of _just_ the Halide generators managed by this
-function.
-
-After the call, `<PACKAGE_NAME>_FOUND` will be set to true if the host generators were imported (and hence won't be
-built). Otherwise, it will be set to false. This variable may be used to conditionally set properties on `<target>`.
+- If `PACKAGE_NAME` is not provided, it defaults to `${PROJECT_NAME}-halide_generators`.
```
steven-johnson, line 919: Drive-by rewrapping of these paragraphs

packaging/CMakeLists.txt
```diff
@@ -23,12 +23,18 @@ set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/tools"
...
+    set(extra_targets_to_export Halide_PyStubs)
+endif ()
+
+install(TARGETS Halide Halide_Generator Halide_LanguageOptions ${extra_targets_to_export}
         EXPORT Halide_Targets
         RUNTIME COMPONENT Halide_Runtime
         LIBRARY COMPONENT Halide_Runtime
         NAMELINK_COMPONENT Halide_Development
         ARCHIVE COMPONENT Halide_Development
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
alexreinking, line 37: This is the default, so it's redundant.

python_bindings/stub/CMakeLists.txt
```diff
@@ -1,12 +1,17 @@
...
-                          VISIBILITY_INLINES_HIDDEN TRUE
-                          POSITION_INDEPENDENT_CODE ON)
-endif ()
+target_link_libraries(Halide_PyStubs PRIVATE Halide::Halide)
+
+# Don't add a dependency on pybind11::pybind11 here: that will add
+# a phantom dependency is propagated into our install setup.
+# All we really need here is a path to the include directories for pybind11,
+# which is publicly documented as `pybind11_INCLUDE_DIRS`,
+target_include_directories(Halide_PyStubs PRIVATE "${HALIDE_INCLUDES}" "${pybind11_INCLUDE_DIRS}")
```
alexreinking, line 10: There's a better solution for this: link to `$<BUILD_INTERFACE:pybind11::pybind11>` in the `target_link_libraries()` command above. Then re-word the comment to explain why the dependency disappears post-build.

packaging/CMakeLists.txt
```diff
@@ -23,12 +23,18 @@ set(Halide_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/tools"
 
 target_sources(Halide_Generator INTERFACE $<INSTALL_INTERFACE:${Halide_INSTALL_TOOLSDIR}/GenGen.cpp>)
 
-install(TARGETS Halide Halide_Generator Halide_LanguageOptions
+set(extra_targets_to_export )
+if (WITH_PYTHON_BINDINGS)
+    set(extra_targets_to_export Halide_PyStubs)
+endif ()
```
alexreinking, line 29: This should go with the `Halide_Python` install rule in `python_bindings/src/halide/CMakeLists.txt`

alexreinking (changes requested):

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -63,6 +63,30 @@ function(add_halide_generator TARGET)
...
+             "${stub_text}")
+
+        Python3_add_library(${TARGET}_pystub MODULE WITH_SOABI "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.py_stub_generated.cpp" ${ARG_SOURCES})
+        set_target_properties(${TARGET}_pystub PROPERTIES
+                              CXX_VISIBILITY_PRESET hidden
+                              VISIBILITY_INLINES_HIDDEN ON
+                              POSITION_INDEPENDENT_CODE ON)
+        target_link_libraries(${TARGET}_pystub PRIVATE Halide::Halide Halide::Python ${ARG_LINK_LIBRARIES})
+        set_target_properties(${TARGET}_pystub PROPERTIES OUTPUT_NAME ${MODULE_NAME})
+        _Halide_target_export_single_symbol(${TARGET}_pystub "PyInit_${MODULE_NAME}")
```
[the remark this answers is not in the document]
alexreinking: I'll try this in a follow-up PR. Not worth holding up here.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -63,6 +63,30 @@ function(add_halide_generator TARGET)
...
+
+        file(WRITE
+             "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.py_stub_generated.cpp"
+             "${stub_text}")
+
+        Python3_add_library(${TARGET}_pystub MODULE WITH_SOABI "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.py_stub_generated.cpp" ${ARG_SOURCES})
+        set_target_properties(${TARGET}_pystub PROPERTIES
+                              CXX_VISIBILITY_PRESET hidden
+                              VISIBILITY_INLINES_HIDDEN ON
+                              POSITION_INDEPENDENT_CODE ON)
```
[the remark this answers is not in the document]
steven-johnson: Sure, but it doesn't hurt and it keeps symmetry with code elsewhere

packaging/CMakeLists.txt
steven-johnson: I think that CMake complained if LIBRARY DESTINATION was missing, or at least it did at one point, let me recheck

python_bindings/stub/CMakeLists.txt
steven-johnson: Cool, let me try that

packaging/CMakeLists.txt
steven-johnson: Will do

steven-johnson: OK, when I do that, I don't get Halide::PyStubs redirected at all in the install -- do we need to add `Halide_Python` here instead?

steven-johnson: So this change I couldn't make work at all. Either I misunderstand or, uh, something.

alexreinking: I'll take a look

cmake/HalideGeneratorHelpers.cmake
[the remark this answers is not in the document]
alexreinking: 🤷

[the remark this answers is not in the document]
steven-johnson: ...if you have an objection then say so -- seems to me a good redundancy

alexreinking (approved):

PR #6953, opened 2022-08-18
Build fixes for manylinux2014

alexreinking:
The `Development` component is the combination of two sub-components:

1. `Development.Module` - the parts necessary to create native Python modules
2. `Development.Embed` - the parts necessary to embed the Python interpreter into your application.

The manylinux images don't include `Development.Embed` and we don't need it anyway. So this fixes the build in that scenario. 

Also disables building tests and tutorials when building wheels because they're not needed.

### DIFF
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -30,7 +30,7 @@ set(PYBIND11_VER 2.6.2 CACHE STRING "The pybind11 version to use (or download)")
 # Dependencies
 ##
 
-find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
+find_package(Python3 REQUIRED COMPONENTS Interpreter Development.Module)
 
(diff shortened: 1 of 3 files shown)


python_bindings/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ set(PYBIND11_VER 2.6.2 CACHE STRING "The pybind11 version to use (or download)")
 # Dependencies
 ##
 
-find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
+find_package(Python3 REQUIRED COMPONENTS Interpreter Development.Module)
```
steven-johnson, line 33: Probably worth adding comment here about why

steven-johnson (approved):

python_bindings/CMakeLists.txt
alexreinking: Done

PR #6966, opened 2022-08-23
CMake packaging fixes

alexreinking:
This PR makes a few fixes to our CMake package (as used by `find_package`).

* The package now includes a hint as to the location of `HalideHelpers`, so that errors regarding not being able to find it are rarer (e.g. setting `Halide_DIR` instead of `Halide_ROOT`).
* `Halide_ASAN_ENABLED` is added to the package to allow consumers to confirm that they found the right Halide when they are using ASAN in their own projects.
* The handling of optional components was fixed. Previously, it incorrectly used the `REQUIRED`-ness of the whole Halide package itself, not the individual component.
  * PNG and JPEG dependencies are now optional by default. If any are missing, `find_package(Halide REQUIRED)` will still succeed. If you need a specific dependency for `Halide::ImageIO`, you can make it required by writing, e.g. `find_package(Halide REQUIRED PNG)`.
* Variables and macros that are internal to `HalideConfig.cmake` are now explicitly deleted with `unset` and a re-definition trick. This keeps our package API in line with what's documented.

### DIFF
--- a/packaging/common/HalideConfig.cmake
+++ b/packaging/common/HalideConfig.cmake
@@ -12,14 +12,9 @@ macro(Halide_find_component_dependency comp dep)
         set(Halide_quiet QUIET)
     endif ()
 
-    set(Halide_required)
-    if (${CMAKE_FIND_PACKAGE_NAME}_FIND_REQUIRED_${comp})
-        set(Halide_required REQUIRED)
-    endif ()
+    find_package(${dep} ${ARGN} ${Halide_quiet})
 
-    find_package(${dep} ${ARGN} ${Halide_quiet} ${Halide_required})
-
-    if (NOT ${dep}_FOUND)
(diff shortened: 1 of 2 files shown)


packaging/common/HalideConfig.cmake
```diff
@@ -114,3 +109,37 @@ else ()
...
+unset(Halide_known_components)
+
+# Delete paths to generated CMake files
+unset(Halide_shared_deps)
+unset(Halide_shared_targets)
+unset(Halide_static_deps)
+unset(Halide_static_targets)
+
+# Delete internal macros -- CMake saves redefined macros and functions with a
+# single underscore prefixed so, for example, Halide_fail is still available as
```
steven-johnson, line 130: wtaf

steven-johnson (approved):

PR #6968, opened 2022-08-23
Use CMake target to handle vendored SPIRV headers

alexreinking:
I think I did a bad job communicating the expected implementation in my review of #6882, but this is what I meant. Skipping buildbots because they don't test SPIRV yet. I have tested this locally.

Pinging @derek-gerstmann in case there's something I'm missing about the way it was previously done.


steven-johnson (approved):

alexreinking:
I'm cancelling buildbots that start to save resources.

PR #6961, opened 2022-08-22
Fix possible overflow in saturating_cast bounds inference

rootjalex:
The issue @steven-johnson noted on #6900 is a result of bounds inference producing integer overflow in the calculation of func value bounds. This PR should fix that issue.

PR #6970, opened 2022-08-23
Lower saturating_cast in bounds inference

rootjalex:
This should revert bounds inference on `saturating_cast` to the behavior of pre - #6900 .

@steven-johnson could you test to see if this fixes things?

The issues detected were not a direct result of #6900 - before that PR, there was still `signed_integer_overflow` inside of Bounds.cpp, but it was being hidden by complex `Cast` handling inside the bounds visitor. @abadams and I are discussing a fix to the underlying issue, but it is incredibly involved and requires possibly using infinite-precision integers in some parts of the compiler.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -7,6 +7,7 @@
 #include "Debug.h"
 #include "Deinterleave.h"
 #include "ExprUsesVar.h"
+#include "FindIntrinsics.h"
 #include "Func.h"
 #include "IR.h"
 #include "IREquality.h"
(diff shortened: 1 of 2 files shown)


steven-johnson:
Should this be applied *in addition to* https://github.com/halide/Halide/pull/6961, or *instead of*?

rootjalex:
Sorry for the lack of clarification - this is instead of #6961

steven-johnson:
OK, looks like this one is passing -- Let's just insert that fix for OpenGLCompute too; I monkey-patched it for my testing

steven-johnson (approved):
LGTM pending the fix to OpenGLCompute as well

abadams:
A proper fix seems intractable without getting some sort of repro for the harder failure cases. AJ's earlier attempt at a fix really should have worked.

abadams:
Maybe we need a test-case reducer like bugpoint

steven-johnson:
> A proper fix seems intractable without getting some sort of repro for the harder failure cases

I can try to devote more time to the other repro case, but don't rely on it, it may be hard to disentangle

rootjalex:
> LGTM pending the fix to OpenGLCompute as well

Just added.

> AJ's earlier attempt at a fix really should have worked.

To clarify for @steven-johnson , Andrew and I both think that https://github.com/halide/Halide/pull/6961/commits/129b464d2c5845497820520cd7e2b328aa5ba522 really should have worked, and cannot figure out what could have changed by #6900 that was not solved by that commit. It's very understandable if you can't disentangle that, but I think looking at that change would probably be easier than looking at the entire #6961 change.

rootjalex:
The only failure on the buildbots appears unrelated, I think we should merge this in as a temporary fix for now. I've opened #6974 to track the fundamental issue.

PR #6971, opened 2022-08-24
Small refactor to remove confusion between CodeGen_LLVM and CodeGen_Internal.

zvookin:
Also remove ```element_count```as it was only used in a single pattern which is now in the ```CodeGen_LLVM::get_splat``` routine.


zvookin:
Ugh, some extra stuff in here. One sec.

PR #6973, opened 2022-08-24
Small refactor to remove confusion between CodeGen_LLVM and CodeGen_Internal.

zvookin:
Also remove ```element_count``` as it was only used in a single pattern which is now in the ```CodeGen_LLVM::get_splat``` routine.

Also allows control of vscale vs. fixed types for vector types based on state in CodeGen_LLVM.

### DIFF
--- a/src/CodeGen_Internal.cpp
+++ b/src/CodeGen_Internal.cpp
@@ -16,16 +16,6 @@ using std::string;
 
 using namespace llvm;
 
-int get_vector_num_elements(llvm::Type *t) {
-    if (t->isVectorTy()) {
-        auto *vt = dyn_cast<llvm::FixedVectorType>(t);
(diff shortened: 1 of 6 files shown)


src/CodeGen_Internal.cpp
```diff
@@ -16,48 +16,13 @@ using std::string;
...
-            default:
-                internal_error << "There is no llvm type matching this floating-point bit width: " << t << "\n";
-                return nullptr;
-            }
-        } else if (t.is_handle()) {
-            return llvm::Type::getInt8PtrTy(*c);
-        } else {
-            return llvm::Type::getIntNTy(*c, t.bits());
-        }
+int get_vector_num_elements(llvm::Type *t) {
```
zvookin, line 19: Pretty sure this is now dead. Testing deleting it now.

steven-johnson:
LGTM pending green.

steven-johnson (approved):

PR #6977, opened 2022-08-25
Fix XCode by wrapping weights in an OBJECT library

### LINKED ISSUE #6976: CMake and the "new build system" of Xcode
I was trying to get an Xcode project to build Halide:
```
cmake -S ../../ -DLLVM_DIR=/opt/homebrew/opt/llvm/lib/cmake/llvm -DHalide_SHARED_LLVM=YES -G Xcode
```
and I got this:
```
CMake Error in src/autoschedulers/adams2019/CMakeLists.txt:
  The custom command generating

    Halide/build/xcode/src/autoschedulers/adams2019/baseline.cpp

  is attached to multiple targets:

    retrain_cost_model
    Halide_Adams2019

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

I managed to work around this by disabling autoschedulers.
(Btw, there's no CMake toggle to enable/disable autoschedulers support, I had to comment the `add_subdirectory(autoschedulers)` line in `src/CMakeFiles.txt`.)

alexreinking:
The XCode "new build system" doesn't like generated source files to be associated with more than one target. Going through an OBJECT library like this fixes that problem, but also saves us a compilation, so it's a good thing to do anyway.

Fixes #6976

### DIFF
--- a/src/autoschedulers/adams2019/CMakeLists.txt
+++ b/src/autoschedulers/adams2019/CMakeLists.txt
@@ -10,6 +10,8 @@ add_custom_command(OUTPUT ${WF_CPP}
                    DEPENDS baseline.weights binary2cpp
                    VERBATIM)
 
+add_library(Adams2019_weights_obj OBJECT ${WF_CPP})
+
 # cost_model, train_cost_model
 add_executable(cost_model.generator cost_model_generator.cpp)
 target_link_libraries(cost_model.generator PRIVATE Halide::Generator)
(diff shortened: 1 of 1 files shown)


slomp (approved):
Thanks for the super quick fix!

alexreinking:
> Thanks for the super quick fix!

Thanks for the super clear bug report! 🙂

steven-johnson (approved):

PR #6984, opened 2022-08-29
Change an assert to match code, hopefully.

zvookin:
Ran into this assert. The entire destructuring of For names things has to go, but the code only accesses two elements here and I'm not sure why a two level loop next shouldn't be applicable here.

### DIFF
--- a/src/CanonicalizeGPUVars.cpp
+++ b/src/CanonicalizeGPUVars.cpp
@@ -128,7 +128,7 @@ class CanonicalizeGPUVars : public IRMutator {
             (op->for_type == ForType::GPULane)) {
 
             vector<string> v = split_string(op->name, ".");
-            internal_assert(v.size() > 2);
+            internal_assert(v.size() >= 2);
 
             CountGPUBlocksThreads counter(v[0] + "." + v[1]);
             op->body.accept(&counter);


steven-johnson (approved):

src/CanonicalizeGPUVars.cpp
```diff
@@ -128,7 +128,7 @@ class CanonicalizeGPUVars : public IRMutator {
             (op->for_type == ForType::GPULane)) {
 
             vector<string> v = split_string(op->name, ".");
-            internal_assert(v.size() > 2);
+            internal_assert(v.size() >= 2);
 
             CountGPUBlocksThreads counter(v[0] + "." + v[1]);
```
abadams, line 133: The arg to CountGPUBlocksThreads is a prefix identifying a Func and a stage. It has no Var. So there should be at least three fields after splitting by '.'

abadams: Maybe a loop name has been normalized in some way?

zvookin: I'm compiling hand written IR to shader's directly as part of an experiment. The entire idea of an invariant on the name here is problematic, but I'm not sure why it needs three levels for the job it is doing. I.e. is it enforcing an invariant of this code or a thing that should be true for other reasons?

abadams: Rereading the code, it's looking for a prefix that all loops corresponding to the same Func and Stage are going to have in common, so that it can identify just those loops as belonging to a single loop nest for that Func/Stage. If the prefix is the entire loop name (v.size() == 2), then it can't possibly do anything useful, because the counter won't find any matching loops. This whole chunk of canonicalize_gpu_vars doesn't really make any sense if there aren't nested loops that share common prefixes of the form "func_name.stage_idx." 

I believe in your case it won't find any matching loops so it will name every GPU loop as if it's the innermost one of that type (i.e. thread/block/etc), and you'll have problems if you have nested loops of the same type.

steven-johnson:
This PR has been sitting here for a while -- should it land?

zvookin:
Andrew and I agreed that the entire thing being checked here is probably no longer needed. Planning to test that hypothesis and do it on this PR if it works.

steven-johnson:
Any update on this PR?

steven-johnson:
Ping for status.

abadams:
I'm testing the hypothesis that the thing being checked is no longer needed. Will report back shortly.

abadams:
Replaced by #7386

PR #6985, opened 2022-08-29
Add `add_halide_runtime` rule

### LINKED ISSUE #6981: CMake should have an `add_halide_runtime()` rule
Currently we don't have a clean way to add a Halide Runtime library without a Generator attached; we should really consider adding one, because:
- Advanced users could use it to reduce redundant build time (since Generators default to produce their own bespoke runtimes, even when identical)
- `add_halide_python_extension_library()` requires all participating targets to have *identical* runtimes, which is unpleasantly hackish at present

steven-johnson:
Fixes #6981

### DIFF
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -270,8 +270,7 @@ function(add_halide_library TARGET)
         set(ARG_USE_RUNTIME Halide::Runtime)
     elseif (NOT ARG_USE_RUNTIME)
         # If we're not using an existing runtime, create one.
-        _Halide_add_halide_runtime("${TARGET}.runtime" FROM ${ARG_FROM}
-                                   TARGETS ${ARG_TARGETS})
+        add_halide_runtime("${TARGET}.runtime" TARGETS ${ARG_TARGETS})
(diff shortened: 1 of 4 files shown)


cmake/HalideGeneratorHelpers.cmake
```diff
@@ -515,8 +514,28 @@ endfunction()
...
+    # If no TARGETS argument, use Halide_TARGET instead
+    if (NOT ARG_TARGETS)
+        set(ARG_TARGETS "${Halide_TARGET}")
+    endif ()
+
+    # Ensure all targets are tagged with "no_runtime",
+    # so that GCD calculation doesn't get confused.
+    list(TRANSFORM ARG_TARGETS APPEND "-no_runtime")
+
+    # Create a Generator that is GenGen.cpp and nothing else; all it can do is generate a runtime.
```
alexreinking, line 533: TODO: ship this and don't build it lazily like this

alexreinking (approved):
Looks good to me

cmake/HalideGeneratorHelpers.cmake
steven-johnson: ...I don't understand?

PR #6986, opened 2022-08-31
Improve error-handling in Python Extensions

steven-johnson:
Currently, Python Extensions don't make any effort to override `halide_error`, so the default (which aborts) is generally used... this is very unfriendly. This modifies the standard Python Extension glue code to hook halide_error, saving the text in a thread local, and then throwing a Python exception after the extension's AOT call is finished (if an error occurred, of course).

Also does a drive-by default hooking of `halide_print` to ensure that it goes to whatever Python thinks that `stdout` is.

(Note that it would be really nice if we could use closures of some sort for halide_error, halide_print, etc so that we could save context in the actual Python module, rather than in a thread-local global var, but this currently isn't possible without nontrivial refactoring in the Halide runtime.)


alexreinking (approved):
LGTM

PR #6991, opened 2022-08-31
Refactor buffer-unpacking code in PythonExtensionGen

steven-johnson:
This moves most of the interesting code into the common module block, so we don't risk duplicating code for extensions that contain multiple function definitions.


alexreinking (approved):

alexreinking:
I can't help but wonder what it would take for more of this code to be visible to clang-tidy

PR #6992, opened 2022-09-01
Fix compiler warnings in Elf.cpp

steven-johnson:
Some versions of GCC will complain that there is a possible use of uninitialized field `Sym<>::st_info` here; that's technically true, in that it is a bitfield that we previously set via two calls, so it temporarily could use uninitialized bits, but those would immediately be overwritten by well-defined bits. That said, the API could have been misused, so I collapsed Sym::set_type and Sym::set_bindings into a single call to avoid this warning.

While I was there, I did a little hygiene on Rel<> and Rela<> as well, as there was an unused-but-similarly-dubious API there. ~~Also added some C++17 `if constexpr` love.~~

### DIFF
--- a/src/Elf.cpp
+++ b/src/Elf.cpp
@@ -139,11 +139,11 @@ struct Rel {
     typedef typename T::addr_t addr_t;
     typedef typename T::addr_off_t addr_off_t;
 
-    addr_t r_offset;
-    addr_t r_info;
+    const addr_t r_offset;
+    const addr_t r_info;
 
     uint32_t r_type() const {
-        if (sizeof(addr_t) == 8) {
+        if constexpr (sizeof(addr_t) == 8) {
             return r_info & 0xffffffff;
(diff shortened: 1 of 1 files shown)


src/Elf.cpp
```diff
@@ -139,56 +139,49 @@ struct Rel {
...
     typedef typename T::addr_off_t addr_off_t;
 
-    addr_t r_offset;
-    addr_t r_info;
+    const addr_t r_offset;
+    const addr_t r_info;
 
     uint32_t r_type() const {
-        if (sizeof(addr_t) == 8) {
+        if constexpr (sizeof(addr_t) == 8) {
```
zvookin, line 146: I'm not sure what these are here for. I don't think the affect code generation and putting the assertion on a sizeof check seems more noise than useful to me.

zvookin (approved):

PR #6995, opened 2022-09-05
Couple small fixes to update RISC V to current LLVM flags and enable vscale use.

zvookin:
The vector extension is no longer experimental and the logic for picking the vector size and vscale was not hooked up.

### DIFF
--- a/src/CodeGen_RISCV.cpp
+++ b/src/CodeGen_RISCV.cpp
@@ -25,6 +25,7 @@ protected:
     string mabi() const override;
     bool use_soft_float_abi() const override;
     int native_vector_bits() const override;
+    int target_vscale() const override;
 };
 
 CodeGen_RISCV::CodeGen_RISCV(const Target &t)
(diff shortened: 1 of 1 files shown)


src/CodeGen_RISCV.cpp
```diff
@@ -50,7 +51,7 @@ string CodeGen_RISCV::mattrs() const {
     string arch_flags = "+m,+a,+f,+d,+c";
 
     if (target.has_feature(Target::RVV)) {
-        arch_flags += ",+experimental-v";
+        arch_flags += ",+v";
```
steven-johnson, line 54: Does this need to be conditional on LLVM_VERSION?

steven-johnson (approved):

src/CodeGen_RISCV.cpp
zvookin: Interesting question. I'm tempted to say we don't want to support something that was experimental. Like if one is using RISC V vector, one wants to be at least this recent on LLVM. (And it will not be a build time issue.) Note it's been broken for maybe a few months so anyone actually using it was making this change by hand. Think it is a safe bet to not conditionalize.

steven-johnson:
LGTM

PR #6856, opened 2022-07-19
Add GPU autoscheduler

aekul:
This is a draft PR to merge the anderson2021 GPU autoscheduler (#5602). It includes the code (there is some overlap with adams2019, but it will probably be a substantial amount of work to deduplicate them), some tests, utility scripts for generating data/statistics/etc. (these can be removed if desired), and baseline weights (trained on a V100).

PR #7003, opened 2022-09-08
Clean up Adams2019 CMake file

steven-johnson:
This does a few things:
- The main thing it does is add a unique prefix to all the targets in Adams2019; apparently CMake has a global namespace for targets, and when they collide (as happened in #6856) you get amusing results.
- Added `add_adams2019_test` helper to clean up boilerplate. Note that setting HL_TARGET was never necessary for any of these tests.

### DIFF
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -80,7 +80,7 @@ else ()
 endif ()
 
 if (TARGET Halide_Adams2019)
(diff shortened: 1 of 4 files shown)


alexreinking (approved):
We really need to reorganize the autoscheduler tests under... `tests/`...

steven-johnson:
> We really need to reorganize the autoscheduler tests under... `tests/`...

And rename them to `Mullapudi`1

PR #7010, opened 2022-09-12
Rework Python Extension C++ code (again)

steven-johnson:
My previous effort was too clever for itself: while it worked for Halide's build systems, some other build systems (e.g. Blaze) are much more finicky about the C++ files you build Python extensions from, and re-using the same C++ files with different preprocessor settings turns out to be too problematic there, for reasons that aren't important here.

Anyway, the important part here was to rework so that (1) All the C++ source files needed are compiled exactly once (2) All the C++ source files needed can be compiled with the same set of preprocessor definitions

To that end, I have extended GenGen's `-r` flag to allow using `-e python_extension`; this emits the bare module-registration code by itself. So now, we generate the Python Extension code as before, but define HALIDE_PYTHON_EXTENSION_OMIT_MODULE_DEFINITION to defeat the standalone module registration for each one, then also compile in the new 'standalone' registration, with HALIDE_PYTHON_EXTENSION_MODULE and HALIDE_PYTHON_EXTENSION_FUNCTIONS defined to fill in the blanks.

Also, a little drive-by cleanup in CodeGen_C to make extern "C" blocks more findable, and some restructuring in PyExtGen.

### DIFF
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -672,7 +672,14 @@ endfunction()
 function(_Halide_gengen_ensure)
     # Create a Generator that is GenGen.cpp and nothing else; all it can do is generate a runtime.
     if (NOT TARGET _Halide_gengen)
-        add_executable(_Halide_gengen )
+
+        # add_executable requires at least one source file for some
+        # configs (e.g. Xcode), because, uh, reasons, so we'll create
+        # an empty one here to satisfy it
+        set(empty "${CMAKE_CURRENT_BINARY_DIR}/_Halide_gengen.empty.cpp")
+        file(WRITE "${empty}" "/* nothing */\n")
+
+        add_executable(_Halide_gengen "${empty}")
         target_link_libraries(_Halide_gengen PRIVATE Halide::Generator)
         _Halide_place_dll(_Halide_gengen)
     endif ()
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -454,30 +454,22 @@ function(add_halide_python_extension_library TARGET)
         message(FATAL_ERROR "${TARGET} requires all libraries to use the same Halide Runtime, but saw ${len}: ${runtimes}")
     endif ()
 
-    # Module def code is the same in all of them, so arbitrarily choose the first
-    list(GET pycpps 0 module_definition_cpp)
-    add_library(${TARGET}_module_definition OBJECT ${module_definition_cpp})
-    set_target_properties(${TARGET}_module_definition PROPERTIES
-                          CXX_VISIBILITY_PRESET hidden
-                          VISIBILITY_INLINES_HIDDEN ON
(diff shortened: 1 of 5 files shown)


cmake/HalideGeneratorHelpers.cmake
```diff
@@ -680,3 +668,13 @@ function(_Halide_target_export_single_symbol TARGET SYMBOL)
         GNU_LD "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.ldscript"
     )
 endfunction()
+
+function(_Halide_gengen_ensure)
+    # Create a Generator that is GenGen.cpp and nothing else; all it can do is generate a runtime.
+    if (NOT TARGET _Halide_gengen)
+        add_executable(_Halide_gengen )
```
alexreinking, line 675: This is going to cause problems on Xcode, as it doesn't have a source file associated with it. Unfortunately need an empty, dummy source, here.

python_bindings/test/generators/user_context_generator.cpp
```diff
@@ -1,3 +1,4 @@
+
```
alexreinking, line 1: Why?

alexreinking (changes requested):

cmake/HalideGeneratorHelpers.cmake
steven-johnson: Er, ok, but this is exactly what we were already doing -- I just moved the code around. Were we broken before?

(Side note: wtf, cmake, how is it possible this isn't supported yet)

python_bindings/test/generators/user_context_generator.cpp
steven-johnson: Scalpel left in patient

cmake/HalideGeneratorHelpers.cmake
steven-johnson: Done

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -680,3 +668,20 @@ function(_Halide_target_export_single_symbol TARGET SYMBOL)
...
+
+function(_Halide_gengen_ensure)
+    # Create a Generator that is GenGen.cpp and nothing else; all it can do is generate a runtime.
+    if (NOT TARGET _Halide_gengen)
+
+        # add_executable requires at least one source file for some
+        # configs (e.g. Xcode), because, uh, reasons, so we'll create
+        # an empty one here to satisfy it
+        set(empty "${CMAKE_CURRENT_BINARY_DIR}/_Halide_gengen.empty.cpp")
+        file(WRITE "${empty}" "/* nothing */\n")
```
alexreinking, line 680: This needs to be in an `if (NOT EXISTS "${empty}")` to avoid redundant rebuilds on every reconfigure.

alexreinking (changes requested):

cmake/HalideGeneratorHelpers.cmake
steven-johnson: PTAL

alexreinking (approved):

PR #7021, opened 2022-09-15
Allow explicit inheritance for python Generators

steven-johnson:
Changes Python Generators to require explicit inheritance from hl.Generator (rather than forbidding it). 

Not particularly happy about this change, but it seems to be the short path to getting code that is compliant with pytype. 

Hopefully we'll find a better workaround and can soften this requirement.

### DIFF
--- a/python_bindings/src/halide/_generator_helpers.py
+++ b/python_bindings/src/halide/_generator_helpers.py
@@ -765,7 +765,6 @@ def alias(**kwargs):
 
     return alias_impl
 
-
 def generator(name:str=""):
     # This code relies on dicts preserving key-insertion order, which is only
     # guaranteed for all Python implementations as of v3.7.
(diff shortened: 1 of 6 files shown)


alexreinking:
I am accepting this PR because it's externally important that these features land soon. _However,_ I am alarmed by a few things:

1. No Python type-checking tools are standard and type-checking is still far from the norm for Python. 
2. There was no discussion early on that passing `pytype` was a requirement. If it had been known, we would have had a chance to find a solution that appeases `pytype` while maintaining good UX.
3. There is still no GHA workflow or buildbot step that runs `pytype`. Playing telephone with other companies' CI systems is unsustainable.
4. This patch makes the UX worse.
5. This patch doesn't fix a real type error.

alexreinking (approved):

PR #7022, opened 2022-09-16
Appease Python linter

steven-johnson:
Apparently it's "more Pythonic" to use "not in" vs "not ... in", etc.


alexreinking (approved):
Yeah, I noticed this but thought it was too pedantic if we weren't going to buy into Python linting. :man_shrugging:

PR #7031, opened 2022-09-20
Codegen_C for user_context

steven-johnson:
While adding a user_context to the `async_parallel` generator for other purposes, I found some subtle bugs in Codegen_C regarding possible aliasing and/or const-correctness of the internal 'ucon' arguments when calling or unpacking a closure for async usage. This fixes it.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -2069,7 +2069,12 @@ void CodeGen_C::visit(const Variable *op) {
         // This is the name of a global, so we can't modify it.
         id = op->name;
     } else {
-        id = print_name(op->name);
+        // This substitution ensures const correctness for all calls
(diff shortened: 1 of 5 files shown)


src/CodeGen_C.cpp
```diff
@@ -2069,7 +2069,12 @@ void CodeGen_C::visit(const Variable *op) {
         // This is the name of a global, so we can't modify it.
         id = op->name;
     } else {
-        id = print_name(op->name);
+        // This substitution ensures const correctness for all calls
+        if (op->name == "__user_context") {
```
vksnk, line 2073: Worth adding a helper function to check if the name is "__user_context?

vksnk (approved):

steven-johnson:
Failures are unrelated, merging

PR #7034, opened 2022-09-20
Add stack-size-canary test to apps/fft's CMake file

steven-johnson:
This was apparently meant as a canary for stack size usage, but the necessary setting only happened in the Makefile, not the CMake file.

Also, drive-by fix in Makefile to ignore warnings about `-ObjC++` being ignored, which apparently can be the case with current AppleClang configs.

### DIFF
--- a/apps/fft/CMakeLists.txt
+++ b/apps/fft/CMakeLists.txt
@@ -33,7 +33,7 @@ add_halide_library(fft_inverse_c2c FROM fft.generator
 # flag on os x, and testing this on one platform is sufficient.
 set(LDFLAGS "")
 if (CMAKE_SYSTEM_NAME MATCHES "Darwin")
-    set(LDFLAGS "-Wl,-stack_size;-Wl,0x100000")
+    set(LDFLAGS "LINKER:-stack_size,0x100000")
 endif()
 
 # Main executable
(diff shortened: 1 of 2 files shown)


apps/fft/CMakeLists.txt
```diff
@@ -28,13 +28,23 @@ add_halide_library(fft_inverse_c2c FROM fft.generator
                    GENERATOR fft
                    PARAMS direction=frequency_to_samples size0=16 size1=16 input_number_type=complex output_number_type=complex)
 
+# Use a small stack, to stress out the compiler and act as a canary
+# for stack bloat problems. This can be accomplished with a linker
+# flag on os x, and testing this on one platform is sufficient.
+set(LDFLAGS "")
+if (CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    set(LDFLAGS "-Wl,-stack_size;-Wl,0x100000")
```
alexreinking, line 36: For future-proofing portability.
```suggestion
    set(LDFLAGS "LINKER:-stack_size,0x100000")
```

steven-johnson: Done

alexreinking (approved):

PR #7038, opened 2022-09-20
Add a terminate_handler to try to report unhandled exceptions

### LINKED ISSUE #6959: Halide built with exceptions can produce test failures that don't report errors

steven-johnson:
Fixes #6959.

### DIFF
--- a/src/Error.cpp
+++ b/src/Error.cpp
@@ -3,6 +3,7 @@
 #include "Util.h"  // for get_env_variable
 
 #include <csignal>
+#include <mutex>
 
 #ifdef _MSC_VER
 #include <io.h>
--- a/src/Error.cpp
+++ b/src/Error.cpp
@@ -104,6 +104,36 @@ InternalError::InternalError(const char *msg)
 
 namespace Internal {
 
+void unhandled_exception_handler() {
+    // Note that we use __cpp_exceptions (rather than HALIDE_WITH_EXCEPTIONS)
+    // to maximize the change of dealing with uncaught exceptions in weird
+    // build situations (i.e., exceptions enabled via C++ but HALIDE_WITH_EXCEPTIONS
+    // is somehow not set).
+#ifdef __cpp_exceptions
+    // This is a trick: rethrow the pending (unhandled) exception
+    // so that we can see what it is and log `what` before dying.
+    if (auto ce = std::current_exception()) {
+        try {
+            std::rethrow_exception(ce);
+        } catch (Error &e) {
+            // Halide Errors are presume to be nicely formatted as-is
+            std::cerr << e.what() << std::flush;
+        } catch (std::exception &e) {
+            // Arbitrary C++ exceptions... who knows?
+            std::cerr << "Uncaught exception: " << e.what() << "\n"
+                      << std::flush;
+        } catch (...) {
+            std::cerr << "Uncaught exception: <unknown>\n"
+                      << std::flush;
+        }
+    }
+#else
+    std::cerr << "unhandled_exception_handler() called but Halide was compiled without exceptions enabled; this should not happen.\n"
+              << std::flush;
+#endif
+    abort();
+}
+
 // Force the classes to exist, even if exceptions are off
 namespace {
 CompileError _compile_error("");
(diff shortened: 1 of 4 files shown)


src/Error.cpp
```diff
@@ -172,6 +201,11 @@ ErrorReport::~ErrorReport() noexcept(false) {
     }
 
 #ifdef HALIDE_WITH_EXCEPTIONS
+    static std::once_flag init;
+    std::call_once(init, []() {
+        std::set_terminate(UnhandledExceptionHandler);
```
alexreinking, line 206: I don't think this belongs in libHalide... library code generally shouldn't trample consumers' termination handlers. Between these two locations, I think we'll have everything we care about covered.

1. Enabling this in GenGen
2. Enabling this in our tests via an object library, the same way we do the abort-to-exit-code-1 translation for our error tests in `expect_abort.cpp`

src/Error.cpp
```diff
@@ -148,6 +149,34 @@ ErrorReport::ErrorReport(const char *file, int line, const char *condition_strin
...
+
+// This is a trick: rethrow the pending (unhandled) exception
+// so that we can see what it is and log `what` before dying.
+// Strictly speaking, standard C++ doesn't allow you to do a try/throw/catch
+// within a terminate handler, but in practice, Clang and GCC (at least)
+// allow it, so we'll use it to improve quality-of-life for our tests.
+void UnhandledExceptionHandler() {
+    try {
+        throw;
+    } catch (Error &e) {
```
alexreinking, line 163: Neat, but might be better to do:
```
if (auto exn = std::current_exception()) {
    try {
        std::rethrow_exception(exn);
    } catch (Error &e) {
        // ...
    }
}

abort();
```
This way we don't fail to rethrow an exception, in case `terminate` was called for a reason besides failure-to-catch

alexreinking:
> This is not standard C++,

Why do you say that? Consulting the [C++17 standard](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf), sec 18.3p7-8, I read:

> (7) A handler is considered active when initialization is complete for the parameter (if any) of the catch clause. [ Note: The stack will have been unwound at that point. — end note ] Also, an implicit handler is considered active when std::terminate() is entered due to a throw. A handler is no longer considered active when the catch clause exits.
> (8) The exception with the most recently activated handler that is still active is called the currently handled exception.

So by my reading, there is a notion of an active exception while `std::terminate` is running. Paragraph 7 _defines_ this implicit handler. It should be legal to re-throw as long as _that_ throw doesn't leave the terminate handle.

src/Error.cpp
steven-johnson: > Why do you say that? Consulting the [C++17 standard](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf), sec 18.3p7-8, I read:

Interesting -- I had been under the impression that trying to do anything fancy in std::terminate was (technically) UB.

steven-johnson: > Neat, but might be better to do:
> 
> ```
> if (auto exn = std::current_exception()) {
> ```

Hmm... yeah, I'll give that a try.

src/Error.cpp
steven-johnson: Done, PTAL

alexreinking (approved):
LGTM when green

PR #7039, opened 2022-09-21
Handle widen_right_* intrinsics in bounds inference

vksnk:
It seems widen_right_add can appear in load indices and this was causing difficulties [here](https://github.com/halide/Halide/blob/main/src/HexagonOptimize.cpp#L2010), i.e. load+shuffle wasn't generated somewhere where it was happening before.


rootjalex (approved):
Thanks for the catch!

Issue #7044, opened 2022-09-23
A Generator's `add_requirements` get lost when used via cpp_stubs or Python's `call()` method


steven-johnson:
When you call `Generator::add_requirement` (in either C++ or Python), we add the requirements to a list in the `Pipeline` we create; then, when lowering, convert those into a series of `AssertStmts` at the top of the block.

A subtle issue here comes in when you want to re-use a Generator at language-composition time, either via `cpp_stubs` or via the equivalent `call()` method in our Python Generators: instead of returning the `Pipeline`, these return the output(s) as a list of `Func`, extracted from the Pipeline... so the requirements are lost, meaning a possibly-vital error check is now missing.

Thinking out loud: perhaps we should:
- move the `add_requirements` method to `Func`
- have `Pipeline` aggregate all the requirements in its output Funcs and emit them at the top of its block (taking care to remove duplicates)

@abadams -- what do you think?

abadams:
That might be weird in the case where the consumer only uses one of the Func outputs, and the requirement refers to things only used by the other Func output.

I guess we could have Func::add_requirement, and Pipeline::add_requirement would just add its requirement to the first output Func? Then they'd get lost if you only use the second output. All output funcs? Then you'd get potentially unwanted requirements again. Hrm.

steven-johnson:
> Hrm

Yeah, the solution isn't immediately obvious, but the status quo definitely feels like a problem too.

PR #7045, opened 2022-09-23
add_requirement() maintenance

steven-johnson:
This PR started out as a quick fix to add Python bindings for the `add_requirements` methods on Pipeline and Generator (which were missing), but expanded a bit to fix other issues as well:
- The implementation of `Generator::add_requirement` was subtly wrong, in that it only worked if you called the method after everything else in your `generate()` method. Now we accumulate requirements and insert them at the end, so you can call the method anywhere.
- We had C++ methods that took both an explicit `vector<Expr>` and also a variadic-template version, but the former required a mutable vector... and fixing this to not require that ended up creating ambiguity about which overloaded call to use. Added an ugly enable_if thing to resolve this.

While working on this, also identified Issue #7044, which I haven't yet attempted to address.

(Side note #1: overloading methods to have both templated and non-templated versions with the same name is probably something to avoid in the future.)

(Side note #2: we should probably thing more carefully about using variadic templates in our public API in the future; we currently use it pretty heavily, but it tends to be messy and hard to reason about IMHO.)

### DIFF
--- a/src/Generator.h
+++ b/src/Generator.h
@@ -3444,9 +3444,14 @@ public:
         return p;
     }
 
-    template<typename... Args>
-    HALIDE_NO_USER_CODE_INLINE void add_requirement(Expr condition, Args &&...args) {
-        get_pipeline().add_requirement(condition, std::forward<Args>(args)...);
+    void add_requirement(const Expr &condition, const std::vector<Expr> &error_args);
+
+    template<typename... Args,
+             typename = typename std::enable_if<Internal::all_are_printable_args<Args...>::value>::type>
+    inline HALIDE_NO_USER_CODE_INLINE void add_requirement(const Expr &condition, Args &&...error_args) {
+        std::vector<Expr> collected_args;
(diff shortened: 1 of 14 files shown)


src/Generator.h
```diff
@@ -3634,7 +3639,13 @@ class GeneratorBase : public NamesInterface, public AbstractGenerator {
     std::unique_ptr<GeneratorParamInfo> param_info_ptr;
 
     std::string generator_registered_name, generator_stub_name;
-    Pipeline pipeline;
+    Pipeline pipeline_;
```
abadams, line 3642: Why the new underscores? I thought we typically didn't do that for member variables.

abadams (approved):

src/Generator.h
steven-johnson: Ah, was trying to track down usage by doing the 'rename and see what fails to compile' trick. Will revert.

steven-johnson:
Failures are unrelated

PR #7046, opened 2022-09-23
Fix false positive use after free warning.

zvookin:
Latest clang is giving a use after free warning for printing a pointer value after it is freed. The compiler is not exactly wrong in that a pointer is being passed to something after it is freed, but the something is just printing the pointer's value for debugging purposes. Cast through an intptr_t to avoid the warning.


derek-gerstmann (approved):
Ahh, thanks for the heads up! Nice to see pointer usage tracking diagnostics in clang too!

steven-johnson (approved):

steven-johnson:
We should probably file a bug on this; that's totally legit code and should not be flagged

PR #7048, opened 2022-09-23
Allow call_intrin to call an LLVM intrinsic with void return type.

zvookin:
Specific use case is calling vector predicated store. Factored out of [the vector predication branch](https://github.com/halide/Halide/tree/llvm_vector_predication_intrinsics) for ease of code review and such.


abadams (approved):

steven-johnson (approved):
LGTM pending green

steven-johnson:
(but please fix clang-format)

Issue #5756, opened 2021-02-18
Improve StmtToHtml output


dsharletg:
We currently generate pretty bad HTML output:
- It's *huge*. I'm looking at a 10MB html file right now. The stmt file for the same pipeline is a few hundred KB.
- It brings chrome to its knees. Expanding/collapsing a node is slow and clunky.

I think there are ways to generate similar syntax highlighted and collapsible/expandable HTML that is far less bloated and easier for the browser to handle, but I only know enough HTML to be dangerous.

steven-johnson:
Is the HTML output regularly used? (Honest question; I *never* use it but I guess you must do so otherwise you wouldn't have noticed)

dsharletg:
I tend to not use it *because* it's not very good, I think due in large part to these issues. I just tried using it because I really needed it to understand a large piece of IR, and that's when I filed this issue.

abadams:
I use it for grokking large pipelines too, and it's really annoying how large and slow it is. I think a talented js person could make it substantially better.

steven-johnson:
Tagging as a good candidate for a contributor project

BachiLi:
I used it a lot when I was learning how Halide compilation works. Indeed it is less useful for larger pipelines.

abadams:
https://github.com/halide/Halide/wiki/%5BGSoC-2021%5D-A-(Better)-Interactive-Viewer-for-Halide-Pseudocode

dsharletg:
I just had an idea: maybe we could tweak the stmt syntax to make it more C-like (e.g. emit produce/consume as comments, make allocate a function call/return value, etc.), and then one could just tell a text editor to treat .stmt as .c.

Maybe just those things would make it close enough. Things like lets would be a problem but maybe not enough to break syntax highlighting so much it isn't useful.

rootjalex:
iirc, Fredo's class also recommends using it to understand Halide a bit, and students complained about it.

Also might be worth it to make a syntax highlighter that works on different IDEs for .stmt files?

alexreinking:
> Also might be worth it to make a syntax highlighter that works on different IDEs for .stmt files?

This is very easy and a good idea.

mcourteaux:
I just switched to latest Halide, and the HTML output has gotten these new features with the assembly, and this weird middle column. While the idea looks interesting with the three panes, this is EXTREMELY slow. Loading the page takes longer than compiling the pipeline.


[...]

PR #7056, opened 2022-09-28
New Feature: Halide Program IR Visualizer

darya-ver:
# New Feature: Halide Program IR Visualizer #

This PR is committing a new feature that generates an HTML page, with extension `.stmt.viz.html`, containing a visualization of Halide's IR. Here is an example of what the HTML page looks like: 

Screen Shot 2022-09-23 at 11 29 43 AM

You can also see the actual html page here: https://darya-ver.github.io/files/project_files/visualize_halide_example_blur3x3.html.

---
## File Changes ##

This PR is changing the following *existing files*:

* `Makefile`: added the new files into `SOURCE_FILES` and `HEADER_FILES`.

* `src/CodeGen_LLVM.h` // `src/CodeGen_LLVM.cpp`: added logic in `CodeGen_LLVM::visit(const ProducerConsumer *op)` and `CodeGen_LLVM::visit(const For *op)` to add unique identifiers to the markers in the generated LLVM code to more easily map the Halide IR code to the LLVM code.

* `src/Generator.cpp`: added logic for `-e stmt_viz` flag to generate the HTML page. Also added `assembly` flag if `stmt_viz` flag is present so that the assembly code is also generated.

* `src/Module.h` // `src/Module.cpp`: added new logic to call `Internal::print_to_viz()` if the `-e stmt_viz` flag is present.

This PR is adding the following *new files*:

* `src/StmtViz.h` // `src/StmtViz.cpp`: this is the main file that calls all other files. It contains the logic to generate the HTML page, and defines in the `print_to_viz()` function that is called from `src/Module.cpp`.

* `src/FindStmtCost.h` // `src/FindStmtCost.cpp`: added a new file that contains the logic to find the cost of a Halide statement. This is used to determine the color of the nodes in the visualization.

* `src/GetAssemblyInfoViz.h` // `src/GetAssemblyInfoViz.cpp`: added a new file that contains the logic to copy over the contents of the `.s` assembly file into a div in the HTML page, and finds the mapping of Halide IR to LLVM code.

* `src/GetStmtHierarchy.h` // `src/GetStmtHierarchy.cpp`: added a new file that contains the logic to generate the statement hierarchy of the nodes in the visualization.

* `src/IRVisualization.h` // `src/IRVisualization.cpp`: added a new file that contains the logic to generate the main visualization in the middle column.

---
## How to Navigate Webpage ##

Looking at the screenshot above, if you click the information button at the top right of the screen, you will see the following popup, which has information about what each column represents and what features exist within them. 

Screen Shot 2022-09-23 at 1 51 06 PM

---
## How to Generate HTML File ##

To use this new feature, your code needs to be a `Generator`. Once you compile the generator, run it with the `-e stmt_viz` flag. This will generate an HTML page in the directory that you've specified in the `-o` flag, with extension `.stmt.viz.html`.

For example, if you've compiled a generator called `my_generator` and you want to generate the HTML page in the `bin/` directory, you would run the following command:

```
./my_generator -e stmt_viz -o bin [rest of the flags]
```

where `[rest of flags]` are any another other flags that you need to add to correctly run the generator (like `-g`, `-f`, `target`, etc).

---
## TODO before merging into main ##

* Merge the `stmt_viz` and `stmt_html` flags into 1 so that we get rid of original IR visualization.


src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+int FindStmtCost::get_if_node_cost(const IRNode *op, bool inclusive, bool is_computation) const {
+    if (op->node_type != IRNodeType::IfThenElse) {
+        internal_error << "\n"
+                       << "FindStmtCost::get_if_node_cost: " << print_node(op)
+                       << "node is not IfThenElse"
+                       << "\n\n";
+    }
+
+    int cost;
+    const IfThenElse *if_then_else = dynamic_cast<const IfThenElse *>(op);
```
steven-johnson, line 196: Don't use dynamic_cast<> in Halide code -- we aren't always build with RTTI enabled, so this won't work -- and there is a *much* better way to do this with Halide IR nodes, namely, the `.as<>()` method (which is heavily optimized):

```
    const IfThenElse *if_then_else = op->as<IfThenElse>();  // nullptr if op is not of this type
```

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+        children_dmc += get_data_movement_cost(child, inclusive);
+    }
+
+    vector<int> costs_children = {children_cc, children_dmc};
+
+    return costs_children;
+}
+
+void FindStmtCost::set_costs(
+    bool inclusive, const IRNode *node, vector<const IRNode *> children,
```
steven-johnson, line 235: Do you really want to pass this vector by value here? Wouldn't it be preferable to pass as a const ref?

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+        }
+    }
+
+    // inclusive and exclusive costs are the same
+    set_costs(true, op, children);
+    set_costs(false, op, children);
+}
+
+void FindStmtCost::visit(const Let *op) {
+
```
steven-johnson, line 505: Style nit, here and many other places: the first line of a function body should never be a blank line.

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+    std::function<int(int)> calculate_cc = [count_cost](int children_cost) {
+        return count_cost * (NORMAL_NODE_CC + children_cost);
+    };
+
+    std::function<int(int)> calculate_dmc = [count_cost](int children_cost) {
+        return count_cost * (NORMAL_NODE_DMC + children_cost);
+    };
+
+    // inclusive and exclusive costs are the same
+    set_costs(true, op, {op->value.get()}, calculate_cc, calculate_dmc);
```
steven-johnson, line 541: The pattern of calling `set_costs` twice with the same args (except for the first) happens so many times that it seems worthy of its own little helper function.

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+    op->min.accept(this);
+    op->extent.accept(this);
+    op->body.accept(this);
+
+    current_loop_depth -= 1;
+
+    set_costs(true, op, {op->min.get(), op->extent.get(), op->body.get()});
+    set_costs(false, op, {op->min.get(), op->extent.get()});
+
+    // TODO: complete implementation of different loop types
```
steven-johnson, line 581: These should all be `internal_assert`

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+
+    set_costs(true, op, {op->body.get()});
+    set_costs(false, op, {});
+}
+
+string FindStmtCost::print_node(const IRNode *node) const {
+    stringstream s;
+    s << "Node in question has type: ";
+    IRNodeType type = node->node_type;
+    if (type == IRNodeType::IntImm) {
```
steven-johnson, line 819: Sure looks like a `switch` statement would be more appropriate here.

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,930 @@
...
+
+    stringstream name;
+    name << op->producer_name;
+
+    set_costs(true, op, {op->body.get()});
+    set_costs(false, op, {});
+}
+
+string FindStmtCost::print_node(const IRNode *node) const {
+    stringstream s;
```
steven-johnson, line 816: nit: `ostringstream` might be slightly better since it's output-only

src/FindStmtCost.h
```diff
@@ -0,0 +1,147 @@
...
+#include "Module.h"
+
+#include <stdexcept>
+#include <unordered_map>
+
+using namespace Halide;
+using namespace Internal;
+using namespace std;
+
+#define NORMAL_NODE_CC 1
```
steven-johnson, line 17: Prefer to use either `constexpr int` or `enum` for numeric constants; use #define only when unavoidable

src/FindStmtCost.h
```diff
@@ -0,0 +1,147 @@
...
+#include "Error.h"
+#include "ExternFuncArgument.h"
+#include "Function.h"
+#include "IRVisitor.h"
+#include "Module.h"
+
+#include <stdexcept>
+#include <unordered_map>
+
+using namespace Halide;
```
steven-johnson, line 13: `using namespace` in a .h file is strictly prohibited and should ~never be done.

src/FindStmtCost.h
```diff
@@ -0,0 +1,147 @@
...
+#define FINDSTMTCOST_H
+
+#include "Error.h"
+#include "ExternFuncArgument.h"
+#include "Function.h"
+#include "IRVisitor.h"
+#include "Module.h"
+
+#include <stdexcept>
+#include <unordered_map>
```
steven-johnson, line 11: Be aware that Halide needs to be able to compile with exceptions enabled *and* disabled -- and I don't see what this is used for. Was it included accidentally?

src/Generator.cpp
```diff
@@ -850,6 +850,15 @@ gengen
             output_types.insert(OutputFileType::registration);
             output_types.insert(OutputFileType::static_library);
         } else {
+            // if emit_flags contains "stmt_viz", add "assembly" to the list
+            for (const auto &f : emit_flags) {
```
steven-johnson, line 854: Hrm... not sure about this. I think it would be preferable to just fail with a message that specifying `stmt_viz` requires that `assembly` also be specified, if for no other reason than it makes it easier to integrate with Blaze/Bazel, which is very picky about undeclared files being generated. (Other Halide contributors may disagree here.)

src/GetAssemblyInfoViz.cpp
```diff
@@ -0,0 +1,196 @@
+
```
steven-johnson, line 1: nit: no blank lines at the top of a source file

src/GetAssemblyInfoViz.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+#include "GetAssemblyInfoViz.h"
+
+using namespace std;
```
steven-johnson, line 5: In general, don't use `using namespace` in Halide source at all, at least inside the `src/` directory. This code should probably all be in the `Halide::Internal` namespace anyway, so the usings would be redundant.

src/GetAssemblyInfoViz.cpp
```diff
@@ -0,0 +1,196 @@
...
+        }
+        assembly_file.close();
+    }
+
+    assembly_HTML << "</pre>\n";
+    assembly_HTML << "</div>\n";
+}
+
+string GetAssemblyInfoViz::get_assembly_filename(const string &filename) {
+    string assembly_filename = "./" + filename;
```
steven-johnson, line 71: Why do we want a `./` prefix here -- isn't a plain name preferable?

src/GetAssemblyInfoViz.cpp
```diff
@@ -0,0 +1,196 @@
...
+    for (auto &marker : producer_consumer_markers) {
+        add_line_number_prod_cons(assembly_line, marker, line_number);
+    }
+}
+
+void GetAssemblyInfoViz::add_line_number_for_loop(string &assembly_line,
+                                                  AssemblyInfoForLoop &marker, int line_number) {
+    // start of for loop
+    if (std::regex_search(assembly_line, marker.regex_start)) {
+
```
steven-johnson, line 89: nit, here and elsewhere: don't have a blank line at the start of a braced block

src/GetAssemblyInfoViz.cpp
```diff
@@ -0,0 +1,196 @@
...
+void GetAssemblyInfoViz::visit(const ProducerConsumer *op) {
+    producer_consumer_count++;
+
+    string assembly_marker = "%\"";
+    assembly_marker += to_string(producer_consumer_count);
+    assembly_marker += op->is_producer ? "_produce " : "_consume ";
+    assembly_marker += op->name;
+
+    // replace all $ with \$
+    std::regex dollar("\\$");
```
steven-johnson, line 133: Do you need to use `regex` for this? Seems like the `replace_all()` function in Util.h would probably do the same thing a bit more cheaply

src/GetStmtHierarchy.cpp
```diff
@@ -0,0 +1,724 @@
...
+        html << open_node(op, "atomic");
+        html << node_without_children(nullptr, op->mutex_name);
+        html << close_node();
+    }
+}
+
+string GetStmtHierarchy::generate_stmt_hierarchy_js() {
+    stringstream stmt_hierarchy_js;
+
+    stmt_hierarchy_js
```
steven-johnson, line 604: This is fine as-is, but it might be cleaner to use a C++11 raw string literal instead.

src/GetStmtHierarchy.cpp
```diff
@@ -0,0 +1,724 @@
...
+        << "    tooltip.addEventListener('blur', () => { \n"
+        << "        hideTooltip(tooltip); \n"
+        << "    } \n"
+        << "    ); \n"
+        << "} \n";
+
+    return stmt_hierarchy_js.str();
+}
+
+const string GetStmtHierarchy::stmt_hierarchy_css = "\n \
```
steven-johnson, line 629: (1) you definitely want a raw string literal here
(2) declaring this as `const char *` or `const char[]` instead of `std::string` would avoid unnecessary memory allocations

src/GetStmtHierarchy.cpp
```diff
@@ -0,0 +1,724 @@
...
+.tf-custom-stmtHierarchy .tf-nc { border-radius: 5px; border: 1px solid; font-size: 12px; border-color: rgb(200, 200, 200);} \n \
+.tf-custom-stmtHierarchy .end-node { border-style: dashed; font-size: 12px; } \n \
+.tf-custom-stmtHierarchy .tf-nc:before, .tf-custom-stmtHierarchy .tf-nc:after { border-left-width: 1px; border-color: rgb(200, 200, 200);} \n \
+.tf-custom-stmtHierarchy li li:before { border-top-width: 1px; border-color: rgb(200, 200, 200);}\n \
+.tf-custom-stmtHierarchy { font-size: 12px; } \n \
+div.nodeContent { display: flex; } \n \
+div.nodeName { padding-left: 5px; } \n \
+";
+
+const string GetStmtHierarchy::stmt_hierarchy_collapse_expand_JS = "\n \
```
steven-johnson, line 645: see previous comment

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+void GetReadWrite::generate_sizes(const Module &m) {
+    traverse(m);
+}
+
+StmtSize GetReadWrite::get_size(const IRNode *node) const {
+    auto it = stmt_sizes.find(node);
+
+    // errors if node is not found
+    if (it == stmt_sizes.end()) {
+        internal_error << "\n\nGetReadWrite::get_size - Node not found in stmt_sizes: "
```
steven-johnson, line 22: `internal_error` will *never* return, so the `return StmtSize()` is unnecessary

zvookin:
Wow. Very cool. Will try it out.

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+            read_write_table_ss << "</tr>";
+        }
+    }
+
+    // close table
+    read_write_table_ss << "</table>";
+
+    return read_write_table_ss.str();
+}
+string IRVisualization::allocate_table(vector<string> &allocation_sizes) const {
```
steven-johnson, line 549: nit: blank line between functions/methods

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+    stringstream allocate_table_ss;
```
steven-johnson, line 550: nit, here and elsewhere: probably prefer `ostringstream` for a stream that is only written to

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+    html << "<div class='functionViz'>";
+    func.body.accept(this);
+    html << "</div>";
+
+    html << close_function_box_div();
+}
+void IRVisualization::visit(const Variable *op) {
+    // if op->name starts with "::", remove "::"
+    string var_name = op->name;
+    if (var_name[0] == ':' && var_name[1] == ':') {
```
steven-johnson, line 828: ...what if var_name.size() < 2?

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+
+    html << allocate_div_header(op, header, anchor_name, table_rows);
+
+    op->body.accept(this);
+
+    html << close_box_div();
+}
+
+string IRVisualization::generate_ir_visualization_js() {
+    stringstream irVizJS;
```
steven-johnson, line 1243: `ir_viz_js` would really fit Halide style better

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+        ss << "Atomic type";
+    } else {
+        ss << "Unknown type";
+    }
+
+    return ss.str();
+}
+
+const string IRVisualization::scroll_to_function_JS_viz_to_code = "\n \
+// scroll to function - viz to code\n \
```
steven-johnson, line 1409: this also should probably be `const char *` and a raw string literal

src/IRVisualization.cpp
```diff
@@ -0,0 +1,1585 @@
...
+ \n \
+    // change content for 1 second   \n \
+    setTimeout(function () { \n \
+        scrollToObject.style.backgroundColor = 'transparent'; \n \
+        scrollToObject.style.fontSize = '12px'; \n \
+    }, 1000); \n \
+} \n \
+";
+
+const string IRVisualization::ir_viz_CSS = "\n \
```
steven-johnson, line 1443: ditto

src/IRVisualization.h
```diff
@@ -0,0 +1,189 @@
...
+#include "IRVisitor.h"
+
+#include <set>
+#include <unordered_map>
+
+using namespace std;
+using namespace Halide;
+using namespace Internal;
+
+#define MAX_CONDITION_LENGTH 35
```
steven-johnson, line 14: Don't use #define for numeric constants -- `constexpr` or `enum`, please

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2389 @@
...
+namespace {
+template<typename T>
+string to_string(T value) {
+    std::ostringstream os;
+    os << value;
+    return os.str();
+}
+
+class StmtToViz : public IRVisitor {
+
```
steven-johnson, line 32: nit: no blank lines at the start of a struct/class declaration

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2389 @@
...
+    StmtHierarchyInfo get_stmt_hierarchy_html(const Stmt &op) {
+        StmtHierarchyInfo stmt_hierarchy_info = get_stmt_hierarchy.get_hierarchy_html(op);
+        string &html = stmt_hierarchy_info.html;
+        string popup = generate_stmt_hierarchy_popup(html);
+        stmt_hierarchy_info.html = popup;
+
+        return stmt_hierarchy_info;
+    }
+    StmtHierarchyInfo get_stmt_hierarchy_html(const Expr &op) {
```
steven-johnson, line 114: nit: consistently put one blank line between method bodies

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2389 @@
...
+
+        tooltip_count++;
+        s << "<button class='iconButton assemblyIcon' ";
+        s << "id='button" << tooltip_count << "' ";
+        s << "aria-describedby='tooltip" << tooltip_count << "' ";
+        s << "onclick='populateCodeMirror(" << assembly_line_num_start << ", "
+          << assembly_line_num_end << ");'>";
+        s << "<svg xmlns='http://www.w3.org/2000/svg' width='16' height='16' fill='currentColor' "
+             "class='bi bi-filetype-raw' viewBox='0 0 16 16'>";
+        s << "<path fill-rule='evenodd' d='M14 4.5V14a2 2 0 0 1-2 2v-1a1 1 0 0 0 1-1V4.5h-2A1.5 "
```
steven-johnson, line 357: ... a comment about what this shape looks like would be nice :-)

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2389 @@
...
+    void visit(const FloatImm *op) override {
+        stream << open_span("FloatImm Imm");
+        stream << Expr(op);
+        stream << close_span();
+    }
+
+    void visit(const StringImm *op) override {
+        stream << open_span("StringImm");
+        stream << "\"";
+        for (unsigned char c : op->value) {
```
steven-johnson, line 464: `std::string` is defined as being `char`, so this is doing an implicit conversion to `unsigned char`. Presumably no compiler complains?

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2389 @@
...
+                    stream << "t";
+                    break;
+                case '\r':
+                    stream << "r";
+                    break;
+                case '\n':
+                    stream << "n";
+                    break;
+                default:
+                    string hex_digits = "0123456789ABCDEF";
```
steven-johnson, line 486: This is creating (and destroying) this string for *every character* you output. It should really be a `static const char [16]`.

src/IRVisualization.h
zvookin, line 14: Is it possible to use ```constexpr``` for these?

src/IRVisualization.h
```diff
@@ -0,0 +1,189 @@
...
+struct StmtSize {
+    map<string, string> writes;
+    map<string, string> reads;
+
+    bool empty() const {
+        return writes.size() == 0 && reads.size() == 0;
+    }
+};
+
+/*
```
zvookin, line 28: It would be great to do the doxygen style comments so the classes show up in the documentation.

src/IRVisualization.h
zvookin: And if these do not need to be exported, possibly move them into the .cpp file where they are used.

steven-johnson (commented):
At first glance, this looks amazing -- nice work! I did a quick skim, and have lots of style nits and a handful of real issues.  (Most important is probably that 'using namespace' should be avoided, *especially* in .h files, as it creates ambiguities in name resolution.)

There are a lot of long string literals containing HTML, JS, etc. -- I idly wonder if some/any/all of them would be better off as source-files-on-disk rather than embedded string literals, but don't have any strong feeling about that yet.

Also, we'll need to update the CMake build rules (those are actually the canonical rules) before landing.

steven-johnson:
(please run the `run-clang-format.sh` script to apply a lot of our style requirements)

steven-johnson:
Playing with the UI in Chrome, I can't seem to resize the column width by grabbing the green vertical dividers -- the cursor changes as though I can resize them there, but click-and-drag does nothing.

zvookin:
After a quick glance here, this looks fairly well structured and isolated from the rest of the code. I'm probably not the best person to comment on the internals, but I'm not seeing anything that worries me.

I believe the new files will need to be added to CMakeLists.txt.

src/FindStmtCost.h
alexreinking: And if they can be moved to a single TU, then they should be, too.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2389 @@
...
+                   << "\n"
+                   << "\n";
+}
+
+void print_to_viz(const string &filename, const Module &m) {
+
+    StmtToViz sth(filename, m);
+
+    sth.generate_html(filename, m);
+    cout << "Done generating HTML IR Visualization - printed to: " << filename << endl;
```
alexreinking, line 2385: This should be printed only at some more-verbose-than-default `debug()` level.

src/FindStmtCost.cpp
slomp: It's been a while, but it looks like `as<T>` is no longer an interface of `IRNode`; I see one in `IRHandle`, though.

slomp: Is it just a matter of locally wrapping he `IRNode*` into an `IRHandle` just to have access to the `as<T>()` interface?

abadams: get_if_node_cost should take a const IfThenElse *, and the cast should happen in the caller as a C-style cast. If you wrap it in an IRHandle to use `as` it'll do a pointless refcount.

src/FindStmtCost.h
```diff
@@ -0,0 +1,148 @@
...
+#include "Function.h"
+#include "IRVisitor.h"
+#include "Module.h"
+
+#include <unordered_map>
+
+namespace Halide {
+namespace Internal {
+
+constexpr int NORMAL_NODE_CC = 1;
```
abadams, line 15: This is the style we use for preprocessor macros, and the scope is a bit broad. I suggest moving these inside StmtCost and changing them to NormalNodeDMC and similar, so that usage looks like StmtCost::NormalNodeDMC

src/FindStmtCost.cpp
slomp: How about something like `string GetReadWrite::print_node(const IRNode *node) const { ... }`?

alexreinking:
Related issue: #5756

steven-johnson:
This is requesting to merge from an external fork into a local branch, so I don't think we can get the buildbots to run on this correctly. Ideally we should convert this into an 'ordinary' PR to merge it into Halide `main` branch, but I'll leave that to the original author to deal with lest I mess something up. In the meantime, I've made a sub-branch of this (with no additional changes) in https://github.com/halide/Halide/pull/7062 to get the buildbots to run.

slomp:
> This is requesting to merge from an external fork into a local branch, so I don't think we can get the buildbots to run on this correctly. Ideally we should convert this into an 'ordinary' PR to merge it into Halide `main` branch, but I'll leave that to the original author to deal with lest I mess something up. In the meantime, I've made a sub-branch of this (with no additional changes) in #7062 to get the buildbots to run.

Yup, the idea is to merge from a fork branch (Darya's own `StmtCost` feature branch) to the true Halide repo, but in a non-main branch (`darya-ver/ir-viz`). Once it's there and we address the first batch of reviews, we'll merge, and then create a PR from `darya-ver/ir-viz` to `main`, where we can get more comprehensive tests.

slomp:
@steven-johnson And thanks for running clang-tidy too :-)

src/FindStmtCost.cpp
```diff
@@ -0,0 +1,1025 @@
...
+                                                             << "FindStmtCost::get_if_node_cost: " << print_node(op)
+                                                             << "node is not IfThenElse"
+                                                             << "\n\n";
+
+    int cost;
+    const IfThenElse *if_then_else = dynamic_cast<const IfThenElse *>(op);
+
+    if (is_computation) {
+        if (inclusive) {
+            int computation_cost = get_computation_cost(if_then_else->condition.get(), inclusive) +
```
maaz139, line 194: Unclear why inclusive cost ignores ignores the else-case. Discussing the cost-model is perhaps tangential to this PR but we should add a comment explaining the intention here.

maaz139 (approved):
Addressed most of the comments from other reviews and re-ran the clang-tidy script.

I plan to do some more code revisions but I'll defer those for the next PR (into Halide main).

abadams:
Please don't merge yet. I want to do a pass.

abadams:
Two big things: 

There appear to be a bunch of unrelated wasm and python float16 changes in here.

The header files are missing file-level documentation and doxygen-style non-trivial class and method comments throughout.

abadams:
My other big concern is that I don't think computing and showing approximate costs is useful enough to justify the maintenance cost of the big chunk of code that computes it. I can't think of a case where I would find it helpful.

maaz139:
> There appear to be a bunch of unrelated wasm and python float16 changes in here.

I traced the changes in wasm and I believe they were pulled into Darya's fork from main. The `ir-viz` branch that this PR pushes the new code probably needs to be updated as well.

abadams:
Oh I mistakenly thought this was a PR into main. Yeah we can merge then continue to work on it.

PR #7059, opened 2022-10-04
Improve MSAN under JIT

steven-johnson:
Running JIT code under MSAN is (still) not officially supported, as it can (and does) give false positives; that said, this PR reduces the false positives in some cases.

Specifically:
(1) if we are building a JIT shared runtime (ie, we are jitting), and we are compiling with MSAN enabled (detected via preprocessor), we should always use the real MSAN stubs, even if the Target::MSan feature has been cleared (because JITModule::make_module() clears it). 
(2) Target::get_jit_from_environment() should add any detected sanitizer bits to the target sniffed from HL_JIT_TARGET

### DIFF
--- a/src/LLVM_Runtime_Linker.cpp
+++ b/src/LLVM_Runtime_Linker.cpp
@@ -1034,7 +1034,18 @@ std::unique_ptr<llvm::Module> get_initial_module_for_target(Target t, llvm::LLVM
                 }
             }
 
-            if (t.has_feature(Target::MSAN)) {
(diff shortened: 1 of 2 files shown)


steven-johnson:
Gentle review ping

abadams:
This is for when the Halide compiler itself is compiled under a sanitizer and you're executing JIT-compiled code? I'm surprised that works at all.

abadams (approved):

Issue #7063, opened 2022-10-04
tests for Mullapudi2016 aren't very good


steven-johnson:
The existing tests for the Mullapudi2016 autoscheduler aren't very useful: most of them basically just run and dump the schedule and loop nest to stdout (and maybe runtimes too), but don't attempt to validate them. Thus, they mostly amount to "can we produce a schedule without crashing and will it run in a non-crazy amount of time". This is better than nothing, I guess, but we can do better.

PR #7064, opened 2022-10-05
Autoscheduler test reorg, part 1

steven-johnson:
The end goal here is to move the tests for all autoschedulers into `test/autoschedulers/<name>`. This part handles the tests for the Mullapudi2016 autoscheduler:
- Moves from test/auto_schedule
- Silences a lot of the irrelevant noise that the tests emit to stdout (see also #7063)
- Ensure that the tests that run manual-and-auto tests actually check the times for plausibility (where "plausible" means "no worse than the current status quo")
- Changes the test group label from `auto_schedule` to `autoschedulers` (note that `auto_schedule` is retained for now but will be eliminated once everything is migrated)


test/autoschedulers/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(HalideTestHelpers)
```
alexreinking, line 1: This shouldn't be necessary as it's already included in the CMakeLists.txt above this one... why is it here?

test/autoschedulers/mullapudi2016/reorder.cpp
```diff
@@ -161,51 +164,48 @@ int main(int argc, char **argv) {
...
         double auto_time = run_test_1(true);
 
-        std::cout << "======================\n"
-                  << "Manual time: " << manual_time << "ms\n"
-                  << "Auto time: " << auto_time << "ms\n"
-                  << "======================\n";
-
-        if (auto_time > manual_time * slowdown_factor) {
-            fprintf(stderr, "Warning: Auto-scheduler is much much slower than it should be.\n");
+        const double slowdown_factor = 15.0;  // TODO: whoa
```
alexreinking, line 171: lol, I wonder which commit slowed this down by ~3x

alexreinking (changes requested):
Looks pretty good, just one oddity I want to address.

test/autoschedulers/CMakeLists.txt
alexreinking: If the answer is "removing it breaks things" then please open an issue and we can land this PR. Otherwise, remove this line.

test/autoschedulers/mullapudi2016/reorder.cpp
steven-johnson: We never bothered checking the results before, we just logged to stderr

alexreinking: Presumably it wasn't worse than 6x slower when it was written, though?

test/autoschedulers/CMakeLists.txt
steven-johnson: Copy-paste error, removed

alexreinking (approved):
LGTM

PR #7065, opened 2022-10-05
Autoscheduler test reorg, part 2

### LINKED ISSUE #4053: FR: reorganize the autoscheduler(s) in the codebase

steven-johnson:
Move the Li2018 tests to tests/autoschedulers. Drive-by fix to the Python test. Fixes #4053


alexreinking (approved):
This is a slam-dunk :+1:

alexreinking:
...but please add "Fixes" to the PR description for the relevant issues :slightly_smiling_face:

PR #7067, opened 2022-10-05
Autoscheduler test reorg, part 3

### LINKED ISSUE #4053: FR: reorganize the autoscheduler(s) in the codebase

steven-johnson:
Fixes #4053


src/autoschedulers/adams2019/CMakeLists.txt
```diff
@@ -128,27 +66,15 @@ add_executable(adams2019_weightsdir_to_weightsfile weightsdir_to_weightsfile.cpp
...
-    target_link_libraries(adams2019_test_apps_autoscheduler PRIVATE Halide::Halide Halide::Tools ${CMAKE_DL_LIBS})
-
-    add_adams2019_test(adams2019_test_apps_autoscheduler
-                       COMMAND adams2019_test_apps_autoscheduler $<TARGET_FILE:Halide_Adams2019> ${CMAKE_CURRENT_SOURCE_DIR}/baseline.weights
-                       LABELS multithreaded
-                       ENVIRONMENT "LD_LIBRARY_PATH=$<TARGET_FILE_DIR:Halide_Adams2019>:$ENV{LD_LIBRARY_PATH}")
-endif ()
-
-##
+# Tests for private/internal functionality of Adams2019 (vs for public functionality,
```
alexreinking, line 69: These should _still_ be moved below the `tests/` tree somewhere. At the very least, these shouldn't be built if the user says `-DWITH_TESTS=OFF`

test/autoschedulers/adams2019/CMakeLists.txt
```diff
@@ -0,0 +1,84 @@
...
+    set_tests_properties(${NAME}
+                         PROPERTIES
+                         LABELS "${ARGS_LABELS}"
+                         ENVIRONMENT "${ENVIRONMENT}")
+endfunction()
+
+
+##
+# Tests and demos
+# TODO(#4053): move these to a separate folder since they're tests.
```
alexreinking, line 37: Is this still todo? :slightly_smiling_face:

alexreinking (changes requested):

src/autoschedulers/adams2019/CMakeLists.txt
steven-johnson: > These should _still_ be moved below the `tests/` tree somewhere. 

Should they? It seems weird to have tests for non-exported, internal-only stuff in a different hierarchy

> At the very least, these shouldn't be built if the user says `-DWITH_TESTS=OFF`

Agreed, fixing

alexreinking (approved):

PR #7080, opened 2022-10-11
Fix issue 7076

steven-johnson:


steven-johnson:
PTAL

steven-johnson:
Gentle review ping

src/ScheduleFunctions.cpp
```diff
@@ -2243,7 +2287,14 @@ bool validate_schedule(Function f, const Stmt &s, const Target &target, bool is_
         err << "Func \"" << f.name() << "\" is computed at the following invalid location:\n"
             << "  " << schedule_to_source(f, store_at, compute_at) << "\n"
             << "Legal locations for this function are:\n";
-        for (const auto &site : sites) {
+        for (size_t i = 0; i < sites.size(); i++) {
+            const auto &site = sites[i];
+            if (invalid_sites.count(i)) {
+                if (debug::debug_level() > 0) {
```
abadams, line 2293: Was this intentionally left in?

src/ScheduleFunctions.cpp
```diff
@@ -2224,9 +2226,51 @@ bool validate_schedule(Function f, const Stmt &s, const Target &target, bool is_
...
+        string compute_at_fname = sites[compute_idx].loop_level.func();
+        int possibly_invalid_idx = compute_idx;
+        for (size_t i = compute_idx + 1; i < sites.size(); i++) {
+            if (!sites[i].is_gpu_block) {
+                continue;
+            }
+            string site_fname = sites[i].loop_level.func();
+            if (site_fname == compute_at_fname) {
+                err << "Functions that are compute_at() a gpu_block() loop must specify the innermost gpu_block() loop for that Func.\n";
+                invalid_sites.insert(possibly_invalid_idx);
```
abadams, line 2265: I'm confused. Why keep a set of invalid sites, as opposed to just removing them from the set of valid sites?

steven-johnson: When I first wrote this I thought it might be confusing to the user to just leave out those locations (since they would be valid for cpu code), and thought maybe they should be listed specially for that reason (i.e., "These locations aren't usable with GPU because blah blah")... but maybe I'm overthinking it. I'll simplify it as you suggest.

src/ScheduleFunctions.cpp
steven-johnson: Yes, though it probably should be pruned (see comment below)

src/ScheduleFunctions.cpp
steven-johnson: (but there actually was a good reason to make a list of invalid sites: if we prune the invalid sites as we find them, we might invalidate the indices for for `compute_idx` and `store_idx`)

steven-johnson:
PTAL

steven-johnson:
Monday morning review ping

Issue #7084, opened 2022-10-14
We need a simd_op_check equivalent for GPU targets


steven-johnson:
To verify that (e.g.) dot-product operations are generated when expected, etc

PR #7085, opened 2022-10-14
Generate dot() in the Metal backend

vksnk:
Basically, it will generate dot() call for vector_reduce(Add, mul(float, float)). I've tested it locally to make sure it is actually generated. It'd be nice to have something similar to simd_op_check for GPU targets, but  it doesn't exist (#7084).

### DIFF
--- a/src/CodeGen_Metal_Dev.cpp
+++ b/src/CodeGen_Metal_Dev.cpp
@@ -225,7 +225,7 @@ void CodeGen_Metal_Dev::CodeGen_Metal_C::visit(const Min *op) {
 }
 
 void CodeGen_Metal_Dev::CodeGen_Metal_C::visit(const VectorReduce *op) {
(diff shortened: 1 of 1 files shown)


steven-johnson:
Monday morning review ping

shoaibkamil:
We also should think longer-term how we want to do these kinds of pattern matches.  @rootjalex @abadams In the future, should we consider moving some of these kinds of rules into earlier passes and do them via pattern-matching?

steven-johnson:
Marking for backport to release/15.x

rootjalex:
> I was planning on doing this for ARM/HVX as well

And I know HVX already has HexagonOptimize - but I want to turn some of those passes into proper term-rewriting systems - the current model of "exact pattern goes to specific intrinsic" is rather restrictive and does not support many of the rules that my project has generated.

PR #7098, opened 2022-10-17
Attempt to fix pip build issues

steven-johnson:


.github/workflows/pip.yml
```diff
@@ -4,11 +4,13 @@
 
 name: Build PyPI package
 
-on:
-  push:
-    branches: [ main ]
-  release:
-    types: [ created ]
+#on:
```
vksnk, line 7: Is this commented out code needed?

vksnk (approved):

.github/workflows/pip.yml
steven-johnson: It is (unfortunately) the only way I've found to test GitHub Actions (ie, by temporarily changing the trigger action to allow it to actually run before checking it in). I should have done this before requesting approval, sorry.

PR #7102, opened 2022-10-18
Remove API wart on call_intrin

steven-johnson:
The optional `scalable_vector_result` arg on one of the overloads of `call_intrin()` was apparently completely redundant, as it could (and arguably should) be inferred from `result_type` instead (as one call site did). Note the one TODO that I think is correct but needs evaluation. 

Also did drive-by inlining of args-vectors to hopefully reduce temporaries.


steven-johnson:
PTAL

PR #7109, opened 2022-10-20
Python tutorial cleanup

steven-johnson:
(This PR is not yet complete; opening as a Draft PR to get discussion and input)

The tutorials for the python bindings are in need of a lot of love. As of right now, this PR does this:

- Fix horribly wrong usage instructions
- Format everything for PEP8 (mostly, though some >80c lines are there)
- Fix wrong comments leftover from C/C++ port
- Ensure images are installed so the tutorials work there too as-is
- Updated imageio usage to avoid deprecation warnings
- Add README_python.md to the install

There are a couple of gotchas here, though:
(1) A lot of the tutorials were never ported to Python; we only go thru 14, while C++ goes up to 21. This isn't likely a big deal, porting them is mostly plug-n-chug. But...
(2) The structure of the stuff in the tutorials relating to AOT and Generators needs a complete overhaul. When the tutorials were first written, Generators didn't exist, so lesson 10 talks about AOT via direct calls to `compile_to_file()`; lesson 11 is about cross-compilation, using the same approach; lesson 12 is about GPU usage, and as a convenience there, it uses a generator-like class that is not a Generator. 

At this point in time we really want to steer people towards Generators for AOT use, so this PR actually removes lessons 10-12 from Python entirely for now. What I think we (well, I) should do here is this:
- remove existing lessons 10-12
- move the existing lesson 15 (Generators) to lesson 10 and bring it up to date, focusing only on JIT usage (since all the tutorials up to then are JIT)
- make lesson 11 be about AOT (via Generators)
- make lesson 12 be about cross-compilation
- make lesson 13 be about GPU
- renumber the rest of the lessons appropriately

I think this would be a more useful ordering of lessons for recent Halide, but it would (at least initially) diverge from the C++ tutorials in terms of the numbering, etc. (In the medium-to-long term we would bring the C++ versions up to date with the Python versions.) 

What do folks think?


steven-johnson:
Note that as part of this, I updated the use of `imageio`... it seems that as of imageio v2.16, they offer new and old apis, which must be imported as `imageio.v3` or `imageio.v2` respectively; if you just import and use the naked `imageio` but have 2.16+ installed, you get a noisy and puzzling deprecation warning emitted to stdout, which sucks for tutorials.

All the options here seem bad:
- We could just leave stuff as-is, but that would mean that folks running with a newer (Feb 2022 +) install of imageio would get a distracting deprecation warning
- We could play nicely with both older and newer versions of imageio by wrapping the import in a `try` block; that would work, but would be a distracting wart at the top of each tutorial
- We could settle on using either the v2 or v3 API, but that means that folks with older imageio installs fail outright

For the moment, I settled on the last one (using v2)... which forced me to upgrade the version on all the buildbots, since they were mostly running 2.9.

I'd welcome feedback on this.

python_bindings/tutorial/lesson_10_generators_1.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # The result is a Func, so we can just realize() it like any other:
+    output_buffer = output_func.realize([w, h])
+
+    for yy in range(h):
+        for xx in range(w):
+            assert output_buffer[xx, yy] == 192
+
+
+# If you like, you can put multiple Generators in the same source file. This
```
abadams, line 123: Are the first two sentences in this paragraph still helpful in the python context?

python_bindings/tutorial/lesson_10_generators_1.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def use_second_generator():
+    # We can override values for GeneratorParams by specifying an optional by-name-only argument,
+    # generator_params, to the .call() method. It takes a simple Python dict as its value.
+
+    # We'll read in an image here so that the effects of the code are easier to see.
+
+    im = imageio.imread(os.path.join(os.path.dirname(__file__), "images/gray.png"))
+    # We need to ensure that the image is arranged in 'F' order, so that the innermost axis
```
abadams, line 217: I think this paragraph is potentially a big point of confusion. Normally I use the default C order for numpy arrays and just expect the indexing to be reversed between Halide vs numpy, so numpy_array[y, x] == halide_buffer[x, y], in the same way that we have c_array[y][x] == halide_buffer[x, y] when using C.

IIUC, F order only changes how the indices are interpreted when using numpy's operator[], so this is unrelated to whether or not things vectorize cleanly in Halide, unless we're matching indexing order between the two, which we don't do in C.

abadams: Specifically, I think this version is confusing because people expect to write numpy_array[row, col], and in lesson 1 we explain that in Halide we index as [col, row]. If they try to index the numpy_array as [col, row] here as they're used to, they'll get the transpose of what they want.

python_bindings/tutorial/lesson_10_generators_1.py
steven-johnson: Nope. Removing.

python_bindings/tutorial/lesson_10_generators_1.py
steven-johnson: > this paragraph is potentially a big point of confusion

I don't disagree, but am not sure what better options we have:
- because of the rotation, I wanted an image that would be obvious from inspection that it 'worked'
- imageio.imread() apparently insists on reading grayscale images (or at least *this* one) gets read with stride=1 on the outmost axis rather than the innermost, so we can't vectorize() in the usual way with this image
- I suppose I could upgrade to the RGB image, it would just make the example more verbose?
- Or maybe just construct a buffer in-memory and draw a simple arrow or gradient on it?

abadams: imread loads an image such that the first row of the image is dense in memory.

Halide with default memory layouts and indexing conventions also expects the first row of the image to be dense in memory.

These should be compatible. My suspicion is that the numpy array -> Halide buffer conversion is converting the numpy strides/sizes arrays in order whereas it should be flipping the order.

abadams: My guess is asking numpy for 'F' order just permutes the numpy arrays sizes/strides array accordingly.

anuejn:
What's the state of this? Can I help to bring this PR over the finish line? I have the feeling that already merging this as-is is a huge improvement over the status quo and would have made my life starting with halides python bindings a lot easier :)

abadams:
Steven is currently away, so it's unlikely to get any attention from him in the next month or so. If you'd like to take it over and get it over the finish line, that would be welcome.

anuejn:
What would be needed for this? Just a rebase to the current main?

abadams:
For a start yeah, but the thing it really needs is for a python user of Halide to go through it and check it all makes sense.

PR #7111, opened 2022-10-21
Add support for generating LLVM vector predication intrinsics.

zvookin:
LLVM now has intrinsics for length bounded, predicated, vector operations. These are a generalization of the current vector IR LLVM has long had and will likely become the preferred way to generate pretty much all vector code in the future. (With appropriately restricted arguments to target existing fixed width SIMD instruction sets.)

Code has been tested in the riscv_update branch and will be used by the RISC V backend. PR has been separated out to make for a smaller, easier to review PR.

This required adding some helper methods to CodeGen_LLVM as the intrinsics have specific type mangling and argument pattterns that can be handled in common code.


src/CodeGen_LLVM.h
```diff
@@ -593,9 +652,9 @@ class CodeGen_LLVM : public IRVisitor {
     llvm::Function *add_argv_wrapper(llvm::Function *fn, const std::string &name,
                                      bool result_in_argv, std::vector<bool> &arg_is_buffer);
 
-    llvm::Value *codegen_dense_vector_load(const Type &type, const std::string &name, const Expr &base,
-                                           const Buffer<> &image, const Parameter &param, const ModulusRemainder &alignment,
-                                           llvm::Value *vpred = nullptr, bool slice_to_native = true);
+    llvm::Value *codegen_vector_load(const Type &type, const std::string &name, const Expr &base,
```
steven-johnson, line 655: At some point in the future we need to try to clean up this and related APIs; IMHO having 6-9 args (w/ the last three having defaults) makes for unreasonably hard to read code at the call site (but that is of course irrelevant for this PR)

src/CodeGen_LLVM.h
```diff
@@ -552,6 +562,55 @@ class CodeGen_LLVM : public IRVisitor {
...
+
+    /** Generate an intrisic call if use_llvm_vp_intrinsics is true
+     * and length is greater than 1. If generated, assigns result
+     * of vp intrinsic to value and returns true if it an instuction
+     * is generated, otherwise returns false. */
+    bool try_vector_predication_intrinsic(const std::string &name, llvm::Type *llvm_result_type,
+                                          int32_t length, MaskVariant mask, std::vector<VPArg> args);
+
+    /** Controls use of vector predicated intrinsics for vector operations. */
+    bool use_llvm_vp_intrinsics;
```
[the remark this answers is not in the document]
steven-johnson: is this ever expected to vary during the life of the codegen? if not, I'd be tempted to declare it 'const' and make it an optional arg to the CG_LLVM ctor, but maybe I'm overthinking it...

src/CodeGen_LLVM.cpp
```diff
@@ -1951,7 +2032,7 @@ void CodeGen_LLVM::visit(const Load *op) {
 
         llvm::Type *load_type = llvm_type_of(op->type.element_of());
         if (ramp && stride && stride->value == 1) {
-            value = codegen_dense_vector_load(op);
+            value = codegen_dense_vector_load(op, nullptr);
```
steven-johnson, line 2035: (1) nullptr is the default for `vpred` so this isn't necessary
(2) if it's desirable to be explicit anyway, please comment a la `/*vpred*/ nullptr`

src/CodeGen_LLVM.cpp
```diff
@@ -2006,8 +2087,8 @@ void CodeGen_LLVM::visit(const Load *op) {
                 int lanes_i = std::min<int>(slice_lanes, op->type.lanes() - i);
                 Expr slice_base = simplify(base + load_base_i);
 
-                Value *load_i = codegen_dense_vector_load(op->type.with_lanes(load_lanes_i), op->name, slice_base,
-                                                          op->image, op->param, align, nullptr, false);
+                Value *load_i = codegen_vector_load(op->type.with_lanes(load_lanes_i), op->name, slice_base,
```
steven-johnson, line 2090: (same comment as previous... and perhaps add comments to the pre-existing `nullptr` and `false` args too)

src/CodeGen_LLVM.cpp
```diff
@@ -4075,101 +4237,137 @@ void CodeGen_LLVM::codegen_vector_reduce(const VectorReduce *op, const Expr &ini
...
+                    name = "fmul";
                     takes_initial_value = true;
                     if (!initial_value.defined()) {
                         initial_value = make_one(op->type);
                     }
                     break;
                 case VectorReduce::Min:
-                    name << "fmin";
+                    name = "fmin";
+                    // TODO(zvookin): For signed case, whether this is Inf or the max floating-point value depends on strict_float. (Or maybe it is QNaN in strict_float.)
```
steven-johnson, line 4276: Do these need Issues opened?

src/CodeGen_LLVM.cpp
```diff
@@ -4075,101 +4237,137 @@ void CodeGen_LLVM::codegen_vector_reduce(const VectorReduce *op, const Expr &ini
...
                     break;
                 default:
                     break;
                 }
             }
-            name << ".v" << val.type().lanes() << (op->type.is_float() ? 'f' : 'i') << bits;
 
-            string intrin_name = name.str();
+            if (use_llvm_vp_intrinsics) {
+                string vp_name = "llvm.vp.reduce.";
```
steven-johnson, line 4323: Dumb q, why not just `string vp_name = "llvm.vp.reduce." + name;`?

src/CodeGen_LLVM.cpp
```diff
@@ -4935,5 +5134,108 @@ llvm::Constant *CodeGen_LLVM::get_splat(int lanes, llvm::Constant *value,
     return ConstantVector::getSplat(ec, value);
 }
 
+std::string CodeGen_LLVM::mangle_llvm_vector_type(llvm::Type *type) {
```
steven-johnson, line 5137: Would be worth a short comment here explaining what the format looks like, with some examples

steven-johnson (approved):
LGTM with various style comments.

src/CodeGen_LLVM.cpp
```diff
@@ -2352,12 +2437,30 @@ llvm::Value *CodeGen_LLVM::codegen_dense_vector_load(const Type &type, const std
...
         Value *vec_ptr = builder->CreatePointerCast(elt_ptr, slice_type->getPointerTo());
 
-        Instruction *load_inst;
-        if (vpred != nullptr) {
-            Value *slice_mask = slice_vector(vpred, i, slice_lanes);
-            load_inst = builder->CreateMaskedLoad(slice_type, vec_ptr, llvm::Align(align_bytes), slice_mask);
+        Value *slice_mask = (vpred != nullptr) ? slice_vector(vpred, i, slice_lanes) : nullptr;
+
+        Instruction *load_inst = nullptr;
+        if (stride) {
```
abadams, line 2443: I'm confused. Based on the calling code, stride is non-null if it's a compile-time constant, which means we're calling the strided load intrinsic when the stride is the constant 1 (i.e. a dense vector load), and we're calling llvm.vp.load otherwise (e.g. a load vectorized across y)?

src/CodeGen_LLVM.h
```diff
@@ -464,9 +473,10 @@ class CodeGen_LLVM : public IRVisitor {
                              llvm::Function *intrin, std::vector<Expr>);
     llvm::Value *call_intrin(const llvm::Type *t, int intrin_lanes,
```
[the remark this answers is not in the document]
zvookin: Looking at the PR now. I didn't realize there was one and thought we'd decided to wait on changes to this API. My apologies.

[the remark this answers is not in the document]
steven-johnson: My PR is super-low-priority, just ignore it for now, I'll update it later.

src/CodeGen_LLVM.cpp
zvookin: Added detail in comment on the declaration in the header.

src/CodeGen_LLVM.cpp
zvookin: I modified the calling location to not do this for the dense case. Ideally it doesn't matter as LLVM should strength reduce the vector predication instructions for constant arguments, but some experiments should it isn't very good about it for constant one masks and constant length that matches the fixed SIMD width on e.g. x86.

PR #7119, opened 2022-10-25
RISC V vector predication support intrinsics support

zvookin:
Have RISC V use LLVM vector predication intrinsics and add support for some architecture specific intrinsics.


src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+    llvm::Function *wrapper =
+        llvm::Function::Create(wrapper_ty, llvm::GlobalValue::InternalLinkage,
+                               wrapper_name, module.get());
+    llvm::BasicBlock *block =
+        llvm::BasicBlock::Create(module->getContext(), "entry", wrapper);
+    llvm::IRBuilderBase::InsertPoint here = builder->saveIP();
+    builder->SetInsertPoint(block);
+
+    // Set vector fixed-point rounding flag if needed for intrinsic.
+    bool round_down = intrin.flags & RISCVIntrinsic::RoundDown;
```
steven-johnson, line 338: Do we need explicit `() != 0` here? The current flag values are fine, but if we ever get bits above the first 8, not sure if the language promises this to work :-/

src/CodeGen_RISCV.cpp
```diff
@@ -1,24 +1,114 @@
...
+struct RISCVIntrinsic {
+    const char *riscv_name;
+    FixedOrScalableVector ret_type;
+    const char *name;
+    FixedOrScalableVector arg_types[max_intrinsic_args];
+    int flags;
+    enum {
+        AddVLArg = 1 << 0,          // Add a constant full size vector length argument
+        RoundDown = 1 << 1,         // Set rounding mode to down (rdn) before intrinsic.
+        RoundUp = 1 << 2,           // Set rounding mode to up (rdu) before intrinsic.
```
rootjalex, line 65: Could you provide a link in the comments that points to documentation of the rounding modes?

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+    // Set vector fixed-point rounding flag if needed for intrinsic.
+    bool round_down = intrin.flags & RISCVIntrinsic::RoundDown;
+    bool round_up = intrin.flags & RISCVIntrinsic::RoundUp;
+    if (round_down || round_up) {
+        internal_assert(!(round_down && round_up));
+        llvm::Value *rounding_mode = llvm::ConstantInt::get(xlen_type, round_down ? 2 : 0);
+        // TODO: When LLVM finally fixes the instructions to take rounding modes,
+        // this will have to change to passing the rounding mode to the intrinsic.
+        llvm::FunctionType *csrw_llvm_type = llvm::FunctionType::get(void_t, {xlen_type}, false);
+        llvm::InlineAsm *inline_csrw = llvm::InlineAsm::get(csrw_llvm_type, "csrw vxrm,${0:z}", "rJ,~{memory}", true);
```
steven-johnson, line 346: ...I had no idea you could do inline asm this way in LLVM!

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+        builder->CreateCall(inline_csrw, {rounding_mode});
+    }
+
+    // Call the LLVM intrinsic.
+    // TODO: This does not work for reductions.
```
steven-johnson, line 351: Should we open an Issue for this? Or add `internal_assert(not-a-reduction);`?

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+
+    // Always inline these wrappers.
+    wrapper->addFnAttr(llvm::Attribute::AlwaysInline);
+
+    builder->restoreIP(here);
+
+    wrapper->addFnAttr(llvm::Attribute::ReadNone);
+    wrapper->addFnAttr(llvm::Attribute::NoUnwind);
+
+    // llvm::verifyFunction(*wrapper);
```
steven-johnson, line 367: Is this meant to be commented out?

steven-johnson (commented):
Looks good, mostly nits

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+    {"vaaddu", Type::UInt, "halving_add", {Type::UInt, Type::UInt}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::RoundDown},
+    {"vaadd", Type::Int, "rounding_halving_add", {Type::Int, Type::Int}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::RoundUp},
+    {"vaaddu", Type::UInt, "rounding_halving_add", {Type::UInt, Type::UInt}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::RoundUp},
+    {"vwadd", {Type::Int, 2}, "widening_add", {Type::Int, Type::Int}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::MangleReturnType},
+    {"vwaddu", {Type::UInt, 2}, "widening_add", {Type::UInt, Type::UInt}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::MangleReturnType},
+    {"vwsub", {Type::Int, 2}, "widening_sub", {Type::Int, Type::Int}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::MangleReturnType},
+    {"vwsubu", {Type::UInt, 2}, "widening_sub", {Type::UInt, Type::UInt}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::MangleReturnType},
+    {"vwmul", {Type::Int, 2}, "widening_mul", {Type::Int, Type::Int}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::MangleReturnType},
+    {"vwmulu", {Type::UInt, 2}, "widening_mul", {Type::UInt, Type::UInt}, RISCVIntrinsic::AddVLArg | RISCVIntrinsic::MangleReturnType},
```
rootjalex, line 200: Does LLVM currently offer an intrinsic for `vwmulsu`?

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+
+llvm::Function *CodeGen_RISCV::define_riscv_intrinsic_wrapper(const RISCVIntrinsic &intrin,
+                                                              int bit_width_scale) {
+    int effective_vscale = target_vscale();
+
+    llvm::Type *xlen_type = target.bits == 32 ? i32_t : i64_t;
+
+    // Produce intrinsic name and
+    std::vector<llvm::Type *> llvm_arg_types;
+    std::string mangled_name = "llvm.riscv.";
```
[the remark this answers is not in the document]
zvookin: Because that will not compile without casting one of the ```const char *``` values to ```std::string``` and if I have to mention the type again anyway, this strikes me as a bit clearer.

src/CodeGen_RISCV.cpp
zvookin: Conversion to ```bool``` is a comparison against zero. If the value is zero, it's false, otherwise true. Normally I would add the comparison explicitly, but for this usage, it feels idiomatic to treat the ```&``` operator as set inclusion. As in I feel adding the explicit test makes the code harder to read, not easier.

src/CodeGen_RISCV.cpp
zvookin: It introduces a fairly significant code motion barrier, at least for this usage, and thus is highly undesirable. I"m pretty sure this is what inline assembly in C/C++ turns into in LLVM IR.

src/CodeGen_RISCV.cpp
zvookin: Reductions are now handled through the vector predication intrinsics, hence the reason I had that TODO there is obsolete. (It was tracking that I needed to figure out a way to make it work.) There are likely more to be handled, but it doesn't really need an assert as there's no way to specify an intrinsic that does a reduction in this part of the code. This is a thing to consider when revamping ```call_intrin```, but I don't think it needs a tracking issue as there is an ```is_reduction``` argument now. (I.e. I probably fixed the underlying issue there too.)

src/CodeGen_RISCV.cpp
```diff
@@ -1,24 +1,114 @@
...
 using std::string;
+using std::vector;
 
 #if defined(WITH_RISCV)
 
 namespace {
 
+constexpr int max_intrinsic_args = 4;
+
+struct FixedOrScalableVector {
```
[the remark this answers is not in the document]
zvookin: I changes the name to something that is not much shorter but is more descriptive.

rootjalex:
@steven-johnson Do the buildbots test RISC-V cross-compilation?

If so, this PR should probably make an addition to simd_op_check, to check for the added vector instructions in this PR.

src/CodeGen_RISCV.cpp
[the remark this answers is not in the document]
steven-johnson: Gotcha

steven-johnson:
> @steven-johnson Do the buildbots test RISC-V cross-compilation?
> 
> If so, this PR should probably make an addition to simd_op_check, to check for the added vector instructions in this PR.

Not currently, no. We should definitely add that.

src/CodeGen_RISCV.cpp
zvookin: Not in inline comments as it's pretty long and to a PDF so I can't (easily) link directly to the section anyway. Strikes me as the sort of thing where it's pretty obvious what they do and one has to know a bit about the RISC V vector architecture to work on the code at all. I added "vxrm" to make it clear these are the fixed-point rounding mode.

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+    builder->SetInsertPoint(block);
+
+    // Set vector fixed-point rounding flag if needed for intrinsic.
+    bool round_down = intrin.flags & RISCVIntrinsic::RoundDown;
+    bool round_up = intrin.flags & RISCVIntrinsic::RoundUp;
+    if (round_down || round_up) {
+        internal_assert(!(round_down && round_up));
+        llvm::Value *rounding_mode = llvm::ConstantInt::get(xlen_type, round_down ? 2 : 0);
+        // TODO: When LLVM finally fixes the instructions to take rounding modes,
+        // this will have to change to passing the rounding mode to the intrinsic.
```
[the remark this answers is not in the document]
abadams: Is there an existing LLVM issue for it? Having to use inline assembly to use any of the ops that rely on a rounding flag seems ... unfinished.

zvookin:
Added an issue for simd_op_check coverage. https://github.com/halide/Halide/issues/7122

zvookin:
Yeah, I'd rant about not having the rounding mode flags in the intrinsic at
the start is badly broken, but it's sort of how things are with LLVM and
RISC V. Really they ought to be in the opcode, but you know, bits are
expensive. I put in a reference to the one PR on LLVm and will dig up
another in a bit.

src/CodeGen_RISCV.cpp
zvookin: Added. Wasn't sure if there is normalization of the argument ordering and didn't see it happening anywhere so I added support for handling either order.

zvookin:
Had to do a number of bug fixes on the vector predication work as I messed up on the test coverage previously. It works better with the code in this branch...

zvookin:
I believe I have addressed the feedback raised. In a couple cases, by opening tracking issues, but mostly by incorporating the changes suggested or fixing the problems raised. If I've missed anything please let me know. I also did some better testing. Obviously that should get rolled into the real tests, which is effectively the simd_op_check issue. This should be ready to land.

steven-johnson (approved):
LGTM pending green

src/CodeGen_RISCV.cpp
rootjalex: There isn’t on main. I found it useful to add normalization on a project branch I’m working on, I’ll probably make a PR to add that and clean up some patterns later.

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+};
+
+void CodeGen_RISCV::init_module() {
+    CodeGen_Posix::init_module();
+
+    int effective_vscale = target_vscale();
+    if (effective_vscale != 0) {
+        for (const RISCVIntrinsic &intrin : intrinsic_defs) {
+            if (intrin.ret_type.type_pattern == FixedOrScalableVector::WildcardWidths) {
+                for (int scale = 0; scale < 4; scale++) {
```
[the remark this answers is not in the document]
rootjalex: @zvookin a comment stating something like “iterate through 8, 16, 32, 64 bit return types” could be helpful here

rootjalex (approved):
Also LGTM with a nit

src/CodeGen_RISCV.cpp
```diff
@@ -97,7 +187,188 @@ int CodeGen_RISCV::target_vscale() const {
...
+                                                                         target.vector_bits));
+                    }
+                    llvm::Function *intrin_impl = define_riscv_intrinsic_wrapper(intrin, bit_width_scale);
+                    declare_intrin_overload(intrin.name, ret_type, intrin_impl, arg_types);
+                }
+            } else {
+                llvm::Function *intrin_impl = define_riscv_intrinsic_wrapper(intrin, 1);
+                Type ret_type = concretize_fixed_or_scalable(intrin.ret_type, 1,
+                                                             target.vector_bits);
+                std::vector<Type> arg_types;
```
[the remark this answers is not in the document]
zvookin: Refactored to avoid allocating the vector for each intrinsic.

PR #7124, opened 2022-10-25
Avoid deprecation warnings in imageio

steven-johnson:
(Cherry-picked from WIP #7109)

It seems that as of imageio v2.16, they offer new and old apis, which must be imported as imageio.v3 or imageio.v2 respectively; if you just import and use the naked imageio but have 2.16+ installed, you get a noisy and puzzling deprecation warning emitted to stdout, which is not great, especially for tutorials.

All the options here seem bad:

- We could just leave stuff as-is, but that would mean that folks running with a newer (Feb 2022 +) install of imageio would get a distracting deprecation warning
- We could play nicely with both older and newer versions of imageio by wrapping the import in a try block; that would work, but would be a distracting wart at the top of each tutorial
- We could settle on using either the v2 or v3 API, but that means that folks with older imageio installs fail outright

- For the moment, I settled on the last one (using v2), along with requirements.txt... which forced me to upgrade the version on all the buildbots, since they were mostly running 2.9.

I'd welcome feedback on this.

### DIFF
--- a/python_bindings/test/apps/bilateral_grid_shell.py
+++ b/python_bindings/test/apps/bilateral_grid_shell.py
@@ -6,7 +6,7 @@ from bilateral_grid import bilateral_grid
 from bilateral_grid_Adams2019 import bilateral_grid_Adams2019
 from bilateral_grid_Li2018 import bilateral_grid_Li2018
 from bilateral_grid_Mullapudi2016 import bilateral_grid_Mullapudi2016
-import imageio
+import imageio.v2 as imageio
 import numpy as np
 import os
 import sys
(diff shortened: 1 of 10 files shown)


steven-johnson:
(Adding @jiawen here to get comments from a Pythonista... our main Python go-to on the team, Alex Reinking, is taking time away from the project to finish his PhD thesis)

steven-johnson:
(This is probably obsoleted by https://github.com/halide/Halide/pull/7125 -- don't land this without checking on that first.)

steven-johnson:
Obsoleted by #7125

PR #7127, opened 2022-10-26
add 'reverse_axes' options to Buffer conversions

steven-johnson:
This adds some options to the changes in #7127 to allow for easier backwards compatibility with previous versions of the Python bindings:
- The default when creating `hl.Buffer` from a Python buffer is to reverse the axes. There is now an optional named arg to that ctor that you can specify `reverse_axes = False` if you want the behavior of pre-Halide-15 (ie, don't reverse).
- The default when creating `np.array`, `np.ndarray` etc from an `hl.Buffer` is to reverse the axes. There is now a new method on `hl.Buffer`, `reverse_axes()`, which simply returns another Buffer that is a view onto the same memory, but with the order of the axes reverse. This allows you to keep the behavior of pre-Halide-15 (ie, don't reverse) by doing something like `np.array(halidebug.reverse_axes(), ...)`.

### DIFF
--- a/python_bindings/src/halide/halide_/PyBuffer.cpp
+++ b/python_bindings/src/halide/halide_/PyBuffer.cpp
@@ -233,7 +233,6 @@ py::object buffer_setitem_operator(Buffer<> &buf, const std::vector<int> &pos, c
         throw py::value_error("Incorrect number of dimensions.");
     }
 // TODO: add bounds checking?
-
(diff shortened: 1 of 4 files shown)


steven-johnson:
(LMK if this would be easier to review if I went ahead and combined it with #7125)

abadams (approved):

PR #7129, opened 2022-10-26
[x86] Generate AVX512 fixed-point instructions

### LINKED ISSUE #7002: Saturating instructions not generated on AVX512
We do not currently generate the AVX512 saturating_add or saturating_sub instructions, instead we generate multiple calls to the AVX2 variants.

This was noted on a significant performance loss on AVX512 using the `absd` optimization from #6545 , which was traced to not properly generating `saturating_sub` AVX512 instructions.

rootjalex:
This PR adds support for generating `saturating_(add | sub)` and `pmulh(rs)` on Skylake and Cannonlake (i.e. for AVX512BW). It also increases simd_op_check test coverage of fixed-point operations on those archs.

I also did a bit of clean-up on the way:

- Use LLVM intrinsics for x86 `abs`
- Remove explicit use of saturating_pmulhrs, do a proper rewrite (as a temporary fix before #6884 lands).

I did not add `abs` to codegen because it doesn't appear that LLVM currently exposes non-masked versions of AVX512 `abs` variants.

Fixes #7002

### DIFF
--- a/src/CodeGen_X86.cpp
+++ b/src/CodeGen_X86.cpp
@@ -111,13 +111,13 @@ struct x86Intrinsic {
 
 // clang-format off
 const x86Intrinsic intrinsic_defs[] = {
-    {"abs_i8x32", UInt(8, 32), "abs", {Int(8, 32)}, Target::AVX2},
-    {"abs_i16x16", UInt(16, 16), "abs", {Int(16, 16)}, Target::AVX2},
-    {"abs_i32x8", UInt(32, 8), "abs", {Int(32, 8)}, Target::AVX2},
+    {"llvm.x86.avx2.pabs.b", UInt(8, 32), "abs", {Int(8, 32)}, Target::AVX2},
+    {"llvm.x86.avx2.pabs.d", UInt(16, 16), "abs", {Int(16, 16)}, Target::AVX2},
+    {"llvm.x86.avx2.pabs.w", UInt(32, 8), "abs", {Int(32, 8)}, Target::AVX2},
     {"abs_f32x8", Float(32, 8), "abs", {Float(32, 8)}, Target::AVX2},
-    {"abs_i8x16", UInt(8, 16), "abs", {Int(8, 16)}, Target::SSE41},
-    {"abs_i16x8", UInt(16, 8), "abs", {Int(16, 8)}, Target::SSE41},
-    {"abs_i32x4", UInt(32, 4), "abs", {Int(32, 4)}, Target::SSE41},
+    {"llvm.x86.ssse3.pabs.b.128", UInt(8, 16), "abs", {Int(8, 16)}, Target::SSE41},
+    {"llvm.x86.ssse3.pabs.d.128", UInt(16, 8), "abs", {Int(16, 8)}, Target::SSE41},
+    {"llvm.x86.ssse3.pabs.w.128", UInt(32, 4), "abs", {Int(32, 4)}, Target::SSE41},
     {"abs_f32x4", Float(32, 4), "abs", {Float(32, 4)}},
 
     {"round_f32x4", Float(32, 4), "round", {Float(32, 4)}, Target::SSE41},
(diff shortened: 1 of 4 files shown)


src/CodeGen_X86.cpp
```diff
@@ -111,29 +111,37 @@ struct x86Intrinsic {
...
+    {"llvm.x86.ssse3.pabs.w.128", UInt(32, 4), "abs", {Int(32, 4)}, Target::SSE41},
     {"abs_f32x4", Float(32, 4), "abs", {Float(32, 4)}},
 
     {"round_f32x4", Float(32, 4), "round", {Float(32, 4)}, Target::SSE41},
     {"round_f64x2", Float(64, 2), "round", {Float(64, 2)}, Target::SSE41},
     {"round_f32x8", Float(32, 8), "round", {Float(32, 8)}, Target::AVX},
     {"round_f64x4", Float(64, 4), "round", {Float(64, 4)}, Target::AVX},
 
+    {"llvm.sadd.sat.v64i8", Int(8, 64), "saturating_add", {Int(8, 64), Int(8, 64)}, Target::AVX512_Cannonlake},
+    {"llvm.sadd.sat.v64i8", Int(8, 64), "saturating_add", {Int(8, 64), Int(8, 64)}, Target::AVX512_Skylake},
```
steven-johnson, line 129: These look identical except for the feature flags -- what's up with that? Is it just that we want to enable it on either of these features? (other similar ones down below)

src/CodeGen_X86.cpp
```diff
@@ -600,6 +619,22 @@ void CodeGen_X86::visit(const Call *op) {
         }
     }
 
+    // Special case of saturating_pmulhrs.
```
steven-johnson, line 622: Maybe add a sentence about why it is a special case, as it's not immediately obvious to me

src/CodeGen_X86.cpp
rootjalex: Yes - every arch that implements AVX512BW has these instructions, which afaict is only Skylake and Cannonlake.

src/CodeGen_X86.cpp
rootjalex: Bleh, this was a bad comment. Basically we just want to generate `pmulhrs` when we see the saturating variant, plus a check for the case where saturation would trigger. I will update the comment.

rootjalex: @steven-johnson is the comment added in [060ec46](https://github.com/halide/Halide/pull/7129/commits/060ec46c7251364e282758a7db8a4d164b17e7e7) sufficient?

steven-johnson: LGTM

src/CodeGen_X86.cpp
steven-johnson: I guess we don't support using a mask of features there? I guess since we support > 64 features we'd need to use a bitset or similar. Bleah. I guess this is fine since it's not common, just feels a little weird. Maybe add a comment.

rootjalex: It might be worth added an AVX512BW feature flag, that gets added for Cannonlake/Skylake. I don't know what it normally takes to justify adding new feature flags.

rootjalex: Oops, I think that was a timing conflict, my message wasn't in reply to your follow-up, I didn't see that. Yes, I'll add a comment. And yes, I wish it could be a mask of features.

steven-johnson: > AVX512BW

There was a debate about this a while back (ie: should AVX512 be based around capabilities or variants). If you wish to try to argue the case again, let me know :-)

rootjalex: It would make instruction selection simpler :') But I won't argue too hard. I also just realized that Skylake appears to be a strict subset of Cannonlake (based on `complete_x86_target`), so I'm just gonna remove the Cannonlake extras.

rootjalex: fixed in [f6eb8aa](https://github.com/halide/Halide/pull/7129/commits/f6eb8aad4d6fbfebf2c077bd7cab71677dff8346)

steven-johnson:
Several legit failures here

rootjalex:
Can't quite figure out why the JIT doesn't like `ssse3.pabs` instructions. I see them used in LLVM tests (i.e. [here](https://github.com/llvm/llvm-project/blob/7841f557ea3df1987ceb1c033e458790f5828795/llvm/test/CodeGen/X86/ssse3-intrinsics-x86-upgrade.ll#L24)). Gonna revert the use of those for now, but will still change the .ll to use `llvm.abs`.

rootjalex:
Ugh, same deal with the `avx2.pabs` instructions (despite showing up in LLVM tests [here](https://github.com/llvm/llvm-project/blob/e4ec6ce8a75c208b49b163c81cda90dc7373c791/llvm/test/CodeGen/X86/avx2-intrinsics-x86-upgrade.ll#L537-#L561)). I will revert that change and add a comment, but I don't know why these intrinsics in particular are an issue.

rootjalex:
Just updated the AVX512_Skylake `pabs` generation, with a fix to `complete_x86_target` thanks to @abadams

rootjalex:
Only test failure appears unrelated

src/CodeGen_X86.cpp
```diff
@@ -600,6 +617,24 @@ void CodeGen_X86::visit(const Call *op) {
...
+    if (expr_match(saturating_pmulhrs, op, matches)) {
+        // Rewrite so that we can take advantage of pmulhrs.
+        internal_assert(matches.size() == 2);
+        const Expr &a = matches[0];
+        const Expr &b = matches[1];
+        Expr pmulhrs = i16(rounding_shift_right(widening_mul(a, b), 15));
+        // Handle edge case of possible overflow.
+        Expr i16_min = op->type.min();
+        Expr i16_max = op->type.max();
+        Expr expr = select((a == i16_min) && (b == i16_min), i16_max, pmulhrs);
```
abadams, line 633: On AVX2 or <32 vector lanes, the following is better (assuming it's correct)
```
Expr mask = select(max(a, b) == i16_min), cast<int16_t>(-1), cast<int16_t>(0));
Expr expr = mask ^ pmulhrs;
```
This should be four instructions (vpmaxsw, vcmpeqw, vpmulhrsw, vpxor)

The original is five (vcmpeqw, vcmpeqw, vpand, vpmulhrs, vpclendvb)

For avx512 and vectorized at least 32-wide, it seems like what you already have is better, because the comparison result goes into a mask register.

rootjalex: That rewrite formally verifies for me. I will add it!

rootjalex: Added in [41f09f8](https://github.com/halide/Halide/pull/7129/commits/41f09f83f7e108aa4856401db09a24d6ec559a70)

steven-johnson (approved):
LGTM, tests pass on my AVX512 Linux box

PR #7133, opened 2022-10-28
Rewrite python_bindings/apps

steven-johnson:
The apps were pretty old; none used Generators (though they really should have); most of them had very limited scheduling options. I updated all of them based on the equivalent C++ app, moved them to be Generator based (and self-contained instead of mixed in with the tests), and nuked the `erode` app entirely since it was of dubious use and had no C++ counterpart. Also formatted all of them with Black (https://github.com/psf/black), which is one of most popular Python formatters. The goal here is to ensure that the apps here are good examples of "best practices" for Halide Python use as of Halide 15.


python_bindings/test/apps/bilateral_grid_generator.py
```diff
@@ -0,0 +1,196 @@
...
+            pass
+        elif g.target().has_gpu_feature():
+            # 0.50ms on an RTX 2060
+
+            xi, yi, zi = _vars("xi", "yi", "zi")
+
+            # Schedule blurz in 8x8 tiles. This is a tile in
+            # grid-space, which means it represents something like
+            # 64x64 pixels in the input (if s_sigma is 8).
+            (blurz.compute_root().reorder(c, z, x, y).gpu_tile(x, y, xi, yi, 8, 8))
```
abadams, line 129: Is it worth it to use parens for the one-liner schedules, or is it just confusing?

python_bindings/test/apps/blur_app.py
```diff
@@ -0,0 +1,51 @@
+"""
+Bilateral histogram.
```
abadams, line 2: Wrong comment?

python_bindings/test/apps/bilateral_grid_generator.py
steven-johnson: Not sure. I basically wrapped all of the multi-schedule bits in parens and let Black do its thing, and, well, this is what it does. Not sure I like it but apparently that's the way the Python ecosystem seems to be moving. (Google is moving towards `pyink` internally, which is basically a fork of Black with a handful of minor changes to minimize churn.)

python_bindings/test/apps/local_laplacian_generator.py
```diff
@@ -0,0 +1,224 @@
+"""
+Fast image interpolation using a pyramid.
```
abadams, line 2: This comment looks like it's from the interpolate app

python_bindings/test/apps/blur_app.py
steven-johnson: oopsie, fixing

python_bindings/test/apps/local_laplacian_generator.py
steven-johnson: oopsie, fixing

python_bindings/test/apps/bilateral_grid_generator.py
steven-johnson: I decided to remove all the parens from the cases that fit on a single line.

steven-johnson:
Monday Morning Review Ping

abadams:
Are these apps run in CI? I can't figure out which builder does it.

steven-johnson:
> Are these apps run in CI? I can't figure out which builder does it.

Yes, anything that runs the `python` tests exercises these.

abadams (approved):

PR #7134, opened 2022-10-28
[WASM] Use rounding_mul_shift_right for q15mulr_sat_s pattern

rootjalex:
Noticed that the `q15mulr_sat_s` rule didn't use the corresponding intrinsic - meaning we'd lower `rounding_mul_shift_right` just to match the rule before.

### DIFF
--- a/src/CodeGen_WebAssembly.cpp
+++ b/src/CodeGen_WebAssembly.cpp
@@ -187,7 +187,7 @@ void CodeGen_WebAssembly::visit(const Call *op) {
 
     // clang-format off
     static const Pattern patterns[] = {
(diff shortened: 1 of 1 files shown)


steven-johnson:
Do we need an updated test in simd_op_check?

rootjalex:
No, this change should have not introduce any code generation change - it just makes code generation faster.

`find_intrinsics` will have already lifted the target pattern to a `rounding_mul_shift_right`. Before, when CodeGen_LLVM lowers that intrinsic (because CodeGen_WebAssembly doesn't pattern match it), the exact expected pattern is reproduced and then matched to the pattern that has the lowered form. Now we just side-step the lowering.

steven-johnson (approved):

rootjalex:
Test failure appears unrelated

PR #7146, opened 2022-11-03
Attempt to fix Python layout once and for all

steven-johnson:
TL;DR: our current Python structure requires use of relative paths (not ideal) and a lot of fiddling with PYTHONPATH (extremely not ideal). This restructures things so that PYTHONPATH ~never has to be set, and all imports are via absolute Python package/module specifications, rather than relative. Most of this is just code motion.


abadams:
After discussing with Alex (and hopefully not mangling what I learned), I believe the correct layout is:

python_bindings/src/halide/foo.py
python_bindings/src/halide/halide_/foo.cpp
python_bindings/tests
python_bindings/apps
python_bindings/tutorial

i.e. having a 'halide' package makes sense, but it should be under src, and the tests apps and tutorials should be external users of it, not part of it. That way they can run against a system-installed halide, or against the build-tree halide using PYTHONPATH.

abadams:
This seems somewhat official, and recommends the layout above. https://packaging.python.org/en/latest/tutorials/packaging-projects/

steven-johnson:
That is indeed the current layout, but it does not work well in Google's ecosystem. I believe what I have proposed is a better compromise between different styles. Clearly we should discuss this further.

abadams:
Oh, so it is. I didn't realize we had the extra 'halide' in there.

steven-johnson:
I'm going to abandon this (for now), in favor of #7179, which should allow me to solve the most pressing of my issues with hopefully less controversy.

PR #7149, opened 2022-11-03
Add "argv_signature" to C header generation

steven-johnson:
Add a new call, `$funcname_argv_signature()`, to our generated C header files. This returns a const char * one character per expected argument to the 'argv' call; each character is used to indicate the type of the argument. The character encoding is (mostly) borrowed from Python/NumPy, with some minor extensions.

This is intended to provide a lightweight mechanism to write reusable marshal/unmarshal code across language/processor/etc boundaries. (Note that while the existing metadata provides equivalent information, it is also heavier-weight to process, and hard to replicate from scratch; this, on the other hand, provides a convention that can be constructed simply by inspection, or even copy/pasted into source files that may not easily be able to process halide metadata.)

Note that this is being implemented as an inline function, rather than a constant or a #define; I think this is the safest way to implement this in a way that works with C (C99 or later) as well as C++.

This doesn't need to land yet, but I'm about to try building something on top of it and if there are loud objections I'd rather hear sooner than later.

```
!   bfloat16
e   float16
f   float32
d   float64
b   int8
h   int16
i   int32
q   int64
?   bool
B   uint8
H   uint16
I   uint32
Q   uint64
P   void * / handle
@   halide_buffer_t * (as input)
#   halide_buffer_t * (as output)

```

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -1494,10 +1494,82 @@ void CodeGen_C::forward_declare_type_if_needed(const Type &t) {
     forward_declared.insert(t.handle_type);
 }
 
+namespace {
+
+// Emit a string with one character per expected argument to the 'argv' call;
+// each character is used to indicate the type of the argument.
+// Scalar arguments are (mostly) borrowed from Python.
+// This is intended to provide a lightweight mechanism to write reusable marshal/unmarshal
+// code across language/processor/etc boundaries. (Note that while the existing
+// metadata provides equivalent information, it is also heavier-weight to process,
+// and hard to replicate from scratch; this, on the other hand, provides a convention
+// that can be constructed simply by inspection.
+//
+// !   bfloat16
+// e   float16
+// f   float32
+// d   float64
+// b   int8
+// h   int16
+// i   int32
+// q   int64
+// ?   bool
+// B   uint8
+// H   uint16
+// I   uint32
+// Q   uint64
+// P   void * / handle
+// @   halide_buffer_t * (as input)
+// #   halide_buffer_t * (as output)
+//
+char type_to_argv_signature_char(const halide_type_t &t) {
+
+#define HANDLE_CASE(CODE, BITS, CHAR)        \
+    case halide_type_t(CODE, BITS).as_u32(): \
+        return CHAR;
+
+    switch (t.element_of().as_u32()) {
+        HANDLE_CASE(halide_type_bfloat, 16, '!')  // not defined in Python or NumPy, so we just invent this
+        HANDLE_CASE(halide_type_float, 16, 'e')   // this is what NumPy uses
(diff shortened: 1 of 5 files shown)


steven-johnson:
It occurred to me that I could actually do all this downstream, using C++ template techniques to sniff the types of the function, and build up the string I want... with one important exception, namely, it's not currently possible to (safely) tell input buffers from output buffers based on the type signature (both are non-const halide_buffer_t*).

I wonder: is there any way we could tweak the codegen signature to give input and output buffers unique C++ types, without breaking backwards compatibility?

abadams:
I can't help but think that we should just make it easier to iterate over the metadata somehow. Can you provide any more context? If you're doing it from python accessing the metadata should be straight-forward with ctypes.

steven-johnson:
> Can you provide any more context

I'm writing an auto-thunk layer on top of it: basically, for a case where I want to autogenerate the equivalent of an RPC call, to marshall stuff back and forth across an architecture boundary, without clogging up the interior of Halide. I want something with as little overhead as possible, hence the desire for a small, flat 'structure' to provide me this info at runtime.

steven-johnson:
> This is rather stringy. 

Yeah -- I put this out there as the quickest path to try out my idea. It definitely works, but having a strictly type-based solution would clearly be better (since it's pretty much impossible to add actual code that way accidentally). 

As I mentioned above, I could accomplish what I need here by just being able to distinguish input vs output buffers from the type system (rather than using `halide_buffer_t *` for both); unfortunately, I don't think there is a way to modify the existing call signature that won't compromise some existing use case(s) in minor ways (I'd love to hear otherwise). (Well, I take that back; I can imagine some C++-specific trickery with ephemeral struct types that auto-convert from halide_buffer_t*, but it would make for some ugly and confusing header file declarations.)

Probably a better approach would be to emit (say) a templated struct that has all the relevant types as a parameter pack, which can be used for type-sniffing purposes downstream, using a technique similar to what we already use in Callable.h. (Since this struct wouldn't be used for anything else and would never be instantiated, no code impact). Obviously we'd add bespoke types to this to handle the buffer variations.

Re: doing codegen for a general marshaling approach, I pondered that, but I wasn't sure that I knew the shape of enough real-world use cases to design something that would be long-lasting; the benefit to these sort of approaches is that it is much less likely to be a support pain in the long term.

Re: using custom enumerations, yeah, that's literally what I want to do internally... and actually started typing out the enum until I realized I couldn't make the buffer-direction distinction.

Unless someone has a brilliant suggestion for how to annotate the existing calls, I'll probably try out the "generate a struct with the type annotations" in a separate PR and see how that looks.

steven-johnson:
Consider https://github.com/halide/Halide/pull/7157 as an alternative.

steven-johnson:
Closing this in favor of https://github.com/halide/Halide/pull/7170 instead.

PR #7151, opened 2022-11-07
Slightly improve error message for non-integer RDom min/extent

### LINKED ISSUE #6948: Error message when you use something other than an int32 for an RDom extent is bad
It just says 
```Error: Extent cannot be losslessly cast to an int32: ((uint32)diameter - (uint32)1)```

which says nothing about RDom bounds

abadams:
Fixes #6948 

Before: 

Extent cannot be losslessly cast to an int32: 3.400000f

After: 

RDom extent cannot be represented as an int32: 3.400000f


steven-johnson (approved):

steven-johnson:
Ready to land?

PR #7153, opened 2022-11-07
Implement _halide_buffer_to_string for C Backend, plus other fixes

steven-johnson:
The `debug` feature flag is great for debugging in odd situations where you just don't know what is actually being passed to the code... except that the C backend didn't actually emit all the yummy details of the buffer, just its pointer. This fixes that. It also includes two driveby fixes:
- We didn't call `cache.clear()` between internal functions in the C backend, so the cache could try to re-use something declared in a previous (internal, closure) function and would fail to compile. Easy fix. (I'm surprised we haven't seen this fail before now.)
- the runtime `halide_type_to_string` function didn't handle `bfloat`.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -273,7 +273,6 @@ void _halide_buffer_to_string(char* dst, size_t n, const halide_buffer_t *buf) {
 } // namespace
 )INLINE_CODE";
 
-
 class TypeInfoGatherer : public IRGraphVisitor {
 private:
     using IRGraphVisitor::include;
@@ -1931,7 +1930,6 @@ void CodeGen_C::compile(const LoweredFunc &f, const MetadataNameMap &metadata_na
 
             // Return success.
             stream << get_indent() << "return 0;\n";
-
         }
 
         // Ensure we use open/close_scope, so that the cache doesn't try to linger
@@ -2094,7 +2092,6 @@ void CodeGen_C::open_scope() {
     stream << get_indent();
     indent++;
     stream << "{\n";
-
 }
 
 void CodeGen_C::close_scope(const std::string &comment) {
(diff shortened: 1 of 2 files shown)


src/CodeGen_C.cpp
```diff
@@ -219,6 +219,60 @@ class HalideFreeHelper {
 } // namespace
 )INLINE_CODE";
 
+const string post_halide_runtime_globals = R"INLINE_CODE(
+namespace {
+void _halide_buffer_to_string(char* dst, size_t n, const halide_buffer_t *buf) {
```
abadams, line 224: Is it not possible to just call the halide_buffer_to_string already in the runtime?

steven-johnson: There isn't any guarantee that that version will be available for use (e.g., if we are using the C backend because we can't target it via LLVM)

abadams: Wait, does the C backend not require a runtime? I thought there were plenty of calls it could make into the runtime still.

abadams: (After offline discussion)

You need a runtime to use the C backend (e.g. for halide_malloc!). People can either generate one in the standard way, though that raises the question of why they're using the C backend, or they can attempt to compile our runtime sources directly and link that in, or they can just implement runtime functions piecemeal as they need them. I assume the latter is happening here. I don't think we should start inlining duplicate implementations of our runtime functionality into the generated C, because it doesn't solve the problem of needing a runtime unless you implement everything (e.g. halide_do_par_for), and whatever we do implement is a piece of repeated functionality to maintain and test. So I think we should just emit a call to halide_buffer_to_string, and the downstream environment with a custom runtime should implement halide_buffer_to_string.

vksnk: Xtensa backend (from xtensa-codegen branch) is based on C backend, and we indeed compile and link runtime sources directly, so I agree that in this case we should also just emit a call and then link corresponding runtime piece.

steven-johnson: ok, I'll split this PR into several pieces

PR #7157, opened 2022-11-08
Add a struct with type signatures to output, as compile-time-only alternative to #7149

steven-johnson:
The idea here is that we emit info about the call signatures via the type system, rather than as data.

Pros:
    - Can encode more info (eg, buffer input/output but also time and dimensions)
    - Guaranteed not to add to code size if unused
    - In theory, could generate a lot of interesting stuff via metaprogramming using this

Cons:
    - Not usable from straight C
    - Template metaprogramming to make use of this can be verbose and fiddly (see the test in metadata_tester_aottest.cpp)

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -326,7 +326,11 @@ CodeGen_C::CodeGen_C(ostream &s, const Target &t, OutputKind output_kind, const 
         // If it's a header, emit an include guard.
         stream << "#ifndef HALIDE_" << c_print_name(guard) << "\n"
                << "#define HALIDE_" << c_print_name(guard) << "\n"
+               << "\n"
                << "#include <stdint.h>\n"
+               << "#ifdef __cplusplus\n"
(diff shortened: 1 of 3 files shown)


zvookin:
Please document the names and types that are being added here. Likely both in the description and somewhere in persistent documentation. I think this is fairly simple as it is just a tuple of the types with buffers getting two ne types to distinguish in/out.

One design question is whether the types should be taken from the signature itself and only extra info (input/output) from the added tuple. No idea whether this simplifies things or makes it worse.

An obvious thing here is to add a mechanism to put user provided information in the tuple. I'd likely do this by allowing the generator to specify the type that goes in the tuple explicitly. To make this work better, we'd likely want to use struct types for everything, rather than e.g. ```float``` for a float argument, we'd use ```HalideArgument<float>```. This allows inheritance to add extra info while still maintaining floatness. Hence the question of whether you want to do the ```HalideArgument``` wrapper now.

PR #7170, opened 2022-11-16
Add optional "function_info" header output

steven-johnson:
At first glance, this looks like a subset of what is already provided by the `_metadata()` functionality: describing the argument attributes of an AOT-generated Halide function. However, _metadata() is suboptimal for some use cases: Because it's expressed as ordinary data, we can only process it at runtime; the new fuctionality is expressed as a `constexpr` data structure, meaning we can process it at *compile* time if we so choose. (This is quite useful for producing automatic call wrappers, etc).

At first I considered adding this to the normal `.h` file, but moving it into a new file is cleaner in a few ways:
- It maintains the 'C-only' nature of the existing .h files (adding this would have imposed a C++17-only section on them)
- Splitting into a new file means no existing users are affected by this change at all

Note also that this is deliberately not replicating all of the existing `_metadata()` functionality (it's just the argument signature, but no e.g. estimates or default values, etc). This approach means that it is probably more sensible to add several separate constexpr "getters" to this file, rather than trying to mash everything together into one clumsy structure. (With _metadata(), there was an incentive to keep the surface area of the API small, even if that meant combining somewhat-unrelated concerns; there is no such incentive here.)


zvookin:
I may be missing something, but I don't see any Doxygen here. This should be mentioned in documentation, preferably with a simple example that shows what to expect in this new header.

steven-johnson:
> I may be missing something, but I don't see any Doxygen here. This should be mentioned in documentation, preferably with a simple example that shows what to expect in this new header.

Yeah, busted, I was lazy. I'll update this today.

zvookin:
Is "c_function_info_header" a good name? If I understand, using the header requires C++. Possibly just "function_info_header"? Something like "cpp_compile_time_info_header" or  "cpp_compile_time_metadata_header" is highly descriptive, but wordy. (Tempted to leave out "header" but it may be a load bearing word here.)

steven-johnson:
PTAL:
- Renamed c_function_info_header -> function_info_header
- Added docs about the new output to README_cmake.md
- Added some docs + example to the generated  header files
- Added an internal test in CodeGen_C
- Fixed the test in metadata_tester_aottest to actually be `constexpr` rather than merely `const` (oops)

I think I technically satisfied your request, but it's a little unsatisfying: we don't really have a good central place that describes in detail what all of the OutputTypes really generate. We really need such a place, but rather than shave that yak right now, I chose to put details into the README and the generated header. Good enough? Not sure, I need feedback.

steven-johnson:
Monday Morning Review Ping

src/CodeGen_C.cpp
```diff
@@ -221,6 +222,39 @@ class HalideFreeHelper {
 } // namespace
 )INLINE_CODE";
 
+const char *const constexpr_argument_info_docs = R"INLINE_CODE(
+/**
+ * This function returns a constexpr array of information about a Halide-generated
+ * function's argument signature (e.g., number of arguments, type of each, etc).
+ * Technically this is a subset of the information provided by the existing _metadata
+ * function, except that (as a constexpr data structure), meaning one can process
```
abadams, line 230: This sentence seems like it got a little garbled.

abadams:
Another piece of useful constexpr data is the value of the generator params (so that you can enforce that they stay in sync between the code that sets them and any c++ code that also needs to know the value). Not suggesting you add it to this PR, but we should have an idea of how that would look to make sure we're not making a decision here that would make it awkward later.

src/runtime/HalideRuntime.h
```diff
@@ -2046,6 +2048,25 @@ HALIDE_ALWAYS_INLINE constexpr halide_type_t halide_type_of<int64_t>() {
     return halide_type_t(halide_type_int, 64);
 }
 
+#ifndef COMPILING_HALIDE_RUNTIME
+
+namespace HalideFunctionInfo {
```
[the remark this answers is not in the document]
steven-johnson: Done

src/CodeGen_C.cpp
steven-johnson: Done

steven-johnson:
> Another piece of useful constexpr data is the value of the generator params (so that you can enforce that they stay in sync between the code that sets them and any c++ code that also needs to know the value). Not suggesting you add it to this PR, but we should have an idea of how that would look to make sure we're not making a decision here that would make it awkward later.

What I had in mind was basically just adding more `inline constexpr` functions as we need them. For GeneratorParams, the simplest thing would likely be to have something that returns `array<pair<string_view, string_view>>>`, where the values are key-value pairs (with all values being stringified, regardless of type).

steven-johnson:
PTAL

abadams:
Imagine having a numeric generator param that you want to use to control the size of a stack array in c++ because for whatever reason those two things need to be in sync. Doing this:
```
float my_array[my_halide_pipeline_generator_param_foo];
```
is much nicer than.... is atoi constexpr?

But yeah, we can just add more appropriately named constexpr things later.

steven-johnson:
> is much nicer than.... is atoi constexpr?

Dunno, but there are lots of ways we could finesse it. `std::variant` is constexpr so we could use that instead of string if we wanted, but there's something to be said for keeping it simple, too. I think that's a decision we can reasonably make when we actually need the functionality.

steven-johnson:
See also https://github.com/halide/Halide/issues/7180

abadams (approved):

PR #7171, opened 2022-11-17
Add fopen() bottleneck to runtime

steven-johnson:
Prefer using `fopen64()` on Linux systems.

Also, drive-by sorting of the list of initmods that was supposed to be kept sorted.


abadams (approved):

zvookin:
I've chatted with Steven directly on this. We need to avoid confusion between 32-bit and 64-bit platforms and fopen vs. fopen64. (fopen64 allows files that are greater than 2GB. It probably should have been named "fopen_lfs" or something. Plain fopen may or may not have a 32-bit restriction based on platform so calling it fopen_32 is a misnomer. There is much  opportunity for confusion around this in the future, hence the requirement for careful naming and good documentation.)

Ideally we can just call fopen64 everywhere on Linux and nowhere else. It is supported on Windows, but I'm not sure if the same concept applies or they just put it in for compatibility. It does not exist on some platforms because they never had the issue in the first place or still don't support large files. Linux uses the C preprocessor to control which ABI thing code compiles to, which is frankly an egregiously bad decision.

steven-johnson:
Renamed for clarity (thanks @zvookin), PTAL

zvookin:
Also, for context, we're doing this for compliance reasons not because we hit an actual bug. I suppose >2GB trace files could be a thing however.

Issue #7174, opened 2022-11-22
[RISC-V] Data layout mismatch between module


dkurt:
Works fine with LLVM 15.0.2 but fails on LLVM https://github.com/llvm/llvm-project/commit/c67b710379402a48b0a158fa6247ae421ff9aa77

`Module: e-m:e-p:64:64-i64:64-i128:128-n64-S128`
`Engine:  e-m:e-p:64:64-i64:64-i128:128-n32:64-S128`

```
Internal Error at /home/dkurt/Halide/src/JITModule.cpp:286 triggered by user code at : Warning: data layout mismatch between module (e-m:e-p:64:64-i64:64-i128:128-n64-S128) and what the execution engine expects (e-m:e-p:64:64-i64:64-i128:128-n32:64-S128)
```

<details>
<summary>Reproducer</summary>

```cpp
#include "Halide.h"

using namespace Halide;

static int sz = 64;
static int numPrints = 17;

int main(int argc, char** argv) {
    Func brighter("brighter");
    Var x("x"), y("y");

    Buffer<uint8_t> input(sz, sz);
    for (int j = 0; j < numPrints; j++) {
        for (int i = 0; i < numPrints; i++) {
            input(i, j) = i + j;
            std::cout << (int)input(i, j) << " ";
        }
        std::cout << std::endl;
    }
    std::cout << std::endl;

    brighter(x, y) = input(x, y) * 2;

    brighter.bound(x, 0, sz).bound(y, 0, sz);

    Target target = get_host_target();
    std::cout << target << std::endl;

    brighter.print_loop_nest();

    Buffer<uint8_t> output(sz, sz);
    try {
        brighter.realize(output, target);
    } catch(Halide::InternalError& ex) {
        std::cout << ex.what() << std::endl;
    }

    for (int j = 0; j < numPrints; j++) {
        for (int i = 0; i < numPrints; i++) {
            std::cout << (int)output(i, j) << " ";
        }
        std::cout << std::endl;
    }
    std::cout << std::endl;
}
```

</details>

PR #7175, opened 2022-11-22
Use n32:64 in RISC-V data layout

dkurt:
related: https://github.com/llvm/llvm-project/commit/974e2e690b4024c2677dde26cc76ec31e0047c1d
resolves https://github.com/halide/Halide/issues/7174

### DIFF
--- a/src/LLVM_Headers.h
+++ b/src/LLVM_Headers.h
@@ -35,7 +35,6 @@
 #include <llvm/Analysis/TargetTransformInfo.h>
 #include <llvm/Bitcode/BitcodeReader.h>
 #include <llvm/Bitcode/BitcodeWriter.h>
-#include <llvm/ExecutionEngine/JITEventListener.h>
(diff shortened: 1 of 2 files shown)


steven-johnson:
LGTM, please push one more commit to trigger the buildbots

zvookin:
Only question here is whether this should be conditional in the LLVM version. It looks like it was kind of always wrong. I guess the way to find out would be to make the change unconditional and see if if it still works with LLVM 15, but that may not be worth it. (The buildbots will not test that and I don't have an LLVM 15 tree handy.) Going to LGTM this.

zvookin (approved):

steven-johnson:
LGTM once buildbots are green.

dkurt:
@zvookin, good point. Not sure if that will work but I can check with LLVM 15 as well.

steven-johnson:
Thanks for the fix!

PR #7183, opened 2022-11-29
Use i1 as result type for llvm.vp.icmp

dkurt:
`icmp` return type should be i1 but not i32: https://llvm.org/docs/LangRef.html#llvm-vp-icmp-intrinsics

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1677,6 +1677,7 @@ void CodeGen_LLVM::visit(const EQ *op) {
         codegen(EQ::make(cast(t, op->a), cast(t, op->b)));
         return;
     }
+    t = UInt(1, t.lanes());
 
     Value *a = codegen(op->a);
     Value *b = codegen(op->b);
@@ -1697,6 +1698,7 @@ void CodeGen_LLVM::visit(const NE *op) {
         codegen(NE::make(cast(t, op->a), cast(t, op->b)));
         return;
     }
+    t = UInt(1, t.lanes());
 
     Value *a = codegen(op->a);
     Value *b = codegen(op->b);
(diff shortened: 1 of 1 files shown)


steven-johnson:
The buildbots seem to disagree with you, this is failing literally everywhere

dkurt:
@steven-johnson, sorry, forgot that type is still used in `if` statement. Fixed that.

steven-johnson:
Looks like it has broken Hexagon/HVX codegen (see https://buildbot.halide-lang.org/master/#/builders/42/builds/549/steps/29/logs/correctness_simd_op_check_hvx):

Failed: vdelta(v*,v*) did not generate for target=hexagon-32-noos-hvx-hvx_128-no_asserts-no_bounds_query-no_runtime vector_width=512. Instead we got:

attn @pranavb-ca for suggestions

dkurt:
Thanks! I've tried debug Hexagon locally but I'm not able to download the SDK.

steven-johnson:
> Thanks! I've tried debug Hexagon locally but I'm not able to download the SDK.

Update: this may be unrelated to your PR, I'm getting an apparently-similar failure in main with top-of-tree LLVM. Investigating.

pranavb-ca:
@aankit-ca - fyi.

steven-johnson:
The "failure" can be ignored, see https://reviews.llvm.org/rG073d5e5945c428e20db0884943e6dcb7ff2158df  -- output is different but better, I will update the testcase today. In the meantime, this PR is good to go.

steven-johnson (approved):
LGTM but adding @abadams as he knows this better than I

abadams:
Actually I think we want @zvookin

dkurt:
Does RISC-V instance on buildbot is good option? QEMU can be used to run JIT and AOT tests. I've tried collect some brief proposals here: https://github.com/halide/build_bot/issues/214.

zvookin:
I'll try to get to this shortly, but I don't think this is the right thing. (The right thing is for the internals of that function to always use i1 and to cast the result to the requested result type. Also, needs a test, which is on me.)

dkurt:
see https://github.com/halide/Halide/pull/7205

PR #7187, opened 2022-11-29
[xtensa] Improved gather_load with IVP_GATHER

Aelphy:


src/CodeGen_Xtensa.cpp
```diff
@@ -633,16 +633,13 @@ HALIDE_ALWAYS_INLINE void store_variable<native_vector_u8, uint8_t, VECTOR_WIDTH
...
-    int offsets[Lanes];
-    store<OffsetType, int32_t, Lanes>(offset, &offsets[0], 0);
-    for (int i = 0; i < Lanes; i++) {
-        tmp[i] = ((const BaseType*)base)[offsets[i]];
-    }
-
-    return *((VectorType *)tmp);
+template <typename VectorType, typename OffsetType, typename BaseType, int Lanes, bool isTCM>
+HALIDE_ALWAYS_INLINE VectorType gather_load(const void *base, const OffsetType& offset){
+  if (isTCM) {
```
vksnk, line 638: We probably don't want to have an if statement here. I think slight modification of your approach might fix it: I would keep a single gather_load function (instead of gather_load_tcm and gather_load_default), add a new template parameter isTCM (already there), all template specializations which are implemented with intrinsics will have it set to true and then emit additional true or false template parameter at the handler of Allocate node. This way only gather_loads with matching types AND is_tcm parameter set to true will get their corresponding specialization called, but everything else will default to scalarized implementation.

src/CodeGen_Xtensa.cpp
```diff
@@ -2363,6 +2360,117 @@ HALIDE_ALWAYS_INLINE native_vector_f32_x2 halide_xtensa_concat_from_native(const
...
+  // NOTE(aelphy): the intrinsic for gathering 8-bit elements extends them to 16-bit, and the conversion back to 8-bit is needed
+  return convert<native_vector_u8, native_vector_u16_x2>(native_vector_u16_x2(native_vector_u16_x2::from_native_vector, output1, output2));
+}
+
+template<>
+HALIDE_ALWAYS_INLINE HALIDE_MAYBE_UNUSED native_vector_i16 gather_load_tcm<native_vector_i16, native_vector_i32_x2, int16_t, VECTOR_WIDTH_U16>(const void *base, const native_vector_i32_x2& offset) {
+  return IVP_GATHERDNX16(
+    IVP_GATHERANX16(
+      (const int16_t*) base,
+      convert<native_vector_u16, native_vector_i32_x2>(offset) << 1
```
vksnk, line 2429: Could you please add a comment that this shift is needed because offsets are expected to be in bytes?

src/CodeGen_Xtensa.cpp
Aelphy: One comment, here. Since we emit gather_load during visiting Load node, I think we should emit is_tcm aslo there and not in Allocate node.

src/CodeGen_Xtensa.cpp
Aelphy: done

steven-johnson:
Ready to land?

PR #7189, opened 2022-11-30
Use aligned_alloc() as default allocator for halide_malloc() on most platforms

steven-johnson:
C++17 includes C11, which includes `aligned_alloc` as part of the standard library. We should move to prefer using it over `malloc()` in `halide_malloc()`, since we currently have to do an overallocate-and-adjust trick with malloc() to get what we want.

We can't move to it everywhere, alas, because:
- Older versions of Android (before Android 9) and OSX (before 10.15) don't support them regardless of your compiler version
- Microsoft has stated that they are unlikely to *ever* support it, as they map malloc() directly onto Win32 APIs and apparently those APIs don't offer any aligned-allocation options, oddly enough
- Hexagon/QuRT I left as-is, because I'm not sure about whether `aligned_alloc` is offered there and if so, what version. (Would welcome input from @pranavb-ca et al on this)

That said, on ~all modern Unixy platforms, it should be available and arguably preferable, so this PR makes it the default, with a `no_aligned_alloc` feature to allow dropping back to using plain malloc()-with-hackery. (This option is ignored for Windows and QuRT, of course.)

It's worth pointing out that this subtlely modifies the (implicit) contract of `halide_malloc`: in addition to guaranteeing that the returned pointer is aligned to `halide_malloc_alignment()`, the `aligned_alloc()` allocator also requires/guarantees that the memory returned is an integral number of align-sizes long (rounding up, obviously). In order to avoid subtle behavior differences between the two malloc-based allocators remaining (posix and qurt) I modified them to also make the same guarantee. (I guess I should document this and make sure that all the overrides of `halide_malloc()` I know of in Google and other places make this same guarantee...)

I considered adding a new `halide_aligned_alloc()` runtime function instead, and then migrating all the calls we make for buffer memory there (since that's the only place we really care about this alignment), leaving `halide_malloc()` to be used for whatever places we need to allocate memory that doesn't need special alignment... but (1) I'm not sure if there are enough of the latter to be worthwhile, and (2) we'd have to 'dumb down' `halide_malloc()` to make the exercise worthwhile, and that would leave us risk of hard-to-repro crashes if we got one of the allocations wrong.

This will likely need some torture testing, since it's possible that some platforms offer `aligned_alloc` but with inferior performance.

### DIFF
--- a/src/runtime/posix_allocator.cpp
+++ b/src/runtime/posix_allocator.cpp
@@ -24,7 +24,7 @@ WEAK void *halide_default_malloc(void *user_context, size_t user_size) {
     // in size, which we can always fit entirely into 2*alignment.
     const size_t requested_size = (aligned_size <= alignment) ?
                                       (alignment * 2) :
-                                      (aligned_size + alignment - 1);
+                                      (aligned_size + alignment);
 
     // malloc() and friends must return a pointer aligned to at least
     // alignof(std::max_align_t); we can't reasonably check that in
--- a/src/runtime/posix_allocator.cpp
+++ b/src/runtime/posix_allocator.cpp
@@ -19,12 +19,7 @@ WEAK void *halide_default_malloc(void *user_context, size_t user_size) {
     // so that all allocators follow the behavior of aligned_alloc() and
     // return aligned pointer *and* aligned length.
     const size_t aligned_size = align_up(user_size, alignment);
-
-    // We can save a bit of space by special-casing allocations < alignment
-    // in size, which we can always fit entirely into 2*alignment.
-    const size_t requested_size = (aligned_size <= alignment) ?
-                                      (alignment * 2) :
-                                      (aligned_size + alignment);
+    const size_t requested_size = (aligned_size + alignment);
 
     // malloc() and friends must return a pointer aligned to at least
     // alignof(std::max_align_t); we can't reasonably check that in
--- a/src/runtime/posix_allocator.cpp
+++ b/src/runtime/posix_allocator.cpp
@@ -3,6 +3,19 @@
 
 #include "printer.h"
 
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+
+// Read into a global to avoid making a call to halide_malloc_alignment()
+// in every halide_malloc() call (halide_malloc_alignment() is required to
+// return the same value every time).
+WEAK size_t _alignment = (size_t) halide_malloc_alignment();
+
+}  // namespace Internal
+}  // namespace Runtime
+}  // namespace Halide
+
 extern "C" {
 
 extern void *malloc(size_t);
@@ -10,7 +23,7 @@ extern void free(void *);
 
 WEAK void *halide_default_malloc(void *user_context, size_t user_size) {
     // Allocate enough space for aligning the pointer we return.
-    const size_t alignment = halide_malloc_alignment();
+    const size_t alignment = Halide::Runtime::Internal::_alignment;
 
     // All bets are off if alignment is smaller than a pointer.
     halide_debug_assert(user_context, alignment >= sizeof(void *));
--- a/src/runtime/posix_allocator.cpp
+++ b/src/runtime/posix_allocator.cpp
@@ -10,7 +10,7 @@ namespace Internal {
 // Read into a global to avoid making a call to halide_malloc_alignment()
 // in every halide_malloc() call (halide_malloc_alignment() is required to
 // return the same value every time).
-WEAK size_t _alignment = (size_t) halide_malloc_alignment();
+WEAK size_t _alignment = (size_t)halide_malloc_alignment();
 
 }  // namespace Internal
 }  // namespace Runtime
--- a/src/runtime/posix_allocator.cpp
+++ b/src/runtime/posix_allocator.cpp
@@ -8,7 +8,7 @@ extern "C" {
 extern void *malloc(size_t);
 extern void free(void *);
 
-WEAK void *halide_default_malloc(void *user_context, size_t x) {
+WEAK void *halide_default_malloc(void *user_context, size_t user_size) {
     // Allocate enough space for aligning the pointer we return.
     const size_t alignment = halide_malloc_alignment();
 
--- a/src/runtime/posix_allocator.cpp
+++ b/src/runtime/posix_allocator.cpp
@@ -24,7 +24,7 @@ WEAK void *halide_default_malloc(void *user_context, size_t user_size) {
     // in size, which we can always fit entirely into 2*alignment.
     const size_t requested_size = (aligned_size <= alignment) ?
                                       (alignment * 2) :
-                                      (aligned_size + sizeof(void *) + alignment - 1);
+                                      (aligned_size + alignment - 1);
 
     // malloc() and friends must return a pointer aligned to at least
     // alignof(std::max_align_t); we can't reasonably check that in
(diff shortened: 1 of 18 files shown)


abadams:
On platforms where it's available, why would we ever *not* use aligned_alloc?

Is the target feature just for testing it out?

steven-johnson:
> On platforms where it's available, why would we ever _not_ use aligned_alloc?

- On some platforms we can't determine for sure... e.g. Android, where it is only available with SDK level 28 (or so), regardless of C++17. 
- It's also possible that a platform has implemented aligned_alloc() but it's dramatically slower than malloc().
 
> Is the target feature just for testing it out?

For now, yes.

src/runtime/qurt_allocator.cpp
```diff
@@ -12,10 +12,10 @@ namespace Internal {
 
 WEAK void *aligned_malloc(size_t alignment, size_t size) {
     // We also need to align the size of the buffer.
-    size = (size + alignment - 1) & ~(alignment - 1);
+    size = align_up(size, alignment);
 
     // Allocate enough space for aligning the pointer we return.
-    void *orig = malloc(size + alignment);
+    void *orig = malloc(size + alignment * 2);
```
abadams, line 18: Why *2? Adding one alignment worth should be enough to fit a pointer, surely.

steven-johnson: Oops, scalpel left in patient, fixing

test/correctness/allocator_alignment.cpp
```diff
@@ -0,0 +1,104 @@
...
+            copy(x, y) = img_(x, y);
+
+            out_(x, y) = u32_sat(copy(x, y) + offset_);
+
+            copy.compute_root().vectorize(x, natural_vector_size<uint32_t>());
+            out_.vectorize(x, natural_vector_size<uint32_t>());
+        }
+    };
+
+    // Make it large enough so that we won't attempt to use stack instead of heap
```
abadams, line 40: Could also use copy.store_in(MemoryType::Heap)

steven-johnson: Done

test/correctness/allocator_alignment.cpp
```diff
@@ -0,0 +1,104 @@
...
+            assert(out4(i, j) == i * 10 + j + offset4);
+        }
+    }
+
+    // Now run a benchmark, but don't check it
+    double time = Halide::Tools::benchmark(10, 100, [&]() {
+        check(c(in2, offset4, out4));
+    });
+    const float megapixels = (W * H) / (1024.f * 1024.f);
+    printf("Benchmark: %dx%d -> %f mpix/s for %s\n", W, H, megapixels / time, t.to_string().c_str());
```
abadams, line 89: Should this be a performance test?

test/correctness/allocator_alignment.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    const GeneratorContext context(t);
+
+    auto gen = TestGen1::create(context);
+    Callable c = gen->compile_to_callable();
+
+    const uint32_t offset1 = 42;
+    Buffer<uint32_t> out1(W, H);
+    check(c(in1, offset1, out1));
```
abadams, line 61: Can Callables return non-zero without a custom error handler? I thought we just threw an exception.

steven-johnson: Probably not, but it's good hygiene.

test/correctness/allocator_alignment.cpp
steven-johnson: Maybe eventually? Need to see real-world examples first to see where we can expect to use it.

src/runtime/posix_allocator.cpp
```diff
@@ -8,16 +8,36 @@ extern "C" {
...
+    // Always round allocations up to alignment size,
+    // so that all allocators follow the behavior of aligned_alloc() and
+    // return aligned pointer *and* aligned length.
+    const size_t aligned_size = align_up(user_size, alignment);
+
+    // We can save a bit of space by special-casing allocations < alignment
+    // in size, which we can always fit entirely into 2*alignment.
+    const size_t requested_size = (aligned_size <= alignment) ?
+                                      (alignment * 2) :
+                                      (aligned_size + sizeof(void *) + alignment - 1);
```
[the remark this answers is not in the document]
abadams: The worst case is when malloc returns an aligned pointer. We need to shift the base a whole alignment bytes up, and we'll use those bytes for the pointer. If we need to shift the base less than alignment bytes up, then we still fit into the allocation, and we'll still have at least a void * worth of room before the start.

src/runtime/qurt_allocator.cpp
```diff
@@ -10,18 +10,30 @@ namespace Halide {
...
-    void *orig = malloc(size + alignment);
+WEAK void *aligned_malloc(size_t alignment, size_t user_size) {
+    // Always round allocations up to alignment size,
+    // so that all allocators follow the behavior of aligned_alloc() and
+    // return aligned pointer *and* aligned length.
+    const size_t aligned_size = align_up(user_size, alignment);
+
+    // We can save a bit of space by special-casing allocations < alignment
+    // in size, which we can always fit entirely into 2*alignment.
+    const size_t requested_size = (aligned_size <= alignment) ?
```
abadams, line 21: Same comment here

test/correctness/allocator_alignment.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace {
+
+void check(int r) {
+    assert(r == 0);
+}
+
+void run_test(Target t) {
+    // Must call this so that the changes in cached runtime are noticed
+    Halide::Internal::JITSharedRuntime::release_all();
```
abadams, line 16: I'm surprised we don't invalidate the cache if the target changes.

src/runtime/posix_allocator.cpp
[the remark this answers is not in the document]
steven-johnson: Done

test/correctness/allocator_alignment.cpp
steven-johnson: Me too. Want me to open an issue?

steven-johnson: https://github.com/halide/Halide/issues/7195

steven-johnson:
PTAL

src/runtime/posix_allocator.cpp
```diff
@@ -8,16 +8,36 @@ extern "C" {
...
+    halide_debug_assert(user_context, alignment >= sizeof(void *));
+
+    // Always round allocations up to alignment size,
+    // so that all allocators follow the behavior of aligned_alloc() and
+    // return aligned pointer *and* aligned length.
+    const size_t aligned_size = align_up(user_size, alignment);
+
+    // We can save a bit of space by special-casing allocations < alignment
+    // in size, which we can always fit entirely into 2*alignment.
+    const size_t requested_size = (aligned_size <= alignment) ?
```
abadams, line 25: Isn't this just: requested_size = aligned_size + alignment? I doubt the -1 is going to do anything.

steven-johnson: Done

abadams (approved):

pranavb-ca:
@steven-johnson  - Thanks for this. You are right, the pointer math tricks can be avoided. I think we can replace this with memalign on QuRT. I am working from India at present but only for a few days more before I am out on vacation, so, I'll get to this for Qurt once I am back in early Jan.

steven-johnson:
attn @abadams: please take a look at https://github.com/halide/Halide/pull/7189/commits/54caa29752129122c27acf8a7103902be7ec1a0d

steven-johnson:
Closing this in favor of https://github.com/halide/Halide/pull/7206

PR #7190, opened 2022-11-30
Use aligned_alloc() as default allocator for HalideBuffer.h on most platforms

steven-johnson:


abadams (approved):

steven-johnson:
Added HALIDE_RUNTIME_BUFFER_ALLOCATION_ALIGNMENT to allow the default alignment of 128 to be configurable at compile time. I think this is probably ready to land (pending green); it doesn't need to land in tandem with #7189.

abadams (approved):

PR #7193, opened 2022-11-30
Add a -v flag to generator_main()

steven-johnson:
This is a simple thing that just logs the path to all generated file(s) to stdout if `-v=1` is specified. It's intended for people running Generators directly from the commandline, and is intended as a more user-friendly alternative to HL_DEBUG_CODEGEN=1. No makefiles, etc specify it at present, but I anticipate using it in some tooling in the future.

Example usage:
```
$ resize_image_bilinear.generator_binary -v 1 -o /tmp -g resize_image_bilinear -n resize_image_bilinear_uint16 -f resize_image_bilinear_uint16 -e assembly,c_header,llvm_assembly,registration,static_library,stmt 'target=arm-64-android' 'input.type=uint16' 'output.type=uint16'
Generated file: /tmp/resize_image_bilinear_uint16.s
Generated file: /tmp/resize_image_bilinear_uint16.h
Generated file: /tmp/resize_image_bilinear_uint16.ll
Generated file: /tmp/resize_image_bilinear_uint16.registration.cpp
Generated file: /tmp/resize_image_bilinear_uint16.a
Generated file: /tmp/resize_image_bilinear_uint16.stmt
```

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -672,6 +672,7 @@ gengen
      bugs and/or degenerate cases don't stall build systems. Defaults to 900
      (=15 minutes). Specify 0 to allow ~infinite time.
 
+ -v  If nonzero, log the path to all generated files to stdout.
 )INLINE_CODE";
 
     std::map<std::string, std::string> flags_info = {
@@ -683,6 +684,7 @@ gengen
         {"-o", ""},
         {"-p", ""},
         {"-r", ""},
+        {"-v", "0"},
         {"-t", "900"},  // 15 minutes
     };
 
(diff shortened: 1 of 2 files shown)


abadams:
Most tools would have verbosity given by `-v`, not `-v 1`. Is the issue that our argument parsing always expects an argument on every flag?

steven-johnson:
> Most tools would have verbosity given by `-v`, not `-v 1`. Is the issue that our argument parsing always expects an argument on every flag?

Yeah. I could special-case it for -v, though.

steven-johnson:
>  Most tools would have verbosity given by -v, not -v 1.

I took a stab at allowing -v and -d to omit the argument, but the problem is that there are cases that are somewhat tricky to parse: if the following `argv` starts with `-` (or there is no next value) then default to `1` is easy, but what if the next arg is the name of a GeneratorParam (which doesn't begin with `-`)? We don't have a set of valid GeneratorParam names at that point. So how should we parse each of the following:

```
generator -v target=host   # We don't know at flag-parsing time that `target=host` isn't meant as the value for -v
generator -v nonexistent_generator_param=foo # Ditto
generator -v 22=33 
```

In hindsight, we should have specified GeneratorParams via flags of some sort, but no way we are gonna change that now.

I'm not sure there's a clean way to resolve this without adding some path to extract the known GeneratorParam names at flag-parsing time, and while I could do that, it would be kinda fragile. LMK.

abadams:
bleh, Let's just stick with -v 1 then.

abadams (approved):

PR #7198, opened 2022-12-05
[xtensa] Added initial support for float16_t

Aelphy:


steven-johnson:
attn @zvookin for related non-xtensa work

steven-johnson:
LGTM

PR #7205, opened 2022-12-06
Fix issue with vector predicated comparison and select instructions.

zvookin:
Add start of RISC V simd_op_check test.


steven-johnson:
CMakeLists.txt needs updating, otherwise looks good pending green

steven-johnson (approved):

Issue #7208, opened 2022-12-07
Some powerpc simd ops aren't being generated


steven-johnson:
correctness_simd_op_check_powerpc hasn't been getting tested regularly and has bitrotted; there are several ops that simd_op_check_powerpc tests for that aren't getting generated. This is a tracking issue to note the breakage.

Issue #7207, opened 2022-12-07
simd_op_check tests should be runnable via crosscompilation without setting HL_TARGET


steven-johnson:
simd_op_check used to be a single monolithic test, which decided which architecture to run based on the value of HL_TARGET. We recently split the tests up into architecture-specific variants, but the logic for the right host config is still there, meaning that running (say) simd_op_check_arm on x86 does... nothing at all. 

The reasonable default should probably instead be to test one (or more) likely targets for that test (eg perhaps x86 should test several simd variants), and just skip the jit testing: it is still valuable to verify that the expected instructions are being generated.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** The request was that each architecture-specific simd_op_check test compile for a fixed list of targets regardless of `HL_TARGET`, and skip only the JIT-run portion when the host does not match. That is precisely how the tests work now.

**Evidence:** Commit 16421a7ae, "Revise simd_op_check tests to ignore HL_TARGET (#7207) (#7216)". Today `SimdOpCheckTest::main(argc, argv, targets_to_test)` (`test/correctness/simd_op_check.h:544-591`) loops over an explicit target list passed by each test (e.g. `test/correctness/simd_op_check_powerpc.cpp:137-148`, `simd_op_check_arm.cpp` bottom), and `can_run_code()` (`test/correctness/simd_op_check.h:104-120`) gates only the execution step on host compatibility. Running `correctness_simd_op_check_powerpc` on this x86 host compiles all eight PowerPC targets and prints `Success!`.

Closing: fixed by #7216 (commit 16421a7ae). Each simd_op_check test now takes an explicit list of targets (`SimdOpCheckTest::main(..., targets_to_test)` at `test/correctness/simd_op_check.h:544`) and ignores `HL_TARGET`, running the JIT check only when `can_run_code()` says the host matches.

PR #7211, opened 2022-12-07
Fix bitrot in PowerPC testing

steven-johnson:
(See #7208)

- DataLayout was wrong (and has been for a long time)
- simd_op_check_powerpc had errors. Some were easy to fix; the rest I commented out with a TODO since this backend doesn't appear to be in active use.

(Want to fix this in preparation for fixing #7207)

(Note to reviewers: this isn't being run on the buildbots, so there shouldn't be any failures -- if you want to test right now, run locally and set HL_TARGET appropriately)

### DIFF
--- a/test/correctness/simd_op_check_powerpc.cpp
+++ b/test/correctness/simd_op_check_powerpc.cpp
@@ -61,9 +61,11 @@ public:
             check("vsububs", 16 * w, u8(max(i16(u8_1) - i16(u8_2), 0)));
             check("vsubuhs", 8 * w, u16(max(i32(u16_1) - i32(u16_2), 0)));
             check("vsubuws", 4 * w, u32(max(i64(u32_1) - i64(u32_2), 0)));
-            check("vsububs", 16 * w, absd(i8_1, i8_2));
-            check("vsubuhs", 16 * w, absd(i16_1, i16_2));
-            check("vsubuws", 16 * w, absd(i32_1, i32_2));
+            // TODO: not getting generated in recent LLVM builds.
+            // https://github.com/halide/Halide/issues/7208
+            // check("vsububs", 16 * w, absd(i8_1, i8_2));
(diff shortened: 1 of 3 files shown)


test/correctness/simd_op_check_powerpc.cpp
```diff
@@ -61,9 +61,11 @@ class SimdOpCheckPowerPC : public SimdOpCheckTest {
             check("vsububs", 16 * w, u8(max(i16(u8_1) - i16(u8_2), 0)));
             check("vsubuhs", 8 * w, u16(max(i32(u16_1) - i32(u16_2), 0)));
             check("vsubuws", 4 * w, u32(max(i64(u32_1) - i64(u32_2), 0)));
-            check("vsububs", 16 * w, absd(i8_1, i8_2));
-            check("vsubuhs", 16 * w, absd(i16_1, i16_2));
-            check("vsubuws", 16 * w, absd(i32_1, i32_2));
+            // TODO: not getting generated in recent LLVM builds.
+            // https://github.com/halide/Halide/issues/7208
+            // check("vsububs", 16 * w, absd(i8_1, i8_2));
```
abadams, line 66: https://github.com/halide/Halide/commit/bc5b5575576e480fdcceb3ab7cc46bdc69e7b64c

It looks like the cause was me accidentally putting these ones in the powerpc section instead of the x86 section! I don't think they're even correct, because the or-two-saturating-subtracts trick only works for unsigned.

abadams: I'll take the liberty of pushing a fix to this branch, moving these tests to the right place.

steven-johnson: Wait, x86 doesn't have any instructions of those names, does it?

abadams: It's vpsubusb on x86, which looked close enough to fool me.

zvookin (approved):

PR #7216, opened 2022-12-07
Revise simd_op_check tests to ignore HL_TARGET (#7207)

steven-johnson:
The simd_op_check tests have historically only run using the value of HL_TARGET, which mean that the coverage they had was low (since HL_TARGET is only set to values that are runnable on at least one buildbot). This change completely disconnects these tests from HL_TARGET; instead, each test now tests for a range of targets appropriate to the architecture being tested. On all platforms, they still compile to assembly and verify that the correct instructions are generated; additionally, if the host platform can JIT for the given target, it verifies that the results are as expected.

### DIFF
--- a/test/correctness/simd_op_check_arm.cpp
+++ b/test/correctness/simd_op_check_arm.cpp
@@ -1078,5 +1078,11 @@ private:
 }  // namespace
 
 int main(int argc, char **argv) {
-    return SimdOpCheckTest::main<SimdOpCheckARM>(argc, argv);
+    return SimdOpCheckTest::main<SimdOpCheckARM>(
+        argc, argv,
+        {
+            Target("arm-32-linux"),
+            Target("arm-64-linux"),
+            Target("arm-64-linux-arm_dot_prod"),
+        });
 }
--- a/test/correctness/simd_op_check_x86.cpp
+++ b/test/correctness/simd_op_check_x86.cpp
@@ -567,10 +567,11 @@ public:
             check("vpminsq", 8, min(i64_1, i64_2));
         }
         if (use_avx512 && target.has_feature(Target::AVX512_SapphireRapids)) {
-            check("vcvtne2ps2bf16*zmm", 32, cast(BFloat(16), f32_1));
-            check("vcvtneps2bf16*ymm", 16, cast(BFloat(16), f32_1));
-            check("vcvtneps2bf16*xmm", 8, cast(BFloat(16), f32_1));
-            check("vcvtneps2bf16*xmm", 4, cast(BFloat(16), f32_1));
+            // TODO: fails LLVM verification in LLVM16
+            // check("vcvtne2ps2bf16*zmm", 32, cast(BFloat(16), f32_1));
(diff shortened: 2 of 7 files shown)


steven-johnson:
(Note that all the RISCV tests are going to fail until the buildbots add RISCV to their LLVM16 build, probably by tomorrow)

test/correctness/simd_op_check_arm.cpp
```diff
@@ -548,19 +548,31 @@ class SimdOpCheckARM : public SimdOpCheckTest {
                         // If we're widening the type by a factor of four
                         // as well as reducing by a factor of four, we
                         // expect vpaddl followed by vpadal
-                        check(arm32 ? "vpaddl.s8" : "saddlp", 8, sum_(i32(in_i8(f * x + r))));
-                        check(arm32 ? "vpaddl.u8" : "uaddlp", 8, sum_(i32(in_u8(f * x + r))));
-                        check(arm32 ? "vpaddl.u8" : "uaddlp", 8, sum_(u32(in_u8(f * x + r))));
+                        if (target.has_feature(Target::ARMDotProd)) {
```
rootjalex, line 551: Were we never testing with this feature before?

steven-johnson: Nope!

test/correctness/simd_op_check_x86.cpp
```diff
@@ -567,10 +567,11 @@ class SimdOpCheckX86 : public SimdOpCheckTest {
             check("vpminsq", 8, min(i64_1, i64_2));
         }
         if (use_avx512 && target.has_feature(Target::AVX512_SapphireRapids)) {
-            check("vcvtne2ps2bf16*zmm", 32, cast(BFloat(16), f32_1));
-            check("vcvtneps2bf16*ymm", 16, cast(BFloat(16), f32_1));
-            check("vcvtneps2bf16*xmm", 8, cast(BFloat(16), f32_1));
-            check("vcvtneps2bf16*xmm", 4, cast(BFloat(16), f32_1));
+            // TODO: fails LLVM verification in LLVM16
```
rootjalex, line 570: Should this be an open issue?

test/correctness/simd_op_check_arm.cpp
rootjalex: Yikes. Great catch!

test/correctness/simd_op_check_x86.cpp
steven-johnson: ...yep, that was it, fix pushed, PTAL

steven-johnson: ...Whoops, nope, I was mistaken. There were bugs in can_run_the_code, but this was unrelated. Opened issue https://github.com/halide/Halide/issues/7219 (feel free to claim it).

rootjalex: It may take me a few weeks to get to. I'll try to take a look over the holidays!

PR #7223, opened 2022-12-09
Use `aligned_alloc()` when possible

steven-johnson:
Currently, all our `halide_malloc()` implementations just use `malloc()`/`free()`, user overallocation tricks to ensure the right alignment. This PR adds a new implementation, which uses the C11/C++17 `aligned_alloc()` call instead. By default, we use this implementation on all Unixy platforms, with a new Feature, `no_aligned_alloc`, to allow forcing the use of `malloc()` instead. This is necessary because while ~all modern Linux versions support this, Android doesn't support it till API >= 28, and OSX doesn't support it till >= 10.15. (The QuRT allocator will continue to use `malloc()` for now, pending some post-holiday investigation by QC.)

We also add a Windows-specific variant that uses their `_aligned_malloc()`/`_aligned_free()` calls; IIRC, the MSVC team has stated that they are unlikely to ever support the standard `aligned_alloc()` calls, for reasons that aren't important here, but do support these as a partial workaround.

This will likely need some torture testing, since it's possible that some platforms offer `aligned_alloc()` implementations that have inferior performance to `malloc()`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -751,6 +751,7 @@ HEADERS = $(HEADER_FILES:%.h=$(SRC_DIR)/%.h)
 
 RUNTIME_CPP_COMPONENTS = \
   aarch64_cpu_features \
+  aligned_alloc_aligned_alloc \
   alignment_128 \
   alignment_32 \
   alignment_64 \
@@ -830,6 +831,7 @@ RUNTIME_CPP_COMPONENTS = \
   trace_helper \
   tracing \
   wasm_cpu_features \
+  windows_aligned_alloc \
   windows_clock \
   windows_cuda \
   windows_d3d12compute_arm \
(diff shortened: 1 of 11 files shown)


mcourteaux (changes requested):
I don't see why we would add an alternative routine for Windows. [The docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=msvc-170#remarks) state that the `_aligned_malloc` routine just calls `malloc`, and does pretty much the same as our own `halide_internal_aligned_alloc` wrapper, if you look at ReactOS reverse engineered code: https://doxygen.reactos.org/dd/d29/align_8cpp_source.html#l00147 

Additionally, I think this is a great opportunity to clear up the names of these runtime modules. I propose something like:

 - `posix_aligned_alloc`: your new implementation (see below)
 - `malloc_aligned_alloc`: the old implementation using the user overallocation approach

This would also delete the `windows_aligned_alloc` module.

Regarding the Android issue, it seems that `posix_memalign` is actually introduced in Android API 16: https://stackoverflow.com/a/44880554/155137 Accoring to https://apilevels.com/, level 16 was introduced in 2012, and is run by 99.9% of devices. So the `posix_aligned_alloc` module could actually call the `posix_memalign` function. This function is [also available on macOS since 10.6](https://github.com/alexey-lysiuk/macos-sdk/blob/main/MacOSX10.6.sdk/usr/include/stdlib.h#L173), which is really old by now.

PR #7224, opened 2022-12-09
Increase __clang_major__ check in Float16.h to 16

steven-johnson:


zvookin:
(Per issue we ran into internally. Looks like 15 may not support this on all platforms or it got added at some point in 15's development and wasn't there all along. Or maybe it's still iffy in 16 on some platforms too.)

zvookin (approved):

PR #7225, opened 2022-12-09
Improved halide_popcount

Aelphy:
This is a slightly more efficient version of a popcount, which has as many iterations as many ones are in the binary representation.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -156,7 +156,9 @@ inline int halide_popcount_fallback(T a) {
     int bits_set = 0;
     while (a != 0) {
         bits_set += 1;
-        // NOTE(aelphy): remove least significant zeros and the first met one
+        // this is Brian Kernigan's bit counting algorithm
+        // https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetKernighan)
+        // removes the least significant one
         a &= a - 1;
     }
     return bits_set;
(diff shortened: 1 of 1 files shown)


src/CodeGen_C.cpp
```diff
@@ -155,8 +155,9 @@ template<typename T>
 inline int halide_popcount(T a) {
     int bits_set = 0;
     while (a != 0) {
```
steven-johnson, line 157: FWIW, ~all modern compilers should have popcount as an intrinsic -- see `popcount64` in Util.cpp, maybe we should just adapt that here

steven-johnson:
What's the story on this, ready to land?

src/CodeGen_C.cpp
```diff
@@ -152,15 +152,39 @@ inline float float_from_bits(uint32_t bits) {
 }
 
 template<typename T>
-inline int halide_popcount(T a) {
+inline int halide_popcount_fallback(T a) {
```
vksnk, line 155: I would leave a link to the source for this algorithm for the reference (found this one https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetKernighan, for example).

src/CodeGen_C.cpp
```diff
@@ -152,15 +152,39 @@ inline float float_from_bits(uint32_t bits) {
...
     int bits_set = 0;
     while (a != 0) {
-        bits_set += a & 1;
-        a >>= 1;
+        bits_set += 1;
+        // NOTE(aelphy): remove least significant zeros and the first met one
```
vksnk, line 159: I would just make it a regular comment (i.e. remove NOTE(aelphy) part).

vksnk (approved):

steven-johnson:
Where does this PR stand?

PR #7227, opened 2022-12-09
Remove rogue string suffix in simd_op_check_arm.cpp

steven-johnson:
Interestingly, it compiles here, but in some compilers it will fail with "unexpected token".


abadams (approved):

rootjalex:
Looks like all the store checks in this file have the same rogue suffix - can you remove the others too (i.e. [here](https://github.com/halide/Halide/blob/16421a7aea8b5ca3d28b70ac34c5d969807b91b6/test/correctness/simd_op_check_arm.cpp#L1053) and above) ? I would but don't want to push to your branch.

steven-johnson:
> Looks like all the store checks in this file have the same rogue suffix - can you remove the others too (i.e. [here](https://github.com/halide/Halide/blob/16421a7aea8b5ca3d28b70ac34c5d969807b91b6/test/correctness/simd_op_check_arm.cpp#L1053) and above) ? I would but don't want to push to your branch.

done

PR #7230, opened 2022-12-12
Explicitly stage strided loads

abadams:
This PR adds a new compiler pass that converts strided loads into dense loads followed by shuffles. 

For a stride of two, the trick is to do a dense load of twice the size, and then extract either the even or odd lanes. This was previously done in codegen, where it was challenging, because it's not easy to know there if it's safe to do the double-sized load, as it either loads one element beyond or before the original load. We used the alignment of the ramp base to try to tell if it was safe to shift backwards, and we added padding to internal allocations so that for those at least it was safe to shift forwards. Unfortunately the alignment of the ramp base is usually unknown if you don't know anything about the strides of the input, and adding padding to allocations was a serious wart in our memory allocators.

This PR instead actively looks for evidence elsewhere in the Stmt (at some location which definitely executes whenever the load being transformed executes) that it's safe to read further forwards or backwards in memory. The evidence is in the form of a load at the same base address with a different constant offset. It also clusters groups of these loads so that they do the same dense load each and extract the appropriate slice of lanes. For loads from external buffers, it just does a shorter load, and for loads from internal allocations that we can't shift forwards or backwards, it adds padding to the allocation explicitly, via a new padding field on Allocate nodes.

Edit: Some backends don't like non-power-of-two sized vectors, so for loads from external buffers, we now split the load into two overlapping dense loads of half the size, then shuffle out appropriate lanes from each and concat the results.


steven-johnson:
The excellent PR description of what's going on now (and what we used to do) should be captured in a code comment somewhere.

src/CodeGen_ARM.cpp
```diff
@@ -770,6 +840,22 @@ void CodeGen_ARM::init_module() {
...
 
+void CodeGen_ARM::compile_func(const LoweredFunc &f,
+                               const string &simple_name,
+                               const string &extern_name) {
+    CodeGen_Posix::begin_func(f.linkage, simple_name, extern_name, f.args);
+
+    Stmt body = f.body;
+
+    if (target.os != Target::IOS && target.os != Target::OSX) {
+        body = SubstituteInStridedLoads().mutate(body);
```
rootjalex, line 851: A comment explaining why we don't perform this substitution for IOS or OSX would be helpful here

steven-johnson (approved):
LGTM -- don't see any red flags inside Google.

src/CodeGen_ARM.cpp
```diff
@@ -1123,6 +1209,28 @@ void CodeGen_ARM::visit(const Load *op) {
...
+    // vld2, vld3, vld4. So here we explicitly do the load as a single big dense
+    // load.
+    int stride = op->slice_stride();
+    const Load *load = op->vectors[0].as<Load>();
+    if (load &&
+        op->vectors.size() == 1 &&
+        2 <= stride && stride <= 4 &&
+        op->slice_begin() < stride &&
+        load->type.lanes() == stride * op->type.lanes() &&
+        target.os != Target::IOS && target.os != Target::OSX) {
```
rootjalex, line 1226: This condition seems really cheap to check, might be worth moving into a top-level if, to reduce the need to do the other checks? probably not super important, I just thought it was strange to check this last

src/StageStridedLoads.cpp
```diff
@@ -0,0 +1,366 @@
...
+                int64_t offset = 0;
+                const Add *base_add = base.as<Add>();
+                const int64_t *offset_ptr = base_add ? as_const_int(base_add->b) : nullptr;
+                if (offset_ptr) {
+                    base = base_add->a;
+                    offset = *offset_ptr;
+                }
+                // TODO: We do not yet handle nested vectorization here for
+                // ramps which have not already collapsed. We could potentially
+                // handle more interesting types of shuffle than simple flat slices.
```
rootjalex, line 99: Should this TODO be an open issue?

src/StageStridedLoads.cpp
```diff
@@ -0,0 +1,366 @@
...
+    const IRNode *scope = nullptr;
+
+    Scope<const Allocate *> allocation_scope;
+
+    std::map<const IRNode *, const IRNode *> parent_scope;
+
+protected:
+    void visit(const Load *op) override {
+        if (is_const_one(op->predicate)) {
+            Expr idx = substitute_in_all_lets(simplify(common_subexpression_elimination(op->index)));
```
rootjalex, line 85: Why is this series of transformations necessary? Hasn't simplification already been run before this pass?

rootjalex: A comment might help here

src/StageStridedLoads.cpp
```diff
@@ -0,0 +1,366 @@
...
+        }
+        IRVisitor::visit(op);
+    }
+
+    void visit(const For *op) override {
+        if (can_prove(op->extent > 0)) {
+            // The loop body definitely runs
+
+            // TODO: worry about different iterations of the loop somehow not
+            // providing the evidence we thought it did.
```
rootjalex, line 118: (same as above) should this TODO be an issue? Or a blocker for this PR? I'm also not quite sure what the TODO means.

src/StageStridedLoads.cpp
```diff
@@ -0,0 +1,366 @@
...
+    using IRMutator::visit;
+};
+
+}  // namespace
+
+Stmt stage_strided_loads(const Stmt &s) {
+    Stmt stmt = s;
+    // The following pass assumes that same variable name is the same variable,
+    // which is violated by the prior unrolling pass. Fix it up first.
+    stmt = uniquify_variable_names(stmt);
```
rootjalex, line 194: Should we just run this immediately following `unroll_loops` in Lower.cpp? There have been bugs with violations of this before iirc (in the bounds stuff I did a few years ago).

test/correctness/simd_op_check_arm.cpp
```diff
@@ -291,35 +291,39 @@ class SimdOpCheckARM : public SimdOpCheckTest {
...
-
-            // VLD4     X       -       Load Four-Element Structures
-            check(arm32 ? "vld4.8" : "ld4", 32 * w, in_i8(x * 4));
-            check(arm32 ? "vld4.8" : "ld4", 32 * w, in_u8(x * 4));
-            check(arm32 ? "vld4.16" : "ld4", 16 * w, in_i16(x * 4));
-            check(arm32 ? "vld4.16" : "ld4", 16 * w, in_u16(x * 4));
-            check(arm32 ? "vld4.32" : "ld4", 8 * w, in_i32(x * 4));
-            check(arm32 ? "vld4.32" : "ld4", 8 * w, in_u32(x * 4));
-            check(arm32 ? "vld4.32" : "ld4", 8 * w, in_f32(x * 4));
+            if (target.os != Target::IOS && target.os != Target::OSX) {
```
rootjalex, line 294: In the `else` case, should we make sure that we generate the regular load / shuffle commands we expect?

rootjalex (approved):
Lgtm with nits

PR #7237, opened 2022-12-13
[xtensa] DMA support improvements.

vksnk:
This includes multiple related changes:

* all transactions are 2D.
* Each buffer will use a separate DMA channel.
* For the case when destination is an output buffer, we can delay the wait for completion until the beginning of it's producer.

### DIFF
--- a/src/CodeGen_Xtensa.cpp
+++ b/src/CodeGen_Xtensa.cpp
@@ -90,11 +90,16 @@ extern int32_t halide_release_dma();
 #endif
 
 class ScopedDmaInitializer {
-  bool is_valid_ = false;
+  bool is_valid_;
  public:
   ScopedDmaInitializer(int channel_count) {
     is_valid_ = (halide_init_dma(channel_count) == 0);
   }
+
+  ScopedDmaInitializer() = delete;
+  ScopedDmaInitializer(const ScopedDmaInitializer&) = delete;
+  ScopedDmaInitializer& operator=(const ScopedDmaInitializer&) = delete;
+  ScopedDmaInitializer(ScopedDmaInitializer&&) = delete;
 
   ~ScopedDmaInitializer() {
     if (is_valid_) {
@@ -102,7 +107,7 @@ class ScopedDmaInitializer {
     }
   }
 
-  bool is_valid() { return is_valid_; }
+  bool is_valid() const { return is_valid_; }
 };
 
 )INLINE_CODE";
--- a/src/CodeGen_Xtensa.cpp
+++ b/src/CodeGen_Xtensa.cpp
@@ -52,8 +52,9 @@ private:
 
 protected:
     void visit(const Call *op) override {
-        if (op->name == "halide_xtensa_copy_1d") {
+        if ((op->name == "halide_xtensa_copy_1d") || (op->name == "halide_xtensa_copy_2d")) {
             uses_dma = true;
+            max_channel_no = std::max<int>(max_channel_no, *as_const_int(op->args[0]));
         }
 
         IRGraphVisitor::visit(op);
@@ -61,6 +62,7 @@ protected:
 
 public:
     bool uses_dma = false;
+    int max_channel_no = 0;
 };
 
 void CodeGen_Xtensa::add_platform_prologue() {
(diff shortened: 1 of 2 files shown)


src/CodeGen_Xtensa.cpp
```diff
@@ -77,9 +79,10 @@ extern "C" {
 
 extern void *halide_tcm_malloc(void *user_context, size_t x);
 extern void halide_tcm_free(void *user_context, void *ptr);
-extern int halide_init_dma();
-extern int32_t halide_xtensa_copy_1d(void* dst, int32_t dst_base, void* src, int32_t src_base, int extent, int item_size);
-extern int32_t halide_xtensa_wait_for_copy(int32_t id);
+extern int halide_init_dma(int channel_count);
+extern int32_t halide_xtensa_copy_1d(int32_t channel, void* dst, int32_t dst_base, void* src, int32_t src_base, int extent, int item_size);
```
steven-johnson, line 83: nit: it's a little weird to mix `int` and `int32_t` here; I'm sure it's fine, it's just that reading it made me wonder if the difference was necessary or just arbitrary.

src/CodeGen_Xtensa.cpp
```diff
@@ -88,8 +91,8 @@ extern int halide_release_dma();
 
 class ScopedDmaInitializer {
  public:
-  ScopedDmaInitializer() {
-    int status = halide_init_dma();
+  ScopedDmaInitializer(int channel_count) {
+    int status = halide_init_dma(channel_count);
```
steven-johnson, line 95: ...shouldn't we do something if there's an error? Just ignoring it seems dangerous.

steven-johnson (approved):

src/CodeGen_Xtensa.cpp
vksnk: Fixed, changed them all into int32_t.

src/CodeGen_Xtensa.cpp
vksnk: Added is_valid() method and corresponding error propogation.

src/CodeGen_Xtensa.cpp
```diff
@@ -77,25 +79,30 @@ extern "C" {
...
   }
 
   ~ScopedDmaInitializer() {
-    halide_release_dma();
+    if (is_valid_) {
+      halide_release_dma();
+    }
   }
+
+  bool is_valid() { return is_valid_; }
```
steven-johnson, line 105: probably want const method

src/CodeGen_Xtensa.cpp
```diff
@@ -77,25 +79,30 @@ extern "C" {
...
+extern int32_t halide_xtensa_copy_2d(int32_t channel, void *dst, int32_t dst_base, int32_t dst_stride, void *src, int32_t src_base, int32_t src_stride, int32_t extent0, int32_t extent1, int32_t item_size);
+extern int32_t halide_xtensa_wait_for_copy(int32_t channel);
+extern int32_t halide_release_dma();
 
 #ifdef __cplusplus
 }  // extern "C"
 #endif
 
 class ScopedDmaInitializer {
+  bool is_valid_ = false;
```
steven-johnson, line 93: Arguably redundant since we always initialize it in the ctor?

src/CodeGen_Xtensa.cpp
```diff
@@ -77,25 +79,30 @@ extern "C" {
...
  public:
-  ScopedDmaInitializer() {
-    int status = halide_init_dma();
-    (void)status;
+  ScopedDmaInitializer(int channel_count) {
```
steven-johnson, line 95: Since this is an RAII helper, probably good to `=delete` the default ctor and the move/copy ctors

src/CodeGen_Xtensa.cpp
```diff
@@ -195,7 +202,13 @@ void CodeGen_Xtensa::compile(const LoweredFunc &f, const std::map<std::string, s
             UsesDmaCopy uses_dma;
             body.accept(&uses_dma);
             if (uses_dma.uses_dma) {
-                stream << "ScopedDmaInitializer dma_initializer;\n";
+                stream << get_indent() << "ScopedDmaInitializer dma_initializer(" << uses_dma.max_channel_no + 1 << ");\n";
+                stream << get_indent() << "if (!dma_initializer.is_valid()) {\n";
+                stream << get_indent() << "halide_error("
```
steven-johnson, line 207: Random thought: maybe generate a local function of the form

int halide_error_dma_failure(void* user_context) {
    halide_error(user_context, "DMA failure");
    return halide_error_code_generic_error;
}

then you can make the failure clause simpler, and (possibly) slightly less code.

steven-johnson (approved):

PR #7243, opened 2022-12-14
Hoist vector slices using rewrite rules

abadams:
This lets us add associative variants more easily, which are helpful in the work on staging strided loads.


rootjalex:
correctness_intrinsics is giving me an infinite loop:
```
    frame #146848: 0x00000001087776c0 libHalide.dylib`Halide::Internal::(anonymous namespace)::deinterleave(Halide::Expr, int, int, int, Halide::Internal::Scope<void> const&) + 364
    frame #146849: 0x0000000108777b8c libHalide.dylib`Halide::Internal::extract_lane(Halide::Expr const&, int) + 92
    frame #146850: 0x0000000108c6397c libHalide.dylib`Halide::Internal::Simplify::visit(Halide::Internal::Shuffle const*, Halide::Internal::Simplify::ExprInfo*) + 1340
    frame #146851: 0x0000000108c1cfb8 libHalide.dylib`Halide::Internal::Simplify::visit(Halide::Internal::Mul const*, Halide::Internal::Simplify::ExprInfo*) + 9128
    frame #146852: 0x0000000108b29694 libHalide.dylib`Halide::Expr Halide::Internal::Simplify::simplify_let<Halide::Internal::Let, Halide::Expr>(Halide::Internal::Let const*, Halide::Internal::Simplify::ExprInfo*) + 9456
    frame #146853: 0x0000000108a8f500 libHalide.dylib`Halide::Internal::simplify(Halide::Expr const&, bool, Halide::Internal::Scope<Halide::Internal::Interval> const&, Halide::Internal::Scope<Halide::Internal::ModulusRemainder> const&) + 56
    frame #146854: 0x00000001087776c0 libHalide.dylib`Halide::Internal::(anonymous namespace)::deinterleave(Halide::Expr, int, int, int, Halide::Internal::Scope<void> const&) + 364
<repeated>
```

src/IRMatch.h
```diff
@@ -2080,6 +2080,66 @@ HALIDE_ALWAYS_INLINE auto cast(halide_type_t t, A &&a) noexcept -> CastOp<declty
...
+        lanes.make_folded_const(lanes_val, ty, state);
+        int l = (int)lanes_val.u.i64;
+        return Shuffle::make_slice(vec.make(state, type_hint), b, s, l);
+    }
+
+    constexpr static bool foldable = false;
+
+    HALIDE_ALWAYS_INLINE
+    SliceOp(Vec v, Base b, Stride s, Lanes l)
+        : vec(v), base(b), stride(s), lanes(l) {
```
rootjalex, line 2127: Could you add `static_assert`s here that enforce that base/stride/lanes are all `foldable`, so the templating error produced if one is not `foldable` is less verbose and points to the solution (instead of "missing call to `make_folded_const` errors).

src/Simplify_Add.cpp
```diff
@@ -112,6 +112,13 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
              rewrite(y*x + y*z, y*(x + z)) ||
              rewrite(x*c0 + y*c1, (x + y*fold(c1/c0)) * c0, c1 % c0 == 0) ||
              rewrite(x*c0 + y*c1, (x*fold(c0/c1) + y) * c1, c0 % c1 == 0) ||
+
+             rewrite(slice(x, c0, c1, c2) + slice(y, c0, c1, c2), slice(x + y, c0, c1, c2)) ||
+             rewrite(slice(x, c0, c1, c2) + (z + slice(y, c0, c1, c2)), slice(x + y, c0, c1, c2) + z) ||
+             rewrite(slice(x, c0, c1, c2) + (slice(y, c0, c1, c2) + z), slice(x + y, c0, c1, c2) + z) ||
+             rewrite(slice(x, c0, c1, c2) + (z - slice(y, c0, c1, c2)), slice(x - y, c0, c1, c2) + z) ||
+             rewrite(slice(x, c0, c1, c2) + (slice(y, c0, c1, c2) - z), slice(x + y, c0, c1, c2) - z) ||
```
rootjalex, line 120: Are these always valid via a cost model for shuffling (across all backends)? For example, suppose the stride was 8 (and lanes also 8) - wouldn't the sum of length-8 vectors + 2 shuffles be cheaper than a sum of length-64 vectors + 1 shuffle?

rootjalex: (same question for the Min/Max/Mul/Sub rules below)

src/IRMatch.h
```diff
@@ -2080,6 +2080,66 @@ HALIDE_ALWAYS_INLINE auto cast(halide_type_t t, A &&a) noexcept -> CastOp<declty
...
+    constexpr static IRNodeType max_node_type = IRNodeType::Shuffle;
+    constexpr static bool canonical = Vec::canonical && Base::canonical && Stride::canonical && Lanes::canonical;
+
+    template<uint32_t bound>
+    HALIDE_ALWAYS_INLINE bool match(const BaseExprNode &e, MatcherState &state) const noexcept {
+        if (e.node_type != IRNodeType::Shuffle) {
+            return false;
+        }
+        const Shuffle &v = (const Shuffle &)e;
+        return v.vectors.size() == 1 &&
```
rootjalex, line 2103: `hoist_slice_vector` doesn't seem to have the restriction that slices are of a single input vector, so doesn't this change weaken simplification?

src/Simplify_Add.cpp
abadams: If stride is 8 and lanes is 8, then you're extracting one element from each vector, so the 2 shuffles are actually 16 shuffles. It's 16 shuffles and one add vs 8 adds and 8 shuffles.

abadams: More generally, for any binary single-cycle op that consumes two inputs, it should be better to hoist the slice.

rootjalex: Okay cool. Sgtm

src/IRMatch.h
abadams: Shuffle::make_slice takes a single vector as an argument. I doubt there are any cases in practice of slicing multiple vectors.

abadams: Shuffle::make_slice takes a single vector as an argument. I doubt there are any cases in practice of slicing multiple vectors.

src/IRMatch.h
abadams: Done

abadams:
Turns out the shuffle visitor wants to sink extract_element calls, and extract_element counts as a degenerate slice_vector, so it was also being hoisted. I restricted my rewrite rules to non-degenerate slice_vectors

src/IRMatch.h
rootjalex: Okay

rootjalex:
> Turns out the shuffle visitor wants to sink extract_element calls, and extract_element counts as a degenerate slice_vector, so it was also being hoisted. I restricted my rewrite rules to non-degenerate slice_vectors

Might be worth adding a comment about this on this above the rules.

src/Simplify_Add.cpp
```diff
@@ -113,6 +113,9 @@ Expr Simplify::visit(const Add *op, ExprInfo *bounds) {
              rewrite(x*c0 + y*c1, (x + y*fold(c1/c0)) * c0, c1 % c0 == 0) ||
              rewrite(x*c0 + y*c1, (x*fold(c0/c1) + y) * c1, c0 % c1 == 0) ||
 
+             // Hoist shuffles. The Shuffle visitor wants to sink
+             // extract_elements to the leaves, and those count as degenerate
+             // shuffles, so only hoist shuffles that grab more than one lane.
```
rootjalex, line 118: “shuffles” -> “slices”? (here and below)

steven-johnson:
Any update on this PR?

steven-johnson:
How risky do you think this is? Does it need a preflight test inside google?

abadams:
It's probably worth a preflight test.

steven-johnson:
At least one failure inside google:

```
Internal Error at third_party/halide/halide/src/IR.cpp:42 triggered by user code at : Condition failed: a.type() == b.type(): Add of mismatched types
*** SIGABRT received by PID 424 (TID 424) on cpu 30 from PID 424; stack trace: ***
PC: @     0x7fa7c0bf1347  (unknown)  gsignal
    @     0x7fa6bfc876dd        256  base/process_state.cc:1202 FailureSignalHandler()
    @     0x7fa801f071c0  (unknown)  (unknown)
    @     0x7fa77537b003        160  third_party/halide/halide/src/Error.cpp:221 Halide::Internal::ErrorReport::~ErrorReport()
    @     0x7fa775567736        160  third_party/halide/halide/src/IR.cpp:42 Halide::Internal::Add::make()
    @     0x7fa77588f0ea        192  third_party/halide/halide/src/IRMatch.h:736 Halide::Internal::IRMatcher::Rewriter<>::build_replacement<>()
    @     0x7fa775860dd9        224  third_party/halide/halide/src/IRMatch.h:2916 Halide::Internal::Simplify::visit()::$_0::operator()()
    @     0x7fa775857cfa        256  third_party/halide/halide/src/Simplify_Add.cpp:40 Halide::Internal::Simplify::visit()
    @     0x7fa775838a2f        128  third_party/halide/halide/src/IRVisitor.h:184 Halide::Internal::VariadicVisitor<>::dispatch_expr<>()
    @     0x7fa775b875f1        224  third_party/halide/halide/src/IRVisitor.h:341 Halide::Internal::Simplify::visit()
    @     0x7fa775838a51        128  third_party/halide/halide/src/IRVisitor.h:186 Halide::Internal::VariadicVisitor<>::dispatch_expr<>()
    @     0x7fa775af8143        288  third_party/halide/halide/src/IRVisitor.h:341 Halide::Internal::Simplify::visit()
    @     0x7fa775838a95        128  third_party/halide/halide/src/IRVisitor.h:190 Halide::Internal::VariadicVisitor<>::dispatch_expr<>()
    @     0x7fa77585771f        256  third_party/halide/halide/src/IRVisitor.h:341 Halide::Internal::Simplify::visit()
    @     0x7fa775838a2f        128  third_party/halide/halide/src/IRVisitor.h:184 Halide::Internal::VariadicVisitor<>::dispatch_expr<>()
```

rootjalex:
Seems like all of the rules need a check that the lanes of `x` and `y` match

abadams:
Good point, that's probably the issue.

steven-johnson:
(re-requested review so this doesn't accidentally get merged yet)

rootjalex:
@steven-johnson Could you check inside Google again, when you have the chance? I think the bug you're seeing should be fixed now.

steven-johnson:
will do

steven-johnson:
Running check now. Does this PR need a test?

rootjalex:
> Running check now. Does this PR need a test?

Good point. I'll try to find time this afternoon to add a test for the improved hoisting and another one for the bug you found.

abadams:
Thanks for taking this over, AJ.

rootjalex:
Failures seem unrelated? One build failure and a mullapaudi issue (mullapudi2016). @steven-johnson is performance_inner_loop_parallel sometimes flaky? Or is that something I should investigate?

steven-johnson:
Both of those are known to be flaky, unfortunately We need to track them down but they aren't related.

steven-johnson:
Looks clean inside Google.

PR #7261, opened 2022-12-29
Inline into extern function args during bounds inference

### LINKED ISSUE #7260: Inliner doesn't properly handle all Expr args for `define_extern()` Funcs
Given something like this:

```
class MyGen {
  ...  
  Input<Buffer<float, 2>> arg_{"arg"};
  ...
};
void MyGen::generate() {
     for (int i = ...) {
      Expr arg_0 = arg_(i, 0);
      Expr arg_1 = arg_(i, 1);
      ...
      Func foo;
      foo.define_extern("foo", {arg_0, arg_1, ...}, ...);
}
```

...you'll fail to build with something like `Symbol not found: arg_im.min.0`.

After a bit of sleuthing, I think the issue here is that while we do correctly inline the `arg_im` function in `schedule_functions()`, we don't take into account that `foo.extern_arguments()` contains Exprs that refer to inlined functions. While `validate_schedule()` attempts to validate that none of the extern args are inlined, it only does so for Funcs... not for Exprs that reference inlined Funcs, like we have here.

There are ways to hack this into working (e.g. wrap arg_0 in a compute_root()'ed Func), but they are awkward.

(Genuinely surprised we've never seen this one before!)

abadams:
Fixes #7260

### DIFF
--- a/src/BoundsInference.cpp
+++ b/src/BoundsInference.cpp
@@ -234,7 +234,7 @@ protected:
             if (to_inline.count(f)) {
                 auto args = mutate(op->args);
                 Expr body = get_qualified_body(f, op->value_index);
-                const vector<string> func_args = f.args();
+                const vector<string> &func_args = f.args();
                 for (size_t i = 0; i < args.size(); i++) {
                     body = Let::make(f.name() + "." + func_args[i], args[i], body);
                 }
--- a/src/BoundsInference.cpp
+++ b/src/BoundsInference.cpp
@@ -199,6 +199,23 @@ bool is_fused_with_others(const vector<vector<Function>> &fused_groups,
     return false;
 }
 
+class Inliner : public IRMutator {
+public:
+    std::set<Function, Function::Compare> to_inline;
+
+    Expr visit(const Call *op) override {
+        if (op->func.defined()) {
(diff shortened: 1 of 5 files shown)


src/BoundsInference.cpp
```diff
@@ -199,6 +201,32 @@ bool is_fused_with_others(const vector<vector<Function>> &fused_groups,
     return false;
 }
 
+class Inliner : public IRMutator {
```
steven-johnson, line 204: nit: worth inserting a comment here about why this is necessary in addition to the previously existing inliner elsewhere

steven-johnson (changes requested):
Looks good so far, but every bot is failing...

abadams:
Failing the test python_correctness_realize_warnings because it only prints the warning twice instead of three times. I'm confused about why it would print it more than once. Any ideas?

abadams:
I just changed the test to assert that the warning is printed at least once, and that nothing else is printed.

steven-johnson:
I guess the more interesting question here is how/why this PR caused any difference in that at all, whether or not the test made good sense in the first place...

steven-johnson (approved):
LGTM

PR #7264, opened 2023-01-05
Update change following LLVM WASM change f841ad30d77eeb4c51663e68efefdb734c7a3d07

alinas:
Update test following LLVM WASM change https://github.com/llvm/llvm-project/commit/f841ad30d77eeb4c51663e68efefdb734c7a3d07

### DIFF
--- a/test/correctness/simd_op_check_wasm.cpp
+++ b/test/correctness/simd_op_check_wasm.cpp
@@ -387,11 +387,11 @@ public:
                 // check("v128.load32_zero", 2 * w, in_u32(0));
                 // check("v128.load64_zero", 2 * w, in_u64(0));
 
-                // Load vector with identical lanes
-                check("v128.load8_splat", 16 * w, in_u8(0));
-                check("v128.load16_splat", 8 * w, in_u16(0));
-                check("v128.load32_splat", 4 * w, in_u32(0));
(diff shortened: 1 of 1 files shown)


test/correctness/simd_op_check_wasm.cpp
```diff
@@ -387,11 +387,11 @@ class SimdOpCheckWASM : public SimdOpCheckTest {
                 // check("v128.load32_zero", 2 * w, in_u32(0));
                 // check("v128.load64_zero", 2 * w, in_u64(0));
 
-                // Load vector with identical lanes
-                check("v128.load8_splat", 16 * w, in_u8(0));
-                check("v128.load16_splat", 8 * w, in_u16(0));
-                check("v128.load32_splat", 4 * w, in_u32(0));
-                check("v128.load64_splat", 2 * w, in_u64(0));
+                // Load vector with identical lanes generates *.splat.
+                check("i8x16.splat", 16 * w, in_u8(0));
```
steven-johnson, line 391: This will probably need to have a check for `internal::llvm_version()` so that tests running against LLVM15 (which we still test on main) will pass, there should be an example of that elsewhere in one of the simd_op_checks

zvookin (approved):

PR #7269, opened 2023-01-10
Optimize Module::compile() for some edge cases

steven-johnson:
Avoid redundant `compile_to_buffer()` calls for output requests that can't possibly ever need them.

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -538,11 +538,10 @@ void Module::compile(const std::map<OutputFileType, std::string> &output_files) 
     // ever rely on submodules (e.g.: toplevel declarations in C/C++), don't bother resolving
     // the submodules, which can call compile_to_buffer().
     const auto should_ignore_submodules = [](const std::map<OutputFileType, std::string> &output_files) {
-        auto output_files_copy = output_files;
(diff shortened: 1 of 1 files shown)


src/Module.cpp
```diff
@@ -534,10 +534,21 @@ void Module::compile(const std::map<OutputFileType, std::string> &output_files)
         Internal::print_to_html(output_files.at(OutputFileType::stmt_html), *this);
     }
 
+    // Minor but worthwhile optimization: if all of the output files are of types that won't
+    // ever rely on submodules (e.g.: toplevel declarations in C/C++), don't bother resolving
+    // the submodules, which can call compile_to_buffer().
+    const auto should_ignore_submodules = [](const std::map<OutputFileType, std::string> &output_files) {
+        auto output_files_copy = output_files;
```
abadams, line 541: A non-mutating way to do this would be `output_files.size() == output_files.count(...) + output_files.count(...) + output_files.count(...)`

steven-johnson: Lazy programmer was lazy, done

abadams (approved):

PR #7270, opened 2023-01-11
Fix declaration of extern fns in C backend that use __user_context

steven-johnson:
TL;DR: the headers declare it `void const *` but our internal declarations declared it `void *`; if we had a C++-backend piece of code that had a define_extern to another C++-backend piece of code (and both were C++ and both used user_context), the name mangling was wrong and linking would fail. Whew.


zvookin:
(This changes the internal code to match the header and thus should not be a breaking change. The expectation is what it changes would not have worked before.)

zvookin (approved):

steven-johnson:
grr, this is apparently trickier than meets the eye, as 'fixing' the all-C++ case broke ~every other case -- ie, apparently LLVM-generated code assumes ucon is `void*` instead of `void const *`. (Why we declare ucon differently in different cases is a clear example of unforced error on our part.) It may be that just uniformly making it plain `void *` in all cases would mostly fix things, but this feels like a rathole. 

I'm going to convert this to a draft PR until I have a bit more time to look at it (the case I wanted to fix is probably not going to land anytime soon).

PR #7275, opened 2023-01-13
Unify user_context name-and-type everywhere

steven-johnson:
For reasons that aren't clear, we sometimes emitted user_context as `void*` or `Handle()` (most places), but sometimes as `const void *` (Codegen_C). This attempts to unify it to be plain `void *` everywhere, to avoid the need for const-cast and also void possibilities of c++ name-mangling linker mismatches in some obscure cases.

This is technically a breaking change, since existing code may be passing in `const void *` pointers; I think this is arguably acceptable, since the breakage will be very obvious and very straightforward to fix.

Alternately, I considered unifying on `const void *`, but that seems more restrictive at the call site, and makes for a weird mismatch with our runtime functions (which are all uniformly `void *`).

I'm not entirely happy with the breakage, but the status quo is clearly wrong, and really, has always been.

### DIFF
--- a/src/Callable.cpp
+++ b/src/Callable.cpp
@@ -62,7 +62,7 @@ Callable::Callable(const std::string &name,
 
     contents->quick_call_check_info.reserve(contents->jit_cache.arguments.size());
     for (const Argument &a : contents->jit_cache.arguments) {
-        const auto qcci = (a.name == "__user_context") ?
+        const auto qcci = (a.name == user_context_name()) ?
                               Callable::make_ucon_qcci() :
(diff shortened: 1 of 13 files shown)


steven-johnson:
Unfortunately, this change would be problematic downstream: there is a *lot* of user code inside Google that assumes it can pass in a const-something-ptr as user_context; adding const_cast (or similar changes) everywhere would be a nightmare. I'm going to try going in the other direction, and making all user-generated code be consistently `const void*` in all cases (maintaining the runtime code as `void *` for now).

PR #7276, opened 2023-01-13
(V2) Unify user_context name-and-type everywhere

steven-johnson:
For reasons that aren't clear, we sometimes emitted user_context as `void*` or `Handle()` (most places), but sometimes as `const void *` (Codegen_C). This attempts to unify it to be `const void *` for all user-generated code, to avoid the need for const-cast and also void possibilities of c++ name-mangling linker mismatches in some obscure cases.


steven-johnson:
This looks more promising, PTAL

abadams (approved):

steven-johnson:
The gotcha with this one is that ~every `define_extern()` callback that is currently defined as `void *user_context` needs to become `const void *user_context` -- a lot of churn, not sure yet if a dealbreaker.

steven-johnson:
> The gotcha with this one is that ~every `define_extern()` callback that is currently defined as `void *user_context` needs to become `const void *user_context` -- a lot of churn, not sure yet if a dealbreaker.

...and, by implication, any `extern` function that calls a halide runtime routine (eg halide_print, halide_error) needs to do a const_cast. Oy. Maybe we were better off keeping everything void*. What a mess.

steven-johnson:
OK, so both #7275 and #7276 PRs are fatally flawed, in that both require (1) a lot of churn, (2) a lot of const_casts (just in different places), and (3) are still fundamentally unsatisfactory in that we have untyped void* being thrown around everywhere. I'm going to leave these open as draft PRs for the moment, but I don't think either should land. 

For the moment, we have a situation where this (mostly) works, with a weird corner case that only turns up in one compiler case that I may be able to finesse around, so I may put this on the back burner. For the longer term, if we accept that anything that attempts to clean up this mess will require nontrivial churn no matter what, I think I have a design idea that might be worth exploring -- I'll need to write a design proposal for it, though.

PR #7277, opened 2023-01-14
[xtensa] set is_tcm to False when working with buffer

Aelphy:


vksnk (commented):
It seems that we also use external_buffers.count(op->name) to do the same check elsewhere in this function, so I would use it here as well to make it more uniform (even if it's equivalent to what you have).

vksnk (approved):

PR #7279, opened 2023-01-16
[HVX] Fix EliminateInterleaves

### LINKED ISSUE #7100: [HVX] EliminateInterleaves is broken
On a benchmark I see the following pattern enter EliminateInterleaves:
```
x + interleave(y)
```
The output is:
```
interleave(deinterleave(x) + y)
```

I understand that EliminateInterleaves is supposed to push `interleave` towards the top of an `Expr`, but producing two swizzles from one swizzle seems like it does not follow a good cost model. I think the issue is simple to fix, this line:
https://github.com/halide/Halide/blob/8f210f1167e7454d7e33b38a1f9e583c29695070/src/HexagonOptimize.cpp#L1419

should probably use a `>` instead of a `>=`. I'm only opening this issue in case there's something more complicated that I don't understand happening - otherwise I'll just open a PR to fix this.

I also think FuseInterleaves should be merged with EliminateInterleaves, as they seem to essentially perform the same function, but FuseInterleaves only tries to fix widening ops while EliminateInterleaves tries to fix narrowing ops (in the `Call` visitor).

@pranavb-ca @steven-johnson

rootjalex:
Minor fix for an edge case that introduces more shuffling.

Fixes #7100


pranavb-ca (approved):

steven-johnson:
Huh, the win32 failure is one I've never seen before:

```
FAILED: bin/li2018_demo_gradient_autoscheduler.exe 
li2018_demo_gradient.runtime.lib(li2018_demo_gradient.runtime.obj) : error LNK2019: unresolved external symbol ___floatundisf referenced in function _halide_profiler_report_unlocked
```

rootjalex:
> Huh, the win32 failure is one I've never seen before:
> 
> ```
> FAILED: bin/li2018_demo_gradient_autoscheduler.exe 
> li2018_demo_gradient.runtime.lib(li2018_demo_gradient.runtime.obj) : error LNK2019: unresolved external symbol ___floatundisf referenced in function _halide_profiler_report_unlocked
> ```

That is super weird, and entirely unrelated to this PR... Strange.

steven-johnson:
> > Huh, the win32 failure is one I've never seen before:
> > ```
> > FAILED: bin/li2018_demo_gradient_autoscheduler.exe 
> > li2018_demo_gradient.runtime.lib(li2018_demo_gradient.runtime.obj) : error LNK2019: unresolved external symbol ___floatundisf referenced in function _halide_profiler_report_unlocked
> > ```
> 
> That is super weird, and entirely unrelated to this PR... Strange.

Maybe related to something new/different in LLVM codegen?

steven-johnson:
> Runtime Function: float __floatundisf (unsigned long i)
> Runtime Function: double __floatundidf (unsigned long i)
> Runtime Function: long double __floatunditf (unsigned long i)
> Runtime Function: long double __floatundixf (unsigned long i)
> These functions convert i, an unsigned long, to floating point.
> 
>

rootjalex:
That test is this file: https://github.com/halide/Halide/blob/main/test/autoschedulers/li2018/demo_generator.cpp , right? I don't even see an int -> float conversion in that code.

rootjalex:
Oh it's in the profiler. Hmm

steven-johnson:
So I experimentally crosscompiled `_halide_profiler_report_unlocked` to x86-32-windows-profile -- I don't see any calls to  floatundisf there, but looking at the .ll file, there are a number of `uitofp` instructions, which could be the culprit. That said... that would be used all over the place, so why is the failure being reported only there?

steven-johnson:
Anyway, this is clearly unrelated, feel free to land.

steven-johnson:
Still finishing the bisect but the culprit is almost certainly https://reviews.llvm.org/D141074

PR #7282, opened 2023-01-17
Fix issue in find_package in cross-compilation for no OS

stevesuzuki-arm:
When using toolchain where Threads libs are not available, which is the case in baremetal target cross-compilation, we were not able to load even HalideHelpers pacakge.

### DIFF
--- a/README_cmake.md
+++ b/README_cmake.md
@@ -750,6 +750,8 @@ Variables that control package loading:
 | Variable             | Description                                                                                                                                                                   |
 |----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `Halide_SHARED_LIBS` | override `BUILD_SHARED_LIBS` when loading the Halide package via `find_package`. Has no effect when using Halide via `add_subdirectory` as a Git or `FetchContent` submodule. |
+| `Halide_RUNTIME_NO_THREADS` | skip linking of Threads libraray to runtime. Should be set if your toolchain does not support it (e.g. baremetal). |
+| `Halide_RUNTIME_NO_DL_LIBS` | skip linking of DL libraray to runtime. Should be set if your toolchain does not support it (e.g. baremetal). |
 
 Variables set by the package:
 
--- a/apps/HelloBaremetal/cmake/toolchain.noos-arm32-sample.cmake
+++ b/apps/HelloBaremetal/cmake/toolchain.noos-arm32-sample.cmake
@@ -29,8 +29,8 @@ set(CMAKE_EXE_LINKER_FLAGS "-z noexecstack")
 set(Halide_TARGET "arm-32-noos-semihosting")
 
 # To prevent Threads and DL libs from being linked to runtime, as this toolchain doesn't have them
-set(HALIDE_RUNTIME_NO_THREADS ON)
-set(HALIDE_RUNTIME_NO_DL_LIBS ON)
+set(Halide_RUNTIME_NO_THREADS ON)
+set(Halide_RUNTIME_NO_DL_LIBS ON)
 
 # Switch for baremetal specific build steps
 set(BAREMETAL ON)
--- a/apps/HelloBaremetal/cmake/toolchain.noos-arm32-sample.cmake
+++ b/apps/HelloBaremetal/cmake/toolchain.noos-arm32-sample.cmake
@@ -22,11 +22,15 @@ set(CMAKE_CXX_FLAGS "${C_COMMON_FLAGS}")
 set(CMAKE_C_FLAGS "${C_COMMON_FLAGS}")
 set(CMAKE_ASM_FLAGS "${C_COMMON_FLAGS}")
 
-# To surpress linker warning "missing .note.GNU-stack section" by GCC 12
+# To suppress linker warning "missing .note.GNU-stack section" by GCC 12
 set(CMAKE_EXE_LINKER_FLAGS "-z noexecstack")
 
 # Halide target for Halide Generator
 set(Halide_TARGET "arm-32-noos-semihosting")
+
+# To prevent Threads and DL libs from being linked to runtime, as this toolchain doesn't have them
+set(HALIDE_RUNTIME_NO_THREADS ON)
+set(HALIDE_RUNTIME_NO_DL_LIBS ON)
 
 # Switch for baremetal specific build steps
 set(BAREMETAL ON)
--- a/cmake/HalideGeneratorHelpers.cmake
+++ b/cmake/HalideGeneratorHelpers.cmake
@@ -216,7 +216,7 @@ function(add_halide_library TARGET)
     ##
 
     set(options C_BACKEND GRADIENT_DESCENT)
-    set(oneValueArgs FROM GENERATOR FUNCTION_NAME NAMESPACE USE_RUNTIME AUTOSCHEDULER HEADER ${extra_output_names})
+    set(oneValueArgs FROM GENERATOR FUNCTION_NAME NAMESPACE USE_RUNTIME AUTOSCHEDULER HEADER ${extra_output_names} NO_THREADS NO_DL_LIBS)
     set(multiValueArgs TARGETS FEATURES PARAMS PLUGINS)
     cmake_parse_arguments(ARG "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
@@ -314,7 +314,16 @@ function(add_halide_library TARGET)
         set(ARG_USE_RUNTIME Halide::Runtime)
     elseif (NOT ARG_USE_RUNTIME)
         # If we're not using an existing runtime, create one.
-        add_halide_runtime("${TARGET}.runtime" TARGETS ${ARG_TARGETS} FROM ${ARG_FROM})
+
+        # To forward NO_THREADS/NO_DL_LIBS args to add_halide_runtime()
+        if (DEFINED ARG_NO_THREADS)
+            set(CALL_ARG_NO_THREADS NO_THREADS ${ARG_NO_THREADS})
+        endif ()
+        if (DEFINED ARG_NO_DL_LIBS)
+            set(CALL_ARG_NO_DL_LIBS NO_DL_LIBS ${ARG_NO_DL_LIBS})
+        endif ()
+
+        add_halide_runtime("${TARGET}.runtime" TARGETS ${ARG_TARGETS} FROM ${ARG_FROM} ${CALL_ARG_NO_THREADS} ${CALL_ARG_NO_DL_LIBS})
         set(ARG_USE_RUNTIME "${TARGET}.runtime")
     elseif (NOT TARGET ${ARG_USE_RUNTIME})
         message(FATAL_ERROR "Invalid runtime target ${ARG_USE_RUNTIME}")
@@ -556,7 +565,7 @@ endfunction()
 
 function(add_halide_runtime RT)
     set(options "")
-    set(oneValueArgs FROM)
+    set(oneValueArgs FROM NO_THREADS NO_DL_LIBS)
     set(multiValueArgs TARGETS)
     cmake_parse_arguments(ARG "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
@@ -581,6 +590,15 @@ function(add_halide_runtime RT)
             add_executable(_Halide_gengen ALIAS ${ARG_FROM})
         endif()
     endif()
+
+    # The default of NO_THREADS/NO_DL_LIBS is OFF unless HALIDE_RUNTIME_NO_THREADS/NO_DL_LIBS is defined globally
+    if (NOT DEFINED ARG_NO_THREADS)
+        set(ARG_NO_THREADS ${HALIDE_RUNTIME_NO_THREADS})
+    endif ()
+    if (NOT DEFINED ARG_NO_DL_LIBS)
+        set(ARG_NO_DL_LIBS ${HALIDE_RUNTIME_NO_DL_LIBS})
+    endif ()
+
     # Ensure _Halide_gengen is defined
     _Halide_gengen_ensure()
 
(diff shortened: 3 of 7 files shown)


packaging/common/HalideHelpersConfig.cmake
```diff
@@ -2,10 +2,11 @@ cmake_minimum_required(VERSION 3.22)
...
 
-include(CMakeFindDependencyMacro)
-
-set(THREADS_PREFER_PTHREAD_FLAG TRUE)
-find_dependency(Threads)
+if (NOT (DEFINED Halide_NO_THREADS AND Halide_NO_THREADS))
+    include(CMakeFindDependencyMacro)
+    set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+    find_dependency(Threads)
+endif()
```
alexreinking, line 10: If `Threads` is only needed for the runtime libraries and not for libHalide, then this logic should be moved to HalideGeneratorHelpers, in the if condition above.

It's a bit smelly here that the package config file references user-settable variables when determining which dependencies are needed by the package.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -608,7 +608,10 @@ function(add_halide_runtime RT)
         _Halide_fix_xcode("${RT}")
     endif ()
 
-    target_link_libraries("${RT}" INTERFACE Halide::Runtime Threads::Threads ${CMAKE_DL_LIBS})
+    target_link_libraries("${RT}" INTERFACE Halide::Runtime)
+    if (NOT (DEFINED Halide_NO_THREADS AND Halide_NO_THREADS))
+        target_link_libraries("${RT}" INTERFACE Threads::Threads ${CMAKE_DL_LIBS})
+    endif ()
     _Halide_add_targets_to_runtime("${RT}" TARGETS ${ARG_TARGETS})
```
alexreinking, line 615: There should probably be something here like `if (NOT TARGET Threads::Threads)` followed by a `find_package(Threads REQUIRED)`

README_cmake.md
```diff
@@ -751,6 +751,7 @@ Variables that control package loading:
 | Variable             | Description                                                                                                                                                                   |
 |----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `Halide_SHARED_LIBS` | override `BUILD_SHARED_LIBS` when loading the Halide package via `find_package`. Has no effect when using Halide via `add_subdirectory` as a Git or `FetchContent` submodule. |
+| `Halide_NO_THREADS` | skip checking dependency of Threads libraray. Should be set if your toolchain does not support it (e.g. baremetal). In such a situation, some of the Halide targets using Threads (e.g. `Halide::RunGenMain`) cannot be built. |
```
alexreinking, line 754: This comment makes me think that the correct solution is to put something more like

```
set(Halide_NO_THREADS "@HALIDE_NO_THREADS@")
```

In the package config file and to make sure early in the build that the variable has a value (to avoid definedness checks)

alexreinking (changes requested):
My comments are a bit mixed, but I wanted to keep my whole thought process around.

What I'm not sure about is whether the lack of threads must be fixed upon package export (i.e. does it constitute a package variant?) or if it can truly be set by the loader of the package.

In the latter case, we should move targets that depend on threads to a separate export set, so they aren't even created in the user's build when `Halide_NO_THREADS` is set.

README_cmake.md
stevesuzuki-arm: I'm not sure if I understand the points correctly. Should `Halide_NO_THREADS` be set inside package config file depending on find_package(Threads) fails or not, rather than by the user calling find_package(Halide or HalideHelpers)? Is that what you mean to say?

My assumption when I uploaded this PR is the usage shown in #7286.
About the dependency, the targets which depends on Threads seem to be `Halide::Halide`, `Halide::Generator` and `Halide::RunGenMain`. And `add_halide_runtime` links Threads to application runtime.

alexreinking: Let me ask my question a different way: would re-naming this variable to `Halide_RUNTIME_NO_THREADS` more accurately describe its use?

It doesn't seem valid to exclude `find_dependency(Threads)` on the same path that loads `Halide::Halide`.

stevesuzuki-arm: I agree with your points. The problem is `find_package(HalideHelpers)` for runtime and `find_package(Halide)` for Generator/libHalide go through the same path.
What we want to do is to generate runtime and function (image processing) even if toolchain does not have `Threads`. So we need some mechanism for `find_package` that doesn't break in that particular use case.

alexreinking:
In #7297 we removed the dependency on `Threads::Threads` for `Halide::Halide`, `Halide::Generator`, and `Halide::RunGenMain`. I wonder if this change can be improved in light of that.

stevesuzuki-arm:
> In #7297 we removed the dependency on `Threads::Threads` for `Halide::Halide`, `Halide::Generator`, and `Halide::RunGenMain`. I wonder if this change can be improved in light of that.

The latest commit no longer uses `Halide_NO_THREADS`.

Alternative approach where user explicitly declares is to add the option argument `NO_THREADS` to `add_halide_library` and `add_halide_runtime` to skip linking `Threads`.

steven-johnson:
Where does this PR stand?

steven-johnson:
Ping for status

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -608,7 +608,16 @@ function(add_halide_runtime RT)
...
-    target_link_libraries("${RT}" INTERFACE Halide::Runtime Threads::Threads ${CMAKE_DL_LIBS})
+    target_link_libraries("${RT}" INTERFACE Halide::Runtime)
+    if (NOT TARGET Threads::Threads)
+        set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+        find_package(Threads QUIET)
+    endif()
+    if (TARGET Threads::Threads)
+        target_link_libraries("${RT}" INTERFACE Threads::Threads ${CMAKE_DL_LIBS})
+    else()
+        message(STATUS "Package Threads not found, skip linking it to Halide::Runtime")
```
alexreinking, line 619: This is going to search for threads on every invocation if it is not present. Thus this message will print once per call and be quite noisy. Can we change this to write to a global property like `Halide_RUNTIME_NO_THREADS` that can be checked to avoid the search and warning?

Also, I'm not sure why `${CMAKE_DL_LIBS}` is only applied when `Threads::Threads` is around. Shouldn't it always be included?

alexreinking (changes requested):
Simple change for less-noisy UX.

alexreinking:
Overall, I'm a little bit concerned about _how_ this PR determines whether or not a runtime will get a dependency on `Threads::Threads`.

The "real" problem as I understand it is that, currently, pipeline libraries that don't actually use threads in their schedules get a spurious dependency on `Threads::Threads`. Most of the time, this is benign, but on target machines that don't support threads in the first place, this becomes an error. Right?

Thus, it seems like the most correct solution would be to only add `Threads::Threads` if any schedule in the library uses `.parallel()`. However, this would be _extremely_ complex to implement without at least requiring all Halide generators to be in a separate package from the libraries they produce, which makes the common cases as hard as the cross-compiling cases. It would also require work on the generator side to quickly check that property (but not run the whole compiler, which would be slow). Not awesome.

So we're electing to use a proxy instead. In this PR, that proxy is "whether or not the `Threads` package can be found". I think that's too loose, especially for larger projects.

Better would be to take a two step approach:

1. Add an argument to `add_halide_library` that passes through to `add_halide_runtime` (both functions have the argument) that optionally disables threads. Say, `NO_THREADS`. When `NOT NO_THREADS` and `NOT TARGET Threads::Threads`, then we `find_package(Threads REQUIRED)` inside `add_halide_runtime`.
2. The default for `NO_THREADS` is `NO` unless the variable `Halide_RUNTIME_NO_THREADS` is defined, in which case it provides the default value.

This is pretty similar to a prior implementation, with the major difference being the function argument qualification so that invocations of `add_halide_library` that the user _knows_ will not use threads can have this property specified locally/declaratively in the call.

steven-johnson:
> only add `Threads::Threads` if any schedule in the library uses `.parallel()`.

(1) .... or `async()`
(2) Doing so would also require upgrading LLVM_Runtime_Linker; it currently will unconditionally add threading-runtime-support on all platforms that can possibly support it, regardless of whether the thing being linked actually uses any of those features.

alexreinking:
> Doing so would also require upgrading LLVM_Runtime_Linker; it currently will unconditionally add threading-runtime-support on all platforms that can possibly support it, regardless of whether the thing being linked actually uses any of those features.

Maybe this means that whether or not threads are required can be accurately computed from the target string?

stevesuzuki-arm:
> 1. Add an argument to `add_halide_library` that passes through to `add_halide_runtime` (both functions have the argument) that optionally disables threads. Say, `NO_THREADS`. When `NOT NO_THREADS` and `NOT TARGET Threads::Threads`, then we `find_package(Threads REQUIRED)` inside `add_halide_runtime`.
> 2. The default for `NO_THREADS` is `NO` unless the variable `Halide_RUNTIME_NO_THREADS` is defined, in which case it provides the default value.
> 

I have updated the patch accordingly. I appreciate your efforts in your busy time!

apps/HelloBaremetal/cmake/toolchain.noos-arm32-sample.cmake
```diff
@@ -22,11 +22,15 @@ set(CMAKE_CXX_FLAGS "${C_COMMON_FLAGS}")
...
+# To suppress linker warning "missing .note.GNU-stack section" by GCC 12
 set(CMAKE_EXE_LINKER_FLAGS "-z noexecstack")
 
 # Halide target for Halide Generator
 set(Halide_TARGET "arm-32-noos-semihosting")
 
+# To prevent Threads and DL libs from being linked to runtime, as this toolchain doesn't have them
+set(HALIDE_RUNTIME_NO_THREADS ON)
+set(HALIDE_RUNTIME_NO_DL_LIBS ON)
+
```
alexreinking, line 34:
```suggestion
# To prevent Threads and DL libs from being linked to runtime, as this toolchain doesn't have them
set(Halide_RUNTIME_NO_THREADS ON)
set(Halide_RUNTIME_NO_DL_LIBS ON)

```

Note the letter case in `Halide` vs `HALIDE`

README_cmake.md
```diff
@@ -750,6 +750,8 @@ Variables that control package loading:
 | Variable             | Description                                                                                                                                                                   |
 |----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `Halide_SHARED_LIBS` | override `BUILD_SHARED_LIBS` when loading the Halide package via `find_package`. Has no effect when using Halide via `add_subdirectory` as a Git or `FetchContent` submodule. |
+| `Halide_RUNTIME_NO_THREADS` | skip linking of Threads libraray to runtime. Should be set if your toolchain does not support it (e.g. baremetal). |
+| `Halide_RUNTIME_NO_DL_LIBS` | skip linking of DL libraray to runtime. Should be set if your toolchain does not support it (e.g. baremetal). |
```
alexreinking, line 754: Can you sketch an example where unconditionally linking to `${CMAKE_DL_LIBS}` is a problem? My understanding is that it is very often empty, and indeed should be empty when the target platform does not support it.

alexreinking (changes requested):

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -581,6 +590,15 @@ function(add_halide_runtime RT)
...
         endif()
     endif()
+
+    # The default of NO_THREADS/NO_DL_LIBS is OFF unless HALIDE_RUNTIME_NO_THREADS/NO_DL_LIBS is defined globally
+    if (NOT DEFINED ARG_NO_THREADS)
+        set(ARG_NO_THREADS ${HALIDE_RUNTIME_NO_THREADS})
+    endif ()
+    if (NOT DEFINED ARG_NO_DL_LIBS)
+        set(ARG_NO_DL_LIBS ${HALIDE_RUNTIME_NO_DL_LIBS})
+    endif ()
```
alexreinking, line 600:
```suggestion
    if (NOT DEFINED ARG_NO_THREADS)
        set(ARG_NO_THREADS ${Halide_RUNTIME_NO_THREADS})
    endif ()
    if (NOT DEFINED ARG_NO_DL_LIBS)
        set(ARG_NO_DL_LIBS ${Halide_RUNTIME_NO_DL_LIBS})
    endif ()
```

stevesuzuki-arm: Fixed.

apps/HelloBaremetal/cmake/toolchain.noos-arm32-sample.cmake
stevesuzuki-arm: Fixed.

README_cmake.md
stevesuzuki-arm: In `HelloBaremetal` sample app, `dl` is set to `CMAKE_DL_LIBS`. This doesn't change even if `set(CMAKE_DL_LIBS "")` in toolchain file. It seems to be overwritten in [CMakeGenericSystem.cmake](https://github.com/Kitware/CMake/blob/c5a945dce8e1a4fb1b1db91c2dae03795c74074b/Modules/CMakeGenericSystem.cmake#L22)

alexreinking: @stevesuzuki-arm -- can you try one thing for me?

Try setting `CMAKE_SYSTEM_NAME` in the HelloBaremetal app to `Generic-ELF` instead of `none`. If that works, let's remove the `Halide_RUNTIME_NO_DL_LIBS` control

alexreinking: Okay, I've spoken with the CMake devs. Ben Boeckel thinks CMake is doing the wrong thing here, but fixing it would require a policy. So this is fine for now.

alexreinking (approved):

README_cmake.md
stevesuzuki-arm: That's good to know. JFYI, I tried `Generic` and `Generic-ELF` but neither of them solved DL issue.

alexreinking: Thanks... sorry about the comment churn.

steven-johnson:
So is this ready to land?

alexreinking:
> So is this ready to land?

Yep, merged.

PR #7283, opened 2023-01-17
Fix error of add_halide_generator in cross-compilation

stevesuzuki-arm:
In case the project name is CamelCase, add_halide_generator() was not able to find the generator package, because CMake searches <name>Config.cmake or <lower-case-name>-config.cmake


alexreinking (approved):

alexreinking:
Landing this... thanks for the patch! 🙂

PR #7284, opened 2023-01-17
Fix issue in add_halide_runtime in cross-compilation

stevesuzuki-arm:
add_halide_runtime() tries to build generator executable, but it fails if we are working with cross-compiler toolchain. By using existing generator set as "FROM", we can work around this.


alexreinking (approved):
This seems good for now. The medium-term vision is to ship a `gengen` utility binary in its own package that can produce runtimes.

PR #7285, opened 2023-01-17
Make auto scheduler libs available in HalideHelpers package

stevesuzuki-arm:
find_package(HalideHelpers) allows us to use add_halide_library(). But auto scheduler libs are not available unless they are in Halide-Interfaces.cmake.
Note:
Those libraries are not actually linked to the target application, but need to be available for add_custom_command call.

### DIFF
--- a/packaging/CMakeLists.txt
+++ b/packaging/CMakeLists.txt
@@ -33,7 +33,7 @@ install(TARGETS Halide Halide_Generator Halide_LanguageOptions
 
 if (TARGET Halide_Adams2019)
     install(TARGETS Halide_Adams2019 Halide_Li2018 Halide_Mullapudi2016
-            EXPORT Halide_Targets
+            EXPORT Halide_Interfaces
             LIBRARY DESTINATION ${Halide_INSTALL_PLUGINDIR} COMPONENT Halide_Runtime
             NAMELINK_COMPONENT Halide_Development)
 endif ()


packaging/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ install(TARGETS Halide Halide_Generator Halide_LanguageOptions
 
 if (TARGET Halide_Adams2019)
     install(TARGETS Halide_Adams2019 Halide_Li2018 Halide_Mullapudi2016
```
alexreinking, line 35: This doesn't seem quite right, but might be okay to get people unstuck... it makes me nervous, though.

The package decomposition I want in the end is as follows:

1. `HalideHelpers` (interfaces) includes all the platform-independent CMake helpers, header-only/source-injection libraries, etc.
2. `HalideBuildTools` includes all the libraries and executables intended to run on the host system. This is where the (to-be-created) `gengen` tool and the autoschedulers should go in the end.
3. `Halide` includes the libraries meant to be linked into the target applications.

alexreinking (changes requested):
Not rejecting necessarily, just want to talk about this so we're all on the same page. I've had plans for adding XC support for a while, but they were delayed by my PhD graduation.

packaging/CMakeLists.txt
stevesuzuki-arm: It's good to know that there is a plan and design which seems to make more sense. I admit this patch is a kind of hack that may not follow the existing design intension. My short term goal is to make it possible to compile the application like #7286 by CMake. If there is any other approach that could be less unacceptable, I'm happy to try and test.

steven-johnson:
Where does this PR stand?

steven-johnson:
What's the status on this?

stevesuzuki-arm:
> What's the status on this?

Options:
1. Suspend until proper (and more invasive) rework of package decomposition which @alexreinking is considering is introduced
1. Merge as it is as a temporary workaround to get people unstack

alexreinking:
Hey @stevesuzuki-arm 

Really sorry this is so delayed. I'm at the end of my job search/negotiations and it's just taken up way more time than I've budgeted.

I'm okay with merging this as a stop-gap, even if it isn't ultimately included in a numbered release (15 will branch before this. 16 will hopefully branch after my more invasive fixes).

steven-johnson:
OK, I vote we open a tracking Issue for the proper packaging rework (if one doesn't exist already) and land this.

alexreinking (approved):

stevesuzuki-arm:
> OK, I vote we open a tracking Issue for the proper packaging rework (if one doesn't exist already) and land this.

#7344 is filed for further rework.

Issue #7287, opened 2023-01-17
Halide should drop MIPS support


steven-johnson:
The primary reason for adding this was for Android, but Android hasn't supported MIPS since NDK r17 (~2018), and even MIPS-the-company no longer supports this architecture (they have pivoted to RISC-V). We should prune it to simplify things.

abadams:
Note also that our MIPS support is currently untested.

steven-johnson:
It was always untested in the past, but it's currently untested, too

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** MIPS support was removed; the architecture no longer exists in the `Target::Arch` enum and there is no MIPS codegen file.

**Evidence:** Commit e8e148163 "Drop support for MIPS (#7287) (#7289)"; src/Target.h:39-47 lists only `X86, ARM, Hexagon, POWERPC, WebAssembly, RISCV`, and `src/CodeGen_MIPS.cpp` no longer exists.

Done in e8e148163 — `Target::Arch` (src/Target.h:39-47) no longer has a MIPS entry and `CodeGen_MIPS.cpp` is gone. Closing.

PR #7289, opened 2023-01-17
Drop support for MIPS (#7287)

steven-johnson:
Please see https://github.com/halide/Halide/issues/7287 for details.


zvookin:
Why?

PR #7294, opened 2023-01-19
[Xtensa] 8-bit arithmetic improvements + some other smaller changes

Aelphy:
Changes from Cadence


src/XtensaOptimize.cpp
```diff
@@ -937,8 +938,11 @@ class MatchXtensaPatterns : public IRGraphMutator {
             {"halide_xtensa_narrow_i48_with_rounding_shift_i16", i16(rounding_shift_right(i32(wild_i48x), wild_i32))},
             {"halide_xtensa_narrow_i48_with_rounding_shift_u16", u16(rounding_shift_right(u32(wild_i48x), wild_u32))},
 
+#if 0  // Somehow this pattern is preventing a i16(i32(i48x) >> wild_i32) to be recognized as a PACKVRNRNX48
```
steven-johnson, line 941: Do we need to open an Issue for this?

vksnk: I think I also ran into this issue recently and have a fix.

steven-johnson:
Ready to land or is the `if 0` going to be dealt with first?

src/CodeGen_Xtensa.cpp
```diff
@@ -4057,6 +4105,10 @@ void CodeGen_Xtensa::visit(const Allocate *op) {
             }
         }
 
+        if (op->type.bits() == 24) {
```
vksnk, line 4108: This doesn't look right, looks like workaround for something and should be done in schedule instead.

src/XtensaOptimize.cpp
vksnk: This needs to be resolved before submission, I'll double check if it's the same issue I had.

steven-johnson:
Where does this PR stand?

steven-johnson:
Monday Morning Review Ping

src/XtensaOptimize.cpp
vksnk: Resolved now, but you'll need to sync.

Aelphy:
Seems ready from my side

vksnk (approved):

PR #7300, opened 2023-01-24
Restructure CMake dependencies

alexreinking:
This (draft) PR makes several significant changes / improvements to the way we handle dependencies.

### General

The `dependencies` folder is no more; custom `FindXYZ.cmake` modules inside `cmake/` have been written where upstream modules aren't suitable.

Project options have been collected at the top level for easy enumeration.

### LLVM

The LLVM-backend-related options (e.g. `TARGET_RISCV`)  are now correctly modeled as ternary options with the following alternatives:

1. `REQUIRED` or a CMake truthy value (like `ON`) -- the backend must be present or the build will fail
2. `OPTIONAL` -- if the backend is present, include it, otherwise do nothing
3. CMake falsey value (like `OFF`) -- do not include the backend

### SPIR-V

The SPIRV headers now use the upstream CMake package rather than vendoring it ourselves.

### WABT

We now use the upstream CMake package (that I worked on while at Google) instead of using FetchContent

### V8

Instructions for building V8 from source have been added and a find module written that is compatible with both those instructions and the Ubuntu package.


alexreinking:
Look away until I mark "ready for review". I did this work a while ago and some of the changes are currently regressions / went stale during my hiatus.

abadams:
I know you said to look away, but I just wanted to pipe up in favor of us keeping the spir-v header in tree instead of treating it as an externally-fetched thing. It's a spec so it should be stable, and the fewer external deps the better.

alexreinking:
This is the problem I'm trying to solve: the way we handle external dependencies is inconsistent. Here's the status quo:

| Dependency | Consumption method |
|--|--|
| `pybind11` | User-selected  between`find_package` or `FetchContent` (requires internet connection) |
| `wabt` | `FetchContent` only (requires internet connection) |
| `SPIRV-Headers` | vendored header in `dependencies/spirv` sub-tree |
| `V8` | set ad-hoc CMake variables (as an alternative to using `wabt`) |
| `LLVM` | upstream `find_package` support (with lots of hacks around that broken mess) |

Among these methods, most have serious drawbacks:

1. Ad-hoc CMake variables aren't very discoverable or user-friendly.
2. Vendoring code is an obstacle for new package maintainers and a headache for existing package maintainers.
3. We've already had users complain about requiring an internet connection to build Halide (apparently these users don't care about wasm or use only V8). So `FetchContent` isn't a great default solution.

That leaves us with `find_package`, which is IME the best option, anyway. All but V8 provide first-party CMake packages that are intended to be used by downstreams. I wrote custom `find_package` support for V8 in this PR.

> the fewer external deps the better

This PR doesn't change the number of external deps, only the way they interact with the build process. I appreciate the desire for simple UX and I think there are a few good options for handling dependencies in today's C++ ecosystem.

1. All of our dependencies (including llvm and the spirv headers) are available on vcpkg; we could provide a top-level `vcpkg.json` file so that one command (`vcpkg install` iirc?) installs the right stuff locally for an all-features build.
2. This PR includes a "dependency provider" compatible with CMake 3.24+ in `cmake/dependencies.cmake` that redirects our `find_package` calls to `FetchContent`. Maybe including that in a preset is enough.
3. We could look into Conan 2.0, which provides binary packages for many platforms and appears to be a big improvement over Conan 1.x (which really didn't grok CMake at all). This could either be an alternative or supplement to vcpkg.
4. We could use git submodules and prefer them when they're initialized (so that source tarballs and standard clones don't vendor anything by default).

----

Anyway, maybe we can justify carving out an exception for vendoring the single SPIR-V header on the basis that it's a spec. But hopefully this communicates my thinking a bit more clearly.

Issue #7306, opened 2023-01-27
Float16 emulations is easily missed


stevesuzuki-arm:
In Func definition, we can use `float16_t` or `Float(16)` type for 16 bit float data and arithmetic.  Some of the target devices support native float 16 bit operations which are enabled by adding the corresponding feature explicitly (e.g. `arm_fp16`, `f16c`).
A common mistake of user is to forget adding that feature. In that case, Halide code-gen kindly but silently replaces float16 operations with the sequence of 32 bit operations to emulate the equivalent. As you might imagine, the performance is order of magnitude slow, but it is prone to be missed or not easy for user to figure out the reason instantly.

Related issue : #7101

PR #7307, opened 2023-01-27
Warn emulated float16 equivalent is generated

stevesuzuki-arm:
Improvement for the issue #7306

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -1448,7 +1448,7 @@ void CodeGen_LLVM::visit(const Cast *op) {
                 "(b)float16 type operation is emulated, which is likely to slow down the performance. "
                 "If your target supports native (b)float16 operations, "
                 "it could be improved by adding Target feature to enable it.\n";
-            onetime_warnings.try_emplace(__LINE__, warn_msg);
+            onetime_warnings.try_emplace(WarningKind::EmulatedFloat16, warn_msg);
             Expr equiv = lower_float16_cast(op);
             internal_assert(equiv.type() == op->type);
             codegen(equiv);
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -710,6 +710,11 @@ void CodeGen_LLVM::compile_func(const LoweredFunc &f, const std::string &simple_
     debug(1) << "Generating llvm bitcode for function " << f.name << "...\n";
     f.body.accept(this);
 
+    // Show one time warning and clear it.
+    for (auto it = onetime_warnings.begin(); it != onetime_warnings.end(); it = onetime_warnings.erase(it)) {
+        user_warning << "In function " << f.name << ", " << it->second;
+    }
+
     // Clean up and return.
     end_func(f.args);
 }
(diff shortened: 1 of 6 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -1439,6 +1444,11 @@ void CodeGen_LLVM::visit(const Cast *op) {
         debug(4) << "Emulating cast from " << src << " to " << dst << "\n";
         if ((src.is_float() && src.bits() < 32) ||
             (dst.is_float() && dst.bits() < 32)) {
+            string warn_msg =
+                "(b)float16 type operation is emulated, which is likely to slow down the performance. "
+                "If your target supports native (b)float16 operations, "
+                "it could be improved by adding Target feature to enable it.\n";
+            onetime_warnings.try_emplace(__LINE__, warn_msg);
```
steven-johnson, line 1451: Why use __LINE__ here? Presumably it will always be the same value (i.e., 1451 at this point in code history)

steven-johnson (commented):
LGTM but the mechanism for making it "one-time" (via __LINE__) seems kinda wonky.

src/CodeGen_LLVM.cpp
stevesuzuki-arm: It was the mechanism for making it one-time, as you think, but I found it would be a problem if the same line number in different files add different warnings. Now it is replaced with enum.

steven-johnson (approved):

steven-johnson:
(I think this is good to land but waiting on a hopefully-all-green result now that some LLVM fixes have landed from upstream)

steven-johnson:
(I merged https://github.com/halide/Halide/pull/7313 in here so that we could get a not-complete-broken Windows build :-/)

src/CodeGen_LLVM.cpp
```diff
@@ -1439,6 +1444,11 @@ void CodeGen_LLVM::visit(const Cast *op) {
         debug(4) << "Emulating cast from " << src << " to " << dst << "\n";
         if ((src.is_float() && src.bits() < 32) ||
             (dst.is_float() && dst.bits() < 32)) {
+            string warn_msg =
+                "(b)float16 type operation is emulated, which is likely to slow down the performance. "
+                "If your target supports native (b)float16 operations, "
+                "it could be improved by adding Target feature to enable it.\n";
+            onetime_warnings.try_emplace(WarningKind::EmulatedFloat16, warn_msg);
```
abadams, line 1451: Is the error message itself not a valid key? I.e. could it be a std::set instead of a std::map?

steven-johnson:
This is weird: the new 'warning' is being emitted and yet the test is failing? https://buildbot.halide-lang.org/master/#/builders/11/builds/7/steps/13/logs/warning_emulated_float16

steven-johnson:
> This is weird: the new 'warning' is being emitted and yet the test is failing? https://buildbot.halide-lang.org/master/#/builders/11/builds/7/steps/13/logs/warning_emulated_float16

Ah, the failure is because we are writing to /dev/null... which is Not A Thing on Windows. I'll push a fix.

src/CodeGen_LLVM.cpp
stevesuzuki-arm: It could. I think the downside is the cost of computing key equivalence proportional to the length of string, while we can remove enum WarningKind from the code instead. Do you have any preference?

steven-johnson: I think this is fine as-is, let's merge it

PR #7309, opened 2023-01-30
Minimal updates to allow Halide building with LLVM17

steven-johnson:
(Opening as draft initially until Buildbots build the new LLVM versions)


alexreinking (approved):
LGTM when ready

Issue #7314, opened 2023-01-31
Expose call_argv_fast from Callable


twesterhout:
Dear Halide team,

Would it be possible to make [`Halide::Callable::call_argv_fast`](https://github.com/halide/Halide/blob/main/src/Callable.h#L284) public? Since `Callable` is currently not used in the tutorials and is not displayed by Doxygen, this change shouldn't affect too many users. And, of course, a big warning could be added to inform users about possible segfaults if `call_argv_fast` is invoked with wrong argument types.

### Motivation

I'm working on Haskell wrappers for Halide as part of Tweag Open Source Fellowship. Haskell has static typing and very good support for polymorphism, and that means that although I know the types of arguments when invoking the Callable, I can't pass them to C++ as template arguments. So what I'd like to implement is similar to what the Python bindings do in [`PyCallable`](https://github.com/halide/Halide/blob/main/python_bindings/src/halide/halide_/PyCallable.cpp#L54).
If there are other preferred methods of calling Halide pipelines with low overhead when argument types are unknown until the runtime of the program, any pointers will be highly appreciated.

Thanks,
Tom

steven-johnson:
This sounds pretty awesome!

I'd definitely be willing to make it public, we'd just have to add the right amount of caveats in the documentation. (Of course, it's not really any more unsafe than the existing _argv calls provided by AOT codegen, so it's really a question of ensuring people realize it's public only for code of the sort you are talking about).

Please feel free to open a PR, and please keep us up to date on this work to the extent you can :-)

PR #7315, opened 2023-01-31
Make Callable::call_argv_fast public

twesterhout:
Addresses #7314.

I'm wondering, is the argv-based calling convention is documented somewhere? If so, I can add a reference to the doc string.

### DIFF
--- a/src/Callable.h
+++ b/src/Callable.h
@@ -385,7 +385,7 @@ public:
      * This function relies on the same calling convention as the argv-based
      * functions generated for ahead-of-time compiled Halide pilelines.
      *
-     * Very rough specifications ofthe calling convention (but check the source
+     * Very rough specifications of the calling convention (but check the source
      * code to be sure):
      *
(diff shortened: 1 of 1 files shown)


steven-johnson:
> is the argv-based calling convention is documented somewhere

Probably not, at least not as well as it should be. Basically:
- All scalar arguments are passed by *pointer*, not by value, regardless of size
- All buffer arguments (input or output) are passed as `halide_buffer_t*`
- All arguments are passed in the same order as they appear in the C function argument list

steven-johnson (approved):
LGTM pending green buildbots.

twesterhout:
One of the Windows builds is failing. It can't have anything to do with the latest commits, can it? I've only changed the wording in the comments...

steven-johnson:
> One of the Windows builds is failing. It can't have anything to do with the latest commits, can it? I've only changed the wording in the comments...

Yes, unfortunately our Windows buildbots are in need of some love and that's one of the flakes. This is fine to land.

PR #7317, opened 2023-02-02
Replace some push_backs with emplace_back

abadams:


steven-johnson (approved):
Did you find these by inspection or via tooling?

abadams:
I tried out the latest clang-tidy and it found these using modernize-use-emplace

PR #7321, opened 2023-02-06
Fix segfault by nonconstant bound in Adams2019

stevesuzuki-arm:
Fix segmentation fault in Adams2019 in case the estimate or bound of Func is set to nonconstant Expr.


abadams (approved):

steven-johnson:
32-bit breakages are irrelevant

PR #7322, opened 2023-02-06
Fix LLVM 17+ build integration on 32-bit systems

### LINKED ISSUE #7319: dependencies/llvm/CMakeLists.txt incompatible with latest llvm-project
According to https://reviews.llvm.org/D139752#4006602 , latest llvm-project would cause `<command-line>:0:70: error: token "=" is not valid in preprocessor expressions`. According to https://reviews.llvm.org/D139752#4104371 , it's likely a fault in Halide `dependencies/llvm/CMakeLists.txt`.

alexreinking:
Fixes #7319

### DIFF
--- a/dependencies/llvm/CMakeLists.txt
+++ b/dependencies/llvm/CMakeLists.txt
@@ -28,6 +28,12 @@ if (LLVM_PACKAGE_VERSION VERSION_GREATER 17.0)
     message(WARNING "Halide is not tested on LLVM versions beyond 17.0")
 endif ()
 
+# LLVM_DEFINITIONS is a space-separated list instead of a more typical
+# CMake semicolon-separated list. For a long time, CMake could handle
+# this transparently but, since LLVM 17, the flag -D_FILE_OFFSET_BITS=64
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):
LGTM, nice find. We should probably circle back to the LLVM patch review to tell them the resolution of this. (Also: is there any point to reporting this to the CMake maintainers as a bug?)

alexreinking:
> We should probably circle back to the LLVM patch review to tell them the resolution of this.

Sure -- I just need to reset my LLVM reviews password...

> Also: is there any point to reporting this to the CMake maintainers as a bug?

If we can minimize the test case to something convincing, then sure.

alexreinking:
I left a comment on the related LLVM issue: https://reviews.llvm.org/D139752#4107678

alexreinking:
Actually, upon closer inspection, it has to do with the presence of an `=` in the list. When there's an `=`, it seems that CMake does not split on spaces automatically anymore. We've never built Halide on AIX or SunOS, which are the only other build scenarios where an = can appear in `LLVM_DEFINITIONS`.

Issue #7325, opened 2023-02-08
Allow Buffer<_Float16> to mean Buffer<float_16_t> on compilers that support it


steven-johnson:
This would make runtime specification of float16 buffers at runtime a lot simpler for most users. Is there any reason we shouldn't do this? (I think it would just take another overload of `halide_type_of<>()` in HalideRuntime.h)

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** The request was a `halide_type_of<>()` overload for `_Float16` on compilers that support the type. That overload exists, guarded by a `HALIDE_CPP_COMPILER_HAS_FLOAT16` feature-detection block.

**Evidence:** Commit 4156c5ab3 ("Allow _Float16 as alias for float16_t in halide_type_of\<>() (#7325)", PR #7326); `src/runtime/HalideRuntime.h:2327-2332` (`template<> halide_type_of<_Float16>()` returning `halide_type_t(halide_type_float, 16)`), with detection at `src/runtime/HalideRuntime.h:102-134`.

Fixed by #7326 (commit `4156c5ab3`): `src/runtime/HalideRuntime.h:2327` defines `halide_type_of<_Float16>()` whenever `HALIDE_CPP_COMPILER_HAS_FLOAT16` is set, so `Buffer<_Float16>` works on compilers with the type. Closing.

PR #7326, opened 2023-02-08
Allow _Float16 as alias for float16_t in halide_type_of<>() (#7325)

steven-johnson:


test/correctness/float16_t.cpp
```diff
@@ -290,43 +307,71 @@ int main(int argc, char **argv) {
...
         float16_t f(1.0f16);
         _Float16 f2 = (_Float16)f;
         if (f2 != 1.0f16) {
-            printf("Roundtrip of 16-bit float via _Float16 failed.\n");
+            fprintf(stderr, "Roundtrip of 16-bit float via _Float16 failed.\n");
             return -1;
         }
     }
+#else
+    printf("Only clang supports _Float16 literals, skipping roundtrip test\n");
```
terryheo, line 370: I can't get this. In GCC, we can't use `_Float16` even if `HALIDE_CPP_COMPILER_HAS_FLOAT16` is set?

steven-johnson: As far as I can tell, even the versions of gcc that support `_Float16` don't support 16-bit-floating-point-constants via the `f16` suffix. (Maybe I don't have it configured correctly, but I also can't find any gcc docs that suggest that it does.)

terryheo: Got it. Thanks for the explanation.

terryheo, line 370: It wasn't sufficient to me. Maybe
`Only clang supports _Float16 constant literal 'f16', , skipping roundtrip test` ?

src/Float16.h
```diff
@@ -71,7 +58,9 @@ struct float16_t {
 #ifdef HALIDE_CPP_COMPILER_HAS_FLOAT16
     /** Cast to compiler's built-in _Float16 type. */
     explicit operator _Float16() const {
-        return *(const _Float16 *)&data;
+        _Float16 result;
```
terryheo, line 61: Just wonder why this change is needed. The old logic doesn't work?

terryheo (approved):
LGTM

src/Float16.h
steven-johnson: The old logic was technically a violation of strict aliasing rules; in practice, ~all C++ compilers would work fine with it, but one of ours complained, and using memcpy() is definitely safe (and likely to be optimized into the same code).

test/correctness/float16_t.cpp
steven-johnson: Done

steven-johnson:
Ping to @abadams and/or @zvookin

abadams (approved):

steven-johnson:
I'm going to go ahead and land this to unblock some work; hopefully @zvookin can chime in when he has a chance, since he spent a lot of time on the original issue here.

PR #7328, opened 2023-02-08
[xtensa] Improvements to CodeGen_Xtensa

Aelphy:
Fixed common_int and common_uint for Q8. Added new types support for load_predicated, store_predicated, halide_xtensa_interleave (for u16 also added Q8 support for native_vector_u16_x3). Improved convert f32 to u32 vectors with proper intrinsic. Cleaned up.


src/CodeGen_Xtensa.cpp
```diff
@@ -858,6 +896,46 @@ HALIDE_ALWAYS_INLINE native_vector_i32_x2 load_predicated<native_vector_i32_x2,
...
+    native_vector_u16 vmask = IVP_MOVNX16T(native_vector_u16(1), native_vector_u16(0), predicate);
+    uint8_t __attribute__((aligned(XCHAL_VISION_SIMD8))) mask[2 * VECTOR_WIDTH_F32];
+    aligned_store<native_vector_u16, uint16_t, 2 * VECTOR_WIDTH_F32>(vmask, &mask[0], 0);
+
+    float __attribute__((aligned(XCHAL_VISION_SIMD8))) output[2 * VECTOR_WIDTH_F32];
+    for (int i = 0; i < 2 * VECTOR_WIDTH_F32; i++) {
+        if (mask[i] == 1) {
+            output[i] = ((const float*)base)[offsets[i]];
+        } else {
+            output[i] = 0;
```
steven-johnson, line 912: Does mask=0 for a predicated load mean "set this entry to zero in the output", or "leave this entry unchanged in the output"? In predicated_store, mask=0 definitely means "leave this entry unchanged".

src/CodeGen_Xtensa.cpp
```diff
@@ -3125,13 +3339,13 @@ void CodeGen_Xtensa::visit(const Div *op) {
         string sb = print_expr(op->b);
         // Just cast to clang vector types and use division defined on them.
         if (is_native_xtensa_vector<uint8_t>(op->type, target)) {
-            print_assignment(op->type, "(common_uint8x64_t)" + sa + " / (common_uint8x64_t)" + sb);
+            print_assignment(op->type, "(common_uint8x" + std::to_string(op->type.lanes()) + "_t)" + sa + " / (common_uint8x" + std::to_string(op->type.lanes()) + "_t)" + sb);
```
vksnk, line 3342: I think ``` "common_" + print_type(op->type())``` will be better.

src/CodeGen_Xtensa.cpp
```diff
@@ -1424,6 +1571,28 @@ HALIDE_ALWAYS_INLINE native_vector_u16_x3 halide_xtensa_interleave_u16(const nat
       58, 0,  22, 1,  23, 48, 59, 2,  24, 3,  25, 49, 60, 4,  26, 5,
       27, 50, 61, 6,  28, 7,  29, 51, 62, 8,  30, 9,  31, 52, 63, 10};
   unsigned long long int_16B_c3_step_1_msk = 0xffffffff55555555ULL;
+  #elif XCHAL_VISION_TYPE == 8
```
vksnk, line 1574: Is this also from programming guide?

src/CodeGen_Xtensa.cpp
```diff
@@ -900,6 +978,29 @@ HALIDE_ALWAYS_INLINE void store_predicated<native_vector_u8, native_vector_i32_x
     }
 }
 
+template <>
```
vksnk, line 981: Not for this PR, but we should try to come with a way to generate these (more) automatically, because there is a lot of repetition.

src/CodeGen_Xtensa.cpp
Aelphy: Found the same behaviour in other load_predicated variants in the same file

src/CodeGen_Xtensa.cpp
Aelphy: Yes

src/CodeGen_Xtensa.cpp
```diff
@@ -328,13 +328,23 @@ inline int GetCycleCount() {
...
 typedef int32_t common_int32x16_t __attribute__((ext_vector_type(16)));
 typedef uint32_t common_uint32x16_t __attribute__((ext_vector_type(16)));
+#elif XCHAL_VISION_TYPE == 8
+typedef int8_t common_int8x128_t __attribute__((ext_vector_type(128)));
+typedef uint8_t common_uint8x128_t __attribute__((ext_vector_type(128)));
+typedef int16_t common_int16x64_t __attribute__((ext_vector_type(64)));
+typedef uint16_t common_uint16x64_t __attribute__((ext_vector_type(64)));
+typedef int32_t common_int32x32_t __attribute__((ext_vector_type(32)));
+typedef uint32_t common_uint32x32_t __attribute__((ext_vector_type(32)));
+#endif
```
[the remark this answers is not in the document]
Aelphy: done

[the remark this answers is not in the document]
Aelphy: done

src/CodeGen_Xtensa.cpp
```diff
@@ -818,6 +834,28 @@ HALIDE_ALWAYS_INLINE native_vector_i16 load_predicated<native_vector_i16, native
     return *((native_vector_i16 *)output);
 }
 
+template<>
+HALIDE_ALWAYS_INLINE native_mask_i16_x2 convert<native_mask_i16_x2, native_mask_i8>(const native_mask_i8& src);
+
+template <>
+HALIDE_ALWAYS_INLINE native_vector_i16_x2 load_predicated<native_vector_i16_x2, native_vector_i32_x4, native_mask_i8 , int16_t, 2 * VECTOR_WIDTH_I16>(const void *base, const native_vector_i32_x4& offset, const native_mask_i8& predicate) {
```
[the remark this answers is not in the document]
Aelphy: done

src/CodeGen_Xtensa.cpp
steven-johnson: I hear you, but I'm not sure if those are right, either; these are meant to implement the `Load` IR node in Halide, and the implementation of these in Codegen_C leaves the entry unchanged for mask=0. I'm not sure if that's what the LLVM implementation does as well, but I rather strongly suspect so. Adding @abadams to comment.

abadams: They should be set to an undefined value, so zero is safe. That's what the llvm backend does.

steven-johnson: OK, in that case I'd argue that

`output[i] = mask[i] ? ((const float*)base)[offsets[i]] : 0;` is slightly better style, but probably results in identical codegen

src/CodeGen_Xtensa.cpp
steven-johnson: These aren't declared as 'static', so we are creating a new instance every time we call this function -- is the overhead significant?

PR #2060, opened 2017-05-18
Don't reuse llvm_module for compiling to assembly, bitcode

dpalermo:
Reusing the llvm_module is effectively optimizing the generated code
a second (or third) time producing assembly & bitcode that doesn't
match the object code.  By starting with a fresh copy of the llvm
module for each compile to step, the generated code matches between
all targets.

PR #7329, opened 2023-02-08
Fixes for top-of-tree LLVM

steven-johnson:
Note that there will likely be subsequent changes/fixes to LLVM_Output.cpp; this is a quick fix to get buildbots unbroken.

### DIFF
--- a/src/LLVM_Headers.h
+++ b/src/LLVM_Headers.h
@@ -28,7 +28,7 @@
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/ADT/StringMap.h>
 #include <llvm/ADT/StringRef.h>
-#if LLVM_VERSION < 17
+#if LLVM_VERSION < 170
 #include <llvm/ADT/Triple.h>
 #endif
 #include <llvm/ADT/Twine.h>
@@ -83,7 +83,7 @@
 #include <llvm/Support/raw_ostream.h>
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/Target/TargetOptions.h>
-#if LLVM_VERSION >= 17
+#if LLVM_VERSION >= 170
 #include <llvm/TargetParser/Triple.h>
 #endif
 #include <llvm/Transforms/IPO.h>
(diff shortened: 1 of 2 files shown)


abadams (approved):

steven-johnson:
Some comments from LLVM folks that I don't know how to answer, but maybe @abadams does:

- `I'm wondering why [createAlwaysInlinerLegacyPass()] needs to be added -- I'd assume that you'd already have optimized the IR by then -- and alwaysinline calls are already inlined`... this code predates my joining Halide, I think, so I dunno. @abadams, any idea?

- `[createStripDeadDebugInfoPass()] probably should be part of the optimization pipeline instead of the codegen pipeline` -- that specific call doesn't exist in top of tree anymore but something similar does, I can try adding it elsewhere

abadams:
> Some comments from LLVM folks that I don't know how to answer, but maybe @abadams does:
> 
> * `I'm wondering why [createAlwaysInlinerLegacyPass()] needs to be added -- I'd assume that you'd already have optimized the IR by then -- and alwaysinline calls are already inlined`... this code predates my joining Halide, I think, so I dunno. @abadams, any idea?

My memory is hazy, but I believe at some point it was necessary to get the runtime alwaysinline functions inlined early enough for other things to work. No idea if that's still the case a decade later.

abadams:
and I have no idea what dead stripping debug info is all about

alexreinking (approved):

steven-johnson:
> My memory is hazy, but I believe at some point it was necessary to get the runtime alwaysinline functions inlined early enough for other things to work. No idea if that's still the case a decade later.

That was present in the first instance of this file in 2014 -- I can't figure out if that was branched/copied from another file prior to that (maybe lost in the History Rewrite Rebase years ago?) -- but it likely doesn't hurt anything, so let's leave it.

> and I have no idea what dead stripping debug info is all about

That was added in https://github.com/halide/Halide/pull/2060 -- based on the comments, it looks like the strip pass was added in an attempt to fix an error, but didn't actually fix it, and (apparently) just got left in? I vote we don't bother trying to re-add it for top-of-tree LLVM and see if it matters, since it looks like it was moot anyway. I'll add a comment to that effect.

PR #7330, opened 2023-02-08
Fix a subtle uninitialized-memory-read in Buffer::for_each_value()

steven-johnson:
When we flattened dimensions in for_each_value_prep(), we would copy from one past the end, meaning the last element contained uninitialized garbage. (This wasn't noticed as an out-of-bounds read because we overallocated in structure in for_each_value_impl()). This garbage stride was later used to advance ptrs in for_each_value_helper()... but only on the final iteration, so even if the ptr was wrong, it didn't matter, as the ptr was never used again. Under certain MSAN configurations, though, the read would be (correctly) flagged as uninitialized.

This fixes the MSAN bug, and also (slightly) improves the efficiency by returning the post-flattened number of dimensions, potentially reducing the number of iterations f for_each_value_helper() needed.

### DIFF
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -2228,7 +2228,6 @@ private:
                 }
                 i--;
                 d--;
-                t[d].extent = 1;
             }
         }
 
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -2176,6 +2176,9 @@ private:
     template<int N>
     HALIDE_NEVER_INLINE static std::pair<int, bool> for_each_value_prep(for_each_value_task_dim<N> *t,
                                                                         const halide_buffer_t **buffers) {
+        const int dimensions = buffers[0]->dimensions;
+        assert(dimensions > 0);
+
         // Check the buffers all have clean host allocations
         for (int i = 0; i < N; i++) {
             if (buffers[i]->device) {
@@ -2188,11 +2191,6 @@ private:
                        "Buffer passed to for_each_value has no host or device allocation");
             }
         }
-
-        const int dimensions = buffers[0]->dimensions;
-
-        // Caller currently enforces this
-        assert(dimensions > 0);
 
         // Extract the strides in all the dimensions
         for (int i = 0; i < dimensions; i++) {
@@ -2231,11 +2229,12 @@ private:
             }
         }
 
+        // Note that we assert() that dimensions > 0 above
+        // (our one-and-only caller will only call us that way)
+        // so the unchecked access to t[0] should be safe.
         bool innermost_strides_are_one = true;
-        if (dimensions > 0) {
-            for (int i = 0; i < N; i++) {
-                innermost_strides_are_one &= (t[0].stride[i] == 1);
-            }
+        for (int i = 0; i < N; i++) {
+            innermost_strides_are_one &= (t[0].stride[i] == 1);
         }
 
         return {d, innermost_strides_are_one};
--- a/src/runtime/HalideBuffer.h
+++ b/src/runtime/HalideBuffer.h
@@ -2172,9 +2172,10 @@ private:
         }
     }
 
+    // Return pair is <new_dimensions, innermost_strides_are_one>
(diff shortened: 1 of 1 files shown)


src/runtime/HalideBuffer.h
```diff
@@ -2235,26 +2239,30 @@ class Buffer {
...
+                Buffer<>::for_each_value_helper(f, new_dims - 1,
+                                                innermost_strides_are_one,
+                                                t,
+                                                data(), (other_buffers.data())...);
+            }
+            // else fall thru
         }
+
+        // zero-dimensional case
+        f(*data(), (*other_buffers.data())...);
```
abadams, line 2266: This line looks like it runs unconditionally, instead of just for dimensions() == 0

src/runtime/HalideBuffer.h
```diff
@@ -2235,26 +2239,30 @@ class Buffer {
...
-                                            data(), (other_buffers.data())...);
-        } else {
-            f(*data(), (*other_buffers.data())...);
+            auto [new_dims, innermost_strides_are_one] = Buffer<>::for_each_value_prep(t, buffers);
+            if (new_dims > 0) {
+                Buffer<>::for_each_value_helper(f, new_dims - 1,
+                                                innermost_strides_are_one,
+                                                t,
+                                                data(), (other_buffers.data())...);
+            }
```
abadams, line 2261: Should there be a return after the call to for_each_value_helper?

abadams (approved):

src/runtime/HalideBuffer.h
```diff
@@ -2235,26 +2239,30 @@ class Buffer {
             }
         }
 
-        return innermost_strides_are_one;
+        return {d, innermost_strides_are_one};
```
vitalybuka, line 2242: this d can be after d-- on 2226, is this expected?

src/runtime/HalideBuffer.h
```diff
@@ -2219,7 +2223,7 @@ class Buffer {
             }
             if (flat) {
                 t[i - 1].extent *= t[i].extent;
-                for (int j = i; j < d; j++) {
+                for (int j = i; j < d - 1; j++) {
```
vitalybuka, line 2226: t[d-1].extent will be set on line 2231, what is about .stride?

src/runtime/HalideBuffer.h
steven-johnson: Yeah, I forgot to hit 'save' before pushing, I presume you saw the subsequent fix

src/runtime/HalideBuffer.h
steven-johnson: See above

vitalybuka (approved):
Added few comments, but I know nothing about this code.

The patch fixes sanitizer issue we had.

src/runtime/HalideBuffer.h
steven-johnson: Yes, this is deliberate; if we 'flatten' multiple dimensions into one, we want to know the new, smaller number of dimensions.

src/runtime/HalideBuffer.h
steven-johnson: No, we want to leave that alone -- @abadams to confirm.

abadams: Do we still need to set the extent, given that we're not going to iterate over that dimension?

steven-johnson: Oooh, good point. We can probably elide that now. Let me do some hackery to verify.

steven-johnson:
I now have another thing I'm not sure is optimal; this section in for_each_value_prep():

```
        bool innermost_strides_are_one = true;
        if (dimensions > 0) {
            for (int i = 0; i < N; i++) {
                innermost_strides_are_one &= (t[0].stride[i] == 1);
            }
        }
```

I presume the `if (dimensions > 0)` check is just for paranoia? Currently this function should never be called when dimensions == 0 (there's only one caller). I added an `assert()` to indicate this is expected as an invariant.

PR #7334, opened 2023-02-09
Add __restrict to pointer var decls in C backend

steven-johnson:
(Backported from the Xtensa branch -- I suspect that all interesting compilers support __restrict properly at this point, but let's find out)


steven-johnson:
Closing this because the existing __restrict usage that is being backported is almost certainly incorrect (it's marking things as __restrict that should not be).

PR #7338, opened 2023-02-10
C backend: use static_cast<>/reinterpret_cast<> instead of C-style casts

steven-johnson:
Another stylistic change that shouldn't affect object code at all, but IMHO makes the C code a bit easier to read and reason about. (Disclaimer: not sure if this will affect the GPU, etc backends that share this code -- do Cuda et al support this syntax?)

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -143,14 +143,14 @@ inline bool is_finite_f32(float x) {return isfinite(x);}
 inline bool is_finite_f64(double x) {return isfinite(x);}
 
 template<typename A, typename B>
-inline A reinterpret(const B &b) {
+inline A halide_reinterpret_bits(const B &b) {
     static_assert(sizeof(A) == sizeof(B), "type size mismatch");
     A a;
     memcpy(&a, &b, sizeof(a));
     return a;
 }
 inline float float_from_bits(uint32_t bits) {
-    return reinterpret<float, uint32_t>(bits);
+    return halide_reinterpret_bits<float, uint32_t>(bits);
 }
 
 template<typename T>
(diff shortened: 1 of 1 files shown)


src/CodeGen_C.cpp
```diff
@@ -1375,14 +1375,14 @@ string CodeGen_C::print_type(Type type, AppendSpaceIfNeeded space_option) {
...
     if (type.is_handle() || e.type().is_handle()) {
-        // Use a c-style cast if either src or dest is a handle --
+        // Use a reinterpret_cast<> if either src or dest is a handle --
         // note that although Halide declares a "Handle" to always be 64 bits,
         // the source "handle" might actually be a 32-bit pointer (from
-        // a function parameter), so calling reinterpret<> (which just memcpy's)
+        // a function parameter), so calling halide_reinterpret_bits<> (which just memcpy's)
         // would be garbage-producing.
-        oss << "(" << print_type(type) << ")";
+        oss << "reinterpret_cast<" << print_type(type) << ">";
```
abadams, line 1383: I don't think a reinterpret cast is valid here, because the bit-width might change.

abadams:
Doesn't this just make the generated code more verbose? We don't tend to use static_cast inside libHalide so I'm not sure why we'd prefer to generate it.

steven-johnson:
> Doesn't this just make the generated code more verbose? We don't tend to use static_cast inside libHalide so I'm not sure why we'd prefer to generate it.

My theory here was that when generating code, we can select the right cast more appropriately. (Note that this was opened as a draft...)

steven-johnson:
This breaks various GPU backends, and is definitely not worth conditionalizing for that -- closing as a failed experiment.

PR #7340, opened 2023-02-10
Remove Xtensa::compile(LoweredFunc), add Xtensa::preprocess_function_body()

steven-johnson:
This removes a nice chunk of redundant code (and adds some corner cases that were missing from the Xtensa version).

### DIFF
--- a/src/CodeGen_Xtensa.cpp
+++ b/src/CodeGen_Xtensa.cpp
@@ -24,6 +24,8 @@ using std::ostringstream;
 using std::string;
 using std::vector;
 
+namespace {
+
 std::string intrinsic_suffix_for_type(Type t) {
     if (t.is_int() && (t.bits() == 8)) {
         return "2NX8";
@@ -64,6 +66,8 @@ public:
     bool uses_dma = false;
     int max_channel_no = 0;
 };
+
+}  // namespace
 
 void CodeGen_Xtensa::add_platform_prologue() {
     const char *headers = R"INLINE_CODE(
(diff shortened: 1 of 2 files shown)


src/CodeGen_Xtensa.cpp
```diff
@@ -123,186 +127,25 @@ void CodeGen_Xtensa::compile(const Module &module) {
...
-                if (arg.is_buffer()) {
-                    stream << "_buffer";
-                }
-                comma = ", ";
-            }
-            stream << ");\n";
-            close_scope("");
-        }
-    }
+void CodeGen_Xtensa::compile(const LoweredFunc &func, const MetadataNameMap &metadata_name_map) {
```
vksnk, line 131: Why is still needed?

steven-johnson: Without it, we get

```
/Users/srj/GitHub/Halide/src/CodeGen_Xtensa.h:21:10: warning: 'Halide::Internal::CodeGen_Xtensa::compile' hides overloaded virtual function [-Woverloaded-virtual]
    void compile(const Module &module);
         ^
/Users/srj/GitHub/Halide/src/CodeGen_C.h:69:18: note: hidden overloaded virtual function 'Halide::Internal::CodeGen_C::compile' declared here: different number of parameters (2 vs 1)
    virtual void compile(const LoweredFunc &func, const MetadataNameMap &metadata_name_map);
```

(arguably, we should rename these to avoid overloading, to avoid this nonsense -- I may propose that as a followup)

vksnk (approved):

PR #7345, opened 2023-02-13
Fix tuple output bounds checks

### LINKED ISSUE #7343: Apparent write beyond end of buffer in jit-compiled code
The following code writes beyond the end of Buffer r0 inside the JIT-compiled code. Given the schedule, I believe the minimum viable output size is 1024 + 16, starting at -1, not 1024 starting at zero, so it should be failing a bounds check, but it's not.

```
    Var x;
    const int size = 1024;
    Func h;
    h(x) = {0, 0};
    RDom r(0, size);
    h(r) = {h(r - 1)[0], 0};

    Var xo, xi;
    h.split(x, xo, xi, 16, TailStrategy::RoundUp);

    Buffer<int> r0(size);
    Buffer<int> r1(size);
    h.realize({r0, r1});
```

abadams:
Tuple outputs weren't getting appropriate bounds checks due to overzealous culling of uninteresting code in the add_image_checks pass.

Fixes #7343

### DIFF
--- a/src/AddImageChecks.cpp
+++ b/src/AddImageChecks.cpp
@@ -106,11 +106,21 @@ class TrimStmtToPartsThatAccessBuffers : public IRMutator {
     using IRMutator::visit;
 
     Expr visit(const Call *op) override {
-        touches_buffer |= (buffers.count(op->name) > 0);
+        touches_buffer |=
+            (buffers.count(op->name) > 0) ||
+            (buffers.count(op->name + "." + std::to_string(op->value_index)));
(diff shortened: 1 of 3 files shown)


steven-johnson:
https://buildbot.halide-lang.org/master/#/builders/101/builds/56 looks like a real failure

steven-johnson (approved):

abadams:
Note that this adds a missing bounds check, so it's possible that it'll break previously working code that either had non-faulting out-of-bounds accesses or overconservative deduced bounds.

PR #7347, opened 2023-02-14
Change early-bound default args in Python bindings to late-bound

steven-johnson:
In PyBind11, if you specify a default argument for a method, it is evaluated when the Python module is initialized, *not* when the method is called (as you might expect in C++). For defaults that are just constants/literals, this is no big deal, but when calling get_*_target_from_environment, this means it is called at module init time -- also normally not a big deal (since the values ~never change at runtime anyway), with one big exception (no pun intended): if the function throws an exception (e.g. via calling user_assert() or similar), that exception is thrown at Module-initialization time, which is a much more inscrutable crash, and one that is very hard to recover from.

This may seem unlikely, but can happen pretty easily if you set (say) HL_JIT_TARGET=host-cuda (or other gpu) and the given GPU runtime isn't present on the given system; the current behavior is basically "make if impossible for the libHalidePython bindings to run", whereas what we want is "runtime exception thrown when you call the method".

This changes the relevant methods to use `Target()` as the default value, and inside the method wrapper, if the value passed equals `Target()`, it replaces the value with the righ `get_*_target_from_environment()` call.

(This turned up while doing some testing of https://github.com/halide/Halide/pull/6924 on a system without Vulkan available)

This change should probably be backported to the 15.x branch as well.

### DIFF
--- a/python_bindings/src/halide/halide_/PyBuffer.cpp
+++ b/python_bindings/src/halide/halide_/PyBuffer.cpp
@@ -587,27 +587,27 @@ void define_buffer(py::module &m) {
             })
 
             .def(
-                "copy_to_device", [](Buffer<> &b, const Target &t) -> int {
-                    return b.copy_to_device(t);
+                "copy_to_device", [](Buffer<> &b, const Target &target) -> int {
+                    return b.copy_to_device(to_jit_target(target));
                 },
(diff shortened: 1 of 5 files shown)


steven-johnson:
All clean, PTAL

abadams (approved):

PR #7352, opened 2023-02-15
Fix Python error handling

steven-johnson:
Error handling in the Python bindings wasn't quite right for JIT:

We previously replaced halide_error() to throw a C++ exception. Sounds good, but unfortunately, doesn't work reliably: if called from jitted code (which doesn't know about C++ exceptions), the throw statement may be unable to find the enclosing try block (which is outside jitted code), meaning it will call std::terminate. Now, instead, we just leave the JIT error handler unset, and call with an explicit JITUserContext with a custom print handler; in theory, this meant that the code in JITFuncCallContext::finalize() would check for an error after the call into jitted code, and call `halide_runtime_error` if so (which would then trigger an all-in-C++-exception). Unfortunately...

(2) JITFuncCallContext is broken by design; it mutates the input JITUserContext, so that trying to use the same JITUserContext for two calls in a row leaves you with a JITUserContext with (at least) the error_handler set. Since at least one of the realize() calls does this twice (once for bounds query, once for execution), this means that an error in the second call would never be seen, since finalize() only reported errors if there wasn't a custom error handler on input. Per @abadams suggestion, we work around this by treating 'JITErrorBuffer::handler' as 'no custom error handler', which is mostly true. (But really, JITFuncCallContext and JITUserContext are a hard-to-reason-about mess and arguably need to rethought entirely.)

(3) Removed entirely-unnecessary overrides of runtime print and error handlers from PyStubImpl; despite the comments, this code is unnecessary.

(This should probably be backported to the 15.x branch)

### DIFF
--- a/python_bindings/src/halide/halide_/PyError.cpp
+++ b/python_bindings/src/halide/halide_/PyError.cpp
@@ -28,7 +28,8 @@ public:
 
 }  // namespace
 
-PyJITUserContext::PyJITUserContext() : JITUserContext() {
+PyJITUserContext::PyJITUserContext()
+    : JITUserContext() {
     handlers.custom_print = halide_python_print;
     // No: we don't want a custom error function.
     // If we leave it as the default, realize() and infer_input_bounds()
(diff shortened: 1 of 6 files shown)


alexreinking (approved):
Looks good to me. Once you backport this to release/15.x, I'll speedrun the release process.

PR #7355, opened 2023-02-15
Remove ParamMap (Fix #7121)

### LINKED ISSUE #7121: Should ParamMap be Deprecated/Removed?
`ParamMap` was added years ago as a way to provide thread-safety when using `realize()` and friends with the JIT. The introduction of `compile_to_callable()` in Halide 15 provides an alternate (and arguably better/cleaner) way to accomplish the same thing; therefore, I'd argue that we should consider a deprecation/removal plan for ParamMap. Would love to hear feedback from other folks on this. (Especially if use ParamMap, either now or in the past, of course.)

steven-johnson:
This PR proposes to remove ParamMap entirely for Halide 16; it was added to provide a threadsafe way to provide parameteres to the JIT, but `compile_to_callable()` now does this in a much less intrusive way.

Normally I'd propose deprecating something like this first (eg via ifdef), but in this case it would be extremely intrusive to do so, so this PR just rips off the band-aid and removes it entirely -- feedback welcome on this.


test/performance/thread_safe_jit.cpp
```diff
@@ -12,11 +12,15 @@
 using namespace Halide;
 using namespace Halide::Tools;
 
+// The Halide compiler is currently not guaranteed to be thread safe.
```
abadams, line 15: It's not? I thought we fixed all the non-threadsafe stuff a long time ago.

abadams:
Sure maybe an ifdef is a pain, but can't we mark all the ParamMap-consuming methods with deprecation warnings for a version before ripping them out?

steven-johnson:
> can't we mark all the ParamMap-consuming methods with deprecation warnings for a version before ripping them out

I guess we could, we'd just have to introduce new overloads for all of them (basically doubling the number of overloads to realize() and infer_input_bounds()), otherwise you'd get deprecation warnings all over the place.

IIRC we know of exactly one user for ParamMap in the first place -- we should contact them specifically and get their feedback.

test/performance/thread_safe_jit.cpp
abadams: Should be fine if no Halide objects are shared across threads. I.e. we don't have any non-threadsafe hidden global state (though temporaries will get non-deterministic names due to atomic counters used to name them).

But yeah the individual objects may have non-threadsafe state, so if you share Halide objects across threads and mutate them bad things will happen.

abadams:
What about adding a deprecation warning to all ParamMap methods except for empty_map() and the default constructor?

steven-johnson:
> What about adding a deprecation warning to all ParamMap methods except for empty_map() and the default constructor?

Looks promising, see https://github.com/halide/Halide/pull/7357

steven-johnson:
Closing this in favor of https://github.com/halide/Halide/pull/7357

PR #7356, opened 2023-02-15
Use a std::unique_ptr for the IR Builder

abadams:
instead of a raw owning pointer

### DIFF
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -287,7 +287,7 @@ void CodeGen_LLVM::initialize_llvm() {
 
 void CodeGen_LLVM::init_context() {
     // Ensure our IRBuilder is using the current context.
-    builder.reset(new IRBuilder<>(*context));
+    builder = std::make_unique<IRBuilder<>>(*context);
 
     // Branch weights for very likely branches
     llvm::MDBuilder md_builder(*context);
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -287,8 +287,7 @@ void CodeGen_LLVM::initialize_llvm() {
 
 void CodeGen_LLVM::init_context() {
     // Ensure our IRBuilder is using the current context.
-    delete builder;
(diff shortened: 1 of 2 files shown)


src/CodeGen_LLVM.cpp
```diff
@@ -287,8 +287,7 @@ void CodeGen_LLVM::initialize_llvm() {
 
 void CodeGen_LLVM::init_context() {
     // Ensure our IRBuilder is using the current context.
-    delete builder;
-    builder = new IRBuilder<>(*context);
+    builder.reset(new IRBuilder<>(*context));
```
steven-johnson, line 290: Suggestion: use `std::make_unique` instead? (Your code isn't wrong but as of C++14 make_unique is in theory the 'preferred' way to do this)

src/CodeGen_LLVM.cpp
```diff
@@ -978,7 +973,7 @@ llvm::Function *CodeGen_LLVM::add_argv_wrapper(llvm::Function *fn,
     std::vector<llvm::Value *> wrapper_args;
     for (llvm::Function::arg_iterator i = fn->arg_begin(); i != fn->arg_end(); i++) {
         // Get the address of the nth argument
-        llvm::Value *ptr = CreateConstGEP1_32(builder, i8_t->getPointerTo(),
+        llvm::Value *ptr = CreateConstGEP1_32(builder.get(), i8_t->getPointerTo(),
```
steven-johnson, line 976: I'm surprised the `.get()` is necessary, doesn't a unique_ptr have an implicit conversion?

steven-johnson (approved):
LGTM but IMHO using `std::make_unique` would be preferable

src/CodeGen_LLVM.cpp
abadams: Apparently not. I get compiler errors without it.

src/CodeGen_LLVM.cpp
abadams: Done

PR #7376, opened 2023-02-24
Minor cleanup of GPUCompilationCache

steven-johnson:
While tracking down an apparently-unrelated threading bug in the webgpu backend, I made some tweaks to this code that I think are worth keeping. The main one of importance is that `release_hold()` and `release_context()` really should acquire the mutex -- they weren't before -- so now all public methods are properly mutexed.

The other changes are mostly cosmetic:
- Moved helper methods to be private rather than public
- Changed the id value size to be `uintptr_t` rather than `uint32_t`; the space allocated for them is sizeof(void*). (Not sure this moves the needle but it felt right.)
- Removed unused ctor for CachedCompilation

Also, drive-by change in printer.h to capture some logging improvements.

### DIFF
--- a/src/runtime/gpu_context_common.h
+++ b/src/runtime/gpu_context_common.h
@@ -116,23 +116,8 @@ class GPUCompilationCache {
         return true;
     }
 
-public:
-    HALIDE_MUST_USE_RESULT bool lookup(ContextT context, void *state_ptr, ModuleStateT &module_state) {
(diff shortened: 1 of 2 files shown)


steven-johnson:
Looks like at least one test (correctness_gpu_object_lifetime_1) is hanging...

abadams (approved):

Issue #7374, opened 2023-02-24
Inefficient loops generated with 2D RDom 


adrian-lebioda:
Hello,

I encountered issue that loops generated from reduction domain where inefficient. Halide properly detected stop condition for `x` rdom variable loop and limited iteration in this dimension however for `y` rdom variable not only it did not detect it properly as loop limit but also put if branch inside inner loop.

I encountered this issue in v14.0.0 and also on release/15.x branch.

I was able to reduce code to this reproduction case (probably it could be reduced further as at some point I just complicated things to try to trigger it, however I think main point is complicating `totalEntries` expression):

```cpp
#include "Halide.h"

using namespace Halide;

class BugGenerator : public Generator<BugGenerator>
{
public:
    static constexpr int MAX_ENTRIES = 512;

    Input<Buffer<std::uint32_t, 2>> input {"input"};
    Input<Buffer<bool, 3>> valid {"valid"};
    Output<Buffer<std::uint32_t, 2>> result {"result"};

    Func internal {"internal"};
    Var x, y, index;

    void generate()
    {
        const Expr entry1 = input(x / 2, y / 2);
        const Expr entry2 = input(x / 2, y / 2);

        const auto dim0 = input.dim(0);
        const auto dim1 = input.dim(1);
        const Expr entry3 = input(
            clamp(cast<int>(entry1), dim0.min(), dim0.min() + dim0.extent()),
            clamp(cast<int>(entry2), dim1.min(), dim1.min() + dim1.extent()));
        const Expr totalEntries = 1 + 2 * entry1 + entry3;

        Func entries;
        entries(x, y, index) = Tuple(cast<int>(input(x, y) % 10), y, index);

        RDom rdom(0, MAX_ENTRIES, 0, MAX_ENTRIES, "internalRDOM");
        rdom.where(rdom.y < totalEntries);
        rdom.where(rdom.x < totalEntries);

        const Tuple entry = entries(x, y, rdom.y);
        rdom.where(valid(entry.as_vector()));

        internal(x, y, index) = cast<std::uint32_t>(0);

        const auto currentIndex = clamp(totalEntries - rdom.x - 1, 0, MAX_ENTRIES);
        const Expr currentValue = internal(x, y, currentIndex);
        rdom.where(currentValue < 1024);

        internal(x, y, currentIndex) = select(currentValue < 512,
            currentValue + cast<std::uint32_t>(rdom.y),
[...]

PR #7379, opened 2023-02-27
Bounds visitor for div was missing single_point mutated case

adrian-lebioda:
So further investigating issue I reported #7374 I found that also `Div` visit misses this case which results in not bounded loop being generated.


abadams (approved):
Thanks for the fix!

PR #7380, opened 2023-02-27
Add Callable default ctor + `defined()` method

steven-johnson:
This allows it to behave like other Halide IR pieces.

### DIFF
--- a/src/Callable.cpp
+++ b/src/Callable.cpp
@@ -47,7 +47,11 @@ void destroy<CallableContents>(const CallableContents *p) {
 }  // namespace Internal
 
 Callable::Callable()
-    : contents(new CallableContents) {
+    : contents(nullptr) {
+}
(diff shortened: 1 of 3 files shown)


src/Callable.cpp
```diff
@@ -197,6 +203,8 @@ Callable::FailureFn Callable::check_fcci(size_t argc, const FullCallCheckInfo *a
 }
 
 int Callable::call_argv_checked(size_t argc, const void *const *argv, const QuickCallCheckInfo *actual_qcci) const {
+    internal_assert(defined());
```
abadams, line 206: I think this should be a user_assert with an error message, because a user might actually try to call a default-constructed callable

steven-johnson: Done

abadams (approved):

PR #7384, opened 2023-02-28
Specify a full type for llvm::IRBuilder

vksnk:
This is a partial revert of https://github.com/halide/Halide/commit/32468447f761250574b1a7af5af77f1b2ad18668, otherwise I see a lot of random crashes inside LLVM in Google testing.


abadams (approved):

steven-johnson:
How very weird that it's not failing here

steven-johnson:
Don't you also need to change this line in the .cpp file?

> builder = new IRBuilder<>(*context)

PR #7386, opened 2023-03-01
Destringify CanonicalizeGPUVars

abadams:
This new implementation takes the high-water marks of each type of GPU loop, instead of filtering using the prefix of the loop name.

Alternative to #6984

### DIFF
--- a/src/CanonicalizeGPUVars.cpp
+++ b/src/CanonicalizeGPUVars.cpp
@@ -29,51 +29,42 @@ string get_block_name(int index) {
 }
 
 class CountGPUBlocksThreads : public IRVisitor {
-    string prefix;  // Producer name + stage
-
     using IRVisitor::visit;
 
     void visit(const For *op) override {
-        if (starts_with(op->name, prefix)) {
-            if (op->for_type == ForType::GPUBlock) {
-                nblocks++;
-            } else if (op->for_type == ForType::GPUThread) {
-                nthreads++;
-            } else if (op->for_type == ForType::GPULane) {
-                nlanes++;
-            }
-        }
+        // Figure out how much to increment each counter by based on the loop
+        // type.
+        int db = op->for_type == ForType::GPUBlock;
+        int dl = op->for_type == ForType::GPULane;
+        int dt = op->for_type == ForType::GPUThread;
+
+        // The threads counter includes lanes loops
+        dt += dl;
+
+        // Increment counters
+        nb += db;
+        nl += dl;
+        nt += dt;
+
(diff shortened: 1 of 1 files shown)


steven-johnson:
+1 to destringification :-)

src/CanonicalizeGPUVars.cpp
```diff
@@ -29,51 +29,42 @@ string get_block_name(int index) {
...
-            nthreads = std::max(then_nthreads, nthreads);
-            nlanes = std::max(then_nlanes, nlanes);
-        }
+        // Decrement counters
+        nb -= db;
+        nl -= dl;
+        nt -= dt;
     }
 
+    int nb = 0, nt = 0, nl = 0;
```
rootjalex, line 63: I think this needs more documentation. How are these counters different from `nblocks`, `nthreads`, and `nlanes`? I can't tell from the existing code / comments in the `For` visitor.

src/CanonicalizeGPUVars.cpp
```diff
@@ -29,51 +29,42 @@ string get_block_name(int index) {
...
+
+        // The threads counter includes lanes loops
+        dt += dl;
+
+        // Increment counters
+        nb += db;
+        nl += dl;
+        nt += dt;
+
+        // Update the high water mark.
```
rootjalex, line 49: Is "high water mark" a known term? I have no idea what that means.

steven-johnson: https://en.wikipedia.org/wiki/High_water_mark
https://encyclopedia2.thefreedictionary.com/high-water+mark
https://stackoverflow.com/questions/45489405/what-are-high-and-low-water-marks-in-bit-streaming

etc

src/CanonicalizeGPUVars.cpp
```diff
@@ -46,8 +50,8 @@ class CountGPUBlocksThreads : public IRVisitor {
         nl += dl;
         nt += dt;
 
-        // Update the high water mark.
-        nblocks = std::max(nb, nblocks);
+        // Update the maximum counter values seen.
+        Nblocks = std::max(nb, nblocks);
```
rootjalex, line 54: this seems like a typo? `Nblocks` -> `nblocks`

abadams: There's some key command in emacs that capitalizes a word that I accidentally hit from time to time.

abadams:
PTAAL

steven-johnson (approved):

PR #7388, opened 2023-03-01
hannk: Provide weak symbol functions to use op profiling

terryheo:
You can add your own profiler with strong symbols.

### DIFF
--- a/apps/hannk/interpreter/model.cpp
+++ b/apps/hannk/interpreter/model.cpp
@@ -5,6 +5,8 @@
 #include <cmath>
 #include <list>
 
+#if HANNK_PROFILER
+
 #define WEAK __attribute__((weak))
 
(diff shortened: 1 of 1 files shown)


apps/hannk/interpreter/model.cpp
```diff
@@ -5,6 +5,15 @@
 #include <cmath>
 #include <list>
 
+#define WEAK __attribute__((weak))
```
steven-johnson, line 8: This won't compile on Windows where weak linking doesn't exist, so the buildbots here will likely break. At a minimum, wrap these changes in `#ifndef _MSC_VER`.

terryheo: Added compiler guard with HANNK_PROFILER

steven-johnson (approved):

PR #7397, opened 2023-03-07
Revise pip.yml to produce true nightlies

steven-johnson:
Instead of trying to produce new package(s) for every commit (which sometimes happens multiple times per day), it seems more reasonable (and less confusing) to produce nightlies instead. This PR is configured to trigger every night at 8:15 PST.

Note that both the current status quo, and this PR, have a nontrivial defect that is hard to fix: they need to build against an LLVM image that we have in a Docker image; for release branches, that's a relatively fixed target, but for top-of-tree Halide, it really should be a working  snapshot of top-of-tree LLVM. (At the time of this PR, Halide main should really have nightlies built against LLVM 17.x work-in-progress, but this PR builds against LLVM 15.0.7 because it's the most recent thing we have available. This is highly suboptimal.)


alexreinking:
We probably want to go for every commit on release branches and for PRs to release branches.

We also want to run the real deal when a release is created.

steven-johnson:
> We probably want to go for every commit on release branches and for PRs to release branches.
> 
> We also want to run the real deal when a release is created.

Agreed.

steven-johnson:
This has been sitting out here so long I forget where it stands.

alexreinking:
This has been superseded by #8405 🎉

PR #7399, opened 2023-03-09
Cleanup of src/runtime/internal and test/runtime

steven-johnson:
- Don't include .cpp files.
- Don't use header-only "libraries" that rely on include order or being included only once (to wit: test/runtime/common.h -> common.cpp)
- All files should explicitly #include what they need, even if they think it's already included (ie, order of include files should not matter)
- In src/runtime/internal, change all `halide_abort_if_false` -> `halide_debug_assert`
- in test/runtime, add HALIDE_CHECK to common.h and use it for tests instead of `halide_abort_if_false`


derek-gerstmann:
Yes, this is much cleaner. Thanks for disentangling the header dependencies.

Looks good to me.

derek-gerstmann (approved):

PR #7402, opened 2023-03-10
Docs/update readme python

Ttayu:
- There is no `imsave` function in `halide/imageio.py`. Use `imwrite`
- If you assume pip installed `imageio`, the following error will occur:
   ```
   ValueError: Image must be 2D (grayscale, RGB, or RGBA).
   ```


steven-johnson (approved):
Thanks for the fix!

PR #7404, opened 2023-03-10
Modify runtime calls to always return a valid halide_error_code_t value

steven-johnson:
Currently, the return values from our runtime code are a mishmash -- there's lots of code that returns any random nonzero value to indicate an error. This isn't wrong per se, but it's not clean, and it's desirable that the return values are predictable. This PR doesn't change the call signature of any (public) Halide Runtime functions, but modifies the internal logic so that all return values are valid values of `enum halide_error_code_t`.

Generally, there should be minimal change to the code otherwise, although I did leave in a few drive-by changes that I couldn't resist (e.g., better error-checking when dynamically loading symbols).

My long-term goal here is to eventually propose changing the signature of runtime functions that return errors to actually return `enum halide_error_code_t`; as you might imagine, making that transition might be controversial for a number of reasons. This PR is intended to be a way to make such a future transition easier to reason about, while arguably improving the code quality of the runtime slightly.

(This PR might be too large to review conveniently -- if so, I can break it into smaller pieces.)


src/CMakeLists.txt
```diff
@@ -424,6 +424,7 @@ target_compile_options(
         Halide
         PRIVATE
         $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wall>
+        $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Werror>
```
abadams, line 427: I'm a little surprised we didn't already have this, if I'm understanding "PRIVATE" correctly. @alexreinking does this make sense?

alexreinking: It does not make sense to do this. One should _never_ hard-code `-Werror` in a `CMakeLists.txt` file. We just don't have enough control over downstream toolchains as an open source project for this to be safe.

This should be set in CI/buildbots via the [`CMAKE_COMPILE_WARNING_AS_ERROR`](https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html) cache variable.

alexreinking (changes requested):
Remove `-Werror` from CMakeLists.txt

src/runtime/cuda.cpp
```diff
@@ -153,16 +166,15 @@ WEAK int halide_default_cuda_acquire_context(void *user_context, CUcontext *ctx,
...
         }
 
         {
             ScopedMutexLock spinlock(&context_lock);
             local_val = context;
             if (local_val == nullptr) {
-                CUresult error = create_cuda_context(user_context, &local_val);
-                if (error != CUDA_SUCCESS) {
-                    return error;
+                if (auto result = create_cuda_context(user_context, &local_val); result != halide_error_code_success) {
```
abadams, line 176: stylistically this looks awkward to me. What do you think about a line break after the semicolon when we do this?

src/runtime/cuda.cpp
```diff
@@ -252,75 +264,84 @@ namespace Cuda {
 
 // Helper object to acquire and release the cuda context.
 class Context {
-    void *user_context;
+    void *const user_context;
+    int status_;  // must always be a valid halide_error_code_t value
```
abadams, line 268: Why the underscore?

src/runtime/cuda.cpp
```diff
@@ -676,9 +678,9 @@ WEAK int halide_cuda_device_free(void *user_context, halide_buffer_t *buf) {
             // We don't want to use a buffer freed one stream on
             // another, as there are no synchronization guarantees and
             // everything is async.
-            int result = halide_cuda_get_stream(user_context, ctx.context, &item->stream);
```
abadams, line 679: This is a minor change in contract. If you override halide_cuda_get_stream you should now call halide_error if there's an error (or signal your error some other way). We should clarify the comment in HalideRuntimeCuda.h

src/runtime/cuda.cpp
```diff
@@ -1228,36 +1227,37 @@ WEAK int halide_cuda_device_and_host_free(void *user_context, struct halide_buff
 WEAK int halide_cuda_wrap_device_ptr(void *user_context, struct halide_buffer_t *buf, uint64_t device_ptr) {
     halide_abort_if_false(user_context, buf->device == 0);
     if (buf->device != 0) {
-        return -2;
+        error(user_context) << "halide_cuda_wrap_device_ptr: device is nonzero";
```
abadams, line 1230: suggest changing to "halide_cuda_wrap_device_ptr: device field is already non-zero" to clarify it slightly

src/runtime/cuda.cpp
```diff
@@ -1445,7 +1437,6 @@ WEAK const char *get_error_name(CUresult err) {
     default:
         // This is unfortunate as usually get_cuda_error is called in the middle of
```
abadams, line 1438: Delete stale comment

src/runtime/device_interface.h
```diff
@@ -1,8 +1,10 @@
 #ifndef HALIDE_RUNTIME_DEVICE_INTERFACE_H
 #define HALIDE_RUNTIME_DEVICE_INTERFACE_H
 
+#include "HalideRuntime.h"
+
 #ifndef WEAK
```
abadams, line 6: Can cut the ifndef entirely if including HalideRuntime.h

src/runtime/hexagon_cache_allocator.cpp
```diff
@@ -139,16 +139,16 @@ inline int hexagon_cache_pool_free(void *user_context) {
         if (temp->l2memory != nullptr) {
             err = HAP_cache_unlock(temp->l2memory);
             if (err != QURT_EOK) {
-                error(user_context) << "Hexagon: HAP_cache_unlock failed on pool free\n";
-                return err;
+                error(user_context) << "Hexagon: Hexagon: HAP_cache_unlock failed on pool free";
```
abadams, line 142: Double Hexagon:

src/CMakeLists.txt
steven-johnson: This change was meant to be local-only, to flush out the instances.

(Alex and I have a longstanding disagreement over the value of `-Werror`, but this is not the place to resolve it :-)

src/runtime/cuda.cpp
steven-johnson: I thought the line-break version also looked awkward, but let's switch to that and see what you think.

src/runtime/cuda.cpp
steven-johnson: scalpel left in patient, fixing

src/runtime/cuda.cpp
steven-johnson: Done (also applied to opencl.cpp)

src/runtime/device_interface.cpp
```diff
@@ -594,7 +590,10 @@ WEAK int halide_buffer_copy_already_locked(void *user_context, struct halide_buf
         }
     }
 
-    return err;
+    if (result != halide_error_code_success) {
```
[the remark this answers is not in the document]
steven-johnson: We could do that, but I have an idea for a more general fix for calling halide_error() that I'd like to discuss elsewhere

src/runtime/device_interface.h
steven-johnson: done

src/runtime/device_interface.cpp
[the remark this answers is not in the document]
steven-johnson: > Haven't the device implementations potentially already called halide_error at this point?

But in response to your point:
- It's not currently enforced (or reliably coded) that device implementations do this, hence the possibly-redundant call here. But if they did:
    - If halide_error() calls abort(), having a redundant call doesn't matter because we are dead
    - if halide_error() just logs, we get a redundant log, no biggie

src/runtime/hexagon_cache_allocator.cpp
steven-johnson: Hexagon Hexagon Hexagon Hexagon Hexagon Hexagon ?

steven-johnson:
PTAL

alexreinking (approved):
For my part, I'm happy to approve.

src/runtime/hexagon_host.cpp
```diff
@@ -432,23 +444,26 @@ WEAK int halide_hexagon_run(void *user_context,
...
                                            input_buffers);
     if (input_buffer_count < 0) {
-        return input_buffer_count;
+        error(user_context) << "Hexagon: map_arguments() failed.";
+        return halide_error_code_generic_error;
     }
 
     // Then the output buffers (bit 1 of flags is set).
     remote_buffer *output_buffers = input_buffers + input_buffer_count;
     int output_buffer_count = map_arguments(user_context, arg_count, arg_sizes, args, arg_flags, 0x2, 0x2,
```
abadams, line 453: map_arguments notionally returns an error code, so it will have already called halide_error (though it looks like it can't actually throw an error).

src/runtime/opencl.cpp
```diff
@@ -99,6 +103,15 @@ WEAK char build_options[1024];
 WEAK ScopedSpinLock::AtomicFlag build_options_lock = 0;
 WEAK bool build_options_initialized = false;
 
+template<typename... Args>
+halide_error_code_t error_opencl(void *user_context, cl_int cl_error, const Args &...args) {
+    if (cl_error == CL_SUCCESS) {
+        return halide_error_code_success;
+    }
+    error(user_context).append(get_opencl_error_name(cl_error), args...);
```
abadams, line 111: There might need to be a " " between the opencl error name and the rest of the args. As far as I understand, .append doesn't add spaces. Comment also applies to other gpu runtimes.

src/runtime/cuda.cpp
```diff
@@ -61,46 +61,59 @@ extern "C" WEAK void *halide_cuda_get_symbol(void *user_context, const char *nam
...
-WEAK CUresult create_cuda_context(void *user_context, CUcontext *ctx);
+WEAK const char *get_cuda_error_name(CUresult error);
+WEAK int create_cuda_context(void *user_context, CUcontext *ctx);
+
+template<typename... Args>
+int error_cuda(void *user_context, CUresult cuda_error, const Args &...args) {
+    if (cuda_error == CUDA_SUCCESS) {
+        return halide_error_code_success;
+    }
+    error(user_context).append(get_cuda_error_name(cuda_error), args...);
```
abadams, line 114: My earlier comment was suggesting something like:

```error(user_context).append("CUDA error: ", get_cuda_error_name(cuda_error), " ", args...);```

And then removing "CUDA: " in the error strings below.

I don't think the error line should lead with a cuda enum value name.

File: src/runtime/opencl.cpp (hunk 11 above)
Comment (abadams) [reply src/runtime/opencl.cpp:111]: Also see above comment I just added to cuda.cpp

Comment (steven-johnson) [reply src/runtime/opencl.cpp:111]: > There might need to be a " " between the opencl error name and the rest of the args. As far as I understand, .append doesn't add spaces. Comment also applies to other gpu runtimes.

Yeah, I just now removed the 'sep' argument entirely

File: src/runtime/cuda.cpp (hunk 12 above)
Comment (steven-johnson) [reply src/runtime/cuda.cpp:114]: Ah. Ok.

File: src/runtime/opencl.cpp [hunk 13]
```diff
@@ -1222,7 +1234,8 @@ WEAK int halide_opencl_device_and_host_free(void *user_context, struct halide_bu
 WEAK int halide_opencl_wrap_cl_mem(void *user_context, struct halide_buffer_t *buf, uint64_t mem) {
     halide_abort_if_false(user_context, buf->device == 0);
     if (buf->device != 0) {
-        return -2;
+        error(user_context) << "halide_opencl_wrap_cl_mem: device is nonzero.";
```
abadams, line 1237: Same clarification as in cuda.cpp: "device field is already non-zero"

src/runtime/opencl.cpp
```diff
@@ -1497,7 +1512,7 @@ WEAK const char *get_opencl_error_name(cl_int err) {
     case CL_INVALID_DEVICE_PARTITION_COUNT:
         return "CL_INVALID_DEVICE_PARTITION_COUNT";
     default:
-        return "<Unknown error>";
+        return "OpenCL:<Unknown error>";
```
abadams, line 1515: Remove "OpenCL:" here

src/runtime/opencl.cpp
```diff
@@ -1824,10 +1844,12 @@ WEAK int halide_opencl_image_device_and_host_free(void *user_context, struct hal
 WEAK int halide_opencl_image_wrap_cl_mem(void *user_context, struct halide_buffer_t *buf, uint64_t mem) {
     halide_abort_if_false(user_context, buf->device == 0);
     if (buf->device != 0) {
-        return -2;
+        error(user_context) << "halide_opencl_image_wrap_cl_mem: device is nonzero.";
```
abadams, line 1847: device field is already non-zero

src/runtime/hexagon_host.cpp
steven-johnson: Sure, but the fundamental issue here is that with our current setup, there isn't a good way to detect that other than manual checking, and it's super-fragile (no obvious way to keep this in sync). With the status quo, calling error() too much is better than calling it not enough (since the latter risks dropping errors on the floor).

(FWIW, all this runtime churn is basically lead-up to a proposal for cleaning up the protocol in a way that should make this more tractable)

abadams: I thought our new rule was that if something returns a halide_error_code_t, then it has already called halide_error if necessary.

abadams: Also, if something returns a halide_error_code_t, return that error code upwards, don't rewrite it as the code here is doing.

abadams: (I'm reading map_arguments as intending to return an error code if the return value is negative)

src/runtime/cuda.cpp
```diff
@@ -61,46 +61,59 @@ extern "C" WEAK void *halide_cuda_get_symbol(void *user_context, const char *nam
...
-WEAK CUresult create_cuda_context(void *user_context, CUcontext *ctx);
+WEAK const char *get_cuda_error_name(CUresult error);
+WEAK int create_cuda_context(void *user_context, CUcontext *ctx);
+
+template<typename... Args>
+int error_cuda(void *user_context, CUresult cuda_error, const Args &...args) {
+    if (cuda_error == CUDA_SUCCESS) {
+        return halide_error_code_success;
+    }
+    error(user_context).append("CUDA error: ", get_cuda_error_name(cuda_error), args...);
```
abadams, line 114: Does this still need a " " between the cuda error name and the rest of the args?

src/runtime/hexagon_host.cpp
steven-johnson: > I thought our new rule was that if something returns a halide_error_code_t, then it has already called halide_error if necessary.

That was a rule that I had proposed, but I decided to abandon it (for now) because 
   (1) relying on that rule in a reliable way really needs us to convert all the `int` return types to `halide_error_code_t`, so that the person writing/reading the call site doesn't have to go inspect the callee to be satisfied that the return value really is a `halide_error_code_t`; I am very much in favor of making such a change, but it's quite invasive and making the change would require all downstream users to tweak runtime overrides to match, and while that's likely to be an easy change in almost all cases, it's a change we'd need to discuss more widely and get consensus on
    (2) I think there is a simpler approach we can use, which I haven't yet proposed here (basically, I've been harvesting IMHO-useful changes from a grand experiment and trying to get them landed first)
    
> Also, if something returns a halide_error_code_t, return that error code upwards, don't rewrite it as the code here is doing.

Fun fact: `map_arguments() doesn't return a `halide_error_code_t` at all, it returns an actual count! (See what I said above about having to inspect the callee? Yikes.) 

Another fun fact: `map_arguments()` cannot ever return a negative value, so this clause was never firing in the first place -- it really should be removed entirely.

> (I'm reading map_arguments as intending to return an error code if the return value is negative)

That's what the comment for it says, but (1) it doesn't actually do that, and (2) IMHO it's usually a false economy to handle errors in this way (ie, by using magic values of the return result to signal an error). Returning `nullptr` as failure from thing-returning-pointer is arguably reasonable (since `nullptr` is never a valid ptr), but returning negative values from an int result is tempting fate for a misinterpretation of the results someday. (Obviously, if it could be shown that there was meaningful codesize/performance savings by doing such a thing, it would be worth entertaining, but I will bet you a dollar that that isn't the case here.)

src/runtime/opencl.cpp
abadams, line 111: Similar change should happen here to the cuda change:

```
    error(user_context).append("OpenCL error: ", get_opencl_error_name(cl_error), " ", args...);
```

src/runtime/cuda.cpp
steven-johnson: Fixed

src/runtime/opencl.cpp
steven-johnson: Fixed

abadams (approved):

PR #7405, opened 2023-03-10
Dont use negative values for exit()

steven-johnson:
A program that terminates via `exit(-1)` leaves `$? = 255` (per the exit(3) manpage), as the value is chopped with `& 0377`.

This makes it hard to use `git bisect` to track down bugs, as considers any exit code > 127 (or = 15, oddly enough) equivalent to an `abort()` and terminates the bisect.

IMHO this is perverse behavior on the part of `git bisect`, but it is what it is, so let's revamp our tests to avoid calling `exit(-1);`.


abadams:
I believe we also have a lot of main functions that return -1. Do those need fixing too?

abadams (approved):

PR #7407, opened 2023-03-10
Fix overflow in x86 absd lowering

abadams:
The existing code was codegenning an int32 replacement for an Expr of type uint32. One option was to cast the result to a uint32, but we want wrapping semantics on that subtract, so it's better to just do an unsigned subtraction.

This should fix the breakage with trunk llvm.


steven-johnson (approved):
LGTM pending green

PR #7409, opened 2023-03-11
docs: Use halide.imageio instead of imageio

Ttayu:
I have added the missing fix in the #7402 pull request. (i'm sorry)
Additionally, I have added assert statements. This is because the image created in the first example is of shape `[512, 512, 3]`, while the example being used to compute xor_filter is of shape `[512, 512]`. To avoid confusion for the user, I made it explicit.


steven-johnson (approved):
Thanks for the fix!

PR #7411, opened 2023-03-13
Remove the result code from halide_reuse_device_allocations() et al

steven-johnson:
(Split from #7403)

The current implementation of halide_reuse_device_allocations() and friends cannot ever return a nonzero error code. I'd argue that even if errors could be associated with reusing allocations, it is likely more sensible to return errors from the relevant alloc/free calls.

This PR proposes removing the result code entirely to remove confusion, at the cost of a (minor) API breakage, which should be easy to detect at compile time.

### DIFF
--- a/src/runtime/HalideRuntime.h
+++ b/src/runtime/HalideRuntime.h
@@ -1963,8 +1963,13 @@ extern double halide_float16_bits_to_double(uint16_t);
  *
  * If set to false, releases all unused device allocations back to the
  * underlying device APIs. For finer-grained control, see specific
- * methods in each device api runtime. */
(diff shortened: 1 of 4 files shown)


abadams:
Turning off the allocation cache frees a bunch of device memory, which I believe can fail, e.g. due to a double-free bug.

abadams:
There are a variety of other errors it can return. Worst of all, according to the cuda docs: "Note that this function may also return error codes from previous, asynchronous launches."

steven-johnson:
OK, in that case we need to preserve the return result and ensure that callers actually check the result -- closing this.

Issue #7248, opened 2022-12-20
[WebGPU] Unify Emscripten and dawn-native paths in the runtime


jrprice:
There are a few differences in the native APIs implemented by Emscripten and Dawn that mean Halide's WebGPU backend currently requires a build-time switch to select between the two. Once these differences are resolved, we should be able to remove this conditional logic and the build-time switch.

The differences are:
- Emscripten missing `wgpuCreateInstance` and `wgpuReleaseInstance`
- Emscripten missing `wgpuAdapterGetLimits`
- Waiting for async device init callbacks: `emscripten_sleep` vs `usleep`
- Waiting on other async callbacks: `emscripten_sleep()` vs `wgpuDeviceTick`

steven-johnson:
Is there any plan as to when/if this will get resolved?

steven-johnson:
One comment in https://github.com/halide/Halide/pull/6492 suggests that the lack of support for `wgpuInstanceProcessEvents()` is also tied to this issue, but I don't see it mentioned above

jrprice:
> lack of support for wgpuInstanceProcessEvents() is also tied to this issue

Yeah, this is the expected name for the solution for the `emscripten_sleep()` vs `wgpuDeviceTick()` discrepancy. I've added it to the original comment to make that clearer.

> Is there any plan as to when/if this will get resolved?

We're focusing on getting V1 of WebGPU shipped in Chrome really soon [1], and then we expect to have a renewed focus on making the native vs web story better. So yes, these things will get resolved, but I can't give an exact timeframe right now.

[1]: [Intent to ship: WebGPU (Chromium)](https://groups.google.com/a/chromium.org/g/blink-dev/c/VomzPhvJCxI/m/SUhU9Z0vAgAJ)

steven-johnson:
Ah, gotcha, thanks. FYI, googling for `wgpuInstanceProcessEvents()` shows that it is in Emscripten's current version of `webgpu.h`: https://github.com/emscripten-core/emscripten/blob/main/system/include/webgpu/webgpu.h -- is it already implemented there?

jrprice:
> is it already implemented there?

It's in the header (which I think is just a copy of Dawn's header), but the implementation is still absent:
https://github.com/emscripten-core/emscripten/blob/7de3bc93f0f7b65549215772255779ad8305ce78/src/library_webgpu.js#L2426

steven-johnson:
[...]

PR #7419, opened 2023-03-14
Split WebGPU runtime into two variants (#7248 workaround)

steven-johnson:
Halide promises that you can crosscompile to *any* supported target from a 'stock' build of libHalide. Unfortunately, the initial landing of WebGPU support breaks that promise: we compile the webgpu runtime support (webgpu.cpp) with code that is predicated on `WITH_DAWN_NATIVE` (for Dawn vs Emscripten, respectively).

This means that if you build Halide with `WITH_DAWN_NATIVE` defined, you can *only* target Dawn with that build of Halide; similarly, if you build with `WITH_DAWN_NATIVE` not-defined, you can only target Emscripten. (Trying to use the 'wrong' version will produce link-time errors.)

For people who build everything from source, this isn't a big deal, but for people who just pull binary builds, this is a big problem.

This PR proposes a temporary workaround until the API discrepancies are resolved:
- Compile the existing webgpu.cpp runtime *both* ways
- in LLVM_Runtime_Linker.cpp, select the correct variant based on whether the Target is targeting wasm or not
- Profit!

This is a rather ugly hack, but it should hopefully be (relatively) temporary.

attn @jrprice

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -830,7 +830,8 @@ RUNTIME_CPP_COMPONENTS = \
   trace_helper \
   tracing \
   wasm_cpu_features \
-  webgpu \
+  webgpu_dawn \
+  webgpu_emscripten \
   windows_clock \
   windows_cuda \
   windows_d3d12compute_arm \
--- a/README_webgpu.md
+++ b/README_webgpu.md
@@ -29,8 +29,6 @@ device codegen may be required before it becomes profitable to use.
 
 Halide can generate WebGPU code that can be integrated with WASM code using
 Emscripten.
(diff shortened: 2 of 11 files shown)


jrprice (approved):
LGTM!

Will keep an eye on the native API convergence so we can unify this as soon as possible.

steven-johnson:
Added a few more changes to reflect reality as of this PR, PTAL

steven-johnson:
OSX failure is a performance flake, should be ready to land

vksnk (approved):

PR #7421, opened 2023-03-14
Feature Enhancement: Halide IR HTML Visualization

maaz139:
## Summary
This PR proposes improved printing of Halide IR Stmt code in HTML. In addition to existing features such as syntax highlighting and variable matching, we now provide:
- A visualization tab that highlights high-level control-flow and structure of the program. 
- An assembly tab that displays the assembly code for the program.
- Cost indicators (per line in the program) to emphasize critical portions of the program.
- Interaction between different tabs, allowing users to easily synchronize code with the visualization or assembly.

## Demo Screenshot
ir-viz-demo

## Credit
This PR is built on top of #7056 and most of the code and ideas were contributed by @darya-ver. 

### Modified Files
The following files were modified to introduce the new `stmt_viz` emit flag. When specified, it triggers the generation of `<module>.stmt.viz.html` file. Note: currently this flag requires the `assembly` flag to also be set, not doing so results in an error message and compilation is terminated.
- [src/Module.h](https://github.com/halide/Halide/pull/7421/files#diff-212527e5e3c537f82f90b5459f7dae399dbd6cace14ec40d8bdef0d2cdf73870)
- [src/Module.cpp](https://github.com/halide/Halide/pull/7421/files#diff-c6956448fdbad2b5d83fe9c5f3d4defafc68a645ef9cda30b62f92ef716a0db9)
- [src/Generator.cpp](https://github.com/halide/Halide/pull/7421/files#diff-5b0ae89f895169b02bc53cf1ff95665336538d286147490fd80922d39f4316ff)
 
To jump from a `For` or `ProducerConsumer` node in the visualization to the corresponding line in assembly, we assign unique IDs and print them as comments in the generated assembly: 
- [src/CodeGen_LLVM.h](https://github.com/halide/Halide/pull/7421/files#diff-430144da8edb5d6f8dbc84ea384cb194e6125bd4448ec0573d61dbdd8b095ec8)
- [src/CodeGen_LLVM.cpp](https://github.com/halide/Halide/pull/7421/files#diff-6dad3f193950d162e1218b3bd4b4497a8e1e68111dfedc732f642812fa626df1)

Build changes, primarily the handling of large template files using `binary2cpp`:
- [Makefile](https://github.com/halide/Halide/pull/7421/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)
- [src/CMakeLists.txt](https://github.com/halide/Halide/pull/7421/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4)

### New Files
The following files contain boilerplate html/js/css code used to generate the visualizations.
- [src/irvisualizer/StmtToViz_dependencies.template.html](https://github.com/halide/Halide/pull/7421/files#diff-29da13411552dd05b3df1f965ac44e1fa4fdb2078e3d630559852791311b7329)
- [src/irvisualizer/StmtToViz_javascript.template.html](https://github.com/halide/Halide/pull/7421/files#diff-5c22b409eaceadf4c2412348dbc9050ebdd1c2985f2108bb8f7871269cd9ae98) 
- [src/irvisualizer/StmtToViz_stylesheet.template.html](https://github.com/halide/Halide/pull/7421/files#diff-9be3c1527e2d88661a265f003495b594795bc46915615cf70c51dab3cff876f4) 

The following file contain the new printing code for the IR:
- [src/StmtToViz.cpp](https://github.com/halide/Halide/pull/7421/files#diff-eec62d3f3c6e3c896ffecb5722127cd5b1bdc965cd750a035bde990b5becd24c)
- [src/StmtToViz.h](https://github.com/halide/Halide/pull/7421/files#diff-00857a79867bebfd4634fc0b99100754c3a2eab47b798fca88be7171d6e57795)

The external interface for the new code is a single `print_to_viz` function in `StmtToViz.h`. `StmtToViz.cpp` contains the following new classes:
- CostModel; A basic cost model for the Halide programs. It models compute cost as a basic op-count and the data-movement cost as the number of bits moved.
- AssemblyInfo; Used to map some Halide IR nodes to line-numbers in the assembly file containing the corresponding generated code.
- HTMLCodePrinter; Prints the Stmt IR as code in HTML.
- HTMLVisualizationPrinter; Prints the Stmt IR as a visualization in HTML
- IRVisualizer; The class that generates the output HTML file. Handles all boilerplate code etc.

## Testing
I am able to successfully build the following apps with the `stmt_viz` emit flag. The generated IR/HTML looks correct upon inspection. I am happy to hear ideas for more thorough testing.
    - bgu
    - bilateral_grid
    - blur
    - camera pipe
    - conv_layer
    - depthwise_separable_conv
    - fft
    - hankk
    - harris
    - hexagon_benchmarks
    - hist
    - iir_blur
    - interpolate
    - lens_blur
    - local_laplacian
    - max_filter
    - nl_means
    - resize
    - resnet_50 
    - unsharp
    - wavelet

## Future Work
1. The feature still suffers from some scaling issues for very large IR files. Ex: Resnet_50 generates an html file greater than 15mb and fails to render on my machine. JIT rendering is an option that can be implemented down the road if enough folks find this feature useful.
2. With enough feedback/confidence in the new implementation, we should remove `StmtToHtml` which offers a strict subset of features..


steven-johnson:
(Clarification: does this *replace* #7056 or build on top of it? I'm not entirely sure based on your comment)

maaz139:
@steven-johnson The changes are implemented on top of #7056, although some parts of the code are heavily revised

steven-johnson:
> @steven-johnson The changes are implemented on top of #7056, although some parts of the code are heavily revised

It might be easier to review (and thus, to land) this change if it was collapsed into a single PR.

steven-johnson:
Looks like real failures on Windows:

```
C:\PROGRA~1\MICROS~2\2022\COMMUN~1\VC\Tools\MSVC\1431~1.311\bin\Hostx64\x64\cl.exe  /nologo /TP -DHALIDE_ENABLE_RTTI -DHALIDE_WITH_EXCEPTIONS -DHalide_EXPORTS -DLLVM_VERSION=170 -DUNICODE -DWITH_AARCH64 -DWITH_ARM -DWITH_D3D12 -DWITH_HEXAGON -DWITH_INTROSPECTION -DWITH_METAL -DWITH_NVPTX -DWITH_OPENCL -DWITH_OPENGLCOMPUTE -DWITH_POWERPC -DWITH_RISCV -DWITH_WEBASSEMBLY -DWITH_WEBGPU -DWITH_X86 -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_LIBCPP_ENABLE_ASSERTIONS -D_SCL_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_WARNINGS -D_UNICODE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -IC:\src\build_bot\worker\llvm-17-x86-64-windows\llvm-install\include /DWIN32 /D_WINDOWS /EHsc /O2 /Ob2 /DNDEBUG -MD /W3 /wd4018 /wd4141 /wd4146 /wd4244 /wd4267 /wd4291 /wd4503 /wd4800 -std:c++17 /showIncludes /Fosrc\CMakeFiles\Halide.dir\StmtToViz.cpp.obj /Fdsrc\CMakeFiles\Halide.dir\ /FS -c C:\src\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\src\StmtToViz.cpp
C:\src\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\src\StmtToViz.cpp(3362): error C2026: string too big, trailing characters truncated
C:\src\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\src\StmtToViz.cpp(2723): fatal error C1075: '{': no matching token found

```

maaz139:
> > @steven-johnson The changes are implemented on top of #7056, although some parts of the code are heavily revised
> 
> It might be easier to review (and thus, to land) this change if it was collapsed into a single PR.

The original PR was from Darya's fork and merged into the `ir-viz` branch. We chose not to merge into main as the code still needed some work. This PR revises the ir-viz branch and proposes merging the feature into main. I am unsure how to implement your suggestion to collapse the two PRs. Fwiw, I plan to improve the PR description and include details about the feature from the original PR.

> Looks like real failures on Windows:
> 
> ```
> C:\PROGRA~1\MICROS~2\2022\COMMUN~1\VC\Tools\MSVC\1431~1.311\bin\Hostx64\x64\cl.exe  /nologo /TP -DHALIDE_ENABLE_RTTI -DHALIDE_WITH_EXCEPTIONS -DHalide_EXPORTS -DLLVM_VERSION=170 -DUNICODE -DWITH_AARCH64 -DWITH_ARM -DWITH_D3D12 -DWITH_HEXAGON -DWITH_INTROSPECTION -DWITH_METAL -DWITH_NVPTX -DWITH_OPENCL -DWITH_OPENGLCOMPUTE -DWITH_POWERPC -DWITH_RISCV -DWITH_WEBASSEMBLY -DWITH_WEBGPU -DWITH_X86 -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_LIBCPP_ENABLE_ASSERTIONS -D_SCL_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_WARNINGS -D_UNICODE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -IC:\src\build_bot\worker\llvm-17-x86-64-windows\llvm-install\include /DWIN32 /D_WINDOWS /EHsc /O2 /Ob2 /DNDEBUG -MD /W3 /wd4018 /wd4141 /wd4146 /wd4244 /wd4267 /wd4291 /wd4503 /wd4800 -std:c++17 /showIncludes /Fosrc\CMakeFiles\Halide.dir\StmtToViz.cpp.obj /Fdsrc\CMakeFiles\Halide.dir\ /FS -c C:\src\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\src\StmtToViz.cpp
> C:\src\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\src\StmtToViz.cpp(3362): error C2026: string too big, trailing characters truncated
> C:\src\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\src\StmtToViz.cpp(2723): fatal error C1075: '{': no matching token found
> ```

Thanks for the heads up! I think the very large strings containing JS/CSS used by the feature are unsavory even without this error. I'll look into alternates ways to store this code.

steven-johnson:
> I'll look into alternates ways to store this code.

May I suggest looking at the `binary2cpp` tool that Halide already uses -- e.g. it encodes `HalideRuntime.h` and similar files into static data in Codegen_C.cpp for inclusion in generated code. It would probably be easy to use the same approach here.

steven-johnson:
> > I'll look into alternates ways to store this code.
> 
> May I suggest looking at the `binary2cpp` tool that Halide already uses -- e.g. it encodes `HalideRuntime.h` and similar files into static data in Codegen_C.cpp for inclusion in generated code. It would probably be easy to use the same approach here.

As a timely example of how to do this: https://github.com/halide/Halide/pull/7426

maaz139:
> > > I'll look into alternates ways to store this code.
> > 
> > 
> > May I suggest looking at the `binary2cpp` tool that Halide already uses -- e.g. it encodes `HalideRuntime.h` and similar files into static data in Codegen_C.cpp for inclusion in generated code. It would probably be easy to use the same approach here.
> 
> As a timely example of how to do this: #7426

Perfect. I just pushed a patch that mirrors the solution in your PR.

maaz139:
The build seems to be passing now! I expanded the PR description to help reviewers navigate the changes.

Makefile
```diff
@@ -887,6 +894,7 @@ INITIAL_MODULES = $(RUNTIME_CPP_COMPONENTS:%=$(BUILD_DIR)/initmod.%_32.o) \
                   $(RUNTIME_CPP_COMPONENTS:%=$(BUILD_DIR)/initmod.%_64_debug.o) \
                   $(RUNTIME_EXPORTED_INCLUDES:$(INCLUDE_DIR)/%.h=$(BUILD_DIR)/initmod.%_h.o) \
                   $(C_TEMPLATE_FILES:%=$(BUILD_DIR)/c_template.%.o) \
+				  $(HTML_TEMPLATE_FILES:%=$(BUILD_DIR)/html_template.%.o) \
```
steven-johnson, line 897: nit: fix indentation

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+
+using std::istringstream;
+using std::ostringstream;
+using std::string;
+
+extern "C" unsigned char halide_html_template_StmtToViz_dependencies[];
+extern "C" unsigned char halide_html_template_StmtToViz_stylesheet[];
+extern "C" unsigned char halide_html_template_StmtToViz_javascript[];
+
+// Classes define within this file
```
steven-johnson, line 31: defined

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+class IRVisualizer;
+
+/********************** IRCostModel **********************/
+// A basic cost model for Halide IR. Estimates computation
+// cost through simple op-counting and data-movement cost
+// by counting the number of bits being moved.
+class IRCostModel : public IRVisitor {
+public:
+    IRCostModel()
+        : _max_compute_cost(-1), _max_data_cost(-1), _max_compute_cost_inclusive(-1),
```
steven-johnson, line 46: style nit: Halide never uses an underscore-prefix for member variables. It occasionally uses underscore-suffix (e.g. `max_compute_cost_`), but usually only when necessary to avoid ambiguity with another symbol. Preferred style in Halide is just a plain snake-case name (e.g. `max_compute_cost`). Please update here and elsewhere as needed.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+
+    // Returns the compute cost of a node (estimated using simple op-counting)
+    int get_compute_cost(const IRNode *node, bool _include_subtree_cost) {
+        internal_assert(node != nullptr) << "IRCostModel::get_compute_cost(): node is nullptr\n";
+
+        int cost = -1;
+        if (_compute_cost.count(node)) {
+            cost = _include_subtree_cost ? _compute_cost_inclusive[node] : _compute_cost[node];
+        } else {
+            internal_assert(false) << "IRCostModel::get_compute_cost(): cost lookup failed\n";
```
steven-johnson, line 87: Instead of `internal_assert(false) << foo`, use `internal_error << foo` -- it's identical but terser and cleaner.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+        }
+
+        internal_assert(cost >= 0) << "Cost must not be negative.\n";
+        return cost;
+    }
+
+    // Returns the data movement cost of a node (the number of bits moved in load/store/shuffle ops)
+    int get_datamovement_cost(const IRNode *node, bool _include_subtree_cost) {
```
steven-johnson, line 95: I'm thinking that `data_movement` is a better snake-casification than `datamovement`.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+
+    void visit(const Call *op) override {
+        IRVisitor::visit(op);
+        std::vector<const IRNode *> args;
+        for (const auto &arg : op->args) {
+            args.push_back(arg.get());
+        }
+        set_compute_costs(op, 1, args);
+        // Currently there is no special handling
+        // for intrinsics such as `prefetch`
```
steven-johnson, line 365: Do we think they will need special handling? Does this require a TODO or a tracking issue?

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+        for (const auto &arg : op->args) {
+            args.push_back(arg.get());
+        }
+        args.push_back(op->predicate.get());
+        set_compute_costs(op, 0, args, {});
+        set_data_costs(op, 0, args, {});
+    }
+
+    void visit(const Allocate *op) override {
+        // We do not model allocation/de-allocation costs
```
steven-johnson, line 446: Do we think we will need to at some point? Does this require a TODO or a tracking issue?

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+        // Traverse the module to populate the list of
+        // nodes we need to map and generate their assembly
+        // markers (comments that appear in the assembly code
+        // associating the code with this node)
+        for (const auto &fn : m.functions()) {
+            fn.body.accept(this);
+        }
+
+        // Find markers in asm code
+        istringstream _asm_stream(code);
```
steven-johnson, line 539: *definitely* don't use an underscore-prefix for local variables, ever

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+    // Generate asm markers for Halide loops
+    int _loop_id;
+    int gen_loop_id() {
+        return ++_loop_id;
+    }
+
+    std::regex gen_loop_asm_marker(int id, const string &loop_var) {
+        std::regex dollar("\\$");
+        string marker = "%\"" + std::to_string(id) + "_for " + loop_var;
+        marker = std::regex_replace(marker, dollar, "\\$");
```
steven-johnson, line 580: Does this need a regex to do the substitution? If not using `replace_all` (from Util.h) is likely to be cheaper and simpler to maintain.

EDIT: oh, I see, you're saving the regex for later. Yeah, I guess this works.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+        // Close div holding this module
+        print_closing_tag("div");
+
+        // Pop out to outer scope
+        scope.pop(m.name());
+    }
+
+private:
+    // Handle to output file stream
+    T &stream;
```
steven-johnson, line 693: Since we're holding a mutable reference to an external thing here, probably wise to make this class non-copyable and non-movable to be sure the reference can't go stale

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+    }
+
+    // Prints text to stream
+    void print_text(const string &x) {
+        stream << x;
+    }
+
+    // Prints the button to show or hide a code scope
+    void print_show_hide_icon(int id) {
+        stream << "<div class='show-hide-btn-wrapper'>"
```
steven-johnson, line 907: Here and a few other places: is there any risk that the HTML snippets we inject could get out of sync with the template files in a meaningful way?

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+    // Maaz: This is a legacy function from the old html generator. I made sure it works
+    // with the refactored code but I did not dig into why this function does what it does.
+    void print_cuda_gpu_source_kernels(const string &str) {
+        print_opening_tag("code", "ptx");
+
+        int current_id = -1;
+        bool in_braces = false;
+        bool in_func_signature = false;
+
+        string current_kernel;
```
steven-johnson, line 933: style nit: IMHO, within a given source file, we should prefer to either prefix all the `std` stuff explicitly, or put `using` at the top of the file and omit the prefixes ~everywhere. Mixing the two in this way isn't wrong but makes the code harder to read -- choose one style and stick with it for the file.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+                print_html_element("span", "matched", ")");
+
+                print_toggle_anchor_closing_tag();
+
+                // Add a button to jump to this conditional branch in the viz
+                print_visualization_button("cond-viz-" + std::to_string((uint64_t)nested_if));
+
+                // Update op to the nested if for next loop iteration
+                op = nested_if;
+            }
```
steven-johnson, line 1953: style nit: always use `} else {` (no line break)

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+    void print(const Module &m, AssemblyInfo asm_info) {
+        _assembly_info = std::move(asm_info);
+        for (const auto &fn : m.functions()) {
+            print(fn);
+        }
+    }
+
+private:
+    // Handle to output file stream
+    std::ofstream &stream;
```
steven-johnson, line 2144: Same comment as earlier: probably wise to make this class noncopyable/nonmoveable

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+
+    // Used to track the context within generated HTML
+    std::vector<string> context_stack_tags;
+
+    // Used to translate IR to code in HTML
+    ostringstream ss;
```
steven-johnson, line 2150: Is this ostringsteam used in a way that is accumulates between function calls? At first glance I don't see that, in which case it seems simpler and more robust to just declare it locally where needed, but maybe I'm missing something...

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2900 @@
...
+};
+
+/************** HTMLVisualizationPrinter Class ***************/
+// Visualizes the IR in HTML. The visualization is essentially
+// an abstracted version of the code, highlighting the higher
+// level execution pipeline along with key properties of the
+// execution performed at each stage.
+class HTMLVisualizationPrinter : public IRVisitor {
+public:
+    HTMLVisualizationPrinter(std::ofstream &os)
```
steven-johnson, line 2127: style nit, here and elsewhere: constructors with a single argument should ~always be marked as `explicit`. (In the unusual cases where implicit construction is desired, the ctor should have a comment like `/* not-explicit */` in front of it to indicate this is deliberate.)

steven-johnson (changes requested):
This looks pretty great! Almost all of my comments are style issues that should be very easy to resolve. @abadams, this looks like it's ready for you to review as well.

maaz139:
Thanks Steven! Also adding @slomp as a reviewer since he also worked on this feature with us.

src/irvisualizer/StmtToViz_dependencies.template.html
```diff
@@ -0,0 +1,27 @@
+<!-- Bootstrap links -->
+<link href='https://cdn.jsdelivr.net/npm/bootstrap@5.2.0/dist/css/bootstrap.min.css' rel='stylesheet' integrity='sha384-gH2yIJqKdNHPEq0n4Mqa/HGKIhSkIHeL5AyhkYV8i59U5AR6csBvApHHNl/vI1Bx' crossorigin='anonymous'>
```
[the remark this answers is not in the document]
steven-johnson: > I'd say these are quite reliable and used by millions of websites in production. However, using a CDN means users need an internet connection to view these files properly. It may be worthwhile to add local copies of js/css dependencies just to remove that requirement.

Agreed that would be worthwhile, but doesn't need to block this PR; I'd suggest opening an Issue here to suggest that as future work, so we don't forget about it.

src/StmtToViz.cpp
maaz139: I'm unsure what the best course of action is here. A prefetch performs data-movement that is important to highlight (esp for backends like HVX). Otoh, unless the prefetch parameters are constants, it makes the estimated cost a symbolic expression. I worry that updating the cost model and visualizer code to handle symbolic expressions sensibly will grow the code by a non-trivial amount for perhaps very little gain. We could assign some constant cost to every prefetch to ensure it gets highlighted but that would not be true to the description of the cost model (counting bits moved).

src/StmtToViz.cpp
maaz139: The template files define the javascript and css, whereas what we inject inside `StmtToViz.cpp` is html. Changing the html can definitely break the page as selectors may not match and the buttons will consequently stop working or the styling will not be applied.

Perhaps we can reduce the risk by adding documentation near sensitive code. Let users know that changes to these snippets usually require corresponding changes to the JS.

maaz139:
Addressed the previous round of changes. The build error on `buildbot/halide-testbranch-main-llvm16-x86-64-linux-cmake` seems unrelated to this PR.

steven-johnson (approved):

Makefile
```diff
@@ -55,7 +55,7 @@ endif
 SHELL = bash
 CXX ?= g++
 PREFIX ?= /usr/local
-LLVM_CONFIG ?= llvm-config
+LLVM_CONFIG ?= llvm-config-15
```
abadams, line 58: This doesn't look intentional

src/CodeGen_LLVM.cpp
```diff
@@ -3660,11 +3660,13 @@ void CodeGen_LLVM::return_with_error_code(llvm::Value *error_code) {
 }
 
 void CodeGen_LLVM::visit(const ProducerConsumer *op) {
+    producer_consumer_id++;
+
     string name;
     if (op->is_producer) {
```
abadams, line 3666: Why the underscore in the name, given that the rest of the name is space-separated?

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2892 @@
...
+// by counting the number of bits being moved.
+class IRCostModel : public IRVisitor {
+public:
+    IRCostModel()
+        : max_compute_cost(-1), max_data_cost(-1), max_compute_cost_inclusive(-1),
+          max_data_cost_inclusive(-1) {
+    }
+
+    // Pre-compute all costs to avoid repeated work
+    void comput_all_costs(const Module &m) {
```
abadams, line 47: comput -> compute

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2892 @@
...
+    // We also track the max costs to determine the cost color
+    // intensity for a given line of code
+    int max_compute_cost;
+    int max_data_cost;
+
+    int max_compute_cost_inclusive;
+    int max_data_cost_inclusive;
+
+    /* Utility functions to store node costs in the cost database */
+    void set_compute_costs(const IRNode *node, int node_cost, const std::vector<const IRNode *> &child_nodes) {
```
abadams, line 139: I wonder if these should accepted initializer_lists instead of vectors for efficiency. They always seem to be brace-enclosed literals

abadams: Ah, except for the call visitor

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2892 @@
...
+
+    std::regex gen_prodcons_asm_marker(int id, const std::string &var, bool is_producer) {
+        std::regex dollar("\\$");
+        std::string marker = "%\"" + std::to_string(id) + (is_producer ? "_produce " : "_consume ") + var;
+        marker = std::regex_replace(marker, dollar, "\\$");
+        return std::regex(marker);
+    }
+
+    // Mapping of IR nodes to their asm markers
+    std::map<uint64_t, std::regex> markers;
```
abadams, line 594: Why is this a uint64_t?

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2892 @@
...
+
+    void visit(const For *op) override {
+        // Generate asm marker
+        markers[(uint64_t)op] = gen_loop_asm_marker(gen_loop_id(), op->name);
+        // Continue traversal
+        IRVisitor::visit(op);
+    }
+};
+
+/******************* HTMLCodePrinter Class *******************/
```
abadams, line 616: Please use the following comment style for multi-line comments like this for uniformity with the rest of Halide and for better doxygen docs:

```
/** ClassName
 * description...
 */
```

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2892 @@
...
+    }
+
+    // Prints a button to sync text with visualization
+    void print_visualization_button(std::string id) {
+        stream << "<button class='icon-btn sync-btn' onclick='scrollToViz(\"" << id << "\")'>"
+               << "  <i class='bi bi-arrow-right-square' title='Jump to visualization'></i>"
+               << "</button>";
+    }
+
+    // Maaz: This is a legacy function from the old html generator. I made sure it works
```
abadams, line 924: This comment won't age well. How about:
```
CUDA kernels are embedded into modules as PTX assembly. This routine pretty-prints that assembly format.
```

src/StmtToViz.cpp
steven-johnson: > I wonder if these should accept initializer_lists instead of vectors for efficiency. They always seem to be brace-enclosed literals

<threadjack> when we eventually migrate to C++20 we should probably use `span` for all stuff like this

src/StmtToViz.cpp
maaz139: I used the pointer values as IDs to connect JavaScript with HTML elements

src/CodeGen_LLVM.cpp
maaz139: I think the pattern Darya went for was `<id>_produce|consume <varname>`. I can change it to `<id>_produce|consume_<varname>`. As long as its a uniquely identifying string, it should work for our usecase.

src/StmtToViz.cpp
maaz139: I opened an issue so I remember to do it

src/StmtToViz.cpp
abadams: So they show up as integers in the html source? That would make the generated html non-deterministic, which may interfere with downstream consumers of these files (e.g. if they ever get checked into source the diffs when they update will be pointlessly large). I'd feel better if there were a mapping from const IRNode * to unique ids instead.

src/StmtToViz.cpp
```diff
@@ -0,0 +1,2892 @@
...
+        print_toggle_anchor_opening_tag(id);
+
+        // -- print icon
+        print_show_hide_icon(id);
+
+        // -- print text (fn name and args)
+        //    Note: We wrap the show/hide buttons in a navigation anchor
+        //    that lets us sync text and visualization tabs.
+        print_opening_tag("span", "matched");
+        print_html_element("span", "keyword nav-anchor", "func ", "lowered-func-" + std::to_string((uint64_t)&fn));
```
abadams, line 785: Same issue here

src/StmtToViz.cpp
maaz139: Yes they do. And fair point, this would be really bad for any sort of regression testing. I'll assign unique IDs based on traversal order instead.

maaz139:
Thanks again everyone for you comments. I've opened issues to track the handful of comments that have not already been addressed. Will go ahead and merge this PR now. Please tag me on any issues/feature requests relating to the IR visualizer! :)

steven-johnson:
The arm32 buildbot failure is unrelated -- the overnight LLVM build was busted for 32-bit codegen. OK to land.

PR #7422, opened 2023-03-15
Use env vars instead of CMake vars for Dawn/Node bindings

steven-johnson:
The WebGPU bindings need to know the locations of two external files containing executable code:
- WEBGPU_NATIVE_LIB, a shared library containing the 'native' Dawn implementation, for use with variants of `host-webgpu`
- WEBGPU_NODE_BINDINGS, a Node extension that contains the Node/JS bindings for WebGPU, for use with variants of `wasm-webgpu`

Currently, these are specified via CMake configure-time defines, but this is problematic:

- For JIT compilation, specifying WEBGPU_NATIVE_LIB as a CMake configure-time definition means there is no way to customize what Dawn variant you use (unless you build from source, which would significantly limit the usefulness of prebuilt Halide binaries with WebGPU at the present time)

- for AOT compilation, having these as CMake defines is not unreasonable... assuming you are compiling C++ Generators using CMake. If you aren't using CMake, or are exclusively using the Python bindings, you'll have to figure out the right setup for these on your own, which would be painful.

This PR proposes to replace the CMake defines for these two things with plain old environment variables. I'm not wild about using env vars to point to executable files -- aside from the possible security risk, it's generally problematic IMHO to have build-time options controlled by a silent env var -- but I'm not sure that we have a better option here.

At some point, we may be able to simply assume that Node has the bindings built in, and Dawn (etc) would be available via a predictably-named shared library that can be presumed to be in the current search path... but neither of these is the case at the moment.

attn: @jrprice


steven-johnson:
(This should finally go green once buildbots get to it -- fingers crossed -- so please consider reviewing it)

alexreinking:
CMake should really keep using the cache variable, but initialize it from the environment variable.

No matter what we do, the generated build system (be it Make, Ninja, or VS) won't be sensitive to a change in environment variables (i.e. CMake will not regenerate), so we might as well provide a consistent, platform-independent way of setting these variables (i.e. the cache).

Moreover, this prevents unexpected changes to this value if you have to reconfigure for some _other_ reason and do so from a different environment.

cmake/HalideGeneratorHelpers.cmake
```diff
@@ -718,8 +718,8 @@ function(_Halide_target_link_gpu_libs TARGET VISIBILITY)
     endif ()
 
     if ("${ARGN}" MATCHES "webgpu")
-        if (WEBGPU_NATIVE_LIB)
-            target_link_libraries(${TARGET} INTERFACE ${WEBGPU_NATIVE_LIB})
+        if (DEFINED ENV{HL_WEBGPU_NATIVE_LIB})
+            target_link_libraries(${TARGET} INTERFACE $ENV{HL_WEBGPU_NATIVE_LIB})
         endif ()
```
alexreinking, line 723: There should definitely be some more validation around this. If this library doesn't exist, it should surely be the case that the node version does.

alexreinking:
I would have expected to see something like this in `dependencies/webgpu/CMakeLists.txt`:

```cmake
find_library(
  Halide_WEBGPU_NATIVE_LIB
  NAMES webgpu_dawn wgpu
  HINTS ENV HL_WEBGPU_NATIVE_LIB  # for compatibility with Make
)

find_path(
  Halide_WEBGPU_NODE_BINDINGS
  NAMES dawn.node
  HINTS ENV HL_WEBGPU_NODE_BINDINGS  # for compatibility with Make
)

if (NOT (Halide_WEBGPU_NATIVE_LIB AND NOT Halide_WEBGPU_NODE_BINDINGS))
  message(FATAL_ERROR "At least one of Halide_WEBGPU_NATIVE_LIB or Halide_WEBGPU_NODE_BINDINGS must be set.")
endif ()

if (Halide_WEBGPU_NATIVE_LIB AND NOT TARGET Halide::webgpu)
  add_library(Halide::webgpu UNKNOWN IMPORTED GLOBAL)
  set_target_properties(
    Halide::webgpu
    PROPERTIES
    IMPORTED_LOCATION "${Halide_WEBGPU_NATIVE_LIB}"
  )
endif ()
```

Then you could link to `Halide::webgpu` conditionally if it exists, later.

alexreinking (changes requested):

dependencies/wasm/CMakeLists.txt
```diff
@@ -154,12 +154,16 @@ function(add_wasm_halide_test TARGET)
...
+        if (DEFINED ENV{HL_WEBGPU_NODE_BINDINGS})
+            set(script "\
+                const provider = require('$ENV{HL_WEBGPU_NODE_BINDINGS}')\n\
+                const gpu = provider.create([])\n\
+                const navigator = { gpu: gpu }\n
+                ${script}")
+        else ()
+            message(FATAL_ERROR "You must define the env var HL_WEBGPU_NODE_BINDINGS=/path/to/dawn.node when running wasm tests for Halide.")
+        endif ()
     endif ()
```
alexreinking, line 168: I think you should instead write a launcher script (say `tools/launch_webgpu_test.js` that [reads `process.argv`](https://stackoverflow.com/a/4351548/2137996) to determine which script to `require` and reads `HL_WEBGPU_NODE_BINDINGS` from the environment.

A few benefits:

1. We don't need to generate JavaScript code from CMake
2. CMake doesn't need to try to find `dawn.node` or even know it exists.
3. `HL_WEBGPU_NODE_BINDINGS` is purely an environment variable.

Below, you'd write:

```
add_halide_test("${TARGET}"
                GROUPS ${args_GROUPS}
                COMMAND ${NODE_JS_EXECUTABLE} "${Halide_SOURCE_DIR}/tools/launch_webgpu_test.js" "${TARGET}.js")
```

src/JITModule.cpp
```diff
@@ -117,31 +117,39 @@ void load_metal() {
...
-            break;
-        } else {
-            debug(1) << "not found.\n";
+        debug(1) << (error.empty() ? "found!\n" : "not found.\n");
+        return error;
+    };
+
+    string error;
+
+    auto env_libname = get_env_variable("HL_WEBGPU_NATIVE_LIB");
```
alexreinking, line 131: The error message below should include a reference to `HL_WEBGPU_NATIVE_LIB`.

cmake/HalideGeneratorHelpers.cmake
alexreinking: Ah, but if this is going in HalideGeneratorHelpers, then the logic to find the webgpu library can't go in `dependencies/...`, it needs to either be placed here or factored into a Find module.

steven-johnson:
> CMake should really keep using the cache variable, but initialize it from the environment variable.
> 
> No matter what we do, the generated build system (be it Make, Ninja, or VS) won't be sensitive to a change in environment variables (i.e. CMake will not regenerate), so we might as well provide a consistent, platform-independent way of setting these variables (i.e. the cache).
> 
> Moreover, this prevents unexpected changes to this value if you have to reconfigure for some _other_ reason and do so from a different environment.

I'm a little unclear on precisely what you are suggesting: is the idea that we have a cache variable in CMake that defaults to initialization from an env var? I see what you are getting at, but that opens up a different can of worms, namely, that code using the JIT will immediately recognize changes to the `HL_WEBGPU_NATIVE_LIB` env var, but code using AOT will ignore such changes unless you re-run CMake. (Correct me if I'm wrong, but even an 'ordinary' rebuild via `ninja` etc wouldn't trigger a cached cmake variable to pick up the env var change.) I find this to be unsettling, and while maybe it's not *worse* than the status quo of this PR, it doesn't strike me as *better*. Is there a detail I'm missing here?

steven-johnson:
> I would have expected to see something like this in `dependencies/webgpu/CMakeLists.txt`:

At first I said to myself "but there isn't any such file", but then I realize you were suggesting I add one. Your suggestion looks reasonable, though it does potentially conflict with my concern about binding to a cmake cached var in a previous comment response.

dependencies/wasm/CMakeLists.txt
steven-johnson: Yeah, this one I think I agree with 100%, I'll put this in place shortly.

steven-johnson:
Gentle review ping for the changes (which aren't complete but address concerns partially)

alexreinking:
> Gentle review ping for the changes (which aren't complete but address concerns partially)

This is _better_, but as I wrote in the review comments of #7152, using environment variables for project configuration in CMake is vexing. The generated build systems don't track dependencies on the values of environment variables for the regeneration step; they only track files: the cache file, the CMake code files that were executed as part of the build, and those files which were explicitly added to the `CMAKE_CONFIGURE_DEPENDS` properties (via any of the, like, 10 mechanisms that do this).

I understand you have some concerns about this state of affairs and I probably wouldn't die on this particular hill if the environment variable only affected Halide's own build. However, it's going downstream via `HalideGeneratorHelpers.cmake` and I really don't want to ship CMake anti-patterns to our users.

steven-johnson:
At this point I'm fine with revising it as you describe in #7512 [not #7152 as mentioned above], but I'm unclear on which comments are still relevant, since I've been away from this PR for a while now... 

at one point you mention adding `dependencies/webgpu/CMakeLists.txt`, then subsequently point out that if we use the dependency in HalideGeneratorHelpers that we can't do that after all, and need to add a `Find` module... can I trouble you to synopsize what you think is necessary here to wrap this up The Right Way?

alexreinking:
> can I trouble you to synopsize what you think is necessary here to wrap this up The Right Way?

Absolutely. I'll just do it and put a write up in the comments here.

alexreinking (approved):

steven-johnson:
Looks like the find_package(Halide_WebGPU) is failing on the armbot: https://buildbot.halide-lang.org/master/#/builders/33/builds/238 -- not sure why yet

steven-johnson:
The env var is set correctly:
```
$ ls -l $HL_WEBGPU_NATIVE_LIB
-rwxr-xr-x  1 halidenightly  staff  88849 Mar 13 14:32 /Users/halidenightly/dawn/build-dawn/src/dawn/native/libwebgpu_dawn.dylib
```

alexreinking:
@steven-johnson -- I think that link is invalid after the build-master reset. Is there a new link?

steven-johnson:
> @steven-johnson -- I think that link is invalid after the build-master reset. Is there a new link?

There will be sortly

steven-johnson:
> @steven-johnson -- I think that link is invalid after the build-master reset. Is there a new link?

https://buildbot.halide-lang.org/master/#/builders/76/builds/11
https://buildbot.halide-lang.org/master/#/builders/83/builds/6

steven-johnson:
Doing some debugging by enabling `CMAKE_FIND_DEBUG_MODE`, it seems that CMake expects the hint to be the *directory* to look in, not the full path to the dylib (whereas `HL_WEBGPU_NATIVE_LIB` is explicitly a full path). I'll revise the code accordingly.

steven-johnson:
Hopeful fix pushed

alexreinking (approved):
Sorry it took so long to get to this, but it's good to land!

cmake/FindHalide_WebGPU.cmake
```diff
@@ -0,0 +1,45 @@
...
+  # and specify two common names for the NAMES part.
+  set(WEBGPU_NATIVE_DIR "")
+  set(WEBGPU_NATIVE_NAMES "webgpu_dawn;wgpu")
+endif()
+
+find_library(
+  Halide_WebGPU_NATIVE_LIB
+  NAMES ${WEBGPU_NATIVE_NAMES}
+  HINTS ${WEBGPU_NATIVE_DIR}
+)
```
alexreinking, line 29: I think this can just be:
```suggestion
if (EXISTS "$ENV{HL_WEBGPU_NATIVE_LIB}")
  set(Halide_WebGPU_NATIVE_LIB "$ENV{HL_WEBGPU_NATIVE_LIB}"
      CACHE FILEPATH "")
endif ()

find_library(Halide_WebGPU_NATIVE_LIB NAMES webgpu_dawn wgpu)
```
Since `find_library` will not search if the cache variable already exists.

alexreinking (changes requested):
Sorry, just finished my coffee and thought of this... @steven-johnson - can you test this?

cmake/FindHalide_WebGPU.cmake
alexreinking: @steven-johnson -- I'm looking at https://github.com/halide/Halide/pull/7422/commits/bb01f3c1b8a3d945bb698057165efff8a23991f2 and it looks like this change wasn't applied correctly. The `find_library` call still reads:

```
find_library(
  Halide_WebGPU_NATIVE_LIB
  NAMES ${WEBGPU_NATIVE_NAMES}
  HINTS ${WEBGPU_NATIVE_DIR}
)
```
but both of those variables are undefined. In the diff above, I wrote:
```
find_library(Halide_WebGPU_NATIVE_LIB NAMES webgpu_dawn wgpu)
```

steven-johnson:
ops, fixed?

alexreinking (approved):
LGTM assuming the buildbots like it

PR #7425, opened 2023-03-15
Backport xtensa_codegen to main

steven-johnson:
This backports the branch for Xtensa codegen into main. Note that it deliberately ignores changes to apps/ (which had grown stale); those changes may or may not be backported at a later date.

Aside from `simd_op_check_xtensa`, there are few tests for the new codegen paths; the next step to remedy that will be updating the buildbots to add code coverage for the existing test suite to the extent that is possible via AOT only (since the Xtensa codegen does not support JIT).

### DIFF
--- a/src/Func.h
+++ b/src/Func.h
@@ -2256,6 +2256,9 @@ public:
      */
     Func &async();
 
+    /** TODO: document me */
+    Func &dma();
+
     /** Bound the extent of a Func's storage, but not extent of its
      * compute. This can be useful for forcing a function's allocation
      * to be a fixed size, which often means it can go on the stack.
--- a/src/InjectDmaTransfer.cpp
+++ b/src/InjectDmaTransfer.cpp
@@ -1,0 +1,417 @@
+#include "InjectDmaTransfer.h"
+#include "CSE.h"
+#include "ExprUsesVar.h"
+#include "Function.h"
+#include "IREquality.h"
+#include "IRMutator.h"
+#include "IROperator.h"
+#include "Simplify.h"
+#include "Substitute.h"
+
+namespace Halide {
+namespace Internal {
+
+using std::set;
+using std::string;
+using std::vector;
+
+/** If an integer expression varies linearly with the variables in the
+ * scope, return the linear term. Otherwise return an undefined
+ * Expr. */
+Expr is_linear(const Expr &e, const Scope<Expr> &linear) {
+    if (e.type() != Int(32)) {
+        return Expr();
+    }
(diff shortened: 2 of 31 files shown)


src/InjectDmaTransfer.cpp
```diff
@@ -0,0 +1,417 @@
...
+                body = Block::make(wait_is_done, body);
+
+                return ProducerConsumer::make_produce(op->name, body);
+            }
+        }
+        return IRMutator::visit(op);
+    }
+
+    Stmt visit(const Allocate *op) override {
+        auto it = producers_to_wait.find(op->name);
```
abadams, line 385: This looks like it won't work with Tuples

src/Func.h
```diff
@@ -2256,6 +2256,9 @@ class Func {
      */
     Func &async();
 
+    /** TODO: document me */
```
abadams, line 2259: What's Func::dma()? Is it like Func::copy_to_device()?

src/InjectDmaTransfer.cpp
```diff
@@ -0,0 +1,417 @@
...
+        loop_vars.pop_back();
+        if (loops_to_be_removed.count(op->name) > 0) {
+            loops_to_be_removed.erase(op->name);
+            return mutated.as<For>()->body;
+        }
+        return mutated;
+    }
+
+    Stmt visit(const LetStmt *op) override {
+        // TODO: Not really correct, but probably want to skip lets which
```
abadams, line 166: "Not really correct" sounds alarming

src/InjectDmaTransfer.cpp
```diff
@@ -0,0 +1,417 @@
...
+        std::vector<Expr> dma_extents;
+
+        vector<Expr> store_strides;
+        vector<Expr> value_strides;
+
+        // Compute strides for each of the loop vars.
+        for (const auto &v : loop_vars) {
+            Scope<Expr> local_scope;
+            local_scope.push(v.name, 1);
+            Expr is_linear_store = is_linear(op_index, local_scope);
```
abadams, line 229: op_index and value_index could be combinatorially large due to the substitution above, but is_linear just visits it recursively.

steven-johnson:
This probably needs https://github.com/halide/Halide/pull/7426 to land first, so we can use the technique for long strings there -- the currents string(s) in the new backend are too large for MSVC.

steven-johnson:
My apologies for the inconvenience of reviewers, but for reasons related to keeping some merge procedures simple(r), I'm going to close this PR and re-open it off the `xtensa_codegen` branch.

PR #7426, opened 2023-03-15
Move long boilerplate C/C++ code into template files

steven-johnson:
Codegen_C has a couple of long strings with boilerplate code that is conditionally emitted; at least one of these is too long for a single string literal under MSVC. Let's try moving these into standalone files instead; this may make it easier to use conventional tooling on the C++ code, and make Codegen_C easier to read and think about.

(Note that the not-yet-landed Xtensa branch should also use this approach, if we decide this approach is good, since it has even more such code.)

TODO: probably would be good to augment `binary2cpp` to allow option comments in the source file that are stripped in the output file (e.g. "This file is used in CodeGen_C.cpp for blah blah blah, look out for blah")


Makefile
```diff
@@ -1134,6 +1142,11 @@ $(BUILD_DIR)/initmod_ptx.%_ll.o: $(BUILD_DIR)/initmod_ptx.%_ll.cpp
 $(BUILD_DIR)/initmod.%.o: $(BUILD_DIR)/initmod.%.cpp
 	$(CXX) -c $< -o $@ -MMD -MP -MF $(BUILD_DIR)/$*.d -MT $(BUILD_DIR)/$*.o
 
+$(BUILD_DIR)/c_template.%.o: $(BUILD_DIR)/c_template.%.cpp
+	$(CXX) -c $< -o $@ -MMD -MP -MF $(BUILD_DIR)/$*.d -MT $(BUILD_DIR)/$*.o
+# 	@mkdir -p $(@D)
```
abadams, line 1147: delete commented out lines

abadams:
I think this is probably better, but I'd like to hear from @alexreinking because it's build-related so I'll add him as a reviewer.

vksnk (approved):
Looks great, but I don't really know much about cmake.

vksnk:
> Looks great, but I don't really know much about cmake.

... so it'd be great if someone else reviewed build file changes.

alexreinking:
> I think this is probably better, but I'd like to hear from @alexreinking because it's build-related so I'll add him as a reviewer.

>> Looks great, but I don't really know much about cmake.
> ... so it'd be great if someone else reviewed build file changes.

Looks like an improvement to me. Better to have boilerplate code somewhere it can be separately analyzed.

@steven-johnson - are these files subject to `clang-format`? I suppose they aren't seen by clang-tidy (yet).

steven-johnson:
> are these files subject to clang-format

Not sure. I manually formatted this time.

steven-johnson:
OSX failures are unrelated (will be fixed by https://github.com/halide/Halide/pull/7419), landing

PR #7427, opened 2023-03-16
Always use __atomic in runtime thread operations

steven-johnson:
Upstream LLVM is strongly suggesting we move away from __sync primitives and use __atomic primitives instead.

IIRC, we still use the older __sync primitives on 32-bit because of some tooling issues -- perhaps on arm32 linux?

This PR is really just a test to see if those issues are still the case on our current set of buildbots.

(There are other `__sync` primitives usages in Halide that would need to be migrated as well if we decide to migrate, this is just an initial test)

### DIFF
--- a/src/runtime/synchronization_common.h
+++ b/src/runtime/synchronization_common.h
@@ -88,80 +88,6 @@ ALWAYS_INLINE void if_tsan_post_signal(void *) {
 }
 #endif
 
-#ifdef BITS_32
-ALWAYS_INLINE uintptr_t atomic_and_fetch_release(uintptr_t *addr, uintptr_t val) {
-    return __sync_and_and_fetch(addr, val);
-}
-
-template<typename T>
-ALWAYS_INLINE T atomic_fetch_add_acquire_release(T *addr, T val) {
-    return __sync_fetch_and_add(addr, val);
-}
-
-template<typename T>
-ALWAYS_INLINE bool cas_strong_sequentially_consistent_helper(T *addr, T *expected, T *desired) {
-    T oldval = *expected;
-    T gotval = __sync_val_compare_and_swap(addr, oldval, *desired);
-    *expected = gotval;
-    return oldval == gotval;
-}
-
-ALWAYS_INLINE bool atomic_cas_strong_release_relaxed(uintptr_t *addr, uintptr_t *expected, uintptr_t *desired) {
-    return cas_strong_sequentially_consistent_helper(addr, expected, desired);
-}
-
-ALWAYS_INLINE bool atomic_cas_weak_release_relaxed(uintptr_t *addr, uintptr_t *expected, uintptr_t *desired) {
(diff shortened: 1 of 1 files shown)


abadams:
IIRC the issue with the __atomic primitives is that they might call a library function if they fail to prove alignment (whether or not it's actually aligned). These library functions don't always exist for us. The __sync functions in contrast assume alignment and have UB on misalignment.

abadams:
... and because we compile our runtime in a weird way, llvm can't prove alignment

zvookin:
I may be wrong, but at the time I relented and just used __sync, I convinced myself that there were 32-bit architectures where using __atomic on a 32-bit data type required libatomic. The reason is that __atomic guarantees atomicity across all the operations in the API, even if there is no an appropriate instruction for one of the things in the API. In that case, it has to fall back to a synthesized mechanism rather than just generating an instruction. the __sync API either does not offer than guarantee or can do so more often in hardware because it has a smaller set of operations.

The backup plan here is to implement architecture specific wrappers in .ll for the 32-bit cases where __atomic requires libatomic.

On the flipside, the only architecture where I can definitely recall that being the case was MIPS... So...

zvookin:
If it is just alignment, perhaps adding alignment qualifiers will fixit.

steven-johnson:
You two may wish to take a look at and/or comment on https://reviews.llvm.org/D143813; this isn't (yet) a breaking issue for us, but it's clear that the LLVM folks really don't want people to use __sync anymore. If the feedback above still holds, we should let them know; if now, we should update our code.

steven-johnson:
> llvm can't prove alignment

No, but we could (and probably should) add `__attribute(aligned(N))` to explicitly tell it that we will ensure the right alignment for the relevant structs.

abadams:
I really don't like the failure mode where our runtime gets a lot slower without us noticing because llvm's alignment proof starts failing with some new llvm version and it inserts library calls instead of atomic instructions. I'd prefer a noisy compilation failure to a silent library call insertion, which is what __sync is currently giving us, apparently. Let's try inserting alignment promises and seeing if it silences the error.

abadams:
Actually I don't think that will work. The error isn't about alignment, it's because llvm doesn't have sufficient information on the platform at that stage of compilation to know that a lock-free operation is even possible.

zvookin:
I would expect asserting natural alignment in all the right places would help as it should cover the most conservative architecture. If not, maybe we can push back to get that to work somehow. I'm also guessing part of the issue here is using __sync on 64-bit values for 32-bit architectures, which may just be that we aren't forcing 8-byte alignment or it may be that the instructions are not supported.

Generally the fallback to libatomic is not silent because none of our tests, etc. link against it. But yeah, the entire premise of libatomic is somewhat questionable and a way to fail rather than fallback would be productive.

steven-johnson:
I've captured the discussion points in https://github.com/halide/Halide/issues/7431 -- I'm going to close this PR, let's move any further discussion to the Issue.

PR #7428, opened 2023-03-16
Mark the Halide pipeline structs as aligned(8)

steven-johnson:
When compiling for 32-bit, LLVM assumes that these structs are only 4-aligned (since alignof(uint64_t) == 4 for x86-32), which means some atomic operations on these structs may require library calls. Since these structs are always malloc'ed, and malloc on all our platforms will return an 8-aligned pointer, we can improve this by telling Clang that the struct will always be at least 8-aligned.


steven-johnson:
Monday review ping

abadams (approved):

PR #7429, opened 2023-03-16
Use existing Halide Runtime atomic wrappers everywhere in the runtime

steven-johnson:
We currently have a set of wrappers around the __atomic/__sync primitives used by our threading model; for various reasons, we desire to use the (deprecated) __sync primitives for 32-bit builds instead of the __atomic primitives (see https://github.com/halide/Halide/pull/7427 for some discussion).

This PR attempts to use these abstractions everywhere else in our runtime, for consistency, so that 64-bit builds consistent use the __atomic primitives for (e.g.) profiling and tracing too. This meant:
- Splitting the wrappers into a new header (runtime_atomics.h), as synchronization_common.h can't be included into arbitrary other files, for valid reasons
- Adding wrappers for the necessary primitives
- Modifying the code elsewhere in the runtime

Where new wrappers were needed, I generally defaulted to assuming that SEQ_CST was the safest memory order to use.

Not entirely sure if this is a worthwhile goal or not, but putting this out there for consideration and discussion.

### DIFF
--- a/src/runtime/runtime_atomics.h
+++ b/src/runtime/runtime_atomics.h
@@ -1,0 +1,270 @@
+#ifndef HALIDE_RUNTIME_RUNTIME_ATOMICS_H
+#define HALIDE_RUNTIME_RUNTIME_ATOMICS_H
+
+#include "HalideRuntime.h"
+
+namespace Halide {
+namespace Runtime {
+namespace Internal {
+namespace Synchronization {
+
+namespace {
+
+// TODO: most of these wrappers should do the remove_volatile for secondary arguments;
+// I've only put it in place for the locations necessary at this time.
+template<class T>
+struct remove_volatile { typedef T type; };
+template<class T>
+struct remove_volatile<volatile T> { typedef T type; };
+
+#ifdef BITS_32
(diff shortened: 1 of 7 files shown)


abadams (approved):

src/runtime/runtime_atomics.h
```diff
@@ -0,0 +1,270 @@
+#ifndef HALIDE_RUNTIME_RUNTIME_ATOMICS_H
+#define HALIDE_RUNTIME_RUNTIME_ATOMICS_H
+
```
zvookin, line 3: It is likely worth capturing the issue re: __atomic possibly leading to libatomic references and that being the reason for this support. (I.e. that in order for __atomic to generate hardware instructions rather than calls to libatomic routines, all operations promised by __atomic for the datatype in question must be supported in hardware)

steven-johnson: As in, open an Issue and reference it in this file? Will do.

steven-johnson: PTAL

zvookin (approved):

PR #7432, opened 2023-03-17
Backport xtensa_codegen to main (v2)

steven-johnson:


steven-johnson:
Adding @rootjalex as reviewer specifically for XtensaOptimize, due to its similarity to HexagonOptimize

rootjalex:
> Adding @rootjalex as reviewer specifically for XtensaOptimize, due to its similarity to HexagonOptimize

I am happy to take a look but am traveling this week, might be a few days before I can review this, sorry about that.

src/CodeGen_Xtensa.cpp
```diff
@@ -0,0 +1,1593 @@
...
+            }
+        } else {
+            rhs = id_value;
+        }
+    }
+
+    print_assignment(vector_type, rhs);
+}
+
+void CodeGen_Xtensa::visit(const LE *op) {
```
rootjalex, line 637: This method, the `LT` visitor, the `GE` visitor, and the `GT` visitor all seem to share the same logic, and could probably benefit from a single method that is parameterized by an op string. I think the `Min` and `Max` visitor might be the same exact logic as well, but that's just me squinting at it.

rootjalex: Probably the `EQ` visitor too.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+    return t.is_float() && (t.bits() == 16) && (t.lanes() == vector_size);
+}
+
+template<>
+bool is_native_xtensa_vector<float>(const Type &t, const Target &target) {
+    int vector_size = target.natural_vector_size<float>();
+    return t.is_float() && (t.bits() == 32) && (t.lanes() == vector_size);
+}
+
+bool is_native_vector_type(const Type &t, const Target &target) {
```
rootjalex, line 84: Is there an explanation as to why 8 and 24 bit are full lanes, 16 and 48 are half lanes, and 32 bits are a fourth of the 8 bit lanes? This behavior makes very little sense to me, but I am not familiar with this target.

rootjalex: e.g. I don't understand why this method isn't as simple as `is_double_native_vector_type` below.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+        return "_i32";
+    } else if (t.is_uint() && (t.bits() == 32)) {
+        return "_u32";
+    } else if (t.is_float() && (t.bits() == 32)) {
+        return "_f32";
+    } else if (t.is_float() && (t.bits() == 16)) {
+        return "_f16";
+    }
+
+    return "";
```
rootjalex, line 156: Should this be an error?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+}
+
+struct Pattern {
+    enum Flags {
```
rootjalex, line 160: It's not clear what order these flags are supposed to be applied in (i.e. does `NarrowOp0` happen before `SwapOps01` or after? I assume that earlier flags should be processed earliest, but that's not explicitly stated.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+    } else if (t.is_float() && (t.bits() == 32)) {
+        return "_f32";
+    } else if (t.is_float() && (t.bits() == 16)) {
+        return "_f16";
+    }
+
+    return "";
+}
+
+struct Pattern {
```
rootjalex, line 159: This is a bit more of a comment on Halide's codegen in general, rather than this PR, but this rather extreme set of options suggests to me that we really need to invest in transforming this code (and other instruction selection code) into proper term-to-term rewriting systems, so we don't need to enumerate every possible substitution and rewrite we want.

src/XtensaOptimize.cpp
rootjalex: After looking at the process flags implementation, it seems my intuition was wrong. Thi should probably be documented here.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+                // Predicated addition
+                // NOTE(vksnk): patterns below are for predicated instructions and look like they may
+                // be more efficient, but they are not according to simulator. We will need to check with
+                // Cadence about this.
+                // {"halide_xtensa_pred_add_i8", wild_i8x + select(wild_u1x, wild_i8x, wild_i8x)},
+                // {"halide_xtensa_pred_add_i16", wild_i16x + select(wild_u1x, wild_i16x, wild_i16x)},
+                // {"halide_xtensa_pred_add_i32", wild_i32x + select(wild_u1x, wild_i32x, wild_i32x)},
+
+                {"halide_xtensa_qqqq", slice(wild_i24x256, 0, 1, 128) + slice(wild_i24x256, 128, 1, 128), Pattern::SameOp01},
+                {"halide_xtensa_yyyy", (call("halide_xtensa_xxxx", wild_i24x64, {wild_i24x64, wild_i24x128}) + slice(wild_i24x128, 64, 1, 64)), Pattern::SameOp12},
```
rootjalex, line 710: Why are some methods implemented via the static call methods defined above, and some implemented via `call()`? imo the former is more readable, and the latter should be replaced by the former if possible.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+            }
+        }
+
+        return IRGraphMutator::visit(op);
+    }
+
+    Expr visit(const Div *op) override {
+        if (op->type.is_vector()) {
+            Expr div = op;
+            static const std::vector<Pattern> divs = {
```
rootjalex, line 838: Are there ever cases where these rules trigger? `find_intrinsics` should convert divisions by powers of two into shifts already.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+        //         }
+        //     }
+        // }
+
+        static const std::vector<Pattern> casts = {
+            // Narrowing multiply with shift.
+            // {"halide_xtensa_sat_mul_with_shift_i32", i32(wild_i64x * wild_i64x / wild_i64), Pattern::NarrowOp0 | Pattern::NarrowUnsignedOp1 | Pattern::ExactLog2Op2},
+
+            // Casts from bool.
+            {"halide_xtensa_convert_u1_to_i16", i16(i8(wild_u1x))},
```
rootjalex, line 925: Is it possible to see `i16(u8(wild_u1x))` or `i16(wild_u1x)`?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+
+            // Narrowing with shifting.
+            {"halide_xtensa_narrow_i48_with_shift_i16", i16(i32(wild_i48x) >> wild_i32)},
+            {"halide_xtensa_narrow_i48_with_shift_i16", i16(i32(wild_i48x) / wild_i32), Pattern::ExactLog2Op1},
```
rootjalex, line 929: Same comment on these as the div visitor

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+            {"halide_xtensa_abs_f32", abs(wild_f32x)},
+
+            {"halide_xtensa_avg_u8", halving_add(wild_u8x, wild_u8x)},
+            {"halide_xtensa_avg_i8", halving_add(wild_i8x, wild_i8x)},
+
+            {"halide_xtensa_avg_u16", halving_add(wild_u16x, wild_u16x)},
+            {"halide_xtensa_avg_i16", halving_add(wild_i16x, wild_i16x)},
+
+            // {"halide_xtensa_avg_u32", halving_add(wild_u32x, wild_u32x)},
+            // {"halide_xtensa_avg_i32", halving_add(wild_i32x, wild_i32x)},
```
rootjalex, line 1099: Why are these commented out?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+
+            {"halide_xtensa_avg_round_u8", rounding_halving_add(wild_u8x, wild_u8x)},
+            {"halide_xtensa_avg_round_i8", rounding_halving_add(wild_i8x, wild_i8x)},
+
+            {"halide_xtensa_avg_round_u16", rounding_halving_add(wild_u16x, wild_u16x)},
+            {"halide_xtensa_avg_round_i16", rounding_halving_add(wild_i16x, wild_i16x)},
+
+            // {"halide_xtensa_avg_round_u32", rounding_halving_add(wild_u32x, wild_u32x)},
+            // {"halide_xtensa_avg_round_i32", rounding_halving_add(wild_i32x, wild_i32x)},
```
rootjalex, line 1108: Why are these commented out?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+
+            {"halide_xtensa_widen_right_mul_u64", widen_right_mul(wild_u32x, wild_u16x), Pattern::AccumulatorOutput64},
+
+            {"halide_xtensa_widen_zzzzz", halide_xtensa_widen_mul_u24(wild_u8x256, wild_u8)},
+            {"halide_xtensa_widen_zzzzz", halide_xtensa_widen_mul_u24(concat({wild_u8x64, wild_u8x64, wild_u8x64, wild_u8x64}), repeat_each_element(wild_u8x4, 64))},
+            {"halide_xtensa_widen_zzzzz", halide_xtensa_widen_mul_u24(repeat_each_element(wild_u8x4, 64), wild_u8x256), Pattern::SwapOps01},
+
+            // {"halide_xtensa_rounding_mul_shift_right_i8", rounding_mul_shift_right(wild_i8x, wild_i8x, bc(wild_u8))},
+            // {"halide_xtensa_rounding_mul_shift_right_i16", rounding_mul_shift_right(wild_i16x, wild_i16x, bc(wild_u16))},
+            // {"halide_xtensa_rounding_mul_shift_right_i32", rounding_mul_shift_right(wild_i32x, wild_i32x, bc(wild_u32))},
```
rootjalex, line 1135: And these. Commented out patterns should have explanatory comments.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+            {"halide_xtensa_sat_narrow_i8", i8_sat(wild_i16x)},
+            {"halide_xtensa_sat_narrow_u8", u8_sat(wild_i16x)},
+            {"halide_xtensa_sat_narrow_i16", i16_sat(wild_i32x)},
+
+            {"halide_xtensa_rounding_shift_right_i8", rounding_shift_right(wild_i8x, bc(wild_u8))},
+            // {"halide_xtensa_rounding_shift_right_u8", rounding_shift_right(wild_u8x, bc(wild_u8))},
+            {"halide_xtensa_rounding_shift_right_i16", rounding_shift_right(wild_i16x, bc(wild_u16))},
+            // {"halide_xtensa_rounding_shift_right_u16", rounding_shift_right(wild_u16x, bc(wild_u16))},
+            {"halide_xtensa_rounding_shift_right_i32", rounding_shift_right(wild_i32x, bc(wild_u32))},
+            // {"halide_xtensa_rounding_shift_right_u32", rounding_shift_right(wild_u32x, bc(wild_u32))},
```
rootjalex, line 1170: Same comment here about commented-out patterns

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+    } else {
+        return bounds.max - bounds.min;
+    }
+}
+
+// NOTE(vksnk): this is borrowed from HexagonOptimize.cpp, so
+// eventually need to generalize and share across two places.
+// Replace indirect loads with dynamic_shuffle intrinsics where
+// possible.
+class OptimizeShuffles : public IRMutator {
```
rootjalex, line 1451: If this is directly copied without change, it should probably be moved into a separate file and de-duplicated.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+            Expr r = Select::make(padded_cond, padded_t, padded_f);
+
+            return slice(r, op->type, lanes);
+        }
+
+        return IRMutator::visit(op);
+    }
+
+    // NOTE(vksnk): not very clear if it's a good idea to slice loads/stores.
+    // Expr visit(const Load* op) override {
```
rootjalex, line 1670: This seems like a lot of commented-out code to check into main. Should probably be removed?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+                                           Call::PureExtern);
+                Expr r = Op::make(sliced_a, sliced_b);
+                concat_args.push_back(std::move(r));
+            }
+            return Call::make(op->type,
+                              "halide_xtensa_concat_from_native",
+                              concat_args, Call::PureExtern);
+        }
+
+        // TODO(vksnk): bool handling is maybe sketchy.
```
rootjalex, line 1835: What does this mean?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+            Expr r = Call::make(op->type.with_lanes(width_to_extend), op->name, args, op->call_type);
+
+            return slice(r, op->type, lanes);
+        }
+
+        return IRMutator::visit(op);
+    }
+
+    Expr visit(const VectorReduce *op) override {
+        // TODO(vksnk): Factor it out.
```
rootjalex, line 2052: Factor what out? I assume this entire method, which looks generally useful for most targets.

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+    // NOTE(vksnk): CSE seemed to break loop carry
+    // s = common_subexpression_elimination(s);
+
+    // Use at most 16 vector registers for carrying values.
+    // NOTE(vksnk): loop_carry seems to be a little finicky right now
+    // but looks like something we'd definitely want to have, so
+    // need to figure out where it goes wrong.
+    s = loop_carry(s, 16);
+    s = simplify(s);
+    for (int ix = 0; ix < 10; ix++) {
```
rootjalex, line 2306: How was this constant (and the other constants below) decided on for the number of iterations performed?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2331 @@
...
+    for (int ix = 0; ix < 3; ix++) {
+        s = SimplifySliceConcat().mutate(s);
+    }
+
+    // Extra run to replace cast + concat, etc.
+    for (int ix = 0; ix < 10; ix++) {
+        s = MatchXtensaPatterns(target).mutate(s);
+    }
+    // NOTE(vksnk): looks like we shouldn't do simplification in the end.
+    // s = simplify(common_subexpression_elimination(s));
```
rootjalex, line 2322: Why not? And should this code be removed then?

src/XtensaOptimize.cpp
```diff
@@ -0,0 +1,2208 @@
...
+            binop = Max::make;
+            break;
+        case VectorReduce::And:
+            binop = And::make;
+            break;
+        case VectorReduce::Or:
+            binop = Or::make;
+            break;
+        case VectorReduce::SaturatingAdd:
+            binop = saturating_add;
```
abadams, line 1954: Change this to Halide::saturating_add (the one in the internal namespace is the deprecated wrapper)

src/XtensaOptimize.cpp
```diff
@@ -132,6 +97,16 @@ Type get_native_xtensa_vector(const Type &t, const Target &target) {
...
 }
 
+bool is_native_vector_type(const Type &t, const Target &target) {
+    Type native_vector_type = get_native_xtensa_vector(t, target);
+    return t == native_vector_type;
+}
+
+bool is_double_native_vector_type(const Type &t, const Target &target) {
+    Type native_vector_type = get_native_xtensa_vector(t, target);
+    return t == native_vector_type.with_lanes(native_vector_type.lanes());
```
rootjalex, line 107: Should there be a `* 2` inside the `with_lanes()` call?

vksnk: Thanks, fixed.

PR #7438, opened 2023-03-18
[xtensa] Added uint32 vector mul and fixed i16 -> i32_x2 vector conversion

Aelphy:


src/CodeGen_Xtensa_vectors.template.cpp
```diff
@@ -2205,13 +2218,23 @@ HALIDE_ALWAYS_INLINE native_vector_i32_x4 convert<native_vector_i32_x4, native_v
...
+        IVP_CVT32S2NX24LL(wide),
+        IVP_CVT32S2NX24LH(wide),
+        IVP_CVT32S2NX24HL(wide),
+        IVP_CVT32S2NX24HH(wide));
+}
+
 template<>
 HALIDE_ALWAYS_INLINE native_vector_i32_x2 convert<native_vector_i32_x2, native_vector_i16>(const native_vector_i16 &src) {
+    xb_vec2Nx24 wide = IVP_CVT24S2NX16(0, src);
     return native_vector_i32_x2(native_vector_i32_x2::from_native_vector,
```
vksnk, line 2235: Oh, yeah, that was incorrect before for sure, because it needs to have sign extension. In my experiment a while ago using a shuffle/select was more efficient than going through the wide vector. Maybe we can use the old approach (which is currently incorrect), but add a sign extension as in halide_xtensa_convert_i16_low_i32/halide_xtensa_convert_i16_high_i32?

src/CodeGen_Xtensa.cpp
```diff
@@ -222,6 +222,10 @@ void CodeGen_Xtensa::visit(const Mul *op) {
             string sa = print_expr(op->a);
             string sb = print_expr(op->b);
             print_assignment(op->type, "IVP_PACKLN_2X64W(IVP_MULN_2X32(" + sa + ", " + sb + "))");
+        } else if (is_native_xtensa_vector<uint32_t>(op->type, target)) {
```
vksnk, line 225: How is it different from the branch above? Maybe just use "or" there to handle the new case as well?

vksnk (approved):

PR #7440, opened 2023-03-20
Use pmaddubsw for non-RDom horizontal widening adds

abadams:
We were rewriting the following into a pmaddubsw:

```
RDom r(0, 2);
f(x) += cast<uint16_t>(g(2*x + r));
f.update().atomic().vectorize(r).vectorize(x, 8);
```

But not the equivalent unrolled version, which is more likely to be written:
```
f(x) = cast<uint16_t>(g(2*x)) + g(2*x + 1)
```

This pattern shows up in downsampling code.

This PR fixes it, and also adds support for doing horizontal widening adds of i16 to i32 using pmaddwd, and adds some missing avx512 variants of horizontal adds using pmaddubsw.

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -727,6 +727,7 @@ protected:
         }
 
         const int bits = op->type.bits();
+        const int lanes = op->type.lanes();
         const auto is_x_same_int = op->type.is_int() && is_int(x, bits);
         const auto is_x_same_uint = op->type.is_uint() && is_uint(x, bits);
         const auto is_x_same_int_or_uint = is_x_same_int || is_x_same_uint;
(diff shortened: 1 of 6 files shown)


src/FindIntrinsics.cpp
```diff
@@ -781,6 +782,13 @@ class FindIntrinsics : public IRMutator {
             rewrite(saturating_cast(op->type, rounding_shift_right(widening_mul(x, y), z)),
                     rounding_mul_shift_right(x, y, cast(unsigned_type, z)),
                     is_x_same_int_or_uint && x_y_same_sign && is_uint(z)) ||
+
+            // Rewrite combinations of deinterleaves into horizontal ops
+            rewrite(widening_add(slice(x, 0, 2, lanes), slice(x, 1, 2, lanes)),
+                    h_add(cast(op->type.with_lanes(lanes * 2), x), lanes)) ||
+            rewrite(widening_add(slice(x, 1, 2, lanes), slice(x, 0, 2, lanes)),
+                    h_add(cast(op->type.with_lanes(lanes * 2), x), lanes)) ||
```
rootjalex, line 790: I think it might be useful to have a mutator that attempts to perform this operation more generally, for other operations as well?

PR #7441, opened 2023-03-21
[xtensa] Returned old free_helper

Aelphy:
due to poor performance of xtensa compiler with the new one

### DIFF
--- a/src/CodeGen_C.h
+++ b/src/CodeGen_C.h
@@ -300,7 +300,8 @@ protected:
     void emit_constexpr_function_info(const std::string &function_name,
                                       const std::vector<LoweredArgument> &args,
                                       const MetadataNameMap &metadata_name_map);
-    void emit_halide_free_helper(const std::string &alloc_name, const std::string &free_function);
+    /** Xtensa compiler produces sub-optimal results with this free_helper */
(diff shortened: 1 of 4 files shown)


src/CodeGen_C.h
```diff
@@ -300,7 +300,8 @@ class CodeGen_C : public IRPrinter {
     void emit_constexpr_function_info(const std::string &function_name,
                                       const std::vector<LoweredArgument> &args,
                                       const MetadataNameMap &metadata_name_map);
-    void emit_halide_free_helper(const std::string &alloc_name, const std::string &free_function);
+    /** Xtensa compiler produces sub-optimal results with this free_helper */
```
vksnk, line 303: This comment is specific to Xtensa, so shouldn't be in CodeGen_C

vksnk (approved):

steven-johnson:
> due to poor performance of xtensa compiler with the new one

For future reference, can you be more specific about "Poor Performance"? Compile time, codegen, ...?

steven-johnson:
Also, we should report this as a bug to Xtensa ASAP -- the templated version is extremely basic C++ and there's no way that any modern compiler should produce *worse* code with it (it should be strictly better-or-no-worse)

PR #7443, opened 2023-03-22
Fix bugs in PyTorch codegen.

Yongqi-Zhuo:
There are several problems with the current PyTorch CodeGen.
First, the emitted `.pytorch.h` file incorrectly exposes functions with internal linkage (which is inconsistent with the C codegen), and leads to compile errors when actually compiling the header file. I added a check for internal linkage to fix this.
Next is the problems when using CUDA. Currently when using the generated `.pytorch.h` file, the linker will complain that there is undefined reference to a symbol related to `halide_cuda_device_interface` (the name is mangled when reported by the linker). This is because the forward declaration was not in a `extern "C"` block.
Moreover, there is a really subtle problem when the generated pipeline is invoked by PyTorch (which I actually have described in Matrix). The Halide pipeline runs in a different CUDA stream than that of PyTorch kernels. Afterwards I discovered that `HalidePyTorchCudaHelpers.h` was not included as `apps/HelloPyTorch/setup.py` did, and weakly-linked CUDA handles are not overridden. I fix this by simply including this header in the codegen.

### DIFF
--- a/src/CodeGen_PyTorch.cpp
+++ b/src/CodeGen_PyTorch.cpp
@@ -25,6 +25,7 @@ void CodeGen_PyTorch::compile(const Module &module) {
                           "Please add \"-user_context\" to the generator's target options.\n";
         }
         stream << "#include \"ATen/cuda/CUDAContext.h\"\n";
+        stream << "#include \"HalidePyTorchCudaHelpers.h\"\n";
     }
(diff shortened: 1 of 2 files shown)


steven-johnson:
Thanks -- looks like we haven't been testing this code on buildbots recently, added https://github.com/halide/Halide/issues/7448 in hopes someone will address this.

PR #7446, opened 2023-03-24
Promote fixed-point intrinsics out of the Internal namespace

abadams:
This PR promotes the fixed-point intrinsics out of Halide::Internal:: into Halide::. These are suitable for use in production, and I know of a bunch of production use that just refers to the Halide::Internal:: versions, which is not ideal. In one of our dev meetings we agreed we should just promote them, and I'm finally getting around to doing it.

The PR also adds deprecated wrappers for them in the Internal namespace so that we don't break any existing code.

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -2075,7 +2075,7 @@ Expr CodeGen_C::scalarize_vector_reduce(const VectorReduce *op) {
         binop = Or::make;
         break;
     case VectorReduce::SaturatingAdd:
-        binop = saturating_add;
+        binop = Halide::saturating_add;
         break;
     }
 
--- a/src/CodeGen_LLVM.cpp
+++ b/src/CodeGen_LLVM.cpp
@@ -4188,7 +4188,7 @@ void CodeGen_LLVM::codegen_vector_reduce(const VectorReduce *op, const Expr &ini
         binop = Or::make;
         break;
     case VectorReduce::SaturatingAdd:
-        binop = saturating_add;
+        binop = Halide::saturating_add;
         break;
     }
 
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -14,6 +14,8 @@
 #include "Util.h"
 #include "Var.h"
 
+using namespace Halide::Internal;
+
 namespace Halide {
 
 // Evaluate a float polynomial efficiently, taking instruction latency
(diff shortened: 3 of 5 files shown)


rootjalex (approved):

abadams:
Failure unrelated.

steven-johnson:
This causes (legit) compilation problems in google3. The main issue is that something like

```
    Var x, y, c;

    Func f("f");
    Func g("g");
    f(x, y, c) = (x + y + c);
    g(x, y, c) = (x + y);

    Func h("h");
    h(x, y, c) = Halide::Internal::halving_add(f(x, y, c), g(x, y, c));
```

fails because `f(x, y, c)` converts to a FuncRef (via Func.h:1254)... and while a FuncRef is supposed to implicitly convert to an Expr, it apparently doesn't here, because (I'm guessing) Func.h hasn't been included yet.

We can't (easily) change Func.h to be included by IROperator.h -- trying to do so reveals a large hairball of circular includes that may be fixable but not easily.

OK, so how about we just add a ctor to PseudoExpr that accepts a FuncRef? We can do that, as long as the body of it is in IROperator.cpp (since we can only forward-declare FuncRef in IROperator.h), but... then we get ambiguous calls to these functions in some cases, e.g.

```
/Users/srj/GitHub/Halide/src/InlineReductions.cpp:160:22: error: call to 'saturating_add' is ambiguous
    f(v.free_vars) = saturating_add(f(v.free_vars), e);
```

I'm honestly not sure of the right fix here. (I'm tempted to suggest that we move the "Internal" wrappers behind an #ifdef and force people to opt in to the old-and-deprecated functions, but that's pretty kludgey.)

I'm going to make some more effort to unwind the hairball of circular includes as a way to fix this, but I am not hopeful, so thoughts are welcome here.

steven-johnson:
Actually, a thought did occur to me that might be a clean fix, but you might not like it:
- Convert all internal-to-Halide usage of these functions to be fully namespaced, eg `::Halide::halving_add`
- Remove PseudoExpr entirely and use Expr; this theoretically opens the possibility of ambiguous functions, of course, but all our internal code will be safe, thanks to full-namespacing. Well-behaved external code will be fine since they should be using explicit namespacing. If any external code that is `using namespace` both `Halide` and `Halide::Internal`, it will break spectacularly, of course, but if external code is doing that, I have little sympathy for them.
- After the Internal versions are eventually removed, we can clean the `::Halide::` prefix from our internal usage.

abadams:
How about a templated constructor on PseudoExpr that takes anything and casts it to Expr?

abadams:
I believe that ambiguous call in InlineReductions.cpp needs to be resolved using Halide:: anyway. I probably didn't hit that one because I didn't have the FuncRef version.

PR #7447, opened 2023-03-24
Moves OptimizeShuffles pass into separate file

vksnk:


steven-johnson (approved):
LGTM -- does this have any code tests (or even any Hexagon tests)? If not, we should probably add some, or at least open a tracking issue that notes that we need some.

vksnk:
yes, simd_op_check does check that ``` vlut ``` is generated for clamped indices.

steven-johnson:
Ready to land

PR #7453, opened 2023-03-28
Fix for top-of-tree LLVM

steven-johnson:


alexreinking (approved):

steven-johnson:
(Note, this is pending testing -- this looks right based on internal-to-google tests but I'm waiting to get everything rebuilt locally)

PR #7458, opened 2023-03-29
Fix PseudoExpr for FuncRef (followup to #7446)

steven-johnson:


abadams (approved):

apps/hannk/halide/pool_generator.cpp
```diff
@@ -68,7 +68,7 @@ class AveragePool : public Generator<AveragePool> {
         Expr inv_filter_count =
             u16_sat(((2 << log2_numerator) + filter_count) / (2 * filter_count));
         Expr average =
-            rounding_mul_shift_right(sum(c, x, y, b), inv_filter_count, log2_numerator);
+            ::Halide::rounding_mul_shift_right(sum(c, x, y, b), inv_filter_count, log2_numerator);
```
abadams, line 71: How come this one is ambiguous? It shouldn't have access to the internal namespace here, unless there's still a using hiding in a header somewhere.

steven-johnson: Not clear: 

```
/Users/srj/GitHub/Halide/apps/hannk/halide/pool_generator.cpp:71:13: error: call to 'rounding_mul_shift_right' is ambiguous
            rounding_mul_shift_right(sum(c, x, y, b), inv_filter_count, log2_numerator);
            ^~~~~~~~~~~~~~~~~~~~~~~~
/Users/srj/halide-16-install/include/Halide.h:14169:6: note: candidate function
Expr rounding_mul_shift_right(Expr a, Expr b, int q);
     ^
/Users/srj/halide-16-install/include/Halide.h:14277:13: note: candidate function
inline Expr rounding_mul_shift_right(PseudoExpr a, const Expr &b, int q) {
            ^
```

steven-johnson: Searching in all .h files doesn't show any lurkers that are `using` anything relevant. But this is in the same category as the one in InlineReductions.cpp that you said to just fix (ie, uses a FuncRef).

steven-johnson: (I'm going to go ahead and land this as-is since it is still strictly better than before)

PR #7460, opened 2023-03-29
Revert #7446, #7457, #7458

steven-johnson:
Despite best efforts, these changes break existing downstream code; this temorarily reverts the changes to allow downstream merges. Once we find a better solution we should reapply these.


steven-johnson:
Closing in favor of #7461

PR #7463, opened 2023-03-31
Minor improvements to Xtensa codegen

steven-johnson:
Harvesting some minor code improvements from an experiment that didn't get finished:
- Move all the `is_native_xtensa_vector()` and related code from XtensaOptimize into CodegenXtensa, which is the only caller
- Make these all member methods so they can just use the `get_target()` member, which was the only target they ever wanted
- Simplify the implementation of them a bit, especially `is_native_xtensa_vector()`, which should still be equivalent but terser and smaller codegen
- Use `halide_type_t` instead of `Halide::Type` in many places, since the former is smaller (exactly 32 bits) and is easier to hash; similarly, use `std::unordered_set` for the lookups we need. (Note that Halide::Type implicitly converts to halide_type_t at no cost.)


src/runtime/mini_webgpu.h
```diff
@@ -31,21 +31,21 @@
 #define WEBGPU_H_
 
 #if defined(WGPU_SHARED_LIBRARY)
-#    if defined(_WIN32)
```
vksnk, line 34: This is unrelated formatting change (ok to keep here, just wanted to make sure)

vksnk (approved):

src/runtime/mini_webgpu.h
steven-johnson: No, I'll revert, for some reason my local clang-format isn't reading the .clang-format file correctly

PR #7464, opened 2023-03-31
When doing x86 check for lower_lerp(), check features

steven-johnson:
We only want to do the specific-to-x86-trick there if we are compiling to CPU code on x86... if x86 is the host, but we are compiling to GPU, DSP, etc., we are more likely to want the other case.

(This actually mattered on Xtensa, where we made the incorrect selection due to an arbitrary choice of Target string, and the x86-adapted code choked the Xtensa compiler, while the 'normal' code didn't due to good fixed-point-op optimizations.)

(Related note: Xtensa should probably be its own Target::Arch, rather than a Feature flag...)

### DIFF
--- a/src/Lerp.cpp
+++ b/src/Lerp.cpp
@@ -143,11 +143,24 @@ Expr lower_lerp(Type final_type, Expr zero_val, Expr one_val, const Expr &weight
                 // ((x + 128) / 256 + x + 128) / 256. Note that
                 // overflow is impossible here because the most our
                 // prod_sum can be is 255^2.
-                if (target.arch == Target::X86) {
-                    // On x86 we have no rounding shifts but we do
-                    // have a multiply-keep-high-half. So it's
-                    // actually one instruction cheaper to do the
-                    // division directly.
+                //
+                // On x86 we have no rounding shifts but we do
+                // have a multiply-keep-high-half. So it's
+                // actually one instruction cheaper to do the
+                // division directly.
+                //
+                // Note, however, that it isn't enough to just check for
+                // target.arch == Target::X86, since we could be compiling
+                // for a GPU, DSP, or other similar thing... hence the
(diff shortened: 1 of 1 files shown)


src/Lerp.cpp
```diff
@@ -143,11 +143,24 @@ Expr lower_lerp(Type final_type, Expr zero_val, Expr one_val, const Expr &weight
                 // ((x + 128) / 256 + x + 128) / 256. Note that
                 // overflow is impossible here because the most our
                 // prod_sum can be is 255^2.
-                if (target.arch == Target::X86) {
-                    // On x86 we have no rounding shifts but we do
-                    // have a multiply-keep-high-half. So it's
-                    // actually one instruction cheaper to do the
-                    // division directly.
+                //
```
abadams, line 146: This isn't going to handle mixed host/device code. I guess the Target argument isn't quite and the interface or the lowering needs a small rethink.

steven-johnson: > This isn't going to handle mixed host/device code. 

True, but in that case, usually the device code is much more important and the host code can afford slight suboptimal bits.

> I guess the Target argument isn't quite right and the interface or the lowering needs a small rethink.

Yeah, that's definitely true. In hindsight, perhaps having GPU code be features hanging off the host is problematic for cases like this, but that's a nontrivial rewrite, to say the least. 

(Xtensa should almost certainly be its own arch though; I am going to prep a PR to propose that)

abadams: Actually I think when a hexagon kernel is being compiled, the target arch is Hexagon, due to the way offload works (the submodule gets its own target). We should probably have GPU targets do something similar.

abadams: (so that target.arch == Target::X86 really really means we're compiling code for x86 right now)

PR #7466, opened 2023-03-31
Remove unused 'external_buffers' member in Codegen_Xtensa

steven-johnson:


src/CodeGen_Xtensa.cpp
```diff
@@ -786,9 +786,6 @@ void CodeGen_Xtensa::visit(const Load *op) {
             native_lanes = bytes_in_vector / 2;
         }
         bool is_aligned_load = (op->alignment.modulus % native_lanes == 0) && (op->alignment.remainder % native_lanes == 0);
-        if (external_buffers.count(op->name) > 0) {
```
vksnk, line 789: Ugh, this seems like it was a legit bug. I think ``` if (external_buffers.count(op->name) > 0) ``` should be replaced by ``` if (op->param.defined()) ``` .

src/CodeGen_Xtensa.cpp
```diff
@@ -811,7 +808,7 @@ void CodeGen_Xtensa::visit(const Load *op) {
         // } else {
         string id_index = print_expr(op->index);
         // Is not allocated on the heap and is not a buffer
-        bool is_tcm = !(heap_allocations.contains(name) || external_buffers.count(op->name) > 0);
```
vksnk, line 814: Same here and below.

src/CodeGen_Xtensa.cpp
steven-johnson: yikes, ok, lets hold off on this PR -- I'll let you choose how to clean this up, you are much closer to this code than I

steven-johnson:
Did we ever decide if this needed attention?

PR #7470, opened 2023-03-31
Avoiding having shader kernels depend on host arch/os by unsetting it

abadams:
Alternative to #7464 

Currently how lerp is lowered inside cuda kernels depends on the host architecture. This was unintentional. This PR nukes the host os and arch when compiling shader kernels so that this sort of accidental dependence doesn't occur in future.


steven-johnson (approved):
Will this address Hexagon, Xtensa, etc?

abadams:
Hexagon wasn't a problem because it already used a correct arch and os for the offloaded code. Xtensa is fixed by #7467

PR #7483, opened 2023-04-06
Clean up various autoscheduler tool issues

steven-johnson:
- In CMake, not all of the tooling for Anderson2021 was included in packaging; now it is.
- Since the featurization_to_sample and get_host_target tools are 100% identical across all autoschedulers, they now get build only once, and don't get the `adams2019_` (etc) prefix they formerly did.
- conversely, the autotune_loop.sh and weightsdir_to_weightsfile tools now *do* get autoscheduler-specific prefixes, because they aren't interchangeable.
- Fixes to various scripts etc to use the correct tool names.
- When the Anderson2021 autoscheduler was added, we tried to factor out common coding and tooling, but went a bit too far: the weightsdir_to_weightsfile source was moved into common/, and while it and the related Weights source were identical, they needed to include *different* Featurization.h files, and so basically relied on the build rules building it twice, with a different include path each time. IMHO this is unreasonably fragile and weird, so I moved those sources back into the folders in question, as there isn't any compelling reason to keep these sources exactly in sync anyway.
- Same story for the test_function_dag test.

Note that I started to add support for Anderson2021 to the main Makefile, but it became painful to do -- it works fine for CMake, so I will leave adding Make support for someone who wants to use it there. (Honestly, the Makefile for it is still kind of a mess even after some cleanup; I'm tempted to delete it and just declare that that one is CMake-only...)


src/autoschedulers/adams2019/Makefile
```diff
@@ -36,23 +36,23 @@ $(BIN)/baseline_weights.o: $(BIN)/baseline_weights.cpp
 	$(CXX) -c $< -o $@
 
 AUTOSCHED_COST_MODEL_LIBS=\
-$(BIN)/cost_model/adams2019_cost_model.a \
-$(BIN)/cost_model/adams2019_train_cost_model.a \
+$(BIN)/adams2019_cost_model/adams2019_cost_model.a \
```
abadams, line 39: Why the adams2019 prefix on the cost model directory and generator? This file should be private to this autoscheduler, so it shouldn't need namespacing,

test/autoschedulers/anderson2021/test.cpp
```diff
@@ -14,7 +14,7 @@ int main(int argc, char **argv) {
     AutoschedulerParams params = {"Anderson2021", {{"parallelism", std::to_string(hardware_parallelism)}}};
 
     // Use a fixed target for the analysis to get consistent results from this test.
-    Target target("x86-64-linux-sse41-avx-avx2-cuda");
+    Target target("x86-64-linux-sse41-avx-avx2-cuda-debug");
```
abadams, line 17: Unintentional? This isn't going to change anything in the current test, but if the test is changed to actually run code, it's going to invalidate any performance measurements.

steven-johnson: oops, yes, fixing

src/autoschedulers/adams2019/Makefile
```diff
@@ -52,7 +52,7 @@ $(BIN)/auto_schedule_runtime.a: $(BIN)/cost_model.generator
 
 $(BIN)/cost_model/adams2019_%.a: $(BIN)/cost_model.generator
 	@mkdir -p $(@D)
-	$^ -g $* -o $(BIN)/cost_model -f $* -n adams2019_$* target=$(HL_TARGET)-no_runtime -e stmt,static_library,h,assembly
+	$^ -g $* -o $(BIN)/adams2019_cost_model -f $* -n adams2019_$* target=$(HL_TARGET)-no_runtime -e stmt,static_library,h,assembly
```
abadams, line 55: Is this rename still valid?

steven-johnson: oops, no

steven-johnson:
Monday Morning Review Ping

abadams:
LGTM for my part, but would like Alex to approve for the cmake changes.

alexreinking (approved):
Build changes look good to me across the board.

PR #7493, opened 2023-04-10
Rename `device_copy` -> `DeviceCopy`

steven-johnson:
This is 100% stylistic, but it's bothered me for years: the `device_copy` struct doesn't follow the usual Halide style for struct types, so every time I see it, I want to read it as a function name. Let's rename it to DeviceCopy to help keep me sane?

Also, drive-by change of #define to constexpr since we have C++17 in runtime.

### DIFF
--- a/src/runtime/cuda.cpp
+++ b/src/runtime/cuda.cpp
@@ -893,7 +893,7 @@ WEAK int halide_cuda_device_malloc(void *user_context, halide_buffer_t *buf) {
 }
 
 namespace {
-WEAK int cuda_do_multidimensional_copy(void *user_context, const device_copy &c,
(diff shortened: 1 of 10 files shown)


abadams:
C-style structs in the runtime are usually snake_case. See `git grep '^struct.*{' *`

We have a few proper classes with constructors and destructors (e.g. ScopedSpinLock), and these are inconsistently CamelCase or snake_case.

steven-johnson:
> C-style structs in the runtime are usually snake_case. See `git grep '^struct.*{' src/runtime/*`

...in which case we should rename it to `device_copy_t`

steven-johnson:
Closing because apparently I'm the only one bothered by this.

PR #7494, opened 2023-04-11
fix python binding method export

zhen8838:
1. fix halide var `_0` export
2. fix ImageParam `.in` method export


steven-johnson:
Nice fixes. Should probably get backported to 15.x too next time we do a dot release.

steven-johnson (approved):
LGTM pending green.

steven-johnson:
Failure is unrelated

PR #7496, opened 2023-04-11
d3d12: enforce weak linkage

slomp:
Define all symbols with the weak linkage attribute, or as `static constexpr` when applicable. Should allow for merging runtimes together.

### DIFF
--- a/src/runtime/d3d12compute.cpp
+++ b/src/runtime/d3d12compute.cpp
@@ -73,14 +73,14 @@
 #define HALIDE_D3D12_COMMAND_LIST_TYPE D3D12_COMMAND_LIST_TYPE_DIRECT
 #endif
 
-static void d3d12_debug_dump(error &err);
+WEAK void d3d12_debug_dump();
 
 #define d3d12_panic(...)                               \
     do {                                               \
         error err(nullptr);                            \
         err << __VA_ARGS__ << "\n";                    \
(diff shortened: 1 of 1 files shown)


src/runtime/d3d12compute.cpp
```diff
@@ -73,30 +73,30 @@
...
+        d3d12_debug_dump();                            \
         err << "^^^^^ D3D12  End  Debug Dump ^^^^^\n"; \
         err << "D3D12 HALT !!!\n";                     \
     } while (0)
 
 // v trace and logging utilities for debugging v
 // ! definitely not super thread-safe stuff... !
 // in case there's no 'user_context' available in the scope of a function:
-static void *const user_context = nullptr;
+static constexpr void *const user_context = nullptr;
```
steven-johnson, line 91: changing this to constexpr is less-bad than what was here before, but may I gently suggest that this symbol is a bad idea and should be removed entirely -- it's is just a workaround for needing a `user_context` in places you don't have one. You should really remove this (in a subsequent PR) and either pass the user_context you need down to all the places it is needed, or figure out how to do without. (Or, in some unusual cases, just pass `nullptr` directly... but at least in those cases the passing of `nullptr` will be obvious, rather than obfuscated.)

steven-johnson (approved):

PR #7498, opened 2023-04-11
Pacify clang-tidy

### LINKED ISSUE #7497: Possible error in correctness_lerp
A newer version of clang-tidy reports `signed shift result (0x80000000) sets the sign bit of the shift expression's type ('int') and becomes negative` on lines 181 and 185 of `correctness/lerp.cpp`; someone should investigate.

abadams:
The code was constructing INT_MIN in a way that clang-tidy disapproved of.

Hopefully fixes #7497


steven-johnson (approved):

steven-johnson:
Looks like the arm-32-linux failure is real but in LLVM?

PR #7500, opened 2023-04-12
Add build number to Python wheel before uploading

### LINKED ISSUE #7293: PyPI "daily" updates are broken
We attempt to do "daily" uploads of Halide-for-Python to the PyPI test server, but these don't work, because apparently the test server doesn't ever let you overwrite/replace an upload. Presumably we need to append a timestamp, hash, etc?

dragly:
This change adds a build number based on GitHub Actions' `github.run_id` to the Python wheel before uploading. This should work around the issue that causes the uploads to fail currently.

Fixes #7293

### DIFF
--- a/.github/workflows/pip.yml
+++ b/.github/workflows/pip.yml
@@ -75,10 +75,13 @@ jobs:
             -DWITH_TUTORIALS=NO -DWITH_UTILS=NO -DWITH_PYTHON_STUBS=NO &&
             cmake --build build --target install
 
+      - name: Tag wheels
+        run: wheel tags --build=${{ github.run_id }} ./wheelhouse/*.whl
+
       - uses: actions/upload-artifact@v3
         with:
           name: wheels
-          path: ./wheelhouse/*.whl
+          path: ./wheelhouse/*${{ github.run_id }}*.whl
 
   pip-other:
     name: Package Halide Python bindings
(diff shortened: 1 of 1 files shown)


dragly:
I am not sure this will work since it is a bit hard to test locally. It could be that the `wheel` command is not available where I try to call it, but I think it should be installed as one of the requirements to build Halide with Python bindings.

I was hoping to call wheel as part of the cibuildwheel steps, but I was unable to find any option to add post-build commands (there seems to be no CIBW_AFTER_BUILD).

.github/workflows/pip.yml
```diff
@@ -192,10 +195,13 @@ jobs:
           CIBW_CONFIG_SETTINGS: "--global-option=egg_info --global-option=-b.dev${{ needs.pip-labels.outputs.halide_pypi_label }}"
```
alexreinking, line 195: I think you can add the following here, rather than patching after the fact with `wheel`:

```suggestion
          CIBW_CONFIG_SETTINGS: "--global-option=egg_info --global-option=-b.dev${{ needs.pip-labels.outputs.halide_pypi_label }} --global-option=--build-number --global-option=${{github.run_id}}"
```

alexreinking: Same idea further up.

.github/workflows/pip.yml
```diff
@@ -192,10 +195,13 @@ jobs:
...
           CIBW_ARCHS_MACOS: "universal2"
 
+      - name: Tag wheels
+        run: wheel tags --build=${{ github.run_id }} ./wheelhouse/*.whl
+
       - uses: actions/upload-artifact@v3
         with:
           name: wheels
-          path: ./wheelhouse/*.whl
+          path: ./wheelhouse/*${{ github.run_id }}*.whl
```
alexreinking, line 204: With the change above, I don't think this will be necessary anymore.

alexreinking (changes requested):
If I understand things correctly, the same thing can be accomplished via the `--build-number` option.

.github/workflows/pip.yml
alexreinking: You might need a `--global-option=wheel` in there or to move the options around.

dragly: Pushed a fixup now

alexreinking (approved):

PR #7505, opened 2023-04-12
metal : replacing `arg_sizes` by `arg_types` in kernel run interface

slomp:


slomp:
I sort of wanted to preserve the `arg_sizes` array, but there does not seem to be an `alloca()` equivalent for runtime modules.

steven-johnson:
> I sort of wanted to preserve the `arg_sizes` array, but there does not seem to be an `alloca()` equivalent for runtime modules.

`__builtin_alloca()`?

slomp:
> > I sort of wanted to preserve the `arg_sizes` array, but there does not seem to be an `alloca()` equivalent for runtime modules.
> 
> `__builtin_alloca()`?

Ah, nice! I suppose I could use VLA as well (to be more idiomatic) since this runtime modules gets compiled by clang, and not clang-cl.

steven-johnson:
> Ah, nice! I suppose I could use a VLA as well (to be more idiomatic) since this runtime module gets compiled by clang, and not clang-cl.

No, please don't, VLAs are problematic in general and especially in runtime code -- IIRC we already had to eliminate some that crept in accidentally and caused problems I can't recall now. if __builtin_alloca() isn't adequate then dropping back to plain malloc/free is the right answer here.

slomp:
I'm debugging things locally here, and when I print the fields of each `arg_types[i]`, something is bogus (`bits()` is always 0).
I think there's more I need to change somewhere in the code to make sure the proper arguments get passed to the `halide_metal_run()`. No clue where it may be, though. Ideas?

slomp:
This CI error is probably unrelated to my changes:
https://buildbot.halide-lang.org/master/#/builders/52/builds/273

src/runtime/metal.cpp
```diff
@@ -789,6 +789,22 @@ WEAK int halide_metal_run(void *user_context,
...
+    int num_kernel_args = 0;
+    for (int i = 0; args[i] != nullptr; i++) {
+        ++num_kernel_args;
+    }
+
+    // NOTE(slomp): arg_sizes[] must terminate with a zero!
+    size_t *arg_sizes = (size_t *)__builtin_alloca((num_kernel_args + 1) * sizeof(size_t));
+    for (int i = 0; i < num_kernel_args; i++) {
+        if (arg_is_buffer[i]) {
+            arg_sizes[i] = sizeof(void *);  // or should it always be uint64_t as in line 876 below?
```
steven-johnson, line 801: In practice, sizeof(void*) == sizeof(uint64_t) for all Metal applications since all Metal OS versions are 64-bitonly.

src/runtime/metal.cpp
```diff
@@ -789,6 +789,22 @@ WEAK int halide_metal_run(void *user_context,
 
     set_compute_pipeline_state(encoder, pipeline_state);
 
+    int num_kernel_args = 0;
+    for (int i = 0; args[i] != nullptr; i++) {
+        ++num_kernel_args;
+    }
+
+    // NOTE(slomp): arg_sizes[] must terminate with a zero!
```
steven-johnson, line 797: Why is that necessary? From a casual inspection it appears that the rest of the code here accesses the arg_sizes array by index and not null termination

slomp: I was just trying to preserve the original semantics of `arg_sizes[]`, in case someone changes code in the future and still relies on the fact it is zero-terminated.

I'd say once we have all runtimes using `halide_type_t arg_types[]`, we should probably add an extra `num_args` to the `halide_*_run()` interface.

steven-johnson: I think this could be done more simply by just doing the size calculations inline and skipping the intermediate buffer entirely; it's only a couple of places. See this gist for an example:

https://gist.github.com/steven-johnson/7811391b60fd21672f36be82d4ec9a1f

src/runtime/metal.cpp
slomp: It looks like we are generating 32bits metal modules, though:
```
1>Generating initmod.metal_32.bc
1>Generating initmod.metal_32_debug.bc
```

slomp:
Hmmm:
```
        if (arg_is_buffer[i]) {
            halide_debug_assert(user_context, arg_types[i].bytes() == sizeof(uint64_t));
            ...
        }
```
this assertion triggers (`-debug`).
`arg_types[]` is referring to the underlying numerical type of the buffer elements, and not to `halide_type_handle`.
Should I just get rid of the assert then?

src/runtime/HalideRuntimeMetal.h
```diff
@@ -33,7 +33,7 @@ extern int halide_metal_run(void *user_context,
                             int blocksX, int blocksY, int blocksZ,
                             int threadsX, int threadsY, int threadsZ,
                             int shared_mem_bytes,
-                            size_t arg_sizes[],
+                            struct halide_type_t arg_sizes[],
```
shoaibkamil, line 36: Nit: `arg_sizes` -> `arg_types`

slomp:
current failure:
https://buildbot.halide-lang.org/master/#builders/33/builds/204
```
17/52 Test  #31: mullapudi2016_reorder ...................***Failed  Required regular expression not found. Regex=[Success!
]  4.17 sec
Warning: Outer dim vectorization of var "x" in function "r$1.update(0)"
Autoscheduler time (3) is slower than expected:
======================
Manual time: 9.14377ms
Auto time: 30.0692ms
======================
```
I don' think this is related, or am I wrong?

slomp:
All right, are we good to merge this?

steven-johnson (approved):

Issue #7507, opened 2023-04-13
Deprecate vs Remove `stmt_html` output for Halide 16?


steven-johnson:
With #7421 having landed, there is really not much use for the `stmt_html` output any more.

Ordinarily I'd say we should deprecate it, but (1) it will be hard to emit useful deprecation warnings for all use cases, and (2) I don't recall seeing a single use of it inside Google for several years now -- I wouldn't be surprised if the same is true externally.

So, taking a straw poll here -- does it make more sense to just rip off the band-aid, so to speak, and remove it entirely for Halide 16?

shoaibkamil:
I think it's ok to remove it, but we may want to alias `stmt_html` to output `stmt_viz` or at least provide a reasonable error message-- but that seems just as hard as a deprecation message.

slomp:
One thing to consider: compared to stmt_html, how long does it take to generate? And also, how long does it take to open it in a browser?

abadams:
I propose that to a user of the releases, it should just appear as if the html output got better. So we don't introduce any "viz" names at all (which already collide with HalideTraceViz), and just swap in the new html output and delete the old one at the same time.

steven-johnson:
Tagging @maaz139 as his input would be welcomed here

maaz139:
> One thing to consider: compared to stmt_html, how long does it take to generate? And also, how long does it take to open it in a browser?
The new HTML is definitely slower to load. In general, html files above ~15mb were not working well in my experiments. I think there are ways to improve the HTML performance and it would be worth doing so if people are actively using the feature and running into this issue.

Other than performance, I don't see a reason to keep the old html file around. I would advocate silently swapping the new viz.html for the old html tags as Andrew suggested.

antonysigma:
Chiming in... As a Nvidia GPU programmer, I am in favor of the new `StmtViz` output as long as we can bring back the `gpu_threads`, `gpu_blocks` IR visualization back in Halide 10. Ever since I moved to Halide 14, all the scheduled GPU code are visualized as assembly code.

The strength of `StmtViz` tool is to compare the IR and assembly code side by side, with cross referencing. However, if the IR visualization panel dumps the assembly code like what we do in Halide 14, it kind of defeats the purpose, right? 

The figure below shows how a generic code `blurred_x` , scheduled with `gpu_theads`, is visualized in Halide 10 and Halide 14.

[...]

PR #7508, opened 2023-04-13
Add a few missing support bits for StmtViz

steven-johnson:
This is to augment #7421 and add some missing build-and-test support bits that got overlooked, including output from the CMake generator rules, some Python support issues, README, tutorial, and multitarget output.

### DIFF
--- a/python_bindings/test/correctness/compile_to.py
+++ b/python_bindings/test/correctness/compile_to.py
@@ -43,8 +43,12 @@ def main():
         else:
             assert os.path.isfile(os.path.join(tmpdir, "f_all.o"))
 
-        p = os.path.join(tmpdir, "f.html")
+        p = os.path.join(tmpdir, "f.stmt.html")
         f.compile_to({hl.OutputFileType.stmt_html: p}, args, "f")
+        assert os.path.isfile(p)
+
+        p = os.path.join(tmpdir, "f.stmt.viz.html")
+        f.compile_to({hl.OutputFileType.stmt_viz: p}, args, "f")
         assert os.path.isfile(p)
 
     finally:
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -535,6 +535,10 @@ void Module::compile(const std::map<OutputFileType, std::string> &output_files) 
         debug(1) << "Module.compile(): stmt_html " << output_files.at(OutputFileType::stmt_html) << "\n";
         Internal::print_to_html(output_files.at(OutputFileType::stmt_html), *this);
     }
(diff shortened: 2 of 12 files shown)


python_bindings/test/correctness/compile_to.py
```diff
@@ -43,10 +43,14 @@ def main():
         else:
             assert os.path.isfile(os.path.join(tmpdir, "f_all.o"))
 
-        p = os.path.join(tmpdir, "f.html")
+        p = os.path.join(tmpdir, "f.stmt.html")
```
maaz139, line 46: Should this be `.stmt.viz.html`

steven-johnson: No, this is (still) testing the older `stmt_html` output (which may be deprecated/removed, see #7507)

steven-johnson:
Gentle review ping

src/Module.cpp
```diff
@@ -535,6 +535,10 @@ void Module::compile(const std::map<OutputFileType, std::string> &output_files)
         debug(1) << "Module.compile(): stmt_html " << output_files.at(OutputFileType::stmt_html) << "\n";
         Internal::print_to_html(output_files.at(OutputFileType::stmt_html), *this);
     }
+    if (contains(output_files, OutputFileType::stmt_viz)) {
+        debug(1) << "Module.compile(): stmt_viz " << output_files.at(OutputFileType::stmt_viz) << "\n";
+        Internal::print_to_viz(output_files.at(OutputFileType::stmt_viz), *this);
```
maaz139, line 540: Unclear why these lines were added, this code is already present on line 621 https://github.com/halide/Halide/blob/929b5d7ce952ec12753c0f7637d67375a6c6c5eb/src/Module.cpp#L621

Since the features displays the compiled assembly, we run it after the assembly flag has been processed.

src/Pipeline.cpp
```diff
@@ -340,7 +340,7 @@ void Pipeline::compile_to_lowered_stmt(const string &filename,
                                        StmtOutputFormat fmt,
                                        const Target &target) {
     Module m = compile_to_module(args, "", target);
-    m.compile(single_output(filename, m, fmt == HTML ? OutputFileType::stmt_html : OutputFileType::stmt));
+    m.compile(single_output(filename, m, fmt == HTML ? OutputFileType::stmt_html : (fmt == StmtViz ? OutputFileType::stmt_viz : OutputFileType::stmt)));
```
maaz139, line 343: Similarly, should we first generate the assembly file here if the target format is `stmt_viz`?

steven-johnson:
Closing in favor of #7516

PR #7509, opened 2023-04-14
Remove all usage of `stmt_html` from Halide (Fixes #7507)

### LINKED ISSUE #7507: Deprecate vs Remove `stmt_html` output for Halide 16?

steven-johnson:
This takes the approach of "Remove completely" vs "Deprecate for Halide 16" -- it's not at all clear if this is what the consensus is, but putting it there for consideration.


maaz139:
In terms of features, `.html` files are strict subset of `.viz.html` and therefore should be removed. However, they do scale to larger stmt files, so it may be worth keeping it around until the viz html is optimized sufficiently?

Issue #4088, opened 2019-08-02
Integrating with OSS-Fuzz


Google-Autofuzz:
Greetings Halide developers and contributors,

We’re reaching out because your project is an important part of the open source ecosystem, and we’d like to invite you to integrate with our [fuzzing](https://www.owasp.org/index.php/Fuzzing) service, [OSS-Fuzz]( https://opensource.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html ). OSS-Fuzz is a free fuzzing infrastructure you can use to identify security vulnerabilities and stability bugs in your project, like #4047 and #4003. OSS-Fuzz will:

- Continuously run all the fuzzers you write.
- Alert you when it finds issues.
- Automatically close issues after they’ve been fixed by a commit.

Many widely used [open source projects]( https://github.com/google/oss-fuzz/tree/master/projects ) like OpenSSL, FFmpeg, LibreOffice, and ImageMagick are fuzzing via OSS-Fuzz, which helps them find and remediate [critical issues]( https://bugs.chromium.org/p/oss-fuzz/issues/list?can=1&q=status%3AFixed%2CVerified+Type%3ABug%2CBug-Security+-component%3AInfra+ ). 

Even though typical integrations can be done in < 100 LoC, we have a [reward program]( https://www.google.com/about/appsecurity/patch-rewards/ ) in place which aims to recognize folks who are not just contributing to open source, but are also working hard to make it more secure.

We want to stress that anyone who meets the eligibility criteria and integrates a project with OSS-Fuzz is eligible for a reward.

If you're not interested in integrating with OSS-Fuzz, it would be helpful for us to understand why—lack of interest, lack of time, or something else—so we can better support projects like yours in the future.

If we’ve missed your question in our [FAQ]( https://github.com/google/oss-fuzz/blob/master/docs/faq.md ), feel free to reply or reach out to us at oss-fuzz-outreach@googlegroups.com.

Thanks!

Julien,
OSS-Fuzz Team

alexreinking:
We have some fuzz tests already. I'd be happy to try running them on your service.

alexreinking:
A major issue I'm finding with integrating OSS-Fuzz into our project is that we don't really ingest strings anywhere. Our existing fuzz tests always build well-typed expressions and the error checking in our IR data structures aborts the process if constructing an ill-typed IR node is attempted.

[...]

PR #7512, opened 2023-04-14
Add libfuzzer compatible fuzz harness

silvergasp:
As discussed in #4088 the fuzz_simplify harness is an easy candidate to begin support for coverage-driven feedback required for high-performance fuzz testing.


silvergasp:
I wasn't sure on how best to manage the Cmake build files for fuzz tests, so I'm certainly open to suggestions on some of the changes here :)

alexreinking:
Because of the special toolchain requirements, I think this test should be moved to `test/fuzz/simplify.cpp`. That folder should get its own `CMakeLists.txt` containing the following:

```
tests(GROUPS fuzz
      SOURCES
      simplify.cpp
      )
target_compile_options(fuzz_simplify PRIVATE -fsanitize=fuzzer,address)
target_link_options(fuzz_simplify PRIVATE -fsanitize=fuzzer,address)
```

The `test/CMakeLists.txt` should be modified to contain the following:

```
# near the top:
include(CheckCXXCompilerFlag)

# toward the bottom
check_cxx_compiler_flag("-fsanitize=fuzzer,address" HAS_FUZZ_FLAGS)
cmake_dependent_option(
    WITH_TEST_FUZZ "Build fuzz tests" ON
    HAS_FUZZ_FLAGS OFF
)
```

silvergasp:
Thanks for the quick response, I've made the suggested changes and also added some of the precursors for multi-fuzzer engine builds for oss-fuzz:)

test/CMakeLists.txt
```diff
@@ -61,3 +62,17 @@ else ()
 endif ()
 
 # FIXME: failing_with_issue is dead code :)
+
+check_cxx_compiler_flag("-fsanitize=fuzzer,address" HAS_FUZZ_FLAGS)
```
steven-johnson, line 66: What compiler(s) support this flag? We need documentation on this somewhere, at least in a comment here or elsewhere.

test/fuzz/simplify.cpp
```diff
@@ -1,5 +1,6 @@
 #include "Halide.h"
 #include <array>
+#include <fuzzer/FuzzedDataProvider.h>
```
steven-johnson, line 3: What happens if libfuzzer isn't installed on the build machine -- will this test be able to compile?

silvergasp: Libfuzzer is shipped with clang as part of the llvm project as clang/llvm are already dependencies of this project I don't think that this will be an issue. That specific [header file](https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h) is sort of fuzzer adjacent, and is more of a helper for transforming raw data into structured data. But the header itself is not required to write a fuzzer.

test/CMakeLists.txt
silvergasp: As far as I'm aware if we were to break this up into two flags;
- `-fsanitize=fuzzer` supported by clang+msvc
- `-fsanitize=address` aka ASAN support by clang+gcc+msvc

Note that the actual flags for msvc are slightly different e.g. `\fsanitize` rather than `-fsanitize`

silvergasp: I'll add a comment here :)

test/fuzz/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+tests(GROUPS fuzz
+      SOURCES
+      simplify.cpp
+)
+
+
+target_link_libraries(fuzz_simplify PRIVATE Halide::Halide)
+
+# Allow OSS-fuzz to manage flags directly
+if (DEFINED ENV{LIB_FUZZING_ENGINE})
```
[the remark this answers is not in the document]
silvergasp: The `LIB_FUZZING_ENGINE` is a somewhat well-known variable. It's used to allow for overriding the fuzzing engine entirely. So the goal here is to have libfuzzer+address-sanitizer be the default but then if LIB_FUZZING_ENGINE is declared you can override the fuzzing engine to one of;
- [Centipede](https://github.com/google/centipede)
- [Hongfuzz](https://github.com/google/honggfuzz)
- [AFL++](https://github.com/AFLplusplus/AFLplusplus)
- etc.

Using this variable is also a [precursor to integrating with OSS-fuzz](https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#build-support). So while I could rename it to `HALIDE_LIB_FUZZING_ENGINE` it would probably add some unnecessary indirection further down the line.

silvergasp:
Looks like I've got at least a couple of formatting things to fix. 

On a different note, I've got a corresponding [draft PR to integrate with OSS-fuzz](https://github.com/google/oss-fuzz/pull/10107). A couple of things I'll need for that integration to go through;
- A primary contact email address (associated with a google/gmail account).
- A list of email addresses that you want to have CC'd into automatic bug reports from oss-fuzz.
- A contributor here with write access to Halide to comment mentioning that you approve of integration with oss-fuzz.

If you are happy for me to do so, I can pull the email addresses from the git log, it's worth noting however that these emails will be listed in plaintext in the oss-fuzz repo.

test/CMakeLists.txt
steven-johnson: > \fsanitize

...flags starting with a backslash, truly the end times are nigh :-)

test/fuzz/CMakeLists.txt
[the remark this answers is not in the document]
steven-johnson: OK, cool. Please try to integrate these comments into the CMake file or whatnot so that they are captured in the source rather than just here in the PR -- it will make our lives easier in the future.

test/CMakeLists.txt
alexreinking: > > \fsanitize
> 
> ...flags starting with a backslash, truly the end times are nigh :-)

No, the MSVC flag starts with a `/` not a `\`.

alexreinking: But AFAIK, MSVC understands `-` for the sake of compatibility, too.

test/fuzz/CMakeLists.txt
alexreinking, line 10: Please do not use environment variables this way in CMake. Unlike with autotools, the user does not control when the build system is regenerated; the build system itself does according to CMake's analysis of the build file dependencies, which includes CMake's own implementation. This will cause vexing behavior when a developer upgrades CMake. Imagine:

> I upgraded CMake and now the fuzz tests aren't building! Why? Oops, I didn't set `LIB_FUZZING_ENGINE` in Bash again.

Thus, cache variables should always dominate. To initialize a cache variable from the environment, write:

```suggestion
set(LIB_FUZZING_ENGINE "$ENV{LIB_FUZZING_ENGINE}"
    CACHE STRING "Compiler flags necessary to use libFuzzer instrumentation")

if (LIB_FUZZING_ENGINE)
```

test/CMakeLists.txt
alexreinking: > What compiler(s) support this flag? We need documentation on this somewhere, at least in a comment here or elsewhere.

Part of the point of checking is that it's less necessary to know the complete matrix of compiler vendors and versions. We could add some `message(VERBOSE)`s for debugging when detection fails.

alexreinking (changes requested):

steven-johnson:
> Part of the point of checking is that it's less necessary to know the complete matrix of compiler vendors and versions

No, but when it inevitably fails somewhere, having comments and/or messages to help explain would be nice

test/fuzz/CMakeLists.txt
silvergasp: Cool thanks for the suggestion. I'll add that in. I'm definitely a cmake noob, so I appreciate the guidance.

alexreinking: > I'm definitely a cmake noob,

For your own sanity, stay that way 😆

silvergasp: > > I'm definitely a cmake noob,
> 
> For your own sanity, stay that way 😆

lol yeah, I find cmake in general just... baffling

silvergasp:
Ok so I think there is a bit of [a bug with cmakes `CHECK_CXX_COMPILER_FLAG`](https://cmake.cmake.narkive.com/faC8opxU/check-cxx-compiler-flag-usage) as it doesn't seem to support checking sanitiser flags. Something to do with it needing those sanitizer flags during linking as well, so the check fails. I'm going to have a bit of a go at using CheckCXXSourceCompiles instead as it seems to give a bit more control over how/what flags are used during compiling/linking.

test/CMakeLists.txt
```diff
@@ -61,3 +62,29 @@ else ()
...
+#   with a GCC based project you should consider looking into the LIB_FUZZING_ENGINE
+#   env variable defined in `test/fuzz/CMakeLists.txt`.
+check_cxx_compiler_flag("-fsanitize=fuzzer,address" HAS_FUZZ_FLAGS)
+
+if (not HAS_FUZZ_FLAGS)
+    message(VERBOSE "Compiler does not support libfuzzer/address sanitizer.")
+endif ()
+
+cmake_dependent_option(
+    WITH_TEST_FUZZ "Build fuzz tests" ON
```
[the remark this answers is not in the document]
alexreinking: No. See the documentation here: https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

If `HAS_FUZZ_FLAGS` is false, then `WITH_TEST_FUZZ` must be false, too.

test/CMakeLists.txt
```diff
@@ -61,3 +62,33 @@ else ()
...
+# - Address sanitizer is often used in conjunction with fuzzing as it will detect
+#   common high severity bugs. This sanitizer is used as a "default" for fuzzing
+#   when the sanitizer isn't otherwise specified.
+# - Fuzzer sanitizer will link against libfuzzer and is currently only supported
+#   on clang/msvc and isn't supported with GCC. If you need to use these fuzzers
+#   with a GCC based project you should consider looking into the LIB_FUZZING_ENGINE
+#   env variable defined in `test/fuzz/CMakeLists.txt`.
+set(CMAKE_REQUIRED_LINK_OPTIONS "-fsanitize=fuzzer,address")
+set(CMAKE_REQUIRED_FLAGS "-fsanitize=fuzzer-no-link,address")
+check_cxx_source_compiles("#include <cstdint>\n extern \"C\" int LLVMFuzzerTestOneInput(const uint8_t *Data, std::size_t Size) { return 0; }" HAS_FUZZ_FLAGS)
```
alexreinking, line 76: I'd prefer to see this as a multi-line string
```suggestion
check_cxx_source_compiles([[
#include <cstdint>
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, std::size_t Size) {
  return 0;
}
]] HAS_FUZZ_FLAGS)
```

test/CMakeLists.txt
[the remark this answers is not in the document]
silvergasp: Cool, yeah I've switched back in the latest revision, it looks like it was more a problem with the flag check than how the dependant option was configured.

silvergasp:
I think I've covered most things here. Is [this build failure related to my changes](https://buildbot.halide-lang.org/master/#/builders/52/builds/330). The error is;
```
The following tests FAILED:
	243 - correctness_multi_way_select (Failed)
Errors while running CTest
program finished with exit code 8
elapsedTime=578.168036
```

test/CMakeLists.txt
silvergasp: Done in https://github.com/halide/Halide/pull/7512/files#diff-33394812ba204689144fd2f80832db83853ba1cb32403edb4e15fe4893e675fdR84-R86

steven-johnson (approved):

steven-johnson:
LGTM, waiting on @alexreinking approval

alexreinking (approved):
This is great! Thanks for the contribution!

test/fuzz/simplify.cpp
```diff
@@ -14,76 +15,73 @@ using namespace Halide::Internal;
...
-
 Type fuzz_types[] = {UInt(1), UInt(8), UInt(16), UInt(32), Int(8), Int(16), Int(32)};
 const int fuzz_type_count = sizeof(fuzz_types) / sizeof(fuzz_types[0]);
 
 std::string fuzz_var(int i) {
     return std::string(1, 'a' + i);
 }
 
-Expr random_var() {
-    int fuzz_count = rng() % fuzz_var_count;
```
silvergasp, line 28: TODO: change as per #7546

test/fuzz/simplify.cpp
```diff
@@ -14,76 +15,73 @@ using namespace Halide::Internal;
...
-int get_random_divisor(Type t) {
+int get_random_divisor(FuzzedDataProvider &fdp, Type t) {
     std::vector<int> divisors = {t.lanes()};
     for (int dd = 2; dd < t.lanes(); dd++) {
         if (t.lanes() % dd == 0) {
             divisors.push_back(dd);
         }
     }
 
-    return divisors[rng() % divisors.size()];
```
silvergasp, line 49: Looks fine as per #7546

test/fuzz/simplify.cpp
```diff
@@ -14,76 +15,73 @@ using namespace Halide::Internal;
...
+    return divisors[fdp.ConsumeIntegralInRange<size_t>(0, divisors.size() - 1)];
 }
 
-Expr random_leaf(Type T, bool overflow_undef = false, bool imm_only = false) {
+Expr random_leaf(FuzzedDataProvider &fdp, Type T, bool overflow_undef = false, bool imm_only = false) {
     if (T.is_int() && T.bits() == 32) {
         overflow_undef = true;
     }
     if (T.is_scalar()) {
-        int var = rng() % fuzz_var_count + 1;
```
silvergasp, line 57: Should be `fdp.ConsumeIntegralInRange<int>(0, fuzz_var_count)` as per #7546 NOTE: drop the `+ 1`

test/fuzz/simplify.cpp
```diff
@@ -14,76 +15,73 @@ using namespace Halide::Internal;
...
-            auto v1 = random_var();
+            auto v1 = random_var(fdp);
             return cast(T, v1);
         } else {
             if (overflow_undef) {
                 // For Int(32), we don't care about correctness during
                 // overflow, so just use numbers that are unlikely to
                 // overflow.
-                return cast(T, (int)(rng() % 256 - 128));
+                return cast(T, fdp.ConsumeIntegralInRange<int>(-128, 128));
```
silvergasp, line 64: Should be `fdp.ConsumeIntegralInRange<int>(-128, 127)` as per #7546

test/fuzz/simplify.cpp
```diff
@@ -14,76 +15,73 @@ using namespace Halide::Internal;
...
             } else {
-                return cast(T, (int)(rng() - RAND_MAX / 2));
+                return cast(T, fdp.ConsumeIntegralInRange<int>(-RAND_MAX / 2, RAND_MAX / 2));
```
silvergasp, line 66: Should be `fdp.ConsumeIntegralInRange<int>(-RAND_MAX / 2, RAND_MAX / 2 -1)` as per #7546

test/fuzz/simplify.cpp
```diff
@@ -95,13 +93,13 @@ Expr random_condition(Type T, int depth, bool maybe_scalar) {
...
+    if (maybe_scalar && fdp.ConsumeBool()) {
         T = T.element_of();
     }
 
-    Expr a = random_expr(T, depth);
-    Expr b = random_expr(T, depth);
-    int op = rng() % op_count;
+    Expr a = random_expr(fdp, T, depth);
+    Expr b = random_expr(fdp, T, depth);
+    int op = fdp.ConsumeIntegralInRange<int>(0, op_count);
```
silvergasp, line 102: Should be `fdp.ConsumeIntegralInRange<int>(0, op_count-1)` as per #7546

test/fuzz/simplify.cpp
```diff
@@ -134,60 +132,60 @@ Expr random_expr(Type T, int depth, bool overflow_undef) {
...
-        return random_leaf(T, overflow_undef);
+        return random_leaf(fdp, T, overflow_undef);
     }
 
     const int bin_op_count = sizeof(make_bin_op) / sizeof(make_bin_op[0]);
     const int bool_bin_op_count = sizeof(make_bool_bin_op) / sizeof(make_bool_bin_op[0]);
     const int op_count = bin_op_count + bool_bin_op_count + 5;
 
-    int op = rng() % op_count;
+    int op = fdp.ConsumeIntegralInRange<int>(0, op_count);
```
silvergasp, line 142: Should be `fdp.ConsumeIntegralInRange<int>(0, op_count-1)`

silvergasp (commented):
@steven-johnson would you mind double checking these? I'll have a PR up shortly with these changes.

test/fuzz/simplify.cpp
```diff
@@ -14,76 +15,73 @@ using namespace Halide::Internal;
...
-    int fuzz_count = rng() % fuzz_var_count;
+Expr random_var(FuzzedDataProvider &fdp) {
+    int fuzz_count = fdp.ConsumeIntegralInRange<int>(0, fuzz_var_count);
     return Variable::make(Int(0), fuzz_var(fuzz_count));
 }
 
-Type random_type(int width) {
-    Type T = fuzz_types[rng() % fuzz_type_count];
+Type random_type(FuzzedDataProvider &fdp, int width) {
+    Type T = fdp.PickValueInArray(fuzz_types);
```
steven-johnson, line 31: Hadn't noticed this call existed -- perhaps we should aggressively use it in other cases too where possible (e.g. line 47, line 102)

test/fuzz/simplify.cpp
steven-johnson: The semantics of fuzz_var_count aren't clear to me -- in various for loops we do `for (int i = 0; i < fuzz_var_count; i++)` which implies that we need to limit the range to max=fuzz_var_count-1?

test/fuzz/simplify.cpp
steven-johnson: LGTM

steven-johnson: ...except, RAND_MAX applies to the `rand()` call in stdlib; does libfuzzer also make the same promise about range?

test/fuzz/simplify.cpp
steven-johnson: or better yet, use PickValueInArray()?

silvergasp: True that would be better

test/fuzz/simplify.cpp
steven-johnson: (I realize the fragility was pre-existing before your changes, but improving the resilience here seems easy and worthwhile while we are here)

test/fuzz/simplify.cpp
silvergasp: Yeah looking at this again the `var` variable seems to only get used once on line 56, so I could probably just get rid of it and replace the condition on line 56 with `fdp.ConsumeBool`.

test/fuzz/simplify.cpp
silvergasp: Yeah, looking at this again a second time I'm struggling to understand what is going on here. Where did [this `5`](https://github.com/halide/Halide/pull/7512/files#diff-0bb69bc3a0270944620451c6da8561235b4d1062aa5acea5357e3aa29f99e451R140) come from.

test/fuzz/simplify.cpp
silvergasp: I'm guessing it was +1 in the beginning so that the condition on line 56 was sometimes true/false. In any case it seems like a really roundabout way of doing it. I'm just going to replace it with the consumebool method.

PR #7513, opened 2023-04-15
Improve assembly-file finding logic in StmtToViz

steven-johnson:
(1) Avoid having to guess at location by just passing in the location, since we usually already know it. (2) If we don't know it, be more cautious when constructing it: the output html filename might not match our expectations, and all file extensions must use get_output_info() to work correctly on all platforms.


maaz139 (approved):
Looks good to me!

PR #7515, opened 2023-04-17
Allow emitting `stmt_viz` without specifying `assembly`

steven-johnson:
TL;DR: if we request `stmt_viz` without `assembly`, just generate the latter to a temp file that we dispose of later; this wasn't feasible before since we were previously requiring the assembly output to be generated with the same directory and basename as stmt_viz, but that was fixed.

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -57,12 +57,12 @@ std::map<OutputFileType, const OutputInfo> get_output_info(const Target &target)
 
 namespace {
 
-class TemporaryObjectFileDir final {
+class TemporaryFileDir final {
 public:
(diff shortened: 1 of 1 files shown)


abadams:
Does stmt_viz refer to the assembly file in-place, or does it slurp it into the generated html?

maaz139:
> > Does stmt_viz refer to the assembly file in-place, or does it slurp it into the generated html?
> 
> Pretty sure it just slurps, but I will doublecheck

That is correct, we read it from the assembly file into memory and emit a copy in the html.

abadams (approved):

maaz139 (approved):

PR #7516, opened 2023-04-17
Convert stmt_html output to use stmt_viz output

steven-johnson:
Per discussion on #7507, this entirely removes the "classic" stmt_html output and replaces it with the "new" StmtToViz output. Using `compile_to_lowered_stmt` or requesting `stmt_html` will now always output the new output, and requesting `stmt_viz` output is no longer legal.

(Note that this builds on top of #7515, which must be submitted first.)

It's not clear to me whether https://github.com/halide/Halide/issues/7507#issuecomment-1511761706 is a blocker for this change, or a request to add back already-lost functionality.


abadams:
It's a request to add back lost functionality. Things that used to be in codegen have crept into lowering (e.g. compiling shaders), making the .stmt output lower and lower level over time. We should roll it back a little to before those passes.

steven-johnson:
> It's a request to add back lost functionality

OK, in that case we should probably be good to go in landing this.

Makefile
```diff
@@ -560,7 +560,6 @@ SOURCE_FILES = \
   SpirvIR.cpp \
   SplitTuples.cpp \
   StageStridedLoads.cpp \
-  StmtToHtml.cpp \
```
maaz139, line 563: Also need to remove the header file at https://github.com/halide/Halide/blob/c94e9e7452c0fca22205935f0b204f49cdd3dfef/Makefile#L741

maaz139 (commented):
StmtToHtml.h needs to be removed from the Makefile, otherwise looks good to me!

maaz139 (approved):

src/Generator.cpp
```diff
@@ -790,12 +790,12 @@ gengen
...
         } else {
-            // if emit_flags contains "stmt_viz" but not "assembly", throw an error
-            bool has_stmt_viz = std::find(emit_flags.begin(), emit_flags.end(), "stmt_viz") != emit_flags.end();
+            // if emit_flags contains "stmt_html" but not "assembly", throw an error
+            bool has_stmt_html = std::find(emit_flags.begin(), emit_flags.end(), "stmt_html") != emit_flags.end();
             bool has_assembly = std::find(emit_flags.begin(), emit_flags.end(), "assembly") != emit_flags.end();
 
-            user_assert(!has_stmt_viz || has_assembly)
-                << "Output flag `stmt_viz` requires the `assembly` flag to also be set.";
+            user_assert(!has_stmt_html || has_assembly)
```
maaz139, line 797: Is this still needed considering #7515 has landed?

steven-johnson: Oh, good catch! Removed.

PR #7517, opened 2023-04-17
Forbid assigning to Buffer(Expr) by introducing an intermediate type.

### LINKED ISSUE #7514: Update definition of Buffer takes precedence over any prior usage
Consider this example where each pixel needs to undergo two tests. The final value of the pixel must be determined by the minimum value of the two tests.

```C++
Var x, y;
Func f;
int32_t width = 100;
int32_t height = 100;
Buffer<uint8_t> output(width, height);
Buffer<double> minBuffer(width, height);
minBuffer.fill(1e10);
f(x, y) = cast<uint8_t>(0);
std::vector<double> zs = {0, 1};
for (const double& z : zs)
{
    Expr zTest = Expr(z) < minBuffer(x, y);
    f(x, y) = select(zTest, cast<uint8_t>(Expr(z) * Expr(255)), f(x, y));
    minBuffer(x, y) = select(zTest, Expr(z), minBuffer(x, y));
}
f.realize(output);
```

In this case, the final value of the pixel must be `0` *black image* since the minimum `z = 0`. However, I'm getting back a *white image*, suggesting that the update definition to `minBuffer` is being evaluated before the `zTest` has taken place.

abadams:
Fixes #7514 by triggering a compile-time error instead of silently doing something surprising.

### DIFF
--- a/src/Buffer.h
+++ b/src/Buffer.h
@@ -581,13 +581,13 @@ public:
      * given symbolic coordinate. */
     // @{
     template<typename... Args>
-    const Expr operator()(const Expr &first, Args... rest) const {
+    const Expr operator()(const Expr &first, Args... rest) const {  // NOLINT
         std::vector<Expr> args = {first, rest...};
         return (*this)(args);
     }
 
     template<typename... Args>
-    const Expr operator()(const std::vector<Expr> &args) const {
+    const Expr operator()(const std::vector<Expr> &args) const {  // NOLINT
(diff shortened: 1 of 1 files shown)


abadams:
The issue is that it's a C++ compile-time failure, not a Halide compile-time failure, so I don't think an error test is possible.

steven-johnson:
Ready to land?

PR #7518, opened 2023-04-17
Fix Anderson2021 tests to avoid spurious failures on non-Cuda systems

steven-johnson:
The Anderson2021 autoscheduler is pretty Cuda-specific, so some tests assume it is present; this is pretty much never true on macOS, and annoying spurious failures are annoying. This adds a new flag and capability to RunGenMain to try to sniff out the necessary runtime setup and make it a quiet [SKIP] failure when testing.


steven-johnson:
The Windows errors make no bloody sense: the errors are reported on lines 557 and 559, but the descriptions don't fit the source at all.

steven-johnson:
Finally building, PTAL

tools/RunGenMain.cpp
```diff
@@ -176,6 +176,12 @@ Usage: $NAME$ argument=value [argument=value... ] [flags]
         (This is mainly useful for use with Halide's testing infrastructure,
         which relies on this for successful tests.)
 
+    --skip_bad_environement:
+        If the Generator is build for a runtime environment we don't have available
```
pranavb-ca, line 180: ``s/build/built/``

tools/RunGenMain.cpp
```diff
@@ -499,6 +513,63 @@ int main(int argc, char **argv) {
...
+                {"cuda_capability80", 80},
+                {"cuda_capability86", 86},
+            };
+            for (const auto &it : capabilities) {
+                if (tokens.count(it.first)) {
+                    version_required = std::max(version_required, it.second);
+                }
+            }
+
+            void *cuda_intf = halide_get_symbol("halide_cuda_device_interface");
```
pranavb-ca, line 549: <nit> consider moving this check to the start of this ``if`` block so we exit early in the instances that we need to.

pranavb-ca (changes requested):

steven-johnson:
PTAL

pranavb-ca (approved):

PR #7522, opened 2023-04-19
Make stmt_html generation work correctly for submodules

steven-johnson:


abadams:
I don't understand why the stmt is erased but the stmt_html is not.

steven-johnson:
> I don't understand why the stmt is erased but the stmt_html is not.

We used to emit stmt_html along with stmt, before resolving submodules, because we didn't need to generate and object (etc) code. Now that stmt_html includes assembly code, it happens after the recursive call, so if there are submodules (eg for hexagon), stmt_html never gets emitted.

src/Module.cpp
```diff
@@ -546,7 +546,6 @@ void Module::compile(const std::map<OutputFileType, std::string> &output_files)
     if (!submodules().empty() && !should_ignore_submodules(output_files)) {
         std::map<OutputFileType, std::string> output_files_copy = output_files;
         output_files_copy.erase(OutputFileType::stmt);
```
abadams, line 548: Please add a comment explaining why the stmt output is erased (this would have cleared up my confusion).

steven-johnson:
In practice, submodules are used only to emit Hexagon output -- it's a weird special case. We skip the stmt and stmt_html because the "outer" ones we generate contain the output for the submodules too, so we we don't want to overwrite the more-complete outer ones with the less-complete inner ones. The problem now is that if we take the submodule path, we no longer generate stmt_html at all -- and my proposed fix here is still not correct. 

I'll add some comments to document what's going on and put in a proper fix (and test).

steven-johnson:
OK, this thing was kind of a mess -- TBH I'm not sure why we ever special-cased stmt and stmt_html; the 'blame' goes back to a change by Dillon in 2017, but it's not at all clear why it was necessary, because (as it turns out) just emitting both from the "post-resolved" Module (ie after its submodules have been integrated) works fine and produces nearly identical output -- the only difference appears to be that doing it this way includes references to the embedded buffer with the Hexagon code, which I had to diff to even notice.

This seems to work fine now -- I updated StmtToViz to indicate that it doesn't handle submodules but shouldn't ever need to; I also massively upgraded correctness_cross_compilation to serve as a testing ground (since the error only showed up if you generated HVX code with the `.hexagon()` scheduling directive). PTAL.

abadams (approved):

maaz139 (approved):

PR #7542, opened 2023-05-01
Fix flag check for fuzzers

silvergasp:
On some system size_t isn't available under `<cstdint>`, however it is garaunteed to be available under `<cstddef>` for all systems.


silvergasp:
This fixes some build in https://github.com/google/oss-fuzz/pull/10107 failures caused by different implementations of the standard c++ library. Not sure how this one slipped passed me, it was working fine on my local machine and was working fine at some point on the oss-fuzz docker containers as well. Essentially the check to see if fuzzing was supported was failing due to an unrelated reason, (size_t not defined).

silvergasp:
So that definetely fixes the cmake flag check issue https://github.com/silvergasp/oss-fuzz/actions/runs/4854664389/jobs/8652309043#step:8:8960 but now I'm running into some other funky runtime issue, when actually running the fuzzers on the oss-fuzz build...

steven-johnson (approved):

PR #7543, opened 2023-05-02
fuzz: Port correctness/cse fuzzer over to libfuzzer

silvergasp:


steven-johnson (approved):

test/fuzz/cse.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <fuzzer/FuzzedDataProvider.h>
+#include <time.h>
+
+using namespace Halide;
+using namespace Halide::Internal;
+using std::vector;
+
+Expr random_expr(FuzzedDataProvider &fdp, int depth, vector<Expr> &exprs) {
+    if (depth <= 0) {
+        return fdp.ConsumeIntegralInRange<int>(-5, 4);
```
silvergasp, line 12: Looks fine #7546 origional `rng() % 10 - 5`

test/fuzz/cse.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    if (!exprs.empty() && fdp.ConsumeBool()) {
+        // Reuse an existing expression
+        return exprs[fdp.ConsumeIntegralInRange<size_t>(0, exprs.size() - 1)];
```
silvergasp, line 17: Looks fine, explicitly references size()-1

test/fuzz/cse.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    Expr next;
+    switch (fdp.ConsumeIntegralInRange<int>(0, 8)) {
```
silvergasp, line 21: This looks ok.

silvergasp (commented):
#7546 These all look fine

test/fuzz/cse.cpp
steven-johnson: If the idea is that a value of 8 is intended to hit the default case, we should add a comment to that effect

test/fuzz/cse.cpp
steven-johnson: Or better yet, use PickValueInArray()?

silvergasp: PickValueInArray doesn't work on vectors... see [implementation](https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h#L75-L78) although this is probably annoying enough that I might just create a PR with the LLVM project.

test/fuzz/cse.cpp
silvergasp: yeah that's what it looks like, I was mostly just attempting to replicate the original logic but while I'm at it I can probably make this clearer too.

test/fuzz/cse.cpp
steven-johnson: (But could always just make a local `PickValueInVector()` wrapper to use in these tests, nice for clarity)

PR #7545, opened 2023-05-02
Remove workaround for GCC 4.x.x in cpuid()

dbabokin:
The PR removes the code with workaround for GCC 4.x.x. With later version of GCC (5 and later) and Clang the problem does not exist and "movl %ebx / xchgl %ebx" hack is not needed - the compiler correctly detects the use of `ebx` and stores/restores `ebx` register. The original problem was compiler error message when compiled with `-fPIC` and now it's safe to remove.

I have complete description of the problem here: https://github.com/ispc/ispc/issues/2504 and I've fixed ISPC code as well (it's the code where this snippet was originally borrowed from).

### DIFF
--- a/src/Target.cpp
+++ b/src/Target.cpp
@@ -41,24 +41,12 @@ static void cpuid(int info[4], int infoType, int extra) {
 // CPU feature detection code taken from ispc
 // (https://github.com/ispc/ispc/blob/master/builtins/dispatch.ll)
 
-#ifdef _LP64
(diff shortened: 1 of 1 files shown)


steven-johnson:
> The docs state that C++17 is required for Halide, so GCC 4.x.x is automatically out of the game.

In that case, LGTM pending buildbots

steven-johnson (approved):

Issue #7546, opened 2023-05-04
fuzz tests call ConsumeIntegralInRange() incorrectly


steven-johnson:
ConsumeIntegralInRange() returns a value in the range [min, max], but in several cases we call it in the form [min, max+1] (e.g. simplify.cpp:102, which should be `(0, op_count - 1)`. This causes intermittent segfaults, usually only in `-O3` mode. We should examine all the extant calls to `ConsumeIntegralInRange()` to verify/fix incorrect usage. attn @silvergasp

silvergasp:
Ah my bad, I should have paid closer attention from the get go. I'll use/comment on the original PR #7512 to double check everything.

steven-johnson:
as they say, there are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors

silvergasp:
Haha, I'll have to borrow that joke

PR #7547, opened 2023-05-04
fix(fuzz): Refactor fuzzers to fix off by 1 errors

silvergasp:
Cleanup the fuzzers making them more readable and fix off by one errors caused by incorrect usage of
FuzzedDataProvider::ConsumeIntegralInRange.

Closes: #7546

### DIFF
--- a/test/fuzz/fuzz_helpers.h
+++ b/test/fuzz/fuzz_helpers.h
@@ -1,0 +1,15 @@
+#ifndef HALIDE_FUZZ_HELPERS_H_
+#define HALIDE_FUZZ_HELPERS_H_
+
+#include "fuzzer/FuzzedDataProvider.h"
+#include <vector>
+
+namespace Halide {
+
+template<typename T>
+inline T pick_value_in_vector(FuzzedDataProvider &fdp, std::vector<T> &vec) {
+    return vec[fdp.ConsumeIntegralInRange<size_t>(0, vec.size() - 1)];
+}
+}  // namespace Halide
+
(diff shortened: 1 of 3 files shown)


test/fuzz/simplify.cpp
```diff
@@ -44,26 +46,25 @@ int get_random_divisor(FuzzedDataProvider &fdp, Type t) {
...
         } else {
             if (overflow_undef) {
                 // For Int(32), we don't care about correctness during
                 // overflow, so just use numbers that are unlikely to
                 // overflow.
-                return cast(T, fdp.ConsumeIntegralInRange<int>(-128, 128));
+                return cast(T, fdp.ConsumeIntegralInRange<int>(-128, 127));
             } else {
-                return cast(T, fdp.ConsumeIntegralInRange<int>(-RAND_MAX / 2, RAND_MAX / 2));
+                return cast(T, fdp.ConsumeIntegral<int>());
```
silvergasp, line 67: I wasn't sure about changing this line but I'm 80% sure it makes sense.

test/fuzz/fuzz_helpers.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
steven-johnson, line 1: `#pragma once` is not portable -- the way to do this in halide is to surround with 

```
#ifndef HALIDE_FUZZ_HELPERS_H_
#define HALIDE_FUZZ_HELPERS_H_
...
#endif  // HALIDE_FUZZ_HELPERS_H_
```

steven-johnson (approved):

PR #7549, opened 2023-05-06
fuzz: Add libfuzzer compatible bounds fuzzer

silvergasp:


silvergasp:
See #4088

rootjalex (commented):
This is great, thanks for doing this! Lgtm pending green buildbots.

rootjalex (approved):

test/fuzz/bounds.cpp
```diff
@@ -96,108 +97,98 @@ Expr random_condition(Type T, int depth, bool maybe_scalar) {
     };
     const int op_count = sizeof(make_bin_op) / sizeof(make_bin_op[0]);
```
steven-johnson, line 98: This is unused, it must be removed

steven-johnson:
I took the liberty of harvesting fixes from #7551 and #7548 and applying them here (since our buildbot doesn't test this stuff, I pulled this into Google to test)... unfortunately it's still broken, as running it with default args ends up trying to construct a binary operation with vector elements with mismatched lanes. Gonna debug some more.

test/fuzz/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ tests(GROUPS fuzz
       SOURCES
       simplify.cpp
       cse.cpp
+      bounds.cpp
```
steven-johnson, line 5: Please sort these alphabetically :-)

steven-johnson:
hrm, this may have found a real bug: it seems we are generating two uint32x6:

```
// Both are uint32x6
a=x3(ramp((uint32)b, (uint32)e, 2)) 
b=x2(ramp((uint32)e, (uint32)4117721957, 3))
```

But if we attempt to cast either to uint1x6 (ie, bool), we fail inside cast at IROperator.cpp:1946 because we don't expect the type to be a combination of Broadcast-and-Ramp. Pondering the right way to handle this.

steven-johnson:
See https://github.com/halide/Halide/issues/7556 for possible fix

steven-johnson:
LGTM. There are still issues to be resolved with the fuzzers but this change seems OK.

steven-johnson:
ready to land?

silvergasp:
All good on my end!

PR #7564, opened 2023-05-12
Avoid potentially infinite loop in fuzz/simplify.cpp

steven-johnson:
FuzzedDataProvider is *not* a RNG; there's no guarantee that it won't return the same data to you forever. This means that the loop to find a new subtype may never terminate (eg if the 'random' type returned always matches the input type). This "fixes" it by just adding a count to break out of the loop, in which case we just use the original type. Not sure if there's a more elegant fix?

attn @silvergasp


abadams (approved):

silvergasp:
This looks reasonable to me.

PR #7577, opened 2023-05-17
[vulkan] Change the feature version requirement to v1.3 for correctness_gpu_dynamic_shared

derek-gerstmann:
The linux-buildbot-4 machine only supports v1.2, and that isn't enough to run this test.

Change the feature detection to require v1.3, otherwise skip the test.

### DIFF
--- a/test/correctness/gpu_dynamic_shared.cpp
+++ b/test/correctness/gpu_dynamic_shared.cpp
@@ -20,8 +20,8 @@ int main(int argc, char **argv) {
         assert(interface->compute_capability != nullptr);
         int major, minor;
         int err = interface->compute_capability(nullptr, &major, &minor);
-        if (err != 0 || (major == 1 && minor < 2)) {
-            printf("[SKIP] Vulkan %d.%d is less than required 1.2.\n", major, minor);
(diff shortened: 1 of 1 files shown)


steven-johnson (approved):

steven-johnson:
What would we need to install/update to get v1.3? Is it an nvidia driver issue, or something else?

derek-gerstmann:
> What would we need to install/update to get v1.3? Is it an nvidia driver issue, or something else?

Yes, the latest proprietary driver should have v1.3 support:
https://developer.nvidia.com/vulkan-driver

derek-gerstmann:
> > > What would we need to install/update to get v1.3? Is it an nvidia driver issue, or something else?
> > 
> > 
> > Yes, the latest proprietary driver should have v1.3 support: https://developer.nvidia.com/vulkan-driver
> 
> Feel free to upgrade both of the Linux bots to the latest driver (or I can do it, LMK)

Okay, let’s see how things work with these fixes on the v1.2 driver. I’ve been doing most of my testing on v1.3 so upgrading should be fine anytime.

PR #7578, opened 2023-05-17
Fix various compilation errors with AppleClang 14.0.3

steven-johnson:
Newer versions of Xcode trigger `-Wbitwise-instead-of-logical` for this usage, which we treat as an error


zvookin (approved):

abadams (approved):

steven-johnson:
FYI, this only showed up using `make`, not `cmake` -- we have slightly different sets of warnings enabled/disabled between the two, apparently. :-/

PR #7587, opened 2023-05-21
[Draft] CodeGen_MLIR: Add initial MLIR CodeGen

xerpi:
Initial implementation. Just a proof-of-concept to show that MLIR generation is possible and quite straightforward.
It uses the following MLIR dialects: `arith`, `func`, `memref`, `scf`, `vector`.

Future work:
- Implement the rest of the Halide IR node conversion
- Use MLIR's `affine` dialect instead of `scf`'s if, for, and memory load/stores to take advantage of `affine` transformations and optimizations.


zvookin:
I look forward to reading the PR and will try to do so soon.

MLIR definitely now has enough prebuilt dialect support to be able to support Halide well. The main issue for re: adding MLIR  support in main is that it adds a significantly large dependency to the build. This has a cost on CI/testing.

The main advantage of MLIR would be as a fairly configurable backend to target a broad variety of dialects flexibly, specifically ML operation sets that end up on TPU hardware and such.

zvookin:
I would like to see a quick readme with minimal info on how you set up/installed MLIR and possibly a shell script that generates something through this path and compiles it with mlir-opt or similar.

xerpi:
@zvookin I have updated the pull request.

I have added `compile_to_mlir` methods to `Func` and `Pipeline`, which I'm not sure it's a good idea. What should be the behavior when Halide is compiled without MLIR, to return an error?

In my project, I use MLIR and CIRCT to generate generic RTL code which I wrap with Xilinx-specific wrappers, and I have also implemented a Halide runtime (`xrt.cpp`) for [Xilinx Runtime Library (XRT)](https://www.xilinx.com/products/design-tools/vitis/xrt.html) to interface with the FPGA device (allocate/deallocate memory, copy data device<->host and run kernel).
For that use case, I treat the FPGA as an "accelerator". For that, I have a custom class `OffloadLoopsToAccelerator` which just marks loops to be executed on the FPGA (similarly to `OffloadGPULoops`) and calls the MLIR CodeGen to generate the MLIR code and replaces the loop execution to calls to the XRT runtime backend.

In the code in this pull request, when `CodeGen_MLIR` is invoked via `compile_to_mlir`  the Halide IR will contain nodes that are not supposed to be present in the code running on the accelerator, just the host (such as `_halide_buffer_get_host`), which are not implemented.

Therefore my question is, should we consider the MLIR code as an "accelerator" that can only run loops, or also code that can also run on a host?

xerpi:
> Hey, just wanted to check in and see where this PR stands. Is it waiting on feedback?

Indeed. I have open questions about how to better integrate the MLIR Codegen. See my last message for the details.

steven-johnson:
> > Hey, just wanted to check in and see where this PR stands. Is it waiting on feedback?
> 
> Indeed. I have open questions about how to better integrate the MLIR Codegen. See my previous message for the details.

Ah, gotcha -- unfortunately @zvookin has been away on leave for a couple of months and it looks like no one took over the responsibility here. I'll bring this up and see who can take it over.

xerpi:
No code changes, just a rebase.

xerpi:
I have rebased the code to upstream and also made it emit MLIR code for `Call::buffer_get_min` and `Call::buffer_get_extent`.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,565 @@
...
+    } else if (op->is_intrinsic(Call::shift_right)) {
+        if (op->type.is_int())
+            value = builder.create<mlir::arith::ShRSIOp>(codegen(op->args[0]), codegen(op->args[1]));
+        else
+            value = builder.create<mlir::arith::ShRUIOp>(codegen(op->args[0]), codegen(op->args[1]));
+    } else if (op->is_intrinsic(Call::widen_right_mul)) {
+        mlir::Value a = codegen(op->args[0]);
+        mlir::Value b = codegen(op->args[1]);
+        if (op->type.is_int())
+            b = builder.create<mlir::arith::ExtSIOp>(a.getType(), b);
```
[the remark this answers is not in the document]
xerpi: Good catch, thanks. It seems `op->type` already contains `op->args[1].type().widen()`:
https://github.com/halide/Halide/blob/8bf7827bf08d655f0c842d4fe7271d5bf65600f9/src/IROperator.cpp#L1136-L1138

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,565 @@
...
+        index = codegen(ramp_base);
+    } else {
+        internal_error << "Unsupported store\n";
+    }
+
+    index = builder.create<mlir::arith::IndexCastOp>(builder.getIndexType(), index);
+    if (op->value.type().is_scalar())
+        builder.create<mlir::memref::StoreOp>(value, buffer, mlir::ValueRange{index});
+    else
+        builder.create<mlir::vector::StoreOp>(value, buffer, mlir::ValueRange{index});
```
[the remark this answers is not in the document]
xerpi: Actually, the difference is that if the type is scalar, `mlir::memref::StoreOp` is being generated, and `mlir::vector::StoreOp` otherwise.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,565 @@
...
+void CodeGen_MLIR::Visitor::visit(const Free *op) {
+    builder.create<mlir::memref::DeallocOp>(sym_get(op->name));
+}
+
+void CodeGen_MLIR::Visitor::visit(const Realize *op) {
+    internal_error << "Unimplemented\n";
+}
+
+void CodeGen_MLIR::Visitor::visit(const Block *op) {
+    // Peel blocks of assertions with pure conditions
```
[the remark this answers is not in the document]
xerpi: Actually, I copied the code from `CodeGen_LLVM::visit(const Block *op)`, leaving this as a TODO for later.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,565 @@
...
+        mlir::arith::CmpIPredicate predicate = op->a.type().is_int() ? mlir::arith::CmpIPredicate::sge :
+                                                                       mlir::arith::CmpIPredicate::uge;
+        value = builder.create<mlir::arith::CmpIOp>(predicate, codegen(op->a), codegen(op->b));
+    } else if (op->a.type().is_float()) {
+        value = builder.create<mlir::arith::CmpFOp>(mlir::arith::CmpFPredicate::OGE, codegen(op->a), codegen(op->b));
+    }
+}
+
+void CodeGen_MLIR::Visitor::visit(const And *op) {
+    value = builder.create<mlir::arith::AndIOp>(codegen(NE::make(op->a, make_zero(op->a.type()))),
```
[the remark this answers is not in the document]
xerpi: Indeed.
For reference: in Halide `bool` types seem to be `UInt` with `bits() == 1`.

xerpi:
> First round of feedback.
> 
> In general, there seem to be too many things missing for this to be able to work. Are you managing to get something useful out of this? Perhaps with `no_asserts` ?

Thank you very much for your review, @mcourteaux!

I’ve addressed some of your feedback in the latest changes.
Regarding the minimal state of the MLIR support, I’d like to provide some context:
Initially, about two years ago, this MLIR codegen was designed to lower Halide IR to RTL code using Calyx as an intermediate step (see calyxir/calyx#1585). At that time, Calyx had limited support for floating-point operations, vector types, and lowering of SCF dialect constructs like for loops, which is why I didn’t implement features like variable-size allocations. Recently, the Calyx team has made significant progress in supporting the lowering of higher-level MLIR dialects such as arith, scf, and memref, which positions this MLIR codegen much better for RTL generation.

Additionally, I’ve introduced Xilinx XRT runtime support in a separate PR (#7668). This way, we can launch pre-compiled kernels targeting a Xilinx FPGA directly from Halide.

I've also updated this PR's description to include a minimal code example (which disables asserts, indeed).

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,565 @@
...
+void CodeGen_MLIR::Visitor::visit(const AssertStmt *op) {
+    internal_error << "Unimplemented\n";
+}
+
+void CodeGen_MLIR::Visitor::visit(const ProducerConsumer *op) {
+    codegen(op->body);
+}
+
+void CodeGen_MLIR::Visitor::visit(const For *op) {
+    mlir::Value min = codegen(op->min);
```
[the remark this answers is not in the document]
xerpi: In theory, the code supports `min` being any value, so do you think we can leave this as is?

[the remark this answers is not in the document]
mcourteaux: Yes actually you can leave it.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    mlir_module.print(output);
+}
+
+void CodeGen_MLIR::compile_func(mlir::ImplicitLocOpBuilder &builder, const LoweredFunc &func) {
+    mlir::SmallVector<mlir::Type> inputs;
+    mlir::SmallVector<mlir::Type> results;
+    mlir::SmallVector<mlir::NamedAttribute> funcAttrs;
+    mlir::SmallVector<mlir::DictionaryAttr> funcArgAttrs;
+
+    for (const auto &arg : func.args)
```
mcourteaux, line 48: Halide code style does not allow implicit braces. Please at { }

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    mlir::SmallVector<mlir::Attribute> indicesAttrs(op->lanes);
+    for (int i = 0; i < op->lanes; i++)
+        indicesAttrs[i] = mlir::IntegerAttr::get(elementType, i);
+
+    mlir::DenseElementsAttr indicesDenseAttr = mlir::DenseElementsAttr::get(vectorType, indicesAttrs);
+    mlir::Value indicesConst = builder.create<mlir::arith::ConstantOp>(indicesDenseAttr);
+    mlir::Value splatStride = builder.create<mlir::vector::SplatOp>(vectorType, stride);
+    mlir::Value offsets = builder.create<mlir::arith::MulIOp>(splatStride, indicesConst);
+    mlir::Value splatBase = builder.create<mlir::vector::SplatOp>(vectorType, base);
+    value = builder.create<mlir::arith::AddIOp>(splatBase, offsets);
```
mcourteaux, line 322: Many `Ramp`s that are generated in Halide have constant stride. I am totally unaware of any of the internals of MLIR, but it seems that generating constants instead of MulIOps seems beneficial in that case. Maybe MLIR further optimizes it's IR and this comment is pointless... I don't know.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    value = builder.create<mlir::vector::SplatOp>(mlir_type_of(op->type), codegen(op->value));
+}
+
+void CodeGen_MLIR::Visitor::visit(const Call *op) {
+    if (op->is_intrinsic(Call::bitwise_and)) {
+        value = builder.create<mlir::arith::AndIOp>(codegen(op->args[0]), codegen(op->args[1]));
+    } else if (op->is_intrinsic(Call::shift_left)) {
+        value = builder.create<mlir::arith::ShLIOp>(codegen(op->args[0]), codegen(op->args[1]));
+    } else if (op->is_intrinsic(Call::shift_right)) {
+        if (op->type.is_int())
```
mcourteaux, line 335: { }

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+            value = builder.create<mlir::arith::ShRSIOp>(codegen(op->args[0]), codegen(op->args[1]));
+        else
+            value = builder.create<mlir::arith::ShRUIOp>(codegen(op->args[0]), codegen(op->args[1]));
+    } else if (op->is_intrinsic(Call::widen_right_mul)) {
+        mlir::Value a = codegen(op->args[0]);
+        mlir::Value b = codegen(op->args[1]);
+        mlir::Type widen_type = mlir_type_of(op->type);
+        if (op->type.is_int())
```
mcourteaux, line 343: { }

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+            b = builder.create<mlir::arith::ExtSIOp>(widen_type, b);
+        else
+            b = builder.create<mlir::arith::ExtUIOp>(widen_type, b);
+        value = builder.create<mlir::arith::MulIOp>(a, b);
+    } else if (op->name == Call::buffer_get_host) {
+        value = codegen(op->args[0]);
+    } else if (op->name == Call::buffer_get_min) {
+        mlir::Type type = mlir_type_of(op->type);
+        value = builder.create<mlir::arith::ConstantOp>(type, builder.getIntegerAttr(type, 0));
```
mcourteaux, line 352: This seems wrong. Buffer bounds minima are not guaranteed to be zero.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+}
+
+void CodeGen_MLIR::Visitor::visit(const For *op) {
+    mlir::Value min = codegen(op->min);
+    mlir::Value max = builder.create<mlir::arith::AddIOp>(min, codegen(op->extent));
+    mlir::Value lb = builder.create<mlir::arith::IndexCastOp>(builder.getIndexType(), min);
+    mlir::Value ub = builder.create<mlir::arith::IndexCastOp>(builder.getIndexType(), max);
+    mlir::Value step = builder.create<mlir::arith::ConstantIndexOp>(1);
+
+    mlir::scf::ForOp forOp = builder.create<mlir::scf::ForOp>(lb, ub, step);
```
mcourteaux, line 392: Looks correct, but adding a comment that scf::ForOp is over a half-open interval: lowerbound inclusive, upper bound exclusive, would be good for documentation.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    if (op->value.type().is_scalar()) {
+        index = codegen(op->index);
+    } else if (Expr ramp_base = strided_ramp_base(op->index); ramp_base.defined()) {
+        index = codegen(ramp_base);
+    } else {
+        internal_error << "Unsupported Store: " << Stmt(op);
+    }
+
+    index = builder.create<mlir::arith::IndexCastOp>(builder.getIndexType(), index);
+    if (op->value.type().is_scalar())
```
mcourteaux, line 417: { }

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    } else {
+        codegen(op->first);
+        codegen(op->rest);
+    }
+}
+
+void CodeGen_MLIR::Visitor::visit(const IfThenElse *op) {
+    mlir::scf::IfOp ifOp = builder.create<mlir::scf::IfOp>(codegen(op->condition), op->else_case.defined());
+    {
+        mlir::OpBuilder::InsertionGuard guard(builder);
```
mcourteaux, line 469: I'm not convinced / It doesn't seem evident that restoring the insertion point after the codegen for the then or else branches are done is correct. Can you explain why this is the case, or perhaps point to a bit of documentation in the form of a in-code comment?

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    if (op->else_case.defined()) {
+        mlir::OpBuilder::InsertionGuard guard(builder);
+        builder.setInsertionPointToStart(ifOp.elseBlock());
+        codegen(op->else_case);
+    }
+}
+
+void CodeGen_MLIR::Visitor::visit(const Evaluate *op) {
+    codegen(op->value);
```
mcourteaux, line 482: Will MLIR respect the generated code if the result is discarded? I would assume so, but asking to double check.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    internal_error << "Unimplemented";
+}
+
+void CodeGen_MLIR::Visitor::visit(const Atomic *op) {
+    internal_error << "Unimplemented";
+}
+
+void CodeGen_MLIR::Visitor::visit(const HoistedStorage *op) {
+    internal_error << "Unimplemented";
+}
```
mcourteaux, line 513: Please specify for all of these internal errors what is unimplemented. Knowing what the problem is would require hopping into a debugger, or matching line numbers in the error with the source file. Not user friendly.

src/CodeGen_MLIR.h
```diff
@@ -0,0 +1,133 @@
...
+#define HALIDE_CODEGEN_MLIR_H
+
+/** \file
+ * Defines the code-generator for producing MLIR code
+ */
+
+#include "IRVisitor.h"
+#include "Scope.h"
+
+#include <mlir/Dialect/Arith/IR/Arith.h>
```
mcourteaux, line 11: Includes like these in a public header are avoided in Halide. This CodeGen class does typically not need its internals exposed publicly. Consider moving the class definition to the cpp file. For reference, study `CodeGen_PTX_Dev` as an example.

cmake/FindHalide_LLVM.cmake
```diff
@@ -161,4 +161,18 @@ if (Halide_LLVM_FOUND)
             endif ()
         endif ()
     endforeach ()
+
+    find_package(MLIR CONFIG HINTS "${LLVM_INSTALL_PREFIX}" "${LLVM_DIR}/../mlir" "${LLVM_DIR}/../lib/cmake/mlir")
```
mcourteaux, line 165: I think, for making this mergeable, we should have MLIR as a dependency be optional. Right now, none of the build bots make LLVM builds that have MLIR enabled. @alexreinking or @abadams might now how to make LLVM builds with MLIR enabled on the bots. Either way, for a broader audience, I think this should be optional for now. That would require a few `#if` guards in the code unfortunately.

src/CodeGen_MLIR.cpp
```diff
@@ -0,0 +1,541 @@
...
+    mlir_context.loadDialect<mlir::arith::ArithDialect>();
+    mlir_context.loadDialect<mlir::func::FuncDialect>();
+    mlir_context.loadDialect<mlir::memref::MemRefDialect>();
+    mlir_context.loadDialect<mlir::scf::SCFDialect>();
+    mlir_context.loadDialect<mlir::vector::VectorDialect>();
+}
+
+void CodeGen_MLIR::compile(const Module &module) {
+    mlir::LocationAttr loc = mlir::UnknownLoc::get(&mlir_context);
+    mlir::ModuleOp mlir_module = mlir::ModuleOp::create(loc, module.name());
```
mcourteaux, line 30: How would the MLIR builder/compiler know what is the target? Processor? Instruction Set Extensions? My guess is that the MLIR is target-agnostic. However, much of the Halide lowering passes are making target-specific decisions. I'm not sure what to think about this.

Issue #7524, opened 2023-04-19
Build fails with PTX backend disabled


alexreinking:
The Anderson autoscheduler appears to require the PTX backend of Halide to be available in order to build.

```
[3948/4047] Generating anderson2021_included_schedule_file.h, anderson2021_included_schedule_file.o, anderson2021_included_schedule_file.registration.cpp
FAILED: test/autoschedulers/anderson2021/anderson2021_included_schedule_file.h test/autoschedulers/anderson2021/anderson2021_included_schedule_file.o test/autoschedulers/anderson2021/anderson2021_included_schedule_file.registration.cpp /local/mnt2/workspace/areinkin/dev/Halide/build/debian-release/test/autoschedulers/anderson2021/anderson2021_included_schedule_file.h /local/mnt2/workspace/areinkin/dev/Halide/build/debian-release/test/autoschedulers/anderson2021/anderson2021_included_schedule_file.o /local/mnt2/workspace/areinkin/dev/Halide/build/debian-release/test/autoschedulers/anderson2021/anderson2021_included_schedule_file.registration.cpp 
cd /local/mnt2/workspace/areinkin/dev/Halide/build/debian-release/test/autoschedulers/anderson2021 && /local/mnt2/workspace/areinkin/dev/Halide/build/debian-release/test/autoschedulers/anderson2021/anderson2021_included_schedule_file.generator -n anderson2021_included_schedule_file -d 0 -g included_schedule_file -f anderson2021_included_schedule_file -e c_header,object,registration -p /local/mnt2/workspace/areinkin/dev/Halide/build/debian-release/src/autoschedulers/anderson2021/libautoschedule_anderson2021.so -o . target=x86-64-linux-cuda-cuda_capability_70-no_runtime autoscheduler=Anderson2021
Unhandled exception: Error: PTX not enabled for this build of Halide.
```

abadams:
It's a CUDA-specific autoscheduler, so ideally it should just skip building it if PTX isn't available.

PR #7593, opened 2023-05-25
[Fix for #7524] Skip tests for anderson2021 if PTX is not enabled

pranavb-ca:
(Fixes https://github.com/halide/Halide/issues/7524) Skip tests for anderson2021 because it is a CUDA-specific autoscheduler.


alexreinking (approved):

pranavb-ca:
Is this ok to merge? The error in mullapudi2016_reorder (https://buildbot.halide-lang.org/master/#/builders/6/builds/26) is unrelated this change.

PR #7597, opened 2023-05-31
Fix operator/ on ModulusRemainder

### LINKED ISSUE #7584: Simplification failure in fuzz/simplify
This may be hard to repro outside of Google until https://github.com/halide/Halide/pull/7566 is resolved, but Google fuzzing reported a failure in the simplifier test (ie, the Exprs don't equal each other), see below. I'm be attaching the fuzzer input file (which can be used to repro this once we are capable of building the fuzzers outside of Google.)

This one looks like a legit bug somewhere, and the Exprs below may be simple enough [no pun intended] to investigate now.

```
Running: /tmp/testcase-5126156915376128
Simplified Expr is not equal() to Original Expr!
Var a = -128
Var b = -128
Var c = -128
Var d = -128
Var e = -128
Original Expr is: (((((x2(ramp(-128, -79, 2))/x4(-1))/x4(-71)) + x4(-128)) + x4(-128)) + x4(-128))
Simplified Expr is: x4(-385)
In vector lane 1, original -> simplified:
   -386 -> -386
   -385 -> -385
```
[testcase-5126156915376128.zip](https://github.com/halide/Halide/files/11510118/testcase-5126156915376128.zip)

abadams:
It wasn't reducing the remainder modulo the modulus, which confused trim_bounds_using_alignment in the simplifier.

Fixes #7584

### DIFF
--- a/src/ModulusRemainder.cpp
+++ b/src/ModulusRemainder.cpp
@@ -453,7 +453,9 @@ ModulusRemainder operator/(const ModulusRemainder &a, const ModulusRemainder &b)
 
     if (b.modulus == 0 && b.remainder != 0) {
         if (mod(a.modulus, b.remainder) == 0) {
-            return {a.modulus / b.remainder, div_imp(a.remainder, b.remainder)};
+            int64_t m = a.modulus / b.remainder;
+            int64_t r = mod(div_imp(a.remainder, b.remainder), m);
+            return {m, r};
         }
     }
 


steven-johnson:
> reducing the remainder modulo the modulus

Quis modulo ipsos modulus?

steven-johnson (approved):
LGTM, but do we need a test?

abadams:
This is an obscure corner case in the simplifier. I would argue that the simplifier fuzz tester is the test and worked as intended (yay for fuzzing).

PR #7603, opened 2023-06-01
Silence `psabi` warnings when compiling C++ generated code

steven-johnson:
Some versions of GCC/Clang emit many of these warnings when compiling in some Intel configurations, and they are useless in this context. Make them go away.


cmake/HalideGeneratorHelpers.cmake
```diff
@@ -414,6 +414,9 @@ function(add_halide_library TARGET)
         set_target_properties("${TARGET}" PROPERTIES
                               POSITION_INDEPENDENT_CODE ON
                               LINKER_LANGUAGE CXX)
+        # Silence many useless warnings in generated C++ code compilation
+        target_compile_options("${TARGET}" PRIVATE
+            $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-psabi>)
```
alexreinking, line 419: This needs to be guarded by `Halide_NO_DEFAULT_FLAGS` in case a newer compiler removes the psabi warning and the flag becomes invalid.
```suggestion
        if (NOT Halide_NO_DEFAULT_FLAGS)
            # Silence many useless warnings in generated C++ code compilation
            target_compile_options("${TARGET}" PRIVATE
                $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-Wno-psabi>)
        endif ()
```

alexreinking (changes requested):
Nit about allowing downstreams to control which non-necessary flags we inject.

steven-johnson:
PTAL

alexreinking (approved):
LGTM

PR #7604, opened 2023-06-01
Get the ASAN toolchain working again

steven-johnson:
Various fixes to enable ASAN to finally work (linux x64 only).

Note that this found ~several~ one real ASAN failure in the Anderson2021 autoscheduler tests, which is *not* fixed yet; I'll fix thus in a subsequent PR.


alexreinking:
I'm going to try this locally before doing a full review. There are a few things here that are raising alarm bells for certain toolchain configurations.

alexreinking:
When building with clang 16, I get the following error:
```
[59/1220] Linking CXX executable test/correctness/correctness_float16_t
FAILED: test/correctness/correctness_float16_t 
: && /local/mnt2/workspace/areinkin/.local/llvm-16/bin/clang++ -fsanitize=address -fuse-ld=/local/mnt2/workspace/areinkin/.local/llvm-16/bin/ld.lld test/CMakeFiles/Halide_terminate_handler.dir/common/terminate_handler.cpp.o test/correctness/CMakeFiles/correctness_float16_t.dir/float16_t.cpp.o -o test/correctness/correctness_float16_t  -Wl,-rpath,/local/mnt2/workspace/areinkin/dev/Halide/build/clang-16/src  src/libHalide.so.16.0.0 && :
ld.lld: error: undefined symbol: __extendhfsf2
```

Things I did that allowed the build to finish:

1. Link to the Clang compiler runtime via `-rtlib=compiler-rt`. 
2. Set `-march=haswell` in this toolchain (`CMAKE_{C,CXX}_FLAGS_INIT`) so that native fp16 operations are available and the `__extendhfsf2` function isn't emitted in the first place.
3. Check that a simple test program that uses `_Float16` builds; if not, disable the test.

I don't recommend (1)... The default is `libgcc`, which is sensible for distributing binaries; no guarantee downstream systems have compiler-rt built.

I'm partial to a mix of 2 and 3.

steven-johnson:
I thought that we had to explicitly link to `compiler-rt` anyway for ASAN to work properly. (Maybe I'm confusing it with something?)

FWIW, I didn't even test with LLVM16 -- my assumption here is that we'll only do ASAN builds with top-of-tree, so as long as it works with LLVM/Clang17+, we're fine.

alexreinking:
Okay, if Clang 17 doesn't need any of this, then I'm cool with the change.

cmake/toolchain.linux-x64-asan.cmake
```diff
@@ -33,3 +33,16 @@ set(CMAKE_CROSSCOMPILING_EMULATOR /usr/bin/env)
...
+        COMMAND ${CMAKE_CXX_COMPILER} "-print-file-name=libclang_rt.asan.so"
+        OUTPUT_VARIABLE Halide_SHARED_ASAN_RUNTIME_LIBRARY
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+endif ()
+
+set(Halide_SHARED_ASAN_RUNTIME_LIBRARY "${Halide_SHARED_ASAN_RUNTIME_LIBRARY}"
+    CACHE FILEPATH "Library to preload when running Python tests.")
+
+set(Halide_SANITIZER_ENV_VARS "LD_PRELOAD=${Halide_SHARED_ASAN_RUNTIME_LIBRARY}")
```
alexreinking, line 48:
```suggestion
set(
    Halide_PYTHON_LAUNCHER
    ${CMAKE_COMMAND} -E env ASAN_OPTIONS=detect_leaks=0 LD_PRELOAD=${Halide_SHARED_ASAN_RUNTIME_LIBRARY}
)
```
As described above.

alexreinking (changes requested):
I'm proposing a less ad-hoc extension point that feels better to send downstream. I can open a PR-to-PR if you agree with the change, but would prefer not to do this work.

steven-johnson:
With your changes merged, is this good to go now, or are more changes needed?

alexreinking (approved):
LGTM now

PR #7605, opened 2023-06-01
Add missing deps for some autoscheduler tests

steven-johnson:
Autoscheduler tests that rely on the relevant shared library being available at runtime need to add a dependency to ensure this is the case.

### DIFF
--- a/test/autoschedulers/li2018/CMakeLists.txt
+++ b/test/autoschedulers/li2018/CMakeLists.txt
@@ -25,7 +25,7 @@ tests(GROUPS li2018 autoschedulers auto_schedule
       SOURCES
       test.cpp
       ARGS $<TARGET_FILE:Halide::Li2018>)
-add_dependencies(li2018_test Li2018)
+add_dependencies(li2018_test Halide::Li2018)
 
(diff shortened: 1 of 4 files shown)


test/autoschedulers/li2018/CMakeLists.txt
```diff
@@ -39,6 +40,7 @@ if (WITH_PYTHON_BINDINGS)
         list(TRANSFORM PYTHONPATH PREPEND "PYTHONPATH=path_list_prepend:")
 
         set_tests_properties(li2018_gradient_autoscheduler_test_py PROPERTIES
+                             DEPENDS Li2018
```
alexreinking, line 43: Is there a test named `Li2018`? Should this be the following?
```suggestion
                             DEPENDS li2018_test
```

alexreinking (changes requested):
Just want to clarify something before approving... the `DEPENDS` test property does not establish dependencies on targets, but on other tests.

steven-johnson:
PTAL

alexreinking (approved):

PR #7608, opened 2023-06-02
[Hexagon] Remove v62 support

aankit-ca:
Remove v62 and make v65 as the default isa version for HVX.
@pranavb-ca @steven-johnson

### DIFF
--- a/src/CodeGen_Hexagon.cpp
+++ b/src/CodeGen_Hexagon.cpp
@@ -54,10 +54,6 @@ protected:
                                          std::vector<Type> arg_types,
                                          int flags);
 
-    int is_hvx_v65_or_later() const {
-        return (isa_version >= 65);
-    }
-
     using CodeGen_Posix::visit;
 
     /** Nodes for which we want to emit specific hexagon intrinsics */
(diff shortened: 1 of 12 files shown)


src/CodeGen_Hexagon.cpp
```diff
@@ -54,10 +54,6 @@ class CodeGen_Hexagon : public CodeGen_Posix {
                                          std::vector<Type> arg_types,
                                          int flags);
 
-    int is_hvx_v65_or_later() const {
```
pranavb-ca, line 57: @aankit-ca  - where is ``hvx_v66`` needed? I dont see any use of anything that is available at v66 but not v65, but I may be mistaken.

steven-johnson:
I need to do some checking regarding Google's internal usage here -- not sure if we still require support for the earlier clients (in which case we would either need to do some internal patching to maintain the old codepaths, or considering deferring this change.)

steven-johnson:
I'm a little concerned that the approach here makes it nonobvious that the baseline is generating... previously, if you just specified `hvx` you'd get v60, whereas after this, the same build commands would quietly give you v65. I wonder if a safer approach might be:

- Remove the `hvx_v62` feature flag entirely
- Keep the `hvx_v65`
- For now, require that any target that has `hvx` set also has `hvx_v65` (or later) set

This makes for a more-verbose target string, but it does ensure that the meaning of a given Halide target string doesn't silently change codegen (although it may fail to compile, e.g. if you try to specify v62, but a loud failure is much better than a quiet change to codegen)

steven-johnson:
After some internal consultation, it seems likely that Google will want to continue to support v60/v62 codegen in Halide for another year or so. 

Can you share some detail on the motivation to remove these codegen options?

PR #7613, opened 2023-06-05
Fix PCH build failures

steven-johnson:
(Harvested from #7604 to land separately)


test/CMakeLists.txt
```diff
@@ -7,6 +7,11 @@ add_executable(_test_internal internal.cpp)
 target_link_libraries(_test_internal PRIVATE Halide::Test)
 target_include_directories(_test_internal PRIVATE "${Halide_SOURCE_DIR}/src")
 target_precompile_headers(_test_internal PRIVATE <Halide.h>)
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
alexreinking, line 10: Can we guard this by Halide_CCACHE_BUILD?

alexreinking (changes requested):

test/CMakeLists.txt
steven-johnson: Done

alexreinking (approved):

PR #7601, opened 2023-06-01
[Hexagon] Fix compilation failures hexagon_remote

aankit-ca:
Fix for:
1. Missing include directory for pthread.h
2. Incorrect function signature for qurt_hvx_lock

PR #7614, opened 2023-06-05
Revert "[Hexagon] Fix compilation failures hexagon_remote"

steven-johnson:
Reverts halide/Halide#7601

I'm reverting this because I can't see how the change to known_symbols.cpp could possibly compile anywhere (and it certainly doesn't for me); the `qurt_hvx_mode_t` enum type isn't defined and can't be forward-declared. 

@aankit-ca, please open a new PR with similar changes, but with instructions on how to properly build and test it so I can verify your fix.

### DIFF
--- a/src/runtime/hexagon_remote/Makefile
+++ b/src/runtime/hexagon_remote/Makefile
@@ -74,7 +74,7 @@ CCFLAGS-arm-32-android = -target armv7a-linux-androideabi21
 CCFLAGS-host := ${CCFLAGS} -I../ -I ${HEXAGON_TOOLS_ROOT}/Tools/include/iss/ -fPIC \
 	-L${HEXAGON_TOOLS_ROOT}/Tools/lib/iss/ -lwrapper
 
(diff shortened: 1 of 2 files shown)


vksnk (approved):

pranavb-ca:
@aankit-ca  is out on vacation @steven-johnson. he'll be back next week, but I think what is happening is a mismatch in the SDK/Hexagon toolchain version. I think Ankit used Hexagon SDK >= 5.x and I am guessing you are using an older version (4.x based?)

steven-johnson:
> @aankit-ca is out on vacation @steven-johnson. he'll be back next week, but I think what is happening is a mismatch in the SDK/Hexagon toolchain version. I think Ankit used Hexagon SDK >= 5.x and I am guessing you are using an older version (4.x based?)

Yes -- that is what the existing Makefile is set up to use, and also what we use internally at Google. 

Migrating to use 5.3 is fine, but if we're going to do so, we really need to update the Makefile that builds hexagon_remote to specify that by default (and of course, to ensure it's tested)

PR #7620, opened 2023-06-08
Change argument/type signature of Call::prefetch (Fixes #4211)

### LINKED ISSUE #4211: Call::prefetch's return type indicates the type being prefetched, which does not match the implementation
This change of #3667: https://github.com/halide/Halide/pull/3667/files#diff-a8d8ceb2b87d97fbb8fe6eea48d2178f broke prefetching. Apparently, the return type of `Call::prefetch` indicates the type of prefetching to perform in some places, e.g. https://github.com/halide/Halide/blob/4b545ddb1d725d850350c7d0b0e58af127661158/src/CodeGen_C.cpp#L2281

But not others, e.g.: https://github.com/halide/Halide/blob/4b545ddb1d725d850350c7d0b0e58af127661158/src/CodeGen_LLVM.cpp#L3054

It seems reasonable-ish for prefetch to be defined this way (it looks kind of like a load), but it is difficult to implement. If we want to keep this behavior, then I think we need to make a special case for this function in CodeGen_LLVM and CodeGen_Hexagon.

For now, I am going to send a PR to fix the breakage by adding an exception for `Call::prefetch` to the new assert, but I'm going to leave this issue open because this seems like it needs improvement.

steven-johnson:
As noted in the issue above, the return type of this Call was used weirdly and wrongly: it wasn't the return type of the intrinsic (which is always int32, and always ignored), but rather, the type of the elements to prefetch. This didn't seem to be causing any obvious errors, but it meant we needed some weird special cases in the code. We now insert another element in the args to carry the type needed (with a value of constant zero).

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -1825,20 +1825,23 @@ void CodeGen_C::visit(const Call *op) {
     } else if (op->is_intrinsic(Call::undef)) {
         user_error << "undef not eliminated before code generation. Please report this as a Halide bug.\n";
     } else if (op->is_intrinsic(Call::prefetch)) {
-        user_assert((op->args.size() == 4) && is_const_one(op->args[2]))
+        internal_assert(op->type == Int(32));
+        user_assert((op->args.size() == 5) && is_const_one(op->args[3]))
             << "Only prefetch of 1 cache line is supported in C backend.\n";
 
-        const Expr &base_address = op->args[0];
-        const Expr &base_offset = op->args[1];
-        // const Expr &extent0 = op->args[2];  // unused
-        // const Expr &stride0 = op->args[3];  // unused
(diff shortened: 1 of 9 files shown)


abadams:
When you say "it wasn't the return type of the intrinsic", what intrinsic are you referring to? 

It makes sense to me that a prefetch expression would have the type of the value being fetched, as it's basically a load, except you only want the side effects of loading, not the actual value.

abadams:
Call::prefetch is lowered to _halide_prefetch, but it isn't _halide_prefetch, so the return types don't need to match. The args are different too. The comment on that earlier PR is "It seems reasonable-ish for prefetch to be defined this way (it looks kind of like a load)"

The special-casing that you're removing is only necessary because the Call::prefetch handler in CodeGen_LLVM.cpp:3304 is wrong (it returns a value of the wrong type). I'll fix it.

PR #7627, opened 2023-06-12
Deprecate OpenGLCompute for Halide 16

shoaibkamil:
This marks OpenGLCompute as deprecated, roughly modeled after the commit that did the same for OpenGL.

### DIFF
--- a/src/JITModule.cpp
+++ b/src/JITModule.cpp
@@ -762,7 +762,7 @@ enum RuntimeKind {
     OpenCL,
     Metal,
     CUDA,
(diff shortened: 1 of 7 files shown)


steven-johnson (approved):

derek-gerstmann (approved):
Looks good to me!

derek-gerstmann:
Okay to merge?

steven-johnson:
LGTM

steven-johnson:
Actually, I think there are Python usages that need deprecating too, I'll follow up

PR #7628, opened 2023-06-12
Alternative fix for #4211

abadams:
Call::Prefetch evaluates to a currently-unspecified value of the prefetched type. Let's just make it zero.


steven-johnson:
We should probably fix Codegen_C while we are at it -- it produces code that wouldn't have compiled before (it assumed __builtin_prefetch() returned a value, which it does not), see the changes in #7620

steven-johnson:
Fix needed in latest commit but LGTM otherwise

steven-johnson (approved):
Failures appear unrelated

PR #7629, opened 2023-06-13
[Xtensa] Adds a pass to lift allocations out of the loop

vksnk:
This pass is only called for Xtensa codegen and is only applied to stack or dynamic allocations in local on-chip memory. Potentially, it could be used on other platforms as well if needed (with hopefully only few adjustments).

This requires two traversals of the IR, which are implemented within a single pass, because they need to share a lot of code. Hopefully, the implementation is not too confusing.

I've tested it extensively on our internal pipelines and see significant performance gains in some cases. Ideally, I would like to add a separate test for it, but since this is Xtensa specific, it's going to be dependent on adding testing infrastructure for Xtensa first (I hope I can return to working on it very soon).

### DIFF
--- a/src/XtensaOptimize.cpp
+++ b/src/XtensaOptimize.cpp
@@ -2063,6 +2063,259 @@ public:
     SimplifySliceConcat() = default;
 };
 
+// Lifts Stack or VTCM allocation up the loop nest if possible.
+//
+// The idea is that we traverse a loop nest and look for allocations which are
+// of the right type (Stack or VTCM) and don't depend on any of the Let's
+// defined in the given loop. If these conditions are met then allocation can
+// be lifted out of the given loop (or potentially out of multiple of loops).
+// The difficulty is that due to loop partitioning there might be a multiple
+// instances of the same allocation in different instance of the loops. If all
+// instances can be lifted then we can have a single allocation outside of the
+// loop (we need to take max ot total extents as an extent for it to make sure
+// all instances are satisfied).
+// However, there might be a situation when only some of the allocations can be
+// lifted, which potentially could lead to duplicate allocation being generated.
+// Consider an example:
+// for yo:
+//   for xo:
+//        for x:
+//           allocate tmp[depends on let's inside of the loop x]
+//        for x:
+//           allocate tmp[doesn't depend on let's inside of the loop x]
+//        for x:
+//           allocate tmp[depends on let's inside of the loop x]
+// in which if we only consider a local context, the tmp allocation can be
+// lifted from the second inner loop, but couldn't be lifted from the first and
+// third loops.
+// In order to handle cases like this, LiftAllocations has two passes:
+// ** During the first pass, for each loop we compute a list of allocations which
+// are *not* allowed to be lifted to this loop level. This is done by finding a
+// *last* loop (i.e we are not able to lift allocation from the parent loop)
+// from which we *can* lift the allocation and then updating all of it's parent
+// loop lists to disallow given allocation.
+// ** In the second pass, we simply push allocations as far as we can up the
+// loop nest, while making sure that they satisfy all of the conditions (
+// including absence in the "disallowed" list).
+class LiftAllocations : public IRMutator {
+    using IRMutator::visit;
+
+    // Index of the current loop.
+    int loop_index = 0;
+    // Is this a first sub-pass of the pass?
+    bool first_run = true;
+
+    struct Allocation {
+        std::string name;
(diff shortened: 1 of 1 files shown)


src/XtensaOptimize.cpp
```diff
@@ -2063,6 +2063,284 @@ class SimplifySliceConcat : public IRGraphMutator {
...
+// including absence in the "disallowed" list).
+class LiftAllocations : public IRMutator {
+    using IRMutator::visit;
+
+    // Index of the current loop.
+    int loop_index = 0;
+    // Is this a first sub-pass of the pass?
+    bool first_run = true;
+
+    struct Allocation {
```
abadams, line 2108: Could uses of this struct be replaced with pointers to the original Allocate node?

src/XtensaOptimize.cpp
```diff
@@ -2063,6 +2063,284 @@ class SimplifySliceConcat : public IRGraphMutator {
...
+              total_extent(std::move(total_extent)),
+              const_allocation_size(const_allocation_size),
+              condition(std::move(condition)),
+              new_expr(std::move(new_expr)),
+              free_function(std::move(free_function)),
+              padding(padding) {
+        }
+    };
+
+    struct Loop {
```
abadams, line 2136: Same comment as above, but with For nodes instead.

src/XtensaOptimize.cpp
```diff
@@ -2063,6 +2063,284 @@ class SimplifySliceConcat : public IRGraphMutator {
...
+
+        return IRMutator::visit(block);
+    }
+
+    Stmt visit(const Allocate *op) override {
+        if (!loops.empty() &&
+            (op->memory_type == MemoryType::Stack || op->memory_type == MemoryType::VTCM) && !op->new_expr.defined() &&
+            (op->constant_allocation_size() == 0 ||
+             // Don't lift small allocations, constant-sized allocations.
+             op->constant_allocation_size() > 512)) {
```
abadams, line 2237: 512 seems somewhat arbitrary. Why not lift all constant-sized allocations? I'm not clear on what the downside is.

abadams:
This is a somewhat tricky loop transformation, so I think it needs a bunch of tests.

src/XtensaOptimize.cpp
vksnk: Done

src/XtensaOptimize.cpp
```diff
@@ -2063,6 +2063,284 @@ class SimplifySliceConcat : public IRGraphMutator {
...
+            debug(3) << "Finished moving allocations : \n"
+                     << ""
+                     << "\n";
+        }
+        loops.pop_back();
+        return for_op;
+    }
+
+    Stmt visit(const Block *block) override {
+        if (const Free *maybe_free = block->first.as<Free>()) {
```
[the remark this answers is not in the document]
vksnk: Yes, this is safe, added a comment about it.

src/XtensaOptimize.cpp
```diff
@@ -2063,6 +2063,284 @@ class SimplifySliceConcat : public IRGraphMutator {
...
+                             << " at " << loops[ix].name << "\n";
+                    blocked_allocations[loops[ix].loop_index].insert(op->name);
+                }
+            } else {
+                if (loop_to_lift_to < loops.size()) {
+                    // Combine extents, so we can take a maximum in case there are
+                    // multiple allocations.
+                    Expr total_extent = 1;
+                    for (const auto &e : op->extents) {
+                        total_extent *= e;
```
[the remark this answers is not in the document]
vksnk: Good point, added a comment about it.

src/XtensaOptimize.cpp
vksnk: I did it for For node, but in this case it feels a bit more awkward because the original node is removed.

vksnk:
> This is a somewhat tricky loop transformation, so I think it needs a bunch of tests.

It seems it's quite difficult to add an isolated test for the pass (especially, if it's called inside of the codegen). I guess one way to do some testing is to move this pass earlier in the lowering pipeline, add a custom pass to count Allocate nodes to make sure that number of allocs is as expected. Any other ideas?

abadams:
Any tests that trigger it at all would be good, even just end-to-end ones that check the output values are correct.

steven-johnson:
Where does this PR stand?

vksnk:
Closing in favor of #7915 .

PR #7636, opened 2023-06-14
Bump Halide version to 17.0.0 in main

derek-gerstmann:


steven-johnson:
...should we do this *after* we have created the `release/16.x` branch?

derek-gerstmann:
> ...should we do this _after_ we have created the `release/16.x` branch?

Yes!  I already created the `release/16x` branch right before this PR.

steven-johnson (approved):
Might be things in README etc that need attention?

derek-gerstmann:
> Might be things in README etc that need attention?

I checked, but there's not any Halide version numbers in the general README.md.  Perhaps the platform and toolchain section for tested configs and supported versions of OS-es, minimum LLVM version(s), etc might need a review?

shoaibkamil:
> Perhaps the platform and toolchain section for tested configs and supported versions of OS-es, minimum LLVM version(s), etc might need a review?

Yup-- basically we need to change all references to LLVM versions to bump up the minimum supported version (which will now be LLVM-15) and the supported versions (LLVM-15, LLVM-16, LLVM-17).

run-clang-tidy.sh
```diff
@@ -8,23 +8,23 @@ ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 FIX=$1
 
-# We are currently standardized on using LLVM/Clang14 for this script.
+# We are currently standardized on using LLVM/Clang15 for this script.
```
steven-johnson, line 11: We need to actually try clang-tidy-15 and clang-format-15 before we commit to this -- e.g., if the new version changes formatting in an unacceptable way. Let me do some experimenting.

steven-johnson: OK, the new clang-format looks good (just a few minor changes, all of which are good). clang-tidy-15 doesn't work at all, however :-/

steven-johnson: OK, let me suggest that you revert the changes to these two scripts (but keep the rest); I'm going to try to land updates to these scripts as a separate PR (since there is some other stuff to update too)

steven-johnson (approved):
LGTM if you revert the changes to the format and tidy scripts

PR #7644, opened 2023-06-19
[xtensa] widen ops, convert, division + abs, gather_load improvements

Aelphy:
Adopted changed from Cadence


src/XtensaOptimize.cpp
```diff
@@ -952,11 +990,18 @@ class MatchXtensaPatterns : public IRGraphMutator {
...
+
+            if (op->type.bits() == 16) {
+                return Call::make(op->type, "halide_xtensa_absd_i16",
+                                  {mutate(op->args[0]), mutate(op->args[1])},
+                                  Call::PureExtern);
+            } else if (op->type.bits() == 8) {
+                return Call::make(op->type, "halide_xtensa_absd_u8",
+                                  {mutate(op->args[0]), mutate(op->args[1])},
+                                  Call::PureExtern);
+            }
```
steven-johnson, line 1004: What if neither is true -- does the fall-through case handle it correctly? Or should this produce a compilation error?

steven-johnson (commented):
LGTM but should probably get @vksnk approval too unless this is urgent

vksnk (approved):

PR #7646, opened 2023-06-21
Add a compositing example app

abadams:
This PR adds a new app that does Porter-Duff compositing with a handful of operators. The purpose of the app is to demonstrate two things:

- How to write a mini-interpreter in Halide that uses a bytecode input to switch between different operations.
- Some fixed-point math patterns, though unfortunately for this app float is actually faster on CPUs.

It also makes two changes to the Mullapudi autoscheduler. It silences a warning about some fixed point intrinsics, and it fixes a bug where the autoscheduler was not considering all Exprs associated with a Func, just the RHS, which was making it miss the ops input buffer, which is only ever referenced in an RDom predicate.


steven-johnson:
Lots of legit-looking compile errors

apps/compositing/compositing_generator.cpp
```diff
@@ -0,0 +1,199 @@
...
+// that ingests some byte code to determine what to do. It also demonstrates
+// some fixed-point math patterns useful in compositing.
+class Compositing : public Halide::Generator<Compositing> {
+    Var x{"x"}, y{"y"}, c{"c"};
+
+    static constexpr int num_layers = 6;
+    static constexpr int num_blend_modes = 5;
+
+public:
+    Input<Buffer<uint8_t, 3>[num_layers]> layer_rgba { "layer_rgba" };
```
zvookin, line 16: Comments describing these arguments might be useful.

zvookin:
Might make sense to add one liners to the PR description for the change to RegionCosts and the Mullapudi autoscheduler. (In case someone is looking for info to track down a change in behavior in the future.)

steven-johnson:
C:\build_bot\worker\halide-testbranch-main-llvm17-x86-64-windows-cmake\halide-source\apps\compositing\process.cpp(37): error C2065: 'M_PI': undeclared identifier

steven-johnson (approved):
LGTM pending green, but I'd kinda prefer that the changes to Halide proper get landed separately from the app, for future maintenance reasons -- and do any of those changes need tests?

apps/compositing/process.cpp
```diff
@@ -1,4 +1,5 @@
 #include <chrono>
+#include <cmath>
```
steven-johnson, line 2: Still won't work: https://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio

(╯°□°)╯︵ ┻━┻

abadams:
Failures unrelated. It looks like vulkan is frequently failing to allocate device memory.

steven-johnson:
> Failures unrelated. It looks like vulkan is frequently failing to allocate device memory.

Tagging @derek-gerstmann

PR #7648, opened 2023-06-21
Fix RISCV codegen for top-of-tree LLVM

steven-johnson:
Also add a warning if you try to codegen with older versions of LLVM: many intrinsics have changed in ways that are hard to deal with both ways, and trying to support both would be painful and of dubious value.

### DIFF
--- a/src/CodeGen_RISCV.cpp
+++ b/src/CodeGen_RISCV.cpp
@@ -144,6 +144,9 @@ CodeGen_RISCV::CodeGen_RISCV(const Target &t)
     user_assert(native_vector_bits() > 0) << "No vector_bits was specified for RISCV codegen; "
                                           << "this is almost certainly a mistake. You should add -rvv-vector_bits_N "
                                           << "to your Target string, where N is the SIMD width in bits (e.g. 128).";
+#if LLVM_VERSION < 170
+    user_warning << "RISCV codegen is only tested with LLVM 17.0 or later; it is unlikely to work well with earlier versions of LLVM.\n";
+#endif
 }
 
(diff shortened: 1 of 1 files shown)


src/CodeGen_RISCV.cpp
```diff
@@ -355,30 +364,24 @@ bool CodeGen_RISCV::call_riscv_vector_intrinsic(const RISCVIntrinsic &intrin, co
...
 
+    // TODO: Should handle intrinsics other than binary operators.
+    // Call the LLVM intrinsic.
+    int actual_lanes = op->type.lanes();
+    llvm::Constant *actual_vlen = llvm::ConstantInt::get(xlen_type, actual_lanes);
+
     // Set vector fixed-point rounding flag if needed for intrinsic.
-    bool round_down = intrin.flags & RoundDown;
-    bool round_up = intrin.flags & RoundUp;
     if (round_down || round_up) {
```
zvookin, line 373: I'd probably move to building the vector of arguments and doing one call to the builder, but it's your call.

zvookin (approved):

zvookin:
Also, we should consider just not enabling RISC V, by default, in the build if LLVM is older than 17. Presuming that's not extremely difficult to do.

steven-johnson:
> Also, we should consider just not enabling RISC V, by default, in the build if LLVM is older than 17. Presuming that's not extremely difficult to do.

See also: https://github.com/halide/Halide/issues/7649

steven-johnson:
I decided to make LLVM16 work too, since we still build against it for the buildbots. (Leaving the warning in though.)

PR #7650, opened 2023-06-21
Default RISCV backend to OFF for LLVM < 17

steven-johnson:
LLVM17 is doing a lot of work on the RISCV backend, and the amount of testing done on Halide's LLVM16-based RISCV codegen is very light. It's been suggested that we should default to not enabling the RISCV backend for LLVM16 and earlier because of this (so that people attempting to use Halide for RISCV won't encounter a possible footgun).

This PR just adds the relevant mechanism; whether or not this is the correct decision is not clear. Discussion welcome.


zvookin (approved):

alexreinking (approved):
The mechanism LGTM. I'm fine with the change philosophically, too, though if the situation is bad enough with LLVM <=16, then I'd suggest just dropping support altogether.

steven-johnson:
> if the situation is bad enough with LLVM <=16

Eh, we don't really know -- it hasn't been tested enough. But at this point, the lion's share of the testing will all be in LLVM17, which has a lot of RISCV changes. So I don't think we want to prevent LLVM16 from being used, but we should make it opt-in.

PR #7652, opened 2023-06-22
Fix broken wasm linker flags for LLVM16

steven-johnson:


derek-gerstmann (approved):
Ah, that explains the WASM test failures

Issue #7611, opened 2023-06-04
CMake Fetch Content


dbonattoj:
It would be nice to be able to import Halide in a C++ project using the CMake package FetchContent,
eg:

```CMakeLists.txt
cmake_minimum_required(VERSION 3.22)
project(HalideTests)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED YES)
set(CMAKE_CXX_EXTENSIONS NO)

include(FetchContent)

# Halide
FetchContent_Declare(halide
                     GIT_REPOSITORY https://github.com/halide/Halide
                     GIT_TAG        v15.0.1)
FetchContent_GetProperties(halide)
if(NOT halide_POPULATED)
    message("Cloning HALIDE")
    FetchContent_Populate(halide)
    add_subdirectory(
    ${halide_SOURCE_DIR}
    ${halide_BINARY_DIR})
endif()
FetchContent_MakeAvailable(halide)

#find_package(Halide REQUIRED)

add_executable(HalideTests 
    "${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp"
)
target_link_libraries(HalideTests PRIVATE Halide::Halide)

```

When doing that, it complains that it doesn’t find LLVM,

I tried to use fetchcontent on llvm (which works perfectly but requires a lot of space and compile time), but Halide still didn’t find it.

alexreinking:
The problem I consistently ran into when trying to make this work during my first CMake rewrite was that this statement,

> I tried to use fetchcontent on llvm (which works perfectly but requires a lot of space and compile time)

is not true.

LLVM's build interface is pretty different from the interface it exposes via its CMake package. Unfortunately, this makes it very difficult to achieve good FetchContent support while also depending on LLVM.

Maybe one day LLVM's build will progress to a point where it can be consistently consumed via either interface, but regardless the user experience of using LLVM via FetchContent will always be so bad that it's hard to prioritize.

I have some more pressing build system improvements for Halide planned that should make it more amenable to FetchContent. We can re-evaluate in a few months once those are done.

alexreinking:
[...]

PR #7658, opened 2023-06-23
Dependency wrangling part 0/N: standard CMake modules

alexreinking:
I'm resuming the work I started in #7300.

Since I've been helping other projects with their CMake builds, and rewriting some others myself, I've come to disagree with the popular approach of using a `dependencies/` folder with `IMPORTED_GLOBAL` targets. I now prefer to structure dependencies such that they _always_ use `find_package` at the narrowest scope possible. This is a prerequisite for decent FetchContent support, which our users request from time to time (e.g. #7611). It also promotes integration with other package management systems, since it's now much easier to hook the `find_package` calls.

I'm breaking this into smaller PRs for ease of review.

---

In this step, I'm moving the `find_package` calls to standard CMake modules to their best locations. This includes Threads, OpenGL, OpenMP, JPEG, and PNG.


steven-johnson (approved):

steven-johnson:
Looks like legit build failures

alexreinking:
> Looks like legit build failures

They are... debugging...

alexreinking:
Wheeee -- this is because CMake's `$<TARGET_NAME_IF_EXISTS:PNG::PNG>` generator expression is evaluated in the target environment of the _consuming_ target when in `INTERFACE_LINK_LIBRARIES`. Yet using plain `PNG::PNG` will trigger a lookup in the _linkee's_ target environment. Conditional dynamic scoping... yay... 🙃 I'm opening an issue with upstream.

In the meantime, moving the `find_package(PNG)`/`JPEG` up to the top level should work around this.

steven-johnson:
> The Vulkan failures seem spurious.

Yes, there seem to still be intermittent Vulkan errors, tagging @derek-gerstmann to investigate

alexreinking:
I think this is good to merge now?

PR #7659, opened 2023-06-23
[Hexagon] - Build the hexagon_remote runtime using cmake

pranavb-ca:
This PR builds the hexagon_remote runtime using CMake. This integrates the build of the hexagon_remote runtime into the Halide build if ``TARGET_HEXAGON`` is true.
* I have tested this locally, but for the bots to test this effectively, we'll need to make changes to the bots themselves so that they export the right paths.
* Once this is merged, we can remove the precompiled binaries that are in the repository at ``src/runtime/hexagon_remote/bin``.
* The ``host``, ``src`` and ``v65`` subdirectories are no longer present because their contents are unique and we wont really be storing binaries in the repository anymore anyway.

Since CMake is such a challenge for cross-compilation, please let me know if my approach is the right one.


alexreinking:
Somehow Halide needs to have a dependency on halide_hexagon_host, right? It doesn't _link_ to it, but building Halide should bring halide_hexagon_host up to date, too. Maybe via `add_dependencies`?

alexreinking:
It's also worth noting that this adds a build-time dependency to the Hexagon SDK. The README should be updated with instructions on how to do that.

alexreinking:
For the parts that are cross-compiled with the Hexagon toolchain and then treated opaquely... I think those should be factored into a separate project, along with a toolchain file that points to the Hexagon clang compilers. It can be added to the primary build via `ExternalProject`. This is safe in this case because no matter the host or target system for libHalide, that code must always be compiled with the Hexagon toolchain.

pranavb-ca:
@alexreinking - Thank you for your review.
I am fine with the ``ExternalProject`` idea. Let me sync up with you offline to discuss this more. (I saw your PM about it and will follow up with you).
Also, the buildbots use the Hexagon Toolchain which they get from the SDK presently. So, the buildbots already use the HexagonSDK and so should have access to it in the context of this PR.

alexreinking:
Superseded by #7671

PR #7661, opened 2023-06-23
Enable the misc-use-anonymous-namespace clang-tidy check

steven-johnson:


tools/binary2cpp.cpp
```diff
@@ -12,10 +12,12 @@
 // Embeds a binary blob (from stdin) in a C++ source array of unsigned
 // chars. Similar to the xxd utility.
 
-static int usage() {
+namespace {
+int usage() {
```
abadams, line 16: Weird that this one was static to begin with.

abadams (approved):

PR #7663, opened 2023-06-26
Update for top-of-tree LLVM changes

steven-johnson:


abadams:
Looks like this was removed in https://reviews.llvm.org/D153679

It's not clear to me how to get the same functionality though. LLVM seems to be going through and deleting a bunch of legacy pass manager stuff, which might make our life difficult in the very near future.

abadams:
From the comment a few lines above:

"so don't worry about this 'legacy' tag until there's any indication that the old APIs start breaking."

Looks like the old APIs have started breaking.

abadams:
Looks like the symbol rewriting pass was added in:

https://github.com/halide/Halide/commit/09d8c2e11f3f3808e538387ca781a0e6b720fc67

It's an old commit. Hard to say if anyone is using the feature that this PR removes. Approving to unbreak the build, but it would be good to leave a TODO in the code to reenable the feature when we switch to the new pass manager.

abadams (approved):

PR #7664, opened 2023-06-27
Experiment in removing ptr->int->ptr casts

steven-johnson:
While looking over `clang-tidy` checks that we had disabled, I came across `performance-no-int-to-ptr` and ended up in a fascinating rabbit hole. TL;DR: doing ptr->int->ptr casting reduces opportunities for compiler optimizations (and that's what this check attempts to warn you about). I had no idea this was even a thing, but https://www.ralfj.de/blog/2020/12/14/provenance.html and https://www.ralfj.de/blog/2022/04/11/provenance-exposed.html provided interesting backstory.

Running clang-tidy with this enabled revealed a lot of places in our runtime code that might run afoul of this... but it's not clear whether any of them are on a critical path that makes this inefficiency critical (except, perhaps, for the bit-tagging we do in `synchronization_common`).

Anyway, I experimentally converted a few places to follow what I think are the right guidelines (and they no longer trigger this check). Not sure if this is worth pursuing any further, but I thought it worthwhile to post here as a Draft for people to take a look at.


src/runtime/HalideBuffer.h
```diff
@@ -710,6 +711,27 @@ class Buffer {
...
+    inline SomeType align_up(SomeType value, size_t alignment) {
+        // Doing pointer -> integer -> pointer conversions can force
+        // the compiler to pessimize optimizations (see e.g.
+        // https://www.ralfj.de/blog/2020/12/14/provenance.html),
+        // so, if our compiler has builtins for this, use them;
+        // if not, fall back to casting as necessary.
+#if defined(__has_builtin) && __has_builtin(__builtin_align_up)
+        return __builtin_align_up(value, alignment);
+#else
+        if constexpr (std::is_pointer<SomeType>::value) {
```
abadams, line 724: How about this implementation for align_up:
```
ptr += ((-(uintptr_t)(ptr)) & (alignment - 1));
```
Curious if that passes the clang-tidy checker. It just adds a computed offset to a pointer, where the offset depends on casting the pointer to an int. It never casts an int to a ptr.

PR #7668, opened 2023-07-01
[Draft] runtime: Add initial Xilinx Runtime Library (XRT) support

xerpi:
To be used in the future when a path to generate `xclbin`  files ([Xilinx FPGA bitstream format](https://docs.xilinx.com/r/en-US/ug1393-vitis-application-acceleration/Building-the-Device-Binary)) from Halide kernels exists.
A possible start to this future path is MR https://github.com/halide/Halide/pull/7587, which adds MLIR generation.
After generating MLIR code, [CIRCT](https://github.com/llvm/circt) could be used to generate RTL code.


antonysigma:
@xerpi I am not the reviewer, but I like the CIRCT/XRT idea. As a former Xilinx-accelerated algorithm developer, the pain is to develop a custom linux kernel driver in the programmable system (PS) that talks to the programmable logic (PL).

One example is the spinoff [Halide-HLS project](https://github.com/jingpu/Halide-HLS), which requires a complete userspace/kernelspace software stack to function: https://github.com/stevenbell/ultrazynqbuilder

Could you please elaborate more on the target use case? Is it for cloud FPGA (e.g. AWS F1), or for embedded FPGA (e.g. Zynq or Versal with Yocto). They have a different architecture.

xerpi:
> @xerpi I am not the reviewer, but I like the CIRCT/XRT idea. As a former Xilinx-accelerated algorithm developer, the pain is to develop a custom linux kernel driver in the programmable system (PS) that talks to the programmable logic (PL).
> 
> One example is the spinoff [Halide-HLS project](https://github.com/jingpu/Halide-HLS), which requires a complete userspace/kernelspace software stack to function: https://github.com/stevenbell/ultrazynqbuilder
> 
> Could you please elaborate more on the target use case? Is it for cloud FPGA (e.g. AWS F1), or for embedded FPGA (e.g. Zynq or Versal with Yocto). They have a different architecture.

Thanks for the feedback!
Indeed, XRT is a very nice too since it provides not only a way to launch kernel in the variety of Xilinx devices but also to manage them in a common way.

XRT can target both PCIe and embedded FPGA devices, therefore both targets are supported. In my particular case I was using an embedded Zynq device.

src/runtime/xrt.cpp
```diff
@@ -0,0 +1,628 @@
...
+    for (unsigned i = 0; i < count; i++) {
+        xrtDeviceHandle device = xrtDeviceOpen(i);
+        debug(user_context) << "XRT: create_xrt_context: xrtDeviceOpen: " << device << "\n";
+        if (device != nullptr) {
+            global_device = device;
+            return halide_error_code_success;
+        }
+    }
+
+    return halide_error_code_gpu_device_error;
```
steven-johnson, line 129: Error handling in the Halide runtime is (alas) hard to get right and easy to get wrong (it's needed an overhaul for a while but so far has not happened).

But, generally speaking, the Golden Rule is:
- If you call `error()` or `halide_error()` in the runtime code, you *must* return a nonzero error code from the Runtime API boundary.
- Similarly, if you return a nonzero error code, you must call `error()` or `halide_error()`. 

This is to allow for the fact that a given runtime environment might define  a `halide_error()` implementation that simply calls `abort()` or `exit()`... or it might define one that does nothing (except perhaps log a message), meaning that the result code is the only way for the caller to check for failure.

src/runtime/xrt.cpp
```diff
@@ -0,0 +1,628 @@
...
+    debug(user_context)
+        << "XRT: halide_xrt_device_malloc (user_context: " << user_context
+        << ", buf: " << buf << ")\n";
+
+    if (buf->device) {
+        return halide_error_code_success;
+    }
+
+    XRTContext context(user_context);
+    if (context.error_code)
```
steven-johnson, line 151: Style nit, here and elsewhere: *always* enclose if clauses with {}, even for single statements. Also, do we need to call `error()` here?

src/runtime/xrt.cpp
```diff
@@ -0,0 +1,628 @@
...
+    debug(user_context)
+        << "XRT: halide_xrt_initialize_kernels (user_context: " << user_context
+        << ", state_ptr: " << state_ptr
+        << ", kernel_name: " << kernel_name << ")\n";
+
+    XRTContext context(user_context);
+    if (context.error_code)
+        return context.error_code;
+
+    snprintf(xclbin_name, sizeof(xclbin_name), "%s.xclbin", kernel_name);
```
steven-johnson, line 380: Here and elsewhere, should check the result code for an error -- IIRC, a negative number returned if buffer too small.

src/runtime/xrt.cpp
```diff
@@ -0,0 +1,628 @@
...
+                    return halide_error_code_generic_error;
+                }
+
+                debug(user_context) << "XRT: halide_xrt_run: "
+                                    << "allocated buffer with size: " << (uint64_t)buf->size
+                                    << " at physical address: " << (void *)xrtBOAddress(buf->handle) << "\n";
+
+                if (buf->copy_to_device_pending) {
+                    debug(user_context) << "  buffer has a copy to device pending.\n";
+                    // memset(buffer->host, 0xFF, buf->size);
```
steven-johnson, line 504: Don't leave in commented-out code -- either delete it, or add a comment about why it's intentionally commented out (eg for debugging)

steven-johnson (commented):
Looks good so far, with some style nits and some details about error handling. How will this be tested?

src/runtime/xrt.cpp
xerpi: I have checked other runtime backends and it seems they don't call `error()` here.

xerpi:
> Looks good so far, with some style nits and some details about error handling. How will this be tested?

Thanks for the code review!

Currently, there's no way to test it (also that's why it's a _Draft_).
This is one end of the flow that needs to be implemented to execute Halide kernels on (Xilinx) FPGAs.
The opposite end of such a flow would be to generate MLIR from Halide IR (https://github.com/halide/Halide/pull/7587), which in turn can be converted to RTL by using CIRCT or to HLS by using projects such as https://github.com/hanchenye/scalehls.
Check https://github.com/cucapra/calyx/issues/1585 for more information.

steven-johnson:
OK, so it's pending on (at least) https://github.com/halide/Halide/pull/7587, but I'm not sure what that one is pending -- left a comment there.

PR #7671, opened 2023-07-07
Build Hexagon runtime components using the Hexagon SDK

alexreinking:
This is a reimplementation of #7659 that uses ExternalProject and the Hexagon SDK-provided toolchain files. This will require relatively less maintenance when porting the build between Hexagon SDK versions.

Notable changes:

- The source files in `src/runtime/hexagon_remote` have been arranged into subdirectories `qurt` and `android` based on which toolchain must compile them. This avoids a horrible use of _recursive CMake_.

To-do:

- [x] Use `SYSTEM` includes for the SDK directories.
- [ ] Decide what to do about the new dependency.
  - Should Hexagon be off by default?
  - Should this part of the build be controlled by a new switch?
  - Should `Halide` depend on `hexagon_runtime`?
- [ ] Update GHA CI (clang-tidy missing Hexagon SDK)
- [ ] Add install rules for these artifacts (if need be)
  - Should/can these be part of our binary distribution?
- [ ] Update or remove the Makefile for this part.
- [ ] Remove precompiled binaries from the repository
- [ ] Update documentation
  - Requires Hexagon SDK 4.3.0.0, specifically.

Future work:

- Port to latest Hexagon SDK


steven-johnson:
This PR seems to be massively broken on the buildbots -- is it ready for review?

src/runtime/hexagon_host.cpp
```diff
@@ -147,7 +147,11 @@ WEAK int init_hexagon_runtime(void *user_context) {
     if (!host_lib) {
         host_lib = halide_load_library("libhalide_hexagon_host.dll");
     }
-
+    if (!host_lib) {
+        // This will now cause a more specific error 'halide_error_code_symbol_not_found' down the line.
+        // So, just print this message and continue on instead of returning a generic error here.
+        error(user_context) << "Hexagon: unable to load libhalide_hexagon_host.so";
```
steven-johnson, line 153: (1) Depending on how you have `halide_error()` configured, this may not continue on at all. 
(2) The golden rule of Halide Runtime code is that every call to `error()` or `halide_error()` *must* *also* return a nonzero error code across the runtime API boundary; otherwise, you get different failure modes in different configurations. 

If you always want to continue from here, use `print()` instead of `error()`.

src/runtime/hexagon_remote/qurt/halide_remote.cpp
```diff
@@ -15,8 +15,8 @@ extern "C" {
 #include "known_symbols.h"
 #include "log.h"
 
-const int stack_alignment = 128;
-const int stack_size = 1024 * 1024;
+// const int stack_alignment = 128;
```
steven-johnson, line 18: Why are these commented out?

steven-johnson (changes requested):

pranavb-ca:
> This PR seems to be massively broken on the buildbots -- is it ready for review?

The PR here needs the an environment variable set
``HEXAGON_SDK_ROOT=/home/halidenightly/Qualcomm/Hexagon_SDK/4.3.0/``
that should make the linux cmake bots pass (or at least go beyond the current point of failure)

src/runtime/hexagon_host.cpp
pranavb-ca: Got it.

alexreinking: @pranavb-ca

pranavb-ca:
> > > This PR seems to be massively broken on the buildbots -- is it ready for review?
> > 
> > 
> > The PR here needs the an environment variable set `HEXAGON_SDK_ROOT=/home/halidenightly/Qualcomm/Hexagon_SDK/4.3.0/` that should make the linux cmake bots pass (or at least go beyond the current point of failure)
> 
> Want me to do that, or do you want to?



> > > This PR seems to be massively broken on the buildbots -- is it ready for review?
> > 
> > 
> > The PR here needs the an environment variable set `HEXAGON_SDK_ROOT=/home/halidenightly/Qualcomm/Hexagon_SDK/4.3.0/` that should make the linux cmake bots pass (or at least go beyond the current point of failure)
> 
> Want me to do that, or do you want to?

This is the right place to look right? https://github.com/halide/build_bot/blob/master/master/master.cfg
There are a couple of more changes needed than just setting ``HEXAGON_SDK_ROOT``. Let me take a stab at it tonight.

pranavb-ca:
> > > > This PR seems to be massively broken on the buildbots -- is it ready for review?
> > > 
> > > 
> > > The PR here needs the an environment variable set `HEXAGON_SDK_ROOT=/home/halidenightly/Qualcomm/Hexagon_SDK/4.3.0/` that should make the linux cmake bots pass (or at least go beyond the current point of failure)
> > 
> > 
> > Want me to do that, or do you want to?
> 
> > > > This PR seems to be massively broken on the buildbots -- is it ready for review?
> > > 
> > > 
> > > The PR here needs the an environment variable set `HEXAGON_SDK_ROOT=/home/halidenightly/Qualcomm/Hexagon_SDK/4.3.0/` that should make the linux cmake bots pass (or at least go beyond the current point of failure)
> > 
> > 
> > Want me to do that, or do you want to?
> 
> This is the right place to look right? https://github.com/halide/build_bot/blob/master/master/master.cfg There are a couple of more changes needed than just setting `HEXAGON_SDK_ROOT`. Let me take a stab at it tonight.

@steven-johnson - PTAL -> https://github.com/halide/build_bot/pull/237

pranavb-ca:
There is a problem here that I hadn't thought of earlier.
our windows & osx and ARM builders build Hexagon (because of the underlying LLVM having built the hexagon backend). As a result ``hexagon_remote`` will get built as per this PR (because ``TARGET_HEXAGON`` will be on). But not all builders have Hexagon related properties, namely ``HL_HEXAGON_TOOLS`` set up thereby meaning we cannot set up ``HEXAGON_SDK_ROOT`` the way things stand now.

For context, see https://github.com/halide/build_bot/pull/237#issuecomment-1631562528

src/runtime/CMakeLists.txt
```diff
@@ -339,3 +339,6 @@ add_library(Halide::Runtime ALIAS Halide_Runtime)
 target_include_directories(Halide_Runtime INTERFACE $<BUILD_INTERFACE:${Halide_BINARY_DIR}/include>)
 set_target_properties(Halide_Runtime PROPERTIES EXPORT_NAME Runtime)
 
+if (TARGET_HEXAGON)
```
pranavb-ca, line 342: Since this condition can be true on builders that do not handle hexagon, maybe add another check to see if HEXAGON_SDK_ROOT is set up in the environment. It is not the most elegant thing to do though.

pranavb-ca: @alexreinking @steven-johnson - Any thoughts on this?

pranavb-ca: A simpler solution was to simply not build the hexagon backend in llvm on builders that don't handle hexagon. I have updated my PR -> https://github.com/halide/build_bot/pull/237

steven-johnson:
I think we're still just waiting on the clang-tidy fixes here

pranavb-ca:
> I think we're still just waiting on the clang-tidy fixes here

Yes, we only need to merge main into this branch and those errors should get resolved.

steven-johnson:
> Yes, we only need to merge main into this branch and those errors should get resolved.

I don't have permission to change your branch, you need to do it

pranavb-ca:
> > Yes, we only need to merge main into this branch and those errors should get resolved.
> 
> I don't have permission to change your branch, you need to do it

Unfortunately, this isn't my branch either. It is Alex's as he was helping me out. I couldn't get in touch with him in the office today. I have sent him a message to merge main into his branch, hopefully he'll be online tomorrow.

steven-johnson:
Is this PR still useful, or has it been subsumed by the other PR for this that landed recently?

steven-johnson:
Is this PR still useful, or has it been subsumed by the other PR for this that landed recently?

pranavb-ca:
> Is this PR still useful, or has it been subsumed by the other PR for this that landed recently?

Yes, this PR should be closed because it is subsumed by https://github.com/halide/Halide/pull/7741  which has now been merged.

PR #7672, opened 2023-07-10
Add a select overload for tuples

abadams:
As discussed in the dev meeting, this makes select work for tuples, in the same way as #7646 did for mux.

Should we deprecated tuple_select?


steven-johnson:
Lots of compilation failures (`select` is ambiguous) ... I'll wait for those to be dealt with before reviewing

steven-johnson:
Failures are unrelated.

steven-johnson:
> Should we deprecated tuple_select?

IMHO, yes, and that should be part of this PR.

steven-johnson:
Python bindings need attention too, I'll deal with that

steven-johnson:
asan failure is unrelated and being worked on elsewhere

steven-johnson (approved):

Issue #7681, opened 2023-07-14
BGU app not 100% correct.


mcourteaux:
I'm toying around with the Bilateral Guided Upsampling demo code, and it seems to be not correct all the time.

Specifically, according to my tests:

 - CPU, manual schedule: wrong
 - CPU, auto schedule: wrong
 - GPU, manual schedule: correct
 - GPU, auto schedule: wrong

All the wrong results are wrong in the exact same way. Using this random picture from a tree: 
https://www.gardeningknowhow.com/wp-content/uploads/2020/12/lonely-japanese-cherry.jpg

Running and saving all the different versions gets me this:

CUDA manual schedule:
![test manual_out](https://github.com/halide/Halide/assets/845012/09c73346-563f-4580-b781-089f4b0068ad)

CUDA auto schedule:
![test auto_out](https://github.com/halide/Halide/assets/845012/9c7360e3-a41e-49d4-9258-1766901ee8fe)

CPU manual schedule:
![test manual_out](https://github.com/halide/Halide/assets/845012/e03bd763-a7d8-4470-bb1b-552406ce5225)

CPU auto schedule:
![test auto_out](https://github.com/halide/Halide/assets/845012/4b356988-a9da-4b77-b711-af4136aeff7a)

**Q1**: Notice the red-blue discoloration in the sky. Could this be a data-clipping issue? The colors seem like overflows or something. Here is another example of this problem in another picture of mine:
![image](https://github.com/halide/Halide/assets/845012/e7adfa62-a15e-4194-a171-d617f843f223)

**Q2**: It's been a while since I was hands-on with Halide, but my experience always told me that, despite Halide promising that the result should be the same, regardless of the schedule (if you don't do the race_condition() things and alike). Still here is another example of where this promise seems broken. Is there a FAQ or guide on why this might happen and what to pay attention to regarding these inconsistencies?

**Q3**: Is there a way to reduce the block artifacts here? Looking at the one that worked (the first: CUDA manual schedule), there is still quite some block artifacts visible. Intuition tells me this is because of a lack of diffusion through the bilateral grid (which should ensure the smoothness if I understand the explanatory video correctly), but the blurring settings are fixed.

abadams:
I suspect what's going on is that for some inputs, BGU is numerically unstable. @jiawen and I tried to regularize it correctly, but I suspect there's still something wrong, and it's trying to invert nearly-singular matrices in unpopulated regions of the grid. 

[...]

PR #7684, opened 2023-07-17
Fixed the regularization for BGU.

mcourteaux:
See discussion in #7681.

This patch simplifies the regularization. The most important takeaway is in the comment of the changed code, and a thought I had: https://github.com/halide/Halide/issues/7681#issuecomment-1638095230 :

 > The only thing I'm not sure about is that I changed it to regularize towards the identity function, instead of "towards the average gain in the cell". The problem I have with "the average gain" is that you also need samples to compute the average gain, which is exactly the problem being solved.

@jiawen @abadams I don't know if you want to validate the result by running the experiment setup of your paper? I guess that will be a more formal test.

### DIFF
--- a/apps/bgu/bgu_generator.cpp
+++ b/apps/bgu/bgu_generator.cpp
@@ -395,39 +395,20 @@ public:
             b(2, 2) = blurx(x, y, z, 20);
             b(3, 2) = blurx(x, y, z, 21);
 
-            // Regularize by pushing the solution towards the average gain
-            // in this cell = (average output luma + eps) / (average input luma + eps).
-            const float lambda = 1e-6f;
-            const float epsilon = 1e-6f;
+            // Regularize it with 1/10th of a sample that pulls the result towards the identity function.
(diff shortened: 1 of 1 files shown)


mcourteaux:
To compare the image I reported originally:

(before) https://user-images.githubusercontent.com/845012/253539155-09c73346-563f-4580-b781-089f4b0068ad.jpg

(after) https://user-images.githubusercontent.com/845012/253946895-792e6941-8cb4-475d-9a6c-95d48857d99c.jpg

Open both and toggle back and forth in two tabs. Result looks identical, except for the regularization artifacts.

abadams:
Your argument about the original regularization makes sense to me. I'd appreciate @jiawen 's take on it though.

steven-johnson:
ASAN failures are unrelated

steven-johnson:
Ready to merge?

abadams (approved):

mcourteaux:
:partying_face:

PR #7686, opened 2023-07-17
Fix quadratic algorithm in simplify_correlated_differences

abadams:
This pass called expr_uses_var in a loop while building up a potentially long let chain. This does a quadratic amount of work in the size of the let chain, which stalled compilation for a particular pathological pipeline I encountered.

This changes it to an eager algorithm that tracks the set of free variables and incrementally grows it instead of revisiting the entire expr for each new let added. It is n log(n) in the number of lets instead of n^2


steven-johnson:
asan failure should be addressed by https://github.com/halide/Halide/pull/7687

rootjalex (approved):

abadams:
Failures unrelated

PR #7687, opened 2023-07-17
Fix correctness_float16_t for ASAN builds

steven-johnson:
This appears to be a glitch that has to do with changing ABI for float16 across versions of GCC; we build LLVM with gcc-9 on Linux, but the float16 ABI got changed (and unified in gcc12); since ASAN builds use Clang even on linux, there is a hiccup here.

This is an ugly monkey-patch to work around this issue.

### DIFF
--- a/test/correctness/float16_t.cpp
+++ b/test/correctness/float16_t.cpp
@@ -2,6 +2,17 @@
 
 #include <limits>
 
+#ifdef __linux__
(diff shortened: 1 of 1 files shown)


abadams (approved):

steven-johnson:
ASAN is reporting a lot of leaks that need to be tracked down (https://buildbot.halide-lang.org/master/#/builders/146/builds/11) but that should probably be handled separately, landing

PR #7690, opened 2023-07-18
Revert "Fix correctness_float16_t for ASAN builds"

steven-johnson:
Reverts halide/Halide#7687

Local testing makes it appear this fix is (1) incorrect and (2) no longer needed, opening for retesting on the buildbots.

### DIFF
--- a/test/correctness/float16_t.cpp
+++ b/test/correctness/float16_t.cpp
@@ -2,17 +2,6 @@
 
 #include <limits>
 
-#ifdef __linux__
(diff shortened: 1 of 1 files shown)


shoaibkamil (approved):
LGTM pending passing the buildbots.

steven-johnson:
EDIT: looks like the issue here is that the fix in question is needed for gcc-9 (which is what the buildbots use) but I have gcc-12 locally (and it's hard to install earlier versions, because, uh, reasons), but this is definitely not the right fix

PR #7692, opened 2023-07-19
Don't link to cudart or opencl library. These are loaded dynamically as needed.

abadams:
I found these lines in test/generator/CMakeLists.txt and I don't think they should be necessary because nothing in that folder uses any cuda/opencl APIs except the runtime, which loads those libraries dynamically and resolves things via dlsym. I figured the simplest way to see if they're necessary is to open a PR and see what breaks.

AFAIK the only thing that actually needs any cuda libs linked at compile-time is apps/cuda_mat_mul


abadams:
This failure is why I ran across these lines: https://buildbot.halide-lang.org/master/#/builders/146/builds/21

steven-johnson:
Compilation failures

abadams:
The thing I was missing is that while nothing in test/generator includes cuda.h, a file in test/common does. The tests in question are for checking if GPU contexts generated using the standard APIs (i.e. not generated using our own runtime) can be used by overriding the acquire/release context runtime functions.

PR #7695, opened 2023-07-20
Default WITH_TEST_FUZZ to OFF

steven-johnson:
Just because our compiler supports fuzzing doesn't mean we want to build the fuzz tests, because they won't really build properly without the right preset specified.

(This will be followed up with a change to the buildbot to set WITH_TEST_FUZZ to ON for fuzz tests)

attn: @TH3CHARLie

### DIFF
--- a/src/CodeGen_ARM.cpp
+++ b/src/CodeGen_ARM.cpp
@@ -1053,7 +1053,9 @@ void CodeGen_ARM::visit(const Store *op) {
         std::ostringstream instr;
         vector<llvm::Type *> arg_types;
         llvm::Type *intrin_llvm_type = llvm_type_of(intrin_type);
-#if LLVM_VERSION >= 150
+#if LLVM_VERSION >= 170
+        const bool is_opaque = true;
+#elif LLVM_VERSION >= 150
         const bool is_opaque = llvm::PointerType::get(intrin_llvm_type, 0)->isOpaque();
 #else
         const bool is_opaque = false;
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -94,8 +94,12 @@ else ()
     message(VERBOSE "Compiler supports libfuzzer sanitizer.")
 endif ()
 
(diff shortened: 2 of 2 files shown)


TH3CHARLie:
want to add to the reasons for not enabling them, with fuzzer linked, whenever you call `f.realize` you will get something like this:
```
Internal Error at /home/xuanda/dev/Serializer/Halide/src/JITModule.cpp:226 triggered by user code at : Condition failed: addr: Failed to materialize symbols: { (main, { blur_y_metadata, blur_y, $.blur_y.__inits.0, __orc_init_func.blur_y, sancov.module_ctor_8bit_counters, blur_y_argv }) }
```
Not sure about the exact reason, but most likely because libfuzz will insert some kind of `main` into the code already.

steven-johnson:
Monday morning review ping

alexreinking (changes requested):

steven-johnson:
> What doesn't work, exactly? The idea was that HAS_FUZZ_FLAGS was a sufficient condition on the environment for the fuzz tests to be expected to work. What is the true condition?

The fuzzer tests *must* be built using the `linux-x64-fuzzer` preset

alexreinking:
> The fuzzer tests _must_ be built using the `linux-x64-fuzzer` preset

That's kind of my question... why? What does the preset do that's special? The only non-standard thing I see is that it uses static Halide (so `BUILD_SHARED_LIBS=OFF`),

TH3CHARLie:
> > The fuzzer tests _must_ be built using the `linux-x64-fuzzer` preset
> 
> That's kind of my question... why? What does the preset do that's special? The only non-standard thing I see is that it uses static Halide (so `BUILD_SHARED_LIBS=OFF`),

I believe it's the new flag making it special:https://github.com/halide/Halide/blob/fd9bfc835f658d8f9d64cff59c64a4503298030d/cmake/toolchain.linux-x64-fuzzer.cmake#L20-L25

abadams:
e.g. it injects its own int main!

alexreinking (approved):

PR #7697, opened 2023-07-21
Add a warning if a Generator declares any Outputs before the final Input (Fixes #7669)

### LINKED ISSUE #7669: Python Generators rearrange input and output arguments silently
It appears that python generators will rearrange inputs and outputs so that in the generated function call all input arguments come first, then output arguments, even if this is not the order defined in the generator class.

This may be intended behavior, but I can't find it called out anywhere in the examples or documentation, and it seems to me like an unintuitive result. It seems like either an addition to the documentation or a more descriptive error message would be a helpful addition. I think this would be caught more readily in a C++ setting because of static typing, Python tools can't easily catch this problem because it's dynamically typed. It's also not immediately obvious to me how to inspect the code generated by a generator, but I've only spent a few minutes looking.

Minimal example:
```
import numpy as np
import halide as hl

x = hl.Var("x")
y = hl.Var("y")

@hl.generator(name="testgen")
class testgen(hl.Generator):
    output = hl.OutputBuffer(hl.UInt(8), 2)  # output defined first
    offset = hl.InputScalar(hl.UInt(8))  # input defined second
    
    def generate(self):
        self.output[x,y] = self.offset


if __name__ == "__main__":
    inp = 3
    out = np.empty((256,256), dtype=np.uint8)
    t = hl.Target("host")
    with hl.GeneratorContext(t):
        gen = testgen()
        
        test_filter = gen.compile_to_callable()
        
        test_filter(inp, out)  # <- this works
        
        test_filter(out, inp)  # <- this fails

```

result:
```
  File ...example.py:26
    test_filter(out, 3)  # <- this fails

HalideError: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

steven-johnson:
See https://github.com/halide/Halide/issues/7669 for details

attn @Copper280z

### DIFF
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1237,7 +1237,8 @@ GeneratorParamInfo::GeneratorParamInfo(GeneratorBase *generator, const size_t si
 
     const char *const io_order_warning = "Generators will always produce code that orders all Inputs before all Outputs; "
                                          "this Generator declares the Inputs and Outputs in a different order, so the calling convention may not be as expected. "
-                                         "(A future version of Halide will make this illegal, and require all Inputs to be declared before all Outputs.)";
+                                         "A future version of Halide will make this illegal, and require all Inputs to be declared before all Outputs. "
+                                         "(You can avoid this requirement by overriding Generator::allow_out_of_order_inputs_and_outputs().)";
 
     bool outputs_seen = false;
     auto vf = ObjectInstanceRegistry::instances_in_range(generator, size);
--- a/src/Generator.cpp
+++ b/src/Generator.cpp
@@ -1218,10 +1218,6 @@ GeneratorBase::~GeneratorBase() {
 }
 
 GeneratorParamInfo::GeneratorParamInfo(GeneratorBase *generator, const size_t size) {
-    std::vector<void *> vf = ObjectInstanceRegistry::instances_in_range(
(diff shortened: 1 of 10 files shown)


src/Generator.cpp
```diff
@@ -1239,45 +1235,67 @@ GeneratorParamInfo::GeneratorParamInfo(GeneratorBase *generator, const size_t si
...
-    for (auto *v : vg) {
-        auto *param = static_cast<GeneratorParamBase *>(v);
-        internal_assert(param != nullptr);
-        user_assert(is_valid_name(param->name())) << "Invalid GeneratorParam name: " << param->name();
-        user_assert(!names.count(param->name())) << "Duplicate GeneratorParam name: " << param->name();
-        names.insert(param->name());
-        internal_assert(param->generator == nullptr || param->generator == generator);
-        param->generator = generator;
-        filter_generator_params.push_back(param);
+    const char *const io_order_warning = "Generators will always produce code that orders all Inputs before all Outputs; "
```
abadams, line 1238: The warning should probably include instructions to override that method if you explicitly want to have outputs declared before inputs

steven-johnson: Not sure if I entirely agree with this -- part of me wants to push back and say that anyone who *really* needs it will figure it out. But I'm probably overthinking it...

abadams (approved):

PR #7700, opened 2023-07-24
Fix leaks caused by self-referential parameter constraints

abadams:
This should fix memory leaks caused by Parameters that have constraints that refer back to the same Parameter.

Note that it *doesn't* fix leaks caused by Parameter A having a constraint that refers to Parameter B which has a constraint that refers back to Parameter A.

### DIFF
--- a/src/Dimension.cpp
+++ b/src/Dimension.cpp
@@ -50,42 +50,42 @@ Expr Dimension::stride() const {
     return Variable::make(Int(32), s.str(), param);
 }
 
-Dimension Dimension::set_extent(Expr extent) {
+Dimension Dimension::set_extent(const Expr &extent) {
     // Propagate constant bounds into estimates as well.
     if (is_const(extent)) {
         param.set_extent_constraint_estimate(d, extent);
     }
-    param.set_extent_constraint(d, std::move(extent));
+    param.set_extent_constraint(d, extent);
     return *this;
 }
 
-Dimension Dimension::set_min(Expr min) {
+Dimension Dimension::set_min(const Expr &min) {
     // Propagate constant bounds into estimates as well.
     if (is_const(min)) {
         param.set_min_constraint_estimate(d, min);
     }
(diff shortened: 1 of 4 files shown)


steven-johnson (approved):
Does this allow us to re-enable leak detection on the bots, or is there more (known) work to be done?

abadams:
There's more work to be done.

steven-johnson:
```
/home/halidenightly/build_bot/worker/halide-testbranch-main-llvm17-x86-64-linux-make/halide-source/src/Parameter.cpp: In function ‘Halide::Expr Halide::Internal::remove_self_references(const Halide::Internal::Parameter&, const Halide::Expr&)’:
/home/halidenightly/build_bot/worker/halide-testbranch-main-llvm17-x86-64-linux-make/halide-source/src/Parameter.cpp:206:14: error: ‘virtual Halide::Expr Halide::Internal::remove_self_references(const Halide::Internal::Parameter&, const Halide::Expr&)::RemoveSelfReferences::visit(const Halide::Internal::Variable*)’ can be marked override [-Werror=suggest-override]
  206 |         Expr visit(const Variable *var) {
      |              ^~~~~
/home/halidenightly/build_bot/worker/halide-testbranch-main-llvm17-x86-64-linux-make/halide-source/src/Parameter.cpp: In function ‘Halide::Expr Halide::Internal::restore_self_references(const Halide::Internal::Parameter&, const Halide::Expr&)’:
/home/halidenightly/build_bot/worker/halide-testbranch-main-llvm17-x86-64-linux-make/halide-source/src/Parameter.cpp:229:14: error: ‘virtual Halide::Expr Halide::Internal::restore_self_references(const Halide::Internal::Parameter&, const Halide::Expr&)::RestoreSelfReferences::visit(const Halide::Internal::Variable*)’ can be marked override [-Werror=suggest-override]
  229 |         Expr visit(const Variable *var) {
      |              ^~~~~
```

abadams:
Failures unrelated

PR #7701, opened 2023-07-24
Fix float16 warning for older clangs

abadams:
This routine was missing a float cast when compiled with older clangs. Also removed the warning, because I have now tested it.


steven-johnson (approved):
Thanks :-)

steven-johnson:
Failures are unrelated

PR #7705, opened 2023-07-24
Fix leaks in test/correctness/memoize.cpp

abadams:
This was a little tricky, because it turns out any custom allocator used with memoized allocations needs to be compatible with the vanilla halide_free, because that's what's going to get called when the runtime shuts down.

Edit: The above issue is now resolved by just digging out the malloc/free from the runtime and wrapping them. Making a compatible one just didn't work on windows, as the malloc that gets called from the test seems to be different to the malloc that is resolved by the JIT-compiled runtime.

Built on #7700


steven-johnson (approved):

steven-johnson:
Looks like an error on correctness_memoize on Windows... https://buildbot.halide-lang.org/master/#/builders/148/builds/60

PR #7699, opened 2023-07-24
[Debug only; Do not merge] Debug the ASAN error w/ Anderson2021

antonysigma:
Provide a minimal reproducible test case where the Anderson2021 autoscheduler failed with segmentation fault. Primarily the L2 norm user defined function.

Contact me if you wish to rebase the changes on top of Halide `v16.0.0`.

Expected ASAN error:
```
env LD_LIBRARY_PATH=/home/antony/Projects/ProxImaL/proximal/halide/subprojects/Halide-16.0.0-x86-64-linux/lib /home/antony/Projects/ProxImaL/proximal/halide/build-clang/src/user-problem/solver-generator -o /home/antony/Projects/ProxImaL/proximal/halide/build-clang/src/user-problem -g debug_norm -e static_library,h target=host-cuda-cuda_capability_75 -p autoschedule_anderson2021 autoscheduler=Anderson2021 autoscheduler.parallelism=32
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1236944==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000070 (pc 0x7f00983507ce bp 0x000000000008 sp 0x7ffe56413000 T0)
==1236944==The signal is caused by a READ memory access.
==1236944==Hint: address points to the zero page.
    #0 0x7f00983507ce in Halide::Internal::Autoscheduler::LoopNest::compute_strides(Halide::Internal::Autoscheduler::LoadJacobian const&, int, Halide::Internal::Autoscheduler::FunctionDAG::Node const*, Halide::Internal::IntrusivePtr<Halide::Internal::Autoscheduler::BoundContents const> const&, Halide::Internal::Autoscheduler::ThreadInfo const&, bool) const (/home/antony/Projects/ProxImaL/proximal/halide/subprojects/Halide-16.0.0-x86-64-linux/lib/libautoschedule_anderson2021.so+0x657ce)
[...]

Issue #7606, opened 2023-06-01
ASAN reports out-of-bounds read error in `anderson2021_test_apps_autoscheduler`


steven-johnson:
ASAN is reporting an out-of-bounds read error in SearchSpace.cpp:125; examination reveals that we are reading past the end of a `std::vector` (in this case, `c->size.size()` is 1, but `l.pure_dim` is 1)... so this is a real bug.

Unfortunately I have no idea where to even begin debugging this... assigning to @abadams for now to see if he can route it to the right person.

(ASAN isn't necessary to repro this bug, just add something like `internal_assert(l.pure_dim < c->size.size());` in the code)

(But if you want to run with asan, just add `--preset linux-x64-asan` to the cmake and ctest commands)

Yongqi-Zhuo:
Any updates on this? :)
Looking forward to using this in production

aekul:
Will take a look at this in the next few days.

aekul:
I'm not able to reproduce this. I added an assert and built Halide with CMake but things work fine. Is that all I should have to do to reproduce it?  @steven-johnson or @Yongqi-Zhuo are using the default settings for the autoscheduler?

aekul:
I'm having trouble getting ASAN to build. I've tried LLVM 15 and 16, but can't build either of them. Error is:
```
CMake Error at cmake/Modules/CompilerRTUtils.cmake:434 (string):
  string sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  CMakeLists.txt:118 (construct_compiler_rt_default_triple)
```
Tried Googling but can't find any way to resolve it. "compiler-rt" seems to be the issue. Is `-DLLVM_ENABLE_RUNTIMES="compiler-rt"` required for ASAN?

steven-johnson:
Yikes, yes, sorry (I thought this was in the readme)... for ASAN you must build LLVM with

    -D LLVM_ENABLE_PROJECTS="clang;lld;clang-tools-extra" 
    -D LLVM_ENABLE_RUNTIMES="compiler-rt;libcxx;libcxxabi;libunwind" 

(I have only tested recently with top-of-tree LLVM but 16 *should* work)

aekul:
Building LLVM works with those changes, thanks. But now when I compile Halide, I get `Float16.h:43:24: error: _Float16 is not supported on this target` when using `--preset linux-x64-asan` or `--preset release`. Any ideas? I'm using a PowerPC machine in case that's relevant.

[...]

PR #7706, opened 2023-07-25
Attempt to fix #7703

### LINKED ISSUE #7703: Anderson2021 autoscheduler plays dubious games with unique_ptr<>
While investigating #7606 I noticed this code in the Anderson2021 codebase:

```
std::unique_ptr<ThreadInfo> GPULoopInfo::create_thread_info() {
    internal_assert(at_or_inside_block());
    internal_assert(at_or_inside_thread());

    auto max_thread_counts = current_block_loop->get_union_thread_counts(nullptr);
    std::unique_ptr<ThreadInfo> new_thread_info = std::make_unique<ThreadInfo>(
        current_thread_loop->vectorized_loop_index,
        current_thread_loop->size,
        current_thread_loop->stage->loop,
        max_thread_counts);
    thread_info = new_thread_info.get();
    return new_thread_info;
}
```

since `thread_info` is a member of `this`, and declared as a plain `const ThreadInfo*`, after this call, it refers to an *unowned* pointer... the ownership is in the returned unique_ptr. Since the only two call sites immediately assign the result to a local unique_ptr, this member var is guaranteed to be `stale` at the end of the calling routine.

I'm not sure what the right fix is here because I don't know the intended semantics -- `shared_ptr` might be appropriate, but it's not clear to me whether the returned result is intended to be mutable or not, or what constraints there are on mutability. 

The current code looks absolutely unsafe, however, and needs fixing; if there are similar code patterns elsewhere in this codebase, they need examination as well.

steven-johnson:
This fixes #7703, in the sense that we no longer play dubious games with unique_ptr. As a bonus, the failure in #7699 is now reproducible without using ASAN (it becomes an assertion failure). Not sure if it addresses #7606 or not. 

Not sure if we should land this as-is or wait for a full fix -- landing this will inject *reliable* failures on the buildbots.

### DIFF
--- a/src/autoschedulers/anderson2021/GPULoopInfo.h
+++ b/src/autoschedulers/anderson2021/GPULoopInfo.h
@@ -46,8 +46,8 @@ struct GPULoopInfo {
     const ThreadInfo *create_thread_info();
 
     // Note: if create_thread_info() has not been called yet, this will return nullptr.
-    // TODO: should this just assert-fail if thread_info is null? I don't know if
-    // downstream code might legitimately check for null to mean "no thread_info exists".
+    // (Note that this is an unusual but legitimate situation, so it should *not*
+    // assert-fail if the value is null.)
     const ThreadInfo *get_thread_info() const {
         return thread_info.get();
     }
--- a/src/autoschedulers/anderson2021/GPULoopInfo.h
+++ b/src/autoschedulers/anderson2021/GPULoopInfo.h
@@ -7,6 +7,7 @@
  * hierarchy of blocks, threads, etc. Useful when computing GPU features
  */
 
+#include <memory>
 #include <vector>
 
 #include "Halide.h"
@@ -19,7 +20,7 @@ namespace Autoscheduler {
 struct LoopNest;
 
 struct GPULoopInfo {
-    GPULoopInfo(const LoopNest *root)
+    explicit GPULoopInfo(const LoopNest *root)
         : root{root} {
     }
 
@@ -30,7 +31,6 @@ struct GPULoopInfo {
     int64_t num_blocks = 1;
     int64_t total_outer_serial_extents = 1;
     int64_t total_inner_serial_extents = 1;
-    const ThreadInfo *thread_info = nullptr;
 
     void update(const Target &target, const LoopNest *loop);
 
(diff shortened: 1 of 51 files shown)


steven-johnson:
@aekul -- if you don't have the bandwidth to look at this, if there are any pointers you can give to get me started on fixing it, I'd appreciate it; I have some co-workers who want to try this out, but I don't want to give it to them until we can work out the crashy bits :-)

aekul:
@steven-johnson Just started looking into this. Using a `shared_ptr` is probably the way to go but I'll investigate a bit more in case I'm overlooking something.

aekul:
Using a `shared_ptr` makes sense here. The `ThreadInfo` is intended to be created once (the first time `compute_features` is called on the thread loop) and then shared across multiple `GPULoopInfo` objects that are created in the recursive calls to `compute_features`.

There's an assumption in the code that when accessing the `ThreadInfo` object, we should be at a loop nest inside a thread loop, so the object should be non-null. If that's not the case, then it indicates a bug not in the `ThreadInfo` creation code but in how that particular Func is handled e.g. in https://github.com/halide/Halide/pull/7699 the issue is that scalars are not handled properly.

@steven-johnson were these changes finalized or were there other things you planned to do? If they're final, I'll approve. This will probably create merge conflicts with https://github.com/halide/Halide/pull/7726 though.

steven-johnson:
Let's get #7726 landed and then revisit this with that in mind -- I approved #7726 but want to let the buildbots do at least some spot-builds before landing it, hopefully later today

steven-johnson:
PTAL -- look at the TODO comment for `get_thread_info()` too.

src/autoschedulers/anderson2021/GPULoopInfo.h
```diff
@@ -42,9 +42,23 @@ struct GPULoopInfo {
...
     std::vector<int64_t> get_inner_serial_loop_extents(const LoopNest *loop_nest) const;
 
-    std::unique_ptr<ThreadInfo> create_thread_info();
+    // assert-fails if create_thread_info() has *already* been called.
+    const ThreadInfo *create_thread_info();
+
+    // Note: if create_thread_info() has not been called yet, this will return nullptr.
+    // TODO: should this just assert-fail if thread_info is null? I don't know if
+    // downstream code might legitimately check for null to mean "no thread_info exists".
+    const ThreadInfo *get_thread_info() const {
```
aekul, line 51: This can legitimately return null for scalars after https://github.com/halide/Halide/pull/7726 (e.g. [here](https://github.com/halide/Halide/blob/main/src/autoschedulers/anderson2021/LoopNest.cpp#L978)) so I don't think it should assert-fail.

src/autoschedulers/anderson2021/LoopNest.cpp
```diff
@@ -4031,8 +4031,7 @@ double LoopNest::max_idle_lane_wastage(const Target &target, GPULoopInfo gpu_loo
     std::unique_ptr<ThreadInfo> thread_info;
```
aekul, line 4031: This looks unused now so can be removed

aekul (approved):
LGTM. Thanks for fixing this

src/autoschedulers/anderson2021/LoopNest.cpp
steven-johnson: Done.

src/autoschedulers/anderson2021/GPULoopInfo.h
steven-johnson: Comment added.

steven-johnson:
Failures are unrelated, landing

PR #7717, opened 2023-07-26
Ensure that multitarget AOT builds have consistent random sequence

steven-johnson:
If a Generator uses random_float() (or the int or uint versions), and is used in a multitarget build, we weren't resetting the counters for random generation between each subtarget... meaning that each subtarget would get a different random sequence, leading to some ery hard-to-debug test failures when running on different hardware variants. This PR ensures that the relevant counters are all reset before each subtarget is generated, so that each should see the same sequence of random number generation.

### DIFF
--- a/src/Function.cpp
+++ b/src/Function.cpp
@@ -299,10 +299,11 @@ public:
     }
 };
 
-// A counter to use in tagging random variables
-std::atomic<int> rand_counter{0};
-
 }  // namespace
+
+// A counter to use in tagging random variables.
+// Note that this will be reset by Internal::reset_random_counters().
+std::atomic<int> random_variable_counter = 0;
 
 Function::Function(const FunctionPtr &ptr)
     : contents(ptr) {
@@ -542,7 +543,7 @@ void Function::define(const vector<string> &args, vector<Expr> values) {
     }
 
     // Tag calls to random() with the free vars
-    int tag = rand_counter++;
+    int tag = random_variable_counter++;
     vector<VarOrRVar> free_vars;
     free_vars.reserve(args.size());
     for (const auto &arg : args) {
@@ -752,7 +753,7 @@ void Function::define_update(const vector<Expr> &_args, vector<Expr> values) {
             free_vars.emplace_back(RVar(check.reduction_domain, i));
         }
     }
-    int tag = rand_counter++;
+    int tag = random_variable_counter++;
     for (auto &arg : args) {
         arg = lower_random(arg, free_vars, tag);
     }
--- a/src/IROperator.cpp
+++ b/src/IROperator.cpp
@@ -2593,10 +2593,26 @@ Expr mod_round_to_zero(Expr x, Expr y) {
                       Call::PureIntrinsic);
 }
 
+namespace {
+
+std::atomic<int> random_float_counter = 0;
+std::atomic<int> random_uint_counter = 0;
+
+}  // namespace
+
+namespace Internal {
+
+void reset_random_counters() {
+    random_float_counter = 0;
+    random_uint_counter = 0;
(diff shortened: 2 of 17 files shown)


steven-johnson:
> Do we want to just reset the random counters, or do we want to centralize and reset all state (i.e. unique name counters too)?

I'd be very much in favor of that. I'm not sure what else to add, though -- unique-name stuff, anything else? (And to instantly add scope-creep -- could we move all such global state into a single struct that is referenced by a thread-local pointer? :-) Having compiler state in random globals has always irked me from a design perspective, but if we could at least collapse it into a single global struct it would be less bad...)

abadams:
In theory IR can be generated from multiple threads and then used in a single pipeline, so I don't think we can use thread-local state.

steven-johnson:
> In theory IR can be generated from multiple threads and then used in a single pipeline, so I don't think we can use thread-local state.

Yeah, I guess that would take some more careful thinking. But moving all the mutable global state into a single struct seems like a worthwhile goal regardless.

TH3CHARLie:
To add more motivation for this PR, in libfuzzer, each invocation of the fuzzer API will create the same pipeline but with incrementing suffix like `blur$800`

steven-johnson:
Closing in favor of #7722

steven-johnson:
Reopening this as #7722 ran into some sticky issues, and the original impetus for investigating this was just the random sequences.

src/IROperator.h
```diff
@@ -351,6 +351,11 @@ Expr requirement_failed_error(Expr condition, const std::vector<Expr> &args);
 
 Expr memoize_tag_helper(Expr result, const std::vector<Expr> &cache_key_values);
 
+/** Reset the counters used for random-number seeds in random_float/int/uint
+ * when no seed is explicitly passed in; this is used for multitarget compilation
```
abadams, line 355: The counter is added even if a seed is passed in. The seed is additional.

src/IROperator.cpp
```diff
@@ -2593,10 +2593,26 @@ Expr mod_round_to_zero(Expr x, Expr y) {
                       Call::PureIntrinsic);
 }
 
+namespace {
+
+std::atomic<int> random_float_counter = 0;
```
abadams, line 2598: Now that it's a global, this can just be one counter, and the even/odd nonsense below can be removed.

src/Function.cpp
```diff
@@ -299,11 +299,12 @@ class FreezeFunctions : public IRGraphVisitor {
...
 };
 
-// A counter to use in tagging random variables
-std::atomic<int> rand_counter{0};
-
 }  // namespace
 
+// A counter to use in tagging random variables.
+// Note that this will be reset by Internal::reset_random_counters().
+std::atomic<int> random_variable_counter = 0;
```
abadams, line 306: The only effect of this tag seems to be to give each compiled Pipeline a separate set of random numbers, in addition to each individual call to random_whatever being distinct. resetting the random counters in Module.cpp below seems like it would just totally undo this...

So if this counter is important, we'd better figure out why, because this PR breaks it.

abadams: Disregard, this is a per-Func counter, not a per-Pipeline counter.

src/IROperator.h
steven-johnson: Done.

src/Function.cpp
abadams: Yes

steven-johnson:
PTAL

steven-johnson:
Clean aside from the since-fixed bogus clang-tidy failure, PTAL

abadams (approved):

PR #7718, opened 2023-07-27
Change default generator timeout to infinite

abadams:
Timeouts should really be the responsibility of the build system. Leaving the feature in because it's useful for things like autotuning loops.


steven-johnson (approved):

abadams:
Failures unrelated.

PR #7726, opened 2023-07-31
Fix handling of thread features for scalars in Anderson2021

aekul:
This fixes the issue in https://github.com/halide/Halide/pull/7699. The cause is that the autoscheduler assumes that every Func is surrounded by a block loop and a thread loop, which means it assumes there is always a non-null `ThreadInfo` object when the features are computed. But in this case, the output of the generator is a single number so the autoscheduler is currently not creating a block and thread loop for it, so the ThreadInfo object ends up being null.

The main thing this PR does is add some special cases to `compute_features` for handling scalars. It was also necessary to change some function signatures to accept a `ThreadInfo*` instead of a `ThreadInfo&` so that it can test whether the passed object is null or not.

### DIFF
--- a/src/autoschedulers/anderson2021/LoopNest.cpp
+++ b/src/autoschedulers/anderson2021/LoopNest.cpp
@@ -718,7 +718,7 @@ double LoopNest::compute_local_mem_stride(double stride, double bytes) const {
 
 // Get the stride over "node's" storage and its element-wise stride for a unit
 // increment in the given thread loops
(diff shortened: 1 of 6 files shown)


steven-johnson (approved):
LGTM, but I presume that #7703 is still an issue -- after this lands, do we want to consider #7706 as a fix for it, or do you have an alternate suggestion?

steven-johnson:
(Oh, and thanks SO much for taking the time to look at this!)

steven-johnson:
This looks like it may have injected #7729 based on a quick bisection -- I thought that `anderson2021_test_apps_autoscheduler` had run on the testbots, but maybe I misread due to other false-positive failures :-/

aekul:
No problem. Thanks for looking into the test failure

Issue #7731, opened 2023-07-31
buildbots aren't testing anderson2021 autoscheduler


steven-johnson:
While investigating https://github.com/halide/Halide/issues/7729 I realized that we aren't testing anderson2021 on the bots at all: we only try to test it if Cuda is enabled... but we don't do any autoscheduler testing with Cuda because we formerly had no autoschedulers that worked with it. This will be simple to fix in the build config but we obviously shouldn't do so until #7729 is fixed (otherwise we'll get constant failures).

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** Filed because the anderson2021 tests were only registered when a CUDA target was enabled, so no bot ever ran them; the blocker (#7729) has since been fixed. Today the tests are registered whenever LLVM has NVPTX (true for all CI configs) and carry granular labels that the CI `ctest` invocations pick up.

**Evidence:** `test/autoschedulers/CMakeLists.txt:9` gates on `"NVPTX" IN_LIST Halide_LLVM_COMPONENTS`; labels added by bc30d6f83 (#7732); blocker #7729 fixed by 831fd1a47 (#7734); `.github/workflows/testing-linux.yml:108` runs `ctest -LE performance`, which includes the `anderson2021` label. I built and ran `anderson2021_test_apps_autoscheduler` from this checkout and it printed `Success!`.

Closing: anderson2021 tests are now registered whenever LLVM has NVPTX (`test/autoschedulers/CMakeLists.txt:9`) and are picked up by the CI `ctest -LE performance` runs; the blocking bug #7729 was fixed in #7734. The old buildbot fleet has also been replaced by GitHub Actions since this was filed.

PR #7732, opened 2023-08-01
Revise labels on autoscheduler tests

steven-johnson:
This is step 1 in fixing https://github.com/halide/Halide/issues/7731: it replaces the `autoschedulers` tag with more granular ones, so that we can modify the build script to test the right autoscheduler(s) for a given backend. (Note that the `autoschedulers` tag was unused by the buildbots, which only used the generic `auto_schedule` tag.)

Step 2 will be to modify the buildbot script after this lands to use the new tags above.

Step 3 will be to remove the `auto_schedule` tag.


steven-johnson:
Failures are unrelated Vulkan/D3D12 failures, PTAL

steven-johnson:
Failures unrelated, gentle review ping plz

abadams (approved):

PR #7735, opened 2023-08-01
Fix leak on cloning functions with update defs

abadams:
When cloning a Func with an update def, the remapping map resulting from the deep copy may already contain a key for the wrapped function pointing to a strong reference to itself. The reasons are unclear to me, but it means that emplace silently does nothing and we get a memory leak because the cloned Func's update definition has a strong self-reference after the remapping is applied.

We want to replace it with a weak reference, so this PR changes things to use operator[] instead of emplace.


steven-johnson (approved):
LGTM but please leave a comment here so it doesn't get changed back

PR #7719, opened 2023-07-27
Making HLSL code-gen a couple orders of magnitude faster...

slomp:
Title says it all!

The redundant call to `print_expr()` ends up triggering an exponential chain of events that ultimately gets discarded.

PR #7737, opened 2023-08-01
Revert "Making HLSL code-gen a couple orders of magnitude faster..."

abadams:
Reverts halide/Halide#7719


steven-johnson:
So it looks like that PR did not inject the bug -- so maybe WinWorker2 really is just too old to be useful here (which wouldn't be surprising). To be a broken record here, we really need more Windows bots; WinWorker3 is decent when it is running but it has proven to be unstable over time. Desperately need Windows testing we can rely on and stop spending endless person-hours fixing.

steven-johnson:
See https://github.com/halide/build_bot/pull/250

Issue #7724, opened 2023-07-28
CodeGen performance: stringstream is quite slow!


slomp:
While profiling the HLSL code gen, I noticed that `ostringstream` was much slower than plain `std::string` concatenation (coupled with `std::to_string()` conversions), at least on MSVC's standard library implementation.

In the case of `print_type_maybe_storage`, using `ostringstream` is about an order of magnitude slower:
![image](https://github.com/halide/Halide/assets/526886/87b5e2af-6ded-4797-8f8b-ad93f30024a0)

In the case of `print_reinforced_cast`, it's about 5x:
![image](https://github.com/halide/Halide/assets/526886/2c2c2b84-db84-4576-8eb8-e45e59fd81d0)

Overall, looking at the top-level `add_kernel` impact of these two experiments, I got about 25% speedup:
![image](https://github.com/halide/Halide/assets/526886/272d9356-bcb1-426f-9f55-c4d16f4e8826)

slomp:
I was trying to come up with a hack to implement a `std::string` wrapper that behaves like an `std::ostream`, but because we store `std::ostream&` in `IRPrinter`, there's no way to have any subsequent `<<` be intercepted by the wrapper class.

We would have to come up with our own stream interface and replace `std::osream&` members. (We may still need an `ostream& operator << (ostream&, OurStreamClass&)` for compatibility between stream interfaces.)

antonysigma:
If `std::stringstream` is too resource heavy, perhaps we should to move to the `std::format` API? 

The idea is that, instead of writing things like:
```c++
void IRPrinter::visit(const Reinterpret *op) {
    stream << "reinterpret<" << op->type << ">(";
    print(op->value);
    stream << ")";
}
```

We write this:
```c++
#include <format>
...
void IRPrinter::visit(const Reinterpret *op) {
    using nostd::format_literals;
    nostd::print(stream, "reinterpret<{type:s}>({value:d})", 
        "type"_a=op->type,
        "value"_a=op->value);
}
```

**The pros:** The new API claims to be more efficient because since C++17, the format string `"reinterpret<{type:s}>({value:d})"` is stored as `std::string_view`, and in-turn compiled into `.rodata` section of the binary. Secondly, at the highest optimization level `-O3`, the `std::format` claims to invoke `constexpr` comptile-time evaluation to parse the format string, completely bypasses the traditional hacks of `ostream& operator<<` and `std::to_string()` and other formatting hints `std::hex | std::set_fill`. 

[...]

PR #7738, opened 2023-08-03
Experimenting with an ostringstream-like object that wraps around std::string

slomp:
This is a rough draft, not for merging.
The implementation is a bit messy, since I was trying to be quick and make as few changes as possible.
This is just to illustrate issue https://github.com/halide/Halide/issues/7724.

To get to the optimal path, enable `SLOMP_REPLACE_ISOLATED_STRINGSTREAMS` in `CodeGen_C.h` by setting it to `(1)`, and disable `SLOMP_HALIDE_STREAM_REDIRECT_TO_OSTREAM` in `IRPrinter.cpp` by setting it to `(0)`.

Across the board, on both MSVC and MacOS+clang+libc++, replacing `ostringstream` by `string` concatenation is anywhere from 1.5x to 10x faster. I'll post some profiling pictures soon. With the replacement, the various `print` routines move down the hot-spot list and things like cache insertions and such move to the top (as expected).

Is it really worth? Maybe. This is mostly for GPU code (and statement files). I am not sure about the overall impact on other code-gen paths. Moreover, the real bottlenecks are the various `substitute` routines. which dwarf the print routines anyway.

## MacOS + clang + libc++

before:
Screenshot 2023-08-03 at 11 07 16 AM

after:
Screenshot 2023-08-03 at 11 06 39 AM

## MSVC

before:
image

after:
image

### DIFF
--- a/src/CodeGen_C.cpp
+++ b/src/CodeGen_C.cpp
@@ -15,12 +15,38 @@
 #include "Util.h"
 #include "Var.h"
 
+#define TRACY_GLUE_ENABLE (0)
+#include "debug/tracy_profiler_glue.hpp"
+
 namespace Halide {
 namespace Internal {
 
+#if SLOMP_REPLACE_ISOLATED_STRINGSTREAMS
+struct ostringstream {
+    std::string stream;
+
+    template<typename T>
+    ostringstream& operator << (const T& t) {
+        stream += std::to_string(t);
+        return *this;
+    }
+    ostringstream& operator << (const char* str) {
+        stream += str;
+        return *this;
+    }
+    ostringstream& operator << (const std::string& str) {
+        stream += str;
+        return *this;
+    }
+
+    const std::string& str() { return stream; }
+};
+#else
+using std::ostringstream;
+#endif
+
 using std::map;
 using std::ostream;
-using std::ostringstream;
 using std::string;
 using std::vector;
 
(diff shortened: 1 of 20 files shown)


steven-johnson:
Not to cast any shade on this work, but is there existing work we can just recycle here?

e.g., 
- llvm:raw_ostream
- absl::StrCat() [which is the internal-to-Google equivalent, since use of `std::stream` and friends is pretty much forbidden in the Google style guide in favor of this -- using this directly would require too much code churn since it uses call syntax instead of `<<` but the underlying implementation might be good to adapt]

slomp:
> Not to cast any shade on this work, but is there existing work we can just recycle here?
> 
> e.g.,
> 
> * llvm:raw_ostream
> * absl::StrCat() [which is the internal-to-Google equivalent, since use of `std::stream` and friends is pretty much forbidden in the Google style guide in favor of this -- using this directly would require too much code churn since it uses call syntax instead of `<<` but the underlying implementation might be good to adapt]

Certainly, it was never my intention to merge this.
I just want to have a record here for people to try it out if curious, and to document findings.

A proper implementation can be made much cleaner.
I would just advocate against bringing more unnecessary dependencies, and to avoid black-boxes that we have no control over in the future.

steven-johnson:
Random thoughts: 
- If we're looking into improving string-manipulation time, migrating towards using `std:: string_view` where appropriate would probably be a worthwhile (incremental) win.
- Improving codegen time is always nice, but is it problematically slow for AOT? (i.e., I guess if you are jitting it matters a lot more...)

steven-johnson:
Should we leave this open or close it?

slomp:
It was a fun experiment, but on the grand scheme of things, we have bigger fish to fry as far as accelerating code-generation as a whole.

PR #7739, opened 2023-08-03
[vulkan] Fix SPIR-V IR references causing leaks

derek-gerstmann:
Removed parent/owner back references (which weren't actually needed)
Fixed ownership transfers
Explicitly destruct and clear contents

### DIFF
--- a/src/SpirvIR.cpp
+++ b/src/SpirvIR.cpp
@@ -357,12 +357,12 @@ SpvBlock::~SpvBlock() {
 
 void SpvBlock::add_instruction(SpvInstruction inst) {
     check_defined();
-    contents->instructions.push_back(inst);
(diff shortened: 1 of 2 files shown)


steven-johnson (approved):
Thanks so much for this! LGTM pending green.

derek-gerstmann:
> Thanks so much for this! LGTM pending green.

Thanks for finding these!  The ASAN toolchain is a massive help!

abadams:
I believe after this PR and the memoize fix one (which is still failing on windows, unfortunately), that's all the leaks.

steven-johnson:
> I believe after this PR and the memoize fix one (which is still failing on windows, unfortunately), that's all the leaks.

Sweet! LMK how the Windows issues go.

PR #7746, opened 2023-08-04
Run clang-tidy on macOS runners instead of Linux

steven-johnson:
The current macOS runners have twice the RAM and more CPU power.

Also, drive-by change to allow specifying the parallelism that the run-clang-tidy script should use (defaults to nproc)


steven-johnson:
(Ignore the change in Generator.cpp, it's just there to trigger the check, will be reverted)

alexreinking (approved):

PR #7749, opened 2023-08-07
Fix a couple of clang-tidy errors

steven-johnson:
Newer versions of clang-tidy can warn about things that are referenced only via transitive includes (e.g. "no header providing "Halide::Expr" is directly included"); this is not an error per se, but having everything referenced in a .cpp file directly included is good practice to avoid flaky include issues on different systems.

### DIFF
--- a/src/Dimension.cpp
+++ b/src/Dimension.cpp
@@ -1,5 +1,6 @@
 #include "Dimension.h"
 
+#include "Expr.h"
 #include "IR.h"
 #include "IROperator.h"
 #include <utility>
--- a/src/Parameter.cpp
+++ b/src/Parameter.cpp
@@ -5,6 +5,7 @@
 #include "IR.h"
 #include "IRMutator.h"
 #include "IROperator.h"
+#include "Util.h"
 
 namespace Halide {
 namespace Internal {


steven-johnson:
Monday morning review ping

abadams:
I'm confused about why this touches so few files. Don't we ubiquitously include IR.h by itself, instead of IR.h + Expr.h ?

I once applied iwyu to Halide and it made a pretty big mess.

abadams:
My experience with iwyu led me to believe transitive inclusion is a good thing. Forbidding it encourages packing things into fewer, larger headers, which is bad for a variety of reasons. Allowing it makes it feasible to factor things into more smaller headers (e.g. PrefetchDirective.h, FunctionPtr.h) without adding a giant list of #includes to everything. 

I'd really prefer a checker that tells you when a header include inside another header isn't actually necessary, and can be accomplished with forward-declaration of a few types instead.

abadams:
Looks like most of our lowering passes just include IRMutator.h, and expect to get IR.h, Expr.h, ModulusRemainder.h, PrefetchDirective.h, FunctionPtr.h, and all the other assorted structs that briefly exist on the stack when mutating IR nodes. Surely a checker that forbids transitive includes would add all of these. Maybe the clang-tidy checker has some sort of more considered take on the issue.

steven-johnson:
> My experience with iwyu led me to believe transitive inclusion is a good thing

From everything I can tell, the Clang ecosystem of correctness checkers is aggressively moving towards avoiding the reliance on transitive inclusion. IMHO, I've never seen a list of #includes so long that it was a problem.

> Surely a checker that forbids transitive includes would add all of these

As I said, I can't explain it, with "it's not fully baked yet" the assumption I have.

I don't want to spend any time debating whether this is a good thing or not; if we don't want to do this sort of thing upstream, I'll just have to add the extra #includes on the Google side via the set of Google-specific patches we make when importing. (Otherwise we'll get nagged every time we make a change in our code, because everything not forbidden is compulsory, apparently.)

Issue #7519, opened 2023-04-18
stmt and stmt_html output are too low level


abadams:
For example, they show already-compiled PTX assembly for cuda kernels instead of stmt ir, because those have already been offloaded. As more and more of codegen creeps into lowering, this problem will get worse. We need to identify a point in lowering at which the stmt should be preserved for stmt and stmt_html output. I propose just after the custom passes and before hexagon kernel offload.

See #7507

antonysigma:
![232558221-25826389-9ccc-4663-881b-1ed649d2367c](https://github.com/halide/Halide/assets/12185630/c293c0a0-8536-4f0d-b78f-e9539e957984)

Uploading the screenshots to illustrate the stmt output, and why they are too low level since Halide 14.0 (?)

antonysigma:
> For that to happen we'd have to emit the .stmt file before the offload gpu pass, e.g. in Lower.cpp around line 441. I guess we'd have to stash that earlier stmt in the Module somewhere for use in the .stmt outputs.

@maaz139 Other than line 441 of `lower.cpp`, is there anywhere I can contribute as a Halide beginner? I suppose I can add an optional argument `intent=stmt | codegen` for a start?

```c++
Module lower(const std::vector<Function> &output_funcs, ..., Halide::my_intent_t = Stmt_only);
```

mcourteaux:
I'd like to chime in, and state that **_I :heart: the PTX code in the Stmt HTML_** to get a good understanding of what's happening in a single CUDA thread. I examined these for months when working on something for work. Much like how one wants to check the generated assembly for CPU code, you'd want to validate what got generated for the GPU code.

In my opinion, the compilation pipeline should be respected for what it is: a pipeline. The code can be useful to inspect at different stages. Statement IR can be useful, PTX code can be useful. Just like we are able to get a `stmt` file showing IR and an `s` file showing the CPU assembly, I believe we should be able to see an output PTX output file. I'd imagine `.s` and `.ptx`. Or perhaps `.x86-64.s` and `.ptx.s`? Maybe there are existing conventions here already that I'm unaware of.

Alternatively, we output the PTX code in the assembly file, but that would cause us to lose the colors, and in general is hard to navigate, as the `s` file is less verbose and intuitive.

Considering all the discussions regarding the VizIR HTML, the problems it has, and the fact that now PTX code is likely to be another variable, I believe a possible solution could be to have more generator emit options: now there is only `stmt` and `stmt_html`. I think the following would be useful:

[...]

PR #7753, opened 2023-08-10
[Help wanted] Emit stmt ahead of GPU/DSP/parfar offloading

antonysigma:
Stash the stmt right after the "final" (?) lowered form, ahead of DSP or GPU offloading. Also skipped the extraction of `parfor` into individual functions. The stmt is stored in `Halide::Module::contents->pseudo_code`. Modify IRPrinter to emit the pseudo-code as desired.

~Current status: the new code can emit stmt from the generator command-line argument `path/to/generator ... -e stmt`, but not `-e html`.~

Note: I have no idea how VizStmt estimate the compute and data movement costs for HVX/GPU `Stmt`. ~Help wanted to resolve the following bug:~

Resolves: #7519 . 

cc'ed @maaz139 .

### DIFF
--- a/src/Lower.cpp
+++ b/src/Lower.cpp
@@ -423,6 +423,8 @@ void lower_impl(const vector<Function> &output_funcs,
     debug(1) << "Lowering after final simplification:\n"
              << s << "\n\n";
 
+    const Stmt pseudo_code{s};
+
     if (!custom_passes.empty()) {
         for (size_t i = 0; i < custom_passes.size(); i++) {
             debug(1) << "Running custom lowering pass " << i << "...\n";
@@ -559,6 +561,8 @@ void lower_impl(const vector<Function> &output_funcs,
     }
 
     result_module.append(main_func);
+
+    result_module.stashPseudoCode(LoweredFunc{pipeline_name, public_args, pseudo_code, linkage_type});
 
     auto *logger = get_compiler_logger();
     if (logger) {
--- a/src/StmtToViz.cpp
+++ b/src/StmtToViz.cpp
@@ -46,9 +46,7 @@ public:
     // Pre-compute all costs to avoid repeated work
     void compute_all_costs(const Module &m) {
         // Compute all node costs
-        for (const auto &fn : m.functions()) {
-            fn.body.accept(this);
-        }
+        m.pseudoCode().body.accept(this);
 
         // Compute the max cost for each category
         max_compute_cost = -1;
@@ -526,9 +524,7 @@ public:
         // nodes we need to map and generate their assembly
         // markers (comments that appear in the assembly code
         // associating the code with this node)
-        for (const auto &fn : m.functions()) {
-            fn.body.accept(this);
-        }
+        m.pseudoCode().body.accept(this);
 
         // Find markers in asm code
         std::istringstream asm_stream(code);
@@ -672,9 +668,7 @@ public:
         }
 
         // Print module functions
-        for (const auto &fn : m.functions()) {
-            print(fn);
-        }
+        print(m.pseudoCode());
 
         // Close indented div holding body code
         print_closing_tag("div");
@@ -2190,9 +2184,7 @@ public:
 
     void print(const Module &m, AssemblyInfo asm_info) {
         assembly_info = std::move(asm_info);
-        for (const auto &fn : m.functions()) {
-            print(fn);
-        }
+        print(m.pseudoCode());
     }
 
 private:
--- a/src/StmtToViz.cpp
+++ b/src/StmtToViz.cpp
@@ -734,7 +734,7 @@ private:
             print_html_element("span", "matched", " {");
 
             // Open indented div to hold buffer data
-            print_opening_tag("div", "indent BufferData", id);
+            print_opening_tag("div", "indent BufferData collapsed-block", id);
 
             std::string str((const char *)buf.data(), buf.size_in_bytes());
             if (starts_with(buf.name(), "cuda_")) {
--- a/src/IRPrinter.cpp
+++ b/src/IRPrinter.cpp
@@ -79,9 +79,7 @@ ostream &operator<<(ostream &stream, const Module &m) {
     for (const auto &b : m.buffers()) {
         stream << b << "\n";
     }
-    for (const auto &f : m.functions()) {
-        stream << f << "\n";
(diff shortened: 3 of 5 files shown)


src/Lower.cpp
```diff
@@ -423,6 +423,8 @@ void lower_impl(const vector<Function> &output_funcs,
     debug(1) << "Lowering after final simplification:\n"
              << s << "\n\n";
 
+    const Stmt pseudo_code{s};
+
```
antonysigma, line 427: Question to self: what does "final" mean here? Are we okay to skip parallel functions lowering?

src/StmtToViz.cpp
```diff
@@ -740,7 +734,7 @@ class HTMLCodePrinter : public IRVisitor {
             print_html_element("span", "matched", " {");
 
             // Open indented div to hold buffer data
-            print_opening_tag("div", "indent BufferData", id);
+            print_opening_tag("div", "indent BufferData collapsed-block", id);
```
antonysigma, line 737: Question: why are we printing `buffer cuda_gpu_source_kernels` in the StmtVisualizer's Stmt panel? I suppose I can collapse the assembly code by default.

src/Lower.cpp
maaz139: I think printing any IR other than what is present after *all* lowering steps may be confusing or misleading (if thats is the only output). 

Perhaps the answer is allowing users to view the IR after each IR stage or more realistically at a few key checkpoints during the lowering pipeline?

src/StmtToViz.cpp
maaz139: In this instance, the stmt tab is just mimicking the basic IR printer behavior.

Given that we now have an assembly tab when using the html output, perhaps we can instead simply add a jump to the corresponding line in assembly. I think collapsing for now is fine as well.

src/StmtToViz.cpp
```diff
@@ -76,14 +74,14 @@ class IRCostModel : public IRVisitor {
     int get_compute_cost(const IRNode *node, bool include_subtree_cost) {
         internal_assert(node != nullptr) << "IRCostModel::get_compute_cost(): node is nullptr\n";
 
-        int cost = -1;
+        int cost = 0;
         if (compute_cost.count(node)) {
             cost = include_subtree_cost ? compute_cost_inclusive[node] : compute_cost[node];
         } else {
-            internal_error << "IRCostModel::get_compute_cost(): cost lookup failed\n";
+            debug(2) << "IRCostModel::get_compute_cost(): cost lookup failed, node_type=" << int(node->node_type) << "\n";
```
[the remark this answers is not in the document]
maaz139: I haven't had a chance to confirm this yet, but I think the problem is line 2188. The HTMLVisualizationPrinter is not printing the stashed psuedocode over which the costmodel was run.

https://github.com/halide/Halide/blob/1277688d9a170fd20a415e8c9539feb32c63624d/src/StmtToViz.cpp#L2188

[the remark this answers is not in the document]
antonysigma: Yes. Replacing the line with `print(m.pseudoCoder());` resolves the "cost lookup failed" issue. Closing this thread.

src/Lower.cpp
antonysigma: I am okay to have additional panel in the Visualizer output. I took the before and after screenshots of this PR. Could you please point out which figure, (b) or (c), do you prefer?

For my usecase (reasoning about `gpu_thread` and `gpu_block` sizes), I only read the colored pseudo-code output. The visualizer panel is optional to me. I don't need cross-reference buttons in the pseudo-code panel.

image

src/StmtToViz.cpp
antonysigma: Re: collapsing the block. Great. 

Re: button to cross-reference IR with assembly. As far as I know, the feature is absent in Halide 16.0.0. Line 2164 appears to destroy the bookmarks on the fly. Not sure why.

I don't use that feature by the way. Wondering who else is using it?
 https://github.com/halide/Halide/blob/027547f713f6a01c7b51cb59baa665219f89751d/src/StmtToViz.cpp#L2164

src/Lower.cpp
antonysigma: Update: discussion moved to https://github.com/halide/Halide/issues/7519#issuecomment-1676051118 . Needing a consensus on the number of IR checkpoints to show simutaneously in the StmtViz panels.

src/StmtToViz.cpp
mcourteaux: > Question: why are we printing buffer cuda_gpu_source_kernels in the StmtVisualizer's Stmt panel? I suppose I can collapse the assembly code by default.

It boils down to the other comment of @maaz139. Because that really _is_ the lowered IR. See: https://github.com/halide/Halide/pull/7753/files#r1290639013

I think it would be wrong to take it out.

mcourteaux:
I have been thinking -- like most of you -- quite a bit about this. I believe it just makes sense to indeed pick a few point in the lowering process, stash the IR away into the Module for later, and emit multiple files. Each file representing one point in the lowering process.

I'm absolutely not sure what to do with the PTX code and the assembly tab. It _is_ assembly, and therefore could go there, but it's not _the_ assembly of the pipeline. It really is a buffer containing assembly for the GPU. Maybe... Perhaps... It is more sensible to have a "Buffers" tab?

Overall, I believe the HTML-way of doing things is getting convoluted, and I'm thinking of an approach where we just dump a very information-rich IR tree to some file format and have a custom DearImGui-based tool be the visualizer? I don't know what the recently-merged "experimental serializer" is exactly capable of, but maybe piggybacking on that to dump flatbuffers of the IR at a few points of the lowering process can be a nice trick here to build such a tool. @TH3CHARLie Can the serializer you built be used for dumping an IR tree?

antonysigma:
Hi @maaz139 , thank you for [resolving the assertion](https://github.com/halide/Halide/pull/7753#discussion_r1290674781) `cost lookup failed` in the draft PR. The most recent version can now generate the Halide IR right ahead of the GPU/HVX/`parfor` offloading. The patch served my immediate needs. Thank you.

I am closing this PR for now, till a consensus is reached on the `stmt_html` requirements. Please move the discussions to #7519 .

Issue #7712, opened 2023-07-26
VizIR is super slow.


mcourteaux:
Loading a page takes me about 12 seconds on a high-end desktop.
I recorded what's happening in Chromium with the performance profiler in the Developer Tools.

[Trace-20230726T110312.json.zip](https://github.com/halide/Halide/files/12170163/Trace-20230726T110312.json.zip)

The first half of the time (below in green rectangle) is spent in `init_tooltips`, which calls out to some jQuery which takes *all* the time in `querySelectorAll`. The piece of offending code, according to the profile, is this one:
```js
        function init_tooltips(btns, prefix) {
            var re = /(?:\-([^-]+))?$/;
            for (var i = 0; i < btns.size(); i++) {
                const button = btns[i];
                const tooltip = $(prefix + re.exec(button.id)[1])[0];
                button.is_clicked = false;
                button.addEventListener('mouseenter', ()=>{
                    if (!button.is_clicked)
                        showTooltip(button, tooltip);
                }
                );
                button.addEventListener('mouseleave', ()=>{
                    if (!button.is_clicked)
                        hideTooltip(tooltip);
                }
                );
                button.addEventListener('click', ()=>{
                    if (button.is_clicked) {
                        hideTooltip(tooltip);
                        button.is_clicked = false;
                    } else {
                        showTooltip(button, tooltip);
                        button.is_clicked = true;
                    }
                }
                );
            }
        }
```
The culprit is the `$()` call, I assume. 

Second half of the time is spent in the same `querySelectorAll` but without a parent call.

![image](https://github.com/halide/Halide/assets/845012/afb1b313-5b65-488e-ad22-f46782d1657e)

mcourteaux:
@maaz139 I think you made this tool?

mcourteaux:
I'm working on this. Will open a Draft PR where I can document progress.

mcourteaux:
#7713

PR #7754, opened 2023-08-10
StmtViz: Search for tooltip only in the child node

antonysigma:
Further cut ~5 second of StmtVisualizer rendering by searching for the tooltip text-box in the child node of the current button. Previously, the script compose the global ID with regular expression, and then search the entire DOM causing delays.

Quoting the comments from #7713:
> ... It still is for like a second unresponsive every time you resize the panes, but I couldn't figure out how to fix that.

Resolves: https://github.com/halide/Halide/issues/7712 .
See also: #7713 .

cc'ed @maaz139  and @mcourteaux .

### DIFF
--- a/src/StmtToViz.cpp
+++ b/src/StmtToViz.cpp
@@ -1260,14 +1260,14 @@ private:
         }
 
         stream << "<div id='" << id << "' class='cost-btn CostColor" << line_costc << "'"
-               << "   aria-describedby='tooltip-" << id << "'"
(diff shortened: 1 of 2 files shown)


maaz139 (approved):
Good catch! Change looks good to me.

steven-johnson (approved):

mcourteaux:
Oooh, let's try this! Will report back soon.

PR #7755, opened 2023-08-11
Speedup page loading of VizStmt.

mcourteaux:
- Disabled line numbers in the syntax highlihgting of the assembly. This was a massive bottleneck upon opening the assembly tab.
 - ~~Made syntax highlighting on-demand with a button.~~
 - Also, inspired by @antonysigma, I replaced some of the element-grabbing in the `toggleViz` to directly take the element instead of running a selector query. This was a big speedup.
 - Updated scrolling logic to deal with the lack of line number divs. 
 - Prevented the last update recalculation when collapsing the assembly tab upon loading the page.
 - Deleted all the tooltip JavaScript, and now just make them appear on `:hover`. This significantly lowers the amount of "listeners" in the document. (Inspired by @antonysigma as well for this).
 - Finally, cleanup some dangling dependencies that were not used and are no longer used as per this PR.

Tagging @maaz139 @antonysigma @steven-johnson.

### DIFF
--- a/src/StmtToViz.cpp
+++ b/src/StmtToViz.cpp
@@ -2897,6 +2897,7 @@ private:
     // Generate tab 3/3: Generated assembly code
     void generate_assembly_tab(const Module &m) {
         stream << "<div id='assembly-tab'>\n";
+        stream << "<button id='highlight-button' onclick='highlightAssembly();'>Highlight</button>\n";
         stream << "<div id='assemblyContent' class='shj-lang-asm'>\n";
         stream << "<pre>\n";
         stream << asm_stream.str();
--- a/src/StmtToViz.cpp
+++ b/src/StmtToViz.cpp
@@ -2897,7 +2897,6 @@ private:
     // Generate tab 3/3: Generated assembly code
     void generate_assembly_tab(const Module &m) {
         stream << "<div id='assembly-tab'>\n";
-        stream << "<button id='highlight-button' onclick='highlightAssembly();'>Highlight</button>\n";
         stream << "<div id='assemblyContent' class='shj-lang-asm'>\n";
         stream << "<pre>\n";
         stream << asm_stream.str();
--- a/src/irvisualizer/StmtToViz_dependencies.template.html
+++ b/src/irvisualizer/StmtToViz_dependencies.template.html
@@ -17,7 +17,7 @@
 <script type="module">
   import { highlightAll } from 'https://cdn.jsdelivr.net/gh/speed-highlight/core/dist/index.js';
   import 'https://cdn.jsdelivr.net/gh/speed-highlight/core/dist/languages/asm.js';
-  window.highlightAll = highlightAll;
+  highlightAll({hideLineNumbers: true});
 </script>
 <!-- adapted from the default style from: https://unpkg.com/@speed-highlight/core@1.1.11/dist/themes/default.css -->
 <style type='text/css'>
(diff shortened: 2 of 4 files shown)


mcourteaux:
Follow up to #7754 by @antonysigma.

mcourteaux:
I wonder @maaz139, what's the use of the visualization of the IR. I don't look at it really. The only reason I HAVE to look at it is to click the "Jump to Assembly" buttons in the VizIR tree.

antonysigma:
> The only reason I HAVE to look at it is to click the "Jump to Assembly" buttons in the VizIR tree.

Likewise. If only we can have jump buttons between the IR view and the assembly view, like the [Compiler Explorer project](https://godbolt.org/z/y99L50).

I am preoccupied with #7519 and the draft PR right now, but I will join the discussion later this month.

mcourteaux:
Compatibility problem. Firefox does not support "computedStyleMap". F*$king webtech...

mcourteaux:
Fixed it.

mcourteaux:
I used the patch in this PR yesterday for a day, and I think that, given that the syntax highlighting became fast again (due to removing line numbers), I don't think having an extra click to run the highlighter on demand is useful. I'll change this to be highlighted by default again.

antonysigma:
I tested it today. Yes, the assembly code tab indeed is more responsive than before.

Attaching the demo here.
[demo.stmt.html.gz](https://github.com/halide/Halide/files/12383982/demo.stmt.html.gz)

mcourteaux:
@steven-johnson Is there a reason for waiting? I believe these changes are outside of the scope of the discussion about where to go with this VizIR thing. I have more improvements coming for this.

steven-johnson:
Sorry, didn't realize this was ready to land. Landing now.

steven-johnson (approved):

maaz139:
> I wonder @maaz139, what's the use of the visualization of the IR. I don't look at it really. The only reason I HAVE to look at it is to click the "Jump to Assembly" buttons in the VizIR tree.

I think Darya's idea was to offer a simplified view into the program that reveals high-level control/data flow. To what extent it achieves that goal, I think only user feedback like yours can tell us. Happy to hear about how you use the HTML IR file and if there are things you look for that we could summarize in the visualization.

mcourteaux:
> > I wonder @maaz139, what's the use of the visualization of the IR. I don't look at it really. The only reason I HAVE to look at it is to click the "Jump to Assembly" buttons in the VizIR tree.
> 
> I think Darya's idea was to offer a simplified view into the program that reveals high-level control/data flow. To what extent it achieves that goal, I think only user feedback like yours can tell us. Happy to hear about how you use the HTML IR file and if there are things you look for that we could summarize in the visualization.

@maaz139 

I'm always dragging it out of the way, and using it to Jump to the Assembly. I really haven't found any use for it. I hate to say it, but to me its useless and rather annoying. The code-like stmt HTML is what I am always looking at. I'm currently working on a PR which cleans this up, and will allow you to generate "stmt + assembly", and "stmt + vizir + assembly" versions. Both will feature direct jump-to-assembly buttons right from the stmt. The 

I was looking into generating more jump buttons, as after optimization passes sometimes the basic block labels that are generated during LLVM CodeGen are removed from the assembly output. I was trying to work out how to add proper debug information with source locations, but I can't seem to figure out this LLVM DIBuilder stuff and how to get it in the actual outputted `.o` and `.a` files.

As I am not optimistic about figuring out how to get the debug info working, I will probably leave that for a later attempt, and make a PR that only does the first bit.

Since I added the new jump-buttons (the blue ones in the Stmt code), I like to work like this:

image

mcourteaux:
Opened PR #7793.

PR #7760, opened 2023-08-14
[serialization] Add support to serialize to memory, and a basic serialization tutorial

derek-gerstmann:


src/Serialization.h
```diff
@@ -5,6 +5,16 @@
 
 namespace Halide {
 
+/**
+ * Serialize a Halide pipeline into the given data buffer.
+ * params is an optional map which can be used to bind external parameters to objects in the pipeline by name
+ */
+void serialize_pipeline(const Pipeline &pipeline, std::vector<uint8_t> &data, std::map<std::string, Internal::Parameter> &params);
```
steven-johnson, line 12: Food for future thought: this is in the `Halide` namespace and so is a public API... but it uses `Internal::Parameter`, so it really isn't quite public. Would be nice to figure out how to avoid needing to mix these.

tutorial/lesson_22_serialization.cpp
```diff
@@ -0,0 +1,169 @@
+// Halide tutorial lesson 22: Serialization
+
+// This lesson describes how to serialize pipelines into a binary format 
+// which can be saved on disk, and later deserialized and loaded for 
+// evaluation.
```
steven-johnson, line 5: We should add a note to the effect that "Serialization is *experimental* in Halide 17 and is subject to change; we recommend that you avoid relying on it for production work at this time". (On that note: I wonder if we should consider moving the API into something like `Halide::Experimental` namespace?)

tutorial/lesson_22_serialization.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+// On linux, you can compile this tutorial and run it like so:
+// g++ lesson_22*.cpp -g -I <path/to/Halide.h> -L <path/to/libHalide.so> -lHalide -lpthread -ldl -o lesson_22 -std=c++17
+// LD_LIBRARY_PATH=<path/to/libHalide.so> ./lesson_22
+
+// On os x:
+// g++ lesson_22*.cpp -g -I <path/to/Halide.h> -L <path/to/libHalide.so> -lHalide -o lesson_22 -std=c++17
+// DYLD_LIBRARY_PATH=<path/to/libHalide.dylib> ./lesson_22
+
+// If you have the entire Halide source tree, you can also build it by
```
steven-johnson, line 19: ...It makes me sad that our Tutorials are still relying entirely on Make, but that's a discussion for another day :-/

abadams:
A Julia set is a cool example, but I don't think it belongs in a tutorial about serialization. The tutorial about serialization shouldn't have anything that a reader needs to try to understand in it that's not either introduced in an earlier lesson or directly related to serialization (e.g. the Complex struct). It would be better to use e.g. the simpler blur pipeline from lesson 7, but using an ImageParam instead of the direct Buffer reference so that the tutorial can also handle reconnecting parameters.

src/Deserialization.h
```diff
@@ -15,8 +15,20 @@ namespace Halide {
  */
 Pipeline deserialize_pipeline(const std::string &filename, const std::map<std::string, Internal::Parameter> &external_params);
 
+/**
+ * Deserialize a Halide pipeline from an input stream.
```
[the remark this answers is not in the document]
derek-gerstmann: I'll make another pass at the doc strings and make sure they are Doxygen compatible.

src/Serialization.cpp
```diff
@@ -28,8 +28,12 @@ class Serializer {
 public:
     Serializer() = default;
 
+    // Serialize the given pipeline into the given filename
     void serialize(const Pipeline &pipeline, const std::string &filename);
 
+    // Serialize the given pipeline into given the data buffer
```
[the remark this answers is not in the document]
derek-gerstmann: It currently inserts the resulting binary serialization into the start of the vector.  I'll add a comment that the contents will get discarded, and update the code accordingly.

src/Serialization.cpp
```diff
@@ -1453,17 +1457,35 @@ void Serializer::serialize(const Pipeline &pipeline, const std::string &filename
 
     uint8_t *buf = builder.GetBufferPointer();
     int size = builder.GetSize();
+
+    if (buf != nullptr && size > 0) {
```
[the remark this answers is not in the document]
derek-gerstmann: Yep, sure!  See above ...

src/Serialization.h
derek-gerstmann: Good point ... it seems like we may need to revise this a little bit so its only using public class interfaces.

tutorial/lesson_22_serialization.cpp
derek-gerstmann: Okay!  I'll add update the comments.  Having an experimental namespace seems like a worthwhile thing ... maybe worth a separate PR.

tutorial/lesson_22_serialization.cpp
derek-gerstmann: They build with CMake too ...

derek-gerstmann:
> A Julia set is a cool example, but I don't think it belongs in a tutorial about serialization. The tutorial about serialization shouldn't have anything that a reader needs to try to understand in it that's not either introduced in an earlier lesson or directly related to serialization (e.g. the Complex struct). It would be better to use e.g. the simpler blur pipeline from lesson 7, but using an ImageParam instead of the direct Buffer reference so that the tutorial can also handle reconnecting parameters.

Yeah, I went back and forth on this ... I'm happy to change it to a simple blur.  My reasoning was that we should have a reasonably complex pipeline to serialize, but I agree that there's too much unrelated material for a tutorial about serialization.  I'll change it!

src/Serialization.cpp
```diff
@@ -1453,17 +1457,35 @@ void Serializer::serialize(const Pipeline &pipeline, const std::string &filename
...
+        result.reserve(size);
+        result.insert(result.begin(), buf, buf + size);
+    } else {
+        user_error << "failed to serialize pipeline!\n";
+    }
+}
+
+void Serializer::serialize(const Pipeline &pipeline, const std::string &filename) {
+    std::vector<uint8_t> data;
```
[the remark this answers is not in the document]
derek-gerstmann: It looks like the Flatbuffer API has limited streaming functionality ... it requires you to always read/write to/from an in-memory data buffer.  But it can tell you how large the buffer should be on a per-object basis, so you could do per object data buffer allocations and read/write to a stream manually.

abadams:
My understanding was that it was supposed to create new Parameters for you in that case.

abadams:
Looks like Param<> acts as a scalar Param with a dynamic type, so for the Internal::Parameter API issue I think the methods should take separate maps for scalar and buffer params, e.g:

```
Pipeline deserialize_pipeline(const std::string &filename, 
    const std::map<std::string, OutputImageParam> &external_buffer_params, 
    const std::map<std::string, Param<>> &external_scalar_params);
```

Open to other opinions though. What we discussed in the dev meeting was just making Internal::Parameter public, but that increases our public API surface.

TH3CHARLie:
That sounds like a good plan, and can be implemented with minimal changes

steven-johnson:
Can we please introduce aliases for these maps and use them everywhere, since these are unwieldy, eg

```
using BufferParamMap = std::map<std::string, OutputImageParam>;
using ScalarParamMap = std::map<std::string, Param<>>;

Pipeline deserialize_pipeline(const std::string &filename, 
    const BufferParamMap &external_buffer_params, 
    const ScalarParamMap> &external_scalar_params);
```

steven-johnson:
> Open to other opinions though. What we discussed in the dev meeting was just making Internal::Parameter public, but that increases our public API surface.

I'm starting to think that might be preferable, assuming we do the relevant API cleanup on Parameter per the discussion

steven-johnson:
This PR doesn't appear to be very close to ready -- decisions about making `Parameter` a public API are up in the air, and the bots are largely broken. Should it be converted to draft?

derek-gerstmann:
Yes, I’ll make this to a draft until we sort out the API changes.

derek-gerstmann:
Added https://github.com/halide/Halide/issues/7862

derek-gerstmann:
Updated to include changes to make parameter map optional for serialize(...) #7849

src/Serialization.h
derek-gerstmann: Updated

derek-gerstmann:
Ready to merge ... we'll address https://github.com/halide/Halide/issues/7862 in another PR.

derek-gerstmann:
@alexreinking @abadams Friendly ping for approval

alexreinking (approved):

PR #7769, opened 2023-08-16
Define `cast<i32>(u32)` overflow behavior

rootjalex:
The simplifier normalizes `reinterpret<i32>(u32)` (which has defined behavior) to `cast<i32>(u32)`, which has undefined behavior. Currently, the following code produces inconsistent output:
```
{
    Expr x = UIntImm::make(UInt(32), 0xFFFFFFF0u);
    Expr e = i32(x + 1);
    e = simplify(e);
    std::cout << e << std::endl;

    e = reinterpret(Int(32), x + 1);
    e = simplify(e);
    std::cout << e << std::endl;
}
{
    Expr x = Variable::make(UInt(32), "x");
    Expr e = i32(x + 1);
    e = simplify(e);
    e = substitute("x", UIntImm::make(UInt(32), 0xFFFFFFF0), e);
    e = simplify(e);
    std::cout << e << std::endl;

    e = reinterpret(Int(32), x + 1);
    e = simplify(e);
    e = substitute("x", UIntImm::make(UInt(32), 0xFFFFFFF0), e);
    e = simplify(e);
    std::cout << e << std::endl;
}
```
Output:
```
signed_integer_overflow(0)
-15
signed_integer_overflow(1)
signed_integer_overflow(2)
```

@abadams and I agreed to simply define `cast<i32>(u32)` to have the same 2's complement wrapping behavior as `reinterpret<i32>(u32)`. I think the only place this behavior is encoded is in the simplifier (LLVM codegen produces a no-op for either case because LLVM types don't have signedness).

Now the output of the above code is:
```
-15
-15
-15
-15
```

### DIFF
--- a/src/Simplify_Cast.cpp
+++ b/src/Simplify_Cast.cpp
@@ -74,14 +74,21 @@ Expr Simplify::visit(const Cast *op, ExprInfo *bounds) {
         } else if (op->type.is_int() &&
                    const_uint(value, &u) &&
                    op->type.bits() < value.type().bits()) {
-            // uint -> int
+            // uint -> int narrowing
             // Recursively call mutate just to set the bounds
             return mutate(make_const(op->type, safe_numeric_cast<int64_t>(u)), bounds);
         } else if (op->type.is_int() &&
                    const_uint(value, &u) &&
-                   op->type.bits() >= value.type().bits()) {
-            // uint -> int with less than or equal to the number of bits
+                   op->type.bits() == value.type().bits()) {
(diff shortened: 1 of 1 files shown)


steven-johnson:
SGTM

steven-johnson:
No idea if these are related to this change or not, but:

https://github.com/halide/Halide/issues/7810
https://github.com/halide/Halide/issues/7811

(Just reported by Google-internal fuzz testing after our most recent merge from GitHub)

PR #7771, opened 2023-08-17
Fix out of bounds access in anderson2021_test_apps_autoscheduler

aekul:
This fixes https://github.com/halide/Halide/issues/7606. If an update stage has fewer pure vars than the pure definition, it can end up reading out of bounds. When reading the size of the update stage's loop, instead of using the corresponding pure dimension, it should just use its index within the update stage's loops.


steven-johnson:
Thanks for the fix!

PR #7776, opened 2023-08-18
Don't introduce reinterprets in find/lower intrinsics

rootjalex:
As discussed with @abadams , later lowering stages expect to see int/uint reinterprets to have been normalized to casts. This PR removes the places in `find_intrinsics` and `lower_intrinsics` that introduce `reinterpret`, and use `cast` instead.

### DIFF
--- a/src/FindIntrinsics.cpp
+++ b/src/FindIntrinsics.cpp
@@ -157,16 +157,19 @@ Expr to_rounding_shift(const Call *c) {
                 return rounding_shift(cast(add->type, add->args[0]), b);
             }
         }
-        // Also need to handle the annoying case of a reinterpret wrapping a widen_right_add
+
+        // Also need to handle the annoying case of a reinterpret cast wrapping a widen_right_add
         // TODO: this pattern makes me want to change the semantics of this op.
-        if (const Reinterpret *reinterp = a.as<Reinterpret>()) {
-            if (reinterp->type.bits() == reinterp->value.type().bits()) {
-                if (const Call *add = Call::as_intrinsic(reinterp->value, {Call::widen_right_add})) {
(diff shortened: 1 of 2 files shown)


rootjalex:
Failures appear unrelated

abadams (approved):

pranavb-ca:
Thank you for this @rootjalex. @prasmish- Isn't this essentially what we were talking about this morning? Can you please take a look if it helps your case?

rootjalex:
@pranavb-ca No problem!

Btw, I have a PR that I am working on that is aiming to improve handling of reinterpret/casts that do reinterprets, for ARM and HVX codegen. If you / your team are working on something similar, it might be useful to coordinate. I've been running into a lot of headaches with HexagonOptimize for that PR.

rootjalex:
An example:
```
x_i16 + widening_mul(y_u8, (int8)3)
```
On ARM, it's better to rewrite this to:
```
x_i16 + i16(widening_mul(y_u8, (uint8)3))
```
and target `umlal`. However, on HVX, this pattern maps to `vmpybusv_acc`, so we shouldn't lift to the reinterpreted `widening_mul` (although, for this particular example, `vmpyubv_acc` could be used for the operation as well). The general theme is this: on non-HVX architectures, it's almost always better to rewrite `widening_mul(u, i)` to `reinterpret(widening_mul(u, u))` if possible (i.e. the signed operand is a positive constant), but this causes all sorts of issues with HVX codegen, which is generally better at handling mixed-signed `widening_mul`. I almost want to fix this by having `find_intrinsics` accept the `Target` and only apply certain rules if not targeting HVX

pranavb-ca:
Thanks, @rootjalex. 
@prasmish is working on this on our end. I have spoken to him and he'll keep you posted.

PR #7783, opened 2023-08-18
Profiler should report Func count as well as time

steven-johnson:
Sometimes it's useful to know how many times a Func is evaluated, independent of its time. This adds that to the profiler.


abadams:
That's not a realization count though, that's a count of how many times the profiler woke up and that particular func was running. It should be strictly less informative than the time, because it adds profiler wake-up timer noise.

steven-johnson:
> how many times the profiler woke up and that particular func was running

...yeah, I see that now, alas (after using it a bit and seeing results that aren't scaling in the way I expected). Phooey. Closing this.

Issue #7491, opened 2023-04-10
How to evaluate/compare the GPU auto-scheduler performance btw Anderson2021 and Mullapudi2016?


antonysigma:
Hi I noticed the PR #5602 has been merged to the main branch. Thank you!

If I were to compare the performance between `Li2018` and `Anderson2021`, how do I configure `machine_param=n_threads,cache_size,compute_cache_tradeoff` accordingly to equalize the autoscheduler environment?

Similarly, how do I configure the `machine_param` to compare the performance between `Mullapudi2016` and `Anderson2021`?

---

To clarify, the current version of `Maullapudi2016` can only generate CPU schedules. But, the original paper suggested a manual intervention step to generate GPU schedules. That is, to edit the generated `*.schedule.h` code to replace all `.vectorize(...).parallel(...)` with `gpu_threads(...).gpu_tiles(...)` in accordance to the following rules. In the past, I found it highly effective for algorithms having less than ~5 Halide stages.

> We configure the auto-scheduler to target the GPU by setting the PARALLELISM_THRESHOLD to 128, VECTOR_WIDTH to 32, and
CACHE_SIZE to 48 KB. Additionally, we add two new parameters
TARGET_THREADS_PER_BLOCK and MAX_THREADS_PER_BLOCK whose values are set to 128 and 2048 respectively. These parameters enable the
auto-scheduler to avoid tiling configurations that generate too few
or too many threads per GPU thread block. The inlining, tiling, and
grouping processes are otherwise similar to the CPU case. Groups
resulting from merging are mapped to CUDA kernels by designating
the outer tile loops as GPU block grid dimensions and the inner tile
loops as GPU thread block dimensions. All intermediate buffers
within a group are allocated in GPU shared memory.
http://graphics.cs.cmu.edu/projects/halidesched/

antonysigma:
> how do I configure `machine_param=n_threads,cache_size,compute_cache_tradeoff` accordingly to equalize the autoscheduler environment?

Found it. since Halide 16.0, the `machine_param` command line argument becomes a named struct. Previously, we specify the params with comma separated tuple.

https://github.com/halide/Halide/blob/083927020a594a66f6e45ccd407ad922d867906b/src/autoschedulers/anderson2021/CostModel.h#L18

Closing this thread.

Issue #7696, opened 2023-07-21
We should fix Generator warnings in autoscheduler and apps


steven-johnson:
When building these, we get a large number of warnings of the form

```
Warning: Insufficient parallelism
Warning: Outer dim vectorization
Warning: Update definition 0 of function foo has not been scheduled
```

...these are warnings that got added to Halide after the code in question was written, but we never went back to fix the warnings.  :-( We really should strive to have all of Halide build with zero warnings.

abadams:
The first two examples are from the Mullapudi autoscheduler and are not actionable. We should demote to autoscheduler diagnostics. The third one is something we should fix in the app.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_FIXED):** All three named warnings have been dealt with as the thread proposed: the two Mullapudi2016 warnings were demoted to `debug(1)` diagnostics, and the apps no longer emit the unscheduled-update warning in practice.

**Evidence:** `src/autoschedulers/mullapudi2016/AutoSchedule.cpp:3432` is now `debug(1) << "Insufficient parallelism for "` (demoted in `9f6ec17ac`, #7990) and `:3059` is now `debug(1) << "Outer dim vectorization of var "` (demoted in `5aa891a78`, #7992); both were `user_warning` as of `71eb4ee7f` (2023-07-20). Built 21 apps (bilateral_grid, camera_pipe, lens_blur, max_filter, interpolate, nl_means, stencil_chain, harris, unsharp, conv_layer, bgu, iir_blur, hist, resize, depthwise_separable_conv, fft, resnet_50, blur, local_laplacian, linear_algebra) from a clean state: zero `Warning:` lines from any Generator. Several apps carry explicit `.unscheduled()` calls (e.g. `apps/hist/hist_generator.cpp:184`, `apps/iir_blur/iir_blur_generator.cpp:57`, `apps/fft/fft.cpp:875`).

[...]

PR #7787, opened 2023-08-21
GPU autoscheduling with Mullapudi2016: the reference implementation

antonysigma:
**Rationale:**

1. To compare the GPU auto-scheduling performance of `Mullapudi2016` against `Li2018` and `Anderson2021`.

2. To reduce the following claims to practice, quoting the original Mullapudi2016 article:

>  **Portability to Different Architectures: GPU Performance:** The inlining, tiling, and grouping processes are otherwise similar to the CPU case. Groups resulting from merging are mapped to CUDA kernels by designating the outer tile loops as GPU block grid dimensions and the inner tile loops as GPU thread block dimensions. All intermediate buffers
within a group are allocated in GPU shared memory

3. To implement the so-call "single level tiling only" limitation in the `Mullapudi2016` and `Sioutas2020` algorithms, according to the findings in the Anderson2021 paper:

> [Mullapudi et al] develops an automatic scheduling technique using a heuristic cost model and a greedy stage grouping algorithm... but its search space is smaller compared to ours among other reasons because it only supports a single level of tiling, and as we discuss in Section 6.2, this excludes a number of high performance schedules. 

---
**Change summary:**

Reverse engineer the GPU scheduling feature as stated in Section 5.4 of Mullapudi's article:

Mullapudi, Adams, Sharlet, Ragan-Kelley, Fatahalian. Automatically scheduling Halide image processing pipelines. ACM Transactions on Graphics, 35(4), 83pp 1–11. https://doi.org/10.1145/2897824.2925952

When `target=cuda` is detected in the code generator command line arguments, intercept all `vectorize`, `parallel` scheduling calls requested by the auto-vectorization algorithm and the auto-parallelization algo with the class `GPUTilingDedup` for deferred execution.

Implement the class `GPUTilingDedup` to ensure all Halide gpu schedule calls are idempotent: no matter how many times the Stage is `vectorized`, `reordered`, `parallel`, and then `reordered` again, the `reorder` and `gpu_threads()` schedules are called exactly once.

Also, intercept all `split` and `reorder` scheduling calls by Mullapudi's auto-splitting algorithm.

Implement the clss `GPUTileHelper` to enforce atomic transaction of the gpu schedules. If the current stage is `compute_root`, mark all auto-split inner dimensions as `gpu_threads`, and outer dimensions as `gpu_blocks`. If the Stage is `compute_at` another Stage, mark all `vectorize` dimensions as `gpu_threads`.

If auto-splitting of the current stage does not result in any tile, implement a rudimentary tiling having tile size = vector_length x parallel_factor.

If Mullapudi does not call any split, vectorize, or parallel schedules, assume scalar reduction routine. Implement it on the GPU via `single_thread`.

cc'ed @aekul , @jrk, @abadams .

**See also:** https://github.com/halide/Halide/issues/7491

### DIFF
--- a/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
+++ b/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
@@ -1514,10 +1514,12 @@ struct Partitioner {
     // Bounds of each function stage in the pipeline. These bounds are inferred from the
     // estimates of the outputs and other functions in the pipeline.
     const map<string, Box> &pipeline_bounds;
+    // The active target
+    const Target &target;
     // Parameters of the machine model that is used for estimating the cost of each
     // group in the pipeline.
     const ArchParams &arch_params;
-    // Dependency analysis of the pipeline. This support queries on regions
+    // Dependency analysis of the pipeline. This supports queries on regions
     // accessed and computed for producing some regions of some functions.
     DependenceAnalysis &dep_analysis;
     // The arithmetic and memory costs of evaluating the expressions which define
@@ -1527,6 +1529,7 @@ struct Partitioner {
     const vector<Function> &outputs;
 
     Partitioner(const map<string, Box> &_pipeline_bounds,
+                const Target &_target,
                 const ArchParams &_arch_params,
                 const vector<Function> &_outputs,
                 DependenceAnalysis &_dep_analysis,
@@ -1638,12 +1641,12 @@ struct Partitioner {
     // visible to the user. Additionally, functions like sum and maximum are not
     // user visible. More thought needs to go into interaction between the user and
     // auto scheduling.
-    void generate_cpu_schedule(const Target &t, AutoSchedule &sched);
+    void generate_cpu_schedule(AutoSchedule &sched);
 
     // Same as \ref Partitioner::generate_cpu_schedule, but this generates and
     // applies schedules for a group of function stages.
 
-    void generate_group_cpu_schedule(const Group &g, const Target &t,
+    void generate_group_cpu_schedule(const Group &g,
                                      const map<FStage, DimBounds> &group_loop_bounds,
                                      const map<string, Box> &group_storage_bounds,
                                      const set<string> &inlines,
@@ -1656,13 +1659,13 @@ struct Partitioner {
         const Group &g, Stage f_handle, int stage_num, const Definition &def,
         bool is_group_output, const VarOrRVar &v, const Expr &factor, const string &in_suffix,
         const string &out_suffix, map<string, Expr> &estimates, AutoSchedule &sched,
-        const Target &t, GPUTilingDedup *gpu_tiling = nullptr);
+        GPUTilingDedup *gpu_tiling = nullptr);
 
     // Loop over the dimensions of function stage 'f_handle' starting from innermost
     // and vectorize the first pure dimension encountered.
     std::optional<pair<VarOrRVar, VarOrRVar>> vectorize_stage(
         const Group &g, Stage f_handle, int stage_num, Definition def,
-        const Function &func, bool is_group_output, const Target &t, set<string> &rvars,
+        const Function &func, bool is_group_output, set<string> &rvars,
         map<string, Expr> &estimates, AutoSchedule &sched, GPUTilingDedup &gpu_tiling);
 
     // Reorder the dimensions to preserve spatial locality. This function
@@ -1671,7 +1674,7 @@ struct Partitioner {
     // This takes the strides along each dimension as input.
     void reorder_dims(Stage f_handle, int stage_num, Definition def,
                       map<string, Expr> strides, AutoSchedule &sched,
-                      const Target &t, GPUTilingDedup &gpu_tiling);
+                      GPUTilingDedup &gpu_tiling);
 
     // Helper functions to display partition information of the pipeline.
     void disp_pipeline_costs();
@@ -1769,11 +1772,12 @@ void Partitioner::disp_pipeline_costs() {
 // Construct a partitioner and build the pipeline graph on which the grouping
 // algorithm operates.
 Partitioner::Partitioner(const map<string, Box> &_pipeline_bounds,
+                         const Target &_target,
                          const ArchParams &_arch_params,
                          const vector<Function> &_outputs,
                          DependenceAnalysis &_dep_analysis,
                          RegionCosts &_costs)
-    : pipeline_bounds(_pipeline_bounds), arch_params(_arch_params),
+    : pipeline_bounds(_pipeline_bounds), target(_target), arch_params(_arch_params),
       dep_analysis(_dep_analysis), costs(_costs), outputs(_outputs) {
     // Place each stage of a function in its own group. Each stage is
     // a node in the pipeline graph.
@@ -2775,7 +2779,7 @@ pair<VarOrRVar, VarOrRVar> Partitioner::split_dim(
     const Group &g, Stage f_handle, int stage_num, const Definition &def,
     bool is_group_output, const VarOrRVar &v, const Expr &factor, const string &in_suffix,
     const string &out_suffix, map<string, Expr> &estimates, AutoSchedule &sched,
-    const Target &t, GPUTilingDedup *gpu_tiling) {
+    GPUTilingDedup *gpu_tiling) {
     // Create new variables for the split dimensions
     string arg_name = v.name();
     string inner_name = arg_name + in_suffix;
@@ -2824,7 +2828,7 @@ pair<VarOrRVar, VarOrRVar> Partitioner::split_dim(
         strategy = TailStrategy::GuardWithIf;
     }
 
-    if (t.has_gpu_feature() && gpu_tiling) {
+    if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule && gpu_tiling) {
         gpu_tiling->has_split(v, outer, inner, factor, strategy);
     }
 
@@ -2852,7 +2856,7 @@ pair<VarOrRVar, VarOrRVar> Partitioner::split_dim(
 
 std::optional<pair<VarOrRVar, VarOrRVar>> Partitioner::vectorize_stage(const Group &g, Stage f_handle, int stage_num,
                                                                        Definition def, const Function &func, bool is_group_output,
-                                                                       const Target &t, set<string> &rvars,
+                                                                       set<string> &rvars,
                                                                        map<string, Expr> &estimates, AutoSchedule &sched,
                                                                        GPUTilingDedup &gpu_tiling) {
     vector<Dim> &dims = def.schedule().dims();
@@ -2861,7 +2865,7 @@ std::optional<pair<VarOrRVar, VarOrRVar>> Partitioner::vectorize_stage(const Gro
     // Set the vector length as the maximum of the natural vector size of all
     // values produced by the function.
     const auto vec_len = [&]() -> int {
-        if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+        if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
             /** Section 5.4 of the Mullapudi2016 article: We configure the
              * auto-scheduler to target the GPU by set- ting the ...,
              * VECTOR_WIDTH to 32.
@@ -2871,7 +2875,7 @@ std::optional<pair<VarOrRVar, VarOrRVar>> Partitioner::vectorize_stage(const Gro
 
         int vec_len = 0;
         for (const auto &type : func.output_types()) {
-            vec_len += t.natural_vector_size(type);
+            vec_len += target.natural_vector_size(type);
         }
         return vec_len;
     }();
@@ -2898,14 +2902,14 @@ std::optional<pair<VarOrRVar, VarOrRVar>> Partitioner::vectorize_stage(const Gro
 
         VarOrRVar vec_var(vec_dim_name, is_rvar);
         auto [inner, outer, accepted] = [&]() -> std::tuple<VarOrRVar, VarOrRVar, bool> {
-            if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+            if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
                 VarOrRVar inner{vec_var.name() + "_vi", vec_var.is_rvar}, outer{vec_var.name() + "_vo", vec_var.is_rvar};
                 const bool accepted = gpu_tiling.can_vectorize(vec_var, outer, inner, vec_len);
                 return {inner, outer, accepted};
             }
 
             auto split_vars = split_dim(g, f_handle, stage_num, def, is_group_output, vec_var, vec_len,
-                                        "_vi", "_vo", estimates, sched, t);
+                                        "_vi", "_vo", estimates, sched);
 
             f_handle.vectorize(split_vars.first);
             sched.push_schedule(f_handle.name(), stage_num,
@@ -2960,7 +2964,7 @@ inline bool operator!=(const vector<Dim> &dims, const vector<VarOrRVar> &orderin
 }
 
 void Partitioner::reorder_dims(Stage f_handle, int stage_num, Definition def,
-                               map<string, Expr> strides, AutoSchedule &sched, const Target &t, GPUTilingDedup &gpu_tiling) {
+                               map<string, Expr> strides, AutoSchedule &sched, GPUTilingDedup &gpu_tiling) {
     vector<Dim> &dims = def.schedule().dims();
     internal_assert(dims.size() > 1);
     vector<pair<string, int>> order;
@@ -3055,7 +3059,7 @@ void Partitioner::reorder_dims(Stage f_handle, int stage_num, Definition def,
     }
 
     if (dims != ordering) {
-        if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+        if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
             gpu_tiling.canReorder(ordering);
         } else {
             f_handle.reorder(ordering);
@@ -3085,7 +3089,7 @@ public:
 };
 
 void Partitioner::generate_group_cpu_schedule(
-    const Group &g, const Target &t,
+    const Group &g,
     const map<FStage, DimBounds> &group_loop_bounds,
     const map<string, Box> &group_storage_bounds,
     const set<string> &inlines,
@@ -3149,7 +3153,7 @@ void Partitioner::generate_group_cpu_schedule(
         map<string, Expr> strides =
             analyze_spatial_locality(g.output, group_storage_bounds, inlines);
         if (!strides.empty()) {
-            reorder_dims(f_handle, g.output.stage_num, def, strides, sched, t, gpu_tiling);
+            reorder_dims(f_handle, g.output.stage_num, def, strides, sched, gpu_tiling);
         }
     }
 
@@ -3173,7 +3177,7 @@ void Partitioner::generate_group_cpu_schedule(
             } else {
                 pair<VarOrRVar, VarOrRVar> tile_vars =
                     split_dim(g, f_handle, g.output.stage_num, def, true, v,
-                              tile_size, "_i", "_o", stg_estimates, sched, t, &gpu_tiling);
+                              tile_size, "_i", "_o", stg_estimates, sched, &gpu_tiling);
 
                 inner_dims.push_back(tile_vars.first);
                 outer_dims.push_back(tile_vars.second);
@@ -3208,7 +3212,7 @@ void Partitioner::generate_group_cpu_schedule(
         }
 
         if (dims != ordering) {
-            if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+            if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
                 gpu_tiling.canReorder(ordering);
             } else {
                 f_handle.reorder(ordering);
@@ -3219,10 +3223,10 @@ void Partitioner::generate_group_cpu_schedule(
     }
 
     {
-        auto vectorized_split = vectorize_stage(g, f_handle, g.output.stage_num, def, g_out, true, t,
+        auto vectorized_split = vectorize_stage(g, f_handle, g.output.stage_num, def, g_out, true,
                                                 rvars, stg_estimates, sched, gpu_tiling);
 
-        if (t.has_gpu_feature() && vectorized_split && arch_params.experimental_gpu_schedule) {
+        if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule && vectorized_split) {
             auto [v_i, v_o] = *vectorized_split;
             inner_dims.emplace_back(v_i);
             outer_dims.emplace_back(v_o);
@@ -3268,7 +3272,7 @@ void Partitioner::generate_group_cpu_schedule(
             if ((iter != stg_estimates.end()) && iter->second.defined()) {
                 if (!seq_var.empty()) {
                     VarOrRVar seq(seq_var, (rvars.find(seq_var) != rvars.end()));
-                    if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+                    if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
                         gpu_tiling.canReorder({seq, v});
                     } else {
                         f_handle.reorder(seq, v);
@@ -3277,7 +3281,7 @@ void Partitioner::generate_group_cpu_schedule(
                                             {seq_var, var});
                     }
                 }
-                if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+                if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
                     auto parallelized_split = gpu_tiling.can_parallelize(v, iter->second);
                     if (parallelized_split) {
                         auto split_vars = *parallelized_split;
@@ -3300,7 +3304,7 @@ void Partitioner::generate_group_cpu_schedule(
         debug(1) << "Insufficient parallelism for " << f_handle.name() << "\n";
     }
 
-    if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+    if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
         gpu_tiling.apply(sched);
     }
 
@@ -3382,20 +3386,20 @@ void Partitioner::generate_group_cpu_schedule(
             map<string, Expr> mem_strides =
                 analyze_spatial_locality(mem, group_storage_bounds, inlines);
             if (!mem_strides.empty()) {
-                reorder_dims(mem_handle, mem.stage_num, mem_def, mem_strides, sched, t, gpu_tiling2);
+                reorder_dims(mem_handle, mem.stage_num, mem_def, mem_strides, sched, gpu_tiling2);
             }
         }
 
         vectorize_stage(g, mem_handle, mem.stage_num, mem_def, mem.func, false,
-                        t, mem_rvars, mem_estimates, sched, gpu_tiling2);
+                        mem_rvars, mem_estimates, sched, gpu_tiling2);
 
-        if (t.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
+        if (target.has_gpu_feature() && arch_params.experimental_gpu_schedule) {
             gpu_tiling2.apply(sched);
         }
     }
 }
 
-void Partitioner::generate_cpu_schedule(const Target &t, AutoSchedule &sched) {
+void Partitioner::generate_cpu_schedule(AutoSchedule &sched) {
     // Grab the group bounds early as they rely on the dimensions of the group
     // outputs which will be altered by modifying schedules.
     map<FStage, map<FStage, DimBounds>> loop_bounds = group_loop_bounds();
@@ -3419,7 +3423,7 @@ void Partitioner::generate_cpu_schedule(const Target &t, AutoSchedule &sched) {
 
     // Realize schedule for each group in the pipeline.
     for (const auto &g : groups) {
-        generate_group_cpu_schedule(g.second, t, get_element(loop_bounds, g.first),
+        generate_group_cpu_schedule(g.second, get_element(loop_bounds, g.first),
                                     get_element(storage_bounds, g.first), inlines, sched);
     }
 }
@@ -3845,7 +3849,7 @@ string generate_schedules(const vector<Function> &outputs, const Target &target,
     }
 
     debug(2) << "Initializing partitioner...\n";
-    Partitioner part(pipeline_bounds, arch_params, outputs, dep_analysis, costs);
+    Partitioner part(pipeline_bounds, target, arch_params, outputs, dep_analysis, costs);
 
     // Compute and display reuse
     /* TODO: Use the reuse estimates to reorder loops
@@ -3895,7 +3899,7 @@ string generate_schedules(const vector<Function> &outputs, const Target &target,
     debug(2) << "Initializing AutoSchedule...\n";
     AutoSchedule sched(env, top_order);
     debug(2) << "Generating CPU schedule...\n";
-    part.generate_cpu_schedule(target, sched);
+    part.generate_cpu_schedule(sched);
 
     std::ostringstream oss;
     oss << sched;
--- a/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
+++ b/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
@@ -1226,7 +1226,7 @@ public:
     }
 
     /** Indicate the default dimension order of the Func. */
-    void setInitialOrder(const Func &func) {
+    void set_initial_order(const Func &func) {
         debug(2) << f.name() << ".initialOrder()\n";
 
         ordering.clear();
@@ -3140,7 +3140,7 @@ void Partitioner::generate_group_cpu_schedule(
     }
 
     GPUTilingDedup gpu_tiling{false, f_handle, g.output.stage_num};
-    gpu_tiling.setInitialOrder(Func(g_out));
+    gpu_tiling.set_initial_order(Func(g_out));
 
     // Reorder the dimensions for better spatial locality (i.e. smallest stride
     // is innermost). If we only have one dimension (excluding __outermost),
@@ -3366,7 +3366,7 @@ void Partitioner::generate_group_cpu_schedule(
             }
         }
         GPUTilingDedup gpu_tiling2{true, mem_handle, mem.stage_num};
-        gpu_tiling2.setInitialOrder(Func(mem.func));
+        gpu_tiling2.set_initial_order(Func(mem.func));
 
         // Reorder the dimensions for better spatial locality. If we only have
         // one dimension (excluding __outermost), there is nothing to reorder.
--- a/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
+++ b/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
@@ -46,7 +46,7 @@ struct ArchParams {
     ArchParams(const Target &target, const AutoschedulerParams &params_in) {
         ParamParser parser(params_in.extra);
 
-        bool experimental_gpu_schedule;
+        bool experimental_gpu_schedule = false;
         parser.parse("experimental_gpu_schedule", &experimental_gpu_schedule);
 
         is_gpu_schedule = target.has_gpu_feature() && experimental_gpu_schedule;
--- a/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
+++ b/src/autoschedulers/mullapudi2016/AutoSchedule.cpp
@@ -24,14 +24,27 @@ namespace {
 
 struct ArchParams {
     /** Maximum level of parallelism avalaible. */
-    int parallelism = 16;
+    int parallelism{};
 
     /** Size of the last-level cache (in bytes). */
-    uint64_t last_level_cache_size = 16 * 1024 * 1024;
+    uint64_t last_level_cache_size{};
 
     /** Indicates how much more expensive is the cost of a load compared to
      * the cost of an arithmetic operation at last level cache. */
-    float balance = 40;
+    float balance{};
+
+    /** If GPU target is detected, but machine parameters are not specified, *
+     * make a realistic estimate based on consumer-grade GPUs (Nvidia GTX *
+     * 1660/Turing), or low-cost scientific-grade GPUs (Nvidia K40/Tesla).
+     *
+     * Section 5.4 of the Mullapudi2016 article: We configure the auto-scheduler
+     * to target the GPU by setting the PARALLELISM_THRESHOLD to 128, ..., and
+     * CACHE_SIZE to 48 KB.
+     */
+    constexpr ArchParams(bool has_gpu_feature)
+        : parallelism(has_gpu_feature ? 128 : 16), last_level_cache_size(has_gpu_feature ? 48 * 1024 : 16 * 1024 * 1024),
+          balance(has_gpu_feature ? 20 : 40) {
+    }
 };
 
 // Substitute parameter estimates into the exprs describing the box bounds.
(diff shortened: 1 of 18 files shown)


src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -2789,25 +3221,37 @@ void Partitioner::generate_group_cpu_schedule(
...
             } else {
                 break;
             }
         }
     }
 
-    if (can_prove(def_par < arch_params.parallelism)) {
+    if (can_prove(def_par < arch_params.parallelism && def_par != 1)) {
         user_warning << "Insufficient parallelism for " << f_handle.name() << "\n";
     }
```
antonysigma, line 3249: Question for reviewers: why are we permitting `parallel(y)` having `Func.bound(y, 1)`, i.e. having an extent of one?

See also: #7696 .

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+                << z.outer.name() << "}, {"  //
+                << x.inner.name() << ", "
+                << y.inner.name() << ", "
+                << z.inner.name() << "}, {"  //
+                << x.factor << ", "
+                << y.factor << ", "
+                << z.factor << "})";
+
+            if (is_compute_at) {
+                f.gpu_threads(x.inner);
```
abadams, line 1009: This could be f.gpu_threads(x.inner, y.inner, z.inner)

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+                << y.v.name() << ", "  //
+                << x.outer.name() << ", "
+                << y.outer.name() << ", "  //
+                << x.inner.name() << ", "
+                << y.inner.name() << ", "  //
+                << x.factor << ", "
+                << y.factor << ")";
+
+            if (is_compute_at) {
+                f.gpu_threads(x.inner);
```
abadams, line 975: This could be f.gpu_threads(x.inner, y.inner)

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+
+    bool is_initial_order = true;
+    std::vector<VarOrRVar> ordering;
+
+    std::set<std::string> has_split;
+    std::set<std::string> outer_vars;
+    std::set<std::string> inner_vars;
+
+    /** True if Func::parallelize(v_o) is already handled by gpu_blocks() before. */
+    bool isOuter(const std::string &variable_name) const {
```
abadams, line 1088: We use snake case for methods: is_outer

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+        return outer_vars.find(variable_name) != outer_vars.end();
+    }
+
+    /** True if Func::vectorize(v_i) is already handled by gpu_threads() before. */
+    bool isInner(const std::string &variable_name) const {
```
abadams, line 1093: is_inner

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+        return inner_vars.find(variable_name) != inner_vars.end();
+    }
+
+    bool isUpdate() const {
```
abadams, line 1097: is_update

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+        return f.name().find("update") != std::string::npos;
+    }
+
+    void markGPUThreads(AutoSchedule &sched) const {
```
abadams, line 1101: mark_gpu_threads

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+            // vectorized dimension is treated as a thread in GPU. No need to
+            // further split it to match the natural_vector_size() of CPUs.
+            inner_vars.emplace(v.name());
+            return;
+        }
+
+        parallelize.try_emplace(var, split_t{v, vo, vi, factor, TailStrategy::Auto});
+    }
+
+    /** Mark the current dimension is already splitted by Mullapudi2016's
```
abadams, line 1176: past tense of "split" is just "split", not "splitted"

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -880,6 +880,413 @@ struct AutoSchedule {
...
+
+            split_t new_entry{entry};
+            new_entry.factor = simplify(min(threads_budget, new_entry.factor));
+
+            helper.applySplit(new_entry);
+            threads_budget = simplify(max(threads_budget / new_entry.factor, 1));
+        }
+
+        if (!is_already_split) {
+            // TODO(Antony): gpu_threads only when is_compute_at
```
abadams, line 1284: Unresolved TODO?

antonysigma: TODO comment removed. It is resolved.

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
antonysigma: Done. Thank you for pointing it out.

antonysigma (commented):
> Thanks for this! IIRC the original GPU version of this autoscheduler was what we charitably describe as "research code", and was never fit for production.

Hi @abadams , 

Thank you for reviewing the code, and dotting the i's and t's. I concur that GPU scheduling is an experimental feature, and should be highlighted as such in the `user_warning`. Could you please show me where to warn the user?

I am also open to an additional option `bool ArchParams::emit_gpu_schedules = false;`, passable in the generator command line interface. Though, I highly doubt if anyone would go through the hassle of setting `target=host-cuda-cuda_capability_??` just to disable GPU auto-scheduler.

My primary goal is get this PR upstreamed, so that everybody can benefit from the auto-scheduler comparison and other studies. The generated `demo.schedule.h` can be sub-optimal; we all expect the end users will tweak it for their use cases.

abadams:
As this is an attempted reconstruction of his GPU autoscheduler, I should probably tag @ravi-teja-mullapudi to see if this looks sane, because this will affect how people cite and compare to his work in future.

steven-johnson:
Several bot failures with:

```
/home/halidenightly/build_bot/worker/halide-testbranch-main-llvm18-x86-32-linux-make/halide-source/src/autoschedulers/mullapudi2016/AutoSchedule.cpp:2830:21: error: unused variable ‘types’ [-Werror=unused-variable]
```

antonysigma:
> Several bot failures with:
> 
> ```
> /home/halidenightly/build_bot/worker/halide-testbranch-main-llvm18-x86-32-linux-make/halide-source/src/autoschedulers/mullapudi2016/AutoSchedule.cpp:2830:21: error: unused variable ‘types’ [-Werror=unused-variable]
> ```

Done removing the offending line. I also rebased the changes on top of `main`.

antonysigma:
@steven-johnson and @abadams , thank you for testing the PR on the CI. Yes, the failure is triggered by the CMake build option `-DHalide_TARGET=host-[metal|gpu]`. I didn't know we can do that. I like the feature. I will reproduce it on my machine.

The build failure: `Functions that are compute_at() a gpu_block() loop cannot have their own gpu_block() loops` stems from the over-estimated "L2 cache size per thread" machine parameter; the value should have been ~70kB instead of 16MB. It is described in the original paper as a limitation, not a bug.

But yeah, we should have a better exception handling mechanism for this actionable error. I need help to improve the user experience.

steven-johnson:
Updated to main branch to fix OSX WebGPU failures

antonysigma:
Update: The GPU scheduling extension for Mullapudi2016 passes all Buildbot tests except for `autograd_grad.generator` and `local_laplacian_generator`.

1. `autograd_grad` triggers `std::bad_alloc` out-of-memory error on an average consumer-grade computers (i.e. <= 16GB RAM). But it passes the Buildbot tests.

2. `local_laplacian_generator` triggers a subtle `!name.empty()` exception in the Halide IR. 

@abadams Yeah I agreed the Buildbot CI jobs ensure production quality auto-schedulers, which is not original goal of the Mullapudi2016's GPU extensions. I will switch this PR to a draft, and work on issue 2 later next week.

steven-johnson:
This PR is a year old at this point -- is it defunct?

antonysigma:
> This PR is a year old at this point -- is it defunct?

Hi @steven-johnson this branch is still active. I actively use it to generate working schedules for imaging processing pipelines on GPUs like Jetson and RTX cards. 

I also rebase the branch to upstream's top of tree on a monthly basis.

I recalled that we cannot merge it because it doesn't pass a few CI tests. There are test cases on GitHub actions where the following combinations results in autoscheduler exception:

- target set to `host-cuda`
- the halide algorithm under test is related deep learning use cases, e.g. tensor multiplications
- mullapudi2016 autoscheduler is enabled.

Could you show me a way to skip these combos in GitHub actions? Mullapudi2016 is designed for camera ISP use cases, so it makes sense for the GPU reference implementation to report exceptions for deep learning algorithm pipelines.

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -2799,30 +3265,45 @@ void Partitioner::generate_group_cpu_schedule(
...
+    }
+
+    if (t.has_gpu_feature()) {
+        gpu_tiling.apply(sched);
+    }
 
     // Find the level at which group members will be computed.
-    int tile_inner_index = dims.size() - outer_dims.size() - 1;
+    internal_assert(dims.size() >= outer_dims.size());
+    const auto tile_inner_index = std::max(int(dims.size() - outer_dims.size()) - 1, 0);
```
antonysigma, line 3306: For tensor multiplications, the assertion `dims.size() > outer_dims.size()` fails because the GPU schedule wants to map outer dimensions to `gpu_blocks`. If the assertion is ignored, integer `tile_inner_index` becomes invalid (i.e. value less than zero).

Here, I am trying to clamp the `tile_inner_index` to zero. This may break more test case. Help wanted here.

antonysigma:
@steven-johnson Update: state of the CI jobs `halide-testbench` passes almost all tests, except the following:

```
The following tests FAILED:
	  1 - bgu_filter (Subprocess aborted)
	139 - iir_blur_filter (Subprocess aborted)
	433 - nl_means_process (Subprocess aborted)
```

I suspect Mullapudi2016 is not designed for algorithms that are not chained stencil pipelines. Is there a way to skip benchmarks on the CI runners based on CMake option `-DHalide_target=host-cuda` ?

Details of the failure when `Halide_target=host-cuda`:
```
bgu Filer - Error: CUDA error: CUDA_ERROR_ILLEGAL_ADDRESS cuCtxSynchronize failed
iir_blur: Error: Input buffer input is accessed at 5, which is beyond the max (3) in dimension 2
nl_means: Error: CUDA error: CUDA_ERROR_OUT_OF_MEMORY cuMemAlloc failed
```

And when `Halide_target=host-metal`:
```
bgu_filter[39113:94564599] Metal API Validation Enabled
-[MTLDebugComputeCommandEncoder _validateThreadsPerThreadgroup:]:1267: failed assertion `(threadsPerThreadgroup.width(12) * threadsPerThreadgroup.height(21) * threadsPerThreadgroup.depth(4))(1008) must be <= 896. (kernel threadgroup size limit)'
Manually-tuned time: 19.9856ms
Required regular expression not found. Regex=[Success!]
```

antonysigma:
> Several bot failures with:
> 
> ```
> /home/halidenightly/build_bot/worker/halide-testbranch-main-llvm18-x86-32-linux-make/halide-source/src/autoschedulers/mullapudi2016/AutoSchedule.cpp:2830:21: error: unused variable ‘types’ [-Werror=unused-variable]
> ```


> As this is an attempted reconstruction of his GPU autoscheduler, I should probably tag @ravi-teja-mullapudi to see if this looks sane, because this will affect how people cite and compare to his work in future.

Hi @steven-johnson and @abadams ,

Now I think of it, do you recommend that we toggle the Mullapudi2016-GPU feature with an autoscheduler parameter? If so, I can add a parameter `experimental_gpu_schedule=false` in the PR.

My goal is to merge the PR ahead of the Halide v20.0 release.

As of today, the CI failure is primarily caused by test-bench, in which the generated schedules over-estimated the GPU shared memory & register pressures in the GPU hardware. It is a known issue as stated in the Mullapudi2016 and Anderson2021 articles.

Regards,
Antony

mcourteaux:
Wouldn't it be better to simply skip the test if the target has a GPU?
You can check the JIT target features, and print `[SKIP] Explanation`. Performance will still be measured in the CPU version of performance tests.

antonysigma:
Hi @mcourteaux, Finally! Thank you for reviewing the PR so we can move forward!

After almost a year of polishing the PR, I prefer to break up the PR into three separate milestones:

1. (this PR, done) to ignore the combination `Mullapudi2016` + GPU target by default. Buildbot and Testbench will emit cpu-only instructions to pass all tests. (already done, please read the all pass status of the Github actions and the Buildbot testbench reports).

2. (work in progress, https://github.com/antonysigma/Halide/pull/2) to edit the testbench logic at `apps/*/` for selected pipelines. Enable Techbench performance tests for these selected pipelines in the  Mullapudi2016 + GPU target environment.

3. (to be determined) Modify the `mullapudi2016` code to support infinite impulse response (IIR) algorithms. Or, write exception handling logic to make it fail gracefully.

So, just wondering if @abadams can approve of the 3-milestone idea, and then merge this PR first.

> Wouldn't it be better to simply skip the test if the target has a GPU? You can check the JIT target features, and print `[SKIP] Explanation`. Performance will still be measured in the CPU version of performance tests.

I think I already did. Please read the testbench report attached to this PR. It shows all pass.

> Performance will still be measured in the CPU version of performance tests.

Speaking of the test logic at the path `halide/apps/`. There is a way to instruct the `mullaudi2016` autoscheduler to emit GPU code that passes the CI tests on the Buildbot server. The trick is to manually tune `last_level_cache` value for every single pipeline in the `apps/` folder. Like this:

```diff
                   GENERATOR bilateral_grid
                   STMT bilateral_grid_auto_schedule_STMT
                   SCHEDULE bilateral_grid_auto_schedule_SCHEDULE
-                  AUTOSCHEDULER Halide::Mullapudi2016)
+                  AUTOSCHEDULER Halide::Mullapudi2016
+                  # When target=host-cuda or host-metal, limit the GPU shared
+                  # memory per block to avoid gpu kernel launch failure.
+                  PARAMS autoscheduler.last_level_cache_size=20000 autoscheduler.experimental_gpu_schedule=1
                   )

# Main executable
add_executable(bilateral_grid_process filter.cpp)
```

The rest of the commits are at the separate branch, waiting to be submitted to Halide team as a new PR. I am waiting for this PR to merge first.
https://github.com/antonysigma/Halide/pull/2/files#diff-21cb9c0fecc15770841ef2e485fb060036613561a1a981bfbd3e328d74a78b2bR24

> This error is clearly a scheduling issue. I have not read the Mullapudi paper, but this seems like a bug in the implementation? This looks like an .unroll(5), or .vectorize(5), or .split(5) over a dimension with length 3 (RGB, in this case, I believe). Perhaps a missing TailStrategy?
> Edit: is this what you fixed in [ca98dee](https://github.com/halide/Halide/pull/7787/commits/ca98deef62f267b1660599f77d9d60b3b6ef7396)?

Yes, help wanted here to get to the bottom of the issue, but.... only if the code reviewers find it necessary to pass all Testbench tests to approve and merge the PR.

Personally, I favor the idea of skipping GPU-targeted test for `iir_blur` in particular.

From my understanding, the Mullapudi2016's paper clearly stated that his/her work is based on the underlying assumption of stencil-chain only operations, i.e. the pipeline only consists of finite impulse responses (FIRs). So, from first principles, I totally expect `Mullapudi2016` to fail/struggle on infinite impulse response (IIR) blur pipeline *by design*. Isn't the pipeline name in concern called `iir_blur`?

mcourteaux:
> The trick is to manually tune `last_level_cache` value for every single pipeline in the `apps/` folder. Like this:

I wouldn't want to do a deep-dive into the paper, so can you explain why tweaking this value is solving anything? This sounds like it's supposed to be a hardware parameter. Is there perhaps a deeper underlying reason why tweaking this number makes the difference between launch failures and launch successes? Ideally, the autoscheduler _just works_ of course, regardless of that value. The value would only impact the performance of the schedule, but not it's ability to run or crash.



> I prefer to break up the PR into three separate milestones:

I like your proposal. If this is as close as you can get it to the reference implementation of the paper, I see no reason to not go forward with this part, and indeed exclude it from performance tests on the build bots. I would consider a correctness test though. There is a folder `test/autoschedulers/mullapudi2016/`. I think we would all like to see at least one or two tests added there that demonstrate the autoscheduler working on GPU targets.

> > This error is clearly a scheduling issue. I have not read the Mullapudi paper, but this seems like a bug in the implementation? This looks like an .unroll(5), or .vectorize(5), or .split(5) over a dimension with length 3 (RGB, in this case, I believe). Perhaps a missing TailStrategy?
> > Edit: is this what you fixed in [ca98dee](https://github.com/halide/Halide/pull/7787/commits/ca98deef62f267b1660599f77d9d60b3b6ef7396)?
> 
> Yes, help wanted here to get to the bottom of the issue, but.... only if the code reviewers find it necessary to pass all Testbench tests to approve and merge the PR.
> 
> Personally, I favor the idea of skipping GPU-targeted test for `iir_blur` in particular.
> 
> From my understanding, the Mullapudi2016's paper clearly stated that his/her work is based on the underlying assumption of stencil-chain only operations, i.e. the pipeline only consists of finite impulse responses (FIRs). So, from first principles, I totally expect `Mullapudi2016` to fail/struggle on infinite impulse response (IIR) blur pipeline _by design_. Isn't the pipeline name in concern called `iir_blur`?

Well, the error talks about dimension 2, which, looking at the iir_blur app, concerns the color channels dimension. The IIR stuff happens on the x and y dimensions; i.e., dimensions 0 and 1. So this scheduling issue on dimension 2 might be a simple bug after all.

> > Wouldn't it be better to simply skip the test if the target has a GPU? You can check the JIT target features, and print `[SKIP] Explanation`. Performance will still be measured in the CPU version of performance tests.
> 
> I think I already did. Please read the testbench report attached to this PR. It shows all pass.

You are right. I didn't realize this was an app-issue. Apps and autoschedulers are different. The experimental flag seems like a good solution to this experimental scheduling strategy.

test/autoschedulers/mullapudi2016/param.cpp
```diff
@@ -119,21 +93,42 @@ int main(int argc, char **argv) {
...
+
+    const auto run_autoscheduler = [&](Pipeline &&pipeline, const bool debug = false) {
+        // Auto-schedule the pipeline
+        pipeline.apply_autoscheduler(target, autoscheduler_params);
+
+        // Inspect the schedule (only for debugging))
+        if (debug) {
+            pipeline.print_loop_nest();
+        }
+    };
```
antonysigma: >  I would consider a correctness test though. There is a folder test/autoschedulers/mullapudi2016/. I think we would all like to see at least one or two tests added there that demonstrate the autoscheduler working on GPU targets.

@mcourteaux Here you are, the correctness tests for the Mullapudi2016's experimental GPU auto-scheduling feature.

The changes are quite intrusive to the existing test fixture though. This is only one out of 15 tests in the path `test/autoscheduler/mullapudi2016/`.

Expected output:
```shell
inja test/autoschedulers/mullapudi2016/all src/autoschedulers/mullapudi2016/all && ctest -L mullapudi2016_gpu
[4/4] Linking CXX executable test/autoschedulers/mullapudi2016/mullapudi2016_gpu_param
Test project /home/antony/Projects/halide/build-clang18
    Start 35: mullapudi2016_gpu_param
1/1 Test #35: mullapudi2016_gpu_param ..........   Passed    0.78 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
mullapudi2016_gpu    =   0.78 sec*proc (1 test)

Total Test time (real) =   0.81 sec
```

antonysigma (commented):
Hi @mcourteaux I modified one of the correctness test script as per your instructions. Please refer to the inline comments.

> After spending almost two years polishing the PR, I prefer to break up the PR into three separate milestones... https://github.com/halide/Halide/pull/7787#issuecomment-2922895667

> From @mcourteaux . I like your proposal. If this is as close as you can get it to the reference implementation of the paper, I see no reason to not go forward with this part, and indeed exclude it from performance tests on the build bots. 

Hi @abadams ,

What is the exact criteria for getting PR approval and merge for things like an upgrade to autoschedulers? Last time, you and I agreed to suspend the PR because the option `autoscheduler.experimental_gpu_schedule` was hardcoded to `true`. Now I relented and made the option `false` by default. Is it sufficient to be granted an approval?

I also vaguely recalled that the PR is also waiting for the approval from the original author @ravi-teja-mullapudi because the changes make impact his (academic) track record.  Have you received any replies from him since then? https://github.com/halide/Halide/pull/7787#issuecomment-1688846876 

-Antony

test/autoschedulers/mullapudi2016/param.cpp
```diff
@@ -119,21 +93,42 @@ int main(int argc, char **argv) {
...
 
+    // We don't actually care about the content of the 3rd argument. Enable
+    // experimental GPU scheduling whenever a 3rd argument is present.
+    const bool using_experimental_gpu_schedule = (argc >= 3);
+    const Halide::AutoschedulerParams autoscheduler_params{"Mullapudi2016",
+                                                           {{"experimental_gpu_schedule", using_experimental_gpu_schedule ? "1" : "0"}}};
+
+    Target target = get_jit_target_from_environment();
+    if (using_experimental_gpu_schedule) {
+        target.set_feature(Target::CUDA);
```
mcourteaux: Why not `if (target.has_gpu_feature()) { experimental_gpu_schedule = true; }`?

test/autoschedulers/mullapudi2016/param.cpp
mcourteaux: Thanks, this looks mostly good. Except I'm not sure why you don't check if the jit target has a GPU. If so, enable the experimental GPU scheduling feature.

To make the changes less intrusive, the AutoschedulerParams could have just been a global for which you add that boolean value. Then all of the tests stayed the same, except for the 4 `apply_autoscheduler` lines.

test/autoschedulers/mullapudi2016/param.cpp
antonysigma: > Except I'm not sure why you don't check if the jit target has a GPU. If so, enable the experimental GPU scheduling feature.

Sure. Quick question: does our repo's Buildbot covers all the following scenarios in the correctness tests?

| Legend:
| A: `target.has_gpu_features() == true`
| B: `autoscheduler.experimental_gpu_schedule == "1"`

1. `not A && not B`: the autoscheduler should correctly emit CPU-only schedules.
2. `A && not B`: should not emit GPU schedules.
3. `A && B`: should always emit GPU schedules.
4. `not A && B`: should not emit GPU schedules.

test/autoschedulers/mullapudi2016/param.cpp
antonysigma: > To make the changes less intrusive, the AutoschedulerParams could have just been a global for which you add that boolean value. Then all of the tests stayed the same, except for the 4 apply_autoscheduler lines.

Okay. I will use a global boolean variable `using_experimental_gpu_schedule` then. It is old-fashioned but it works.

abadams:
Sorry, I've been slammed with non-Halide things. I have marked it with the dev_meeting tag, so that we can internally sync on what else needs to happen to merge this. The only (extremely superficial) thing I see is that there are some camelCase methods that should be snake_case. @alexreinking may be interested in this autoscheduler and is better qualified than me to give approval.

Ravi has had plenty of time to voice any objections, so I don't think we need to worry about that. Besides, this doesn't change the behavior of the CPU mode, right?

alexreinking:
I'll review

antonysigma:
> Besides, this doesn't change the behavior of the CPU mode, right?

@abadams That is correct.

test/autoschedulers/mullapudi2016/reorder.cpp
```diff
@@ -81,7 +82,15 @@ double run_test_2(bool auto_schedule) {
...
-        p.apply_autoscheduler(target, {"Mullapudi2016"});
+        //
+        // Disabling this experimental GPU feature because the autoscheduler correctly
+        // designates dimension 'z' as the stride axis in the GPU grid-stride loop,
+        // which improves thread retention. However, it fails to reorder 'z' inside
+        // the gpu_blocks 'xo' and 'yo', which is required for proper loop nesting
+        // and successful code generation.
+        //
+        // Reference: https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/
+        p.apply_autoscheduler(target, get_autoscheduler_params(false));
```
antonysigma: I will work on this feature in the next PR.

test/autoschedulers/mullapudi2016/cost_function.cpp
```diff
@@ -48,7 +49,22 @@ int main(int argc, char **argv) {
...
+    // GPU kernel launch failure happens when Likely because (i) most Halide
+    // users can only afford consumer-grade GPUs (e.g. Nvidia GTX1660), and (ii)
+    // the Autoscheduler heristics is over-confident about the actual L2/share
+    // memory usage of the generated GPU kernels.
+    constexpr unsigned last_level_cache_estimate_for_gpu = 25'000;
+    constexpr unsigned using_default_values = 0;
+
+    AutoSchedulerResults results = p.apply_autoscheduler(target,
+                                                         get_autoscheduler_params(target.has_gpu_feature(),
+                                                                                  target.has_gpu_feature() ? last_level_cache_estimate_for_gpu : using_default_values));
```
antonysigma: The need of `last_level_cache_size` manual tuning is required by design. Related to #8637 .

cc'ed @mcourteaux

test/autoschedulers/mullapudi2016/data_dependent.cpp
```diff
@@ -40,13 +41,18 @@ int main(int argc, char **argv) {
...
     // g.print_loop_nest();
 
     // Run the schedule
-    Buffer<uint16_t> out = p.realize({input.width() - 1, input.height()});
+    if (!target.has_gpu_feature()) {
+        // Skipping GPU testing because of the following bug:
+        //
+        // Internal error at halide/src/CodeGen_LLVM.cpp:3024 Condition failed: append_string:
+        Buffer<uint16_t> out = p.realize({input.width() - 1, input.height()});
+    }
```
antonysigma: Not sure what that means. It successfully generates the loop nest, but failed to generate LLVM bitcode for JIT execution. I don't expect 100% test coverage for experimental GPU features. So we are fine.

antonysigma (commented):
Hi @alexreinking , 

Thank you for reviewing the PR. It has accumulated almost two years worth of PR review comments. Let me know which ones to revolve so we can expedite the approval.

Also, per @mcourteaux 's request, I submitted new changes a few minutes ago so as to boost the correctness test coverage to ~90%. 

Please refer to the inline comments on why the experimental GPU feature fails for the remaining 10% test cases. Personally I am not too concerned. We can follow up those edge cases on the next PRs.

-Antony

test/autoschedulers/mullapudi2016/cost_function.cpp
mcourteaux: Typo "heristics". Can you also format the arguments on the next line, to avoid these extremely long lines? (Put target on a new line, and put target.has_gpu_feature() on a new line. That will look nicer I expect.

test/autoschedulers/mullapudi2016/data_dependent.cpp
mcourteaux: Please indicate this clearer using the printing of `[SKIP] Mullapudi for GPU test skipped because of bug`. The cmake tester (ctest) understands `[SKIP]`. You don't run the test, but simply exit with exit code 0 (most of the time simply `return 0;` out of `main()`). You can leave the comment explaining the issue of course. But printing `Success!` is misleading.

See `correctness/gpu_param_allocation.cpp` for an example.

test/autoschedulers/mullapudi2016/reorder.cpp
```diff
@@ -31,7 +32,7 @@ double run_test_1(bool auto_schedule) {
         // Provide estimates on the pipeline output
         r.set_estimates({{0, W}, {0, H}, {0, 3}});
         // Auto-schedule the pipeline
-        p.apply_autoscheduler(target, {"Mullapudi2016"});
+        p.apply_autoscheduler(target, get_autoscheduler_params(target.has_gpu_feature()));
```
mcourteaux: Write a GPU schedule below in the manual scheduling case as well. Otherwise we compare GPU autoschedule against CPU manual schedule.

test/autoschedulers/mullapudi2016/reorder.cpp
mcourteaux: Don't disable the GPU feature here. Simply disable the comparison in the main function  in at line 188 when there is GPU.

test/autoschedulers/mullapudi2016/reorder.cpp
```diff
@@ -121,7 +130,15 @@ double run_test_3(bool auto_schedule) {
...
-        p.apply_autoscheduler(target, {"Mullapudi2016"});
+        //
+        // Disabling this experimental GPU feature because the autoscheduler correctly
+        // identifies dimension 'c' as the stride axis for the GPU grid-stride loop,
+        // which helps retain threads efficiently. However, it fails to reorder 'c'
+        // inside the loop nests of gpu_blocks 'xo' and 'yo', which is necessary for
+        // successful code generation.
+        //
+        // Reference: https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/
+        p.apply_autoscheduler(target, get_autoscheduler_params(false));
```
mcourteaux: Same two comments as above. Add a manual GPU schedule; and disable the test in the main function for now. (Don't print SKIP however, as there is more than 1 test in this file).

mcourteaux (changes requested):

test/autoschedulers/mullapudi2016/param.cpp
mcourteaux: Buildbots will run all tests without GPU, and then again with CUDA GPU, and then again with OpenCL GPU, and then again with Vulkan GPU. Similar things happen on the macos build bots which run the test suite with CPU and then again with GPU Metal.

Your tests define whether experimental_gpu_schedule is 1 or 0. Not the build bots.

mcourteaux:
> Also, per @mcourteaux 's request, I submitted new changes a few minutes ago so as to boost the correctness test coverage to ~90%.
> 
> Please refer to the inline comments on why the experimental GPU feature fails for the remaining 10% test cases. Personally I am not too concerned. We can follow up those edge cases on the next PRs.

Thanks for this update. This is much better than your initial approach. I have a few comments regarding disabling tests, but we're almost there.

test/autoschedulers/mullapudi2016/cost_function.cpp
antonysigma: > Can you also format the arguments on the next line, to avoid these extremely long lines? 

@alexreinking Manual code formatting requires manual override with `// clang-format off`. Is it frown upon by reviewers?

mcourteaux: I believe you can simply add a newline after an opening `(`, and clang-format will accept that and move stuff to the left.

test/autoschedulers/mullapudi2016/reorder.cpp
antonysigma: Re: benchmarking. I prefer to bundle benchmark-related changes with the `apps/` changes in the next PR. I am more comfortable with small scale and frequent PR submission and approval cycles.

For your information, I already have working changes at `apps/` similar to the one below, waiting to be submitted for a formal code review. I am simply waiting for this PR to finish to submit a new one.

```diff
                   GENERATOR bilateral_grid
                   STMT bilateral_grid_auto_schedule_STMT
                   SCHEDULE bilateral_grid_auto_schedule_SCHEDULE
-                  AUTOSCHEDULER Halide::Mullapudi2016)
+                  AUTOSCHEDULER Halide::Mullapudi2016
+                  # When target=host-cuda or host-metal, limit the GPU shared
+                  # memory per block to avoid gpu kernel launch failure.
+                  PARAMS autoscheduler.last_level_cache_size=20000 autoscheduler.experimental_gpu_schedule=1
                   )
```

Re: concerns about mixing correctness and  benchmarking test logic.

These files at path `test/autoscheduler/mullapudi2016/` looks ancient to me... It may have written before we have the TestBench on the Buildbot and the benchmarking stuff at path `apps/`. If you insist on writing benchmarking logic inter-mixed with correctness tests, I can certainly do that but preferably in the next PR. My concern is about losing momentum (again) because of the ever changing PR approval criteria.

test/autoschedulers/mullapudi2016/reorder.cpp
antonysigma: Re: skip test but don't print `[SKIP]` to console. Done.

Re: concerns about intermixing benchmarking logic and correctness test logic. Please read my replies above.

test/autoschedulers/mullapudi2016/reorder.cpp
```diff
@@ -121,7 +140,15 @@ double run_test_3(bool auto_schedule) {
...
-        p.apply_autoscheduler(target, {"Mullapudi2016"});
+        //
+        // Disabling this experimental GPU feature because the autoscheduler correctly
+        // identifies reduction domain 'r.x' as the stride axis for the GPU grid-stride loop,
+        // which helps retain threads efficiently. However, it fails to reorder 'r.x'
+        // inside the loop nests of gpu_blocks 'xo' and 'yo', which is necessary for
+        // successful code generation.
+        //
+        // Reference: https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/
+        p.apply_autoscheduler(target, get_autoscheduler_params(target.has_gpu_feature()));
```
mcourteaux: Perhaps update the comment to reflect the test is disabled below in main().

test/autoschedulers/mullapudi2016/reorder.cpp
```diff
@@ -184,6 +211,11 @@ int main(int argc, char **argv) {
     }
 
     {
+        if (get_jit_target_from_environment().has_gpu_feature()) {
+            std::cout << "Mullapudi for GPU test for Test Case 3 skipped because of reordering bug.\n";
+            return 0;
+        }
+
```
mcourteaux: Move this just out of the `{ }` block, remove the `return`, and use ` else ` to then enter the actual test. Code flow wise, it feels much cleaner; disabling one test shouldn't depend on the fact that it's the last one.

test/autoschedulers/mullapudi2016/reorder.cpp
mcourteaux: It's not really a matter of performance benchmarking. The tests simply check if the autoscheduler doesn't do something wildly wrong, by allowing a 2x performance hit comapred to the manual schedule. If it's slower than half the speed, it's considered a correctness issue.

If you already have schedules available for a follow-up PR, which you feel are out-of-scope for this PR, I guess we can solve this next PR.

test/autoschedulers/mullapudi2016/param.cpp
```diff
@@ -119,7 +120,7 @@ int main(int argc, char **argv) {
         return 0;
     }
 
-    if (argc != 2) {
+    if (argc < 2) {
```
mcourteaux: Why did you change this? argv[0] is always the program name. So argc is always >= 1.

antonysigma (commented):
Re: Last mile changes.

@mcourteaux  Done. I appreciate if you review the latest changes again. 

Re: intermixing correctness test cases and performance regression tests in the `test/autoscheduler/` folder.

> It's not really a matter of performance benchmarking. The tests simply check if the autoscheduler doesn't do something wildly wrong, by allowing a 2x performance hit comapred to the manual schedule. If it's slower than half the speed, it's considered a correctness issue.

> If you already have schedules available for a follow-up PR, which you feel are out-of-scope for this PR, I guess we can solve this next PR.

I see. So Halide developers treats performance regression tests as part of correctness tests. Thanks for the reminder.

Yes, I do prefer to write regression tests (with manual GPU schedules) in the next PR. And, stick to the following road map to ensure small-scale and frequent PRs and merges:

1. (this PR) Experimental GPU schedule support, with correctness tests;
2. Enable experimental GPU features for the pipelines in `apps/*/`; manual tuning of the `last_level_cache_size` estimates for complex pipelines.
3. Write manual GPU schedules for the test cases at `test/autoscheduler/mullapudi2016/*.cpp`;
4. Resolve the `reorder` bug having "reduction axes sandwiched between `gpu_blocks`;
5. (Optional) Resolve the `apps/iir_blur/` error having vectorized width larger than the channel count (=3).
6. (Optional) Resolve other unknown issues not anticipated by the original Mullapudi paper, e.g. `CodeGen_NVPTX` does not support more than 4 `gpu_blocks`; needing `fuse()` to fit the 4 gpu clock limit imposed by the hardware.

By small scale PRs, I mean those having less than ~1,000 words of accumulated code review comments and conversations. It avoid losing momentum whenever new reviewers join the chat.

-Antony

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
mcourteaux: This perhaps should be a comment in the code for future reference:

`// note(antonysigma): <same text here>`

mcourteaux (approved):
Testing: LGTM.
AutoScheduler: Not reviewed in depth, but code changes look fine, style wise.

test/autoschedulers/mullapudi2016/get_autoscheduler_params.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <Halide.h>
+#include <optional>
```
alexreinking: Should add headers for `get_autoscheduler_params` implementation:
```suggestion
#include <Halide.h>

#include <map>
#include <optional>
#include <string>
```

test/autoschedulers/mullapudi2016/get_autoscheduler_params.hpp
```diff
@@ -0,0 +1,27 @@
...
+    if (using_experimental_gpu_schedule && gpu_params.has_value()) {
+        params["last_level_cache_size"] = std::to_string(gpu_params.value().last_level_cache_size);
+        params["parallelism"] = std::to_string(gpu_params.value().parallelism);
+    }
+    return {
+        "Mullapudi2016",  //
+        std::move(params)};
+}
+
+}  // namespace
```
alexreinking: Please use trailing newlines in files.

test/autoschedulers/mullapudi2016/fibonacci.cpp
```diff
@@ -22,11 +23,11 @@ double run_test(bool auto_schedule) {
 
     if (auto_schedule) {
         // Auto-schedule the pipeline
-        p.apply_autoscheduler(target, {"Mullapudi2016"});
+        p.apply_autoscheduler(target, get_autoscheduler_params(target.has_gpu_feature()));
     }
 
     // Inspect the schedule (only for debugging))
-    // g.print_loop_nest();
+    g.print_loop_nest();
```
alexreinking: What about this?

```suggestion
    debug(1) << [&] {
        g.print_loop_nest();
        return "";
    }();
```

Otherwise, drop the print.

test/autoschedulers/mullapudi2016/get_autoscheduler_params.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <Halide.h>
+#include <optional>
+
+namespace {
+
+struct Mullapudi2016Params {
+    unsigned last_level_cache_size{};
+    unsigned parallelism{};
+};
```
alexreinking: I think the naming in this header is misleading. This struct is named very much like the canonical structs for the other autoschedulers (e.g. `Anderson2021Params`) but is only visible to tests. This makes it hard to grep the codebase and to read the test code.

This name should be changed to include the word `Test`.

test/autoschedulers/mullapudi2016/get_autoscheduler_params.hpp
```diff
@@ -0,0 +1,27 @@
...
+        {"experimental_gpu_schedule", using_experimental_gpu_schedule ? "1" : "0"}};
+
+    if (using_experimental_gpu_schedule && gpu_params.has_value()) {
+        params["last_level_cache_size"] = std::to_string(gpu_params.value().last_level_cache_size);
+        params["parallelism"] = std::to_string(gpu_params.value().parallelism);
+    }
+    return {
+        "Mullapudi2016",  //
+        std::move(params)};
+}
```
alexreinking: Same naming issue as above -- this is really only for testing. I'd rather see it named `get_mullapudi2016_test_params` or the like. While we're changing it, `using_experimental_gpu_schedule` is always equal to `target.has_gpu_feature()`. I think this function should just take the `target` to keep the call sites clean.

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -23,15 +23,31 @@ using std::vector;
...
+     * 1660/Turing), or low-cost scientific-grade GPUs (Nvidia K40/Tesla).
+     *
+     * Section 5.4 of the Mullapudi2016 article: We configure the auto-scheduler
+     * to target the GPU by setting the PARALLELISM_THRESHOLD to 128, ..., and
+     * CACHE_SIZE to 48 KB.
+     */
+    constexpr ArchParams(bool has_gpu_feature)
+        : parallelism(has_gpu_feature ? 128 : 16), last_level_cache_size(has_gpu_feature ? 48 * 1024 : 16 * 1024 * 1024),
+          balance(has_gpu_feature ? 20 : 40) {
+    }
```
alexreinking: I'm a little concerned about the connection between `has_gpu_feature` and `experimental_gpu_schedule` as it is used throughout. In this instance, `ArchParams(true)` can be called even while `experimental_gpu_schedule` is false. I don't see why that should be valid. Elsewhere, both `t.has_gpu_feature()` and `experimental_gpu_schedule` are redundantly checked.

I don't think this is the appropriate place to keep default values now that there are multiple regimes (CPU/GPU).

I would like parsing and configuration to be further decoupled.

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -887,6 +903,419 @@ struct AutoSchedule {
...
+        Expr factor;
+        TailStrategy strategy;
+    };
+
+    GPUTileHelper(Stage &_f, uint32_t n)
+        : f(_f), stage_num(n) {
+    }
+
+    /** Indicate the need to split the dimensions with `gpu_tile()` method. */
+    void applySplit(const split_t &x) {
```
alexreinking: We prefer `snake_case` to `camelCase` in our codebase.

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -887,6 +903,419 @@ struct AutoSchedule {
...
+        return "";
+    }
+}
+
+/** Apply gpu_threads and gpu_blocks as an atomic transaction. */
+class GPUTileHelper {
+public:
+    /** A data structure documenting the split factor and tail strategy. */
+    struct split_t {
+        VarOrRVar v;
```
alexreinking: Types ending in `_t` are reserved by POSIX. A name like `split_info` is legal and fine.

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -887,6 +903,419 @@ struct AutoSchedule {
...
+        debug(2) << f.name() << ".split(" << v.name() << "," << factor << ")\n";
+        is_split.emplace(v.name());
+        outer_vars.emplace(vo.name());
+        inner_vars.emplace(vi.name());
+
+        parallelize.try_emplace(v.name(), split_t{v, vo, vi, factor, strategy});
+    }
+
+    /** Indicate the default dimension order of the Func. */
+    void setInitialOrder(const Func &func) {
```
alexreinking: More `camelCase`

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -887,6 +903,419 @@ struct AutoSchedule {
...
+
+    /** Indicate to desire to reorder the dimensions.
+     *
+     * Func::reorder() is called by the auto-parallelization algorithm multiple
+     * times, as it seeks to over-subscribe the available CPU cores
+     * aggressively. For GPU schedule, we only need the very last reorder() call
+     * to map the tile orders. Here, we always cache the current dimension
+     * order, and override the previous ones.
+     */
+    void canReorder(const std::vector<VarOrRVar> &vars) {
```
alexreinking, line 1259: More `camelCase`

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -887,6 +903,419 @@ struct AutoSchedule {
...
+            // tile size, we simply mark the inner dims as gpu_threads();
+            // similarly, outer dims as gpu_blocks().
+            mark_gpu_threads(sched);
+            return;
+        }
+
+        GPUTileHelper helper{f, stage_num};
+        Expr threads_budget = max_n_threads;
+
+        // Traverse the dimensions, ordered by the variable names (x, y, z) in lexilogical order.
```
alexreinking:
```suggestion
        // Traverse the dimensions, ordered by the variable names (x, y, z) in lexicographic order.
```

src/autoschedulers/mullapudi2016/AutoSchedule.cpp
```diff
@@ -2392,25 +2824,18 @@ pair<VarOrRVar, VarOrRVar> Partitioner::split_dim(
         strategy = TailStrategy::GuardWithIf;
     }
 
-    f_handle.split(v, outer, inner, factor, strategy);
+    if (t.has_gpu_feature() && gpu_tiling) {
```
alexreinking: Should we be checking `arch_params.experimental_gpu_schedule` here? The inconsistency here is troubling.

alexreinking (changes requested):
I have some style nitpicks, but my biggest concern is with the "experimental" gating -- I would like to see simpler checks at the key algorithmic points that boil down to "are we scheduling for GPU or not?" without redundantly (and inconsistently!) checking that the experimental flag is enabled.

I also think fixing the naming issues in the test code is important.

alexreinking:
@antonysigma - I really want to get this merged, so I made the changes I requested myself. Please take a look and review my changes. If we're both happy, we can land this.

alexreinking (approved):
Approved pending :+1: from @antonysigma

antonysigma:
> I really want to get this merged, so I made the changes I requested myself. Please take a look and review my changes. If we're both happy, we can land this.

@alexreinking Thank you very much for initiating the code style changes. The buildbot failed after the changes. How do I assist you to pass all tests?

I also noticed Halide PR doesn't enforce linear git history by "squash and rebase merge" per PR. It doesn't impact me as much, but good to know about the PR policy here. Thanks.

alexreinking:
I think the failures are due to some new nightly LLVM 20 regression, not to my changes. I'll investigate further.

We do enforce a linear history by squashing.

abadams:
I'm investigating the llvm regression

alexreinking:
Merged in main with LLVM fixes. Can merge this when green

alexreinking:
One flaky performance test. Merging!

PR #7790, opened 2023-08-21
[ARM] Distribute shifts as muls

rootjalex:
ARM/LLVM currently lowers `a + widening_shl(b, c)` into `(u | s)shll` and `add`, this PR lowers it to `(u | s)mlal`. I reused HVX machinery for the distribution, `DistributeShiftsAsMuls`. HVX lowering is not expected to change at all.


steven-johnson (commented):
Nice to have it factored into common code; I suspect that RISC-V will (eventually) benefit from something similar

src/DistributeShifts.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "IRMatch.h"
+#include "IRMutator.h"
+#include "Simplify.h"
+
+namespace Halide {
+namespace Internal {
+
+// Distribute constant RHS widening shift lefts as multiplies.
+// TODO: This is an extremely unfortunate mess. I think the better
+// solution is for the simplifier to distribute constant multiplications
```
abadams, line 14: distributing constant multiplications goes in the wrong direction in terms of the simplifier's reduction order. It adds more ops! I think doing this transformation is very architecture and data-type specific, and we should do it as late as possible.

src/DistributeShifts.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+#include "IR.h"
+
+namespace Halide {
+namespace Internal {
+
+// Distributes shifts as multiplies. If `polynomials_only` is set,
+// then only distributes the patterns `a + widening_shl(b, c)` /
+// `a - widening_shl(b, c)` and `a + b << c` / `a - b << c`.
```
abadams, line 15: Comment should show what the 'after' IR is, because it's not clear what 'distributes' means here.

src/DistributeShifts.cpp
rootjalex: This comment is copied over directly from HexagonOptimize!

src/DistributeShifts.h
```diff
@@ -0,0 +1,23 @@
...
+ * A tool to distribute shifts as multiplies, useful for some backends. (e.g. ARM, HVX).
+ */
+
+#include "IR.h"
+
+namespace Halide {
+namespace Internal {
+
+// Distributes shifts as multiplies. If `polynomials_only` is set,
+// then only distributes the patterns `a + widening_shl(b, c)` /
```
abadams, line 14: I'm a bit confused by the term polynomial here too. I don't see where the polynomial is. Is there a better word?

rootjalex: Perhaps. I’m trying to say that a + b << c is lowered to a + (b * (1 << c)) but just b << c will not be (on ARM)

rootjalex: Do you know of a better word? “affine” perhaps?

abadams: multiply_add?

src/DistributeShifts.cpp
```diff
@@ -20,12 +20,12 @@ namespace Internal {
 // widening_mul_add operands.
 class DistributeShiftsAsMuls : public IRMutator {
 public:
-    DistributeShiftsAsMuls(const bool polys_only)
-        : polynomials_only(polys_only) {
+    DistributeShiftsAsMuls(const bool _multiply_adds)
```
abadams, line 23: It's legal for the arg to have the same name as the member

src/DistributeShifts.cpp
abadams: Please update it anyway. It's not a great comment, and you being the last person to touch it changes the meaning of "I"

abadams:
There's a nifty clang tidy failure

src/DistributeShifts.h
```diff
@@ -0,0 +1,23 @@
...
+/** \file
+ * A tool to distribute shifts as multiplies, useful for some backends. (e.g. ARM, HVX).
+ */
+
+#include "IR.h"
+
+namespace Halide {
+namespace Internal {
+
+// Distributes shifts as multiplies. If `multiply_adds` is set,
```
abadams, line 13: Use a doxygen-style comment here.

abadams:
lgtm pending green bots

rootjalex:
The `find_intrinsics` rule that helps with mixed-sign `widening_shift_left` appears to have broken `simd_op_check_hvx`, so I am reverting it. This was to handle cases like:
```
i16_1 + i16(u8_2) * 2 -> umlal
```
@pranavb-ca I'd be happy to hear if you have any ideas about fixing this. I think the particular issue was that `2 * (i16(u8_1) + i16(u8_2))` got mapped to 2 `vmpy`s with the change, where it used to be 1 `vmpa`.

rootjalex:
Failures appear unrelated

Issue #7786, opened 2023-08-21
Halide schedule have undef before code generation which should have been eliminated 


TH3CHARLie:
Schedule that produces this internal error:
```cpp
    Func input("input");
    Func blur_x("blur_x");
    Func blur_y("blur_y");
    Var x("x"), y("y"), xo("xo"), xi("xi"), xio("xio"), xii("xii"), xiiyif("xiiyif"), yo("yo"), yi("yi");
    input(x, y) = x + y;
    blur_x(x, y) = (input(x - 1, y) + input(x, y) + input(x + 1, y)) / 3;
    blur_y(x, y) = (blur_x(x, y - 1) + blur_x(x, y) + blur_x(x, y + 1)) / 3;
    blur_y.split(x, xo, xi, 2, TailStrategy::Predicate).split(xi, xio, xii, 4, TailStrategy::Auto).vectorize(xii).split(y, yo, yi, 8, TailStrategy::PredicateStores).fuse(xii, yi, xiiyif);
    blur_x.split(y, yo, yi, 1, TailStrategy::RoundUp).split(x, xo, xi, 1, TailStrategy::RoundUp).compute_root();
    Pipeline p({blur_y});
    Buffer<int> buf = p.realize({width, height});
```
Compile this generates:
```
Error: undef not eliminated before code generation. Please report this as a Halide bug.
```

abadams:
Fixed by #7791

PR #7791, opened 2023-08-21
Don't inject undef() in the simplifier

abadams:
We shouldn't be using undef() in the simplifier. This replaces a load with a constant false predicate with a zero instead.

I also added a guard around some dubious logic about out of bounds loads. out of bounds loads may be reachable if they have a false predicate, so I changed this simplification to only trigger if the load is unpredicated.


TH3CHARLie:
This fixes #7786

steven-johnson (approved):

PR #7762, opened 2023-08-15
Add support to the makefile for serialization

abadams:

PR #7792, opened 2023-08-21
[serialization] Add serialization support to generator interface 

derek-gerstmann:
Added new 'hlpipe' output type.
Added new `emit_hlpipe(...)` method to generator class interface.
Added new `pre_serialization()/post_serializtion()` methods to set a flag to bypass target validation during build_pipeline()

Note that generators that currently use `natural_vector_size(...)` must have a target specified, otherwise they will emit a user_error during serialization.  Since almost all our generators seem to rely upon `natural_vector_size(...)` it seems like we might need to allow some way to inject a placeholder into the serialization format to allow the target to be unspecified, and the vector size to be resolved when the pipeline is realized.


src/Generator.cpp
```diff
@@ -996,10 +998,19 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
     const bool cpp_stub_only = args.output_types.size() == 1 &&
                                args.output_types.count(OutputFileType::cpp_stub) == 1;
-    if (!cpp_stub_only) {
-        // It's ok to leave targets unspecified if we are generating *only* a cpp_stub
+
+    const bool hlpipe_and_cpp_stub_only = args.output_types.size() == 2 &&
+                                          (args.output_types.count(OutputFileType::cpp_stub) == 1) &&
+                                          (args.output_types.count(OutputFileType::hlpipe) == 1);
+
+    const bool target_is_optional = hlpipe_only || cpp_stub_only || hlpipe_and_cpp_stub_only;
```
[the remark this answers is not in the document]
steven-johnson: Yeah, common and critical

src/AbstractGenerator.h
```diff
@@ -177,6 +177,21 @@ class AbstractGenerator {
      */
     virtual bool emit_cpp_stub(const std::string &stub_file_path) = 0;
 
+    /** Emit a Serialized Halide Pipeline (.hlpipe) file to the given path. Not all Generators support this.
+     *
+     * If you call this method, you should not call any other AbstractGenerator methods
+     * on this instance, before or after this call.
+     *
+     * If the Generator is capable of emitting a hlpipe, do so and return true. (Errors
```
steven-johnson, line 185: "an hlpipe" instead of "a hlpipe"? (How are you pronouncing it?)

src/AbstractGenerator.h
```diff
@@ -177,6 +177,21 @@ class AbstractGenerator {
...
+     * during stub emission should assert-fail rather than returning false.)
```
steven-johnson, line 186: s/stub emission/hlpipe emission/

src/Generator.cpp
```diff
@@ -996,10 +998,19 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
 
+    const bool hlpipe_only = args.output_types.size() == 1 &&
+                             args.output_types.count(OutputFileType::hlpipe) == 1;
+
     const bool cpp_stub_only = args.output_types.size() == 1 &&
                                args.output_types.count(OutputFileType::cpp_stub) == 1;
-    if (!cpp_stub_only) {
-        // It's ok to leave targets unspecified if we are generating *only* a cpp_stub
+
+    const bool hlpipe_and_cpp_stub_only = args.output_types.size() == 2 &&
```
steven-johnson, line 1007: Is there a real-world use case for emitting these two and nothing else? I can't think of one. Unless we know of one, I don't know why we should special case it. (Especially since hlpipe requires a Target, unlike cppstub, so they don't really go together)

src/Generator.cpp
```diff
@@ -996,10 +998,19 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
+                                          (args.output_types.count(OutputFileType::cpp_stub) == 1) &&
+                                          (args.output_types.count(OutputFileType::hlpipe) == 1);
+
+    const bool target_is_optional = hlpipe_only || cpp_stub_only || hlpipe_and_cpp_stub_only;
+    if (!target_is_optional) {
+        // It's ok to leave targets unspecified if we are generating *only* a cpp_stub and/or a hlpipe
```
steven-johnson, line 1013: remove "and/or a hlpipe"

src/Generator.cpp
```diff
@@ -1047,17 +1058,23 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
     if (!args.generator_name.empty()) {
         const std::string base_path = args.output_dir + "/" + args.file_base_name;
         debug(1) << "Generator " << args.generator_name << " has base_path " << base_path << "\n";
-        if (args.output_types.count(OutputFileType::cpp_stub)) {
-            // When generating cpp_stub, we ignore all generator args passed in, and supply a fake Target.
+
+        if (args.output_types.count(OutputFileType::cpp_stub) || args.output_types.count(OutputFileType::hlpipe)) {
```
steven-johnson, line 1062: remove the hlpipe case here

src/Generator.cpp
```diff
@@ -1047,17 +1058,23 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
+                gen->emit_cpp_stub(output_files[OutputFileType::cpp_stub]);
+            }
+            if (args.output_types.count(OutputFileType::hlpipe)) {
+                gen->emit_hlpipe(output_files[OutputFileType::hlpipe]);
+            }
         }
 
-        // Don't bother with this if we're just emitting a cpp_stub.
-        if (!cpp_stub_only) {
+        // Don't bother with this if we're just emitting a cpp_stub or hlpipe
```
steven-johnson, line 1076: fix comment

src/Generator.cpp
```diff
@@ -1402,7 +1419,7 @@ void GeneratorBase::pre_generate() {
     advance_phase(GenerateCalled);
     GeneratorParamInfo &pi = param_info();
     user_assert(!pi.outputs().empty()) << "Must use Output<> with generate() method.";
-    user_assert(get_target() != Target()) << "The Generator target has not been set.";
+    user_assert(get_target() != Target() || is_serializable()) << "The Generator target has not been set.";
```
steven-johnson, line 1422: per above comments, pretty sure we want to remove the inserted `|| is_serializable()` here

src/Generator.h
```diff
@@ -3478,6 +3484,9 @@ class GeneratorBase : public NamesInterface, public AbstractGenerator {
 
     const size_t size;
 
+    // Flag to enable serialization (and bypass asserts for unbound target)
```
steven-johnson, line 3487: per previous comments, this seems no longer necessary or desirable

src/Module.cpp
```diff
@@ -49,6 +49,7 @@ std::map<OutputFileType, const OutputInfo> get_output_info(const Target &target)
         {OutputFileType::static_library, {"static_library", is_windows_coff ? ".lib" : ".a", IsSingle}},
         {OutputFileType::stmt, {"stmt", ".stmt", IsMulti}},
         {OutputFileType::stmt_html, {"stmt_html", ".stmt.html", IsMulti}},
+        {OutputFileType::hlpipe, {"hlpipe", ".hlpipe", IsSingle}},
```
steven-johnson, line 52: Please insert in alphabetical order, as is the case with the rest of this

src/Module.h
```diff
@@ -41,6 +41,7 @@ enum class OutputFileType {
     static_library,
     stmt,
     stmt_html,
+    hlpipe
```
steven-johnson, line 44: alpha order, please

steven-johnson (changes requested):
Since schedules depend pretty critically on the Target, we definitely want to require a target to be set. 

That said, perhaps we could/should add an option to the string-to-target conversion that will produce the equivalent of `Target()` -- i.e., shorthand for `arch_unknown-os_unknown-0`, which is a "Target" that can't be used usefully for codegen, but perhaps could be useful for some serialization purposes?

src/Generator.cpp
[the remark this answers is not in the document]
derek-gerstmann: That's very true.  I'll change this so that the target is required. Thx!

derek-gerstmann:
> Since schedules depend pretty critically on the Target, we definitely want to require a target to be set.
> 
> That said, perhaps we could/should add an option to the string-to-target conversion that will produce the equivalent of `Target()` -- i.e., shorthand for `arch_unknown-os_unknown-0`, which is a "Target" that can't be used usefully for codegen, but perhaps could be useful for some serialization purposes?

Doing a quick pass through our generator tests, only a few can actually be serialized *without* specifying a target, so I'm not sure how useful it would be.

src/Generator.cpp
derek-gerstmann: Fixed!

src/Generator.h
derek-gerstmann: Fixed!

src/Generator.cpp
derek-gerstmann: Yes, I've separated these, and now handle the hlpipe separately within a target specific context.

src/AbstractGenerator.h
derek-gerstmann: Sure ... I've been calling it an "HL" Pipe, so "an" seems more appropriate here.

src/Generator.cpp
```diff
@@ -1047,27 +1050,49 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
+
         if (args.output_types.count(OutputFileType::cpp_stub)) {
-            // When generating cpp_stub, we ignore all generator args passed in, and supply a fake Target.
+            // When generating cpp_stub we ignore all generator args passed in, and supply a fake Target.
             // (CompilerLogger is never enabled for cpp_stub, for now anyway.)
             const Target fake_target = Target();
             auto gen = args.create_generator(args.generator_name, GeneratorContext(fake_target));
             auto output_files = compute_output_files(fake_target, base_path, args.output_types);
-            gen->emit_cpp_stub(output_files[OutputFileType::cpp_stub]);
+            if (args.output_types.count(OutputFileType::cpp_stub)) {
```
steven-johnson, line 1073: Isn't this redundant to the check on line 1067?

src/Generator.cpp
```diff
@@ -1047,27 +1050,49 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
+                const Target &target = args.targets[i];
+                const std::string hlpipe_path = use_target_suffix ? (base_path + "-" + target.to_string()) : base_path;
+                auto output_files = compute_output_files(target, hlpipe_path, args.output_types);
+                auto gen = generator_factory(args.function_name, target);
+                gen->emit_hlpipe(output_files[OutputFileType::hlpipe]);
+            }
         }
 
-        // Don't bother with this if we're just emitting a cpp_stub.
+        // Don't bother with this if we're just emitting a cpp_stub
```
steven-johnson, line 1091: Why did you remove the period? This is a complete sentence, so it should have proper punctuation.

src/Generator.cpp
```diff
@@ -1047,27 +1050,49 @@ void execute_generator(const ExecuteGeneratorArgs &args_in) {
...
             auto gen = args.create_generator(args.generator_name, GeneratorContext(fake_target));
             auto output_files = compute_output_files(fake_target, base_path, args.output_types);
-            gen->emit_cpp_stub(output_files[OutputFileType::cpp_stub]);
+            if (args.output_types.count(OutputFileType::cpp_stub)) {
+                gen->emit_cpp_stub(output_files[OutputFileType::cpp_stub]);
+            }
+        }
+
+        if (args.output_types.count(OutputFileType::hlpipe)) {
+            // When serializing a halide pipeline, target is required (since the schedule may be target dependent)
```
steven-johnson, line 1079: These are complete sentences, so they really should end with proper punctuation.

src/Generator.h
```diff
@@ -3478,6 +3478,9 @@ class GeneratorBase : public NamesInterface, public AbstractGenerator {
...
+    bool serializable = false;
```
steven-johnson, line 3482: Shouldn't this be removed now?

src/Generator.cpp
derek-gerstmann: Yeah, this is redundant now.

src/Generator.cpp
derek-gerstmann: Okay, will fix!

src/Generator.h
derek-gerstmann: Yes, this is leftover.  I'll remove this.

src/Generator.cpp
derek-gerstmann: Will fix.

steven-johnson:
Looks like the Make builds are failing to link...? Maybe https://github.com/halide/Halide/pull/7762?

abadams:
Some stuff probably needs to be guarded with #ifdef WITH_SERIALIZATION

steven-johnson (approved):
LGTM -- bot failures are unrelated, ready to land unless @abadams objects

PR #7802, opened 2023-08-24
Unbreak WebGPU builds

steven-johnson:
It seems that newer versions of emcc and/or dawn no longer have the `wgpuDeviceSetDeviceLostCallback()` fn available. This just comments these out temporarily, to unbreak the builds; it's not clear if these are no longer necessary or if there is an alternate API to accomplish this, but while I'm investigating, this is better than nothing.

attn: @jrprice

### DIFF
--- a/src/runtime/webgpu.cpp
+++ b/src/runtime/webgpu.cpp
@@ -242,12 +242,13 @@ namespace {
 
 halide_error_code_t init_error_code = halide_error_code_success;
 
-void device_lost_callback(WGPUDeviceLostReason reason,
-                          char const *message,
-                          void *user_context) {
-    error(user_context) << "WGPU device lost (" << reason << "): "
-                        << message << "\n";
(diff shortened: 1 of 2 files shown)


steven-johnson:
It appears that Emscripten removed this from their `webgpu_cpp` library in https://github.com/emscripten-core/emscripten/commit/412fd754da4d0b2885c15a2acd47f9629e281066 on Jun 23 2023, but:
- It's not clear to me why, the comments aren't helpful
- the most-recent previous change to that source file was https://github.com/emscripten-core/emscripten/commit/ca3e10c7933ac45ec041a96be0fea28145e9aa06 **almost a year earlier** (Aug 2022) so there's a lotta Dawn changes in between those

steven-johnson:
This is not quite good enough, but I have a proper and complete fix on the way shortly.

Issue #7312, opened 2023-01-31
Support WASM Relaxed SIMD instructions


shoaibkamil:
Halide should support the WASM [Relaxed SIMD instructions](https://github.com/WebAssembly/relaxed-simd).  The current proposal is nearing final acceptance, and is implemented in the emscripten/LLVM toolchain, so should be testable.  Speedups for certain use cases (neural network inference, for example) show speedups of 30-40% on desktop CPUs and 2-3x on ARM phones.

Design-wise, there's a couple options on how we should support these instructions.  Given a target with explicit support for relaxed SIMD, we could:
1) add user-callable intrinsics for the relaxed instructions; or
2) automatically rewrite sequences into relaxed instructions when appropriate

(or both).  Implementing automatic rewrites may be possible since we can, in many cases, infer ranges on values and thus prove that no non-determinism occurs.  An initial prototype, however, should probably just add the intrinsics.

steven-johnson:
I haven't looked at the spec for Relaxed SIMD closely enough to get a feel for whether it fits into the "fast-math" mold (ie, could we just assume it's OK to use these instructions when strict-float isn't in use)... I defer to your judgement here. That said, it makes me nervous / unhappy if they have added instructions that don't reasonably fit into any existing mold.

As a short-term stopgap, adding explicit intrinsics for these seems like a pragmatic expedient, but in the long run I think we'd prefer to avoid doing that, as it seems like a road to writing Halide code that starts to diverge wildly (in the non-schedule section) over time.

(FYI, I realize I've done most of the work on the wasm backend to date, but at present I don't know if my work priorities include making this happen in a timely manner; do you have any resources in mind for implementing this?)

shoaibkamil:
> I haven't looked at the spec for Relaxed SIMD closely enough to get a feel for whether it fits into the "fast-math" mold (ie, could we just assume it's OK to use these instructions when strict-float isn't in use)... I defer to your judgement here. That said, it makes me nervous / unhappy if they have added instructions that don't reasonably fit into any existing mold.

The proposal has 3 categories of instructions:
> 1.  Integer instructions where the inputs are interpreted differently (e.g. swizzle, 4-D dot-product)
> 2. Floating-point instructions whose behavior for out-of-range and NaNs differ (e.g. float-to-int conversions, float min/max)
> 3. Floating-point instructions where the precision or order of operations differ (e.g. FMA, reciprocal instructions, sum reduction)

[...]

PR #7805, opened 2023-08-24
[x86 & HVX & WASM] Use bounds inference for saturating_narrow instruction selection

rootjalex:
There are cases where bounds inference allows us to safely reinterpret unsigned values as signed values, in order to target x86 saturating narrow instructions. I give an example below. I don't add tests to simd_op_check_x86 because we can't test for removing `min` instructions via the existing mechanism.

Example:
```
Var x("x");
ImageParam in0(UInt(8), 1);
Func out("out");
out(x) = u8_sat(u16(in0(x)) * 5);
out.vectorize(x, 32);
Target x86("x86-64-linux-avx-avx2-fma-sse41");
out.compile_to_assembly("x86-test0.asm", out.infer_arguments(), x86);
```

Before:
```
vpmovzxbw	(%rax,%rcx), %ymm2      # ymm2 = mem[0],zero,mem[1],zero,mem[2],zero,mem[3],zero,mem[4],zero,mem[5],zero,mem[6],zero,mem[7],zero,mem[8],zero,mem[9],zero,mem[10],zero,mem[11],zero,mem[12],zero,mem[13],zero,mem[14],zero,mem[15],zero
vpmovzxbw	16(%rax,%rcx), %ymm3    # ymm3 = mem[0],zero,mem[1],zero,mem[2],zero,mem[3],zero,mem[4],zero,mem[5],zero,mem[6],zero,mem[7],zero,mem[8],zero,mem[9],zero,mem[10],zero,mem[11],zero,mem[12],zero,mem[13],zero,mem[14],zero,mem[15],zero
vpmullw	%ymm0, %ymm3, %ymm3
vpmullw	%ymm0, %ymm2, %ymm2
vpminuw	%ymm1, %ymm2, %ymm2
vpminuw	%ymm1, %ymm3, %ymm3
vpackuswb	%ymm3, %ymm2, %ymm2
vpermq	$216, %ymm2, %ymm2              # ymm2 = ymm2[0,2,1,3]
vmovdqu	%ymm2, (%r13,%rcx)
addq	$32, %rcx
cmpq	%rcx, %rdx
jne	.LBB219_44
```

After (removed 2 `vpminuw` instructions):
```
vpmovzxbw	(%rax,%rcx), %ymm1      # ymm1 = mem[0],zero,mem[1],zero,mem[2],zero,mem[3],zero,mem[4],zero,mem[5],zero,mem[6],zero,mem[7],zero,mem[8],zero,mem[9],zero,mem[10],zero,mem[11],zero,mem[12],zero,mem[13],zero,mem[14],zero,mem[15],zero
vpmovzxbw	16(%rax,%rcx), %ymm2    # ymm2 = mem[0],zero,mem[1],zero,mem[2],zero,mem[3],zero,mem[4],zero,mem[5],zero,mem[6],zero,mem[7],zero,mem[8],zero,mem[9],zero,mem[10],zero,mem[11],zero,mem[12],zero,mem[13],zero,mem[14],zero,mem[15],zero
vpmullw	%ymm0, %ymm2, %ymm2
vpmullw	%ymm0, %ymm1, %ymm1
vpackuswb	%ymm2, %ymm1, %ymm1
vpermq	$216, %ymm1, %ymm1              # ymm1 = ymm1[0,2,1,3]
vmovdqu	%ymm1, (%r13,%rcx)
addq	$32, %rcx
cmpq	%rcx, %rdx
jne	.LBB219_44
```


steven-johnson:
> I don't add tests to simd_op_check_x86

Understood, but is there any other reasonable way to add checks for this? This sort of stuff is often pretty fragile.

src/Type.cpp
```diff
@@ -6,7 +6,6 @@ namespace Halide {
 
 using std::ostringstream;
 
-namespace {
 uint64_t max_uint(int bits) {
```
steven-johnson, line 9: These three functions could probably be `constexpr` -- wouldn't help in the case added to CodegenX86 but might move the needle elsewhere?

steven-johnson (approved):
LGTM but I feel like if we don't have some way to test this, we'll eventually find it broken :-/

src/Type.cpp
rootjalex: Done!

rootjalex:
> > I don't add tests to simd_op_check_x86
> 
> Understood, but is there any other reasonable way to add checks for this? This sort of stuff is often pretty fragile.

I thought about adding a "check_not" to simd_op_check, because that's the only way I could think to test this sort of optimization, but that seems like heavy functionality to use for just this case (I'm also not sure we can verify that this optimization runs by saying "don't give me a `vpmin`", because other parts of the generated code could contain a `vpmin` and we don't want false failures). I'm open to suggestions (especially because I intend to make a similar PR for HVX).

rootjalex:
We can do the same thing for HVX saturating narrow instructions, which I was going to do in a separate PR, but it requires the same changes to Bounds.cpp and Type.h that that the x86 version does, so I am making this one PR.

steven-johnson (approved):

rootjalex:
@steven-johnson Luckily, the improvement to HVX can be tested (unlike x86). Unluckily, this revealed that constant bounds inference does not work on u32 like I expected (which is part of why @abadams opened #7807)

src/CodeGen_X86.cpp
```diff
@@ -631,6 +631,37 @@ void CodeGen_X86::visit(const Call *op) {
...
+    // Search for saturating casts where the inner value can be
+    // reinterpreted to signed, so that we can use existing
+    // saturating_narrow patterns.
+    for (const auto &pattern : reinterpret_patterns) {
+        if (expr_match(pattern.pattern, op, matches)) {
+            const Expr &expr = matches[0];
+            Expr upper_bound = find_constant_bound(expr, Direction::Upper);
+            if (const uint64_t *bound = as_const_uint(upper_bound)) {
+                Type t = matches[0].type();
+                if (*bound <= (uint64_t)max_int(t.bits())) {
```
abadams, line 652: Would this be cleaner: `t.with_code(halide_type_int).can_represent(*bound)`

src/HexagonOptimize.cpp
```diff
@@ -256,6 +258,30 @@ bool process_match_flags(vector<Expr> &matches, int flags) {
         internal_assert(matches.size() >= 3);
         std::swap(matches[1], matches[2]);
     }
+    if (flags & Pattern::SafeReinterpretOp0) {
+        // Use bounds inference to check if the first operand can
+        // be safely reinterpreted.
+        const Type &t = matches[0].type();
+        std::cout << "matched with safe reinterpret\n";
```
abadams, line 265: debugging code left in

pranavb-ca (approved):

steven-johnson:
Is this PR still an active one? It seems pretty stale.

steven-johnson:
What's the story, needs more work?

rootjalex:
This has gotten stale, but I am trying to find time to update it. Andrew and I discussed some issues with using bounds inference in this way, but I think they are fixed by using the new infrastructure in #8179. I will aim to do this in the next two weeks, I've finally started to catch up on my insane backlog.

abadams:
I think it should be fairly straight-forward to adapt it to use ConstantInterval, now that the ConstantInterval refactor is in. You could do it now and there wouldn't be much of a merge conflict, or you could wait for #8155 

Here's how I use ConstantInterval to do something similar: https://github.com/halide/Halide/pull/8155/files#diff-6cd3cc01186f9ce4abaa1a98c9201bd440983ecc5620fbccae383df1116bdce6R704

All your uses are about something being safe to reinterpret as signed/unsigned, so I think you want a few instances of `const bool safe_to_reinterpret = some_expr.type().with_code(...).can_represent(constant_integer_bounds(some_expr))`

rootjalex:
Yeah, realized it wasn't a ton of work, and was bored during an ASPLOS workshop. This is good for review now! Also, the u32 HVX tests now work, because the new ConstantInterval stuff is stronger than find_constant_bounds on u32. Nice side effect.

src/CodeGen_X86.cpp
```diff
@@ -667,6 +668,38 @@ void CodeGen_X86::visit(const Call *op) {
...
+    // reinterpreted to signed, so that we can use existing
+    // saturating_narrow patterns.
+    for (const auto &pattern : reinterpret_patterns) {
+        if (expr_match(pattern.pattern, op, matches)) {
+            const Expr &expr = matches[0];
+            const Type &t = expr.type();
+            // TODO: might want to keep track of scope of bounds information.
+            const ConstantInterval ibounds = constant_integer_bounds(expr);
+            const Type reint_type = Int(t.bits());
+            if (ibounds.max_defined && reint_type.can_represent(ibounds.max)) {
```
abadams, line 690: You already make the type t.with_code(halide_type_int) below, so I suggest making reint_type that instead (just in case vector lanes has some bearing later), and then saying reint_type.can_represent(ibounds) instead so that the reader doesn't have to think about whether the min needs checking too.

src/HexagonOptimize.cpp
```diff
@@ -260,6 +263,27 @@ bool process_match_flags(vector<Expr> &matches, int flags) {
...
+        // Use bounds inference to check if the first operand can
+        // be safely reinterpreted.
+        const Expr &expr = matches[0];
+        const Type &t = expr.type();
+        if (t.is_int()) {
+            // A signed integer can be reinterpreted as unsigned if strictly positive.
+            // TODO: might want to keep track of scope of bounds information.
+            const ConstantInterval ibounds = constant_integer_bounds(expr);
+            const Type reint_type = UInt(t.bits());
+            return ibounds.min_defined && reint_type.can_represent(ibounds.min);
```
abadams, line 276: Same comment here. Use `reint_type.can_represent(ibounds)` (or just `reint_type.can_represent(constant_integer_bounds(expr))`) so the reader doesn't have to think about whether the max needs checking.

src/HexagonOptimize.cpp
```diff
@@ -260,6 +263,27 @@ bool process_match_flags(vector<Expr> &matches, int flags) {
...
+        } else {
+            internal_assert(t.is_uint());
+            // An unsigned integer can be reinterpreted as signed if bounded by int max.
+
+            // TODO: might want to keep track of scope of bounds information.
+            const ConstantInterval ibounds = constant_integer_bounds(expr);
+            const Type reint_type = Int(t.bits());
+            return ibounds.max_defined && reint_type.can_represent(ibounds.max);
```
abadams, line 284: Same comment here.

src/CodeGen_X86.cpp
rootjalex: `reint_type.can_represent(ibounds)` requires that the interval is bounded, but I only need an upper bound here (and only need a lower bound below), so I think that is less powerful than here. I will fix the `t.with_code(halide_type_int)` stuff though.

abadams: constant_integer_bounds uses the bounds of the type if it can't find something better, so constant_integer_bounds(some_int_or_uint_expr) always returns a defined upper and lower bound, with one exception: If it's a uint64, the upper bound may not be defined.

It's only really when you start doing math on the ConstantInterval objects themselves (which represent infinite integers) that you get undefined upper or lower bounds.

abadams (approved):

rootjalex:
@abadams Sorry to add a change after your approval, but I just saw a wasm talk and was reminded that we can do this for wasm too. It's basically a copy-paste from x86, but would be nice to have eyes just to be safe.
@steven-johnson just want to ping you about the wasm change, if you're curious to look

abadams:
The big TODO for wasm now is exploiting constant_integer_bounds to hit the relaxed simd instructions in case where we know they don't have UB

rootjalex:
> The big TODO for wasm now is exploiting constant_integer_bounds to hit the relaxed simd instructions in case where we know they don't have UB

Yep, I think #7312 is for tracking that. This is still a nice side benefit of constant bounds for wasm though

rootjalex:
I don't think relaxed SIMD has actually shipped in wasm yet, so that's still waiting on the wasm folks

steven-johnson:
> I don't think relaxed SIMD has actually shipped in wasm yet, so that's still waiting on the wasm folks

https://webassembly.org/features/

TL;DR: shipped in Chrome, behind flags elsewhere

PR #7814, opened 2023-08-28
Fix bounds inference for uint -> int casts

### LINKED ISSUE #7807: Bounds inference handles casts from uint32 to int32 incorrectly
See https://github.com/halide/Halide/blob/main/src/Bounds.cpp#L284

It assumes they can't overflow, but we recently changed that from UB to wrapping semantics in #7769 

Now I'm not clear if a case that would have overflowed would have compiled correctly previously (because it would have created a UB expression inside bounds inference), but now it's definitely not right. Here's a failure case that segfaults instead of throwing a bounds error:

```

#include "Halide.h"

using namespace Halide;

    Var x;
    Func f, g, h;
    Param<uint32_t> p;

    f(x) = x;

    g(x) = f(x) + f(x + p);

    f.compute_root();

    p.set(0x80000000U);
    g.realize({128});

    return 0;
}

```

### LINKED ISSUE #7811: Fuzzer failure in fuzz_simplify
Running fuzz_simplify on the enclosed file produces a failure:

```
$ ./build/linux-x64-fuzzer/test/fuzz/fuzz_simplify /tmp/testcase-5562105212239872 
Simplified Expr is not equal() to Original Expr!
Var a = -128
Var b = -128
Var c = -128
Var d = -128
Var e = -128
Original Expr is: (max(int32(select((uint32)3980344417 == uint32((int0)a), uint32((int0)a) - uint32((int0)a), (uint32)2147483648)), -128) - int32((int0)c))
Simplified Expr is: (int32(select(uint32((int0)a) == (uint32)3980344417, (uint32)0, (uint32)2147483648)) - int32((int0)c))
In vector lane 0, original -> simplified:
   (max(int32(select((uint32)3980344417 == uint32((int0)a), uint32((int0)a) - uint32((int0)a), (uint32)2147483648)), -128) - int32((int0)c)) -> 0
   (int32(select(uint32((int0)a) == (uint32)3980344417, (uint32)0, (uint32)2147483648)) - int32((int0)c)) -> -2147483520
fuzz_simplify: /usr/local/google/home/srj/GitHub/Halide/test/fuzz/simplify.cpp:359: int LLVMFuzzerTestOneInput(const uint8_t *, size_t): Assertion `test_expression(fdp, test, samples)' failed.
==747475== ERROR: libFuzzer: deadly signal
    #0 0x555f442592c4  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x39ae2c4)
    #1 0x555f441d8eb8  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x392deb8)
    #2 0x555f441bb3f3  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x39103f3)
    #3 0x7f3de617b53f  (/lib/x86_64-linux-gnu/libc.so.6+0x3c53f) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #4 0x7f3de61c912b  (/lib/x86_64-linux-gnu/libc.so.6+0x8a12b) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #5 0x7f3de617b4a1  (/lib/x86_64-linux-gnu/libc.so.6+0x3c4a1) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #6 0x7f3de61654b1  (/lib/x86_64-linux-gnu/libc.so.6+0x264b1) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #7 0x7f3de61653d4  (/lib/x86_64-linux-gnu/libc.so.6+0x263d4) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #8 0x7f3de61743a1  (/lib/x86_64-linux-gnu/libc.so.6+0x353a1) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #9 0x555f4425b314  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x39b0314)
    #10 0x555f441bc970  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x3911970)
    #11 0x555f441a6134  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x38fb134)
    #12 0x555f441abd02  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x3900d02)
    #13 0x555f441d9722  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x392e722)
    #14 0x7f3de61666c9  (/lib/x86_64-linux-gnu/libc.so.6+0x276c9) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #15 0x7f3de6166784  (/lib/x86_64-linux-gnu/libc.so.6+0x27784) (BuildId: f4017039b18cb668db130b83647b6a0dbefd4414)
    #16 0x555f4419cb20  (/usr/local/google/home/srj/GitHub/Halide/build/linux-x64-fuzzer/test/fuzz/fuzz_simplify+0x38f1b20)
```
[testcase-5562105212239872.zip](https://github.com/halide/Halide/files/12457068/testcase-5562105212239872.zip)

abadams:
Fixes #7807
Fixes #7811

Note that this may cause previously-functioning pipelines to throw a compile-time error, in cases where they cast a uint32 to an int32 for use in an index expression and were relying on Halide to assume the result is positive. Now that the uint32 -> int32 cast is defined to wrap, the result may not be positive. An example might be the expression:

```
lut(min(cast<int>(some_u32_param), 256))
```

Previously Halide would have treated this as bounded between 0 and 256. Now it's no longer bounded below, because the uint32 could have been large enough to wrap.

If this causes too much carnage in production pipelines we may have to revert the decision to define wrapping behavior for uint32 -> int32 casts.

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -281,9 +281,10 @@ private:
         bool could_overflow = true;
         if (to.can_represent(from) || to.is_float()) {
             could_overflow = false;
-        } else if (to.is_int() && to.bits() >= 32) {
-            // If we cast to an int32 or greater, assume that it won't
-            // overflow. Signed 32-bit integer overflow is undefined.
+        } else if (from.is_float() && to.is_int() && to.bits() >= 32) {
+            // If we cast from float to an int32 or greater, assume that it
+            // won't overflow. Signed 32-bit integer overflow on floating point
+            // casts is undefined. (Casting from uint is defined to wrap).
             could_overflow = false;
         } else if (a.is_bounded()) {
             if (from.can_represent(to)) {
--- a/test/correctness/bounds_of_abs.cpp
+++ b/test/correctness/bounds_of_abs.cpp
@@ -45,6 +45,19 @@ int main(int argc, char **argv) {
     f5 = lambda(x, input(cast<int>(clamp(abs(1.0f / (x + .1f)), -50, 50))));
     check(f5, input, 0, 51);
 
+    // Verify that casting the result of abs of a uint32 back to an int32 is
+    // considered unbounded below - it may produce the most negative int if the
+    // cast wraps. We won't phrase this like the tests above, because we would
+    // expect it to fail to compile with an unbounded access error.
+    Internal::Scope<Internal::Interval> scope;
+    scope.push(x.name(), Internal::Interval::everything());
+    Internal::Interval i =
+        Internal::bounds_of_expr_in_scope(cast<int>(abs(x)), scope);
+    if (i.has_lower_bound()) {
+        printf("Interval should not have had a lower bound\n");
+        return 1;
+    }
+
     printf("Success!\n");
     return 0;
 }


steven-johnson:
Testing this in Google, I get many many failures of the form `Error: Buffer foo may be accessed in an unbounded way in dimension 0`

abadams:
The question now is whether to roll back us defining uint -> int casts as having wrapping semantics, or to patch those pipelines. I'd like to slowly reduce the number of types of UB we have over time, so if patching seems tractable (e.g. by inserting a max with zero around the index), that would be best.

steven-johnson:
I'll have to figure out how many pipelines it is. (there are *lots* of failures, but maybe it's just a handful of shared pipelines.) What is the likely failure pattern I'm looking for?

abadams:
Access to an input named "foo" using a computed uint32 index. This is a bit of a guess but I think the needed patch is something like:

Before:

```foo(cast<int>(some_u32_expr), ...)```

After:

```foo(saturating_cast<int>(some_u32_expr), ...)```

**Comment by steven-johnson** [comment]
> Access to an input named "foo" using a computed uint32 index. This is a bit of a guess but I think the needed patch is something like:

The first case I'm looking at doesn't have anything obviously of that form, but there are a lot of helper functions. I wonder if there is a bottleneck for trapping most/all int->uint cases that would help?

**Comment by steven-johnson** [comment]
Running the failure case with HL_DEBUG_CODEGEN=4 reveals something possibly interesting: just before the failure, I see

```
simplify let t749 = (int16)1023 in...
new let t749.s = (undefined) in ... (int16)1023 ...
```

undefined?

**Comment by abadams** [comment]
That undefined is fine. It means the simplifier has determined the new let is not necessary to insert.

**Comment by steven-johnson** [comment]
The immediate failure looks like this:

```
Input to CSE ((uint16)2048 == (uint16)2048)
After removing lets: ((uint16)2048 == (uint16)2048)
Include: ((uint16)2048 == (uint16)2048); should extract: 1
Include: (uint16)2048; should extract: 0
Include: (uint16)2048; should extract: 0
Canonical form without lets ((uint16)2048 == (uint16)2048)
0: (uint16)2048, 0
1: ((uint16)2048 == (uint16)2048), 1
With variables ((uint16)2048 == (uint16)2048)
With lets: ((uint16)2048 == (uint16)2048)
In image input region touched is:
  0: max(min(((0*16) + input.min.0) + 0, (input.extent.0 + input.min.0) + -1), input.min.0) .. max(min(((((input.extent.0 + -1)/16)*16) + input.min.0) + 15, (input.extent.0 + input.min.0) + -1), input.min.0)
  1: max(min(((0*4) + input.min.1) + 0, (input.extent.1 + input.min.1) + -1), input.min.1) .. max(min(((((input.extent.1 + -1)/4)*4) + input.min.1) + 3, (input.extent.1 + input.min.1) + -1), input.min.1)
Injecting constraints for input.0
Injecting constraints for input.1
In image original_input region touched is:
  0: (void *)neg_inf .. (void *)pos_inf
  1: (void *)neg_inf .. (void *)pos_inf
User error triggered at third_party/halide/halide/src/AddImageChecks.cpp:383
```

**Comment by abadams** [comment]
I pushed a temporary change that should give a more useful error

**Comment by steven-johnson** [comment]
<spooky voice> The failure is coming from INSIDE HALIDE! </>

Look at `mirror_interior()` and tell me where the uint32->int32 cast is (implicitly) creeping in:
```
            Expr limit = extent - 1;
            Expr coord = arg_var - min;   // Enforce zero origin.
            coord = coord % (2 * limit);  // Range is 0 to 2w-1
            coord = coord - limit;        // Range is -w, w
            coord = abs(coord);           // Range is 0, w
            coord = limit - coord;        // Range is 0, w
            coord = coord + min;          // Restore correct min

            // The boundary condition probably doesn't apply
            coord = select(arg_var < min || arg_var >= min + extent, coord,
                           clamp(likely(arg_var), min, min + extent - 1));
```

(Spoiler alert: `abs()` in Halide always returns an unsigned.)

(In hindsight, I really wish that Halide had adopted the Golang property of never doing implicit type conversions...)

**Comment by steven-johnson** [comment]
Unfortunately, if I change the bad line to `coord = saturating_cast<int>(abs(coord));`, we still fail with the same result (input is touched in an infinite way), but none of the "dodgy" check warnings you inserted trigger this time (they did previously, which is how I found the mirror issue). I'm gonna guess there is more `abs()` nonsense going on.

**Comment by abadams** [comment]
Can repro! I'll investigate.

**Comment by abadams** [comment]
Possible fix pushed

**Comment by steven-johnson** [comment]
OK, that fixes the unbounded access... now I get failures with `Error: Signed integer overflow occurred during constant-folding. Signed integer overflow for int32 and int64 is undefined behavior in Halide.`  Let me investigate that.

**Comment by steven-johnson** [comment]
Yeah, this is another one where it is gonna be tricky to find these errors in existing code without some helpers.

**Comment by abadams** [comment]
I pushed a user_warning for when bounds inference does something different to what it used to do.

**Comment by steven-johnson** [comment]
The first one I found basically amounted to code that used explicit casting to do a 'widening' mul (this code predates the `widening_mul()` etc operators:

```
    // f and h are both funcs with type i16
    Expr r = i32(f(x, y, 0));
    Expr g = i32(f(x, y, 1));
    Expr b = i32(f(x, y, 2));
    Expr result = h(0, c) * r + h(1, c) * g + h(2, c) * b;
```

switching it to use `widening_mul` seems to heal it (not sure if result is correct), but many others to track down. Will update later.

**Comment by abadams** [comment]
I'm a little alarmed that widening_mul heals that, because any 16-bit -> i32 cast should be bounded, because i32 can represent i16 and u16. Also, widening_mul just means the same thing.

**Comment by steven-johnson** [comment]
> Also, widening_mul just means the same thing.

The evidence suggests otherwise? Not sure I can give you an easy repro case from it but maybe using that snippet and seeing how the calculated bounds differ would be useful.

**Comment by steven-johnson** [comment]
> I pushed a user_warning for when bounds inference does something different to what it used to do.

I'll apply that and retry.

**Comment by steven-johnson** [comment]
> I'm a little alarmed that widening_mul heals that, because any 16-bit -> i32 cast should be bounded, because i32 can represent i16 and u16. Also, widening_mul just means the same thing.

Update: maybe that's because the Call visitor in Bounds calculation doesn't have a case for `widening_mul`, so it just defaults to max-of-type for an unknown Func, which in this case is '+/- inf'. 

If I add a specialization that is similar to `widening_mul_right`, eg

```
        } else if (op->is_intrinsic(Call::widening_mul)) {
            Expr mul = Mul::make(cast(op->type, op->args[0]), cast(op->type, op->args[1]));
            mul.accept(this);
        }
```

..then using `widening_mul` **also** fails with signed-integer-overflow.

**Comment by steven-johnson** [comment]
Also interesting (though maybe not 100% relevant here): we do specialize for shift_left, but only for signed integers, not for unsigned... in the case of this code, we have a u32 << u32 case where LHS is in 0..65535 and RHS is in 0..16, so bounds should be 0..0xFFFF0000... but since we don't handle it, bounds is neg_inf..pos_inf

**Comment by steven-johnson** [comment]
This is fairly old, where does it stand?

**Comment by abadams** [comment]
This one is pending figuring out what breaks inside of google

**Comment by steven-johnson** [comment]
> I think we were knocking cases down one by one, but there are more failures that we have yet to diagnose.

ok, I will pull it in and see where we stand

**Comment by steven-johnson** [comment]
So the fact that `abs()` has always returns a uint means that this change will likely mean that ~every call to it will need attention (e.g., something like `abs(y)` -> `max(abs(y), 0)` to ensure no negative values).

**Comment by abadams** [comment]
I see, so people are doing things like my_func(abs(expr)), and that's implicitly my_func(i32(abs(expr))), and the expr could be the most-negative integer if we know nothing about it, in which case that cast now returns the most-negative integer instead of being UB.

**Comment by abadams** [comment]
Wait, that's can't be the example, because if we know nothing about my_expr in that case the access is unbounded anyway. I guess there are cases where we know an upper bound on my_expr but not a lower bound?

**Comment by steven-johnson** [comment]
> I see, so people are doing things like my_func(abs(expr)), and that's implicitly my_func(i32(abs(expr))), and the expr could be the most-negative integer if we know nothing about it, in which case that cast now returns the most-negative integer instead of being UB.

yeah, this is pervasive too, it makes me question whether this is the right design for abs() -- maybe we should change it to return the same type for signed. alternately, add `iabs()` which does something like `clamp(abs(x), 0, max-for-type)` for int...

**Comment by steven-johnson** [comment]
> Wait, that can't be the example, because if we know nothing about expr in that case the access is unbounded anyway. I guess there are cases where we know an upper bound on expr but not a lower bound?

in this case the full expr is `Expr mirror_y = clamp(min(abs(y), 2 * max_y + 1 - y), 0, max_y);`

**Comment by abadams** [comment]
If this really becomes the show-stopper, we could declare that abs of the most negative 32/64-bit integer is UB (which it already is in main if you cast the result back to int due to the signed integer overflow), and then handle int casts of abs specially in bounds inference.

**Comment by steven-johnson** [comment]
I don't know if it's a showstopper, but in the first failure I looked at, literally every failure warning was about abs()

**Comment by steven-johnson** [comment]
I tried writing a shim like so:

```
Halide::Expr iabs(Halide::Expr i) {
  if (i.type().is_int()) {
    return cast<int32_t>(min(abs(i), cast<uint32_t>(i.type().max())));
  } else {
    return abs(i);
  }
}
```

but it doesn't help, e.g.

```
Formerly bounded cast is no longer bounded:
 Cast: int32(min((uint32)abs((blur_x.s0.x.guarded/2) + -1), (uint32)2147483647))
 Bounds of arg: (uint32)0 min(max(uint32(((4 - min(input.extent.0, 1))/2)), uint32(((min(max(input.extent.0, 1), min(max(input.extent.0, 1) - min(input.extent.0, 1), 7) + min(max(input.extent.0, 1), min(input.extent.0, 1) + (((max(input.extent.0, 1) - min(input.extent.0, 1))/8)*8))) + -3)/2))), (uint32)2147483647)
 Bounds of output: (void *)neg_inf (void *)pos_inf
```

not sure if this is just a false positive in the sniffing code or not? but I still fail with a buffer being accessed in an unbounded way.

steven-johnson:
`ContainsDodgyIntCast` may not be smart enough, it doesn't check to see if the value in question has been constrained to a legal range. Maybe something like changing `can_represent()` -> `can_prove(op->value >= op->type.min() && op->value <= op->type.max())`?

abadams:
This is technically a bug fix, but it seemed like it caused too many problems in production due to the abs issue. Not sure what to do about it.

PR #7825, opened 2023-08-30
[Hexagon] - Fix 8-bit unsigned saturating downcasts for HVX (Fixes #7806)

### LINKED ISSUE #7806: [HVX] Failure on seemingly valid schedule
I'm getting a failure related to interleaving on a simple HVX schedule:
```
#include "Halide.h"
#include <iostream>
#include <stdio.h>

using namespace Halide;

void test0() {
    Var x("x");
    ImageParam in0(UInt(8), 1);
    Target hvx("hexagon-32-noos-hvx_128-hvx_v66");
    Func out("out");
    out(x) = u8_sat(u16(in0(x)) * 5);
    out.vectorize(x, 128);
    out.compile_to_assembly("hvx-test0.asm", out.infer_arguments(), hvx);
}

int main(int arch, char **argv) {
    test0();
}
```

```
ajroot@Alexanders-MacBook-Pro Halide % ./bin/correctness_hvx
Internal Error at /Users/ajroot/projects/Halide/src/CodeGen_Hexagon.cpp:1883 triggered by user code at :
Condition failed: op->type.lanes() % (native_vector_bits() * 2 / op->type.bits()) == 0:
zsh: abort      ./bin/correctness_hvx
```

@pranavb-ca do you know what might fix this?

pranavb-ca:
(Fixes #7806)
We do not have a way of reliably lowering the following expression
to LLVM bitcode for HVX.

```
u8_sat(uint16x)
```
where uint16x is a vector (preferably a HVX double vector) with element type uint16.
Since there is no native HVX instruction to do this, this PR introduces two helper functions in hvx_128.ll to perform this
operation. One function interleaves its input (trunc_satub.vuh) and the other does not (pack_satub.vuh)

This patch also removes declaration of some intrinsics not used any longer in hvx_128.ll


rootjalex:
Is this equivalent to calling lower_intrinsic(expr) on any unsigned -> unsigned saturating casts? Or is there something extra happening here due to the lack of unsigned min?

pranavb-ca:
> Is this equivalent to calling lower_intrinsic(expr) on any unsigned -> unsigned saturating casts? Or is there something extra happening here due to the lack of unsigned min?

Both. 

In the case of ``halide.hexagon.pack_satub.vuh``, I'd think it would be equivalent to lower_intrinsics(expr). Of course, I haven't tested whether the expanded code resulting from ``lower_saturating_cast`` would turn into code as tight as ``halide.hexagon.pack_satub.vuh``. I could give that a try though.

In the case of ``halide.hexagon.pack_satuh.vuw``, there is something extra happening because of the lack of unsigned min for unsigned words. This case is probably best handled using and hvx specific implementation in bitcode (``hvx_128.ll``)

steven-johnson:
Looks like still two hvx failures: https://buildbot.halide-lang.org/master/#/builders/163/builds/157

pranavb-ca:
> > Is this equivalent to calling lower_intrinsic(expr) on any unsigned -> unsigned saturating casts? Or is there something extra happening here due to the lack of unsigned min?
> 
> Both.
> 
> In the case of `halide.hexagon.pack_satub.vuh`, I'd think it would be equivalent to lower_intrinsics(expr). Of course, I haven't tested whether the expanded code resulting from `lower_saturating_cast` would turn into code as tight as `halide.hexagon.pack_satub.vuh`. I could give that a try though.
> 
> In the case of `halide.hexagon.pack_satuh.vuw`, there is something extra happening because of the lack of unsigned min for unsigned words. This case is probably best handled using and hvx specific implementation in bitcode (`hvx_128.ll`)

@rootjalex  -  I was out on vacation and am back now. I took a look at this briefly today. while ``lower_intrinsic(u8_sat(u16x)`` does a fairly good job, as I had suspected the output code is not as tight as converting to ``halide.heaxgon.pack_satub.vuh``. The reason is not that ``halide.hexagon.pack.satub.vuh``.

 
``lower_intrinsic(u8_sat(u16x)`` gets expanded to ``halide.hexagon.pack.vh(min(u16x, 255)``. In cases where u16x is a native interleave i.e ``u16x`` is essentially ``halide.hexagon.interleave.vuh(...)``, ``EliminateInterleaves`` separates the saturation (``min(.., 255)``) from the truncation (``halide.hexagon.pack.vh``). As a result we saturate and truncate separately.

So, I think sticking to what I have done is going to yield better results. Let me know what you think.

rootjalex:
> `lower_intrinsic(u8_sat(u16x)` gets expanded to `halide.hexagon.pack.vh(min(u16x, 255)`. In cases where u16x is a native interleave i.e `u16x` is essentially `halide.hexagon.interleave.vuh(...)`, `EliminateInterleaves` separates the saturation (`min(.., 255)`) from the truncation (`halide.hexagon.pack.vh`). As a result we saturate and truncate separately.

This seems like a case for improving `EliminateInterleaves`, but this fix seems fine for now, until that can be done.

rootjalex (approved):

PR #7829, opened 2023-09-01
[api] Promote Internal::Parameter to Halide::Parameter 

derek-gerstmann:
As per the discussion regarding refactoring the Serialization API: 
https://github.com/halide/Halide/pull/7760


steven-johnson (commented):
From a quick skim, it looks like this change is simpy moving it into the right namespace -- we also need to take a close look at the methods that Parameter exposes (including the ctors); IMHO we want to prune it down to the bare minimum we think are necessary (enforcing via `friend` access if necessary), because otherwise we'll end up with a lot of weird usage we didn't expect (https://www.hyrumslaw.com/).

Once that is done, we will also need to provide a proper Python wrapper for the resulting API; the current Python wrapper (in PyParam.cpp) is deliberately a bare-minimum wrapper that was put in place only for some internal use, but with Parameter become a public Halide object, we need to ensure the Python wrapper provides the same.

derek-gerstmann:
> From a quick skim, it looks like this change is simpy moving it into the right namespace -- we also need to take a close look at the methods that Parameter exposes (including the ctors); IMHO we want to prune it down to the bare minimum we think are necessary (enforcing via `friend` access if necessary), because otherwise we'll end up with a lot of weird usage we didn't expect (https://www.hyrumslaw.com/).
> 
> Once that is done, we will also need to provide a proper Python wrapper for the resulting API; the current Python wrapper (in PyParam.cpp) is deliberately a bare-minimum wrapper that was put in place only for some internal use, but with Parameter become a public Halide object, we need to ensure the Python wrapper provides the same.

Agreed. I intentionally left obvious internal methods (like `remove_self_references` and `restore_self_references`) hidden, but I didn't got so far as to change the class interface, since I wasn't entirely sure of the impact.

derek-gerstmann:
I made the raw_buffer(), scalar_address(), and scalar_raw_value() access methods protected, and added Pipeline and Serializer as protected friends.

I believe these are the only potentially controversial class methods which expose addresses and internal values.

src/Generator.cpp
```diff
@@ -1645,7 +1645,7 @@ bool GeneratorBase::emit_hlpipe(const std::string &hlpipe_file_path) {
         debug(1) << "Applying autoscheduler " << asp.name << " to Generator " << name() << " ...\n";
         auto_schedule_results = pipeline.apply_autoscheduler(context.target(), asp);
     }
-    std::map<std::string, Internal::Parameter> params;  // FIXME: Remove when API allows this to be optional
+    std::map<std::string, Parameter> params;  // FIXME: Remove when API allows this to be optional
```
steven-johnson, line 1648: Is there an issue tracking this?

steven-johnson (changes requested):
LGTM but I think we need to also flush out the Python wrapper for `Parameter` -- as mentioned elsewhere, the Python wrapper is currently just a minimal stub for internal use, but with this change, we need to promote it to full public status, and provide all the public methods in Python as well as C++.

src/Generator.cpp
derek-gerstmann: Just added this as #7849

derek-gerstmann:
> LGTM but I think we need to also flush out the Python wrapper for `Parameter` -- as mentioned elsewhere, the Python wrapper is currently just a minimal stub for internal use, but with this change, we need to promote it to full public status, and provide all the public methods in Python as well as C++.

Should we update the python bindings in this PR or open a new one?

steven-johnson:
> > LGTM but I think we need to also flush out the Python wrapper for `Parameter` -- as mentioned elsewhere, the Python wrapper is currently just a minimal stub for internal use, but with this change, we need to promote it to full public status, and provide all the public methods in Python as well as C++.
> 
> Should we update the python bindings in this PR or open a new one?

Unless there is an urgent reason we need to land this quickly, I think it would be better to work out the Python bindings in this PR as well; I've found it a useful exercise in the past to winnow out the calls that are actually necessary for the public API from the ones that aren't.

python_bindings/src/halide/halide_/PyParameter.cpp
```diff
@@ -0,0 +1,107 @@
...
+            "set_scalar", [](Parameter &parameter, TYPE value) -> void {
+                parameter.set_scalar<TYPE>(value);
+            },
+            py::arg("value"));
+}
+
+}  // namespace
+
+void define_parameter(py::module &m) {
+
```
steven-johnson, line 24: nit: no blank line at start of func, please

steven-johnson (approved):
LGTM pending green

steven-johnson:
LGTM

PR #7830, opened 2023-09-01
Iterate over lets in the correct order in VectorizeLoops

### LINKED ISSUE #7779: vectorize loops triggers name not in scope error
A simple repro:
```cpp
#include "Halide.h"

using namespace Halide;

    Var x("x"), y("y"), yo("yo"), yi("yi"), yoi("yoi"), yoio("yoio"), yoii("yoii");
    Func f("f");
    f(x, y) = x + y;
    f.split(y, yo, yi, 8, TailStrategy::Auto).split(yo, yo, yoi, 4, TailStrategy::RoundUp).vectorize(yoi).vectorize(yi).split(yoi, yoio, yoii, 2, TailStrategy::Auto);
    Pipeline p({f});
    std::cout << "before realize\n";
    Buffer<int> buf = p.realize({128, 128});
    std::cout << "realize done\n";
    return 0;
}
```

This code will trigger the following internal error under `HL_DEBUG_CODEGEN=2`
```
Vectorizing...
Subcontext threw exception. Rethrowing...
Internal Error at /home/xuanda/dev/Serializer/Halide-clean/src/Scope.h:134 triggered by user code at : Name not in Scope: f.s0.y.yo.widened.f.s0.y.yi
{
  f.s0.y.yi.base.widened.f.s0.y.yo.yoi.yoii
  f.s0.y.yo.widened.f.s0.y.yo.yoi.yoii
}
```

vksnk:
We need to iterate over vars in insertion order, but the Scope doesn't preserver the order.

Fixes #7779

### DIFF
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -960,8 +960,8 @@ class VectorSubs : public IRMutator {
 
             vectorized_vars.push_back({op->name, min, (int)extent_int->value});
             update_replacements();
-            // Go over lets which were vectorized in order of occurance and update them according
-            // to the current loop level.
+            // Go over lets which were vectorized in the order of their occurance and update
+            // them according to the current loop level.
             for (auto let = containing_lets.begin(); let != containing_lets.end(); let++) {
                 // Skip if this var wasn't vectorized.
                 if (!scope.contains(let->first)) {
--- a/src/VectorizeLoops.cpp
+++ b/src/VectorizeLoops.cpp
@@ -960,7 +960,7 @@ class VectorSubs : public IRMutator {
 
             vectorized_vars.push_back({op->name, min, (int)extent_int->value});
             update_replacements();
-            // Go over lets which were vectorized in the order of their occurance and update
+            // Go over lets which were vectorized in the order of their occurrence and update
             // them according to the current loop level.
             for (auto let = containing_lets.begin(); let != containing_lets.end(); let++) {
                 // Skip if this var wasn't vectorized.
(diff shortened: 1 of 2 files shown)


src/VectorizeLoops.cpp
```diff
@@ -962,17 +962,23 @@ class VectorSubs : public IRMutator {
             update_replacements();
             // Go over lets which were vectorized and update them according to the current
             // loop level.
-            for (auto it = scope.cbegin(); it != scope.cend(); ++it) {
-                string vectorized_name = get_widened_var_name(it.name());
-                Expr vectorized_value = mutate(it.value());
+            for (auto let = containing_lets.begin(); let != containing_lets.end(); let++) {
```
steven-johnson, line 965: Do we need to add comment(s) here and below about why we are doing this?

src/VectorizeLoops.cpp
```diff
@@ -960,9 +960,10 @@ class VectorSubs : public IRMutator {
 
             vectorized_vars.push_back({op->name, min, (int)extent_int->value});
             update_replacements();
-            // Go over lets which were vectorized and update them according to the current
-            // loop level.
+            // Go over lets which were vectorized in order of occurance and update them according
```
steven-johnson: nit: s/occurance/occurrence/

steven-johnson (approved):

src/VectorizeLoops.cpp
vksnk: I updated the comments, but it really was just my mistake incorrectly assuming that Scope preserves insertion order (which it does but only within *specific* name).

src/VectorizeLoops.cpp
vksnk: Done

vksnk:
Looks like failure are unrelated, going to  merge in.

Issue #7831, opened 2023-09-01
onnx app is failing on one of the x86 mac builders


abadams:
I will temporarily disable that app on Mac. Opening an issue to track it.

abadams:
> [!NOTE]
> **Automated triage:** this issue was flagged by Claude as likely safe to close, for the reason below. The reasoning was checked against the current source, and a second pass tried and failed to refute it, but no human verified this specific text. **If the reason is wrong, please reopen the issue and say so** — some fraction of these are expected to be wrong.

**Reason (CLOSE_STALE):** The issue records only "the onnx app fails on an x86 mac builder" with no error text, no repro, and no comments. The machine it concerns no longer exists: build_bot removed all macOS workers, and `apps/` is not built by the GitHub Actions macOS jobs.

**Evidence:** halide/build_bot commits `a4eb1a4b` and `e3858d0c` (2026-07-24) removed macOS workers and macOS worker scripts from `master.cfg`; the current `master.cfg` has no onnx handling at all. `apps/` is a standalone project not referenced from the top-level `CMakeLists.txt`, and `.github/workflows/testing-macos.yml` builds only the `ci-macos-*` presets. The app itself has since been reworked (`52ae4dd4c`, ONNX 1.19.0) and is enabled unconditionally at `apps/CMakeLists.txt:74`.

Closing as stale: the x86 macOS buildbot this refers to no longer exists (macOS workers were removed from build_bot in July 2026, and the GHA macOS jobs don't build `apps/`), and the issue records no error output to act on. Please reopen with logs if `apps/onnx` fails on macOS today.

PR #7832, opened 2023-09-01
Skip onnx app on mac

abadams:
See #7831


steven-johnson (approved):

abadams:
Last I checked the bot is broken in other ways, so this is sort of pointless because it won't unbreak the build.

PR #7837, opened 2023-09-05
Enable emission of float16/32 casts on x86

### LINKED ISSUE #4166: llvm backends by default use i16 instead of half for float16s.
This is due to problems upstream in llvm: https://bugs.llvm.org/show_bug.cgi?id=43065

We should switch to half types universally once llvm support is better.

### LINKED ISSUE #7836: Generator with F16C target emits emulation code for float16_t->float casts
Checked with Halide-16.0.0. Generator target: `target=x86-64-linux-avx-avx2-f16c-fma-sse41` raises the following warning for float16 ->f32 conversion:
```
Warning: In function my_gen, (b)float16 type operation is emulated, which is likely to slow down the performance. If your target supports native (b)float16 operations, it could be improved by adding Target feature to enable it.
```

Generator code:
```
#include "Halide.h"

using namespace Halide;

class HalfFloatMatMul : public Halide::Generator<HalfFloatMatMul> {
public:
    Input<Buffer<float16_t, 2>> A{"A"};
    Input<Buffer<float16_t, 2>> B{"B"};

    Output<Buffer<float, 2>> output{"C"};

    void generate() {
        RDom r(0, A.dim(0).extent(), "r");
    	matmul(x, y) += cast<float>(A(r, y)) * cast<float>(B(r, x));
        output(x, y) = matmul(x, y);
    }

    void schedule() {
        if (using_autoscheduler()) {
            A.set_estimates({{0, 1024}, {0, 1024}});
            B.set_estimates({{0, 1024}, {0, 1024}});
            output.set_estimates({{0, 1024}, {0, 1024}});
        } else {
            matmul.compute_root();
        }
    }

private:
    Func matmul;
    Var x{"x"}, y{"y"};
};

HALIDE_REGISTER_GENERATOR(HalfFloatMatMul, my_gen)
```

I expect to see `vcvtph2ps` (`_mm256_cvtph_ps`) in generated assembly. Is it possible to change anything to emit it?

abadams:
Fixes #7836
Fixes #4166

These weren't enabled yet because the last time this code was touched, float16 wasn't a legal type in llvm on x86 (See https://bugs.llvm.org/show_bug.cgi?id=43065)

### DIFF
--- a/test/correctness/simd_op_check_x86.cpp
+++ b/test/correctness/simd_op_check_x86.cpp
@@ -644,9 +644,9 @@ int main(int argc, char **argv) {
         {
             Target("x86-32-linux"),
             Target("x86-32-linux-sse41"),
-            // Always turn on f16c when using avx. Haswell had avx without f16c,
-            // but f16c is orthogonal to everything else, so there's no real
-            // reason to test avx without it.
+            // Always turn on f16c when using avx. Sandy Bridge had avx without
+            // f16c, but f16c is orthogonal to everything else, so there's no
+            // real reason to test avx without it.
             Target("x86-64-linux-sse41-avx-f16c"),
             Target("x86-64-linux-sse41-avx-f16c-avx2"),
             // See above: don't test avx512 without extra features, the test
--- a/test/correctness/simd_op_check_x86.cpp
+++ b/test/correctness/simd_op_check_x86.cpp
@@ -45,6 +45,7 @@ public:
     void check_sse_and_avx() {
         Expr f64_1 = in_f64(x), f64_2 = in_f64(x + 16), f64_3 = in_f64(x + 32);
         Expr f32_1 = in_f32(x), f32_2 = in_f32(x + 16), f32_3 = in_f32(x + 32);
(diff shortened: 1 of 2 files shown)


vksnk (approved):

test/correctness/simd_op_check_x86.cpp
```diff
@@ -638,14 +644,17 @@ int main(int argc, char **argv) {
         {
             Target("x86-32-linux"),
             Target("x86-32-linux-sse41"),
-            Target("x86-64-linux-sse41-avx"),
-            Target("x86-64-linux-sse41-avx-avx2"),
+            // Always turn on f16c when using avx. Haswell had avx without f16c,
```
AngryLoki, line 647: Actually, Haswell had F16C (together with newly-added AVX2+FMA). Even IvyBridge supported F16C (without AVX2 or FMA). Only SandyBridge was pure AVX.

abadams: Oops, I had haswell and ivy bridge in the wrong order in my head. I'll fix the comment.

abadams:
Failures unrelated.

PR #7838, opened 2023-09-05
[tutorials] Add tutorial on JIT compile/execute performance

derek-gerstmann:
Compares performance of realize(), compile_jit(), compile_callable(), compile_module() and shows benefits of JIT cache.


tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
+// Halide tutorial lesson 22: JIT compilation performance
+
+// This lesson demonstrates the various performance implications of the
+// various Halide methods of doing "Just-In-Time" compilation.
+
+// On linux, you can compile and run it like so:
```
steven-johnson, line 6: possibly  dumb q: have we verified that all these instructions still work

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+// in a shell at the top of the halide source tree.
+
+#include "Halide.h"
+#include "halide_benchmark.h"
+#include <stdio.h>
+
+using namespace Halide;
+using namespace Halide::Tools; // for benchmark()
+
+// Lets define a helper function to construct a simple pipeline that we'll use for our performance tests
```
steven-johnson, line 26: s/Lets/Let's/

Also, nit here and elsewhere: comments that appear to be intended to be complete sentences should probably end in punctuation.

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+        std::cout << "Execute Pipeline (realize): " << int(count / t) << " times/sec\n";
+    }
+
+    // This time let's explicitly invoke the JIT compile, before we realize and execute the pipeline
+    {
+        Pipeline pipeline = make_pipeline();
+        Buffer<uint16_t> result(1024, 1024);
+
+        // Let's explicitly JIT compile before we realize, which will cache the generated code
```
steven-johnson, line 86: Suggestion: add comments indicating the rough expected improvement on a typical desktop system

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+        auto arguments = pipeline.infer_arguments();
+        const Target target = get_jit_target_from_environment();
+        Callable callable = pipeline.compile_to_callable(arguments, target);
+
+        // The Callable object acts as a convienient way of invoking the compiled code like
+        // a function call, using an argv like syntax for the argument list. It also caches 
+        // the JIT compiled code, so there's no code generation overhead when invoking the
+        // callable object and executing the pipeline.
+
```
steven-johnson, line 113: Suggestion: add a clause after this one that uses `compile_to_std_function()`; its benefit is mostly cleaner type checking/coercion, but the call overhead is also slightly lower than `Callable` so it may measure as (very slightly) faster (but even if not, having an example that demos it is worthwhile)

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+    Var x("x"), y("y");
+
+    // Fill the input with a linear combination of the coordinate values
+    input(x, y) = cast<uint16_t>(x + y);
+    input.compute_root();
+
+    // Transpose the rows and cols 
+    output(x, y) = input(y, x);
+
+    // Schedule it ... there's a number of possibilities here to do an efficient block-wise transpose
```
abadams, line 40: Nit: there's -> there are

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+    Var xi("xi"), yi("yi");
+    
+    // Let's focus on 8x8 subtiles, and then vectorize across X, and unroll across Y
+    output.tile(x, y, xi, yi, 8, 8).vectorize(xi).unroll(yi);
+
+    // For more advanced scheduling:
```
abadams, line 46: If you're going to mention the advanced schedule, I'd just used it.

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+    {
+        Pipeline pipeline = make_pipeline();
+        Buffer<uint16_t> result(1024, 1024);
+
+        auto arguments = pipeline.infer_arguments();
+        const Target target = get_jit_target_from_environment();
+        Callable callable = pipeline.compile_to_callable(arguments, target);
+
+        // The Callable object acts as a convienient way of invoking the compiled code like
+        // a function call, using an argv like syntax for the argument list. It also caches
```
abadams, line 110: argv-like

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,174 @@
...
+    {
+        Pipeline pipeline = make_pipeline();
+
+        // Create an 1024x1024 output buffer to hold the results 
+        Buffer<uint16_t> result(1024, 1024);
+
+        // Lets realize the pipeline (which implicitly JIT compiles, dynamically links and executes the code)
+        size_t count = 0;
+        double t = benchmark([&]() {
+            pipeline.realize(result);
```
abadams, line 74: This isn't really different to the case below because realize is only going to compile on first run through the benchmarking loop. I think you should put make_pipeline in the benchmarking loop here, because that's a common mistake people make that is much slower than building the pipeline once and realizing it multiple times.

abadams:
I think the cases to test here are the ones that match real usage patterns:

1) Defining and compiling the whole pipeline every time you want to run it (i.e. in the benchmarking loop)
2) Defining the pipeline outside the benchmarking loop, and realizing it repeatedly.
3) (optional) Same as 2), but calling compile_jit() outside the loop, saying what it does, and saying why the time isn't actually different to case 2 (benchmark() runs multiple times and takes a min, and realize only compiles on the first run)
4) Compiling to a callable outside the benchmarking loop and showing that it has lower overhead than case 3 (if indeed it does. If not we may need to change the example so that it does, e.g. by adding a real input buffer.)

Another subtlety we could consider is the difference between allocating the output buffer outside the benchmarking loop vs using the form of realize that allocates an output for you, e.g. realize({1024, 1024});

derek-gerstmann:
> 

Ah, yes ... great suggestions!  I'll swap out the contrived examples and add the above use cases you listed.

derek-gerstmann:
Updated with new cases as suggested.

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,263 @@
...
+        // much of a difference to the previous case.  However, the first call to realize() will implicitly
+        // JIT-compile and cache the generated code associated with the Pipeline object, which is basically 
+        // what we've done here. Each subsequent call to realize uses the cached version of the native code, 
+        // so there's no additional overhead, and the cost is amortized as we re-use the pipeline.
+        std::cout << "Execute Pipeline (compile before realize): " << int(count / t) << " times/sec\n";
+
+
+        // Another subtlety is the creation of the result buffer ... the declaration implicitly
+        // allocates memory. Instead, we could use the realize({1024, 1024}) call which use the
+        // buffer managed by the pipeline object for the outputs ...
```
steven-johnson, line 136: Another tiny style nit here and elsewhere: I think it's more correct to omit the space before an ellipsis, i.e., `for the outputs...` rather than `for the outputs ...`

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+        count = 0;
+        t = benchmark(samples, iterations, [&]() {
+            pipeline.realize(result);
+            ++count;
+        });
+
+        // On a MacBook Pro M1, this should be much more efficient ... ~200000 times/sec (or 10-12% faster).
+        std::cout << "Execute Pipeline (re-use buffer with realize): " << int(count / t) << " times/sec\n";
+
```
steven-johnson, line 159: Nit: unnecessary blank line before the end of the block

tutorial/lesson_22_jit_performance.cpp
```diff
@@ -0,0 +1,263 @@
...
+    {
+        Pipeline pipeline = make_pipeline();
+        const Target target = get_jit_target_from_environment();
+
+        // Here, we can ask the pipeline for its argument list (these are either Params,
+        // ImageParams, or Buffers) so that we can construct a Callable object with the same 
+        // calling convention.
+        auto arguments = pipeline.infer_arguments();
+
+        // The Callable object acts as a convienient way of invoking the compiled code like
```
steven-johnson, line 172: convenient

(side note: it's 2023, why don't we have smart spellcheckers that are useful for typos in code comments?)

steven-johnson (approved):
More minor nits but otherwise this is great, LGTM

Issue #7839, opened 2023-09-06
host target detection shouldn't set features that are implied by other features


abadams:
More generally, we need to be smarter about features that imply other features.

If you manually set the target, e.g. x86-64-linux-avx2, then the x86 backend completes this target by setting all implied features to produce x86-64-linux-sse41-avx-f16c-fma-avx2 and uses that to make instruction selection decisions. So you don't need to manually set implied features in a build(*).

Host target detection on such a machine will return the latter completed target rather than the former. This is getting increasingly ridiculous. Sapphire rapids avx512 implies everything in cannonlake avx512, skylake avx512, and all the way down, so on a sapphire rapids machine, host target detection returns:

x86-64-linux-sse41-avx-f16c-fma-avx2-avx512-avx512_skylake-avx512_cannonlake-avx512_sapphire_rapids

zen4's avx512 support is a superset of cannonlake and a subset of sapphire rapids, so once that's merged we'd generate:

x86-64-linux-sse41-avx-f16c-fma-avx2-avx512-avx512_skylake-avx512_cannonlake-avx512_zen4-avx512_sapphire_rapids

This discourages us from being too granular with what generations of thing we support, which is silly. Target strings shouldn't be size O(n) after n generations. If we wanted to be very precise on arm, we'd have to have target strings like:

arm-64-linux-aarchv81-aarchv82-aarchv84-aarchv85 ....

I think we should stop manually setting implied features, and rely on the backends to do the feature bit completion on a local copy of the target as necessary if it's convenient for them to make instruction selection decisions. The only thing we'd really need to change to enact this is to stop setting implied features when we detect the host target. The rest is a matter of usage rather than a code change.

A gotcha I anticipate is if a user schedule switches based on something like get_host_target().has_feature(Target::AVX). We might address this with a new target accessor (supports_feature?) that returns true for implied features too. Or maybe has_feature should return true if the feature is implied. Target.cpp would have to keep a list of feature implications (e.g. avx2 -> avx), and that would also let us centralize target completion for the backends.

Yet another option is to set implied features but not print them when converting a target to a string. I think it's probably better if the set flags in the struct mirrors the string exactly though.

(*) though making that change would break schedules that do things like has_feature(Target::AVX) and want that to return true for avx2, so for existing codebases (e.g. google3) we may not want to change the target strings in use.

zvookin:
[...]

Issue #7834, opened 2023-09-03
illegal instructions issued on ryzen 7800x3d with target "host"


abadams:
We detect this chip as avx512 sapphire rapids.

The specific example I found was vpdpwssd on an xmm register, which is part of avx512_vnni and avx512vl. Zen 4 claims to support these, so I'm confused.

Another example was vmovw from an xmm register to a general purpose register, which is supposed to be part of avx512_fp16. This is not supposed to be supported by zen 4, so that makes more sense.

abadams:
I believe my confusion was that the vpdpwssd on an xmm register is actually part of avx_vnni, which is not implied by avx512_vnni. Sapphire rapids supports avx_vnni but Zen4 does not.

PR #7840, opened 2023-09-06
Zen4 support

abadams:
This PR adds support for Zen4 chips. This is currently broken, because we are misdetecting them as sapphire rapids (#7834). There were also some avx512_bf16 things broken in llvm that we weren't aware of because we have no buildbots that support it.

It turns out the subset of avx512 supported by Zen4 is a strict superset of cannonlake, and a strict subset of sapphire rapids, so it was easy to slot into the target feature list. However, I think we need to have a discussion about how long these x86 target strings are getting (#7839)

The other thing in this PR we should discuss is feature-bit-based host CPU detection vs detecting specific microarchitectures and setting the right flags. There were comments in the existing source indicating we should do the latter, so that's what I did. I'm not completely sold on that though because it doesn't seem very future-proof.

This PR is necessary to bring online a new Zen4 buildbot.


vksnk (approved):

steven-johnson:
> The other thing in this PR we should discuss is feature-bit-based host CPU detection vs detecting specific microarchitectures and setting the right flags

I have felt all along that the former was the better way to go, but IIRC most other folks felt the other way was better -- given the existence of Zen4 as an in-betweener I'd think we should move to feature bits and redefine the existing microarch flags as combinations of those

abadams:
I think you might be referring to target flags being microarch vs features, but I was actually referring to how host detection is implemented. 

When we call the CPUID instruction on x86, should we be sniffing the bits that tell you what avx512 extensions exist, or should we be sniffing the bits that describe what the vendor and microarchitecture is? Either way should produce the same target string in theory, but they have different failure modes when things go wrong.

On the target flag issue, it's just not feasible to do individual feature bits with the way avx512 variants work. The Zen4 CPU would have the target string x86-64-linux-sse41-f16c-fma-avx-avx2-avx512-avx512f-avx512dq-avx512ifma-avx512cd-avx512bw-avx512vl-avx512vbmi-avx512vbmi2-avx512vnni-avx512vpclmulqdq-avx512bitalg-avx512vpopcntdq. Sapphire rapids would be even longer, because it adds avx512fp16, avxvnni, and a few more I think. It's much more usable to identify sets and subsets and have a single flag that means "all the avx512 extensions present on zen4".

zvookin:
My preferred design is to do things via individual feature bits but provide a way to map short names to a set of feature bits. The canonical form is fully expanded and the shortened names should be used only at a user facing command/UI level. Anything used for equality comparison or tagging in a build system etc. must use the expanded form.

Generally anything in a build file or source code wants to use an expanded form for clarity and maintainability. (E.g. the thing here where Zen4 has features in between two other microarchitectures. This is really easy if all the feature bits the compiler knows about are surfaced.)

The expansion could be a helper function or a Python script. It invariably leads to wanting a more powerful language like "The foo architecture minus the bar feature." That's fine in an external tool but shouldn't go into the target specification of the main compiler. I want both usability and uncomplicated specification at the compiler interface. Having an explicit conversion that is only used for user facing stuff is probably the best way to do this.

abadams:
I will gently point out though that discussion of redesigning what the avx512 feature flags look like in the Target struct is largely independent of this PR, and I don't want it to derail the issues I actually wanted to discuss here. The changes in this PR don't pose any challenge to the status quo. #7839 is more related, but that's really about what we should do when some target flags imply other target flags, and is also not about refactoring how we've done the avx512 target flags.

zvookin:
Thinking about it more, the PR is likely the right approach for now.

How the feature flags are done directly informs the question of how to detect the information. If the processor architecture is in the feature name, it makes sense to derive it from the information identifying the processor. If the feature just names a specific ISA component, it should probably be detected via feature detection. The tradeoff is that newer processors can refuse to run code that they do support running.

The area is surprisingly complicated and thinking about it more, I'm leaning toward Target supporting both chip names and features with the compiler and runtime working in terms of an exhaustive list of features. So this PR likely doesn't make things much more difficult as backward compatibility is easy to provide if there is an explicit conversion anyway. 

(The alternative here would involve calling ```Target::AVX512_Cannonlake``` a baseline and adding feature specific flags for additions since then. ```Target:::AVX512_SapphireRapids``` would set all the new feature flags and there either wouldn't be a Zen4 specific feature or it would also set the ones it supports. It isn't strictly necessary to have bits for features that are irrelevant to Halide at the moment. They could be added later. All in all, probably not a better bet.)

abadams:
Failures unrelated.

PR #7842, opened 2023-09-08
Update arguments in driver.cpp to match what correctness/simd_op_check has

vksnk:
Float16 and bfloat16 buffers were missing.


steven-johnson (approved):
Nice catch!

PR #7852, opened 2023-09-18
Build rules for Xtensa tests with cstub library

vksnk:
This PR adds some testing for Xtensa target and requires a cstub library which can be obtained from the vendor.

Only tests/generator are supported, because Xtensa codegen is based on C++ codegen and doesn't support JIT execution. There are a few tests which fail and excluded right now, so need to be fixed later, but I don't think any of them are really critical.

To run the tests:
```
make test_aotcpp_xtensa_generator
```
In addition, apps/simd_op_check will compile and verify correctness of operations listed in test/correctness/simd_op_check_xtensa (which in turn will verify that specific intrinsic was generated for corresponding pattern).

To run the app:
```
make clean && make xtensa -j 12 && ./bin/driver-xtensa
```

Finally, this is only added for Make build, I will repeat the exercise for CMake in a follow-up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1286,6 +1286,7 @@ test_correctness_multi_gpu: correctness_gpu_multi_device
 # 3) Externally-written JIT-based tests
 GENERATOR_AOT_TESTS = $(GENERATOR_EXTERNAL_TESTS:$(ROOT_DIR)/test/generator/%_aottest.cpp=generator_aot_%)
 GENERATOR_AOTCPP_TESTS = $(GENERATOR_EXTERNAL_TESTS:$(ROOT_DIR)/test/generator/%_aottest.cpp=generator_aotcpp_%)
+GENERATOR_AOTXTENSA_TESTS = $(GENERATOR_EXTERNAL_TESTS:$(ROOT_DIR)/test/generator/%_aottest.cpp=generator_aotcpp_xtensa_%)
 GENERATOR_JIT_TESTS = $(GENERATOR_EXTERNAL_TESTS:$(ROOT_DIR)/test/generator/%_jittest.cpp=generator_jit_%)
 
 # multitarget test doesn't make any sense for the CPP backend; just skip it.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1316,6 +1317,60 @@ GENERATOR_AOTCPP_TESTS := $(filter-out generator_aotcpp_gpu_multi_context_thread
...
+# Tests below work, but need to disable parallel in the schedule.
+# TODO(vksnk): figure out what's wrong with parallel in this case.
+GENERATOR_AOTXTENSA_TESTS := $(filter-out generator_aotcpp_xtensa_example,$(GENERATOR_AOTXTENSA_TESTS))
+GENERATOR_AOTXTENSA_TESTS := $(filter-out generator_aotcpp_xtensa_mandelbrot,$(GENERATOR_AOTXTENSA_TESTS))
+GENERATOR_AOTXTENSA_TESTS := $(filter-out generator_aotcpp_xtensa_pyramid,$(GENERATOR_AOTXTENSA_TESTS))
+
+# Xtensa doesn't have float64 vectors
+GENERATOR_AOTXTENSA_TESTS := $(filter-out generator_aotcpp_xtensa_templated,$(GENERATOR_AOTXTENSA_TESTS))
+
+# Needs define_extent
```
steven-johnson, line 1338: For at least some of these, we should probably create Issue(s) and tag them with comment, so there is some tracking of the eventual fixes

test/correctness/simd_op_check_xtensa.cpp
```diff
@@ -111,13 +111,15 @@ class SimdOpCheckXtensa : public SimdOpCheckTest {
         // some of these could overflow that limit. (Omitting the spaces is
         // a bit of a band-aid here; a better solution would probably be
         // to allow arbitrary names that don't match, but for now, this will do.)
-        check("convert<float16x32_t,float32x32_t>", vector_width / 2, f16(f32_1));
-        check("convert<float32x32_t,float16x32_t>", vector_width / 2, f32(f16_1));
+        // TODO(vksnk): float16 doesnt't seem to be supported well by cstubs library.
```
steven-johnson, line 114: probably deserves an Issue

steven-johnson (approved):

Makefile
vksnk: Done.

test/correctness/simd_op_check_xtensa.cpp
vksnk: Done

Issue #7847, opened 2023-09-14
Calling realize() within an AOT Generator is problematic


steven-johnson:
A user inside Google recently had some confusion about unexpected results from a Generator they wrote; the issue turned out to be that some Halide-helper-library code that they were indirectly referencing called `realize()` to produce a LUT, and while this call referenced an `Input<int>` to the Generator, the LUT was (apparently) quietly produced with the value of zero. 

This was an easy workaround (ie, use `compute_root()` instead of `realize()`), but there are at least two issues here worth investigating:

- Should it be an error to call `.realize()` (or any similar JIT-invocation) during AOT Generator evaluation? While I can envision not-reasonable use cases (such as LUT creation, as in this case), they all seem like corner cases that have better solutions in the first plce (such as using compute_root()).
- Should it be possible to JIT-compile with a Parameter that doesn't have an explicit value set? (For `ImageParam` I suspect we don't allow this, but perhaps `Param<>` is just quietly defaulting to a value of zero?) I'd argue that the answer *should* be "no" in all cases.

zvookin:
Calling ```realize``` within an AOT generator is a standard technique. I used to cover it in presentations I did on how to use Halide. Moreover, in as much as Halide provides JIT as an API, not being able to JIT in this context would signal something broken elsewhere in Halide.

Does the threadsafe Callable mechanism make things better by requiring all the arguments get passed in at the call site?

abadams:
I also think we want to allow realize within AOT generators.

It might be that we never track if a scalar param has been set, and are just defaulting to zero. We should probably add a flag for that so that we can throw an error, like we do for buffers.

steven-johnson:
Fixing this for `Param<>` in the JIT case is indeed pretty trivial (see https://github.com/halide/Halide/pull/7853); as you suggested, adding a flag to track it works fine, because when 

However, this does *not* fix it for the Generator/AOT case (which is how we discovered it in the first place); the failure mode we saw was basically something like this:

```
class ShouldFail : public Halide::Generator<ShouldFail> {
public:
    Input<Buffer<uint8_t, 1>> input{"input"};
    Input<float> scalar_input{"scalar_input"};
    Output<Buffer<uint8_t, 1>> output{"output"};

    void generate() {
        Func lut_fn("lut_fn");
        lut_fn(x) = u8_sat(x * scalar_input / 255.f);

        // This should always fail, because it depends on a scalar input
        // that *cannot* have a valid value at this point.
[...]

PR #7853, opened 2023-09-18
Prevent use of uninitialized scalar Parameters in JIT code (#7847, partial)

steven-johnson:


test/correctness/func_clone.cpp
```diff
@@ -159,14 +159,6 @@ int update_defined_after_clone_test() {
         return 1;
     }
 
-    Buffer<int> im = g.realize({200, 200});
```
abadams, line 162: Prefix this block of code with param.set(false) instead of deleting it.

steven-johnson: Done

src/Parameter.cpp
```diff
@@ -20,6 +20,7 @@ struct ParameterContents {
     std::vector<BufferConstraint> buffer_constraints;
     Expr scalar_default, scalar_min, scalar_max, scalar_estimate;
     const bool is_buffer;
+    bool data_ever_set = false;
```
TH3CHARLie, line 23: this is the place we realize whenever we add sth to the frontend IR, serialization needs to be updated at the same time

steven-johnson: I don't understand

TH3CHARLie: `data_ever_set` looks like a field that needs to be consistent across serialization (i.e. A parameter that has a data set should also be known as data-ever-set after serialization roundtrip). In this case, we need to add this field to the serialization implementation to properly serialize it.

steven-johnson: Hmm... didn't think about that. 

I'm actually kinda unsure as to whether we should be saving the scalar value of the Param (if any) to the Serialization or not -- I know we do now (and it's required for the serialization-round-trip-via-JIT hack), but *conceptually*, it seems like the wrong thing to do. 

A Parameter (either scalar or buffer) is conceptually just a placeholder with required-type information. Is it really desirable (in the general sense) to save whatever happens to be stuff in there?

TH3CHARLie: it does seem weird to me at first as well, had to admit it.

But now I treat this new field in the PR same/similar to the purpose of `defined`, then it makes more sense

steven-johnson:
PTAL -- this change ballooned quite a bit but I think this is all better now:
- Use `std::optional<halide_scalar_value_t>` instead of uint64-and-bool for the scalar data field.
- Mark the Flatbuffer field as `= null` so that the API for it uses `std::optional` as well. (Note that there is no binary format change here since all Flatbuffer fields are optional by default.)
- Make the recently-added-for-deserialization ctors for Parameter private, with friend access for Deserializer; also remove the redundant `is_buffer` arguments since each of them has only one legal value anyway (and delete the Python wrappers for them entirely)
- Tighten the implementation of `Parameter:scalar_expr()` and `Parameter::scalar()`

PTAL

src/Pipeline.cpp
```diff
@@ -808,6 +825,11 @@ void Pipeline::prepare_jit_call_arguments(RealizationArg &outputs, const Target
                                           bool is_bounds_inference, JITCallArgs &args_result) {
     user_assert(defined()) << "Can't realize an undefined Pipeline\n";
 
+    FindVariablesWithUnsetParams find_unset;
```
abadams, line 828: I think this code is a hot path, so we can't do IR traversal here.

src/Pipeline.cpp
```diff
@@ -70,6 +70,23 @@ std::string sanitize_function_name(const std::string &s) {
     return name;
 }
 
+class FindVariablesWithUnsetParams : public IRVisitor {
+    using IRVisitor::visit;
+
+    void visit(const Variable *op) override {
+        if (op->param.defined() && !op->param.is_buffer() && op->param.name() != "__user_context") {
+            user_assert(op->param.has_scalar_value())
```
abadams, line 78: I don't think this error message is right. Won't it trigger on all attempts to realize a pipeline with unset params, regardless of whether or not you're in a generator? I think the much more likely scenario is someone using the JIT forgetting to set the value of a Param<T>. Also, you can absolutely compile_to_callable without setting the scalar params.

src/Pipeline.cpp
steven-johnson: Blarg. That's no good. I'm open for better suggestions.

src/Pipeline.cpp
steven-johnson: > Won't it trigger on all attempts to realize a pipeline with unset params, regardless of whether or not you're in a generator?

Yes. That's a feature, not a bug.

> I don't think this error message is right.

Yeah, this is a bit confusing -- IIRC the reason for wording it this way is that the "other" cases (realize, callable) end up being caught via `read_only_scalar_address()` noting the invalid value and failing (apparently they don't add the interposed Variable, I don't recall why, it's been a few days).

steven-johnson:
OK, so upon further review:

`FindVariablesWithUnsetParams` was added to try to address the case of using `compile_to_callable()` inside a Generator, but I think it's basically impossible to (mis)use `compile_to_callable()` in this way inside a Generator; you'd have to construct your own `Argument` directly to use in `compile_to_callable()` (since Generator Input fields don't expose any conversion to Argument or Param<T> or Parameter)... and even then, there isn't a way to invoke the Callable with the Input, because the Callable requires a `float` argument (an `Expr` won't do). So I don't think we even need that traverser (or the uninitialized_param_3 test) at all.

abadams (approved):

PR #7865, opened 2023-09-25
Respect input buffer constraints in root-level bounds inference exprs

### LINKED ISSUE #7761: Missing potential optimizations when input/output buffers are constant extents
If you explicitly constrain a Generators input buffer extents to specific values (eg we know that `input_k` is always 128x16, so we do `input_k_.dim(0).set_extent(128).dim(1).set_extent(16);` 

- we *do* inject the correct assertions to verify that, but... 

- we don't seem to propagate those constants elsewhere, e.g. when calculating extents of intermediate Func that are based on that, e.g. `let O_block.s1.y.max = let t283 = min(input_k.extent.1, 16) in max(-1 - t283, max(t283, 1) + -1)` [hey, we know that input_k.extent.1 is a constant!]
 
- we still get (redundant) assertion  checks that `input_k.min.1` is in range (in addition to the must-be-exact-value check),

abadams:
Fixes #7761

### DIFF
--- a/test/correctness/respect_input_constraint_in_bounds_inference.cpp
+++ b/test/correctness/respect_input_constraint_in_bounds_inference.cpp
@@ -1,0 +1,23 @@
+#include "Halide.h"
+
+using namespace Halide;
+
+int main(int argc, char **argv) {
+    ImageParam im(Float(32), 1);
+    Func f, g;
+    Var x;
+
+    f(x) = x;
+    g(x) = f(x % im.dim(0).extent());
+    im.dim(0).set_extent(16);
+
+    // Given the constraint, we know the bounds of f should be less than 16, so
+    // the compiler should be happy placing it in a register. This is just a way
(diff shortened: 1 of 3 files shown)


test/correctness/respect_input_constraint_in_bounds_inference.cpp
```diff
@@ -0,0 +1,23 @@
...
+int main(int argc, char **argv) {
+    ImageParam im(Float(32), 1);
+    Func f, g;
+    Var x;
+
+    f(x) = x;
+    g(x) = f(x % im.dim(0).extent());
+    im.dim(0).set_extent(16);
+
+    // Given the constraint, we know the bounds of f should be less than 16, so
```
steven-johnson, line 14: Suggestion: add a variant where bounds of f is >= (say) 1024 and make it an error test?

steven-johnson (approved):

PR #7866, opened 2023-09-26
Handle unreachable code in bounds inference

### LINKED ISSUE #7756: bounds-query can quietly return wrong answer when specialize_fail() is used
The scenario: we have code that does a bilinear resize, with specialize() for various channel counts, planar/chunky options, etc.

While tweaking the schedule, `specialize_fail()` was added at the end, to avoid codegen for unsupported specialize options; after this was done, bounds-query mode starting failing when this function was used for `define_extern()`: it doesn't return an error code, but it just quietly gives bogus results (generally, all-zero bounds). I'm assuming this is because the `define_extern()` case passes in bounds that have `stride=0` for all dimensions (which is an unsupported case after this change).

If I remove the specialize_fail(), or call bounds-query with supported stride values, things work correctly.

abadams:
See https://github.com/halide/Halide/issues/7756#issuecomment-1736001570

Fixes #7756

### DIFF
--- a/src/Bounds.cpp
+++ b/src/Bounds.cpp
@@ -2190,7 +2190,7 @@ private:
     // Map variable name to all other vars which values depend on that variable.
     map<VarInstance, set<VarInstance>> children;
 
-    bool in_producer{false};
+    bool in_producer{false}, in_unreachable{false};
     map<std::string, Expr> buffer_lets;
 
     using IRGraphVisitor::visit;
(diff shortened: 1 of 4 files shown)


steven-johnson (approved):

abadams:
Failure unrelated (that test is getting awfully close to failing on main, I'll investigate separately)

